PR #3271, opened 2023-01-11
feat(docs): add remote cache counter

tknickman:
Adds a remote cache time saved ticket to Turborepo docs

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -18,6 +18,7 @@
     "@heroicons/react": "1.0.6",
     "@mdx-js/react": "^2.1.4",
     "@react-aria/ssr": "3.3.0",
+    "@react-spring/web": "^9.6.1",
     "@sentry/nextjs": "^7.17.3",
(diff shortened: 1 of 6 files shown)


docs/package.json
```diff
@@ -18,6 +18,7 @@
     "@heroicons/react": "1.0.6",
     "@mdx-js/react": "^2.1.4",
     "@react-aria/ssr": "3.3.0",
+    "@react-spring/web": "^9.6.1",
```
mehulkar, line 21: What's the cost of this extra dep? Does it automatically get treeshaken out on other pages?

mehulkar (approved):

docs/package.json
tknickman: Just ran a build analysis before and after, and the difference is hardly noticeable, around 0.1kB additional.

ForsakenHarmony:
I feel like hours would be more meaningful

19 million minutes is a very big number

PR #3275, opened 2023-01-12
Make pre-commit hoook executable

mehulkar:
Seeing this on git commit:

```
hint: The '.husky/pre-commit' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```

File permissions were changed in daaeabfb502d3b1f6c1acfae31d0d2cf8a740463, might have been accidental @ForsakenHarmony ?

PR #3276, opened 2023-01-12
Mark .husky/pre-commit as executable

wbinnssmith:
Prior to this change, committing would result in this message from git:

```
hint: The '.husky/pre-commit' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```


wbinnssmith:
Closing in favor of #3275

PR #3277, opened 2023-01-12
nextest: time out leaky check after 500ms

wbinnssmith:
nexttest considers tests leaky if they do not close stdout and stderr within 100ms by default. I've noticed that the tailwind basic test is the only test considered flakey on macOS. Increasing this timeout to 500ms locally makes it no longer marked as leaky.

It's still possible the integration tests leak subprocesses on Windows, but this should clear up confusion on other environments.

Test Plan: `cargo nextest run -p next-dev test_crates_next_dev_tests_integration_next_tailwind_basic` without it being marked leaky on macOS.


mehulkar (approved):
Fine with me, but if this is turbopack related, maybe added this new dir to codeowners also?

ForsakenHarmony (approved):
LGTM, adding the file to `CODEOWNERS` would be good as mentioned above

PR #3281, opened 2023-01-12
fix windows paths with --root

sokra:


kwonoj (approved):

crates/next-dev/src/lib.rs
```diff
@@ -265,8 +265,9 @@ async fn source(
     let project_relative = project_dir.strip_prefix(&root_dir).unwrap();
     let project_relative = project_relative
         .strip_prefix(MAIN_SEPARATOR)
-        .unwrap_or(project_relative);
-    let project_path = fs.root().join(project_relative);
+        .unwrap_or(project_relative)
+        .replace(MAIN_SEPARATOR, "/");
+    let project_path = fs.root().join(&project_relative);
```
jridgewell, line 270: This should have been detected by the `join` call, but we run our tests strip out `debug_assert`s.

jridgewell: https://github.com/vercel/turbo/blob/368ca0ba914080a4932859374a3b6043b07fb0ab/.github/workflows/test.yml#L482

PR #3282, opened 2023-01-12
use the first result when resolving extends in tsconfig.json

sokra:


crates/turbopack-ecmascript/src/typescript/resolve.rs
```diff
@@ -72,8 +72,9 @@ pub async fn read_tsconfigs(
                         RequestVc::parse(Value::new(extends.to_string().into())),
                         resolve_options,
                     )
+                    .primary_assets()
                     .await?;
-                    if let ResolveResult::Single(asset, _) = *result {
+                    if let Some(&asset) = result.iter().next() {
```
alexkirsz, line 77: Add a comment explaining why we select the first result specifically.

alexkirsz (approved):

PR #3284, opened 2023-01-12
add basic webpack loader support

sokra:
add the minimum to run a simple raw-loader (see test case)

* add `experimental.turbopackWebpackLoaders` to next.config.js
  * key is extension like `".mdx"`, value is an array of loaders like `["mdx-loader"]`

### DIFF
--- a/crates/turbopack-core/src/server_fs.rs
+++ b/crates/turbopack-core/src/server_fs.rs
@@ -6,36 +6,36 @@ use turbo_tasks_fs::{
 };
 
 #[turbo_tasks::value]
-pub struct DevServerFileSystem {}
+pub struct ServerFileSystem {}
 
 #[turbo_tasks::value_impl]
-impl DevServerFileSystemVc {
+impl ServerFileSystemVc {
     #[turbo_tasks::function]
     pub fn new() -> Self {
-        Self::cell(DevServerFileSystem {})
+        Self::cell(ServerFileSystem {})
     }
 }
 
 #[turbo_tasks::value_impl]
-impl FileSystem for DevServerFileSystem {
+impl FileSystem for ServerFileSystem {
     #[turbo_tasks::function]
     fn read(&self, _fs_path: FileSystemPathVc) -> Result<FileContentVc> {
-        bail!("Reading is not possible from the virtual filesystem for the dev server")
+        bail!("Reading is not possible from the marker filesystem for the server")
     }
 
     #[turbo_tasks::function]
     fn read_link(&self, _fs_path: FileSystemPathVc) -> Result<LinkContentVc> {
-        bail!("Reading is not possible from the virtual filesystem for the dev server")
+        bail!("Reading is not possible from the marker filesystem for the server")
     }
 
     #[turbo_tasks::function]
     fn read_dir(&self, _fs_path: FileSystemPathVc) -> Result<DirectoryContentVc> {
-        bail!("Reading is not possible from the virtual filesystem for the dev server")
+        bail!("Reading is not possible from the marker filesystem for the server")
     }
 
     #[turbo_tasks::function]
     fn write(&self, _fs_path: FileSystemPathVc, _content: FileContentVc) -> Result<CompletionVc> {
-        bail!("Writing is not possible to the virtual filesystem for the dev server")
+        bail!("Writing is not possible to the marker filesystem for the server")
     }
 
     #[turbo_tasks::function]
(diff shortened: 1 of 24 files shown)


crates/turbopack/src/module_options/module_options_context.rs
```diff
@@ -22,6 +44,7 @@ pub struct ModuleOptionsContext {
     pub enable_styled_components: bool,
     pub enable_styled_jsx: bool,
     pub enable_postcss_transform: Option<PostCssTransformOptions>,
+    pub enable_webpack_loaders: Option<WebpackLoadersOptions>,
```
alexkirsz, line 47: It seems weird that this is worded as a boolean (`enable_something`) instead of just `webpack_loaders`, like we have for other options. Same comment goes for PostCSS.

crates/turbopack-node/js/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "@vercel/turbopack-node",
+  "version": "0.0.0",
+  "description": "turbopack node runtime",
+  "license": "UNLICENSED",
+  "private": true,
+  "scripts": {
+    "check": "tsc --noEmit"
+  },
+  "dependencies": {
+    "loader-runner": "^4.3.0"
```
alexkirsz, line 11: How do we ensure this is installed by the user? Is it one of Next.js' transitive dependencies? Or should we compile it and distribute it with turbopack?

alexkirsz (approved):
A few nits and questions, but otherwise :ship-it:

crates/turbopack-core/src/server_fs.rs
```diff
@@ -6,36 +6,36 @@ use turbo_tasks_fs::{
...
     }
 }
 
 #[turbo_tasks::value_impl]
-impl FileSystem for DevServerFileSystem {
+impl FileSystem for ServerFileSystem {
     #[turbo_tasks::function]
     fn read(&self, _fs_path: FileSystemPathVc) -> Result<FileContentVc> {
-        bail!("Reading is not possible from the virtual filesystem for the dev server")
+        bail!("Reading is not possible from the marker filesystem for the server")
```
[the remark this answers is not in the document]
sokra: We only used it for paths and do not read/write from it. A virtual filesystem would allow to read/write...

Brooooooklyn:
It looks like it would be easier to implement this with napi-rs, @sokra do you mind if I make a prototype based on this with napi-rs?

crates/turbopack/src/module_options/module_options_context.rs
sokra: yeah. `pub enable_styled_jsx: bool,` could also be `pub styled_jsx: bool,`

crates/turbopack-node/js/package.json
sokra: Currently the user need to install it to use this feature. That's fine for the first iteration...

sokra:
> It looks like it would be easier to implement this with napi-rs, @sokra do you mind if I make a prototype based on this with napi-rs?

A few questions?

* Can we use napi-rs even if you run the standalone binary `next-dev.exe` without being called from next.js?`
* Can we create a separate Node.js Isolate with napi-rs that can be restarted separately?
* Can we provide the JS source files from in-memory without them being written to disk at all?

Brooooooklyn:
> Can we use napi-rs even if you run the standalone binary next-dev.exe without being called from next.js?`

Why do we need to distribute turbopack as a binary? IMO the entry file can be a JavaScript file that calls the napi-rs wrapped Rust function like next-swc.

> Can we create a separate Node.js Isolate with napi-rs that can be restarted separately?

We can use `worker_threads` API to do that, and I think if we can call JavaScript function directly in Rust, we can avoid restarting Node.js process scenario in most cases; just re-call the JavaScript function is fine.

> Can we provide the JS source files from in-memory without them being written to disk at all?

Yes, sure; we can keep all the ability in turbopack with napi-rs, including the memory fs, resolve alias, and even dev-server. 
Building turbopack with napi-rs just changes the way to distribute turbopack, we will not lose any features and abilities in turbopack.

wbinnssmith (approved):

crates/turbopack-node/src/transforms/webpack.rs
```diff
@@ -0,0 +1,175 @@
...
+struct ProcessWebpackLoadersResult {
+    content: AssetContentVc,
+    assets: Vec<VirtualAssetVc>,
+}
+
+#[turbo_tasks::function]
+fn webpack_loaders_executor(project_root: FileSystemPathVc, context: AssetContextVc) -> AssetVc {
+    EcmascriptModuleAssetVc::new(
+        VirtualAssetVc::new(
+            project_root.join("__turbopack__/webpack-loaders-executor.ts"),
```
ForsakenHarmony, line 106: Shouldn't that be `[embedded_modules]/@vercel/turbopack-node`

PR #3288, opened 2023-01-12
Introduce next-dev-tests to rebuild automatically on test input changes

kdy1:
I verified that the crate is rebuilt if I make a change of inputs.


crates/next-dev-tests/build.rs
```diff
@@ -0,0 +1,9 @@
+use turbo_tasks_build::{generate_register, rerun_if_glob};
+
+fn main() {
+    generate_register();
+    // The test/integration crate need to be rebuilt if any snapshots are added.
+    // Unfortunately, we can't have the build.rs file operate differently on
+    // each file, so the entire next-dev crate needs to be rebuilt.
```
jridgewell, line 7: Nit: this comment needs to be updated.

crates/next-dev-tests/Cargo.toml
```diff
@@ -0,0 +1,57 @@
+[package]
+name = "next-dev-tests"
```
jridgewell, line 2: Can we remove any deps from `next-dev/Cargo.toml`?

jridgewell (approved):

crates/next-dev-tests/Cargo.toml
kdy1: Some of them are used by benchmarks, so I removed only unused ones

sokra (approved):

PR #3193, opened 2023-01-06
Add support for dynamic `import()`s on the server-side

alexkirsz:
This PR separates Turbopack's `runtime.js` into two:

- the Browser/DOM implementation, which uses `<script>` and `<link>` tags for dynamically loading chunks;
- the Node.js/CommonJS implementation, which uses `require` instead.

~~The two runtimes share a lot of code, but I'm not sure what's the best approach for sharing this logic, or even if we should, so for now they are two separate entities.~~

The shared code between the two runtimes is now in a separate file to avoid duplication.

PR #3297, opened 2023-01-12
Fix deduplication when importing module statically and dynamically

jridgewell:
The `processed_assets` `IndexSet` was performing deduplication on the `AssetVc`, which is the same between static (`import ... from "foo"`) and dynamic (`import("foo")`) imports of the same specifier. But, the _reference_ to those assets is different, and generates different chunking semantics. In order for both to succeed, we need to process both assets.

Coupled with #3193's ability to load chunks on the server side, this fixes fixes WEB-381.


crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -407,7 +409,7 @@ async fn chunk_content_internal<I: FromChunkableAsset>(
                 for asset in assets
                     .await?
                     .iter()
-                    .filter(|asset| processed_assets.insert(**asset))
+                    .filter(|asset| processed_assets.insert((chunking_type, **asset)))
```
sokra, line 412: This causes `Placed` and `PlacedOrParallel` to generate two chunk items.

Can we add a `processed_assets.insert((Placed, **asset))` into the placed code path in `PlacedOrParallel` and a `processed_assets.insert((Parallel, **asset))` into the parallel code path.

sokra (approved):

PR #3298, opened 2023-01-12
Implement `experimental.resolveAlias` map for next.config.js

wbinnssmith:
Fixes WEB-416.

This implements loading an alias map set as `experimental.resolveAlias` in `next.config.js`.

Test Plan: Added an integration test.

### DIFF
--- a/crates/next-core/src/next_import_map.rs
+++ b/crates/next-core/src/next_import_map.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use turbo_tasks::Value;
 use turbo_tasks_fs::{glob::GlobVc, FileSystemPathVc};
+use turbopack::module_options::ResolveAliasOptionsVc;
 use turbopack_core::resolve::{
     options::{ImportMap, ImportMapVc, ImportMapping, ImportMappingVc, ResolvedMap, ResolvedMapVc},
     AliasPattern,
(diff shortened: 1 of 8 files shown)


crates/next-core/src/next_import_map.rs
```diff
@@ -16,13 +18,18 @@ use crate::{
...
-) -> ImportMapVc {
+    next_config: NextConfigVc,
+) -> Result<ImportMapVc> {
     let mut import_map = ImportMap::empty();
 
-    insert_next_shared_aliases(&mut import_map, project_path);
+    insert_next_shared_aliases(
+        &mut import_map,
+        project_path,
+        &resolve_alias_options_to_alias_map(&*next_config.resolve_alias_options().await?).await?,
```
wbinnssmith, line 31: This is a little awkward: `insert_next_shared_aliases` is a non-turbo, non-async function. Any suggestions that might clean this up?

jridgewell: You should be able to call a non-turbo function with Vcs, it just needs to be async. The dependency tracking will be assigned to the closest turbo function in the call stack (this function in this case).

crates/next-core/src/next_import_map.rs
```diff
@@ -307,3 +335,17 @@ fn request_to_import_mapping(context_path: FileSystemPathVc, request: &str) -> I
 fn external_request_to_import_mapping(request: &str) -> ImportMappingVc {
     ImportMapping::External(Some(request.to_string())).into()
 }
+
+pub async fn resolve_alias_options_to_alias_map(
+    resolve_alias_options: &ResolveAliasOptions,
+) -> Result<IndexMap<String, Vec<String>>> {
+    let mut alias_map = IndexMap::new();
+    for (key, aliases_vc) in &resolve_alias_options.alias_map {
```
jridgewell, line 343: Can't we just clone the full thing and return it?

jridgewell (approved):

sokra (approved):

Issue #3272, opened 2023-01-11
[turborepo] replays task output even with "outputMode": "new-only"


AlexAegis:
### What version of Turborepo are you using?

1.7.0

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Linux

### Describe the Bug

"outputMode": "new-only" is not used and I get my cached output replayed regardless.

This was working in `1.6.3` and broke in `1.7.0`

I tried two repositories and both behaved this way since the update

### Expected Behavior

To only output the hash on cache hit

### To Reproduce

Set a task to only output on cache misses

```json
{
	"$schema": "https://turborepo.org/schema.json",
	"pipeline": {
		"test": {
			"env": ["VITEST"],
			"outputs": ["coverage"],
			"outputMode": "new-only"
		}
	}
}
```

Launch the task twice, on the second run (provided the first was successful) it should only output hashes.

Example:

```sh
git clone https://github.com/AlexAegis/advent-of-code.git
cd advent-of-code
git checkout d55c481
pnpm lint:format
pnpm lint:format | grep output
```

On the last, grepped command you should see supressed output messages and not `lint:format: cache hit, replaying output`

(That commit is the head right now with succeeding lint and test tasks, in case I break it.)

### Reproduction Repo

https://github.com/AlexAegis/advent-of-code

PR #3299, opened 2023-01-12
Do not submit a default --output-logs argument from Rust 

mehulkar:
Specifying the CLI arg overrides outputMode from Task Definitions,
so we want to ensure that the user explicitly passes it, and not provide a default.

Fixes https://github.com/vercel/turbo/issues/3272


chris-olszewski (approved):
LGTM

mehulkar:
installing sponge is failing on some of the args

Issue #3257, opened 2023-01-11
[turborepo] Cannot find module 'turbo' when using docker to build


aindong:
### What version of Turborepo are you using?

1.7.0

### What package manager are you using / does the bug impact?

Yarn v1

### What operating system are you using?

Mac

### Describe the Bug

I have been using turbo repo and building my apps using docker and was working perfectly with v1.6.3 until 1.7.0 released, the `turbo` cli can't be found after doing `yarn global add turbo` on docker. 

```
 => [installer 1/8] WORKDIR /app                                                                                                                0.2s
 => [runner 2/4] RUN addgroup --system --gid 1001 nodejs                                                                                        3.1s
 => [builder 2/4] RUN yarn global add turbo                                                                                                    30.7s
 => [runner 3/4] RUN adduser --system --uid 1001 ***                                                                                        2.0s
 => [installer 2/8] COPY .gitignore .gitignore                                                                                                  0.5s
 => [builder 3/4] COPY . .                                                                                                                      7.3s
 => ERROR [builder 4/4] RUN turbo prune --scope=*** --docker                                                                     1.6s
------
 > [builder 4/4] RUN turbo prune --scope=*** --docker:
#17 1.472 /app/node_modules/turbo/bin/turbo:13
#17 1.472     throw e;
#17 1.472     ^
#17 1.472
#17 1.472 Error: Cannot find module 'turbo'
#17 1.472 Require stack:
#17 1.472 - /app/node_modules/turbo/node-platform.js
#17 1.472 - /app/node_modules/turbo/bin/turbo
#17 1.472     at Function.Module._resolveFilename (node:internal/modules/cjs/loader:985:15)
#17 1.472     at Function.resolve (node:internal/modules/cjs/helpers:109:19)
#17 1.472     at downloadedBinPath (/app/node_modules/turbo/node-platform.js:44:44)
#17 1.472     at generateBinPath (/app/node_modules/turbo/node-platform.js:67:15)
#17 1.472     at Object.<anonymous> (/app/node_modules/turbo/bin/turbo:7:5)
#17 1.472     at Module._compile (node:internal/modules/cjs/loader:1155:14)
#17 1.472     at Object.Module._extensions..js (node:internal/modules/cjs/loader:1209:10)
#17 1.472     at Module.load (node:internal/modules/cjs/loader:1033:32)
#17 1.472     at Function.Module._load (node:internal/modules/cjs/loader:868:12)
[...]

PR #3301, opened 2023-01-13
fix(examples): ignore node_modules in docker

tknickman:
Fixes https://github.com/vercel/turbo/issues/3257


nathanhammond (approved):
This should work, but also I suspect that we should maybe change which side we do the prune on.

PR #3302, opened 2023-01-13
chore: Port logout to be pure Rust

chris-olszewski:
With this PR `turbo logout` can run without jumping to Go.

Notes:
 - Used `config-rs` for our config reading. It doesn't support writing configs to disk out of the box, but this can be solved with a wrapper struct. This can probably be abstracted to be generic over different configs, but we should hold off until that seems beneficial.
 - `termcolor` is used for coloring output, this is the same crate that `env_logger` uses for coloring outputs. The `UI` struct is expected to be passed into any subcommands that need pretty output.
 - A slight behavior change is that we no longer delete the entire config file and instead make the token null (`"token": null`). If this is unwanted, I can change it back to the old behavior, but I wanted to test out `UserConfig::set_token` before someone else tried using it.


crates/turborepo-lib/src/commands/logout.rs
```diff
@@ -0,0 +1,23 @@
...
+            .ok_or_else(|| anyhow::anyhow!("Unable to find default config path"))?,
+        None,
+    )?;
+
+    if let Err(err) = config.set_token(None) {
+        error!("could not logout. Something went wrong: {}", err);
+        return Err(err);
+    }
+
+    ui.info(">>> Logged out", Some(&GREY))?;
```
ForsakenHarmony, line 21: Would prefer using `log::info!` for this

I think we can delete the `UI` struct if we used another color library

crates/turborepo-lib/Cargo.toml
```diff
@@ -9,19 +9,25 @@ license = "MPL-2.0"
...
 dunce = "1.0"
 env_logger = "0.10.0"
+lazy_static = "1.4.0"
 log = "0.4.17"
 predicates = "2.1.1"
 semver = "1.0"
 serde = { version = "1.0.145", features = ["derive"] }
 serde_json = "1.0.86"
 serde_yaml = "0.8.26"
+termcolor = "1.1.3"
```
ForsakenHarmony, line 31: This is the third (to my knowledge) color library we use in this project, I think we should decide on a single one to use everywhere

https://crates.io/crates/console might be a good option
- has a function for inline styles https://docs.rs/console/0.15.4/console/fn.style.html
- you can enable/disable color globally https://docs.rs/console/0.15.4/console/fn.set_colors_enabled.html (for `FORCE_COLOR`)
  - also detects if colors are supported on its own

crates/turborepo-lib/src/config.rs
```diff
@@ -0,0 +1,132 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::Result;
+use config::{Config, Environment};
+// Go's xdg implementation uses FOLDERID_LocalAppData for config home
+// https://github.com/adrg/xdg/blob/master/paths_windows.go#L28
+// Rust xdg implementations uses FOLDERID_RoamingAppData for config home
```
ForsakenHarmony, line 7: Should probably migrate at some point (not now)

crates/turborepo-lib/src/config.rs
```diff
@@ -0,0 +1,132 @@
...
+// We use cache_dir so we can find the config dir that the Go code uses
+#[cfg(windows)]
+use dirs_next::cache_dir as config_dir;
```
ForsakenHarmony, line 10: Shouldn't this be `data_local_dir`?
```suggestion
use dirs_next::data_local_dir as config_dir;
```

ForsakenHarmony (changes requested):

ForsakenHarmony (changes requested):
LGTM apart from the logging

crates/turborepo-lib/src/config.rs
```diff
@@ -0,0 +1,132 @@
...
+#[cfg(not(windows))]
+use dirs_next::config_dir;
+use serde::{Deserialize, Serialize};
+
+pub fn default_user_config_path() -> Option<PathBuf> {
+    config_dir().map(|p| p.join("turborepo").join("config.json"))
```
ForsakenHarmony, line 16: If this `Option` being `None` is always an error, it might be better to return the error from here

```suggestion
pub fn default_user_config_path() -> Result<PathBuf> {
    config_dir().map(|p| p.join("turborepo").join("config.json")).context("default config path not found")
```

same goes for `data_dir` below

crates/turborepo-lib/src/commands/logout.rs
```diff
@@ -0,0 +1,23 @@
...
+
+use crate::{
+    config::{default_user_config_path, UserConfig},
+    ui::{GREY, UI},
+};
+
+pub fn logout(ui: UI) -> Result<()> {
+    let mut config = UserConfig::load(
+        &default_user_config_path()
+            .ok_or_else(|| anyhow::anyhow!("Unable to find default config path"))?,
```
ForsakenHarmony, line 12:
```suggestion
        &default_user_config_path()
            .context("Unable to find default config path")?,
```

something anyhow provides

crates/turborepo-lib/Cargo.toml
chris-olszewski: I chose `termcolor` since we're already using `env_logger`'s color functionality that uses it: https://github.com/rust-cli/env_logger/blob/main/src/fmt/writer/termcolor/extern_impl.rs

crates/turborepo-lib/src/config.rs
chris-olszewski: Both resolve to the same path on Windows, but `data_local_dir` is a better name.

crates/turborepo-lib/src/commands/logout.rs
chris-olszewski: Turborepo has logs which respect verbosity level and then UI print statements which get written regardless, for UI info/error/warn/output just map to default styles applied to the string.

crates/turborepo-lib/Cargo.toml
chris-olszewski: `console` looks good, but we would be adding a new color crate to our transitive deps. I'll look at switching out `colored` in favor of this.

crates/turborepo-lib/src/config.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Inner struct that matches the config file schema
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq, Default)]
+struct UserConfigInner {
+    token: Option<String>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UserConfig {
```
NicholasLYang, line 31: Could we add a comment explaining why we have a `disk_config` and `config`? I'm a little confused as to what's going on.

NicholasLYang (commented):
Just some clarifications and nits

crates/turborepo-lib/src/config.rs
```diff
@@ -0,0 +1,132 @@
...
+    disk_config: UserConfigInner,
+    config: UserConfigInner,
+    path: PathBuf,
+}
+
+impl UserConfig {
+    /// Loads the user config from the given path,
+    pub fn load(path: &Path, token: Option<&str>) -> Result<Self> {
```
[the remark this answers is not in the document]
NicholasLYang: In general I'm thinking we should try to have doc comments on public functions unless they're very very obvious.

crates/turborepo-lib/src/config.rs
```diff
@@ -0,0 +1,134 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::{Context, Result};
+use config::{Config, Environment};
+#[cfg(not(windows))]
+use dirs_next::config_dir;
```
arlyon, line 6: I used `directories` when writing the rust-go-rust sandwich proof of concept, will use `directories-next` (which depends on `dirs-next`) instead.

NicholasLYang (approved):

PR #3304, opened 2023-01-13
Disable mdx parsing in turbotrace

Brooooooklyn:
It's not necessary to enable `MDX` parsing in turbotrace, and there was a parsing error that caused the build to fail when I was trying to enable turbotrace in `verce/site`: https://vercel.com/vercel/vercel-site/7pFWUVMWVAv3jC44WD2nBnwrbJpM


crates/node-file-trace/src/lib.rs
```diff
@@ -259,6 +259,7 @@ async fn input_to_modules<'a>(
         env,
         ModuleOptionsContext {
             enable_types: true,
+            enable_mdx: false,
```
sokra, line 262: false is the default, so this is not really needed. But you need to set it to `true` for next-dev ModuleOptionContextes

crates/turbopack/tests/node-file-trace/integration/mdx/index.js
```diff
@@ -0,0 +1 @@
+import Example from "./example.mdx";
```
sokra, line 1: Doesn't this break in node.js?

Also this test case is not listed in the `node-file-trace.rs` test file which calls all test cases.

sokra (changes requested):

crates/turbopack/tests/node-file-trace/integration/mdx/index.js
Brooooooklyn: It's existed for testing directly via `cargo run --bin node-file-trace`. I didn't have any idea how to test it with node-file-trace. Do you have any suggestions? Should I remove these files?

ForsakenHarmony: It should be added here
https://github.com/vercel/turbo/blob/7baa4d41f3efb5eb13ed48eec2ea8911e0759ab2/crates/turbopack/tests/node-file-trace.rs#L46-L240

ForsakenHarmony:
Why do we not want to parse in turbotrace? Don't we still want to resolve the imports of the mdx files?

crates/turbopack/tests/node-file-trace/integration/mdx/index.js
Brooooooklyn: @ForsakenHarmony I know it could be added here, but it could not be executed by Node.js directly, so the test case would always fail

ForsakenHarmony: That's not a good idea imo, maybe we should have snapshot tests for mdx

Brooooooklyn:
> Why do we not want to parse in turbotrace? Don't we still want to resolve the imports of the mdx files?

Align the behavior with https://github.com/vercel/nft. 

Turbotrace will run two phases in next.js projects:

1. tracing webpack entries
2. tracing webpack output

tracing the webpack entries is because we need to trace static assets like CSS/images that could not be traced in output. So, ignoring MDX is totally safe in this step.

@ForsakenHarmony

ForsakenHarmony:
@Brooooooklyn what happens when you import css from mdx?

sokra:
You could use https://mdxjs.com/packages/node-loader/ from node to import mdx. In which case we also need to trace dependencies in mdx, so mdx need to be parsed.

Brooooooklyn:
> what happens when you import css or an image from mdx?

In this case, they can use `unstable_includeFiles` to include it: https://nextjs.org/docs/advanced-features/output-file-tracing#caveats

Brooooooklyn:
@sokra added the `enable_mdx` option to node-file-trace, I will disable it in `vercel-site` for now because of the `mdx-rs` bug, and enable it again when `mdx-rs` fix it.

sokra (approved):

PR #3306, opened 2023-01-13
add first version of rgr sandwich

arlyon:
Hi!

This is a proof-of-concept for the rust-go-rust sandwich. It successfully ports the `GetTurboDataDir` function onto the rust side, and exposes it via a single statically-linked library `turborepo_ffi` along with bindings for it. It is expected that anything we need via FFI and the accompanying glue code can live here, so we can put the needed functionality into the right places in rust-land and just re-expose a simple API here.

There are still a number of things that we need to do to finish this work:

- [x] set up gitignore and decide how to integrate the various parts into the toolchain. we use `swig` to generate the go bindings which needs to be re-run when changing any function signatures, so we'll have to decide if we want that to be a required part of the toolchain or optional. things that need to be ignored: libturborepo_ffi.a, bindings.h (?), messages.pb.go (?), swig_wrap.cxx (?), turborepo_ffi.go (?)
- [ ] ~~update the build.rs file for turborepo to statically build and move `libturborepo_ffi` and associated artifacts into the go module~~ (handled in the makefile)
- [x] set up proto generation properly for the go side
- [x] clean up go.mod (some bits leaked in from testing)
- [x] figure out a strategy for opting in by setting the go build `-tag` (perhaps enable it by default for canary builds...?)

That said, it's working on both ends using a nice generic interface.

https://github.com/vercel/turbo/blob/66f3d527f57d09594b1275964bca90e58162cb92/crates/turborepo-ffi/src/lib.rs#L5-L32

https://github.com/vercel/turbo/blob/66f3d527f57d09594b1275964bca90e58162cb92/cli/internal/ffi/util.go#L10-L26

Usage looks something like the following:

https://github.com/vercel/turbo/blob/66f3d527f57d09594b1275964bca90e58162cb92/crates/turborepo-ffi/src/lib.rs#L41-L50

https://github.com/vercel/turbo/blob/66f3d527f57d09594b1275964bca90e58162cb92/cli/internal/fs/get_turbo_data_dir_rust.go#L11-L18

### Implementation Details

- we use go build tags to define which code path to use. When defining the 'go' tag, then the native go implementation is used. When defining the 'rust' tag, then we use ffi. This can be controlled in make via the cli using `GO_TAG=rust`
- the Buffer type is a simple rust struct that contains a lenth and a pointer to some data. we can wrap that in some simple code to serialize and deserialize a particular protobuf message, handling cleanup etc.

### How to try it out:

```sh
make turbo GO_TAG=rust
```

The rust implementation of GetTurboDataDir will print out the path whenever it is called, to verifiy it is being used.


crates/turborepo-ffi/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    len: u32,
+    data: *mut u8,
+}
+
+impl<T: prost::Message> From<T> for Buffer {
+    fn from(value: T) -> Self {
+        let mut bytes = value.encode_to_vec();
+        let data = bytes.as_mut_ptr();
+        let len = bytes.len() as u32;
+        std::mem::forget(bytes);
```
chris-olszewski, line 17: This feels off to me. If I'm understanding `mem::forget` this means `Buffer::from(message)` assumes the constructed `data` will be passed to another language that will free the data once it's used. Would it be safer to forget the right before we pass it back?

Maybe [`ManuallyDrop`](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html) would be good here to signal this expectation?

crates/turborepo-ffi/build.rs
```diff
@@ -0,0 +1,13 @@
...
+fn main() -> Result<()> {
+    let crate_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+
+    cbindgen::Builder::new()
+        .with_crate(crate_dir)
+        .generate()
+        .expect("Unable to generate bindings")
+        .write_to_file("bindings.h");
+
+    prost_build::compile_protos(&["messages.proto"], &["."])?;
```
tknickman, line 11: Should we make sure this is done on commit?

cli/Makefile
```diff
@@ -17,15 +17,37 @@ GO_FLAGS += -trimpath
...
+	 && cp ffi.go ../../cli/internal/ffi\
+	 && cp -r ffi/proto ../../cli/internal/ffi
+
+# since swig 4.1.0 intgosize is no longer needed, but gh
+# actions on ubuntu and windows still use 4.0.2
+swig: turborepo-ffi
+	cd ../crates/turborepo-ffi && swig -go -cgo -intgosize 64 -c++ swig.i
+
+turborepo-ffi:
+	cd ../crates/turborepo-ffi && cargo build --release
```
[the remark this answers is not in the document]
arlyon: Main goal was consistency really, but I believe specifying the package would work too

crates/turborepo-ffi/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+        // protobuf has a fairly strict schema so overrunning or underrunning the byte
+        // array will not cause any major issues, that is to say garbage in
+        // garbage out
+        let mut slice = unsafe { std::slice::from_raw_parts(self.data, self.len as usize) };
+        T::decode(&mut slice)
+    }
+}
+
+// #[no_mangle]
+// pub extern "C" fn glob(args: Buffer) -> Buffer {
```
NicholasLYang, line 35: What's happening with this code? Are we porting globs too?

NicholasLYang (approved):
Looks great! Only a nit about some commented out code.

chris-olszewski (approved):

cli/internal/ffi/ffi.go
```diff
@@ -0,0 +1,69 @@
...
+	b.len = C.uint(len(bytes))
+	b.data = (*C.uchar)(C.CBytes(bytes))
+	return b
+}
+
+func GetTurboDataDir() string {
+	buffer := C.get_turbo_data_dir()
+	resp := ffi_proto.TurboDataDirResp{}
+	Unmarshal(buffer, resp.ProtoReflect().Interface())
+	fmt.Printf("rust turbo data dir: %s\n", resp.Dir)
```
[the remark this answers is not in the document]
arlyon: perhaps we can get rid of it when we start releasing builds with it? Its handy to prove the execution path is set up

cli/Makefile
```diff
@@ -17,15 +17,31 @@ GO_FLAGS += -trimpath
...
-	CGO_ENABLED=1 go build $(GO_FLAGS) -o go-turbo$(EXT) ./cmd/turbo
+go-turbo$(EXT): $(GENERATED_FILES) $(SRC_FILES) go.mod turborepo-ffi-install
+	CGO_ENABLED=1 go build $(GO_FLAGS) -tags $(GO_TAG) -o go-turbo$(EXT) ./cmd/turbo
+
+.PHONY: turborepo-ffi-install
+turborepo-ffi-install: turborepo-ffi turborepo-ffi-proto
+	cd ../crates/turborepo-ffi\
+	 && cp bindings.h ../../cli/internal/ffi\
+	 && cp ../../target/release/libturborepo_ffi.a ../../cli/internal/ffi\
+	 && cp -r ffi/proto ../../cli/internal/ffi
```
ForsakenHarmony, line 38:
```suggestion
	cp -r \
		../crates/turborepo-ffi/bindings.h \
		../crates/turborepo-ffi/ffi/proto \
		../target/release/libturborepo_ffi.a \
		internal/ffi
```
I'd do it that way

cli/Makefile
```diff
@@ -17,15 +17,31 @@ GO_FLAGS += -trimpath
...
+
+turborepo-ffi:
+	cd ../crates/turborepo-ffi && cargo build --release
+
+turborepo-ffi-proto:
```
ForsakenHarmony, line 43: these make targets should also have a `.PHONY` as they don't generate a file of the same name?

cli/Makefile
```diff
@@ -17,15 +17,31 @@ GO_FLAGS += -trimpath
...
 GO_FILES = $(shell find . -name "*.go")
 SRC_FILES = $(shell find . -name "*.go" | grep -v "_test.go")
 GENERATED_FILES = internal/turbodprotocol/turbod.pb.go internal/turbodprotocol/turbod_grpc.pb.go
 
 turbo: go-turbo$(EXT)
 	cargo build --manifest-path ../crates/turborepo/Cargo.toml
 
-go-turbo$(EXT): $(GENERATED_FILES) $(SRC_FILES) go.mod
-	CGO_ENABLED=1 go build $(GO_FLAGS) -o go-turbo$(EXT) ./cmd/turbo
+go-turbo$(EXT): $(GENERATED_FILES) $(SRC_FILES) go.mod turborepo-ffi-install
```
ForsakenHarmony, line 30: Needs a new target that can get called from the `turborepo-lib` build script without potentially calling `cargo`

cli/internal/ffi/bindings.h
```diff
@@ -0,0 +1,11 @@
+#include <stdarg.h>
```
ForsakenHarmony, line 1: if these files are generated they probably should not be in the .gitignore

crates/turborepo-ffi/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "turborepo-ffi"
```
ForsakenHarmony, line 2: Add this as a dependency to `turborepo`

cli/internal/ffi/bindings.h
arlyon: Should, or shouldn't be? We have been checking in go proto files so I decided to include it.

cli/Makefile
arlyon: I am experimenting with just always building turborepo-ffi in its own `target` dir since it's not _really_ part of the dependency tree...

cli/internal/ffi/bindings.h
nathanhammond: (Current state seems correct to me.)

crates/turborepo-ffi/messages.proto
```diff
@@ -0,0 +1,24 @@
...
+message GlobResp {
+  oneof response {
+    GlobRespList files = 1;
+    string error = 2;
+  }
+}
+
+message GlobRespList {
+  repeated string files = 1;
+}
```
nathanhammond, line 24: I believe that these are unused right now? I believe we should drop them from this PR.

cli/Makefile
nathanhammond: It's a transitive dependency, but Go doesn't know how to build it.

I don't have a strong opinion here.

cli/Makefile
```diff
@@ -17,15 +17,33 @@ GO_FLAGS += -trimpath
...
+	      ../crates/turborepo-ffi/target/release/libturborepo_ffi.a \
+	      ./internal/ffi
+
+.PHONY: turborepo-ffi
+turborepo-ffi:
+	cd ../crates/turborepo-ffi && CARGO_TARGET_DIR=./target cargo build --release
+
+.PHONY: turborepo-ffi-proto
+turborepo-ffi-proto:
+	cd ../crates/turborepo-ffi && protoc --go_out=. messages.proto
```
nathanhammond, line 46: Mild preference for this, which is untested but I think it works:

```suggestion
	cd ../crates/turborepo-ffi && protoc --go_out=$(CLI_DIR)/ffi/proto messages.proto
```

cli/internal/ffi/ffi.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := proto.Unmarshal(bytes, c); err != nil {
+		return err
+	}
+
+	b.Free()
+
+	return nil
+}
+
+// Marshal consumes a proto.Message and returns a bufferfire
```
nathanhammond, line 29: Is a bufferfire similar to a dumpster fire?

cli/internal/ffi/ffi.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"reflect"
+	"unsafe"
+
+	ffi_proto "github.com/vercel/turbo/cli/internal/ffi/proto"
+	"google.golang.org/protobuf/proto"
+)
+
+// Unmarshal consumes a buffer and parses it into a proto.Message
+func Unmarshal[M proto.Message](b C.Buffer, c M) error {
```
nathanhammond, line 18: I _think_ this is our first generic into the Go codebase. I don't think it matters anymore, but I want to make sure we didn't just break linting.

/cc @gsoltis

nathanhammond, line 18: Can we rename these variables? We're not at a limit on characters and `b C.Buffer, c M` is unnecessarily cryptic.

cli/internal/ffi/ffi.go
```diff
@@ -0,0 +1,75 @@
...
+
+	return toBuffer(bytes)
+}
+
+func (c C.Buffer) Free() {
+	C.free(unsafe.Pointer(c.data))
+}
+
+// rather than use C.GoBytes, we use this function to avoid copying the bytes,
+// since it is going to be immediately Unmarshalled into a proto.Message
```
nathanhammond, line 46: This is clever. I still kinda prefer `GoBytes` in spite of the bonus copy.

nathanhammond (commented):
This landed while I was mid-review.

PR #3313, opened 2023-01-13
style(docs): fix prettier

kwonoj:
This is so minor, but seeing CI check failure earliest with this prettier check can be confusing so attempt to fix it.


mehulkar (approved):
saw this too, thanks for the fix

PR #3320, opened 2023-01-15
Add minimal source map for css modules locals js

ahabhgk:


crates/turbopack-css/src/module_asset.rs
```diff
@@ -193,21 +197,37 @@ impl EcmascriptChunkItem for ModuleChunkItem {
...
                 }
             }
         }
         .cell())
     }
 }
+
+fn generate_minimal_source_map(filename: String, source: String) -> ParseResultSourceMapVc {
+    let sm: Arc<SourceMap> = Default::default();
+    sm.new_source_file(FileName::Custom(filename), source);
```
jridgewell, line 230: Since this isn't really a source map, could this just be an empty `source` string?

ahabhgk: There will be a error displayed in the chrome's dev tools, maybe we can leave a message as a placeholder of the `source` string, any suggestions for the message?
Screen Shot 2023-01-16 at 10 43 40

jridgewell: Actually, I guess this is fine.

crates/turbopack-css/src/module_asset.rs
```diff
@@ -193,21 +197,37 @@ impl EcmascriptChunkItem for ModuleChunkItem {
...
+    let map = ParseResultSourceMap::new(sm, vec![(BytePos(1), LineCol { line: 0, col: 0 })]);
```
sokra, line 231: For breakpoints and stepping to work correctly you at least need to generate a SourceMapping per line.

sokra (changes requested):

crates/turbopack-css/src/module_asset.rs
ahabhgk: I changed this to use parse and emit to generate the source map, which made the source map not minimal and have more cost for performance, or do you suggest create the mappings with traversing the `source` line by line? It seems inconvenient but more performance.

jridgewell: I think his suggestion was to add multiple items to this `vec` to represent each line of the output source file. If your interested in undo the changes, this should be as simple as a `source.lines()` and build up all the lines.

ahabhgk: I will undoing the changes and have a try later when I have free time.

crates/turbopack-tests/tests/snapshot/css/css/output/8697f_foo_style.module.css.js.map
```diff
@@ -1,4 +1,6 @@
 {
   "version": 3,
-  "sections": []
+  "sections": [
+    {"offset": {"line": 4, "column": 0}, "map": {"version":3,"sources":["/crates/turbopack-tests/tests/snapshot/css/css/input/node_modules/foo/style.module.css.js"],"sourcesContent":["__turbopack_export_value__({\n  \"foo-module-style\": \"foo-module-style◽[project]/crates/turbopack-tests/tests/snapshot/css/css/input/node_modules/foo/style.module.css\",\n});\n"],"names":[],"mappings":"AAAA;AAA4B;AAC4G"}},
```
sokra:
```suggestion
    {"offset": {"line": 4, "column": 0}, "map": {"version":3,"sources":["/crates/turbopack-tests/tests/snapshot/css/css/input/node_modules/foo/style.module.css.js"],"sourcesContent":["__turbopack_export_value__({\n  \"foo-module-style\": \"foo-module-style◽[project]/crates/turbopack-tests/tests/snapshot/css/css/input/node_modules/foo/style.module.css\",\n});\n"],"names":[],"mappings":"AAAA;AACA;AACA"}},
```

sokra (approved):
Was just about to tell you about `lines()` and line endings, but you already fixed it

sokra:
Thanks

PR #3324, opened 2023-01-16
improve resolveAlias to allow conditions to allow to apply aliases to…

sokra:
… client/server only


crates/next-core/src/next_import_map.rs
```diff
@@ -28,6 +32,7 @@ pub async fn get_next_client_import_map(
         &mut import_map,
         project_path,
         next_config.resolve_alias_options(),
+        true,
```
ForsakenHarmony, line 35: Maybe we could use an enum, booleans are quite hard to read in function calls

wbinnssmith (approved):

PR #3325, opened 2023-01-16
rename turbopackWebpackLoaders to turbopackLoaders

sokra:
since they don't have exactly the same API


ForsakenHarmony:
But do we want to keep this as the official loader API?
Probably not, so `turbopackLoaders` doesn't seem like the best name

sokra:
> But do we want to keep this as the official loader API?

It's only to get front working. We need to decide on an official API/configuration.

PR #3326, opened 2023-01-16
improve SSR data handling for pages

sokra:


ForsakenHarmony (approved):

sokra:
> Does it make a difference? `_app` and `_document` would have been parsed already
> 
> This kinda seems like a micro optimization

For client side navigation we only need data, so we can create smaller bundles and faster execution.

But the real reason is that the data transform removes the components from all files in the pages directory, including `_app`. It crashed when proving undefined for App.

PR #3327, opened 2023-01-16
add stack traces to console.error and console.warn

sokra:
also see WEB-429 to reduce duplication


ForsakenHarmony (approved):
Why do we want this?

sokra:
see #4028

PR #3331, opened 2023-01-16
fix error overlay sometimes now showing

ForsakenHarmony:


ForsakenHarmony:
* main
  * **PR #3331** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3331" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3331?utm_source=stack-comment).

crates/next-core/js/src/dev/hmr-client.ts
```diff
@@ -116,6 +116,15 @@ function aggregateUpdates(
   const key = resourceKey(msg.resource);
   const aggregated = chunksWithUpdates.get(key);
 
+  if (msg.type === "issues" && aggregated == null && hasIssues) {
+    // add an empty record to make sure we don't call `onBuildOk`
+    chunksWithUpdates.set(key, {
```
ForsakenHarmony, line 121: We check if there are entries in the map below, so even if we have no updates we insert an empty record to make sure it doesn't think everything is fine

crates/next-core/js/src/dev/hmr-client.ts
```diff
@@ -116,6 +116,15 @@ function aggregateUpdates(
   const key = resourceKey(msg.resource);
   const aggregated = chunksWithUpdates.get(key);
 
+  if (msg.type === "issues" && aggregated == null && hasIssues) {
```
jridgewell, line 119: Nit: Looks like this can be merged with the if statements below?

jridgewell (approved):

wbinnssmith:
Any chance this fixes WEB-439?

crates/next-core/js/src/dev/hmr-client.ts
ForsakenHarmony: could be, but imo makes it harder to read

ForsakenHarmony:
@wbinnssmith it might

PR #3333, opened 2023-01-16
allow hiding error overlay with resolve errors

ForsakenHarmony:


ForsakenHarmony:
* main
  * **PR #3333** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3333" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3333?utm_source=stack-comment).

crates/next-core/js/src/overlay/internal/container/Errors.tsx
```diff
@@ -154,8 +154,10 @@ export function Errors({ issues, errors }: ErrorsProps) {
   );
 
   const hasIssues = issues.length !== 0;
-  const hasIssueWithError = issues.some((issue) =>
-    ["bug", "fatal", "error"].includes(issue.severity)
+  const hasIssueWithError = issues.some(
+    (issue) =>
+      ["bug", "fatal", "error"].includes(issue.severity) &&
+      !(issue.severity === "error" && issue.category === "resolve")
```
alexkirsz, line 160: Could you add a short comment explaining why we don't consider resolve issues as errors?

alexkirsz (approved):

crates/next-core/js/src/overlay/internal/container/Errors.tsx
sokra: I think we shouldn't use the category for that decision. This should be encoded in the `severity` only. So:
* reduce resolve errors to `warning`, or
* allow to hide `error`s in general

I feel like resolve errors are a bit weak as `warnings`. So I would prefer keeping them errors.

Maybe we need a new severity type that is between `fatal` and `error`.

For reference the descriptions of the error types:

```
            IssueSeverity::Fatal => "unrecoverable problem",
            IssueSeverity::Error => "problem that cause a broken result",
            IssueSeverity::Warning => "problem should be addressed in short term",
```

There would be space for a `problem that cause a partial broken result`...

sokra (changes requested):

crates/next-core/js/src/overlay/internal/container/Errors.tsx
ForsakenHarmony: I accidentally unmarked this as a draft, just wanted to keep this around because I needed it at some point

PR #3334, opened 2023-01-16
refactor content sources to avoid evaluating content before serving

sokra:
fixes WEB-141

This changes the ContentSource API.

Before a ContentSource returned a tuple of specificity and content.
Now it returns a tuple of specificity and a get_content function.

The old way made it very inefficient to combine multiple ContentSource.
All combined ContentSources were ask before, all returned specificity and content, and they are ordered after that. That required to compute all contents, even if it is not really used after ordering.

Now we order specificity and a get_content function and only compute the content (by calling get_content) with the final result.

It also changes how data needs are expressed in the `get_content` function. Instead of asking it over and over again with more data, it statically returns the needed data with `vary` and the `get` function is only called with that data. This is technically more limited than the old way, but one can workaround that if needed. But practically we probably never need that. On the other hand it improves the number of function calls, since the `vary` method is only called once and the `get` only once per request. Before data needs required at least 2 `get` calls per request.

### DIFF
--- a/crates/turbopack-dev-server/src/source/conditional.rs
+++ b/crates/turbopack-dev-server/src/source/conditional.rs
@@ -1,11 +1,13 @@
 use anyhow::Result;
-use turbo_tasks::{primitives::StringVc, State};
+use turbo_tasks::{primitives::StringVc, State, Value};
 use turbopack_core::introspect::{Introspectable, IntrospectableChildrenVc, IntrospectableVc};
 
 use super::{
-    ContentSource, ContentSourceContent, ContentSourceData, ContentSourceResultVc, ContentSourceVc,
+    combined::CombinedContentSource, ContentSource, ContentSourceData, ContentSourceDataVaryVc,
+    ContentSourceResult, ContentSourceResultVc, ContentSourceVc, GetContentSourceContent,
+    GetContentSourceContentVc,
 };
-use crate::source::ContentSourcesVc;
+use crate::source::{ContentSourceContentVc, ContentSourcesVc};
 
 /// Combines two [ContentSource]s like the [CombinedContentSource], but only
 /// allows to serve from the second source when the first source has
@@ -41,25 +43,31 @@ impl ConditionalContentSourceVc {
 impl ContentSource for ConditionalContentSource {
     #[turbo_tasks::function]
     async fn get(
-        &self,
+        self_vc: ConditionalContentSourceVc,
         path: &str,
         data: turbo_tasks::Value<ContentSourceData>,
     ) -> Result<ContentSourceResultVc> {
-        let first = self.activator.get(path, data.clone());
-        if let ContentSourceContent::NotFound = &*first.await?.content.await? {
-            if !*self.activated.get() {
-                return Ok(first);
-            }
+        let this = self_vc.await?;
+        if !*this.activated.get() {
+            let first = this.activator.get(path, data.clone());
+            let first_value = first.await?;
+            return Ok(match &*first_value {
+                &ContentSourceResult::Result {
+                    get_content,
+                    specificity,
+                } => ContentSourceResult::Result {
+                    get_content: ActivateOnGetContentSource(this, get_content).cell().into(),
+                    specificity,
+                }
+                .cell(),
+                _ => first,
+            });
         }
-        self.activated.set(true);
-        let second = self.action.get(path, data);
-        let first_specificity = first.await?.specificity.await?;
-        let second_specificity = second.await?.specificity.await?;
-        if first_specificity >= second_specificity {
-            Ok(first)
-        } else {
-            Ok(second)
+        Ok(CombinedContentSource {
+            sources: vec![this.activator, this.action],
         }
+        .cell()
+        .get(path, data))
     }
 
     #[turbo_tasks::function]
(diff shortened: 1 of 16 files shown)


crates/turbopack-dev-server/src/source/conditional.rs
```diff
@@ -41,25 +43,31 @@ impl ConditionalContentSourceVc {
...
         path: &str,
         data: turbo_tasks::Value<ContentSourceData>,
     ) -> Result<ContentSourceResultVc> {
-        let first = self.activator.get(path, data.clone());
-        if let ContentSourceContent::NotFound = &*first.await?.content.await? {
-            if !*self.activated.get() {
-                return Ok(first);
-            }
+        let this = self_vc.await?;
+        if !*this.activated.get() {
```
jridgewell, line 51: Or does this refactor solve the performance penalty by not activating the main source yet?

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -240,10 +284,12 @@ impl Default for BodyVc {
...
 
 /// Filter function that describes which information is required.
-#[derive(Debug, Clone, PartialEq, Eq, TraceRawVcs, Serialize, Deserialize)]
+#[derive(
+    Debug, Clone, PartialEq, Eq, TraceRawVcs, Hash, PartialOrd, Ord, Serialize, Deserialize,
+)]
 pub enum ContentSourceDataFilter {
     All,
-    Subset(HashSet<String>),
+    Subset(BTreeSet<String>),
```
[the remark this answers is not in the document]
sokra: I use `BTreeSet` because it's `Hash, PartialOrd, Ord` which HashSet is not.

The order is not relevant, so it's not `IndexSet`.

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -280,13 +326,30 @@ impl ContentSourceDataFilter {
             ContentSourceDataFilter::Subset(set) => set.contains(key),
         }
     }
+
+    pub fn fulfills(
```
alexkirsz, line 330: Please comment this item.

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -311,6 +374,28 @@ impl ContentSourceDataVary {
         ContentSourceDataFilter::extend_options(&mut self.query, &other.query);
         ContentSourceDataFilter::extend_options(&mut self.headers, &other.headers);
     }
+
+    pub fn fulfills(&self, other: &ContentSourceDataVary) -> bool {
```
alexkirsz, line 378: Please add a comment to this public item.

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -311,6 +374,28 @@ impl ContentSourceDataVary {
...
+        if other.cache_buster && !self.cache_buster {
+            return false;
+        }
+        if !ContentSourceDataFilter::fulfills(&self.query, &other.query) {
+            return false;
+        }
+        if !ContentSourceDataFilter::fulfills(&self.headers, &other.headers) {
+            return false;
+        }
+        true
```
alexkirsz, line 397: This could easily break if we were to add new extensions to `ContentSourceDataVary`.

I would argue that either:
* `fulfills` should be a trait that is automatically derived; or
* we use destructuring assignment in this function to not forget any fields.

```rust
let ContentSourceDataVary { method: other_method, /* etc., but no `..` */ } = &other;
```

But this clashes with `placeholder_for_future_extensions`.

crates/turbopack-dev-server/src/update/stream.rs
```diff
@@ -41,28 +40,26 @@ fn extend_issues(issues: &mut Vec<PlainIssueReadRef>, new_issues: Vec<PlainIssue
...
-    Ok(match *content_source_result.await?.content.await? {
-        ContentSourceContent::NotFound => None,
-        ContentSourceContent::HttpProxy(_) => {
-            panic!("HTTP proxying is not supported in UpdateStream")
-        }
-        ContentSourceContent::Static(content) => Some(content),
-        ContentSourceContent::NeedData(_) => {
-            bail!("this might only happen temporary as get_content_wrapper resolves the data")
+    Ok(match &*content {
+        ContentSourceResult::NeedData(..) => unreachable!(),
```
alexkirsz, line 55: The `while` above should be a `match` within a `loop` to avoid this `unreachable`:

```rust
loop {
    match &*content {
        ContentSourceResult::NeedData(data) => {
          content = ...
        }
        ContentSourceResult::NotFound => return ...,
        ContentSourceResult::Result { .. } => return ...,
    }
}
```

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -37,32 +37,58 @@ pub struct ProxyResult {
...
-            specificity: SpecificityVc::not_found(),
-            content: ContentSourceContent::NotFound.cell(),
-        }
-        .cell()
+        ContentSourceResult::NotFound.cell()
+    }
+}
+
+#[turbo_tasks::value_trait]
+pub trait GetContentSourceContent {
```
alexkirsz, line 87: Please document this trait and how it interacts with `ContentSource`. I'm having trouble understanding why we need the additional indirection.

From looking at the trait definition, rather than a `GetContentSourceContent`, this feels more like a `DeferredContent`, i.e. a content source's content that we know exists but is yet to be computed.

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -37,32 +37,58 @@ pub struct ProxyResult {
...
+    NotFound,
+    NeedData(NeededData),
+    Result {
+        specificity: SpecificityVc,
+        get_content: GetContentSourceContentVc,
+    },
+}
+
+#[turbo_tasks::value_impl]
+impl ContentSource for ContentSourceResult {
```
alexkirsz, line 50: image

crates/turbopack-dev-server/src/source/conditional.rs
```diff
@@ -116,3 +124,20 @@ impl Introspectable for ConditionalContentSource {
         ))
     }
 }
+
+#[turbo_tasks::value(serialization = "none", eq = "manual", cell = "new")]
+struct ActivateOnGetContentSource(ConditionalContentSourceReadRef, GetContentSourceContentVc);
```
alexkirsz, line 129: Please use named struct fields here.

alexkirsz (approved):

jridgewell (approved):
I'd like @alexkirsz's items to be addressed, but LGTM.

PR #3335, opened 2023-01-16
reduce function count

sokra:
avoid creating very many resolve tasks by resolving the arguments before calling methods on it


ForsakenHarmony (approved):
too many footguns in turbo tasks 🙃

jridgewell:
Should this be a macro attribute on the `#[turbo_tasks::function]`?

PR #3347, opened 2023-01-17
Expose `process.turbopack` to js run from turbopack-node

wbinnssmith:
For now, this can allow `next.config.js` to understand where it's been run and apply different options. We should probably have better compatibility in the future, which would make this unnecessary for this usecase.

Test Plan: Modified a `next.config.js` to change its behavior based on the presence of this option. Verified non-turbo case doesn't have this exposed.

### DIFF
--- a/crates/turbopack-node/js/src/globals.ts
+++ b/crates/turbopack-node/js/src/globals.ts
@@ -1,0 +1,2 @@
+// @ts-ignore
+process.turbopack = {};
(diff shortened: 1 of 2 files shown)


sokra (changes requested):
We need something compile-time evaluatable to shortcircut e. g. `if(!process.turbopack)` and skip resolving references in that block.
Sadly we are missing the infrastructure for that still. So this will be a bigger task

sokra:
Was merged

PR #3079, opened 2022-12-19
fix(shim): Fixed cwd parsing bug

NicholasLYang:
If `cwd` is passed as a single chunk, e.g. `--cwd=./path/to/foo`, the `env::arg` iterator provides it as a single arg. We need to parse out the `cwd` from that arg.

PR #3348, opened 2023-01-17
fix(turborepo): Passing a relative path to --cwd when using --skip-infer.

NicholasLYang:
Ran into this while testing changes. In general we should figure out better abstractions for relative and absolute paths in Rust.

### DIFF
--- a/crates/turborepo-lib/src/cli.rs
+++ b/crates/turborepo-lib/src/cli.rs
@@ -3,6 +3,7 @@ use std::{env, io, mem, path::PathBuf, process};
 use anyhow::{anyhow, Result};
 use clap::{ArgAction, CommandFactory, Parser, Subcommand, ValueEnum};
 use clap_complete::{generate, Shell};
+use dunce::canonicalize as fs_canonicalize;
 use log::error;
 use serde::Serialize;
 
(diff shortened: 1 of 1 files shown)


mehulkar (commented):
What problem does this solve? I read through what this library does here: https://docs.rs/dunce/latest/dunce/#functions, but I'm not sure (based on PR title and diff) what `--cwd` input this addresses. 

Are we _trying_ to pass relative path? Or does this _handle_ when a relative path is passed? In the case of the latter, does this deserve a dedicated prysk integration test?

tknickman (approved):

tknickman:
@mehulkar more details here, we overlooked this one from this original fix: https://github.com/vercel/turbo/pull/3079/files

NicholasLYang:
@mehulkar  yeah basically we need to pass absolute paths to Go and we're not currently doing that when we have the `--skip-infer` flag

mehulkar:
Curious, why this additional normalization isn't also in the `shim.rs`, only in cli/run handling?

NicholasLYang:
It is, but in this case with `--skip-infer` we bypass the shim entirely. In the future I would prefer a better abstraction where we have a type called say, `AbsolutePath` that when constructed canonicalizes the path automatically, but that is a larger task (maybe a good first Rust issue if anybody's interested!)

PR #3350, opened 2023-01-18
ci(actions): report failed --turbo next.js tests

kwonoj:
Closes WEB-168.

This PR changes the behavior of next.js integration test in this repo to run everything with `--turbo` by default, instead of manually opt in per each PR. Since current next.js integration tests are expected to fail with turbopack, PR adds changes

1. Do not fail CI when next.js integration test fails
So we can keep CI checker green for  the daily workflow

2. Report failed tests in the PR comment 
Instead, report failed tests in the comment, like https://github.com/vercel/turbo/pull/3350#issuecomment-1387356900

This is largely similar to what next.js upstream does for failing report, but does not use exact same webhook since the way CI pipeline sets up is bit different so we can't use it out of the box. However, the logic in the custom action is short & straightforward enough to keep internally.

One thing to note is this report uses default automatic github auth token (user `github-actions` bot). This means report comment **will not be created if PR is created from fork**, but I expect that's acceptable limitation.

There maybe some improvements we want to add, like diff-between-main but probably not a high priority & maybe we can fix all the turbopack faster than having those.


kwonoj:
For the ref, original webhook is at https://github.com/vercel/next-pr-webhook

sokra (approved):
Note that it still doesn't run all next.js tests since it only runs group 1 and 2 or 4 resp. 7 resp 25 total groups

ForsakenHarmony (commented):
forgot to submit

PR #3351, opened 2023-01-18
fix(docs): remove empty array instruction

tknickman:
`outputs: []` is no longer required to cache logs, and can be omitted entirely.

### DIFF
--- a/docs/pages/repo/docs/core-concepts/caching.mdx
+++ b/docs/pages/repo/docs/core-concepts/caching.mdx
@@ -71,7 +71,7 @@ To override the default cache output behavior, pass an array of globs to a [`pip
 }
 ```
 
-If your task does not emit any files (e.g. unit tests with Jest), set `outputs` to an empty array (i.e. `[]`), and Turborepo will cache the logs for you.
+If your task does not emit any files (e.g. unit tests with Jest), Turborepo will automatically cache the logs for you.
 
 When you run `turbo run build test`, Turborepo will execute your build and test scripts,
 and cache their `output`s in `./node_modules/.cache/turbo`.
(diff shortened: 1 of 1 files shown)


docs/pages/repo/docs/core-concepts/caching.mdx
```diff
@@ -71,7 +71,7 @@ To override the default cache output behavior, pass an array of globs to a [`pip
 }
 ```
 
-If your task does not emit any files (e.g. unit tests with Jest), set `outputs` to an empty array (i.e. `[]`), and Turborepo will cache the logs for you.
+If your task does not emit any files (e.g. unit tests with Jest), Turborepo will automatically cache the logs for you.
```
nathanhammond, line 74:
```suggestion
Turborepo automatically records and caches the logs of every task. You do not need to specify anything to get this behavior. If your task does not emit any files (e.g. unit tests with Jest) you can simply omit `outputs`.
```

nathanhammond (approved):
Just a slightly different wording pitch.

docs/pages/repo/docs/core-concepts/caching.mdx
```diff
@@ -71,7 +71,7 @@ To override the default cache output behavior, pass an array of globs to a [`pip
 }
 ```
 
-If your task does not emit any files (e.g. unit tests with Jest), set `outputs` to an empty array (i.e. `[]`), and Turborepo will cache the logs for you.
+Turborepo automatically records and caches the logs of every task. You do not need to specify anything to get this behavior. If your task does not emit any files (e.g. unit tests with Jest) you can simply omit `outputs`.
```
mehulkar, line 74: I like the rewording around the logs, but feel like we should keep the continuity from caching file outputs and explain how that relates to caching logs after, rather than jumping directly to logs.

```suggestion
If your task does not emit any files (e.g. unit tests with Jest) you can omit `outputs`. Even without any file outputs, Turborepo automatically records and caches the logs of every task. If no inputs change (i.e. if there is a cache hit), subsequent runs will replay these logs.
```

mehulkar (approved):
Small suggestion, but overall 👍🏾 🚢 . Thanks for catching this!

PR #3352, opened 2023-01-18
feat: add support for pnpm v6 lockfiles

chris-olszewski:
Add support for pnpm lockfile v6 (https://github.com/pnpm/pnpm/pull/5810)

Notable changes in the lockfile v6:
 - Lockfile keys are encoded as `/name@version` as opposed to `/name/version`
 - Specifiers are inline with resolved versions as opposed to being in their own field
 - Packages with multiple peer dependencies no longer have the peer dependencies hashed

Notes:
 - We currently just attempt to parse the v5 structure and if that fails we attempt v6. We do this since lockfile version can be decided by a combination of `pnpm` version and `.npmrc` contents. In the future once `pnpm@8` is widely used we can switch the ordering of this if it becomes a performance bottle neck
 - A lot of the implementation was copy/pasta for the *slightly* different structure. I could have gone the route of making most of these functions work on `map[string]interface{}`, but I don't love doing type assertions all over the place.


mehulkar (approved):
Should we start porting the tests around these lockfiles to integration tests at some point? Otherwise looks good

PR #3353, opened 2023-01-18
fix(prettier): format

tknickman:
Fix prettier


kwonoj (approved):

nathanhammond (approved):
If this file is generated from somewhere, we should excluded. If this file is manually created and maintained, this is fine.

PR #3343, opened 2023-01-17
add full next.js config

ForsakenHarmony:

PR #3361, opened 2023-01-18
external node_modules for SSR

ForsakenHarmony:
Adds resolve plugins (currently only running `after_resolve`, could add `resolve` later)

Had to fix `ResolveResult::merge_alternatives` because it merged two external results into `Unresolvable`

### DIFF
--- a/crates/next-core/src/next_server/resolve.rs
+++ b/crates/next-core/src/next_server/resolve.rs
@@ -68,7 +68,7 @@ impl ResolvePlugin for ExternalCjsModulesResolvePlugin {
 
         // check if we can resolve the package from the root dir (might be hidden by
         // pnpm)
-        let FileJsonContent::Content(_) = *self
+        let FileJsonContent::Content(resolved_package) = &*self
             .root
             .join(&format!("node_modules/{}/package.json", package_name))
             .read_json()
@@ -76,6 +76,11 @@ impl ResolvePlugin for ExternalCjsModulesResolvePlugin {
         else {
             return Ok(ResolveResultOptionVc::none());
         };
+
+        // only mark external if the package.json files are the same
+        if resolved_package != package {
+            return Ok(ResolveResultOptionVc::none());
+        }
 
         // mark as external
         Ok(ResolveResultOptionVc::some(
--- a/crates/next-core/src/next_server/resolve.rs
+++ b/crates/next-core/src/next_server/resolve.rs
@@ -1,0 +1,72 @@
+use anyhow::Result;
+use turbo_tasks::primitives::StringsVc;
+use turbo_tasks_fs::{glob::GlobVc, FileJsonContent, FileSystemPathVc};
+use turbopack_core::resolve::{
+    find_context_file, package_json,
+    parse::RequestVc,
+    plugin::{ResolvePlugin, ResolvePluginConditionVc, ResolvePluginVc, ResolveResultOptionVc},
+    FindContextFileResult, ResolveResult, SpecialType,
+};
+
+#[turbo_tasks::value]
+pub(crate) struct ExternalCjsModulesResolvePlugin {
+    root: FileSystemPathVc,
+    transpiled_packages: StringsVc,
+}
+
+#[turbo_tasks::value_impl]
+impl ExternalCjsModulesResolvePluginVc {
+    #[turbo_tasks::function]
+    pub fn new(root: FileSystemPathVc, transpiled_packages: StringsVc) -> Self {
+        ExternalCjsModulesResolvePlugin {
+            root,
+            transpiled_packages,
+        }
+        .cell()
+    }
+}
+
+#[turbo_tasks::value_impl]
+impl ResolvePlugin for ExternalCjsModulesResolvePlugin {
+    #[turbo_tasks::function]
+    fn condition(&self) -> ResolvePluginConditionVc {
+        ResolvePluginConditionVc::new(self.root, GlobVc::new("**/node_modules"))
+    }
+
+    #[turbo_tasks::function]
+    async fn after_resolve(
+        &self,
+        fs_path: FileSystemPathVc,
+        _request: RequestVc,
+    ) -> Result<ResolveResultOptionVc> {
+        let transpiled_glob = packages_glob(self.transpiled_packages).await?;
+
+        // always bundle transpiled modules
+        if transpiled_glob.execute(&fs_path.await?.path) {
+            return Ok(ResolveResultOptionVc::none());
+        }
+
+        // check `package.json` for `"type": "module"`
+        if let FindContextFileResult::Found(package_json, _) =
+            &*find_context_file(fs_path.parent(), package_json()).await?
+        {
+            if let FileJsonContent::Content(package) = &*package_json.read_json().await? {
+                if let Some("module") = package["type"].as_str() {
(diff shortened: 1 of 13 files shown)


ForsakenHarmony:
* main
  * **PR #3343** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3343" target="_blank">Graphite</a> 
    * **PR #3361** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3361" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3361?utm_source=stack-comment).

crates/turbopack-core/src/resolve/plugin.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    #[turbo_tasks::function]
+    pub fn none() -> Self {
+        ResolveResultOption(None).cell()
+    }
+}
+
+#[turbo_tasks::value_trait]
+pub trait ResolvePlugin {
```
alexkirsz, line 55: Should this be a `ResolvePlugin` or a straight up `Resolver`? I'm asking because in https://github.com/vercel/turbo/blob/89ead9b9d4a78b27c3eb100b4de580ed7ddd82c0/crates/turbopack-core/src/resolve/mod.rs#L576 we already have three clear passes:
1. `ImportMap(Resolver)`
2. `Request(Resolver)`
3. `FallbackImportMap(Resolver)`

And so it seems like this would essentially add intermediate resolvers between 2. and 3.

crates/next-core/src/next_server/resolve.rs
```diff
@@ -0,0 +1,98 @@
...
+        let FindContextFileResult::Found(package_json, _) =
+            *find_context_file(fs_path.parent(), package_json()).await?
+        else {
+            // can't find package.json
+            return Ok(ResolveResultOptionVc::none());
+        };
+        let FileJsonContent::Content(package) = &*package_json.read_json().await? else {
+            // can't parse package.json
+            return Ok(ResolveResultOptionVc::none());
+        };
```
alexkirsz, line 57: This could be a single `find_and_read_package_json` method that could be reused in `crates/turbopack-core/src/resolve/mod.rs`.

crates/turbopack/src/resolve_options_context.rs
```diff
@@ -40,6 +43,7 @@ pub struct ResolveOptionsContext {
     /// A list of rules to use a different resolve option context for certain
     /// context paths. The first matching is used.
     pub rules: Vec<(ContextCondition, ResolveOptionsContextVc)>,
+    pub plugins: Vec<ResolvePluginVc>,
```
alexkirsz, line 46: Please add a comment to this option.

crates/turbopack-core/src/resolve/plugin.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn condition(&self) -> ResolvePluginConditionVc;
+    fn after_resolve(&self, fs_path: FileSystemPathVc, request: RequestVc)
+        -> ResolveResultOptionVc;
+}
```
alexkirsz, line 59: Please add comments explaining the role of this trait and of each of its items.

crates/turbopack-core/src/resolve/plugin.rs
```diff
@@ -0,0 +1,59 @@
+use anyhow::Result;
+use turbo_tasks::primitives::BoolVc;
+use turbo_tasks_fs::{glob::GlobVc, FileSystemPathVc};
+
+use crate::resolve::{parse::RequestVc, ResolveResultVc};
+
+#[turbo_tasks::value]
+pub struct ResolvePluginCondition {
```
alexkirsz, line 8: Please add comments to these public items.

crates/next-core/src/next_server/resolve.rs
ForsakenHarmony: Was looking into it, but because I wanted both path and content it would require a custom return type

matching on the `FileJsonContent::Content` is also only possible here because `package` is a `&serde_json::Value` (can't return that directly)

crates/turbopack-core/src/resolve/plugin.rs
ForsakenHarmony: Not sure if I understand the question, what's a `Resolver`? Would that be a new type?

The `ResolvePlugin` right now is kinda part of pass 2, we already have the `resolved_map` in there.
And it would later on also become part of pass 1

Was thinking the import map could be implemented as a `ResolvePlugin` (not that we would necessarily want that)

alexkirsz: `Resolver` is just another name for `ResolvePlugin`. I'm saying the current resolving logic can be separated into 3 `ResolvePlugins`.

crates/turbopack-core/src/resolve/plugin.rs
ForsakenHarmony: done

crates/turbopack-core/src/resolve/plugin.rs
ForsakenHarmony: done

crates/turbopack/src/resolve_options_context.rs
ForsakenHarmony: done

alexkirsz (approved):

crates/next-core/src/next_server/resolve.rs
```diff
@@ -0,0 +1,98 @@
...
+        // check if we can resolve the package from the root dir (might be hidden by
+        // pnpm)
+        let FileJsonContent::Content(resolved_package) = &*self
+            .root
+            .join(&format!("node_modules/{}/package.json", package_name))
+            .read_json()
+            .await?
+        else {
+            return Ok(ResolveResultOptionVc::none());
+        };
```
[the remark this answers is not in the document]
ForsakenHarmony: We don't get the original request here, so that's hard to do

Are there any cjs modules that have an inner package.json?

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -153,6 +154,17 @@ impl ResolveResult {
...
             },
+            ResolveResult::Special(special, list) => match other {
+                ResolveResult::Special(special2, list2) if special2 == special => {
+                    list.extend(list2.iter().cloned());
+                }
+                ResolveResult::Unresolveable(list2) => {
+                    list.extend(list2.iter().cloned());
+                }
+                _ => {
+                    *self = ResolveResult::Unresolveable(take(list));
```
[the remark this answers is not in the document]
ForsakenHarmony: good spot

crates/next-core/src/next_server/resolve.rs
```diff
@@ -0,0 +1,98 @@
...
+            // can't find package.json
+            return Ok(ResolveResultOptionVc::none());
+        };
+        let FileJsonContent::Content(package) = &*package_json.read_json().await? else {
+            // can't parse package.json
+            return Ok(ResolveResultOptionVc::none());
+        };
+
+        // always bundle esm modules
+        if let Some("module") = package["type"].as_str() {
```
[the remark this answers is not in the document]
ForsakenHarmony: I see, done

PR #3367, opened 2023-01-18
feat(codemod): turbo migrate

tknickman:
## Details:
* New migrate command that will run all codemods needed to move from turbo `x` to turbo `y`  and will (optionally) upgrade turbo for you as well
* Rewritten transforms using new transform class that automatically handles logging, determining when to write the file, file diffs, consistent outputs, and consistent logging
* When running with --print, we now show changed file diffs complete with colors
* 100% test coverage for all transforms
* A codemod for adding new codemods (:yodog:)
    * `pnpm add-transform`. Adds tests, fixtures, and new transform boilerplate.
* Moved from meow to commander.js for arg parsing

## Reviewers Guide
This PR is huge. But it’s almost all tests and fixtures. I have split the commits for easier review:
1. [new arg parsing](https://github.com/vercel/turbo/commit/21ff5ae0bc928fe9485dbda8304ba4c4a80fa75d)
2. [rewrite transforms](https://github.com/vercel/turbo/commit/441de33121ecc3be23e32866184d1348f4b7cef0)
3. [add codemod generation](https://github.com/vercel/turbo/commit/45070d2d63bfd75dff13cf07b7649e4b840dabc6)
4. [add migrate command](https://github.com/vercel/turbo/commit/3cd8ed6dba9bcbd84ede5a6aec1322a128bd58d8)
5. [add tests](https://github.com/vercel/turbo/commit/cdbb17c0e26bd8da6ac0efdc4bfd176116c86694)

### FAQ
**Q**. Why not just use `jscodeshift`? 
**A**. It would be major overkill. We don’t need AST’s. Everything we touch (and likely will touch for the foreseeable future) is JSON.

**Q**. Why add an extra class for file transforms?
**A**. This abstracts away the need to care about the CLI options or handle output. Writing a new transform is as simple as - 1. read the file, 2. make changes, 3. pass it to runner.modifyFile. It does the rest (logging, determining when to write, file diffing, consistent outputs, consistent logging)

## TODO:

- [ ] Update codemod docs and in-product reccomendations


packages/turbo-codemod/package.json
```diff
@@ -17,41 +17,46 @@
...
     "lint": "eslint src/**/*.ts",
-    "check-types": "tsc --noEmit"
+    "check-types": "tsc --noEmit",
+    "add-transformer": "plop"
   },
   "dependencies": {
+    "axios": "0.27.2",
     "chalk": "2.4.2",
-    "execa": "5.1.1",
+    "commander": "^9.5.0",
```
mehulkar, line 26: create-turbo uses `meow`. Do we want to either use commander there or use meow here?

packages/turbo-codemod/src/commands/migrate/index.ts
```diff
@@ -0,0 +1,197 @@
...
+    } else {
+      console.log(`Upgrading turbo with ${chalk.bold(upgradeCommand)}`, os.EOL);
+      execSync(upgradeCommand, { cwd: selectedDirectory });
+    }
+  } else {
+    console.log(`Upgrade turbo with ${chalk.bold(upgradeCommand)}`, os.EOL);
+  }
+
+  endMigration({ success: true });
+}
```
mehulkar, line 197: It would be somewhat cool to generate a patch at the end of each migration. Might get a little hairy to get that intermediate diffs if they're all going to be unstaged. Could move to a new branch and do `git commit` in between each step also. The main idea being that you want to see the change each codemod made, in addition to the sum total. Might also help _us_ field debug requests.

packages/turbo-codemod/src/commands/migrate/steps/getLatestVersion.ts
```diff
@@ -0,0 +1,23 @@
...
+
+async function getLatestVersion({
+  to,
+}: MigrateCommandOptions): Promise<string | undefined> {
+  if (to) {
+    return Promise.resolve(to);
+  }
+
+  try {
+    const result = await axios.get(`${REGISTRY}/turbo`);
```
mehulkar, line 15: Can we use fetch? (or undici for older Node support)

packages/turbo-codemod/src/transforms/create-turbo-config.ts
```diff
@@ -1,64 +1,70 @@
...
 }
+
+const transformerMeta = {
+  name: `${TRANSFORMER}: ${DESCRIPTION}`,
+  value: TRANSFORMER,
+  introducedIn: INTRODUCED_IN,
+  transformer,
+};
+
+export default transformerMeta;
```
mehulkar, line 70: bit weird to export an object as the default export (reminiscent of CJS). Could also do named exports for each of the keys:

```
export const name = `TRANSFORMER...`
export const value = TRANSFORMER
// etc
```

mehulkar (approved):
Nothing super blocking, but left a few inline comments. Nice work!

packages/turbo-codemod/package.json
tknickman: Yea good callout. `meow` is _super_ light weight. Which is great for very basic CLI's (single command). But once you get more than a single command and need nested help etc, `meow` gets to be pretty verbose. Commander has a nice API, handles a lot of this for you, and is zero dependencies so it's a good choice. 

I think when we revisit create-turbo we could consider using it there as well.

packages/turbo-codemod/src/commands/migrate/steps/getLatestVersion.ts
tknickman: Fetch isn't native to node (_yet_, it's partially in 18, but we can't rely on that). So we would need a dependency here regardless, and axios is the defacto standard for node (we also use axios in docs and turbopack already).

packages/turbo-codemod/src/transforms/create-turbo-config.ts
tknickman: This is done for simpler type safety. Because we dynamically require the transform files, we establish an API that a transformer _must_ satisfy. A single export of the `Transform` type:

```
export type Transformer = {
  name: string;
  value: string;
  introducedIn: string;
  transformer: (args: TransformerArgs) => TransformerResults;
};
```

packages/turbo-codemod/src/commands/migrate/index.ts
tknickman: By default, this will already show you a condensed table view of the files modified, with the additions and deletions _per transform_. You can also pass in `--print` to optionally see the diffs of the files themselves for each transform.

packages/turbo-codemod/src/commands/migrate/steps/getLatestVersion.ts
mehulkar: Agreed, we need a dependency, but we should we looking to move towards undici, since it's effectively the polyfill for spec-compliant fetch in Node. (It _is_ the implementation pulled into node itself).

packages/turbo-codemod/src/commands/migrate/index.ts
arlyon: A temp branch is a cool idea, I haven't seen any other codemod tools do that.

packages/turbo-codemod/src/commands/migrate/steps/getTurboUpgradeCommand.ts
```diff
@@ -0,0 +1,182 @@
...
+    };
+  }
+
+  return {
+    installType: undefined,
+    isUsingWorkspaces,
+  };
+}
+
+/*
```
arlyon, line 126: Minor thing, but I believe JSDoc expects /** for doc comments which I believe is a convention most js docgen tools now expect also.

arlyon (approved):
Small comment otherwise looks good.

packages/turbo-codemod/src/commands/migrate/index.ts
mehulkar: > see the diffs of the files themselves for each transform.

oh that's cool, if that's possible. Is it a full diff _per_ transform? Are you storing somewhere in between each transform? (no need to point to code, just confirming!)

packages/turbo-codemod/src/commands/migrate/steps/getTurboUpgradeCommand.ts
tknickman: Ah good call, will fix!

packages/turbo-codemod/src/commands/migrate/index.ts
tknickman: Here's the output without print:
https://app.warp.dev/block/EDiMe0dRf4ChltYW9tm71a

And here's the same run, with print:
https://app.warp.dev/block/kfIHdgynjlfIPNR63MJiPj

mehulkar: awesome 👍🏾

PR #2706, opened 2022-11-15
Composable turbo.json

mehulkar:
This feature allows monorepos to place a `turbo.json` file in workspaces and add a `"extends"` key to extend from the root configuration to add/remove to the pipeline. After this PR, individual workspaces should be able to:

- add keys to task defintiions (e.g. changing `dependsOn` or `outputs` configs)
- reset keys in task definitions to defaults (`e.g. `dependsOn: []`)
- add tasks (e.g add a `lint` command only in the workspace that implements it)

<details>
<summary>Limitations</summary>

Some of these limitations are edge cases, but still listing them out. You cannot:

- extend config other than the `pipeline` tasks. For example, `remoteCache`, `globalEnv`, etc cannot be changed for individual workspaces.
- extend from other workspaces.
- configure the root `turbo.json` to extend from anything.
- add or subtract to existing array configs (e.g. if your root config has `dependsOn: ['something']`, you can't append to this array, you can only overwrite it.)
- declare tasks in the `package#task` syntax in workspaces. The `package` is inferred.

These limitations are based on a lot of research to keep the design space open, gather use cases, and avoid making the wrong decisions before putting the feature into the hands of real users. In short: to make it possible to ship this sooner!

</details>

<details>
<summary>Example</summary>

- `./turbo.json` (root)
	```json
	{
	"pipeline": {
			"build": {
				"outputs": ["dist/**"],
				"dependsOn": ["lint"]
			}
		}
	}
	```

- `./packages/ui/turbo.json` (`ui` workspace)

	```json
	{
		"extends": ["//"],
		"pipeline": {
			"build": {
				"dependsOn": ["lint", "compile"],
				"outputs": ["lib/**"]
			}
		}
	}
	```

With these configs, when you run `turbo build`, all workspaces except the `ui` workspace will cache the `dist/` directory and run their `lint` commands first. The `ui` workspace will run `lint`  and `compile` first, and cache the `lib/` directory instead.
</details>

<details>
<summary>Review Notes</summary>

- Implementation is not super robust, it just looks for a workspace config and root config and merges them. I'd like to get this out as feature-complete work, before adding any crawling behavior for future work.
- Read the test cases (`*.t` files) to see what use cases this supports
- Happy to pair-review anything.

</details>

<details>
<summary>Tests</summary>

- [x] When workspace adds a new _task_ that is not defined in the root config
- [x] When workspace adds new keys into task config that are not defined in the root
- [x] When workspace has a task definition, but omits some keys that root config has
[...]

PR #3368, opened 2023-01-18
Factor out ReadTurboConfig method

mehulkar:
We need this method to be able to read arbitrary turbo.json files and not fallback on package.json config or synthesizing one from the package.json.

Some work extracted from #2706.


cli/internal/fs/turbo_json.go
```diff
@@ -153,37 +160,20 @@ func (to TaskOutputs) Sort() TaskOutputs {
...
 
-	// If the configFile doesn't exist, but a legacy config does, log that it's deprecated and return an error
-	if hasLegacyConfig {
-		log.Printf("[DEPRECATED] \"turbo\" in package.json is deprecated. Migrate to %s by running \"npx @turbo/codemod create-turbo-config\"\n", configFile)
-	}
-
-	// If there's no turbo.json and no turbo key in package.json, return an error.
-	return nil, errors.Wrapf(os.ErrNotExist, "Could not find %s. Follow directions at https://turbo.build/repo/docs to create one", configFile)
+	// If there's no turbo.json, return an error.
+	return nil, errors.Wrapf(os.ErrNotExist, "Could not find %s", configFile)
```
chris-olszewski, line 176: Was there a reason to drop the link to the docs? It's causing the unit tests to fail.

chris-olszewski (approved):
Overall looks good to me as long as error message change was unintentional.

cli/internal/fs/turbo_json.go
mehulkar: The goal was to make it more general purpose since this method will be used for turbo.json's other than the root one. But let me double check we didn't lose the error message for the existing use case.

mehulkar: Great catch, it's back now.

PR #3372, opened 2023-01-18
feature(turborepo): Port login without sso team

NicholasLYang:


crates/turborepo-lib/src/client.rs
```diff
@@ -0,0 +1,126 @@
...
+            base_url: base_url.as_ref().to_string(),
+        })
+    }
+
+    fn make_url(&self, endpoint: &str) -> String {
+        format!("{}{}", self.base_url, endpoint)
+    }
+}
+
+fn user_agent() -> String {
```
arlyon, line 118: Could we perhaps make this static?

crates/turborepo-lib/src/ui/mod.rs
```diff
@@ -35,12 +37,44 @@ impl UI {
...
+
+        let mut out = Vec::new();
+        for (i, c) in text.char_indices() {
+            let (r, g, b) = Self::rainbow_rgb(i);
+            out.push(format!(
+                "\x1b[1m\x1b[38;2;{};{};{}m{}\x1b[0m\x1b[0;1m",
+                r, g, b, c
+            ));
+        }
+        out.push(RESET.to_string());
```
chris-olszewski, line 64: Not blocking, but we use [`tiny_gradient`](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/shim.rs#L484) in the shim for coloring. It would be nice to consolidate our gradient logic

crates/turborepo-lib/src/retry.rs
```diff
@@ -0,0 +1,46 @@
+use std::future::Future;
+
+use anyhow::anyhow;
+use tokio::time::sleep;
+
+const MIN_SLEEP_TIME: u64 = 2;
+const MAX_SLEEP_TIME: u64 = 10;
```
chris-olszewski, line 7: Just for better readability
```suggestion
const MIN_SLEEP_TIME_SECS: u64 = 2;
const MAX_SLEEP_TIME_SECS: u64 = 10;
```

crates/turborepo-lib/src/retry.rs
```diff
@@ -0,0 +1,46 @@
...
+
+        let sleep_period = (2_u64)
+            .pow(retry_count)
+            .clamp(MIN_SLEEP_TIME, MAX_SLEEP_TIME);
+        sleep(std::time::Duration::from_secs(sleep_period)).await;
+    }
+
+    Err(anyhow!(
+        "skipping HTTP Request, too many failures have occurred"
+    ))
```
chris-olszewski, line 45: Can we add the last error we hit to this message?

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -0,0 +1,106 @@
...
+    println!();
+    println!(
+        "{}",
+        ui.apply(
+            CYAN.apply_to("To connect to your Remote Cache, run the following in any turborepo:")
+        )
+    );
+    println!();
+    println!("{}", ui.apply(BOLD.apply_to("  npx turbo link")));
+    println!();
```
chris-olszewski, line 66: Can we consolidate these `println`s?

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -0,0 +1,106 @@
...
+    let repo_config = base.repo_config()?;
+    let login_url_base = repo_config.login_url();
+    debug!("turbo v{}", get_version());
+    debug!("api url: {}", repo_config.api_url());
+    debug!("login url: {login_url_base}");
+
+    let redirect_url = format!("http://{DEFAULT_HOST_NAME}:{DEFAULT_PORT}");
+    let login_url = format!("{login_url_base}/turborepo/token?redirect_uri={redirect_url}");
+    println!(">>> Opening browser to {login_url}");
+
```
chris-olszewski, line 30: I noticed we're not having a spinner displaying `Waiting for your authorization...` in the console anymore. Is there something preventing us from using [`indicatif`](https://docs.rs/indicatif/latest/indicatif/index.html) to keep this functionality? I've had good luck with the spinner and custom tick strings.

crates/turborepo-lib/src/client.rs
```diff
@@ -0,0 +1,127 @@
...
+use reqwest::StatusCode;
+use serde::Deserialize;
+
+use crate::{get_version, retry::retry_future};
+
+#[async_trait]
+pub trait UserClient {
+    fn set_token(&mut self, token: String);
+    async fn get_user(&self) -> Result<UserResponse>;
+}
```
chris-olszewski, line 15: I'm assuming this trait is here for testing, but I'm not seeing any use this. Will we need to refactor the `login` to use this trait before we can add tests?

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -0,0 +1,106 @@
...
+    config::{default_user_config_path, UserConfigLoader},
+    get_version,
+    ui::{BOLD, CYAN},
+};
+
+const DEFAULT_HOST_NAME: &str = "127.0.0.1";
+const DEFAULT_PORT: u16 = 9789;
+
+pub async fn login(base: CommandBase) -> Result<()> {
+    let repo_config = base.repo_config()?;
```
chris-olszewski, line 21: We need to add `--__test-run` otherwise the integration tests open up a browser and will hang for user input.

chris-olszewski (changes requested):
Overall looks good. The only blocker is the `--__test-run` support so our integration tests keep working. It would also be great if we could port the `Test_run` unit test to Rust so we don't drop test coverage.

crates/turborepo-lib/src/commands/login.rs
NicholasLYang: Ah good catch. Meant to add a spinner library

crates/turborepo-lib/src/ui/mod.rs
NicholasLYang: Yeah...I'm personally of the opinion that we should keep this because it has a fallback for Apple Terminal and has color stripping.

chris-olszewski (approved):

chris-olszewski:
Looks like failures are from a bad `Cargo.lock`?

PR #2712, opened 2022-11-15
Remove default task outputs when outputs is not defined

mehulkar:
This behavior is frameworks-specific and unintuitive to new users.
Specifically, this removes the confusion between omitting the outputs
key and setting to an empty array. Additionally, this will allow us to
more clearly define turbo.json composition, so locally-scoped turbo.json
configs can override outputs, without rationalizing what happens to this
implicit behavior.

Stacked PRs:

- [ ] **(This one)** #2712 
- [ ] #2714
- [ ] Release `turbo@1.7`
- [ ] #2713

PR #3377, opened 2023-01-19
Print resolvedTaskDefinition in DryRun

mehulkar:
This PR adds a `resolvedTaskDefinition` key to the Dry Run output. Today, this allows users to see the applied defaults for each key in their tasks. In the future, this will allow users to introspect the task config when it's composed from multiple sources.

- [x] write tests


cli/integration_tests/basic_monorepo/dry_run.t
```diff
@@ -0,0 +1,97 @@
...
+    Cached (Local)         = false                                                                                             
+    Cached (Remote)        = false                                                                                             
+    Directory              = apps/my-app                                                                                       
+    Command                = echo 'building'                                                                                   
+    Outputs                =                                                                                                   
+    Log File               = apps/my-app/.turbo/turbo-build.log                                                                
+    Dependencies           =                                                                                                   
+    Dependendents          =                                                                                                   
+    ResolvedTaskDefinition = {"outputs":[],"cache":true,"dependsOn":[],"inputs":[],"outputMode":0,"env":[],"persistent":false}
```
mehulkar, line 25: I thought about _not_ printing json here and trying have nested lines, but I think we want json, since it's trying to reflect user's input.

cli/integration_tests/basic_monorepo/dry_run.t
```diff
@@ -0,0 +1,97 @@
...
+    Directory              = packages/util                                                                                     
+    Command                = echo 'building'                                                                                   
+    Outputs                =                                                                                                   
+    Log File               = packages/util/.turbo/turbo-build.log                                                              
+    Dependencies           =                                                                                                   
+    Dependendents          =                                                                                                   
+    ResolvedTaskDefinition = {"outputs":[],"cache":true,"dependsOn":[],"inputs":[],"outputMode":0,"env":[],"persistent":false} 
+
+# Validate output of my-app#build task
+  $ ${TURBO} run build --dry=json | jq '.tasks' | jq 'map(select(.taskId == "my-app#build"))' | jq '.[0]'
```
mehulkar, line 42: there might be a better way to do this `jq` filter

cli/internal/fs/turbo_json.go
```diff
@@ -48,14 +48,13 @@ type RemoteCacheOptions struct {
...
 type rawTask struct {
-	Outputs *[]string `json:"outputs,omitempty"`
-
-	Cache      *bool               `json:"cache,omitempty"`
-	DependsOn  []string            `json:"dependsOn,omitempty"`
-	Inputs     []string            `json:"inputs,omitempty"`
-	OutputMode util.TaskOutputMode `json:"outputMode,omitempty"`
-	Env        []string            `json:"env,omitempty"`
-	Persistent bool                `json:"persistent,omitempty"`
+	Outputs    []string            `json:"outputs"`
```
mehulkar, line 51: changed this from a pointer, because I had trouble using it as a pointer in the new code, and I'm not sure if it needs to be one. I can change it back if it breaks something

cli/internal/run/dry_run.go
```diff
@@ -282,30 +289,32 @@ func commandLooksLikeTurbo(command string) bool {
...
+	Hash                   string             `json:"hash"`
+	CacheState             cache.ItemStatus   `json:"cacheState"`
+	Command                string             `json:"command"`
+	Outputs                []string           `json:"outputs"`
+	ExcludedOutputs        []string           `json:"excludedOutputs"`
+	LogFile                string             `json:"logFile"`
+	Dir                    string             `json:"directory"`
+	Dependencies           []string           `json:"dependencies"`
+	Dependents             []string           `json:"dependents"`
+	ResolvedTaskDefinition *fs.TaskDefinition `json:"resolvedTaskDefinition"`
```
mehulkar, line 304: Need to bikeshed the `ResolvedTaskDefinition` key name a bit. 

My main concern is that "resolved" is misleading because we aren't going to include things like turbo's injected env vars or expand globs in inputs/outputs. I think that is ok, since this is a "task definition", and that "fully resolved state" is more for "cache hit status".

cli/internal/run/dry_run.go
```diff
@@ -282,30 +289,32 @@ func commandLooksLikeTurbo(command string) bool {
...
-	Dependencies    []string         `json:"dependencies"`
-	Dependents      []string         `json:"dependents"`
+	TaskID                 string             `json:"taskId"`
+	Task                   string             `json:"task"`
+	Package                string             `json:"package"`
+	Hash                   string             `json:"hash"`
+	CacheState             cache.ItemStatus   `json:"cacheState"`
+	Command                string             `json:"command"`
+	Outputs                []string           `json:"outputs"`
+	ExcludedOutputs        []string           `json:"excludedOutputs"`
```
mehulkar, line 299: `Outputs` and `ExcludedOutputs` are redundant now, and I would like to remove them at some point in the future.

cli/integration_tests/basic_monorepo/dry_run.t
chris-olszewski, line 42:
```suggestion
  $ ${TURBO} run build --dry=json | jq '.tasks | map(select(.taskId == "my-app#build")) | .[0]'
```
You could also use `.tasks | .[] | select(.taskId == ...)` if you want to omit the `.[0]` at the end, but that would just be a style choice.

cli/internal/fs/turbo_json.go
chris-olszewski: A comment from #2712 that got deleted seems to indicating this was a pointer so it would get unmarshaled as `nil` instead of `[]string{}`:
```
// We can't use omitempty for Outputs, because it will
// always unmarshal into an empty array, which means something different from nil.
```
I think we don't need to differentiate between the two now that we don't supply default outputs.

cli/internal/run/dry_run.go
```diff
@@ -142,18 +143,19 @@ func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.Complete
...
-			Dependents:      stringDescendents,
+			TaskID:                 packageTask.TaskID,
+			Task:                   packageTask.Task,
+			Package:                packageTask.PackageName,
+			Hash:                   hash,
+			CacheState:             itemStatus,
+			Command:                command,
+			Dir:                    packageTask.Pkg.Dir.ToString(),
+			Outputs:                packageTask.TaskDefinition.Outputs.Inclusions, // TODO: remove this?
+			ExcludedOutputs:        packageTask.TaskDefinition.Outputs.Exclusions, // TODO: remove this?
```
chris-olszewski, line 154: Maybe add some more context around why these are redundant and can be removed in the future if we're gonna leave the `TODO` here.

cli/internal/fs/turbo_json.go
```diff
@@ -301,6 +300,46 @@ func (c *TaskDefinition) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// MarshalJSON deserializes JSON into a TaskDefinition
+func (c *TaskDefinition) MarshalJSON() ([]byte, error) {
```
chris-olszewski, line 304: Quick glancing through the code I think we aren't altering `TaskDefinition`s after construction, but since non-determinism has gotten us before: we might want to sort these slices before marshalling them. If you've done some closer reading of the code to verify that the slices remain sorted after creation, then feel free to ignore this comment.

cli/internal/run/dry_run.go
chris-olszewski: IMO I think the name pretty clear

chris-olszewski (approved):

cli/internal/fs/turbo_json.go
mehulkar: I had understood that comment to be about `omitempty`, rather than the pointer (which is why i deleted the comment). Also that comment used to be in the Unmarshal method, rather than where the pointer was defined, so that seems to support the theory
https://github.com/vercel/turbo/blob/v1.6.0/cli/internal/fs/turbo_json.go#L215-L219

cli/internal/fs/turbo_json.go
mehulkar: Good call. That seems like a good idea, even just for determinism in tests, but I don't _think_ it should affect task hashes, right? 🤔

mehulkar: Wrote a test first and it seemed to pass deterministically, but added a sort in anyway

cli/internal/fs/turbo_json.go
nathanhammond: This is all correct, and a good change. Using weird Go implementation details to hack serialization is a smell, and I'm happy to have it gone.

cli/internal/fs/turbo_json.go
nathanhammond: I wonder how visible this change will be. I _think_ it effectively changes the execution order of peer tasks?

None of that should be relied upon (and we should probably intentionally insert randomness into that at execution time to prevent people relying on it) but that doesn't mean people aren't implicitly relying on it right now.

cli/internal/fs/turbo_json.go
```diff
@@ -301,6 +300,54 @@ func (c *TaskDefinition) UnmarshalJSON(data []byte) error {
...
+	// Initialize with empty arrays, so we get empty arrays serialized into JSON
```
nathanhammond, line 305: For elegance I _kinda_ want to omit empty arrays when printing, but that may make it harder for people to use `--dry=json`. So I'm not actually proposing this.

cli/internal/fs/turbo_json.go
mehulkar: Based on my experience with test runs, the order is indeterminate already (see https://github.com/Nicoretti/prysk/issues/154, for example), so I don't think the sort is guaranteeing anything.

Also, for future readers, to be clear, `Marshal` doesn't play a part in taskGraph exectution at all. I'm guessing Nathan is commenting on sorting lists in general (and in this case the Unmarshal part)

cli/internal/fs/turbo_json.go
mehulkar: Yeah, IMO it's actually important to print the empty arrays, so it's never a question about what the whole config is. In other words, missing key can mean "we don't have this config" or "this config has an empty value", and that makes debugging a bit harder.

PR #3378, opened 2023-01-19
Factor out method to get resolved TaskDefinition

mehulkar:
The way we resolve a taskDefinition will change
once we start allowing workspaces to create their own config so it makes sense to factor this method out first.

### DIFF
--- a/cli/internal/graph/graph.go
+++ b/cli/internal/graph/graph.go
@@ -44,28 +44,45 @@ func (g *CompleteGraph) GetPackageTaskVisitor(ctx gocontext.Context, visitor fun
 			return fmt.Errorf("cannot find package %v for task %v", packageName, taskID)
 		}
 
-		// first check for package-tasks
-		taskDefinition, ok := g.Pipeline[fmt.Sprintf("%v", taskID)]
-
-		if !ok {
-			// then check for regular tasks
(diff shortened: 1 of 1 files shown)


cli/internal/graph/graph.go
```diff
@@ -44,28 +44,45 @@ func (g *CompleteGraph) GetPackageTaskVisitor(ctx gocontext.Context, visitor fun
...
+
 		return visitor(ctx, packageTask)
 	}
 }
+
+// getResolvedTaskDefinition currently just looks for the definition in the Pipeline
+// defined in the Graph. Later, this will get Pipelines defined in the task's workspace as well.
+func (g *CompleteGraph) getResolvedTaskDefinition(taskID string, taskName string) (*fs.TaskDefinition, error) {
+	return getTaskFromPipeline(g.Pipeline, taskID, taskName)
+}
```
mehulkar, line 69: This is where we'll implement the magic in #2706

nathanhammond (approved):
I'm not convinced we won't want to pre-resolve everything before getting here, but this doesn't materially change anything and seems fine.

mehulkar:
> I'm not convinced we won't want to pre-resolve everything before getting here

Let's chat about that, considering the other conversation we had today about pre-calculated task hashes as well. Want to make sure I'm not overlooking anything.

PR #3380, opened 2023-01-19
Target via synthesized task

nathanhammond:
What if we implemented targets by doing it as a synthesized volatile task which inserts itself as a dependency in the correct places?


gsoltis:
Can we implement this without putting it into the UI as a task (graph, dry run, etc.) and without needing to add the concept of volatile tasks?

PR #3387, opened 2023-01-19
fix: update cargo lockfile

chris-olszewski:
`Cargo.lock` on main appears to be [out of date](https://github.com/vercel/turbo/actions/runs/3963061842/jobs/6790483844#step:9:276), these are the results of `cargo check`


kwonoj (approved):

kwonoj:
I think we need to update CI to use frozen lockfile?

kwonoj:
https://github.com/vercel/turbo/pull/3388

PR #3388, opened 2023-01-19
ci(actions): ensure cargo check with frozen lockfile

kwonoj:
Avoid cases like https://github.com/vercel/turbo/pull/3387


chris-olszewski (approved):

sokra:
@kwonoj I had to revert that. I think that's a good idea, but I broke CI. Not even this PR passed the CI.

Maybe you want to use `--locked` instead of `--frozen`

kwonoj:
@sokra weird, --frozen is alias to --locked.

https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-option

Also PR correctly passed ci. Ref: https://github.com/vercel/turbo/actions/runs/3963353615/jobs/6792718578

kwonoj:
Looks like `The --frozen flag also prevents Cargo from attempting to access the network to determine if it is out-of-date.` is related, but --locked supposed to behave same. In any case, fine to revert and dig later.

sokra:
It seem to complain about not being about to make network requests. I guess it need to make requests when it need to download the crates from the registry. Maybe it only passes with rust cache, but fails on downloading when there is no cache or crates are missing...

PR #3389, opened 2023-01-20
Minimal next-build implementation

Brooooooklyn:
We want to implement the most basic way of running `next build` with turbopack. For that, we use the "dumb bundler" approach and run the usual next build steps in next.js with one exception: Instead of running webpack, we invoke turbopack via napi.
So on the turbopack side, we are only responsible for bundling and emitting the bundled files (for all pages) to disk into the `_next` folder, including the manifest that next.js expect from us.
For the start, we don't want to run any optimization at all. We just emit the files as we would serve them for development (maybe disable HMR if that leads to issues). We already have the logic for emitting assets to disk (e. g. for node.js execution or node-modules-trace), and we can use that. But we need to come up with additional code for generating the manifests.
No node.js rendering has to be done in turbopack. This will be handled by next.js.
There are a few optional things we could do that might help in testing and developing:
Create a next-build CLI in turbopack.
Add a flag/subcommand to next.js to only run static generation with the existing `_next` directory.


Brooooooklyn:
Close in https://github.com/vercel/next.js/pull/46602

PR #3396, opened 2023-01-20
add edge compilation, runtime config parsing and edge apis

sokra:


crates/next-core/src/util.rs
```diff
@@ -66,3 +73,155 @@ pub async fn foreign_code_context_condition(next_config: NextConfigVc) -> Result
...
+    fn detail(&self) -> StringVc {
+        self.detail
+    }
+}
+
+#[turbo_tasks::function]
+pub async fn parse_config_from_source(module_asset: AssetVc) -> Result<NextSourceConfigVc> {
+    if let Some(ecmascript_asset) = EcmascriptModuleAssetVc::resolve_from(module_asset).await? {
+        if let ParseResult::Ok { program, .. } = &*ecmascript_asset.parse().await? {
+            if let Program::Module(module) = &program {
```
jridgewell, line 138: Are scripts supported by Next?

crates/next-core/src/util.rs
```diff
@@ -66,3 +73,155 @@ pub async fn foreign_code_context_condition(next_config: NextConfigVc) -> Result
...
+                for item in &module.body {
+                    if let Some(decl) = item
+                        .as_module_decl()
+                        .and_then(|mod_decl| mod_decl.as_export_decl())
+                        .and_then(|export_decl| export_decl.decl.as_var())
+                    {
+                        for decl in &decl.decls {
```
jridgewell, line 145: I had commented before in slack that we could write a generic AST evaluator, and this might allow us to parse the `next.config.js` file without actually spinning up node. We have a similar desire here, and we could make this more generic by using a function that takes the export were interested in, and returning a `Result<JsonValue>` of the statically analyzed value.

jridgewell (approved):

crates/next-core/src/util.rs
sokra: When it has an export it's not a script

crates/next-core/src/util.rs
sokra: Technically we have such an evaluator for the static analysis. Maybe we can use that here instead.

crates/next-core/src/util.rs
jridgewell: I understand, but that doesn't answer the question. Does next support `exports.handler = function () {}` CJS edge functions? Are devs forced to use ESM? Do we need to add a Script branch to this function?

sokra: Afaik we only support config in that format:

```
export const config = {
  runtime: "edge"
};
```

sokra: https://nextjs.org/docs/advanced-features/react-18/switchable-runtime#edge-api-routes

sokra: next.js also looks for export declarations: https://github.com/vercel/next.js/blob/8f05c0c07e0796b6e4e8c3610a49e3828d1ee8ef/packages/next-swc/crates/core/src/page_config.rs#L74

crates/next-core/src/util.rs
sokra: did that

jridgewell (approved):

crates/turbopack-node/src/lib.rs
```diff
@@ -200,9 +201,9 @@ pub async fn get_renderer_pool(
     )
     .await?;
 
-    emit(intermediate_asset, intermediate_output_path).await?;
+    emit(intermediate_asset, output_root).await?;
 
-    let cwd = intermediate_output_path.root();
+    let cwd = output_root;
```
ForsakenHarmony, line 206: This is completely broken, though to be fair it was partially broken before (workspaces wouldn't have worked)

PR #3179, opened 2023-01-05
switch to husky, fix prettier

ForsakenHarmony:

PR #3403, opened 2023-01-20
don't run `cargo fmt --all` in pre-commit hook

wbinnssmith:
This action started failing because `cargo fmt` was added to the pre-commit hook and the tool is not available without running Rust setup. Instead, we should just skip applying hooks when creating commits from this workflow.

~Since the create-pull-request doesn't have a way of disabling hooks, simply remove the file before husky installs it into place.~ Let `cargo fmt` run on individual files through lint-staged, but don't run it on everything on every precommit.

Test Plan: Manual workflow run on this branch which opened #3401

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,4 +2,3 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 pnpm exec lint-staged
-cargo fmt --all


mehulkar (commented):
is it possible to only run `cargo fmt` when relevant rust files change? this looks like it's not going to touch any of those files, so the pre-commit hook shouldn't apply. (I'm not sure if lint-staged is already supposed to be doing that?)

wbinnssmith:
> is it possible to only run `cargo fmt` when relevant rust files change? this looks like it's not going to touch any of those files, so the pre-commit hook shouldn't apply. (I'm not sure if lint-staged is already supposed to be doing that?)

it actually does already: https://github.com/vercel/turbo/blob/a3bfa40ccd67862b3640731d06a2529a455fd2e5/package.json#L51

...but `cargo fmt --all` was added to the hook in #3179: https://github.com/vercel/turbo/commit/daaeabfb502d3b1f6c1acfae31d0d2cf8a740463#diff-d2bc4bbf14eadc292d84e0ef5f7a93115c23557f533809fc80b896934291529dR5

@ForsakenHarmony was there a reason to format everything on every change? `cargo fmt` is probably fast enough to do so, but it requires the rust toolchain in situations like these.

ForsakenHarmony:
@wbinnssmith we can remove it

ForsakenHarmony:
Why do we have the google font data in our repo though, I thought we wanted to use it from the next package directly

wbinnssmith:
> Why do we have the google font data in our repo though, I thought we wanted to use it from the next package directly (`@next/font/dist/google/font-data.json`)

Yeah, we do. We just haven't gotten to it yet, and this was breaking.

ForsakenHarmony (approved):

PR #3408, opened 2023-01-20
feat(codemod): use transform as default command

tknickman:
Ensures codemod backwards compatibility


mehulkar (approved):
Want to note this in the README also?

tknickman:
Good call @mehulkar, will do.

PR #3410, opened 2023-01-20
Refactor task dependencies retrieval to support composable configs

mehulkar:
This means we look up a task's dependencies when iterating per-workspace, rather than per-task. This will allow us to customize each task per workspace based on additional config that may be present in that workspace.

After this change, `core.Task` is essentually a tuple that contains the Task name and the Task Definition. It's the perfect
place to implement #2706, since it's a layer _above_ the underlying TaskDefinition, so we can create the construct however we want to.

Follow up to #3405


cli/internal/core/engine.go
```diff
@@ -161,10 +159,29 @@ func (e *Engine) generateTaskGraph(pkgs []string, taskNames []string, tasksOnly
...
+
+		for _, dependency := range task.TaskDefinition.TaskDependencies {
+			// If the current task is a workspace-specific task (including root Task)
+			// and its dependency is _also_ a workspace-specific task, we need to add
+			// a reference to this dependency directly into the engine.
+			// TODO @mehulkar: Why do we need this?
+			if isPackageTask && util.IsPackageTask(dependency) {
+				if err := e.AddDep(dependency, taskName); err != nil {
+					return err
+				}
```
mehulkar, line 173: This part is kind of strange, hopefully we can make some improvements so that packageTasks don't have a separate code path some time in the future. I'm not sure they need to be tracked separately. (Especially now that we're doing this work per-workspace.

chris-olszewski (approved):

PR #3411, opened 2023-01-20
Rename TopologicGraph to WorkspaceGraph

mehulkar:
TopologicGraph in the engine sounds like it's for tasks, since the engine is for executing tasks, but it's really the workspace graph.

### DIFF
--- a/cli/internal/context/context.go
+++ b/cli/internal/context/context.go
@@ -210,7 +210,7 @@ func BuildPackageGraph(repoRoot turbopath.AbsoluteSystemPath, rootPackageJSON *f
 	for _, pkg := range c.WorkspaceInfos {
 		pkg := pkg
 		populateGraphWaitGroup.Go(func() error {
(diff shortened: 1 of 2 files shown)


NicholasLYang:
Just to be clear, topological in this context means topologically sorted? Seems like there's other variables in this context using that term and I'm not really sure what it implies. Otherwise LGTM

NicholasLYang:
Hmm yeah maybe this is me overthinking this but topological is a really general word. I'd be in favor of removing it entirely unless we're talking about topological sorting.

NicholasLYang (approved):

tknickman (approved):
Can you also rename some of the functions in `context.go` (like `populateTopologicGraphForPackageJSON` for example) - I think that's the last place we refer to TopologicalGraph

mehulkar:
@tknickman great call, updated!

PR #3428, opened 2023-01-23
Store TaskDefinitions in the Graph by Task ID

mehulkar:
Moves TaskDefinitions in the `CompleteGraph` struct, so we can look it up during execution.
We want to do this here, because _during_ execution, The TaskGraph has already been constructed,
so it is too late to change the definition at the point. Although this commit does not try to
change any TaskDefinitions, we will implement that for composable configs in the future.

- [x] #3410
- [x] #3405


cli/internal/graph/graph.go
```diff
@@ -38,51 +41,21 @@ type CompleteGraph struct {
...
-	// first check for package-tasks
-	taskDefinition, ok := pipeline[taskID]
-	if !ok {
-		// then check for regular tasks
-		fallbackTaskDefinition, notcool := pipeline[taskName]
-		// if neither, then bail
-		if !notcool {
-			// Return an empty fs.TaskDefinition
-			return nil, fmt.Errorf("No task defined in pipeline")
+		taskDefinition, ok := g.TaskDefinitions[taskID]
```
mehulkar, line 48: Instead of looking for a taskDefinition in the Pipeline, we now look for it in the `TaskDefinitions` map. The TaskDefinitions in this map will later become composed from multiple configs.

cli/internal/run/run.go
```diff
@@ -388,7 +389,7 @@ func (r *run) initCache(ctx gocontext.Context, rs *runSpec, analyticsClient anal
 }
 
 func buildTaskGraphEngine(g *graph.CompleteGraph, rs *runSpec) (*core.Engine, error) {
-	engine := core.NewEngine(&g.WorkspaceGraph)
+	engine := core.NewEngine(g)
```
mehulkar, line 392: This is not great, and it really blurs the line between `Engine` and `Graph`. This line already seems pretty blurry to me, so I made the call to pass the whole thing in, rather than just TaskDefinitions, Pipeline, etc. There's some room for improvement with this architecture after Composable Turbo is done.

cli/internal/core/engine.go
```diff
@@ -33,13 +33,17 @@ type Engine struct {
...
 }
 
 // NewEngine creates a new engine given a topologic graph of workspace package names
-func NewEngine(topologicalGraph *dag.AcyclicGraph) *Engine {
+func NewEngine(completeGraph *graph.CompleteGraph) *Engine {
 	return &Engine{
+		completeGraph:    completeGraph,
 		Tasks:            make(map[string]*Task),
-		TopologicGraph:   topologicalGraph,
+		TopologicGraph:   &completeGraph.WorkspaceGraph,
```
mehulkar, line 46: This field can be removed and we could access from `completeGraph` directly, but I don't want to go that far in this PR. (See also #3411 which renames this).

cli/internal/core/engine_test.go
```diff
@@ -107,29 +121,42 @@ func TestDependenciesOnUnspecifiedPackages(t *testing.T) {
...
+		Pipeline:        pipeline,
+		TaskDefinitions: map[string]*fs.TaskDefinition{},
+		WorkspaceInfos: graph.WorkspaceInfos{
+			"app1": &fs.PackageJSON{},
+			"app2": &fs.PackageJSON{},
+			"libA": &fs.PackageJSON{},
+			"libB": &fs.PackageJSON{},
+			"libC": &fs.PackageJSON{},
+			"libD": &fs.PackageJSON{},
+		},
```
mehulkar, line 154: All these engine tests are really coupled to the implementation, so they have to change when things change. I would like to convert a lot of them to integration tests with prysk, so we can validate what we really need to validate.

cli/internal/core/engine.go
```diff
@@ -142,10 +148,36 @@ func (e *Engine) generateTaskGraph(pkgs []string, taskNames []string, tasksOnly
...
 		pkg, taskName := util.GetPackageTaskFromId(taskID)
+
+		// This for lopo adds in the Root Node as a taskID (i.e. __ROOT__#build)
+		// into the traversalQueue. But in order to look up the information about
+		// the root workpsace, we need to use the package name "//". Re-assign
+		// the pkgName here to account for that.
+		// TODO(mehulkar): Is it safe to move this logic into GetPackageTaskFromId()?
+		if pkg == ROOT_NODE_NAME {
+			pkg = util.RootPkgName
+		}
```
mehulkar, line 159: I'm not sure if this check should instead be to _skip_ the __ROOT_ node instead? All tests are passing, so it doesn't seem like this is a problem, but in other places, we have checks to skip this root node.

cli/internal/taskhash/taskhash.go
```diff
@@ -174,6 +174,8 @@ func (th *Tracker) CalculateFileHashes(allTasks []dag.Vertex, workerCount int, r
 			continue
 		}
 
+		// TODO(mehulkar): Once we start composing turbo.json, we need to change this
+		// to look in the graph for TaskDefinitions, rather than the root pipeline.
```
mehulkar, line 178: I have another ticket for this. It shouldn't break anything since there is only one turbo.json so far.

chris-olszewski (approved):
Looks good to me, glad to see more tests!

cli/integration_tests/topological_deps/run.t
```diff
@@ -0,0 +1,26 @@
+Setup
+  $ . ${TESTDIR}/../setup.sh
+  $ . ${TESTDIR}/setup.sh $(pwd)
+
+Check my-app#build output
```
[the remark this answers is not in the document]
mehulkar: There was a step along the way where it was just plain broken (in this branch), and I saw there were no tests for `dependsOn: [^build]`. So my intention wasn't to check that the tasks were run in order, but that they ran at all! Running in topological order is  also tested here though, so I guess so 😄

cli/internal/core/engine.go
mehulkar: I _think_ it would be hit some times. there's a difference between a `//#build` and `__ROOT__#build` task. The first is user-configured, the second is some virtual node that we insert into the graph (and I can't remember why). I think this rewrite might actually be problematic because the next condition will be exercised for the virtual node also. 

I'll spend some more time understanding this / adding a test case in.

cli/internal/core/engine.go
mehulkar: just so I know / and to learn Go, what could cause a data race?

cli/internal/core/engine.go
mehulkar: I was actually able to remove this piece of code. It was left over from doing actual composition in #2706, but in this refactor it's not necessary, so I can delay doing this work.

PR #3435, opened 2023-01-23
Fix data requests for dynamic routes

alexkirsz:
In addition to the fix, I've extracted the path regex/params matching logic from `turbopack-node` to `next-core`. `turbopack-node` now only declares a value trait which `next-core` implements.


crates/next-core/src/page_source.rs
```diff
@@ -301,11 +306,11 @@ async fn create_page_source_for_file(
...
-        let data_pathname = StringVc::cell(format!(
-            "_next/data/development/{}",
-            get_asset_path_from_route(&pathname.await?, ".json")
-        ));
-        let data_path_regex = regular_expression_for_path(data_pathname);
+        let data_params_matcher = NextParamsMatcherVc::new(
+            pathname,
+            OptionStringVc::cell(Some("_next/data/development/".to_string())),
+            OptionStringVc::cell(Some(".json".to_string())),
+        );
```
alexkirsz, line 313: This is part 1 of the fix: ensuring that the params matcher ignores both the prefix and suffix of data requests.

crates/next-core/src/page_source.rs
```diff
@@ -332,16 +337,16 @@ async fn create_page_source_for_file(
...
                 ssr_entry,
                 runtime_entries,
                 fallback_page,
             ),
             create_node_rendered_source(
                 specificity,
                 server_root,
-                data_pathname,
-                data_path_regex,
+                pathname,
```
alexkirsz, line 348: This is part 2: Next.js has an internal Regex to detect dynamic routes, which won't match `/[dynamic].json` but will match `/[dynamic]` and `/[dynamic]/`.

alexkirsz: This is what Next.js does: the pathname is the original URL pathname (e.g. `/[test]`), and not the data pathname (`/_next/data/development/thing.json`).

alexkirsz:
* main
  * **PR #3435** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3435" target="_blank">Graphite</a>  👈
    * **PR #3448** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3448" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3435?utm_source=stack-comment).

crates/next-core/src/params_matcher/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+#[turbo_tasks::value_impl]
+impl NextParamsMatcherVc {
+    /// Converts a filename within the server root into a regular expression
+    /// with named capture groups for every dynamic segment.
+    #[turbo_tasks::function]
+    pub async fn new(
+        pathname: StringVc,
+        prefix: OptionStringVc,
+        suffix: OptionStringVc,
```
sokra, line 28: `Option<StringVc>` is also supported as argument to turbo functions, so you could use that

sokra (approved):

Issue #3419, opened 2023-01-21
[turborepo] with-nextjs example - outdated next.config.js


bibblebabl:
### What version of Turborepo are you using?

1.7.0

### What package manager are you using / does the bug impact?

Yarn v1

### What operating system are you using?

Mac

### Describe the Bug

Running `yarn build` in the repo we are getting following output:

```sh
➜  with-docker git:(main) yarn build
yarn run v1.22.19
$ turbo run build
• Packages in scope: api, eslint-config-custom, eslint-config-custom-server, jest-presets, logger, tsconfig, ui, web
• Running build in 8 packages
• Remote caching disabled
web:build: cache miss, executing b34dec01510f1d21
logger:build: cache hit, replaying output 4e5a1d77328cdb6f
logger:build: $ tsc
api:build: Skipping cache check for api#build, outputs have not changed since previous run.
api:build: cache hit, replaying output 6ca19da0ceb56e5f
api:build: $ tsc
web:build: $ next build
web:build: warn  - Invalid next.config.js options detected:
web:build:   - The root value has an unexpected property, transpilePackages, which is not in the list of allowed properties (amp, analyticsId, assetPrefix, basePath, cleanDistDir, compiler, compress, crossOrigin, devIndicators, distDir, env, eslint, excludeDefaultMomentLocales, experimental, exportPathMap, generateBuildId, generateEtags, headers, httpAgentOptions, i18n, images, onDemandEntries, optimizeFonts, output, outputFileTracing, pageExtensions, poweredByHeader, productionBrowserSourceMaps, publicRuntimeConfig, reactStrictMode, redirects, rewrites, sassOptions, serverRuntimeConfig, staticPageGenerationTimeout, swcMinify, trailingSlash, typescript, useFileSystemPublicRoutes, webpack).
web:build:
web:build: See more info here: https://nextjs.org/docs/messages/invalid-next-config
web:build: warn  - You have enabled experimental feature (outputFileTracingRoot) in next.config.js.
web:build: warn  - Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use at your own risk.
web:build:
web:build: info  - Linting and checking validity of types...
web:build: info  - Creating an optimized production build...
web:build: Failed to compile.
web:build:
web:build: ../../packages/ui/Button.tsx
web:build: Module parse failed: The keyword 'interface' is reserved (3:0)
[...]

PR #3436, opened 2023-01-23
chore(examples): update with-docker lock

tknickman:
Fixes https://github.com/vercel/turbo/issues/3419

Updates Next.js to latest so we get support for experimental `transpilePackages`


arlyon:
transpilePackages is great!

arlyon (approved):

tunguyen-ct:
Hi @tknickman , can you make this example work with PNPM Workspace? That would be very helpful.

PR #3437, opened 2023-01-23
css modules: use a terser selector format that requires less escaping

wbinnssmith:
Previously, generated classnames would include literal `/` characters, which need escaping when generating selectors. While users should still escape selectors, this avoids it for this case and makes generated code easier to read.


crates/turbopack-tests/Cargo.toml
```diff
@@ -31,3 +31,8 @@ turbopack-env = { path = "../turbopack-env" }
 
 [build-dependencies]
 turbo-tasks-build = { path = "../turbo-tasks-build" }
+
+[features]
```
wbinnssmith, line 35: Added this section so `cargo test -p turbopack-tests --features native-tls` is possible. cc @kwonoj

jridgewell (approved):

ForsakenHarmony (approved):
next uses `filename_class__hash`, haven't checked `node_modules`

sokra (changes requested):
I don't like this change. Before it was obvious from a class name from which source file it comes from. Now this is no longer the case.

I don't see the benefit of reading generated code, since the user will never read it anyway. It's source mapped and links to the source will point to the original css. Instead they might read the html or Dom where escaping is not required.

ForsakenHarmony:
@sokra readability of the classes in the HTML tree is terrible when there are multiple classes applied to one element

sokra (approved):

Issue #219, opened 2021-12-12
Add `--grouped` flag to group log output 


jaredpalmer:
### Describe the feature you'd like to request

> Output of steps are intermingled and can be a little hard to read

<img width="618" alt="CleanShot 2021-12-12 at 18 03 32@2x" src="https://user-images.githubusercontent.com/4060187/145733177-98d75d05-49d9-4fdd-9362-a9c808493a78.png">

https://twitter.com/yangshunz/status/1470019087669489665

### Describe the solution you'd like

- [ ] Add a `--grouped` flag that will not stream, but buffer and print all logs of a task once in a single group after a task is completed. This should be the default log behavior in non-interactive terminal processes. Our [`logstreamer`](https://github.com/vercel/turborepo/tree/main/cli/internal/logstreamer) abstraction already has an option for this, but need to make sure that prints to stdout/stderr are not intermingled. Ideally, writes should print First In, First Out, so a (new) queue/channel might be needed inside of [`run_state.go`](https://github.com/vercel/turborepo/blob/main/cli/internal/run/run_state.go#L85).
- [ ] Update documentation

### Describe alternatives you've considered

- `--buffered` flag name

marklawlor:
Could this be enabled selectively?

For example, I may wish to only group the `build` tasks but stream the `dev` ones? That way if `dev` depends on `build`, it will initially group the `build` logs but then switch to streaming for the `dev` tasks.

meehawk:
@jaredpalmer is this taken up? If not I'd love to take this up. Any guidance would be awesome

szk2s:
Would love this. Our CI log is almost useless.

georeith:
+1 could really use this for test output

ruettenm:
We have the same problem in a bigger project (turbo.json n around 300 lines and 25 package.json in the mono repo). It's always a bit hard to find something in the log :-)

This feature would be really helpful for us ❤️

robaca:
It also would be cool, if in that case, an additional parameter would toggle turbo to output GitHub Workflow Group statements for every task, maybe even with a nice `title`.

```
::group::{task name}
::endgroup::
```

(see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines)

mehulkar:
Couple naming options that have been floated internally:

- `--log-stream: natural | grouped`
- `--log-order: stream | grouped`

[...]

PR #3439, opened 2023-01-23
Add two log output modes, `stdout-full` and `stdout-new-only`

### LINKED ISSUE #901: Do not prefix log output

1.1.7-canary.5

npm

Mac

### Describe the Bug

Screenshot 2022-03-18 at 15 37 50

All the task log lines are prefixed with the scope, causing the logs to be hard to read.

### Expected Behavior

Please do not prefix every log line.

### To Reproduce

turbo run build --scope=foo

rafaeltab:
## Add two log output modes, `stdout-full` and `stdout-new-only`
The new `stdout-full` mode will not prefix the output generated by the commands run by Turborepo. It will still prefix turbos messages, such as cache hit, cache miss, and cache bypass. This allows the user to identify what was and wasn't cached while preserving the original output. 

The new `stdout-new-only` mode combines the behavior of `new-only` and `stdout-full` by only logging non-cached tasks, without prefixing the output.

### stdout-full 
Command: `turbo run build --output-logs=stdout-full --filter=util --force`
Output:
```
\xe2\x80\xa2 Packages in scope: util
\xe2\x80\xa2 Running build in 1 packages
\xe2\x80\xa2 Remote caching disabled
util:build: cache bypass, force executing 6dec18f9f767112f
  
> build 
> echo 'building'
  
building
  
 Tasks:    1 successful, 1 total
Cached:    0 cached, 1 total
  Time:    354ms
```

### stdout-new-only
Command: `turbo run build --output-logs=stdout-new-only --filter=util`
Output:
```
\xe2\x80\xa2 Packages in scope: util
\xe2\x80\xa2 Running build in 1 packages
\xe2\x80\xa2 Remote caching disabled
util:build: cache hit, suppressing output 6dec18f9f767112f

 Tasks:    1 successful, 1 total
Cached:    1 cached, 1 total
  Time:    5ms >>> FULL TURBO
```

Command: `turbo run build --output-logs=stdout-new-only --filter=util --force`
Output:
```
\xe2\x80\xa2 Packages in scope: util
\xe2\x80\xa2 Running build in 1 packages
\xe2\x80\xa2 Remote caching disabled
util:build: cache bypass, force executing 6dec18f9f767112f
  
> build
> echo 'building'
  
building
  
 Tasks:    1 successful, 1 total
Cached:    0 cached, 1 total
  Time:    243ms
```

## Side notes
I also added some integration tests for the `output-logs` option

Closes #901


mehulkar (commented):
Thanks for the contribution @rafaeltab! The implementation and tests look great, but I don't love the user-facing API.

1. "stdout" is inaccurate, as it will stream `stderr` also
2. Adding `stdout-*` options means that we need 2n options for each "filter view" of logs.

What do you think instead about adding a new flag `--pretty` and making it on by default? Then, `--output-logs=stdout-full` would be the same as `--output=logs=full`, etc. Some advantages of this:

- we don't need to capture _what_ turbo means by pretty output, we can do prefixes, colors, or something else in the future
- following `git log --pretty`, in the future we could enable something like `--pretty='%some-micro-syntax-for-fine-grain-control%'`

I'm not sold on this idea, since adding a new flag adds some overhead, but I hope this gives us something to talk about!

rafaeltab:
I agree, the name stdout doesn't really convey the idea that prefixes are removed either. A pretty flag could cover this, however, we do not want to make this too complex either. Perhaps some baked-in pretty configurations, with the option to go custom would be best. Some ideas that come to mind are

`turbo run build` & `turbo run build --pretty` & `turbo run build --pretty=normal`: Current output

`turbo run build --pretty=none`: Remove prefixes, basically the current result from `stdout-full`

`turbo run build --pretty=no-color`: This would remove the color from the prefixes we have currently, it should not affect the logs themselves.

`turbo run build --pretty=folder-name`: This would log the folder name of the workspace instead of the package name

`turbo run build --pretty="%Cred%pn%Creset: %O"`: Custom format, print the package name in red
`%Cred` to get color red, `%pn` for package name, `%Creset` to reset the color, `%o` for output

If we were to go for this option this would require a rewrite for a large portion of the logging system.
And of course more investigation, I believe that would fall out of the scope for this PR though, so how about implementing `--pretty=none`, `--pretty`, and `--pretty=normal` for now? That way the team can have some time to think this idea over properly, and probably create an internal or public RFC.

This would also allow the current stdout behavior to be combined with errors-only, which was left out in this PR.

mehulkar:
> how about implementing --pretty=none, --pretty, and --pretty=normal for now?

I think we can start with only true/false (defaulting to true).  My suggestion to expand the flag was to demonstrate _how_ `--pretty` would be more open for expansion, rather than to ask for this! Definitely don't go down the rabbit hole rewriting logging! :)

cc @gsoltis / @vercel/turbo-oss to help make a decision/bikeshed here as well.

One more thing to point out is that without prefixing, logs for tasks running in parallel will be mixed up. The prefixes are the only thing that let you differentiate between them right now. That may be a reason to add `--pretty=no-color` (which would leave prefixes) in addition to true/false options. Or it could just be documented if we decide to tackle #219 at some point.

rafaeltab:
I was absolutely not planning to rewrite logging just yet, the 3 possible options I could implement all represent either the no prefix logging or the normal logging, so the current implementation could almost completely work for them. Then when you guys decide you want to go all the way we won't have any backwards compatibility issues.

If you still prefer a simple boolean, that is also still an option. Would that be `--pretty` and `--pretty false` or `--no-pretty` or both.

rafaeltab:
The logs will indeed be mixed up, and `--grouped` would solve this, I am probably going to try picking that up once this is finished as well. It also seems to be the case that people don't want to remove the prefix because they think it is ugly, but because it screws up some CI mechanisms (such as set-output for github).

gsoltis:
How about a separate boolean flag like `--raw-output` or `--no-prefix-logs` or something like that? That way it's separate from which logs to print. Would that cover what we need?

rafaeltab:
@gsoltis that would absolutely cover what we need. I think it's more a question of what do we want to add in the future when it comes to logging, and what do we think looks and feels better.

I believe `--no-prefix`, `--no-prefix-logs`, `--raw-output` & `--pretty=false` all cover what we currently need. The `--pretty` option allows for more future flexibility, and `--no-prefix`/`--no-prefix-logs` reflects the current behavior in the best way.

Do we want more flexibility, or de we want more clarity?

gsoltis:
I see. Let's go with `--raw-logs`. I think that communicates what we're doing and optimizes for clarity now. If we need more flexibility in the future, we can either deprecate or mark as incompatible with whatever flags we end up needing then.

zomars:
Can we take over this PR?

rafaeltab:
I have been a bit busy recently (I will be moving out of my parents' home to live with a friend of mine), I will continue development for this PR this week.

mehulkar:
Hey @rafaeltab thanks for your work on this. We talked internally and landed on `—log-prefix`. I’ve started that work in https://github.com/vercel/turbo/pull/3851 and it’s my top priority tomorrow. We expect it to land in a 1.8.x patch release! Feel free to leave any feedback there! Also happy to share more of our thinking around the naming choice, etc.

rafaeltab:
Alright, it's understandable since I hadn't worked on it for a little while. The naming choice seems straightforward as well. I'll close this PR then. I'll probably end up picking up grouped output in the near future.

PR #3440, opened 2023-01-23
chore: Port configuration files to Rust

chris-olszewski:
A full Rust port of configuration reading and writing. This PR is probably best reviewed in each individual commit.

This PR establishes, but doesn't enforce a pattern of `ConfigLoader`, `Config`, and `ConfigInner`. In the future we can generalize this to reduce boilerplate, but considering we have two configs at the moment that seemed like overkill.

- `ConfigInner`: This is the configuration file layout on disk. This is just a data object and shouldn't have any logic
- `Config`: Handles the applying of defaults and writing config updates to disk. Internally it holds onto a copy of the config that comes from the file exclusively to avoid accidentally writing a value that was provided via command line or environment variable to the configuration file.
- `ConfigLoader`: Handles the loading of the `Config` structure this is where the command line overrides are applied and can be used to mock out the environment to allow for testing without polluting the test proc's environment.

Notes:
 - We need to use a handrolled `MappedEnvironment` to provide custom mappings from environment variable names to config file keys. e.g. `TURBO_TEAM` gets mapped to `teamslug` in `config.json`


mehulkar (approved):
Really like this. I can't read Rust very well yet, but I like `Config`/`ConfigInner`/`ConfigLoader` pattern. `Config` and `ConfigInner` are a bit vaguely named though and I wouldn't know the difference without reading more. I suggest `ConfigFile` for ConfigInner instead?

Otherwise, this looks good to me! (you may want to have another Rustacean look at the code as well before merge)

NicholasLYang (approved):
This all makes sense to me. The only issues I see is that the `replace` function is a little ambiguous as to which direction it's replacing.

PR #3442, opened 2023-01-24
Move E2E tests to main test.yml workflow

mehulkar:


.github/workflows/test.yml
```diff
@@ -213,6 +213,34 @@ jobs:
 
       - name: Integration Tests
         run: pnpm -- turbo run integration-tests --filter=cli
+
+  go_e2e:
+    name: Go E2E Tests
+    needs: determine_jobs
+    if: needs.determine_jobs.outputs.go == 'true'
+    timeout-minutes: 30
```
chris-olszewski, line 221: It looks like this might not be enough time to build and run the tests on Windows 😢

chris-olszewski (dismissed):

.github/workflows/test.yml
```diff
@@ -213,6 +213,34 @@ jobs:
 
       - name: Integration Tests
         run: pnpm -- turbo run integration-tests --filter=cli
+
+  go_e2e:
+    name: Go E2E Tests
+    needs: determine_jobs
+    if: needs.determine_jobs.outputs.go == 'true'
```
chris-olszewski, line 220: We should also run if anything in `crates/turborepo*` changes along with global changes like `Cargo.lock`

.github/workflows/test.yml
mehulkar: it's the same as it is now? https://github.com/vercel/turbo/blob/01e143d80720b532369383e12c60f87368811185/.github/workflows/pr-go-e2e-filtered.yml#L20

.github/workflows/test.yml
mehulkar: updated and also included prysk tests for this condition if you want to take a look

tknickman:
You'll need to mark these as required as well (like here): https://github.com/vercel/turbo/pull/3519

mehulkar:
@tknickman done

chris-olszewski (approved):

.github/workflows/test.yml
```diff
@@ -82,13 +82,19 @@ jobs:
...
         uses: technote-space/get-diff-action@v6
         with:
           PATTERNS: |
             cli/**
+      - name: Go E2E related changes
+        id: go_e2e
+        uses: technote-space/get-diff-action@v6
+        with:
+          PATTERNS: |
+            shim/**
```
NicholasLYang, line 95: shim is no longer needed. We should probably add `turbo-updater` too

mehulkar: removing shim, but can we rename turbo-updater to turborepo-updater instead? Why the naming inconsistency?

mehulkar:
Merging this. Rust tests failed, but @jridgewell confirmed that they're failing at the moment. All other checks are green

PR #3446, opened 2023-01-24
Handle routing by calling Next.js code

jridgewell:
This implements routing by using a-yet-to-be-implemented API exposed by Next.js. The API follows something similar to:

```typescript
type MakeResolver = (config: NextConfig) => Resolver;
type Resolver = (IncomingMessage, ServerResponse) => Promise<void> | void;

import { makeResolver } from "next/dist/...";
const resolver  = makeResolver(nextConfig as object);

// Later, once we have a request we'd like to route:
// We don't care what the promise resolved to, we just want it to settle.
await resolver(req, res);
```

The resolver can do 1 of 3 things with this:
1. ~~Return a redirect response~~ Removed
2. Return a rewrite response
3. Stream a middleware response

> ~~1. Return a redirect response~~

<details>

First, ensure a `x-nextjs-route-result: 1` header is present on the response. Then return a JSON encoded body:

```typescript
{
  url: string,
  statusCode: u16,
  headers: Record<string, string>
  isRedirect: true
}
```

The Rust server will then respond with a redirect using to the appropriate location.

</details>

> 2. Return a rewrite response

First, ensure a `x-nextjs-route-result: 1` header is present on the response. Then return a JSON encoded body:

```typescript
{
  url: string,
  headers: Record<string, string>
}
```

The Rust server will use this updated URL to request content from our handlers.

> 3. Stream a middleware response

Ensure `x-nextjs-route-result` header **is not present** on the response. All headers will be sent back, and the body will be streamed back to the browser.

- - -

TODO:
- ~~Do `headers` actually matter to a `redirect`~~ Yes?
- ~~Does `statusCode` actually matter to a `rewrite`?~~ No
- ~~We can't handle streaming body responses yet.~~ Mocked by buffering.

Fixes WEB-228


crates/turbopack-dev-server/src/lib.rs
```diff
@@ -419,24 +422,8 @@ pub(crate) fn resource_to_data(
...
-                    }
-                    Entry::Occupied(mut e) => {
-                        e.get_mut().extend_with_string(header_value);
-                    }
-                }
-            }
-        }
-        data.headers = Some(headers);
+    if vary.headers.is_some() {
+        data.headers = Some(Headers::default())
```
sokra, line 426: Why was this changed? Seems like it's no longer passing the headers from the ResourceIdentifier to data. This is needed for RSC HMR which sends the `rsc: 1` header over web socket.

jridgewell: Damn, this was a change I made earlier when extracting `Headers` into its own struct that I forgot to revert.

crates/next-core/js/src/entry/router.ts
```diff
@@ -0,0 +1,148 @@
...
+  statusCode: number;
+  headers: string[];
+};
+
+export default async function route(
+  ipc: Ipc<RouterRequest, IpcOutgoingMessage>,
+  routerRequest: RouterRequest
+) {
+  // Deferring the import allows us to not error while we wait for Next.js to implement.
+  const { makeResolver } = await import("next/dist/server/router");
```
ijjk, line 58:
```suggestion
  const { makeResolver } = await import("next/dist/server/router.js");
```
Looks like this needs to include extension.

CleanShot 2023-01-25 at 15 13 00@2x

crates/next-core/js/src/entry/router.ts
```diff
@@ -0,0 +1,148 @@
...
+  const nextConfig = loadNextConfig();
```
ijjk, line 59:
```suggestion
  const nextConfig = await loadNextConfig();
```

crates/next-core/js/src/entry/router.ts
```diff
@@ -0,0 +1,148 @@
...
+  //   'x-nextjs-route-result: 1' to signal the body has the JSON payload with result,
+  //       else it streams the response headers/body as raw bytes.
+  // body (if 'x-nextjs-route-result: 1'):
+  // {
+  //   url: '/', // resolved url (includes query info if applicable)
+  //   headers: {}, // response headers to send down
+  //   statusCode: 200, //
+  //   isRedirect: false, //
+  // }
+  const resolveRoute = makeResolver(nextConfig);
```
ijjk, line 76:
```suggestion
  const resolveRoute = await makeResolver(nextConfig);
```
Let's allow this to return a promise for initializing the custom routes which are async.

crates/next-core/js/src/entry/config/next.js
```diff
@@ -1,13 +1,16 @@
...
 
   nextConfig.generateBuildId = await nextConfig.generateBuildId?.();
-  nextConfig.headers = await nextConfig.headers?.();
-  nextConfig.redirects = await nextConfig.redirects?.();
-  nextConfig.rewrites = await nextConfig.rewrites?.();
+
+  const { headers, redirects, rewrites } = await loadCustomRoutes(nextConfig);
+  nextConfig.headers = headers;
+  nextConfig.redirects = redirects;
+  nextConfig.rewrites = rewrites;
```
ijjk, line 13: This should be handled on the Next.js side

jridgewell: Removed.

crates/next-core/Cargo.toml
```diff
@@ -17,6 +17,7 @@ mime = "0.3.16"
 once_cell = "1.13.0"
 qstring = "0.7.2"
 regex = "1.6.0"
+reqwest = { workspace = true, features = ["stream"] }
```
sokra, line 20: Is this still needed?

sokra (approved):

wbinnssmith (approved):

crates/next-core/Cargo.toml
jridgewell: Removed in a follow up. This was part of the initial work where we would make an actual HTTP request to a router server spun up by Next.js during init.

crates/next-core/js/src/entry/router.ts
```diff
@@ -0,0 +1,155 @@
...
+  } catch (e) {
+    ipc.sendError(e as Error);
+  }
+}
+
+async function handleClientResponse(
+  _ipc: Ipc<RouterRequest, IpcOutgoingMessage>,
+  clientResponsePromise: Promise<IncomingMessage>
+): Promise<MessageData | void> {
+  const clientResponse = await clientResponsePromise;
```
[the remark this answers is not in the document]
jridgewell: Refactored.

crates/next-core/js/src/entry/router.ts
```diff
@@ -0,0 +1,155 @@
...
+      routerRequest.headers
+    );
+
+    // Send the clientRequest, so the server parses everything. We can then pass
+    // the serverRequest to Next.js to handle.
+    clientRequest.end();
+
+    const [_, response] = await Promise.all([
+      resolveRoute(serverRequest, serverResponse),
+      handleClientResponse(ipc, clientResponsePromise),
```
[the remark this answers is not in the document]
jridgewell: Refactored.

jridgewell (commented):
Review requests addressed in #3501

PR #3448, opened 2023-01-24
Handle 404 page when getStaticProps.notFound = true or the route is not found

alexkirsz:
This specifically supports the case where `getStaticProps.notFound = true` or a matching route is not found, but not the case where  `getStaticPaths.fallback = false` and the route is not enumerated;

This separate case will be implemented in another PR.


alexkirsz:
* main
  * **PR #3435** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3435" target="_blank">Graphite</a> 
    * **PR #3448** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3448" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3448?utm_source=stack-comment).

sokra (changes requested):
Isn't it better to respond with `Rewrite("/404")` instead of compiling the error and 404 page into every page?

sokra (approved):

crates/next-core/js/src/internal/page-server-handler.tsx
```diff
@@ -188,6 +169,19 @@ export default function startHandler({
       headers: renderData.headers,
     } as any;
     const res: ServerResponse = new ServerResponseShim(req) as any;
+
+    // Both _error and 404 should receive a 404 status code.
```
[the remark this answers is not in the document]
alexkirsz: When navigating to `/_error` in Next.js, a 404 status is always returned.

crates/turbopack-ecmascript/src/references/esm/module_id.rs
```diff
@@ -75,6 +78,17 @@ impl CodeGenerateable for EsmModuleIdAssetReference {
...
             );
+        } else {
+            // If the referenced asset can't be found, replace the expression with null.
+            // This can happen if the referenced asset is an external, or doesn't resolve
+            // to anything.
+            visitors.push(
+                create_visitor!(self.ast_path.await?, visit_mut_expr(expr: &mut Expr) {
+                    *expr = Expr::Lit(Lit::Null(Null {
+                        span: DUMMY_SP
+                    }));
```
jridgewell, line 89: Nit:
```suggestion
                    *expr = quote!("null" as Expr);
```

crates/turbopack-ecmascript/src/references/esm/module_id.rs
```diff
@@ -1,5 +1,8 @@
 use anyhow::Result;
-use swc_core::ecma::ast::Expr;
+use swc_core::{
+    common::DUMMY_SP,
+    ecma::ast::{Expr, Lit, Null},
+};
```
jridgewell, line 5: Nit:
```suggestion
use swc_core::{ecma::ast::Expr, quote};
```

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1242,11 +1242,22 @@ impl File {
...
+        self.meta.status_code
+    }
+
     pub fn with_content_type(mut self, content_type: Mime) -> Self {
         self.meta.content_type = Some(content_type);
         self
     }
 
+    /// Sets the status code associated with this file.
+    pub fn with_status_code(mut self, status_code: u16) -> Self {
```
jridgewell, line 1256: I think this is the wrong level for this API, and I'm glad the follow up PR is removing it.

crates/next-core/src/page_source.rs
```diff
@@ -400,6 +433,106 @@ async fn create_page_source_for_file(
     })
 }
 
+async fn get_not_found_page(pages_dir: FileSystemPathVc) -> Result<Option<AssetVc>> {
+    for ext in ["js", "ts", "jsx", "tsx", "mdx"] {
```
jridgewell, line 437: I think this needs to updated to hook into the next-config's extensions?

crates/next-core/src/next_route_matcher/prefix_suffix.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Creates a new [PrefixSuffixMatcher].
+    pub fn new(prefix: String, suffix: String, inner: T) -> Self {
+        Self {
+            prefix,
+            suffix,
+            inner,
+        }
+    }
+
+    fn strip_prefix_and_suffix<'a, 'b>(&'a self, path: &'b str) -> Result<Option<&'b str>> {
```
jridgewell, line 31: Nit: Can we unnest this?

```suggestion
    fn strip_prefix_and_suffix<'a, 'b>(&'a self, path: &'b str) -> Option<&'b str> {

File: crates/turbopack-node/src/route_matcher.rs [hunk 7]
```diff
@@ -0,0 +1,15 @@
+use indexmap::IndexMap;
+use turbo_tasks::primitives::BoolVc;
+
+#[turbo_tasks::value(transparent)]
+pub struct Params(Option<IndexMap<String, String>>);
+
+/// Extracts parameters from a URL path.
+#[turbo_tasks::value_trait]
+pub trait RouteMatcher {
```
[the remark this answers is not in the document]
alexkirsz: It belongs at the same level as `NodeRenderContentSource`, which is the only thing that actually calls into it, and that's in `turbopack-node` at the moment.

crates/next-core/js/src/entry/error.tsx
```diff
@@ -0,0 +1,4 @@
+import Error, { getStaticProps } from "@vercel/turbopack-next/internal/_error";
+
+export default Error;
+export { getStaticProps };
```
ForsakenHarmony, line 4: this is actually not supported by next.js

https://nextjs.org/docs/advanced-features/custom-error-page#caveats

PR #3449, opened 2023-01-24
Add RSS auto-discovery to Turbo Blog

jschuur:
RSS auto-discovery means you can just paste the blog URL into a feed reader and it will identify the RSS feed automatically for easy subscription (especially since the feed URL isn't promoted right now).

See also https://github.com/vercel/turbo/discussions/3447


mehulkar (approved):

jschuur:
Sorry, I was cleaning up repos on my end, and thought deleting my fork doesn't affect the PR, but it looks like it closed it.

tknickman (approved):

mehulkar:
The linter is failing on docs/theme.config.js (you should be able to repro with `turbo run check:prettier` locally). Not sure what the issue is but `pretter -w ./docs/theme.config.js` will probably fix it.

PR #3450, opened 2023-01-24
fix chunk optimization

sokra:
before it just merged all stuff into one chunk when over the chunk limit


crates/turbopack-core/src/chunk/optimize.rs
```diff
@@ -2,18 +2,22 @@ use std::{cell::RefCell, mem::take, rc::Rc};
```
ForsakenHarmony, line 2: would be nice to add some module level docs that give a high level overview of what's happening here

crates/turbopack-ecmascript/src/chunk/optimize.rs
```diff
@@ -1,4 +1,7 @@
-use std::{cmp::Ordering, collections::HashSet};
+use std::{
```
ForsakenHarmony, line 1: would be nice to add some module level docs that give a high level overview of what's happening here (same as above)

ForsakenHarmony (approved):

crates/turbopack-ecmascript/src/chunk/optimize.rs
```diff
@@ -269,11 +272,69 @@ async fn merge_duplicated_and_contained(
...
+    target_count: usize,
+) -> Result<Vec<EcmascriptChunkVc>> {
+    let mut remaining = chunks.len();
+    let mut chunks = chunks.into_iter().enumerate().collect::<Vec<_>>();
+    chunks.sort_by_key(|&(i, (_, k))| (k, i));
+    let mut chunks_by_source = chunks
+        .group_by(|(_, (_, a)), (_, (_, b))| a == b)
+        .map(|slice| {
+            (
+                slice[0].1 .1,
```
jridgewell, line 286: Weird space
```suggestion
                slice[0].1.1,
```

crates/turbopack-ecmascript/src/chunk/optimize.rs
```diff
@@ -269,11 +272,69 @@ async fn merge_duplicated_and_contained(
     Ok(())
 }
 
-async fn merge_by_size(chunks: &[EcmascriptChunkVc]) -> Result<Vec<EcmascriptChunkVc>> {
+async fn merge_to_limit(
+    chunks: Vec<(EcmascriptChunkVc, Option<ChunksVc>)>,
+    target_count: usize,
+) -> Result<Vec<EcmascriptChunkVc>> {
+    let mut remaining = chunks.len();
+    let mut chunks = chunks.into_iter().enumerate().collect::<Vec<_>>();
```
jridgewell, line 280: I think the enumerate (and the nested tuple complications below) is unnecessary? I assume you're doing it to ensure a stable sorting of the chunks, but Rust's sort it stable already.

jridgewell (changes requested):

crates/turbopack-core/src/chunk/optimize.rs
```diff
@@ -169,6 +197,19 @@ impl ContainmentTree {
...
 
+        fn skip_unnessary_nodes(trees: &mut IndexMap<FileSystemPathVc, Rc<RefCell<Node>>>) {
+            for tree in trees.values_mut() {
+                let mut tree = tree.borrow_mut();
+                if tree.chunks.len() == 0 && tree.children.len() == 1 {
+                    let child = tree.children.pop().unwrap();
+                    let mut child = child.borrow_mut();
+                    tree.path = child.path;
+                    tree.chunks.append(&mut child.chunks);
+                    tree.children.append(&mut child.children);
```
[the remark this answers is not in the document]
sokra: Good point

crates/turbopack-ecmascript/src/chunk/optimize.rs
sokra: cargo fmt adds this. I guess it should not be a number literal

crates/turbopack-core/src/chunk/optimize.rs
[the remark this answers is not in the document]
sokra: Seems like the AI wrote that correctly^^

crates/turbopack-ecmascript/src/chunk/mod.rs
```diff
@@ -126,6 +126,7 @@ impl EcmascriptChunkVc {
             .resolve()
             .await?;
         for path in paths {
+            let path = path.resolve().await?;
```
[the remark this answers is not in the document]
sokra: I don't see how this is possible...

crates/turbopack-core/src/chunk/optimize.rs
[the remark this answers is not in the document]
jridgewell: Ohh, I missed the `pop()` call.

jridgewell (approved):

ForsakenHarmony:
@sokra why did you ignore the clippy warnings?

PR #3451, opened 2023-01-24
force entrypoint file name for node.js execution to be .js

sokra:


crates/turbopack-node/src/evaluate.rs
```diff
@@ -69,7 +69,13 @@ pub async fn get_evaluate_pool(
     .as_asset();
 
     let module_path = module_asset.path().await?;
-    let path = intermediate_output_path.join(module_path.file_name());
+    let file_name = module_path.file_name();
+    let file_name = if file_name.ends_with(".js") {
```
jridgewell, line 73: Should we be forcing `.cjs` instead of `.js`? I assume there will be `require()` calls inside.

jridgewell (approved):

crates/turbopack-node/src/evaluate.rs
sokra: We could. We already write a package.json with type commonjs into the folder

PR #3454, opened 2023-01-24
fix(release): use turbo to build packages

tknickman:
We need to use turbo to build our packages because some of them now have internal dependencies that need to be built first.

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -88,9 +88,9 @@ cmd/turbo/version.go: ../version.txt
 	mv cmd/turbo/version.go.txt cmd/turbo/version.go
 
 build: install
-	cd $(CLI_DIR)/../packages/create-turbo && pnpm install --filter=create-turbo && pnpm run build
-	cd $(CLI_DIR)/../packages/turbo-codemod && pnpm install --filter=@turbo/codemod && pnpm run build
-	cd $(CLI_DIR)/../packages/turbo-ignore && pnpm install --filter=turbo-ignore && pnpm run build
+	cd $(CLI_DIR)/../ && pnpm install --filter=create-turbo && pnpm turbo build --filter=create-turbo...
+	cd $(CLI_DIR)/../ && pnpm install --filter=@turbo/codemod && pnpm turbo build --filter=@turbo/codemod...
+	cd $(CLI_DIR)/../ && pnpm install --filter=turbo-ignore && pnpm turbo build --filter=turbo-ignore...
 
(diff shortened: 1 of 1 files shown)


NicholasLYang (approved):

cli/Makefile
```diff
@@ -88,9 +88,9 @@ cmd/turbo/version.go: ../version.txt
 	mv cmd/turbo/version.go.txt cmd/turbo/version.go
 
 build: install
-	cd $(CLI_DIR)/../packages/create-turbo && pnpm install --filter=create-turbo && pnpm run build
-	cd $(CLI_DIR)/../packages/turbo-codemod && pnpm install --filter=@turbo/codemod && pnpm run build
-	cd $(CLI_DIR)/../packages/turbo-ignore && pnpm install --filter=turbo-ignore && pnpm run build
+	cd $(CLI_DIR)/../ && pnpm install --filter=create-turbo && pnpm turbo build --filter=create-turbo...
+	cd $(CLI_DIR)/../ && pnpm install --filter=@turbo/codemod && pnpm turbo build --filter=@turbo/codemod...
+	cd $(CLI_DIR)/../ && pnpm install --filter=turbo-ignore && pnpm turbo build --filter=turbo-ignore...
```
mehulkar, line 93: I haven't used `--filter` much. I see the [`...` documented here](https://turbo.build/repo/docs/core-concepts/monorepos/filtering#include-dependencies-of-matched-workspaces), but isn't this also default behavior? Or is it not default behavior when `--filter` is used?

mehulkar (approved):

cli/Makefile
tknickman: Yea good question, it's a little confusing, but there is a difference. However practice it’s not usually applicable. Say you only have this in your turbo.json (no depends on):

```json
"build": {
   "outputs": ["dist/**", ".next/**"]
 }
```

In this case, you will get different results depending on the filter.
For example:
`turbo build --filter={app}` will only build `{app}`.
But,
`turbo build --filter={app}...` will build `{app}` , and any workspace dependency listed in `{app}`’s `package.json`

However, if you had (the more common):
```json
"build": {
   "dependsOn": ["^build"],
   "outputs": ["dist/**", ".next/**"]
 }
```

`--filter={app}` and `--filter={app}...` are going to behave the same.

mehulkar: ahhh, thanks for that explanation, that makes sense

PR #3459, opened 2023-01-24
port(turborepo): Ported CommandBase from Go

NicholasLYang:
Since we have all of the config and UI logic, we should have the Rust equivalent of Go's `cmdbase`, i.e. a single struct that contains all of the helper structs for running a command.


crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -1,2 +1,40 @@
...
+pub struct CommandBase {
+    pub repo_root: PathBuf,
+    pub ui: UI,
+    pub repo_config: RepoConfig,
+    pub user_config: UserConfig,
+}
+
+impl CommandBase {
+    pub fn new(args: Args, repo_root: PathBuf) -> Result<Self> {
+        let repo_config_path = repo_root.join(".turbo").join("config.json");
```
chris-olszewski, line 25: This operations should probably live in `config/repo.rs` for better discoverability

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -1,2 +1,40 @@
...
+};
+
 pub(crate) mod bin;
 pub(crate) mod logout;
+
+pub struct CommandBase {
+    pub repo_root: PathBuf,
+    pub ui: UI,
+    pub repo_config: RepoConfig,
+    pub user_config: UserConfig,
```
chris-olszewski, line 20: Not blocking, but it would be nice if we were lazy about reading these configs to avoid unnecessary fs operations.

crates/turborepo-lib/src/cli.rs
```diff
@@ -422,9 +422,11 @@ pub fn run(repo_state: Option<RepoState>) -> Result<Payload> {
         clap_args.cwd = Some(repo_state.root);
     }
 
-    if let Some(cwd) = &clap_args.cwd {
-        clap_args.cwd = Some(fs_canonicalize(cwd)?);
-    }
+    let repo_root = if let Some(cwd) = &clap_args.cwd {
+        fs_canonicalize(cwd)?
```
chris-olszewski, line 426: Don't we still need to normalize `clap_args.cwd` so Go gets a normalized path?

chris-olszewski (approved):
Only real concern I have is that we're no longer normalizing `clap_args.cwd`

crates/turborepo-lib/src/cli.rs
NicholasLYang: Ah good catch

crates/turborepo-lib/src/commands/mod.rs
NicholasLYang: We could have these be methods and store the args in the struct. Would be a little more copying but not a big deal

PR #3460, opened 2023-01-24
ci(workflow): reusable next.js integration test

kwonoj:
We'll need to run next.js test in multiple place (PR, new next.js release..). This PR makes running next.js integration test reusable without duplicating same logics.

Also this PR changes default behavior of next.js integration test to **always pull latest version of next.js**.

### DIFF
--- a/.github/workflows/nextjs-integration-test.yml
+++ b/.github/workflows/nextjs-integration-test.yml
@@ -1,0 +1,158 @@
+# Reusable workflow to execute certain version of Next.js integration tests
+# with turbopack.
+#
+# Refer test.yml for how this workflow is being initialized
+# - Workflow can specify `inputs.version` to specify which version of next.js to use, otherwise will use latest release version.
+name: Turbopack Next.js integration test
+
+on:
+  workflow_call:
+    inputs:
+      # Allow to specify Next.js version to run integration test against.
+      # If not specified, will use latest release version including canary.
+      version:
+        required: false
+        type: string
+
+jobs:
+  # Build debug build of next-dev to use in integration test.
+  rust_build_dev:
+    strategy:
+      fail-fast: false
+    runs-on: ubuntu-latest-16-core-oss
+    name: Rust building debug next-dev for next.js integration test
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Rust
+        uses: ./.github/actions/setup-rust
+        with:
+          targets: x86_64-unknown-linux-musl
+          cache-key: dev-x86_64-unknown-linux-musl
+          save-cache: true
+
+      - name: Install musl tools
+        run: |
+          wget https://github.com/napi-rs/napi-rs/releases/download/linux-musl-cross%4011.2.1/x86_64-linux-musl-native.tgz -O musl.tgz
+          tar -xvzf musl.tgz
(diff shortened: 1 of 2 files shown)


.github/workflows/nextjs-integration-test.yml
```diff
@@ -0,0 +1,160 @@
...
+          ./next-dev --display-version
+
+      - name: Install dependencies
+        run: |
+          corepack disable
+          pnpm install
+          pnpm run build
+
+      - run: |
+          docker run --rm -v $(pwd):/work mcr.microsoft.com/playwright:v1.28.1-focal /bin/bash -c "cd /work && ls && curl https://install-node.vercel.app/v${{ matrix.node }} | FORCE=1 bash && node -v && npm i -g pnpm@${PNPM_VERSION} && /work/next-dev --display-version && __INTERNAL_CUSTOM_TURBOPACK_BINARY=${NEXT_DEV_BIN} __INTERNAL_NEXT_DEV_TEST_TURBO_GLOB_MATCH=${NEXT_DEV_TEST_GLOB} NEXT_TEST_JOB=1 NEXT_TEST_MODE=dev xvfb-run node run-tests.js --type development --timings -g ${{ matrix.group }}/4 >> /proc/1/fd/1"
```
wbinnssmith, line 135: While it would introduce some indirection, maybe reference a script to wrap all this would reduce duplication with below

kwonoj: This part I specifically allowed duplication, as this is 95% copy-paste from next.js's integration CI workflow so we can update easily later by recopy again if there are changes. I hope to simplify it maybe later, but for now put it as-is for those reason.

.github/workflows/nextjs-integration-test.yml
```diff
@@ -0,0 +1,158 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: vercel/next.js
+          ref: ${{ env.NEXTJS_VERSION }}
+
+      - uses: actions/cache@v3
+        id: restore-build
+        with:
+          path: ./*
+          key: ${{ github.sha }}-${{ github.run_number }}
```
ForsakenHarmony, line 103: What exactly is this supposed to do?

I think as is, this will cache the entire repository and never restore anything

.github/workflows/nextjs-integration-test.yml
```diff
@@ -0,0 +1,158 @@
...
+      - name: Download binary
+        uses: actions/download-artifact@v3
+        with:
+          path: artifacts
+
+      - name: Validate next-dev binary
+        run: |
+          ls -r ${{ github.workspace }}/artifacts
+          chmod +x ${{ github.workspace }}/artifacts/debug-next-dev-linux-musl/next-dev
+          cp ${{ github.workspace }}/artifacts/debug-next-dev-linux-musl/next-dev .
```
ForsakenHarmony, line 123:
```suggestion
          mv ./artifacts/debug-next-dev-linux-musl/next-dev .
          chmod +x ./next-dev
```

ForsakenHarmony (approved):

.github/workflows/nextjs-integration-test.yml
kwonoj: it's probably copy-paste from next.js upstream, I'll try to double check later.

PR #3464, opened 2023-01-25
partially support next metadata

sokra:
This adds support for metadata in pages and layouts.

Full metadata support also needs support for implicit metadata with files named `icon.ico` etc.

This PR also improves the test suite and adds a basic test case for app dir support


crates/next-dev-tests/tests/integration.rs
```diff
@@ -207,49 +216,152 @@ async fn create_browser(is_debugging: bool) -> Result<(Browser, JoinHandle<()>)>
...
+                        return Ok(value);
+                    }
+                } else {
+                    return Err(anyhow!("Binding events channel ended unexpectedly"));
+                }
+                bindings_next = binding_events.next();
+            }
+            event = &mut network_next => {
+                if let Some(event) = event {
+                    println!("{} [{}]", event.response.url, event.response.status);
```
ForsakenHarmony, line 348: maybe add a prefix so it's obvious this is just a network debug log

crates/next-dev-tests/tests/integration.rs
```diff
@@ -207,49 +216,152 @@ async fn create_browser(is_debugging: bool) -> Result<(Browser, JoinHandle<()>)>
...
+    let mut console_events = page
+        .event_listener::<EventConsoleApiCalled>()
+        .await
+        .context("Unable to listen to console events")?;
+    let mut network_response_events = page
+        .event_listener::<EventResponseReceived>()
+        .await
+        .context("Unable to listen to response received events")?;
+
+    page.evaluate_expression(format!("window.location='http://{addr}'"))
```
ForsakenHarmony, line 256: is there a reason for this change and for not using `page.goto(format!("http://{}", addr))`?

crates/next-dev-tests/tests/integration.rs
```diff
@@ -43,6 +51,7 @@ lazy_static! {
     // Allows for interactive manual debugging of a test case in a browser with:
     // `TURBOPACK_DEBUG_BROWSER=1 cargo test -p next-dev-tests -- test_my_pattern --nocapture`
     static ref DEBUG_BROWSER: bool = env::var("TURBOPACK_DEBUG_BROWSER").is_ok();
+    static ref BROWSER_HEAD: bool = env::var("TURBOPACK_BROWSER_HEAD").is_ok();
```
ForsakenHarmony, line 54: where is this used?

ForsakenHarmony (changes requested):

crates/next-dev-tests/tests/integration.rs
sokra: I copied that from the benchmarks. The difference is that this doesn't wait for the load event to happen. I guess it doesn't matter that much here.

wbinnssmith (approved):

jridgewell (approved):

crates/next-core/js/src/entry/app-renderer.tsx
```diff
@@ -112,6 +112,26 @@ async function runOperation(renderData: RenderData) {
...
+        type: "layout",
+        layer: i,
+        mod: () => info.layout!.module,
+        path: `layout${i}.js`,
+      });
+    }
+    if (info.page) {
+      metadata.push({
+        type: "page",
+        layer: i - 1,
```
alexkirsz, line 129: Could you explain the `i - 1` here?

PR #3469, opened 2023-01-25
Improve CI recipe docs

mehulkar:
- Use latest moniker for turbo dependency
- Change test task to depend on build task of the same workspace

    The general setup is for packages to build themselves before running their own tests, rather than building their dependencies


styfle (approved):
Thanks!

docs/pages/repo/docs/ci/circleci.mdx
```diff
@@ -34,7 +34,7 @@ And a `turbo.json`:
       "dependsOn": ["^build"]
     },
     "test": {
-      "dependsOn": ["^build"]
+      "dependsOn": ["build"]
```
gsoltis, line 37: I think we should identify some common test setups before categorically recommending this

styfle: I think those recommendations can come in a follow up PR.

Right now, the docs need to be fixed for correctness so that you don't accidentally have build and test running concurrently.

gsoltis: I think that's the heart of the matter. Whether or not build and test running concurrently is correct or incorrect is context dependent. `turbo` will do either, but you need to tell it which one is correct for your setup. 

Put another way, it's not currently accidental that build and test are running concurrently, that's the intention. That intention is based on what we expect is the common case, not needing to build the workspace being tested before running test. We expect most setups aren't testing the built artifact by default, and are instead doing things like running unit tests. For the situations where that doesn't hold, you need to update your config to account for that.

mehulkar: Since there isn't a clear consensus but the existing docs are confusing, I've updated this PR to remove `dependsOn` entirely from this "Recipe" and added a note instead. Personally I think you're both right and I don't think one configuration is more correct than the other.

@gsoltis let me know if you want to spend more time researching the top frameworks and what they require (linear ticket would be good in that case). Otherwise, I'll either wait for an approval or close this down next week!

docs/components/ci-recipe-turbo-config.mdx
```diff
@@ -0,0 +1,12 @@
+```jsonc
+{
+  "$schema": "https://turbo.build/schema.json",
+  "pipeline": {
+    "build": {
+      "dependsOn": ["^build"]
+    },
+    // Add dependsOn if your tests need to run a `build` or another script first
```
styfle, line 8: Lets add another example with the correct usage. I could see someone ending up in the same place on accident by doing 

```
    "test": {
      "dependsOn": ["^build"]
    },
```

instead of the expected

```
    "test": {
      "dependsOn": ["build"]
    },
```

docs/pages/repo/docs/ci/circleci.mdx
gsoltis: I don't think removing `dependsOn` is the way to go, as you likely depend on _something_ for tests. My preference is to leave the `dependsOn` as is, with `^build`, and to add a section about integration testing or specifically testing against built artifacts that depend on `build`.

gsoltis: A quick check on MochaJS, jest, and jasmine finds that all have setups that do not require an independent build step. If you're using typescript, you may need to do something with a test-framework-specific plugin, or use ts-node, but even then your test suite doesn't depend on your build command, which is likely different.

docs/components/ci-recipe-turbo-config.mdx
nzpopa: I spotted the same thing https://discord.com/channels/818588653005176832/920705046147129344/1070478043283464202

But it's not yet clear to me what is the difference between:

```
"lint": {
      "dependsOn": ["^lint"]
    },
```

and

```
"lint": {}
```

mehulkar:
Closing this down as it's not ready for merge yet

PR #3474, opened 2023-01-25
fix(shim): use custom version parser

tknickman:
Fixes hanging canaries

### DIFF
--- a/crates/turborepo-lib/src/shim.rs
+++ b/crates/turborepo-lib/src/shim.rs
@@ -14,7 +14,7 @@ use chrono::offset::Local;
 use dunce::canonicalize as fs_canonicalize;
 use env_logger::{fmt::Color, Builder, Env, WriteStyle};
 use log::{debug, Level, LevelFilter};
-use semver::{Version, VersionReq};
+use semver::Version;
 use serde::{Deserialize, Serialize};
 use tiny_gradient::{GradientStr, RGB};
 use turbo_updater::check_for_updates;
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/shim.rs
```diff
@@ -35,7 +35,12 @@ static TURBO_PURE_OUTPUT_ARGS: [&str; 6] = [
 ];
 
 static TURBO_SKIP_NOTIFIER_ARGS: [&str; 4] = ["--help", "--h", "--version", "--v"];
-static SUPPORTS_SKIP_INFER_SEMVER: &str = ">=1.7.0-canary.0";
```
tknickman, line 38: According to the rust semver crate, `1.7.1-canary.0 >= 1.7.0-canary.0` is `false` so this was causing turbo to infinitely call itself because no `skip-infer` arg was passed.

We fix this by writing our own comparator instead of using the semver `matches`.

NicholasLYang (approved):

crates/turborepo-lib/src/shim.rs
nathanhammond: I spent a few moments seeing what it would take to make this work the way npm does.

```rust
extern crate semver;

use semver::Version;
use semver::VersionReq;

fn main() {
    // Purposely set into the future.
    let req = VersionReq::parse(">= 1.7.0-canary.3").unwrap();

    let input = "1.7.0-canary.0"; // errors for older canaries on the same patch
    // let input = "1.7.0-canary.4"; // works for future canaries on the same patch
    // let input = "1.7.0"; // works for the patch being canaried

    // let input = "1.7.1-canary.0"; // works for canaries on the next patch
    // let input = "1.8.1-canary.0"; // works for future minor canaries
    // let input = "1.8.1"; // works for future minors

    let parsed = &Version::parse(input).unwrap();

    // Let's remove the the prerelease information.
    let reduced = &Version::new(parsed.major, parsed.minor, parsed.patch);
    
    // This behavior is actually 100% correct for _any_ single comparator.
    // (Ignore the actual implementation, needs to be done in a loop to support multiple comparators.)
    let check_reduced = !(req.comparators[0].major == parsed.major && req.comparators[0].minor.unwrap() == parsed.minor && req.comparators[0].patch.unwrap() == parsed.patch);

    // The comparison.
    assert!(req.matches(parsed) || (check_reduced && req.matches(reduced)));
}
```

PR #3477, opened 2023-01-25
`changesets` action: `yarn` -> `pnpm`

anthonyshew:
Looks like we missed the action here when we moved our examples from yarn to pnpm.

Thanks, @colbyfayock for letting us know!

### DIFF
--- a/examples/with-changesets/.github/workflows/release.yml
+++ b/examples/with-changesets/.github/workflows/release.yml
@@ -21,14 +21,14 @@ jobs:
           node-version: 16.x
 
       - name: Install Dependencies
-        run: yarn
+        run: pnpm i
 
       - name: Create Release Pull Request or Publish to npm
         id: changesets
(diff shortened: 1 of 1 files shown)


examples/with-changesets/.github/workflows/release.yml
```diff
@@ -21,14 +21,14 @@ jobs:
...
-        run: yarn
+        run: pnpm i
 
       - name: Create Release Pull Request or Publish to npm
         id: changesets
         uses: changesets/action@v1
         with:
           # This expects you to have a script called release which does a build for your packages and calls changeset publish
-          publish: yarn release
+          publish: pnpm publish
```
gaspar09, line 31: This should still reference the `release` command in the package.json

anthonyshew: 🤦

gaspar09 (approved):

PR #3482, opened 2023-01-26
`changesets` action: add pnpm install

colbyfayock:
pnpm needs to be set up before it can be installed inside Actions


gaspar09 (approved):
Thank you

PR #3483, opened 2023-01-26
fix: pnpm prune misses root dependency

### LINKED ISSUE #2798: [turborepo] `turbo prune` creates broken `pnpm-lock.yaml`

1.6.3

pnpm

Mac

### Describe the Bug

`turbo prune --scope="<app-name>" --docker` creates broken/incomplete `pnpm-lock.yaml` file, which leads to `ERR_PNPM_ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY` and `ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE` errors .

### Expected Behavior

Pruned `pnpm-lock.yaml` lock file should still contain every needed dependency and pruned output should also include packages that a referenced from the project root (but are not inside the `--scope`) 

### To Reproduce

```sh
git clone https://github.com/dlehmhus/pnpm-prune-root-issue-example
cd pnpm-prune-root-issue-example/
docker build -f apps/web/Dockerfile -t "pnpm-prune-example" .
```
or, to inspect the output:
```sh
pnpm exec turbo prune --scope="web" --docker
```
if you take a look at the generated `pnpm-lock.yaml` you can see two issues:

1. `apps/web` references `13.0.4_biqbaboplfbrettd7655fr4n2y`, but that dependency doesn't exists in the pruned `pnpm-lock.yaml`
2. `.` (projet root) links to `link:packages/package-for-ci` but that doesn't exist in `out/json`

### Reproduction Repo

https://github.com/dlehmhus/pnpm-prune-root-issue-example

chris-olszewski:
Fixes #2798 which comprised of two issues:
 - We eagerly checked if a version specifier was already resolved in the package list. This can lead to grabbing a version of the package that doesn't have the expected peer dependencies. This was as easy as moving this check after we have gone down the happy path.
 - Not copying internal dependencies of the monorepo root. We now internally add `//` to the prune scope to make sure we get any internal dependencies.

For reviewers: I highly recommend reviewing the first commit in this PR initially as it unifies the pnpm lockfile implementations. I realized that my code split was unnecessary and made fixing bugs for lockfile v5 and v6 very difficult.


cli/internal/lockfile/pnpm_lockfile.go
```diff
@@ -14,82 +14,94 @@ import (
...
-	Dependencies         map[string]DependencyV6     `yaml:"dependencies,omitempty"`
-	OptionalDependencies map[string]DependencyV6     `yaml:"optionalDependencies,omitempty"`
-	DevDependencies      map[string]DependencyV6     `yaml:"devDependencies,omitempty"`
-	DependenciesMeta     map[string]DependenciesMeta `yaml:"dependenciesMeta,omitempty"`
-	PublishDirectory     string                      `yaml:"publishDirectory,omitempty"`
+// SpecifierMap is a type wrapper that overrides IsZero for Golang's map
+// to match the behavior that pnpm expects
+type SpecifierMap map[string]string
+
+// IsZero is used to check whether an object is zero to
```
NicholasLYang, line 56: Out of curiosity, what is the default functionality of `IsZero`? If a map is nil or empty?

cli/internal/lockfile/pnpm_lockfile.go
```diff
@@ -327,11 +326,12 @@ func pruneImporters(importers map[string]ProjectSnapshot, workspacePackages []tu
 		workspace := workspacePath.ToUnixPath().ToString()
 		importer, ok := importers[workspace]
 
-		if !ok {
-			return nil, fmt.Errorf("Unable to find import entry for workspace package %s", workspace)
+		// If a workspace has no dependencies *and* it is only depended on by the
```
NicholasLYang, line 329: Is this the only situation where a workspace won't show up as an importer?

NicholasLYang (approved):
I don't see any issues, at least to the extent that I'm understanding the code. Just a couple clarifying questions.

cli/internal/lockfile/pnpm_lockfile.go
chris-olszewski: The default behavior is `len(map) == 0` which is true if the map is nil or empty, but we need to differentiate between the two.

cli/internal/lockfile/pnpm_lockfile.go
chris-olszewski: It's the only time I've seen a workspace omitted from the importers. If there's another scenario where `pnpm` omits a workspace, then this change will catch that as well.

PR #3489, opened 2023-01-26
[FIX]: VSCode "Parsing error: Cannot find module 'next/babel" PNPM ESLint

ImBIOS:
# 🪲 Error Reported

```bash
Parsing error: Cannot find module 'next/babel'
```

# 💻 Where does the Error Spawn?

Requirements:

1. PNPM
2. ESLint
3. Monorepo (Turborepo)
4. Next.js

If all of the above requirements ✅ checked, you can see the Error in every ESLint config file in the children directories.

# 🤔 Cause

Still needed to figure out why does this happen. My hypothesis as it is because of the PNPM node_modules design that is read by VSCode's ESLint extension. But it's because ESLint cannot resolve `next/babel`.

# 🚀 How this solution works?

This works because `require.resolve` uses the node's module resolution to return the absolute path to a module. It can provide that absolute path to any importing modules.

### DIFF
--- a/examples/kitchen-sink/packages/eslint-config-custom-server/index.js
+++ b/examples/kitchen-sink/packages/eslint-config-custom-server/index.js
@@ -7,9 +7,6 @@ module.exports = {
   parserOptions: {
     ecmaVersion: "latest",
     sourceType: "module",
-    babelOptions: {
-      presets: [require.resolve('next/babel')],
-    },
   },
   overrides: [
     {
--- a/examples/kitchen-sink/packages/eslint-config-custom-server/index.js
+++ b/examples/kitchen-sink/packages/eslint-config-custom-server/index.js
@@ -7,6 +7,9 @@ module.exports = {
   parserOptions: {
     ecmaVersion: "latest",
     sourceType: "module",
+    babelOptions: {
+      presets: [require.resolve('next/babel')],
+    },
   },
   overrides: [
     {
--- a/examples/with-create-react-app/packages/eslint-config-custom/index.js
+++ b/examples/with-create-react-app/packages/eslint-config-custom/index.js
@@ -13,9 +13,4 @@ module.exports = {
       version: "detect",
     },
   },
-  parserOptions: {
-    babelOptions: {
-      presets: [require.resolve('next/babel')],
-    },
-  },
 };
--- a/examples/with-create-react-app/packages/eslint-config-custom/index.js
+++ b/examples/with-create-react-app/packages/eslint-config-custom/index.js
@@ -13,4 +13,9 @@ module.exports = {
       version: "detect",
     },
   },
+  parserOptions: {
+    babelOptions: {
+      presets: [require.resolve('next/babel')],
+    },
+  },
 };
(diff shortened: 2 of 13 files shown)


ImBIOS:
Please don't merge it, I still want to add this fix to every affected examples as much as I could.

ImBIOS:
Good 👍 to go now, I already edit all potentially affected examples and edited the error I found in `examples/basic` 🚀

examples/kitchen-sink/packages/eslint-config-custom-server/index.js
```diff
@@ -7,6 +7,9 @@ module.exports = {
   parserOptions: {
     ecmaVersion: "latest",
     sourceType: "module",
+    babelOptions: {
+      presets: [require.resolve('next/babel')],
```
tknickman, line 11: We shouldn't need this here, this config is for server apps only and doesn't extend from next

examples/with-create-react-app/packages/eslint-config-custom/index.js
```diff
@@ -13,4 +13,9 @@ module.exports = {
       version: "detect",
     },
   },
+  parserOptions: {
+    babelOptions: {
+      presets: [require.resolve('next/babel')],
```
tknickman, line 18: We shouldn't need this one either, this example is for CRA and not Next (doesn't extend from next)

examples/with-docker/packages/eslint-config-custom-server/index.js
```diff
@@ -7,6 +7,9 @@ module.exports = {
   parserOptions: {
     ecmaVersion: "latest",
     sourceType: "module",
+    babelOptions: {
+      presets: [require.resolve('next/babel')],
```
tknickman, line 11: One more, this config shouldn't be required here since we're not extending from next for this config.

examples/with-svelte/packages/eslint-config-custom/index.js
```diff
@@ -9,11 +9,14 @@ module.exports = {
   },
   parserOptions: {
     sourceType: 'module',
-    ecmaVersion: 2020
+    ecmaVersion: 2020,
+    babelOptions: {
+      presets: [require.resolve('next/babel')],
```
tknickman, line 14: Can remove, this example is for svelte, not next.

examples/with-vite/packages/eslint-config-custom/index.js
```diff
@@ -16,4 +16,9 @@ module.exports = {
   rules: {
     "@typescript-eslint/no-non-null-assertion": "off",
   },
+  parserOptions: {
+    babelOptions: {
+      presets: [require.resolve('next/babel')],
```
tknickman, line 21: Can remove, this example is for vite, not next.

tknickman (changes requested):
Thanks for the PR! left a few comments, but would love to get this one in.

tknickman (approved):

tknickman:
Awesome, thanks!

tunguyen-ct:
@tknickman @ImBIOS This fix can silent the error of VSCode ESLint extension, however the VSCode ESLint extension also stop working.

You can see the Output of ESLint extention:

Screenshot 2023-04-29 at 14 07 45

```
Error: Cannot find module 'next/babel'
```

ImBIOS:
I think we should find the root cause and solve it.

PR #3492, opened 2023-01-26
fix(turborepo): Add `.npmrc` support to `turbo prune`

### LINKED ISSUE #2871: [turborepo] `turbo prune` is not including `.npmrc` when using `pnpm`

1.6.3

pnpm

Mac

### Describe the Bug

When running `turbo prune --scope="<app-name>" --docker` and using a `.npmrc` file to configure `pnpm install` to use settings like `enable-pre-post-scripts=true` or `shamefully-hoist=true` the `.npmrc` file is missing in the `out` directory

### Expected Behavior

If exists a `.npmrc` should be copied to `out`

### To Reproduce

1. Use `pnpm` as package manager 
2. Create an `.npmrc` file with some config. e.g:
```sh
enable-pre-post-scripts=true
shamefully-hoist=true
```
3. Run `turbo prune --scope="<app-name>" --docker`
4. Inspect `out` folder, no `.npmrc` exists

### Reproduction Repo

_No response_

Kuirak:
closes #2871 

Copying `.npmrc` file if it exists to keep packageManager settings the same during deploy/Docker builds.

### DIFF
--- a/cli/internal/prune/prune.go
+++ b/cli/internal/prune/prune.go
@@ -190,6 +190,17 @@ func (p *prune) prune(opts *turbostate.PrunePayload) error {
 		}
 	}
 
+	if fs.FileExists(".npmrc") {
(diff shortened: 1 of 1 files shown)


chris-olszewski (approved):
Thanks for implementing this!

PR #3498, opened 2023-01-26
fix(shim): better error message for missing go bin

tknickman:


crates/turborepo/src/main.rs
```diff
@@ -1,20 +1,47 @@
...
+    let turbo_path = fs_canonicalize(current_exe()?)?;
+    let mut go_binary_path = turbo_path.clone();
     go_binary_path.pop();
     #[cfg(windows)]
     go_binary_path.push("go-turbo.exe");
     #[cfg(not(windows))]
     go_binary_path.push("go-turbo");
 
-    let serialized_args = serde_json::to_string(&args)?;
+    if !go_binary_path.exists() {
```
NicholasLYang, line 22: I'm confused, if the go binary doesn't exist we log out that we found it?

tknickman: Big oof. Was using this to test the error message format without removing my local go-bin. Thanks.

mehulkar:
@tknickman fyi, I think this broke integration tests. We can catch these once we get https://github.com/vercel/turbo/pull/3442 merged! The log lines in the integration tests don't match any more

PR #3500, opened 2023-01-26
port(turbo): Link

NicholasLYang:
Ported `link`. Required porting more `APIClient` functionality


crates/turborepo-lib/src/client.rs
```diff
@@ -12,14 +12,58 @@ use crate::{get_version, retry::retry_future};
...
+    OverLimit,
+    #[serde(rename = "paused")]
+    Paused,
+}
+
+/// Membership is the relationship between the logged-in user and a particular
+/// team
+#[derive(Debug, Clone, Deserialize)]
+pub struct Membership {
+    role: String,
```
chris-olszewski, line 36: The API docs list out the possible values `role` can take: https://vercel.com/docs/rest-api#interfaces/team-limited

I believe it would be safe to limit ourselves to those values via an enum since we're querying a specific version of the api.

crates/turborepo-lib/src/client.rs
```diff
@@ -12,14 +12,58 @@ use crate::{get_version, retry::retry_future};
 pub trait UserClient {
     fn set_token(&mut self, token: String);
     async fn get_user(&self) -> Result<UserResponse>;
+    async fn get_teams(&self) -> Result<TeamsResponse>;
```
chris-olszewski, line 15: I noticed we're getting rid of the separate interfaces for link and login that are present in the Go impl. I'm fine as long as testability isn't negatively affected.

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -0,0 +1,206 @@
...
+    let homedir = homedir_path.to_string_lossy();
+    println!(">>> Remote Caching");
+    println!();
+    println!("  Remote Caching shares your cached Turborepo task outputs and logs across");
+    println!("  all your team’s Vercel projects. It also can share outputs");
+    println!("  with other services that enable Remote Caching, like CI/CD systems.");
+    println!("  This results in faster build times and deployments for your team.");
+    println!(
+        "  For more info, see {}",
+        UNDERLINE.apply_to("https://turbo.build/repo/docs/core-concepts/remote-caching")
```
chris-olszewski, line 36:
```suggestion
        base.ui.apply(UNDERLINE.apply_to("https://turbo.build/repo/docs/core-concepts/remote-caching"))
```

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -0,0 +1,206 @@
...
+
+enum SelectedTeam<'a> {
+    User,
+    Team(&'a Team),
+}
+
+pub async fn link(mut base: CommandBase, modify_gitignore: bool) -> Result<()> {
+    let homedir_path = home_dir().ok_or_else(|| anyhow!("could not find home directory."))?;
+    let homedir = homedir_path.to_string_lossy();
+    println!(">>> Remote Caching");
```
chris-olszewski, line 28: Once again going to ask to combine these `println` statements to avoid locking/unlocking stdout unnecessarily

crates/turborepo-lib/src/cli.rs
```diff
@@ -461,8 +461,20 @@ pub async fn run(repo_state: Option<RepoState>) -> Result<Payload> {
...
+        Command::Link { no_gitignore } => {
+            if clap_args.test_run {
+                println!("Link test run successful");
+                return Ok(Payload::Rust(Ok(0)));
+            }
+
+            let modify_gitignore = !*no_gitignore;
+            let base = CommandBase::new(clap_args, repo_root)?;
+
+            link::link(base, modify_gitignore).await?;
```
chris-olszewski, line 473: I think we should be logging the error here before returning it to match the Go impl

chris-olszewski (changes requested):
We should rip out the Go impl with this PR. Would also really like to see some tests (not blocking though) since the integration test doesn't hit any of the logic inside `link`

crates/turborepo-lib/src/commands/link.rs
NicholasLYang: Does color include underline? I explicitly didn't apply this because I wasn't sure if the color config should apply to underline

crates/turborepo-lib/src/client.rs
NicholasLYang: Yeah it doesn't really make sense to have traits with overlapping methods, so we might as well combine them.

chris-olszewski (approved):

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -0,0 +1,206 @@
+use std::{
+    fs,
+    fs::{File, OpenOptions},
+    io,
+    io::{BufRead, Write},
+    os::unix::fs::PermissionsExt,
```
chris-olszewski, line 6: We need to configure this to only be imported on `not(windows)`

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -0,0 +1,206 @@
...
+    // We return an error no matter what
+    return Err(anyhow!("link after enabling caching"));
+}
+
+fn add_turbo_to_gitignore(base: &CommandBase) -> Result<()> {
+    let gitignore_path = base.repo_root.join(".gitignore");
+
+    if !gitignore_path.exists() {
+        let mut gitignore = File::create(gitignore_path)?;
+        #[cfg(unix)]
```
chris-olszewski, line 189: Should we include setting the permissions on macos as well?

crates/turborepo-lib/src/client.rs
```diff
@@ -13,27 +13,38 @@ pub trait UserClient {
     fn set_token(&mut self, token: String);
     async fn get_user(&self) -> Result<UserResponse>;
     async fn get_teams(&self) -> Result<TeamsResponse>;
-    async fn get_caching_status(&self, team_id: &str) -> Result<CachingStatus>;
+    async fn get_caching_status(&self, team_id: &str) -> Result<CachingStatusResponse>;
```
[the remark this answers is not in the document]
NicholasLYang: It's not particularly important. At some point we might add pagination, at which point the response type will have more importance.

crates/turborepo-lib/src/commands/link.rs
NicholasLYang: Isn't macOS also unix?

PR #3501, opened 2023-01-26
Review fixes for Next routing

jridgewell:
This completes a few review requests from #3446.


crates/next-core/js/src/entry/router.ts
```diff
@@ -84,24 +84,25 @@ export default async function route(
...
-    const [_, response] = await Promise.all([
-      resolveRoute(serverRequest, serverResponse),
-      handleClientResponse(ipc, clientResponsePromise),
-    ]);
-    server.close();
+    await resolveRoute(serverRequest, serverResponse);
+    const response = await handleClientResponse(
+      ipc,
+      await clientResponsePromise
+    );
```
sokra, line 91: This is not the same as before. Is this still correct in case of streaming? Now we wait for `resolveRoute` to finish before starting to process the response.

jridgewell: Yah, you're write. I was assuming that `resolveRoute` wouldn't await anything after is started writing because it currently doesn't, but that assumption could change at any point. I've updated, and added comments.

wbinnssmith (approved):

PR #3503, opened 2023-01-26
fix: Skip over repo root package for package inferrence

chris-olszewski:
Once again map iteration is the culprit. The failure of `infer_pkg.t` is caused by package inference iterating over the root package before the correct candidate package. This results in `pkgPath == repoRoot` and since `fullInferencePath` is necessarily contained in the repo root, the root package will always be selected as the exact package. The fix is just skipping over the root package.

First commit adds a root package to the tests. The package inference cases will fail in a non-deterministic fashion since this bug depends on the repo root package being iterated over before the package that should be inferred. The next commit is a fix.

### DIFF
--- a/cli/internal/scope/scope.go
+++ b/cli/internal/scope/scope.go
@@ -165,7 +165,8 @@ func calculateInference(repoRoot turbopath.AbsoluteSystemPath, rawPkgInferenceDi
 		if err != nil {
 			return nil, err
 		}
-		if inferredPathIsBelow {
+		// We skip over the root package as the inferred path will always be below it
+		if inferredPathIsBelow && pkgPath != repoRoot {
 			// set both. The user might have set a parent directory filter,
 			// in which case we *should* fail to find any packages, but we should
 			// do so in a consistent manner
--- a/cli/internal/scope/scope_test.go
+++ b/cli/internal/scope/scope_test.go
@@ -80,6 +80,10 @@ func TestResolvePackages(t *testing.T) {
 	graph.Connect(dag.BasicEdge("app2", "libC"))
 	graph.Connect(dag.BasicEdge("app2-a", "libC"))
 	workspaceInfos := internalGraph.WorkspaceInfos{
+		"//": {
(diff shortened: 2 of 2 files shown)


gsoltis (approved):

cli/internal/scope/scope_test.go
```diff
@@ -80,6 +80,10 @@ func TestResolvePackages(t *testing.T) {
 	graph.Connect(dag.BasicEdge("app2", "libC"))
 	graph.Connect(dag.BasicEdge("app2-a", "libC"))
 	workspaceInfos := internalGraph.WorkspaceInfos{
+		"//": {
+			Dir:  turbopath.AnchoredSystemPath("").ToSystemPath(),
+			Name: "monorepo",
+		},
```
mehulkar, line 86: maybe we need a test helper to construct all the internal bits that make up a valid monorepo representation... it's quite a pain to dig through the code to do this right. There's a beginning of one here: https://github.com/vercel/turbo/blob/3bbbc471d3f8b9f6c141d620ae6f18dd206fd8cd/cli/internal/core/engine_persistent_deps_test.go#L448-L449

cli/internal/scope/scope.go
```diff
@@ -165,7 +165,8 @@ func calculateInference(repoRoot turbopath.AbsoluteSystemPath, rawPkgInferenceDi
 		if err != nil {
 			return nil, err
 		}
-		if inferredPathIsBelow {
+		// We skip over the root package as the inferred path will always be below it
```
mehulkar, line 168: I don't know much about this feature, but is this true? If you run `turbo run` from the root of the repo, will inference be skipped altogether?

mehulkar (approved):
Looks good to me

cli/internal/scope/scope.go
chris-olszewski: `turbo run` from repo root will cause `rawPkgInferenceDir` to be `""` which triggers a return at the beginning of this function

Issue #3139, opened 2022-12-26
[turbopack] "Error: invalid type: string" when using variable on next/font


ZakariaBenali:
### What version of Turbopack are you using?

next: 13.1.1

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Linux

### Describe the Bug

Using the variable property on next/font with Turbopack causes the following error:
`[500] error: invalid type: string "--font-inter", expected a sequence at line 1 column 95`

The font is declared as per the docs: 
```
import { Inter } from '@next/font/google'
const inter = Inter({
  subsets:['latin'],
  variable: '--font-inter',
})
```

**Important to note:**   this only happens when you add the variable property, and you don't have to apply the font style to the app to experience the issue, just declaring the font like so with variable property causes the app to crash.

### Expected Behavior

The next/font with variable property works fine without Turbopack.

### To Reproduce

1. Simply create a fresh app using the latest version of next.js (13.1.1 by the time I'm writing this issue). 
2. Declare any google font on _app.tsx as per the doc: 
```
import { Inter } from '@next/font/google'
const inter = Inter({
  subsets:['latin'],
  variable: '--font-inter',
})
```
3. Run the app on turbo mode 'pnpm dev --turbo'

pan93412:
I found the exact issue – seems like a wrong declaration? Not sure why the developer declare this as a `Vec<T>` while specification says that this should be a string: <https://nextjs.org/docs/api-reference/next/font#variable>

https://github.com/vercel/turbo/blob/aa7e60b85ed4acbd9d56fbab1b06ad7cfce9e69a/crates/next-core/src/next_font_google/options.rs#LL23C9-L23C17

Changing `variable: 'str'` to `variable: ['str']` suppress this issue. `@ts-expect-error` may be needed.

beldur:
I have the same issue. Can't use `@next/font/google` with a variable.

So this documentation breaks with turbopack: https://nextjs.org/docs/api-reference/next/font#css-variables

minega25:
Hello There,

Still facing the same issue when setting fonts with tailwindcss according to below instructions: https://nextjs.org/docs/basic-features/font-optimization#with-tailwind-css

PR #3505, opened 2023-01-27
@next/font/google: support setting a css variable for fonts

wbinnssmith:
Fixes WEB-501

Fixes https://github.com/vercel/turbo/issues/3139

This adds proper support for the [`variable` property](https://nextjs.org/docs/api-reference/next/font#variable) when constructing fonts. This results in:

* defining a selector in the css module for a classname defining the css custom property (css variable), whose value is the font family
* adding a `variable` property to the resulting js object whose value is the classname used in the selector above

Test Plan: Added an integration test.

### DIFF
--- a/crates/next-core/src/next_font_google/mod.rs
+++ b/crates/next-core/src/next_font_google/mod.rs
@@ -3,6 +3,7 @@ use indexmap::IndexMap;
 use indoc::formatdoc;
 use once_cell::sync::Lazy;
 use turbo_tasks::primitives::{OptionStringVc, OptionU16Vc, StringVc, U32Vc};
+#[allow(unused_imports, /* reason = "this is used in tests" */)]
 use turbo_tasks_env::{CommandLineProcessEnvVc, ProcessEnv};
 use turbo_tasks_fetch::fetch;
 use turbo_tasks_fs::{FileContent, FileSystemPathVc};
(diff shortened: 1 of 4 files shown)


jridgewell:
Fixes https://github.com/vercel/turbo/issues/3139?

jridgewell (approved):

crates/next-core/src/next_font_google/mod.rs
```diff
@@ -3,6 +3,7 @@ use indexmap::IndexMap;
 use indoc::formatdoc;
 use once_cell::sync::Lazy;
 use turbo_tasks::primitives::{OptionStringVc, OptionU16Vc, StringVc, U32Vc};
+#[allow(unused_imports, /* reason = "this is used in tests" */)]
```
sokra, line 6:
```suggestion
#[cfg(debug_assertions)]
```

But best just move the `use ...` into the block below.

crates/next-dev-tests/tests/integration/next/font-google/basic/pages/index.js
```diff
@@ -25,27 +28,50 @@ function runTests() {
...
+}
+
+function getRuleMatchingClassName(className) {
+  const selector = `.${CSS.escape(className)}`;
+
+  let matchingRule;
+  for (const stylesheet of document.querySelectorAll("link[rel=stylesheet]")) {
+    const sheet = stylesheet.sheet;
+    if (sheet == null) {
+      continue;
```
sokra, line 65: There is a race condition with the loading of stylesheets here. This makes the tests flaky.

Maybe add a `await new Promise(resolve => stylesheet.onload = resolve);` or similar here?

sokra (approved):
Nitpicks can be fixed in follow-up PR

beldur:
Thanks a lot! 🙇

PR #3508, opened 2023-01-27
Fix prettier for examples

alexkirsz:
CI currently fails on main because of Prettier.


arlyon (approved):
Approved pending CI :)

alexkirsz:
The one job this fixes is filtered out for this particular PR, which is the reason this issue didn't surface in the PR. I'll fix this in this PR as well.

ForsakenHarmony (approved):

PR #3499, opened 2023-01-26
add support for pageExtensions in next.config.js

ForsakenHarmony:

PR #3510, opened 2023-01-27
add pageExtensions support to appDir

ForsakenHarmony:
The previous PR #3499 only implemented support for `pages` directory.

The `app` directory previously had no validation at all, it's added with this PR.


ForsakenHarmony:
* main
  * **PR #3510** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3510" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3510?utm_source=stack-comment).

crates/next-core/src/app_source.rs
```diff
@@ -344,131 +349,149 @@ async fn create_app_source_for_directory(
...
+    let DirectoryContent::Entries(entries) = &*input_dir.read_dir().await? else {
+        return Ok(CombinedContentSource { sources }.cell());
+    };
+
+    let allowed_extensions = &*page_extensions.await?;
+
+    for (name, entry) in entries.iter() {
+        if let &DirectoryEntry::File(file) = entry {
+            if let Some((name, ext)) = name.rsplit_once('.') {
+                if !allowed_extensions.iter().any(|allowed| allowed == ext) {
```
alexkirsz, line 362: Should this perhaps only apply to `page`/`error`/`loading`?

ForsakenHarmony: I don't think so, why?

alexkirsz: It feels weird that `pageExtensions` would apply to things that aren't technically pages, like `head` or `layout`, but I guess it makes sense to support these too, so 🤷

ForsakenHarmony: Well, we're just copying next.js behaviour

It treats all of them the same as far as I can tell https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack/loaders/next-app-loader.ts#L12-L12

jridgewell, line 362: Should we be representing this with a `BSet` instead?

jridgewell (approved):

sokra (approved):

PR #3512, opened 2023-01-27
Allow controlling depth of Vc debugging

alexkirsz:
Adds a mechanism for controlling the depth of `Vc` debugging, to avoid printing out huge structs when we're only interested in some superficial overview of it.

This will only affect `Vc`s: if a `Vc` contains a non-`Vc`, very deep struct, the struct will be fully printed out. This is because you can't control the printing depth of the default Rust `Debug` trait.

Marking this as a draft for now as I intend to change the name and arguments of `dbg_depth`.

```rust
eprintln!("{:?}", some_vc.dbg_depth(2).await?) // will now only print up to a depth of 2
```

### DIFF
--- a/crates/turbo-tasks-build/src/lib.rs
+++ b/crates/turbo-tasks-build/src/lib.rs
@@ -397,23 +397,27 @@ impl<'a> RegisterContext<'a> {
 
     /// Declares the default derive of the `ValueDebug` trait.
     fn register_debug_impl(&mut self, ident: &Ident, dbg_ty: DebugType) -> std::fmt::Result {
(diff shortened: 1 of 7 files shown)


jridgewell (approved):
😍😍😍

alexkirsz:
I meant to change `dbg_depth` to `dbg_opts` and pass it an options struct instead, but:
* it's unclear whether that will ever be used for anything other than `depth`; and
* this struct would need to be `FromTaskInput` to allow passing it as an argument to `ValueDebugVc`.

So I'll merge this as-is for now and perhaps revisit later.

PR #3513, opened 2023-01-27
docs: mentioned missing info regarding putting npm organization while…

mbsc26:
… importing workspace

For coherence on the page, I mentioned the missing information of adding the npm organization or user space as well if it is there in the imported workspace's `package.json` file as part of it's `name` during import of a workspace in another workspace.

For example, if it is `@mycompany/shared-utils` instead of `shared-utils`.

This case is not mentioned in the 'Workspaces which depend on each other' section.

Currently the 'Workspaces which depend on each other' section only explains the case where npm organization or user space is not part of the imported workspace's `name`

### DIFF
--- a/docs/pages/repo/docs/handbook/workspaces.mdx
+++ b/docs/pages/repo/docs/handbook/workspaces.mdx
@@ -147,6 +147,43 @@ For instance, if we want `apps/docs` to import `packages/shared-utils`, we'd nee
   of our packages change.
 </Callout>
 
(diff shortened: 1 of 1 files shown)


nathanhammond:
Hey @mbsc26! Thank you for the PR.

We explicitly specify the name of the package in a previous step in this walkthrough. If the user chooses to add a namespace to the package name we assume that they're also capable of knowing that they will need to further adjust the following steps accordingly.

```json filename="packages/shared-utils/package.json"
{
  "name": "shared-utils"
}
```

I'm going to close this as it is advanced usage beyond what we intend to cover in our walkthrough.

PR #3515, opened 2023-01-27
chore: add option to dump args sent to go

chris-olszewski:
I used this yesterday when debugging package inference and found it helpful for launching the Go debugger with the correct args. I figure it might be helpful for other devs as well.


mehulkar (commented):
Is this meant to be only for dev usage? Can we include it in `-vvv` output instead of adding an env var?

tknickman (approved):

PR #3516, opened 2023-01-27
ci: allow opting-in to a dry run for stage 3 of release

arlyon:
This adds an additional flag into the workflow that allows you to run a dry run of the release process, skipping the final publish step. The artifacts can be downloaded and tested, and a subsequent run on the same branch will download those prebuilt binaries rather than starting from scratch. For an example, see:

1. https://github.com/arlyon/turbo/actions/runs/4026998158 (dry run)

### DIFF
--- a/.github/workflows/turborepo-release-step-3.yml
+++ b/.github/workflows/turborepo-release-step-3.yml
@@ -5,6 +5,9 @@ on:
     inputs:
       release_branch:
         description: "Staging branch to run release from"
+      dry_run:
+        description: "Do a dry run, skipping the final publish step. The outputs of a previous dry-run will be reused when uploading properly, but you can delete the old ones to force a re-build."
+        type: boolean
 
 env:
   RELEASE_TURBO_CLI: true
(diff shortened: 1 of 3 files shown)


cli/Makefile
```diff
@@ -172,20 +172,22 @@ publish-turbo: clean build
 	npm config set --location=project "//registry.npmjs.org/:_authToken" $(NPM_TOKEN)
 
 	# Publishes the native npm modules.
-	goreleaser release --rm-dist -f combined-shim.yml
+	goreleaser release --rm-dist -f combined-shim.yml $(SKIP_PUBLISH)
```
tknickman, line 175: [Nice](https://goreleaser.com/quick-start/#release-flags)

.github/workflows/turborepo-release-step-3.yml
```diff
@@ -5,6 +5,9 @@ on:
     inputs:
       release_branch:
         description: "Staging branch to run release from"
+      dry_run:
+        description: "Do a dry run, skipping the final publish step. The outputs of a previous dry-run will be reused when uploading properly, but you can delete the old ones to force a re-build."
```
tknickman, line 9: What would the use case be for reusing outputs from a previous `dry-run`? It makes sense when you're running `dry-run=true` followed by `dry-run=false`, but if you run two dry-runs in a row you would want new artifacts right?

tknickman (approved):
One question, but overall looks great

arlyon:
Closing in favour of https://github.com/vercel/turbo/pull/3720

PR #3517, opened 2023-01-27
fix: pnpm prune misses git dependency

### LINKED ISSUE #2870: [turborepo] `turbo prune` creates `pnpm-lock.yaml` with missing `git` dependencies

1.6.3

pnpm

Mac

### Describe the Bug

`pnpm exec turbo prune --scope="<app-name>" --docker` creates broken `pnpm=lock.yaml` which leads to `ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY` when running `pnpm install`

### Expected Behavior

Pruned `pnpm-lock.yaml` file should still contain git dependencies

### To Reproduce

```sh
git clone https://github.com/peerigon/turbo-pnpm-prune-git-dependency-issue
cd turbo-pnpm-prune-git-dependency-issue/
docker build -f apps/web/Dockerfile -t "pnpm-prune-example" .
```

or to inspect output

```sh
pnpm exec turbo prune --scope="web" --docker
```

The resulting `pnpm-lock.yaml` is missing the entry for `github.com/peerigon/dashboard-icons/ce27ef933144e09cef3911025f3649040a8571b6` in the `packages:` key

### Reproduction Repo

https://github.com/peerigon/turbo-pnpm-prune-git-dependency-issue

chris-olszewski:
Fixes #2870

The resolved version of a `github:` dependency is the lockfile key of the package so now when resolving a package we check and see if we can find a package by using just the version instead of `/name@version`. 

This is built on top of #3483 so review that first.


cli/internal/lockfile/pnpm_lockfile.go
```diff
@@ -238,6 +238,16 @@ func (p *PnpmLockfile) ResolvePackage(workspacePath turbopath.AnchoredUnixPath,
 		return Package{Key: key, Version: version, Found: true}, nil
 	}
 
+	if entry, ok := p.Packages[resolvedVersion]; ok {
+		var version string
+		if entry.Version != "" {
```
NicholasLYang, line 243: Again, just to clarify, when would we have a version in the entry? If it's a git dependency would there be a version at all?

chris-olszewski: The test data added to this PR came from pnpm which does include a version in a git dependency entry. In the case of the git dependency we rely on the key changing as the commit hash is part of the key. This will catch a dependency change since we use both the key and version when determining if a package changed.

NicholasLYang (approved):

PR #3429, opened 2023-01-23
Move Go Integration tests into main test Github workflow

mehulkar:
e2e tests are still separate, there's a separate ticket for those.

PR #3519, opened 2023-01-27
fix(ci): require go integration

tknickman:
Introduced when moving go integration tests into our main test workflow:

https://github.com/vercel/turbo/pull/3429

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -904,6 +904,7 @@ jobs:
       - go_unit
       - go_examples
       - go_e2e
+      - go_integration
       - rust_lint
       - rust_check
       - rust_test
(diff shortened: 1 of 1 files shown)


chris-olszewski (approved):
Since we're now triggering these from the main we should update the patterns for `go` to include `crates/turborepo*` so we trigger on Rust changes that directly affect our CLI. It looks like right now we just run if `cli/` changes. This does mean we will run Go unit tests when Rust code changes though 😢

arlyon (approved):
Approved pending CI passing :)

mehulkar (changes requested):
I think we want #3442 instead

mehulkar (approved):
Never mind, this makes sense now

tknickman:
@chris-olszewski yeaaaa, then we're going to run these no matter what. We really need to break these down by product instead of language

PR #3524, opened 2023-01-27
Allow manual workflow dispatch and send a basic message through Slack

wbinnssmith:
This:

* adds `workflow_dispatch` to `on` for the integration test workflow
* adds a basic Slack message that's sent when integration tests complete

It seems like we have to land this change before manual workflow runs become available in the GitHub UI.


kwonoj (approved):

kwonoj:
This'll allow unblock my further task as well, let me automerge 🚢 and test if it works

PR #3526, opened 2023-01-28
Reduce memory usage when writing files

jridgewell:
This PR reduces the our total memory usage (hopefully) for files that we write to disk.

When we write a file, we obviously have the input bytes (the bytes we want to write to disk). But we also read the file to see if the file needs to be written at all. This means we have _another_ copy of the bytes, the ones that came from the disk.

Worse, our write is strongly dependent on the read's `RopeVc`. Meaning we're stuck with that read memory until some indeterminate time. And any time we update the written contents, we're gonna get more bytes stuck in our Vc cache.

The goal here is to track the file, so that we can rewrite it if it's modified/deleted. Why not have an API specific to tracking, without actually reading any contents? This PR does exactly that. And now that we don't need to read contents to track, we can avoid using the `read()` API (and its `RopeVc`) in order to do our new-old contents comparison.

### DIFF
--- a/crates/turbo-tasks-fs/src/lib.rs
+++ b/crates/turbo-tasks-fs/src/lib.rs
@@ -74,6 +74,9 @@ pub trait FileSystem: ValueToString {
     fn read(&self, fs_path: FileSystemPathVc) -> FileContentVc;
     fn read_link(&self, fs_path: FileSystemPathVc) -> LinkContentVc;
     fn read_dir(&self, fs_path: FileSystemPathVc) -> DirectoryContentVc;
+    fn track(&self, _fs_path: FileSystemPathVc) -> CompletionVc {
+        CompletionVc::immutable()
+    }
     fn write(&self, fs_path: FileSystemPathVc, content: FileContentVc) -> CompletionVc;
     fn write_link(&self, fs_path: FileSystemPathVc, target: LinkContentVc) -> CompletionVc;
     fn metadata(&self, fs_path: FileSystemPathVc) -> FileMetaVc;
@@ -467,6 +470,13 @@ impl FileSystem for DiskFileSystem {
     }
 
     #[turbo_tasks::function]
+    async fn track(&self, fs_path: FileSystemPathVc) -> Result<CompletionVc> {
+        let full_path = self.to_sys_path(fs_path).await?;
+        self.register_invalidator(&full_path, true);
+        Ok(CompletionVc::new())
+    }
+
+    #[turbo_tasks::function]
     async fn write(
         &self,
         fs_path: FileSystemPathVc,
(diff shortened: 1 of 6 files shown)


sokra (changes requested):
That would fix the loop, but it breaks writing the file again when switching back to the previous write.

Competing writes is the real problem we need to solve.

jridgewell:
@sokra: I've reworked this PR so that we still perform tracking of the written file, but now we do not suffer the memory bloat from reading the disk file into a Vc.

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -74,6 +74,9 @@ pub trait FileSystem: ValueToString {
     fn read(&self, fs_path: FileSystemPathVc) -> FileContentVc;
     fn read_link(&self, fs_path: FileSystemPathVc) -> LinkContentVc;
     fn read_dir(&self, fs_path: FileSystemPathVc) -> DirectoryContentVc;
+    fn track(&self, _fs_path: FileSystemPathVc) -> CompletionVc {
+        CompletionVc::immutable()
```
sokra, line 78: The default impl can call `read()` to track it.

sokra (approved):

crates/turbo-tasks-fs/src/lib.rs
sokra: This really need to be changed otherwise the default impl behaves incorrectly

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -474,17 +488,21 @@ impl FileSystem for DiskFileSystem {
...
-            .await
-            .with_context(|| format!("reading old content of {}", full_path.display()))?;
 
-        if *content == *old_content {
+        // Track the file, so that we will rewrite it if it ever changes.
+        fs_path.track().await?;
+
+        // We perform an untracked read here, so that this write is not dependent on the
+        // read value (and the memory it holds). It's possible the read can be freed if
+        // no other task tries to read it, which is entirely likely for a output file.
```
sokra, line 497: Why "possible"? This will always happen when `old_content` goes out of scope. It's impossible that anyone else will hold a reference to it since it will created just for us in `read_file`. We do not share the memory with anyone. Neither is rust a GC language.

jridgewell: This was a leftover comment from my first attempt at this, where I used `path.read().into_untracked().await?` directly. Because it was a Vc's value, it could be held by others that also `.read()`, but that's not the case with the current impl. Removed.

crates/turbo-tasks-fs/src/lib.rs
jridgewell: Done.

jridgewell: Actually, I'm removing this default impl entirely. The default gets us back into doubling the memory again (write -> track -> read means we have a write content and a read content in memory).

PR #3527, opened 2023-01-29
Doc for environment variables.

anthonyshew:
Folks have asked about how to handle environment variables. Our current recommendation is to use `dotenv-cli`.  A Turborepo-native solution for environment variables is something we are all looking forward to.


arlyon (approved):
This seems like a reasonable stopgap, however with global turbo available I'd also like to see this handled ootb, sinc (I assume) global turbo will not be able to respect this setting. Perhaps this is a target for 1.8 (or 1.9) @gsoltis?

PR #3529, opened 2023-01-30
feat(ci): published smoke tests

tknickman:
Automatically run smoke tests when new versions of Turborepo are published. 

This tests the following matrix: 

```
test: ["local", "global", "both"]
packageManager: ["yarn", "pnpm", "npm"]
localVersion: ["latest", "canary"]
globalVersion: ["latest", "canary"]
```

Where `test` described how `turbo` is installed (local is local install only, global is global install only, and both is both a local and global install. 

This effectively runs 36 tests ensuring our local turbo is cooperating with global turbo across all package managers, and install scenarios. 

Example run from development: https://github.com/tknickman/turborepo/actions/runs/4040066963

**TODO**: Slack notifications on failure

### DIFF
--- a/scripts/smoke-published.ts
+++ b/scripts/smoke-published.ts
@@ -1,0 +1,369 @@
+#!/usr/bin/env node
+
+import { execSync, ExecSyncOptions } from "child_process";
+
+type PackageManager = "yarn" | "npm" | "pnpm";
+type InstallType = "local" | "global";
+type VersionInput = {
+  version: string;
+  type: InstallType;
+};
+type Condition = {
+  expected: string;
+  operator: "includes" | "notIncludes" | "startsWith";
+};
+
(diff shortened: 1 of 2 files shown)


scripts/smoke-published.ts
```diff
@@ -0,0 +1,369 @@
...
+  };
+
+  console.log(
+    `Running test: "${testName}" with local version: "turbo@${localVersion}" and global version: turbo@${globalVersion} using ${packageManager}`
+  );
+  tests[testName]({ local, global, packageManager });
+  console.log("Tests passed!");
+}
+
+test();
```
arlyon, line 369: Does it make sense to check if this is the main module before running this? Might be useful in case others want to use this code.

arlyon (approved):
One minor thing.

PR #3531, opened 2023-01-30
Avoids emitting the same output from multiple processes of the same process pool

sokra:
It uses a `shared` set of `(line: [u8], occurences: u32)` to avoid emitting lines that has been emitted by other processes already.

It still supports emitting the same line multiple times from one process.

The `shared` data might be later used for showing logging from a worker pool. It contains the merged output of all processes.

Fixes WEB-489


crates/turbopack-node/src/pool.rs
```diff
@@ -55,11 +60,61 @@ impl Drop for RunningNodeJsPoolProcess {
...
+                let new_line = {
+                    let mut shared = shared.lock().unwrap();
+                    shared.insert((line.clone(), occurances))
+                };
+                if new_line {
+                    if final_stream.write(&*line).await.is_err() {
+                        // Whatever happened with stdout/stderr, we can't write to it anymore.
+                        break;
+                    }
+                }
```
ForsakenHarmony, line 101: Move this below the match, the other 2 cases break the loop already

crates/turbopack-node/src/pool.rs
```diff
@@ -55,11 +60,61 @@ impl Drop for RunningNodeJsPoolProcess {
...
+                    Entry::Occupied(mut e) => {
+                        let c = e.get_mut();
+                        *c += 1;
+                        *c
+                    }
+                    Entry::Vacant(e) => {
+                        e.insert(0);
+                        0
+                    }
+                };
```
ForsakenHarmony, line 91:
```suggestion
                let occurances = own_output.entry(Arc::clone(&line))
                    .and_modify(|e| { *e += 1 })
                    .or_insert(0);
```

Shouldn't it insert `1` though?

crates/turbopack-node/src/pool.rs
```diff
@@ -55,11 +60,61 @@ impl Drop for RunningNodeJsPoolProcess {
...
+                let new_line = {
+                    let mut shared = shared.lock().unwrap();
+                    shared.insert((line.clone(), occurances))
+                };
+                if new_line {
```
ForsakenHarmony, line 96: So if I'm understanding this correctly, a new line gets logged whenever there are more than the previous amount of occurrences?

This also means if one execution logs something 2 times and the next logs it 3 times, it will only output 1 time for the next execution? That seems like confusing behavior

sokra: Yeah, it could be a bit confusing in some edge cases... Do you have a better idea?

crates/turbopack-node/src/pool.rs
```diff
@@ -55,11 +60,53 @@ impl Drop for RunningNodeJsPoolProcess {
...
+            }
+            Err(err) => {
+                eprintln!("error reading from stream: {}", err);
+                break;
+            }
+            Ok(_) => {}
+        }
+        let line = Arc::from(take(&mut buffer).into_boxed_slice());
+        let occurance_number = *own_output
+            .entry(Arc::clone(&line))
```
jridgewell, line 87: Can we store a hash instead of the raw bytes? Wouldn't that save memory overall?

crates/turbopack-node/src/pool.rs
```diff
@@ -55,11 +60,53 @@ impl Drop for RunningNodeJsPoolProcess {
...
+            .and_modify(|c| *c += 1)
+            .or_insert(0);
+        let new_line = {
+            let mut shared = shared.lock().unwrap();
+            shared.insert((line.clone(), occurance_number))
```
jridgewell, line 92: Why do we need `occurance_number`? Wouldn't just a `HashSet` work?

jridgewell (changes requested):

crates/turbopack-node/src/pool.rs
sokra: I wanted to allow logging lines multiple times from a process.

crates/turbopack-node/src/pool.rs
sokra: That work work (with a small risk of collisions), but that would loose the ability to read the logged output from the browser ui in future.

crates/turbopack-node/src/pool.rs
```diff
@@ -55,11 +60,53 @@ impl Drop for RunningNodeJsPoolProcess {
...
+                break;
+            }
+            Err(err) => {
+                eprintln!("error reading from stream: {}", err);
+                break;
+            }
+            Ok(_) => {}
+        }
+        let line = Arc::from(take(&mut buffer).into_boxed_slice());
+        let occurance_number = *own_output
```
alexkirsz, line 88:
```suggestion
        let occurrence_number = *own_output
```

crates/turbopack-node/src/pool.rs
```diff
@@ -55,11 +60,53 @@ impl Drop for RunningNodeJsPoolProcess {
...
+            .entry(Arc::clone(&line))
+            .and_modify(|c| *c += 1)
+            .or_insert(0);
+        let new_line = {
+            let mut shared = shared.lock().unwrap();
+            shared.insert((line.clone(), occurance_number))
```
alexkirsz, line 94:
```suggestion
            shared.insert((line.clone(), occurrence_number))
```

crates/turbopack-node/src/pool.rs
```diff
@@ -55,11 +60,53 @@ impl Drop for RunningNodeJsPoolProcess {
...
+async fn handle_output_stream(
+    stream: impl AsyncRead + Unpin,
+    shared: SharedOutputSet,
+    mut final_stream: impl AsyncWrite + Unpin,
+) {
+    let mut buffered = BufReader::new(stream);
+    let mut own_output = HashMap::<Arc<[u8]>, u32>::new();
+    let mut buffer = Vec::new();
+    loop {
+        match buffered.read_until(b'\n', &mut buffer).await {
```
alexkirsz, line 77: Do we expect non-UTF-8 sequences here? Otherwise let's just use `buffered.lines()`.

crates/turbopack-node/src/pool.rs
```diff
@@ -55,11 +60,53 @@ impl Drop for RunningNodeJsPoolProcess {
...
+type SharedOutputSet = Arc<Mutex<IndexSet<(Arc<[u8]>, u32)>>>;
+
+/// Pipes the `stream` from `final_stream`, but uses `shared` to deduplicate
+/// lines that has beem emitted by other `handle_output_stream` instances with
+/// the same `shared` before.
+async fn handle_output_stream(
+    stream: impl AsyncRead + Unpin,
+    shared: SharedOutputSet,
+    mut final_stream: impl AsyncWrite + Unpin,
+) {
```
alexkirsz, line 72: If stream A reports:

```
error A:
   shared line
   non-shared line (A)
```

while stream B reports:

```
error B:
  shared line
  non-shared line (B)
```

We would end up with:

```
error A:
  shared line
  non-shared line (A)
error B:
  non-shared line (B)
```

I expect this to also happen with empty lines printed for formatting purposes.

Should we instead only dedupe the start of a stream, up until the first different character?

crates/turbopack-node/src/pool.rs
sokra: I had this before, but lines is weird since it doesn't include the line terminator, which requires adding it manually before piping which requires extending the Vec by one byte, which potentially copies the data... This read_until is better

ForsakenHarmony:
Why did you merge this with 0 approvals and 1 request for change?

PR #3535, opened 2023-01-30
Document global turbo package inference behavior

gsoltis:
- adds package inference to the docs on filtering workspaces
 - adds package inference to the `dev` topic in the handbook, which AFAICT is where we introduce the concept of filtering.


gsoltis:
Marked as on hold because we need `1.7.1` released first

docs/pages/repo/docs/handbook/dev.mdx
```diff
@@ -95,13 +95,22 @@ This means that users of your `dev` task _don't need to worry about codegen or m
 
 ## Running `dev` only in certain workspaces
 
-To run a `dev` task in only certain workspaces, you should use the [`--filter` syntax](/repo/docs/core-concepts/monorepos/filtering). For example:
+Let's assume you want to run the `dev` task in the `docs` workspace, located at `<root>/apps/docs`.
+With a global install of `turbo`, e.g. via `npm install --global turbo`, you can do
```
chris-olszewski, line 99: Should we use the `selected-pkg-manager` tab here to provide all of the possible install commands?

chris-olszewski (approved):
Only suggested change is providing multiple install instructions for global turbo

docs/pages/repo/docs/handbook/dev.mdx
gsoltis: Good idea, I'll look into it!

gsoltis: I actually decided to drop the package-manager specific part, including the reference to global turbo, in favor of just demonstrating it. I think we probably want the global turbo piece to live in a more explicit "installing turbo" section

gsoltis:
Dropping hold, `1.7.1` is live.

PR #3536, opened 2023-01-30
Note that path globs must be relative

gsoltis:
Paths in configuration are intended to work across machines, and thus must be relative to some anchor in the monorepo (`turbo.json` location or workspace where a task is being run).

### DIFF
--- a/docs/pages/repo/docs/reference/configuration.mdx
+++ b/docs/pages/repo/docs/reference/configuration.mdx
@@ -17,6 +17,11 @@ You can configure the behavior of `turbo` by adding a `turbo.json` file in your 
 A list of file globs for implicit global hash dependencies. The contents of these files will be included in the global hashing algorithm and affect the hashes of all tasks.
 This is useful for busting the cache based on `.env` files (not in Git) or any root level file that impacts workspace tasks (but are not represented in the traditional dependency graph (e.g. a root `tsconfig.json`, `jest.config.js`, `.eslintrc`, etc.)).
 
+<Callout type="info">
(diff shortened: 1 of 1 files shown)


docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -17,6 +17,11 @@ You can configure the behavior of `turbo` by adding a `turbo.json` file in your
 A list of file globs for implicit global hash dependencies. The contents of these files will be included in the global hashing algorithm and affect the hashes of all tasks.
 This is useful for busting the cache based on `.env` files (not in Git) or any root level file that impacts workspace tasks (but are not represented in the traditional dependency graph (e.g. a root `tsconfig.json`, `jest.config.js`, `.eslintrc`, etc.)).
 
+<Callout type="info">
+  Note that these must be relative paths from the location of `turbo.json`, and they should be valid for any machine where
```
tknickman, line 21: Minor nit, the `Callout` implies the "Note" in my opinion. But def not blocking

```suggestion
These must be relative paths from the location of `turbo.json`, and they should be valid for any machine where
```

tknickman (approved):

mehulkar (approved):
Looks good. Same is true for outputs/inputs, so another suggestion might be to change the type in the docs. E.g. `type: RelativePath[]` instead of `type: string[]`.

gsoltis:
@mehulkar I've filed a task internally to define relative path globs as a core concept, at which point I think it would be appropriate to update the types here and have them point to that document.

PR #3537, opened 2023-01-30
Add print_task_invalidation feature

jridgewell:
I used this to track down the infinite looping behavior #3525 and #3526. Running `cargo run --features print_task_invalidation --bin next-dev` will print something similar to:

```
invalidated Task { id: 16875, name: ConditionalContentSource::get }
Task { id: 16875, name: ConditionalContentSource::get } invalidates:
	Task { id: 16873, name: CombinedContentSource::get }
	Task { id: 16832, name: CombinedContentSource::get }
	Task { id: 16815, name: CombinedContentSource::get }
	Task { id: 16807, name: CombinedContentSource::get }
	Task { id: 15384, name: get_from_source }
invalidated Task { id: 16873, name: CombinedContentSource::get }
invalidated Task { id: 16832, name: CombinedContentSource::get }
invalidated Task { id: 16815, name: CombinedContentSource::get }
invalidated Task { id: 16807, name: CombinedContentSource::get }
invalidated Task { id: 15384, name: get_from_source }
invalidated Task { id: 71537, name: DiskFileSystem::read }
invalidated Task { id: 71517, name: DiskFileSystem::write }
invalidated Task { id: 72217, name: DiskFileSystem::read }
invalidated Task { id: 72214, name: DiskFileSystem::write }

### DIFF
--- a/crates/auto-hash-map/src/set.rs
+++ b/crates/auto-hash-map/src/set.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::hash_map::RandomState,
-    fmt::Debug,
+    fmt::{Debug, Display},
     hash::{BuildHasher, Hash},
     marker::PhantomData,
 };
@@ -18,6 +18,20 @@ impl<K, H> Default for AutoSet<K, H> {
     fn default() -> Self {
         Self {
             map: Default::default(),
+        }
+    }
+}
+
+impl<K: Display, H> Display for AutoSet<K, H> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.is_empty() {
+            write!(f, "{{}}")
+        } else {
+            write!(f, "{{ ")?;
+            for i in self.iter() {
+                Display::fmt(i, f)?;
+            }
+            write!(f, " }}")
         }
     }
 }
--- a/crates/auto-hash-map/src/set.rs
+++ b/crates/auto-hash-map/src/set.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::hash_map::RandomState,
-    fmt::{Debug, Display},
+    fmt::Debug,
     hash::{BuildHasher, Hash},
     marker::PhantomData,
 };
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: crates/auto-hash-map/src/set.rs [hunk 1]
```diff
@@ -22,6 +22,20 @@ impl<K, H> Default for AutoSet<K, H> {
...
+            write!(f, "{{}}")
+        } else {
+            write!(f, "{{ ")?;
+            for i in self.iter() {
+                Display::fmt(i, f)?;
+            }
+            write!(f, " }}")
+        }
+    }
+}
```
alexkirsz, line 37: Why implement `Display` instead of using `Debug`?

crates/turbo-tasks-memory/src/output.rs
```diff
@@ -60,32 +60,8 @@ impl Output {
...
-                    change = Some(target);
-                }
-            }
-            OutputContent::Empty | OutputContent::Error(_) | OutputContent::Panic(_) => {
-                change = Some(target);
-            }
-        };
-        if let Some(target) = change {
-            self.assign(OutputContent::Link(target), turbo_tasks)
-        }
```
alexkirsz, line 88: What does this fix or change?

jridgewell: This does 2 things:
1. Simplifies the equality
2. Removes the equality check from the `link` method

The simplification is just that the old code manually implemented equality, and it wasn't needed. It's much easier to just defer to the equality of the `RawVc` inside the `OutputContent::Link`.

The split out of the `link` method is necessary because I need to check if any invalidation happens in `crates/turbo-tasks-memory/src/task.rs` so that I can log while I still have access to the `Task`.

crates/auto-hash-map/src/set.rs
```diff
@@ -22,6 +22,20 @@ impl<K, H> Default for AutoSet<K, H> {
     }
 }
 
+impl<K: Display, H> Display for AutoSet<K, H> {
```
alexkirsz, line 25: Is this impl still needed? It looks like you're iterating over the set manually now in `execution_result`.

alexkirsz (approved):

crates/auto-hash-map/src/set.rs
jridgewell: You're right, done.

PR #3539, opened 2023-01-31
chore: add snapshot test for graph output

chris-olszewski:
Given discussion about the stability of the graph output I figured we should have a snapshot test to catch any changes.


mehulkar (approved):
We have some of this coverage in https://github.com/vercel/turbo/blob/main/cli/internal/core/engine_test.go also, which I think we need to convert to integration tests.

PR #3540, opened 2023-01-31
turbopackLoaders: accept basic loader options

wbinnssmith:
Fixes WEB-525

This allows webpack loaders defined by `experimental.turbopackLoaders` in `next.config.js` to pass serializable options to loaders.

It expands `turbopackLoaders` to accept not just loader names for a given extension, but also `{loader: string, options: Object}`. These options are verified to be serializable and passed to `loader-runner` and exposed as `getOptions()` to the loader.

Test Plan: Added an integration test for basic options. Verified the serializable check by passing `class Foo {}; new Foo()` as an option value and verified the error was displayed.


jridgewell (approved):

pete-willard:
Sorry for necroing a thread but I think it would be beneficial to reflect this change in the the example in [the docs](https://nextjs.org/docs/app/api-reference/next-config-js/turbo#webpack-loaders). Thanks for the change!

PR #3542, opened 2023-01-31
Wrap CSS chunk items in a @layer

alexkirsz:
In Turbopack, as a consequence of our lazy compilation model, CSS chunks can contain duplicate CSS chunk items. This can cause issues with precedence. Take the following example:

Initial CSS chunk:
```css
/* ... */

/* chunk item A */
h1 {
  font-size: 2rem;
}

/* ... */

/* other chunk item */
h1 {
  font-size: 4rem;
}

/* ... */
```

Dynamic CSS chunk (loaded after the first page load completes)
```css
/* ... */

/* chunk item A */
h1 {
  font-size: 2rem;
}

/* ... */
```

In this example, when the page first loads, the following rule will be applied:
```css
h1 {
  font-size: 4rem;
}
```

But as soon as the dynamic CSS chunk loads, the following rule will be applied instead:
```css
h1 {
  font-size: 2rem;
}
```

However, from the order of rules in the initial load, we know that the former should still apply.

We can remedy this particular issue by wrapping each CSS chunk item into its own [`@layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) (thanks @sokra for the idea!). This ensures that when a CSS chunk item is re-encountered at a later time, it is automatically de-duplicated thanks to the inherent CSS layering algorithm.

This is not an issue in Next.js as we can't have duplicated CSS chunk items.


crates/turbopack-css/src/chunk/mod.rs
```diff
@@ -345,6 +346,23 @@ impl CssChunkContextVc {
...
+        let mut s = chunk_item.to_string().await?.clone_value();
+        if !layer.is_empty() {
+            if s.ends_with(')') {
+                s.pop();
+                write!(s, ", {layer})")?;
+            } else {
+                write!(s, " ({layer})")?;
+            }
+        }
+        Ok(ModuleId::String(s).cell())
```
alexkirsz, line 364: This implementation is shared with `EcmascriptChunkContextVc`.

alexkirsz:
Adding this comment from @sokra if another issue arises in the future:
> Assuming the order in one chunk is correct (it's probably not, but that's something we can fix).
And assuming the order in which chunks of one chunk group are loaded is correct (it's probably not due to optimization, the optimization need to take order into account).
We can put every css chunk item in a @layer named by it's module id (we need to add module ids for CSS for that, similar to ecmascript).
Just by doing that, we magically will have correct ordering. But I think it's worth explaining why.
When using @layer the first time it will add this layer to the global order of layers. The next time you use the same layer name, it will place the css into that existing layer.
So once we load a duplicate CSS chunk item, it will place it into the existing layer, which is earlier in the order.

crates/next-dev-tests/tests/integration/next/font-google/basic/pages/index.js
```diff
@@ -67,13 +66,35 @@ async function getRuleMatchingClassName(className) {
...
+  for (const rule of rules) {
+    if (rule instanceof CSSStyleRule && rule.selectorText === selector) {
+      return rule;
+    }
+
+    if (rule instanceof CSSLayerBlockRule) {
+      const res = getRuleMatchingClassNameRec(selector, rule.cssRules);
+      if (res != null) {
+        return res;
+>>>>>>> 4c1c09ba1 (Fix tests)
```
wbinnssmith, line 94: merge marker 🙂

alexkirsz: Whoops 👻

sokra (approved):
Sounds good. We should keep in mind that this currently breaks users using `@layer`. We can later fix that by moving our `@layer` into the user layer.

crates/next-dev-tests/tests/integration/next/font-google/basic/pages/index.js
```diff
@@ -67,13 +66,34 @@ async function getRuleMatchingClassName(className) {
       });
     }
 
-    for (const rule of stylesheet.sheet.cssRules) {
-      if (rule.selectorText === selector) {
-        matchingRule = rule;
-        break;
+    const sheet = stylesheet.sheet;
+
+    if (sheet == null) {
```
wbinnssmith, line 71: can sheet be null if we've waited for the stylesheet to load above? Should this just be an invariant?

wbinnssmith (approved):

crates/next-dev-tests/tests/integration/next/font-google/basic/pages/index.js
alexkirsz: This check came back with the merge, but it should no longer be necessary indeed.

PR #3550, opened 2023-01-31
improve error reporting for fatal errors

sokra:
This adds the turbo-tasks function name as `context` of all fatal errors propagated from executing or reading turbo-tasks functions.

That should help in debugging fatal errors.


ForsakenHarmony (approved):

cli/cmd/turbo/version.go
```diff
@@ -1,3 +1,3 @@
 package main
 
-const turboVersion = "1.7.1"
+const turboVersion = "1.7.1-canary.7"
```
mehulkar, line 3: is this a legit change? I think you may need to pull in main?

mehulkar (commented):
Guessing you need to pull in main. Bunch of turborepo tests being deleted and version number changing

sokra:
> Guessing you need to pull in main. Bunch of turborepo tests being deleted and version number changing

yeah, something weird happened here... should be fixed now...

PR #3552, opened 2023-01-31
include omit entries in hash

sokra:


crates/turbopack-ecmascript/src/chunk/mod.rs
```diff
@@ -977,6 +977,8 @@ impl Asset for EcmascriptChunk {
                 hasher.write_value(id.await?);
                 need_hash = true;
             }
+            // Marker to end evaluate
+            hasher.write_value(true);
```
jridgewell, line 981: Should this have written a length header instead of a bool footer?

jridgewell (approved):

crates/turbopack-ecmascript/src/chunk/mod.rs
alexkirsz: Both `evaluate` and `omit` could write length headers (or `0` when N/A).

PR #3431, opened 2023-01-23
Combine TaskSummary with RunSpec for 🪄

mehulkar:
### Dry Runs
- [x] #3377
- [x] https://github.com/vercel/turbo/pull/4095
- [x] #3957
- [x] #3934
- [x] https://github.com/vercel/turbo/pull/4020
- [x] https://github.com/vercel/turbo/pull/4090
- [x] https://github.com/vercel/turbo/pull/3941
- [ ] hashes of dependencies (will have to decide whether to do this at a package or task level)
- [x] https://github.com/vercel/turbo/pull/4047
- [ ] https://github.com/vercel/turbo/pull/4165

### Real Runs
- [x] #4069
- [x] https://github.com/vercel/turbo/pull/4116
- [x] https://github.com/vercel/turbo/pull/4165
- [ ] exit codes per task
- [ ] parent task exit code

## Open Qs
- [ ] Can we output exitCodes, expandedOutputs on dry runs if there was a cache hit?
- [ ] Should we separate env vars that are included by default from turbo?
- [ ] group `expandedInputs` and `expandedOutputs` by their "source" (glob, filename, turbo)

### Pre-work
- https://github.com/vercel/turbo/pull/3555
- https://github.com/vercel/turbo/pull/4052

PR #3555, opened 2023-01-31
Move helpers for getting task ancestors/descendants to engine

mehulkar:
This starts to clean up / consolidate the differences between the information Dry Run gathers and the information RealRun has, so we can consolidate to a RunSummary. Extracted from #3431.

### DIFF
--- a/cli/internal/core/engine.go
+++ b/cli/internal/core/engine.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/vercel/turbo/cli/internal/fs"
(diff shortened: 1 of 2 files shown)


cli/internal/core/engine.go
```diff
@@ -383,3 +384,41 @@ func (e *Engine) ValidatePersistentDependencies(graph *graph.CompleteGraph) erro
...
+		// Don't leak out internal ROOT_NODE_NAME nodes, which are just placeholders
+		if !strings.Contains(dep.(string), ROOT_NODE_NAME) {
+			stringAncestors = append(stringAncestors, dep.(string))
+		}
+	}
+	// TODO(mehulkar): Why are ancestors not sorted, but GetTaskGraphDescendants sorts?
+	return stringAncestors, nil
+}
+
+// GetTaskGraphDescendants gets all the ancestors for a given task in the graph.
```
arlyon, line 407: Should this be 'gets all the descendants'?

arlyon (approved):

cli/internal/core/engine.go
mehulkar: it absolutely should. good catch 🤦🏾

PR #3558, opened 2023-01-31
chore: Removed `Cargo.lock` from CI for PRs

NicholasLYang:


.github/workflows/test.yml
```diff
@@ -57,17 +57,25 @@ jobs:
...
+            rust-toolchain
+
       - name: Rust related changes
         id: rust
         uses: technote-space/get-diff-action@v6
         with:
           PATTERNS: |
             pnpm-lock.yaml
             package.json
+            Cargo.**
```
mehulkar, line 75: Is this still supposed to be here?

NicholasLYang: Yeah because there are more general Rust jobs that should run if `Cargo.lock` changes.

sokra (approved):

PR #3562, opened 2023-01-31
Load TaskDefinitions lazily when building taskGraph

mehulkar:
Previously, before the `Engine` struct is Prepared for execution, Turbo would store a map of `core.Task` instances so it could look up Task Definitions. This change removes that behavior and instead looks up `TaskDefinition`s during task graph construction. This bakes in the assumption that there is no longer _one_ place for Tasks to be defined (which is a pre-requisite for composable configs). This implementation hardcodes looking up the `turbo.json` from the root workspace, but in a future PR, we'll be able to update this to use the Workspace and start looking for task definitions there. 

Pre-reqs:
- https://github.com/vercel/turbo/pull/3566
- https://github.com/vercel/turbo/pull/3592
- https://github.com/vercel/turbo/pull/3593

### DIFF
--- a/cli/integration_tests/single_package/dry-run.t
+++ b/cli/integration_tests/single_package/dry-run.t
@@ -8,7 +8,7 @@ Check
   Tasks to Run
   build
     Task                   = build                                                                                                       
-    Hash                   = 7bf32e1dedb04a5d                                                                                            
+    Hash                   = 46d68c621aadbb40                                                                                            
     Cached (Local)         = false                                                                                                       
     Cached (Remote)        = false                                                                                                       
     Command                = echo 'building' > foo                                                                                       
@@ -23,7 +23,7 @@ Check
     "tasks": [
       {
         "task": "build",
-        "hash": "7bf32e1dedb04a5d",
+        "hash": "46d68c621aadbb40",
         "cacheState": {
           "local": false,
           "remote": false
--- a/cli/integration_tests/single_package/dry-run.t
+++ b/cli/integration_tests/single_package/dry-run.t
@@ -8,7 +8,7 @@ Check
   Tasks to Run
   build
     Task                   = build                                                                                                       
-    Hash                   = 46d68c621aadbb40                                                                                            
+    Hash                   = 7bf32e1dedb04a5d                                                                                            
     Cached (Local)         = false                                                                                                       
     Cached (Remote)        = false                                                                                                       
     Command                = echo 'building' > foo                                                                                       
@@ -23,7 +23,7 @@ Check
     "tasks": [
       {
         "task": "build",
-        "hash": "46d68c621aadbb40",
+        "hash": "7bf32e1dedb04a5d",
         "cacheState": {
           "local": false,
           "remote": false
--- a/cli/internal/context/context.go
+++ b/cli/internal/context/context.go
@@ -362,7 +362,6 @@ func (c *Context) parsePackageJSON(repoRoot turbopath.AbsoluteSystemPath, pkgJSO
 			existing := c.WorkspaceInfos[pkg.Name]
 			return fmt.Errorf("Failed to add workspace \"%s\" from %s, it already exists at %s", pkg.Name, pkg.Dir, existing.Dir)
 		}
-
 		c.WorkspaceInfos[pkg.Name] = pkg
 		c.WorkspaceNames = append(c.WorkspaceNames, pkg.Name)
 	}
--- a/cli/internal/context/context.go
+++ b/cli/internal/context/context.go
@@ -356,12 +356,13 @@ func (c *Context) parsePackageJSON(repoRoot turbopath.AbsoluteSystemPath, pkgJSO
 			return fmt.Errorf("parsing %s: %w", pkgJSONPath, err)
 		}
 
+		// TODO(mehulkar): Figure out if adding duplicate names to the WorkspaceGraph is a problem.
+		c.WorkspaceGraph.Add(pkg.Name)
 		if c.WorkspaceInfos[pkg.Name] != nil {
 			existing := c.WorkspaceInfos[pkg.Name]
 			return fmt.Errorf("Failed to add workspace \"%s\" from %s, it already exists at %s", pkg.Name, pkg.Dir, existing.Dir)
 		}
 
-		c.WorkspaceGraph.Add(pkg.Name)
 		c.WorkspaceInfos[pkg.Name] = pkg
 		c.WorkspaceNames = append(c.WorkspaceNames, pkg.Name)
 	}
--- a/cli/internal/context/context.go
+++ b/cli/internal/context/context.go
@@ -360,8 +360,6 @@ func (c *Context) parsePackageJSON(repoRoot turbopath.AbsoluteSystemPath, pkgJSO
 		if err != nil {
 			return err
 		}
-
-		// TODO(mehulkar): Figure out if adding duplicate names to the WorkspaceGraph is a problem.
 		c.WorkspaceGraph.Add(pkg.Name)
 		pkg.PackageJSONPath = turbopath.AnchoredSystemPathFromUpstream(relativePkgJSONPath)
 		pkg.Dir = turbopath.AnchoredSystemPathFromUpstream(filepath.Dir(relativePkgJSONPath))
--- a/cli/internal/context/context.go
+++ b/cli/internal/context/context.go
@@ -351,13 +351,20 @@ func (c *Context) parsePackageJSON(repoRoot turbopath.AbsoluteSystemPath, pkgJSO
 	defer c.mutex.Unlock()
 
 	if pkgJSONPath.FileExists() {
-		pkg, err := fs.ReadPackageJSON(repoRoot, pkgJSONPath)
+		pkg, err := fs.ReadPackageJSON(pkgJSONPath)
 		if err != nil {
 			return fmt.Errorf("parsing %s: %w", pkgJSONPath, err)
 		}
 
+		relativePkgJSONPath, err := repoRoot.PathTo(pkgJSONPath)
+		if err != nil {
+			return err
+		}
+
 		// TODO(mehulkar): Figure out if adding duplicate names to the WorkspaceGraph is a problem.
 		c.WorkspaceGraph.Add(pkg.Name)
+		pkg.PackageJSONPath = turbopath.AnchoredSystemPathFromUpstream(relativePkgJSONPath)
+		pkg.Dir = turbopath.AnchoredSystemPathFromUpstream(filepath.Dir(relativePkgJSONPath))
 		if c.WorkspaceInfos[pkg.Name] != nil {
 			existing := c.WorkspaceInfos[pkg.Name]
 			return fmt.Errorf("Failed to add workspace \"%s\" from %s, it already exists at %s", pkg.Name, pkg.Dir, existing.Dir)
--- a/cli/internal/context/context.go
+++ b/cli/internal/context/context.go
@@ -351,22 +351,17 @@ func (c *Context) parsePackageJSON(repoRoot turbopath.AbsoluteSystemPath, pkgJSO
 	defer c.mutex.Unlock()
 
 	if pkgJSONPath.FileExists() {
-		pkg, err := fs.ReadPackageJSON(pkgJSONPath)
+		pkg, err := fs.ReadPackageJSON(repoRoot, pkgJSONPath)
 		if err != nil {
 			return fmt.Errorf("parsing %s: %w", pkgJSONPath, err)
 		}
 
-		relativePkgJSONPath, err := repoRoot.PathTo(pkgJSONPath)
-		if err != nil {
-			return err
-		}
-		c.WorkspaceGraph.Add(pkg.Name)
-		pkg.PackageJSONPath = turbopath.AnchoredSystemPathFromUpstream(relativePkgJSONPath)
-		pkg.Dir = turbopath.AnchoredSystemPathFromUpstream(filepath.Dir(relativePkgJSONPath))
 		if c.WorkspaceInfos[pkg.Name] != nil {
 			existing := c.WorkspaceInfos[pkg.Name]
 			return fmt.Errorf("Failed to add workspace \"%s\" from %s, it already exists at %s", pkg.Name, pkg.Dir, existing.Dir)
 		}
+
+		c.WorkspaceGraph.Add(pkg.Name)
 		c.WorkspaceInfos[pkg.Name] = pkg
 		c.WorkspaceNames = append(c.WorkspaceNames, pkg.Name)
 	}
--- a/cli/internal/core/engine.go
+++ b/cli/internal/core/engine.go
@@ -67,6 +67,61 @@ type EngineBuildingOptions struct {
 	TasksOnly bool
 }
 
+// EngineExecutionOptions controls a single walk of the task graph
+type EngineExecutionOptions struct {
+	// Parallel is whether to run tasks in parallel
+	Parallel bool
+	// Concurrency is the number of concurrent tasks that can be executed
+	Concurrency int
+}
+
+// Execute executes the pipeline, constructing an internal task graph and walking it accordingly.
+func (e *Engine) Execute(visitor Visitor, opts EngineExecutionOptions) []error {
+	var sema = util.NewSemaphore(opts.Concurrency)
+	return e.TaskGraph.Walk(func(v dag.Vertex) error {
+		// Each vertex in the graph is a taskID (package#task format)
+		taskID := dag.VertexName(v)
+
+		// Always return if it is the root node
+		if strings.Contains(taskID, ROOT_NODE_NAME) {
+			return nil
+		}
+
+		// Acquire the semaphore unless parallel
+		if !opts.Parallel {
+			sema.Acquire()
+			defer sema.Release()
+		}
+
+		return visitor(taskID)
+	})
+}
+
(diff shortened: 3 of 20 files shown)


cli/internal/core/engine.go
```diff
@@ -107,31 +98,54 @@ func (e *Engine) Execute(visitor Visitor, opts EngineExecutionOptions) []error {
 }
 
 func (e *Engine) getTaskDefinition(taskName string, taskID string) (*Task, error) {
-	if task, ok := e.Tasks[taskID]; ok {
-		return task, nil
+	pipeline, err := e.getPipelineFromWorkspace(util.RootPkgName)
```
mehulkar, line 101: Hardcoding the lookup from root pipeline, but we'll expand it to starting in the actual packageTask's workspace.

cli/internal/context/context.go
```diff
@@ -357,18 +357,13 @@ func (c *Context) parsePackageJSON(repoRoot turbopath.AbsoluteSystemPath, pkgJSO
...
 		}
 
-		relativePkgJSONPath, err := repoRoot.PathTo(pkgJSONPath)
-		if err != nil {
-			return err
-		}
+		// TODO(mehulkar): Figure out if adding duplicate names to the WorkspaceGraph is a problem.
 		c.WorkspaceGraph.Add(pkg.Name)
-		pkg.PackageJSONPath = turbopath.AnchoredSystemPathFromUpstream(relativePkgJSONPath)
-		pkg.Dir = turbopath.AnchoredSystemPathFromUpstream(filepath.Dir(relativePkgJSONPath))
```
mehulkar, line 371: Assign this during unmarshalling, so the root package.json can also get these props set (we need the Dir so we can look up turbo config for the root workspace)

cli/internal/core/engine.go
```diff
@@ -272,19 +286,14 @@ func (e *Engine) generateTaskGraph(pkgs []string, taskNames []string, tasksOnly
...
+// AddTask adds root tasks to the engine so they can be looked up later.
+func (e *Engine) AddTask(taskName string) {
+	if util.IsPackageTask(taskName) {
+		pkg, taskName := util.GetPackageTaskFromId(taskName)
 		if pkg == util.RootPkgName {
 			e.rootEnabledTasks.Add(taskName)
 		}
 	}
-
-	e.Tasks[task.Name] = task
```
mehulkar, line 286: this was the "up front" store of known tasks. We resolve this lazily now as we go through packageTasks.

cli/internal/core/engine.go
```diff
@@ -58,24 +67,6 @@ type EngineBuildingOptions struct {
 	TasksOnly bool
 }
 
-// Prepare constructs the Task Graph for a list of packages and tasks
-func (e *Engine) Prepare(options *EngineBuildingOptions) error {
```
mehulkar, line 62: Renamed `generateTaskGraph()` to `Prepare()` , since this is an empty shell of a function now.

cli/internal/core/engine.go
```diff
@@ -58,24 +67,6 @@ type EngineBuildingOptions struct {
...
-	pkgs := options.Packages
-	tasks := options.TaskNames
-	if len(tasks) == 0 {
-		// TODO(gsoltis): Is this behavior used?
-		for key := range e.Tasks {
-			tasks = append(tasks, key)
-		}
-	}
```
mehulkar, line 70: This functionality is removed. As far as I can tell, this code would run _all_ known tasks if `turbo run` is run without any other arguments. But we're already catching this before this point, so this could should never be hit. cc @gsoltis let me know if you know any other history here.

cli/internal/lockfile/npm_lockfile.go
```diff
@@ -103,7 +103,18 @@ var _ Lockfile = (*NpmLockfile)(nil)
...
-	_, ok := l.Packages[workspacePath.ToString()]
+	// workspacePath comes from Turbo's understanding of the directory a workspace lives in.
+	// This is the relative path from the root of the repo.
+	// The root package directory's path is ".", but the NpmLockfile (i.e. package-lock.json)
+	// stores this "package" as an empty string (""). So when we want to look up this package
+	// we need to correct the uncoming workspacePath from "." to "".
+	workspacePathCorrected := workspacePath.ToString()
+	if workspacePathCorrected == "." {
+		workspacePathCorrected = ""
+	}
```
mehulkar, line 114: I _think_ this is the right fix, but cc @chris-olszewski , in case you have any other suggestions. Only the npm integration test was failing, so I checked the other package managers, and it looks like they _do_ use `"."` as the specifier for the root package, so this correction is only needed for npm?

- pnpm: https://github.com/vercel/turbo/blob/c164ab77b382fd50910bf314154556326e4618a5/cli/integration_tests/lockfile_aware_caching/pnpm/pnpm-lock.yaml#L5
- berry: https://github.com/vercel/turbo/blob/c164ab77b382fd50910bf314154556326e4618a5/cli/integration_tests/lockfile_aware_caching/berry/yarn.lock#L38
- yarn: doesn't seem to have the monorepo packages at all: https://github.com/vercel/turbo/blob/c164ab77b382fd50910bf314154556326e4618a5/cli/integration_tests/lockfile_aware_caching/yarn/yarn.lock

cli/internal/context/context.go
```diff
@@ -357,18 +357,13 @@ func (c *Context) parsePackageJSON(repoRoot turbopath.AbsoluteSystemPath, pkgJSO
...
+		pkg, err := fs.ReadPackageJSON(repoRoot, pkgJSONPath)
 		if err != nil {
 			return fmt.Errorf("parsing %s: %w", pkgJSONPath, err)
 		}
 
-		relativePkgJSONPath, err := repoRoot.PathTo(pkgJSONPath)
-		if err != nil {
-			return err
-		}
+		// TODO(mehulkar): Figure out if adding duplicate names to the WorkspaceGraph is a problem.
```
chris-olszewski, line 365: Should we open up a task for this investigation?

cli/internal/core/engine.go
```diff
@@ -422,3 +431,45 @@ func (e *Engine) GetTaskGraphDescendants(taskID string) ([]string, error) {
...
+	}
+
+	// TODO(mehulkar) what is this value when it's a non-root workspace?
+	dir := e.completeGraph.WorkspaceInfos[workspaceName].Dir
+
+	var pkgJSON *fs.PackageJSON
+	repoRoot := e.completeGraph.RepoRoot
+
+	// dirAbsolute := turbopath.AbsoluteSystemPath(dir)
+	dirAbsolutePath := repoRoot.UntypedJoin(dir.ToString())
```
chris-olszewski, line 448:
```suggestion
	dirAbsolutePath := dir.RestoreAnchor(repoRoot)
```

cli/internal/core/engine.go
```diff
@@ -422,3 +431,45 @@ func (e *Engine) GetTaskGraphDescendants(taskID string) ([]string, error) {
...
 	return stringDescendents, nil
 }
+
+func (e *Engine) getPipelineFromWorkspace(workspaceName string) (*fs.Pipeline, error) {
+	cachedPipeline, ok := e.pipelines[workspaceName]
+	if ok {
+		return cachedPipeline, nil
+	}
+
+	// TODO(mehulkar) what is this value when it's a non-root workspace?
```
chris-olszewski, line 441: Is this TODO still needed?

cli/internal/core/engine.go
```diff
@@ -422,3 +431,45 @@ func (e *Engine) GetTaskGraphDescendants(taskID string) ([]string, error) {
...
+	dir := e.completeGraph.WorkspaceInfos[workspaceName].Dir
+
+	var pkgJSON *fs.PackageJSON
+	repoRoot := e.completeGraph.RepoRoot
+
+	// dirAbsolute := turbopath.AbsoluteSystemPath(dir)
+	dirAbsolutePath := repoRoot.UntypedJoin(dir.ToString())
+
+	// We need to get a package.json, because turbo.json can "synthesize" tasks from it
+	// for single-package repos.
```
chris-olszewski, line 451: Does the root `package.json` still need to be read for monorepos?

chris-olszewski (approved):
Overall looks good. 

Biggest concern is any fallout from changing `PackageJSON.Dir` from `""` to `"."` for the root `package.json`. I couldn't find any additional places where we were checking against "", but it's still something to be aware of.

cli/internal/fs/package_json.go
```diff
@@ -59,12 +60,30 @@ func (r *Workspaces) UnmarshalJSON(data []byte) error {
...
+	relativePkgJSONPath, err := repoRoot.PathTo(pkgJSONFilePath)
+	if err != nil {
+		return nil, err
+	}
+
+	// assign PackageJSONPath and Dir
+	// TODO(mehulkar) These could go into WorkspaceInfos or some other Workspace construct
+	// rather than onto the fs.PackageJSON itself (which should just be the unmarshaled contents of package.json)
+	pkg.PackageJSONPath = turbopath.AnchoredSystemPathFromUpstream(relativePkgJSONPath)
+	pkg.Dir = turbopath.AnchoredSystemPathFromUpstream(filepath.Dir(relativePkgJSONPath))
```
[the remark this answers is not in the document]
mehulkar: I don't love it either, but I also felt it was weird that the `fs.PackageJSON` struct sometimes has these fields and sometimes doesn't. Is the TODO I left here enough to indicate that we should come back to it? I can go the route of the functional options, but I think that will reach into quite a few files. 

Another option is to manually set this for the root workspace:

https://github.com/vercel/turbo/blob/f1fc4bdb74940d1c6eb3b1b0ee10308e7937045a/cli/internal/run/run.go#L158-L159

like it's being set for other workspaces:

https://github.com/vercel/turbo/blob/f1fc4bdb74940d1c6eb3b1b0ee10308e7937045a/cli/internal/context/context.go#L370-L371

The problem with this is then I would not touch anything in prune, and it could become confusing why the Root Workspace is different for the `run` and `prune` and other commands.

cli/internal/context/context.go
mehulkar: Will do. in practice it probably doesn't matter since the next line will throw an error, but I didn't want to change this behavior in this PR

cli/internal/core/engine.go
mehulkar: removed it. i know the answer now, but it's still something to grapple with when I want to look up configs in other, non-root workspaces.

cli/internal/core/engine.go
mehulkar: Yes, but only because it logs a warning when it sees a `turbo` key in there (which was the old way to define the pipeline). we've removed support for that already, but there's still a warning. I'd like to drop that warning eventually, and when that happens we can have two methods, one that synthesizes (and needs package.json), and one that doesn't

cli/internal/lockfile/npm_lockfile.go
mehulkar: Remove one of these instances in the `pnpm_lockfile.go` and didn't see any others

chris-olszewski:
Removing on hold as pnpm changes were the only thing that needed to happen in this PR. Rest are fine as future work

cli/internal/core/engine.go
gsoltis: I think it was used in a test at one point, but if the tests are passing, then it's fine to nuke.

cli/integration_tests/single_package/dry-run.t
```diff
@@ -8,7 +8,7 @@ Check
   Tasks to Run
   build
     Task                   = build                                                                                                       
-    Hash                   = 7bf32e1dedb04a5d                                                                                            
+    Hash                   = 46d68c621aadbb40
```
gsoltis, line 11: Can you describe what it is that is causing these hashes to change? Do we need to bump the global cache key?

cli/internal/fs/package_json.go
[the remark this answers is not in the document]
mehulkar: I think we'd have to do it everywhere. I do want a use case to use that pattern, but at the moment I'd rather get this merged and get back to #2706 so we can keep 1.8 release on schedule

cli/integration_tests/single_package/dry-run.t
mehulkar: It's because the Workspace Directories are part of the task hash, and in this case, the Root Workspace changed it's representation of a Directory (from `""` to `"."`)

https://github.com/vercel/turbo/blob/bbdb5e592cae2a557a4c6adfb423216d461bc09d/cli/internal/taskhash/taskhash.go#L297

<details>
<summary>
Only the Single Package hashes were changed here, since their workspace is the root directory. Root Tasks would also bust their cache,but I'm guessing we don't have any dry run tests for those yet.
</summary>

```
$ git d main --stat integration_tests/                                       
cli/integration_tests/single_package/dry-run.t           |  4 ++--
cli/integration_tests/single_package/run.t               |  4 ++--
cli/integration_tests/single_package_deps/dry-run.t      |  8 ++++----
cli/integration_tests/single_package_deps/run.t          | 12 ++++++------
cli/integration_tests/single_package_no_config/dry-run.t |  4 ++--
cli/integration_tests/single_package_no_config/run.t     |  4 ++--
6 files changed, 18 insertions(+), 18 deletions(-)
```
</details>

I was a little worried about this too and meant to ask if we have a policy around cache busting in minor releases (another reason why I feel like a release branch would be a good idea, btw).

> Do we need to bump the global cache key?

What did you mean by this? I know [there's a key](https://github.com/vercel/turbo/blob/bbdb5e592cae2a557a4c6adfb423216d461bc09d/cli/internal/run/global_hash.go#L20), but what would bumping it do in this situation?

ForsakenHarmony:
rebase gone wrong?

mehulkar:
> rebase gone wrong?

"Update Branch" from UI and then `git merge main` from local. Since both merges happened at different states of `main`, I think that aded a bunch of commits. Rebased on top of main now. Sorry for the noise!

cli/integration_tests/single_package/dry-run.t
mehulkar: Reverted the thing that made the taskhash change

cli/internal/core/engine.go
```diff
@@ -27,24 +27,33 @@ type Visitor = func(taskID string) error
...
+	TaskGraph        *dag.AcyclicGraph
 	PackageTaskDeps  map[string][]string
 	rootEnabledTasks util.Set
 
 	// completeGraph is the CompleteGraph. We need this to look up the Pipeline, etc.
 	completeGraph *graph.CompleteGraph
+
+	// Map of packageName to pipeline. We resolve task definitions from here
+	// but we don't want to read from the filesystem every time
+	pipelines map[string]*fs.Pipeline
```
gsoltis, line 39: Does this need to be `*fs.Pipeline` instead of just `fs.Pipeline`? `map` (the underlying type of `fs.Pipeline`) is a reference type, so I think you should be fine to pass it by value, right?

cli/internal/core/engine.go
```diff
@@ -422,3 +431,45 @@ func (e *Engine) GetTaskGraphDescendants(taskID string) ([]string, error) {
 	sort.Strings(stringDescendents)
 	return stringDescendents, nil
 }
+
+func (e *Engine) getPipelineFromWorkspace(workspaceName string) (*fs.Pipeline, error) {
```
gsoltis, line 435: I think you should be able to return `fs.Pipeline` without the pointer

cli/internal/core/engine.go
```diff
@@ -422,3 +431,45 @@ func (e *Engine) GetTaskGraphDescendants(taskID string) ([]string, error) {
...
+
+	var pkgJSON *fs.PackageJSON
+	repoRoot := e.completeGraph.RepoRoot
+
+	dirAbsolutePath := dir.RestoreAnchor(repoRoot)
+
+	// We need to get a package.json, because turbo.json can "synthesize" tasks from it
+	// for single-package repos.
+	if workspaceName == util.RootPkgName {
+		var err error
```
gsoltis, line 453: Not a huge deal, but usually having to declare `err` like this is an indication that it would be better to use a different temp variable elsewhere. In this case it would be something instead of `pkgJSON`. The reason is that accidentally shadowing or reusing `err` is a somewhat common source of Go logic errors, so you almost never want to deviate from the standard pattern of `myVar, err := ...`

cli/internal/run/run.go
```diff
@@ -268,14 +272,19 @@ func (r *run) run(ctx gocontext.Context, targets []string) error {
...
+		r.opts.runOpts.singlePackage,
+	)
 
 	if err != nil {
 		return errors.Wrap(err, "error preparing engine")
 	}
 	tracker := taskhash.NewTracker(
 		g.RootNode,
 		g.GlobalHash,
+		// TODO(mehulkar): remove g,Pipeline, because we need to get task definitions from CompleteGaph instead
```
gsoltis, line 287: > need

What is the need? Are there any implications for not doing this now? Or is it just a code cleanup thing?

gsoltis (approved):

cli/internal/run/run.go
mehulkar: Th is passed in so the taskhash can include the task definition. With composable, we don't want to use the root definition, we'll want to use the _resolved_ definition. 

Since this work included reducing where / how we use `g.Pipeline` (i.e. the root pipeline), I was leaving a paper trail for myself for other usages of the root Pipeline that will disappear. [The compiler doesn't protect me from unused arguments](https://play.golang.com/p/lhzeLpiKsDj), so I wanted to be sure that I stopped passing it in when the `taskhash` internals are updated.

PR #3572, opened 2023-02-01
ci(renovate): trying to enable swc_* updates

kwonoj:
Trying to progress WEB-510. PR updates renovate config to include specific dependencies we want to get an update.

This might need some verification once it lands.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -18,6 +18,11 @@
     {
       "matchLanguages": ["golang"],
       "enabled": false
+    },
+    {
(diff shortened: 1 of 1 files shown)


ForsakenHarmony:
prettier is broken, do your git hooks not work @kwonoj ?

kwonoj:
> do your git hooks not work

looks like so? not sure why.

kwonoj:
https://github.com/vercel/turbo/pull/3575 for quick amend.

PR #3574, opened 2023-02-01
Export __N_SSG and __N_SSP from the error component

alexkirsz:
`export *` would be preferable once supported.


crates/next-dev-tests/tests/integration/next/app/metadata/input/next.config.js
```diff
@@ -2,4 +2,5 @@ module.exports = {
   experimental: {
     appDir: true,
   },
+  transpilePackages: ["@turbo/pack-test-harness"],
```
alexkirsz, line 5: Any idea why I need to include this for this test, but it's not necessary for the 404 test?

sokra: 🤷

sokra (approved):

crates/next-dev-tests/tests/integration/next/app/metadata/input/next.config.js
alexkirsz: This was because these two cases have a `tsconfig.json` that overrides the top-level `tsconfig.json`, which includes the test harness in its paths list (hence it's compiled without needing transpilePackages).

PR #3582, opened 2023-02-01
fix: Add rustls-tls as default feature

NicholasLYang:
Might as well have this as a default feature to avoid accidental runtime errors.

### DIFF
--- a/crates/turborepo/Cargo.toml
+++ b/crates/turborepo/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 license = "MPL-2.0"
 
 [features]
+default = ["rustls-tls"]
 native-tls = ["turborepo-lib/native-tls"]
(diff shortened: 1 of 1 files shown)


kwonoj (changes requested):
Copying slack comment for the visiblity:

>We can't, as said in the comment https://github.com/vercel/turbo/blob/main/Cargo.toml#L103
this'll break next-swc's platform support since cargo will merge all of default enabled features even if we can't build.
Reason we explicitly opt in for specific and then have buidl time checker to not to allow not enabling both
https://github.com/vercel/turbo/blob/dfc6ded5f9d14e4d8c1f5d9ccedae98b99e1ad27/crates/turbo-tasks-fetch/build.rs#L8

NicholasLYang:
Gotcha, opened #3585 instead

PR #3585, opened 2023-02-01
fix(turborepo): Check for TLS feature at build-time

NicholasLYang:


mehulkar (approved):

.github/workflows/test.yml
```diff
@@ -405,7 +405,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --workspace --all-targets --locked
+          args: --workspace --all-targets --locked --features rustls-tls
```
sokra, line 408: You could make it a default feature, so it doesn't require passing the feature to every command: https://github.com/vercel/turbo/blob/b1d013abdb539f52f41cd3e09a5a91ffb2175ce8/crates/next-dev/Cargo.toml#L27

jridgewell: Unfortunately we can't: https://github.com/vercel/turbo/pull/3582#pullrequestreview-1279753622

PR #3590, opened 2023-02-01
Move integration test entries to input/

wbinnssmith:
Moves all next-dev integration test code into a parent `input/` directory, like our snapshot tests. This prepares for introducing `issues` snapshots to next-dev integration tests.

Test Plan: CI


crates/next-dev-tests/tests/integration/next/font-google/basic/input/pages/index.js
```diff
@@ -18,7 +18,7 @@ export default function Home() {
 function runTests() {
   it("returns structured data about the font styles from the font function", () => {
     expect(interNoArgs).toEqual({
-      className: "className__inter_34ab8b4d__77c0d301",
```
alexkirsz, line 21: Is this indicative of some hash instability?

wbinnssmith: I'm pretty sure this is because right now we hash the filepath instead of the full content, and that changed in this PR.

sokra (approved):

PR #3591, opened 2023-02-02
Send Next.js Integration test results to Slack

wbinnssmith:
This uses Slack's GitHub action [0] to send the contents of the summary markdown file to a custom workflow for posting in Slack.

Since the markdown file includes newlines and quotes, this uses `jq` to escape it and format a new json file that the action reads.

Test Plan: Manually run the workflow against this branch

[0] https://github.com/slackapi/slack-github-action


.github/workflows/nextjs-integration-test.yml
```diff
@@ -193,3 +182,29 @@ jobs:
           path: |
             nextjs-test-results.json
             test-summary.md
+
+  post_to_slack:
```
kwonoj, line 186: I think we need to move this to on-nextjs-release-publish.yml, since we don't plan to spam channel per each PR's test results for now.

.github/workflows/nextjs-integration-test.yml
```diff
@@ -61,8 +61,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [16, 18]
-        group: [1, 2]
+        node: [16]
```
kwonoj, line 64: we need to revert this?

.github/actions/next-integration-stat/src/index.ts
```diff
@@ -570,8 +569,7 @@ function getTestSummary(
 
   // Store plain textbased summary to share into Slack channel
   // Note: Likely we'll need to polish this summary to make it more readable.
-  if (shouldShareTestSummaryToSlack) {
-    let textSummary = `*Next.js integration test status with Turbopack*
+  let textSummary = `*Next.js integration test status with Turbopack*
```
kwonoj, line 572: need to reenable `shouldShareTestSummaryToSlack`?

.github/workflows/on-nextjs-release-publish.yml
```diff
@@ -33,3 +41,34 @@ jobs:
...
+            echo "should_continue=true" >> $GITHUB_OUTPUT
+          else
+            echo "should_continue=false" >> $GITHUB_OUTPUT
+          fi
+      - name: Build payload
+        if: steps.summary_check.outputs.should_continue == 'true'
+        run: |
+          SUMMARY_MD=$(cat test-summary.md)
+
+          jq --null-input --arg text "$SUMMARY_MD" '{"text": $text}' > payload.json
```
kwonoj, line 67: if we need json payload, we can actually make test-summary.md itself to be a json instead. Should we do it (in following PR)?

kwonoj (approved):

.github/workflows/on-nextjs-release-publish.yml
```diff
@@ -7,6 +7,13 @@ on:
     # This is the event type defined by next.js upstream's `repository_dispatch` workflow dispatches.
     types: [nextjs-release-published]
   workflow_dispatch:
+    inputs:
+      version:
```
kwonoj, line 11: Would you able to add some conditions in https://github.com/vercel/turbo/blob/b11becb48f75b1ef735ea99c7c84a0d6486e25af/.github/workflows/upload-nextjs-integration-test-results.yml#L56

to skip actual commit if this is triggered via manual dispatch? Currently we have some unexpected results uploaded https://github.com/vercel/turbo/tree/nextjs-integration-test-data/test-results without `RESULT_SUBPATH` substitution for the manual dispatch.

PR #3592, opened 2023-02-02
Capture engine/run unit tests with prysk

mehulkar:
This converts more unit tests to integration tests. The purpose of this change is to make it easier to land #3562, which breaks all these unit tests, and it's a lot of unnecessary work to fix as unit tests. Additionally, because we're lazy loading task definitions in that PR, these unit tests would need access to a file system, breaking their "unit" nature, anyway.

- These are not as 1:1 as the previous PR, but attempts to capture the scenarios in spirit.
- Asserted with `--graph`, because that's what the underlying tests were doing also. I could execute, but parallel execution makes it a bit more challenging to assert against logs.
- Combine with existing `topological-deps` integration tests.

Previous similar work:
- #3566


cli/integration_tests/task-dependencies/overwriting.t
```diff
@@ -0,0 +1,41 @@
...
+  workspace-a:build: > build
+  workspace-a:build: > echo 'build workspace-a'
+  workspace-a:build: 
+  workspace-a:build: build workspace-a
+
+workspace-b#generate DID NOT run
+  $ cat tmp.log | grep "workspace-b:generate"
+  [1]
+
+workspace-a#build ran
```
NicholasLYang, line 34:
```suggestion
workspace-b#build ran
```

cli/integration_tests/task-dependencies/topological.t
```diff
@@ -24,3 +24,19 @@ Check my-app#build output
   Cached:    0 cached, 2 total
     Time:\s*[\.0-9]+m?s  (re)
   
+
+
+Graph
```
NicholasLYang, line 29: What's the goal of this test? I'm not really that well versed in this part of the codebase.

mehulkar: This is merged in from here: https://github.com/vercel/turbo/blob/main/cli/integration_tests/topological_deps/graph.t
The underlying config is:

```json
{
  "pipeline": {
    "build": {
      "dependsOn": ["^build"]
    },
    "//#build": {
      "dependsOn": []
    }
  }
}

So I think the idea is to validate that the task graph is generated correctly

NicholasLYang (approved):

PR #3594, opened 2023-02-02
Automate nightly Turbopack release tagging

jridgewell:
This adds a new workflow which automatically releases nightly versions of turbopack. Coupled with https://github.com/vercel/next.js/pull/45506, we'll be able to fully automate our release process into Next.js.

1. The process kicks of with cron every midnight-ish PST.
2. We find the latest commit that passes CI
3. We bump the latest tag version (more on this in a bit)
4. We generate a new GH release with this version

The automated tagging version will break with our current `turbopack-YYMMDD.XYZ` release process just a bit. The tag action that I'm using can't generate a date-like tag, it can only handle semver (with optional prefix). So I've set it up to generate `turbopack-MAJOR.MINOR.PATCH-nightly.XYZ` tags. The `turbopack-` here differentiates us from Turborepo, and the `nightly` tag denotes this a non-release software.

I hope that I've configured this correctly to always bump patches, but I'm not 100% confident until it runs.

Fixes WEB-502

### DIFF
--- a/.github/workflows/turbopack-nightly-release.yml
+++ b/.github/workflows/turbopack-nightly-release.yml
@@ -29,6 +29,10 @@ jobs:
           # commits from Turborepo, but that'd require a custom
           # action.
           custom_release_rules: ""
+          # Ensure we generate prerelease versions, not stables.
+          release_branches: ""
+          pre_release_branches: "main"
+          append_to_pre_release_tag: "alpha"
 
       - name: Create GitHub release
         uses: ncipollo/release-action@v1
--- a/.github/workflows/turbopack-nightly-release.yml
+++ b/.github/workflows/turbopack-nightly-release.yml
@@ -23,6 +23,12 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           commit_sha: ${{ steps.green_commit.outputs.result }}
           tag_prefix: "turbopack-"
+          # Override the default conventional commit messages
+          # so that we always generate patch releases.
+          # This would be easier if we could filter Turbopack
+          # commits from Turborepo, but that'd require a custom
+          # action.
+          custom_release_rules: ""
 
       - name: Create GitHub release
         uses: ncipollo/release-action@v1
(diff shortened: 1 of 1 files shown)


tknickman:
@jridgewell let us know if it would be helpful for us to start prefixing turborepo tags - we're just squatting on the vanilla semver tags now. Same goes for Github releases.

jridgewell:
> let us know if it would be helpful for us to start prefixing turborepo tags

From a technical standpoint, no it's not necessary. From a product standpoint, maybe?

tknickman:
Yea I think it might be beneficial to help everyone understand the state of each project

.github/workflows/turbopack-nightly-release.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - name: Checkout latest commit that passes CI
+        uses: talentpair/last-green-commit-action@d95cfa8
+        id: green_commit
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Tag nightly
+        id: tag_version
+        uses: mathieudutour/github-tag-action@v6.1
```
sokra, line 21: I guess you could also use `git tag turbopack...` and `git push --tags` which gives you full freedom in naming tags.

.github/workflows/turbopack-nightly-release.yml
```diff
@@ -0,0 +1,42 @@
...
+          release_branches: ""
+          pre_release_branches: "main"
+          append_to_pre_release_tag: "alpha"
+
+      - name: Create GitHub release
+        uses: ncipollo/release-action@v1
+        with:
+          tag: ${{ steps.tag_version.outputs.new_tag }}
+          name: Turbopack Release ${{ steps.tag_version.outputs.new_tag }}
+          body: ${{ steps.tag_version.outputs.changelog }}
```
sokra, line 42: We probably don't want to include turborepo changes here...

jridgewell: That's the difficult part, because we don't currently have a good list of things that are Turbopack vs Turborepo. Splitting this into better defined directories would be helpful.

sokra: Yeah. I guess we probably don't need a changelog for the nightlies anyway

jridgewell: Removed

sokra (approved):

PR #3595, opened 2023-02-02
update next.js to 13.1.7-canary.2

sokra:


sokra:
[0a42a71](https://github.com/vercel/turbo/pull/3595/commits/0a42a713f9a9fe52e9ab5ff2d24e710008b41a27) can be reverted when https://github.com/vercel/turbo/pull/3597 is merged

crates/next-dev/benches/bundlers/turbopack/mod.rs
```diff
@@ -49,7 +49,7 @@ impl Bundler for Turbopack {
         npm::install(
             install_dir,
             &[
-                NpmPackage::new("next", "13.0.8-canary.2"),
+                NpmPackage::new("next", "13.1.7-canary.2"),
```
sokra, line 52: This enables the next.js router for the benchmarks, which causes a slowdown

crates/next-core/src/next_import_map.rs
```diff
@@ -236,9 +236,7 @@ pub fn get_next_client_resolved_map(
         // Temporary hack to replace the hot reloader until this is passable by props in next.js
         (
             context,
-            GlobVc::new(
-                "**/*/next/dist/client/components/react-dev-overlay/hot-reloader-client.js",
-            ),
+            GlobVc::new("**/next/dist/client/components/react-dev-overlay/hot-reloader-client.js"),
```
ForsakenHarmony, line 239: locally rustfmt does this differently for me

crates/turbo-tasks-build/src/lib.rs
```diff
@@ -160,8 +160,10 @@ pub fn generate_register() {
 
 pub fn rerun_if_glob(globs: &str, root: &str) {
     let cwd = env::current_dir().unwrap();
-    let globs = cwd.join(globs);
-    let mut seen = HashSet::from([cwd.join(root)]);
+    let globs = cwd.join(globs.replace('/', PATH_SEP.to_string().as_str()));
+    let root = cwd.join(root.replace('/', PATH_SEP.to_string().as_str()));
```
sokra, line 164: Windows

crates/turbo-tasks-build/src/lib.rs
```diff
@@ -160,8 +160,10 @@ pub fn generate_register() {
...
+    println!("cargo:rerun-if-changed={}", root.display());
```
sokra, line 165: The root directory should also be watched

alexkirsz, line 165: Shouldn't that be handled by the `ancestors` iter above? Or is that for the case where `root` is empty?

alexkirsz (approved):

ForsakenHarmony (approved):

crates/next-dev/benches/bundlers/turbopack/mod.rs
ForsakenHarmony: Would be nice if we could declare this in some central place somehow

crates/turbo-tasks-build/src/lib.rs
sokra: The root is in `seen` from the beginning, so it's ignored below

crates/next-core/src/next_import_map.rs
sokra: The CI checks fmt, so it's fine^^

PR #3597, opened 2023-02-02
Share request resolving logic between http and update servers

alexkirsz:
With the addition of the Next.js routing layer at the root of our content sources, we need the update server to also be able to follow Next.js rewrites, like the HTTP server.

This requires the following:
* Rewrites need to be able to indicate a source where to "resume" after rewriting. Otherwise, we would enter an infinite loop of the Next.js layer rewriting to the Next.js layer.
* The resolving logic from `process_request_with_content_source` needs to be extracted into its own function, so it may be called both from the HTTP server (which expects fully resolved `ReadRef`s) and the update server (which needs a `VersionedContentVc`): this is where `resolve_source_request` comes in.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -318,9 +318,9 @@ checksum = "7a40729d2133846d9ed0ea60a8b9541bccddab49cd30f0715a1da672fe9a2524"
 
 [[package]]
 name = "async-trait"
-version = "0.1.63"
+version = "0.1.64"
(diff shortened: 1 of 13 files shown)


sokra (changes requested):
doesn't compile

sokra (approved):

jridgewell (commented):
Excellent, thank you for getting this across the finish line!

PR #3600, opened 2023-02-02
Support for getStaticPaths.fallback = false

alexkirsz:
Adds support for `getStaticPaths` returning `{ fallback: false }` by building out all static paths and checking whether the resolved path is included in that list.

`buildStaticPaths` requires the `next.config.js` path, as well as other config options such as locale. I created WEB-546 to track being able to pass them to the handlers.

### DIFF
--- a/crates/next-core/js/src/internal/page-server-handler.tsx
+++ b/crates/next-core/js/src/internal/page-server-handler.tsx
@@ -10,6 +10,7 @@ import type { IncomingMessage, ServerResponse } from "node:http";
 import { renderToHTML, RenderOpts } from "next/dist/server/render";
 import { getRedirectStatus } from "next/dist/lib/redirect-status";
 import { PERMANENT_REDIRECT_STATUS } from "next/dist/shared/lib/constants";
+import { buildStaticPaths } from "next/dist/build/utils";
 import type { BuildManifest } from "next/dist/server/get-page-files";
 import type { ReactLoadableManifest } from "next/dist/server/load-components";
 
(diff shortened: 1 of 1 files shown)


ForsakenHarmony (approved):

crates/next-core/js/src/internal/page-server-handler.tsx
```diff
@@ -290,6 +299,26 @@ export default function startHandler({
...
 
+function createNotFoundResponse(isDataReq: boolean): IpcOutgoingMessage {
+  if (isDataReq) {
+    return {
+      type: "response",
+      // Returning a 404 status code is required for the client-side router
+      // to redirect to the error page.
+      statusCode: 404,
+      body: '{"notFound":true}',
+      headers: [["Content-Type", MIME_APPLICATION_JAVASCRIPT]],
```
jridgewell, line 310: Should this be JSON?

alexkirsz: The definition is
```rust
headers: Vec<(String, String)>,`
```

jridgewell: I mean the `MIME_APPLICATION_JAVASCRIPT`, should that be returning JSON content-type instead?

alexkirsz: Yes indeed it should. I'll fix this in my `basePath` PR since it also touches this part of the code.

PR #3605, opened 2023-02-02
fix(cli): ci detection

tknickman:
Improve CI detection

### DIFF
--- a/cli/internal/ci/ci.go
+++ b/cli/internal/ci/ci.go
@@ -4,7 +4,7 @@ package ci
 
 import "os"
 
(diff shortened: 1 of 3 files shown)


cli/internal/ci/ci.go
```diff
@@ -18,6 +18,7 @@ func Name() string {
 
 // Info returns information about a CI vendor
 func Info() Vendor {
+	// key=val required
```
arlyon, line 21: What does this comment mean? Maybe flesh it out

arlyon (approved):
Looks great

cli/internal/ci/ci.go
tknickman: Ah good call - will do.

PR #3611, opened 2023-02-02
fix: mismatched pnpm lockfile version resolution

### LINKED ISSUE #3564: [turborepo] turbo prune generating broken pnpm lockfile

1.7.1

pnpm

Windows

### Describe the Bug

We started seeing CI failures today with pnpm reporting a broken lockfile when running `pnpm install --frozen-lockfile` on a pruned workspace.  After trying to ascertain the problem, we realized that earlier CI runs were using turbo 1.7.0 and the latest CI runs were using turbo 1.7.1.  I was able to replicate locally by using turbo 1.7.1 and I've compared the two lock files and the lockfile generated by 1.7.1 is missing a few dependencies which were present in the 1.7.0 lockfile.  The below is a snippet of a diff with 1.7.0 on the left and 1.7.1 on the right:
image
There are a few other dependencies missing, but the total diff isn't substantial (lockfile size went from 11223 lines to 11100).  However, it is enough to cause a broken lockfile as these do appear to be dependencies which are needed.

I'm happy to provide whatever other information is helpful, but I'm not sure exactly what will help at this stage.

### Expected Behavior

Generate a valid lockfile for the pruned workspace.

### To Reproduce

I'm not sure how easily this can be replicated and might depend on what dependencies are in the lockfile.  In my case, I simply ran `pnpm exec turbo prune --scope=<workspace_name>` and then tried to run `pnpm install --frozen-lockfile` from the `out` folder.

### Reproduction Repo

_No response_

chris-olszewski:
Fixes #3564

I created a regression when adding support for pnpm `lockfileVersion: 6` that caused us to not include some indirect dependencies of a workspace. A package would be missing only if a workspace had a direct dependency on the package that was also included as an indirect dependency where the version ranges didn't overlap. See the new test fixture for a basic version of how this could occur.

Huge thanks to @shawnmcknight for providing a reproduction!


arlyon (approved):
Nice and quick turn-around @chris-olszewski :) Approved.

PR #3612, opened 2023-02-02
feature(turborepo): SSO Login

NicholasLYang:


crates/turborepo-lib/src/client.rs
```diff
@@ -92,26 +96,20 @@ pub struct UserResponse {
...
 }
 
-#[async_trait]
-impl UserClient for APIClient {
-    fn set_token(&mut self, token: String) {
-        self.token = token
-    }
-
-    async fn get_user(&self) -> Result<UserResponse> {
+impl APIClient {
```
NicholasLYang, line 103: Changed from having a `token` field in `APIClient` because we don't want to require a token to construct `APIClient`, as we need a client to do SSO login

arlyon (approved):
Code looks nice. It may complicate things but we could perhaps use generics over the token to conditionally enable certain APIs. I am using a similar pattern for the daemon, since the connector needs a partially-initialized client to perform the handshake.

https://github.com/vercel/turbo/commit/aa1cdfcc7554d3aca12f4a28ee761f127bc14334#diff-1bcb3c075e5dbf2612e90606bd587d15aa912e7cddefcb7a88b56b116d34521eR66-R85

TBH though it may be 'good enough' to just pass the token in...

NicholasLYang:
Oh cool! That's a nice pattern. I'll see if it works here

PR #3615, opened 2023-02-03
Hoist DryRunSummary fields to PackageTask

mehulkar:
The `PackageTask` abstraction has some statically analyzable information about the task, and DryRunSummary collects the rest. Hoisting as much of this information up to the PackageTask will make it easier to make it available on `RealRun`s without duplicating code. Ideally, DryRun doesn't do anything except print the information we already have.

This PR extracts some bits of #3431 without changing any behavior


cli/internal/run/dry_run.go
```diff
@@ -102,10 +102,11 @@ func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.Complete
 			return err
 		}
 
-		command, ok := packageTask.Command()
-		if !ok {
-			command = "<NONEXISTENT>"
+		command := "<NONEXISTENT>"
```
arlyon, line 105: Is it worth extracting this var out while we're here? I think a variable name would help me understand what this value signifies.

arlyon (commented):
No complaints but I'll leave it to someone else with more context to give the final :heavy_check_mark:

NicholasLYang (approved):

cli/internal/run/dry_run.go
mehulkar: for sure! pushed up a commit to explain. Doing to set automerge on this PR, but if it stil doesn't make sense, let me know and we can figure out how to improve the commends / naming.

PR #3618, opened 2023-02-03
make sure node.js worker processes are cleaned up correctly on errors

sokra:
Before it tried to reuse processes even if they crashed which leads to "connection is closed" errors on send and recv


crates/turbopack-node/src/pool.rs
```diff
@@ -363,11 +371,17 @@ impl NodeJsOperation {
...
     fn drop(&mut self) {
-        if let Some(process) = self.process.take() {
-            self.processes
-                .lock()
-                .unwrap()
-                .push(NodeJsPoolProcess::Running(process));
+        if let Some(mut process) = self.process.take() {
+            if self.in_bad_state {
+                if let Some(mut child) = process.child.take() {
+                    let _ = child.start_kill();
```
ForsakenHarmony, line 377: Why not `child.kill()`?

crates/turbopack-node/src/evaluate.rs
```diff
@@ -140,13 +140,28 @@ pub async fn evaluate(
...
-        .send(EvalJavaScriptOutgoingMessage::Evaluate {
-            args: args.iter().map(|v| &**v).collect(),
-        })
-        .await?;
+
+    // Sending to workers can fail if reused workers are in a bad state.
+    // To we retry picking workers from the pools until we succeed.
+    let mut i = 1;
+    let mut operation = loop {
+        let mut operation = pool.operation().await?;
```
ForsakenHarmony, line 150: `pool.operation()` should not return a process in a "bad state"

ForsakenHarmony (changes requested):

crates/turbopack-node/src/evaluate.rs
```diff
@@ -140,13 +140,28 @@ pub async fn evaluate(
...
+
     let args = args.into_iter().try_join().await?;
-    operation
-        .send(EvalJavaScriptOutgoingMessage::Evaluate {
-            args: args.iter().map(|v| &**v).collect(),
-        })
-        .await?;
+
+    // Sending to workers can fail if reused workers are in a bad state.
+    // To we retry picking workers from the pools until we succeed.
```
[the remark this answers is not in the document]
sokra: They might crash randomly even without reporting an error, e.g. when during idle some leftover timeout exits the process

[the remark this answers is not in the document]
ForsakenHarmony: see my comment about `pool.operation` below for that case

crates/turbopack-node/src/evaluate.rs
sokra: Not sure how to ensure that. We could check the worker with some ping message, but that costs a roundtrip and it still could crash right after that, so we would still need to handle this case and retry.

crates/turbopack-node/src/pool.rs
sokra: I copied that from our kill_and_wait message. I think @alexkirsz wrote that

ForsakenHarmony: oh I see, this is a tokio process, we can't synchronously kill that

crates/turbopack-node/src/pool.rs
```diff
@@ -363,11 +371,17 @@ impl NodeJsOperation {
...
+                }
+            } else {
+                self.processes
+                    .lock()
+                    .unwrap()
+                    .push(NodeJsPoolProcess::Running(process));
+            }
         }
```
ForsakenHarmony, line 385:
```suggestion
            if !self.in_bad_state {
                self.processes
                    .lock()
                    .unwrap()
                    .push(NodeJsPoolProcess::Running(process));
            }
        }
```

`self.process` will already kill the child

crates/turbopack-node/src/evaluate.rs
ForsakenHarmony: At the very least, we could check `child.try_wait()`

crates/turbopack-node/src/pool.rs
ForsakenHarmony: actually, we can remove the other `child.kill` drop impls above and add `cmd.kill_on_drop(true);` to `NodeJsPoolProcess::new`

jridgewell (approved):

crates/turbopack-node/src/evaluate.rs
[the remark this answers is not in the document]
ForsakenHarmony: Was the loop needed for the issue you had?

How do we ensure here that the node.js process crash is recoverable?

crates/turbopack-node/src/evaluate.rs
```diff
@@ -140,13 +140,28 @@ pub async fn evaluate(
...
+        let result = operation
+            .send(EvalJavaScriptOutgoingMessage::Evaluate {
+                args: args.iter().map(|v| &**v).collect(),
+            })
+            .await;
+        if result.is_ok() {
+            break operation;
+        } else if i == 10 {
+            result?;
+            break operation;
```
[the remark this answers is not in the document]
ForsakenHarmony: a `match` would be better

crates/turbopack-node/src/pool.rs
ForsakenHarmony: I'm fine with the loop for now, but I'd like this changed

crates/turbopack-node/src/evaluate.rs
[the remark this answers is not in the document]
sokra: indeed

jridgewell (approved):

crates/turbopack-node/src/pool.rs
```diff
@@ -359,15 +373,30 @@ impl NodeJsOperation {
...
 
 impl Drop for NodeJsOperation {
     fn drop(&mut self) {
-        if let Some(process) = self.process.take() {
-            self.processes
-                .lock()
-                .unwrap()
-                .push(NodeJsPoolProcess::Running(process));
+        if let Some(mut process) = self.process.take() {
+            if self.kill_on_drop {
```
ForsakenHarmony, line 385: this is way worse
just add `cmd.kill_on_drop(true);` in `NodeJsPoolProcess::new`, we always want that, no?

this is a feature of tokio

ForsakenHarmony:
```rs
            if !self.in_bad_state {
                self.processes
                    .lock()
                    .unwrap()
                    .push(NodeJsPoolProcess::Running(process));
            }
```
would be more obvious

sokra: ah I see

crates/turbopack-node/src/pool.rs
```diff
@@ -134,6 +134,7 @@ impl NodeJsPoolProcess {
         cmd.envs(env);
         cmd.stderr(Stdio::piped());
         cmd.stdout(Stdio::piped());
+        cmd.kill_on_drop(true);
```
ForsakenHarmony, line 137: Remove the 2 `impl Drop` above

crates/turbopack-node/src/pool.rs
```diff
@@ -324,22 +327,34 @@ impl NodeJsOperation {
...
+        Ok(message)
     }
 
     pub async fn send<M>(&mut self, message: M) -> Result<()>
     where
         M: Serialize,
     {
+        // We temporary set kill_on_drop to true to make sure the process is killed when
+        // this operation fail
+        let old = self.allow_process_reuse;
```
ForsakenHarmony, line 350: I feel like we should error if this is already set, no?

ForsakenHarmony (changes requested):

crates/turbopack-node/src/pool.rs
```diff
@@ -311,35 +294,48 @@ pub struct NodeJsOperation {
...
+    ) -> Result<T> {
+        let process = self
+            .process
             .as_mut()
-            .context("Node.js operation already finished")
+            .context("Node.js operation already finished")?;
+
+        let result = f(process).await;
+        if result.is_err() {
+            self.allow_process_reuse = false;
```
ForsakenHarmony, line 312: Same comment once again, shouldn't we check and throw an error if `with_process` is called when `allow_process_reuse` is false?

sokra: fixed

crates/turbopack-node/src/evaluate.rs
```diff
@@ -140,13 +140,32 @@ pub async fn evaluate(
...
+            Err(err) if i == 10 => {
+                return Err(err);
+            }
+            Err(_) => {
+                // retry
+                i += 1;
+                continue;
+            }
+        }
+    };
```
alexkirsz, line 167: Wrap this in a `retry_future` call. We at least one utility function somewhere for this. Otherwise the meaning of the code is obscured by the retry boilerplate.

Also, put the `10` in a constant somewhere.

crates/turbopack-node/src/evaluate.rs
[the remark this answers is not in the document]
alexkirsz: @sokra any updates on this?

crates/turbopack-node/src/pool.rs
```diff
@@ -311,35 +294,52 @@ pub struct NodeJsOperation {
     #[allow(dead_code)]
     permit: OwnedSemaphorePermit,
     processes: Arc<Mutex<Vec<NodeJsPoolProcess>>>,
+    allow_process_reuse: bool,
 }
 
 impl NodeJsOperation {
-    fn process_mut(&mut self) -> Result<&mut RunningNodeJsPoolProcess> {
-        self.process
+    async fn with_process<'a, F: Future<Output = Result<T>> + Send + 'a, T>(
```
alexkirsz, line 301: This API could also use a `ProcessGuard<'a>`, right?

crates/turbopack-node/src/evaluate.rs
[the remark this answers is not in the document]
sokra: > How do we ensure here that the node.js process crash is recoverable?

We throw away the crashed process and pick an new one from the pool, or even create a fresh one.

crates/turbopack-node/src/pool.rs
sokra: I tried, but it can't capture the `Result` Err in a great way.

```
let guard = op.process();

let result = some_stuff_that_might_fail(*guard).await?;

// You would need to actively flag the guard as successful
guard.ok();

// So dropping guard knows that it should not set allow_process_reuse to false
```

Compare to:

```
let result = op
  .with_process(
      |process| async move { some_stuff_that_might_fail(process).await },
  )
  .await?;
```

ForsakenHarmony (approved):

PR #3619, opened 2023-02-03
add helper for walking a tree concurrently and deterministic

sokra:
Similar to our `try_join` helper this adds a `try_flat_map_recursive_join` helper, which allows to async expand a tree structure into all nodes. It will call the async mapper function concurrently to allow parallelism. It will handle circular and duplicate references and return all nodes in a determinstic way (breath-first).


crates/turbo-tasks/src/join_iter_ext.rs
```diff
@@ -80,6 +135,16 @@ where
...
 }
 
+pub trait TryFlatMapRecursiveJoinIterExt<T, C, F, CI>: Iterator
+where
+    T: Hash + PartialEq + Eq + Clone,
+    C: Fn(T) -> F,
+    F: Future<Output = Result<CI>>,
+    CI: IntoIterator<Item = T>,
+{
+    fn try_flat_map_recursive_join(self, get_children: C) -> TryFlatMapRecursiveJoin<T, C, F, CI>;
```
alexkirsz, line 145: Add a comment to this method/to the trait.

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -124,26 +124,41 @@ impl ChunkGroupVc {
...
-                        if *pc.is_loaded_in_parallel().await? {
-                            let result = r.resolve_reference();
-                            for a in result.primary_assets().await?.iter() {
-                                if let Some(chunk) = ChunkVc::resolve_from(a).await? {
-                                    queue.push(chunk);
-                                }
+        let chunks = [self.await?.entry]
+            .into_iter()
+            .try_flat_map_recursive_join(|chunk| async move {
+                Ok(chunk
```
alexkirsz, line 130: Split up these huge method chains into intermediate items for readability purposes imo.

```rust
                let references = chunk
                    .references()
                    .await?;
                let parallel_chunks = references.iter().map(|r| ...);
```
etc.

Same comment for https://github.com/vercel/turbo/pull/3619/files#diff-920fc3f07551ceb49ecd59a0730dae802cb4f7b906b576defad3149d50933deeR27 and https://github.com/vercel/turbo/pull/3619/files#diff-b6de4264e8fc5bf158812dc1ebd1697428cb28e16adad1eda0cdd339e96b9151R138

While rust-analyzer might show the intermediate types, it doesn't show on GH and the flow of the transformation is unclear.

alexkirsz (approved):

PR #3623, opened 2023-02-03
Unwrap sequence await callee in analyzer

Brooooooklyn:
Close WEB-555


sokra (changes requested):
It would be great to have an node-file-trace test case for that

crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -449,6 +449,8 @@ impl EvalContext {
                 SyntaxContext::empty(),
             )),
 
+            Expr::Await(AwaitExpr { arg, .. }) => self.eval(arg),
```
sokra, line 452: Technically it's not super correct. e. g. `(await { then: () => ({ v: 42 }), v: 43 }).v` evaluates to 43 instead of 42, but I'm fine with that for now...

sokra (approved):

PR #3625, opened 2023-02-03
docs: Increase sidebar text contrast in dark mode

wbinnssmith:
This increases the contrast of the sidebar text against its dark background by slightly lightening it. It now passes WCAG AA guidelines.

Test Plan: https://webaim.org/resources/contrastchecker/


arlyon (approved):
Seems reasonable to avoid forking the theme. Much more readable :)

PR #3628, opened 2023-02-03
ci(workflow): fix summary count error

kwonoj:
Minor fix for calculation error.


ForsakenHarmony (approved):
maybe add this action to the codeowners file

PR #3629, opened 2023-02-03
Various Rope updates and fixes

ForsakenHarmony:
Updates and fixes a few items:
- Fix: Prevents a Rope from containing an empty `Bytes` section
- Perf: Use a `Box<[RopeElem]>` instead of a `Vec<RopeElem>`, reducing memory useage
- Perf: Implement fast path for `Rope::to_str()` when containing a single `Shared` element

The first one is the important change. The assumption while iterating (for streaming, equality checks, etc) is that the only time we can hit an empty `Bytes` section is when we reach the end of the entire Rope. If an empty section is hit in the middle of a rope, we can't tell if we've reached the end or we're in the middle without calling N more times (eg, what if an empty section followed an empty section, then we'd need to call 2 more times to see).

### DIFF
--- a/crates/turbo-tasks-fs/src/rope.rs
+++ b/crates/turbo-tasks-fs/src/rope.rs
@@ -4,6 +4,7 @@ use std::{
     fmt::Debug,
     io::{self, BufRead, Read, Result as IoResult, Write},
     mem, ops,
+    ops::Deref,
     pin::Pin,
     sync::Arc,
     task::{Context as TaskContext, Poll},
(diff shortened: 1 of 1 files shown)


crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -508,12 +515,9 @@ impl AsyncRead for RopeReader {
 }
 
 impl BufRead for RopeReader {
-    /// Returns the full buffer without coping any data. The same bytes will
-    /// continue to be returned until [consume] is called to either consume a
-    /// partial amount of bytes (in which case the Bytes will advance beyond
-    /// them) or the full amount of bytes (in which case the next Bytes will be
-    /// returned).
     fn fill_buf(&mut self) -> IoResult<&[u8]> {
```
ForsakenHarmony, line 518: I changed / removed these comments because the rustdocs on the traits were better (in my opinion) and more useful from the perspective of the user of the api

ForsakenHarmony:
* main
  * **PR #3629** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3629" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3629?utm_source=stack-comment).

crates/turbo-tasks-fs/src/rope.rs
ForsakenHarmony: No strong preference

jridgewell (approved):

PR #3632, opened 2023-02-03
feat: port npm lockfile to Rust

chris-olszewski:
Adds a Rust port of the npm lockfile implementation only used when `GO_TAG=rust` is set.

Notes:
 - The crawling of the lockfile graph is done in the language of the lockfile impl to avoid incurring ffi overhead for every transitive dependency. This change requires moving this function into the lockfile package instead of context in Go (this is probably where it should've belonged all along)
 - We explicitly make an ffi function in Rust for each lockfile implementation. This adds boilerplate, but as this boundary should be short lived (only around until package graph gets moved into Rust) this is acceptable instead of coming up with a more generic solution.
 - If we stick with passing the lockfile contents for other implementations then we will be able to share the protobuf message definitions


NicholasLYang (commented):
We need to add turborepo-lockfile to the excluded packages for turbopack workflows [here](https://github.com/vercel/turbo/blob/main/.github/workflows/test.yml#L606) and [here](https://github.com/vercel/turbo/blob/main/.github/workflows/test.yml#L670). We also need to add it to the [turborepo related changes ](https://github.com/vercel/turbo/blob/main/.github/workflows/test.yml#L105)and exclude it from the turbopack related changes (right above the previous link). I'll try to document this but in the future we should figure out a way to automate it.

cli/internal/ffi/lockfile.go
```diff
@@ -0,0 +1,68 @@
...
+
+	ffi_proto "github.com/vercel/turbo/cli/internal/ffi/proto"
+)
+
+// NpmTransitiveDeps returns the transitive external deps of a given package based on the deps and specifiers given
+func NpmTransitiveDeps(content []byte, pkgDir string, unresolvedDeps map[string]string) ([]*ffi_proto.LockfilePackage, error) {
+	return transitiveDeps(npmTransitiveDeps, content, pkgDir, unresolvedDeps)
+}
+
+func npmTransitiveDeps(buf C.Buffer) C.Buffer {
```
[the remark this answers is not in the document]
chris-olszewski: Go type system? For whatever reason `C.npm_transitive_closure` isn't of type `func(C.Buffer) C.Buffer` so I just manually wrap it and Go compiles

crates/turborepo-lockfiles/src/npm.rs
```diff
@@ -0,0 +1,365 @@
...
+            .get(key)
+            .map(|pkg| {
+                pkg.dep_keys()
+                    .filter_map(|name| {
+                        Self::possible_npm_deps(key, name)
+                            .into_iter()
+                            .find_map(|possible_key| {
+                                self.packages.get(&possible_key).map(|entry| {
+                                    let version = entry.version.as_deref().ok_or_else(|| {
+                                        Error::MissingVersion("".into(), possible_key.clone())
```
arlyon, line 89: Shall we consider making the workspace an option

crates/turborepo-lockfiles/src/npm.rs
```diff
@@ -0,0 +1,365 @@
...
+                self.packages.get(&key).map(|pkg| {
+                    let version = pkg.version.clone().unwrap_or_default();
+                    Ok(Package { key, version })
+                })
+            })
+            .next()
+            .transpose()
+    }
+
+    fn all_dependencies(&self, key: &str) -> Result<Option<HashMap<String, &str>>, Error> {
```
arlyon, line 78: Does it make sense to not transpose here? That way the top level option defines the existence of the key, rather than the alternative which is 'either an optional key or one of the deps is missing a version'. Something like this:

```suggestion

    /// Gets all the dependencies of a package, if it exists in the graph.
    /// If None, the package key does not exist.
    /// If Some(Error), the package exists, but its dependencies are invalid.
    fn all_dependencies(&self, key: &str) -> Option<Result<Hashmap<String, &str>, Error>> {
```

crates/turborepo-lockfiles/src/npm.rs
```diff
@@ -0,0 +1,365 @@
...
+        &self,
+        workspace_packages: &[String],
+        packages: &[String],
+    ) -> Result<Self, Error> {
+        let mut pruned_packages = HashMap::with_capacity(packages.len());
+        for pkg_key in packages {
+            let pkg = self.get_package(pkg_key)?;
+            pruned_packages.insert(pkg_key.to_string(), pkg.clone());
+        }
+        if let Some(root) = self.packages.get("") {
```
arlyon, line 135: Does is make sense to encode the root into the type system? I'm pretty sure with pointer types this has no perf overhead as rust just encodes the Root case to be the null value, but it is of course a bit more complicated.

```rust
enum Package<'a> {
  Root,
  Named(&'a str),
}
```

crates/turborepo-lockfiles/src/npm.rs
```diff
@@ -0,0 +1,365 @@
...
+                    None
+                } else {
+                    Some(parent)
+                }
+            })
+    }
+}
+
+impl NpmPackage {
+    pub fn dep_keys(&self) -> impl Iterator<Item = &String> {
```
arlyon, line 184: Love this :)

arlyon (approved):
Some minor bits, but overall looks great. Address them if you'd like, they're not critical.

cli/internal/ffi/lockfile.go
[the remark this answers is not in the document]
arlyon: Very odd. Would cleaning the build cache maybe fix this? Logically it seems fine but I'm paranoid of leaving it in because it is so innocuous looking.

arlyon:
As a general comment, maybe we can run the two code paths in parallel (instead of mutex) and compare the outputs so that we can thoroughly fuzz this?

Cargo.toml
```diff
@@ -78,6 +79,11 @@ default-members = [
   "crates/turbopack-swc-utils",
   "crates/turbopack",
   "crates/turbopack-tests",
+  "crates/turborepo",
```
NicholasLYang, line 82: We don't actually want to do this because the turbopack workflows depend on only the turbopack crates being default members if I remember correctly.

chris-olszewski: Oops this snuck in with a merge conflict

cli/internal/ffi/lockfile.go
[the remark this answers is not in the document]
chris-olszewski: Clearing build cache doesn't help. Looking at the error message it looks like cgo functions are a different type than regular Go functions which isn't too surprising:
`cannot assign C.npm_transitive_closure (cgo function of type func(p0 _Ctype_struct_Buffer) (r1 _Ctype_struct_Buffer)) to func(_Ctype_struct_Buffer) _Ctype_struct_Buffer`

I've looked at it doesn't appear like there's a good way to describe taking a cgo function as a parameter in Go.

crates/turborepo-lockfiles/src/npm.rs
chris-olszewski: That's getting a little too clever for my taste as I want `NpmLockfile` to be a transparent in-memory version of the `package-lock.json` on disk and `package-lock.json` has a `""` entry in the `packages` section

crates/turborepo-lockfiles/src/npm.rs
chris-olszewski: This is a slight oddity coming from trying to keep this comparable to Go so we can ship changes to both implementations while we're supporting both. In reality, this should be a panic as a version will only be missing if an external package depends on a workspace. I don't want to panic as this is getting called via FFI and would rather let Go handle the error.

Once we stop maintaining/shipping the Go version of the lockfiles there's a layer of Rust we should apply to these implementations as we can rely on types more, but I think that should wait until we're down to one implementation.

crates/turborepo-lockfiles/src/npm.rs
chris-olszewski: I've been missing my iterators while I've been in Go land

cli/internal/ffi/lockfile.go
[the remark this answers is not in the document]
NicholasLYang: Haha gotta love Go

crates/turborepo-lockfiles/src/npm.rs
arlyon: > package-lock.json has a "" entry in the packages section

Didn't know that, I assumed that that was a turbo convention, TY :)

PR #3634, opened 2023-02-04
Import Next's CJS AsyncLocalStorage modules

jridgewell:
The node evaluation always renders with `type: "commonjs"` and `require()` calls, so we'll always import the CJS files. But here we're importing the ESM files. That means we have 2 distinct instances of `requestAsyncStorage` in our node instance, and they cannot properly communicate with the other.

Fixes WEB-543

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ default-members = [
   "crates/next-binding",
   "crates/next-core",
   "crates/next-dev",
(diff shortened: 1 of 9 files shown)


sokra (approved):
Could we add a test case for this to not regress on that?

jridgewell:
Added test case

PR #3642, opened 2023-02-06
Update minor tailwind example issues.

nathanhammond:
The UI package had a few residual config issues from tweaks.

### DIFF
--- a/examples/with-tailwind/packages/ui/package.json
+++ b/examples/with-tailwind/packages/ui/package.json
@@ -15,9 +15,11 @@
     "dev": "tsup --watch",
     "check-types": "tsc --noEmit"
   },
+  "peerDependencies": {
+    "react": "^18.2.0"
+  },
   "devDependencies": {
     "@types/react": "^18.0.26",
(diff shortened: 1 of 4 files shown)


examples/with-tailwind/packages/ui/package.json
```diff
@@ -15,9 +15,11 @@
     "dev": "tsup --watch",
     "check-types": "tsc --noEmit"
   },
+  "peerDependencies": {
+    "react": "^18.2.0"
+  },
```
nathanhammond, line 20: Controversial. But I think it's worth doing.

examples/with-tailwind/packages/ui/tsup.config.ts
```diff
@@ -8,6 +8,6 @@ export default defineConfig((options: Options) => ({
   dts: true,
   minify: true,
   clean: true,
-  external: ["react", "react-dom"],
```
nathanhammond, line 11: This enumeration was unnecessary and should have instead been populated by tools. (Human error is expected, this is the kind of thing that should have been automatic.)

PR #3646, opened 2023-02-06
split tests into two steps, report ok and done separately

sokra:
This removes some long running tests (mac and windows) from the required PR checks.

Now the CI workflow is split into two parts, `Ok` (required) for the required checks and `Done` (non-blocking) for all checks.

The PR comment will show

🟢 CI likely successful 🟢
A few longer running steps are still running, but they should not be considered as blocking.

See workflow summary for details

after the required checks are done.

Once all checks are done, it will be updated to 🟢 CI successful 🟢 or to ⚠️ CI failed ⚠️ depending on the full result

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1081,7 +1081,7 @@ jobs:
             local name=$2
             if [ "$result" = "cancelled" ]; then
               cancelled=true
-            elif [ "$result" != "success" && "$result" != "skipped" ]; then
+            elif [ "$result" != "success" ] && [ "$result" != "skipped" ]; then
               echo "- $name" >> failures.md
               failure=true
             fi
@@ -1198,7 +1198,7 @@ jobs:
             local name=$2
             if [ "$result" = "cancelled" ]; then
               cancelled=true
-            elif [ "$result" != "success" && "$result" != "skipped" ]; then
+            elif [ "$result" != "success" ] && [ "$result" != "skipped" ]; then
               echo "- $name" >> failures.md
               failure=true
             fi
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -982,122 +1060,179 @@ jobs:
...
+          subjob ${{needs.determine_jobs.result}} "Determining jobs"
+          subjob ${{needs.go_lint.result}} "Go lints"
+          subjob ${{needs.go_unit.result}} "Go unit tests"
+          subjob ${{needs.go_examples.result}} "Go examples"
+          subjob ${{needs.go_e2e.result}} "Go e2e tests"
+          subjob ${{needs.go_integration.result}} "Go integration tests"
+          subjob ${{needs.rust_prepare.result}} "Rust prepare"
+          subjob ${{needs.rust_lint.result}} "Rust lints"
+          subjob ${{needs.rust_check.result}} "Rust checks"
+          subjob ${{needs.turbopack_rust_test1.result}} "Turbopack Rust tests (linux)"
```
alexkirsz, line 1098: It's weird that this name is different from that of the job itself (Turbopack Rust testing on ubuntu).

alexkirsz (approved):

tknickman (approved):
Love it, thanks!

.github/workflows/test.yml
sokra: We might move more tests from 2 to 1 when they have faster runners... I think it's more part 1 and 2, instead of linux in particular

alexkirsz: Maybe rename to (blocking) and (non-blocking)?

PR #3653, opened 2023-02-06
Reuse TurboTasks cross turbotrace runs

Brooooooklyn:
Allow reusing `TurboTasks` cross multiple `turbotrace` runs, can create `TurboTasks` in the next.js side and pass it back to reuse it.

Partially Close WEB-562

### DIFF
--- a/crates/node-file-trace/src/lib.rs
+++ b/crates/node-file-trace/src/lib.rs
@@ -343,7 +343,10 @@ fn process_input(dir: &Path, context: &str, input: &[String]) -> Result<Vec<Stri
         .collect()
 }
 
-pub async fn start(args: Arc<Args>) -> Result<Vec<String>> {
+pub async fn start(
+    args: Arc<Args>,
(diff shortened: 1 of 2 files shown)


jridgewell (approved):

cli/internal/ffi/proto/messages.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.28.0
```
sokra, line 3: Why was this changed in this PR? Why is this file committed at all if it's generated?

sokra (changes requested):

cli/internal/ffi/proto/messages.pb.go
Brooooooklyn: it seems to be changed by my `rust-analyzer`, reverted.

sokra (approved):

PR #3656, opened 2023-02-06
fix: No longer error on pnpm lockfile specifier mismatch

### LINKED ISSUE #3638: [turborepo] Turbo prune failing

1.7.3

pnpm

Linux

### Describe the Bug

Our CD started failing after the introduction of turbo@1.7.3, which we're using to prune the monorepo to reduce size. It looks like it's got to do with a specific dependency, namely `@nomiclabs/hardhat-ethers@npm:hardhat-deploy-ethers@0.3.0-beta.13`, though `typescript@4.8.2` appears in the log as well.

Tested on 1.7.0-1.7.2, works perfectly there.

image

### Expected Behavior

image

### To Reproduce

```
pnpm i -g turbo@1.7.3
git clone https://github.com/sushiswap/sushiswap.git
cd sushiswap
turbo prune --scope=earn
```

### Reproduction Repo

http://github.com/sushiswap/sushiswap

chris-olszewski:
Fixes #3638

I added a regression in #3611 where we would error if there was a specifier mismatch and the specifier wasn't totally resolved. This was an unintended change to the previous behavior would indicate it didn't find a package resolution.

Tested against [provided reproduction](https://github.com/sushiswap/sushiswap)


mehulkar (commented):
Is a test case possible to write for this?

chris-olszewski:
> Is a test case possible to write for this?

Yes, isolating it took longer than expected. Underlying issue is don't have proper support for [overrides](https://pnpm.io/package_json#pnpmoverrides). I added basic support that at least properly supports the repro we were provided, it's not a complete fix, but it's a move in the right direction.

mehulkar (approved):

PR #3658, opened 2023-02-06
Warn on absolute paths used in turbo.json

gsoltis:
In this PR:
 - un-export `readTurboConfig`
 - add warnings via global `log` to instances of absolute paths used in `turbo.json` `globalDependencies`, and per-task `inputs`, and `outputs`

Known issues:
 - This is definitely not the best way to add this feature, but is minimally disruptive. During the port, we ought to ensure that we are parsing `turbo.json` and then in a separate pass validating the data so we can wire errors and warnings to the proper ui infrastructure
 - We currently read `turbo.json` twice, as well as the root `package.json`. Until we refactor that, warnings will show up twice.

### DIFF
--- a/cli/internal/fs/turbo_json.go
+++ b/cli/internal/fs/turbo_json.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 
@@ -163,7 +164,7 @@ func LoadTurboConfig(dir turbopath.AbsoluteSystemPath, rootPackageJSON *PackageJ
 	}
 
 	var turboJSON *TurboJSON
-	turboFromFiles, err := ReadTurboConfig(dir.UntypedJoin(configFile))
+	turboFromFiles, err := readTurboConfig(dir.UntypedJoin(configFile))
 
 	if !includeSynthesizedFromRootPackageJSON && err != nil {
 		// If the file didn't exist, throw a custom error here instead of propagating
@@ -250,8 +251,8 @@ func (to TaskOutputs) Sort() TaskOutputs {
 	return TaskOutputs{Inclusions: inclusions, Exclusions: exclusions}
 }
 
-// ReadTurboConfig reads turbo.json from a provided path
-func ReadTurboConfig(turboJSONPath turbopath.AbsoluteSystemPath) (*TurboJSON, error) {
+// readTurboConfig reads turbo.json from a provided path
+func readTurboConfig(turboJSONPath turbopath.AbsoluteSystemPath) (*TurboJSON, error) {
 	// If the configFile exists, use that
 	if turboJSONPath.FileExists() {
 		turboJSON, err := readTurboJSON(turboJSONPath)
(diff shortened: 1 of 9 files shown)


cli/internal/fs/turbo_json.go
```diff
@@ -184,8 +185,8 @@ func (to TaskOutputs) Sort() TaskOutputs {
 	return TaskOutputs{Inclusions: inclusions, Exclusions: exclusions}
 }
 
-// ReadTurboConfig reads turbo.json from a provided path
-func ReadTurboConfig(turboJSONPath turbopath.AbsoluteSystemPath) (*TurboJSON, error) {
+// readTurboConfig reads turbo.json from a provided path
+func readTurboConfig(turboJSONPath turbopath.AbsoluteSystemPath) (*TurboJSON, error) {
```
mehulkar, line 189: Can you leave this public? It was in preparation for #2706, so i'll just have to revert it again

mehulkar (commented):
Want to add a prysk test case? Would be good to get validation during port

gsoltis:
Re: `prysk` test, I started to add one, then decided I didn't want a test encoding the double warning.

Maybe this ought to wait until #2706 is merged.

mehulkar:
> Maybe this ought to wait until https://github.com/vercel/turbo/pull/2706 is merged.

Up to you, but other than exporting the method, I don't see a reason to wait. Is there any reason this PR needs the method to be un-exported?

gsoltis:
The exporting isn't the issue. If we're going to include a test, I'd like the refactor to only read `turbo.json` once to happen so we aren't encoding a double warning in our tests. I expect the refactor doesn't make sense to do until after #2706 lands

gsoltis:
@mehulkar please take another look. I've rebased this on to your PR. It looks like we don't need to re-export `readTurboConfig`, unless you've got another use for it coming.

I was also able to add an integration test that is at least close to the behavior we want to preserve. The `log` outputs are slightly different from the rest of our UI, but I think close enough.

mehulkar (commented):
Needs a rebase to be sure, but at my last reading, the diff looked good

cli/internal/fs/turbo_json.go
```diff
@@ -465,6 +472,12 @@ func (btd *BookkeepingTaskDefinition) UnmarshalJSON(data []byte) error {
 		// Note that we don't require Inputs to be sorted, we're going to
 		// hash the resulting files and sort that instead
 		btd.definedFields.Add("Inputs")
+		// TODO: during rust port, this should be moved to a post-parse validation step
+		for _, input := range task.Inputs {
+			if filepath.IsAbs(input) {
+				log.Printf("[WARNING] Using an absolute path in \"inputs\" (%v) will not work and will be an error in a future version", input)
+			}
```
mehulkar, line 479: I really wish we had the infrastructure to make this an error _now_, and add a config to opt into the old (incorrect) behavior.

cli/internal/fs/turbo_json.go
```diff
@@ -551,11 +564,15 @@ func (c *TurboJSON) UnmarshalJSON(data []byte) error {
 		envVarDependencies.Add(value)
 	}
 
+	// TODO: In the rust port, warnings should be refactored to a post-parse validation step
```
mehulkar, line 567: For what it's worth we have a `TurboJSON.Validate()` method that can do this now. Since this isn't throwing an error though, it would have to be adapted slightly to emit warnings rather than errors.

mehulkar (approved):
Looks good, thanks for waiting and for adding a test!

PR #3659, opened 2023-02-06
ci(workflow): expand integration test groups

kwonoj:
closes WEB-545, WEB-571.

This PR mimic upstream next.js test group config to expand turbopack's integration integration to match with it. To reduce some of duplication, this PR also sets up reusable workflow to setup  / build next.js and share its build artifact via github cache. Since those build artifacts are one-off usage per each test, it doesn't use upload/download artifact which is too slow & heavyweight for those purpose.

### DIFF
--- a/.github/actions/next-integration-stat/src/index.ts
+++ b/.github/actions/next-integration-stat/src/index.ts
@@ -71,13 +71,12 @@ const BOT_COMMENT_MARKER = `<!-- __marker__ next.js integration stats __marker__
 // Header for the test report.
 const commentTitlePre = `## Failing next.js integration test suites`;
 
-// Download logs for a job in a workflow run by reading redirect url from workflow log response.
-async function fetchJobLogsFromWorkflow(
+async function findNextJsVersionFromBuildLogs(
   octokit: Octokit,
   token: string,
   job: Job
-): Promise<{ nextjsVersion: string; logs: string; job: Job }> {
-  console.log("Checking test results for the job ", job.name);
+): Promise<string> {
+  console.log("Checking logs for the job ", job.name);
 
   // downloadJobLogsForWorkflowRun returns a redirect to the actual logs
   const jobLogRedirectResponse =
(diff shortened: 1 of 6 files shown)


kwonoj:
This may or may not fix WEB-544 as well. I need to dig further once this PR lands.

.github/actions/next-integration-stat/src/index.ts
```diff
@@ -136,35 +177,46 @@ function collectFailedTestResults(
...
-          name: failedTest,
-          data: JSON.parse(testData),
-        };
-      } catch (_) {
-        console.log(`Failed to parse test data`);
-        return null;
+      let failedSplitLogs = logs.split(`failed to pass within`);
+      const ret = [];
+
+      while (!!failedSplitLogs && failedSplitLogs.length >= 1) {
```
wbinnssmith, line 183: Can this ever be falsey?

kwonoj: yes we consume it here https://github.com/vercel/turbo/pull/3659/files#diff-0e0c2f9cb52224d172853cfbed972f9ac25436bf53078850cd7b3a7168fef0e7R184 so eventually it'll drain

.github/actions/next-integration-stat/src/index.ts
```diff
@@ -566,13 +654,13 @@ function getTestSummary(
 
   if (fixedTests.length > 0) {
     ret += `\n:white_check_mark: **Fixed tests:**\n\n${fixedTests
-      .map((t) => `\t- ${t}`)
+      .map((t) => (t.length > 5 ? `\t- ${t}` : t))
```
wbinnssmith, line 657: Maybe a comment explaining why 5 is significant?

kwonoj: it's actually no significant random number 😅 I just picked small enough to ensure it have corresponding filename in the logs.

wbinnssmith (approved):

PR #3666, opened 2023-02-07
docs: add package.json and turbo.json to a pruned directory example

m-shaka:
I fixed two mistakes.
- `./out/json` has `package.json` and `./out/full` has `package.json` and `turbo.json` in fact.
- You don't need to copy `turbo.json` manually in Dockerfile because `turbo.json` is already copied in [the previous line](https://github.com/vercel/turbo/compare/main...m-shaka:turbo:docs/fix-pruned-directory-structure?expand=1#diff-3d80bc97cc73148cf5956531eb631ce3e7f58860e778c4b52cf2e3499f4c4f05R139)

### DIFF
--- a/docs/pages/repo/docs/handbook/deploying-with-docker.mdx
+++ b/docs/pages/repo/docs/handbook/deploying-with-docker.mdx
@@ -87,14 +87,17 @@ First, we want to copy over only what we need to install the packages. When runn
 ```txt
 out
 ├── json
-│   └── apps
-│       └── docs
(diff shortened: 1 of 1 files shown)


arlyon:
Ah! You are right. We at some point started processing turbo.json in the prune command and forgot to update the docs. I am approving this, thanks for the contribution 🎉

arlyon (approved):

mehulkar (approved):

PR #3671, opened 2023-02-07
feat: port daemon client to rust

arlyon:
We still need to maintain the go client until run is ported, as the go side needs it to save and request changed files. Now that the daemon client is ported, the next step is to port the globbing tools needed for the daemon server followed by the server itself.

Please note this is currently using a fork of the pidlock crate.


crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,278 @@
...
+use notify::{Config, Event, EventKind, RecommendedWatcher, Watcher};
+use sysinfo::{ProcessExt, ProcessRefreshKind, RefreshKind, SystemExt};
+use tokio::{net::UnixStream, sync::mpsc, time::timeout};
+use tonic::transport::Endpoint;
+
+use super::{client::proto::turbod_client::TurbodClient, DaemonClient, DaemonError};
+
+#[derive(Debug)]
+pub struct DaemonConnector {
+    pub dont_start: bool,
```
NicholasLYang, line 19: I find it a little confusing having negatives as booleans. Could we rephrase these as positives?

NicholasLYang (commented):
Looks great! Just a couple questions

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -0,0 +1,62 @@
...
+                uptime_ms: status.uptime_msec,
+                log_file: status.log_file.into(),
+                pid_file: client.connect_settings.pid_file.clone(),
+                sock_file: client.connect_settings.sock_file.clone(),
+            };
+            if *json {
+                println!("{}", serde_json::to_string_pretty(&status).unwrap());
+            } else {
+                println!("Daemon log file: {}", status.log_file.to_string_lossy());
+                println!("Daemon uptime: {}s", status.uptime_ms / 1000);
```
chris-olszewski, line 46: I realize this will be a UI change, but I really like using [`humantime`](https://docs.rs/humantime/latest/humantime/)

crates/turborepo-lib/Cargo.toml
```diff
@@ -26,23 +26,38 @@ axum-server = "0.4.4"
...
 dirs-next = "2.0.0"
 dunce = "1.0"
 env_logger = "0.10.0"
+futures = "0.3.26"
+hex = "0.4.3"
 indicatif = "0.17.3"
 lazy_static = "1.4.0"
 log = "0.4.17"
+notify = "5.1.0"
+pidlock = { version = "0.1.4", path = "../../../pidlock" }
```
chris-olszewski, line 42:
```suggestion
pidlock = { git = "https://github.com/arlyon/pidlock", branch = "feat/get-owner" }
```

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -0,0 +1,62 @@
...
+        sock_file: base.daemon_file_root().join("turbod.sock"),
+    };
+
+    let mut client = connector.connect().await?;
+
+    match command {
+        DaemonCommand::Restart => {
+            client.restart().await?;
+        }
+        DaemonCommand::Start => {} // no-op
```
chris-olszewski, line 30: A comment explaining that `connector.connect()` will start the server would be useful as to why this is a noop.

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,278 @@
...
+            match (client.handshake().await, &self.dont_kill) {
+                (Ok(_), _) => return Ok(client.with_connect_settings(self)),
+                // should be able to opt out of kill
+                (Err(DaemonError::VersionMismatch), true) => {
+                    return Err(DaemonError::VersionMismatch)
+                }
+                (Err(DaemonError::VersionMismatch), false) => {
+                    self.kill_live_server(client, pid).await?;
+                    continue;
+                }
```
chris-olszewski, line 66: Could we hide the matching on `dont_kill` by putting this logic into a `maybe_kill_live_server`?
```suggestion
                Err(DaemonError::VersionMismatch) => {
                    self.maybe_kill_live_server(client, pid).await?;
                    continue;
                }
```

crates/turborepo-lib/src/daemon/connector.rs
nathanhammond: I'm like 90% sure this is because the Go implementation cares about zero-states.

crates/turborepo-lib/Cargo.toml
```diff
@@ -26,23 +26,38 @@ axum-server = "0.4.4"
 chrono = { version = "0.4.23", features = ["serde"] }
 clap = { version = "4.0.22", features = ["derive"] }
 clap_complete = "4.0.6"
+command-group = { version = "2.0.1", features = ["with-tokio"] }
```
nathanhammond, line 29: From the same effort:
https://github.com/watchexec/watchexec

In case we can learn from them.

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -0,0 +1,62 @@
...
+    daemon::{DaemonConnector, DaemonError},
+};
+
+/// Runs the daemon command.
+pub async fn main(command: &DaemonCommand, base: &CommandBase) -> Result<(), DaemonError> {
+    let (dont_start, dont_kill) = match command {
+        DaemonCommand::Status { .. } => (true, true),
+        DaemonCommand::Restart | DaemonCommand::Stop => (true, false),
+        DaemonCommand::Start => (false, false),
+    };
```
nathanhammond, line 15: Is this idiomatic Rust? (I'm not the person to ask.) It feels very strange to have the assignment be the driver of behavior.

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -102,4 +104,14 @@ impl CommandBase {
...
     }
+
+    pub fn daemon_file_root(&self) -> PathBuf {
+        std::env::temp_dir().join("turbod").join(self.repo_hash())
+    }
+
+    fn repo_hash(&self) -> String {
+        let mut hasher = Sha256::new();
+        hasher.update(self.repo_root.to_str().unwrap().as_bytes());
+        hex::encode(&hasher.finalize()[..8])
```
nathanhammond, line 115: The Go implementation is 16 characters, I didn't check in to see what this did but the eight stands out to me. Can we get a very simple test of this function to demonstrate the behavior?

(I'm also hoping that `std::env::temp_dir()` and Go's `os.TempDir()` are 1:1, but if they're not, oh well.)

***

Also, in a future world, I'd prefer if this had an identifier that was distinct from file path. Something like:

`git rev-list --max-parents=0 --topo-order HEAD`

The first SHA spit out by this would be perfect for addressing.

(Failing `git` would fallback to file path.)

crates/turborepo-lib/src/daemon/client.rs
```diff
@@ -0,0 +1,134 @@
...
+    GrpcFailure(tonic::Code),
+    #[error("missing response")]
+    MissingResponse,
+    #[error("could not read pid file")]
+    PidFile,
+    #[error("could not connect: {0}")]
+    Timeout(#[from] tokio::time::error::Elapsed),
+    #[error("daemon is not running and will not be started")]
+    NotRunning,
+}
```
nathanhammond, line 124: Do these show up in our UI? If so I'd love to make them a bit more accessible.

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// dont_kill is unset and one of these cases is hit:
+    /// 1. the versions do not match
+    /// 2. the server is not running
+    /// 3. the server is unresponsive
+    pub async fn connect(self) -> Result<DaemonClient<DaemonConnector>, DaemonError> {
+        for _ in 0..Self::CONNECT_RETRY_MAX {
+            let pid = self.get_or_start_daemon().await?;
+            debug!("got daemon with pid: {}", pid);
+
+            let path = match self.wait_for_socket().await {
```
nathanhammond, line 46: I _think_ there is a race condition here.

Does the `wait_for_file` call come after `get_or_start_daemon`? Is it possible that we could end up waiting for something to happen separately from this process?

Also, what's the latency and overhead cost on `wait_for_file` vs. just optimistically reading it after it should already be there?

nathanhammond, line 46: This is called in a loop. Is it possible that the timeout implementation could have the `wait_for_file` registration miss the timeout gaps?

nathanhammond (commented):
First pass.

crates/turborepo-lib/src/cli.rs
```diff
@@ -475,10 +475,19 @@ pub async fn run(repo_state: Option<RepoState>) -> Result<Payload> {
 
             Ok(Payload::Rust(Ok(0)))
         }
+        Command::Daemon {
+            command: Some(command),
+            ..
+        } => {
+            let command = command.clone();
```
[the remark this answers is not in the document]
arlyon: We are matching on clap_args and then passing the whole thing into CommandBase, while command is borrowed on the line immediately after, so we need to clone it.

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,278 @@
...
+                    ..
+                }),
+                WaitAction::Missing,
+            ) => {
+                if paths.iter().any(|p| {
+                    p.file_name()
+                        .map(|f| file_name.as_os_str().eq(f))
+                        .unwrap_or_default()
+                }) {
+                    futures::executor::block_on(async {
```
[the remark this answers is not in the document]
arlyon: The closure you pass to notify is synchronous, while we are using async channels. `executor::block_on` lets us run a future (`tx.send`) in a synchronous context.

crates/turborepo-lib/src/daemon/connector.rs
arlyon: Agreed. Decided to keep the old phrasing but I will change it :)

crates/turborepo-lib/src/daemon/client.rs
arlyon: some of them will bubble up. I am going to clean up the error paths and expose the underlying ones a little better

crates/turborepo-lib/src/commands/daemon.rs
arlyon: I didn't want the DaemonConnector to be aware of the Command, and this looks to me like quite a concise and declarative way to set the flags.

crates/turborepo-lib/src/daemon/connector.rs
arlyon: The go implementation just used a backoff for this, but I wanted to get notify in the codebase somewhere ahead of the file watcher stuff.

inotify / kqueue and friends should be pretty fast. The server takes up to a few 100ms on my machine to set up the socket so I would assume that there is more IO activity from polling than setting up a watcher. I think when the daemon server lands I will refactor this a little more into a common utility.

arlyon: Yes. `wait_for_file` _does_ check if the file exists just in case. I will move that condition to _after_ the watcher is registered so that it is impossible for the file to be created between that exists check and the watcher.

arlyon: Notify also supports a polling instead which should be pretty much identical to what we had on the go side, but with an extra runtime cost.

crates/turborepo-lib/src/commands/mod.rs
gsoltis: If we use something other than file path, we need to account for multiple copies of the same repository.

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,337 @@
...
+
+            let mut client = DaemonClient {
+                client: conn,
+                connect_settings: (),
+            };
+
+            match client.handshake().await {
+                Ok(_) => {
+                    return {
+                        debug!("connected in {}ns", time.elapsed().as_micros());
```
arlyon, line 91: Oops, will fix.

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,337 @@
...
+        // this creates a new process group for the given command
+        // in a cross platform way, directing all output to /dev/null
+        let mut group = tokio::process::Command::new(binary_path)
+            .arg("daemon")
+            .stderr(Stdio::null())
+            .stdout(Stdio::null())
+            .group()
+            .kill_on_drop(false)
+            .spawn()
+            .map_err(|_| DaemonConnectorError::Fork)?;
```
gsoltis, line 140: Here, and elsewhere, it looks like we're losing error details. Can we log at debug level what the error was so that we have somewhere to start when something isn't working?

gsoltis (changes requested):
As a general comment, we should port the tests as well. I know at least `connector` has a set of unit tests.

chris-olszewski (commented):
This looks really solid to me, I just have a few nits and questions that are mostly for to better my understanding.

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -0,0 +1,67 @@
...
+        DaemonCommand::Stop => {
+            client.stop().await?;
+        }
+        DaemonCommand::Status { json } => {
+            let status = client.status().await?;
+            let status = DaemonStatus {
+                uptime_ms: status.uptime_msec,
+                log_file: status.log_file.into(),
+                pid_file: client.connect_settings.pid_file.clone(),
+                sock_file: client.connect_settings.sock_file.clone(),
```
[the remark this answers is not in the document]
gsoltis: They should not. These files need to be at a well-known location.

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,548 @@
...
+        std::fs::write(&sock, "").unwrap();
+
+        let connector = DaemonConnector {
+            pid_file: pid,
+            sock_file: sock,
+            can_kill_server: true,
+            can_start_server: false,
+        };
+
+        assert_matches!(
```
gsoltis, line 541: Can you add an assertion about what happens to the pid file?

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,548 @@
...
+            connector
+                .kill_dead_server(Pid::from(proc.id().unwrap() as usize))
+                .await,
+            Ok(())
+        );
+    }
+}
```
gsoltis, line 548: Is it possible to port `TestKillLiveServer` from Go as well?

crates/turborepo-lib/Cargo.toml
```diff
@@ -26,26 +27,46 @@ axum-server = "0.4.4"
...
 glob-match = "0.2.1"
+hex = "0.4.3"
 hostname = "0.3.1"
+humantime = "2.1.0"
 indicatif = { workspace = true }
 lazy_static = { workspace = true }
 log = { workspace = true }
-predicates = { workspace = true }
+notify = "5.1.0"
+pidlock = { git = "https://github.com/arlyon/pidlock", branch = "master" }
```
gsoltis, line 46: I think we need to vendor and review this if we want to use it.

gsoltis (changes requested):

crates/turborepo-lib/src/daemon/connector.rs
arlyon: Yes, but I was planning on doing that when porting the server to rust because trying to launch a go server (and potentially rust binary) and all the path shenanigans around that feels like a nightmare given how short-lived it'll be.

crates/turborepo-lib/Cargo.toml
```diff
@@ -26,26 +27,46 @@ axum-server = "0.4.4"
...
+futures = "0.3.26"
 glob-match = "0.2.1"
+hex = "0.4.3"
 hostname = "0.3.1"
+humantime = "2.1.0"
 indicatif = { workspace = true }
 lazy_static = { workspace = true }
 log = { workspace = true }
-predicates = { workspace = true }
+notify = "5.1.0"
```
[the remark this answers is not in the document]
arlyon: It can't hurt. I will flip that flag.

crates/turborepo-lib/Cargo.toml
arlyon: I am waiting for some PRs but I will vendor for now and we can consider breaking it out again once they merge. (https://github.com/rockstar/pidlock/pulls)

crates/turborepo-lib/src/daemon/client.rs
```diff
@@ -0,0 +1,134 @@
...
+        self.client
+            .notify_outputs_written(proto::NotifyOutputsWrittenRequest {
+                hash,
+                output_globs,
+                output_exclusion_globs,
+            })
+            .await?;
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
arlyon: Good point!

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,548 @@
...
+
+            let conn = match self.get_connection(self.sock_file.clone()).await {
+                Err(DaemonConnectorError::Watcher(_) | DaemonConnectorError::Socket(_)) => continue,
+                rest => rest?,
+            };
+
+            let mut client = DaemonClient {
+                client: conn,
+                connect_settings: (),
+            };
```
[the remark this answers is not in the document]
arlyon: That small changed has cleaned up a lot, thanks!

crates/turborepo-lib/src/daemon/connector.rs
arlyon: I will do it in this PR

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,548 @@
...
+            can_kill_server: false,
+            can_start_server: false,
+        };
+
+        assert_matches!(
+            connector.kill_dead_server(Pid::from(usize::MAX)).await,
+            Ok(())
+        );
+
+        assert!(!connector.pid_file.exists(), "pid file should be deleted");
```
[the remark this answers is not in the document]
arlyon: When the pid lock is opened, the library deletes the lock if there is no process that owns it. Should we change that behaviour?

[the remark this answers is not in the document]
gsoltis: As discussed, yes, let's avoid deleting the pid file from the client to avoid race conditions. In the future we can add a flag or subcommand to allow the user to specify that we should attempt a cleanup of bad or stale files.

gsoltis (changes requested):
Let's add tests for the pid file behavior and update paths to be the correct marker types.

.vscode/settings.json
```diff
@@ -21,7 +21,7 @@
     "files.insertFinalNewline": false,
     "files.trimTrailingWhitespace": false
   },
-  "rust-analyzer.linkedProjects": ["nextpack/Cargo.toml"],
+  // "rust-analyzer.linkedProjects": ["nextpack/Cargo.toml"],
```
gsoltis, line 24: Fixship: I don't think this PR hsould make this change, but the turbopack team is removing this @jridgewell

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -0,0 +1,664 @@
...
+    }
+
+    fn pid_lock(&self) -> pidlock::Pidlock {
+        pidlock::Pidlock::new(self.pid_file.clone().into_path_buf())
+    }
+}
+
+#[cfg(target_os = "windows")]
+fn win(
+    path: Arc<PathBuf>,
```
gsoltis, line 264: Fixship: windows tests are failing, I think this is now an `AbsoluteNormalizedPath`

gsoltis (approved):
I left a couple of fixship comments, and the windows tests are currently failing (see comments).

Other than that, I think this is good. Thanks for being patient getting this done. The infrastructure around the daemon has traditionally been an area of pretty heavy review focus.

PR #3675, opened 2023-02-07
refactor ResolveResult to allow mixing special and assets in alternatives

sokra:


crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -444,13 +450,14 @@ async fn chunk_content_internal<I: FromChunkableAsset>(
...
                 let mut inner_chunk_groups = Vec::new();
 
-                for asset in assets
-                    .await?
-                    .iter()
-                    .filter(|asset| processed_assets.insert((chunking_type, **asset)))
-                {
-                    let asset: &AssetVc = asset;
-
+                for asset in result.await?.primary.iter().filter_map(|result| {
```
jridgewell, line 453: Nit: Can we assign this RHS to a variable. The `for` nesting here becomes weird because of the multi-line function.

crates/turbopack-ecmascript/src/references/node.rs
```diff
@@ -72,7 +72,7 @@ impl AssetReference for DirAssetReference {
         let context_path = self.source.path().await?;
         // ignore path.join in `node-gyp`, it will includes too many files
         if context_path.path.contains("node_modules/node-gyp") {
-            return Ok(ResolveResult::Alternatives(Vec::new(), vec![]).into());
+            return Ok(ResolveResult::assets_with_references(Vec::new(), vec![]).into());
```
jridgewell, line 75: Looks like we're intentionally not resolving this? So 
```suggestion
            return Ok(ResolveResult::unresolveable().into());
```

jridgewell (approved):

PR #3676, opened 2023-02-07
improve error reporting of JSON parsing

sokra:
also make source context generation more reusable

This adds a source context snippet of the JSON that failed parsing and the json path to the error.

```
An error occurred while generating the chunk item [project]/crates/turbopack-tests/tests/snapshot/imports/json/input/invalid.json (json)
  at Execution of module_factory failed
  at Execution of JsonChunkItem::content failed
  at Unable to make a module from invalid JSON: expected `,` or `}` at line 3 column 26
  at nested.?
     1 | {
     2 |   "nested": {
       |                          v
     3 |     "this-is": "invalid" // lint-staged will remove trailing commas, so here's a comment
       |                          ^
     4 |   }
     5 | }
```


crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1403,10 +1408,15 @@ impl FileContent {
...
-                Err(_) => FileJsonContent::Unparseable,
-            },
+            FileContent::Content(file) => {
+                let de = &mut serde_json::Deserializer::from_reader(file.read());
+                match serde_path_to_error::deserialize(de) {
+                    Ok(data) => FileJsonContent::Content(data),
+                    Err(e) => FileJsonContent::Unparseable(
+                        box UnparseableJson::from_serde_path_to_error(e),
+                    ),
+                }
```
jridgewell, line 1418: Should this be using `parse_json_rope_with_source_context`?

jridgewell (approved):

crates/turbo-tasks-fs/src/lib.rs
sokra: No, since we don't want it wrapped in an anyhow error here, but return a structured error.

crates/turbo-tasks-fs/src/json.rs
```diff
@@ -0,0 +1,120 @@
...
+            path: Some(e.path().to_string()),
+            start_location: Some((inner.line() - 1, inner.column() - 1)),
+            end_location: None,
+        }
+    }
+
+    pub fn write_with_content(&self, writer: &mut impl Write, text: &str) -> std::fmt::Result {
+        writeln!(writer, "{}", self.message)?;
+        if let Some(path) = &self.path {
+            writeln!(writer, "  at {}", path)?;
```
alexkirsz, line 58: I think the `at` is redundant here. We can remove it and the error would read more easily.

crates/turbopack-json/src/lib.rs
```diff
@@ -116,20 +118,36 @@ impl EcmascriptChunkItem for JsonChunkItem {
...
-        Ok(EcmascriptChunkItemContent {
-            inner_code: inner_code.into(),
-            ..Default::default()
+        let content = self.module.path().read();
+        let data = content.parse_json().await?;
+        match &*data {
+            FileJsonContent::Content(data) => {
+                let js_str_content = serde_json::to_string(data)?;
+                let inner_code =
+                    format!("__turbopack_export_value__(JSON.parse({js_str_content}));");
```
ForsakenHarmony, line 127: This is broken, it no longer "stringifies the string" and just puts the raw JSON into the file

PR #3681, opened 2023-02-07
ci(workflow): adjust summary report, drain all logs correctly

kwonoj:
Closes WEB-544.

PR fixes minor issue to drain logs correctly. 

Also PR adjusts summary to not to print fixed tests - upstream test sometimes flaky and displaying some test as `fixed` intermittently can cause some confusions. For now, disabling those summary. Still we display `new failed`, false negatives are relatively not occur frequently and we can try for a while to see.

### DIFF
--- a/.github/actions/next-integration-stat/index.js
+++ b/.github/actions/next-integration-stat/index.js
@@ -16005,10 +16005,11 @@
         })
         .map((logs) => {
           var _a, _b, _c, _d;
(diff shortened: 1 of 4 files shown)


kwonoj:
Captured note:

image

kwonoj:
`css-client-nav/test` : flaky in upstream

PR #3686, opened 2023-02-07
Add a page about installing turbo, including both global and repo-local

gsoltis:
Refactor some of the installation details to a common spot.
Include a note about scoping to a workspace via global turbo
Include a note about pinning versions via local install


docs/pages/repo/docs/getting-started/add-to-project.mdx
```diff
@@ -40,11 +40,13 @@ npm create vite@latest
   </Tab>
   <Tab>
     ```bash
-    pnpm install turbo --global
+    pnpm add turbo --global
```
tknickman, line 43: `install` should be correct here. (add is just `yarn`)

docs/pages/repo/docs/installing.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+<Tabs items={['npm', 'yarn', 'pnpm']} storageKey="selected-pkg-manager">
+  <Tab>
+    ```bash
+    npm install turbo --dev
+    ```
+  </Tab>
+  <Tab>
+    ```bash
+    yarn add turbo --dev --ignore-workspace-root-check
```
tknickman, line 73: I don't think we should recommend using `--ignore-workspace-root-check` if we're also instructing to install "...as a dev dependency at the root of the repository:"

docs/pages/repo/docs/installing.mdx
```diff
@@ -0,0 +1,85 @@
...
+title: Installing Turborepo
+description: Learn how to install Turborepo for use with your repository
+---
+
+import Callout from "../../../components/Callout";
+import { Tabs, Tab } from '../../../components/Tabs'
+
+# Install Turborepo
+
+`turbo` works with [Yarn](https://classic.yarnpkg.com/lang/en/), [npm](https://www.npmjs.com/), and [pnpm](https://pnpm.io/) on the following operating systems:
```
NicholasLYang, line 11:
```suggestion
`turbo` works with [yarn](https://classic.yarnpkg.com/lang/en/), [npm](https://www.npmjs.com/), and [pnpm](https://pnpm.io/) on the following operating systems:
```

docs/pages/repo/docs/installing.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+- macOS darwin 64-bit (Intel), ARM 64-bit (Apple Silicon)
+- Linux 64-bit, ARM 64-bit
+- Windows 64-bit, ARM 64-bit
+
+<Callout>
+  Note: Linux builds of `turbo` link against `glibc`. For Alpine Docker environments, you will need to ensure libc6-compat is installed as well, via `RUN apk add --no-cache libc6-compat`
```
NicholasLYang, line 18: Do we still link against glibc? I thought we switched to musl?

tknickman: You do still need `libc6-compat` in alpine.

docs/pages/repo/docs/installing.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+Once you have a globally installed copy of `turbo`, you will be able to run directly from workspace
+directories.
+
+```bash
+cd <repository root>/apps/docs
+turbo build
+```
+
+is equivalent to
```
mehulkar, line 52:
```suggestion
is equivalent to the [filtering syntax](../core-concepts/monorepos/filtering)
```suggestion

File: docs/pages/repo/docs/installing.mdx [hunk 6]
```diff
@@ -0,0 +1,85 @@
...
+
+```bash
+cd <repository root>
+turbo build --filter=docs
+```
+
+## Install Per Repository
+
+You may wish to pin the version of Turborepo used within a repository, especially if you are
+collaborating with other developers. In that case, add `turbo` as a dev dependency at the root
```
mehulkar, line 62:
```suggestion
You may wish to pin the version of Turborepo used within a repository, especially [if you are
collaborating with other developers](../core-concepts/remote-caching). In that case, add `turbo` as a dev dependency at the root
```

mehulkar (approved):

docs/pages/repo/docs/getting-started/add-to-project.mdx
gsoltis: It might work, but the `pnpm` docs suggest that [install](https://pnpm.io/cli/install) is for installing everything, whereas [add](https://pnpm.io/cli/add) is for adding a new dependency

docs/pages/repo/docs/installing.mdx
gsoltis: I'm not sure I understand. `yarn` will report an error if you don't add that flag. You use that flag to dependencies "at the root of the repository": https://classic.yarnpkg.com/en/docs/cli/add#toc-yarn-add-ignore-workspace-root-check-w

docs/pages/repo/docs/installing.mdx
gsoltis: I cut and paste'd this, and it's a little bit of a simplification. The real answer is something like:
"The Go executable portion of Turborepo expects to dynamically link against glibc. This means that Alpine Linux users will need to ensure the `libc6-compat` package is installed. To install `libc6-compat` when creating a Docker image based on Alpine Linux, include the line `RUN apk add --no-cache libc6-compat` in your Dockerfile"

FWIW I think that's a bit too much detail, but I'm open to suggestions on a rewording here if anyone has some good ideas.

docs/pages/repo/docs/getting-started/add-to-project.mdx
tknickman: Ah TIL. That's a bit annoying since they mirror NPM in _almost_ everything else....

docs/pages/repo/docs/installing.mdx
tknickman: Ah I've never actually seen these full commands, I've only seen their shorthand (`-w` for pnpm and `-W` for yarn). 

All good.

PR #3687, opened 2023-02-07
docs(turborepo): Enumerate supported install scenarios

NicholasLYang:


CONTRIBUTING.md
```diff
@@ -89,6 +89,33 @@ TURBO_BINARY_PATH=~/repos/vercel/turbo/cli/turbo.exe npm link turbo
 
 If you're using a different package manager replace npm accordingly.
 
+## Manually testing `turbo`
+
+Before releasing, it's recommended to test the `turbo` binary manually.
+Here's a checklist of testing strategies to cover:
+
+- Test `login`, `logout`, `login --sso-team`, `link`, `unlink`
```
chris-olszewski, line 97: Just adding this here, I'll make a script that enumerates all of the package managers and update this in another PR.
```suggestion
- Test `login`, `logout`, `login --sso-team`, `link`, `unlink`
- Test `prune` (Note `turbo` here is the unreleased turbo binary)
  - `npx create-turbo --use-pnpm prune-test && cd prune-test`
  - `turbo --skip-infer prune --scope=docs && cd out && pnpm install --frozen-lockfile`
  - `turbo --skip-infer build`
```

CONTRIBUTING.md
```diff
@@ -89,6 +89,33 @@ TURBO_BINARY_PATH=~/repos/vercel/turbo/cli/turbo.exe npm link turbo
...
+- Test every command with `--skip-infer`.
```
chris-olszewski, line 98: Is it necessary to test this codepath separate from the global only installation scenario?
(Calling this out as it brings down the test matrix this section describes from 60 tests to 30)

CONTRIBUTING.md
```diff
@@ -89,6 +89,33 @@ TURBO_BINARY_PATH=~/repos/vercel/turbo/cli/turbo.exe npm link turbo
...
+- Test `--dry-run` and `--graph`.
+- Test with and without daemon.
+
+There are also multiple installation scenarios worth testing:
+
+- Global-only. `turbo` is installed as global binary, no local `turbo` in repository.
+- Local-only. `turbo` is installed as local binary, no global `turbo` in PATH. turbo` is invoked via a root package script.
+- Global + local. `turbo` is installed as global binary, and local `turbo` in repository. Global `turbo` delegates to local `turbo`
+- npx. `turbo` is invoked via `npx turbo`.
+- npx + local. `turbo` is invoked via `npx turbo` in a repository with a local `turbo` binary.
```
chris-olszewski, line 108: Do these scenarios need to be tested separately?
- If there isn't a local install, then `npx turbo` is the same as `turbo` (assuming `PATH` is correctly setup) which should be the same codepath as global only turbo
- If there's a local install then `npx turbo` goes down the same code path as local only since `npx turbo` invokes the local binary directly even if there's a global turbo present

chris-olszewski (changes requested):
I think we can slim down this matrix a bit to make it more reasonable for a single person to do manual QA. If these additional test scenarios are needed we should probably be explicit about what behavior we're testing so the tester can be on the lookout and not just accepting any non-zero exit code.

CONTRIBUTING.md
NicholasLYang: Yeah probably not

CONTRIBUTING.md
NicholasLYang: Hmm I guess the only difference with `npx turbo` on a local install is that the local turbo is also doing the inference and calling itself.

NicholasLYang: That does seem worth testing actually. The first situation might not be worth testing though

chris-olszewski: I'm pretty sure that's the same codepath as local only where turbo is invoked via a package script.

chris-olszewski (approved):

PR #3689, opened 2023-02-08
Combine next.config.js Turbopack options into a nested object

wbinnssmith:
This combines Turbopack-specific options into a nested object `experimental.turbopack`, exposing `loaders` and `resolveAlias` properties.

x-ref: https://github.com/vercel/next.js/pull/45560/files/631b637f42757e20aabe84dc933527dc66b55173#r1098123643

Test Plan: Integration tests for loaders and resolveAlias.

### DIFF
--- a/crates/next-core/js/src/entry/config/next.js
+++ b/crates/next-core/js/src/entry/config/next.js
@@ -17,9 +17,9 @@ const loadNextConfig = async (silent) => {
   nextConfig.exportPathMap = nextConfig.exportPathMap && {};
   nextConfig.webpack = nextConfig.webpack && {};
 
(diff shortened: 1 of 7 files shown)


kwonoj (approved):
q: if config have `experimental.turbopack`, does it automatically enables `--turbo` or it is still manual thing?

alexkirsz:
> q: if config have `experimental.turbopack`, does it automatically enables `--turbo` or it is still manual thing?

For gradual adoption, I think we'll want to allow people to have both a normal config and a turbopack config side-by-side. In which case `--turbo` should still be opt-in.

PR #3690, opened 2023-02-08
Next Router Middleware Support

jridgewell:
This updates our Next.js router, passing the `edgeInfo` manifest generated from the `middleware.js` file (or any other configured page extension).

Fixes WEB-277
Fixes WEB-370


crates/next-core/src/router.rs
```diff
@@ -154,16 +206,20 @@ fn route_executor(context: AssetContextVc, project_path: FileSystemPathVc) -> As
...
     let project_path = wrap_with_next_js_fs(project_root);
     let context = node_evaluate_asset_context(Some(get_next_build_import_map(project_path)));
-    let router_asset = route_executor(context, project_path);
-    // TODO this is a hack to get these files watched.
-    let extra_configs = extra_configs(context, project_path);
+
+    let configs = config_assets(context, project_path, next_config.page_extensions());
+    let router_asset = route_executor(context, project_path, configs);
+
+    // TODO this is a hack to get this files watched.
```
alexkirsz, line 221:
```suggestion
    // TODO this is a hack to get these files watched.
```

sokra (changes requested):

sokra (changes requested):
Add a test case to next-dev-tests

alexkirsz (approved):

crates/next-core/js/src/entry/router.ts
```diff
@@ -6,6 +6,7 @@ import loadNextConfig from "@vercel/turbopack-next/entry/config/next";
 import { makeResolver } from "next/dist/server/router.js";
 
 import "next/dist/server/node-polyfill-fetch.js";
+import * as middleware from "MIDDLEWARE_CONFIG";
```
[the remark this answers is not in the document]
sokra: Yes. Let's call `import "."` tech dept from now on^^

crates/next-core/src/router.rs
sokra: Best would be to either bundle that or at least trace that. Same goes for the externals, we should trace them. But not sure if this means we loose the performance benefit from making them external^^

crates/next-core/js/src/entry/router.ts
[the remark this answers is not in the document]
jridgewell: Can you explain why this is beneficial? How is middleware executed in a different way?

jridgewell:
Added tests (and addressed some review comments), but the tests are failing because Next.js doesn't handle the middleware yet.

crates/next-core/js/src/entry/router.ts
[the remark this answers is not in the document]
sokra: It's executed "on the edge". Next.js has some emulation for that via `vm`. There are different globals available in middleware and you can't use `require` etc.

cli/internal/ffi/proto/messages.pb.go
```diff
@@ -144,7 +144,6 @@ type GlobResp struct {
 	unknownFields protoimpl.UnknownFields
 
 	// Types that are assignable to Response:
-	//
```
sokra, line 147: huh?

alexkirsz: I'm having the same issue. I think our pre commit hooks are changing this file, which might indicate something's wrong in main.

crates/next-core/js/src/entry/router.ts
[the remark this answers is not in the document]
jridgewell: I'm still not understanding why this is necessary. We're not using `require` here, this will be transformed into the same `__turbopack_import__('[project-with-next]/my-app/middleware.ts (ecmascript)')` as every other import in this file. I don't understand why importing the middleware file requires special behavior.

cli/internal/ffi/proto/messages.pb.go
jridgewell: 😮‍💨 https://vercel.slack.com/archives/C02HEJASXGD/p1675715378860089

crates/next-core/src/router.rs
```diff
@@ -115,37 +131,73 @@ impl From<RouterIncomingMessage> for RouterResult {
...
+    Ok(EcmascriptModuleAssetVc::new(
+        config_asset,
         context,
         Value::new(EcmascriptModuleAssetType::Typescript),
         EcmascriptInputTransformsVc::cell(vec![EcmascriptInputTransform::TypeScript]),
         context.environment(),
     )
-    .as_ecmascript_chunk_placeable();
-    Ok(EcmascriptChunkPlaceablesVc::cell(vec![config_chunk]))
+    .into())
```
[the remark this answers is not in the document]
jridgewell: Done.

crates/next-core/src/next_server/context.rs
```diff
@@ -107,6 +108,26 @@ pub async fn get_server_resolve_options_context(
                 ..resolve_options_context
             }
         }
+        ServerContextType::Middleware => {
+            let resolve_options_context = ResolveOptionsContext {
+                enable_node_modules: true,
+                enable_node_externals: true,
+                enable_node_native_modules: true,
```
sokra, line 115: Node externals and node native modules are not supported in middleware

crates/next-core/src/next_server/context.rs
```diff
@@ -107,6 +108,26 @@ pub async fn get_server_resolve_options_context(
...
+                module: true,
+                custom_conditions: vec!["development".to_string()],
+                import_map: Some(next_server_import_map),
```
sokra, line 118: This import map probably defines some externals which we don't want here.

There must be no externals in an edge compilation.

crates/next-core/src/router.rs
```diff
@@ -114,55 +140,155 @@ impl From<RouterIncomingMessage> for RouterResult {
...
+    page_extensions: StringsVc,
+) -> Result<InnerAssetsVc> {
+    let mut inner = HashMap::new();
+
+    let middleware_config = get_config(context, project_path, middleware_files(page_extensions));
+    inner.insert(
+        "MIDDLEWARE_CHUNK_GROUP".to_string(),
+        context.with_transition("next-edge").process(
+            middleware_config.into(),
+            Value::new(TypeScript(TypeScriptReferenceSubType::Undefined)),
```
sokra, line 196: It doesn't really matter currently, but it should be EcmaScript import

sokra (approved):

crates/next-core/js/src/entry/router.ts
```diff
@@ -65,7 +68,16 @@ async function getResolveRoute(
     true
   );
 
-  return await makeResolver(dir, nextConfig);
+  const edgeInfo = {
+    name: "edge",
+    paths: middlewareChunkGroup.map((chunk: string) =>
```
ijjk, line 73: Shouldn't this be called `files` and be relative to the `distDir` to match our middleware plugin 
x-ref: https://github.com/vercel/next.js/blob/7d8f85bd8dbd767dfac0d92eed651877268e62cb/packages/next/src/build/webpack/plugins/middleware-plugin.ts#L35

jridgewell: I copied this from https://github.com/vercel/turbo/blob/805a5ae14f1ab67f3c3e01e758f7b53d44f9d696/crates/next-core/js/src/entry/server-edge-api.tsx#L53-L59, but I'm not sure if it's correct. @sokra?

sokra: I aligned with the return value of `getEdgeFunctionInfo`, which converts the format in the manifest file into one with `paths`

https://github.com/vercel/next.js/blob/7d8f85bd8dbd767dfac0d92eed651877268e62cb/packages/next/src/server/next-server.ts#L1640-L1654

So when this info object is provided it could skip calling `getEdgeFunctionInfo`.

ijjk: I see, yeah aligning on that return type seems alright

jridgewell:
Removing the test case while we wait on Next to land the middleware changes. Once we update to the next release, I'll readd the test.

PR #3691, opened 2023-02-08
Update SWC

Brooooooklyn:
- Ref: https://github.com/swc-project/swc/pull/6910


kdy1:
Note: https://github.com/swc-project/swc/pull/6916 is also important, because `stacker` breaks `armv7` build

kdy1 (approved):
Thanks!

PR #3693, opened 2023-02-08
add ==, !=, ===, !== operator support for evaluation

sokra:
extracted from https://github.com/vercel/turbo/pull/3670

depends on #3685 

adds the equal operators to the evaluator


crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -84,11 +84,74 @@ impl PartialEq for ConstantNumber {
 
 impl Eq for ConstantNumber {}
 
+#[derive(Debug, Clone)]
+
+pub enum ConstantString {
```
alexkirsz, line 89:
```suggestion
#[derive(Debug, Clone)]
pub enum ConstantString {
```

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -145,10 +208,26 @@ impl Default for ConstantValue {
     }
 }
 
+impl From<bool> for ConstantValue {
+    fn from(v: bool) -> Self {
+        if v {
+            ConstantValue::True
+        } else {
+            ConstantValue::False
+        }
```
alexkirsz, line 217:
```suggestion
        match v {
            true => ConstantValue::True,
            false => ConstantValue::False,
        }
```

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -169,8 +248,7 @@ impl Display for ConstantValue {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             ConstantValue::Undefined => write!(f, "undefined"),
-            ConstantValue::StrWord(str) => write!(f, "\"{str}\""),
-            ConstantValue::StrAtom(str) => write!(f, "\"{str}\""),
+            ConstantValue::Str(str) => write!(f, "\"{str}\""),
```
alexkirsz, line 251: Should this use `stringify_js` instead? Although I don't expect we'll ever want to attempt to evaluate the result of this.

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -211,6 +289,25 @@ impl LogicalOperator {
...
+}
+
+impl BinaryOperator {
+    fn joiner(&self) -> &'static str {
+        match self {
+            BinaryOperator::Equal => " == ",
+            BinaryOperator::NotEqual => " != ",
+            BinaryOperator::StrictEqual => " === ",
+            BinaryOperator::StrictNotEqual => " !== ",
+        }
```
alexkirsz, line 307: Looks like there's already a space in:

```rust
            JsValue::Binary(_, a, op, b) => write!(f, "({} {} {})", a, op.joiner(), b),
```

so these should not have leading and trailing spaces.

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -1336,9 +1491,59 @@ impl JsValue {
...
+            JsValue::Binary(_, box a, op, box b) => {
+                let negate = matches!(
+                    op,
+                    BinaryOperator::NotEqual | BinaryOperator::StrictNotEqual
+                );
+                let positive_op = match op {
+                    BinaryOperator::NotEqual => BinaryOperator::Equal,
+                    BinaryOperator::StrictNotEqual => BinaryOperator::StrictEqual,
+                    _ => *op,
+                };
```
alexkirsz, line 1506: This could be a struct method, which returns `(bool, PositiveBinaryOperator)`. That way we don't keep `BinaryOperator::*Not*` in the mix in the matches below, which we know aren't possible at this point.

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -1336,9 +1491,59 @@ impl JsValue {
...
+                        BinaryOperator::Equal,
+                        JsValue::Constant(ConstantValue::Str(a)),
+                        JsValue::Constant(ConstantValue::Str(b)),
+                    ) => Some((a == b) ^ negate),
+                    (
+                        BinaryOperator::Equal,
+                        JsValue::Constant(ConstantValue::Num(a)),
+                        JsValue::Constant(ConstantValue::Num(b)),
+                    ) => Some((a == b) ^ negate),
+                    _ => None,
```
alexkirsz, line 1544: The `^ negate` could happen in a `.map` here instead.

alexkirsz (changes requested):

sokra:
performance will be fixed by follow-up PR: https://github.com/vercel/turbo/pull/3670

alexkirsz (approved):

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -1336,9 +1505,54 @@ impl JsValue {
...
+                    (
+                        PositiveBinaryOperator::StrictEqual,
+                        JsValue::Constant(a),
+                        JsValue::Constant(b),
+                    ) if a.is_value_type() => Some(a == b),
+                    (
+                        PositiveBinaryOperator::StrictEqual,
+                        JsValue::Constant(a),
+                        JsValue::Constant(b),
+                    ) if a.is_value_type() => {
```
jridgewell, line 1523: Can this second branch ever execute? It's the same match as the first.

sokra: yeah, the second branch can be removed. The first branch is an less verbose version of the first one

PR #3695, opened 2023-02-08
Add req.cookies to API requests

alexkirsz:
This PR does a little bit more than what it says on the box. Before, we were parsing URL queries and request headers, only to serialize them some other way afterwards. The reason why we were parsing headers and queries is because we sometimes need to filter them. However, we don't need that for router/page SSR/page API. This PR makes it so we avoid parsing on Turbopack's side through `rawHeaders` and `rawQuery`.

### DIFF
--- a/crates/next-core/js/src/internal/utils.ts
+++ b/crates/next-core/js/src/internal/utils.ts
@@ -1,0 +1,21 @@
+/**
+ * Converts an array of raw header entries to a map of header names to values.
+ */
+export function headersFromEntries(
+  entries: Array<[string, string]>
+): Record<string, string | string[]> {
+  const headers: Record<string, string | string[]> = Object.create(null);
+  for (const [key, value] of entries) {
+    if (key in headers) {
+      const prevValue = headers[key];
+      if (typeof prevValue === "string") {
+        headers[key] = [prevValue, value];
+      } else {
+        prevValue.push(value);
+      }
+    } else {
+      headers[key] = value;
+    }
+  }
+  return headers;
+}
--- a/crates/next-core/js/src/ipc/server.ts
+++ b/crates/next-core/js/src/ipc/server.ts
@@ -1,7 +1,7 @@
 import type { ClientRequest, IncomingMessage, Server } from "node:http";
 import type { AddressInfo } from "node:net";
 import http, { ServerResponse } from "node:http";
-import qs from "node:querystring";
+import { headersFromEntries } from "@vercel/turbopack-next/internal/utils";
 
 /**
  * Creates a server that listens a random port.
@@ -23,8 +23,8 @@ export function makeRequest(
   server: Server,
   method: string,
   path: string,
-  query?: Record<string, string>,
-  headers?: Record<string, string>
+  rawQuery?: string,
+  rawHeaders?: Array<[string, string]>
 ): Promise<{
   clientRequest: ClientRequest;
   clientResponsePromise: Promise<IncomingMessage>;
@@ -106,11 +106,10 @@ export function makeRequest(
       port: address.port,
       method,
       path:
-        query && Object.keys(query).length > 0
-          ? `${path}?${qs.stringify(query)}`
-          : path,
-      headers,
+        rawQuery != null && rawQuery.length > 0 ? `${path}?${rawQuery}` : path,
+      headers: rawHeaders != null ? headersFromEntries(rawHeaders) : undefined,
     });
+
     // Otherwise Node.js waits for the first chunk of data to be written before sending the request.
     clientRequest.flushHeaders();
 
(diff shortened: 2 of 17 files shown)


alexkirsz:
* main
  * **PR #3695** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3695" target="_blank">Graphite</a>  👈
    * **PR #3696** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3696" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3695?utm_source=stack-comment).

crates/turbopack-dev-server/src/source/resolve.rs
```diff
@@ -113,6 +113,18 @@ async fn request_to_data(
     if vary.body {
         data.body = Some(request.body.clone().into());
     }
+    if vary.raw_query {
+        data.raw_query = Some(request.uri.query().unwrap_or("").to_string());
```
jridgewell, line 117: 🎉. Would you update https://github.com/vercel/turbo/blob/780ff56732564617bbc8da300c335365bf404427/crates/turbopack-node/src/source_map/content_source.rs#L53-L58?

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -213,9 +213,14 @@ pub struct ContentSourceData {
     pub url: Option<String>,
     /// query string items, if requested
     pub query: Option<Query>,
+    /// raw query string, if requested
```
jridgewell, line 216:
```suggestion
    /// raw query string, if requested. Does not include the `?`
```

crates/turbopack-dev-server/src/source/resolve.rs
```diff
@@ -113,6 +113,18 @@ async fn request_to_data(
...
+    }
+    if vary.raw_headers {
+        data.raw_headers = Some(
+            request
+                .headers
+                .iter()
+                .map(|(name, value)| Ok((name.to_string(), value.to_str()?.to_string())))
+                .collect::<Result<Vec<_>>>()?,
```
jridgewell, line 125: Does this automatically transpose the individual `Result`s into a single `Result`?

crates/next-core/js/src/ipc/server.ts
```diff
@@ -106,11 +106,10 @@ export function makeRequest(
       port: address.port,
       method,
       path:
-        query && Object.keys(query).length > 0
-          ? `${path}?${qs.stringify(query)}`
-          : path,
-      headers,
+        rawQuery != null && rawQuery.length > 0 ? `${path}?${rawQuery}` : path,
```
jridgewell, line 109:
```suggestion
        rawQuery ? `${path}?${rawQuery}` : path,
```

crates/next-core/js/src/ipc/server.ts
```diff
@@ -106,11 +106,10 @@ export function makeRequest(
...
+      headers: rawHeaders != null ? headersFromEntries(rawHeaders) : undefined,
```
jridgewell, line 110: I've always wished we had an inverse `??` for chaining: `headers: rawHeaders !? headersFromEntries(rawHeaders)`.

crates/next-core/js/src/internal/utils.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Converts an array of raw header entries to a map of header names to values.
+ */
+export function headersFromEntries(
```
jridgewell, line 4: Is there a reason this couldn't have been done on the Rust side? I've never like the bytes/string code in our current `Headers` impl. If possible, I'd love to remove that, and just have a single headers instead of `headers` and `raw_headers`.

crates/turbopack-dev-server/src/source/resolve.rs
alexkirsz: Yep! `.collect` is magic. The first error in the iterator will be reported.

crates/next-core/js/src/ipc/server.ts
alexkirsz: My personal preference is to avoid boolean coercions in JS altogether (bringing it closer to Rust). Otherwise, it's unclear whether I meant to include the empty string in my truthy check or not.

crates/next-core/js/src/internal/utils.ts
alexkirsz: I think this could be moved to Rust without any issues. I'll look into it.

We still need `headers` and `raw_headers` because our `NeedData` implementation is able to filter on header keys and values in order to cache some results.

We do however have quite a few header types across the codebase, so it would be helpful to standardize on only two or three.

jridgewell (approved):

PR #3696, opened 2023-02-08
Add support for ctx.params in getStaticProps/getServerSideProps

alexkirsz:
Dynamic segments should be passed a single string, while catch-all segments should be passed a list of strings.

Furthermore, ctx.params was previously undefined because we weren't passing it forward through the render options.


cli/internal/ffi/proto/messages.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.1
-// 	protoc        v3.21.12
+// 	protoc        v3.21.7
```
mehulkar, line 4: I don't think we're supposed to commit this. 
cc @arlyon

alexkirsz: My bad, this sneaked it. It keeps popping up again, I think this diff might be missing on main?

jridgewell (approved):

PR #3699, opened 2023-02-08
fix: update verbosity integration tests

chris-olszewski:
Fixes breakage introduced in #3661

I'll open another PR that expands when these tests get run to prevent breakages from landing.


mehulkar (approved):
Looks like I had added the turborepo crates to trigger e2e tests, but not integration tests in https://github.com/vercel/turbo/pull/3442/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R91-R98.

PR #3701, opened 2023-02-08
match next.js client side routing (buildManifest rewrites & devPagesManifest HMR)

ForsakenHarmony:


ForsakenHarmony:
* main
  * **PR #3701** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3701" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3701?utm_source=stack-comment).

crates/next-core/js/src/entry/config/next.js
```diff
@@ -13,6 +14,12 @@ const loadNextConfig = async (silent) => {
 
   nextConfig.generateBuildId = await nextConfig.generateBuildId?.();
 
+  const customRoutes = await loadCustomRoutes(nextConfig);
```
ijjk, line 17: We should be creating this on the Next.js side so that we don't have to call `loadCustomRoutes` twice x-ref: https://github.com/vercel/next.js/blob/7654d7bc2d95c8b3dc18cf5e94695375c733e74c/packages/next/src/server/lib/route-resolver.ts#L28

ForsakenHarmony: We load the whole config a second time to call the route resolver right now unfortunately

ForsakenHarmony: We load the whole config a second time to call the route resolver right now unfortunately

crates/next-core/src/next_config.rs
```diff
@@ -161,6 +158,100 @@ enum OutputType {
...
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, TraceRawVcs)]
+#[serde(rename_all = "camelCase")]
+pub enum RedirectStatus {
+    StatusCode(f64),
+    Permanent(bool),
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, TraceRawVcs)]
+#[serde(rename_all = "camelCase")]
+pub struct Redirect {
```
jridgewell, line 215: Do we need to implement all of these if we only care about `Rewrite`?

crates/next-core/src/manifest.rs
```diff
@@ -75,25 +150,30 @@ impl ContentSource for DevManifestContentSource {
...
+                File::from(serde_json::to_string(&serde_json::json!({
                     "pages": pages,
-                }))?
+                }))?)
+                .with_content_type(APPLICATION_JSON)
+            }
+            "_next/static/development/_buildManifest.js" => {
+                let build_manifest = &*self_vc.create_build_manifest().await?;
+
+                File::from(build_manifest.as_str()).with_content_type(APPLICATION_JAVASCRIPT)
```
jridgewell, line 165: Nit:
```suggestion
                File::from(build_manifest.as_str()).with_content_type(APPLICATION_JSON_UTF_8)
```

jridgewell (approved):

crates/next-core/js/src/entry/config/next.js
ForsakenHarmony: @ijjk is there any chance we could get something implemented on the next.js side that would allow the whole config to be serialized to JSON (after loading custom routes probably)?

crates/next-core/src/manifest.rs
ForsakenHarmony: Was that supposed to be `APPLICATION_JAVASCRIPT_UTF_8`?

crates/next-core/src/next_config.rs
ForsakenHarmony: Probably not 🤷‍♀️
But maybe we can pass the serialized version to the next.js router

crates/next-core/src/manifest.rs
jridgewell: Yes 🤦

jridgewell (approved):

crates/next-core/src/manifest.rs
ForsakenHarmony: done

crates/next-core/src/manifest.rs
```diff
@@ -177,3 +182,26 @@ impl ContentSource for DevManifestContentSource {
...
+/// alphabetically this means \[slug] and \[\[catchall]] routes are prioritized
+/// over fixed paths, so we have to override the ordering with this.
+#[derive(Ord, PartialOrd, Eq, PartialEq)]
+enum PageSortKey {
+    Static(String),
+    Slug,
+    CatchAll,
+}
+
+impl From<&str> for PageSortKey {
```
jridgewell, line 193: I think we can bound the lifetime of `PageSortKey` by the `&str`, so we don't need to clone into a `String`?
```suggestion
impl<'a> From<&'a str> for PageSortKey<'a> {
```

jridgewell (approved):

crates/next-core/src/manifest.rs
ForsakenHarmony: I was assuming the `&str` wouldn't survive after returning out of the `sort_by_cached_key` lambda function, but let me check

ForsakenHarmony: yeah

```rs
error: lifetime may not live long enough
  --> crates\next-core\src\manifest.rs:77:39
   |
77 |         routes.sort_by_cached_key(|s| s.split('/').map(PageSortKey::from).collect::<Vec<_>>());
   |                                    -- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'2`
   |                                    ||
   |                                    |return type of closure is Vec<PageSortKey<'2>>
   |                                    has type `&'1 std::string::String`
```

PR #3707, opened 2023-02-08
Issue Reporters

wbinnssmith:
Big thanks to @jridgewell for helping me out with a number of Rust-isms with this change.

This expands `handle_issues` and the `NextDevServerBuilder` to accept an arbitrary `IssueReporter` -- a trait implementing `report_issues` which receives captured issues to send somewhere.

This replaces using a fixed `ConsoleUi` to send issues to stdout/stderr, though `ConsoleUi` now implements `IssueReporter` and is the default implementation of an issue reporter if no other is provided. It also moves the responsibility of detecting fatal errors out of `ConsoleUi` and into `handle_issues` itself.

This lays the foundation for alternative reporters, such as a test reporter to snapshot or assert against issues emitted, or a newline-delimited JSON reporter for other tools to consume.

Co-authored-by: Justin Ridgewell <justin@ridgewell.name>


crates/next-dev/src/lib.rs
```diff
@@ -199,51 +211,70 @@ impl NextDevServerBuilder {
...
-        .await?;
+async fn handle_issues<T: Into<RawVc> + CollectiblesSource + Copy>(
+    source: T,
+    issue_reporter: IssueReporterVc,
+) -> Result<()> {
+    let issues = IssueVc::peek_issues_with_path(source).await?;
+    issue_reporter.report_issues(issues, TransientValue::new(source.into()));
+
+    let mut has_fatal = false;
+    for issue in issues.strongly_consistent().await?.iter() {
```
jridgewell, line 251: The `strongly_consistent` needs to come before the call to `report_issues`, that way we get the same issues for both the reporting and the fatal iteration. We'll need to update `TaskInput` to receive a `ReadRef`, @alexkirsz might be able to help with that.

crates/next-dev/src/lib.rs
```diff
@@ -199,51 +211,70 @@ impl NextDevServerBuilder {
...
         let tasks = turbo_tasks.clone();
+        let issue_reporter_once = Arc::new(Lazy::new(|| {
+            self.issue_reporter
+                .unwrap_or_else(|| {
+                    // Initialize a ConsoleUi reporter if no custom reporter was provided
+                    Box::new(|log_options| ConsoleUi::new(log_options).cell().into())
+                })
+                .get_issue_reporter(log_options)
+        }));
+        let issue_reporter_arc = Arc::new(move || **issue_reporter_once);
```
sokra, line 224: The Lazy is not needed as long `get_issue_reporter` is a turbo-tasks function. As it returns a Vc we should be able to assume that it's one.

Caching that with a Lazy will cause some parents to not be connected and this might cause weird effects.

You can remove the Lazy and rely on turbo-tasks functions.

You probably want to change:

``` diff
-ConsoleUi::new(log_options).cell()
+ConsoleUiVc::new(log_options)
```

So this is cached too and not creates cell over and over.

sokra (changes requested):

crates/next-dev/src/lib.rs
```diff
@@ -192,58 +203,69 @@ impl NextDevServerBuilder {
...
+) -> Result<()> {
+    let issues_vc = IssueVc::peek_issues_with_path(source).await?;
+    let issues = issues_vc.strongly_consistent().await?;
+    issue_reporter.report_issues(
+        TransientInstance::new(issues),
+        TransientValue::new(source.into()),
+    );
 
-    if state.has_fatal {
+    if *issues_vc.has_fatal().await? {
```
sokra, line 250: You probably want to read this strongly_consistent since you are reading issues strongly_consistent.

But there is still no guarantee that issues and fatal are consistent together.

---

So it might be better to use `issues.has_fatal()` instead and about the additional function.

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -346,6 +347,20 @@ impl CapturedIssuesVc {
     pub async fn is_empty(self) -> Result<BoolVc> {
         Ok(BoolVc::cell(self.await?.is_empty()))
     }
+
+    #[turbo_tasks::function]
+    pub async fn has_fatal(self) -> Result<BoolVc> {
+        let mut has_fatal = false;
+
+        for issue in self.strongly_consistent().await?.iter() {
```
sokra, line 355: `strongly_consistent` can only be used on operations. `self` is already resolved to a cell.

Only use strongly_consistent after calling a function.

The correct way to make this strongly consistent would be `issues_vc.has_fatal().strongly_consistent().await?`

crates/turbopack-dev-server/src/lib.rs
```diff
@@ -50,6 +51,19 @@ where
...
+
+impl<T> IssueReporterProvider for T
+where
+    T: Fn() -> IssueReporterVc + Send + Sync + Clone + 'static,
+{
+    fn get_issue_reporter(&self) -> IssueReporterVc {
+        self()
+    }
+}
+
```
sokra, line 66: This is only used in next-dev? Maybe it should be there.

sokra (approved):

crates/node-file-trace/src/lib.rs
```diff
@@ -487,24 +487,27 @@ async fn run<B: Backend + 'static, F: Future<Output = ()>>(
...
     let dir = current_dir().unwrap();
     let tt = create_tt();
-    let console_ui = Arc::new(ConsoleUi::new(LogOptions {
-        current_dir: dir.clone(),
-        show_all,
-        log_detail,
-        log_level: log_level.map_or_else(|| IssueSeverity::Error, |l| l.0),
-    }));
     let task = tt.spawn_root_task(move || {
+        let console_ui = ConsoleUiVc::new(TransientInstance::new(LogOptions {
```
jridgewell, line 491: This is the bug. The `TransientInstance` needs to be extracted out and shared by all calls, that way the same `ConsoleUiVc` will be reused.

crates/node-file-trace/src/lib.rs
```diff
@@ -487,24 +487,27 @@ async fn run<B: Backend + 'static, F: Future<Output = ()>>(
...
         Box::pin(async move {
             let output = main_operation(TransientValue::new(dir.clone()), args.clone().into());
 
-            let console_ui = (*console_ui).clone().cell();
-            console_ui
-                .group_and_display_issues(TransientValue::new(output.into()))
+            let source = TransientValue::new(output.into());
+            let issues = IssueVc::peek_issues_with_path(output)
+                .await?
+                .strongly_consistent()
```
jridgewell, line 506: Actually, it might be this.

PR #3719, opened 2023-02-09
chore(releases): add release config

tknickman:
Generate release notes automatically.


arlyon (approved):
Looks great! We can adjust the specifics later if needed.

PR #3720, opened 2023-02-09
Allow opt-in release dry run for turborepo

arlyon:
This modifies the release pipeline to add an opt-in dry run step that simply ignores the final stage of the publish pipeline. It also updates the release to handle the new tags added for the rust sandwich.


tknickman (approved):
Looks good!

.github/workflows/turborepo-release-step-1.yml
```diff
@@ -1,3 +1,20 @@
+# Turborepo Release Pipeline
+#
+# This release consists of 3 steps. The first, is this one.
+#
+# 1. Kick off the release flow with a manual workflow_dispatch
+#    which increments the version and creates a new branch,
+#    with a commit incrementing version.go as well as version.txt.
+#    For more detail, see stage-release in cli/Makefile
```
mehulkar, line 8:
```suggestion
#    For more detail, see `stage-release` in cli/Makefile
```

.github/workflows/turborepo-release-step-1.yml
```diff
@@ -1,3 +1,20 @@
...
+# 2. Stage two takes the newly-created staging branch and compiles
+#    a production go build, the artifacts for which are stored
```
mehulkar, line 10:
```suggestion
#    a production Go build, the artifacts for which are stored
```

"production go build" could be confused as "the go live build" (rather than Golang). Other than that, I'm not sure if `go` is supposed to be capitalized or not

.github/workflows/turborepo-release-step-3.yml
```diff
@@ -5,6 +5,9 @@ on:
     inputs:
       release_branch:
         description: "Staging branch to run release from"
+      dry_run:
+        description: "Do a dry run, skipping the final publish step."
+        type: boolean
```
mehulkar, line 10: Should we default this to true for more safety?

mehulkar (approved):
minor suggestions, but looks good. thanks for the comments, i was wondering how the release process works!

.github/workflows/turborepo-release-step-3.yml
arlyon: Good question, thoughts @tknickman?

arlyon: I think we should do this separately. I am going to merge now :)

PR #3722, opened 2023-02-09
fix: Turbo updater feature

NicholasLYang:
Gate turbo-updater behind a feature because it uses `ureq` which in turn uses `rustls-tls` which uses `ring` which [doesn't build on Windows ARM](https://github.com/briansmith/ring/issues/1514) (at least until 0.17 is released).


NicholasLYang:
Closed in favor of #3726

PR #3723, opened 2023-02-09
Accept `next/font/google` as well as `@next/font/google`

wbinnssmith:
Aligning with https://github.com/vercel/next.js/pull/45707, this allows users to import fonts from `next/font/google` as well as `@next/font/google`.

Test Plan: Adjusted tests to use `next/font/google` and added another test to verify `@next/font/google` continues to work.


crates/next-core/src/next_font_google/mod.rs
```diff
@@ -82,14 +82,19 @@ impl ImportMappingReplacement for NextFontGoogleReplacer {
...
                                 style: {{
                                     fontFamily: "{}",
                                     {}{}
                                 }},
-                                variable: cssModule.variable
                             }};
+
+                            if (cssModule.variable != null) {{
+                                fontData.variable = cssModule.variable;
+                            }}
```
jridgewell, line 95: Why is this change included?

jridgewell (approved):

hanneslund (approved):

crates/next-core/src/next_font_google/mod.rs
wbinnssmith: It's unrelated, but a small issue issue I fixed while adding the new test. Could have been a new PR I suppose.

PR #3726, opened 2023-02-09
fix: use prerelease update-informer to cut ureq

chris-olszewski:
`update-informer` has some good stuff in the pipeline that allows us to specify the http client we use. This is a quick and dirty add support for the updater on Windows ARM. If we don't want to use a non-release version of `update-informer` than #3722 should be merged instead of this.

I expect a new release of `update-informer` might take a little bit as this is a breaking change due to the updated `Registry` trait definition. 
Screen Shot 2023-02-09 at 11 48 37 AM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7635,7 +7635,6 @@ dependencies = [
  "serde",
  "thiserror",
  "update-informer",
- "ureq",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


NicholasLYang:
So this uses `reqwest`, but does it use the `rustls-tls` or `native-tls` feature for reqwest? When I build this I don't see either `ring` or `openssl` getting built.

chris-olszewski:
> So this uses reqwest, but does it use the rustls-tls or native-tls feature for reqwest? When I build this I don't see either ring or openssl getting built.

I've now piped the `rustls-tls` and `native-tls` features into this crate to mirror the `turborepo-lib` behavior. Is this what you're talking about?

arlyon:
I would maybe prefer this approach but I tend to have a more lax opinion on prerelease stuff.

NicholasLYang:
Yeah that looks good to me. I'll test on my Windows ARM setup and see how it goes

NicholasLYang (approved):
Getting some funky Go errors but the Rust compilation works with this. Nice!

PR #3727, opened 2023-02-09
ci(workflow): set diff_base for release publish event

kwonoj:
Trying to fix slack posting issue.

### DIFF
--- a/.github/workflows/nextjs-integration-test.yml
+++ b/.github/workflows/nextjs-integration-test.yml
@@ -12,6 +12,10 @@ on:
       # If not specified, will use latest release version including canary.
       version:
         type: string
+      # The base of the test results to compare against. If not specified, will try to compare with latest main branch's test results.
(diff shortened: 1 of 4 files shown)


.github/workflows/nextjs-integration-test.yml
```diff
@@ -12,6 +12,10 @@ on:
       # If not specified, will use latest release version including canary.
       version:
         type: string
+      # The base of the test results to compare against. If not specified, will try to compare with latest main branch's test results.
+      diff_base:
+        typg: string
```
wbinnssmith, line 17:
```suggestion
        type: string
```

wbinnssmith (approved):

.github/workflows/nextjs-integration-test.yml
kwonoj: 🤦

PR #3729, opened 2023-02-09
Dedupe TurboJSON and PackageJSON reads

mehulkar:
Cache reads of turbo.json and package.json in the `WorkspaceInfos` struct in CompleteGraph. PackageJSON were already cached there, so it's a natural place to expand the notion of known Workspaces.


gsoltis (approved):
Thanks for adding this!

PR #3730, opened 2023-02-10
Allow overriding persistent flag from workspace

mehulkar:
Will do `cache` key next


cli/internal/core/engine.go
```diff
@@ -118,6 +118,12 @@ func (e *Engine) getTaskDefinition(pkg string, taskName string, taskID string) (
 		}, nil
 	}
 
+	// An error here means turbo.json exists, but didn't define the task.
+	// Fallback to the root pipeline to find the task.
+	if pkg != util.RootPkgName {
```
mehulkar, line 123: I guess this condition wasn't exercised by any of the other tests yet... something about #3729 caused this change, because the refactor returned empty structs in some cases rather than throwing errors.

PR #3734, opened 2023-02-10
Implement custom nightly tag bumping

jridgewell:
This accomplishes a few things:
- Aligns with our date-based versioning, instead of semver
- Removes our reliance on `github-tag-action`, which was designed for semver, and didn't like our old tags
- Allows us to manually rerun the release process, bumping the patch of the current day automatically

Fixes WEB-502


.github/actions/turbopack-bump/action.yml
```diff
@@ -0,0 +1,5 @@
+name: "turbopack-bump"
```
sokra, line 1: Should the action.yml list the `inputs` and `outputs`

sokra (approved):

.github/actions/turbopack-bump/action.yml
jridgewell: Good catch, done.

PR #3736, opened 2023-02-10
improve node_modules as externals handling

sokra:
Run after_resolve with original request

try to resolve original request with node.js options to determine if it can be external

typescript extends need to resolve with bare node.js resolving to avoid custom resolve options

add special pnpm logic to improve externals when pnpm is used


ForsakenHarmony (approved):

crates/next-core/src/next_server/resolve.rs
```diff
@@ -47,8 +90,12 @@ impl ResolvePlugin for ExternalCjsModulesResolvePlugin {
...
         }
 
-        // mjs -> esm module
-        if Some("mjs") == raw_fs_path.extension() {
+        // node.js only supports these file extensions
+        // mjs is an esm module and we can't bundle that yet
+        if !matches!(
+            raw_fs_path.extension(),
+            Some("cjs" | "js" | "node" | "json")
+        ) {
```
ForsakenHarmony, line 98: Could we get rid of this and just use `is_node_resolveable`?

crates/next-core/src/next_server/resolve.rs
```diff
@@ -68,31 +115,39 @@ impl ResolvePlugin for ExternalCjsModulesResolvePlugin {
...
-        Ok(ResolveResultOptionVc::some(
-            ResolveResult::primary(PrimaryResolveResult::OriginalReferenceExternal).cell(),
-        ))
+        // Special behavior for pnpm as we could reference all .pnpm modules by
+        // referencing the `.pnpm` folder as module, e. g.
+        // /node_modules/.pnpm/some-package@2.29.2/node_modules/some-package/dir/file.js
+        // becomes
+        // .pnpm/some-package@2.29.2/node_modules/some-package/dir/file.js
+        if let Some(captures) = PNPM.captures(&fs_path.await?.path) {
+            if let Some(r#match) = captures.get(1) {
```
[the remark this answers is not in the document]
sokra: 🤣

crates/next-core/src/next_server/resolve.rs
sokra: Seems like no:

* `package/style.css` is node.js resolveable, but node.js will still struggle with it.
* It's a very cheap check. `is_node_resolveable` is very expensive since it need to resolve the whole request.
* We also want to avoid `.mjs` until we support ESM.

PR #3737, opened 2023-02-10
fix integration tests on windows

sokra:
For windows it's important that the browser is dropped so that the test can
complete. To do that we need to cancel the spawned task below (which will
drop the browser). For this we are using a JoinSet which cancels all tasks
when dropped.


Brooooooklyn (approved):

crates/turbopack-node/src/pool.rs
```diff
@@ -39,6 +39,14 @@ struct SpawnedNodeJsPoolProcess {
 struct RunningNodeJsPoolProcess {
     child: Option<Child>,
     connection: TcpStream,
+    stdout_buf: Vec<u8>,
```
ForsakenHarmony, line 42: this just keeps getting worse

PR #3738, opened 2023-02-10
share asset and chunking context between server, server data and client navigation

sokra:
This improves the performance for client-side navigation as it can reuse the modules from the initial navigation. Same for server data requests.

### DIFF
--- a/crates/next-core/src/page_source.rs
+++ b/crates/next-core/src/page_source.rs
@@ -95,7 +95,9 @@ pub async fn create_page_source(
         src_pages
     } else {
         return Ok(NoContentSourceVc::new().into());
-    };
+    }
+    .resolve()
+    .await?;
 
     let client_ty = Value::new(ClientContextType::Pages { pages_dir });
     let server_ty = Value::new(ServerContextType::Pages { pages_dir });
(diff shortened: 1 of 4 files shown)


crates/next-core/src/page_source.rs
```diff
@@ -192,23 +193,22 @@ pub async fn create_page_source(
         .collect(),
     );
 
+    let client_context: AssetContextVc = ModuleAssetContextVc::new(
+        transitions,
```
sokra, line 197: All AssetContext get passed the same transitions. So `context_context` is equal to the client context created by the `server_context` + client transition

crates/next-core/src/page_source.rs
```diff
@@ -97,7 +95,9 @@ pub async fn create_page_source(
         src_pages
     } else {
         return Ok(NoContentSourceVc::new().into());
-    };
+    }
+    .resolve()
+    .await?;
 
     let client_ty = Value::new(ClientContextType::Pages { pages_dir });
```
sokra, line 102: the `pages_dir` in `client_ty` was not matching the `pages_dir` in the client_ty created within the client transition, as it wasn't resolved (still pointing to the `join` operation).

Note in future we should prevent unresolved Vcs to be stored in structs without explict opt-in of that.

alexkirsz (approved):

crates/next-core/src/page_source.rs
alexkirsz: This is definitely an easy to make mistake.

PR #3741, opened 2023-02-10
Fix the router when Next.js returns no result

alexkirsz:
I changed the return type of the Next.js router in https://github.com/vercel/next.js/pull/45668 to account for cases where the router does and does not find a matching rewrite.

### DIFF
--- a/crates/next-core/src/router_source.rs
+++ b/crates/next-core/src/router_source.rs
@@ -90,6 +90,9 @@ impl ContentSource for NextRouterContentSource {
                 this.inner
                     .get(path, Value::new(ContentSourceData::default()))
             }
+            RouterResult::None => this
+                .inner
+                .get(path, Value::new(ContentSourceData::default())),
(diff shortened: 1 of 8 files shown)


crates/next-core/src/router_source.rs
```diff
@@ -91,6 +91,9 @@ impl ContentSource for NextRouterContentSource {
                 this.inner
                     .get(path, Value::new(ContentSourceData::default()))
             }
+            RouterResult::None => this
+                .inner
+                .get(path, Value::new(ContentSourceData::default())),
```
jridgewell, line 96: Shouldn't this be returning a 404? Possibly with an emitted issue saying you might have forgotten basepath?

alexkirsz: Since the Next.js router will also return "none" for page/app sources' assets, we still need to check whether the inner source has some content for this path.

Issue #3280, opened 2023-01-12
[turborepo] `1.7.0` with yarn berry doesn't work


kyun:
### What version of Turborepo are you using?

1.7.0

### What package manager are you using / does the bug impact?

Yarn v2/v3 (node_modules linker only)

### What operating system are you using?

Mac

### Describe the Bug

I've upgraded turbo version `1.6.3` -> `1.7.0`. 

I've got error like this
```
thread 'main' panicked at 'Failed to execute turbo.: Os { code: 2, kind: NotFound, message: "No such file or directory" }', crates/turborepo/src/main.rs:23:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Expected Behavior

It should be work well like `1.6.3`

### To Reproduce
**.yarnrc.yml**
```
nodeLinker: pnp
```

```
yarn set version berry
yarn add turbo@1.7.0
yarn dev
```

tknickman:
Reproduced, thanks for the issue - we're taking a look.

tknickman:
@kyun you can unblock by using `nodeLinker: node-modules`, but we're identified the issue and are looking into solutions for installing with `pnp` on our end.

alexghr:
Hey, just to add to this: I'm seeing the same thing when using `pnpm@7.24.3` and `turbo@1.7.0`. I've tried with `node-linker=hoisted` but I still get the same error.

tknickman:
@alexghr could you provide a reproduction? I'm not seeing any issues with `pnpm@7.24.3` locally so want to make sure I'm testing correctly.

alexghr:
Hmm, my weird environment might also be at fault. My machine runs NixOS so ever since Turbo switched to a pre-compiled binary I can't install it directly from `npm` (see #2556 and #2293). I'm using `TURBO_BINARY_PATH` to tell the JS wrapper where the binary is and I install the binary through a Nix flake (so that I can patch it for my system).

I've pushed a branch to this test repo here https://github.com/alexghr/turbo.nix-test/tree/turbo-1.7.0+pnpm. The files are copy-pasted from the `create-turbo-app` template, to which I've added `flake.nix` so that I could install `turbo`.

Commands like `turbo --version` and `turbo --help` do work with this setup.

tknickman:
Awesome, thanks @alexghr I'll take a look.

teamchong:
[...]

PR #3744, opened 2023-02-10
fix(turbo): improved yarn berry pnp support (#3744

tknickman:
This improves yarn berry support when using `nodeLinker: pnp` by using the correct binary path for the child process. This ensure our rust binary is can correctly find the go binary.

Fixes https://github.com/vercel/turbo/issues/3280

### DIFF
--- a/packages/turbo/node-platform.js
+++ b/packages/turbo/node-platform.js
@@ -11,14 +11,17 @@ const path = require("path");
 // external code relies on this.
 const TURBO_BINARY_PATH = process.env.TURBO_BINARY_PATH;
 
+const packageDarwin_arm64 = "turbo-darwin-arm64";
+const packageDarwin_x64 = "turbo-darwin-64";
+
 const knownWindowsPackages = {
   "win32 arm64 LE": "turbo-windows-arm64",
   "win32 x64 LE": "turbo-windows-64",
 };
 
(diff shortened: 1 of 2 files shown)


packages/turbo/node-platform.js
```diff
@@ -50,7 +50,13 @@ function generateBinPath() {
   // path without modifying the code itself. Do not remove this because
   // external code relies on this (in addition to turbo's own test suite).
   if (TURBO_BINARY_PATH) {
-    return TURBO_BINARY_PATH;
+    if (!fs.existsSync(TURBO_BINARY_PATH)) {
+      console.warn(
+        `[turbo] Ignoring bad configuration: TURBO_BINARY_PATH=${TURBO_BINARY_PATH}`
+      );
```
chris-olszewski, line 56: Just confirming, this is lifted from esbuild and not something you ran into during testing, correct?

chris-olszewski (approved):
The suggestions when we find a binary for a different platform are wonderful

packages/turbo/node-platform.js
tknickman: That's correct, but I see the value, so wanted to keep it

tknickman:
Credit where it's due, those are mostly from `esbuild` (what inspired the original install process as well)

PR #3746, opened 2023-02-10
set correct CWD for node processes

ForsakenHarmony:


ForsakenHarmony:
* main
  * **PR #3746** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3746" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3746?utm_source=stack-comment).

sokra (approved):

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1668,12 +1669,19 @@ impl ValueToString for NullFileSystem {
     }
 }
 
-pub async fn to_sys_path(path: FileSystemPathVc) -> Result<Option<PathBuf>> {
-    if let Some(fs) = DiskFileSystemVc::resolve_from(path.fs()).await? {
-        let sys_path = fs.await?.to_sys_path(path).await?;
-        return Ok(Some(sys_path));
+pub async fn to_sys_path(mut path: FileSystemPathVc) -> Result<Option<PathBuf>> {
+    loop {
+        if let Some(fs) = AttachedFileSystemVc::resolve_from(path.fs()).await? {
```
sokra, line 1674: That's a bit nasty... We need to get rid of the attached file system...

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1668,12 +1669,19 @@ impl ValueToString for NullFileSystem {
     }
 }
 
-pub async fn to_sys_path(path: FileSystemPathVc) -> Result<Option<PathBuf>> {
-    if let Some(fs) = DiskFileSystemVc::resolve_from(path.fs()).await? {
-        let sys_path = fs.await?.to_sys_path(path).await?;
-        return Ok(Some(sys_path));
+pub async fn to_sys_path(mut path: FileSystemPathVc) -> Result<Option<PathBuf>> {
+    loop {
```
jridgewell, line 1673: It looks like this can only loop 1 time. Did you mean to continue if we unwrapped an `AttachedFileSystemVc`?

jridgewell (changes requested):

crates/next-core/js/src/entry/server-edge-api.tsx
```diff
@@ -77,7 +79,7 @@ async function runEdgeFunction({
 
   const { run } = require("next/dist/server/web/sandbox");
   const result = (await run({
-    distDir: process.cwd(),
+    distDir: join(process.cwd(), ".next/server/pages"),
```
jridgewell, line 82: Can we pass this in as a param instead of duplicating the rust path?

crates/turbo-tasks-fs/src/lib.rs
ForsakenHarmony: yes, fortunately clippy also found this

jridgewell (approved):

PR #3748, opened 2023-02-10
fix: Made native-tls our default feature for ease of dev workflows

NicholasLYang:


arlyon:
I think this is reasonable. Does it make sense to include how to change the default in the doc?

NicholasLYang:
Yeah that makes sense, I'll add it

chris-olszewski (approved):

PR #3749, opened 2023-02-10
Detect invalid workspace turbo.json configurations

mehulkar:
Adds a generic way to validate. I made it generic because I anticipate that

1. we will want a different set of validations for the root config
2. In this case, the end-user error I want to print _which_ config we're talking about the `fs.TurboJSON` doesn't actually know that. Generally, the outer context will know what the validation error means and how to present it to the user.


cli/internal/core/engine.go
```diff
@@ -446,13 +446,16 @@ func (e *Engine) getTaskDefinitionChain(taskID string, taskName string) ([]fs.Bo
...
-
-			if err == nil {
-				taskDefinitions = append(taskDefinitions, *workspaceDefinition)
+			if workspaceDefinition, ok := workspaceTurboJSON.Pipeline[taskName]; ok {
+				taskDefinitions = append(taskDefinitions, workspaceDefinition)
+			} else {
+				// If the workspace doesn't have the taskName defined, but we detect that
+				// it defines the package#task syntax, throw an error. This is a misconfiguration
+				// and we actively want to prevent it.
+				// TODO(mehulkar): Should we validate this upfront when reading a worksapce turbo.jsoN?
```
gsoltis, line 455: Probably yes. Would it help to have `GetTurboConfigFromWorkspace` return `nil, nil` to signal not found to free up the error for actual errors that should be reported? Another option that is somewhat common is to check if the returned error `os.IsNotExist(err)`. IMHO that puts additional work on the caller, whereas checking that a returned value is `nil` or not is very slightly more obvious to do. Another option is that I've seen would be to change the return type to `(*TurboJSON, bool, error)` and treat the `bool` as found / not found. The plus side to this approach is that it mimics checking a map for a key.

gsoltis (changes requested):

mehulkar:
@SocketSecurity ignore esbuild@0.17.7
@SocketSecurity ignore turbo@1.0.0
@SocketSecurity ignore turbo@1.5.6
@SocketSecurity ignore turbo@1.6.3
@SocketSecurity ignore turbo@1.7.4

PR #3750, opened 2023-02-10
Validate invalid workspace turbo.json up front

mehulkar:


mehulkar:
Pushing these commits into #3749, since we want to go with this approach

PR #3767, opened 2023-02-13
chunks in chunk groups no longer reference parallel chunks

sokra:
optimized chunks no longer reference unoptimized chunks

This fixes a bug where unoptimized chunks where written to the temporary output directory


crates/turbopack-core/src/chunk/chunk_in_group.rs
```diff
@@ -0,0 +1,131 @@
...
+    introspect::{
+        asset::{children_from_asset_references, content_to_details, IntrospectableAssetVc},
+        Introspectable, IntrospectableChildrenVc, IntrospectableVc,
+    },
+    reference::AssetReferencesVc,
+    version::VersionedContentVc,
+};
+
+#[turbo_tasks::value]
+pub struct ChunkInGroup {
```
alexkirsz, line 18: We're going to need a comment on this struct to explain the difference between this and `Chunk`/`ChunkGroup`.

crates/turbopack-core/src/chunk/chunk_in_group.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub inner: ChunkVc,
```
alexkirsz, line 19: Can we have this exposed as a public method instead? Otherwise, also write a comment for this item.

alexkirsz (approved):

PR #3770, opened 2023-02-13
Make stringify_str and stringify_module_id use Serialize

alexkirsz:
Both `stringify_str` and `stringify_module_id` are now simply `stringify_js`. We could skip this and go directly to `serde_json`, but this hides the dependency behind turbopack-ecmascript.

### DIFF
--- a/crates/turbopack-core/src/chunk/mod.rs
+++ b/crates/turbopack-core/src/chunk/mod.rs
@@ -8,7 +8,7 @@ use std::{
 
 use anyhow::{anyhow, Result};
 use indexmap::IndexSet;
-use serde::{de::Visitor, Deserialize, Serialize};
+use serde::{Deserialize, Serialize};
 use turbo_tasks::{
     debug::ValueDebugFormat,
     primitives::{BoolVc, StringVc},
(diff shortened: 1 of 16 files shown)


alexkirsz:
* main
  * **PR #3770** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3770" target="_blank">Graphite</a>  👈
    * **PR #3771** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3771" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3770?utm_source=stack-comment).

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -27,9 +27,8 @@ use crate::{
 };
 
 /// A module id, which can be a number or string
-#[turbo_tasks::value(shared)]
+#[turbo_tasks::value(shared, serialization = "custom")]
 #[derive(Debug, Clone, Hash, DeterministicHash)]
-#[serde(untagged)]
```
ForsakenHarmony, line 32: Shouldn't that be the exact same behavior?

ForsakenHarmony (approved):

crates/turbopack-core/src/chunk/mod.rs
sokra: Yeah. I think the same.

PR #3771, opened 2023-02-13
Add support for CSS module composes: and fix CSS precedence issues

alexkirsz:
Adds support for the CSS module `composes:` rule.

This also fixes a large issue with our chunk ordering and CSS precedence, where the BFS order of our chunks did not match the expected topological ordering.


alexkirsz:
* main
  * **PR #3770** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3770" target="_blank">Graphite</a> 
    * **PR #3771** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3771" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3771?utm_source=stack-comment).

crates/turbopack-css/src/chunk/optimize.rs
```diff
@@ -59,49 +66,53 @@ async fn merge_chunks(first: CssChunkVc, chunks: &[CssChunkVc]) -> Result<CssChu
...
-    // When there are too many chunks, try hard to reduce the number of chunks to
-    // limit the request count.
-    if chunks.len() > TOTAL_CHUNK_MERGE_THRESHOLD {
-        let size = chunks.len().div_ceil(TOTAL_CHUNK_MERGE_THRESHOLD);
-        // TODO be smarter in selecting the chunks to merge
-        // see ecmascript implementation
-        for merged in take(&mut chunks).chunks(size) {
-            chunks.push(merge_chunks(*merged.first().unwrap(), merged).await?);
+/// The maximum number of chunks to merge into a single chunk.
+const CHUNK_MERGE_COUNT: usize = 25;
```
alexkirsz, line 70: This number is arbitrary. I found that making it larger speeds up front's `/home`.

crates/turbopack-css/src/chunk/writer.rs
```diff
@@ -36,6 +36,18 @@ pub async fn expand_imports(
                     close,
                 ));
             }
+            Some(CssImport::Composes(composed_chunk_item)) => {
+                let id = &*composed_chunk_item.to_string().await?;
+                writeln!(code, "/* composes({}) */", id)?;
```
alexkirsz, line 41: I'm not sure whether this should actually happen or not. @sokra ?

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -168,59 +168,167 @@ impl ChunkGroupVc {
...
-                                None
-                            })
-                        })
-                        .try_join()
-                        .await?;
-                }
-            }
-            Ok(result.into_iter().flatten())
-        }
+        let chunks = all_parallel_chunks_in_reverse_topological_order(self.await?.entry).await?;
```
alexkirsz, line 171: This is the part that fixes the chunk ordering issue.

Cargo.toml
```diff
@@ -92,6 +92,7 @@ opt-level = 3
 # ref: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace
 [workspace.dependencies]
 indexmap = { version = "1.9.2" }
+indoc = "1.0"
```
alexkirsz, line 95: We should do this more often.

crates/next-core/src/manifest.rs
```diff
@@ -64,7 +64,11 @@ impl DevManifestContentSourceVc {
             .page_roots
             .iter()
             .copied()
-            .try_flat_map_recursive_join(get_content_source_children)
+            // This call returns a non-deterministic order of content sources,
+            // but we don't care as we sort the result anyway.
+            .try_flat_map_recursive_join(|content_source| {
+                Some(get_content_source_children(*content_source))
```
sokra, line 70: Option<Future> is a bit weird?

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -168,59 +168,167 @@ impl ChunkGroupVc {
...
+        }
+    }
+
+    // To get a valid topological order, we would then reverse the vector.
+    // However, in this case, a reverse topological order is what we want.
+
+    rev_topo_sort
 }
 
 #[turbo_tasks::value_impl]
```
alexkirsz, line 334: <i class='graphite__hidden'>[Re: lines 537 to 537]<br/></i>

I had originally implemented a post-order DFS here instead of the reverse topological sort above, which fixed some issues but not all. It might still be relevant to also switch this to a reverse topological sort. @sokra ?<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/vercel/turbo/3771?utm_source=unchanged-line-comment">Graphite</a>.</span>

crates/turbo-tasks/src/join_iter_ext.rs
```diff
@@ -137,20 +135,29 @@ where
...
+    ///
+    /// Collects all items from the iterator and all items returns by
+    /// `filter_flat_map` into a vector.
+    ///
+    /// `filter_flat_map` will execute concurrently
+    ///
+    /// **Beware:**
+    /// * The order of the returned items is undefined.
+    /// * Circular references must be handled within `filter_flat_map`: return
+    ///   `None` to stop the recursion.
```
sokra, line 153: TBH I don't feel good about that.

Can we merge the topo sort into that method to make it deterministic?

Having a non-deterministic helper method feels like a source of mistakes...

crates/turbo-tasks/src/join_iter_ext.rs
```diff
@@ -67,24 +64,23 @@ where
...
-    C: Fn(T) -> F,
+    C: FnMut(&T) -> Option<F>,
     F: Future<Output = Result<CI>>,
     CI: IntoIterator<Item = T>,
 {
-    set: IndexSet<T>,
-    futures: FuturesOrdered<F>,
-    get_children: C,
+    output: Vec<T>,
+    futures: FuturesUnordered<F>,
```
sokra, line 72:
```suggestion
    futures: FuturesUnordered<WithParent<T, F>>,
    roots: Vec<T>,
```

```
WithParent<T, F> = impl Future<Output = Result<(T, Vec<T>)>>
```

Tuple is `(parent, children)`

So you can keep the edges directly including order...

crates/turbo-tasks/src/join_iter_ext.rs
```diff
@@ -96,17 +92,19 @@ where
...
+                                None => {}
                             }
+                            this.output.push(item);
                         }
                     }
                     Err(err) => return std::task::Poll::Ready(Err(err)),
                 }
             } else {
-                return std::task::Poll::Ready(Ok(take(&mut this.set)));
+                return std::task::Poll::Ready(Ok(std::mem::take(&mut this.output)));
```
sokra, line 107: ... and topo sort it here?

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -168,59 +168,167 @@ impl ChunkGroupVc {
...
+    let (edges, chunks_count) = get_chunks_edges_unsorted(root).await?;
+    let adjacency_map = build_adjacency_map(edges, chunks_count);
+    Ok(reverse_topological_order(root, adjacency_map, chunks_count))
+}
+
+#[derive(Debug)]
+struct ChunkEdge {
+    parent: ChunkVc,
+    child: ChunkVc,
+    child_idx: usize,
```
sokra, line 194: Extra maintaining the index is unnecessary...

sokra (changes requested):

crates/turbopack-core/src/chunk/mod.rs
sokra: What's the difference?

crates/turbopack-css/src/chunk/optimize.rs
sokra: It would be better to limit the number of total chunks for the whole chunk group for a certain amount. That will limit the concurrent requests from the browser. e. g. the ecmascript optimizer limits it to 20.

crates/turbopack-css/src/chunk/writer.rs
```diff
@@ -36,6 +36,18 @@ pub async fn expand_imports(
                     close,
                 ));
             }
+            Some(CssImport::Composes(composed_chunk_item)) => {
```
sokra, line 39: There should be some kind of deduplication of imports here (and in CssImport::Internal) to avoid including a CSS chunk item multiple times into a chunk.

sokra (changes requested):

crates/turbopack-css/src/chunk/writer.rs
sokra: I think a header is useful to separate the chunk items in the chunk.

crates/turbopack-css/src/module_asset.rs
```diff
@@ -62,8 +75,93 @@ impl Asset for ModuleCssModuleAsset {
...
+
+#[turbo_tasks::value_impl]
+impl ModuleCssModuleAssetVc {
+    #[turbo_tasks::function]
+    async fn classes(self) -> Result<ModuleCssClassesVc> {
+        let inner = self.await?.inner;
+        let parse_result = inner.parse().await?;
+        let mut classes = IndexMap::default();
+
+        // TODO(alexkirsz) Should we report an error on parse error here?
```
sokra, line 119: The parse error is already reported, so we don't need another error

crates/turbopack-css/src/module_asset.rs
```diff
@@ -62,8 +75,93 @@ impl Asset for ModuleCssModuleAsset {
...
+        if let ParseResult::Ok { exports, .. } = &*parse_result {
+            for (class_name, export_class_names) in exports {
+                let mut export = Vec::default();
+
+                for export_class_name in export_class_names {
+                    export.push(match export_class_name {
+                        CssClassName::Import { from, name } => ModuleCssClass::Import {
+                            original: name.to_string(),
+                            from: CssModuleComposeReferenceVc::new(
+                                inner.as_resolve_origin(),
```
sokra, line 129:
```suggestion
                                self_vc.as_resolve_origin(),
```

sokra, line 129:
```suggestion
                                self_vc.as_resolve_origin(),
```

crates/turbopack-css/src/module_asset.rs
```diff
@@ -118,35 +219,156 @@ impl ValueToString for ModuleChunkItem {
...
             "{} (css module)",
-            self.module.await?.source.path().to_string().await?
+            self.module
+                .await?
+                .inner
+                .await?
+                .source
+                .path()
+                .to_string()
+                .await?
```
sokra, line 229:
```suggestion
            self.module
                .path()
                .to_string()
                .await?
```

Isn't that enough?

crates/turbopack-css/src/module_asset.rs
```diff
@@ -118,35 +219,156 @@ impl ValueToString for ModuleChunkItem {
...
+        // The proxy reference must come first so it is processed before other potential
+        // references inside of the CSS, like `@import` and `composes:`.
+        // This affects the order in which the resulting CSS chunks will be loaded:
+        // later references are processed first in the post-order traversal of the
+        // reference tree, and as such they will be loaded first in the resulting HTML.
+        let mut references = vec![CssProxyToCssAssetReference {
+            module: CssProxyModuleAsset {
+                module: self.module,
+            }
+            .cell(),
```
sokra, line 247:
```suggestion
            module: self.module,
```

Within `cssProxyToCssAssetReference::resolve_reference()` you can create the `CssProxyModuleAsset`

crates/turbopack-css/src/chunk/writer.rs
alexkirsz: I don't think I saw this comment at all in front, which makes me think this doesn't run at all. I'll check again.

crates/turbo-tasks/src/join_iter_ext.rs
sokra: Yeah, that sounds like a good idea. So you have to opt-in into `Sort::Unordered` and it's visible from usage that it's non-deterministic.

`try_flat_map_recursive_join(Unordered, |...| { ... })`
or
`try_flat_map_recursive_join(Topologically, |...| { ... })`

alexkirsz: Yes. Although I'd probably go for `try_flat_map_recursive_join::<Unordered>::(...)` if that's something we can make generic enough.

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -168,59 +168,167 @@ impl ChunkGroupVc {
...
+/// Also returns the total number of chunks in the graph.
+async fn get_chunks_edges_unsorted(root: ChunkVc) -> Result<(Vec<ChunkEdge>, usize)> {
+    let mut visited = HashSet::new();
+    let edges = [
+        // Start with an edge from the root chunk to itself. This edge will be skipped later,
+        // its only purpose is to start the traversal.
+        ChunkEdge {
+            parent: root,
+            child: root,
+            child_idx: 0,
```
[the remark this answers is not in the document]
alexkirsz: It turns out this is already the case in the current implementation. While the order of the result is non-deterministic, sibling order is preserved. I'll remove `child_idx` entirely.

crates/turbopack-css/src/chunk/writer.rs
```diff
@@ -18,10 +21,16 @@ pub async fn expand_imports(
         "".to_string(),
     )];
     let mut external_imports = vec![];
+    let mut imported_chunk_items: HashSet<CssChunkItemVc> = HashSet::default();
+    let mut composed_chunk_items: HashSet<CssChunkItemVc> = HashSet::default();
```
sokra, line 25: They should share a single set

crates/turbopack-css/src/chunk/optimize.rs
```diff
@@ -59,49 +66,60 @@ async fn merge_chunks(first: CssChunkVc, chunks: &[CssChunkVc]) -> Result<CssChu
...
-        if local.len() > LOCAL_CHUNK_MERGE_THRESHOLD {
-            let merged = take(&mut local);
-            if let Some(first) = merged.first().copied() {
-                local.push(merge_chunks(first, &merged).await?);
-            }
+/// The maximum number of chunks to exist in a single chunk group. The optimizer
+/// will merge chunks into groups until it has at most this number of chunks.
+const MAX_CHUNK_COUNT: usize = 20;
+
+fn aggregate_adjacent_chunks(chunks: &[ChunkVc]) -> Vec<Vec<ChunkVc>> {
```
sokra, line 73: Doc comment

crates/turbopack-css/src/chunk/optimize.rs
```diff
@@ -59,49 +66,60 @@ async fn merge_chunks(first: CssChunkVc, chunks: &[CssChunkVc]) -> Result<CssChu
...
-        // TODO be smarter in selecting the chunks to merge
-        // see ecmascript implementation
-        for merged in take(&mut chunks).chunks(size) {
-            chunks.push(merge_chunks(*merged.first().unwrap(), merged).await?);
-        }
+
+    chunks_vecs
+}
+
+async fn merge_adjacent_chunks(chunks_vc: ChunksVc) -> Result<ChunksVc> {
```
sokra, line 103: Doc comment

crates/turbopack-css/src/module_asset.rs
```diff
@@ -62,8 +75,93 @@ impl Asset for ModuleCssModuleAsset {
...
+            .chain(module_references.iter().copied())
+            .collect();
+
+        Ok(AssetReferencesVc::cell(references))
+    }
+}
+
+#[turbo_tasks::value(transparent)]
+#[derive(Debug, Clone)]
+enum ModuleCssClass {
```
sokra, line 94: Doc comment

crates/turbopack-css/src/references/compose.rs
```diff
@@ -0,0 +1,51 @@
...
+    reference::{AssetReference, AssetReferenceVc},
+    reference_type::CssReferenceSubType,
+    resolve::{origin::ResolveOriginVc, parse::RequestVc, ResolveResultVc},
+};
+
+use crate::references::css_resolve;
+
+#[turbo_tasks::value]
+#[derive(Hash, Debug)]
+pub struct CssModuleComposeReference {
```
sokra, line 14: Doc comment

sokra (approved):

sokra (changes requested):
A test case

crates/turbopack-css/src/chunk/writer.rs
alexkirsz: I thought about it, but imported chunk items can also have opening/closing text that composed items don't (which I'm not sure the purpose of), so what happens if we add a composed chunk item, and *then* the same, imported chunk item, with these open/close attributes?

alexkirsz: I've updated the code and the tests to account for the possibility of two imports of the same file with different import attributes.

sokra (approved):

PR #3774, opened 2023-02-13
Snapshot issues in next-dev-tests

wbinnssmith:
This:
* Adds issue snapshot support to next-dev-tests. This will allow us to assert that certain issues are raised in tests that require next-dev.
* Extracts common snapshot code into a new crates, `turbopack-test-utils`, which is shared between snapshot tests and next-dev-tests.
* Implements an issue reporter that emits issues in a channel to the the integration test code, where they are snapshotted.
* Fixes an issue where next-dev tests that were not Next.js apps would emit non-fatal issues for a missing `pages/` directory.

Co-authored-by: Justin Ridgewell <justin@ridgewell.name>

### DIFF
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -527,7 +527,7 @@ impl NextConfigVc {
     }
 }
 
-fn next_configs() -> StringsVc {
+pub fn next_configs() -> StringsVc {
     StringsVc::cell(
         ["next.config.mjs", "next.config.js"]
             .into_iter()
--- a/crates/next-core/src/router_source.rs
+++ b/crates/next-core/src/router_source.rs
@@ -2,14 +2,20 @@ use std::collections::HashSet;
 
 use anyhow::Result;
 use turbo_tasks::{primitives::StringVc, Value};
-use turbopack_core::introspect::{Introspectable, IntrospectableChildrenVc, IntrospectableVc};
+use turbopack_core::{
+    introspect::{Introspectable, IntrospectableChildrenVc, IntrospectableVc},
+    resolve::{find_context_file, FindContextFileResult},
+};
 use turbopack_dev_server::source::{
     ContentSource, ContentSourceContent, ContentSourceData, ContentSourceDataVary,
     ContentSourceResultVc, ContentSourceVc, NeededData, ProxyResult, RewriteVc,
 };
 use turbopack_node::execution_context::ExecutionContextVc;
 
-use crate::router::{route, RouterRequest, RouterResult};
+use crate::{
+    next_config::next_configs,
+    router::{route, RouterRequest, RouterResult},
+};
 
 #[turbo_tasks::value(shared)]
 pub struct NextRouterContentSource {
@@ -59,6 +65,17 @@ impl ContentSource for NextRouterContentSource {
         data: Value<ContentSourceData>,
     ) -> Result<ContentSourceResultVc> {
         let this = self_vc.await?;
+
+        // The next-dev server can currently run against projects as simple as
+        // `index.js`. If this isn't a Next.js project, don't try to use the Next.js
+        // router.
+        let project_root = this.execution_context.await?.project_root;
+        let find_config_result = &*find_context_file(project_root, next_configs()).await?;
+        if matches!(find_config_result, FindContextFileResult::NotFound(_)) {
+            return Ok(this
+                .inner
+                .get(path, Value::new(ContentSourceData::default())));
+        }
 
         let ContentSourceData {
             method: Some(method),
--- a/crates/next-dev-tests/tests/integration.rs
+++ b/crates/next-dev-tests/tests/integration.rs
@@ -1,3 +1,4 @@
+#![feature(min_specialization)]
 #![cfg(test)]
 extern crate test_generator;
 
@@ -25,16 +26,31 @@ use chromiumoxide::{
 };
 use futures::StreamExt;
 use lazy_static::lazy_static;
-use next_dev::{register, EntryRequest, NextDevServerBuilder};
+use next_dev::{EntryRequest, NextDevServerBuilder};
 use owo_colors::OwoColorize;
 use serde::Deserialize;
 use test_generator::test_resources;
-use tokio::{net::TcpSocket, task::JoinSet};
+use tokio::{
+    net::TcpSocket,
+    sync::mpsc::{channel, Sender},
+    task::{JoinHandle, JoinSet},
+};
 use tungstenite::{error::ProtocolError::ResetWithoutClosingHandshake, Error::Protocol};
-use turbo_tasks::TurboTasks;
-use turbo_tasks_fs::util::sys_to_unix;
+use turbo_tasks::{
+    NothingVc, RawVc, ReadRef, State, TransientInstance, TransientValue, TurboTasks,
+};
+use turbo_tasks_fs::{util::sys_to_unix, DiskFileSystemVc, FileSystem};
 use turbo_tasks_memory::MemoryBackend;
 use turbo_tasks_testing::retry::retry_async;
+use turbopack_core::issue::{
+    CapturedIssues, Issue, IssueReporter, IssueReporterVc, PlainIssueReadRef,
+};
+use turbopack_test_utils::snapshot::snapshot_issues;
+
+fn register() {
+    next_dev::register();
+    include!(concat!(env!("OUT_DIR"), "/register_test_integration.rs"));
+}
 
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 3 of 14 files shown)


crates/next-core/src/router_source.rs
```diff
@@ -60,6 +66,17 @@ impl ContentSource for NextRouterContentSource {
     ) -> Result<ContentSourceResultVc> {
         let this = self_vc.await?;
 
+        // The next-dev server can currently run against projects as simple as
+        // `index.js`. If this isn't a Next.js project, don't try to use the Next.js
+        // router.
+        let project_root = this.execution_context.await?.project_root;
+        let find_config_result = &*find_context_file(project_root, next_configs()).await?;
+        if matches!(find_config_result, FindContextFileResult::NotFound(_)) {
```
jridgewell, line 74: Nit: Expose a function from `crates/next-core/src/next_config.rs` that returns a `BoolVc`.

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -498,7 +505,18 @@ impl PlainIssueVc {
...
         }
 
-        Ok(U64Vc::cell(hasher.finish()))
+        hasher.finish()
+    }
+}
+
+#[turbo_tasks::value_impl]
+impl PlainIssueVc {
+    /// We need deduplicate issues that can come from unique paths, but
```
jridgewell, line 514: Move this to the `PlainIssue` impl

crates/turbo-tasks/src/state.rs
```diff
@@ -133,10 +155,35 @@ impl<T: PartialEq> State<T> {
     }
 }
 
-impl<'a, T> std::ops::Deref for StateRef<'a, T> {
+impl<'a, T> Deref for StateRef<'a, T> {
```
jridgewell, line 158: Nit: Now that only `deref_mut` will invalidate the value, I think we can remove `StateRef` entirely (and possibly rename `StateRefMut` to `StateRef`).

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -362,6 +362,19 @@ impl CapturedIssuesVc {
     pub async fn is_empty(self) -> Result<BoolVc> {
         Ok(BoolVc::cell(self.await?.is_empty()))
     }
+
+    #[turbo_tasks::function]
+    pub async fn has_fatal(self) -> Result<BoolVc> {
+        let mut has_fatal = false;
+        for issue in self.await?.iter() {
+            let severity = *issue.severity().await?;
```
jridgewell, line 370: Can we change `CapturedIssues` to be a Vec of `PlainIssueReadRef`? This still has the potential to race updates.

crates/next-core/src/next_config.rs
```diff
@@ -600,3 +600,11 @@ pub async fn load_next_config(execution_context: ExecutionContextVc) -> Result<N
         }
     }
 }
+
+#[turbo_tasks::function]
+pub async fn has_next_config(context: FileSystemPathVc) -> Result<BoolVc> {
+    Ok(BoolVc::cell(matches!(
+        *find_context_file(context, next_configs()).await?,
+        FindContextFileResult::NotFound(_)
```
jridgewell, line 608: Bug? Shouldn't this be matching `!NotFound`?

crates/next-core/src/router_source.rs
```diff
@@ -60,6 +63,16 @@ impl ContentSource for NextRouterContentSource {
     ) -> Result<ContentSourceResultVc> {
         let this = self_vc.await?;
 
+        // The next-dev server can currently run against projects as simple as
+        // `index.js`. If this isn't a Next.js project, don't try to use the Next.js
+        // router.
+        let project_root = this.execution_context.await?.project_root;
+        if *(has_next_config(project_root).await?) {
```
jridgewell, line 70:
```suggestion
        if !has_next_config(project_root).await? {
```

jridgewell (approved):

crates/next-dev-tests/tests/integration/next/tailwind/basic/issues/lint TP1004 fs.readFileSync(???__star__0__star__, "utf-8") is very dynamic-58d7af.txt
```diff
@@ -17,7 +17,7 @@ PlainIssue {
                             },
                             CellId {
                                 type_id: ValueTypeId {
-                                    id: 31,
+                                    id: 24,
```
jridgewell, line 20: Oh man, I missed that these are being injected. These are all per-process, which I why I now remember I was using `vc.dbg().await?`.

jridgewell (changes requested):

crates/next-core/js/src/entry/router.ts
```diff
@@ -142,7 +142,7 @@ export default async function route(
 }
 
 async function handleClientResponse(
-  _ipc: Ipc<RouterRequest, IpcOutgoingMessage>,
+  ipc: Ipc<RouterRequest, IpcOutgoingMessage>,
```
jridgewell, line 145: Nit: undo

crates/next-dev-tests/tests/integration.rs
```diff
@@ -414,3 +466,42 @@ async fn get_mock_server_future(mock_dir: &Path) -> Result<(), String> {
...
+        let mut has_fatal = false;
+        for issue in captured_issues.iter() {
+            let plain = issue.into_plain().await?;
+            if plain.severity == IssueSeverity::Fatal {
+                has_fatal = true;
+            }
+            issue_tx.send(plain).await?;
+        }
+
+        Ok(BoolVc::cell(has_fatal))
```
jridgewell, line 505: Nit: We can just return `BoolVc::cell(false)`, since this is for testing only.

crates/turbopack-test-utils/src/snapshot.rs
```diff
@@ -0,0 +1,178 @@
...
+            "{}-{}.txt",
+            plain_issue.title.replace('*', "__star__"),
+            &hash[0..6]
+        ));
+        seen.insert(path);
+
+        // Annoyingly, the PlainIssue.source -> PlainIssueSource.asset ->
+        // PlainAsset.path -> FileSystemPath.fs -> DiskFileSystem.root changes
+        // for everyone.
+        let content = format!("{:#?}", plain_issue).replace(workspace_root, "WORKSPACE_ROOT");
```
jridgewell, line 45: This is still not resolved. We can't use the `Debug` trait of the read ref, because it debugs the Cells, not the Values. We need to receive a `Vc` here, call `vc.dbg().await`, and diff that.

jridgewell (changes requested):

crates/turbopack-tests/tests/snapshot.rs
```diff
@@ -103,16 +97,35 @@ async fn run(resource: &'static str) -> Result<()> {
...
+            .await?
+            .strongly_consistent()
+            .await?;
+
+        let plain_issues = captured_issues
+            .iter()
+            .map(|issue_vc| async move {
+                Ok((
+                    issue_vc.into_plain().await?,
+                    (*(issue_vc.into_plain().dbg().await?)).clone(),
```
jridgewell, line 110:
```suggestion
                    issue_vc.into_plain().dbg().await?.clone_value(),
```

crates/next-dev-tests/tests/integration.rs
```diff
@@ -493,15 +492,13 @@ impl IssueReporter for TestIssueReporter {
...
-        let mut has_fatal = false;
         for issue in captured_issues.iter() {
-            let plain = issue.into_plain().await?;
-            if plain.severity == IssueSeverity::Fatal {
-                has_fatal = true;
-            }
-            issue_tx.send(plain).await?;
+            let plain = issue.into_plain();
+            issue_tx
+                .send((plain.await?, (*(plain.dbg().await?)).clone()))
```
jridgewell, line 498: Same `clone_value` change. We could also make the parameter a `ValueDebugStringReadRef` and not need to clone the string.

crates/next-dev-tests/tests/integration.rs
```diff
@@ -470,13 +469,13 @@ async fn get_mock_server_future(mock_dir: &Path) -> Result<(), String> {
...
     #[turbo_tasks(trace_ignore, debug_ignore)]
-    pub issue_tx: State<Sender<PlainIssueReadRef>>,
+    pub issue_tx: State<Sender<(PlainIssueReadRef, ValueDebugString)>>,
 }
 
 #[turbo_tasks::value_impl]
 impl TestIssueReporterVc {
     #[turbo_tasks::function]
-    fn new(issue_tx: TransientInstance<Sender<PlainIssueReadRef>>) -> Self {
+    fn new(issue_tx: TransientInstance<Sender<(PlainIssueReadRef, ValueDebugString)>>) -> Self {
```
jridgewell, line 478: Nit: `ValueDebugStringReadRef` would more consistent with `PlainIssueReadRef`

jridgewell (approved):

PR #3775, opened 2023-02-13
port(turborepo): SCM package

NicholasLYang:


crates/turborepo-ffi/Cargo.toml
```diff
@@ -8,9 +8,17 @@ license = "MPL-2.0"
 crate-type = ["staticlib"]
 
 [dependencies]
+anyhow = "1.0.69"
```
chris-olszewski, line 11: FWIW I'm using `thiserror` in #3632

crates/turborepo-ffi/src/scm/mod.rs
```diff
@@ -0,0 +1 @@
+pub(crate) mod git;
```
chris-olszewski, line 1: This should be its own crate so we can use SCM in Rust without requiring protobuf to be setup

crates/turborepo-ffi/src/lib.rs
```diff
@@ -43,3 +45,46 @@ pub extern "C" fn get_turbo_data_dir() -> Buffer {
     let dir = dirs.data_dir().to_string_lossy().to_string();
     proto::TurboDataDirResp { dir }.into()
 }
+
+#[no_mangle]
+pub extern "C" fn changed_files(buffer: Buffer) -> Buffer {
+    let req: proto::ChangedFilesReq = buffer.into_proto().expect("buffer is valid protobuf");
```
chris-olszewski, line 51: I think returning an error to Go instead of panicking here would be better. Go prints a ton of scary stuff if we panic here

crates/turborepo-ffi/src/scm/git.rs
```diff
@@ -0,0 +1,256 @@
...
+        if let Some(path) = file.path() {
+            // NOTE: In the original Go code, `Rel` works even if the base path is not a
+            // prefix of the original path. In Rust, `strip_prefix` returns an
+            // error if the base path is not a prefix of the original path.
+            // However since we're passing a pathspec to `git2` we know that the
+            // base path is a prefix of the original path.
+            let path = relative_to.map_or(path, |relative_to| {
+                path.strip_prefix(relative_to).unwrap_or(path)
+            });
+            files.insert(path.to_string_lossy().to_string());
```
chris-olszewski, line 62: I don't think we should be silently failing if we get a non-UTF8 path. If we don't fail here, we're going to get a weirder failure up the call stack.
```suggestion
            files.insert(path.to_str().ok_or(anyhow!("non utf8 path"))?.to_string());
```

crates/turborepo-ffi/src/scm/git.rs
```diff
@@ -0,0 +1,256 @@
...
+    } else {
+        None
+    };
+    let diff = repo.diff_tree_to_tree(Some(&from_tree), Some(&to_tree), options.as_mut())?;
+    diff.print(DiffFormat::NameOnly, |_, _, _| true)?;
+
+    for delta in diff.deltas() {
+        let file = delta.old_file();
+        if let Some(path) = file.path() {
+            files.insert(path.to_string_lossy().to_string());
```
chris-olszewski, line 95: Again, I don't think we should be passing the buck on non-UTF8 paths

crates/turborepo-ffi/src/scm/git.rs
```diff
@@ -0,0 +1,256 @@
...
+    let to_commit = repo.find_commit(to_commit_oid)?;
+    let from_tree = from_commit.tree()?;
+    let to_tree = to_commit.tree()?;
+    let mut options = if let Some(relative_to) = relative_to {
+        let mut options = DiffOptions::new();
+        options.pathspec(relative_to);
+        Some(options)
+    } else {
+        None
+    };
```
chris-olszewski, line 88: I'm surprised clippy didn't complain about [manual map](https://rust-lang.github.io/rust-clippy/master/#manual_map)

chris-olszewski (changes requested):
My biggest concerns are breaking the SCM stuff into it's own crate and how we deal with non-UTF8 paths. We should either error on non-UTF8 paths or we should be honest and change `changed_files` to return `Result<HashSet<PathBuf>>`

crates/turborepo-ffi/Cargo.toml
arlyon: I think for FFI thiserror is probably the better choice. turborepo-lib is a little more up in the air.

crates/turborepo-ffi/Cargo.toml
```diff
@@ -8,9 +8,17 @@ license = "MPL-2.0"
...
 directories = "4.0.1"
+dunce = "1.0.3"
```
arlyon, line 13: Cool! Haven't seen this crate around before.

cli/internal/ffi/proto/messages.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.1
-// 	protoc        v3.21.12
```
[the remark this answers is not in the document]
NicholasLYang: Is there a way to fix this so it doesn't depend on our local protobuf version?

crates/turborepo-ffi/src/scm/git.rs
NicholasLYang: Yeah this is something we should standardize as a whole. `to_string_lossy` kinda snuck into our codebase and now it's in a few places.

crates/turborepo-ffi/src/scm/mod.rs
NicholasLYang: Should we have crates for all the dependencies we're porting? That seems a little too granular.

chris-olszewski: I'm in favor of being too granular with our crates for the time being. If the overhead for adding crates is too much then we can add a `turborepo-util`, but I'm concerned that will just become a dumping ground.

.github/workflows/test.yml
```diff
@@ -603,12 +605,12 @@ jobs:
       - name: Build nextest
         timeout-minutes: 120
         run: |
-          cargo nextest run --no-run --workspace --release --exclude turbo --exclude turborepo-ffi --exclude turborepo-lib
+          cargo nextest run --no-run --workspace --release --exclude turbo --exclude turborepo-ffi --exclude turborepo-lib --exclude turborepo-scm
```
chris-olszewski, line 608: Why are we excluding this crate? If I understand we exclude `turbo` and `turborepo-ffi` to skip setting up Go/Protobuf stuff and we skip `turborepo-lib` due to the native vs rust tls choice. If we're excluding it here then should it be added to the `turborepo_rust_test` action?

chris-olszewski (approved):
I don't think we're running tests for these in CI at the moment

.github/workflows/test.yml
NicholasLYang: Yeah I need to add it there. We exclude all turborepo crates so that turbopack doesn't run any of our tests.

Cargo.toml
```diff
@@ -39,6 +39,7 @@ members = [
   "crates/turborepo",
   "crates/turborepo-ffi",
   "crates/turborepo-lib",
+  "crates/turborepo-scm",
```
mehulkar, line 42: Have the Rust folks already though how we're going to divide up crates? Asked differently... why is this not in `turborepo-lib`?

cli/internal/scm/git_go.go
```diff
@@ -1,3 +1,6 @@
+//go:build go || !rust
+// +build go !rust
```
mehulkar, line 2: what does this do?

mehulkar (approved):
Just making sure, this isn't being used anywhere yet right? Do you know if any of our tests exercise a codepath where changed files would be relevant? (maybe a test that uses `--filter`)? If not, could we write a couple in prysk and then actually use this?

mehulkar (approved):
Just making sure, this isn't being used anywhere yet right? Do you know if any of our tests exercise a codepath where changed files would be relevant? (maybe a test that uses `--filter`)? If not, could we write a couple in prysk and then actually use this?

Cargo.toml
NicholasLYang: `turborepo-lib` is the code for the front-end of turbo, i.e. the entry part of the Go sandwich. We could in theory put this in `turborepo-lib` and import it into `turborepo-ffi` but we'd be relying on Rust to optimize away the two copies of `turborepo-lib` (one from the `turbo` crate and one from `turborepo-ffi`).

I do think this is worth discussing further. More crates means in theory faster build times because Rust's compilation unit is a crate (as opposed to other languages where it's a single file). But we will pay the price with the orphan rule, which is basically a restriction around trait implementation. Fortunately for us, our code is not very trait-y, so we may be able to avoid that.

cli/internal/scm/git_go.go
NicholasLYang: There's pragmas that indicate we should turn on the Rust code linked to Go or rely on the old Go code. At some point soon we'll just flip the switch and end up with only the Rust code.

Cargo.toml
mehulkar: > But we will pay the price with the orphan rule, which is basically a restriction around trait implementation. Fortunately for us, our code is not very trait-y, so we may be able to avoid that.

I hope that one day in the very near future I understand what this means :)

> I do think this is worth discussing further.

Sounds good, I'll follow along the discussion in the internal channel!

> `turborepo-lib` is the...entry part of the Go sandwich. 

Gotcha, sounds like it could be named something a bit more obvious then

PR #3778, opened 2023-02-13
Include workspace turbo.json in inputs

mehulkar:
When `inputs` is configured for a task, we want to ensure that changes to the `turbo.json` in that workspace cause a cache miss. This is similar to how we treat the `package.json` in the workspace, and also how we treat the root `turbo.json` for the global hash.


gsoltis (approved):

mehulkar:
Will let tests run in the base branch in #2706.

PR #3779, opened 2023-02-13
Run Turborepo integration/e2e tests on rust changes

mehulkar:


wbinnssmith (approved):

chris-olszewski (changes requested):
#3515 didn't get integration tests because CI was run before #3702 landed. This PR does highlight that we probably can get rid of `go_e2e` step and use `turborepo_rust` and `go` as the patterns of `go_e2e` are almost the union of those two

mehulkar:
> #3515 didn't get integration tests because CI was run before #3702 landed. This PR does highlight that we probably can get rid of `go_e2e` step and use `turborepo_rust` and `go` as the patterns of `go_e2e` are almost the union of those two

I knew we had already done this, but I couldn't find the PR! Thanks for the ref, I'll close this down.

PR #3780, opened 2023-02-13
Fix verbosity test

mehulkar:


cli/integration_tests/basic_monorepo/verbosity.t
```diff
@@ -42,58 +42,162 @@ Verbosity level 2
...
   [-0-9:.TWZ+]+ \[DEBUG] turborepo_lib::shim: Running command as global turbo (re)
   [-0-9:.TWZ+]+ \[DEBUG] turborepo_lib::cli: pkg_inference_root set to "" (re)
-  [-0-9:.TWZ+]+ \[DEBUG] turbo: Found go binary at "[\-\w\/]+" (re)
+  [-0-9:.TWZ+]+ \[DEBUG] turborepo_lib: Args passed to Go binary: (re)
+  {
+    "version": false,
+    "api": null,
+    "color": false,
+    "cpu_profile": null,
+    "cwd": "/private/var/folders/vg/sr4krlws0k12g21phhjwy4z40000gn/T/prysk-tests-1oo0sdqw/verbosity.t",
```
NicholasLYang, line 51: This path will probably change, no? Could we use a regex here?

cli/integration_tests/basic_monorepo/verbosity.t
```diff
@@ -42,58 +42,162 @@ Verbosity level 2
...
+        "scope": [],
+        "since": null,
+        "tasks": [
+          "build"
+        ],
+        "pass_through_args": []
+      }
+    }
+  }
+  [-0-9:.TWZ+]+ \[DEBUG] turbo: Found go binary at "/Users/mehulkar/dev/vercel/turbo/target/debug/go-turbo" (re)
```
NicholasLYang, line 97: Probably need a regex here too

NicholasLYang (changes requested):
We need some regexes so that these tests aren't specific to this run

PR #3784, opened 2023-02-13
Add next-rs package and crates

padmaia:


alexkirsz:
Some modules are now duplicated in this repo as we already ported some of them:
* `next_ssg` is now `next-transform-strip-page-exports` (the new one has changes to better support SSR)
* `next_dynamic` has changes to support Turbopack (see `next-transform-dynamic`)
* `next-font` might have changes as well (cc @wbinnssmith)

PR #3788, opened 2023-02-13
Documentation for Workspace Configs

mehulkar:
Preview of relevant docs changes:
- https://turbo-site-git-mehulkar-turbo-548-composable-turbojson-docs.vercel.sh/repo/docs/core-concepts/monorepos/configuring-workspaces
- https://turbo-site-git-mehulkar-turbo-548-composable-turbojson-docs.vercel.sh/repo/docs/reference/configuration#extends


docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+}
+```
+
+<Callout>
+  For now, the only valid value for the `extends` key is `["//"]`. The
+`//` syntax, you may recall, signifies "root", similar to how it's used to
+[declare root tasks][1].
+</Callout>
+
+Extended configuration in a workspace can reset or override any of [the
```
chris-olszewski, line 38: I interpret reset as a special case of overriding
```suggestion
Extended configuration in a workspace can override any of [the
```

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
+---
+title: Configuring Workspaces
+description: Turborepo lets you configure each workspace
+---
+
+import Callout from "../../../../../components/Callout";
+
+# Configuring Workspaces
+
+Most monorepos can declare a `turbo.json` in the root directory with a uniform
```
anthonyshew, line 10: This felt a little...maybe presumptuous? Not sure what word I want to use.

At any rate, I think we can jump straight to "While uniformity across your codebase tends to make things easier to handle, not all packages and workspaces are built the same. To accommodate this..."

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+Let's say your monorepo has a many Next.js apps, and one Sveltekit app. By
+default, both frameworks create their build with the `build` script in their
+respective `package.json`s. With a single `turbo.json` you can configure this
+with:
+
+```jsonc filename="turbo.json"
+{
+  "pipeline": {
+    "build": {
+      "outputs": [".next/**", ".svelte-kit/**"],
```
chris-olszewski, line 57: Should we add another key to show off that you don't need to repeat keys that don't need overrides?
Something like basic `"inputs": ["src/**/*.js"]`

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+<Callout>
+  For now, the only valid value for the `extends` key is `["//"]`. The
+`//` syntax, you may recall, signifies "root", similar to how it's used to
+[declare root tasks][1].
+</Callout>
+
+Extended configuration in a workspace can reset or override any of [the
+configuration for a pipeline task][2]. If you don't include a key, it will inherit
+the value from the extended `turbo.json`.
```
NicholasLYang, line 40: nit: "extended" here is a little ambiguous. It took me a sec to realize it meant the root `turbo.json`. Maybe we could call it "root" or "the `turbo.json` at the root"?

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+## Examples
+
+To illustrate, let's look at a couple of use cases.
+
+### Different Frameworks
+
+Let's say your monorepo has a many Next.js apps, and one Sveltekit app. By
```
NicholasLYang, line 48:
```suggestion
Let's say your monorepo has multiple Next.js apps, and one SvelteKit app. By
```

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+# Configuring Workspaces
+
+Most monorepos can declare a `turbo.json` in the root directory with a uniform
+set of tasks that apply to all workspaces. But sometimes, a monorepo can contain
+workspaces that need to configure their tasks differently. To accommodate this,
+Turborepo enables you to extend the root configuration with a `turbo.json` in
+any workspace. This flexibility enables a more diverse set of apps and packages
+to co-exist, and allows workspace owners to maintain specialized tasks and
+configuration without affecting other inhabitants of the monorepo.
```
anthonyshew, line 16: "without affecting developers working in the monorepo."

Additionally, do we want to be saying "monorepo" specifically? `turbo` can be used in multirepo situations, as well.

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+## How it Works
+
+To override any of the configuration for a task, add a `turbo.json` file in any
```
anthonyshew, line 20: "any of the configuration for a task from the root" or something of that nature. I'm thinking we need to mention the root oooone more time here.

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+workspace of your monorepo with a top level `extends` key:
```
anthonyshew, line 21: top level -> top-level

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+  "pipeline": {
+    "build": {
+      "dependsOn": ["compile"]
+    },
+    "compile": {}
+  }
+}
+```
+
+With a Workspace override, you can move that `compile` task into the
```
NicholasLYang, line 109: Is the W meant to be capitalized in Workspace?

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+## How it Works
+
+To override any of the configuration for a task, add a `turbo.json` file in any
+workspace of your monorepo with a top level `extends` key:
+
+```jsonc filename="turbo.json"
+{
+  "extends": ["//"],
+  "pipeline": {
+    // custom configuration for the workspace here
```
anthonyshew, line 27: I think we can bring slightly more color here. Something like...

// Overrides for this workspace
// dependsOn, inputs, outputs, etc.

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+  }
+}
+```
+
+<Callout>
+  For now, the only valid value for the `extends` key is `["//"]`. The
```
anthonyshew, line 33: For now, the only valid value for the `extends` key is `["//"]`. The
`//` syntax refers to the "root", similar to how it's used to
[declare root tasks][1].

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
anthonyshew, line 38: will* reset and* override

anthonyshew: A bit more intentional! I *know* what's going to happen.

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
anthonyshew: Looks like an extra "a"

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+}
+```
+
+This works, but it's important to understand that the `my-sveltekit-app#build`
+completely ovewrites `build` for the Sveltekit app. Any other config (such as
+`outputMode`) in the `build` config would be lost. To retain the `build` config,
+you would have to duplicate all the config including the override. Other than
+duplication, the obvious problem with this setup is that it can be difficult to
```
NicholasLYang, line 139:
```suggestion
redundancy, the obvious problem with this setup is that it can be difficult to
```

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+## Examples
+
+To illustrate, let's look at a couple of use cases.
+
+### Different Frameworks
+
+Let's say your monorepo has a many Next.js apps, and one Sveltekit app. By
+default, both frameworks create their build with the `build` script in their
+respective `package.json`s. With a single `turbo.json` you can configure this
```
anthonyshew, line 50: "With a single `turbo.json`,"

comma

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+- Root turbo.json cannot use the `extends` key.
+
+If you have a use case for any of these, please [file an issue][4]!
+
+## Troubleshooting
+
+In large monorepos, it can sometimes be difficult to understand how Turborepo is
+interpreting your configuration. To help, we've added a `resolvedTaskDefinition`
+to the Dry Run output. If you run `turbo run build --dry-run`, for example, the
+output will include the combination of all `turbo.json` configuration that were
```
NicholasLYang, line 175:
```suggestion
output will include the combination of all `turbo.json` configurations that were
```

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
anthonyshew: Also think we need to mention "at the root" or something here.

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+  "pipeline": {
+    "build": {
+      "outputs": [".next/**", ".svelte-kit/**"],
+    }
+  }
+}
+```
+
+Notice how both `.next/**` and `.svelte-kit/**` need to be specified as outputs,
+even though though the Next.js apps do not generate this output. With Workspace
```
anthonyshew, line 64: two `though`s

NicholasLYang (approved):
Looks good! Just a couple nits

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+overrides, you can instead add a new configuration at
```
anthonyshew, line 65: Called it "Worskspace Overrides" here. Hmmm, does that name feel more natural...

anthonyshew: Or perhaps you were using it as a common noun here?

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+}
+```
+
+### Specialized Tasks
+
+In another example, say that the `build` task in one workspace `dependsOn` a
+`compile` task. You could universally declare it as `dependsOn: ["compile"]`.
+This means that your root `turbo.json` has to have a empty `compile` task entry:
```
anthonyshew, line 95: an*

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+    "build": {
+      "dependsOn": ["compile"]
+    },
+    "compile": {}
+  }
+}
+```
+
+With a Workspace override, you can move that `compile` task into the
+`apps/my-custom-app/turbo.json` and remove it from the root.
```
anthonyshew, line 110: Would love to pepper in a little bit of higher-level motivation here.

"remove it from the root...to improve the clarity of your root tasks"
"to move the definition of the task closer to where it is used"

something of that nature

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+## Relationship to `Workspace#Task` syntax
```
anthonyshew, line 112: I'd lowercase the `W` in `Workspace`.

anthonyshew: I'm also thinking it would be more clear as "Comparing to..." instead of "Relationship to..."

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+You may be wondering: how does this relate to [`workspace#task` entries][3] in the root turbo.json?
```
anthonyshew, line 114: This got a little verbose. Let's see how can I do some cutting...

"At first glance, Workspace Configs may sound a lot like the `workspace#task` syntax. However, Workspace Configs let us be more more intentional about our task while writing less code. For example..."

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+      "outputs": [".next/**"],
+    },
+    "my-sveltekit-app#build": {
+      "outputs": [".svelte-kit/**"]
+    }
+  }
+}
+```
+
+This works, but it's important to understand that the `my-sveltekit-app#build`
```
anthonyshew, line 135: Picking up from comment above, seeing how I can do more with less...

"It might have snuck by us that our `my-sveltekit-app#build` task lost the `outputMode` when we really were only meaning to change the cache outputs. We would need to duplicate the `outputMode` key get that configuration back.

If we have many configurations, this ends up being a lot of duplicated code to maintain. With Workspace Configs, you don't need to duplicate any configuration. Instead, you only need to override what needs to be changed."

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
chris-olszewski, line 112: Syntax reads to me as we're using the same feature, just in a different way.
```suggestion
## Relationship to workspace-tasks defined in root `turbo.json`
```

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+completely ovewrites `build` for the Sveltekit app. Any other config (such as
+`outputMode`) in the `build` config would be lost. To retain the `build` config,
+you would have to duplicate all the config including the override. Other than
+duplication, the obvious problem with this setup is that it can be difficult to
+understand over time which duplications are intentional and which ones are added
+just because someone wanted to add a custom `outputs` for their workspace.
+
+With workspace overrides, you don't need duplicate any configuration, just
```
anthonyshew, line 143: "workspace overrides" again! Hmmm, maybe it is more natural...

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,181 @@
...
+  "pipeline": {
+    "build": {
+-      "outputs": [".next/**", ".svelte-kit/**"]
++      "outputs": [".next/**"]
+    }
+  }
+}
+```
+
+### Specialized Tasks
```
chris-olszewski, line 91: In our running tasks docs we use "specific workspace-tasks" as the term for tasks that only appear in a single workspace. Later in this doc we use "Workspace Specific Task". I think we need to unify how we talk about composable turbo.json compared to workspace#task.

chris-olszewski (commented):
Everything is a suggestions. Writing is hard and editing is harder

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: this whole page is inside the monorepos section. workspace overrides doesn't apply to single package repos

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: I meant "reset" and "override" as different things. But I've removed reset entirely based on @chris-olszewski's suggestion that "reset" is a special case of override, which I agree with.

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: not seeing the difference here

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: I was indecisive while writing about "worksapce overrides" vs "workspace config" and then didn't clean up after myself. fixed now

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: Agreed, but this was slightly different. I wanted to show that you can add _new_ tasks to workspaces taht don't exist anywhere else. Workspace-Specific tasks, as of today, are just about specialized configuration of _existing_ tasks.

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: I chose to be ambigious because the "can only extend from root" is framed as a starting point, and eventually we want to extend from any worksapce.

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: Talked about this internally and sticking with this, but open to word smithing!

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: this is noted i the Comparison to Workspace tasks section

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: added some more comments, but refraining from adding speciifc keys, because (1) they get out of date, and (2) there isn't anything to compare specific keys to in this intro section

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: done

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
mehulkar: reworded to talk more about the duplication

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,223 @@
...
+  "pipeline": {
+    "build": {
+      "dependsOn": ["compile"]
+    },
+    "compile": {}
+  }
+}
+```
+
+With a Workspace Configurations, you can move that `compile` task into the
```
anthonyshew, line 117: Little bit of grammar. 

a [plural]

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,223 @@
...
+}
+```
+
+Now, the owners of `my-app`, can have full ownership over their `build` task,
+but continue to inherit any other tasks defined at the root.
+
+## Comparison to workspace-tasks defined in root `turbo.json`
+
+At first glance, Workspace Configurations may sound a lot like the
+[`workspace#task`` syntax][3] in the root `turbo.json`. The features are
```
anthonyshew, line 152: Backticks formatting got messed up here!

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,223 @@
...
+similar, but have one significant difference: when you declare a workspace-task
+in the root ` turbo.json``, it _completely_ overwrites the baseline task
+configuration. With a Workspace Config, the task configuration is merged
+instead. Consider the example of the monorepo with multiple Next.js apps and a
```
anthonyshew, line 156: Line break before `Consider`

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,223 @@
...
+      "outputs": [".svelte-kit/**"]
+    }
+  }
+}
+```
+
+This works, but it's important to understand that the `my-sveltekit-app#build`
+completely ovewrites `build` for the Sveltekit app. Any other config (such as
+`outputMode`) in the `build` config would need to be duplicated. With Workspace
+Configurations, you don't need to duplicate any configuration, just override
```
anthonyshew, line 179: Period instead of comma. "...any configuration. Just override..."

docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -57,6 +57,12 @@ A list of environment variables for implicit global hash dependencies. The conte
 }
 ```
 
+## `extends`
+
+`type: string[]`
+
+The `extends` key is only valid in [Workspace Configuration][1]. It will be ignored in the root `turbo.json`.
+
```
anthonyshew, line 65: Maybe a "To learn more, <link>" to get the full rundown if I landed here from search and still feel lost?

docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -8,13 +8,13 @@ import OutputModeTable from "../../../../components/output-mode-table.mdx";
 
 # Configuration Options (`turbo.json`)
```
arlyon, line 9: Tangential to the PR, but does it not make sense for this to be generated from our schema.json at some point? Not only would it avoid drift but we could move these docs into the schema so they are available in the IDE as well.

arlyon (approved):
Looks great!

docs/pages/repo/docs/reference/configuration.mdx
mehulkar: That makes sense to me

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,266 @@
...
+# Configuring Workspaces
+
+Most monorepos can declare a `turbo.json` in the root directory with a uniform
+[pipeline][2] that apply to all workspaces. Sometimes, a monorepo can contain
+workspaces that need to configure their tasks differently. To accommodate this,
+starting in version 1.8, Turborepo enables you to extend the root configuration
+with a `turbo.json` in any workspace. This flexibility enables a more diverse
+set of apps and packages to co-exist, and allows workspace owners to maintain
+specialized tasks and configuration without affecting other apps and packages of
+the monorepo.
```
molebox, line 17: Do we need this bit `Most monorepos can declare a `turbo.json` in the root directory with a uniform[pipeline][2] that applies to all workspaces.` ? I think its cleaner to jump straight in like this:
```suggestion
From version 1.8 you can extend the root configuration of _any_ workspace with a `turbo.json` file. This flexibility enables a more diverse
set of apps and packages to co-exist, and allows workspace owners to maintain specialized tasks and configuration without affecting other apps and packages of the monorepo.
```

docs/pages/repo/docs/core-concepts/monorepos/configuring-workspaces.mdx
```diff
@@ -0,0 +1,266 @@
...
+```
+
+<Callout>
+  For now, the only valid value for the `extends` key is `["//"]`.
+`//` is a special name used to identify the root directory of the monorepo.
+</Callout>
+
+Configuration in a workspace can override any of [the configuration for a
+pipeline task][2]. If you don't include a key, the configuration is inherited
+from the extended `turbo.json`.
```
molebox, line 45:
```suggestion
Configuration in a workspace can override any of [the configurations for a
pipeline task][2]. If you don't include a key, the configuration is inherited
from the extended `turbo.json`.
```

docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -58,6 +58,13 @@ A list of environment variables for implicit global hash dependencies. The conte
 }
 ```
 
+## `extends`
+
+`type: string[]`
+
+The `extends` key is only valid in Workspace Configurations. It will be
+ignored in the root `turbo.json`. Read [the docs to learn more][1].
```
gsoltis, line 66: Will `extends` be ignored or will it error?

gsoltis (approved):

docs/pages/repo/docs/reference/configuration.mdx
mehulkar: it's ignored. we can make it error though

PR #3789, opened 2023-02-13
Update `$schema` for `turbo.json`.

anthonyshew:
The in-editor recommendations for `turbo.json` that come from `$schema` seem to have fallen behind the API in a few spots. I also took a moment to update some language for improved clarity.

I also had to shave a few yaks to work on this:
- Updated `docs#dev` task to ensure it generates the `schema.json` when we run `pnpm run docs`
- Improve deprecation note for environment variables
- Added `errors-only` to `outputMode` doc


gsoltis (approved):

packages/turbo-types/src/types/config.ts
```diff
@@ -4,71 +4,105 @@ export interface Schema {
...
+   *
+   * Tells turbo to extend your root `turbo.json`
+   * and overrides with the keys provided
+   * in your Workspace Configs.
+   *
+   * Currently, only the "//" value is allowed.
+   *
+   * @default ["//"]
+   */
+  extends?: string[];
```
mehulkar, line 75: Let's separate this into a new PR for a couple reasons:
- we want to hold it till the 1.8 release
- I'm looking into setting this up so the existence of this key prevents `globalEnv`, et al. I think it's [technically possible](https://json-schema.org/draft/2020-12/json-schema-core.html#name-keywords-for-applying-subsche), but we need to figure out how to write the JSON, and then see if the TS interface can even generate it.

turbo.json
```diff
@@ -15,6 +15,9 @@
     "dev": {
       "cache": false
     },
+    "docs#dev": {
+      "dependsOn": ["schema"]
+    },
```
mehulkar, line 20: Now that #2706 is merged, we can actually start moving some of these configs into workspaces! (definitely out of scope to do here though!)

PR #3790, opened 2023-02-14
Add `extends` to JSON schema.

anthonyshew:
Held back from #3789 so it can come in with v1.8!

### DIFF
--- a/packages/turbo-types/src/types/config.ts
+++ b/packages/turbo-types/src/types/config.ts
@@ -164,6 +164,20 @@ export interface Pipeline {
    * @default false
    */
   persistent?: boolean;
+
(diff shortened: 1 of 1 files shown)


mehulkar (changes requested):
Going to put this on hold for a bit. If possible, we want to add some additional validation to the schema. Namely: If the `extends` key is present, `globalEnv`, `globalDependencies` and `remoteCache` are not allowed.

I'll try a couple more things today and make a PR to this branch! (If you figure it out before me, feel free to just push that here also!)

tknickman:
Closing in favor of https://github.com/vercel/turbo/pull/3821

PR #3795, opened 2023-02-14
fix export * warning for client pages

sokra:
This gets rid of this warning when using a client component as page:

```
warning - [analyze] [project-with-next]/src/app/client/page.jsx  unexpected export *
  export * used with module [project-with-next]/src/app/client/page.jsx which has no exports
  Typescript only: Did you want to export only types with `export type { ... } from "..."`?
```

* fix export * warning for client pages
* Detect `__turbopack_export_value__` as CJS style exports
* fix `__turbopack_cjs__` name
* flag turbopackHelper in client proxy

### DIFF
--- a/crates/turbopack-ecmascript/src/references/esm/export.rs
+++ b/crates/turbopack-ecmascript/src/references/esm/export.rs
@@ -167,7 +167,7 @@ impl CodeGenerateable for EsmExports {
                     let ident = ReferencedAsset::get_ident_from_placeable(asset).await?;
 
                     cjs_exports.push(quote_expr!(
-                        "__turbopack__cjs__($arg)",
+                        "__turbopack_cjs__($arg)",
                         arg: Expr = Ident::new(ident.into(), DUMMY_SP).into()
                     ));
                 }
--- a/crates/turbopack-ecmascript/src/references/mod.rs
+++ b/crates/turbopack-ecmascript/src/references/mod.rs
@@ -2134,6 +2134,20 @@ async fn resolve_as_webpack_runtime(
 #[turbo_tasks::value(transparent, serialization = "none")]
 pub struct AstPath(#[turbo_tasks(trace_ignore)] Vec<AstParentKind>);
 
+pub static TURBOPACK_HELPER: &str = "__turbopackHelper";
+
+pub fn is_turbopack_helper_import(import: &ImportDecl) -> bool {
+    import.asserts.as_ref().map_or(true, |asserts| {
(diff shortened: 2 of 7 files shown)


crates/turbopack-ecmascript/src/references/esm/export.rs
```diff
@@ -167,7 +167,7 @@ impl CodeGenerateable for EsmExports {
                     let ident = ReferencedAsset::get_ident_from_placeable(asset).await?;
 
                     cjs_exports.push(quote_expr!(
-                        "__turbopack__cjs__($arg)",
+                        "__turbopack_cjs__($arg)",
```
sokra, line 170: @kdy1 I really wonder why we didn't notice that earlier...

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -2139,7 +2139,22 @@ fn has_cjs_export(p: &Program) -> bool {
...
+                    .and_then(|import| import.asserts.as_ref())
+                    .map_or(true, |asserts| {
+                        asserts.props.iter().any(|assert| {
+                            assert
+                                .as_prop()
+                                .and_then(|prop| prop.as_key_value())
+                                .and_then(|kv| kv.key.as_ident())
+                                .map_or(true, |ident| &*ident.sym != "turbopackHelper")
+                        })
+                    })
```
alexkirsz, line 2155: This should be its own function `is_turbopack_helper_module_decl(module_decl) -> bool`.

crates/turbopack-ecmascript/src/transform/server_to_client_proxy.rs
```diff
@@ -52,7 +53,13 @@ pub fn create_proxy_module(transition_name: &str, target_import: &str) -> Progra
                 })],
                 src: box target_import.into(),
                 type_only: false,
-                asserts: None,
+                asserts: Some(box ObjectLit {
+                    span: DUMMY_SP,
+                    props: vec![PropOrSpread::Prop(box Prop::KeyValue(KeyValueProp {
+                        key: PropName::Ident(Ident::new("turbopackHelper".into(), DUMMY_SP)),
```
alexkirsz, line 59: Should this be named `__turbopackHelper` juuuust in case?

`"turbopackHelper"` should move to a `&'static str` constant that's shared between this module and `is_turbopack_helper_module_decl`.

alexkirsz (approved):
2 nits

PR #3803, opened 2023-02-14
Restore Issue Reporters

wbinnssmith:
This restores issue reporters, addressing a bug that prevented turbo-trace from completing. This moves `ConsoleUiVc::new` into an async block to prevent it from stalling.

Test Plan: Verify `cargo run --bin node-file-trace -- print path/to/my/app` no longer stalls.

### DIFF
--- a/crates/node-file-trace/src/lib.rs
+++ b/crates/node-file-trace/src/lib.rs
@@ -37,12 +37,12 @@ use turbopack::{
     resolve_options_context::ResolveOptionsContext, transition::TransitionsByNameVc,
     ModuleAssetContextVc,
 };
-use turbopack_cli_utils::issue::{ConsoleUi, IssueSeverityCliOption, LogOptions};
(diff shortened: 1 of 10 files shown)


crates/node-file-trace/src/lib.rs
```diff
@@ -487,24 +487,27 @@ async fn run<B: Backend + 'static, F: Future<Output = ()>>(
...
-        log_detail,
-        log_level: log_level.map_or_else(|| IssueSeverity::Error, |l| l.0),
-    }));
     let task = tt.spawn_root_task(move || {
         let dir = dir.clone();
         let args = args.clone();
-        let console_ui = console_ui.clone();
         let sender = sender.clone();
         Box::pin(async move {
+            let console_ui = ConsoleUiVc::new(TransientInstance::new(LogOptions {
```
wbinnssmith, line 495: This was moved here from the surrounding block to address the issue.

jridgewell: Nit: move this down to where it's needed.

jridgewell (approved):

PR #3805, opened 2023-02-14
fix: allow HMR to recover from hydration errors

ForsakenHarmony:
We now store react errors in the error overlay (hydration errors) and if the error overlay receives the "build ok" message after that it will hard reload the page to recover

### DIFF
--- a/crates/next-core/js/src/overlay/internal/ReactDevOverlay.tsx
+++ b/crates/next-core/js/src/overlay/internal/ReactDevOverlay.tsx
@@ -3,7 +3,6 @@ import * as React from "react";
 import type { Issue } from "@vercel/turbopack-runtime/types/protocol";
 
 import * as Bus from "./bus";
-import { TYPE_REACT_ERROR } from "./bus";
 import { ShadowPortal } from "./components/ShadowPortal";
 import { Errors, SupportedErrorEvent } from "./container/Errors";
 import { ErrorBoundary } from "./ErrorBoundary";
@@ -168,7 +167,7 @@ export default function ReactDevOverlay({
   const onComponentError = React.useCallback(
     (error: Error, componentStack: string | null) => {
       Bus.emit({
-        type: TYPE_REACT_ERROR,
+        type: Bus.TYPE_REACT_ERROR,
         error,
         componentStack,
       });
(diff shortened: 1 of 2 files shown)


ForsakenHarmony:
* main
  * **PR #3805** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3805" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3805?utm_source=stack-comment).

sokra:
Could you explain how this works in the PR description?

crates/next-core/js/src/overlay/internal/ReactDevOverlay.tsx
```diff
@@ -3,6 +3,7 @@ import * as React from "react";
 import type { Issue } from "@vercel/turbopack-runtime/types/protocol";
 
 import * as Bus from "./bus";
+import { TYPE_REACT_ERROR } from "./bus";
```
alexkirsz, line 6: Nit: this is the only value we import this way, while we refer to the others as `Bus.*`.

crates/next-core/js/src/overlay/internal/ReactDevOverlay.tsx
```diff
@@ -50,6 +53,13 @@ function pushErrorFilterDuplicates(
 function reducer(state: OverlayState, ev: Bus.BusEvent): OverlayState {
   switch (ev.type) {
     case Bus.TYPE_BUILD_OK: {
+      if (state.reactError != null) {
+        console.warn(
+          "[Fast Refresh] performing full reload because your application had an unrecoverable error"
+        );
+        window.location.reload();
```
alexkirsz, line 60: I'm guessing this is how Next.js does it, but why not try to re-render the component instead of reloading the window?

ForsakenHarmony: That's a good question

crates/next-core/js/src/overlay/internal/ReactDevOverlay.tsx
ForsakenHarmony: fixed

crates/next-core/js/src/overlay/internal/ReactDevOverlay.tsx
ForsakenHarmony: I'll try implementing that

one problem I already see with either implementation right now is the issues from other chunks preventing the `onBuildOk` call

ForsakenHarmony: hmm I think I might go for the hard reload for now and revisit in the future, running into a bunch of edge cases when I try using it without

the react error boundary is a huge pain

alexkirsz (approved):

PR #3816, opened 2023-02-15
Reduce turbotrace pressure

Brooooooklyn:
This pull request reduces the CPU/Memory pressures while running turbotrace.

- Skip parse CSS modules if `EnvironmentIntention` is `Api/Data/Middleware`
- Skip sourcemaps resolving if files are in `node_modules`

Additionally, this pr allows passing `ModuleOptionsContext` and `ResolveOptionsContext` in `fn start`, so next.js can pass `enableJsx: true, enableTypes: true` while tracing entries, and `enableJsx: false, enableTypes: false` while tracing compiled files.

This also resolves deadlocks while reaching `memory_limit`.


crates/node-file-trace/src/lib.rs
```diff
@@ -267,16 +266,23 @@ async fn input_to_modules<'a>(
...
-    let context: AssetContextVc = ModuleAssetContextVc::new(
-        TransitionsByNameVc::cell(HashMap::new()),
-        env,
-        ModuleOptionsContext {
-            enable_types: true,
-            enable_mdx,
-            ..Default::default()
+    let resolve_options_context = match resolve_options {
+        Some(r) => {
+            let mut resolve_options = (&*r.await?).clone();
```
alexkirsz, line 271:
```suggestion
            let mut resolve_options = r.await?.clone_value();
```

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -230,6 +230,7 @@ pub(crate) async fn analyze_ecmascript_module(
             let mut import_references = Vec::new();
 
             let pos = program.span().lo;
+            let is_inside_node_modules = *path.contains("node_modules").await?;
```
alexkirsz, line 233: This seems like a very fragile assertion. What if the path contains `some_folder_that_is_definitely_not_node_modules`? We have a more involved check for "is in directory" in crates/turbopack/src/condition.rs.

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -262,27 +263,30 @@ pub(crate) async fn analyze_ecmascript_module(
...
-                        if let Some(m) = SOURCE_MAP_FILE_REFERENCE.captures(&comment.text) {
-                            let path = &m[1];
-                            // TODO this probably needs to be a field in EcmascriptModuleAsset so it
-                            // knows to use that SourceMap when running code generation.
-                            // The reference is needed too for turbotrace
-                            analysis.add_reference(SourceMapReferenceVc::new(
-                                source.path(),
-                                source.path().parent().join(path),
-                            ))
+            if !is_inside_node_modules {
```
alexkirsz, line 266: Are we sure that we don't want source maps for `node_modules` packages? What about Next.js `transpilePackages`?

crates/turbopack/src/lib.rs
```diff
@@ -146,10 +147,28 @@ async fn apply_module_type(
         ModuleType::Json => JsonModuleAssetVc::new(source).into(),
         ModuleType::Raw => source,
         ModuleType::Css(transforms) => {
-            CssModuleAssetVc::new(source, context.into(), *transforms).into()
+            if matches!(
+                &*intention,
+                EnvironmentIntention::Api
+                    | EnvironmentIntention::Middleware
+                    | EnvironmentIntention::Data
+            ) {
```
alexkirsz, line 155: Perhaps this should be a method on `environment` to check whether a particular module type is supported or not. I'm wary of this match not being exhaustive in the future if we add another environment intention, or custom environments.

crates/node-file-trace/src/lib.rs
```diff
@@ -287,6 +293,12 @@ async fn input_to_modules<'a>(
             ..Default::default()
         }
         .cell(),
+    };
+    let context: AssetContextVc = ModuleAssetContextVc::new(
+        TransitionsByNameVc::cell(HashMap::new()),
```
sokra, line 298: This cell call will make every call of input_to_modules to have a separate AssetContext, so multiple call will not share modules.
Could you move the AssetContext creation into a separate TT function

crates/turbopack-ecmascript/src/references/mod.rs
sokra: I don't think we want to disable that at all in next.js. SourceMaps for everyone

crates/turbo-tasks/Cargo.toml
```diff
@@ -18,7 +18,7 @@ hanging_detection = []
 [dependencies]
 anyhow = "1.0.47"
 auto-hash-map = { path = "../auto-hash-map" }
-concurrent-queue = "1.2.2"
+concurrent-queue = "2.1.0"
```
sokra, line 21: Is this the deadlock fix?

arlyon:
@sokra @alexkirsz is it worth closing this? Since turborepo is on the reviewer list it keeps appearing in my backlog

PR #3818, opened 2023-02-15
feat(@turbo/workspaces): add new workspaces package (#3818

tknickman:
Codemods to convert between workspace managers. Node API will be used by `create-turbo` when cloning examples. 

This is also provided as a standalone CLI.

```
Usage:
  $ npx @turbo/workspace-convert [flags...] [<dir>]
  
  If <dir> is not provided, you will be prompted for it.
  
  Flags:
    --npm           Switch to npm workspaces
    --pnpm          Switch to pnpm workspaces
    --yarn          Switch to yarn workspaces
    --install       Convert lock files and install dependencies
    --dry           Run without making any modifications
    --summary       Show a summary of the workspace at <dir>, including the detected package
                    manager, all workspaces, and their locations
    --help, -h      Show this help message
    --version, -v   Show the version of this script
```


packages/turbo-workspace-convert/__tests__/managers/npm.test.ts
```diff
@@ -0,0 +1,203 @@
...
+        ).toBeUndefined();
+      }
+    );
+  });
+
+  describe("clean", () => {
+    test.each([
+      [true, true],
+      [false, true],
+      [false, true],
```
nathanhammond, line 179: I think we're missing `false, false`?

packages/turbo-workspace-convert/__tests__/managers/npm.test.ts
```diff
@@ -0,0 +1,203 @@
...
+          );
+        }
+      }
+    );
+  });
+
+  describe("read", () => {
+    test.each<[PackageManagers, boolean]>([
+      ["yarn", true],
+      ["npm", false],
```
nathanhammond, line 113: I'd derive this, but don't care a lot.

packages/turbo-workspace-convert/__tests__/managers/pnpm.test.ts
```diff
@@ -0,0 +1,246 @@
...
+import { PackageManagers } from "../../src/types";
+import {
+  extendMatrix,
+  generateTestMatrix,
+  validateWorkspace,
+} from "../test-utils";
+import fs from "fs-extra";
+import { existsSync } from "fs-extra";
+
+const pnpm = MANAGERS.pnpm;
```
nathanhammond, line 14: Splitting these test files really feels like we're going to get significant drift for something that is nearly identical, or could at least be made to have identical interface assertion helpers.

Can we loop 'em in one?

packages/turbo-workspace-convert/__tests__/managers/pnpm.test.ts
```diff
@@ -0,0 +1,246 @@
...
+        false
+      );
+    });
+  });
+
+  describe("clean", () => {
+    test.each([
+      [true, true],
+      [false, true],
+      [false, true],
```
nathanhammond, line 222: `false, false`?

packages/turbo-workspace-convert/__tests__/managers/yarn.test.ts
```diff
@@ -0,0 +1,205 @@
...
+          workspaceRoot: root,
+        })
+      ).toEqual(result);
+    });
+  });
+
+  describe("remove", () => {
+    test.each(
+      extendMatrix([
+        ["npm", "8.19.2", ["apps/*", "packages/*"]],
```
nathanhammond, line 40: (Still want to know why these need enumeration, this time for `npm`.)

packages/turbo-workspace-convert/__tests__/test-utils.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  const matrix: Array<
+    [PackageManagers, PackageManagers, boolean, boolean, boolean]
+  > = [];
+  for (const p1 of packageManagers) {
+    for (const p2 of packageManagers) {
+      for (const i of interactive) {
+        for (const d of dryRun) {
+          for (const inst of install) {
+            matrix.push([p1, p2, i, d, inst]);
```
nathanhammond, line 89: yo dawg.

packages/turbo-workspace-convert/jest.config.js
```diff
@@ -0,0 +1,18 @@
...
+  modulePathIgnorePatterns: ["<rootDir>/node_modules", "<rootDir>/dist"],
+  collectCoverage: true,
+  coverageThreshold: {
+    global: {
+      branches: 83,
+      functions: 80,
+      lines: 86,
+      statements: 86,
+    },
+  },
```
nathanhammond, line 17: hah, magicints for "thou shalt not regress"

packages/turbo-workspace-convert/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "build": "tsup",
+    "dev": "tsup --watch",
+    "test": "jest",
+    "lint": "eslint src/**/*.ts",
+    "check-types": "tsc --noEmit"
+  },
+  "dependencies": {
+    "chalk": "2.4.2",
+    "execa": "5.1.1",
+    "fast-glob": "^3.2.12",
```
nathanhammond, line 27: So, fun story, npm and (pnpm/yarn) use different glob libraries. Just in case you wanted to _also_ include `minimatch` here and try to check compat. 😜

packages/turbo-workspace-convert/src/cli.ts
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env node
+
+import inquirer from "inquirer";
+import path from "path";
+import meow from "meow";
```
nathanhammond, line 5: Are we using `meow` everywhere or just here?

packages/turbo-workspace-convert/src/cli.ts
```diff
@@ -0,0 +1,187 @@
...
+const help = `
+  Usage:
+    $ npx @turbo/workspace-convert [flags...] [<dir>]
+
+  If <dir> is not provided, you will be prompted for it.
+
+  Flags:
+    --npm           Switch to npm workspaces
+    --pnpm          Switch to pnpm workspaces
+    --yarn          Switch to yarn workspaces
```
nathanhammond, line 59: I feel like this shouldn't be a flag.

`npx @turbo/workspace-convert <packageManager> <dir>`

Or if it is it should be `--target=<>`.

packages/turbo-workspace-convert/src/cli.ts
```diff
@@ -0,0 +1,187 @@
...
+    --install       Convert lock files and install dependencies
+    --dry           Run without making any modifications
+    --summary       Show a summary of the workspace at <dir>, including the detected package
```
nathanhammond, line 62: Hm, maybe I'm really just asking for there to be a `convert` subcommand? `summary` could be a no-args, no-flags kind of deal.

Maybe the package name could be `workspace-inspector`?

packages/turbo-workspace-convert/src/cli.ts
```diff
@@ -0,0 +1,187 @@
...
+      version: { type: "boolean", default: false, alias: "v" },
+    },
+  });
+
+  if (flags.help) showHelp();
+  if (flags.version) showVersion();
+  const logger = new Logger({ dry: flags.dry, interactive: true });
+
+  logger.hero();
+  await new Promise((resolve) => setTimeout(resolve, 500));
```
nathanhammond, line 99: 🤨
```suggestion
```

packages/turbo-workspace-convert/src/cli.ts
```diff
@@ -0,0 +1,187 @@
...
+  logger.header("Welcome, let's convert your project.");
+  logger.blankLine();
+  // Figure out the app directory
+  let projectDir = path.resolve(
+    process.cwd(),
+    input.length > 0
+      ? input[0]
```
nathanhammond, line 106: Does `meow` let you deal with sentinel "unset" values or are they indistinguishable from empty string?

(Not a request for change.)

packages/turbo-workspace-convert/src/cli.ts
```diff
@@ -0,0 +1,187 @@
...
+        message: "Should we automatically install after conversion?",
+        default: "yes",
+        choices: ["yes", "no"],
+      },
+    ]);
+    shouldInstall = installResponse.install === "yes";
+  }
+
+  // if it's absolute, don't worry about it
+  const workspaceRoot = path.isAbsolute(projectDir)
```
nathanhammond, line 134: `path.resolve(process.cwd(), 'literally anything')` will always be absolute, no?

In other words, I think this is unnecessary.

packages/turbo-workspace-convert/src/cli.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  let answers: Answers;
+  if (flags.npm) {
+    answers = { packageManager: "npm" };
+  } else if (flags.pnpm) {
+    answers = { packageManager: "pnpm" };
+  } else if (flags.yarn) {
+    answers = { packageManager: "yarn" };
+  } else {
```
nathanhammond, line 157: This is why they shouldn't be individual flags.

`convert --pnpm --npm` will _very_ confusingly return `npm`.

packages/turbo-workspace-convert/src/cli.ts
```diff
@@ -0,0 +1,187 @@
...
+    answers = await inquirer.prompt<{
+      packageManager: PackageManagers;
+    }>([
+      {
+        name: "packageManager",
+        type: "list",
+        message: `Convert from ${project.packageManager} workspaces to:`,
+        choices: ["npm", "pnpm", "yarn"].map((p) => ({
```
nathanhammond, line 165: there can be only one!

packages/turbo-workspace-convert/src/cli.ts
```diff
@@ -0,0 +1,187 @@
...
+              name: "dir",
+              message: "Where is the root of your monorepo?",
+              default: ".",
+            },
+          ])
+        ).dir
+  );
+
+  let shouldInstall = flags.install;
+  if (shouldInstall === undefined && !flags.dry) {
```
nathanhammond, line 120: Make the variable state always map to the meaning.

```suggestion
  let shouldInstall = flags.dry === true ? false : flags.install;
  if (shouldInstall === undefined) {
```

nathanhammond (commented):
Part one.

packages/turbo-workspace-convert/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+    })
+    .map((workspacePackageJson) => {
+      const workspaceRoot = path.dirname(workspacePackageJson);
+      const name = getWorkspaceName({ workspaceRoot });
+      return {
+        name,
+        paths: {
+          root: workspaceRoot,
+          packageJson: workspacePackageJson,
+          nodeModules: path.join(workspaceRoot, "node_modules"),
```
nathanhammond, line 109: This isn't a fact being determined by this function.

```suggestion
```

packages/turbo-workspace-convert/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+}: {
+  dependencyList: DependencyList;
+  project: Project;
+  to: PackageManagerDetails;
+}): { dependencyList: DependencyList; updated: Array<string> } {
+  const updated: Array<string> = [];
+  project.workspaceData.workspaces.forEach((workspace) => {
+    const { name } = workspace;
+    if (dependencyList[name]) {
+      dependencyList[name] = to.name === "pnpm" ? `workspace:*` : "*";
```
nathanhammond, line 128: This should preserve any versions present.

packages/turbo-workspace-convert/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+  options?: Options;
+}) {
+  const workspacePackageJson = getPackageJson({
+    workspaceRoot: workspace.paths.root,
+  });
+
+  // update all dependencies
+  const stats: Record<string, Array<string>> = {
+    dependencies: [],
+    devDependencies: [],
```
nathanhammond, line 156: `peerDependencies`, `optionalDependencies`

packages/turbo-workspace-convert/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+  project: Project;
+  to: PackageManagerDetails;
+  logger: Logger;
+  options?: Options;
+}) {
+  const workspacePackageJson = getPackageJson({
+    workspaceRoot: workspace.paths.root,
+  });
+
+  // update all dependencies
```
nathanhammond, line 153: I'd appreciate a couple more comment lines in this function because it took me longer than I wanted to read.

packages/turbo-workspace-convert/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+      project,
+      to,
+    });
+    workspacePackageJson["dependencies"] = dependencyList;
+    stats.dependencies = updated;
+  }
+
+  const numDevDependencies = stats.devDependencies.length;
+  const numDependencies = stats.dependencies.length;
+  const suffix = (num: number) => `${num <= 1 ? "y" : "ies"}`;
```
nathanhammond, line 180: Too clever by half, switch on the whole word?

Also 0 gets "ies"

packages/turbo-workspace-convert/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  // both updated
+  if (numDependencies > 0 && numDevDependencies > 0) {
+    logger.workspaceStep(
+      `updating ${chalk.green(
+        numDevDependencies.toString()
+      )} devDependenc${suffix(numDependencies)} and ${chalk.green(
```
nathanhammond, line 187: This is either `devDependencies` or "dev dependencies". Using the key name shouldn't follow plural rules.

packages/turbo-workspace-convert/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+        workspace.paths.packageJson
+      )}`
+    );
+  }
+
+  if (!options?.dry) {
+    fs.writeJSONSync(workspace.paths.packageJson, workspacePackageJson, {
+      spaces: 2,
+    });
+  }
```
nathanhammond, line 227: This doesn't feel very much like a "utility". There is application logic tucked in here.

packages/turbo-workspace-convert/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+    });
+    workspacePackageJson["dependencies"] = dependencyList;
+    stats.dependencies = updated;
+  }
+
+  const numDevDependencies = stats.devDependencies.length;
+  const numDependencies = stats.dependencies.length;
+  const suffix = (num: number) => `${num <= 1 ? "y" : "ies"}`;
+
+  // both updated
```
nathanhammond, line 182: This whole `if`/`else if` strategy needs to change to allow for `optionalDependencies` and `peerDependencies`

packages/turbo-workspace-convert/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  return { dependencyList, updated };
+}
+
+function updateDependencies({
+  project,
+  workspace,
+  to,
+  logger,
+  options,
```
nathanhammond, line 141: `rm` to force this back into being a utility function.

```suggestion
  logger,
  options,
```

nathanhammond (commented):
Part two.

packages/turbo-workspace-convert/src/install.ts
```diff
@@ -0,0 +1,41 @@
+import execa from "execa";
+import ora from "ora";
+import { satisfies } from "semver";
+import { PACKAGE_MANAGERS, InstallArgs } from "./types";
+
+async function install(args: InstallArgs) {
```
nathanhammond, line 6: I'm tempted to say that this should clean up `node_modules` first for any resolution algorithm path between workspace root and the workspace path. (e.g. `workspaceRoot/node_modules`, `workspaceRoot/packages/node_modules`, `workspaceRoot/packages/a/node_modules`)

Running an install on top of another package manager's install seems like a bad day waiting to happen.

packages/turbo-workspace-convert/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  options,
+}: {
+  root: string;
+  to: PackageManagers;
+  options?: Options;
+}) {
+  const logger = new Logger({ ...options, interactive: false });
+  const workspaceRoot = path.isAbsolute(root)
+    ? root
+    : path.relative(process.cwd(), root);
```
nathanhammond, line 35:
```suggestion
  const workspaceRoot = path.resolve(process.cwd(), root);
```

packages/turbo-workspace-convert/src/getWorkspaceDetails.ts
```diff
@@ -0,0 +1,26 @@
+import managers from "./managers";
+import { Project } from "./types";
+
+async function getWorkspaceDetails({
+  workspaceRoot,
+  fallback = "yarn",
```
nathanhammond, line 6: We should just error here instead of having a default.

If the issue is needing to force it to parse as a particular package manager which is known, but not detectable, that's a different method and we shouldn't tack it on here.

packages/turbo-workspace-convert/src/convert.ts
```diff
@@ -0,0 +1,57 @@
...
+  // create new workspace data
+  await managers[to.name].create({ project, to, logger, options });
+
+  logger.mainStep("Installing dependencies");
+  if (options?.install) {
+    await managers[to.name].convertLock({ project, logger, options });
+    await install({ project, to, logger, options });
+  } else {
+    logger.subStep(
+      chalk.yellow("Skipping install (pass --install to override)")
```
nathanhammond, line 49: Everything else is `--skip-install`, I think we should default in that direction even if it's a bit more-dangerous.

packages/turbo-workspace-convert/src/logger.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  blankLine() {
+    this.logger();
+  }
+
+  error(...args: any[]) {
+    console.error(...args);
+  }
+
+  workspaceSummary({
```
nathanhammond, line 112: I don't love that some of this work is living inside the logger, but that's just an aside.

nathanhammond (commented):
I haven't reviewed `pnpm.ts`, `npm.ts`, and `yarn.ts`. I'll get that on round two.

packages/turbo-workspace-convert/__tests__/managers/pnpm.test.ts
```diff
@@ -0,0 +1,246 @@
...
+      });
+      expect(existsSync(path.join(project.paths.root, "pnpm-lock.yaml"))).toBe(
+        false
+      );
+    });
+  });
+
+  describe("clean", () => {
+    test.each([
+      [true, true],
```
mehulkar, line 220: named keys in an object would be easier to read

packages/turbo-workspace-convert/src/logger.ts
```diff
@@ -0,0 +1,172 @@
...
+    Object.keys(workspacesByDirectory).forEach((directory, idx) => {
+      renderDirectory({
+        number: idx + 1,
+        directory,
+        workspaces: workspacesByDirectory[directory],
+      });
+    });
+    this.blankLine();
+  }
+}
```
mehulkar, line 172: Logger seems like something we want in other CLI tools too. Can we put it into the utils package?

mehulkar (approved):
I don't have any strong objections to this, but couple questions:

- Why are we making this package? I didn't see any chatter about it, so I'm curious. I'm under the impression that switching package managers is pretty easy and in user-land, so unless I'm missing some nuances, do we really want to maintain another 3,000 lines of code?
- Nathan mentioned this inline in one place too, but for the external API, this reads a bit better to me:

    ```bash
    # --target flag instead of --npm, et al
    # convert command instead of part of the package name
    npx @turbo/workspaces convert --target=pnpm
    ```

     Also leaves open the opportunity to do more things with Workspaces in the future, which I can imagine more use cases for.

tknickman:
@mehulkar curious what you would suggest instead of a package? Just keep it in `create-turbo`? Even in that case we'll need to maintain the same code (with the exception of the cli file), so why not do it in its own package where we have a clear separation of concerns, and it's made available as utils and a cli that could be useful for others (and us) as well and act as promo for turbo. There's also a lot in here that we will consume from other packages (for example, codemod can use the package mangaer detection utils, eslint-plugin can use the workspace detect utils as well).

Also, out whole ethos is keeping your monorepo broken into small, distinct parts that are easily testable, buildable, and cacheable.

packages/turbo-workspace-convert/__fixtures__/yarn/basic/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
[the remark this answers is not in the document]
tknickman: This is the real lock file from this test repo believe it or not.

packages/turbo-workspace-convert/__tests__/managers/npm.test.ts
```diff
@@ -0,0 +1,203 @@
...
+        }
+      }
+    );
+  });
+
+  describe("remove", () => {
+    test.each(
+      extendMatrix([
+        ["npm", "8.19.2", undefined],
+        ["yarn", "1.22.19", ["apps/*", "packages/*"]],
```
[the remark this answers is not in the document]
tknickman: The to version is required to determine the install method, not required for this test, but it's part of the API. The expected workspaces are what I'm using to validate if removed worked correctly or not. Could be more clear though.

packages/turbo-workspace-convert/__tests__/managers/npm.test.ts
tknickman: Yea I'm going to clean these up a bit more

packages/turbo-workspace-convert/__tests__/managers/pnpm.test.ts
tknickman: We can, pnpm tends to be it's own special snowflake for a few cases. But I'll look into consolidating

packages/turbo-workspace-convert/__tests__/managers/pnpm.test.ts
tknickman: Agreed. Jest does support something similar to this, but it's pretty buggy from what I found (the name replacement wasn't working - and there's a few issues about it as well)

packages/turbo-workspace-convert/src/install.ts
tknickman: Yep node modules will be gone by the time we get here.

packages/turbo-workspace-convert/src/utils.ts
tknickman: Fixed

packages/turbo-workspace-convert/src/utils.ts
tknickman: You're right. this is bigger than just a util function, I broke it out and cleaned it up

tknickman (commented):
@nathanhammond addressed, going to clean up some of the test matrix generation in a second pass, but everything else should be much cleaner.

packages/turbo-workspace-convert/__fixtures__/npm/basic/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "private": true,
+  "workspaces": [
+    "apps/*",
+    "packages/*"
+  ],
+  "engines": {
+    "node": ">=14.0.0"
+  },
+  "dependencies": {},
+  "packageManager": "npm@8.19.2"
```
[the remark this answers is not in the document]
nathanhammond: The answer is no, I just didn't know as I was reading top to bottom.

packages/turbo-workspaces/__tests__/managers/npm.test.ts
```diff
@@ -0,0 +1,204 @@
...
+        ).toBeUndefined();
+      }
+    );
+  });
+
+  describe("clean", () => {
+    test.each([
+      [true, true],
+      [false, true],
+      [false, true],
```
nathanhammond, line 179: Is there any reason you're running this one twice?

packages/turbo-workspaces/__tests__/managers/pnpm.test.ts
```diff
@@ -0,0 +1,251 @@
...
+        false
+      );
+    });
+  });
+
+  describe("clean", () => {
+    test.each([
+      [true, true],
+      [false, true],
+      [false, true],
```
nathanhammond, line 226: Same question: why is this duplicated?

packages/turbo-workspaces/__tests__/managers/yarn.test.ts
```diff
@@ -0,0 +1,206 @@
...
+        ).toBeUndefined();
+      }
+    );
+  });
+
+  describe("clean", () => {
+    test.each([
+      [true, true],
+      [false, true],
+      [false, true],
```
nathanhammond, line 181: Same comment.

packages/turbo-workspaces/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@turbo/workspaces",
```
nathanhammond, line 2: 👍

packages/turbo-workspaces/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "version": "0.0.0",
+  "description": "Tools for working with package manager workspaces",
+  "homepage": "https://turborepo.org",
```
nathanhammond, line 5:
```suggestion
  "homepage": "https://turbo.build/repo",
```

packages/turbo-workspaces/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "license": "MPL-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/vercel/turborepo",
```
nathanhammond, line 9:
```suggestion
    "url": "https://github.com/vercel/turbo",
```

packages/turbo-workspaces/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "directory": "packages/use-workspaces"
```
nathanhammond, line 10:
```suggestion
    "directory": "packages/turbo-workspaces"
```

packages/turbo-workspaces/package.json
```diff
@@ -0,0 +1,62 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/vercel/turborepo/issues"
```
nathanhammond, line 13:
```suggestion
    "url": "https://github.com/vercel/turbo/issues"
```

packages/turbo-workspaces/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "strip-ansi": "^6.0.1",
+    "ts-jest": "^27.1.1",
+    "tsconfig": "workspace:*",
+    "tsup": "^5.10.3",
+    "turbo-test-utils": "workspace:*",
+    "typescript": "^4.5.5"
+  },
+  "files": [
+    "dist"
+  ]
```
nathanhammond, line 61: I'll allow it. 😜

packages/turbo-workspaces/src/cli.ts
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env node
+
+import chalk from "chalk";
+import { Command } from "commander";
+
+import { summary, convert } from "./commands";
+import cliPkg from "../package.json";
+import { ConvertError } from "./errors";
+
+const codemodCli = new Command();
```
nathanhammond, line 10: Optional, bikeshed:
```suggestion
const workspaceModCli = new Command();
```

packages/turbo-workspaces/README.md
```diff
@@ -0,0 +1,44 @@
+# `@turbo/workspace-convert`
+
+Easily convert your monorepo between package managers. Currently only supports monorepos using either npm, yarn, or pnpm workspaces.
+
+To get started, open a new shell and run:
+
+## CLI
+
+```sh
+npx @turbo/workspace-convert [flags...] [<dir>]
```
nathanhammond, line 10:
```suggestion
npx @turbo/workspaces convert [flags...] [<dir>]
npx @turbo/workspaces summary [flags...] [<dir>]
```

packages/turbo-workspaces/README.md
```diff
@@ -0,0 +1,44 @@
+# `@turbo/workspace-convert`
```
nathanhammond, line 1:
```suggestion
# `@turbo/workspaces`
```

packages/turbo-workspaces/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+Then follow the prompts you see in your terminal.
+
+## Node API
+
+Methods are also available via the Node API:
+
+```js
+import { convertMonorepo, getWorkspaceDetails } from "@turbo/workspace-convert";
```
nathanhammond, line 20:
```suggestion
import { convertMonorepo, getWorkspaceDetails } from "@turbo/workspaces";
```

packages/turbo-workspaces/README.md
```diff
@@ -0,0 +1,44 @@
...
+    to: "pnpm",
+    options: {
+      dry: false,
+      install: true,
+    },
+  });
+}
+```
+
+See [src/index.ts](src/index.ts) for details
```
nathanhammond, line 40: "read the source" isn't documentation. 😜

```suggestion
```

packages/turbo-workspaces/src/cli.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// migrate
+codemodCli
+  .command("convert")
+  .description("Convert project between workspace managers")
+  .argument("[path]", "Project root")
+  .argument("[package-manager]", "Package manager to convert to")
+  .option(
+    "--skip-install",
+    "Install new version of turbo after migration",
```
nathanhammond, line 25:
```suggestion
    "Do not run a package manager install after conversion",
```

nathanhammond (commented):
Part one, 42/61.

packages/turbo-workspaces/src/getWorkspaceDetails.ts
```diff
@@ -0,0 +1,29 @@
...
+    directory: root,
+  });
+  if (!exists) {
+    throw new ConvertError(
+      `Could not find directory at ${workspaceRoot}. Ensure the directory exists.`
+    );
+  }
+
+  for (const { detect, read } of Object.values(managers)) {
+    if (await detect({ workspaceRoot })) {
```
nathanhammond, line 21: I'm not sure that this is identical with what we do elsewhere. (`turbo`, `vc`)

This checks for lockfile, and then `package.json` on a per-package basis. I think we usually check `package.json` as a first pass, and then try to deduce from lockfile presence second in yarn > npm > pnpm order.

Can you confirm? This note is from memory.

packages/turbo-workspaces/src/getWorkspaceDetails.ts
```diff
@@ -0,0 +1,29 @@
...
+      return read({ workspaceRoot });
+    }
+  }
+
+  throw new ConvertError(
+    "Could not determine workspace manager. Ensure a lockfile is present."
```
nathanhammond, line 27:
```suggestion
    "Could not determine workspace manager. Add `packageManager` to `package.json` or ensure a lockfile is present."
```

packages/turbo-workspaces/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+}: {
+  root: string;
+  to: PackageManager;
+  options?: Options;
+}) {
+  const logger = new Logger({ ...options, interactive: false });
+  const [project, availablePackageManagers] = await Promise.all([
+    getWorkspaceDetails({ root }),
+    getAvailablePackageManagers(),
+  ]);
```
nathanhammond, line 20: 🎉

Thanks!

packages/turbo-workspaces/src/install.ts
```diff
@@ -0,0 +1,42 @@
+import execa from "execa";
+import ora from "ora";
+import { satisfies } from "semver";
+import { ConvertError } from "./errors";
+import { PACKAGE_MANAGERS, InstallArgs } from "./types";
+
+async function install(args: InstallArgs) {
+  const { to, logger, options } = args;
+  let packageManager = PACKAGE_MANAGERS[to.name].find((manager) =>
+    satisfies(to.version, manager.semver)
```
nathanhammond, line 10: This kinda stuff is going to be a pain in Rust land. We're going to need `node-compatibile-semver`.

packages/turbo-workspaces/src/install.ts
```diff
@@ -0,0 +1,42 @@
...
+      text: "Installing dependencies...",
+      spinner: {
+        frames: logger.installerFrames(),
+      },
+    }).start();
+
+    try {
+      await execa(`${packageManager.command}`, packageManager.installArgs, {
+        cwd: args.project.paths.root,
+      });
```
nathanhammond, line 31:
```suggestion
      await packageManager.install(args.project.paths.root);
```

packages/turbo-workspaces/src/managers/index.ts
```diff
@@ -0,0 +1,11 @@
+import pnpm from "./pnpm";
+import npm from "./npm";
+import yarn from "./yarn";
+import { ManagerHandler, PackageManager } from "../types";
+
+const MANAGERS: Record<PackageManager, ManagerHandler> = {
+  pnpm,
+  npm,
+  yarn,
```
nathanhammond, line 9: Confirm iteration priority order!

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -0,0 +1,188 @@
...
+  getWorkspacePackageManager,
+} from "../utils";
+
+/*
+ Check if a given project using npm workspaces
+
+ Verify by checking for the existence of:
+  1. package-lock.json
+  2.
+*/
```
nathanhammond, line 28: A few English issues here.

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -0,0 +1,188 @@
...
+  return fs.existsSync(lockFile) || packageManager === "npm";
+}
+
+/*
+ Read workspace data from npm workspaces into generic format
+*/
+async function read(args: ReadArgs): Promise<Project> {
+  const isNpm = await detect(args);
+  if (!isNpm) {
+    throw new ConvertError("Not an npm workspaces project");
```
nathanhammond, line 43:
```suggestion
    throw new ConvertError("Not an npm project");
```

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  const packageJson = getPackageJson(args);
+  return {
+    name: getWorkspaceName(args),
+    packageManager: "npm",
+    paths: {
+      root: args.workspaceRoot,
+      packageJson: path.join(args.workspaceRoot, "package.json"),
+      lockfile: path.join(args.workspaceRoot, "package-lock.json"),
+      nodeModules: path.join(args.workspaceRoot, "node_modules"),
```
nathanhammond, line 54: What about a getter for these dependent properties?

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -0,0 +1,188 @@
...
+
+      // collect all workspace node_modules directories
+      const allModulesDirs = [
+        project.paths.nodeModules,
+        ...project.workspaceData.workspaces.map((w) => w.paths.nodeModules),
+      ];
+      try {
+        logger.subStep(`removing "node_modules"`);
+        await Promise.all(
+          allModulesDirs.map((dir) => fs.rm(dir, { recursive: true }))
```
nathanhammond, line 141: What about `force: true` to avoid the `ENOENT` check?

```suggestion
          allModulesDirs.map((dir) => fs.rm(dir, { recursive: true, force: true }))
```

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -0,0 +1,188 @@
...
+Remove npm workspace data
+
+Removing npm workspaces involves:
+  1. Removing the workspaces field from package.json
+*/
+async function remove(args: RemoveArgs): Promise<void> {
+  const { project, options, to, logger } = args;
+
+  logger.mainStep(`Removing npm workspaces`);
+  if (to.name !== "yarn") {
```
nathanhammond, line 123: How about we _always_ remove, and then _always_ create?

I think that'll be required for dealing with the bonus weird `npm` format anyway.

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -0,0 +1,188 @@
...
+}
+
+async function clean(args: CleanArgs): Promise<void> {
+  const { project, logger, options } = args;
+
+  logger.subStep(
+    `removing ${path.relative(project.paths.root, project.paths.lockfile)}`
+  );
+  if (!options?.dry) {
+    fs.removeSync(project.paths.lockfile);
```
nathanhammond, line 167:
```suggestion
    fs.rmSync(project.paths.lockfile, { force: true });
```

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+}
+
+// converts existing, non npm lockfile to a npm lockfile
```
nathanhammond, line 171: "converts"

So far as I know only pnpm supports conversion.

packages/turbo-workspaces/src/managers/pnpm.ts
```diff
@@ -0,0 +1,199 @@
...
+    fs.existsSync(workspaceFile) ||
+    packageManager === "pnpm"
+  );
+}
+
+// read workspace data from pnpm workspaces into generic format
+async function read(args: ReadArgs): Promise<Project> {
+  const isPnpm = await detect(args);
+  if (!isPnpm) {
+    throw new ConvertError("Not a pnpm workspaces project");
```
nathanhammond, line 42:
```suggestion
    throw new ConvertError("Not a pnpm project");
```

packages/turbo-workspaces/src/managers/pnpm.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  return {
+    name: getWorkspaceName(args),
+    packageManager: "pnpm",
+    paths: {
+      root: args.workspaceRoot,
+      packageJson: path.join(args.workspaceRoot, "package.json"),
+      lockfile: path.join(args.workspaceRoot, "pnpm-lock.yaml"),
+      workspaceConfig: path.join(args.workspaceRoot, "pnpm-workspace.yaml"),
+      nodeModules: path.join(args.workspaceRoot, "node_modules"),
```
nathanhammond, line 53: Getters?

packages/turbo-workspaces/src/updateDependencies.ts
```diff
@@ -0,0 +1,127 @@
...
+      dependencyList[name] =
+        to.name === "pnpm" ? `workspace:${version}` : version;
+      updated.push(name);
+    }
+  });
+
+  return { dependencyList, updated };
+}
+
+export default function updateDependencies({
```
nathanhammond, line 41: 🤨 Don't love this method but not a blocker. Feels like it should be per-package-manager, with a default implementation.

packages/turbo-workspaces/src/managers/pnpm.ts
```diff
@@ -0,0 +1,199 @@
...
+    // collect all workspace node_modules directories
+    const allModulesDirs = [
+      project.paths.nodeModules,
+      ...project.workspaceData.workspaces.map((w) => w.paths.nodeModules),
+    ];
+
+    try {
+      logger.subStep(`removing "node_modules"`);
+      await Promise.all(
+        allModulesDirs.map((dir) => fs.rm(dir, { recursive: true }))
```
nathanhammond, line 138:
```suggestion
        allModulesDirs.map((dir) => fs.rm(dir, { recursive: true, force: true }))
```

packages/turbo-workspaces/src/managers/pnpm.ts
```diff
@@ -0,0 +1,199 @@
...
+    );
+    if (!options?.dry && fs.existsSync(project.paths.lockfile)) {
+      try {
+        await execa("pnpm", ["import"], {
+          stdio: "ignore",
+          cwd: project.paths.root,
+        });
+      } catch (err) {
+        console.error(project.paths.lockfile, err);
+        fs.removeSync(project.paths.lockfile);
```
nathanhammond, line 184:
```suggestion
        fs.removeSync(project.paths.lockfile, {force: true});
```

packages/turbo-workspaces/src/managers/pnpm.ts
```diff
@@ -0,0 +1,199 @@
...
+        project.paths.lockfile
+      )} to pnpm-lock.yaml`
+    );
+    if (!options?.dry && fs.existsSync(project.paths.lockfile)) {
+      try {
+        await execa("pnpm", ["import"], {
+          stdio: "ignore",
+          cwd: project.paths.root,
+        });
+      } catch (err) {
```
nathanhammond, line 182: Nit: `err` and `e` used interchangeably. global replace pretty please?

```suggestion
      } catch (err) {
```

packages/turbo-workspaces/src/managers/yarn.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  const packageJson = getPackageJson(args);
+  return {
+    name: getWorkspaceName(args),
+    packageManager: "yarn",
+    paths: {
+      root: args.workspaceRoot,
+      packageJson: path.join(args.workspaceRoot, "package.json"),
+      lockfile: path.join(args.workspaceRoot, "yarn.lock"),
+      nodeModules: path.join(args.workspaceRoot, "node_modules"),
```
nathanhammond, line 49: getters?

packages/turbo-workspaces/src/managers/yarn.ts
```diff
@@ -0,0 +1,198 @@
...
+  2. Setting the packageManager field in package.json
+*/
+async function create(args: CreateArgs): Promise<void> {
+  const { project, to, logger, options } = args;
+
+  logger.mainStep(`Creating yarn workspaces`);
+  const packageJson = getPackageJson({ workspaceRoot: project.paths.root });
+  logger.rootHeader();
+  // workspaces
+  if (project.packageManager !== "npm") {
```
nathanhammond, line 75: Always create?

packages/turbo-workspaces/src/managers/yarn.ts
```diff
@@ -0,0 +1,198 @@
...
+    )}`
+  );
+  packageJson.packageManager = `${to.name}@${to.version}`;
+
+  if (!options?.dry) {
+    fs.writeJSONSync(project.paths.packageJson, packageJson, { spaces: 2 });
+  }
+
+  // if we're converting from pnpm, we need to update the workspace package.json files
+  if (project.packageManager === "pnpm") {
```
nathanhammond, line 99:
```suggestion
```

The downstream function should be able to handle this.

packages/turbo-workspaces/src/managers/yarn.ts
```diff
@@ -0,0 +1,198 @@
...
+      // collect all workspace node_modules directories
+      const allModulesDirs = [
+        project.paths.nodeModules,
+        ...project.workspaceData.workspaces.map((w) => w.paths.nodeModules),
+      ];
+
+      try {
+        logger.subStep(`removing "node_modules"`);
+        await Promise.all(
+          allModulesDirs.map((dir) => fs.rm(dir, { recursive: true }))
```
nathanhammond, line 145:
```suggestion
          allModulesDirs.map((dir) => fs.rm(dir, { recursive: true, force: true }))
```

packages/turbo-workspaces/src/updateDependencies.ts
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  const allDependencyKeys: Array<keyof PackageJsonDependencies> = [
+    "dependencies",
+    "devDependencies",
+    "peerDependencies",
+    "optionalDependencies",
+  ];
+
+  allDependencyKeys.forEach((depKey) => {
```
nathanhammond, line 74: This cleaned up _nicely_.

packages/turbo-workspaces/src/updateDependencies.ts
```diff
@@ -0,0 +1,127 @@
...
+  const toLog = (key: keyof PackageJsonDependencies) => {
+    const total = stats[key].length;
+    if (total > 0) {
+      return `${chalk.green(total.toString())} ${key}`;
+    }
+    return undefined;
+  };
+
+  const allChanges = allDependencyKeys.map(toLog).filter(Boolean);
+  const workspaceLocation = `./${path.relative(
```
nathanhammond, line 97:
```suggestion
  const workspaceLocation = `${path.relative(
```

We allow passing `dir` so it isn't `./`.

packages/turbo-workspaces/src/utils.ts
```diff
@@ -0,0 +1,119 @@
+import fs from "fs-extra";
+import path from "path";
+import glob from "fast-glob";
+import yaml from "js-yaml";
+import semver from "semver";
+import { PackageJson, Workspace } from "./types";
+import { ConvertError } from "./errors";
+
+// adapted from https://github.com/nodejs/corepack/blob/main/sources/specUtils.ts#L14
```
nathanhammond, line 9: Use `y` to get a permalink.
https://github.com/nodejs/corepack/blob/cae770694e62f15fed33dd8023649d77d96023c1/sources/specUtils.ts#L14

packages/turbo-workspaces/src/utils.ts
```diff
@@ -0,0 +1,119 @@
...
+function getPackageJson({
+  workspaceRoot,
+}: {
+  workspaceRoot: string;
+}): PackageJson {
+  const packageJsonPath = path.join(workspaceRoot, "package.json");
+  try {
+    return fs.readJsonSync(packageJsonPath, "utf8");
+  } catch (_) {
+    throw new ConvertError(`no "package.json" found at ${workspaceRoot}`);
```
nathanhammond, line 21: This error can be either `ENOENT` or `JSON.parse`.

packages/turbo-workspaces/src/utils.ts
```diff
@@ -0,0 +1,119 @@
...
+  workspaceRoot,
+}: {
+  workspaceRoot: string;
+}): string | undefined {
+  const { packageManager } = getPackageJson({ workspaceRoot });
+  if (packageManager) {
+    const match = packageManager.match(PACKAGE_MANAGER_REGEX);
+    if (match && match.length === 3 && semver.valid(match[2])) {
+      return match[1];
+    }
```
nathanhammond, line 35: 6 is afraid of 7 because 7 8 9.

Can we do `const [major, minor, patch] = match;` or something to that effect?

packages/turbo-workspaces/src/utils.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function getPnpmWorkspaces({
+  workspaceRoot,
+}: {
+  workspaceRoot: string;
+}): Array<string> {
+  const workspaceFile = path.join(workspaceRoot, "pnpm-workspace.yaml");
+  if (fs.existsSync(workspaceFile)) {
+    const workspaceConfig = yaml.load(fs.readFileSync(workspaceFile, "utf8"));
```
nathanhammond, line 60: This can `throw`, wrap it all in `try` and ditch the `existsSync`?

packages/turbo-workspaces/src/utils.ts
```diff
@@ -0,0 +1,119 @@
...
+      Array.isArray(workspaceConfig.packages)
+    ) {
+      return workspaceConfig.packages as Array<string>;
+    }
+  }
+
+  return [];
+}
+
+function expandWorkspaces({
```
nathanhammond, line 74: Troll: should be per package manager. 😜

(This is fine, but the API difference between the glob libs is basically zero.)

packages/turbo-codemod/src/commands/migrate/steps/getTurboUpgradeCommand.ts
```diff
@@ -14,9 +14,9 @@ type InstallType = "dependencies" | "devDependencies";
 function getGlobalBinaryPaths(): Record<PackageManager, string | undefined> {
   return {
     // we run these from the homedir to avoid corepack interference
```
arlyon, line 16: This comment should be updated

packages/turbo-workspaces/__tests__/managers/npm.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      await npm.clean({
+        project,
+        logger: new Logger({ interactive, dry }),
+        options: {
+          interactive,
+          dry,
+        },
+      });
+
+      if (dry) {
```
arlyon, line 197: Shall we just collapse this if block

packages/turbo-workspaces/__tests__/managers/yarn.test.ts
```diff
@@ -0,0 +1,206 @@
...
+          project,
+          to: { name: to, version },
+          logger: new Logger({ interactive, dry }),
+          options: {
+            interactive,
+            dry,
+          },
+        });
+
+        if (dry) {
```
arlyon, line 67: Similar here: .toEqual(dry ? ... : ...)

packages/turbo-workspaces/__tests__/managers/npm.test.ts
tknickman: Good call 👍🏼

packages/turbo-workspaces/src/utils.ts
tknickman: 😄 this is another "wait for a bug report before investing the time here" imo

packages/turbo-workspaces/src/utils.ts
tknickman: TIL

packages/turbo-workspaces/README.md
tknickman: 😄 fair.

packages/turbo-workspaces/README.md
tknickman: This was a lazy rename I will admint

packages/turbo-workspaces/__tests__/managers/npm.test.ts
tknickman: Nope, swapped to deriving these.

packages/turbo-workspaces/src/cli.ts
tknickman: Good catch, updates

packages/turbo-workspaces/src/commands/summary/types.ts
```diff
@@ -0,0 +1,4 @@
+export type SummaryCommandArgument = string;
+export type SummaryCommandOptions = {
+  json: boolean;
```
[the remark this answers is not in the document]
tknickman: This was a placeholder for something I ended up not adding, removing for now until/if we need it later

packages/turbo-workspaces/src/install.ts
tknickman: Yea, agreed

packages/turbo-workspaces/src/managers/npm.ts
tknickman: Correct

packages/turbo-workspaces/src/updateDependencies.ts
tknickman: it is, I'm explicitly making this relative to root to avoid really long paths in the output.

packages/turbo-workspaces/src/updateDependencies.ts
tknickman: Yea, this really needed a second pass.

packages/turbo-workspaces/src/managers/npm.ts
tknickman: I did think about that, but my thinking here was that I wanted to try and touch as little as possible on the way from a to b. 

However, it does make these functions a little more gross since each package manager needs to care about more than just itself. 

I'll play with this and see how it feels.

packages/turbo-workspaces/src/utils.ts
tknickman: I didn't catch this because I wanted this to bubble to root. If we can't read this, that's a blocker

packages/turbo-workspaces/src/getWorkspaceDetails.ts
tknickman: Yep you're right, we used to do the opposite in turbo but swapped once we started to enforce the package manager value in 1.1, I'll invert this

packages/turbo-workspaces/src/install.ts
tknickman: Will address in a followup

tknickman:
Updated, shipping this as first version and will address further feedback in followups

packages/turbo-workspaces/src/utils.ts
nathanhammond: destructuring > what the hell index is this? 😜

packages/turbo-workspaces/src/utils.ts
nathanhammond: Yeah, they're ~compatible anyway.

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -138,42 +134,47 @@ async function remove(args: RemoveArgs): Promise<void> {
...
+ */
 async function convertLock(args: ConvertArgs): Promise<void> {
-  const { project } = args;
+  const { project, options } = args;
 
   // remove the lockfile
-  fs.removeSync(project.paths.lockfile);
+  if (!options?.dry) {
+    fs.rmSync(project.paths.lockfile, { force: true });
+  }
```
nathanhammond, line 177: Don't know who caught that one, but good catch.

PR #3819, opened 2023-02-15
Nightly releases: generate basic release body

jridgewell:
The nightly releases are using the last commit message for the release body, which is a bit weird. This generates a "compare at https://github.com/vercel/turbo/compare/turbopack-230214.2...turbopack-230215.1" listing the commits between the last and new tag.

### DIFF
--- a/.github/actions/turbopack-bump/src/index.ts
+++ b/.github/actions/turbopack-bump/src/index.ts
@@ -51,6 +51,12 @@ async function run() {
 
   core.setOutput("new_tag", nextTag);
   await createTag(octokit, nextTag, commitSha);
+
+  // TODO: generate real release notes
+  core.setOutput(
+    "changelog",
+    `See the commit diff at https://github.com/vercel/turbo/compare/${lastTag.name}...${nextTag}`
+  );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


.github/actions/turbopack-bump/src/index.ts
```diff
@@ -51,6 +51,12 @@ async function run() {
 
   core.setOutput("new_tag", nextTag);
   await createTag(octokit, nextTag, commitSha);
+
+  // TODO: generate real release notes
```
tknickman, line 55: You could use the `release.yml` ([docs](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes)) if you'd like. Currently we're using that for Turborepo, but we're generating them via an API call on NPM publish of a new turborepo package - and the API supports specifying where the location of the release config ([docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#generate-release-notes-content-for-a-release)).

So if you'd prefer, ya'll can use `.github/release.yml`, and we can create another one (`.github/turborepo-release-yml`)

tknickman (approved):

.github/actions/turbopack-bump/src/index.ts
jridgewell: Oh that's great! I'll look into using this once I finish my current tasks and get back to the release work. Thanks!

PR #3821, opened 2023-02-15
feat(types): update turbo.json schema

tknickman:
Updated schema to support extends

### DIFF
--- a/packages/turbo-types/src/types/config.ts
+++ b/packages/turbo-types/src/types/config.ts
@@ -1,8 +1,36 @@
 /* This file generates the `schema.json` file. */
-export interface Schema {
+export type Schema = RootSchema | WorkspaceSchema;
+
+export interface BaseSchema {
   /** @default https://turbo.build/schema.json */
   $schema?: string;
+  pipeline: {
+    /**
+     * The name of a task that can be executed by turbo. If turbo finds a workspace
+     * package with a package.json scripts object with a matching key, it will apply the
+     * pipeline task configuration to that npm script during execution.
+     */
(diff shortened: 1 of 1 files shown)


mehulkar (commented):
Does this work? 

Looking at the [generated `/schema.json`][1], and plugging into [a validator][2], it doesn't seem to allow the `extends` key.

[1]: https://turbo-site-git-tomknickman-turbo-782-add-new-schemajson-5fbcb2.vercel.sh/schema.json
[2]: https://www.jsonschemavalidator.net/

Although I'm not sure if the `/schema.json` on the preview link is actually updated or not

mehulkar:
Looks like the preview `/schema.json` wasn't updated. Generated locally and tried it out and it works!

mehulkar (approved):
The error messaging isn't great with this approach, but I think that's ok.

tknickman:
Yea, the error messaging is the one issue with keeping it all in one schema. It ends up giving you a weird error about one of the schema specific keys being an extra key, and then an opaque error that it doesn't match AnyOf the options. 

But I believe it's the best we can do with json.schemas

mehulkar:
> But I believe it's the best we can do with json.schemas

I have a hunch that `if`, `else` can do it. Something like:

```
{
    "dependencies": {
        "extends": {
        "not": {
            "required": ["globalEnv", "globalDependencies", "remoteCache"]
        }
        }
    },

    // or
    "if": {},
    "then": {},
    "else": {}
}
````

but i wasn't able to figure it out very quickly yesterday in my first go

PR #3823, opened 2023-02-15
loaders: use pre-compiled loader-runner distributed with Next.js

wbinnssmith:
The benefit of this depends on https://github.com/vercel/next.js/pull/45962, but it remains compatible with apps including `loader-runner`.

This first attempts to require `loader-runner` from the app's installed version of Next.js, falling back to requiring the package directly. We should probably eventually remove this fallback once all compatible versions of Next.js include the precompiled version, as that has a more predictable version of the package.

Test Plan: Linked a local copy of Next.js including https://github.com/vercel/next.js/pull/45962 to an app without `loader-runner` that uses loaders and verified loaders ran.


crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -18,9 +18,15 @@ type LoaderConfig =
       options: { [k: string]: unknown };
     };
 
-const { runLoaders } = __turbopack_external_require__(
-  "loader-runner"
-) as typeof import("loader-runner");
+let runLoaders: typeof import("loader-runner");
+try {
+  ({ runLoaders } = __turbopack_external_require__(
+    "next/dist/compiled/loader-runner"
```
sokra, line 24: It's a bit weird to have a next.js specific import inside of `turbopack-node`...
I imagine the webpack loader executor could be general purpose.

Take a looks how the postcss transform is implemented. It also uses the next.js packaged version postcss, but that import lives on the next.js side

sokra (changes requested):

crates/next-core/src/next_build.rs
```diff
@@ -19,3 +20,17 @@ pub async fn get_postcss_package_mapping(
     ])
     .cell())
 }
+
+#[turbo_tasks::function]
+pub async fn get_external_next_compiled_package_mapping(
+    package_name: StringVc,
```
sokra, line 26:
```suggestion
    package_name: &str,
```

sokra (approved):

PR #3824, opened 2023-02-15
port(turborepo): Unlink

NicholasLYang:


crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -51,6 +52,18 @@ impl CommandBase {
         Ok(())
     }
 
+    // NOTE: This deletes the repo config file. It does *not* remove the
+    // `RepoConfig` struct from `CommandBase`. This is fine because we
+    // currently do not have any commands that delete the repo config file
+    // and then attempt to read from it.
+    pub fn delete_repo_config(&mut self) -> Result<()> {
```
chris-olszewski, line 59: Since we're just deleting the file I think being explicit about that in the name would be good. (Is there a reason we aren't clearing the config?)
```suggestion
    pub fn delete_repo_config_file(&self) -> Result<()> {
```

crates/turborepo-lib/src/cli.rs
```diff
@@ -477,10 +477,23 @@ pub async fn run(repo_state: Option<RepoState>) -> Result<Payload> {
...
+        Command::Unlink { .. } => {
+            if clap_args.test_run {
+                println!("Unlink test run successful");
+                return Ok(Payload::Rust(Ok(0)));
+            }
+
+            let mut base = CommandBase::new(clap_args, repo_root)?;
+
+            if let Err(err) = unlink::unlink(&mut base) {
+                error!("error: {}", err.to_string())
```
chris-olszewski, line 489: Should we be returning a non-zero exit code here?

chris-olszewski (approved):
Possible exit code changes on error is the only thing I'm concerned about

crates/turborepo-lib/src/commands/mod.rs
NicholasLYang: Basically we'd need to turn the `OnceCell` into a `RefCell` because we want the ability to borrow it and replace the value, while still providing an immutable API.

PR #3826, opened 2023-02-15
feature: Added Vercel API error handling

NicholasLYang:


mehulkar (approved):
I'm learning Rust, but I don't see anything glaring here!

crates/turborepo-lib/src/client.rs
```diff
@@ -30,8 +36,10 @@ pub enum CachingStatus {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct CachingStatusResponse {
-    pub status: CachingStatus,
+#[serde(rename_all = "lowercase")]
+pub enum CachingStatusResponse {
```
chris-olszewski, line 40: All of these `_Response` enums feel like they should be `Result<CachingStatus, APIError>`. Could write our own version of [`Response::error_for_status`](https://docs.rs/reqwest/latest/src/reqwest/async_impl/response.rs.html#341-348) that checks that status code and if it's an error we deserialize to the `APIError` message?

NicholasLYang: Yeah sure. It's a little frustrating because the field for the valid path isn't a consistent one like `{ "ok": {...} }` but varies depending on the resource. Hence the redundant enums

NicholasLYang: Do you think there's a way we can do this generically? I could make a macro ofc but I don't think that's worth it.

NicholasLYang:
Closed in favor of #3898

PR #3828, opened 2023-02-15
Error if none of the tasks passed to CLI were found

mehulkar:
Previously, a `validate` method would assert that tasks passed to the CLI exist. After switching to Lazy Loading Tasks and changing how errors bubble, I had removed this validation behavior. This PR adds it back.

### DIFF
--- a/cli/integration_tests/basic_monorepo/dry_run.t
+++ b/cli/integration_tests/basic_monorepo/dry_run.t
@@ -104,10 +104,6 @@ Check my-app#build output
 
 $ Non-existent tasks don't throw an error
   $ ${TURBO} run doesnotexist --dry=json
-  {
-    "packages": [
-      "my-app",
-      "util"
-    ],
-    "tasks": []
-  }
+   ERROR  run failed: error preparing engine: Could not find "doesnotexist" in project
+  Turbo error: error preparing engine: Could not find "doesnotexist" in project
+  [1]
--- a/cli/integration_tests/basic_monorepo/dry_run.t
+++ b/cli/integration_tests/basic_monorepo/dry_run.t
@@ -102,8 +102,8 @@ Check my-app#build output
     }
   }
 
-$ Non-existent tasks don't throw an error
+Tasks that don't exist throw an error
   $ ${TURBO} run doesnotexist --dry=json
-   ERROR  run failed: error preparing engine: Could not find "doesnotexist" in project
-  Turbo error: error preparing engine: Could not find "doesnotexist" in project
+   ERROR  run failed: error preparing engine: Could not find the following tasks in project: doesnotexist
+  Turbo error: error preparing engine: Could not find the following tasks in project: doesnotexist
   [1]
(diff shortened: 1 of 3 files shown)


cli/internal/core/engine.go
```diff
@@ -180,14 +186,29 @@ func (e *Engine) Prepare(options *EngineBuildingOptions) error {
...
+		missingList = append(missingList, t)
+	}
+	sort.Strings(missingList)
+
+	if len(missingList) > 0 {
+		if len(missingList) == 1 {
+			return fmt.Errorf("Could not find \"%s\" in project", missingList[0])
+		}
+		return fmt.Errorf("Could not find the following tasks in project: %s", strings.Join(missingList, ", "))
+	}
```
nathanhammond, line 210: I appreciate the intent here but I'd prefer that the error message be more-easily searchable as opposed to being 100% grammatically correct.

```suggestion
	if len(missingList) > 0 {
		return fmt.Errorf("Could not find the following tasks in project: %s", strings.Join(missingList, ", "))
	}
```

cli/internal/core/engine.go
```diff
@@ -158,6 +158,12 @@ func (e *Engine) Prepare(options *EngineBuildingOptions) error {
 
 	traversalQueue := []string{}
 
+	// make a map of taskNames passed in
+	missing := map[string]bool{}
+	for _, taskName := range taskNames {
+		missing[taskName] = true
+	}
```
nathanhammond, line 165: Let's use our Set implementation rather than rolling one inline.

```suggestion
	missing := util.SetFromStrings(taskNames)
```

cli/internal/core/engine.go
```diff
@@ -180,14 +186,29 @@ func (e *Engine) Prepare(options *EngineBuildingOptions) error {
 
 					return err
 				}
-
+				// delete Taskanem if it was found
+				delete(missing, taskName)
```
nathanhammond, line 190:
```suggestion
				missing.Delete(taskName)
```

cli/internal/core/engine.go
```diff
@@ -180,14 +186,29 @@ func (e *Engine) Prepare(options *EngineBuildingOptions) error {
...
 		}
 	}
 
 	visited := make(util.Set)
 
+	// validate that all tasks passed were found
+	missingList := []string{}
+	for t := range missing {
+		missingList = append(missingList, t)
+	}
```
nathanhammond, line 202:
```suggestion
	missingList := missing.UnsafeListOfStrings()
```

nathanhammond (approved):
🛠🚢

cli/integration_tests/basic_monorepo/dry_run.t
```diff
@@ -104,10 +104,6 @@ Check my-app#build output
 
 $ Non-existent tasks don't throw an error
```
gsoltis, line 105: I think this comment is wrong now

gsoltis (approved):

PR #3832, opened 2023-02-16
Fix routing cases

jridgewell:
This fixes 2 bugs that I've found:
1. If the `middleware.ts` file doesn't exist, then we get an unhelpful `unable to resolve relative "."` error during `next-edge` transition processing
   - This is fixed by not processing the `VirtualAssetVc` with the edge transition
2. If you're using an older Next version, then the router doesn't have a `type` field
   - Because the `type` field is empty, we hit neither the `rewrite` nor `none` cases and fall through to the middleware case, returning an empty response with no body.
   - This is fixed by treating `{ url: string, headers: … }` as `{ type: 'rewrite', url: string, headers: … }`

### DIFF
--- a/crates/next-core/js/src/entry/router.ts
+++ b/crates/next-core/js/src/entry/router.ts
@@ -156,17 +156,18 @@ async function handleClientResponse(
     const data = JSON.parse(buffer) as RouteResult;
 
     switch (data.type) {
+      case "none":
+        return {
+          type: "none",
+        };
       case "rewrite":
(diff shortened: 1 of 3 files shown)


sokra (approved):

sokra:
> 2\. If you're using an older Next version, then the router doesn't have a `type` field

That's can't really happen in the normal flow as turbopack is published with next.js, but fine fixing it anyway

PR #3834, opened 2023-02-16
Move workspace-specific tasks to workspace configs

mehulkar:
Trying out Workspace Configs in our own repo. Some things I tested:

```
pnpm run turbo run e2e
pnpm run turbo run e2e-prebuild
pnpm run turbo run integration-tests
pnpm run turbo run docs
```

### DIFF
--- a/cli/turbo.json
+++ b/cli/turbo.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "../docs/public/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "build": {
+      "outputs": ["turbo", "turbo.exe"]
+    },
+
+    "e2e": {
+      "outputs": [],
+      "inputs": [
+        "cli/**/*.go",
+        "cli/go.mod",
+        "cli/go.sum",
+        "cli/scripts/e2e/e2e.ts",
+        "shim/**/*.rs"
+      ]
+    },
+    "e2e-prebuilt": {
+      "inputs": [
+        "cli/**/*.go",
+        "cli/go.mod",
+        "cli/go.sum",
+        "cli/scripts/e2e/e2e.ts",
+        "shim/**/*.rs"
+      ]
+    },
+    "integration-tests": {}
+  }
+}
--- a/docs/turbo.json
+++ b/docs/turbo.json
@@ -1,0 +1,28 @@
+{
+  "$schema": "./public/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "build": {
+      "env": [
+        "SENTRY_DSN",
+        "TURBOREPO_SFDC_CAMPAIGN_ID",
+        "TRAY_URL",
+        "CONVERTKIT_API_KEY",
+        "CONVERTKIT_API_SECRET"
+      ],
+      "outputs": [".next/**"],
+      "dependsOn": ["^build", "docs#rss", "docs#schema"]
+    },
+
+    "dev": {
+      "dependsOn": ["schema"]
+    },
+    "schema": {
+      "outputs": ["public/schema.json"]
+    },
+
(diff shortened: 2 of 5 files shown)


docs/turbo.json
```diff
@@ -0,0 +1,28 @@
...
+    },
+
+    "dev": {
+      "dependsOn": ["schema"]
+    },
+    "schema": {
+      "outputs": ["public/schema.json"]
+    },
+
+    "rss": {
```
mehulkar, line 24: This doesn't seem to be hooked up to anything, although a task exists. when does it run?

cli/turbo.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "../docs/public/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "build": {
+      "outputs": ["turbo", "turbo.exe"]
+    },
+
```
nathanhammond, line 8:
```suggestion
```
This is definitely not going to do what we want. We should probably pop this out because it doesn't work like this.

packages/create-turbo/turbo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "../../docs/public/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "test": {
+      "dependsOn": ["build"]
+    }
```
nathanhammond, line 7: We don't have a `build` task available here.

packages/eslint-plugin-turbo/turbo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "../../docs/public/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "test": {
+      "dependsOn": ["build"]
```
nathanhammond, line 6: Same, there's no build task in the pipeline.

nathanhammond (approved):
This is a fine incremental step, but we're not actively using much of these things. We need to move a lot more stuff before this works.

cli/turbo.json
mehulkar: it _should_ do the same thing as it did before this change. are you saying it's already broken in main or because of this change?

packages/create-turbo/turbo.json
mehulkar: It should inherit the `build` task from root config! this is additive. This is the test that's intended to prove this case works: https://github.com/vercel/turbo/blob/main/cli/integration_tests/composable_config/composing-add-tasks.t

packages/eslint-plugin-turbo/turbo.json
mehulkar: same comment as previous! it should work

mehulkar:
> This is a fine incremental step, but we're not actively using much of these things. We need to move a lot more stuff before this works.

Can you say more about what you mean? I can understand that we're not actively using these tasks much, but what do you mean by "we need to move a lot more stuff before this works?

packages/create-turbo/turbo.json
mehulkar: <details>
  <summary>pnpm run turb run test --filter=create-turbo --graph</summary>

```
digraph {
	compound = "true"
	newrank = "true"
	subgraph "root" {
		"[root] ___ROOT___#build" -> "[root] ___ROOT___"
		"[root] create-turbo#build" -> "[root] tsconfig#build"
		"[root] create-turbo#test" -> "[root] create-turbo#build"
		"[root] create-turbo#test" -> "[root] tsconfig#build"
		"[root] tsconfig#build" -> "[root] ___ROOT___#build"
	}
}
```

</details>

tknickman (approved):

PR #3836, opened 2023-02-16
Make `chunk_content_internal` parallel

alexkirsz:
This switches the `chunk_content_internal` function from a sequential BFS to a parallel BFS (+ reverse topological sort at the end).

I expected this to make some difference in performance, as traversing references in parallel can lead to better CPU usage (see #3771), but in practice our benchmarks show no significant difference.

Real apps might be a different story, but I didn't notice any particular performance improvement on vercel.com either.

This implementation is not perfect (we're making more calls to `get_children` than strictly necessary), but I think it's enough to measure a potential performance improvement.

Marking this as a draft for now as it's more complicated than the current implementation and there's no clear win to adopting this.

### DIFF
--- a/crates/turbo-tasks/src/graph/get_children.rs
+++ b/crates/turbo-tasks/src/graph/get_children.rs
@@ -2,12 +2,14 @@ use std::{collections::HashSet, future::Future};
 
 use anyhow::Result;
 
+use super::GraphTraversalControlFlow;
+
 /// A trait that allows a graph traversal to get the children of a node.
-pub trait GetChildren<T, A = ()> {
+pub trait GetChildren<T, A = !, Impl = ()> {
     type Children: IntoIterator<Item = T>;
     type Future: Future<Output = Result<Self::Children>>;
 
-    fn get_children(&mut self, item: &T) -> Option<Self::Future>;
+    fn get_children(&mut self, item: &T) -> GraphTraversalControlFlow<Self::Future, A>;
 }
 
 // The different `Impl*` here are necessary in order to avoid the `Conflicting
@@ -17,7 +19,7 @@ pub trait GetChildren<T, A = ()> {
 
 pub struct ImplRef;
 
-impl<T, C, F, CI> GetChildren<T, ImplRef> for C
+impl<T, C, F, CI> GetChildren<T, !, ImplRef> for C
 where
     C: FnMut(&T) -> F,
     F: Future<Output = Result<CI>>,
@@ -26,14 +28,14 @@ where
     type Children = CI;
     type Future = F;
 
-    fn get_children(&mut self, item: &T) -> Option<Self::Future> {
-        Some((self)(item))
+    fn get_children(&mut self, item: &T) -> GraphTraversalControlFlow<Self::Future> {
+        GraphTraversalControlFlow::Continue((self)(item))
     }
 }
 
 pub struct ImplRefOption;
 
-impl<T, C, F, CI> GetChildren<T, ImplRefOption> for C
+impl<T, C, F, CI> GetChildren<T, !, ImplRefOption> for C
 where
     C: FnMut(&T) -> Option<F>,
     F: Future<Output = Result<CI>>,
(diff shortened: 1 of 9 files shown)


crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -362,194 +367,305 @@ pub trait FromChunkableAsset: ChunkItem + Sized + Debug {
...
+    I: FromChunkableAsset + Eq + std::hash::Hash + Clone,
+{
+    let mut processed_assets: HashSet<(ChunkingType, AssetVc)> = HashSet::default();
+
+    // TODO(alexkirsz) We shouldn't need a clone here.
+    let additional_entries = if let Some(additional_entries) = additional_entries {
+        additional_entries.await?.clone_value().into_iter()
+    } else {
+        vec![].into_iter()
+    };
```
sokra, line 587: It should work this way:

```suggestion
    let additional_entries = if let Some(additional_entries) = additional_entries {
        Some(additional_entries.await?)
    } else {
        None
    };
    let additional_entries = additional_entries.into_iter().map(|list| list.iter()).flatten();
```

The first declarations stores the owned thing. The second can access it by reference.
Also [`Option::into_iter()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.iter) is useful here.

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -362,194 +367,305 @@ pub trait FromChunkableAsset: ChunkItem + Sized + Debug {
...
+
+    let entries = [entry].into_iter().chain(additional_entries);
+
+    // TODO(alexkirsz) We shouldn't need a clone here.
+    for entry in entries.clone() {
```
sokra, line 592:
```suggestion
    let entries = once(&entry).chain(additional_entries);

    // cloning a Iterator over &AssetVc is cheap
    for entry in entries.clone() {
```

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -362,194 +367,305 @@ pub trait FromChunkableAsset: ChunkItem + Sized + Debug {
...
+
+    let context = ChunkContentContext {
+        chunking_context,
+        entry,
+        split,
+    };
+    let state = Arc::new(Mutex::new(ChunkContentState {
+        processed_assets,
+        chunk_items_count: 0,
+    }));
```
sokra, line 615: Technically this should be doable without a Mutex (that's a bit expensive), by checking these constraints after handing a Future from FuturesUnordered. I guess this should be part of the trait.

crates/turbo-tasks/src/graph/get_children.rs
```diff
@@ -2,12 +2,14 @@ use std::{collections::HashSet, future::Future};
...
+use super::GraphTraversalControlFlow;
+
 /// A trait that allows a graph traversal to get the children of a node.
-pub trait GetChildren<T, A = ()> {
+pub trait GetChildren<T, A = !, Impl = ()> {
     type Children: IntoIterator<Item = T>;
     type Future: Future<Output = Result<Self::Children>>;
 
-    fn get_children(&mut self, item: &T) -> Option<Self::Future>;
+    fn get_children(&mut self, item: &T) -> GraphTraversalControlFlow<Self::Future, A>;
```
sokra, line 12:
``` rs
fn join_children(&mut self, children: Children) -> GraphTraversalControlFlow<Children, A>
```

This could be called after the future completes and since it `&mut self` it can do some counting without a Mutex

crates/turbopack-core/src/chunk/mod.rs
alexkirsz: Doing this (and the below) results in the following error (which actually gets reported in turbopack-ecmascript--ugh):
```
error: implementation of `Iterator` is not general enough
   --> crates/turbopack-ecmascript/src/chunk/mod.rs:332:1
    |
332 | #[turbo_tasks::function]
    | ^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `Iterator` is not general enough
    |
    = note: `Iterator` would have to be implemented for the type `std::option::Iter<'0, ReadRef<Assets, Vec<AssetVc>>>`, for any lifetime `'0`...
    = note: ...but `Iterator` is actually implemented for the type `std::option::Iter<'1, ReadRef<Assets, Vec<AssetVc>>>`, for some specific lifetime `'1`
    = note: this error originates in the attribute macro `turbo_tasks::function` (in Nightly builds, run with -Z macro-backtrace for more info)

error: implementation of `FnOnce` is not general enough
   --> crates/turbopack-ecmascript/src/chunk/mod.rs:332:1
    |
332 | #[turbo_tasks::function]
    | ^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `FnOnce` is not general enough
    |
    = note: closure with signature `fn(&'0 ReadRef<Assets, Vec<AssetVc>>) -> std::slice::Iter<'_, AssetVc>` must implement `FnOnce<(&ReadRef<Assets, Vec<AssetVc>>,)>`, for any lifetime `'0`...
    = note: ...but it actually implements `FnOnce<(&ReadRef<Assets, Vec<AssetVc>>,)>`
    = note: this error originates in the attribute macro `turbo_tasks::function` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I'll take another look at it later.

alexkirsz:
Re: benchmarks, I'm not seeing any significant change between this PR and `main` when benchmarking locally for 100, 1000, and 10000 modules on the `bench_hmr_to_commit/Turbopack SSR` case. According to this PR's benchmark results, I should see a ~2ms difference.

I think this is fine to merge as is, and we might have to take another look at optimizing this later on if we think we're spending too much time in `chunk_content_internal`.

alexkirsz:
What I am seeing is a much more stable `bench_hmr_to_commit/Turbopack SSR/10000` **setup** step. On my MBP, on `main`, it can take anywhere between `9s` to `270s` (likely because of the exponential backoff for HMR setup, which I might have to relax a bit).

On this branch, it takes a consistent `9s` on every run.

crates/turbopack-core/src/chunk/mod.rs
alexkirsz: See https://github.com/vercel/turbo/pull/3836#discussion_r1109930532

sokra (approved):

crates/turbopack-core/Cargo.toml
```diff
@@ -17,6 +17,7 @@ browserslist-rs = { workspace = true }
 futures = "0.3.25"
 indexmap = { workspace = true }
 lazy_static = "1.4.0"
+parking_lot = "0.12.1"
```
sokra, line 20: I think this is no longer needed

alexkirsz:
Only failing test is `test_crates_next_dev_tests_tests_integration_next_app_metadata` but it's passing locally, it's just super flaky. Force merging now.

PR #3838, opened 2023-02-16
no longer build arm64 go builds

arlyon:
> 🚨 note: I am assuming here that npm is intelligent enough to offer up the x86 package automatically, but there may be more we need to do here..

Rust's only toolchain on windows arm is msvc, which complicates linking now that we have the rust sandwich, as go uses gnu. Rather than attempt to get go building with msvc, we can instead just rely on windows' x86 emulation and publish one set of binaries for that platform until go is removed from our toolchain altogether.

### DIFF
--- a/.github/workflows/turborepo-release-step-3.yml
+++ b/.github/workflows/turborepo-release-step-3.yml
@@ -141,12 +141,17 @@ jobs:
           mv go-artifacts/turbo-go-cross-${{ inputs.release_branch }}/turbo_linux_arm64/bin/* cli/dist-linux-arm64
           chmod a+x cli/dist-linux-arm64/turbo
           chmod a+x cli/dist-linux-arm64/go-turbo
+
+          # rust doesn't have a toolchain for arm + windows + gnu, so we just use the exe from the amd64 build
+          # and rely on windows' arm JITer to do the work for us. this is because the go exe cannot be build w/ msvc
+          cp -r go-artifacts/turbo-go-cross-${{ inputs.release_branch }}/turbo_windows_amd64_v1/bin/* cli/dist-windows-arm64
+          chmod a+x cli/dist-windows-arm64/turbo.exe
+          chmod a+x cli/dist-windows-arm64/go-turbo.exe
+
           mv go-artifacts/turbo-go-cross-${{ inputs.release_branch }}/turbo_windows_amd64_v1/bin/* cli/dist-windows-amd64
           chmod a+x cli/dist-windows-amd64/turbo.exe
           chmod a+x cli/dist-windows-amd64/go-turbo.exe
-          mv go-artifacts/turbo-go-cross-${{ inputs.release_branch }}/turbo_windows_arm64/bin/* cli/dist-windows-arm64
-          chmod a+x cli/dist-windows-arm64/turbo.exe
-          chmod a+x cli/dist-windows-arm64/go-turbo.exe
+
           mv go-artifacts/turbo-go-darwin-${{ inputs.release_branch }}/turbo_darwin_amd64_v1/bin/* cli/dist-darwin-amd64
           chmod a+x cli/dist-darwin-amd64/turbo
           chmod a+x cli/dist-darwin-amd64/go-turbo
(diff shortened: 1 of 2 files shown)


NicholasLYang (approved):

mehulkar (commented):
I don't have any experience in this area, but if it works, that seems good to me. It looks like it's our `postinstall` script that chooses the right binary to download though. Maybe edit this to always serve the x86 build? 
https://github.com/vercel/turbo/blob/564a0f0dab223fbaf2ad4a7a3a5332aa50e27eb0/packages/turbo/node-platform.js#L32-L36

chris-olszewski (changes requested):
I'm almost positive we need to change our postinstall logic that resides in `packages/turbo/node-platform.js` to now look for the `turbo-windows-64` when on `arm64` instead of `turbo-windows-arm64` like it does now. 

We should also confirm that `goreleaser release --rm-dist -f combined-shim.yml --skip-publish` produces a `turbo-windows-64` where `"cpu": ["x64", "arm64"]`

.github/workflows/turborepo-release-step-3.yml
```diff
@@ -144,9 +144,6 @@ jobs:
           mv go-artifacts/turbo-go-cross-${{ inputs.release_branch }}/turbo_windows_amd64_v1/bin/* cli/dist-windows-amd64
           chmod a+x cli/dist-windows-amd64/turbo.exe
           chmod a+x cli/dist-windows-amd64/go-turbo.exe
-          mv go-artifacts/turbo-go-cross-${{ inputs.release_branch }}/turbo_windows_arm64/bin/* cli/dist-windows-arm64
```
gsoltis, line 147: I think we want to continue shipping a windows/arm package, since we'll eventually be building it natively in Rust. So I think maybe this ought to be 
`mv go-artifacts/turbo-go-cross-${{ inputs.release_branch }}/turbo_windows_amd64_v1/bin/* cli/dist-windows-arm64`

And keep the following lines?

cli/combined-shim.yml
```diff
@@ -16,6 +16,11 @@ builds:
       - arm64
     goamd64:
       - v1
+    ignore:
```
gsoltis, line 19: I think we still want to build the windows/arm package, it just has x86 binaries in it.

cli/cross-release.yml
```diff
@@ -22,7 +22,6 @@ builds:
     targets:
       - linux_arm64
       - linux_amd64
-      - windows_arm64
```
gsoltis, line 25: I think we can delete the overrides entry below as well.

gsoltis (changes requested):

gsoltis (approved):

.github/workflows/turborepo-release-step-3.yml
```diff
@@ -144,9 +144,13 @@ jobs:
...
-          chmod a+x cli/dist-windows-arm64/turbo.exe
-          chmod a+x cli/dist-windows-arm64/go-turbo.exe
+
+          # rust doesn't have a toolchain for arm + windows + gnu, so we just use the exe from the amd64 build
+          # and rely on windows' arm JITer to do the work for us. this is because the go exe cannot be build w/ msvc
+          mv go-artifacts/turbo-go-cross-${{ inputs.release_branch }}/turbo_windows_amd64_v1/bin/* cli/dist-windows-amd64
+          chmod a+x cli/dist-windows-amd64/turbo.exe
+          chmod a+x cli/dist-windows-amd64/go-turbo.exe
```
chris-olszewski, line 152: If I'm understanding this we want to use the AMD builds for ARM so I think we should be copying to the arm directory and not back over the amd directory. I think we also need to change one of the `mv`s to a copy since the two packages will be sharing the same binary.
```suggestion
          mv go-artifacts/turbo-go-cross-${{ inputs.release_branch }}/turbo_windows_amd64_v1/bin/* cli/dist-windows-arm64
          chmod a+x cli/dist-windows-arm64/turbo.exe
          chmod a+x cli/dist-windows-arm64/go-turbo.exe
```

chris-olszewski (approved):

PR #3839, opened 2023-02-16
Expose module_options and resolve_options in turbotrace

Brooooooklyn:
This pr allows passing `ModuleOptionsContext` and `ResolveOptionsContext` in fn start, so next.js can pass enableJsx: true, enableTypes: true while tracing entries, and `enableJsx: false`, `enableTypes: false` while tracing compiled files.

### DIFF
--- a/crates/node-file-trace/src/lib.rs
+++ b/crates/node-file-trace/src/lib.rs
@@ -26,7 +26,8 @@ use turbo_tasks::{
     NothingVc, TaskId, TransientInstance, TransientValue, TurboTasks, TurboTasksBackendApi, Value,
 };
 use turbo_tasks_fs::{
-    glob::GlobVc, DirectoryEntry, DiskFileSystemVc, FileSystem, FileSystemVc, ReadGlobResultVc,
+    glob::GlobVc, DirectoryEntry, DiskFileSystemVc, FileSystem, FileSystemPathVc, FileSystemVc,
+    ReadGlobResultVc,
 };
 use turbo_tasks_memory::{
     stats::{ReferenceType, Stats},
@@ -127,11 +128,6 @@ pub struct CommonArgs {
     #[cfg_attr(feature = "cli", clap(short, long))]
     #[cfg_attr(feature = "node-api", serde(default))]
     exact: bool,
-
-    /// Whether to enable mdx parsing while tracing dependencies
-    #[cfg_attr(feature = "cli", clap(long))]
-    #[cfg_attr(feature = "node-api", serde(default))]
-    enable_mdx: bool,
 
     /// Enable experimental garbage collection with the provided memory limit in
     /// MB.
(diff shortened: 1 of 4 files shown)


crates/node-file-trace/src/lib.rs
```diff
@@ -493,8 +452,15 @@ async fn run<B: Backend + 'static, F: Future<Output = ()>>(
...
         let sender = sender.clone();
+        let module_options = module_options.clone();
+        let resolve_options = resolve_options.clone();
         Box::pin(async move {
-            let output = main_operation(TransientValue::new(dir.clone()), args.clone().into());
+            let output = main_operation(
+                TransientValue::new(dir.clone()),
+                args.clone().into(),
+                module_options.map(|m| m.into()),
+                resolve_options.map(|r| r.into()),
```
sokra, line 462: You `cell` these values into the root task. And there will be a new root task for every call to `run`. So everything see different instances of `module_options` and `resolve_options` and they can't be shared.

You also don't want to `cell` stuff in `main_operation` for the same reason.

You want to pull all these `AssetContextVc` creation related stuff into a separate method and get passed `module_options` and `resolve_options` by value. (`Value<ModuleOptionsContext>`)

sokra (changes requested):

crates/node-file-trace/src/lib.rs
Brooooooklyn: `Value<ModuleOptionsContext>` is unable to pass `#[turbo_tasks::function]` because it lack of implementation for `Hash, Ord` and it's hard to implement these trait for `ModuleOptionsContext` and `ResolveOptionsContext` because they contain many other structs in their type definition.

Brooooooklyn: Let's pass in `ModuleOptionsContextVc` and `ResolveOptionsContextVc` in the first place as the arguments of `fn run` if we want to share them.

crates/turbopack/src/module_options/module_options_context.rs
```diff
@@ -41,25 +41,42 @@ impl WebpackLoadersOptions {
 #[turbo_tasks::value(shared)]
 #[derive(Default, Clone)]
 pub struct ModuleOptionsContext {
+    #[serde(default)]
```
jridgewell, line 44: Are the `serde(default)`s necessary if the struct implements `Default` itself?

crates/node-file-trace/src/lib.rs
```diff
@@ -437,6 +398,8 @@ pub async fn start(
                 println!("graph.html written");
             }
         },
+        module_options,
```
jridgewell, line 401: Looks like the `cfg(feature = "persistent_cache")` guarded `run` above this isn't updated?

crates/node-file-trace/src/lib.rs
```diff
@@ -353,6 +312,8 @@ fn process_input(dir: &Path, context: &str, input: &[String]) -> Result<Vec<Stri
 pub async fn start(
     args: Arc<Args>,
     turbo_tasks: Option<&Arc<TurboTasks<MemoryBackend>>>,
+    module_options: Option<ModuleOptionsContextVc>,
```
jridgewell, line 315: This won't work. You can't create a `Vc` outside of the turbo tasks context. If `Value<ModuleOptionsContext>` won't work, try `TransientInstance<ModuleOptionsContext>`

jridgewell (changes requested):

crates/node-file-trace/src/lib.rs
Brooooooklyn: `cfg(feature = "persistent_cache")` hasn't been maintained for a long time, and couldn't be compiled already. I think it's okay to leave it here and refactor it when we have a plan for it.

crates/turbopack/src/module_options/module_options_context.rs
Brooooooklyn: Yes, need it

crates/node-file-trace/src/lib.rs
Brooooooklyn: Cool

crates/node-file-trace/src/lib.rs
```diff
@@ -497,8 +459,15 @@ async fn run<B: Backend + 'static, F: Future<Output = ()>>(
...
         let args = args.clone();
         let sender = sender.clone();
+        let module_options = module_options.clone();
+        let resolve_options = resolve_options.clone();
         Box::pin(async move {
-            let output = main_operation(TransientValue::new(dir.clone()), args.clone().into());
+            let output = main_operation(
+                TransientValue::new(dir.clone()),
+                args.clone().into(),
+                TransientInstance::new(module_options.unwrap_or_default()),
```
jridgewell, line 468: We still need to hoist this `TransientInstance::new(module_options.unwrap_or_default())` outside the `tt.spawn_root_task`, that way all root tasks can share the transient wrapper around the same `module_options` instance. The current way will create unequal TransientInstances for every task, and kill the caching.

jridgewell (approved):
Almost there, one last blocking comment

PR #3843, opened 2023-02-16
ci(workflow): notify daily test results

kwonoj:
- closes WEB-601

This PR setups workflow to trigger next.js integration test daily, then post its result into slack. The summary from this daily run is simplified - no base, no commit link unfurl.

### DIFF
--- a/.github/actions/next-integration-stat/src/index.ts
+++ b/.github/actions/next-integration-stat/src/index.ts
@@ -158,6 +158,112 @@ async function fetchJobLogsFromWorkflow(
   return { logs, job };
 }
 
+// Store a json payload to share via slackapi/slack-github-action into Slack channel
+async function createSlackPostSummary(payload: {
+  shortCurrentNextJsVersion: string;
+  sha: string;
+  currentTestFailedSuiteCount: number;
+  currentTestPassedSuiteCount: number;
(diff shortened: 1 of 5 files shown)


.github/actions/next-integration-stat/src/index.ts
```diff
@@ -228,14 +334,18 @@ async function getInputs(): Promise<{
...
 }> {
   const token = getInput("token");
   const shouldExpandResultMessages =
     getInput("expand_result_messages") === "true";
-  const shouldDiffWithMain = getInput("diff_base") === "main";
-  if (getInput("diff_base") !== "main" && getInput("diff_base") !== "release") {
-    console.error('Invalid diff_base, must be "main" or "release"');
+  const diffBase = getInput("diff_base");
+  const shouldDiffWithMain = diffBase === "main";
+  // For the daily cron workflow, we don't compare to previous but post daily summary
```
sokra, line 345: Could we compare to the previous day?

kwonoj: Technically, but searching last day is bit wonky and also as shared in the meeting current comparision gives not much usef info due to number inconsistency, so probably will attempt it later once we have better stable test results.

wbinnssmith (approved):

PR #3846, opened 2023-02-16
feat(eslint): support turbo.json workspace configs

tknickman:
Support turbo.json workspace configs. 

Still to do:
- [x] Support workspace specific linting

There exists one edge case here, if a user forgets to add an "extend" to workspace config, we _can_ treat it as root. I punted on this one for now since it's not this packages domain to solve that (at the moment)


mehulkar (approved):

packages/turbo-utils/__tests__/getTurboRoot.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  );
+
+  test.each([
+    [""],
+    ["apps"],
+    ["apps/docs"],
+    ["apps/web"],
+    ["packages"],
+    ["packages/ui"],
+    ["not-a-real/path"],
```
nathanhammond, line 27: You mind exploding these and then `path.join(root, ...repoPath)`?

packages/turbo-utils/src/getTurboConfigs.ts
```diff
@@ -0,0 +1,89 @@
+import fs from "fs";
```
nathanhammond, line 1: Yeah, I really want "resolved config for workspace" to come from `turbo` itself so that we don't have to keep this in sync and require particular versions for particular versions of Turborepo.

nathanhammond:
(I had and deleted a comment about the accidental roots thing. It's a bug in their config but it is unfortunate.

packages/eslint-plugin-turbo/docs/rules/no-undeclared-env-vars.md
```diff
@@ -65,10 +65,10 @@ Examples of **correct** code for this rule:
...
-| `turboConfig` | No       | Auto-detect | Resolved `turbo.json` configuration                                                                                                         | `require("./turbo.json")`                    |
-| `allowList`   | No       | []          | An array of strings (or regular expressions) to exclude. NOTE: an env variable should only be excluded if it has no effect on build outputs | `["MY_API_TOKEN", "^MY_ENV_PREFIX_[A-Z]+$"]` |
+| Option             | Required | Default     | Details                                                                                                                                     | Example                                      |
+| ------------------ | -------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------- |
+| `turboConfigPaths` | No       | Auto-detect | Array of paths to `turbo.json` configuration files                                                                                          | `["./turbo.json"), "./apps/web/turbo.json"]` |
```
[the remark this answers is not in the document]
tknickman: Yea I'm kind of with you here. There's also a ton of code that could be cleaned up if we deprecated this.

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -64,31 +67,52 @@ function create(context: Rule.RuleContext): Rule.RuleListener {
...
-  });
+  const filePath = getPhysicalFilename();
+  const turboConfigPaths = options?.[0]?.turboConfigPaths;
+  const allTurboVars =
+    getEnvVarDependencies({
+      turboConfigPaths,
+      cwd,
+    }) || {};
+
+  const globalTurboVars = allTurboVars["//"];
```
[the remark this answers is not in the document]
tknickman: I thought about that, but this gets called on every change in a repo - so we would need to cache that. But then we would need a good strategy to bust that cache. If only we were really good at caching...

packages/eslint-plugin-turbo/docs/rules/no-undeclared-env-vars.md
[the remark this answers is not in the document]
tknickman: yanked it

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -63,32 +71,54 @@ function create(context: Rule.RuleContext): Rule.RuleListener {
...
-        data: { envKey },
-      });
+      // if we have a workspace config, check that too
+      if (workspaceTurboVars && workspaceTurboVars.has(envKey)) {
+        return {};
+      } else {
+        context.report({
+          node,
+          message:
+            "${{ envKey }} is not listed as a dependency in any turbo.json",
```
mehulkar, line 118: not a blocker, but it would probably be helpful to say which workspaces we looked in

packages/eslint-plugin-turbo/lib/utils/getEnvVarDependencies.ts
```diff
@@ -21,39 +20,56 @@ function findDependsOnEnvVars({
...
-    }),
-    ...globalEnv,
-  ];
-
-  Object.values(pipeline).forEach(({ env, dependsOn }) => {
-    if (dependsOn) {
-      allEnvVars.push(...findDependsOnEnvVars({ dependencies: dependsOn }));
+
+  const envVars: Record<string, Set<string>> = {
+    "//": new Set(),
```
mehulkar, line 33: I'm guessing `//` means root here also, might be good to add a comment or a named constant

packages/turbo-utils/src/getTurboConfigs.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export type TurboConfigs = Array<{
+  config: Schema;
+  turboConfigPath: string;
+  workspacePath: string;
+  isRootConfig: boolean;
+}>;
+
+// A quick and dirty workspace parser
+// TODO: after @turbo/workspace-convert is merged, we can leverage those utils here
```
mehulkar, line 19: Might be a good reason to make that package `@turbo/workspaces` instead of workspace-convert (single use case)

mehulkar (approved):
The code looks good at first pass, if we just want to get it out the door! Since it isn't tied to the version of the `turbo` package and it's pre-1.0, I'd be comfortable shipping it and resolving any issues that come up in more patches.

PR #3849, opened 2023-02-17
improve performance of TaskScopes and collectibes

sokra:


crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -261,6 +275,97 @@ impl MemoryBackend {
...
+                    id,
+                    task,
+                    root_scoped,
+                    turbo_tasks,
+                )
+            }
+        }
+    }
+
+    /// # Safty
```
alexkirsz, line 315:
```suggestion
    /// # Safety
```

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -261,6 +275,97 @@ impl MemoryBackend {
...
+            turbo_tasks,
+        ) {
+            // fast pass without creating a new task
+            task
+        } else {
+            // slow pass with key lock
+            let id = turbo_tasks.get_fresh_task_id();
+            let task =
+                Task::new_read_collectibles(id, scope_id, trait_type, turbo_tasks.stats_type());
+            // Safety: We have a fresh task id that nobody knows about yet
```
alexkirsz, line 300: This could perhaps be ensured via a newtype `FreshTaskId`/`FreshTask`. Only once you `insert` the task do you get a `Task` with a `TaskId`. `remove`ing the task gives you back a `FreshTaskId`.

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -591,7 +684,7 @@ pub(crate) enum Job {
     ScheduleWhenDirtyFromScope(AutoSet<TaskId>),
     /// Add tasks from a scope. Scheduled by `run_add_from_scope_queue` to
     /// split off work.
-    AddToScopeQueue(VecDeque<(TaskId, usize)>, TaskScopeId, bool),
+    AddToScopeQueue(VecDeque<TaskId>, TaskScopeId, usize),
```
alexkirsz, line 687: Turn this into a struct variant. It was already unclear what `bool` represented.

crates/turbo-tasks-memory/src/scope.rs
```diff
@@ -681,12 +645,18 @@ impl TaskScopeState {
         match self.collectibles.entry(trait_id) {
             Entry::Occupied(mut entry) => {
                 let (collectibles, dependent_tasks) = entry.get_mut();
-                if collectibles.remove_count(collectible, count) {
+                let old_value = collectibles.get(&collectible);
+                let new_value = old_value - count as isize;
+                if collectibles.remove_count(collectible, count) || new_value < 0 {
```
alexkirsz, line 650: Why the check for `new_value < 0` here? Intuitively it seems like `remove_count` should take care of it.

crates/turbo-tasks-memory/src/task.rs
```diff
@@ -387,6 +418,8 @@ use self::meta_state::{
     FullTaskWriteGuard, TaskMetaState, TaskMetaStateReadGuard, TaskMetaStateWriteGuard,
 };
 
+const SCOPE_OPTIMIZATION_THRESHOLD: usize = 255;
```
alexkirsz, line 421: Add a comment to explain what this const controls. Also, perhaps `(2 << 7) - 1` would be clearer here?

crates/turbo-tasks-memory/src/task.rs
```diff
@@ -672,24 +706,44 @@ impl Task {
...
+                    trait_type_id,
+                    backend,
+                    turbo_tasks,
+                )
+            }
             TaskType::Persistent(ty) => match &**ty {
                 PersistentTaskType::Native(native_fn, inputs) => {
-                    if let PrepareTaskType::Native(bound_fn) = &state.prepared_type {
+                    let future = if let PrepareTaskType::Native(bound_fn) = &state.prepared_type {
                         bound_fn()
```
alexkirsz, line 735: Why not drop state right before caling `bound_fn()` like we do for `::Root` above?

crates/turbo-tasks-memory/src/task.rs
```diff
@@ -1752,14 +1825,21 @@ impl Task {
     ) {
         let mut state = self.full_state_mut();
         if state.children.insert(child_id) {
+            if let TaskScopes::Inner(_, optimization_counter) = &state.scopes {
+                if *optimization_counter * state.children.len() > SCOPE_OPTIMIZATION_THRESHOLD {
```
alexkirsz, line 1829: This check could be a function (reused above):

```rust
should_add_to_scope(optimization_counter, children_count)
```

Fewer opportunities for error, and lets you inline the const in there and comment the function directly.

crates/turbo-tasks/src/manager.rs
```diff
@@ -400,18 +397,25 @@ impl<B: Backend> TurboTasks<B> {
...
                                 )
-                                .await;
-                                (result, duration, instant)
-                            })
-                            .await;
+                            } else {
+                                None
+                            }
+                        })
+                        .await
```
alexkirsz, line 417: This condition is hard to read. Extract the right-hand side into its own variable.

e.g.

```rust
let execution_future = CELL_COUNTERS.scope(...);
if let Some((result, duration, instant)) = execution_future.await {
```

alexkirsz (approved):

crates/turbo-tasks-memory/src/memory_backend.rs
sokra: That's a cool idea...

crates/turbo-tasks-memory/src/scope.rs
sokra: Yeah that was a bug. The read_collectibles need to be invalidated when negative count changes. Each negative count will eliminate one child scope emitted collectible. So changing from -1 to -2 might affect the visible collectibles.

crates/turbo-tasks-memory/src/task.rs
sokra: bound_fn is borrowed from state, so we can't drop it earlier

crates/turbo-tasks-memory/src/scope.rs
alexkirsz: Can you add a comment to explain this here?

PR #3851, opened 2023-02-17
Add --log-prefix=none option

mehulkar:
This adds a `--log-prefix=none` CLI argument to `turbo run` commands that will remove the `<package>:<task>: ` prefix when running tasks (both on execute and replay from cache). This removal will also apply to "single package" repos where the prefix is just `<task>: `

This new option is intended to alleviate some user issues piping logs into other tools, but comes with a Buyer Beware warning that logs from parallel tasks are still interleaved. We may implement a buffering mechanism (long time issue: #219), that will further help with this, but this PR does not do that.


cli/integration_tests/run_logging/log_prefix.t
```diff
@@ -0,0 +1,56 @@
...
+  build app-a
+  
+   Tasks:    1 successful, 1 total
+  Cached:    0 cached, 1 total
+    Time:\s*[\.0-9]+m?s  (re)
+  
+  $ cat app-a/.turbo/turbo-build.log
+
+  > build (esc)
+  > echo 'build app-a' (esc)
```
mehulkar, line 25: Having trouble validating this output with prysk: https://github.com/Nicoretti/prysk/issues/163

crates/turborepo-lib/src/cli.rs
```diff
@@ -352,6 +352,9 @@ pub struct RunArgs {
     /// to identify which packages have changed.
     #[clap(long)]
     pub since: Option<String>,
+    /// Can be: "none" or missing. Internally, the default is not named yet.
+    #[clap(long)]
+    pub log_prefix: Option<String>,
```
mehulkar, line 357: I'm just copying the existing args here, not sure what i'm doing.

crates/turborepo-lib/src/cli.rs
```diff
@@ -352,6 +352,9 @@ pub struct RunArgs {
     /// to identify which packages have changed.
     #[clap(long)]
     pub since: Option<String>,
+    /// Can be: "none" or missing. Internally, the default is not named yet.
+    #[clap(long)]
```
mehulkar, line 356: needs a default

cli/internal/run/run_spec.go
```diff
@@ -74,4 +74,7 @@ type runOpts struct {
 	graphFile     string
 	noDaemon      bool
 	singlePackage bool
+
+	// logPrefix controls whether we should print a prefix in task logs
+	logPrefix string
```
mehulkar, line 79: This should probably be an enum, similar to TaskOutputMode, but I didn't want to get too fancy yet

crates/turborepo-lib/src/cli.rs
```diff
@@ -352,6 +352,11 @@ pub struct RunArgs {
     /// to identify which packages have changed.
     #[clap(long)]
     pub since: Option<String>,
+    /// Use "none" to remove prefixes from task logs. Note that using this
+    /// option means that tasks running in parallel may stream their logs
+    /// mixed together.
```
chris-olszewski, line 357: Wordsmithing this is hard, but this almost makes it sound like only when using this option the logs may be mixed together. Maybe "Note that tasks running in parallel may interleave their logs and prefix is the only way to identify which task produced a log"

cli/integration_tests/run_logging/log_prefix.t
```diff
@@ -0,0 +1,60 @@
+Setup
+  $ . ${TESTDIR}/../setup.sh
+  $ . ${TESTDIR}/setup.sh $(pwd) ./monorepo
+
+# Run for the first time with --log-prefix=none
+  $ ${TURBO} run build --log-prefix=none
```
chris-olszewski, line 6: Add a test for `${TURBO} run build --log-prefix=junk`

mehulkar: eagle eye chris! I thought of this too and figured i'd add a test if someone called it out. I was thinking error on junk/missing value was too harsh.

chris-olszewski (approved):

PR #3853, opened 2023-02-17
Use IndexSet in IntrospectableChildrenVc

jridgewell:
Another small cleanup stemming from the the `basePath` PR. Makes the introspection ordering consistent, which is a nice usability win.


sokra (approved):

sokra (commented):
Could you remove the `children.sort()` in `crates/turbopack-dev-server/src/introspect/mod.rs`

crates/turbopack-dev-server/src/introspect/mod.rs
```diff
@@ -120,7 +120,6 @@ impl ContentSource for IntrospectionSource {
             })
             .try_join()
             .await?;
-        children.sort();
```
jridgewell, line 123: Already removed

PR #3858, opened 2023-02-17
Add turboVersion key to turbo.json

nathanhammond:
This PR is stacked on top of #4182. The relevant changes: https://github.com/vercel/turbo/pull/3858/files/6323397c434ddf0f0ede7b3663b3de4ede36147c..45951366daaec56bc08b67d4cf828a4877ebe735

In order to make it possible for users to force version constraints on global `turbo`, add `turboVersion` to the root `turbo.json`.

### DIFF
--- a/crates/turborepo-lib/Cargo.toml
+++ b/crates/turborepo-lib/Cargo.toml
@@ -42,7 +42,7 @@ reqwest = { workspace = true, default_features = false, features = ["json"] }
 rustc_version_runtime = "0.2.1"
 semver = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
-serde_json = { workspace = true }
+serde_json = { version = "1.0.94", package = "serde_jsonc" }
 serde_yaml = { workspace = true }
 tiny-gradient = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
--- a/crates/turborepo-lib/src/shim.rs
+++ b/crates/turborepo-lib/src/shim.rs
@@ -16,7 +16,7 @@ use const_format::formatcp;
 use dunce::canonicalize as fs_canonicalize;
 use env_logger::{fmt::Color, Builder, Env, WriteStyle};
 use log::{debug, Level, LevelFilter};
-use semver::Version;
+use semver::{Version, VersionReq};
 use serde::{Deserialize, Serialize};
 use tiny_gradient::{GradientStr, RGB};
 use turbo_updater::check_for_updates;
@@ -25,7 +25,7 @@ use crate::{
     cli,
     files::{
         package_json,
-        yarn_rc::{self, YarnRc},
+        yarn_rc::{self, YarnRc}, turbo_json,
     },
     get_version,
     package_manager::Globs,
@@ -55,6 +55,23 @@ fn turbo_version_has_shim(version: &str) -> bool {
     }
 
     version.major > 1
+}
+
+fn turbo_version_in_range(version: &str, root_path: PathBuf) -> bool {
+    let version = Version::parse(version).unwrap();
+
+    turbo_json::read(root_path)
+        .ok()
+        .and_then(|turbo_json| {
+            Version::parse(&turbo_json.turbo_version).ok()
+        })
+        .and_then(|specified_version| {
+            match version == specified_version {
+                true => Some(true),
+                false => None,
+            }
+        })
+        .is_some()
 }
 
 #[derive(Debug)]
@@ -583,11 +600,19 @@ impl RepoState {
                 self.spawn_local_turbo(&canonical_local_turbo, shim_args),
             ))
         } else {
-            try_check_for_updates(&shim_args, get_version());
+            let global_version = get_version();
+            try_check_for_updates(&shim_args, global_version);
+            debug!("Running command as global turbo");
+
+            if !turbo_version_in_range(global_version, &self.root) {
+                return Err(anyhow!(
+                    "Project specifies that version is 9, you're running 8."
+                ));
+            }
+
             // cli::run checks for this env var, rather than an arg, so that we can support
             // calling old versions without passing unknown flags.
             env::set_var(cli::INVOCATION_DIR_ENV_VAR, &shim_args.invocation_dir);
-            debug!("Running command as global turbo");
             cli::run(Some(self))
         }
     }
--- a/crates/turborepo-lib/src/shim.rs
+++ b/crates/turborepo-lib/src/shim.rs
@@ -21,7 +21,16 @@ use serde::{Deserialize, Serialize};
 use tiny_gradient::{GradientStr, RGB};
 use turbo_updater::check_for_updates;
 
-use crate::{cli, get_version, package_manager::Globs, PackageManager, Payload};
+use crate::{
+    cli,
+    files::{
+        package_json,
+        yarn_rc::{self, YarnRc},
+    },
+    get_version,
+    package_manager::Globs,
+    PackageManager, Payload,
+};
 
 // all arguments that result in a stdout that much be directly parsable and
 // should not be paired with additional output (from the update notifier for
@@ -174,25 +183,6 @@ pub enum RepoMode {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-struct PackageJson {
-    version: String,
-}
-
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
-struct YarnRc {
-    pnp_unplugged_folder: PathBuf,
-}
-
-impl Default for YarnRc {
-    fn default() -> Self {
-        Self {
-            pnp_unplugged_folder: [".yarn", "unplugged"].iter().collect(),
-        }
-    }
-}
-
-#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct TurboState {
     bin_path: Option<PathBuf>,
     version: &'static str,
@@ -313,9 +303,7 @@ impl LocalTurboState {
         let yarn_rc_filename =
             env::var_os("YARN_RC_FILENAME").unwrap_or_else(|| OsString::from(".yarnrc.yml"));
         let yarn_rc_filepath = root_path.join(yarn_rc_filename);
-
-        let yarn_rc_yaml_string = fs::read_to_string(yarn_rc_filepath).unwrap_or_default();
-        let yarn_rc: YarnRc = serde_yaml::from_str(&yarn_rc_yaml_string).unwrap_or_default();
+        let yarn_rc: YarnRc = yarn_rc::read(yarn_rc_filepath).unwrap_or_default();
 
         root_path.join(yarn_rc.pnp_unplugged_folder)
     }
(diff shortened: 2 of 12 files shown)


crates/turborepo-lib/src/shim.rs
```diff
@@ -330,11 +348,19 @@ impl RepoState {
...
             ))
         } else {
-            try_check_for_updates(&shim_args, get_version());
+            let global_version = get_version();
+            try_check_for_updates(&shim_args, global_version);
+            debug!("Running command as global turbo");
+
+            if !turbo_version_in_range(global_version, &self.root) {
+                return Err(anyhow!(
+                    "Project specifies that version is 9, you're running 8."
```
nathanhammond, line 357: What's idomatic for passing `version_req_string` back out so that I can use it here?

I could also make it `try_turbo_version_in_range` which has ok/err and precompose it in the function? Or is there a better pattern?

crates/turborepo-lib/src/turbo_json.rs
```diff
@@ -0,0 +1,18 @@
...
+#[derive(Deserialize, Debug)]
+pub struct TurboJson {
+    pub turboVersion: String,
+}
+
+pub fn read_turbo_json<P: AsRef<Path>>(path: P) -> Result<TurboJson, Box<dyn Error>> {
+    let file = File::open(path)?;
+    let reader = BufReader::new(file);
+
+    // FIXME: serde_json doesn't support JSONC.
```
nathanhammond, line 14: https://github.com/serde-rs/json/blob/master/src/de.rs#L248

Anywhere you can call `parse_whitespace` in `serde_json` you could also have a comment. In a fork we could just scan to the end of and then resume eating whitespace.

crates/turborepo-lib/src/shim.rs
```diff
@@ -46,6 +46,24 @@ fn turbo_version_has_shim(version: &str) -> bool {
...
+
+    let version_req_string = match read_turbo_json(root_path) {
+        Ok(turbo_json) => turbo_json.turboVersion,
+
+        // Either not present or it failed to read `turbo.json`.
+        // We'll assume this is *.
+        Err(_) => "*".to_string(),
+    };
+
+    // FIXME: this SemVer matching is "Rust" SemVer matching.
```
nathanhammond, line 60: There are a few comparators that aren't supported but it can be convinced to work.

crates/turborepo-lib/src/turbo_json.rs
```diff
@@ -0,0 +1,18 @@
+use std::{error::Error, fs::File, io::BufReader, path::Path};
+
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug)]
+pub struct TurboJson {
+    pub turboVersion: String,
```
nathanhammond, line 7: The linter yells at me about this. What should I do here?

crates/turborepo-lib/src/shim.rs
```diff
@@ -46,6 +46,24 @@ fn turbo_version_has_shim(version: &str) -> bool {
...
+fn turbo_version_in_range(version: &str, root_path: &Path) -> bool {
+    let version = Version::parse(version).unwrap();
+
+    let version_req_string = match read_turbo_json(root_path) {
+        Ok(turbo_json) => turbo_json.turboVersion,
+
+        // Either not present or it failed to read `turbo.json`.
+        // We'll assume this is *.
+        Err(_) => "*".to_string(),
+    };
```
chris-olszewski, line 58: This is extreme nit and primarily for educational purposes of how we can avoid an allocation if we fail at reading `turbo.json` (the inner `as_str` might be unnecessary, but off the top of my head I can't remember)
```suggestion
    let turbo_json = read_turbo_json(root_path);
    // Either not present or it failed to read `turbo.json`.
    // We'll assume this is *.
    let version_req_string = turbo_json.as_ref().map_or("*", |turbo_json| turbo_json.turboVersion.as_str()).unwrap_or("*");
```

crates/turborepo-lib/src/shim.rs
```diff
@@ -46,6 +46,24 @@ fn turbo_version_has_shim(version: &str) -> bool {
     version.major > 1
 }
 
+fn turbo_version_in_range(version: &str, root_path: &Path) -> bool {
+    let version = Version::parse(version).unwrap();
```
chris-olszewski, line 50: Again, this is overly cautious, but if we somehow shipped a binary where `get_version` wasn't returning a valid semver it's helpful dump what we're failing to parse:
```suggestion
    let version = Version::parse(version).unwrap_or_else(|| panic!("Invalid semver: {}", version));
```

crates/turborepo-lib/src/shim.rs
```diff
@@ -46,6 +46,24 @@ fn turbo_version_has_shim(version: &str) -> bool {
...
+        // Either not present or it failed to read `turbo.json`.
+        // We'll assume this is *.
+        Err(_) => "*".to_string(),
+    };
+
+    // FIXME: this SemVer matching is "Rust" SemVer matching.
+    match VersionReq::parse(version_req_string.as_str()) {
+        Ok(version_request) => version_request.matches(&version),
+        Err(_) => false,
+    }
```
chris-olszewski, line 64: For this I think we'll need to add a `use anyhow::Context` import statement
```suggestion
    let version_req = VersionReq::parse(version_req_string).context("Invalid semver range in turbo.json")?;
    if !version_req.matches(&version) {
        return Err(anyhow!("Project specifies that version is {}, you're running {}.", version_req, version));
    }
    Ok(())
```

crates/turborepo-lib/src/turbo_json.rs
nathanhammond: I _think_ it can be designed in a way that some automated tooling for rebasing, running a test, and then publishing it _might_ work. In other words, I _think_ the change is small enough that we can automate the fork. And then `serde_jsonc` goes up in usage to the point that eventually dtolnay reconsiders his opinion and lands the proven-to-work patch.

nathanhammond: I ended up doing this: https://github.com/nathanhammond/serde_jsonc/

crates/turborepo-lib/src/shim.rs
nathanhammond: By making the whole thing a `Result` this ended up cleaning up automatically.

crates/turborepo-lib/Cargo.toml
```diff
@@ -42,7 +42,7 @@ reqwest = { workspace = true, default_features = false, features = ["json"] }
 rustc_version_runtime = "0.2.1"
 semver = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
-serde_json = { workspace = true }
+serde_json = { version = "1.0.94", package = "serde_jsonc" }
```
nathanhammond, line 45: This is a floating, automated patch on top of `serde_json` for parsing JSONC. No API changes, release cadence is 1:1 (with up to a 1 hour delay because of cron polling).

I didn't elect to push it to everybody else without having a conversation first.

crates/turborepo-lib/src/turbo_json.rs
```diff
@@ -0,0 +1,31 @@
...
+
+impl TurboJson {
+    pub fn check_version(&self, version: &str) -> Result<bool> {
+        let version = Version::parse(version)?;
+
+        if self.turbo_version.is_empty() {
+            return Ok(true);
+        }
+
+        let version_request = VersionReq::parse(&self.turbo_version)?;
```
nathanhammond, line 28: This is a Cargo-style SemVer comparison string, which is _slightly_ different from what `node` uses.

We can't choose "both", so we should make an active decision about which pattern we want to support.

crates/turborepo-lib/src/shim.rs
```diff
@@ -407,11 +409,35 @@ impl RepoState {
                 self.spawn_local_turbo(&canonical_local_turbo, shim_args),
             ))
         } else {
-            try_check_for_updates(&shim_args, get_version());
+            let global_version = get_version();
```
nathanhammond, line 412: I feel like this should be a proc-macro. We should be confirming that it is valid at compile time.

But that's not for this PR.

crates/turborepo-lib/src/turbo_json.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub fn read_turbo_json(path: &Path) -> Result<TurboJson> {
+    let turbo_json_string = fs::read_to_string(path)?;
+    let turbo_json = serde_json::from_str(&turbo_json_string)?;
+
+    Ok(turbo_json)
+}
+
+impl TurboJson {
+    pub fn check_version(&self, version: &str) -> Result<bool> {
```
nathanhammond, line 21: I moved this into the `TurboJson` implementation since that seemed like a good shared home for it.

nathanhammond (commented):
Reviewer's Guide!

crates/turborepo-lib/Cargo.toml
```diff
@@ -42,7 +42,7 @@ reqwest = { workspace = true, default_features = false, features = ["json"] }
 rustc_version_runtime = "0.2.1"
 semver = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
-serde_json = { workspace = true }
```
mehulkar, line 45: - why don't we need to keep the `worksapce = true` bit?
- does adding the `turboVersion` require reading comments (i.e. does this soft fork need to be in this PR or can we evaluate separately for reading comments)?

crates/turborepo-lib/src/turbo_json.rs
```diff
@@ -0,0 +1,31 @@
...
+        let version = Version::parse(version)?;
+
+        if self.turbo_version.is_empty() {
```
mehulkar, line 24:
```suggestion
        // version is not required
        if self.turbo_version.is_empty() {
```

crates/turborepo-lib/src/shim.rs
```diff
@@ -407,11 +409,35 @@ impl RepoState {
...
+            try_check_for_updates(&shim_args, global_version);
+            debug!("Running command as global turbo");
+
+            // Absence of turbo.json is not an error per business logic.
```
mehulkar, line 416: It's an error downstream in Go for monorepos, btw. Single package repos don't need turbo.json.

mehulkar (approved):
There might be a good additional test here in https://github.com/vercel/turbo/blob/d2d0d3e8815c7ebe7f55075c13d2052826d7d96c/scripts/smoke-published.ts#L321-L331, which is the only place I know where we test global + local interaction. Some other inline comments.

I don't love the soft fork for serde_json, but we do have a hard requirement to keep allowing comments, so if this is the only way, let's go for it.

crates/turborepo-lib/Cargo.toml
nathanhammond: 1. `workspace = true` tells cargo to inherit the version from the root `cargo.toml`. I'd have to make this change at the root in order to get that sorted.
2. This is the first time that we're actually reading `turbo.json` in Rust so that's why this change is included in this PR.

crates/turborepo-lib/src/shim.rs
nathanhammond: Right, that was my meaning. I just mean that "we don't stop here."

crates/turborepo-lib/src/turbo_json.rs
nathanhammond: This line of code gets dropped in the version that it will be rebased on top of:

https://github.com/nathanhammond/turbo/pull/8/files#diff-3c4e635078eed173e0c6dafe5eb719efc6d1d9e2777f09fc95c4c6a826baf231

nathanhammond:
@mehulkar 

> is the only place I know where we test global + local interaction

I've got *way* more tests for global + local interaction over in #4123. Landing order is actually:

1. #4123
2. https://github.com/nathanhammond/turbo/pull/8
3. This one. (#3858)

> I don't love the soft fork for serde_json, but we do have a hard requirement to keep allowing comments, so if this is the only way, let's go for it.

The decision for JSONC is between:
- `serde_jsonc` - a _very_ clean addition-only patch on top of `serde_json` (one of the most-used crates in the world) with an automated release setup to stay up-to-date with `serde_json`.
- [`jsonc-parser`](https://github.com/dprint/jsonc-parser) - which has fewer all-time downloads than `serde_json` had last week and is only intermittently maintained.

I personally find `serde_jsonc` to be the lower risk for compatibility and maintenance.

nathanhammond:
Open points of discussion:
- [ ] Which ecosystem's SemVer?
- [ ] `serde_jsonc` vs. `jsonc-parser`

anthonyshew:
We've no implemented this by other means.

PR #3859, opened 2023-02-17
Update getting started to depend on turbo.json for order

kevinsimper:
With the improvements to dependOn, it seems to be that this is the new best practice and not having the order defined in the `npx turbo run ...`

This would make it easier for new adopters how to structure their turborepo project


nathanhammond:
Thank you for your attention to detail! This is a great update and helps teach people to think in DAGs.

nathanhammond (approved):

PR #3866, opened 2023-02-17
Fix issue with merging dependsOn from Workspace Configs

mehulkar:
Overriding `dependsOn` with _both_ a task dependency and a topo dependency seems to work, but _just_ one topo dependency doesn't work.. not sure why that would be, but test case seems to prove it also.

### DIFF
--- a/cli/integration_tests/composable_config/composing-override-values-deps.t
+++ b/cli/integration_tests/composable_config/composing-override-values-deps.t
@@ -6,43 +6,37 @@ Setup
 # The workspace does not have a turbo.json config. This test checks that both regular dependencies
 # and Topological dependencies are retained from the root config.
 
-# 1. First run, assert that dependet tasks run `dependsOn`
-  $ ${TURBO} run override-values-task-with-deps --filter=override-values > tmp.log
-# Validate in pieces. `omit-key` task has two dependsOn values, and those tasks
-# can run in non-deterministic order. So we need to validate the logs in the pieces.
-  $ cat tmp.log | grep "in scope" -A 2
+# Run override-values-task-with-deps. In the root turbo.json it has two dependsOn values
+# but in the workspace, we override to dependsOn: []. This test validates that only the
+# top level task "override-values-task-with-deps" should run. None of the dependencies should run.
+  $ ${TURBO} run override-values-task-with-deps --filter=override-values
   \xe2\x80\xa2 Packages in scope: override-values (esc)
   \xe2\x80\xa2 Running override-values-task-with-deps in 1 packages (esc)
   \xe2\x80\xa2 Remote caching disabled (esc)
-
-  $ cat tmp.log | grep "override-values:override-values-task-with-deps"
-  override-values:override-values-task-with-deps: cache miss, executing 28b15810eec6414c
+  override-values:override-values-task-with-deps: cache miss, executing 79d527063be89252
   override-values:override-values-task-with-deps: 
   override-values:override-values-task-with-deps: > override-values-task-with-deps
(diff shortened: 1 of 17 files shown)


cli/integration_tests/composable_config/composing-override-values-deps.t
```diff
@@ -42,3 +42,14 @@ Setup
    Tasks:    3 successful, 3 total
   Cached:    0 cached, 3 total
     Time:\s*[\.0-9]+m?s  (re)
+
+  $ ${TURBO} run override-values-task-with-deps-2 --filter=override-values --dry=json | jq '.tasks | map(select(.taskId == "override-values#override-values-task-with-deps-2")) | .[0].resolvedTaskDefinition'
+  {
+    "outputs": [],
+    "cache": true,
+    "dependsOn": [],
```
mehulkar, line 50: As of now, this `dependsOn` is `["^override-values-underlying-topo-task"]`, which is the config from the root config. Not sure why the override from the workspace config doesn't overwrite it.

mehulkar: Works fine with other array configs like `inputs`. Must be due to the custom unmarshaling of dependsOn

cli/internal/fs/turbo_json.go
```diff
@@ -432,17 +433,21 @@ func (btd *BookkeepingTaskDefinition) UnmarshalJSON(data []byte) error {
 	btd.TaskDefinition.TopologicalDependencies = []string{} // TODO @mehulkar: this should be a set
 	btd.TaskDefinition.TaskDependencies = []string{}        // TODO @mehulkar: this should be a set
 
+	// If there was a dependsOn field, add the bookkeeping
+	// we don't care what's in the field, just that it was there
+	// We'll use this marker to overwrite while merging TaskDefinitions.
+	if task.DependsOn != nil {
+		btd.definedFields.Add("DependsOn")
+	}
```
mehulkar, line 441: validated that this fixes the issue in vercel/front I was seeing. Test in that repo:

```bash
$ ~/dev/vercel/turbo/target/debug/turbo run unit-test --skip-infer --filter=playwright --dry=json | jq '.tasks | map(select(.taskId=="playwright#unit-test")) | .[0].resolvedTaskDefinition'
```

result after:

```diff
{
  "outputs": [],
  "cache": true,
-  "dependsOn": ["^build"],
+  "dependsOn": [],
  "inputs": ["**/*.{js,jsx,ts,tsx}"],
  "outputMode": "full",
  "env": [],
  "persistent": false
}
```

cli/integration_tests/composable_config/composing-override-values-deps.t
```diff
@@ -6,39 +6,45 @@ Setup
...
 
-# 1. First run, assert that dependet tasks run `dependsOn`
-  $ ${TURBO} run override-values-task-with-deps --filter=override-values > tmp.log
-# Validate in pieces. `omit-key` task has two dependsOn values, and those tasks
-# can run in non-deterministic order. So we need to validate the logs in the pieces.
-  $ cat tmp.log | grep "in scope" -A 2
+# Run override-values-task-with-deps. In the root turbo.json it has two dependsOn values
+# but in the workspace, we override to dependsOn: []. This test validates that only the
+# top level task "override-values-task-with-deps" should run. None of the dependencies should run.
+  $ ${TURBO} run override-values-task-with-deps --filter=override-values
```
mehulkar, line 12: I guess this test was just written wrong... maybe the result of a large PR where I was `-i` often.

gsoltis (approved):

PR #3877, opened 2023-02-20
Parallelize cache state processing.

nathanhammond:
In a dry run we walk the graph without concurrency. That's fine, except that we check the state of the cache (which can involve a network call) in the same loop.

This splits the checking of the cache state into a followup process.

### DIFF
--- a/cli/internal/run/dry_run.go
+++ b/cli/internal/run/dry_run.go
@@ -37,7 +37,7 @@ const missingTaskLabel = "<NONEXISTENT>"
 type dryRunSummary struct {
 	GlobalHashSummary *globalHashSummary `json:"globalHashSummary"`
 	Packages          []string           `json:"packages"`
-	Tasks             []taskSummary      `json:"tasks"`
+	Tasks             []*taskSummary     `json:"tasks"`
 }
 
 type globalHashSummary struct {
@@ -103,13 +103,13 @@ func DryRun(
 	// We walk the graph with no concurrency.
 	// Populating the cache state is parallelizable.
 	// Do this _after_ walking the graph.
-	withCacheState, err := checkCacheState(turboCache, taskSummaries)
+	err = populateCacheState(turboCache, taskSummaries)
 	if err != nil {
 		return err
 	}
 
 	// Assign the Task Summaries to the main summary
-	summary.Tasks = withCacheState
+	summary.Tasks = taskSummaries
 
 	// Render the dry run as json
 	if dryRunJSON {
@@ -129,8 +129,8 @@ func DryRun(
 	return nil
 }
 
-func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.CompleteGraph, taskHashes *taskhash.Tracker, rs *runSpec, base *cmdutil.CmdBase) ([]taskSummary, error) {
-	taskIDs := []taskSummary{}
+func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.CompleteGraph, taskHashes *taskhash.Tracker, rs *runSpec, base *cmdutil.CmdBase) ([]*taskSummary, error) {
+	taskIDs := []*taskSummary{}
 
 	dryRunExecFunc := func(ctx gocontext.Context, packageTask *nodes.PackageTask) error {
 		deps := engine.TaskGraph.DownEdges(packageTask.TaskID)
@@ -160,7 +160,7 @@ func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.Complete
 			return err
 		}
 
-		taskIDs = append(taskIDs, taskSummary{
+		taskIDs = append(taskIDs, &taskSummary{
 			TaskID:                 packageTask.TaskID,
 			Task:                   packageTask.Task,
 			Package:                packageTask.PackageName,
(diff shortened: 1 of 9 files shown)


cli/internal/run/dry_run.go
```diff
@@ -68,15 +69,22 @@ func DryRun(
...
 	)
 
 	if err != nil {
 		return err
 	}
 
+	// We walk the graph with no concurrency.
+	// Populating the cache state is parallelizable.
+	// Do this _after_ walking the graph.
+	withCacheState, err := checkCacheState(turboCache, taskSummaries)
```
mehulkar, line 81: Looks like `taskSummaries` is being mutated, so the return/reassign isn't necessary. Can we just comment that `taskSummaries` is being mutated instead of having faux immutability? What's idiomatic here btw? E.g. `sort.Strings()` doesn't try to hide that it's mutating the input.

cli/internal/run/dry_run.go
```diff
@@ -173,6 +175,42 @@ func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.Complete
...
+
+	wg := &sync.WaitGroup{}
+	for i := 0; i < parallelRequestCount; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for index := range queue {
+				task := &taskSummaries[index]
+				itemStatus, err := turboCache.Exists(task.Hash)
+				task.CacheState = itemStatus // TODO(mehulkar): Move this to PackageTask
```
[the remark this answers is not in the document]
nathanhammond: We don't call `Exists` in `RealRun`, we just try (and possibly fail) to load the cache. We still should capture that state, but the struct should be constructed in an entirely different way.

cli/internal/run/dry_run.go
```diff
@@ -173,6 +175,42 @@ func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.Complete
 	return taskIDs, nil
 }
 
+func checkCacheState(turboCache cache.Cache, taskSummaries []taskSummary) ([]taskSummary, error) {
+	maxParallelRequests := 8
```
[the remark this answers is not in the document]
mehulkar: so why 8 instead of 6?

mehulkar (approved):
This is fine. We'll likely need to move this to PackageTask for run summary, but I don't want to block you

cli/internal/run/dry_run.go
```diff
@@ -173,6 +175,43 @@ func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.Complete
...
+	wg := &sync.WaitGroup{}
+	for i := 0; i < parallelRequestCount; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for index := range queue {
+				task := taskSummaries[index]
+				itemStatus, err := turboCache.Exists(task.Hash)
+				task.CacheState = itemStatus
+				returnErr = err
```
chris-olszewski, line 201: Should we be using [`ErrGroup`](https://pkg.go.dev/golang.org/x/sync/errgroup) here? I'm still shakey on what's fine to do in goroutines, but every goroutine writing to the same var seems off.

cli/internal/run/dry_run.go
```diff
@@ -173,6 +175,43 @@ func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.Complete
...
+
+	wg := &sync.WaitGroup{}
+	for i := 0; i < parallelRequestCount; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for index := range queue {
+				task := taskSummaries[index]
+				itemStatus, err := turboCache.Exists(task.Hash)
+				task.CacheState = itemStatus
```
chris-olszewski, line 200: Should we only be altering the cache state if there's an error checking the cache?

cli/internal/run/dry_run.go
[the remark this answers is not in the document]
nathanhammond: Because I decided that Turborepo users were more-likely to be on good connections, and we control the server.

cli/internal/run/dry_run.go
nathanhammond: Since I blow up the entire dry run for a single error (this matches current state, BTW) I only need a "has Error" flag. But realizing that clobbering with `nil` is going to end badly.

cli/internal/run/dry_run.go
nathanhammond: The codepath underneath the hood guarantees `ItemStatus{}, error`, so it's safe to set... but also we blow up the entire dry run if there _is_ an error, so it's fine.

cli/internal/run/dry_run.go
nathanhammond: Fixed here: https://github.com/vercel/turbo/pull/3877/commits/42a40c708961c838f5231eda6a68f7c6e20196ac

Good catch.

cli/internal/run/dry_run.go
nathanhammond: Arguably, `Exists` throwing an error should just mean `false`; always assuming that the error is repeatable with `false` being a valid outcome.

nathanhammond: We only call `cache.Exists` in `dry_run.go`. It could map an error to `false`, which is conservative, doesn't unnecessarily blow up a dry run call, and would provide the same expected outcome:

Anyway, here is that alternative:
https://github.com/nathanhammond/turbo/pull/7/files

chris-olszewski (approved):
I think this is good to ship either as is or with the alternative merged in

mehulkar:
Closing this in favor of #4110 which resolves merge conflicts. I would have pushed a merge commit here, but I don't think I have write access to your fork @nathanhammond. Hope you don't mind!

mehulkar:
Closing this in favor of #4110 which resolves merge conflicts. I would have pushed a merge commit here, but I don't think I have write access to your fork @nathanhammond. Hope you don't mind!

PR #3878, opened 2023-02-20
fix consistency issue with reading collectibles

sokra:
connect original task to collectibles reading to fix strongly consistency

move read collectibles tasks to task and scope instead of a separate cache

reading collectibles is no longer strongly consistent by default, it's opt-in now

### DIFF
--- a/crates/turbo-tasks-memory/src/scope.rs
+++ b/crates/turbo-tasks-memory/src/scope.rs
@@ -340,8 +340,11 @@ impl TaskScope {
         self_id: TaskScopeId,
         trait_id: TraitTypeId,
         reader: TaskId,
-    ) -> (CountHashSet<RawVc>, Vec<TaskScopeId>) {
+    ) -> Result<(CountHashSet<RawVc>, Vec<TaskScopeId>), EventListener> {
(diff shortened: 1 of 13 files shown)


crates/turbo-tasks-memory/src/scope.rs
```diff
@@ -393,11 +408,12 @@ impl TaskScope {
             "Scope dependent tasks not correctly cleaned up: {:?}",
             state.dependent_tasks
         );
-        assert!(
-            state.collectibles.is_empty(),
-            "Scope collectibles not correctly cleaned up: {:?}",
-            state.collectibles
-        );
+        // TODO read_collectibles_tasks need to be cleaned up
```
alexkirsz, line 411: Is there a linear issue for this?

alexkirsz (approved):

crates/turbo-tasks-memory/src/scope.rs
sokra: https://linear.app/vercel/issue/WEB-615/the-read-collectibles-in-taskscope-is-not-cleaned-up-when-the

PR #779, opened 2022-02-25
fix: upstream eslint-plugin-react issue causing next build and lint to fail

jaredpalmer:
Fixes upsteam https://github.com/yannickcr/eslint-plugin-react/issues/3215

PR #3880, opened 2023-02-20
refactor: remove `react/jsx-key` ESLint rule

albertothedev:
It was added in vercel/turbo#779 as a temporary fix but forgotten about.


nathanhammond (approved):
Thank you so much for your attention to detail!

PR #3882, opened 2023-02-20
add possible unknown mutation to objects and arrays

sokra:
Before the static evaluation didn't consider the fact that objects and arrays can be mutated. So we can't just assume for sure that they have certain properties/items. Instead we add an unknown mutation alternative to handle that.

This avoids `if(obj.prop)` to be replaced with `if(true)` when prop is initialized with true. It might be modified in future.

### DIFF
--- a/crates/turbopack-ecmascript/src/analyzer/mod.rs
+++ b/crates/turbopack-ecmascript/src/analyzer/mod.rs
@@ -266,9 +266,17 @@ pub enum JsValue {
     // NESTED VALUES
     // ----------------------------
     /// An array of nested values
-    Array(usize, Vec<JsValue>),
+    Array {
+        total_nodes: usize,
+        items: Vec<JsValue>,
+        mutable: bool,
+    },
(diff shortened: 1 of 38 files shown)


crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -381,18 +389,20 @@ impl Display for JsValue {
         match self {
             JsValue::Constant(v) => write!(f, "{v}"),
             JsValue::Url(url) => write!(f, "{}", url),
-            JsValue::Array(_, elems) => write!(
+            JsValue::Array { items, mutable, .. } => write!(
                 f,
-                "[{}]",
-                elems
+                "{}[{}]",
+                if *mutable { "" } else { "frozen " },
```
alexkirsz, line 395: olaf

alexkirsz (approved):

ForsakenHarmony (approved):

PR #3887, opened 2023-02-21
Introduce TraitRefs, remove cell_local

alexkirsz:
`.cell_local`'s semantics are very confusing and can be the cause of particularly hard-to-debug issues. Furthermore, they can't be nested correctly, which is necessary for my future HMR work. Instead, @sokra proposed that we should always use `ReadRef`s instead. 

This PR introduces a way to generate `ReadRef`-like `TraitRef`s for a given trait Vc. The purpose of these refs is only to be passed around, to be converted back into a `Vc` eventually. You can't call any other method on a `TraitRef`. This serves the same purpose as `cell_local`, as you can "snapshot" a value pointed to by a trait Vc through a trait ref, and re-cell that snapshot at a later time (see the tests).

However, this has the same pitfall as `cell_local`: nested `Vc`s will still update. @sokra proposed we should have `ValueSnapshot` types which flatten nested `Vc`s recursively, but this will have to wait until we have time to tackle tech debt.

This also introduces the `ReadRef::cell` method, which works the same way as `TraitRef::cell` but can be used on `ReadRef`s instead.

### DIFF
--- a/crates/turbo-tasks/src/backend.rs
+++ b/crates/turbo-tasks/src/backend.rs
@@ -9,17 +9,16 @@ use std::{
     time::{Duration, Instant},
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
 use serde::{Deserialize, Serialize};
 
 pub use crate::id::BackendJobId;
 use crate::{
     event::EventListener, manager::TurboTasksBackendApi, primitives::RawVcSetVc, raw_vc::CellId,
     registry, task_input::SharedReference, FunctionId, RawVc, ReadRef, TaskId, TaskIdProvider,
-    TaskInput, TraitTypeId,
+    TaskInput, TraitRef, TraitTypeId, ValueTraitVc,
 };
 
-/// Different Task types
 pub enum TaskType {
     /// Tasks that only exist for a certain operation and
     /// won't persist between sessions
@@ -162,6 +161,24 @@ impl CellContent {
             .downcast()
             .ok_or_else(|| anyhow!("Unexpected type in cell"))?;
         Ok(unsafe { ReadRef::new_transparent(data) })
+    }
+
+    /// # Safety
+    ///
+    /// The caller must ensure that the CellContent contains a vc that
+    /// implements T.
+    pub unsafe fn cast_trait<T>(self) -> Result<TraitRef<T>>
+    where
+        T: ValueTraitVc,
+    {
+        let shared_reference = self.0.ok_or_else(|| anyhow!("Cell is empty"))?;
+        if shared_reference.0.is_none() {
+            bail!("Cell content is untyped");
+        }
+        Ok(
+            // Safety: We just checked that the content is typed.
+            unsafe { TraitRef::new(shared_reference) },
+        )
     }
 
     pub fn try_cast<T: Any + Send + Sync>(self) -> Option<ReadRef<T>> {
--- a/crates/turbo-tasks/src/backend.rs
+++ b/crates/turbo-tasks/src/backend.rs
@@ -167,7 +167,7 @@ impl CellContent {
     ///
     /// The caller must ensure that the CellContent contains a vc that
     /// implements T.
-    pub unsafe fn cast_trait<T>(self) -> Result<TraitRef<T>>
+    pub fn cast_trait<T>(self) -> Result<TraitRef<T>>
     where
         T: ValueTraitVc,
     {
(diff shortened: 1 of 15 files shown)


ForsakenHarmony (approved):

crates/turbo-tasks/src/trait_ref.rs
```diff
@@ -0,0 +1,64 @@
...
+    _t: PhantomData<Arc<T>>,
+}
+
+impl<T> TraitRef<T> {
+    /// # Safety
+    ///
+    /// The caller must ensure that:
+    /// * the shared reference points to a vc that implements the value trait T;
+    /// * the shared reference is typed.
+    pub(crate) unsafe fn new(shared_reference: SharedReference) -> Self {
```
sokra, line 28:
```suggestion
    pub(crate) fn new(shared_reference: SharedReference) -> Self {
```

crates/turbo-tasks/src/raw_vc.rs
```diff
@@ -493,6 +487,48 @@ impl<T: Any + Send + Sync, U: Any + Send + Sync> Future for ReadRawVcFuture<T, U
...
+impl<T> Cast for TraitCast<T>
+where
+    T: ValueTraitVc,
+{
+    type Output = TraitRef<T>;
+
+    fn cast(content: CellContent) -> Result<Self::Output> {
+        // Safety: Constructor ensures the cell content points to a value that
+        // implements T
+        unsafe { content.cast_trait::<T>() }
```
sokra, line 528:
```suggestion
        content.cast_trait::<T>()
```

crates/turbo-tasks/src/backend.rs
```diff
@@ -164,6 +163,24 @@ impl CellContent {
...
+    where
+        T: ValueTraitVc,
+    {
+        let shared_reference = self.0.ok_or_else(|| anyhow!("Cell is empty"))?;
+        if let None = shared_reference.0 {
+            bail!("Cell content is untyped");
+        }
+        Ok(
+            // Safety: We just checked that the content is typed.
+            unsafe { TraitRef::new(shared_reference) },
```
sokra, line 180:
```suggestion
            TraitRef::new(shared_reference),
```

crates/turbopack-core/src/version.rs
```diff
@@ -27,7 +27,8 @@ pub trait VersionedContent {
         // By default, since we can't make any assumptions about the versioning
         // scheme of the content, we ask for a full invalidation, except in the
         // case where versions are the same. And we can't compare `VersionVc`s
-        // directly since `.cell_local()` breaks referential equality checks.
+        // directly since going through a `TraitRef<VersionVc>` breaks referential
+        // equality checks.
```
sokra, line 31: Actually this works now. We can compare `VersionTraitRef`s via pointer equality. `SharedReference` implements `Eq` so you can just derive that. Also `Hash` and `PartialEq`.

crates/turbopack-dev-server/src/update/stream.rs
```diff
@@ -99,31 +99,30 @@ async fn compute_update_stream(
 
 #[turbo_tasks::value]
 struct VersionState {
-    inner: State<VersionVc>,
+    #[turbo_tasks(trace_ignore)]
+    version: State<TraitRef<VersionVc>>,
```
sokra, line 103: Much better 👍

sokra (approved):

crates/turbo-tasks/src/trait_ref.rs
alexkirsz: I marked all of these as unsafe as there are some assumptions the caller must verify in order to use these methods soundly. It's not related to memory safety, but I think it's still a useful tool to ensure we're not calling into any sensitive methods by accident.

alexkirsz: I can remove all these `unsafe` if we don't want to mix the two.

alexkirsz:
* main
  * **PR #3887** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3887" target="_blank">Graphite</a>  👈
    * **PR #3974** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3974" target="_blank">Graphite</a> 
      * **PR #3975** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3975" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3887?utm_source=stack-comment).

PR #3893, opened 2023-02-21
ci: Remove build debug turborepo

NicholasLYang:


tknickman (approved):

.github/workflows/test.yml
```diff
@@ -285,14 +285,14 @@ jobs:
...
+      - uses: ./.github/actions/setup-turborepo-environment
         with:
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
           target: ${{ matrix.os.name }}
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
 
       - name: E2E Tests
         # Turbo has already been built in previous step, no need to rebuild
-        run: pnpm -- turbo-prebuilt run e2e-prebuilt --filter=cli
+        run: pnpm -- turbo run e2e-prebuilt --filter=cli
```
mehulkar, line 295: So this will build turbo individually right?

mehulkar (approved):
I'm not sure I understand why this is better. `pnpm -- turbo` will also build turbo right? (as opposed to build before and then `pnpm -- turbo-prebuilt` ? I'm probably just missing some context though, so I'm ok with trying this out

.github/workflows/test.yml
NicholasLYang: We basically build a first version of turbo to then run our individual tasks, some of which involve building turbo again. We do this on every single workflow so that's a lot of redundant turbo building. It's like building rustc from source for every workflow run.

PR #3897, opened 2023-02-21
Implement Rewrite Headers support during routing

jridgewell:
The final piece to supporting Rewrites is implementing support for overwriting **response** headers. These overwritten headers need to be collected until we resolve into a final `ContentSourceContent::Static` (which gets turned into `ResolveSourceRequestResult`, then `GetFromSourceResult`) to be returned into the HTTP response.

I chose to implement this as an external list from the headers already present in a `ContentSourceContent::Static`. It felt wrong to `content.await?` just to overwrite the returned headers, and then repackage that as a new `ContentSourceContentVc`. These `header_overwrites` exist only as a byproduct of route resolution, and specifically the exact path of rewrites we go through to get there. They're not intrinsic to the `ContentSourceContent` itself.

Fixes WEB-617
Fixes WEB-299


sokra (changes requested):
Test case

wbinnssmith (approved):

jridgewell:
I'm working on the test cases for Rewrite/Redirect/Headers now.

jridgewell:
Tests implemented for Rewrite/Redirect, but the Headers support [**just landed**](https://github.com/vercel/next.js/commit/185a8921c878fb6fff46d9fc299683ee689d3415#diff-b8d76362f52dfa3141163a91f6dc67dbfe72242de22db1c388c387698031eb6eR142) in Next side and I can't test it until we update our canary dep.

jridgewell:
I've manually confirmed this works once by duplicating https://github.com/vercel/next.js/commit/185a8921c878fb6fff46d9fc299683ee689d3415.

padmaia (approved):

jridgewell:
Merging over some test flake.

sokra:
> Merging over some test flake.

kodiak could handle this now. It won't block on mac or windows CI anymore.

PR #2642, opened 2022-11-08
Copy `public` directory to docker

sandumorari:
Added copy `public` directory command to Dockerfile in order for nextjs to be able to read static files from it.

PR #3899, opened 2023-02-21
Add empty `public` folder to stop Docker builds breaking

a3957273:
Otherwise it can't create a cache key and the builds fail.  This error affects both `buildkit` and `moby` Docker builds.


a3957273:
This error was created in this PR:

https://github.com/vercel/turbo/pull/2642/files

It can be reproduced by running:

```
npx degit vercel/turbo/examples/with-docker with-docker
cd with-docker
yarn install
COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker-compose -f docker-compose.yml build
```

tknickman (approved):

PR #3902, opened 2023-02-21
Error messaging when using nonexistent task name

mehulkar:


cli/internal/core/engine.go
```diff
@@ -160,6 +160,10 @@ func (e *Engine) Prepare(options *EngineBuildingOptions) error {
 	// we can just exit early.
 	// TODO(mehulkar): but we still need to validate bad task names?
 	if len(pkgs) == 0 {
+		for _, taskName := range taskNames {
+			// look for task name in all workspaces and exit if it isn't found?
+			fmt.Printf("[debug] taskName %#v\n", taskName)
+		}
```
mehulkar, line 166: It doesn't seem like a great idea to load all configs just to validate that a task name is valid, even though we _know_ we're not going to execute it. Is there any reason to implement this?

PR #3907, opened 2023-02-22
Support NeedData responses in SourceMap/SourceMapTrace ContentSources

jridgewell:
This implements a new `ContentSourceProcessor` trait and `WrappedContentSource` struct that allows our `ContentSource`s to register some processor to transform the eventual `ContentSourceContent` that an inner `ContentSource` returns. Yah, it's a bit of a mouthful with lots of very similar sounding names.

Essentially, the old `SourceMapContentSource` and `NextSourceMapTraceContentSource` wrapped some inner `ContentSource`, and would request content from them, and process that content into a source map JSON or trace JSON. But, they didn't implement the `NeedData` response, so it only supported very primitive inner content sources.

This PR extracts that knowledge into a single `WrappedContentSource`, which will recursively wrap every `ContentSourceResult` until we land on a fully resolved `ContentSourceResult::Result`. At that point, it hands off the work to a `WrappedGetContentSource`, which performs the processing on whatever content is returned by the inner `GetContentSourceContent`. If you can't tell yet, I'm making the description intentionally verbose to highlight just how similar our struct/trait names are.

Fixes WEB-614

### DIFF
--- a/crates/next-core/src/next_image/mod.rs
+++ b/crates/next-core/src/next_image/mod.rs
@@ -5,8 +5,8 @@ use turbo_tasks::{primitives::StringVc, Value};
 use turbopack_core::introspect::{Introspectable, IntrospectableVc};
 use turbopack_dev_server::source::{
     query::QueryValue, ContentSource, ContentSourceContent, ContentSourceData,
-    ContentSourceDataFilter, ContentSourceDataVary, ContentSourceResult, ContentSourceResultVc,
-    ContentSourceVc, NeededData, ProxyResult,
+    ContentSourceDataFilter, ContentSourceDataVary, ContentSourceResultVc, ContentSourceVc,
+    NeededData, ProxyResult,
 };
 
 /// Serves, resizes, optimizes, and re-encodes images to be used with
@@ -65,11 +65,7 @@ impl ContentSource for NextImageContentSource {
         // TODO: consume the assets, resize and reduce quality, re-encode into next-gen
         // formats.
         if let Some(path) = url.strip_prefix('/') {
-            let asset = this.asset_source.get(path, Default::default());
-            let inner = asset.await?;
-            if let ContentSourceResult::Result { .. } = &*inner {
-                return Ok(asset);
-            }
+            return Ok(this.asset_source.get(path, Default::default()));
         }
 
         // TODO: This should be downloaded by the server, and resized, etc.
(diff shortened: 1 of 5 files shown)


crates/next-core/src/next_image/mod.rs
```diff
@@ -65,11 +65,7 @@ impl ContentSource for NextImageContentSource {
         // TODO: consume the assets, resize and reduce quality, re-encode into next-gen
         // formats.
         if let Some(path) = url.strip_prefix('/') {
-            let asset = this.asset_source.get(path, Default::default());
-            let inner = asset.await?;
-            if let ContentSourceResult::Result { .. } = &*inner {
-                return Ok(asset);
-            }
+            return Ok(this.asset_source.get(path, Default::default()));
```
jridgewell, line 68: I can't think of a case where we really want to fall into the `HttpProxy` redirect below.

crates/turbopack-dev-server/src/source/source_maps.rs
```diff
@@ -66,55 +68,77 @@ impl ContentSource for SourceMapContentSource {
...
-    #[turbo_tasks::function]
-    fn ty(&self) -> StringVc {
-        StringVc::cell("source map content source".to_string())
-    }
-
-    #[turbo_tasks::function]
-    fn details(&self) -> StringVc {
-        StringVc::cell("serves chunk and chunk item source maps".to_string())
+        let content = sm.to_rope().await?;
+        let asset = AssetContentVc::from(File::from(content).with_content_type(APPLICATION_JSON));
```
jridgewell, line 141: Unrelated fix to add `content-type: application/json`.

jridgewell (commented):
@sokra I don't know how to make a test case for this one, because as far as I know the inner sources for both of these don't need additional data. Because you filled the ticket, I'm assuming you have a test app that can hit it? Can you push that into this PR?

sokra (approved):

sokra:
> @sokra I don't know how to make a test case for this one, because as far as I know the inner sources for both of these don't need additional data. Because you filled the ticket, I'm assuming you have a test app that can hit it? Can you push that into this PR?

@alexkirsz filled the ticket. I assume one hit this when the Stack Trace points to a path to resolves to a next.js page.

I think it's caused by that config is next.js config in front:

``` js
      beforeFiles: [
            {
              source: '/:path*.map',
              destination: '/404',
            },
```

alexkirsz:
I've seen the error both on front and on a much smaller `pages/` app without the above config.

PR #3910, opened 2023-02-22
Include chunk path in WithClientChunks

alexkirsz:
`WithClientChunksChunkItem` currently only includes paths to chunks' references, but not chunks themselves.

Fixes WEB-621.


sokra (approved):

sokra:
clipy

PR #3912, opened 2023-02-22
separate app and pages structure extraction from source generation

sokra:
We need to invalidate router node.js process when routes change for correctness

In future we can pass the routes list to next.js instead of only invalidating @jridgewell 

We also need the separated structure for next build @alexkirsz

### DIFF
--- a/crates/next-core/src/app_structure.rs
+++ b/crates/next-core/src/app_structure.rs
@@ -1,0 +1,316 @@
+use std::collections::HashMap;
+
+use anyhow::{bail, Result};
+use turbo_tasks::{
+    primitives::{StringVc, StringsVc},
+    CompletionVc, ValueToString,
+};
+use turbo_tasks_fs::{
+    DirectoryContent, DirectoryEntry, File, FileContentVc, FileSystemEntryType, FileSystemPathVc,
+};
+use turbopack_core::issue::{Issue, IssueSeverity, IssueSeverityVc, IssueVc};
+use turbopack_dev_server::source::specificity::SpecificityVc;
+
+use crate::{
+    app_render::{LayoutSegment, LayoutSegmentVc, LayoutSegmentsVc},
+    embed_js::wrap_with_next_js_fs,
+    next_config::NextConfigVc,
+};
+
+#[turbo_tasks::value]
+pub enum AppStructureItem {
+    Page {
+        segment: LayoutSegmentVc,
+        segments: LayoutSegmentsVc,
+        url: FileSystemPathVc,
+        specificity: SpecificityVc,
+        page: FileSystemPathVc,
+    },
+}
+
+#[turbo_tasks::value_impl]
+impl AppStructureItemVc {
+    #[turbo_tasks::function]
+    pub async fn routes_changed(self) -> Result<CompletionVc> {
+        match *self.await? {
+            AppStructureItem::Page { url, .. } => url.await?,
+        };
+        Ok(CompletionVc::new())
+    }
+}
+
+#[turbo_tasks::value]
+pub struct AppStructure {
+    pub directory: FileSystemPathVc,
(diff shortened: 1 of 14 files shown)


crates/next-core/src/app_structure.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub async fn routes_changed(self) -> Result<CompletionVc> {
+        match *self.await? {
+            AppStructureItem::Page { url, .. } => url.await?,
+        };
+        Ok(CompletionVc::new())
+    }
+}
+
+#[turbo_tasks::value]
+pub struct AppStructure {
```
alexkirsz, line 43: Comment `AppStructure`, its fields, and `AppStructureItem`.

crates/next-core/src/app_structure.rs
```diff
@@ -0,0 +1,316 @@
...
+
+#[turbo_tasks::value_impl]
+impl AppStructureVc {
+    #[turbo_tasks::function]
+    pub async fn directory(self) -> Result<FileSystemPathVc> {
+        Ok(self.await?.directory)
+    }
+
+    #[turbo_tasks::function]
+    pub async fn routes_changed(self) -> Result<CompletionVc> {
```
alexkirsz, line 57: Explain somewhere that this `routes_changed` method is a way to recursively watch for route changes, by marking leaf route URLs as dependencies.

crates/next-core/src/pages_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub async fn routes_changed(self) -> Result<CompletionVc> {
+        if let Some(pages_structure) = *self.await? {
+            pages_structure.routes_changed().await?;
+        }
+        Ok(CompletionVc::new())
+    }
+}
+
+#[turbo_tasks::function]
+pub async fn find_pages_structure(
```
alexkirsz, line 102: Comment this public item.

crates/next-core/src/pages_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+        match *self.await? {
+            PagesStructureItem::Page { url, .. } => url.await?,
+            PagesStructureItem::Api { url, .. } => url.await?,
+        };
+        Ok(CompletionVc::new())
+    }
+}
+
+#[turbo_tasks::value]
+pub struct PagesStructure {
```
alexkirsz, line 62: Comment this item, its fields, and `PagesStructureItem` and its fields.

crates/turbo-tasks/src/completion.rs
```diff
@@ -14,9 +14,19 @@ impl Default for CompletionVc {
...
+    #[turbo_tasks::function]
+    pub fn immutable() -> Self {
+        CompletionVc::cell(Completion)
+    }
+}
+
 // no #[turbo_tasks::value_impl] to inline new into the caller task
 // this ensures it's re-created on each execution
 impl CompletionVc {
+    /// This is always be a new completion and invalidates the reading task.
```
alexkirsz, line 29:
```suggestion
    /// This will always be a new completion and invalidates the reading task.
```

crates/turbopack-node/src/pool.rs
```diff
@@ -105,6 +105,7 @@ impl NodeJsPoolProcess {
         let port = listener.local_addr().context("getting port")?.port();
         let mut cmd = Command::new("node");
         cmd.current_dir(cwd);
+        cmd.arg("--async-stack-traces");
```
alexkirsz, line 108: Was this change meant to be in this PR? Can add a comment to explain why we want this?

alexkirsz (approved):

crates/turbopack-node/src/pool.rs
sokra: no should be removed

PR #3919, opened 2023-02-22
fix(examples): svelte (#3919

tknickman:
Svelte example looks like it's not deploying due to some dependency drift.


mehulkar (approved):

tknickman:
Forcing this one - example built and deployed, no need to wait for the rest of CI here.

PR #3920, opened 2023-02-22
fix(examples): kitchen-sink (#3920

tknickman:
Update lock and remix app for kitchen sink example


tknickman:
Merging this one through to unblock the template.

PR #3923, opened 2023-02-22
implement route.js support

sokra:
https://beta.nextjs.org/docs/routing/route-handlers

### DIFF
--- a/crates/next-core/js/src/internal/edge.ts
+++ b/crates/next-core/js/src/internal/edge.ts
@@ -1,0 +1,143 @@
+import { join } from "path";
+
+import type {
+  BaseNextRequest,
+  BaseNextResponse,
+} from "next/dist/server/base-http";
+import type {
+  NodeNextRequest,
+  NodeNextResponse,
+} from "next/dist/server/base-http/node";
+import { parse, ParsedUrlQuery } from "querystring";
+import type { Params } from "next/dist/shared/lib/router/utils/route-matcher";
+import { FetchEventResult } from "next/dist/server/web/types";
+import { getCloneableBody } from "next/dist/server/body-streams";
+
+export async function runEdgeFunction({
+  edgeInfo,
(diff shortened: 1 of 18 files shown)


crates/next-core/js/src/internal/edge.ts
```diff
@@ -0,0 +1,143 @@
+import { join } from "path";
```
alexkirsz, line 1: Link which file in `next.js` you adapted for this code, so we can more easily keep track of updates/origin code.

crates/next-core/src/app_structure.rs
```diff
@@ -184,6 +195,21 @@ async fn get_app_structure_for_directory(
 
     let layout = files.get("layout");
 
+    if let (Some(_), Some(route_path)) = (page, route) {
+        AppStructureIssue {
+            severity: IssueSeverity::Error.into(),
+            path: route_path,
+            message: StringVc::cell(
+                "It's not possible to have a page and a route in the same directory.".to_string(),
```
alexkirsz, line 203: Add that "the route will be ignored" in the error message.

alexkirsz (changes requested):

crates/next-core/src/app_source.rs
```diff
@@ -581,3 +668,60 @@ impl NodeEntry for AppRenderer {
...
+            this.intermediate_output_path.join("chunks"),
+            this.server_root.join("_next/static/assets"),
+            this.context.compile_time_info().environment(),
+        )
+        .layer("ssr")
+        .css_chunk_root_path(this.server_root.join("_next/static/chunks"))
+        .build();
+
+        Ok(NodeRenderingEntry {
+            module: EcmascriptModuleAssetVc::new(
```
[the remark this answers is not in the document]
sokra: This one has no inner assets

PR #3934, opened 2023-02-23
Add global hash inputs to dry run

mehulkar:


cli/internal/run/dry_run.go
```diff
@@ -33,8 +34,33 @@ const missingTaskLabel = "<NONEXISTENT>"
...
+	globalFileHashMap    map[turbopath.AnchoredUnixPath]string
+	rootExternalDepsHash string
+	hashedSortedEnvPairs []string
+	globalCacheKey       string
+	pipeline             fs.PristinePipeline
+}) *globalHashSummary {
+	return &globalHashSummary{
+		GlobalFileHashMap:    ghInputs.globalFileHashMap,
+		RootExternalDepsHash: ghInputs.rootExternalDepsHash,
+		HashedSortedEnvPairs: ghInputs.hashedSortedEnvPairs,
```
tknickman, line 60: We'll need to deal with these obscuring these values right?

mehulkar: yeah, hmm. since everyone using turbo-ignore is going to print this in vercel, i think for this PR i'll just leave it out. good catch

cli/internal/run/dry_run.go
```diff
@@ -33,8 +34,32 @@ const missingTaskLabel = "<NONEXISTENT>"
 // DryRunSummary contains a summary of the packages and tasks that would run
 // if the --dry flag had not been passed
 type dryRunSummary struct {
-	Packages []string      `json:"packages"`
-	Tasks    []taskSummary `json:"tasks"`
+	GlobalHashSummary *globalHashSummary `json:"globalHashSummary"`
```
mehulkar, line 37: bikeshed this key name

cli/internal/run/dry_run.go
```diff
@@ -215,6 +240,23 @@ func displayDryTextRun(ui cli.Ui, summary *dryRunSummary, workspaceInfos graph.W
...
+	ui.Info(util.Sprintf("${CYAN}${BOLD}Global Hash Inputs${RESET}"))
+	fmt.Fprintln(w1, util.Sprintf("  ${GREY}Global Files\t=\t%d${RESET}", fileCount))
+	fmt.Fprintln(w1, util.Sprintf("  ${GREY}External Dependencies Hash\t=\t%s${RESET}", summary.GlobalHashSummary.RootExternalDepsHash))
+	fmt.Fprintln(w1, util.Sprintf("  ${GREY}Global Cache Key\t=\t%s${RESET}", summary.GlobalHashSummary.GlobalCacheKey))
+	if bytes, err := json.Marshal(summary.GlobalHashSummary.Pipeline); err == nil {
+		fmt.Fprintln(w1, util.Sprintf("  ${GREY}Root pipeline\t=\t%s${RESET}", bytes))
+	}
+	if err := w1.Flush(); err != nil {
+		return err
+	}
```
mehulkar, line 258: example of what this looks like:
CleanShot 2023-02-23 at 11 50 33@2x

cli/internal/run/dry_run.go
```diff
@@ -215,6 +240,23 @@ func displayDryTextRun(ui cli.Ui, summary *dryRunSummary, workspaceInfos graph.W
...
 	}
 
+	fileCount := 0
+	for range summary.GlobalHashSummary.GlobalFileHashMap {
+		fileCount = fileCount + 1
+	}
+	w1 := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
+	ui.Output("")
+	ui.Info(util.Sprintf("${CYAN}${BOLD}Global Hash Inputs${RESET}"))
+	fmt.Fprintln(w1, util.Sprintf("  ${GREY}Global Files\t=\t%d${RESET}", fileCount))
```
mehulkar, line 250: added a length here instead of the full list (json has the full list), since it could be really large. i'm not sure anyone will care about the full expanded list except to try to parse/operate on it

cli/internal/run/dry_run.go
tknickman: Fine by me

cli/internal/run/dry_run.go
tknickman: I would dim these key / vals as well, but not a blocker.

cli/internal/run/dry_run.go
tknickman: Always something we could revisit later if needed.

tknickman (approved):

cli/internal/run/dry_run.go
mehulkar: I did that in a previous pass, but then felt it was more like the Packages In scope section rather than tasks section. I'll make a note of this and come back to it once all the pieces are in

PR #3935, opened 2023-02-23
fix(docs): loading eager for blog avatars

tknickman:
Avatar's shouldn't load lazily (they're always above the fold)

### DIFF
--- a/docs/components/Avatar.tsx
+++ b/docs/components/Avatar.tsx
@@ -8,10 +8,10 @@ export const Avatar = ({ name, picture, twitterUsername }: AuthorDetails) => (
         src={picture}
         height={32}
         width={32}
-        loading="eager"
         title={name}
         className="w-full rounded-full"
         alt={name}
+        priority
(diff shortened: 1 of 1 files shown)


docs/components/Avatar.tsx
```diff
@@ -8,7 +8,7 @@ export const Avatar = ({ name, picture, twitterUsername }: AuthorDetails) => (
         src={picture}
         height={32}
         width={32}
-        loading="lazy"
+        loading="eager"
```
mehulkar, line 11: leaving out the attr would be the same too

mehulkar (approved):

docs/components/Avatar.tsx
tknickman: Lazy is actually the default https://nextjs.org/docs/api-reference/next/image#loading

tknickman: (swapped to priority instead)

PR #3939, opened 2023-02-23
ci: Use turbo to run CI

NicholasLYang:


cli/turbo.json
```diff
@@ -2,20 +2,47 @@
...
+    "lint:rust": {
+      "inputs": [
+        "../crates/**/*.rs"
       ]
     },
+    "lint": {
+        "dependsOn": ["lint:go", "lint:rust"]
+    },
+    "test": {
+        "dependsOn": ["test:go", "test:rust"]
```
mehulkar, line 30: always contentious, but could depend on `lint` here if we want to do a quick pass before incurring the test cost

mehulkar, line 30: I would add `e2e` here too. If the `inputs` are setup right, we should get cached results when it doesn't need to re-run

mehulkar (commented):
Another thing we should _probably_ do is to make `create-turbo` and `turbo-ignore` tests depend on `cli#build`. If the binary changes with no other changes to the JS packages, we should re-validate (I don't think we do this right now)

PR #3944, opened 2023-02-24
Automatic babel-loader followup

wbinnssmith:
This:
* Addresses feedback from https://github.com/vercel/turbo/pull/3862
* Exempts VirtualAssets from webpack loaders. This addresses a bug where next-hydrate.tsx could not be read.
	* Adds `ModuleRuleCondition::ResourceIsVirtualAsset` to filter these


crates/turbopack/src/module_options/module_rule.rs
```diff
@@ -23,8 +24,8 @@ impl ModuleRule {
         self.effects.iter()
     }
 
-    pub fn matches(&self, path: &FileSystemPath, reference_type: &ReferenceType) -> bool {
```
jridgewell, line 26: Nit: I think we should still pass in the `&FileSystemPath`, since most rules just need that. We'd be able to skip the async delay for most things (I'm assuming a an async function that doesn't `.await` anything is measurably faster than one that does, even if that `.await` is immediately ready).

crates/next-core/src/babel.rs
```diff
@@ -73,7 +74,9 @@ pub async fn maybe_add_babel_loader(
             };
 
             if !has_babel_loader {
-                if !*(is_babel_loader_available(project_root).await?) {
+                if !has_emitted_babel_resolve_issue
+                    && !*(is_babel_loader_available(project_root).await?)
```
jridgewell, line 78: Is the `*(…)` necessary? I'd assume the `!` knows to deref the value.

crates/turbopack/src/module_options/rule_condition.rs
```diff
@@ -33,15 +38,33 @@ impl ModuleRuleCondition {
...
+    ) -> Result<bool> {
+        Ok(match self {
             ModuleRuleCondition::All(conditions) => {
-                conditions.iter().all(|c| c.matches(path, reference_type))
+                for condition in conditions {
+                    if !condition.matches(source, path, reference_type).await? {
+                        return Ok(false);
+                    }
+                }
+                return Ok(true);
```
jridgewell, line 55: Nit:
```suggestion
                true
```

crates/turbopack/src/module_options/rule_condition.rs
```diff
@@ -33,15 +38,33 @@ impl ModuleRuleCondition {
...
             }
             ModuleRuleCondition::Any(conditions) => {
-                conditions.iter().any(|c| c.matches(path, reference_type))
+                for condition in conditions {
+                    if condition.matches(source, path, reference_type).await? {
+                        return Ok(true);
+                    }
+                }
+                return Ok(false);
```
jridgewell, line 63: Nit:
```suggestion
                false;
```

jridgewell (approved):

PR #3949, opened 2023-02-24
ci(workflow): daily next.js test runs against default branch

kwonoj:
This PR switches daily next.js integration test to use latest default branch (canary) instead of specific published version.


mehulkar (commented):
Could you add web-tooling team as the codeowners for these workflows? Would help reduce some notifications!

kwonoj:
Would this https://github.com/vercel/turbo/pull/3949/commits/a7cda402b208863630243957cee73b34d72109e2 work?

Issue #3943, opened 2023-02-24
[turborepo] [eslint-plugin-turbo] eslint-pugin-turbo can't read jsonc turbo.json


domdomegg:
### What version of Turborepo are you using?

v1.8.2

### What package manager are you using / does the bug impact?

npm

### What operating system are you using?

Mac

### Describe the Bug

As per https://github.com/vercel/turbo/pull/3812, turbo.json can contain comments.

However, `eslint-pugin-turbo` has a rule `turbo/no-undeclared-env-vars` which attempts to find and read a `turbo.json` (to find env vars Turbo is aware of) with JSON.parse, which breaks with comments.

### Expected Behavior

`eslint-plugin-turbo` can handle turbo.json files with comments.

### To Reproduce

Follow turbo setup instructions

Add a comment to your turbo.json, e.g.

```jsonc
// this is a comment
{
  "$schema": "https://turbo.build/schema.json",
  "pipeline": {
  },
}
```

Run `npx turbo lint`

domdomegg:
Looking through the source code to try to fix this, it seems like @tknickman has already made the necessary code changes in https://github.com/vercel/turbo/pull/3846/files#diff-5e060cf84306ac1371af2dda2a506b709e04c0a4b6a11f43df6399b0fceee5cfR62

Looks like it might need a release of eslint-plugin-turbo to get this completed though.

tknickman:
Hey @domdomegg, thanks for the repot. I'm taking a look at this, we should def support reading with jsonc in the plugin

PR #3950, opened 2023-02-24
fix(eslint): better messaging and bug fix (#3950

tknickman:
A few fixes for the eslint plugin. 

1. Fix error message to remove the leading `$` since that was deprecated in 1.5
2. Provides more explicit error messaging 
3. Fixes a bug that could cause the rule to exit early if no global deps were provided
4. Cleans up / standardizes tests and mocks

This should also fix https://github.com/vercel/turbo/issues/3943


tknickman:
Published this as a canary for a few discord members to try out: https://www.npmjs.com/package/eslint-plugin-turbo/v/0.0.9-canary.0

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -76,12 +77,20 @@ function create(context: Rule.RuleContext): Rule.RuleListener {
...
   const filePath = getPhysicalFilename();
-  const allTurboVars =
-    getEnvVarDependencies({
-      cwd,
-    }) || {};
+  const allTurboVars = getEnvVarDependencies({
+    cwd,
+  });
+
+  // if this returns null, something went wrong reading from the turbo config
```
nathanhammond, line 84:
```suggestion
  // if allTurboVars is null, something went wrong reading from the turbo config
```

packages/eslint-plugin-turbo/__fixtures__/configs/single/turbo.json
```diff
@@ -0,0 +1,21 @@
+{
```
nathanhammond, line 1: Needs more comments. 😜

nathanhammond (approved):
🛠🚢

wladiston:
Not sure if the benchmark is testing it but the `turbo/no-undeclared-env-vars` rule seems to be extremely slow on the `0.0.9`. 
Confirming it does work with the new turbo.json under the project's folder 👍 
Screenshot 2023-02-27 at 3 44 02 pm

PR #3952, opened 2023-02-24
feat(turbo-utils): bundle to esm (#3952

tknickman:
Now that we're throwing a few more things in turbo-utils, we should bundle it to esm instead of cjs to support code splitting. This is an internal package, and everything that consumes this bundles to cjs, so compatibility isn't an issue, and this will ensure each package that uses a util from turbo-utils only gets what it needs instead of the entire package.

I tested this locally to see the bundle size difference. 

For turbo-ignore, the bundle size drops from `245K` to `229K`. No _huge_ (~6%) but worth it.

### DIFF
--- a/packages/turbo-utils/package.json
+++ b/packages/turbo-utils/package.json
@@ -15,6 +15,7 @@
   "bugs": {
     "url": "https://github.com/vercel/turbo/issues"
   },
+  "module": "dist/index.mjs",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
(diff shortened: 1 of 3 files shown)


mehulkar (approved):
I'm not familiar... what does `tsup` give you over `esbuild` itself? ([docs homepage](https://tsup.egoist.dev/) says it's powered by esbuild)

Other than that, lgtm!

tknickman:
Just simplicity (it just wraps esbuild). It's great for our examples / internal packages that don't need to be overly complicated.

PR #3955, opened 2023-02-24
Implement faster Rope equality

jridgewell:
This optimizes for more reference equality checks. The old code allowed for this to happen for `Local(Bytes)` checks, but it didn't implement reference checks for `Shared(InnerRope)` shared data. Instead, it'd iterate every `RopeElem` inside the shared rope and perform equality on those.


jridgewell:
Ping @alexkirsz for review, since you were unlucky enough to emoji the benchmarks 😉

crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -22,10 +22,10 @@ static EMPTY_BUF: &[u8] = &[];
 
 /// A Rope provides an efficient structure for sharing bytes/strings between
 /// multiple sources. Cloning a Rope is extremely cheap (Arc and usize), and
-/// the sharing contents of one Rope can be shared by just cloning an Arc.
+/// the sharing contents of one Rope can be done by just cloning an Arc.
```
alexkirsz, line 25:
```suggestion
/// sharing the contents of one Rope can be done by just cloning an Arc.
```

crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -350,6 +350,77 @@ impl<'de> Deserialize<'de> for Rope {
...
+        // If we reach the end of iteration without finding a mismatch (or early
+        // breaking), then we know the ropes are either equal or not equal.
+        if index == len {
+            // We know that any remaining RopeElem in the InnerRope must contain content, so
+            // if either one contains more RopeElem than they cannot be equal.
+            return left.len() == right.len();
+        }
+
+        // At this point, we need to do slower contents equality. It's possible we'll
+        // still get some memory reference quality for Bytes.
```
alexkirsz, line 391:
```suggestion
        // still get some memory reference equality for Bytes.
```

alexkirsz (approved):
:shipit:

PR #1229, opened 2022-05-12
Workspace abstraction

nathanhammond:
What if we moved hashing into the workspace? And also sped things up?

Closes TURBO-1619

PR #3957, opened 2023-02-24
Include expanded inputs config for each task in dry run summary

mehulkar:
This PR adds
- an `expandedInputs` key for each task in the `--dry=json` output. 
- `Inputs Files Considered` in `--dry` text output with a number of files considered. 

    Similar to #3934, only include a count, because the entire list of files could be long.

### Review Notes
I would recommend reviewing one commit at a time, as each commit also includes an explanation, and together there is some refactor noise that can make the actual feature hard to follow.

### Deferred

- Organize `expandedInputs` by the _source config_ (e.g. glob or file)

    This is deferred because we're not sure we want this for v0 Run Summaries and we'll want to do it for `expandedOutputs` as well. Since it will involve reaching into the `globby` package, I think we can do it together to make sure there are no blind spots.

    <details><summary>Example</summary>

    ```jsonc
    {
      // this is output for a single task that has configured `inputs: ['src/**/*.js', 'some-specific-file.txt']`
      // ...all the other keys are omitted for brevity ...
      "expandedInputs": {
        "_turbo_implicit": {
          "package.json": "f2a5d2525f3996a57680180a7cd9ad7310e4dec0",
          "turbo.json.json": "f2a5d2525f3996a57680180a7cd9ad7310e4dec0",
        },
        "src/**/*.js": {
          "src/app/foo.js": "f2a5d2525f3996a57680180a7cd9ad7310e4dec0",
          "src/app/bar.js": "f2a5d2525f3996a57680180a7cd9ad7310e4dec1",
        },
        "some-specific-file.txt": {
          "some-specific-file.txt": "f2a5d2525f3996a57680180a7cd9ad7310e4dec0"
        }
      }
    ```
    If `inputs` wasn't configured, maybe the key would be `_git_tracked` or something like that.
    </details>


cli/internal/taskhash/taskhash.go
```diff
@@ -192,8 +197,9 @@ func (th *Tracker) CalculateFileHashes(
 		hashTasks.Add(pfs)
 	}
 
-	hashes := make(map[packageFileHashKey]string)
-	hashQueue := make(chan *packageFileSpec, workerCount)
+	hashes := make(map[PackageFileHashKey]string)
+	hashObjects := make(map[PackageFileHashKey]map[turbopath.AnchoredUnixPath]string)
```
chris-olszewski, line 202: We could add a size hint of `len(hashTasks)` to these maps: https://github.com/uber-go/guide/blob/master/style.md#specifying-map-capacity-hints

cli/internal/taskhash/taskhash.go
```diff
@@ -204,18 +205,18 @@ func (th *Tracker) CalculateFileHashes(
 
 	for i := 0; i < workerCount; i++ {
 		hashErrs.Go(func() error {
-			for packageFileSpec := range hashQueue {
-				pkg, ok := workspaceInfos.PackageJSONs[packageFileSpec.pkg]
+			for PackageFileSpec := range hashQueue {
```
chris-olszewski, line 208: I think the find/replace went wrong here and we now have a loop variable with the same identifier as the struct which makes it hard to read.

cli/internal/run/dry_run.go
```diff
@@ -304,6 +305,12 @@ func displayDryTextRun(ui cli.Ui, summary *dryRunSummary, workspaceInfos workspa
 		fmt.Fprintln(w, util.Sprintf("  ${GREY}Log File\t=\t%s\t${RESET}", task.LogFile))
 		fmt.Fprintln(w, util.Sprintf("  ${GREY}Dependencies\t=\t%s\t${RESET}", strings.Join(dependencies, ", ")))
 		fmt.Fprintln(w, util.Sprintf("  ${GREY}Dependendents\t=\t%s\t${RESET}", strings.Join(dependents, ", ")))
+
+		count := 0
+		for range task.ExpandedInputs {
+			count = count + 1
+		}
```
chris-olszewski, line 312: Could this be `len(task.ExpandedInputs)`?

cli/internal/graph/graph.go
```diff
@@ -61,8 +64,13 @@ func (g *CompleteGraph) GetPackageTaskVisitor(ctx gocontext.Context, visitor fun
 			TaskDefinition:  taskDefinition,
 			Outputs:         taskDefinition.Outputs.Inclusions,
 			ExcludedOutputs: taskDefinition.Outputs.Exclusions,
+			ExpandedInputs:  make(map[turbopath.AnchoredUnixPath]string),
 		}
 
+		pfs := taskhash.SpecFromPackageTask(packageTask)
+		expandedInputs := g.TaskHashTracker.PackageInputsExpandedHashes[pfs.ToKey()]
+		packageTask.ExpandedInputs = expandedInputs
```
chris-olszewski, line 72: I think we're exposing a lot of `taskhash` internals just for these three lines. I think we can keep all of the internals private if we add a method to the tracker that does this:
```
func (th *Tracker) GetExpandedInputs(packageTask *nodes.PackageTask) map[turbopath.AnchoredUnixPath]string {
	pfs := specFromPackageTask(packageTask)
	expandedInputs := th.packageInputsExpandedHashes[pfs.ToKey()]
	inputsCopy := make(map[turbopath.AnchoredUnixPath]string, len(expandedInputs))
	for path, hash := range expandedInputs {
		inputsCopy[path] = hash
	}

	return inputsCopy
}
```

If something down the line requires exposing all of these internals then these are some follow ups:
 - I really don't know what the "Go" way is, but should we make `ExpandedInputs` a pointer so we're more explicit that the map is shared between multiple structs? Either that or we should copy the map here: https://github.com/uber-go/guide/blob/master/style.md#copy-slices-and-maps-at-boundaries
 - From reading through the code I think we're okay with `packageTask.ExpandedInputs` being `nil` so we can have the assignment on the same line as the lookup instead of using a local variable.

chris-olszewski (approved):
I think we can keep a lot of the `taskhash` stuff private with a few getter methods and I think that's the right way to go about it. If this makes run summary harder, then ignore my comments 😄 

`Catalog` naming is a bit weird, but naming is hard when you can't use `Workspace` and it's not that important so feel free to ignore my comment.

cli/internal/taskhash/taskhash.go
mehulkar: done (also added to the line above this one)

cli/internal/run/dry_run.go
mehulkar: 🤦🏾 .  I don't know what happened here. I searched for how to do this and might have just ignored the results entirely...?

CleanShot 2023-02-27 at 22 04 46@2x

cli/internal/graph/graph.go
mehulkar: Great idea. The public method makes more sense. I can always expose pieces for future work, but I don't think we should do it here either

cli/internal/taskhash/taskhash.go
mehulkar: reverted, good catch

cli/internal/workspace/workspace.go
```diff
@@ -0,0 +1,10 @@
+// Package workspace contains some utilities around managing workspaces
+package workspace
+
+import "github.com/vercel/turbo/cli/internal/fs"
+
+// Catalog holds information about each workspace in the monorepo.
+type Catalog struct {
```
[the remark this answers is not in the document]
mehulkar: I think `Info` is a bit generic. I was thinking that maybe in the future, we could make this struct more like a single `map[string]Workspace`, and follow through on https://github.com/vercel/turbo/pull/1229. Does Catalog feel more intuitive with that in mind? (`workspace.Group` or `Workspaces` also didn't feel good enough to me). Of course, this abstraction might not actually go any further happen till Rust, but I needed _something_. 

Since you approved already, I'll let this one be and ping you in chat to see if we can come up with a better name

mehulkar:
> The following steps have failed in CI:
> 
> * Go examples

These have been broken since #3946
> * Go e2e tests

This is the windows one that timed out, it does that sometimes


Merging this PR!

PR #3974, opened 2023-02-27
Split the ES chunk module into smaller, more focused modules

alexkirsz:
This diff splits the `turbopack-ecmascript/src/chunk/mod.rs` and `turbopack-ecmascript/src/loader.rs` modules into smaller modules, usually centered around one type.

The goal of this PR is to make the ecmascript crate easier to navigate and review, as well as reduce the amount of merge conflicts that can arise from modifying imports in two separate diffs within the same module.

This PR also changes the visibility of most modules within the crate to be `pub(crate)` so they're available to the generated `register` function, but not accidentally leaked outside of the crate. The same is done on a per-type basis with `pub(crate)` or `pub(super)`.

PR #3975, opened 2023-02-27
Merge `EcmascriptChunkUpdate`s before sending them to the client

alexkirsz:
This diff:
* introduces the `VersionedContentMerger` trait, which allows for merging the updates of versioned contents within the same chunk group;
* implements this for `EcmascriptChunkContent`/`EcmascriptChunkUpdate`, turning them into `EcmascriptMergedChunkContent`/`EcmascriptMergedChunkUpdate`;
* creates a new `ChunkList` asset which is capable of merging chunk updates of chunks within the same chunk group, and create such an asset for dynamic chunks (through manifest/loader_item.rs) and chunk group files assets.

This fixes a bunch of edge cases related to HMR:
* HMR of dynamic imports now works;
* Chunks getting added/deleted/renamed now also works with HMR, since we're listening to updates at the chunk group level;
* CSS chunks get reloaded in the right order, with respect for precedence.

There are still known edge cases with HMR:
* CSS chunks added through HMR are not inserted at the right position to respect precedence (WEB-652).
* Update aggregation is disabled because we don't want a critical issue to stop all HMR (WEB-582) . This would be fixed by applying aggregated updates when dismissing the error modal, but there are some edge cases with this too (e.g. what happens when an HMR update also causes an error on top of an existing error).


alexkirsz:
* main
  * **PR #3887** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3887" target="_blank">Graphite</a> 
    * **PR #3974** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3974" target="_blank">Graphite</a> 
      * **PR #3975** <a href="https://app.graphite.dev/github/pr/vercel/turbo/3975" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/3975?utm_source=stack-comment).

crates/next-core/src/page_loader.rs
```diff
@@ -109,17 +111,28 @@ impl Asset for PageLoaderAsset {
         let this = &*self_vc.await?;
 
         let chunks = self_vc.get_page_chunks().await?;
-
-        let mut data = Vec::with_capacity(chunks.len());
-        for chunk in chunks.iter() {
-            let path = chunk.path().await?;
-            data.push(Value::String(path.path.clone()));
-        }
+        let server_root = this.server_root.await?;
```
alexkirsz, line 114: The way the page loader asset computed paths was incorrect before, as it wasn't basing them upon the server root path. But this doesn't really have any measurable effect today as we don't support base path.

crates/next-core/js/src/dev/hmr-client.ts
```diff
@@ -100,116 +102,334 @@ function handleSocketConnected() {
...
-  for (const [moduleId, entry] of Object.entries(msg.instruction.added)) {
-    const removedDeleted = aggregated.deleted.delete(moduleId);
-    if (aggregated.modified[moduleId] != null) {
-      console.error(
-        `impossible state aggregating updates: module "${moduleId}" was added, but previously modified`
-      );
-      location.reload();
+  aggregated = {
+    resource: msg.resource,
+    update: mergeChunkListUpdates(aggregated.update, msg.instruction),
```
alexkirsz, line 145: Since the HMR updates are now very different (from a simple "EcmascriptChunkUpdate" to a "ChunkListUpdate" that can contain simple chunk updates and complex "EcmascriptMergedChunkUpdate"s), I cut the merging logic into smaller functions that each operate at a specific nested-type's level.

crates/next-core/js/src/dev/hmr-client.ts
```diff
@@ -282,9 +502,14 @@ function handleSocketMessage(msg: ServerMessage) {
   sortIssues(msg.issues);
 
   const hasCriticalIssues = handleIssues(msg);
-  const aggregatedMsg = aggregateUpdates(msg, hasCriticalIssues);
 
-  const runHooks = chunksWithUpdates.size === 0;
+  // TODO(WEB-582) Disable update aggregation for now.
```
alexkirsz, line 506: This was already the case, but I made it explicit here.

crates/next-core/js/src/dev/hmr-client.ts
```diff
@@ -362,46 +597,3 @@ function triggerUpdate(msg: ServerMessage) {
...
-        link.replaceWith(link);
-        break;
-      }
-      case "partial":
-        throw new Error(`partial CSS chunk updates are not supported`);
-      default:
-        throw new Error(`unknown update type \`${update}\``);
-    }
-  });
-}
```
alexkirsz, line 407: CSS updates are now handled at the chunk list level, using the `BACKEND.loadChunk/unloadChunk/reloadChunk` logic.

crates/next-core/js/src/entry/app/index.d.ts
```diff
@@ -5,3 +5,4 @@ export = Anything;
 
 export const __turbopack_module_id__: string | number;
```
sokra, line 6: We don't need this file anymore since we don't have "." imports anymore?

alexkirsz: Indeed. Instead, we could declare our "PAGE"/"INNER_ASSETS" modules (probably would need to rename them to "@vercel/inner/page" or "@turbopack/inner/page").

crates/turbopack-core/src/chunk/list/content.rs
```diff
@@ -0,0 +1,122 @@
...
+            .map(|(merger, contents)| {
+                let merger = merger;
+                async move {
+                    Ok((
+                        merger,
+                        merger
+                            .merge(VersionedContentsVc::cell(contents))
+                            .version()
+                            .into_trait_ref()
+                            .await?,
```
sokra, line 93: We always merge the version of all inner contents... That sounds pretty expensive.

crates/next-core/js/types/globals.d.ts
```diff
@@ -1,6 +1,10 @@
 declare global {
   function __turbopack_require__(name: any): any;
   function __turbopack_load__(path: string): any;
+  function __turbopack_register_chunk_list__(
+    chunkListPath: string,
+    chunksPaths: string[]
+  ): any;
```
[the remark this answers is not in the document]
alexkirsz: I think `__turbopack_load__` is still useful on its own for loading separate assets which don't have a chunk list.

sokra (approved):

PR #3983, opened 2023-02-27
webpack loaders: implement `emitError` and `emitWarning`

wbinnssmith:
This implements webpack loader context api for `emitError` and `emitWarning`, each of which emit Turbopack issues at the appropriate error level. As with webpack’s implementation, `emitError` does _not_ cause builds to fail [0].

This uses next-dev issue snapshots (#3774) to test that issues are emitted.

[0] https://webpack.js.org/api/loaders/#thisemiterror


crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -98,3 +104,17 @@ const transform = (
 };
 
 export { transform as default };
+
+function makeErrorEmitter(severity: "warning" | "error", ipc: Ipc) {
+  return function (error: Error | string) {
+    const errorObject = error instanceof Error ? error : new Error(error);
```
jridgewell, line 110: The stack should be omitted if this isn't an error, so we don't confuse our creation site with the throw site.

crates/turbopack-node/src/lib.rs
```diff
@@ -249,10 +250,25 @@ enum EvalJavaScriptOutgoingMessage<'a> {
...
+        glob: String,
+    },
+    JsonValue {
+        data: String,
+    },
+    EmittedError {
+        severity: IssueSeverity,
+        name: String,
+        message: String,
+        stack: Vec<StackFrame>,
```
jridgewell, line 270: Nit: these 3 fields should just use the `StructuredError` struct so we can share deserialize code.

crates/turbopack-node/js/src/ipc/evaluate.ts
```diff
@@ -23,6 +23,15 @@ type IpcOutgoingMessage =
...
       glob: string;
+    }
+  | {
+      type: "emittedError";
+      severity: "warning" | "error";
+      error: {
+        name: string;
+        message: string;
+        stack: string[];
+      };
```
jridgewell, line 34: We have a `StructuredError` in `crates/turbopack-node/js/src/ipc/index.ts`

crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -98,3 +104,17 @@ const transform = (
...
+function makeErrorEmitter(severity: "warning" | "error", ipc: Ipc) {
+  return function (error: Error | string) {
+    ipc.send({
+      type: "emittedError",
+      severity: severity,
+      error: {
+        name: error instanceof Error ? error.name : "Error",
+        message: error instanceof Error ? error.message : error,
+        stack: error instanceof Error ? parseStackTrace(error.stack) : [],
+      },
```
jridgewell, line 117: Nit:
```suggestion
        error instanceof Error
          ? {
              name: error.name,
              message: error.message,
              stack: parseStackTrace(error.stack),
            }
          : {
              name: "Error",
              message: error,
              stack: [],
            },
```

crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -98,3 +104,17 @@ const transform = (
...
 export { transform as default };
+
+function makeErrorEmitter(severity: "warning" | "error", ipc: Ipc) {
+  return function (error: Error | string) {
+    ipc.send({
+      type: "emittedError",
+      severity: severity,
+      error: {
+        name: error instanceof Error ? error.name : "Error",
+        message: error instanceof Error ? error.message : error,
```
jridgewell, line 115: Maybe this is the bug here?
```suggestion
        message: error instanceof Error ? error.message : String(error),
```

jridgewell (approved):

PR #3987, opened 2023-02-27
chore(repo): add PR template

tknickman:
Add a PR template to encourage adding better descriptions with reproduction steps


tknickman:
I don't feel as if that's a bad thing, but I don't feel strongly enough to not make the change 😄  - will remove!

mehulkar (approved):

PR #3991, opened 2023-02-27
chore: add rust ffi integration tests to CI

chris-olszewski:
As we write Rust code called by Go via FFI we should run integration tests against it. Once we delete the Go code we can delete the `go_integration` job in favor of the Rust FFI one

Waiting to merge until #3996 to reduce noise on PRs

Test plan: Make sure that integration tests run on CI


.github/workflows/test.yml
```diff
@@ -267,6 +267,35 @@ jobs:
...
+
+      - name: Cache Prysk
+        id: cache-prysk
+        uses: actions/cache@v3
+        with:
+          path: cli/.cram_env
+          key: prysk-venv-${{ matrix.os.name }}
+
+      - name: Integration Tests
+        run: GO_TAG=rust pnpm -- turbo run integration-tests --filter=cli
```
NicholasLYang, line 297: Should we include `GO_TAG` in the integration-tests task definition?

chris-olszewski: Good catch!

mehulkar (commented):
- Does this need to be its own job or can we just make the regular integration-tests do this?
- Any way we can move this a shared action and make the only diff here the env var setting to prevent drift?

chris-olszewski:
> * Does this need to be its own job or can we just make the regular integration-tests do this?

It can be, that just means the integration test job will be longer running since we'll be rebuilding turbo serially. The Rust cache should stay around and save us some time, but I'll verify.

>  * Any way we can move this a shared action and make the only diff here the env var setting to prevent drift?

I didn't go down this path just because this should hopefully be a short lived (we kill of the Go code path before 1.9).

mehulkar (approved):

arlyon (approved):

PR #3992, opened 2023-02-27
feat(codemod): support workspace configs

tknickman:
This fixes turbo codemod not building due to the new config schema introduced for workspace configs. 

It also adds workspace config support and tests to the codemods that are relevant (env vars and outputs). This isn't strictly necessary as workspace configs were introduced after the version that required these codemods, but I wanted to add this just in case someone uses an old example from around the web to update a workspace config or similar.


packages/turbo-codemod/src/transforms/set-default-outputs.ts
```diff
@@ -51,17 +64,20 @@ export function transformer({
...
-  }
-
   runner.modifyFile({
     filePath: turboConfigPath,
-    after: turboJson,
+    after: migrateConfig(turboJson),
+  });
+
+  // find and migrate any workspace configs
+  // find and migrate any workspace configs
```
chris-olszewski, line 73:
```suggestion
```

chris-olszewski (approved):

packages/turbo-codemod/src/transforms/set-default-outputs.ts
tknickman: Good catch, thanks!

PR #3993, opened 2023-02-27
fix: make inputs relative to workspace root

chris-olszewski:
Inputs are relative to workspace and not repo root so we need to drop the `cli` prefix here. I removed the `shim/**` input as that directory no longer exists and to correctly capture these inputs we should Turbo-fy our crates instead of listing out all of the directories out here. Also, I'm not sure how we feel about a task depending on a file outside of the workspace.

Test plan:
```
pnpm -- turbo run e2e --filter=cli
echo "console.log('big change')" >> cli/scripts/e2e/e2e.ts
pnpm -- turbo run e2e --filter=cli
```
should no longer produce a cache hit on the second run

### DIFF
--- a/cli/turbo.json
+++ b/cli/turbo.json
@@ -9,20 +9,18 @@
     "e2e": {
       "outputs": [],
       "inputs": [
-        "cli/**/*.go",
-        "cli/go.mod",
-        "cli/go.sum",
-        "cli/scripts/e2e/e2e.ts",
-        "shim/**/*.rs"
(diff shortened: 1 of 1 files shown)


mehulkar (approved):
Nice catch. Was this broken before Workspace Configs too, or did I mess up when I ported?

chris-olszewski:
> Nice catch. Was this broken before Workspace Configs too, or did I mess up when I ported?

Broken before your port

PR #3996, opened 2023-02-28
fix: scm make patterns relative

chris-olszewski:
Since the Go `ChangedFiles` used the command line `git` it appears that the CLI did some light data processing and if we passed the absolute repo root as the pathspec via CLI we wouldn't have issue. It looks like `libgit2` doesn't do this and expects well formed pathspecs so we shouldn't be passing in absolute paths.

This PR will attempt to remove the repo root prefix from `relative_to`. An alternative would be to change Go's behavior to no longer pass the repo root as the default, but that's a larger change and the Rust and Go implementations of `ChangedFiles` would still differ in the same way, we'd just avoid hitting this behavior in turbo.

Test plan:
`GO_TAG=rust make integration-tests` now passes as well as the new unit test


NicholasLYang:
Only question I have is if the Go version does any normalization so that if you pass `~/repos/turbo` say for repo_root and `/Users/Nicholas/repos/turbo` for relative_to, this still works?

chris-olszewski:
> Only question I have is if the Go version does any normalization so that if you pass ~/repos/turbo say for repo_root and /Users/Nicholas/repos/turbo for relative_to, this still works?


The paths we get are all `turbopath.AbsoluteSystemPath` ([scm construction](https://github.com/vercel/turbo/blob/main/cli/internal/scm/scm.go#L50), [usage in scope](https://github.com/vercel/turbo/blob/main/cli/internal/scope/scope.go#L202)) so I don't think that's ever used by Go.

NicholasLYang (approved):

arlyon (approved):

arlyon:
I am going to drop a ticket on linear for adding some types around relative and absolute paths, but I think this can merge.

PR #2428, opened 2022-10-27
Fixes #2280 adds a client timeout

tm1000:
This fixes #2280 by adding a client timeout config option.

This is my *first* attempt at go, FYI. So I'll take any feedback you can give me (and I'm also going to have my coworkers who work in Go daily give me some feedback too)

All tests passed and so did the e2e tests

PR #3998, opened 2023-02-28
Client timeout

### LINKED ISSUE #2096: Remote Caching: Unhandled HTTP Exception & Client Timeout Feature Request

1.5.1

pnpm

Mac

### Describe the Bug

While implementing a remote cache service, we were running into issues with some cache objects never being uploaded despite the logs reporting a successful write. Server-side, however, no http requests were coming through.

Screen Shot 2022-09-22 at 7 58 54 PM

I recompiled `turbo` locally after adding some additional logging [here](https://github.com/vercel/turborepo/blob/02620b88771b7dc90d3575e6119b3e325fa04a62/cli/internal/client/client.go#L208) (as reflected in the `RESPONSE STATUS` output above)

Currently, the only exception handling is for `403` errors

```
	resp, err := c.HttpClient.Do(req)
	if err != nil {
		return fmt.Errorf("failed to store files in HTTP cache: %w", err)
	}
	defer func() { _ = resp.Body.Close() }()
	if resp.StatusCode == http.StatusForbidden {
		return c.handle403(resp.Body)
	}
```

It would be great if non `200` status codes could likewise be handled/reported in the logs. The `413` issue we could resolve internally, as that was just an `nginx` ingress configuration, but the only way we were able to successfully leverage remote caching for larger cache objects was by changing the [client timeout here](https://github.com/vercel/turborepo/blob/02620b88771b7dc90d3575e6119b3e325fa04a62/cli/internal/client/client.go#L77).

```
	client := &ApiClient{
		baseUrl:      remoteConfig.APIURL,
		turboVersion: turboVersion,
		HttpClient: &retryablehttp.Client{
			HTTPClient: &http.Client{
				Timeout: time.Duration(customTimeout * time.Second),
			},
                         ...
		},
```

This could be made available as part of an official release in one of two ways - either providing the flexibility to set a client timeout (perhaps by environment variable or flag) which would fall back to the default of `20` if it's not set or ensuring no http retry is initiated if the client connection hasn't been closed from the server/an upload is actively in progress. The former would likely be the most straightforward approach for this particular edge case.

I am happy to open a pull request for whatever implementation is preferred if you would be open to adding this feature!

### Expected Behavior

- http errors would be logged and reported by the command line when there are failures uploading to the remote cache
- `writing to cache...` would not appear after failed writes
- `success` would not be the final status if there were any unsuccessful writes
- pending requests would not be suspended in favor of a retry and/or client timeouts could be modified so as to prevent retries that suspend uploads prior to completion

### To Reproduce

This is currently reproducible if you're working with especially large cache outputs (💫  monorepos), but could also potentially be reproduced by synthetically throttling your network connection to simulate a timeout during a remote cache write. 

You can add additional logging to `stdout` in `cli/internal/client/client.go` as further confirmation:

```
	resp, err := c.HttpClient.Do(req)
	if err != nil {
		return fmt.Errorf("failed to store files in HTTP cache: %w", err)
	}

	defer func() { _ = resp.Body.Close() }()
	if resp.StatusCode == http.StatusForbidden {
		return c.handle403(resp.Body)
	}

        if resp.StatusCode != http.StatusOK {
	       fmt.Println("RESPONSE STATUS: ", resp.Status, " URL ", requestURL)
        }
```

As we're running an internal remote cache service, we could also confirm server-side no `PUT` requests were coming through for artifacts reported by the `turbo` cli to have successfully written by referencing the server logs, and that this was only resolved by recompiling the cli with an updated `client` timeout to accommodate those larger cache objects.

nathanhammond:
This commit authored in part by @tm1000.

Closes #1993
Closes #2096
Closes #2428

Notes:
- You may want to look at the original diff: https://github.com/vercel/turbo/pull/2428/files?w=1
- I've completely removed `pflag` at this point. It was only being used in tests.
- I've added a new "config" in Rust land, for the API client. It only accepts `timeout` right now.

### DIFF
--- a/cli/integration_tests/no_args.t
+++ b/cli/integration_tests/no_args.t
@@ -19,22 +19,22 @@ Make sure exit code is 2 when no args are passed
     unlink      Unlink the current directory from your Vercel organization and disable Remote Caching
   
   Options:
-        --version                   
-        --skip-infer                Skip any attempts to infer which version of Turbo the project is configured to use
-        --no-update-notifier        Disable the turbo update notification
-        --api <API>                 Override the endpoint for API calls
-        --color                     Force color usage in the terminal
-        --cpuprofile <CPU_PROFILE>  Specify a file to save a cpu profile
-        --cwd <CWD>                 The directory in which to run turbo
-        --heap <HEAP>               Specify a file to save a pprof heap profile
-        --login <LOGIN>             Override the login endpoint
-        --no-color                  Suppress color usage in the terminal
-        --preflight                 When enabled, turbo will precede HTTP requests with an OPTIONS request for authorization
-        --team <TEAM>               Set the team slug for API calls
-        --token <TOKEN>             Set the auth token for API calls
-        --trace <TRACE>             Specify a file to save a pprof trace
-        --verbosity <COUNT>         Verbosity level
-    -h, --help                      Print help
+        --version                         
+        --skip-infer                      Skip any attempts to infer which version of Turbo the project is configured to use
+        --api <API>                       Override the endpoint for API calls
+        --color                           Force color usage in the terminal
+        --cpuprofile <CPU_PROFILE>        Specify a file to save a cpu profile
+        --cwd <CWD>                       The directory in which to run turbo
+        --heap <HEAP>                     Specify a file to save a pprof heap profile
+        --login <LOGIN>                   Override the login endpoint
+        --no-color                        Suppress color usage in the terminal
+        --preflight                       When enabled, turbo will precede HTTP requests with an OPTIONS request for authorization
+        --remote-cache-timeout <TIMEOUT>  Set a timeout for all HTTP requests
+        --team <TEAM>                     Set the team slug for API calls
+        --token <TOKEN>                   Set the auth token for API calls
+        --trace <TRACE>                   Specify a file to save a pprof trace
+        --verbosity <COUNT>               Verbosity level
+    -h, --help                            Print help
   
   Run Arguments:
         --cache-dir <CACHE_DIR>          Override the filesystem cache directory
--- a/cli/integration_tests/no_args.t
+++ b/cli/integration_tests/no_args.t
@@ -21,6 +21,7 @@ Make sure exit code is 2 when no args are passed
   Options:
         --version                         
         --skip-infer                      Skip any attempts to infer which version of Turbo the project is configured to use
+        --no-update-notifier              Disable the turbo update notification
         --api <API>                       Override the endpoint for API calls
         --color                           Force color usage in the terminal
         --cpuprofile <CPU_PROFILE>        Specify a file to save a cpu profile
--- a/cli/internal/cmdutil/cmdutil.go
+++ b/cli/internal/cmdutil/cmdutil.go
@@ -176,17 +176,17 @@ func (h *Helper) GetCmdBase(cliConfig *turbostate.ParsedArgsFromRust) (*CmdBase,
 	}
 
 	// Primacy: Arg > Env
-	val, ok := os.LookupEnv("TURBO_REMOTE_CACHE_TIMEOUT")
-	if ok {
-		number, err := strconv.ParseUint(val, 10, 64)
-		if err == nil {
-			h.clientOpts.Timeout = number
-		}
-	}
-
 	timeout, err := cliConfig.GetRemoteCacheTimeout()
 	if err == nil {
 		h.clientOpts.Timeout = timeout
+	} else {
+		val, ok := os.LookupEnv("TURBO_REMOTE_CACHE_TIMEOUT")
+		if ok {
+			number, err := strconv.ParseUint(val, 10, 64)
+			if err == nil {
+				h.clientOpts.Timeout = number
+			}
+		}
 	}
 
 	apiClient := client.NewClient(
--- a/cli/internal/cmdutil/cmdutil.go
+++ b/cli/internal/cmdutil/cmdutil.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"strconv"
 	"sync"
 
 	"github.com/hashicorp/go-hclog"
@@ -183,6 +184,22 @@ func (h *Helper) GetCmdBase(cliConfig config.CLIConfigProvider) (*CmdBase, error
 			remoteConfig.TeamID = vercelArtifactsOwner
 		}
 	}
+
+	val, ok := os.LookupEnv("TURBO_REMOTE_CACHE_TIMEOUT")
+	if ok {
+		number, err := strconv.ParseUint(val, 10, 64)
+		if err == nil {
+			h.clientOpts.Timeout = number
+		} else {
+			terminal.Warn(fmt.Sprintf("%s Unable to parse TURBO_REMOTE_CACHE_TIMEOUT: %s", ui.WARNING_PREFIX, prefix, color.YellowString(" %v", err)))
+		}
+	}
+
+	timeout, err := flags.GetUint64("remote-cache-timeout")
+	if flags.Lookup("remote-cache-timeout").Changed && err == nil {
+		h.clientOpts.Timeout = timeout
+	}
+
 	apiClient := client.NewClient(
 		remoteConfig,
 		logger,
--- a/cli/internal/cmdutil/cmdutil.go
+++ b/cli/internal/cmdutil/cmdutil.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/mitchellh/cli"
-	"github.com/spf13/pflag"
 	"github.com/vercel/turbo/cli/internal/client"
 	"github.com/vercel/turbo/cli/internal/config"
 	"github.com/vercel/turbo/cli/internal/fs"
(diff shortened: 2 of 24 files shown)


crates/turborepo-lib/src/config/client.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    #[allow(dead_code)]
+    pub fn load(self) -> Result<ClientConfig> {
+        let Self {
+            remote_cache_timeout,
+        } = self;
+
+        let config: ClientConfigValue = Config::builder()
+            .add_source(Environment::with_prefix("turbo"))
```
nathanhammond, line 52: I don't know how to test this on the Rust side.

I want to ensure that the value is loaded from either `TURBO_REMOTE_CACHE_TIMEOUT` or `--remote-cache-timeout` with the correct primacy.

crates/turborepo-lib/src/config/client.rs
```diff
@@ -0,0 +1,59 @@
...
+            .set_override_option("remote_cache_timeout", remote_cache_timeout)?
+            .build()?
+            .try_deserialize()?;
+
+        Ok(ClientConfig { config })
+    }
+}
```
nathanhammond, line 59: This very studiously ports the `RepoConfig`/UserConfig` pattern over to a new `ClientConfig`.

It's probably overkill, but I decided that the pattern was more important than the actual contents.

crates/turborepo-lib/src/client.rs
```diff
@@ -264,9 +264,9 @@ impl APIClient {
         false
     }
 
-    pub fn new(base_url: impl AsRef<str>) -> Result<Self> {
+    pub fn new(base_url: impl AsRef<str>, timeout: u64) -> Result<Self> {
         let client = reqwest::Client::builder()
-            .timeout(std::time::Duration::from_secs(20))
+            .timeout(std::time::Duration::from_secs(timeout))
```
nathanhammond, line 269: No more magic number!

cli/integration_tests/no_args.t
```diff
@@ -19,21 +19,22 @@ Make sure exit code is 2 when no args are passed
...
+        --heap <HEAP>                     Specify a file to save a pprof heap profile
+        --login <LOGIN>                   Override the login endpoint
+        --no-color                        Suppress color usage in the terminal
+        --preflight                       When enabled, turbo will precede HTTP requests with an OPTIONS request for authorization
+        --remote-cache-timeout <TIMEOUT>  Set a timeout for all HTTP requests
+        --team <TEAM>                     Set the team slug for API calls
+        --token <TOKEN>                   Set the auth token for API calls
+        --trace <TRACE>                   Specify a file to save a pprof trace
+        --verbosity <COUNT>               Verbosity level
+    -h, --help                            Print help
```
arlyon, line 37: What is the cause of all these diffs? They look visually identical and it doesn't look like a tabs / spaces issue? CLRF?

cli/internal/cmdutil/cmdutil.go
```diff
@@ -183,6 +174,21 @@ func (h *Helper) GetCmdBase(cliConfig config.CLIConfigProvider) (*CmdBase, error
 			remoteConfig.TeamID = vercelArtifactsOwner
 		}
 	}
+
+	// Primacy: Arg > Env
+	val, ok := os.LookupEnv("TURBO_REMOTE_CACHE_TIMEOUT")
```
arlyon, line 179: at the risk of extra nesting we can prevent duplicate work here by moving it into the else clause for 188

arlyon (approved):

tm1000:
I'm unfortunately unable to test this, when I attempt to build turborepo I get the following error:

```
root@development:/usr/src/turbo/cli# make turbo
cd ../crates/turborepo-ffi && CARGO_TARGET_DIR=./target cargo build --release
   Compiling turborepo-ffi v0.1.0 (/usr/src/turbo/crates/turborepo-ffi)
error: failed to run custom build command for `turborepo-ffi v0.1.0 (/usr/src/turbo/crates/turborepo-ffi)`

Caused by:
  process didn't exit successfully: `/usr/src/turbo/crates/turborepo-ffi/./target/release/build/turborepo-ffi-085e24f92bcaa906/build-script-build` (exit status: 1)
  --- stderr
  Error: Custom { kind: Other, error: "protoc failed: messages.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.\n" }
make: *** [Makefile:41: turborepo-ffi] Error 101
```

crates/turborepo-lib/src/config/client.rs
NicholasLYang: You can use `env::set_var` to set `TURBO_REMOTE_CACHE_TIMEOUT`, and also pass in a manual value with `with_remote_cache_timeout`. Then check that the correct one is selected. The `UserConfig` struct has similar tests

NicholasLYang (dismissed):
Other than a test for `ClientConfig` it looks great to me! Glad we can delete this Go code.

crates/turborepo-lib/src/config/client.rs
nathanhammond: I added a bunch of tests to make sure I knew how the system was operating. I'mma re-request review to make sure you like the API shape.

crates/turborepo-lib/src/client.rs
```diff
@@ -264,10 +264,13 @@ impl APIClient {
...
-        let client = reqwest::Client::builder()
-            .timeout(std::time::Duration::from_secs(20))
-            .build()?;
+    pub fn new(base_url: impl AsRef<str>, timeout: Option<u64>) -> Result<Self> {
+        let client = match timeout {
+            Some(timeout) => reqwest::Client::builder()
+                .timeout(std::time::Duration::from_secs(timeout))
+                .build()?,
+            None => reqwest::Client::builder().build()?,
+        };
```
nathanhammond, line 273: This is a material difference from an earlier iteration.

In Go if you pass "0" into things you get no timeout. In Rust to do that you have to avoid call `.timeout()`, so I match on an `Option` instead.

crates/turborepo-lib/src/config/client.rs
```diff
@@ -0,0 +1,235 @@
...
+        match self.config.remote_cache_timeout {
+            // Pass 0 to get no timeout.
+            Some(0) => None,
+
+            // Pass any non-zero uint64 to get a timeout of that duration measured in seconds.
+            Some(other) => Some(other),
+
+            // If the _config_ doesn't have a remote_cache_timeout, give them the default.
+            None => Some(DEFAULT_TIMEOUT),
+        }
```
nathanhammond, line 37: More of the implementation for the "0 is no timeout".

nathanhammond (commented):
Material changes from the previous iteration.

Diff of those changes:
https://github.com/vercel/turbo/pull/3998/files/34562e757c562aad6a2ed280a01275f08921cc98..e4c26300db7d7c8d96e6f4ddaa59ae818a0c3a46

crates/turborepo-lib/src/config/client.rs
```diff
@@ -17,12 +19,22 @@ struct ClientConfigValue {
 #[derive(Debug, Clone)]
 pub struct ClientConfigLoader {
     remote_cache_timeout: Option<u64>,
+    environment: Option<HashMap<String, String>>,
 }
 
 impl ClientConfig {
     #[allow(dead_code)]
```
NicholasLYang, line 26: Y'know I'm not entirely sure why we have these everywhere in config.

crates/turborepo-lib/src/config/client.rs
```diff
@@ -42,18 +55,181 @@ impl ClientConfigLoader {
...
+            .set_default("remote_cache_timeout", DEFAULT_TIMEOUT)?
+            .add_source(Environment::with_prefix("turbo").source(environment))
             .set_override_option("remote_cache_timeout", remote_cache_timeout)?
             .build()?
-            .try_deserialize()?;
+            .try_deserialize();
+
+        // This goes wrong when TURBO_REMOTE_CACHE_TIMEOUT can't be deserialized to u64
+        match config_attempt {
+            Err(_) => Ok(ClientConfig {
```
NicholasLYang, line 80: Should we log something here to inform users that we're not using the env variable?

NicholasLYang (approved):

crates/turborepo-lib/src/config/client.rs
nathanhammond: It's an interesting _bonus_ method of setting these things. It makes the most sense for concurrent testing strategies so I actually think it's worth keeping.

crates/turborepo-lib/src/config/client.rs
nathanhammond: Yes, but presently we have no logging anywhere in our config modules. Given that, I'm not going to add this here because we should come up with a consistent strategy for this.

nathanhammond: TURBO-857

nathanhammond:
Testing of a local build appears to behave correctly (if a bit surprising) because the timeout is per-request and it's not immediately clear how many network requests any particular command will trigger. For an intentionally broken network like the one I was testing on it's more-surprising: each serial request must time out in turn. For a normal slow network the behavior will be much-more understandable, only impacting requests that are _actually_ slow like uploads and downloads.

tm1000:
You all are the best. Can't wait for the next release!

PR #2650, opened 2022-11-09
feat(examples): add non-monorepo starter

tknickman:
# Add a non-monorepo example for 1.6

Steps to create
1. `npx create-next-app@latest`
1. `pnpm install next@latest react@latest react-dom@latest eslint-config-next@latest` (to use 13)
1. `turboify™` (add `turbo.json`, add `.turbo` to `.gitignore`)
1. Docs / README / Blog updates

PR #4000, opened 2023-02-28
Convert examples tests to run through prysk

mehulkar:
https://github.com/vercel/turbo/pull/3946 seems to have broken the way our `run-examples.sh` script sets up an `example/` directory to run tests in. I took this opportunity to exercise the same tests through prysk instead, because prysk sets up an isolated directory (`$PWD`) for us by default that we can setup. All our other integration tests also run this way. 

This PR attempts to capture `run-examples` the way it is run today. The goal is to test that some of the `examples/` directories will work out of box with the `@latest` and `@canary` published versions. 

There is a lot more we can improve here if we continue along these lines:

- `smoke-published` could use the same strategy with `create-turbo`
- Existing `cli/integration_tests` could write tests against `examples/` (they will be slower becase they require `npm i`, but it's something to explore
- `Smoke Test` _before_ publish could exercise these example_tests


.github/workflows/test.yml
```diff
@@ -152,9 +152,7 @@ jobs:
         with:
           PATTERNS: |
             examples/**
-            cli/**/*.go
-            cli/go.mod
-            cli/go.sum
```
mehulkar, line 157: I removed trigger on `cli/**`, because the script uses the `turbo` installed in the example, rather than building locally. So PRs that change how turbo works isn't really getting tested against examples. We may still want to do that in a daily cron job though. This setup will make it easier to do so.

arlyon: Makes sense

arlyon (approved):

nathanhammond (changes requested):
We can convert to prysk, but at the same time the change here indicates a behavior change we need to address. Let's prep this PR while I'll get this passing with a current `turbo`:

`pnpm -- turbo run run-example -- "non-monorepo" "yarn"`

gsoltis (approved):
Filed https://linear.app/vercel/issue/TURBO-847/root-inference-is-broken-for-single-packages-with-a-parent-that to track the bugfix for the behavior change.

examples_tests/npm-non-monorepo.t
```diff
@@ -0,0 +1,5 @@
+  $ . ${TESTDIR}/setup.sh non-monorepo npm
+# run twice and make sure it works
+  $ npx turbo build lint > /dev/null
```
gsoltis, line 3: `turbo` is a dev dependency for the `non-monorepo` example, is there a reason you're using `npx` here?

Looks like that was existing behavior though.

gsoltis (approved):

examples_tests/npm-non-monorepo.t
mehulkar: yeah, i'm not sure why it's like that. I don't see an explanation or conversation about it where it was added https://github.com/vercel/turbo/pull/2650. cc @tknickman if you remember!

PR #4005, opened 2023-02-28
add env to spawned processes

### LINKED ISSUE #3876: [turbopack] Node process executed without process environment variables set

turbopack-230213.2

pnpm

Mac

### Describe the Bug

`process.env` has no shell variables when `next.config.js` is loaded inside `next-dev`.
I'm using turbopack from Next.js.

### Expected Behavior

`process.env` should have variables given from the shell.

### To Reproduce

1. Clone the repository.
2. `pnpm i && pnpm dev:turbo`
3. Open `http://localhost:3000/client-side` and you'll see the hydration error.
4. Set `kNextConfigCheckIfEnvIsDefined` to `true` in `next.config.js` and restart the server.
5. Open `http://localhost:3000` and you'll see the server crash.

These errors do not occur in the normal `next dev`.

I made a workaround for this but I'm not sure if this is safe (as `execute()` is called in many places).
https://github.com/3c1u/turbo/tree/fix-process-env-turbopack-node
With this patch, those errors will disappear.

### Reproduction Repo

https://github.com/3c1u/nextjs-env-injection

ForsakenHarmony:
Fixes WEB-517
Fixes #3876

### DIFF
--- a/crates/next-core/js/src/entry/server-edge-api.tsx
+++ b/crates/next-core/js/src/entry/server-edge-api.tsx
@@ -22,7 +22,7 @@ startHandler(async ({ request, response, query, params, path }) => {
       join(process.cwd(), ".next/server/pages", chunk)
     ),
     wasm: [],
-    env: [],
+    env: Object.keys(process.env),
     assets: [],
   };
   await runEdgeFunction({
--- a/crates/next-core/src/app_source.rs
+++ b/crates/next-core/src/app_source.rs
@@ -6,7 +6,7 @@ use std::{
 use anyhow::{anyhow, Result};
 use indexmap::indexmap;
 use turbo_tasks::{TryJoinIterExt, Value, ValueToString};
-use turbo_tasks_env::ProcessEnvVc;
+use turbo_tasks_env::{CustomProcessEnvVc, EnvMapVc, ProcessEnvVc};
 use turbo_tasks_fs::{rebase, rope::RopeBuilder, File, FileContent, FileSystemPathVc};
 use turbopack::{
     ecmascript::EcmascriptInputTransform,
@@ -346,11 +346,11 @@ pub async fn create_app_source(
         output_path,
     );
 
+    let injected_env = env_for_js(EnvMapVc::empty().into(), false, next_config);
+    let env = CustomProcessEnvVc::new(env, next_config.env()).as_process_env();
+
     let server_runtime_entries =
-        vec![
-            ProcessEnvAssetVc::new(project_path, env_for_js(env, false, next_config))
-                .as_ecmascript_chunk_placeable(),
-        ];
+        vec![ProcessEnvAssetVc::new(project_path, injected_env).as_ecmascript_chunk_placeable()];
 
     let fallback_page = get_fallback_page(
         project_path,
@@ -366,6 +366,7 @@ pub async fn create_app_source(
         context_ssr,
         context,
         project_path,
+        env,
         server_root,
         EcmascriptChunkPlaceablesVc::cell(server_runtime_entries),
         fallback_page,
@@ -381,6 +382,7 @@ async fn create_app_source_for_directory(
     context_ssr: AssetContextVc,
     context: AssetContextVc,
     project_path: FileSystemPathVc,
+    env: ProcessEnvVc,
     server_root: FileSystemPathVc,
     runtime_entries: EcmascriptChunkPlaceablesVc,
     fallback_page: DevHtmlAssetVc,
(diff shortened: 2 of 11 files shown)


ForsakenHarmony:
* main
  * **PR #4005** <a href="https://app.graphite.dev/github/pr/vercel/turbo/4005" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/4005?utm_source=stack-comment).

crates/next-core/js/src/entry/server-edge-api.tsx
```diff
@@ -22,7 +22,7 @@ startHandler(async ({ request, response, query, params, path }) => {
       join(process.cwd(), ".next/server/pages", chunk)
     ),
     wasm: [],
-    env: [],
+    env: Object.keys(process.env),
```
ForsakenHarmony, line 25: This was the actual fix I needed for `front`, but fixing the real env also isn't a bad thing

crates/next-core/src/env.rs
```diff
@@ -47,6 +47,29 @@ pub async fn load_env(project_path: FileSystemPathVc) -> Result<ProcessEnvVc> {
     Ok(env)
 }
 
+#[turbo_tasks::function]
+pub async fn full_env(
```
jridgewell, line 51: Can you document how this differs from `env_for_js`

jridgewell: Also, where is this used?

jridgewell (changes requested):

crates/next-core/js/src/entry/server-edge-api.tsx
sokra: Technically it should only list env vars used in code, but I guess for dev mode it doesn't matter

crates/next-core/src/app_source.rs
```diff
@@ -347,11 +347,11 @@ pub async fn create_app_source(
         output_path,
     );
 
+    let injected_env = env_for_js(EnvMapVc::empty().into(), false, next_config);
+    let env = CustomProcessEnvVc::new(env, next_config.env()).as_process_env();
```
sokra, line 351: Why is that needed? shouldn't next config env already be applied to `env`?

crates/next-core/src/env.rs
sokra: `full_env` is a bit misleading as name. This is the `filtered_env` for client?

crates/next-core/src/page_source.rs
```diff
@@ -217,11 +217,12 @@ pub async fn create_page_source(
     )
     .into();
 
+    let injected_env = env_for_js(EnvMapVc::empty().into(), false, next_config);
+    let env = CustomProcessEnvVc::new(env, next_config.env()).as_process_env();
```
sokra, line 221: Same

crates/turbopack-node/src/lib.rs
```diff
@@ -215,7 +217,18 @@ pub async fn get_renderer_pool(
...
-    Ok(NodeJsPool::new(cwd, entrypoint, HashMap::new(), 4, debug).cell())
+    Ok(NodeJsPool::new(
+        cwd,
+        entrypoint,
+        env.read_all()
+            .await?
+            .iter()
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect(),
+        4,
```
sokra, line 228: This should probably not hard-code 4 child processes, but use `available_concurrency` instead

crates/next-core/js/src/entry/server-edge-api.tsx
ForsakenHarmony: that's true, maybe something we can address in the future

crates/next-core/src/env.rs
ForsakenHarmony: This is unused actually, I was looking into different ways to implement this

I wanted to partially get rid of the injected env variables, but maybe I'll just undo that

crates/next-core/src/app_source.rs
ForsakenHarmony: No, they're applied in `env_for_js` and I wanted the unescaped variables for the real env variables

jridgewell: I think we could reorder a few things, then?
```diff
-    let injected_env = env_for_js(EnvMapVc::empty().into(), false, next_config);
+    let env = CustomProcessEnvVc::new(env, next_config.env()).as_process_env();       
+    let injected_env = env_for_js(env, false, next_config);
```

And remove the `next_config.env()` part from `env_for_js`?

jridgewell (approved):

crates/turbopack-node/src/lib.rs
ForsakenHarmony: done

PR #4006, opened 2023-02-28
enable HMR for client-side navigated pages

sokra:
The root level page module might not be able to HMR on its own (e. g. for MDX modules which export a `meta` object, which makes them not pure component file which opts out React Refresh), so we accept the whole loaded page.

This is safe as the page is a component.

Depends on support from the next.js router to allow updating the current page.


anthonyshew:

PR #4008, opened 2023-02-28
[WIP]: speed up rope iteration and serving

jridgewell:
This is just a test to see if `Bytes` equality is secretly very slow. It uses slice equality, which [uses `memcmp`](https://github.com/rust-lang/rust/blob/31f858d9a511f24fedb8ed997b28304fec809630/library/core/src/slice/cmp.rs#L77-L94) for [`&[u8]`](https://github.com/rust-lang/rust/blob/31f858d9a511f24fedb8ed997b28304fec809630/library/core/src/slice/cmp.rs#L226-L227). But maybe `memcmp` doesn't have the obvious optimization for pointer equality? This is the only reason that I can think of leading to the performance improvements in #3955. The rope traversal can't be that expensive, right?

### DIFF
--- a/crates/turbo-tasks-fs/src/rope.rs
+++ b/crates/turbo-tasks-fs/src/rope.rs
@@ -353,7 +353,7 @@ impl<'de> Deserialize<'de> for Rope {
 impl PartialEq for Rope {
     // Ropes with similar contents are equals, regardless of their structure.
     fn eq(&self, other: &Self) -> bool {
-        if Arc::ptr_eq(&self.data, &other.data) {
+        if self.data.ptr_eq(&other.data) {
             return true;
         }
         if self.len() != other.len() {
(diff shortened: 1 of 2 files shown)


crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -686,6 +702,22 @@ impl From<RopeElem> for StackElem {
     }
 }
 
+trait PtrEq {
```
ForsakenHarmony, line 705: I don't think we need a new trait for this?
Could add a small helper function for `Bytes`

crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -388,14 +388,18 @@ impl PartialEq for Rope {
...
 
         // At this point, we need to do slower contents equality. It's possible we'll
-        // still get some memory reference quality for Bytes.
+        // still get some memory reference equality for Bytes.
         let mut left = RopeReader::new(left, index);
         let mut right = RopeReader::new(right, index);
         loop {
-            match (left.fill_buf(), right.fill_buf()) {
+            match (left.next(), right.next()) {
                 // fill_buf should always return Ok, with either some number of bytes or 0 bytes
```
ForsakenHarmony, line 396: this comment probable needs to be changed

ForsakenHarmony (approved):

jridgewell:
Appears this was not the cause, ripe iteration must just be really slow.

PR #4011, opened 2023-02-28
feat(devserver/http): support gzipped responses

kwonoj:
Implements WEB-666.

This PR enables gzip compressions for the responses by default. This can be extended to other compression method (i.e brotli), but for now implements minimal default compression only to pass existing test case.

Fixes `test/integration/compression/test/index.test.js`.

### DIFF
--- a/crates/turbopack-dev-server/src/http.rs
+++ b/crates/turbopack-dev-server/src/http.rs
@@ -1,6 +1,10 @@
 use anyhow::Result;
 use futures::{StreamExt, TryStreamExt};
-use hyper::{header::HeaderName, Request, Response};
+use hyper::{
+    header::{HeaderName, CONTENT_ENCODING},
+    http::HeaderValue,
+    Request, Response,
+};
 use mime_guess::mime;
 use turbo_tasks::TransientInstance;
 use turbo_tasks_fs::{FileContent, FileContentReadRef};
@@ -117,13 +121,21 @@ pub async fn process_request_with_content_source(
                 }
 
                 let content = file.content();
-                header_map.insert(
-                    "Content-Length",
-                    hyper::header::HeaderValue::try_from(content.len().to_string())?,
+                // Grab ropereader stream, coerce anyhow::Error to std::io::Error
+                let stream_ext = content
+                    .read()
+                    .into_stream()
+                    .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err));
+
+                let gzipped_stream = tokio_util::io::ReaderStream::new(
+                    async_compression::tokio::bufread::GzipEncoder::new(
+                        tokio_util::io::StreamReader::new(stream_ext),
+                    ),
                 );
 
-                let bytes = content.read();
-                return Ok(response.body(hyper::Body::wrap_stream(bytes))?);
+                header_map.insert(CONTENT_ENCODING, HeaderValue::from_static("gzip"));
+
+                return Ok(response.body(hyper::Body::wrap_stream(gzipped_stream))?);
             }
         }
         GetFromSourceResult::HttpProxy(proxy_result) => {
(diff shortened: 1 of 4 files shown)


crates/turbopack-dev-server/src/http.rs
```diff
@@ -117,13 +121,21 @@ pub async fn process_request_with_content_source(
...
 
                 let content = file.content();
-                header_map.insert(
-                    "Content-Length",
-                    hyper::header::HeaderValue::try_from(content.len().to_string())?,
+                // Grab ropereader stream, coerce anyhow::Error to std::io::Error
+                let stream_ext = content
+                    .read()
+                    .into_stream()
+                    .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err));
```
jridgewell, line 128: Nit: (with appropriate `use std::io;` statement at the top)
```suggestion
                    .map_err(|err| io::Error::new(io::ErrorKind::Other, err));
```

crates/turbopack-dev-server/src/http.rs
```diff
@@ -117,13 +121,21 @@ pub async fn process_request_with_content_source(
...
+
+                let gzipped_stream = tokio_util::io::ReaderStream::new(
+                    async_compression::tokio::bufread::GzipEncoder::new(
+                        tokio_util::io::StreamReader::new(stream_ext),
```
jridgewell, line 132: Nit: with `use tokio_util::io::{ReaderStream, StreamReader}; use async_compression::tokio::bufread::GzipEncoder;`
```suggestion
                let gzipped_stream = ReaderStream::new(
                    GzipEncoder::new(
                        StreamReader::new(stream_ext),
```

crates/turbopack-dev-server/src/http.rs
```diff
@@ -117,13 +121,21 @@ pub async fn process_request_with_content_source(
...
-                header_map.insert(
-                    "Content-Length",
-                    hyper::header::HeaderValue::try_from(content.len().to_string())?,
+                // Grab ropereader stream, coerce anyhow::Error to std::io::Error
+                let stream_ext = content
+                    .read()
+                    .into_stream()
+                    .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err));
+
+                let gzipped_stream = tokio_util::io::ReaderStream::new(
```
jridgewell, line 130: Not all content should be compressed, mainly images except SVG. Compressing binary formats will only increase the file size. Can we do an inspection for the content-type to ensure it's a textual format?

kwonoj: Probably yes, could have some basic mime check for the content types.

jridgewell: Express's [compression](https://www.npmjs.com/package/compression) middleware uses [compressible](https://www.npmjs.com/package/compressible)

kwonoj: yes, that's I figures. But we should have rust corresponding one instead, isn't it?

jridgewell: The rust crate seems out of date? We could get most of the way by just using the fallback algorithm from `compressible`:

- text/*
- */*+json
- */*+text
- */*+xml

and a few extras for `application/json` and `application/javascript`

kwonoj: You mean only compress for those, correct?

jridgewell: Yes

crates/turbopack-dev-server/src/http.rs
```diff
@@ -92,14 +100,28 @@ pub async fn process_request_with_content_source(
                     );
                 }
 
+                // naively checking if content is `compressible`.
+                let mut should_compress = false;
+                let should_compress_predicate = |mime: &Mime| match (mime.type_(), mime.subtype()) {
```
jridgewell, line 105: We should also be checking `mime.suffix()` for:
- xml
- json
- text

jridgewell (approved):

PR #4015, opened 2023-02-28
feat(overlay): apply data-* header for the error desc

kwonoj:
Partially resolves WEB-671.

There are some test cases use `data-nextjs-dialog-header` to lookup its inner text to verify error output. Due to differences of error overlay layout between turbopack / next-dev, those 2 have different inner texts.

PR applies a workaround, by introducing new data tag `data-nextjs-turbo-dialog-body`. Next.js upstream will need following update to utilize this tag in its test suites.


ststgg:
🚀

PR #4016, opened 2023-02-28
Calculate TaskHash while constructing PackageTask

mehulkar:
### Description

We want to calculate this earlier, because it has some important side effects such as gathering the hashable inputs, etc. By moving this up the chain, we can ensure that more information is available for run summary and dry run summary in a consistent way.

#### Stacked PRs:
1. https://github.com/vercel/turbo/pull/4016 (this one)
2. https://github.com/vercel/turbo/pull/4021
3. https://github.com/vercel/turbo/pull/4020

### Testing Instructions

`cd cli && make integration-tests`

- [X] Auto label

PR #4020, opened 2023-03-01
Add task + framework env vars into Dry Run Summary

mehulkar:
### Description

This PR adds an `environmentVariables` key into Dry Run Summaries. The data is broken down by the source (`configured`, and `inferred`), so it's easy to see how where it's coming from.  This PR does _not_ include global env vars, but that will likely follow the same pattern.

#### Stacked PRs:
1. https://github.com/vercel/turbo/pull/4016
2. https://github.com/vercel/turbo/pull/4021 
3. https://github.com/vercel/turbo/pull/4020 (this one)

### Testing Instructions

The test cases _should_ cover most cases, but we should run against examples also

- [X] Auto label


cli/integration_tests/basic_monorepo/dry_run.t
```diff
@@ -19,37 +19,46 @@ Setup
...
+    Directory                        = packages/util                                                                                                    
+    Command                          = echo 'building'                                                                                                  
+    Outputs                          =                                                                                                                  
+    Log File                         = packages/util/.turbo/turbo-build.log                                                                             
+    Dependencies                     =                                                                                                                  
+    Dependendents                    =                                                                                                                  
+    Inputs Files Considered          = 1                                                                                                                
+    Configured Environment Variables = NODE_ENV=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```
mehulkar, line 57: One thing that's a little weird, is that if the env var is configured, even if it is not _set_, it adds an empty string into the taskHashInput. Since it is part of the task hash calculation, I'm opting to have it show up here also. But this may be confusing to see.

mehulkar, line 57: In this test case, `NODE_ENV` is _not_ set. But it's included here, because the task hash input includes the string `NODE_ENV=""` . I'm not sure if this was intentional when it was done originally. 

I decided to keep it here, but it could be confusing to read. It might say to the reader: "The `NODE_ENV` env var is set". But what it's really saying is "`NODE_ENV` is part of the task hash input".

gsoltis:
Note to self and other reviewers: the changing hashes for the tests in `basic_monorepo` are expected because `turbo.json` was updated.

cli/integration_tests/basic_monorepo/dry_run.t
```diff
@@ -19,37 +19,46 @@ Setup
...
-    Log File                = apps/my-app/.turbo/turbo-build.log                                                                                              
-    Dependencies            =                                                                                                                                 
-    Dependendents           =                                                                                                                                 
-    Inputs Files Considered = 1                                                                                                                               
-    ResolvedTaskDefinition  = {"outputs":["apple.json","banana.txt"],"cache":true,"dependsOn":[],"inputs":[],"outputMode":"full","env":[],"persistent":false} 
+    Task                             = build                                                                                                                           
+    Package                          = my-app                                                                                                                          
+    Hash                             = e8ca4fc486de5b37
```
mehulkar, line 29: task hash changed because turbo.json in fixture changed

cli/internal/run/dry_run.go
```diff
@@ -389,5 +400,11 @@ func (ht *taskSummary) toSinglePackageTask() singlePackageTaskSummary {
...
 		Framework:              ht.Framework,
 		ExpandedInputs:         ht.ExpandedInputs,
+		EnvVars:                ht.EnvVars,
 	}
 }
+
+type taskEnvVarSummary struct {
+	Configured []string `json:"configured"`
+	Inferred   []string `json:"inferred"`
+}
```
mehulkar, line 410: This may also include global env vars that are affecting the task hash (i.e. `globalEnv` and the implicit turbo injected env var `THASH`)

cli/internal/taskhash/taskhash.go
```diff
@@ -332,7 +335,9 @@ func (th *Tracker) CalculateTaskHash(packageTask *nodes.PackageTask, dependencyS
 }
 
 // GetExpandedInputs gets the expanded set of inputs for a given PackageTask
-// Thes was stored during CalculateFilesHash, so that method must run first
+// Thes was stored during CalculateFilesHash(), so that method must run first.
+// Reads from the underlying map in this method do NOT need to be protected by a mutex
```
gsoltis, line 339: I think this comment would be better off attached to the map declaration itself. What you're describing is a property of the map, not this method. Another method that accessed the same map would need the same comment.

IIUC, this map is calculated once and then considered immutable after `CalculateFileHashes` returns.

cli/internal/taskhash/taskhash.go
```diff
@@ -344,3 +349,14 @@ func (th *Tracker) GetExpandedInputs(packageTask *nodes.PackageTask) map[turbopa
 
 	return inputsCopy
 }
+
+// GetEnvVars returns the hashed env vars for a given taskID.
+// This operation is protected by a mutex, because the underlying map where the
```
gsoltis, line 354: As above, I think this comment is better attached to the declaration of the map. The fact that it is protected by the mutex is a property of the map, not of this particular method.

gsoltis (approved):

PR #4021, opened 2023-03-01
Refactor env package to breakdown env vars for task by source

mehulkar:
The ToHashable() method continues to guarantee that the Composite env
vars is deterministic and will not change the input to determining the
task hash.

#### Stacked PRs:
1. https://github.com/vercel/turbo/pull/4016
2. https://github.com/vercel/turbo/pull/4021 (this one)
3. https://github.com/vercel/turbo/pull/4020


cli/internal/env/env.go
```diff
@@ -9,7 +9,55 @@ import (
 	"github.com/vercel/turbo/cli/internal/util"
 )
 
-func getEnvMap() map[string]string {
+// Details contains a map of environment variables broken down by the source
+type Details struct {
```
tknickman, line 13: Not blocking: `Details` by itself doesn't tell me much, it's public as well so would make it a _little_ more explicit. 

```suggestion
type BySource struct {
```

cli/internal/env/env.go
```diff
@@ -20,57 +68,54 @@ func getEnvMap() map[string]string {
...
-func getEnvPairsFromPrefixes(includePrefixes []string, excludePrefix string, allEnvVars map[string]string) []string {
-	allEnvPairs := []string{}
-	for _, includePrefix := range includePrefixes {
-		hashableFrameworkEnvPairs := getEnvPairsFromPrefix(includePrefix, excludePrefix, allEnvVars)
-		allEnvPairs = append(allEnvPairs, hashableFrameworkEnvPairs...)
-	}
-	return allEnvPairs
-}
+// Get returns all sorted key=value env var pairs for both frameworks and from envKeys
+func Get(keys []string, prefixes []string) DetailedMap {
```
tknickman, line 101: Not blocking: this is less descriptive than the old name in my opinion. `env.Get` could easily be confused with just getting all env vars.

cli/internal/env/env.go
```diff
@@ -9,7 +9,55 @@ import (
...
+}
+
+// EnvironmentVariablePairs is a list of "k=v" strings for env variables and their values
+type EnvironmentVariablePairs []string
+
+// ToHashable returns a deterministically sorted set of EnvironmentVariablePairs from an EnvironmentVariableMap
+// This is the value that is used upstream as a task hash input, so we need it to be deterministic
+func (evm EnvironmentVariableMap) ToHashable() EnvironmentVariablePairs {
+	// convert to set to eliminate duplicates
+	unique := make(util.Set, len(evm))
```
tknickman, line 45: In _theory_ we don't need this anymore because our source is a map, so all keys will be unique anyway. So we could eliminate this to save a little overhead of converting back and forth.

cli/internal/env/env.go
```diff
@@ -9,7 +9,55 @@ import (
...
+// DetailedMap contains the composite and the detailed maps of environment variables
+// Composite is used as a taskhash input (taskhash.CalculateTaskHash)
+// Details is used to print out a Dry Run Summary
+type DetailedMap struct {
+	Composite EnvironmentVariableMap
+	Details   Details
+}
+
+// EnvironmentVariableMap is a map of env variables and their values
+type EnvironmentVariableMap map[string]string
```
tknickman, line 27: Not blocking:

I would move this above `EnvironmentVariableMap` just to aid the readable flow

cli/internal/env/env.go
```diff
@@ -9,7 +9,55 @@ import (
...
+	Prefixed EnvironmentVariableMap
+}
+
+// DetailedMap contains the composite and the detailed maps of environment variables
+// Composite is used as a taskhash input (taskhash.CalculateTaskHash)
+// Details is used to print out a Dry Run Summary
+type DetailedMap struct {
+	Composite EnvironmentVariableMap
+	Details   Details
+}
```
tknickman, line 24: Not blocking, more naming bikesheding:

```suggestion
type DetailedMap struct {
	All EnvironmentVariableMap
	BySource EnvironmentVariableBySource
}
```

tknickman (approved):
Logic looks good, a few non-blocking naming comments, but overall :shipit:

cli/internal/env/env.go
mehulkar: good call, removed

cli/internal/env/env.go
mehulkar: I was mostly getting rid of `GetHashableEnv "Pairs"`. `GetHashableEnvVars` seems better!

cli/internal/env/env.go
mehulkar: good ideas, updated

PR #4024, opened 2023-03-01
Add "codeSplitting: false" into package.json - for code-splitting packages/ui exports

cpotey:
### Description

Currently, all exports (if you have more than one component being exported in your `packages/ui/index.tsx`) get included everywhere one import is used.

So say you have two exports of your custom components in `packages/ui/index.tsx`

```
export * from './components/Button';
export * from './components/Input';
```

In your `apps/web`, we choose to only import one of these, say the Button component

`import { Button } from 'ui'`

It will also include the `Input` component and its dependencies in the build too. 

Adding `"codeSplitting": false` to the `packages/ui/package.json` stops this, when building with Webpack. [(Page from Webpack re codesplitting)](https://webpack.js.org/guides/code-splitting/)

I think it would be good to include this line as standard into the example.

### Testing Instructions

You could install [Next Bundle analyzer](https://www.npmjs.com/package/@next/bundle-analyzer) in `apps/web`, add another component into the `packages/ui` folder, import it within a `apps/web` page, and compare a build with/without the `"codeSplitting":false` in the `package.json`

[X] Auto label


arlyon:
Thanks for this, I have approved CI

mehulkar (changes requested):
Thanks for the contribution @cpotey. Since this is not a standard package.json field, I'm inclined to say we should not add it into our example. 

That said, I'm not able to find any documentation about this field either. The link you provided talks about code splitting, but I don't see this particular field mentioned in there. Am I missing something?

cpotey:
@mehulkar I was going crazy there - couldn't find any reference to it either! ...turns out my PR was actually different to what I had in my project in the first place... 

The _"proper change"_ should be `"sideEffects": false,` not `"codeSplitting": false,` the entire time 🤦‍♂️ [Link to docs](https://webpack.js.org/guides/tree-shaking/)

And now to add insult to injury, I can't even seem to replicate this doing what I said in my own environment anymore, when analysing my bundles. Without that line in the `package.json`, it no longer seems to be including other `package/ui` components. I think something was weird in my initial setup!

Must've had a long day when I made this PR 😅 think it's best to just close it! 
Appreciate the response and checking on this though - I'll keep an eye on my builds, and open another if I work out what on earth happened to make me open this in the first place!

mehulkar:
Ah, I need to learn more about `sideEffects: false`. Looks like we do use this here: https://github.com/vercel/turbo/blob/main/examples/with-tailwind/packages/ui/package.json#L4-L6, so my response about it 
 not being a standard field may not be valid. If you feel like it's worth adding, feel free to re-open!

PR #4026, opened 2023-03-01
debug print errors for nested vcs

sokra:
### Description

It's annoying to debug when the whole dbg will throw an error if any nested Vc is in error state.
Instead we want to print the errors here.

### Testing Instructions

[X] Auto label

### DIFF
--- a/crates/turbo-tasks-macros/src/derive/value_debug_format_macro.rs
+++ b/crates/turbo-tasks-macros/src/derive/value_debug_format_macro.rs
@@ -64,7 +64,13 @@ fn format_named(ident: &Ident, fields: &FieldsNamed) -> (TokenStream2, TokenStre
                 vec![#(
                     FormattingField::new(
                         stringify!(#fields_idents),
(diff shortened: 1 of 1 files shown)


crates/turbo-tasks-macros/src/derive/value_debug_format_macro.rs
```diff
@@ -82,7 +88,13 @@ fn format_unnamed(ident: &Ident, fields: &FieldsUnnamed) -> (TokenStream2, Token
...
                 stringify!(#ident),
                 vec![#(
-                    #fields_idents.value_debug_format(depth.saturating_sub(1)).try_to_value_debug_string().await?.await?.to_string(),
+                    match #fields_idents.value_debug_format(depth.saturating_sub(1)).try_to_value_debug_string().await {
+                        Ok(result) => match result.await {
+                            Ok(result) => result.to_string(),
+                            Err(err) => format!("{:?}", err),
+                        },
+                        Err(err) => format!("{:?}", err),
+                    },
```
alexkirsz, line 97: This could move into its own `format_field(ident: &Ident) -> TokenStream2` function.

alexkirsz (approved):

PR #4027, opened 2023-03-01
Add missing function from port.

nathanhammond:
This function missed the ride when we ported it over. This should give us a better error message when we encounter a installation on the wrong platform.

Here is the invocation on `main` of the nonexistent function:
https://github.com/vercel/turbo/blob/main/packages/turbo/node-platform.js#L88


tknickman (approved):
Good catch, thanks

PR #4028, opened 2023-03-01
Capture console statements as one unit including stack trace

sokra:
### Description

fixes the deduplication issue with multiple output

### Testing Instructions

[X] Auto label


crates/next-core/js/src/ipc/index.ts
```diff
@@ -142,3 +142,39 @@ export const IPC = createIpc<unknown, unknown>(parseInt(PORT, 10));
...
+  // @ts-ignore
+  const original = console[name];
+  // @ts-ignore
+  const stdio = process[stream];
+  // @ts-ignore
+  console[name] = (...args: any[]) => {
+    stdio.write(`TURBOPACK_OUTPUT_B\n`);
+    original(...args);
+    if (addStack) {
+      const stack = new Error().stack?.replace(/^.+\n.+\n/, "") + "\n";
```
jridgewell, line 156: Should we use a structured stack here?

crates/turbopack-node/js/src/ipc/index.ts
```diff
@@ -142,3 +142,39 @@ export const IPC = createIpc<unknown, unknown>(parseInt(PORT, 10));
...
+      stdio.write(stack);
+    }
+    stdio.write("TURBOPACK_OUTPUT_E\n");
+  };
+};
+
+improveConsole("error", "stderr", true);
+improveConsole("warn", "stderr", true);
+improveConsole("count", "stdout", true);
+improveConsole("count", "stdout", true);
```
jridgewell, line 167: Ditto.

crates/turbopack-node/src/pool.rs
```diff
@@ -167,7 +234,16 @@ impl NodeJsPoolProcess {
                         .unwrap()
                         .read_to_end(&mut stderr)
                         .await?;
-                    Ok((String::from_utf8(stdout)?, String::from_utf8(stderr)?))
+                    fn clean(buffer: Vec<u8>) -> Result<String> {
+                        Ok(String::from_utf8(buffer)?
+                            .lines()
+                            .filter(|line| {
+                                line.len() != MARKER_STR.len() + 1 && !line.starts_with(MARKER_STR)
```
jridgewell, line 241: Why the `line.len() != MARKER_STR.len() + 1 ` condition?

jridgewell (approved):

crates/turbopack-node/src/pool.rs
sokra: Fast check for the right len

crates/next-core/js/src/ipc/index.ts
sokra: Yep, we should do that in the next iteration when you actually do something valueable with the stack trace.

Issue #3979, opened 2023-02-27
[turborepo] @turbo/codemod migrate sets output for persistent/non-cacheable task


maschwenk:
### What version of Turborepo are you using?

1.8.2

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

When running `pnpx @turbo/codemod migrate` new `outputs` are added to a persistent and on-cached step, which seems antithetical to the purpose of those configs.

![image](https://user-images.githubusercontent.com/1264305/221605033-8141f7e1-d93b-47bb-ac0b-f42dab4eff4d.png)

### Expected Behavior

No `outputs` should be added to a persistent/non-cacheable task

### To Reproduce

`pnpx @turbo/codemod migrate`

on a `turbo.json` that has

```
"dev": {
       "cache": false,
       "persistent": true,
}
```

tknickman:
Good catch @maschwenk - will fix. Thanks!

tknickman:
tl;dr, yes, you're exactly right. 

For some more background - previously, if you left out outputs, we would add in defaults (`"build/**/*", "dist/**/*"`) for you (https://github.com/vercel/turbo/pull/2712 | [specific line](https://github.com/vercel/turbo/pull/2712/files#diff-d47901181dc37fc21211eb843789bff7ba28753b469dd94334d2a4d0d3778f17L106)). 

Because of that, no outputs, and empty outputs weren't equivalent. We modified this in 1.7 ([details](https://turbo.build/blog/turbo-1-7-0#declare-your-outputs-for-improved-clarity)) to make this more explicit, and as part of this changed, provided a codemod for users who were (likely unintentionally) relying on the default outputs we were setting under the hood.

maschwenk:
@tknickman but just to clarify. **Now** that there is no default for `outputs`, `outputs` not set and `outputs` set to `[]` are now functionally the same? I mean maybe you save one method call or something, but the functional outcome is identical?

tknickman:
Yes that's correct.

PR #4029, opened 2023-03-01
fix(codemod): don’t set outputs for cache:false

tknickman:
### Description

Fixes https://github.com/vercel/turbo/issues/3979. We don't need to add outputs to tasks with `cache: false`

### Testing Instructions

This change is fully covered with unit tests.

- [x] Auto label


mehulkar (approved):
This makes sense to me, but I guess it could be confusing this way too. If you upgrade from 1.6 and run this codemod, your _resolved_ config would go from:

```
{ cache: false, outputs: [dist, build] }
```

to:

```
{ cache: false, outputs: [] }
```

In either case, I think the real long term solution is the run summary work, so I have no problem with this!

nathanhammond (approved):

PR #4037, opened 2023-03-01
fix(route_matcher): do not include empty segment

kwonoj:
### Description

Closes WEB-674

if a matcher tries to split path have a slash like `/x/y`, split gives a redundant empty segment `["", 'x', 'y']`.

### DIFF
--- a/crates/next-core/src/next_route_matcher/path_regex.rs
+++ b/crates/next-core/src/next_route_matcher/path_regex.rs
@@ -93,7 +93,7 @@ impl PathRegexBuilder {
         R: AsRef<str>,
     {
         self.push_str(if self.include_slash() {
-            "(/[^?]+)?"
+            "(?:/([^?]+))?"
         } else {
             "([^?]+)?"
         });


crates/next-core/src/next_route_matcher/path_regex.rs
```diff
@@ -52,6 +52,7 @@ impl RouteMatcher for PathRegex {
                                 value
                                     .as_str()
                                     .split('/')
+                                    .filter(|segment| !segment.is_empty())
```
jridgewell, line 55: I think this works, but it's fixing the symptom instead of the problem. The regex that we're using shouldn't capture the `/` when we mush a catchall. Checking in here, we actually do it correctly for `push_catch_all`, but incorrectly for `push_optional_catch_all`.

I think the right fix is:

```diff
diff --git a/crates/next-core/src/next_route_matcher/path_regex.rs b/crates/next-core/src/next_route_matcher/path_regex.rs
index 5482fdd24..688de9a42 100644
--- a/crates/next-core/src/next_route_matcher/path_regex.rs
+++ b/crates/next-core/src/next_route_matcher/path_regex.rs
@@ -93,7 +93,7 @@ impl PathRegexBuilder {
         R: AsRef<str>,
     {
         self.push_str(if self.include_slash() {
-            "(/[^?]+)?"
+            "(?:/([^?]+))?"
         } else {
             "([^?]+)?"
         });

jridgewell (changes requested):

crates/next-core/src/next_route_matcher/path_regex.rs
kwonoj: I was actually under suspicion I might fixing the leaf instead of the root. Updated as suggested.

jridgewell (approved):

PR #4039, opened 2023-03-01
next/font: Align with Next.js integration test suite

wbinnssmith:
This:
* Adds a compile-time feature, `__internal_nextjs_integration_test`, which is set when building Turbopack for the Next.js integration test suite.
* When built with `__internal_nextjs_integration_test`, expects the environment variable `NEXT_FONT_GOOGLE_MOCKED_RESPONSES` to be set with an absolute path to a JavaScript file exporting a mapping of urls to stylesheet text. This is used in place of making real requests to Google Fonts.
* Addresses some small bugs in next/font/google to bring the implementation in line with the urls generated by Next.js without Turbopack.
* Changes classnames generated by next/font to use 6 character hashes, bringing it in alignment with Next.js without Turbopack.

Test Plan:
* Adjusted existing unit tests
* In Next.js, `node run-tests.js -c 1 test/e2e/next-font/index.test.ts`

### Testing Instructions

- [x] Auto label


crates/next-core/Cargo.toml
```diff
@@ -40,3 +40,4 @@ turbo-tasks-build = { path = "../turbo-tasks-build" }
 next-font-local = []
 native-tls = ["turbo-tasks-fetch/native-tls"]
 rustls-tls = ["turbo-tasks-fetch/rustls-tls"]
+__internal_nextjs_integration_test = []
```
kwonoj, line 43: nit: can we leave a comment this is for the test / internal only?

crates/next-core/src/next_font_google/util.rs
```diff
@@ -170,44 +178,54 @@ pub(crate) fn get_stylesheet_url(
...
-        .map(|pair| pair.0)
-        .collect::<Vec<&str>>()
-        .join(",");
-
-    let mut variant_values = variants
-        .iter()
-        .map(|variant| {
-            variant
+    let first_variant = variants.first();
+    println!("first variant {:?}", first_variant);
```
kwonoj, line 182: is this intentional?

hanneslund (approved):

kwonoj (approved):

PR #4047, opened 2023-03-02
Add turboVersion into dry run summary

mehulkar:


cli/integration_tests/basic_monorepo/dry_run_json.t
```diff
@@ -5,6 +5,10 @@ Setup
 # Save JSON to tmp file so we don't need to keep re-running the build
   $ ${TURBO} run build --dry=json > tmpjson.log
 
+# test with a regex that captures what release we usually have (1.x.y or 1.a.b-canary.c)
+  $ cat tmpjson.log | jq .turboVersion
+  "\d\.\d\.\d(-canary\.\d)?" (re)
```
mehulkar, line 10: I don't _really_ care how this gets validated. It could be even more generic as `"1..*" (re)` . As long as we're reasonably sure this key exists and has some content, I don't think we need to go crazy with validation.

nathanhammond: This is fine, and way overkill.

nathanhammond (approved):

tknickman (approved):

PR #4048, opened 2023-03-02
Reimplement RopeReader to borrow Bytes/InnerRope

jridgewell:
I'm hoping this'll eek out a bit more performance from the `RopeReader` and everything dependent on it.

This was complicated a bit by Hyper's `wrap_stream`, which requires that the stream be `'static`. Because the `RopeReader` borrows data from the `Rope`, we were bound by the rope's lifetime. But by holding onto a cloned `InnerRope`, we're able to transmute the borrowed data. This is fine because our held rope will ensure the data for as long as the reader itself is alive.

### DIFF
--- a/crates/turbo-tasks-fs/src/rope.rs
+++ b/crates/turbo-tasks-fs/src/rope.rs
@@ -38,8 +38,8 @@ pub struct Rope {
 
 /// An Arc container for ropes. This indirection allows for easily sharing the
 /// contents between Ropes (and also RopeBuilders/RopeReaders).
-#[derive(Clone, Debug, Default)]
-struct InnerRope(Arc<Box<[RopeElem]>>);
+#[derive(Clone, Debug)]
+struct InnerRope(Arc<[RopeElem]>);
 
 /// Differentiates the types of stored bytes in a rope.
 #[derive(Clone, Debug)]
@@ -120,7 +120,7 @@ impl<T: Into<Bytes>> From<T> for Rope {
         } else {
             Rope {
                 length: bytes.len(),
-                data: InnerRope::from(Box::from([Local(bytes)])),
+                data: InnerRope(Arc::from([Local(bytes)])),
             }
         }
     }
@@ -202,7 +202,7 @@ impl RopeBuilder {
         self.finish();
         Rope {
             length: self.length,
-            data: InnerRope::from(self.committed.into_boxed_slice()),
+            data: InnerRope::from(self.committed),
         }
     }
 }
@@ -453,6 +453,12 @@ impl InnerRope {
     }
 }
 
+impl Default for InnerRope {
+    fn default() -> Self {
+        InnerRope(Arc::from([]))
+    }
+}
+
 impl DeterministicHash for InnerRope {
     /// Ropes with similar contents hash the same, regardless of their
     /// structure. Notice the InnerRope does not contain a length (and any
@@ -465,8 +471,8 @@ impl DeterministicHash for InnerRope {
     }
 }
 
-impl From<Box<[RopeElem]>> for InnerRope {
-    fn from(els: Box<[RopeElem]>) -> Self {
+impl From<Vec<RopeElem>> for InnerRope {
+    fn from(els: Vec<RopeElem>) -> Self {
         if cfg!(debug_assertions) {
             // It's important that an InnerRope never contain an empty Bytes section.
             for el in els.iter() {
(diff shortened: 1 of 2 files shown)


crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -532,31 +538,85 @@ impl DeterministicHash for RopeElem {
...
     /// The Rope's tree is kept as a cloned stack, allowing us to accomplish
     /// incremental yielding.
-    stack: Vec<StackElem>,
+    stack: Vec<StackElem<'a>>,
+
+    /// The leaked root allows us to invert the ownership of Bytes. By leaking
+    /// the Rope's root, we're able to use a 'static lifetime and let the
+    /// RopeReader own the memory and cleanup. When the RopeReader is dropped,
+    /// the InnerRope's Arc will be manually decremented to free the memory.
+    leaked_root: Option<*mut InnerRope>,
```
alexkirsz, line 552: I'm not sure I understand the `Box::leak`/`Box::from_raw` dance. Wouldn't it be enough to keep an `Option<InnerRope>` here?

crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -532,31 +538,85 @@ impl DeterministicHash for RopeElem {
...
-            }
+            vec![StackElem::Shared(inner, index)]
+        };
+        RopeReader {
+            stack,
+            leaked_root: None,
+        }
+    }
+
+    /// Iterates the rope's elements recursively until we find the next Local
```
alexkirsz, line 599:
```suggestion
    /// Iterates over the rope's elements recursively until we find the next Local
```

crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -631,40 +681,30 @@ impl AsyncRead for RopeReader {
...
-                // those bytes are not returned on the next call to [fill_buf].
-                b.advance(amt);
+                debug_assert!(
+                    *offset + amt == bytes.len(),
+                    "cannot over-consume RopeReader"
+                );
+                self.stack.pop();
             }
-        }
+        };
```
alexkirsz, line 703:
```suggestion
        }
```

crates/turbo-tasks-fs/src/rope.rs
jridgewell: Yes, but we still need to turn the `'a` lifetime into a `'static`. Actually, I think this'll work with a `transmute()` and the held `InnerRope`, let me try…

crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -532,31 +538,79 @@ impl DeterministicHash for RopeElem {
...
+
+impl<'a> RopeReader<'a> {
+    /// Constructs a new RopeReader, taking ownership of the Rope's internal
+    /// memory and freeing us from the lifetime constraints of the Rope.
+    fn new(rope: &Rope) -> RopeReader<'static> {
+        // SAFETY: We transmute the InnerRope into a 'static lifetime so that it is
+        // usable without keeping a reference to the Rope alive. This is safe because
+        // the RopeReader itself holds an Arc clone of the data, ensuring the data will
+        // live at least as long as the reader.
+        let borrow = unsafe { std::mem::transmute::<&InnerRope, &'static InnerRope>(&rope.data) };
```
sokra, line 571: This is not safe since you are holding a reference to the passed `InnerRope` and not to the cloned `InnerRope` that's store in `held_root`. You would need to move the `reader.held_root = Some(borrow.clone());` up and borrow from that instead. But that doesn't work for two reasons:
* `new_borrow` returns a new object.
* It would require you to return `Pin<Box<RopeReader<...>>>` since you need to prevent the memory location of `held_root` from moving.

It's probably better to hold references to the content of the `Arc` instead (`&[RopeElem]`) since they won't move as long as you are holding the Arc

sokra: But in the end I would avoid the unsafe logic here at all by adding a `StackElem::Owned(usize)` which looks up the index in `held_root` (probably rename it)

sokra (changes requested):

crates/turbo-tasks-fs/src/rope.rs
jridgewell: > But in the end I would avoid the unsafe logic here at all by adding a StackElem::Owned(usize) which looks up the index in held_root (probably rename it)

I think this'll still require a `transmute`, so that I can cast references from `&self.held_root[index]` into the static lifetime, too.

jridgewell:
This is getting too into weeds for me to feel comfortable. I'm going to close this, and we can reopen in the future if Rope iteration is a bottleneck.

PR #4049, opened 2023-03-02
Re-repair Root Inference

nathanhammond:
Our inference got a little wonky once we started allowing `turbo.json` in workspaces. This PR restores the ideal selection of the root and reduces the number of syscalls as compared to our current implementation.

### DIFF
--- a/crates/turborepo-lib/src/package_manager.rs
+++ b/crates/turborepo-lib/src/package_manager.rs
@@ -3,11 +3,11 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use serde::Deserialize;
 
 #[derive(Debug, Deserialize)]
-struct PnpmWorkspaces {
+struct PnpmWorkspace {
     pub packages: Vec<String>,
 }
 
@@ -60,6 +60,13 @@ pub struct Globs {
     exclusions: Vec<PathBuf>,
 }
 
+impl Globs {
+    pub fn test(&self, root: PathBuf, target: PathBuf) -> bool {
+        // TODO
+        true
+    }
+}
+
 impl PackageManager {
     /// Returns a list of globs for the package workspace.
     /// NOTE: We return a `Vec<PathBuf>` instead of a `GlobSet` because we
@@ -70,25 +77,21 @@ impl PackageManager {
     ///
     /// * `root_path`:
     ///
-    /// returns: Result<Globs, Error>
+    /// returns: Result<Option<Globs>, Error>
     ///
     /// # Examples
     ///
     /// ```
     /// ```
-    pub fn get_workspace_globs(&self, root_path: &Path) -> Result<Globs> {
+    pub fn get_workspace_globs(&self, root_path: &Path) -> Result<Option<Globs>> {
         let globs = match self {
             PackageManager::Pnpm | PackageManager::Pnpm6 => {
                 let workspace_yaml = fs::read_to_string(root_path.join("pnpm-workspace.yaml"))?;
-                let workspaces: PnpmWorkspaces = serde_yaml::from_str(&workspace_yaml)?;
-                if workspaces.packages.is_empty() {
-                    return Err(anyhow!(
-                        "pnpm-workspace.yaml: no packages found. Turborepo requires pnpm \
-                         workspaces and thus packages to be defined in the root \
-                         pnpm-workspace.yaml"
-                    ));
+                let pnpm_workspace: PnpmWorkspace = serde_yaml::from_str(&workspace_yaml)?;
+                if pnpm_workspace.packages.is_empty() {
+                    return Ok(None);
                 } else {
-                    workspaces.packages
+                    pnpm_workspace.packages
                 }
             }
             PackageManager::Berry | PackageManager::Npm | PackageManager::Yarn => {
@@ -96,10 +99,7 @@ impl PackageManager {
                 let package_json: PackageJsonWorkspaces = serde_json::from_str(&package_json_text)?;
 
                 if package_json.workspaces.as_ref().is_empty() {
-                    return Err(anyhow!(
-                        "package.json: no packages found. Turborepo requires packages to be \
-                         defined in the root package.json"
-                    ));
+                    return Ok(None);
                 } else {
                     package_json.workspaces.into()
                 }
@@ -117,10 +117,10 @@ impl PackageManager {
             }
         }
 
-        Ok(Globs {
+        Ok(Some(Globs {
             inclusions,
             exclusions,
-        })
+        }))
     }
 }
 
@@ -135,6 +135,7 @@ mod tests {
         let package_manager = PackageManager::Npm;
         let globs = package_manager
             .get_workspace_globs(Path::new("../../examples/with-yarn"))
+            .unwrap()
             .unwrap();
 
         assert_eq!(
--- a/crates/turborepo-lib/src/package_manager.rs
+++ b/crates/turborepo-lib/src/package_manager.rs
@@ -4,7 +4,6 @@ use std::{
 };
 
 use anyhow::{anyhow, Result};
-use glob_match;
 use serde::Deserialize;
 
 #[derive(Debug, Deserialize)]
@@ -64,20 +63,19 @@ impl Globs {
         let search_value = target
             .strip_prefix(root)?
             .to_str()
-            .ok_or(anyhow!("The relative path is not UTF8."))?;
+            .ok_or_else(|| anyhow!("The relative path is not UTF8."))?;
 
         let includes = &self
             .inclusions
             .iter()
-            .find(|inclusion| glob_match::glob_match(inclusion, search_value))
-            .is_some();
+            .any(|inclusion| glob_match::glob_match(inclusion, search_value));
+
         let excludes = &self
             .exclusions
             .iter()
-            .find(|exclusion| glob_match::glob_match(exclusion, search_value))
-            .is_some();
+            .any(|exclusion| glob_match::glob_match(exclusion, search_value));
 
-        return Ok(*includes && !excludes);
+        Ok(*includes && !excludes)
     }
 }
 
--- a/crates/turborepo-lib/src/package_manager.rs
+++ b/crates/turborepo-lib/src/package_manager.rs
@@ -4,6 +4,7 @@ use std::{
 };
 
 use anyhow::Result;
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/package_manager.rs
```diff
@@ -55,9 +55,16 @@ pub enum PackageManager {
...
     #[allow(dead_code)]
-    exclusions: Vec<PathBuf>,
+    pub exclusions: Vec<PathBuf>,
+}
+
+impl Globs {
+    pub fn test(&self, _root: PathBuf, _target: PathBuf) -> bool {
+        // TODO
+        true
+    }
```
nathanhammond, line 67: This is a big gap in the implementation. Once it works we're all set. @arlyon what's our story on Rust globbing?

arlyon: I'll have more to say when the daemon server lands, but glob-match (https://github.com/devongovett/glob-match) seems to be the way to go.

crates/turborepo-lib/src/shim.rs
```diff
@@ -290,10 +290,14 @@ impl RepoState {
...
         //   2. [0].has_turbo_json && [n].has_turbo_json && [n].is_workspace_root_of(0)
+        //
+        //   If we elect to make any of the changes for early exits we need to expand
+        // the   test suite which presently relies on the fact that the
+        // selection runs in a   loop to avoid creating those test cases.
 
         // We need to perform the same search strategy for _both_ turbo.json and _then_
         // package.json.
-        let search_locations = [InferInfo::has_package_json, InferInfo::has_turbo_json];
+        let search_locations = [InferInfo::has_turbo_json, InferInfo::has_package_json];
```
nathanhammond, line 300: This is why you write tests. 🤦‍♂️

crates/turborepo-lib/src/shim.rs
```diff
@@ -345,23 +351,19 @@ impl RepoState {
...
+                // We have eliminated RepoMode::MultiPackage as an option.
+                // We must exhaustively check before this becomes the answer.
                 let local_turbo_state = LocalTurboState::infer(check_roots[0].path.to_path_buf());
                 return Ok(Self {
                     root: check_roots[0].path.to_path_buf(),
-                    mode: if check_roots[0].package_has_workspaces {
-                        RepoMode::MultiPackage
-                    } else {
-                        RepoMode::SinglePackage
-                    },
```
nathanhammond: Identified that this was unnecessary during testing.

nathanhammond (commented):
Ready to land.

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,156 @@ pub struct RepoState {
...
+            .map(|path| {
+                let workspace_globs = PackageManager::Pnpm
+                    .get_workspace_globs(path)
+                    .unwrap_or_else(|_| {
+                        PackageManager::Npm
+                            .get_workspace_globs(path)
+                            .unwrap_or(None)
+                    });
+
+                InferInfo {
```
NicholasLYang, line 271: If I'm reading this correctly, we allocate a `PathBuf` on every ancestor, but then filter on `has_package_json` and `has_turbo_json`. Could we use `filter_map` and avoid these extra allocations by early returning `None` if `has_package_json` and `has_turbo_json` are both false?

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,156 @@ pub struct RepoState {
...
+                .filter(check_set_comparator)
+                .collect();
+
+            // No potential roots checking by this comparator.
+            if check_roots.is_empty() {
+                continue;
+
+            // If there is only one potential root, that's the winner.
+            } else if check_roots.len() == 1 {
+                let local_turbo_state = LocalTurboState::infer(check_roots[0].path.to_path_buf());
```
NicholasLYang, line 314: I think we can borrow here instead of using `to_path_buf`

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,156 @@ pub struct RepoState {
...
-            .to_path_buf();
-
-        let local_turbo_state = LocalTurboState::infer(&root);
-        Ok(Self {
-            root,
-            mode: RepoMode::SinglePackage,
-            local_turbo_state,
-        })
+        // If we're here we didn't find a valid root.
+        return Err(anyhow!("Root could not be inferred."));
```
NicholasLYang, line 366: Can drop the `return`

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,156 @@ pub struct RepoState {
...
+
+        // We need to perform the same search strategy for _both_ turbo.json and _then_
+        // package.json.
+        let search_locations = [InferInfo::has_turbo_json, InferInfo::has_package_json];
+
+        for check_set_comparator in search_locations {
+            let check_roots: Vec<&InferInfo> = potential_turbo_roots
+                .iter()
+                .filter(check_set_comparator)
+                .collect();
```
NicholasLYang, line 306: More of a note than a serious suggestion, but we could probably do this without using `collect` (and therefore without allocating a Vec) if we use `peek`. We'd call `next` and get the first element.

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,156 @@ pub struct RepoState {
...
+
+            // No potential roots checking by this comparator.
+            if check_roots.is_empty() {
```
NicholasLYang, line 309: 1. If that element is `None`, we `continue` here

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,156 @@ pub struct RepoState {
...
+                continue;
+
+            // If there is only one potential root, that's the winner.
+            } else if check_roots.len() == 1 {
```
NicholasLYang, line 313: 2. If we call `peek` and that is `None`, we know there is only one root.

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,156 @@ pub struct RepoState {
...
+                });
 
-            if is_workspace {
-                let local_turbo_state = LocalTurboState::infer(dir);
+            // More than one potential root.
+            // Closest is not RepoMode::MultiPackage
+            // We attempt to prove that the closest is a workspace of a parent.
+            // Failing that we just choose the closest.
+            } else {
+                for ancestor_infer in &check_roots[1..] {
```
NicholasLYang, line 342: 3. Instead of slicing we consume the rest of the iterator here.

NicholasLYang (changes requested):
This seems a lot more robust and better from a business logic perspective. Have a couple notes around idiomatic Rust

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,154 @@ pub struct RepoState {
...
+                        PackageManager::Npm
+                            .get_workspace_globs(path)
+                            .unwrap_or_else(|_| None)
+                    });
+
+                InferInfo {
+                    path: path.to_owned(),
+                    has_package_json: fs::metadata(path.join("package.json")).is_ok(),
+                    has_turbo_json: fs::metadata(path.join("turbo.json")).is_ok(),
+                    package_has_workspaces: workspace_globs.is_some(),
```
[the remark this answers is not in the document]
nathanhammond: It is derived state; I got tired of writing `is_some()` in an earlier implementation which wasn't as gracefully written. Good catch, updated to remove.

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,154 @@ pub struct RepoState {
...
+impl RepoState {
+    fn generate_potential_turbo_roots(reference_dir: &Path) -> Vec<InferInfo> {
+        // Find all directories that contain a `package.json` or a `turbo.json`.
+        // Gather a bit of additional metadata about them.
+        let potential_turbo_roots: Vec<_> = reference_dir
+            .ancestors()
+            .map(|path| {
+                let workspace_globs = PackageManager::Pnpm
+                    .get_workspace_globs(path)
+                    .unwrap_or_else(|_| {
```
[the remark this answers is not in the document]
nathanhammond: It's a `Result<Option<Globs>, Error>`; has to be unwrapped.

crates/turborepo-lib/src/shim.rs
```diff
@@ -227,114 +226,154 @@ pub struct RepoState {
...
+                    package_has_workspaces: workspace_globs.is_some(),
+                    workspace_globs,
+                }
+            })
+            .filter(|info| info.has_package_json || info.has_turbo_json)
+            .collect();
 
-            if is_workspace {
-                let local_turbo_state = LocalTurboState::infer(dir);
+        return potential_turbo_roots;
```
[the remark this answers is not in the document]
nathanhammond: This happened in a `clippy` update. 👍

crates/turborepo-lib/src/shim.rs
nathanhammond: This is a function I didn't know existed!

crates/turborepo-lib/src/shim.rs
nathanhammond: Thus far my approach here has been to hit things with a stick until rustc decides that I've followed the rules. I appreciate this category of feedback; it helps me build intuition on how to do things.

crates/turborepo-lib/src/shim.rs
nathanhammond: The test coverage actually made this really easy to do at high confidence, so I did it. 👍

crates/turborepo-lib/src/shim.rs
nathanhammond: Clippy didn't catch this one for some reason; I'm guessing because of the macro?

crates/turborepo-lib/src/shim.rs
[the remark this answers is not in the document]
NicholasLYang: Yeah this is a minor nit, just noting you could do:

```
                let workspace_globs = PackageManager::Pnpm
                    .get_workspace_globs(path)
                    .or_else(|_| PackageManager::Npm.get_workspace_globs(path))
                    .unwrap_or(None);
```

And it's a little less nested. But nbd either way!

NicholasLYang (approved):

PR #4050, opened 2023-03-02
Assign Framework to PackageTask struct

mehulkar:
Follow up to https://github.com/vercel/turbo/pull/3941. 

- In https://github.com/vercel/turbo/pull/3957#discussion_r1119512165, @chris-olszewski gave me good feedback that we should expose methods instead of internal maps from the `taskhash` package.
- In https://github.com/vercel/turbo/pull/4020, I saw a race condition and @gsoltis correctly pointed out that reads from the internal map needs to be protected by the mutex. The same race condition currently exists while reading `packageTaskFramework`.
- Assigning to `PackageTask` instead of accessing the map in DryRun is also better, because it will be easier to reuse in Real Run Summaries, and `.Command` also follows the same pattern already.


arlyon:
Excuse ignorance but what does Framework mean in this context?

mehulkar:
@arlyon good q, it’s a very generic word haha. It’s the detected web framework for a workspace: https://github.com/vercel/turbo/blob/main/cli/internal/inference/inference.go. 

We do this so we can auto include environment variables that *should* invalidate cache. https://turbo.build/repo/docs/core-concepts/caching#automatic-environment-variable-inclusion

nathanhammond (approved):
I'm not entirely sure why this was implemented differently to begin with. 🤷‍♂️

As an aside: we probably should have be using `sync.Map` for this. This is not a change request since the code already works and we're going to scuttle the Go portion.

arlyon (approved):

tknickman (approved):

PR #4053, opened 2023-03-02
avoid bundling and execution of modules just of watching

sokra:
### Description

* use completions to signal additional invalidations for evaluation
* avoid adding these modules as runtime entries, since that will unnecessarily code generate, bundle and execute these modules
* gets rid of the watch_files_hack

### Testing Instructions

* change postcss.config, next.config.js, etc. -> changes will be reflected by the app and process restarts

- [x] Auto label

### DIFF
--- a/crates/next-core/src/router.rs
+++ b/crates/next-core/src/router.rs
@@ -4,7 +4,7 @@ use serde::Deserialize;
 use serde_json::json;
 use turbo_tasks::{
     primitives::{JsonValueVc, StringsVc},
-    CompletionVc, Value,
+    CompletionVc, CompletionsVc, Value,
 };
 use turbo_tasks_fs::{
     json::parse_json_rope_with_source_context, to_sys_path, File, FileSystemPathVc,
@@ -12,6 +12,7 @@ use turbo_tasks_fs::{
 use turbopack::{evaluate_context::node_evaluate_asset_context, transition::TransitionsByNameVc};
 use turbopack_core::{
     asset::AssetVc,
+    changed::any_content_changed,
     chunk::dev::DevChunkingContextVc,
     context::{AssetContext, AssetContextVc},
     environment::{EnvironmentIntention::Middleware, ServerAddrVc},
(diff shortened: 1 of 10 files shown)


crates/next-core/src/router.rs
```diff
@@ -342,12 +343,12 @@ pub async fn route(
...
         context,
         intermediate_output_path,
-        Some(next_config),
+        None,
         vec![
             JsonValueVc::cell(request),
             JsonValueVc::cell(dir.to_string_lossy().into()),
         ],
-        routes_changed,
+        CompletionsVc::all(vec![next_config_changed, routes_changed]),
```
alexkirsz, line 351: Unrelated to the current PR, but it feels like `::all` should be a static method on `CompletionVc` instead of `CompletionsVc`.

crates/turbo-tasks/src/completion.rs
```diff
@@ -49,8 +49,19 @@ pub struct Completions(Vec<CompletionVc>);
...
 #[turbo_tasks::value_impl]
 impl CompletionsVc {
+    /// Merges multiple completions into one. The passed list will be part of
+    /// the cache key, so this function should not be used with varying lists.
+    ///
+    /// Varying lists should use `CompletionsVc::cell(list).completed()`
+    /// instead.
     #[turbo_tasks::function]
-    pub async fn all(self) -> anyhow::Result<CompletionVc> {
+    pub fn all(completions: Vec<CompletionVc>) -> CompletionVc {
```
alexkirsz, line 58: Unfortunate that we can't use `[CompletionVc; N]` here. This seems like a very easy mistake to make.

crates/turbopack-core/src/changed.rs
```diff
@@ -0,0 +1,43 @@
...
+    .map(content_changed)
+    .collect();
+
+    Ok(CompletionsVc::cell(completions).completed())
+}
+
+/// Returns a completion that changes when the content of the given asset
+/// changes.
+#[turbo_tasks::function]
+pub async fn content_changed(asset: AssetVc) -> Result<CompletionVc> {
```
alexkirsz, line 39: I wouldn't mark this as `pub` until we have another use for it.

alexkirsz (approved):

PR #4062, opened 2023-03-03
fix next.js source map resolving

sokra:
### Description

Before the error overlay didn't show the source mapped file.
Now it does.

* The wrapped ContentSources need to wrap a Rewrite result.
* Instead of returning a ContentSourceResult from a different ContentSource, one need to use Rewrite to get the data correct.

### Testing Instructions

Throw an error in an next.js app that triggers the error overlay with stack trace.


crates/turbopack-node/src/source_map/content_source.rs
```diff
@@ -81,7 +78,15 @@ impl ContentSource for NextSourceMapTraceContentSource {
             self_vc.await?.asset_source,
             NextSourceMapTraceContentProcessorVc::new(id, line, column, frame.name).into(),
         );
-        Ok(wrapped.as_content_source().get(path, Default::default()))
+        Ok(ContentSourceResultVc::exact(
+            ContentSourceContent::Rewrite(
```
jridgewell, line 82: I don't understand why these need to be rewrites. We're not expecting query params, right? We're not changing headers. This should be equivalent to just doing the `.get()` directly, right?

jridgewell (changes requested):

crates/turbopack-dev-server/src/source/wrapping_source.rs
```diff
@@ -117,6 +129,24 @@ impl GetContentSourceContent for WrappedGetContentSourceContent {
...
+            return Ok(ContentSourceContent::Rewrite(
+                Rewrite {
+                    path_and_query: rewrite.path_and_query.clone(),
+                    source: Some(
+                        WrappedContentSourceVc::new(
+                            rewrite.source.unwrap_or(self.inner_source),
+                            self.processor,
+                        )
+                        .into(),
+                    ),
```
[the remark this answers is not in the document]
sokra: We won't hit ourselves since we changed the `path`. So we need to make sure to always provide a source.

crates/turbopack-node/Cargo.toml
```diff
@@ -27,6 +27,7 @@ turbopack-core = { workspace = true }
 turbopack-dev-server = { workspace = true }
 turbopack-ecmascript = { workspace = true }
 url = { workspace = true }
+urlencoding = { workspace = true }
```
jridgewell, line 30:
```suggestion
```

crates/turbopack-dev-server/src/source/source_maps.rs
```diff
@@ -40,6 +42,17 @@ impl SourceMapContentSourceVc {
     }
 }
 
+fn encode_pathname_to_url(pathname: &str) -> String {
```
jridgewell, line 45: Can you add a comment on why we need to encode the pieces between `/`s?

jridgewell (approved):

PR #4066, opened 2023-03-03
Support exclusions in inputs

gsoltis:
### Description

 - Support exclusion globs (`!some-glob`) in `inputs` specifications
 - Modify file hashing for a package to only consider `git status` in the case where we aren't using `inputs`. `inputs` already uses `git hash-object` which operates on the state of the working tree, as opposed to `git ls-tree` which operates on a particular commit. This additionally fixes an issue where a dirty file not included in `inputs` could get included in the hash.
 - Fix manual file hashing to use the correct prefix on globs (and to support exclusions)
 - Modify globby to allow excluding individual files, not just directories

Note the `globby_test.go` behavior change: previously specifying `foo/**` would ignore files nested under `foo/`, it will now also include `foo` itself. This has the consequence that if `foo` is a file and not a directory, it is going from not-ignored to ignored.

We have behavior to force excludes in globby to apply to directories, but AFAICT we don't document this behavior. My preference would be to drop it so that we have a unified globbing syntax. However, other options include:
 1. The code I have here, with the aforementioned side-effect
 2. Special-casing user input that ends in `**` to maintain both behaviors

### Testing Instructions

A new integration test uses `inputs` with an exclusion to exclude a file from the hash. The excluded file is modified in the test, then hash matching and full turbo are verified.

Globby has a new test as well for excluding a single file, vs a directory. See above for additional globby test details

`taskhash_test.go` has an additional edge case to handle excluding files in `inputs` specifications that fall back to manual file hashing.

### DIFF
--- a/cli/integration_tests/inputs_test/monorepo/apps/my-app/package.json
+++ b/cli/integration_tests/inputs_test/monorepo/apps/my-app/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "my-app",
+  "scripts": {
+    "build": "echo 'building'"
+  },
+  "dependencies": {
+    "util": "*"
+  }
+}
--- a/cli/integration_tests/inputs_test/monorepo/package.json
+++ b/cli/integration_tests/inputs_test/monorepo/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "monorepo",
+  "workspaces": [
+    "apps/**",
+    "packages/**"
+  ]
+}
--- a/cli/integration_tests/inputs_test/monorepo/turbo.json
+++ b/cli/integration_tests/inputs_test/monorepo/turbo.json
@@ -1,0 +1,16 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "globalDependencies": ["foo.txt"],
+  "globalEnv": ["SOME_ENV_VAR"],
+  "pipeline": {
+    "build": {
+      "env": ["NODE_ENV"],
+      "outputs": []
+    },
+    // this comment verifies that turbo can read .json files with comments
+    "my-app#build": {
+      "inputs": ["*.txt", "!excluded.txt"],
+      "outputs": ["banana.txt", "apple.json"]
+    }
+  }
+}
--- a/cli/integration_tests/inputs_test/run.t
+++ b/cli/integration_tests/inputs_test/run.t
@@ -1,0 +1,37 @@
+Setup
+  $ . ${TESTDIR}/../setup.sh
+  $ . ${TESTDIR}/setup.sh $(pwd)
+
+Running build for my-app succeeds
+  $ ${TURBO} run build --filter=my-app
+  \xe2\x80\xa2 Packages in scope: my-app (esc)
+  \xe2\x80\xa2 Running build in 1 packages (esc)
+  \xe2\x80\xa2 Remote caching disabled (esc)
+  my-app:build: cache miss, executing b7cf611b2774bb9a
+  my-app:build: 
+  my-app:build: > build
+  my-app:build: > echo 'building'
+  my-app:build: 
+  my-app:build: building
+  
+   Tasks:    1 successful, 1 total
+  Cached:    0 cached, 1 total
+    Time:\s*[\.0-9]+m?s  (re)
+  
+Update exluded file and try again
+  $ echo "new excluded value" > apps/my-app/excluded.txt
+  $ ${TURBO} run build --filter=my-app
+  \xe2\x80\xa2 Packages in scope: my-app (esc)
+  \xe2\x80\xa2 Running build in 1 packages (esc)
+  \xe2\x80\xa2 Remote caching disabled (esc)
+  my-app:build: cache hit, replaying output b7cf611b2774bb9a
+  my-app:build: 
+  my-app:build: > build
+  my-app:build: > echo 'building'
+  my-app:build: 
+  my-app:build: building
+  
+   Tasks:    1 successful, 1 total
+  Cached:    1 cached, 1 total
+    Time:\s*[\.0-9]+m?s >>> FULL TURBO (re)
+  
--- a/cli/internal/globby/globby.go
+++ b/cli/internal/globby/globby.go
@@ -101,7 +101,10 @@ func globWalkFs(fsys iofs.FS, fsysRoot string, basePath string, includePatterns 
 		// This will not error as it follows the call to checkRelativePath.
 		iofsRelativePath, _ := fs.IofsRelativePath(fsysRoot, excludePath)
 
-		// Excludes operate on entire folders.
+		// In case this is a file pattern and not a directory, add the exact pattern.
+		processedExcludes = append(processedExcludes, iofsRelativePath)
+		// TODO: we need to either document or change this behavior
+		// Excludes operate on entire folders, so we also exclude everything under this in case it represents a directory
 		processedExcludes = append(processedExcludes, filepath.Join(iofsRelativePath, "**"))
 	}
 
--- a/cli/internal/globby/globby_test.go
+++ b/cli/internal/globby/globby_test.go
@@ -414,7 +414,6 @@ func TestGlobFilesFs(t *testing.T) {
 			wantAll: []string{
 				"/repos/some-app/dist",
 				"/repos/some-app/dist/index.html",
-				"/repos/some-app/dist/js",
 			},
 			wantFiles: []string{
 				"/repos/some-app/dist/index.html",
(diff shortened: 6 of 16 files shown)


cli/internal/globby/globby.go
```diff
@@ -101,7 +101,10 @@ func globWalkFs(fsys iofs.FS, fsysRoot string, basePath string, includePatterns
 		// This will not error as it follows the call to checkRelativePath.
 		iofsRelativePath, _ := fs.IofsRelativePath(fsysRoot, excludePath)
 
-		// Excludes operate on entire folders.
+		// In case this is a file pattern and not a directory, add the exact pattern.
+		processedExcludes = append(processedExcludes, iofsRelativePath)
+		// TODO: we need to either document or change this behavior
+		// Excludes operate on entire folders, so we also exclude everything under this in case it represents a directory
 		processedExcludes = append(processedExcludes, filepath.Join(iofsRelativePath, "**"))
```
tknickman, line 108: Agree with you, it would be great to have one unified globbing syntax. However since this is a behavior change we should probably special case it so we can still support the existing behavior, and then move to deprecate that fully in a future version.

tknickman (approved):

cli/internal/globby/globby_test.go
```diff
@@ -660,6 +659,24 @@ func TestGlobFilesFs(t *testing.T) {
...
+				includePatterns: []string{"*.txt"},
+				excludePatterns: []string{"excluded.txt"},
+			},
+			wantAll: []string{
+				"/repos/some-app/included.txt",
+			},
+			wantFiles: []string{
+				"/repos/some-app/included.txt",
+			},
+		},
```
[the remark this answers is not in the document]
gsoltis: I don't think this test is correct, as the directories don't match the include pattern, so I wouldn't expect them to be in the `wantAll` list.

gsoltis:
I've updated the PR to maintain the current behavior while also allowing for excluding files. So there is no longer a change to the globby test, just the addition of single files to it.

I continue to think we should fix the special case for exclusions, since it's odd to support two different globbing syntaxes, especially in the same config field, but I will leave that for a separate PR.

cli/internal/globby/globby.go
```diff
@@ -101,7 +101,10 @@ func globWalkFs(fsys iofs.FS, fsysRoot string, basePath string, includePatterns
 		// This will not error as it follows the call to checkRelativePath.
 		iofsRelativePath, _ := fs.IofsRelativePath(fsysRoot, excludePath)
 
-		// Excludes operate on entire folders.
+		// In case this is a file pattern and not a directory, add the exact pattern.
+		processedExcludes = append(processedExcludes, iofsRelativePath)
+		// TODO: we need to either document or change this behavior
```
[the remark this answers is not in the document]
gsoltis: I'm not a fan of `"outputs": ["glob-syntax-1", "!glob-syntax-2"]`. We're not using `.gitignore` syntax for inclusions, so I don't think we should use it for exclusions either.

cli/integration_tests/inputs_test/monorepo/turbo.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "globalDependencies": ["foo.txt"],
+  "globalEnv": ["SOME_ENV_VAR"],
+  "pipeline": {
+    "build": {
+      "env": ["NODE_ENV"],
+      "outputs": []
+    },
+    // this comment verifies that turbo can read .json files with comments
```
mehulkar, line 10: can we delete this, we already have a test for it

mehulkar, line 10: can we delete this, we already have a test for it

cli/integration_tests/inputs_test/monorepo/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "monorepo",
+  "workspaces": [
+    "apps/**",
+    "packages/**"
```
mehulkar, line 5: can we make this fixture simpler and not have packages?

cli/integration_tests/inputs_test/monorepo/apps/my-app/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "my-app",
+  "scripts": {
+    "build": "echo 'building'"
+  },
+  "dependencies": {
+    "util": "*"
```
mehulkar, line 7: can we make this fixture simpler and not have a dependency?

cli/integration_tests/inputs_test/monorepo/turbo.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "globalDependencies": ["foo.txt"],
+  "globalEnv": ["SOME_ENV_VAR"],
```
mehulkar, line 4: are either of these relevant?

cli/integration_tests/inputs_test/monorepo/turbo.json
```diff
@@ -0,0 +1,16 @@
...
+  "pipeline": {
+    "build": {
+      "env": ["NODE_ENV"],
```
mehulkar, line 7: probably not relevant?

cli/integration_tests/inputs_test/monorepo/turbo.json
```diff
@@ -0,0 +1,16 @@
...
+      "outputs": []
+    },
+    // this comment verifies that turbo can read .json files with comments
+    "my-app#build": {
```
mehulkar, line 11: just use the `build` task, since we don't need the util package

cli/integration_tests/inputs_test/run.t
```diff
@@ -0,0 +1,37 @@
...
+  my-app:build: 
+  my-app:build: building
+  
+   Tasks:    1 successful, 1 total
+  Cached:    0 cached, 1 total
+    Time:\s*[\.0-9]+m?s  (re)
+  
+Update exluded file and try again
+  $ echo "new excluded value" > apps/my-app/excluded.txt
+  $ ${TURBO} run build --filter=my-app
```
mehulkar, line 23: don't need `--filter` if we make the fixture simpler

mehulkar (commented):
some drive by comments about the test fixture. i would like to keep them as simple as possible for the test at hand.

cli/internal/globby/globby_test.go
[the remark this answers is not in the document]
nathanhammond: d'oh, I wrote the test in GitHub, I updated the includePatterns in the above example. Mostly wanting a bit of nestedness in a test case.

cli/internal/globby/globby.go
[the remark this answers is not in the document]
nathanhammond: I'm 100% on board with aligning the syntax, I was pointing out that we could elect to align in the _other_ direction, which would have the bonus feature of being (likely) backwards compatible.

So that would result in these two being identical:
- `includes: "foo"` (where foo is a dir) 
- `includes: "foo/**"`

Unfortunately we didn't really design ourselves into our current state and attaching a design now is much more complicated.

cli/internal/globby/globby.go
```diff
@@ -101,7 +101,13 @@ func globWalkFs(fsys iofs.FS, fsysRoot string, basePath string, includePatterns
 		// This will not error as it follows the call to checkRelativePath.
 		iofsRelativePath, _ := fs.IofsRelativePath(fsysRoot, excludePath)
 
-		// Excludes operate on entire folders.
+		// In case this is a file pattern and not a directory, add the exact pattern.
+		// In the event that the user has already specified /**,
+		if !strings.HasSuffix(iofsRelativePath, "**") {
```
nathanhammond, line 106: So we still add `/**` to everything, but in the event that your current thing does not end in `**` we _also_ add it to the exclusions without appending `/**`.

I'd love to get tests constraining the behavior for more situations:
- `**/` => gets added as `**/` and `**/**`?
- `foo**` => gets added as `foo**/**` => is this a gap in the per-file exclusion?
- `**foo` => gets added as `**foo`, `**foo/**`

nathanhammond (approved):
🛠🚢: we need more tests around the edges of this behavior.

I don't see anything wrong with the implementation, but I can't be sure how it functions in all scenarios once plumbed through doublestar without more tests. We'll need those tests come porting time.

cli/internal/globby/globby.go
gsoltis: > `foo**` => gets added as `foo**/**` => is this a gap in the per-file exclusion?
Good catch. I updated the suffix that we check for from `**` -> `/**`. Paths are already required to be in posix format.

cli/internal/globby/globby.go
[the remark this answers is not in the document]
gsoltis: > attaching a design now is much more complicated.

Definitely agreed

Supporting `.gitignore` (and/or others? `.dockerignore`?) is an interesting idea. If it's breaking, we might need to wait for 2.0. If it's not, that gives us more flexibility....

cli/internal/globby/globby.go
gsoltis: Paths are apparently not required to be in posix format. Now using the appropriate system-specific suffix to look for.

PR #3903, opened 2023-02-22
enable rust codepath in release process

arlyon:
This enables the 'go sandwich' by flipping the go tag in our release pipeline.

PR #4068, opened 2023-03-03
Revert "enable rust codepath in release process"

gsoltis:
Reverts vercel/turbo#3903

Flag rust library off until we sort out compatibility issues


arlyon (approved):
I'll download the canary from npm on Monday and sort out any issues.

Issue #4034, opened 2023-03-01
[turborepo] Non-deterministic `turbo` change detection when using certain workspace configs


maschwenk:
### What version of Turborepo are you using?

1.8.3

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

![image](https://user-images.githubusercontent.com/1264305/222206336-21f4b6f8-a346-449b-a745-a077e2ceaabc.png)

Multiple invocations of `turbo --dry` are giving non-deterministic results. To isolate variables I've disconnected from the internet, used `--force` etc. None of that seemed to make it stable.

I spoke about this with @mehulkar and @gsoltis privately.

### Expected Behavior

If all things equal, multiple invocations of a build should give deterministic results.

### To Reproduce

Don't currently have time to create a reproduction as it's embedded in a very complex monorepo. Turbo setup is pretty vanilla.

```json
{
  "$schema": "https://turbo.build/schema.json",
  "pipeline": {
    "build": {
      "dependsOn": ["^build"],
      "outputs": ["dist/**", "build/**"]
    },
}
```

maschwenk:
```yaml
  - 'apple'
  - 'apple/deployments/*'
  - 'apple/packages/**'
```
I think was the problematic setup in `pnpm-workspace.yaml`. There are `package.json`s at all these levels. I changed it to:

```yaml
  - 'apple/**/*'
```

Which seems stable but y'all do mention in docs that nested workspaces are not supported.

gsoltis:
I'm going to close this in favor of #4178 since it turned out to be a feature that's not currently supported.

maschwenk:
👍🏼 sounds good to me!

Issue #3964, opened 2023-02-25
--log-prefix=pad


gajus:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

Pad logs to the length of the longest workspace+task name.

Consider these logs:

```
contra-temporal:dev: cache bypass, force executing 3d7c2284345252a4
contra-temporal:dev:
contra-temporal:dev: > contra-temporal@ dev /Users/gajus/Developer/contra/gaia-og1/apps/contra-temporal
contra-temporal:dev: > docker compose --profile temporal up
contra-temporal:dev:
contra-redis:dev:
contra-redis:dev: > contra-redis@ dev /Users/gajus/Developer/contra/gaia-og1/apps/contra-redis
contra-redis:dev: > docker run --name contra-redis --rm -p 6382:6379 redis
contra-redis:dev:
contra-database:dev:
contra-database:dev: > contra-database@ dev /Users/gajus/Developer/contra/gaia-og1/apps/contra-database
contra-database:dev: > docker build -t contra-database -f Dockerfile --build-arg DIFF=false ../.. && docker run -p 5435:5432 --name contra-database --rm contra-database
contra-database:dev:
contra-redis:dev: docker: Error response from daemon: Conflict. The container name "/contra-redis" is already in use by container "ad4d87a63c04d75c6c0f117275a261153c810cbbb2b13aabda8c0aaaf5868d74". You have to remove (or rename) that container to be able to reuse that name.
contra-redis:dev: See 'docker run --help'.
contra-redis:dev:  ELIFECYCLE  Command failed with exit code 125.
contra-redis:dev:  WARN   Local package.json exists, but node_modules missing, did you mean to install?
contra-redis:dev: ERROR: command finished with error: command (/Users/gajus/Developer/contra/gaia-og1/apps/contra-redis) pnpm run dev exited (1)
```

If `contra-temporal:dev` is the longest name in the repo, then with the proposed option, these logs would be adjusted to:

```
contra-temporal:dev: cache bypass, force executing 3d7c2284345252a4
contra-temporal:dev:
contra-temporal:dev: > contra-temporal@ dev /Users/gajus/Developer/contra/gaia-og1/apps/contra-temporal
contra-temporal:dev: > docker compose --profile temporal up
contra-temporal:dev:
contra-redis:dev   :
contra-redis:dev   : > contra-redis@ dev /Users/gajus/Developer/contra/gaia-og1/apps/contra-redis
[...]

PR #1707, opened 2022-08-17
Run Summary Prototype

gsoltis:

PR #4069, opened 2023-03-03
Generate a run summary file on turbo run

mehulkar:


cli/internal/runsummary/run_summary.go
```diff
@@ -16,7 +25,21 @@ type RunSummary struct {
...
+// Save saves the run summary to a file
+func (summary *RunSummary) Save(singlePackage bool) error {
+	json, err := summary.FormatJSON(singlePackage)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("[debug] got my json for run summar\n")
+
+	nonce := "abcdef"
+	filename := fmt.Sprintf("turbo-run-%s-%s.json", nonce, time.Now().Format(time.RFC3339))
```
tknickman, line 40: This needs to be prefixed by `vercel-` when in the Vercel CI env (`VERCEL=1`) for us to pick it up automatically.

cli/internal/runsummary/run_summary.go
```diff
@@ -16,7 +25,19 @@ type RunSummary struct {
...
+}
+
+// Save saves the run summary to a file
+func (summary *RunSummary) Save(singlePackage bool) error {
+	json, err := summary.FormatJSON(singlePackage)
+	if err != nil {
+		return err
+	}
+
+	nonce := "abcdef"
```
mehulkar, line 38: i'm thinking we need something determinstic here, so you can run 5 run summaries and (1) identify them to all be at the same state, but (2) not overwrite each other. the timestamp lets you not overwrite, but then if you generate these files across commits/branches, you don't know which is which. 

We could also include this in the file contents itself, but I think we need a non-git solution also / something that comes from turbo itself.

cli/internal/runsummary/run_summary.go
mehulkar: done

cli/internal/nodes/packagetask.go
```diff
@@ -21,10 +19,7 @@ type PackageTask struct {
 	Outputs         []string
 	ExcludedOutputs []string
 	LogFile         string
-	ExpandedInputs  map[turbopath.AnchoredUnixPath]string
 	Hash            string
-	HashedEnvVars   env.DetailedMap
-	Framework       string
```
mehulkar, line 27: these pieces were originally added just to pass on to a task summary. we don't need them anymore

cli/internal/runsummary/run_summary.go
```diff
@@ -16,7 +25,19 @@ type RunSummary struct {
...
+	filename := fmt.Sprintf("vercel-turbo-run-%s-%s.json", nonce, time.Now().Format(time.RFC3339))
```
mehulkar, line 39: this should _probably_ always go in the root of the directory. e.g. if we're running turbo from a workspace, we need to handle that case.

cli/internal/graph/graph.go
```diff
@@ -152,6 +179,6 @@ func (g *CompleteGraph) GetPackageJSONFromWorkspace(workspaceName string) (*fs.P
 
 // repoRelativeLogFile returns the path to the log file for this task execution as a
 // relative path from the root of the monorepo.
-func repoRelativeLogFile(pt *nodes.PackageTask) string {
-	return filepath.Join(pt.Pkg.Dir.ToStringDuringMigration(), ".turbo", fmt.Sprintf("turbo-%v.log", pt.Task))
+func repoRelativeLogFile(dir turbopath.AnchoredSystemPath, taskName string) string {
```
mehulkar, line 182: changed this so getting the logfile doesn't depend on the `packageTask` struct, since it's also used by taskSummary  now

cli/internal/runsummary/run_summary.go
```diff
@@ -16,7 +25,19 @@ type RunSummary struct {
...
+	return ioutil.WriteFile(filename, json, 0644)
```
nathanhammond, line 40: We need to write to either `<cwd>/.turbo` or change the things we `.gitignore` otherwise `turbo run build && turbo run build` will not play nicely.

cli/internal/runsummary/run_summary.go
nathanhammond, line 39: We need to enable this to be shared without collision. We should probably consider using a KSUID:

https://github.com/segmentio/ksuid

cli/internal/runsummary/run_summary.go
nathanhammond: There are two types things we care about:
- linear (e.g. not-Git)
- non-linear (switch branch)

Linear will "just work" and we don't care. Non-linear is really "facts of the run" type situation. It has to be paired with the _local_ repo for it to be meaningful in all scenarios. Yeah, it makes it easier to dig out older things if the information is in the file name, but git history is a messy thing. (rebase, amend)

nathanhammond (changes requested):

cli/internal/runsummary/run_summary.go
mehulkar: >  It has to be paired with the local repo for it to be meaningful in all scenarios

I think you're optimizing for the case where thees files are shared between teammates. Is that right?

I'm imagining the scenario where you're debugging locally and you want to run turbo many times and compare. In this implementation, that would be pretty tedious, because you have to generate the summary, rename it to something, then run again. If you miss that step, you have to guess based on the timestamp. Having debugged recently with `--profile` (which similarly does not have anything helpful in the filename), I'd really like to help the developer out for this case.

cli/internal/runsummary/run_summary.go
mehulkar: good point. I've moved this into `.turbo/runs` as #1707 did also. Started a thread in slack as well because it impacts vercel integration

cli/internal/runsummary/run_summary.go
mehulkar: Read through [segments blog post](https://segment.com/blog/a-brief-history-of-the-uuid/) about KSUIDs, very interesting! Good with me to use it, although I'm opting to keep the timestamp as well so it's a tad easier to differentiate which file is which

cli/internal/runsummary/run_summary.go
mehulkar: That said, @gaspar09/@mknichel want to include repo/git metadata into the run summary anyway, so that could cover some of my use case. Let's talk more about this internally

nathanhammond: Not even just shared! After a rebase or a commit if we had included anything from `git` in the file name it's ~meaningless unless somebody is a regular user of reflogs. Further, since we're not building from HEAD, we're also including any uncommitted content so having the file name include, say, the previous SHA, is at least misleading (in my most–charitable interpretation).

The UX for what you're describing sounds like you want a `--output-file-prefix` option which you could consistently press ⬆️ in your shell to rerun the same command and have it spit out a reasonable file name. Since it doesn't make sense to run `turbo` over and over _without_ making changes to files I'm pretty strongly opposed to grouping by git-anything. Either you're excluding the state of the index compared to the file system, you made a commit in-between and your proposed auto-prefix is nonsense, or `git checkout -b renamed-branch` (same SHA) and it's no longer relevant.

The only one I'd consider possibly viable as, say, a default value for `--output-filename-prefix` is `git symbolic-ref --short HEAD` ... which is a lot of fun because lots of folks have branches that look like `nathanhammond/feature-name` and now that's a huge pain of a filename.

mehulkar: > Since it doesn't make sense to run turbo over and over without making changes to files

This is exactly the case I want to capture (example debugging use case: https://github.com/vercel/turbo/issues/4034)

1. Run turbo 5 times (without making any changes) and get 5 run summaries.
2. Make a change
3. Run turbo 5 times again and get 5 more run summaries

I want it to be obvious which set of 5 is which.

> The UX for what you're describing sounds like you want a `--output-file-prefix`

That works, but I'm thinking `TURBO_RUN_SUMMARY_DIR` as an env var instead, since we want that for another use case too)

> so having the file name include [...] the previous SHA, is [...] misleading
> I'm pretty strongly opposed to grouping by git-anything

agreed, I wasn't proposing grouping by git formally, just that if I were debugging I could _probably_ scrape by with that information (in lieu of filenames)

nathanhammond (approved):
🛠🚢

cli/internal/runsummary/run_summary.go
```diff
@@ -2,21 +2,63 @@
...
+		ID:                ksuid.New(),
+		TurboVersion:      turboVersion,
+		Packages:          packages,
+		Tasks:             &[]*TaskSummary{},
+		GlobalHashSummary: globalHashSummary,
+	}
+}
+
+// Save saves the run summary to a file
+func (summary *RunSummary) Save(dir turbopath.AbsoluteSystemPath, singlePackage bool) error {
```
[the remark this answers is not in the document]
mehulkar: +10000. but out of scope in this PR. It comes from it being passed into `dry_run.go` , which I carried into RunSummary, keeping it as an arg. We can easily add it to `RunSummary` itself. Will do this in a different PR.

That said, this isn't even the right final answer.... `singlePackage` is basically used to strip package names from tasks, we _should_ structure it internally with a formatter callback or something (something that might be useful for other use cases also https://github.com/vercel/turbo/issues/3964#issuecomment-1447223571)

cli/internal/runsummary/run_summary.go
```diff
@@ -2,21 +2,63 @@
...
+	json, err := summary.FormatJSON(singlePackage)
+	if err != nil {
+		return err
+	}
+
+	filename := fmt.Sprintf("%s-%d.json", summary.ID, time.Now().UnixMilli())
```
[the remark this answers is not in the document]
mehulkar: I feel like this is useful, especially if I switch it around, so it will be sorted with `ls`. I would find it very valuable

cli/internal/run/run.go
```diff
@@ -345,25 +349,23 @@ func (r *run) run(ctx gocontext.Context, targets []string) error {
 		}
 	}
 
+	// RunSummary contains information that is statically analyzable about
+	// the tasks that we expect to run based on the user command.
+	// Currently, we only emit this on dry runs, but it may be useful for real runs later also.
```
mehulkar, line 354:
```suggestion
```

cli/internal/runsummary/run_summary.go
```diff
@@ -2,21 +2,73 @@
...
 type RunSummary struct {
+	ID                ksuid.KSUID        `json:"id"`
 	TurboVersion      string             `json:"turboVersion"`
 	GlobalHashSummary *GlobalHashSummary `json:"globalHashSummary"`
 	Packages          []string           `json:"packages"`
-	Tasks             []TaskSummary      `json:"tasks"`
+	Tasks             *[]*TaskSummary    `json:"tasks"`
+}
+
+// NewRunSummary returns a RunSummary instance with some a ksuid for the ID
```
mehulkar, line 34:
```suggestion
// NewRunSummary returns a RunSummary instance
```

cli/internal/run/dry_run.go
```diff
@@ -153,7 +129,7 @@ func executeDryRun(ctx gocontext.Context, engine *core.Engine, g *graph.Complete
 		return nil, errors.New("errors occurred during dry-run graph traversal")
 	}
 
-	return taskIDs, nil
+	return &taskIDs, nil
```
gsoltis, line 132: I don't think this needs to be a pointer

cli/internal/run/real_run.go
```diff
@@ -88,12 +90,16 @@ func RealRun(
...
-	execFunc := func(ctx gocontext.Context, packageTask *nodes.PackageTask) error {
+	taskSummaries := []*runsummary.TaskSummary{}
+	execFunc := func(ctx gocontext.Context, packageTask *nodes.PackageTask, taskSummary *runsummary.TaskSummary) error {
 		deps := engine.TaskGraph.DownEdges(packageTask.TaskID)
+		taskSummaries = append(taskSummaries, taskSummary)
 		// deps here are passed in to calculate the task hash
 		return ec.exec(ctx, packageTask, deps)
 	}
 
+	runSummary.Tasks = &taskSummaries
```
gsoltis, line 101: I don't think this needs to be a pointer

gsoltis (changes requested):

cli/internal/runsummary/format_json.go
```diff
@@ -7,31 +7,31 @@ import (
...
+	for i, task := range *summary.Tasks {
 		singlePackageTasks[i] = task.toSinglePackageTask()
 	}
 
 	spSummary := &singlePackageRunSummary{Tasks: singlePackageTasks}
 
 	bytes, err := json.MarshalIndent(spSummary, "", "  ")
 	if err != nil {
-		return "", errors.Wrap(err, "failed to render JSON")
+		return []byte{}, errors.Wrap(err, "failed to render JSON")
```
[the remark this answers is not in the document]
mehulkar: works, but curious, why does `nil` satisfy `[]byte` type?

cli/internal/runsummary/run_summary.go
```diff
@@ -2,21 +2,64 @@
...
 const MissingFrameworkLabel = "<NO FRAMEWORK DETECTED>"
 
 // RunSummary contains a summary of what happens in the `turbo run` command and why.
 type RunSummary struct {
+	ID                ksuid.KSUID        `json:"id"`
 	TurboVersion      string             `json:"turboVersion"`
 	GlobalHashSummary *GlobalHashSummary `json:"globalHashSummary"`
 	Packages          []string           `json:"packages"`
-	Tasks             []TaskSummary      `json:"tasks"`
+	Tasks             *[]*TaskSummary    `json:"tasks"`
```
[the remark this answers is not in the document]
mehulkar: i had tried this without a pointer, but the `tasks` end up being an empty array when I do that. I'm not sure why. The diff is here if you see anything obvious: https://github.com/vercel/turbo/pull/4088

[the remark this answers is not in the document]
mehulkar: never mind, i was assign too early

gsoltis (approved):

PR #4077, opened 2023-03-05
feat: add sourcemap for postcss transform

ahabhgk:

### DIFF
--- a/crates/turbopack-core/src/source_transform.rs
+++ b/crates/turbopack-core/src/source_transform.rs
@@ -1,10 +1,24 @@
 use anyhow::Result;
 
-use crate::asset::AssetVc;
+use crate::{asset::AssetVc, source_map::SourceMapVc};
 
 #[turbo_tasks::value_trait]
 pub trait SourceTransform {
-    fn transform(&self, source: AssetVc) -> AssetVc;
+    fn transform(&self, source: SourceTransformedVc) -> SourceTransformedVc;
+}
+
+#[turbo_tasks::value]
+pub struct SourceTransformed {
+    pub source: AssetVc,
+    pub source_map: Option<SourceMapVc>,
+}
+
+#[turbo_tasks::value_impl]
+impl SourceTransformedVc {
+    #[turbo_tasks::function]
+    pub fn new(source: AssetVc, source_map: Option<SourceMapVc>) -> Self {
+        SourceTransformed { source, source_map }.cell()
+    }
 }
 
 #[turbo_tasks::value(transparent)]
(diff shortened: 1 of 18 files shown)


crates/turbopack-core/src/source_transform.rs
```diff
@@ -1,10 +1,24 @@
...
 pub trait SourceTransform {
-    fn transform(&self, source: AssetVc) -> AssetVc;
+    fn transform(&self, source: SourceTransformedVc) -> SourceTransformedVc;
+}
+
+#[turbo_tasks::value]
+pub struct SourceTransformed {
+    pub source: AssetVc,
+    pub source_map: Option<SourceMapVc>,
+}
```
sokra, line 14: I wonder if our `AssetVc` should have a `source_map(): OptionSourceMapVc` instead?

sokra: Or our `AssetContent` should have a `source_map: Option<SourceMapVc>` field?

ahabhgk: Both way seems hold a source map on `AssetVc`, if so should `AssetVc::source_map` or `AssetContent.source_map` needs also generate a "output source map" (`CssChunkSourceMapAsset`), currently the output source map needs implement `Asset` (`impl Asset for CssChunkSourceMapAsset`), or is it only suitable for "input source map" (source map generated by loader, source map in node_modules)?

For this two option I would choose `AssetVc::source_map(): OptionSourceMapVc`, source map can remain lazily generate, not sure `AssetContent.source_map` could do that.

sokra: Just remembered... We already have a `GenerateSourceMap` trait that one can optionally implement to allow to get a source map for an Asset. So I just you could implement that trait on the `PostCssTransformedAsset` to allow to get a source map.

ahabhgk: Tried to implement `GenerateSourceMap` for `PostCssTransformedAsset` before, but the problem is source map could be `None` when error happened in postcss transforming, and not all loaders must return a source map (off-topic from this PR), but `GenerateSourceMap` can't return a `None` for source map.

ahabhgk: For "optionally implement" do you mean optionally implement `GenerateSourceMap` for different Asset based on the transforming result, for example

```rust
enum ProcessPostCssResult {
  NotFound(PostcssNotFoundAssetVc),
  Transformed(PostCssTransformedAssetVc),
}
impl GenerateSourceMap for PostCssTransformedAssetVc {/* ... */}
impl Asset for PostCssTransformedAssetVc {/* ... */}
impl Asset for PostcssNotFoundAssetVc {/* ... */}
```

anthonyshew:

PR #4081, opened 2023-03-06
place all pages in the same intermediate output directory

sokra:
### Description

The isolation is unnecessary and we can reuse files when we share directories

### Testing Instructions

Open multiple pages without client-side navigation in multiple tabs concurrently.


jridgewell (approved):

sokra:
Non-determinism in graph traversal need to be fixed first before merging that

PR #4086, opened 2023-03-06
next/font: Generate fallback fonts

wbinnssmith:
This:
* Generates and uses definitions for fallback system fonts to use in place of user fonts, or as they are loading.
* Reads the font metrics file from the `next` package directly relative to the user's app.
* Fixes a bug where `adjust_font_fallback` defaulted to `false` instead of `true`.
* Begins to refactor next/font code to be more modular in anticipation of `next/font/local` support.
* Changes `handle_resolve_error` to take an origin `FileSystemPathVc` instead of a `ResolveOriginVc` as this is all it used, and a resolve origin isn't always available.

### DIFF
--- a/crates/next-core/src/next_font_google/mod.rs
+++ b/crates/next-core/src/next_font_google/mod.rs
@@ -23,7 +23,7 @@ use self::{
     font_fallback::{get_font_fallback, FontFallback, FontFallbackVc},
     options::{FontWeights, NextFontGoogleOptionsVc},
     stylesheet::build_stylesheet,
-    util::{get_scoped_font_family, FontFamilyType, FontFamilyTypeVc},
+    util::{get_scoped_font_family, FontFamilyType},
 };
 use crate::{
     embed_js::next_js_file_path,
@@ -34,6 +34,7 @@ use crate::{
 };
 
 pub(crate) mod font_fallback;
+pub(crate) mod issue;
 pub(crate) mod options;
 pub(crate) mod request;
 pub(crate) mod stylesheet;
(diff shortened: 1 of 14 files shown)


crates/next-core/src/next_font_google/options.rs
```diff
@@ -24,32 +25,37 @@ pub struct NextFontGoogleOptions {
     pub subsets: Option<Vec<String>>,
 }
 
+impl NextFontGoogleOptionsVc {
+    pub fn new(options: NextFontGoogleOptions) -> Self {
```
jridgewell, line 29: Is this supposed to be a turbo function? Right now, the cell is not cached based on `options`, so every call gets a brand new cell.

crates/next-core/src/next_font_google/mod.rs
```diff
@@ -389,7 +383,7 @@ async fn font_options_from_query_map(query: QueryMapVc) -> Result<NextFontGoogle
         };
 
     self::options::options_from_request(&parse_json_with_source_context(json)?, &FONT_DATA)
-        .map(NextFontGoogleOptionsVc::new)
+        .map(NextFontGoogleOptionsVc::cell)
```
jridgewell, line 386: This still has the same bug. Every time this is called, we're creating a new instance of the cell. If the input value is expected to be equivalent, we want to implement a `#[turbo_tasks::function]fn new(opts: NextFontGoogleOptions) -> NextFontGoogleOptionsVc {}

jridgewell (approved):

PR #4087, opened 2023-03-06
Added alpine specific error message around installing libc6-compat

NicholasLYang:
### Description

If users forget to install `libc6-compat`, they get a "binary does not exist" error, which is very confusing. This PR detects the operating system and prints a message on Alpine Linux asking if users forgot to install `libc6-compat`.

### Testing Instructions

I can't get a Rust binary to build properly on alpine so we might have to test this via a canary unfortunately.


tknickman (approved):

arlyon (approved):

mehulkar (commented):
@NicholasLYang Do we still want this? Our support policy only calls out [Debian based distros](https://turbo.build/repo/docs/support), and Alpine isn't one. Feels weird to have a specific error message for a single distro. (Not against it, just wanted to check in again!)

PR #4090, opened 2023-03-07
Add globalenv to run summary

mehulkar:
### Description

Adds a list of environment variables that affect every task hash into the run summary. These could be configured via `globalEnv` or from turbo itself.

This PR does 3 things:

1. Refactors `env.go` to be able to get env vars matching a regex (and adapts the existing framework prefix matchers to use that)
    1. in a second commit, it uses this new affordance to get global env vars
1. Changes `calculateGlobalHash` implementation to use a named struct and convert to the anonymous struct for hashing. This allows us to keep a handle on `env.DetailedMap`, so we can get "secret" hashed values  for run summaries.
1. Finally, adds the `global` env vars to each task summary.
    
    I chose to add these env vars to the task summary, instead in the global space, so that a single `TaskSummary` contains _everything_ needed to diff between runs[^1] .

## How to Review

- I would recommend reviewing one commit at a time!
- One missing piece here is that `THASH` and `VERCEL_ANALYTICS_ID` are included in `global` part of run summary. It might be useful to separate these into another source? 

[^1]: might revisit `globalFileHashMap` to also do this, but need to figure out the schema and how it interacts with `expandedInputs`


cli/internal/run/global_hash.go
```diff
@@ -100,30 +132,16 @@ func calculateGlobalHash(rootpath turbopath.AbsoluteSystemPath, rootPackageJSON
...
 	}
 
-	globalHashable := struct {
-		globalFileHashMap    map[turbopath.AnchoredUnixPath]string
-		rootExternalDepsHash string
-		hashedSortedEnvPairs []string
-		globalCacheKey       string
-		pipeline             fs.PristinePipeline
-	}{
+	return &GlobalHashable{
```
NicholasLYang, line 121: Any reason we're using a pointer here instead of the struct directly? Mostly just asking out of curiosity

cli/internal/env/env.go
```diff
@@ -81,7 +82,7 @@ func getEnvMap() EnvironmentVariableMap {
 	return envMap
 }
 
-// fromPrefixes returns a map of env vars and their values based on include/exclude prefixes
+// fromKeys returns a map of env vars and their values based on include/exclude prefixes
```
NicholasLYang, line 95: Is this comment still correct? Is it from `include/exclude prefixes` still?

cli/internal/env/env.go
```diff
@@ -91,44 +92,55 @@ func fromKeys(all EnvironmentVariableMap, keys []string) EnvironmentVariableMap
...
+	if len(compileFailures) > 0 {
+		return nil, fmt.Errorf("The following env prefixes failed to compile to regex: %s", strings.Join(compileFailures, ", "))
+	}
+
+	return output, nil
 }
 
 // GetHashableEnvVars returns all sorted key=value env var pairs for both frameworks and from envKeys
-func GetHashableEnvVars(keys []string, prefixes []string) DetailedMap {
+func GetHashableEnvVars(keys []string, matchers []string, excludePrefix string) (DetailedMap, error) {
```
NicholasLYang, line 136: I'm a little confused as to the difference between the `excludePrefix` passed here versus the `excludePrefix` passed to `fromMatching`. It appears that the former is the key while the latter is the value in the environment variable map. Could we make that clear in the naming?

NicholasLYang (approved):
The code pretty much makes sense to me, just asking some clarifying questions

cli/internal/run/global_hash.go
mehulkar: I was using a pointer so I could return nil instead of an empty struct in error scenarios, but it's not really required

cli/internal/env/env.go
mehulkar: yeah, I didn't want to spend _too_ much time getting this designed right. I've pushed up a commit that hopefully makes it a little more clear and makes `fromMatching` not as coupled to how it's used

PR #4093, opened 2023-03-07
Use streaming file reads when comparing for writes

jridgewell:
### Description

Following #3526, this reimplements `write(content)`'s file comparison to stream the contents of the file. Not every file we write is massive, but some of the ones we generate are 100kb+. I hope this can reduce memory pressure a bit by using a consistent 8kb block size, instead of allocating a buffer for the full file contents just to check if we should write.

### Testing Instructions

Wait for benchmarks.

### DIFF
--- a/crates/turbo-tasks-fs/src/lib.rs
+++ b/crates/turbo-tasks-fs/src/lib.rs
@@ -6,6 +6,7 @@
 #![feature(main_separator_str)]
 #![feature(box_syntax)]
 #![feature(round_char_boundary)]
+#![feature(new_uninit)]
 
 pub mod attach;
 pub mod embed;
@@ -21,10 +22,11 @@ pub mod util;
 
 use std::{
     borrow::Cow,
+    cmp::min,
     collections::{HashMap, HashSet},
     fmt::{self, Debug, Display, Formatter},
     fs::FileType,
-    io::{self, ErrorKind},
+    io::{self, BufRead, ErrorKind},
     mem::take,
     path::{Path, PathBuf, MAIN_SEPARATOR},
     sync::{
@@ -47,7 +49,10 @@ use read_glob::read_glob;
 pub use read_glob::{ReadGlobResult, ReadGlobResultVc};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use tokio::{fs, io::AsyncReadExt};
+use tokio::{
+    fs,
+    io::{AsyncBufReadExt, AsyncReadExt},
+};
 use turbo_tasks::{
     mark_stateful,
     primitives::{BoolVc, StringReadRef, StringVc},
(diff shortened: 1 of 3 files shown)


crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1196,6 +1198,69 @@ impl From<File> for FileContentVc {
...
+        };
+        // If the meta is different, we need to rewrite the file to update it.
+        if new_file.meta != old_meta.into() {
+            return Ok(FileComparison::NotEqual);
+        }
+
+        // So meta matches, and we have a file handle. Let's stream the contents to see
+        // if they match.
+        let mut new_contents = new_file.read();
+        let mut old_contents = AsyncBufReader::new(&mut old_file);
```
sokra, line 1238: Performance-wise it would probably be better to avoid the buffered reader here and just read directly from the old_file as much as possible. Probably requires a little bit more complicated compare logic to handle consumed positions.

crates/turbo-tasks-fs/src/util.rs
```diff
@@ -110,3 +126,73 @@ pub fn normalize_request(str: &str) -> String {
...
 }
+
+/// AsyncBufReader adds buffering to any [AsyncRead].
+///
+/// This essentially just implements [AsyncBufRead] over an [AsyncRead], using a
+/// large buffer to store data. As the data is consumed, an offset buffer will
+/// continue to be returned until the full buffer has been consumed. This allows
+/// us to skip the overhead of, eg, repeated sys calls to read from disk as we
+/// process a smaller number of bytes.
+pub struct AsyncBufReader<'a, T: AsyncRead + Unpin + Sized> {
```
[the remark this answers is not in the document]
jridgewell: 😳 I searched the docs for `AsyncBufReader` and not `BufReader`, and implemented this from scratch because I didn't find anything.

crates/turbo-tasks-fs/src/lib.rs
jridgewell: I think you're assuming reading from a file returns a buffer, and we're then copying from that buffer into our `BufReader` buffer. But that's not the case, the `read()` API requires you to maintain a buffer and that gets passed to the system to copy bytes into, reading doesn't return a system buffer.

So we can either maintain our own buffer in this loop, or we can reuse the buffer created as part of `BufReader`.

sokra: I see :+1:

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -6,6 +6,7 @@
 #![feature(main_separator_str)]
 #![feature(box_syntax)]
 #![feature(round_char_boundary)]
+#![feature(new_uninit)]
```
alexkirsz, line 9: No longer necessary.

alexkirsz (approved):

PR #4094, opened 2023-03-07
Fix error message documentation

### LINKED ISSUE #4076: Update docs: Understanding `turbo.json`
I just created a new project and quick search on a similar error brought me here.
```
run failed: error preparing engine: Could not find the following tasks in project: hello
Turbo error: error preparing engine: Could not find the following tasks in project: hello
```
according to the docs, [Understanding `turbo.json`](https://turbo.build/repo/docs/getting-started/create-new#3-understanding-turbojson) the expected error is:
```
task `hello` not found in turbo `pipeline` in "turbo.json".
Are you sure you added it?
```
`dev`, `lint` and `build commands are working as expected.

`v: 1.8.3`

_Originally posted by @Tittoh in https://github.com/vercel/turbo/issues/3918#issuecomment-1454673231_

mehulkar:
Fixes #4076

### DIFF
--- a/docs/pages/repo/docs/getting-started/create-new.mdx
+++ b/docs/pages/repo/docs/getting-started/create-new.mdx
@@ -40,7 +40,7 @@ You can also clone a Turborepo starter repository to get a head start on your mo
 
 ## Full tutorial
 
-This tutorial will walk you through using the Turborepo [`basic` example](https://github.com/vercel/turbo/tree/main/examples/basic). By the end, you'll feel confident with using `turbo`, and know all the basic functionality.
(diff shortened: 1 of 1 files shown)


arlyon:
How feasible is it to restore the old log line here? The reported one is quite ugly...

mehulkar:
> How feasible is it to restore the old log line here? 

@arlyon there are two issues: it's probably possible, but not sure it's worth spending more time on right now. the reason it's ugly is because the errors get wrapped as they bubble up, so I'd probably have to make a custom error struct and make an exception to the wrapping logic. Or revisit the wrapping errors entirely, which feels very out of scope here.

PR #4096, opened 2023-03-07
Inline common code generation

sokra:
### Description

Testing for benchmark purposes


anthonyshew:

PR #4100, opened 2023-03-07
only use recursive watchers on macOS and windows

sokra:
### Description

Only windows and macOS support real recursive file watchers, other OS emulate it by walking the directory structure and watching all directories. But that might be really slow and use up a lot of watchers, so we don't want that. Instead we know exactly which directories are used and can watch selectively directories when they are accessed.

* only use recursive watchers on macOS and windows
* fallback to non-recursively watch read directories on other OS

Note that this implementation still has some bugs when renaming folders, but we can probably figure out these edge cases later...

### DIFF
--- a/crates/turbo-tasks-fs/src/lib.rs
+++ b/crates/turbo-tasks-fs/src/lib.rs
@@ -29,7 +29,7 @@ use std::{
     path::{Path, PathBuf, MAIN_SEPARATOR},
     sync::{
         mpsc::{channel, RecvError, TryRecvError},
-        Arc, Mutex,
+        Arc, Mutex, MutexGuard,
     },
     time::Duration,
 };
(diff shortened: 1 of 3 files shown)


kwonoj (approved):

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -29,7 +29,7 @@ use std::{
     path::{Path, PathBuf, MAIN_SEPARATOR},
     sync::{
         mpsc::{channel, RecvError, TryRecvError},
-        Arc, Mutex,
+        Arc, Mutex, MutexGuard,
```
jridgewell, line 32: `MutexGuard` is unused on Mac/Windows, giving a clippy warning.

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -80,6 +80,64 @@ pub trait FileSystem: ValueToString {
...
+    /// used on a OS that doesn't support recursive watching.
+    #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+    watching: dashmap::DashSet<PathBuf>,
+}
+
+impl DiskWatcher {
+    #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+    fn restore_if_watching(&self, dir_path: &Path, root_path: &Path) -> Result<()> {
+        if self.watching.contains(dir_path) {
+            let mut watcher = self.watcher.lock().unwrap();
```
jridgewell, line 96: We have a lock that's only used inside a loop. If we passed the `HashSet` to this method, we could lock once and iterate

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -145,12 +224,18 @@ impl DiskFileSystem {
         }
 
         watcher_guard.replace(watcher);
+        drop(watcher_guard);
+
+        let disk_watcher = self.watcher.clone();
+        let root_path = self.root_path().to_path_buf();
```
jridgewell, line 230: Ditto unused. Could we make a consistent interface between Mac and Linux so we can just make calls to empty methods?

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -161,7 +246,16 @@ impl DiskFileSystem {
...
+                        Ok(DebouncedEvent::Create(path)) => {
+                            batched_invalidate_path_and_children.insert(path.clone());
+                            batched_invalidate_path_and_children_dir.insert(path.clone());
+                            if let Some(parent) = path.parent() {
+                                batched_invalidate_path_dir.insert(PathBuf::from(parent));
+                            }
+                            #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+                            batched_new_paths.insert(path.clone());
+                        }
+                        Ok(DebouncedEvent::Remove(path)) => {
```
jridgewell, line 258: Do we need to remove a from the `DiskWatcher` in this case?

PR #4101, opened 2023-03-07
fix: ensure that memory is free'd using the same allocator

arlyon:
We were encountering segfaults on certain combinations of go / rust versions. This was determined to be due to differing allocators. Memory reserved on one side of the sandwich and free'd on the other would sometimes work depending on toolchains / platform. This resolves that problem.

To complicate things, this issue would only arise during cross-compilation, and not when building on the target OS / arch.

### Testing Instructions

I am going to run a dry run canary that we can re-test, though I have manualy verified the cross-compilation for win amd64.

### DIFF
--- a/cli/internal/ffi/ffi.go
+++ b/cli/internal/ffi/ffi.go
@@ -25,7 +25,7 @@ func Unmarshal[M proto.Message](b C.Buffer, c M) error {
 		return err
 	}
 
-	b.Free()
+	C.free_buffer(b)
 
 	return nil
 }
--- a/cli/internal/ffi/bindings.h
+++ b/cli/internal/ffi/bindings.h
@@ -8,6 +8,8 @@ typedef struct Buffer {
   uint8_t *data;
 } Buffer;
 
+void free_buffer(struct Buffer buffer);
+
 struct Buffer get_turbo_data_dir(void);
 
 struct Buffer changed_files(struct Buffer buffer);
--- a/crates/turborepo-ffi/src/lib.rs
+++ b/crates/turborepo-ffi/src/lib.rs
@@ -11,6 +11,16 @@ pub struct Buffer {
     data: *mut u8,
 }
 
+#[no_mangle]
+pub extern "C" fn free_buffer(buffer: Buffer) {
+    // SAFETY
+    // we are responsible for freeing the memory allocated on the rust side
+    // we do this by converting the raw pointer to a Vec and letting it drop
+    // this is safe because we know that the memory was allocated by rust
+    // and that the length is correct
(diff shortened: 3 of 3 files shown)


NicholasLYang:
I think we should write up a little bit about ownership across this FFI. It's a bit confusing and it'd be good to make it explicit. If I understand correctly, it goes like this:

When Rust passes a Buffer to Go, Rust "forgets" about the Buffer, Go receives the Buffer, unmarshals it, and then frees it by having Rust drop the Buffer.

When Go passes a Buffer to Rust, Go creates the Buffer and implicitly "forgets" about it, Rust receives the Buffer and drops it when executing `into_proto`.

Is that correct? And if so, in the Go to Rust case, does that mean that Go is allocating but Rust is freeing? That seems incorrect.

arlyon:
...yes. I am also confused by why the go side works but rust doesn't. I just did another dry run (see here for artifacts https://github.com/arlyon/turbo/actions/runs/4355600953) and it works. That said, I will make the behaviour consistent and make sure to add some rigorous docs to the crate.

NicholasLYang:
I guess it works because the Rust code is not re-entered, so Rust doesn't ever touch the memory again, and likely Go never touches it either because it's forgotten about it. The main concern I have is whether this leads to Rust "stealing" memory from Go because Go never to GC that buffer

NicholasLYang (approved):
Looks good! The comments are very helpful too

cli/internal/ffi/ffi.go
```diff
@@ -25,7 +30,12 @@ func Unmarshal[M proto.Message](b C.Buffer, c M) error {
 		return err
 	}
 
-	b.Free()
+	// free the buffer on the rust side
+	//
+	// SAFETY: do not use `C.free_buffer` to free a buffer that has been allocated
+	// on the go side. If you happen to accidentally use the wrong one, you can
+	// expect a segfault on some platforms. This is the only valid callsite.
+	C.free_buffer(b)
```
gsoltis, line 38: I am not certain that this is safe. `toBytes` returns a Go-owned slice. Can you document what happens when Go goes to free it?

cli/internal/ffi/ffi.go
```diff
@@ -42,6 +52,16 @@ func Marshal[M proto.Message](c M) C.Buffer {
 	return toBuffer(bytes)
```
gsoltis, line 52: Similar to above, can you document what happens when Go frees the slice that allocated the byte buffer?

gsoltis (changes requested):
Discussed in person, but left the comments requesting documentation around Go deallocating slices.

gsoltis:
We might be able to remove `toBytes` in favor of [unsafe.Slice](https://pkg.go.dev/unsafe@master#Slice)

gsoltis (approved):

PR #4102, opened 2023-03-07
fix(turbopack): Update app-renderer

shuding:
### Description

This PR mirrors some related manifest changes in Next.js: https://github.com/vercel/next.js/pull/46881.


jridgewell (approved):

crates/next-core/js/src/entry/app-renderer.tsx
```diff
@@ -137,32 +137,25 @@ async function runOperation(renderData: RenderData) {
...
+      get(target, key) {
+        if (key === "__ssr_module_mapping__") {
           return manifest;
         }
-        if (name === "__entry_css_files__") {
+        if (key === "__entry_css_files__") {
           return __entry_css_files__;
         }
-        return new Proxy({}, proxyMethodsForModule(name as string, css));
+        const [file, name] = (key as string).split("#");
```
sokra, line 149: The file could potentially contain `#` and the `name` could also potentially contain `#`. This is pretty unsafe to join with `#`.

Can't we encode all information into the JSON instead of joining with `#`.

jridgewell: I'm not sure what's a Next.js detail and what's a Turbopack detail. Is the `name` really supposed to be JSON, or is that a hack that we did? If we instead did an rsplit here, would that fix our hack without changing Next's behavior?

shuding: The `JSON.parse` is a hack inside Turbopack. That `#` was from React's upstream change: https://github.com/facebook/react/pull/26300 and we had to encode it that way.

`name` can never contain `#` because it's the module's export name (`export const foo`) or `*`. But not sure if a filepath can contain the `#` symbol or not, I think it's unlikely.

sokra: The react change changes from `filename` and `name` as separate entities to a single `id`. next.js creates the `file#name` encoding here: https://github.com/vercel/next.js/blob/7e933a094c3f7a17ae085396633ad6525ccfb590/packages/next/src/build/webpack/loaders/next-flight-loader/module-proxy.ts#L168
We could also choose a different encoding, e. g. JSON.

folders can contain `#`, unlikely but some people do that `/home/blah/#repos` which already led to some trouble because we had some special behavior for fragments in webpack.

even export names will soon be able to support any string. There is a proposal for that.

But anyway...

jridgewell: I changed this to use an rsplit.

PR #4116, opened 2023-03-07
Add TaskExecutionSummary to Run Summary

mehulkar:
## Description
This PR adds:
- an `execution` key to each task summary. `omitEmpty` is used, so the key will not show up in `--dry=json` output, when there was no execution.
- a top level `execution` key to the Run Summary which includes counts for successful, failed, cached tasks. This is the same as the output that we also see in terminal UI in logs.

## Implementation
- I used the existing `RunState`  struct (`run/run_state.go`), moving it into the `runsummary` package. I used the `.Run()` function of this struct to return a `TaskExecutionSummary`, which gets updated over the course of the execution, and then attached that to the `TaskSummary` that we've been building up this whole time. 
- `RunState` (now known as `executionSummary`) and pretty much everything around it is now entirely private to the `runsummary` package, and everything happens through `RunSummary` (namely: `TrackTask()` and `Close()`)

## Review Notes
The diff is not huge, but may be hard to understand because of renames. I tried to avoid these renames until the very end, so stepping through each commit will help review this PR.


arlyon (approved):
What is the motivation behind 4ed06b4ffe8fe5043fc7f1c38e530cc5f47b490a? What values can we expect that aren't captured in the enum?

cli/internal/runsummary/execution_summary.go
```diff
@@ -0,0 +1,201 @@
...
+		Label:  label,
+		Status: targetBuilding,
+	})
+
+	tracer := chrometracing.Event(label)
+
+	// This function can be called with an enum and an optional error to update
+	// the state of a given taskID.
+	tracerFn := func(outcome executionEventName, err error) {
+		defer tracer.Done()
```
arlyon, line 123: Does this mean that the tracerFn is meant to only be called once? Is it perhaps possible to track and return an error if you call it twice? Otherwise add will register the same event twice (and increment Failure, Attempted, etc)

cli/internal/runsummary/execution_summary.go
```diff
@@ -0,0 +1,201 @@
...
+	Status string `json:"status"`
+
+	// Error, only populated for failure statuses
+	Err error `json:"error"`
+}
+
+// executionSummary is the state of the entire `turbo run`. Individual task state in `Tasks` field
+// TODO(mehulkar): Can this be combined with the RunSummary?
+type executionSummary struct {
+	mu        sync.Mutex                       `json:"-"`
```
arlyon, line 79: Might be nice to have a comment explaning which field(s) this mutex covers.

cli/internal/runsummary/execution_summary.go
```diff
@@ -0,0 +1,201 @@
...
+	outputPath := chrometracing.Path()
+	if outputPath == "" {
+		// tracing wasn't enabled
+		return nil
+	}
+
+	name := fmt.Sprintf("turbo-%s.trace", time.Now().Format(time.RFC3339))
+	if filename != "" {
+		name = filename
+	}
```
arlyon, line 183: Saves some work

```suggestion
	if filename == "" {
		filename = fmt.Sprintf("turbo-%s.trace", time.Now().Format(time.RFC3339))
	}
```

cli/internal/runsummary/format_execution_summary.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/vercel/turbo/cli/internal/util"
+)
+
+func (summary *RunSummary) printExecutionSummary(ui cli.Ui) {
+	maybeFullTurbo := ""
+	if summary.ExecutionSummary.Cached == summary.ExecutionSummary.Attempted && summary.ExecutionSummary.Attempted > 0 {
+		terminalProgram := os.Getenv("TERM_PROGRAM")
+		// On the macOS Terminal, the rainbow colors show up as a magenta background
+		// with a gray background on a single letter. Instead, we print in bold magenta
+		if terminalProgram == "Apple_Terminal" {
```
arlyon, line 19: well, that's annoying 😂

arlyon (approved):

mehulkar:
> What is the motivation behind [4ed06b4](https://github.com/vercel/turbo/commit/4ed06b4ffe8fe5043fc7f1c38e530cc5f47b490a)? What values can we expect that aren't captured in the enum?

@arlyon it just shows up as an integer, and we want the stringified version. I saw the same strategy for TaskOutputMode, so I followed the same strategy https://github.com/vercel/turbo/blob/d923d44a1444795538639f921a6af4b7cbb6e583/cli/internal/util/task_output_mode.go#L59-L60

cli/internal/runsummary/execution_summary.go
mehulkar: It's confusingly named, but `tracer` and `tracerFn` are different things. 
- `tracer` is for the _chrometracing_ profile that's generated
- `tracerFn` updates the internal status

this `tracer.Done()` is for chrometracing profile. I'm not entirely sure how it works, but I tested before and after this PR and the generated `--profile` is the same

cli/internal/runsummary/execution_summary.go
mehulkar: i generally prefer not to reassign incoming function args, but in this case I also don't really want to make changes here since this function is not the focus of this PR (it just got moved around)

PR #4120, opened 2023-03-08
chore(labels): add rule to label non-vercel PRs

mehulkar:
This will label external PRs, so we can find them more easily during OSS On Call periods. I think it would also be better to match on a matching GH _team_, rather than the company. 

Labeling by company may not meet all our needs, because 

- Company is configured on a user's profile, and isn't required for employees
- We probably want to triage non-_core_-team PRs, rather than non-employees

But in lieu of the tooling being able to do that, this may be a good start!

- [ ] Create the actual `from: external` label (or whatever we land on)


tknickman:
The bot will auto create the label the first time it needs it, so not required!

PR #4123, opened 2023-03-08
Directly locate the executable.

nathanhammond:
We presently do a _lot_ of indirection when attempting to execute a local copy of `turbo`. This PR switches to directly locating the executable in the places where it can exist.

This PR adds support for local deference for `yarn@>=2.1.0` PnP (with `preferUnplugged` support), which previously didn't work.


tknickman:
> One material change: this will drop global deference to local execution for Yarn/Berry PnP for versions < 2.1 as those executables do not exist on disk outside of the zip file.

Let's chat about this - we're relying on that for a few future projects. Also, local deference is _currently_ working fine for Berry, I'm not sure we should put ourself into a fragmented state.

crates/turborepo-lib/src/shim.rs
```diff
@@ -184,38 +184,87 @@ pub struct LocalTurboState {
...
-                local_turbo_path.display()
-            );
-            return None;
-        }
+        };
 
-        let local_turbo_package_path = repo_root
-            .join("node_modules")
-            .join("turbo")
+        let platform_package_name = format!("turbo-{}-{}", os, arch);
```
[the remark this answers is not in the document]
nathanhammond: We don't actually do architecture fallback anywhere; this would be a new feature. I'm not going to include it in this PR.

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +177,237 @@ struct PackageJson {
...
+    pnp_unplugged_folder: PathBuf,
 }
 
-impl LocalTurboState {
-    pub fn infer(repo_root: &Path) -> Option<Self> {
-        let local_turbo_path = repo_root.join("node_modules").join(".bin").join({
+impl Default for YarnRc {
+    fn default() -> Self {
+        Self {
+            pnp_unplugged_folder: PathBuf::from("").join(".yarn").join("unplugged"),
```
NicholasLYang, line 189: Is this different from `PathBuf::from(".yarn").join("unplugged")`

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +177,237 @@ struct PackageJson {
...
+        Self {
+            bin_path: env::current_exe().ok(),
+            version: get_version(),
+            repo_state: None,
+        }
+    }
+}
+
+impl TurboState {
+    pub fn platform_package_name() -> String {
```
NicholasLYang, line 212: This could maybe be a lazy_static

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +177,237 @@ struct PackageJson {
...
+    bin_path: Option<PathBuf>,
+    version: &'static str,
+    repo_state: Option<RepoState>,
+}
+
+impl Default for TurboState {
+    fn default() -> Self {
+        Self {
+            bin_path: env::current_exe().ok(),
+            version: get_version(),
```
NicholasLYang, line 205: should this be `Self::version()`?

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +177,237 @@ struct PackageJson {
...
+}
+
+impl LocalTurboState {
+    // Hoisted strategy:
+    // - `npm install`
+    // - `yarn`
+    // - `yarn install --flat`
+    // - berry (nodeLinker: "node-modules")
+    //
+    // This also supports people directly depending upon the platform version.
```
NicholasLYang, line 285: Not sure I understand what this means

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +177,237 @@ struct PackageJson {
...
+    // Unplugged strategy:
+    // - berry 2.1+
+    fn generate_unplugged_path(root_path: &Path) -> Option<PathBuf> {
+        let platform_package_name = TurboState::platform_package_name();
+        let unplugged_base_path = Self::get_unplugged_base_path(root_path);
+
+        unplugged_base_path
+            .read_dir()
+            .ok()
+            .and_then(|mut read_dir| {
```
NicholasLYang, line 331: This works great, but just FYI, you can also do:

```
let mut read_dir = unplugged_base_path.read_dir().ok()?;
```

Because `?` works with both `Option` and `Result`

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +177,237 @@ struct PackageJson {
...
+    // binary directly. We can choose to operate this aggressively because the
+    // _worst_ outcome is that we run global `turbo`.
+    //
+    // In spite of that, the only known unsupported local invocation is Yarn/Berry <
+    // 2.1 PnP
+    pub fn infer(root_path: &Path) -> Option<Self> {
+        let platform_package_name = TurboState::platform_package_name();
+        let binary_name = TurboState::binary_name();
+
+        let platform_package_json_path = Path::new("")
```
NicholasLYang, line 362: It might be better to convert from an array according to the [path docs](https://doc.rust-lang.org/std/path/index.html)

crates/turborepo-lib/src/shim.rs
nathanhammond: Nope, but let me fix it in the followup PR: https://github.com/nathanhammond/turbo/pull/8/files#diff-22c6f95f72b511609de1411df92b2a92dd25eb5e7847fce1cafb4243baaea4d5

crates/turborepo-lib/src/shim.rs
nathanhammond: That's entirely unused at this step but I've got it lined up for the third PR in this stack, but I bit off too much and walked it back. To avoid rebasing through the PR stack I want to leave it here for now.

crates/turborepo-lib/src/shim.rs
nathanhammond: If I wrote enough macros I could turn it into an _actual_ static.

I spent too much time investigating doing that today and punted instead. But! I think that https://crates.io/crates/const_format may be a better option.

crates/turborepo-lib/src/shim.rs
nathanhammond: It made the chaining less elegant when I had it that way. Since my target outcome is an `Option<PathBuf>` `?` pulls me all the way out, and `and_then` doesn't do the nice resolution for the chain.

I was using `?` before changing the API to `Option<PathBuf>`, though!

crates/turborepo-lib/src/shim.rs
nathanhammond: It's technically possible to depend directly on `turbo-darwin-arm64` without using `turbo` for indirection; though grabbing the binary is a bit more obnoxious. This sidesteps that entirely.

crates/turborepo-lib/src/shim.rs
nathanhammond: This is another case where we really just need a macro. Absolutely no reason to do any of this at runtime. The closest crate is this: https://github.com/zertosh/path_macro, which is a dtolnay special, but extracted.

But that one isn't actually good enough for me.

nathanhammond: (Updated.)

crates/turborepo-lib/src/shim.rs
nathanhammond: nevermind, fixed her.

crates/turborepo-lib/src/shim.rs
nathanhammond: Adopted `const_format`.

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+        Self {
+            bin_path: env::current_exe().ok(),
+            version: get_version(),
+            repo_state: None,
+        }
+    }
+}
+
+impl TurboState {
+    pub fn platform_package_name() -> &'static str {
```
gsoltis, line 213: Could this whole function be a constant?

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+impl Default for YarnRc {
+    fn default() -> Self {
+        Self {
+            pnp_unplugged_folder: [".yarn", "unplugged"].iter().collect(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TurboState {
```
gsoltis, line 196: Given the confusion around global turbo and local turbo, can you add a comment as to which turbo this is the state of?

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+            #[cfg(not(any(target_os = "macos", target_os = "windows", target_os = "linux")))]
+            {
+                "unknown"
+            }
+        };
+
+        formatcp!("turbo-{}-{}", OS, ARCH)
+    }
+
+    pub fn binary_name() -> &'static str {
```
gsoltis, line 251: I think this could also be a constant?

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+
+impl LocalTurboState {
+    // Hoisted strategy:
+    // - `npm install`
+    // - `yarn`
+    // - `yarn install --flat`
+    // - berry (nodeLinker: "node-modules")
+    //
+    // This also supports people directly depending upon the platform version.
+    fn generate_hoisted_path(root_path: &Path) -> Option<PathBuf> {
```
gsoltis, line 287: Would these functions benefit from our newly-acquired path library?

gsoltis (commented):
Can you add a note about why the integration tests are changing? Is the binary no longer at a predictable location?

Also, remainder of the comments are non-blocking, I'll defer to the rust experts

crates/turborepo-lib/src/shim.rs
nathanhammond: So the `const fn` stuff seems super-cool but so far as I can tell (since these are returning `&'static str`) it doesn't appear like it will make a lot of difference except making the API inconsistent for `version` which needs an iterator that `const fn` can't compile through.

crates/turborepo-lib/src/shim.rs
nathanhammond: Good call, this is "neither" global nor local ... it is the "currently-executing `turbo`".

crates/turborepo-lib/src/shim.rs
nathanhammond: Optimizations at compile time should inline this just the same so far as I can tell.

crates/turborepo-lib/src/shim.rs
nathanhammond: I _very_ carefully didn't type a single `\` or `/` because I didn't have good tools for this. Right now it's assuredly correct since they're all relative path appends to a `root_path` value calculated by Rust, but yeah; that's a further refactoring for after this lands.

I really want something like https://github.com/zertosh/path_macro except better that does all of these "magic paths" into `&'static str`, but that's hard to justify building.

Something like:
`let path = smart_path_macro!(root_path / "node_modules" / "turbo" / "node_modules")`

Would validate that all static segments are good and then spit out:
```
#[cfg(not(windows))]
let path = root_path.join("node_modules/turbo/node_modules")
#[cfg(windows)]
let path = root_path.join("node_modules\\turbo\\node_modules")
```

I'll catch this on the next refactor PR.

nathanhammond:
> Can you add a note about why the integration tests are changing? Is the binary no longer at a predictable location?

The binary _never was_ at a predictable location (this enumerates the locations where you can find it) but we had attempted to defer locating it to the package managers.

That had consequences though:
- Not compatible with any version of PnP.
- Up to three processes worth of indirection: the currently-running `turbo`, `.bin/turbo(.cmd|.ps1)`, `turbo` the JS launcher (_always_ required for Windows), and finally the desired `turbo` or `turbo.exe`. (Setting aside `go-turbo`'s existence.) (This was originally a hypothesis for a fix on Windows ctrl-c nonsense.)
- We can't detect whether outer and inner `turbo` are the same and skip executing the inner `turbo` since the inner `turbo` was _always_ redirected. `pnpm turbo build` would reinvoke the _same_ `turbo` executable, but not in a way we could detect.

By adopting this complexity we already:
- Support PnP.
- Directly invoke inner `turbo` from outer `turbo`.
- Have a faster startup time.

In a later PR that's queued up:
- Can avoid calling ourselves by comparing `TurboState` to `LocalTurboState`.

And in some possible far future:
- Factor out the same lookup pattern to replace the `<package-manager> run <script-name>` and directly invoke the script executables.

***

There was a step in the many number of commits where I repaired the existing tests, made sure they were good with the new implementation, and then ported them over to the new much-more-detailed test harness.

gsoltis (dismissed):
Took another deeper pass at reviewing. Feel free to ignore me if I'm wrong about rust things.

The main thrust of my comments is to add some documentation and strongly prefer using `turborepo-path` library, or document why we can't and what's missing.

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+            .read_dir()
+            .ok()
+            .and_then(|mut read_dir| {
+                // berry includes additional metadata in the filename.
+                // We actually have to find the platform package.
+                read_dir.find_map(|item| match item {
+                    Ok(entry) => {
+                        let file_name = entry.file_name();
+                        if file_name
+                            .to_string_lossy()
```
arlyon, line 339: We should probably decide on a strategy for handling these types of issues.

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+        ];
+
+        // Detecting the package manager is more expensive than just doing an exhaustive
+        // search.
+        for root in search_functions
+            .iter()
+            .filter_map(|search_function| search_function(root_path))
+        {
+            // Needs borrow because of the loop.
+            #[allow(clippy::needless_borrow)]
```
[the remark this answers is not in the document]
arlyon: I think this is clippy getting confused by generics.

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+            let bin_path = root.join(&platform_package_executable_path);
+            match fs_canonicalize(&bin_path) {
+                Ok(bin_path) => {
+                    let resolved_package_json_path = root.join(platform_package_json_path);
+                    let platform_package_json_string =
+                        fs::read_to_string(resolved_package_json_path).ok()?;
+                    let platform_package_json: PackageJson =
+                        serde_json::from_str(&platform_package_json_string).ok()?;
```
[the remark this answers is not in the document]
arlyon: You can read from a slice of bytes instead, which is probably better. Ideally you could just pass bytes straight into the parser but serde_json doesn't have a byte stream api.

arlyon (approved):
Looks great aside from a few minor changes. I am approving anyways :)

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+pub struct TurboState {
+    bin_path: Option<PathBuf>,
+    version: &'static str,
+    repo_state: Option<RepoState>,
+}
+
+impl Default for TurboState {
+    fn default() -> Self {
+        Self {
+            bin_path: env::current_exe().ok(),
```
NicholasLYang, line 205: Do we want to error if we can't get `current_exe`?

crates/turborepo-lib/src/shim.rs
NicholasLYang: Yeah I'd consider taking a look at the path library and seeing if it fits here. If it doesn't, then perhaps we can open linear tickets for features and add them.

NicholasLYang (commented):
I mostly think we should check if `turborepo-paths` can be used here. Up to @nathanhammond  and @gsoltis if we want to merge with system paths and add `turborepo-paths` later, or if we want to block on `turborepo-paths`.

crates/turborepo-lib/src/shim.rs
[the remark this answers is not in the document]
nathanhammond: Yeah, doesn't compile without, and Clippy loses the plot. This is a Clippy bug/limitation.

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+        for root in search_functions
+            .iter()
+            .filter_map(|search_function| search_function(root_path))
+        {
+            // Needs borrow because of the loop.
+            #[allow(clippy::needless_borrow)]
+            let bin_path = root.join(&platform_package_executable_path);
+            match fs_canonicalize(&bin_path) {
+                Ok(bin_path) => {
+                    let resolved_package_json_path = root.join(platform_package_json_path);
```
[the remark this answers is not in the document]
nathanhammond: The early canonicalize is a performance thing since I need both `exists` and to grab the path for disambiguation. This gives me the same outcome as `.exists` with the side effect of returning the canonical value. `fs::read` handles it for us in the other case.

crates/turborepo-lib/src/shim.rs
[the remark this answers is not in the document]
nathanhammond: > Note that counter to intuition, [`serde_json::from_reader`] is usually slower than reading a file completely into memory and then applying [from_str](https://docs.rs/serde_json/latest/serde_json/fn.from_str.html) or [from_slice](https://docs.rs/serde_json/latest/serde_json/fn.from_slice.html) on it. See [issue #160](https://github.com/serde-rs/json/issues/160). - https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html

`from_str` vs. `from_slice` doesn't make a material difference for us here but I can get that last little bit by moving to slice. Will address in #4182 so I don't have to feed this all the way through.

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
 
-        let local_turbo_package_path = repo_root
-            .join("node_modules")
-            .join("turbo")
-            .join("package.json");
+    // Linked strategy:
+    // - `pnpm install`
+    // - `npm install --install-strategy=linked`
+    fn generate_linked_path(root_path: &Path) -> Option<PathBuf> {
+        fs_canonicalize(root_path.join("node_modules").join("turbo").join("..")).ok()
```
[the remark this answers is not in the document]
nathanhammond: This is all over at: https://www.notion.so/vercel/Package-Manager-Installation-Strategies-a810c208e7cd4c7f88e37f1d667d4b6d?pvs=4#6e39fdd848a34134924ac53f1eb9adaf

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+    }
+
+    // The unplugged directory doesn't have a fixed path.
+    fn get_unplugged_base_path(root_path: &Path) -> PathBuf {
+        let yarn_rc_filename =
+            env::var_os("YARN_RC_FILENAME").unwrap_or_else(|| OsString::from(".yarnrc.yml"));
+        let yarn_rc_filepath = root_path.join(yarn_rc_filename);
+
+        let yarn_rc_yaml_string = fs::read_to_string(yarn_rc_filepath).unwrap_or_default();
```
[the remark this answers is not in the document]
nathanhammond: It's the default of the `String` type, similar to a Go zero value.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cb658fac4d4ec47b2cdc88da6441070d

crates/turborepo-lib/src/shim.rs
```diff
@@ -176,50 +178,234 @@ struct PackageJson {
...
+        let yarn_rc: YarnRc = serde_yaml::from_str(&yarn_rc_yaml_string).unwrap_or_default();
```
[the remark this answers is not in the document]
nathanhammond: Nope, we get the default `YarnRc` object if you fail to parse here. This is the one which comes with the default value that we want. (`./.yarn/unplugged`)

crates/turborepo-lib/src/shim.rs
nathanhammond: It doesn't actually matter for this case which is why I chose this serialization:
- The upstream producers of these file paths only generate Unicode.
- If you try to break it by putting invalid Unicode in the middle of the prefix you get the replacement character and we don't match.
- If you try to break it by putting invalid Unicode after the prefix you get the replacement character and we actually do the right thing and proceed.

crates/turborepo-lib/src/shim.rs
nathanhammond: That can fail for lots of reasons, most particularly permissions. It's not an error to not know who we are, so I made it an `Option` instead.

The only thing it means is that we can't prevent invoking inner ("local") `turbo` if it just so happens to be the same one. But! The thing that would _prevent_ us from discovering `current_exe` would almost guarantee that we're not going to discover it when we go through looking for a `turbo` to delegate to.

crates/turborepo-lib/src/shim.rs
nathanhammond: The existing `AbsolutePath` type can be either Windows or Unix. The only thing it validates is `Path.isAbsolute` at construction. I don't actually care about that at all in this context, all paths are constructed explicitly using `join()`.

In the current state it provides me zero value, as demonstrated by the number of issues that we had with the git stuff on Windows.

If you look at the implementation:
```
    pub fn join<P: AsRef<Path>>(&self, other: P) -> AbsolutePathBuf {
        let path = self.0.join(other);
        assert!(path.is_absolute());
        AbsolutePathBuf(path)
    }
```

The baseline rules of `join` guarantee that assertion in advance:
https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.push

I appreciate that we want to push this forward, but that's a task in and of itself, not something to attach as a rider this late. (The lib landed after this work had reached "mostly done" and is only in tree because of a rebase.)

PR #4129, opened 2023-03-08
improve handling of syntax errors in HMR

sokra:
### Description

It's a bit ugly and I dislike the way it's implemented, but this will help for now.

* When using HMR and introducing an parse error to a module, this keeps the previous references and exports type to keep the environment intact and avoid unloading a bunch of modules

depends on https://github.com/vercel/turbo/pull/3975 to work correctly

### DIFF
--- a/crates/turbopack-ecmascript/src/lib.rs
+++ b/crates/turbopack-ecmascript/src/lib.rs
@@ -42,7 +42,7 @@ use swc_core::{
 pub use transform::{
     EcmascriptInputTransform, EcmascriptInputTransformsVc, NextJsPageExportFilter,
 };
-use turbo_tasks::{primitives::StringVc, TryJoinIterExt, Value, ValueToString};
+use turbo_tasks::{primitives::StringVc, ReadRef, TryJoinIterExt, Value, ValueToString};
 use turbo_tasks_fs::FileSystemPathVc;
 use turbopack_core::{
     asset::{Asset, AssetContentVc, AssetOptionVc, AssetVc},
@@ -50,7 +50,7 @@ use turbopack_core::{
     compile_time_info::CompileTimeInfoVc,
     context::AssetContextVc,
     ident::AssetIdentVc,
-    reference::AssetReferencesVc,
+    reference::{AssetReferencesReadRef, AssetReferencesVc},
     resolve::{
         origin::{ResolveOrigin, ResolveOriginVc},
         parse::RequestVc,
@@ -60,8 +60,8 @@ use turbopack_core::{
 pub use self::references::AnalyzeEcmascriptModuleResultVc;
 use self::{
     chunk::{
-        EcmascriptChunkItemContent, EcmascriptChunkItemContentVc, EcmascriptChunkItemOptions,
-        EcmascriptExportsVc,
+        placeable::EcmascriptExportsReadRef, EcmascriptChunkItemContent,
+        EcmascriptChunkItemContentVc, EcmascriptChunkItemOptions, EcmascriptExportsVc,
     },
     parse::ParseResultVc,
 };
@@ -94,7 +94,6 @@ fn modifier() -> StringVc {
 }
 
 #[turbo_tasks::value]
-#[derive(Clone, Copy)]
 pub struct EcmascriptModuleAsset {
     pub source: AssetVc,
     pub context: AssetContextVc,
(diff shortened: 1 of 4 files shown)


crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -95,14 +95,21 @@ fn modifier() -> StringVc {
...
     pub ty: EcmascriptModuleAssetType,
     pub transforms: EcmascriptInputTransformsVc,
     pub compile_time_info: CompileTimeInfoVc,
     pub inner_assets: Option<InnerAssetsVc>,
+    #[turbo_tasks(debug_ignore)]
+    pub last_successful_analyse: turbo_tasks::State<
+        Option<(
+            AnalyzeEcmascriptModuleResultVc,
+            AssetReferencesReadRef,
+            EcmascriptExportsReadRef,
```
alexkirsz, line 110: Put these three into a struct called `MemoizedSuccessfulAnalysis` or something like that.

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -95,14 +95,21 @@ fn modifier() -> StringVc {
...
-#[derive(Clone, Copy)]
 pub struct EcmascriptModuleAsset {
     pub source: AssetVc,
     pub context: AssetContextVc,
     pub ty: EcmascriptModuleAssetType,
     pub transforms: EcmascriptInputTransformsVc,
     pub compile_time_info: CompileTimeInfoVc,
     pub inner_assets: Option<InnerAssetsVc>,
+    #[turbo_tasks(debug_ignore)]
+    pub last_successful_analyse: turbo_tasks::State<
```
alexkirsz, line 106:
```suggestion
    pub last_successful_analysis: turbo_tasks::State<
```

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -158,15 +167,40 @@ impl EcmascriptModuleAssetVc {
...
+                    result,
+                    result_value.references.await?,
+                    result_value.exports.await?,
+                )));
+            } else {
+                if let Some((last_operation, last_references, last_exports)) =
+                    &*this.last_successful_analyse.get()
+                {
+                    let last_operation: turbo_tasks::RawVc = last_operation.into();
+                    last_operation.connect();
```
alexkirsz, line 192: Create and link a task to either: 1. get rid of this, or 2. create a better turbo tasks primitive to do this.

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -158,15 +167,40 @@ impl EcmascriptModuleAssetVc {
     }
 
     #[turbo_tasks::function]
-    pub async fn analyze(self) -> Result<AnalyzeEcmascriptModuleResultVc> {
+    pub async fn analyze(self, failsafe: bool) -> Result<AnalyzeEcmascriptModuleResultVc> {
```
alexkirsz, line 170:
```suggestion
    pub async fn analyze(self, mode: Value<AnalysisMode>) -> Result<AnalyzeEcmascriptModuleResultVc> {
```

```rust
#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, PartialOrd, Ord)]
pub enum AnalysisMode {
    #[default]
    Default,
    /// Keep track of the previous analysis and return that value when a new analysis error occurs.
    /// This is necessary so we don't thrash a whole subgraph of computations on error, only to 
    /// have to recompute it once the operation starts succeeding again.
    MemoizedOnError,
}
```

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -109,11 +109,12 @@ use crate::{
     EcmascriptInputTransformsVc,
 };
 
-#[turbo_tasks::value]
+#[turbo_tasks::value(shared)]
 pub struct AnalyzeEcmascriptModuleResult {
     pub references: AssetReferencesVc,
     pub code_generation: CodeGenerateablesVc,
     pub exports: EcmascriptExportsVc,
+    pub successful: bool,
```
alexkirsz, line 117:
```suggestion
    /// `true` when the analysis was successful.
    pub successful: bool,
```

alexkirsz (changes requested):

crates/turbopack-ecmascript/src/lib.rs
sokra: https://linear.app/vercel/issue/WEB-699/rawvcconnect-should-be-exposed-in-a-different-way

crates/turbopack-ecmascript/src/lib.rs
sokra: I split the logic into two different methods, where one is calling the other.

PR #4134, opened 2023-03-09
fix app dir with latest next.js

sokra:
### Description

* [fix manifest for __ssr_module_mapping__](https://github.com/vercel/turbo/commit/4abd87312d05a74ee858a8f556bfe567f3a9e008)
* [add missing styled-jsx alias](https://github.com/vercel/turbo/commit/d9b4a4df7f81b18e657689f4e182bc9ef2259633)

### Testing Instructions

layout-playground

### DIFF
--- a/crates/next-core/js/src/entry/app-renderer.tsx
+++ b/crates/next-core/js/src/entry/app-renderer.tsx
@@ -137,26 +137,51 @@ async function runOperation(renderData: RenderData) {
     tree = [info.segment, { children: tree }, components];
   }
 
+  const proxyMethodsForModule = (
+    id: string
+  ): ProxyHandler<FlightManifest["__ssr_module_mapping__"][""]> => ({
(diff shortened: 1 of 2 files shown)


crates/next-core/js/src/entry/app-renderer.tsx
```diff
@@ -137,26 +137,51 @@ async function runOperation(renderData: RenderData) {
...
+        if (name === "__ssr_module_mapping__") {
+          return manifest;
+        }
+        if (name === "__entry_css_files__") {
+          return __entry_css_files__;
+        }
+        return new Proxy({}, proxyMethodsForModule(name as string));
+      },
+    };
+  };
```
alexkirsz, line 165: It would be helpful to document these workarounds. From the outside they are very obscure and the data flow is hard to understand, and easy to break.

PR #4135, opened 2023-03-09
improve introspection

sokra:
### Description

Add more details to introspection and make it more robust to errors during introspection

### Testing Instructions

go to http://localhost:3000/__turbopack__/

### DIFF
--- a/crates/turbopack-core/src/introspect/asset.rs
+++ b/crates/turbopack-core/src/introspect/asset.rs
@@ -6,7 +6,10 @@ use turbo_tasks_fs::FileContent;
 use super::{Introspectable, IntrospectableChildrenVc, IntrospectableVc};
 use crate::{
     asset::{Asset, AssetContent, AssetContentVc, AssetVc},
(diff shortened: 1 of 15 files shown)


crates/turbopack-core/src/introspect/asset.rs
```diff
@@ -33,6 +37,31 @@ fn reference_ty() -> StringVc {
...
+
+#[turbo_tasks::function]
+fn separate_reference_ty() -> StringVc {
+    StringVc::cell("separate reference".to_string())
+}
+
+#[turbo_tasks::function]
+fn async_reference_ty() -> StringVc {
+    StringVc::cell("async reference".to_string())
+}
```
alexkirsz, line 63: Would be cool if we had static-like Vcs.

```rust
turbo_tasks::static! {
    static PARALLEL_REFERENCE_TY: StringVc = StringVc::cell("parallel reference".to_string());
}
```

This is essentially the same as a tt::function so probably not worth it.

alexkirsz (approved):

PR #4141, opened 2023-03-09
Refactor ES chunk evaluate logic into a Runtime trait

alexkirsz:
### Description

This refactors the EcmaScript chunk evaluate segment logic into a separate `EcmascriptChunkRuntime` trait.

This is important because we are going to start having more than one runtime evaluation logic with `next build` (and `ncc` and any future bundler-specific work).

### Testing Instructions

Snapshots + Integrations tests


crates/turbopack-ecmascript/src/chunk/content.rs
```diff
@@ -213,110 +224,35 @@ impl EcmascriptChunkContentVc {
...
-                r#"
-                    , ({{ loadedChunks, instantiateRuntimeModule, registerChunkList }}) => {{
-                        if (!(true{condition})) return true;
-                        {chunk_list_register}
-                    {entries_instantiations}
-                    }}
-                "#
-            )?;
+
+        write!(code, "\n}}")?;
```
sokra, line 244: This should be controlled by the runtime.

Other runtimes might not use a global var to communicate between chunks.

e.g. they could use ESM `export` and `import()` to load chunks, or in node.js they could use `module.modules = ...` and `require()` to load chunks.

crates/turbopack-ecmascript/src/chunk/runtime/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// Returns the code for this runtime instance's parameters.
+    ///
+    /// This is separate from the runtime code because multiple runtimes can be
+    /// loaded from different chunks. The runtime code must be the same, but the
+    /// parameters can be different. This is why this method accepts the origin
+    /// chunk as argument, and [`EcmascriptChunkRuntime::code`] does not.
+    fn params(&self, chunk: EcmascriptChunkVc) -> CodeVc;
+
+    /// Returns the code for this runtime.
+    fn code(&self) -> CodeVc;
```
sokra, line 34: I would make it a single method that returns the whole chunk code when you pass a list of module factories.

crates/turbopack-ecmascript/src/chunk/content.rs
alexkirsz: Agreed. This is why I'm also trying to rework our `EcmascriptChunkEntry` system to make it so it can be passed to the runtime.

crates/turbopack-dev/js/src/runtime.dom.js
```diff
@@ -2,8 +2,16 @@
 
 /** @type {RuntimeBackend} */
 const BACKEND = {
-  loadChunk(chunkPath) {
+  loadChunk(chunkPath, source) {
     return new Promise((resolve, reject) => {
+      // We don't need to load runtime chunks, as they're already present in the DOM.
+      // However, we need to wait for them to register themselves within `registerChunk`
+      // before we can start instantiating runtime modules, hence the absense of
+      // `resolve()` in this branch.
```
sokra, line 10: I think the comment is a bit wrong.

```
We don't need to load chunks references from runtime code, as they're ...

File: crates/turbopack-ecmascript/src/chunk/context.rs [hunk 4]
```diff
@@ -2,23 +2,22 @@ use anyhow::Result;
...
+use super::{item::EcmascriptChunkItemVc, EcmascriptChunkRuntimeVc};
 
-#[turbo_tasks::value]
-pub(super) struct EcmascriptChunkContext {
-    context: ChunkingContextVc,
-}
+/// [`EcmascriptChunkContext`] must be implemented by [`ChunkingContext`]
+/// implementors that want to operate on [`EcmascriptChunk`]s.
+#[turbo_tasks::value_trait]
+pub trait EcmascriptChunkContext: ChunkingContext {
```
sokra, line 10:
```suggestion
pub trait EcmascriptChunkingContext: ChunkingContext {
```

crates/turbopack-dev/src/ecmascript/runtime.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+#[turbo_tasks::value(serialization = "auto_for_input")]
+#[derive(Debug, Clone, Copy, PartialOrd, Ord, Hash)]
+pub(crate) enum EcmascriptDevChunkRuntimeMode {
+    /// The main runtime code will be included in the chunk and the main entries
+    /// will be evaluated as soon as the chunk executes.
+    RegisterAndEvaluate,
+    /// The chunk's entries will be registered within an existing runtime.
+    Register,
```
sokra, line 217: This is a bit weird...

Assuming chunk A (main_entries: X) with RegisterAndEvaluate and chunk B (main_entries: Y) with Register will be merged. This would lead to X and Y being executed, while the unmerged versions would only execute X.

I think the entries for execution need to be stored in a separate list.

sokra: Or give chunks a list of entries to execute and determine the runtime mode from the list.

crates/turbopack-ecmascript/src/chunk/item.rs
```diff
@@ -43,26 +38,35 @@ pub struct EcmascriptChunkItemOptions {
 #[turbo_tasks::value_trait]
 pub trait EcmascriptChunkItem: ChunkItem {
     fn content(&self) -> EcmascriptChunkItemContentVc;
-    fn chunking_context(&self) -> ChunkingContextVc;
+    fn chunking_context(&self) -> EcmascriptChunkContextVc;
     fn id(&self) -> ModuleIdVc {
-        EcmascriptChunkContextVc::of(self.chunking_context()).chunk_item_id(*self)
+        self.chunking_context().chunk_item_id(*self)
     }
```
sokra, line 44: We should move that into `impl EcmascriptChunkItemVc {}`. It shouldn't be allowed to change this method from the EcmascriptChunkItem impl

sokra (approved):

PR #4142, opened 2023-03-09
feat(ecmascript): transform ts input with runtime flags

kwonoj:
### Description

Another attempt to close WEB-659.

The crux is same as previous PR, but attempt to change the location where it read config / bubble down the config values to the actual transform stage. Mainly, `enable_typescript_transform` is now accepting an option instead of boolean flag to down to `EcmaInputTransform::Typescript`, and `get_*_module_context()` reads the config value as needed.

### DIFF
--- a/crates/next-core/src/next_server/context.rs
+++ b/crates/next-core/src/next_server/context.rs
@@ -299,7 +299,7 @@ pub async fn get_server_module_options_context(
             ModuleOptionsContext {
                 enable_postcss_transform,
                 enable_webpack_loaders,
-                enable_typescript_transform: Some(Default::default()),
+                enable_typescript_transform: Some(tsconfig),
                 rules: vec![(
                     foreign_code_context_condition,
                     module_options_context.clone().cell(),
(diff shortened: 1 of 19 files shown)


crates/next-core/src/next_server/context.rs
```diff
@@ -296,7 +299,7 @@ pub async fn get_server_module_options_context(
             ModuleOptionsContext {
                 enable_postcss_transform,
                 enable_webpack_loaders,
-                enable_typescript_transform: true,
+                enable_typescript_transform: Some(Default::default()),
```
sokra, line 302: app route doesn't derive from tsconfig?

crates/turbopack/src/module_options/mod.rs
```diff
@@ -107,11 +107,24 @@ impl ModuleOptionsVc {
...
         let vendor_transforms =
             EcmascriptInputTransformsVc::cell(custom_ecmascript_transforms.clone());
-        let ts_app_transforms = if enable_typescript_transform {
-            let mut base_transforms = vec![EcmascriptInputTransform::TypeScript];
+        let ts_app_transforms = if enable_typescript_transform.is_some() {
+            let mut base_transforms = if let Some(transform) = ts_transform {
+                vec![transform.clone()]
+            } else {
+                vec![]
+            };
```
sokra, line 127:
```suggestion
        let ts_app_transforms = if let Some(transform) = ts_transform {
            let mut base_transforms = vec![transform.clone()];
```

sokra (approved):
Much better

PR #4147, opened 2023-03-10
feat(turbo-workspaces): support single package repos (#4147

tknickman:
### Description

1. Adds support for single package repos (needed for create-turbo)
1. Cleans up tests
    - Moved to a single test file
    - Updated test case generation


packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -71,39 +73,57 @@ async function read(args: ReadArgs): Promise<Project> {
...
 
-  // root dependencies
-  updateDependencies({
-    workspace: { name: "root", paths: project.paths },
-    project,
-    to,
-    logger,
-    options,
-  });
+    // write package.json here instead of deferring to avoid negating the changes made by updateDependencies
```
nathanhammond, line 103: Is it possible to gracefully feed forward `packageJson`?

packages/turbo-workspaces/src/managers/pnpm.ts
```diff
@@ -125,16 +136,28 @@ async function create(args: CreateArgs): Promise<void> {
  */
 async function remove(args: RemoveArgs): Promise<void> {
   const { project, logger, options } = args;
+  const hasWorkspaces = project.workspaceData.globs.length > 0;
 
-  logger.mainStep(`Removing pnpm workspaces`);
-  if (project.paths.workspaceConfig) {
+  logger.mainStep(
+    `Removing ${project.packageManager}${hasWorkspaces ? "workspaces" : ""}`
```
nathanhammond, line 142: You've missed this space multiple times. We should reword these messages.

```suggestion
    `Removing ${project.packageManager}${hasWorkspaces ? " workspaces" : ""}`
```

packages/turbo-workspaces/src/managers/yarn.ts
```diff
@@ -70,20 +72,13 @@ async function read(args: ReadArgs): Promise<Project> {
  */
 async function create(args: CreateArgs): Promise<void> {
   const { project, to, logger, options } = args;
+  const hasWorkspaces = project.workspaceData.globs.length > 0;
 
-  logger.mainStep(`Creating yarn workspaces`);
+  logger.mainStep(
+    `Creating ${project.packageManager}${hasWorkspaces ? "workspaces" : ""}`
```
nathanhammond, line 78:
```suggestion
    `Creating ${project.packageManager}${hasWorkspaces ? " workspaces" : ""}`
```

packages/turbo-workspaces/src/managers/yarn.ts
```diff
@@ -94,24 +89,39 @@ async function create(args: CreateArgs): Promise<void> {
   );
   packageJson.packageManager = `${to.name}@${to.version}`;
 
-  if (!options?.dry) {
-    fs.writeJSONSync(project.paths.packageJson, packageJson, { spaces: 2 });
-  }
+  if (hasWorkspaces) {
```
nathanhammond, line 92: This series of things is bothering me having seen it three times, but I don't have an alternative proposal.

`pnpm` being weird makes it probably not worth abstracting out.

packages/turbo-workspaces/src/managers/yarn.ts
```diff
@@ -122,37 +132,44 @@ async function create(args: CreateArgs): Promise<void> {
...
-  const { project, to, logger, options } = args;
+  const { project, logger, options } = args;
+  const hasWorkspaces = project.workspaceData.globs.length > 0;
 
-  logger.mainStep(`Removing yarn workspaces`);
-  if (to.name !== "npm") {
-    const packageJson = getPackageJson({ workspaceRoot: project.paths.root });
-    delete packageJson.workspaces;
+  logger.mainStep(
+    `Removing ${project.packageManager}${hasWorkspaces ? "workspaces" : ""}`
```
nathanhammond, line 139:
```suggestion
    `Removing ${project.packageManager}${hasWorkspaces ? " workspaces" : ""}`
```

packages/turbo-workspaces/__tests__/managers.test.ts
```diff
@@ -0,0 +1,285 @@
...
+
+describe("managers", () => {
+  const { useFixture } = setupTestFixtures({
+    directory: path.join(__dirname, "../"),
+  });
+
+  describe("detect", () => {
+    test.each(generateDetectMatrix())(
+      "$project $type project detected by $manager manager - (expect: $result)",
+      async ({ project, manager, type, result }) => {
```
nathanhammond, line 26: 🎉

nathanhammond (approved):
🛠🚢: The log message spacing is the only thing I think I'm hard blocking on.

packages/turbo-workspaces/src/install.ts
```diff
@@ -65,39 +70,58 @@ export const PACKAGE_MANAGERS: Record<
...
+  return pm;
+}
+
+export default async function install(args: InstallArgs) {
   const { to, logger, options } = args;
-  let packageManager = PACKAGE_MANAGERS[to.name].find((manager) =>
-    satisfies(to.version, manager.semver)
-  );
+
+  const installLogger = logger ?? new Logger(options);
```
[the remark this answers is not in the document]
tknickman: Yep 😄 - the discovery was that we need a stand-alone installer method that can be independent from the `convert` flow specifically for use in `create-turbo`. This is needed when the example we're starting from is _already_ using the requested package manager and doesn't need a `conversion`, but still might need an install. So there is some work in here to allow this install method to be used independent of convert. Then, in create-turbo our logic is just (in python psuedo code for some reason):

```
if needsConversion:
    convert(skipInstall=true)

install()
```

packages/turbo-workspaces/src/index.ts
```diff
@@ -1,10 +1,12 @@
 import { getAvailablePackageManagers } from "turbo-utils";
 import getWorkspaceDetails from "./getWorkspaceDetails";
-import { PackageManager, Options } from "./types";
-import { default as convert } from "./convert";
+import type { PackageManager, Options, InstallArgs, Workspace } from "./types";
+import { default as convertProject } from "./convert";
```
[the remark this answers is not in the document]
tknickman: Good suggestion, will update

packages/turbo-workspaces/src/managers/pnpm.ts
tknickman: Yea, was a copy paste - will centralize this

packages/turbo-workspaces/src/managers/npm.ts
tknickman: We could, this gets a little annoying because we're dealing with _multiple_ `package.json`s but here we only want to feed through root (which may not even need to be updated by `updateDependencies` if there are no root internal deps). But I'll see if I can come up with a better API here to avoid the possible double read / write - it annoys me as well. f

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -71,39 +73,57 @@ async function read(args: ReadArgs): Promise<Project> {
  */
 async function create(args: CreateArgs): Promise<void> {
   const { project, options, to, logger } = args;
+  const hasWorkspaces = project.workspaceData.globs.length > 0;
 
-  logger.mainStep(`Creating npm workspaces`);
+  logger.mainStep(
+    `Creating ${project.packageManager}${hasWorkspaces ? "workspaces" : ""}`
```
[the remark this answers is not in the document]
tknickman: Updated to `Adding npm to {monorepo name}`

packages/turbo-workspaces/jest.config.js
```diff
@@ -10,8 +10,8 @@ module.exports = {
   verbose: true,
   coverageThreshold: {
     global: {
-      branches: 86,
-      functions: 91,
+      branches: 83,
+      functions: 87,
```
nathanhammond: 😲

tknickman: I added some more specific error handling for consumption by create-turbo, I'll go back and patch it up don't worry.

PR #4151, opened 2023-03-10
fix HMR when module evaluation throws

sokra:

### DIFF
--- a/crates/turbopack-ecmascript/js/src/runtime.js
+++ b/crates/turbopack-ecmascript/js/src/runtime.js
@@ -311,9 +311,10 @@ const SourceType = {
  * @param {ModuleId} id
  * @param {SourceType} sourceType
  * @param {ModuleId} [sourceId]
+ * @param {ModuleId[]} [parents]
  * @returns {Module}
  */
-function instantiateModule(id, sourceType, sourceId) {
+function instantiateModule(id, sourceType, sourceId, parents) {
   const moduleFactory = moduleFactories[id];
   if (typeof moduleFactory !== "function") {
     // This can happen if modules incorrectly handle HMR disposes/updates,
@@ -344,7 +345,7 @@ function instantiateModule(id, sourceType, sourceId) {
     exports: {},
     loaded: false,
     id,
-    parents: [],
+    parents: parents || [],
     children: [],
     interopNamespace: undefined,
     hot,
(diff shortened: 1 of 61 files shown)


crates/turbopack-ecmascript/js/src/runtime.js
```diff
@@ -311,9 +311,10 @@ const SourceType = {
  * @param {ModuleId} id
  * @param {SourceType} sourceType
  * @param {ModuleId} [sourceId]
+ * @param {ModuleId[]} [parents]
  * @returns {Module}
  */
-function instantiateModule(id, sourceType, sourceId) {
+function instantiateModule(id, sourceType, sourceId, parents) {
```
alexkirsz, line 317: We should change `SourceType` to be:

```
type SourceType = {
  type: "runtime",
} | {
  type: "parent",
  parentId: ModuleId,
} | {
  type: "update",
  parents: ModuleId[]
}

File: crates/turbopack-ecmascript/js/src/runtime.js [hunk 2]
```diff
@@ -628,8 +630,17 @@ function disposePhase(outdatedModules, disposedModules) {
     disposeModule(moduleId, "clear");
   }
 
+  const outdatedModuleParents = new Map();
+  for (const moduleId of outdatedModules) {
+    const oldModule = moduleCache[moduleId];
+    outdatedModuleParents.set(moduleId, oldModule?.parents);
+    delete moduleCache[moduleId];
```
alexkirsz, line 637: Can you add a comment explaining that we need the `disposeModule` calls above to run so that parents are updated accordingly, before we can take them out of the cache?

I'm wary of having this logic both here and in `disposeModule`. Is there a way we could keep this in a single function somewhere? This seems very fragile.

crates/turbopack-ecmascript/js/src/runtime.js
```diff
@@ -1221,8 +1251,12 @@ function disposeChunk(chunkPath) {
     if (noRemainingChunks) {
       moduleChunksMap.delete(moduleId);
       disposeModule(moduleId, "clear");
+      disposedModules.push(moduleId);
     }
   }
+  for (const moduleId of disposedModules) {
+    delete moduleCache[moduleId];
+  }
```
alexkirsz, line 1259:
```suggestion
```

No longer necessary as it's done in `disposeModule` (can remove `disposedModules`)

alexkirsz (changes requested):

alexkirsz (approved):

PR #4159, opened 2023-03-10
feat(docs): add "!.next/cache/**" everywhere

tknickman:
### Description

Turbo doesn't need to cache this (it's cached by the Vercel build cache) and can make cache unnecessarily large.


mehulkar (approved):
🚢

smorimoto:
This is really important! Our team actually fell for this trap! 😱

PR #4161, opened 2023-03-10
Code changes to support cross compiling musl on linux

gsoltis:
### Description

 - Incorporate musl-built libraries into our build container
 - statically link linux go binaries

### Testing Instructions

Verified using a local build of the build container. Will additionally verify via GH Actions


arlyon:
This looks good. Only issue is that removing -lz in the cgo headers means that non-musl linux builds (ie our tests) fail. I _think_ we can use CGO_CFLAGS for this, and override it.

PR #4162, opened 2023-03-10
Remove env-var linting comments.

anthonyshew:
We had a couple `eslint-disable-next-line turbo/no-undeclared-env-vars` sneak in somehow! Removing those and adding PORT to the `globalEnv`.


tknickman (approved):
👍🏼

PR #4166, opened 2023-03-11
Add `dynamic_embed_contents` feature

jridgewell:
Based on https://vercel.slack.com/archives/C04KC8A53T7/p1678496137453779?thread_ts=1678468241.013549&cid=C04KC8A53T7, our Next.js test suite is failing because the default test compilation is not portable. This is because the `embed_dir` and `embed_file` statically link the directory path into the binary, but do not statically embed the contents. If this same binary is executed on another machine that doesn't follow the exact same directory structure, it'll fail to embed the internal assets.

### DIFF
--- a/crates/turbo-tasks-fs/Cargo.toml
+++ b/crates/turbo-tasks-fs/Cargo.toml
@@ -13,6 +13,9 @@ name = "mod"
 harness = false
 
 [features]
+# Enables dynamic linking (and hot reloading) of embedded files/dirs.
+# A binary built with this option **is not portable**, the directory
+# path will be embedded into the binary.
 dynamic_embed_contents = []
 
 [dependencies]
--- a/crates/turbo-tasks-fs/Cargo.toml
+++ b/crates/turbo-tasks-fs/Cargo.toml
@@ -13,7 +13,7 @@ name = "mod"
 harness = false
 
 [features]
-static_embed = []
+dynamic_embed_contents = []
 
 [dependencies]
 anyhow = { workspace = true }
--- a/crates/turbo-tasks-fs/Cargo.toml
+++ b/crates/turbo-tasks-fs/Cargo.toml
@@ -12,6 +12,9 @@ bench = false
 name = "mod"
 harness = false
 
+[features]
+static_embed = []
+
 [dependencies]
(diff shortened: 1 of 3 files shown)


crates/turbo-tasks-fs/Cargo.toml
```diff
@@ -12,6 +12,9 @@ bench = false
 name = "mod"
 harness = false
 
+[features]
+static_embed = []
```
kwonoj, line 16: maybe worth to leave a 1-line comment for the reference?

kwonoj (approved):

alexkirsz (approved):

sokra:
Need to keep this branch alive for a while...

crates/turbo-tasks-fs/src/embed/dir.rs
```diff
@@ -36,12 +36,12 @@ macro_rules! embed_directory {
 
         // make sure the types the `include_dir!` proc macro refers to are in scope
         use turbo_tasks_fs::embed::include_dir;
-        #[cfg(debug_assertions)]
+        #[cfg(all(debug_assertions, feature = "dynamic_embed_contents"))]
```
ForsakenHarmony, line 39: the issue with this is that every crate that embeds anything now needs the feature, because the macro gets executed in the scope of the crate calling it

crates/turbo-tasks-fs/Cargo.toml
```diff
@@ -12,6 +12,12 @@ bench = false
 name = "mod"
 harness = false
 
+[features]
+# Enables dynamic linking (and hot reloading) of embedded files/dirs.
+# A binary built with this option **is not portable**, the directory
+# path will be embedded into the binary.
+dynamic_embed_contents = []
```
ForsakenHarmony, line 19: this does nothing, as the macro is never used inside this crate, at best it's documentation, at worst misleading

ForsakenHarmony:
it's a bit confusing to me that the next integration tests are using debug builds

jridgewell:
We have `debug_assert!` that I think should still run during integration testing. I'd love to flip this so that we can force file embedding for tests, but I couldn't get a flag to actually apply from the next.js repo.

PR #4174, opened 2023-03-13
Make tests run faster?

mehulkar:
- Use `turbo` to run unit, integration, e2e tests so we can _try_ to get Full Turbo in some places
- at the very least, stop building `turbo` 7 extra times times

    we build turbo just to _kick off_ tests, and then each of the test suites also builds turbo for each matrix run. (2 for unit, 3 for unit and e2e each, and 1 for integration)


.github/workflows/test.yml
```diff
@@ -234,70 +232,10 @@ jobs:
           target: ${{ matrix.os.name }}
           github-token: "${{ secrets.GITHUB_TOKEN }}"
 
+      # TODO: ensure integration tests runs with both GO_TAG: go and GO_TAG: rust
```
NicholasLYang, line 235: This won't matter soon (hopefully)

cli/turbo.json
```diff
@@ -3,19 +3,72 @@
...
+        "package-lock.json"
+      ],
+      "env": ["GITHUB_CI_OS_NAME"]
     },
-    "integration-tests": {
-      "inputs": ["integration_tests/**"],
+
+
+    "test:integration": {
+      // test:integration depends on test:unit. This is just way to specify order
```
NicholasLYang, line 45: Any reason we're specifying order instead of letting them run concurrently?

NicholasLYang (approved):
This makes sense to me. At some point I think it might be nice to have all of our tests running via Turbo in parallel, but moving unit, integration and e2e is a good start.

cli/turbo.json
mehulkar: The logs are a bit harder to parse (before `--log-order=group` lands) and I wasn't 100% sure that it was safe to run in parallel, so i wanted to start in easy mode first.

cli/Makefile
```diff
@@ -148,7 +148,7 @@ corepack:
 	which corepack || npm install -g corepack@latest
 	corepack enable
 
-e2e: corepack install turbo
+e2e: corepack install
```
chris-olszewski, line 151: This is the same as `e2e-prebuilt` now

cli/Makefile
```diff
@@ -313,10 +313,10 @@ $(CRAM_ENV)/bin/prysk: $(CRAM_ENV)/bin/pip
 
 INTEGRATION_TEST_FILES = $(shell find integration_tests -name "*.t")
 
-integration-tests: $(CRAM_ENV)/bin/prysk turbo $(INTEGRATION_TEST_FILES) corepack turbo
+integration-tests: $(CRAM_ENV)/bin/prysk turbo $(INTEGRATION_TEST_FILES) corepack
```
chris-olszewski, line 316: Just calling out that this target still depends on `turbo`

.github/workflows/test.yml
```diff
@@ -210,10 +209,9 @@ jobs:
           working-directory: cli
           args: --new-from-rev=${{ github.event.pull_request.base.sha || 'HEAD~1' }}
 
-  go_unit:
-    name: Go Unit Tests
+  turborepo_tests:
+    name: Go Unit, Integration, and E2E Tests
     needs: determine_jobs
-    if: needs.determine_jobs.outputs.go == 'true'
```
chris-olszewski, line 216: I think we still want to restrict this to run only when either turborepo (Rust or Go) changes

cli/turbo.json
```diff
@@ -3,19 +3,72 @@
   "extends": ["//"],
   "pipeline": {
     "build": {
+      "inputs": ["internal/**"],
```
chris-olszewski, line 6: Not necessarily related to these changes, but in the same ballpark. Should we change our build script? Right now we're just running `make` which produces a ton of junk that's not the turbo executable. If we're being honest we should also list out all of the Rust crates that end up getting used when building our Rust and Go binary

cli/turbo.json
```diff
@@ -3,19 +3,72 @@
...
+      "dependsOn": ["test:e2e"],
+      // Nullifies `outputs` in the root turbo.json
+      "outputs": []
     },
-    "e2e-prebuilt": {
-      "inputs": ["**/*.go", "go.mod", "go.sum", "scripts/e2e/e2e.ts"]
+
+
+    "test:e2e": {
+      // test:e2e depends on test:integration. This is just a way to specify order
```
chris-olszewski, line 23: Reading into this, we need this ordering since `integration-tests` will trigger a turbo binary build which is needed for the `e2e` tests, right?

chris-olszewski (commented):
It feels a little weird duplicating the source files as inputs for the tests. Would it be cleaner to try to properly setup `turbo#build` to track all of the changes that can produce a different turbo binary and then have our tests depend on `build`

cli/Makefile
mehulkar: yeah i don't know why it's duplicated here

.github/workflows/test.yml
mehulkar: i was thinking turborepo can just do this. "always run, but get cache hit" when source files haven't changed

cli/turbo.json
mehulkar: ideally we don't actually need the ordering, since `build` happens before tests are run entirely. the ordering is ONLY for DX right now

cli/turbo.json
mehulkar: added all the crates and stuff (need to verify that it actually invalidates cache on the crates changing)

mehulkar:
> It feels a little weird duplicating the source files as inputs for the tests. Would it be cleaner to try to properly setup turbo#build to track all of the changes that can produce a different turbo binary and then have our tests depend on build

That's a good idea, although actually running `cli#build` is a catch 22, since our entry point is a local build of turbo (via `pnpm -- turbo`, so it's already built

mehulkar:
@chris-olszewski I just opted out of `pnpm -- turbo` and used `pnpm run test` instead, and setup the `test` => `build` dependency as you suggested. I _think_ this should work, but not sure if it impacts other test suites

.github/workflows/test.yml
mehulkar: how soon? Should I find a way to run both in this PR or block this PR until we don't need it anymore? Is https://github.com/vercel/turbo/pull/4181 the thing that makes it so we don't need both?

NicholasLYang: Yeah once we flip the rust switch (hopefully) we won't switch it back and we can start deleting Go code. I'd wait until the switch PR lands and then we can merge this

cli/turbo.json
chris-olszewski: Ah, I forgot that `pnpm -- turbo run test` will end up building turbo right away.

.github/workflows/test.yml
chris-olszewski: My only concern with that is that we'll be waiting on a turbo build before we can decide if we should run.

PR #4176, opened 2023-03-13
Remove the `next-*` crates

sokra:
All next crates are being moved to Next.js repo: vercel/next.js#47019

It also removes next related CI workflows, jobs and actions


crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -220,52 +220,6 @@ impl EcmascriptInputTransform {
                     program.visit_mut_with(&mut resolver(unresolved_mark, top_level_mark, false));
                 }
             }
-            EcmascriptInputTransform::NextJsStripPageExports(export_type) => {
```
jridgewell, line 223: Are we ok with removing these now? I was working on implementing a custom transform API so that Next.js could pass in these transformers on Fri, and aim to finish it today. I was going to leave these in until that work is done so that we haven't already broken the Next.js side.

jridgewell (approved):

PR #4177, opened 2023-03-13
fix(examples): update kitchen-sink dev task

tknickman:
### Description

When running `dev` before running a `build`, the packages won't be built when the apps needed them.

This fixes this by adding:
`"dependsOn": ["^build"]` on `dev` to enforce the correct `build` order _first_ (to make sure the `dev` script can run successfully every time), and then running `dev` to kickoff the watch processes in each workspace

### Testing Instructions

`pnpm dev`

### DIFF
--- a/examples/kitchen-sink/turbo.json
+++ b/examples/kitchen-sink/turbo.json
@@ -21,6 +21,7 @@
       "dependsOn": ["^build"]
     },
     "dev": {
+      "dependsOn": ["^build"],
       "cache": false,
       "persistent": true
     },


nathanhammond (approved):
This is unfortunate; I was under the impression that most of these dev servers were able to handle unbuilt inputs, so that dev was live even through the entire stack.

```
admin (vite) => ui
api => logger
blog (remix) => ui
storefront (next) => ui
```

Do you have a "this doesn't work in X" list?

tknickman:
Good question, I'll compile one and update here

anthonyshew (approved):

PR #4178, opened 2023-03-13
Update docs with example of nested workspaces

maschwenk:
### Description

It would be good to have specific guidance on what defines a "nested workspace". Is the problem with **nested folder structures** or rather **workspaces at different levels of directory depth**? Hopefully that nuance makes sense.

See also 
- #4034


docs/pages/repo/docs/getting-started/existing-monorepo.mdx
```diff
@@ -41,6 +41,8 @@ Specify your `packages` in `pnpm-workspace.yaml`.
 packages:
   - "packages/*"
   - "apps/*"
+  // Note that the below is NOT supported
+  - "nested-workspaces/**"
```
mehulkar, line 45: I think we should make this a `<Callout>` and say something more verbose about [`pnpm` configuration](https://pnpm.io/pnpm-workspace_yaml). 

We can say something like:

"pnpm supports the ability [declare workspaces in subdirectories](https://pnpm.io/pnpm-workspace_yaml) with a `**`. This effectively means that with pnpm you can have nested workspaces. Turborepo does not support this at the moment."

maschwenk: 👍🏼 SGTM

anthonyshew (approved):
Moving this up into the `pnpm`-specific tab and tightening up the language slightly in line with Mehul's suggestion!

Thanks for this contribution! Great addition to the docs.

maschwenk:
@anthonyshew ah sorry for dropping the ball on the suggestion 🙏🏼

PR #4180, opened 2023-03-13
Ensure we record the lockfile name when there's an error unmarshalling

gsoltis:
### Description

 - Adds the name of the lockfile we tried to unmarshal to the error we display when we fail to unmarshal it.

### Testing Instructions

Tested manually by editing Turbo's own lockfile to be invalid

### DIFF
--- a/cli/internal/packagemanager/packagemanager.go
+++ b/cli/internal/packagemanager/packagemanager.go
@@ -5,12 +5,12 @@
 package packagemanager
 
 import (
-	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


gsoltis:
@chris-olszewski Is there an equivalent that we need on the rust side?

gsoltis:
cc @sokra

chris-olszewski (approved):

chris-olszewski:
> @chris-olszewski Is there an equivalent that we need on the rust side?

Nope

PR #4182, opened 2023-03-13
Factor out file parsing

nathanhammond:
The goal is to start slimming down and organizing `shim.rs` as the amount of business logic in it has grown.

### DIFF
--- a/crates/turborepo-lib/src/files/package_json.rs
+++ b/crates/turborepo-lib/src/files/package_json.rs
@@ -1,0 +1,68 @@
+use std::{fs, path::PathBuf};
+
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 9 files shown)


NicholasLYang (approved):

crates/turborepo-lib/src/files/package_json.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub workspaces: Option<Workspaces>,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+#[serde(untagged)]
+pub enum Workspaces {
+    TopLevel(Vec<String>),
+
+    // This only works in old npm.
+    Nested { packages: Vec<String> },
```
tknickman, line 18: 🙏🏼

tknickman (approved):

anthonyshew:
Work was done here so I think this is outdated now! Going to close.

PR #4184, opened 2023-03-13
fix: SCM relative paths in nested turborepo

NicholasLYang:


gsoltis:
Looks like `relative_to` should not be an `Option`, it is always set on the Go side, despite there being logic to handle it being empty: https://github.com/vercel/turbo/blob/main/cli/internal/scope/scope.go#L202

NicholasLYang:
I'm wondering if we should just rename `relative_to`  to `monorepo_root`, because that's really what it is. If we end up using this function in other places that won't necessarily be the case, but for now it's accurate.

gsoltis (approved):

gsoltis:
Re: rename, go for it

crates/turborepo-scm/src/git.rs
```diff
@@ -299,18 +303,60 @@ mod tests {
...
         )?;
         assert_eq!(files, HashSet::from(["bar.js".to_string()]));
 
         Ok(())
     }
 
+    // Tests that we can use a subdir as the monorepo_root path
+    // (occurs when the monorepo is nested inside a subdirectory of git repository)
+    #[test]
+    fn test_changed_files_with_subdir_as_monorepo_root() -> Result<(), Error> {
```
arlyon, line 319: Cool 👍

arlyon (approved):

nathanhammond:
So we have two roots that we care about:
- project root (detected by `RepoState::infer`) (`monorepo_root`) in this PR.
- SCM root

We should be renaming some things:
`RepoState` => `ProjectState`
`repo_root` => `project_root`

But, for bonus points, because of things like `git submodule`, you're not guaranteed that the entire "changed state" will be a child of the closest SCM root. At some point we need to circle back to handle the arbitrary ordering of this inside of `ProjectState::infer`.

Fun tricks: `git rev-parse --show-cdup` shows the `../..` prefix to the closest repository root from cwd.

PR #4186, opened 2023-03-13
fix(turborepo): race condition in config tests

NicholasLYang:
### Description

We had a subtle race condition in our config tests where if a test set the `TURBO_REMOTE_CACHE_TIMEOUT` env var, it would set the `remote_cache_timeout` field on the config objects in other tests. Fixed this by running the test functions serially inside one test.

### DIFF
--- a/crates/turborepo-lib/src/config/client.rs
+++ b/crates/turborepo-lib/src/config/client.rs
@@ -93,7 +93,17 @@ mod test {
 
     use super::*;
 
+    // We group these test functions under one test to
+    // avoid race conditions with environment variables
     #[test]
+    fn all_tests() -> Result<()> {
+        test_client_default()?;
+        test_client_arg_variable()?;
+        test_client_env_variable()?;
+
+        Ok(())
+    }
+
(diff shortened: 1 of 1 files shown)


gsoltis:
Would it be sufficient to delete the test that is setting an environment variable explicitly? It looks like we also test by passing in an hashmap with the same values. If we're pretty confident of our ability to read environment variables, we could keep only the variant that uses the hashmap.

NicholasLYang:
We could do that, but I almost want to keep this test as a signifier that race conditions can happen with config testing. It'll be far too easy to make this mistake again (I made it and Nathan appears to have made it too). And because it's not deterministic, CI won't catch the bug either.

nathanhammond:
Amusingly this is after I explicitly noted that I _couldn't_ do this:

https://github.com/vercel/turbo/blob/c7002c903d7fddbcd46f8eb88b8fc959b7b8ce73/crates/turborepo-lib/src/config/client.rs#L230-L231

@NicholasLYang good catch, thanks for cleaning up.

PR #4188, opened 2023-03-14
Add "nextpack" integration setup

jridgewell:
Hopefully this'll ease our developing in Next.js

```
Nextpack is the development sub-repo, allowing us to develop Next.js
features and Turbopack together. This sub-repo will stitch together the
rust crates in Next.js and the crates in Turbo, so they appear as 1
logical Cargo workspace.

There are two steps:

```bash
# This will clone next.js into ./next.js
# if you wanted, you can link your own already-checked-out next.js repo.
$ ./scripts/init.cjs
# This read the Cargo.toml from Next.js and Turbo, generating a new
# Cargo.toml workspace in this directory.
$ cargo run --bin sync-workspace
```

### Failed to load manifest for workspace member …

You can't have an optional workspace member (or if you can, I can't
figure it out). So the `./next.js/…` crate members are gonna error out
if you try to run a Turbo crate without first cloning next.js.

To fix, run `./scripts/init.js` again. It'll resync your Cargo
workspace.

### Why is `init` written in JS?

See the above. If we wrote it in Rust, any failure in the repo setup
would prevent it from running.

### Why is there a turbo-crates symlink dir?

Cargo cannot contain nested workspaces, and if we used
`../crates/turbopack` as a workspace member, it would error out. By
using a symlink, we avoid the problem.

### How do I run `next dev --turbo` with this?

I'm still working on it.
```

### CODE REVIEW

**Review by sokra — APPROVED** [review]
I added a few things and it should be good to go now:

* Cargo.lock is also synced
* Cargo.lock and Cargo.toml are git ignored
* It prints a list of dependency conflicts when merging
* Adjusted some stuff for windows support
* It automatically installs and updates the lockfile
* It checks for swc_core conflicts

File: nextpack/crates/sync-workspace/src/main.rs [hunk 1]
```diff
@@ -0,0 +1,311 @@
...
+    cargo_toml["workspace"]
+        .as_table_mut()
+        .unwrap()
+        .insert("dependencies".to_string(), Value::Table(dependencies));
+
+    let mut toml_file = File::options()
+        .write(true)
+        .truncate(true)
+        .open("Cargo.toml")?;
+    writeln!(toml_file, "# THIS FILE IS AUTOGENERATED BY np-sync-crates")?;
```
alexkirsz, line 241:
```suggestion
    writeln!(toml_file, "# THIS FILE IS AUTOGENERATED BY sync-workspace")?;
```

nextpack/scripts/init.cjs
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env node
```
alexkirsz, line 1: Maybe rename this to `sync.cjs`, since it needs to be run every time the workspaces are modified.

nextpack/crates/sync-workspace/src/main.rs
```diff
@@ -0,0 +1,311 @@
...
+    writeln!(
+        toml_file,
+        "# Do NOT make changes to this file, instead change"
+    )?;
+    writeln!(toml_file, "# - {}/Cargo.toml, and", TURBO_PATH)?;
+    writeln!(toml_file, "# - {}/Cargo.toml", TURBO_PATH)?;
```
alexkirsz, line 247: This should be `NEXT_PATH`.

```suggestion
    writeln!(toml_file, "# - {}Cargo.toml", TURBO_PATH)?;
    writeln!(toml_file, "# - {}Cargo.toml", NEXT_PATH)?;
```

Also, need to remove the trailing /, otherwise we get the following:

```
# THIS FILE IS AUTOGENERATED BY np-sync-crates
# Do NOT make changes to this file, instead change
# - ./turbo-crates//Cargo.toml, and
# - ./turbo-crates//Cargo.toml
# and run `cargo run --bin sync-workspace`.
```

alexkirsz (changes requested):

nextpack/scripts/init.cjs
```diff
@@ -0,0 +1,31 @@
...
+function exec(command, opts) {
+  console.log(command);
+  execSync(command, {
+    stdio: "inherit",
+    ...opts,
+  });
+}
+
+if (!existsSync("./next.js")) {
+  exec('git clone "https://github.com/vercel/next.js.git" --depth 100');
```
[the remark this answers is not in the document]
alexkirsz: (actually I don't have a strong argument either way, but I'm used to SSH so I changed the remote manually)

jridgewell (commented):
> Cargo.lock is also synced

What does this gain us? Why not let Cargo handle this when it runs after creating the `Cargo.toml`?

> Cargo.lock and Cargo.toml are git ignored

Yah, this works much better since I switched to the node `init.cjs` instead of a cargo bin. Thanks for fixing that.

sokra:
> > Cargo.lock is also synced

> What does this gain us?

It installs the right version of dependencies. Before it was broken due to selecting the latest version, which e.g. ended up with two versions of swc_core

Also it's more consistent with the actual version used in next, and the conflict output tells us to update next.js or turbo lockfiles.

PR #4190, opened 2023-03-14
Run all examples in same Github Action run

mehulkar:
Compiling turbo is the slowest part of running these tests in CI, so instead of doing it 18 times, we can do it once and run the tests all together. We can also parallelize at the test level instead.

Anecdotally, 

**Before**:
Each CLI example took ~10m (x18 jobs). Expand the `test` workflow here in this example: https://github.com/vercel/turbo/pull/4000/checks

**After this**
All `example_tests/` also took 10m23s, but only 2 jobs run. The actual test time is a little bit larger, but since the bulk of the time for the job is building Turbo, I think this is a win. Expand the `test` workflow here: https://github.com/vercel/turbo/pull/4190/checks

## Parallalization

```bash
hyperfine -r=3 \
    --show-output --ignore-failure \
    --export-markdown=turbo-example-tests.md  \
    'pnpm run run-example' 'PARALLEL=true pnpm run run-example'
```

(I benchmarked both by enabling `PARALLEL=true`, but in this PR, only the parallel version is available)

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `pnpm run run-example` | 145.596 ± 0.406 | 145.128 | 145.854 | 2.07 ± 0.02 |
| `PARALLEL=true pnpm run run-example` | 70.198 ± 0.644 | 69.507 | 70.780 | 1.00 |

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -313,7 +313,6 @@ jobs:
     needs: determine_jobs
     if: needs.determine_jobs.outputs.examples == 'true'
     timeout-minutes: 30
-
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


mehulkar:
The integration test failure is not related to this PR, I'm trying to debug that in https://github.com/vercel/turbo/pull/4210

arlyon (approved):
I am for this even if there aren't huge wall clock gains. Help reduce our co2 footprint :footprints:  :earth_africa: :leaves:

tknickman (approved):
We lose testing against different OS's with this approach, but for the sake of speed I'm fine with that.

mehulkar:
> We lose testing against different OS's with this approach, but for the sake of speed I'm fine with that.

The matrix still has the different OS's. You can see in this PR that the examples tests are running both macOS and ubuntu also. (not sure why Windows isn't part of the config already, but I wasn't planning to change that here)

CleanShot 2023-03-20 at 10 42 14@2x

mehulkar:
Closing down now that #5011 is merged.

PR #4191, opened 2023-03-14
Rearrange shim into layers

nathanhammond:
This PR is fourth in a stack. It makes no material changes to the actual code other than copy/paste, changing the paths by which certain properties are accessed, and littering it with `.clone().unwrap()` until it compiled and passed tests again.

This is step four in the process of exploding `shim.rs` into components and making the `turbo` that is being run responsible for itself.

This is an interim-state, show-your-work PR. It can land once all of the `.clone().unwrap()` mess gets cleaned up.

Changes in this PR:
https://github.com/vercel/turbo/pull/4191/files/45951366daaec56bc08b67d4cf828a4877ebe735..a38426d2777ed17f9a2dee37be886910c4fef67b

Something like this is where I'm headed, implementable by both `TurboState` and `LocalTurboState` with `TurboState` cleverly delegating:

```
trait Executable {
    fn platform_package_name() -> String;
    fn binary_name() -> String;
    fn bin_path() -> String;
    fn version() -> String;
    fn run() -> String; // cli::run vs. spawn_local_turbo
    fn is_outdated() -> bool;
    fn has_shim() -> bool;
    fn is_outer_turbo() -> bool;
    fn is_acceptable_version() -> bool;
}
```


anthonyshew:
Just closing up some stales!

PR #4198, opened 2023-03-14
feat(ecmascript): support jsximport/runtime options

kwonoj:
### Description

First part of WEB-719 implementation. PR expands enable_jsx option to accept few more configuration point to jsx transforms, mainly its import source. We'll likely to expand options more in the future.

One thing I noted in the code but didn't make change is `enable_react_refresh`, if it needs to be under the jsx option or should be independent. Given react_refresh only should be enabled when jsx is enabled, for me it makes more sense to move it.

### DIFF
--- a/crates/turbopack-ecmascript/src/transform/mod.rs
+++ b/crates/turbopack-ecmascript/src/transform/mod.rs
@@ -16,7 +16,7 @@ use swc_core::{
         visit::{FoldWith, VisitMutWith},
     },
 };
-use turbo_tasks::primitives::StringVc;
+use turbo_tasks::primitives::{OptionStringVc, StringVc};
 use turbo_tasks_fs::json::parse_json_with_source_context;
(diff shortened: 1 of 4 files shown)


wbinnssmith (approved):

sokra:
This PR doesn't have a next.js counter part...

crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -33,6 +33,10 @@ pub enum EcmascriptInputTransform {
     React {
         #[serde(default)]
         refresh: bool,
+        // swc.jsc.transform.react.importSource
+        import_source: OptionStringVc,
+        // swc.jsc.transform.react.runtime,
+        runtime: OptionStringVc,
```
alexkirsz, line 39: I would rather this be an enum instead, since there are only two valid values. This avoids downstream validation.

PR #4200, opened 2023-03-14
fix: prune pnpm patched dependencies with peer dependencies

### LINKED ISSUE #4133: [turborepo] Turbo prune not handling pnpm patchedDependencies in the workspace root

1.7.1, 1.7.4, 1.8.3

pnpm

Windows

### Describe the Bug

This was previously reported on #2788.  This [comment](https://github.com/vercel/turbo/issues/2788#issuecomment-1416398304) indicated that this problem was fixed in 1.7.1.  However, I was running 1.7.4 and experienced the problem.  I updated to 1.8.3 to see if it had actually been fixed in a later release, but the problem persisted on that release as well.  I tried downgrading to 1.7.1 to see if there had been a regression but the problem existed on that release as well.

Effectively, `turbo prune` does not appear to be handling pnpm patches at all.  After pruning, the root `package.json` that is copied into the `out` folder is left with: `"patchedDependencies": {}` when the original `package.json` had content in the object.  The `out` folder version of `pnpm-lock.yaml` does not have the `patchedDependencies` property present in it at all.  Finally, the `patches` folder, which is necessary to apply the patch, isn't present in the `out` folder after pruning.

I believe there must be something unique about my repo that is triggering this.  I tried to create a minimal reproduction from the ground up and everything worked properly.  Since I was unable to reproduce that way, I took my repo and kept removing pieces of it until I had it relatively small.  It had started with several packages in the workspace, but I was able to eliminate all but one and still reproduce the issue.  I'm not sure what is special about this repo that makes the patching fail to work properly, but hopefully providing it will assist.

### Expected Behavior

The `out` folder will have a `package.json` which retains the `patchedDependencies` property in full, the `pnpm-lock.yaml` file will retain its `patchedDependencies` property in full, and the `patches` folder will be copied in.

### To Reproduce

I have created a reproduction at https://github.com/shawnmcknight/turbo-pnpm-patch-repro.  The steps to reproduce are in the README.

### Reproduction Repo

https://github.com/shawnmcknight/turbo-pnpm-patch-repro

chris-olszewski:
### Description

Fixes #4133 (huge thanks to @shawnmcknight for the reproduction)

The crux of the issue was that if a dependency is patched and peer dependencies than both get encoded into the lockfile key. We can't construct a full expected lockfile key from the patch entry anymore as the patch entry doesn't contain information about peer dependencies. To circumvent this we now scan through the entire package list, parse the key, extract the patch hash (if present), and then check if it matches a patch listed in `patchedDependencies`.

Notes for reviewers:
 - A large portion of this PR is actually porting the lockfile key parsing logic that pnpm itself uses so we can extract the individual pieces that we care about. This logic is in a slightly messy state as the pnpm codebase is currently in a state where it supports v5 and v6 lockfiles. As mentioned in https://github.com/pnpm/pnpm/pull/5810 once pnpm@8 lands this code will start to be refactored. I tried to keep the Go port as 1:1 as possible in hopes to make future maintenance easier.  
 - This PR includes new logic for v6 as pnpm now encodes patches in lockfile keys: https://github.com/pnpm/pnpm/pull/5979

### Testing Instructions

Added new packages to the pnpm lockfile fixtures that are patched *and* have peer dependencies.

Also verify that the changes work as expected with the provided repro:
 - `make turbo`
 - Clone https://github.com/shawnmcknight/turbo-pnpm-patch-repro
 - Remove private dependency: `pushd packages/test && pnpm rm @sheet/core && popd`
 - `turbo_dev --skip-infer prune --scope=test`
 - Verify that `out/pnpm-lock.yaml` contains a `patchedDependencies` section with both `@types/jsonwebtoken@8.5.9` and `moleculer@0.14.28` as entries


cli/internal/lockfile/pnpm_lockfile.go
```diff
@@ -477,17 +470,122 @@ func formatPnpmKeyV6(name string, version string) string {
...
 func getVersionFromKey(key string) string {
-	atIndex := strings.LastIndex(key, "/")
-	if atIndex == -1 || len(key) == atIndex+1 {
-		return ""
+	dp := parseDepPath(key)
+	if dp.peerSuffix != "" {
+		return fmt.Sprintf("%s_%s", dp.version, dp.peerSuffix)
 	}
-	return key[atIndex+1:]
+	return dp.version
```
nathanhammond, line 477: What about moving these into `PnpmLockfile` as methods?

It already has `.isV6`, so you could switch there, or alternatively a `PnpmLockfile` interface with `DecodePnpmLockfile` returning a `PnpmLockfile5` or a `PnpmLockfile6`?

(`PnpmLockfile` would probably end up renamed to something like `RawPnpmLockfile`.)

cli/internal/lockfile/pnpm_lockfile.go
```diff
@@ -477,17 +470,122 @@ func formatPnpmKeyV6(name string, version string) string {
...
+	sepIndex := strings.Index(d.peerSuffix, "_")
+	if sepIndex != -1 {
+		return d.peerSuffix[:sepIndex]
+	}
+	// if a dependency just has a single suffix we can't tell if it's a patch or peer hash
+	// return it in case it's a patch hash
+	return d.peerSuffix
+}
+
+func convertNewToOldDepPath(newPath string) string {
```
nathanhammond, line 580: Is this logic from the same SHA as above? Can you add a comment to make that super-clear?

mehulkar (approved):
Do we want an integration test here instead of unit tests? Maybe using  an even smaller repro of the test repo?

I was not able to try this out on the reproduction repo, because it has some dependencies that are private (`@sheet/core`)

nathanhammond (approved):
I'm not going to block on a significant refactor to code we're going to retire soon, but the way in which we assign functions as instance not-methods feels a little gnarly.

PR #4205, opened 2023-03-15
Throw an error when a task graph cannot run to completion.

### LINKED ISSUE #4195: [turborepo] Do something smart if `persistent: true` tasks will (or have) exhausted available concurrency.
### Describe the Bug

If you have more than 10 concurrent parallel tasks you can starve the rest of your graph and never process it. 

### Expected Behavior

We should do something smart here, probably a combination of both of these:

1. Warn or error in advance. After building the graph we can walk by distance from root by graph branch and identify the maximum number of persistent tasks and hard error.
2. If all execution slots are taken up by persistent tasks, grow available concurrency by one. (Only one non-persistent task is ... not great.)

***

Originally reported in #4107. Thanks, @loick!

nathanhammond:
Throw an error if the task graph will never run to completion.

Setting `--concurrency` for the user to N+1 is likely a worse experience by linearizing everything that _isn't_ persistent, this gives the user enough information to tune their invocation.

Fixes #4195

### DIFF
--- a/cli/integration_tests/composable_config/composing-persistent.t
+++ b/cli/integration_tests/composable_config/composing-persistent.t
@@ -11,9 +11,9 @@ This test covers:
 # persistent-task-1-parent dependsOn persistent-task-1
 # persistent-task-1 is persistent:true in the root workspace, and does NOT get overriden in the workspace
   $ ${TURBO} run persistent-task-1-parent --filter=persistent
-   ERROR  run failed: error preparing engine: Invalid persistent task dependency:
+   ERROR  run failed: error preparing engine: Invalid persistent task configuration:
   "persistent#persistent-task-1" is a persistent task, "persistent#persistent-task-1-parent" cannot depend on it
-  Turbo error: error preparing engine: Invalid persistent task dependency:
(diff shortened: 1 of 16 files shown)


arlyon (approved):

mehulkar (approved):

ObliviousHarmony:
> Setting --concurrency for the user to N+1 is likely a worse experience by linearizing everything that isn't persistent, this gives the user enough information to tune their invocation.

Would you be able to share a little more context around this rationale @nathanhammond? In my experience thus far, any invocation of a persistent task requires me to enter it a second time with `--concurrency` set. I haven't come up with any persistent tasks that you might want to be limited by concurrency by default. I could see using the option to limit it, but, it feels like it should be unlimited by default, or, there should be an option to run it with unlimited concurrency. Would the preferred approach just be to set something obscene like `--concurrency=100000`?

We're currently wrapping `turbo` commands in NPM scripts to avoid needing anyone to learn the syntax. I could set a default concurrency like that, I'm just curious about why the N+1 approach wasn't taken in the first place.

Thanks for the great work either way!

PR #4207, opened 2023-03-15
fix: Turbo SCM with Windows paths

NicholasLYang:
### Description

Fixes a bug where turborepo-scm would return unix style paths on Windows.

### Testing Instructions

`test_changed_files_with_subdir_as_monorepo_root` is modified to test that the path returned is a system path

### DIFF
--- a/crates/turborepo-scm/src/git.rs
+++ b/crates/turborepo-scm/src/git.rs
@@ -370,7 +370,15 @@ mod tests {
             None,
             true,
         )?;
-        assert_eq!(files, HashSet::from(["src/bar.js".to_string()]));
+        #[cfg(not(windows))]
+        {
+            assert_eq!(files, HashSet::from(["src/bar.js".to_string()]));
+        }
+
+        #[cfg(windows)]
+        {
+            assert_eq!(files, HashSet::from(["src\\bar.js".to_string()]));
+        }
 
         commit_file(&repo, Path::new("subdir/src/bar.js"), Some(first_commit))?;
 
--- a/crates/turborepo-scm/src/git.rs
+++ b/crates/turborepo-scm/src/git.rs
@@ -355,6 +355,8 @@ mod tests {
         config.set_str("user.email", "test@example.com")?;
 
         fs::create_dir(repo_root.path().join("subdir"))?;
+        // Create additional nested directory to test that we return a system path
+        // and not a normalized unix path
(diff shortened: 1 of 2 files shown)


crates/turborepo-scm/src/git.rs
```diff
@@ -64,7 +66,38 @@ pub fn changed_files(
     Ok(files)
 }
 
+// Gets the system version of both paths by calling `fs_util::canonicalize`,
```
mehulkar, line 69: Can you update this comment to be specific about "both paths"? Since there are 3 args to the function, I can't tell at a glance which is being normalized

crates/turborepo-scm/src/git.rs
```diff
@@ -64,7 +66,38 @@ pub fn changed_files(
...
+// then strips the `monorepo_root` from the `file_path`.
+fn get_stripped_system_file_path(
+    repo_root: &ProjectRoot,
+    file_path: &Path,
+    monorepo_root: &ProjectRelativePath,
```
chris-olszewski, line 74: Not necessary in this PR, but I find `turbo_root` to be clearer at what this param describes

crates/turborepo-scm/src/git.rs
```diff
@@ -64,7 +66,38 @@ pub fn changed_files(
...
+) -> Result<PathBuf, Error> {
+    // We know the path is relative to the repo root so we can convert it to a
+    // ProjectRelativePath
+    let project_relative_file_path = ProjectRelativePath::new(file_path)?;
+    // Which we then resolve to an absolute path
+    let absolute_file_path = repo_root.resolve(project_relative_file_path);
+    // Then we call canonicalize to get a system path instead of a Unix style path
+    let path = fs_util::canonicalize(absolute_file_path)?;
```
mehulkar, line 82: Just making sure I understand, where is `file_path` coming from that it's not already a platform-specific path? Is it from user config?

mehulkar (approved):

chris-olszewski (commented):
In the future, I'd very much appreciate filling out the PR template. It's especially for changes like this which are meant to solve a non-obvious problem.

crates/turborepo-scm/src/git.rs
NicholasLYang: It's coming from git, which always returns a unix-style path

PR #4208, opened 2023-03-15
track invalidation reasons

sokra:
### Description

tracks reasons for invalidation and reports them as updated event

fixes WEB-722
fixes WEB-221


padmaia:
Could you add a screenshot of what the new messages look like? 🙏

sokra:
image

crates/turbo-tasks-fs/src/invalidation.rs
```diff
@@ -0,0 +1,64 @@
...
+struct WatchChangeType {
+    _non_zero_sized: u8,
+}
+
+static WATCH_CHANGE_TYPE: WatchChangeType = WatchChangeType { _non_zero_sized: 0 };
+
+impl InvalidationReasonType for WatchChangeType {
+    fn description(&self, merge_data: &IndexSet<Cow<'static, str>>) -> Cow<'static, str> {
+        format!(
+            "{} files changed (e. g. {})",
```
alexkirsz, line 28: There should be two modes:
```
1 file changed ({path})
```
and 
```
N files changed (e.g. {path})
```

crates/turbo-tasks/src/invalidation_reason_set.rs
```diff
@@ -0,0 +1,113 @@
...
+        data: Cow<'static, str>,
+    },
+    Multiple {
+        reason_ty: &'static dyn InvalidationReasonType,
+        merge_data: IndexSet<Cow<'static, str>>,
+    },
+}
+
+#[derive(Default)]
+pub struct InvalidationReasonSet {
```
alexkirsz, line 31: Add comments that explains what this represents and how it's used.

crates/turbo-tasks/src/invalidation_reason_set.rs
```diff
@@ -0,0 +1,113 @@
...
+    next_unique_tag: usize,
+    map: IndexMap<MapKey, MapEntry>,
```
alexkirsz, line 33: There should be two maps: a typed one, and an untyped one. Otherwise types get muddy below and you need an `unreachable`.

```suggestion
    typed_map: IndexMap<MapKey, MapEntry>,
    untyped_map: IndexMap<MapKey, MapEntry>,
```

crates/turbo-tasks/src/manager.rs
```diff
@@ -1028,6 +1106,22 @@ fn current_task(from: &str) -> TaskId {
...
+        None
+    }
+}
+
+pub trait InvalidationReasonType: Send + Sync + 'static {
+    fn ptr(&self) -> usize {
+        // To have a valid pointer this must not be a ZST
+        debug_assert!(size_of_val(self) != 0);
+        self as *const _ as *const () as usize
+    }
```
alexkirsz, line 1121: Looks like this is only used for keying. Why not return a `TypeId` instead? This would work for ZSTs as well.

crates/turbo-tasks/src/manager.rs
```diff
@@ -1028,6 +1106,22 @@ fn current_task(from: &str) -> TaskId {
...
+    fn description(&self, merge_data: &IndexSet<Cow<'static, str>>) -> Cow<'static, str>;
```
alexkirsz, line 1122: Same Q as above: why not pass the formatter here?

crates/turbopack-dev-server/src/invalidation.rs
```diff
@@ -0,0 +1,35 @@
...
+    _non_zero_sized: u8,
+}
+
+static SERVER_REQUEST_TYPE: ServerRequestType = ServerRequestType { _non_zero_sized: 0 };
+
+impl InvalidationReasonType for ServerRequestType {
+    fn description(&self, merge_data: &IndexSet<Cow<'static, str>>) -> Cow<'static, str> {
+        let example = merge_data
+            .into_iter()
+            .reduce(|a, b| if b.len() < a.len() { b } else { a })
```
alexkirsz, line 31:
```suggestion
            .min_by_key(|data| data.len())
```

crates/turbo-tasks-fs/src/invalidation.rs
```diff
@@ -0,0 +1,64 @@
+use std::borrow::Cow;
+
+use indexmap::IndexSet;
+use turbo_tasks::{InvalidationReason, InvalidationReasonType};
+
+pub struct WatchChange {
```
alexkirsz, line 6: Add a few short comments explaining what these public structs represent.

alexkirsz (changes requested):

crates/turbo-tasks-fs/src/invalidation.rs
sokra: `InvalidationReasonType::description` will only called for more than 1 merge_data entry.

In the case of a single one, it will use `InvalidationReason::description`.

crates/turbo-tasks/src/invalidation_reason_set.rs
sokra: I initially had that, but that won't keep the order of reasons intact when mixing typed and untyped reasons.

crates/turbo-tasks/src/invalidation_reason_set.rs
```diff
@@ -0,0 +1,113 @@
...
+        });
+        let count = self.map.len();
+        for (i, description) in descriptions.enumerate() {
+            if i > 0 {
+                write!(f, ", ")?;
+                if i == count - 1 {
+                    write!(f, "and ")?;
+                }
+            }
+            write!(f, "{}", description)?;
```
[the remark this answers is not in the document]
sokra: It's used as part of the `update in` message, see here: https://github.com/vercel/next.js/pull/47190

So it's better in a single line.

crates/turbo-tasks/src/invalidation_reason_set.rs
alexkirsz: I see. You could add an ordering `idx` to each entry but that might have its own issues. Let's add a comment to explain why everything is in a single map.

crates/turbo-tasks/src/manager.rs
```diff
@@ -1028,6 +1106,22 @@ fn current_task(from: &str) -> TaskId {
     }
 }
 
+pub trait InvalidationReason: Send + Sync + 'static {
+    fn description(&self) -> Cow<'static, str>;
```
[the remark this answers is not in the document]
sokra: Good idea

PR #4212, opened 2023-03-15
Add a hash of external dependencies to run summary

mehulkar:


cli/internal/runsummary/run_summary.go
```diff
@@ -110,6 +110,7 @@ type TaskSummary struct {
 	Framework              string                                `json:"framework"`
 	EnvVars                TaskEnvVarSummary                     `json:"environmentVariables"`
 	Execution              *TaskExecutionSummary                 `json:"execution,omitempty"` // omit when it's not set
+	ExternalDepsHash       string                                `json:"hashOfExternalDependencies"`
```
mehulkar, line 113: open to bikeshedding the json field name. I almost want to call it `lockfile`, even though it's not the full contents of the lockfile. Since it's a hash, the user won't know what changed anyway, and especially for a lockfile, I don't know that we'll ever need to expand this into which specific packages changes. 

And that aside, it seems that long term we should just hash the entire contents of the lockfile anyway?

tknickman: I think this is fine, more verbose is better in this case.

tknickman (approved):

PR #4216, opened 2023-03-15
Exclude snapshots from VSCode search

mehulkar:
make it easier to find the thing I'm looking for

| Before | After |
| --- | --- |
| CleanShot 2023-03-15 at 15 16 58@2x | CleanShot 2023-03-15 at 15 17 59@2x|

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -21,5 +21,8 @@
     "files.insertFinalNewline": false,
     "files.trimTrailingWhitespace": false
   },
-  "rust-analyzer.linkedProjects": ["nextpack/Cargo.toml"]
+  "rust-analyzer.linkedProjects": ["nextpack/Cargo.toml"],
+  "search.exclude": {
+    "crates/turbopack-tests/tests/snapshot/**": true
+  }
 }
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -21,5 +21,8 @@
     "files.insertFinalNewline": false,
     "files.trimTrailingWhitespace": false
   },
-  "rust-analyzer.linkedProjects": ["nextpack/Cargo.toml"]
+  "rust-analyzer.linkedProjects": ["nextpack/Cargo.toml"],
+  "search.exclude": {
+    "crates/turbopack-tests/tests/snapshot/**": true
+  }
```
jridgewell, line 27: Nit: This is JSON-C
```suggestion
  },
```

jridgewell (approved):

.vscode/settings.json
nathanhammond: A few notes:
- This fails our prettier config check since the file extension is JSON.
- JSON5 is the one that explicitly allows trailing commas.
- The closest thing to a "reference implementation" for JSONC has `allowTrailingComma` off by default.

https://github.com/microsoft/node-jsonc-parser/blob/18b2209dafefc8506165ddfb1b8a9863d09ff09d/src/impl/parser.ts#L22-L26

Here's the failure:
https://github.com/vercel/turbo/actions/runs/4433267850/jobs/7778146913

And the removal of this comma is coming along for the ride in #4221.

PR #4220, opened 2023-03-16
Fix path to turbo

mehulkar:
This along with https://github.com/gsoltis/large-monorepo/pull/1 should make `pnpm run benchmark` work again


benchmark/src/index.ts
```diff
@@ -10,7 +10,7 @@ const REPO_PATH = path.join(process.cwd(), REPO_ROOT);
 const REPETITIONS = 5;
 
 const DEFAULT_EXEC_OPTS = { stdio: "ignore" as const, cwd: REPO_PATH };
-const TURBO_BIN = path.resolve(path.join("..", "cli", "turbo"));
+const TURBO_BIN = path.resolve(path.join("..", "target", "debug", "turbo"));
```
nathanhammond, line 13:
```suggestion
const TURBO_BIN = path.resolve(path.join("..", "target", "release", "turbo"));
```

Running benchmarks against a debug build is gonna be invalid. I don't know if we have a release build on this task or not.

nathanhammond (changes requested):

benchmark/src/index.ts
mehulkar: I had to add a `make` target that used `--release`  to get this. Not sure why we don't already have one or if we should be using it everywhere else also, but I agree we should use an optimized build[^1].

[^1]: I haven't looked up what an optimized build is in Rust/cargo is though

mehulkar: Note that this script expects turbo to already be built, so the new make target doesn't affect anything else

CONTRIBUTING.md
```diff
@@ -7,14 +7,16 @@ Thanks for your interest in contributing to Turbo!
 - [Contributing to Turbo](#contributing-to-turbo)
   - [Contributing to Turborepo](#contributing-to-turborepo)
     - [Building Turborepo](#building-turborepo)
+    - [TLS Implementation](#tls-implementation)
```
mehulkar, line 10: some vscode magic extension added these in. I guess it recognized that this is a ToC and noticed missing headings. Pretty cool

tknickman (approved):

cli/Makefile
```diff
@@ -30,6 +30,9 @@ GENERATED_FILES = internal/turbodprotocol/turbod.pb.go internal/turbodprotocol/t
 turbo: go-turbo$(EXT)
 	cargo build --manifest-path ../crates/turborepo/Cargo.toml
 
+turbo-prod: go-turbo$(EXT)
```
mehulkar, line 33: @NicholasLYang @chris-olszewski @arlyon did you have any other plans to enable this?

NicholasLYang: Not really. The actual release process just directly invokes `cargo` with `--release`. The only concern I have is that we release with `rustls-tls` even though `native-tls` is activated for development workflows. We should probably just switch to `rustls-tls` by changing the default feature in `crates/turborepo/Cargo.toml`. That way we're benchmarking with the correct setup.

mehulkar: Is this good enough? https://github.com/vercel/turbo/pull/4239

mehulkar:
@nathanhammond want to give this your blessing?

nathanhammond (approved):

nathanhammond:
> @nathanhammond want to give this your blessing?

One approval and my thing taken care of is good. 👍

PR #4221, opened 2023-03-16
Make the script error ordering deterministic.

nathanhammond:
`stdout` and `stdin` don't necessarily provide guarantees in terms of ordering. We're looking at both in the logs. To ensure determinism (because this isn't what this test is testing), simply remove the write to `stdout`.


nathanhammond:
Heads up: going to force merge once CI is done because it's test-only and fixes CI.

PR #4222, opened 2023-03-16
add missing turborepo exclude in turbopack tests

sokra:
### Description

new crate was added which need to be excluded for the tests


jridgewell (approved):

sokra:
> Is there a more sustainable solution to this?

We could run all test cases together... and figure out the crates that need to be tested from diff...

PR #4225, opened 2023-03-16
Add cache state to task summaries on real runs

mehulkar:
This PR returns the cacheState all the way down to `execContext.exec()` , and then stores it in the taskHashTracker so we can assign it to the taskSummary. I'm not sure if `taskHashTracker` is the best place to keep this data, but we're storing other things there with the same pattern there for the same purpose, so I felt like it was a good fit.

I'm not easily able to write a test that validates that this works against a Remote cache, but because of how the `cache` package is designed, it should Just Work™.

### DIFF
--- a/cli/integration_tests/basic_monorepo/run_summary/run_summary.t
+++ b/cli/integration_tests/basic_monorepo/run_summary/run_summary.t
@@ -61,7 +61,7 @@ Setup
   [
     "apps/my-app/.turbo/turbo-build.log"
   ]
-
+# validate that cache state updates in second run
   $ echo $FIRST_APP_BUILD | jq '.cacheState'
   {
     "local": false,
--- a/cli/integration_tests/basic_monorepo/run_summary/run_summary.t
+++ b/cli/integration_tests/basic_monorepo/run_summary/run_summary.t
@@ -6,6 +6,17 @@ Setup
   $ rm -rf .turbo/runs
 
   $ TURBO_RUN_SUMMARY=true ${TURBO} run build -- someargs > /dev/null # first run (should be cache miss)
+
+# HACK: Generated run summaries are named with a ksuid, which is a time-sorted ID. This _generally_ works
+# but we're seeing in this test that sometimes a summary file is not sorted (with /bin/ls) in the order we expect
+# causing intermittent test failures. 
+# Add a sleep statement so we can be sure that the second run is a later timestamp,
+# so we can reliably get it with `|head -n1` and `|tail -n1` later in this test.
+# When we start emitting the path to the run summary file that was generated, or a way to specify
+# the output file, we can remove this and look for the file directly.
+# If you find this sleep statement, try running this test 10 times in a row. If there are no
+# failures, it *should* be safe to remove.
+  $ sleep 1
   $ TURBO_RUN_SUMMARY=true ${TURBO} run build -- someargs > /dev/null # run again (expecting full turbo here)
 
 # no output, just check for 0 status code, which means the directory was created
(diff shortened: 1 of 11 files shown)


cli/internal/run/real_run.go
```diff
@@ -123,6 +123,11 @@ func RealRun(
 	// Assign tasks after execution
 	runSummary.Tasks = taskSummaries
 
+	// We walk the graph with no concurrency.
+	// Populating the cache state is parallelizable.
+	// Do this _after_ walking the graph.
+	runSummary.PopulateCacheState(turboCache)
```
mehulkar: It would be nice if we could do this, but we can't because it's after executing the task, so the cache is already populated and should be true. We want values from the ec.exec`

mehulkar:
This test seems to be passing non-deterministically locally. Need to resolve that before merging

cli/internal/cache/cache.go
```diff
@@ -232,17 +232,24 @@ func (mplex *cacheMultiplexer) removeCache(removal *cacheRemoval) {
...
 	// Make a shallow copy of the caches, since storeUntil can call removeCache
 	mplex.mu.RLock()
 	caches := make([]Cache, len(mplex.caches))
 	copy(caches, mplex.caches)
 	mplex.mu.RUnlock()
 
+	// We need to return a composite cache status from multiple caches
+	// Initialize the empty struct so we can assign values to it. This is similar
+	// to how the Exists() method works.
+	combinedCacheState := ItemStatus{}
```
mehulkar, line 245: This is modeled after how `.Exists()` works

cli/integration_tests/basic_monorepo/run_summary/run_summary.t
```diff
@@ -7,14 +7,27 @@ Setup
...
 
+# HACK: Generated run summaries are named with a ksuid, which is a time-sorted ID. This _generally_ works
+# but we're seeing in this test that sometimes a summary file is not sorted (with /bin/ls) in the order we expect
+# causing intermittent test failures. 
+# Add a sleep statement so we can be sure that the second run is a later timestamp,
+# so we can reliably get it with `|head -n1` and `|tail -n1` later in this test.
+# When we start emitting the path to the run summary file that was generated, or a way to specify
+# the output file, we can remove this and look for the file directly.
+# If you find this sleep statement, try running this test 10 times in a row. If there are no
+# failures, it *should* be safe to remove.
```
tknickman, line 18: seems-good.gif

cli/internal/runcache/runcache.go
```diff
@@ -124,19 +125,26 @@ func (tc *TaskCache) RestoreOutputs(ctx context.Context, prefixedUI *cli.Prefixe
...
 		if err != nil {
-			return false, err
+			// If there was an error fetching from cache, we'll say there was no cache hit
+			return cache.ItemStatus{Local: false, Remote: false}, err
 		} else if !hit {
 			if tc.taskOutputMode != util.NoTaskOutput && tc.taskOutputMode != util.ErrorTaskOutput {
 				prefixedUI.Output(fmt.Sprintf("cache miss, executing %s", ui.Dim(tc.hash)))
 			}
-			return false, nil
+			// If there was no hit, we can also say there was no hit
```
tknickman, line 146: confusing comment

cli/internal/runcache/runcache.go
```diff
@@ -162,8 +170,7 @@ func (tc *TaskCache) RestoreOutputs(ctx context.Context, prefixedUI *cli.Prefixe
 	default:
 		// NoLogs, do not output anything
 	}
-
-	return true, nil
+	return cacheStatus, nil
```
tknickman, line 173: Presumably either local or remote is true at this point?

tknickman (approved):

cli/internal/runcache/runcache.go
mehulkar: it's a tautology tom!

cli/internal/runcache/runcache.go
mehulkar: yeah it shuold just be one of them. I'm not loving this cacheStatus struct all around, but I didn't want to change that here, just passing through the information we have

PR #4226, opened 2023-03-16
codespaces need permissions to next.js

sokra:
### Description

The github token needs permissions to push to next.js repo to use codespaces with nextpack


ForsakenHarmony (approved):

sokra:
@jridgewell we also need that for the separate repo

Issue #937, opened 2022-03-24
Scripts are run unnecessarily depending on how turbo is configured


dcherman:
### What version of Turborepo are you using?

1.1.9

### What package manager are you using / does the bug impact?

Yarn v1

### What operating system are you using?

Mac

### Describe the Bug

When running a command like `turbo run test`, I would expect it to build the graph of tasks to run based on packages that actually define a `test` task.  Instead, it seems like it assumes that all packages contain that task which results in needlessly running all tasks that would be dependencies.

In the case of the reproducer below, only the `bar` package contains a `test` script with a dependency on `foo`.  The `build` task in baz was executed despite the output of it not being required for executing the requested tasks.

### Expected Behavior

Only packages that actually contain the tasks to run should be considered when building the task graph.

### To Reproduce

1. Clone https://github.com/dcherman/turborepo-bad-script-reproducer
2. `yarn install`
3. `yarn reproduce`

Since only `bar` has a `test` script, I expected to see the `build` task from `foo` and `bar` run followed by the `test` task from `bar`.  The `build` task of `baz` runs despite it having no `test` task and it not being a dependency of either of the other packages.

Here's an image of the graph that's generated from running `turbo run test --graph`

![image](https://user-images.githubusercontent.com/976608/160013487-6798132d-b8a6-47f3-bd97-17696a6a4a6b.png)

Note that the both the `foo` and `baz` tasks contain a node for the `test` task despite not actually defining a `test` task in package.json.

gsoltis:
I can see the issue here, and I think I'm maybe 99% convinced the behavior should change. I am a little worried about anyone using "synthetic" tasks to trigger upstream packages to do something, but I think we have other ways of handling that (`--include-dependencies` / `--filter=<pkg>...` if it is desired behavior.

jaredpalmer:
See https://turborepo.org/docs/features/filtering and the new `--filter` flag in `turbo@1.2` which should correct this. Sorry about that!

michaelkplai:
I think this discussion may also be related #1106.

I also recently ran into this issue. The section of the documentation that mentions that missing tasks are gracefully ignored led me to assume missing task dependencies would also be ignored. See https://turborepo.org/docs/core-concepts/pipelines#tasks-that-are-in-the-pipeline-but-not-in-some-packagejson.

Since ignoring missing task dependencies would be a breaking change it may make sense to introduce this as a new cli flag or config option rather than overwrite the default behavior.

[...]

PR #4238, opened 2023-03-17
Filter out summaries of tasks that didn't execute

mehulkar:
Related conversation about why we don't just filter tasks that don't have scripts out of the graph entirely: https://github.com/vercel/turbo/issues/937

Doing this only for task summaries, because in Dry Runs, we explicitly include these: https://github.com/vercel/turbo/pull/1225

### DIFF
--- a/cli/internal/run/real_run.go
+++ b/cli/internal/run/real_run.go
@@ -106,9 +106,11 @@ func RealRun(
 		if taskExecutionSummary != nil {
 			taskSummary.ExpandedOutputs = taskHashTracker.GetExpandedOutputs(taskSummary.TaskID)
 			taskSummary.Execution = taskExecutionSummary
+
+			// lock since multiple things to be appending to this array at the same time
 			mu.Lock()
 			taskSummaries = append(taskSummaries, taskSummary)
-			// don't hold the lock while we run ec.exec
+			// not using defer, just release the lock
 			mu.Unlock()
 		}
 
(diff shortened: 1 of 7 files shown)


cli/internal/run/real_run.go
```diff
@@ -94,18 +94,24 @@ func RealRun(
...
+		// taskExecutionSummary will be nil if the task never executed
+		// (i.e. if the workspace didn't implement the script corresponding to the task)
+		// We don't need to collect any of the outputs or execution if the task didn't execute.
+		if taskExecutionSummary != nil {
+			taskSummary.ExpandedOutputs = taskHashTracker.GetExpandedOutputs(taskSummary.TaskID)
+			taskSummary.Execution = taskExecutionSummary
+			mu.Lock()
+			taskSummaries = append(taskSummaries, taskSummary)
+			// don't hold the lock while we run ec.exec
+			mu.Unlock()
```
mehulkar, line 112: I'm not sure if this comment applies anymore. Would the `defer mu.Unlock()` be appropriate if I append after `ec.exec`?  Or is it still not safe? cc @gsoltis

gsoltis: As structured now, the comment is no longer relevant, since a `defer my.Unlock()` immediately after `mu.Lock()` would not hold a lock while `ec.exec` is running. It's a judgement call as to whether to switch to a `defer` here. On the one hand, it is currently trivial to see that it wouldn't hold the lock for anything critical, on the other hand, `defer` is function-scoped, not block-scoped, so it's slightly weird to have it in an `if` block.

Personally I probably wouldn't switch to defer here due to the if block, but it's no longer a huge deal since the whole thing is after `ec.exec`.

mehulkar: done. I was a little unsure how `defer` would work in an anonymous function, but it works as expected https://play.golang.com/p/aEodOfXGPtV

gsoltis (approved):

arlyon (approved):
I have wondered the same thing, this is cool.

PR #4239, opened 2023-03-17
Default to rustls-tls when building turborepo

mehulkar:


arlyon (approved):
Yep, that looks about right. Approved :)

.github/workflows/turborepo-release-step-3.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         run: ${{ matrix.settings.setup }}
 
       - name: Build
-        run: ${{ matrix.settings.rust-build-env }} cargo build --release -p turbo --no-default-features --features rustls-tls --target ${{ matrix.settings.target }}
```
NicholasLYang, line 75: We need to remove the `--no-default-features` flag too

PR #4240, opened 2023-03-17
Make Single Package repo work with globalDependencies config

### LINKED ISSUE #4140: [turborepo] Configuring globalDependencies in a non-monorepo

1.8.3

Yarn v1

Mac

### Describe the Bug

I tried to add turborepo to an existing non-monorepo next project.
The `turbo.json` configuration is as follows:
```json
{
  "globalDependencies": ["**/.env.*local"],
  "pipeline": {
    "build": {
      "outputs": [".next/**"]
    },
    "lint": {},
    "dev": {
      "cache": false,
      "persistent": true
    }
  }
}
```

But when I run
```bash
yarn turbo build
```
I get the following error:

> run failed: failed to collect global hash inputs: package.json: no workspaces found. Turborepo requires Yarn workspaces to be defined in the root package.json
> Turbo error: failed to collect global hash inputs: package.json: no workspaces found. Turborepo requires Yarn workspaces to be defined in the root package.json
> error Command failed with exit code 1.

If I remove
```json
"globalDependencies": ["**/.env.*local"],
```
from the configuration, seems to work instead.

I tried to do the same with the official example of [non-monorepo](https://github.com/vercel/turbo/tree/main/examples/non-monorepo) and the same error occurs.

### Expected Behavior

Being able to use turborepo in a 'non-monorepo' by having the configuration option `globalDependencies` working.

### To Reproduce

* Install [Turborepo non-monorepo starter](https://github.com/vercel/turbo/tree/main/examples/non-monorepo) :

```bash
npx degit vercel/turbo/examples/non-monorepo non-monorepo
cd non-monorepo
yarn
git init . && git add . && git commit -m "Init"
```

* Edit the file `turbo.json` by adding the option `"globalDependencies": ["**/.env.*local"]`.

* Run:
```bash
yarn turbo build
```

### Reproduction Repo

_No response_

mehulkar:
Fixes #4140

Adds a failing test at least. The issue is that when calculating the global hash, we get the ignored files:

https://github.com/vercel/turbo/blob/f8438373f59a60c37f292f6dff25d23635a5e222/cli/internal/run/global_hash.go#L88

and in the `yarn` implementation, that involves reading the workspace globs: https://github.com/vercel/turbo/blob/f8438373f59a60c37f292f6dff25d23635a5e222/cli/internal/packagemanager/yarn.go#L43


cli/integration_tests/single_package/setup.sh
```diff
@@ -4,3 +4,10 @@ SCRIPT_DIR=$(dirname ${BASH_SOURCE[0]})
 TARGET_DIR=$1
 cp -a ${SCRIPT_DIR}/my-pkg/. ${TARGET_DIR}/
 ${SCRIPT_DIR}/../setup_git.sh ${TARGET_DIR}
+
+if [ "$2" != "" ]; then
+  pm="$2"
+  jq --arg pm "$pm" '.packageManager = $pm' "$TARGET_DIR/package.json"
+  # cat "$TARGET_DIR/package.json" | jq ".packageManager=\"$2\"" | sponge "$TARGET_DIR/package.json"
```
chris-olszewski, line 11: Should the `cat` command be uncommented? Reading through `jq` docs, I don't think that command will update the `package.json`.

chris-olszewski (approved):

cli/integration_tests/single_package/setup.sh
mehulkar: `cat` wasn't working either becuase `sponge` apparently isn't installed in CI (I thought it was). So I was experimenting with using `jq` directly, and haven't been successful yet.

mehulkar: finagled it a slightly different way

mehulkar:
Thanks for the contribution @stefanofa!

PR #4242, opened 2023-03-17
feat(examples): add vercel template to meta

tknickman:
### Description

- Fix the copy inside the app to match the directory
- Add the template to meta


examples/with-gatsby/packages/ui/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  root: true,
+  extends: ["custom/react"],
+  rules: {
+    "@typescript-eslint/no-unused-vars": "off",
```
tknickman, line 5: Required because tsup needs react to be in imported, but it's never technically uses

mehulkar: > tsup needs react to be in imported,

wut

mehulkar (approved):

PR #4243, opened 2023-03-17
refactor: Removed includeUntracked because it's always set to true

NicholasLYang:
### Description

I realized that we only call `ChangedFiles` in one place and in that place `includeUntracked` is always set to true. So I just removed it as a parameter.


arlyon (approved):
There is an outstanding failing test around an impl of scm that is now invalid but as long as that is satisfied I think that this is a good change. I assume that the scm interface doesn't have implementations that do care about the distinction?

NicholasLYang:
Nope in fact we should probably remove the entire scm abstraction as there's no real intention to support more than git

PR #4247, opened 2023-03-17
refactor: Rename Rust turbopath types

NicholasLYang:
### Description

Decided to rename the various path types to better align with the Go implementation. It's not exactly 1-1 but it should make it easier to port.

### DIFF
--- a/crates/turborepo-paths/src/absolute_system_path.rs
+++ b/crates/turborepo-paths/src/absolute_system_path.rs
@@ -18,7 +18,7 @@ use std::{
 use thiserror::Error;
 
 #[derive(Error, Debug)]
-enum AbsPathError {
+enum AbsolutePathError {
     #[error("expected an absolute path but got a relative path instead: `{0}`")]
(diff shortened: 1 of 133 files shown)


crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -18,7 +18,7 @@ use std::{
 use thiserror::Error;
```
ForsakenHarmony, line 18: this crate is "owned" by `web-tooling`, should probably change that 🙈

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -18,7 +18,7 @@ use std::{
...
 
 #[derive(Error, Debug)]
-enum AbsPathError {
+enum AbsolutePathError {
```
mehulkar, line 21: should this also be `AbsoluteSystemPathError` ?

mehulkar (approved):

PR #4249, opened 2023-03-17
restore dynamic directory embedding

ForsakenHarmony:


jridgewell (approved):
If you can get the flags working in Next, I'd love if we could just flip the flag so that it HMRs by default and statically embeds by opt-in feature flag.

ForsakenHarmony:
@jridgewell works locally in nextpack, will see if I can enable it by default

jridgewell:
No, I mean instead of having `dynamic_embed_contents`, we should change it to `static_embed_contents`. We wouldn't need a `default` feature that way, and we could still pass `--features static_embed_contents` for the test builds

sokra:
> No, I mean instead of having `dynamic_embed_contents`, we should change it to `static_embed_contents`. We wouldn't need a `default` feature that way, and we could still pass `--features static_embed_contents` for the test builds

the release builds would need to static embed

PR #4250, opened 2023-03-17
Send run summary to API

mehulkar:
This PR adds an `--experimental-space-id` flag. When this is provided, along with the `TURBO_RUN_SUMMARY` env var set, the Run Summary will be posted to the Vercel API in 3 chunks:

1. Creating the Run
2. Posting a Task Summary for each task that was executed
3. Updating the Run to "completed"

## Review Notes
- At the moment, the Run is created at the _end_, this is for simplicity, since the `startAt` timestamp is sent, and we don't need to stream updates, so it's ok to create the Run at the very end. As we add more features (e.g. streaming updates or capturing errors), we will move this to the beginning of the run. I'm defering this work so we can get the barebones out.

## Task list 
- [x] Post to endpoint
- [x] Post start/end and each task individually
- [x] `--experimental` flag
- [x] Suppress flag from help output
- [ ] ~~Suppress flag from Usage message (shown when invalid flags are sent)~~
- [x] Use `run.id` from first response for task POSTs
- [x] Make it async / non blocking
- [ ] ~~Read from config file where `turbo link` stores team id~~
- [ ] https://github.com/vercel/turbo/pull/4309 (separate PR to make sure tests are solid )
- [x] Prune down Run Create `POST` request so it doesn't include Tasks.
- [x] Prevent POST if not linked


cli/internal/runsummary/run_summary.go
```diff
@@ -91,67 +140,88 @@ func (summary *RunSummary) Save(dir turbopath.AbsoluteSystemPath, singlePackage
...
+			if err := json.Unmarshal(resp, vercelRun); err != nil {
+				errs = append(errs, err)
+			} else {
+				runID = vercelRun.ID
+			}
+		}
+	}
+
+	if runID != "" {
+		rsm.postTaskSummaries(runID)
```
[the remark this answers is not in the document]
mehulkar: yeah I wanted to do that as well, but since we're generally defering the "streaming updates" feature, I figured we'd wait to do all that together

cli/internal/client/client.go
```diff
@@ -211,3 +213,97 @@ func (c *APIClient) addTeamParam(params *url.Values) {
 		params.Add("slug", c.teamSlug)
 	}
 }
+
+// JSONPatch sends a byte array (json.marshalled payload) to a given endpoint with PATCH
```
arlyon, line 217: Maybe worth renaming this, as I was expecting https://jsonpatch.com/

cli/internal/client/client.go
```diff
@@ -211,3 +213,97 @@ func (c *APIClient) addTeamParam(params *url.Values) {
...
+	// Set headers
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", c.userAgent())
+
+	if allowAuth {
+		req.Header.Set("Authorization", "Bearer "+c.token)
+	}
+
+	if ci.IsCi() {
+		req.Header.Set("x-artifact-client-ci", ci.Constant())
```
arlyon, line 295: I know headers are case-insensitive but we should go all-camel or no-camel 🐫

cli/internal/client/client.go
```diff
@@ -211,3 +213,97 @@ func (c *APIClient) addTeamParam(params *url.Values) {
...
+	}
+
+	resp, err := c.HTTPClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// If there isn't a response, something else probably went wrong
+	if resp == nil {
```
arlyon, line 304: this seems ok, we just need to make sure we don't end up getting 204 No Content back from a request

cli/internal/runsummary/run_summary.go
```diff
@@ -91,67 +140,86 @@ func (summary *RunSummary) Save(dir turbopath.AbsoluteSystemPath, singlePackage
...
-	ExternalDepsHash       string                                `json:"hashOfExternalDependencies"`
-}
+// record sends the summary to the API
+// TODO: make this work for single package tasks
+func (rsm *Meta) record() []error {
+	errs := []error{}
+
+	// Right now we'll send the POST to create the Run and the subsequent task payloads
+	// when everything after all execution is done, but in the future, this first POST request
+	// can happen when the Run actually starts, so we can send updates to Vercel as the tasks progress.
```
arlyon, line 150: For context, here is the docs for bazel's build event protocol which streams updates (https://bazel.build/remote/bep)

cli/internal/runsummary/run_summary.go
```diff
@@ -91,67 +140,86 @@ func (summary *RunSummary) Save(dir turbopath.AbsoluteSystemPath, singlePackage
...
-	for i, depencency := range ht.Dependencies {
-		dependencies[i] = util.StripPackageName(depencency)
+func (rsm *Meta) postTaskSummaries(runID string) {
+	// We make at most 8 requests at a time.
+	maxParallelRequests := 8
+	taskSummaries := rsm.RunSummary.Tasks
+	taskCount := len(taskSummaries)
+	taskURL := fmt.Sprintf(tasksEndpoint, rsm.spaceID, runID)
+
+	parallelRequestCount := maxParallelRequests
```
arlyon, line 191: We could maybe simplify this slightly by removing this var, at the cost of running up to 7 noop goroutines... up to you

cli/internal/runsummary/vercel.go
```diff
@@ -0,0 +1,45 @@
+package runsummary
+
+type vercelRunResponse struct {
+	ID string
+}
+
+type vercelRunPayload struct {
```
arlyon, line 7: Where does this come from? Are we hand-writing these datastructures?

crates/turborepo-lib/src/cli.rs
```diff
@@ -369,6 +369,10 @@ pub struct RunArgs {
     pub tasks: Vec<String>,
     #[clap(last = true, hide = true)]
     pub pass_through_args: Vec<String>,
+
+    // Pass a string to enable posting Run Summaries to Vercel
+    #[clap(long, hide = true)]
```
arlyon, line 374: shame about it not being hidden in usage, may be worth opening an issue since I can't see one for this exact problem

arlyon (commented):
Some comments. Excited to try this out!

cli/internal/client/client.go
mehulkar: do you have a better name? I considered just `Post` and `Patch`, but since there are other methods here that use, `application/octect-stream` for example, I wanted to be clear that I'm not attempting to co-opt.

cli/internal/client/client.go
mehulkar: Agreed, I think there's more work to be done in the ApiClient, but since this is the existing code, I'll keep it as-is

cli/internal/client/client.go
mehulkar: Yeah, Ideally we should just be checking for 2xx, not specifically 200/201 , but also going to defer this since I've just used existing code and I don't want to affect what's going on

cli/internal/runsummary/run_summary.go
mehulkar: This is pretty much a copy paste of @nathanhammond's implementation of parallelizing cache checks: https://github.com/vercel/turbo/blob/674bd4f5456ab030d5feb74301f6a446770bcc5b/cli/internal/run/dry_run.go#L112. I tend to not want to make small changes like this so if there's an opportunity to consolidate, we don't second-guess ourselves about why those changes exist!

cli/internal/runsummary/vercel.go
mehulkar: yeah pretty much... I can ping you a doc in slack

crates/turborepo-lib/src/cli.rs
mehulkar: My unfamiliarity with Rust/clap confused me about where to file this issue, and I was afraid I'd be asked for reproductions that I wouldn't know how to create 😅 . Agreed we should file something though

tknickman (approved):

mehulkar:
The schema here is not complete, but I'm merging this and will continue filling in the data in future PRs.

PR #4251, opened 2023-03-17
Implement node eval streaming

jridgewell:
### Description

This implements streaming Node evaluation using an fun generic `Stream<T>`, usable within any Vc (I'm kinda surprised it works). Think of it a bit like a `DuplexStream`, it implements both a reader (that can be streaming read with the `Stream` trait) and a writer (that can be sent cross threads). As new values are written to the stream, any readers get woken up to receive the next value.

Included as part of this is a rework of our "bytes" implementations, extracting some logic into a new `turbo-tasks-bytes` crate. Eventually, `Rope` should also move out of `turbo-tasks-fs` into this crate. `BytesValue` are cheaply clonable (they're just a wrapper around `bytes::Bytes`), making them a perfect low-level type to hold the data being sent between processes and threads. They're also easy to convert to and from a real `Bytes`, so that they can be streamed into our `hyper::Server`.

### Testing Instructions

Paired next.js PR: https://github.com/vercel/next.js/pull/47264

Fixes WEB-243


crates/turbo-tasks-bytes/src/stream.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// data.
+    OpenStream {
+        source: Box<dyn StreamTrait<Item = T> + Send + Sync + Unpin>,
+        pulled: Vec<T>,
+    },
+
+    /// A Closed stream state cannot be pushed to, so it's anyone polling can
+    /// read all values at their leisure.
+    Closed(Box<[T]>),
+}
```
sokra, line 36: * It's a little bit weird that the pulled data is either in `OpenStream.pulled` or in `Closed.0`. I would hoist that into a shared field.
* The current design doesn't allow to free memory of the stream, even if all reader have read a part of the stream. That doesn't allow to use the stream with infinite stream. That's currently not needed, but would be a cool feature to have.

So I would do it this way

```suggestion
/// The StreamState actually holds the data of a Stream.
pub struct StreamSegment<T> {
    /// Data in the current segment. More can be pushed into it as it's pulled from the source.
    /// When this reaches a limit, it will create a new segment into `next` rather than pushing more
    /// items to the buffer. This allows to deallocate previous segements once all reader have consumed them.
    buffer: Vec<T>,

    /// What follows after this segment.
    /// Might be `None` when the source is ended/closed.
    next: Option<StreamSegmentNext<T>>,
}

enum StreamSegmentNext<T> {
    /// A stream is tied directly to a source stream, and will lazily pull
    /// new values out as a reader reaches the end of our already-pulled
    /// data.
    Source(Box<dyn StreamTrait<Item = T> + Send + Sync + Unpin>),
    /// Another segment is following. This forms a linked list of segments to
    /// allow deallocating memory of previous segments.
    Segment(Arc<Mutex<StreamSegment<T>>>
}
```

Should streams have a notation of "errored"? If so we might want to add `StreamSegementNext::Error(SharedError)`?

crates/turbo-tasks-bytes/src/stream.rs
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+}
+
+/// Implements [StreamTrait] over our Stream.
+#[derive(Debug)]
+pub struct StreamRead<T> {
+    index: usize,
+    source: Stream<T>,
+}
```
sokra, line 155: I think this isn't necessary and can be merged into the `Stream` struct, by adding the `index` field to this struct.

crates/turbo-tasks-bytes/src/stream.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// Streams allow for streaming values from source to sink.
+///
+/// A Stream implements both a reader (which implements the Stream trait), and a
+/// writer (which can be sent to another thread). As new values are written, any
+/// pending readers will be woken up to receive the new value.
+#[derive(Debug)]
+pub struct Stream<T> {
+    inner: Arc<Mutex<StreamState<T>>>,
+}
```
sokra, line 21:
```suggestion
pub struct Stream<T> {
    inner: Arc<Mutex<StreamSegment<T>>>,
    /// Number of items consumed in the current segment.
    /// Might revert back to 0 when reached the end of a segment
    consumed: usize,
}
```

crates/turbo-tasks-bytes/src/stream.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// Implements [StreamTrait] over our Stream.
+#[derive(Debug)]
+pub struct StreamRead<T> {
+    index: usize,
+    source: Stream<T>,
+}
+
+impl<T: Clone> StreamTrait for StreamRead<T> {
```
sokra, line 157:
```suggestion
impl<T: Clone> StreamTrait for Stream<T> {
```

crates/turbo-tasks-bytes/src/stream.rs
```diff
@@ -0,0 +1,219 @@
...
+use futures::{Stream as StreamTrait, StreamExt};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
+
+/// Streams allow for streaming values from source to sink.
+///
+/// A Stream implements both a reader (which implements the Stream trait), and a
+/// writer (which can be sent to another thread). As new values are written, any
+/// pending readers will be woken up to receive the new value.
+#[derive(Debug)]
```
sokra, line 18:
```suggestion
#[derive(Debug, Clone)]
```

crates/turbo-tasks-bytes/src/stream.rs
```diff
@@ -0,0 +1,219 @@
...
+impl<T> Default for Stream<T> {
+    fn default() -> Self {
+        Self {
+            inner: Arc::new(Mutex::new(StreamState::Closed(Box::new([])))),
+        }
+    }
+}
+
+impl<T: PartialEq> PartialEq for Stream<T> {
+    // A Stream is equal if its the same internal pointer, or both streams are
```
alexkirsz, line 104:
```suggestion
    // A Stream is equal if it's the same internal pointer, or both streams are
```

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -34,9 +37,28 @@ pub struct ProxyResult {
     /// Headers arranged as contiguous (name, value) pairs.
     pub headers: Vec<(String, String)>,
     /// The body to return.
-    pub body: Rope,
+    pub body: Body,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct BodyError {
```
alexkirsz, line 44: Could use `thiserror`.

crates/turbopack-node/src/render/render_proxy.rs
```diff
@@ -75,13 +76,11 @@ async fn run_proxy_operation(
         .send(RenderProxyOutgoingMessage::Headers { data: &data })
         .await?;
 
-    let body = body.await?;
+    let mut body = body.await?.read();
```
sokra, line 79: When `StreamRead` is merged into `Stream`, you would `fork`/`clone` the cached stream to consume it, while leaving the cached stream untouched.

```suggestion
    let mut body = body.await?.clone();
```

crates/turbopack-node/src/evaluate.rs
```diff
@@ -29,17 +32,67 @@ use crate::{
...
+    DirDependency {
+        path: String,
+        glob: String,
+    },
+    EmittedError {
+        severity: IssueSeverity,
+        error: StructuredError,
+    },
+    Value {
+        // TODO: There is like 3 levels of JSON string encoding that goes into returning a value,
```
alexkirsz, line 64: Create an issue to track this and refer to it here as TODO(WEB-XXX)

crates/turbopack-node/src/evaluate.rs
```diff
@@ -29,17 +32,67 @@ use crate::{
...
-    // TODO, support stream in the future
-    Stream(#[turbo_tasks(trace_ignore)] Vec<u8>),
+pub enum JavaScriptEvaluation {
+    /// An Empty is only possible if the evaluation completed successfully, but
+    /// didn't send an error nor value.
+    Empty,
+    /// A Single is only possible if the evaluation returns either an error or
+    /// value, and never sends an intermediate value.
+    Single(EvaluationItem),
+    /// A Stream represents a series of intermediate values and followed by
```
alexkirsz, line 93:
```suggestion
    /// A Stream represents a series of intermediate values followed by
```

crates/turbopack-node/src/evaluate.rs
```diff
@@ -229,13 +285,76 @@ pub async fn evaluate(
     .await
     .map_err(|(e, _)| e)?;
 
+    // If we reach an End or Error value immediately, then we can return an easy
+    // Single response. If not, we need to stream multiple responses out.
+    let output = pull_operation(&mut operation, cwd, context_ident_for_issue).await?;
+    let data = match output {
+        LoopResult::End(data) => {
+            if kill {
+                operation.wait_or_kill().await?;
```
alexkirsz, line 294: Do we need to wait for the operation to be finished before returning here? If not, you can spawn this in a different task.
```suggestion
                tokio::spawn(async move {
                    let _ = operation.wait_or_kill().await;
                });
```

crates/turbopack-node/src/evaluate.rs
```diff
@@ -229,13 +285,76 @@ pub async fn evaluate(
...
+                    break;
+                }
+                LoopResult::Value(data) => {
+                    sender.send(Ok(data.into()))?;
+                }
+            }
+        }
+
+        if kill {
+            operation.wait_or_kill().await?;
```
alexkirsz, line 335: Same comment as above.

crates/turbopack-node/src/evaluate.rs
```diff
@@ -29,17 +32,67 @@ use crate::{
...
+    /// An Empty is only possible if the evaluation completed successfully, but
+    /// didn't send an error nor value.
+    Empty,
+    /// A Single is only possible if the evaluation returns either an error or
+    /// value, and never sends an intermediate value.
+    Single(EvaluationItem),
+    /// A Stream represents a series of intermediate values and followed by
+    /// either an error or a ending value. A stream is never empty.
+    Stream(#[turbo_tasks(trace_ignore, debug_ignore)] JavaScriptStream),
 }
```
sokra, line 96: I think it would be easier to consume when it is always an stream.

```suggestion
pub enum JavaScriptEvaluation(#[turbo_tasks(trace_ignore, debug_ignore)] JavaScriptStream)
```

Maybe add `StreamExt::into_single_value()` which gets the only value from stream and throws when it's empty, more than one value or the stream errored.

crates/turbopack-node/src/evaluate.rs
```diff
@@ -29,17 +32,67 @@ use crate::{
...
+        data: String,
+    },
+    End {
+        data: Option<String>,
+    },
+    Error(StructuredError),
+}
+
+#[derive(Debug)]
+enum LoopResult {
```
alexkirsz, line 75: This is similar to the [ControlFlow](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html) enum, maybe its name could reflect that? e.g. `OperationControlFlow`.

sokra (changes requested):

alexkirsz (approved):

crates/turbo-tasks-bytes/src/stream.rs
jridgewell: > The current design doesn't allow to free memory of the stream, even if all reader have read a part of the stream. That doesn't allow to use the stream with infinite stream. That's currently not needed, but would be a cool feature to have.

That would mean the Vc which holds a stream could never be reused, and any tasks which depend on it need to rerun the node evaluation fully.

I can make join the open/closed stream into the same struct by wrapping the `source` in an `Option<dyn StreamTrait>`.

> Should streams have a notation of "errored"? If so we might want to add StreamSegementNext::Error(SharedError)?

`Stream` is generic, the wrappers can (and do) handle error state with `Result<T>` inside the stream.

crates/turbopack-node/src/render/render_proxy.rs
sokra: > The index can't be used on the original value, because it would only screw up all readers if it were ever read from originally.

Rust mutability prevents reading from the cached value since you don't have a mutable reference to it.

Also readers won't references the original stream. They have a cloned stream with a shared `StreamState`. StreamState is still immutable in that sense since we never remove items from `pulled`.

With

``` rs
pub struct Stream<T> {
	    inner: Arc<Mutex<StreamState<T>>>,
	    /// Number of items consumed in the current segment.
	    /// Might revert back to 0 when reached the end of a segment
	    consumed: usize,
}
```

you would need a `&mut Stream` to read from it, which is only possible if you are the only owner.

crates/turbopack-node/src/evaluate.rs
sokra: hmm?

``` diff
- let JavaScriptEvaluation::Single(Ok(val)) = &*config_value else {
+ let Some(val) = config_value.into_single().await? else {
```

sokra (changes requested):

crates/turbopack-node/src/evaluate.rs
```diff
@@ -229,13 +280,76 @@ pub async fn evaluate(
...
+            );
+            return Ok(data.cell());
+        }
+    };
+
+    // The evaluation sent an initial intermediate value without completing. We'll
+    // need to spawn a new thread to continually pull data out of the process,
+    // and ferry that along.
+    let (sender, stream) = JavaScriptStream::new_open(vec![Ok(data.into())]);
+    tokio::spawn(async move {
```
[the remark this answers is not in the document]
sokra: https://tokio.rs/tokio/tutorial/streams#async-stream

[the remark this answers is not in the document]
jridgewell: Done.

crates/turbopack-node/src/render/render_proxy.rs
jridgewell: > Rust mutability prevents reading from the cached value since you don't have a mutable reference to it.

What I mean is imagine you clone from an already mutated stream:

```rust
let mut s = Stream(…);

// This advances `s.index`
s.next.await?;

// Inherits the advanced index
s.clone()
```

So by mutating the index, any clones that want to get access to the full stream would be broken. I don't see why we want to store this on the struct itself anyways, we don't store iteration index as part of a `Vec<T>`. If you want to read and iterate, you alloc your own `Iter` to hold your position.

crates/turbopack-node/src/evaluate.rs
jridgewell: Currently the configs handle errors by defaulting, moving the unwrap into the `?` would prevent that.

jridgewell: I figured out a return value that makes this palatable.

One side effect of this is that any code that handles either a single value or a stream of values must ensure they fully close the stream so the `NodeJsOperation` is released back to the pool. Eg, the `route` method can take receive either a single result for `Rewrite`, or a stream for `Middleware` responses. I can't call `into_single` for this, because I need the individual results for the middleware. For the rewrite response, I just called the `read.next().await` a single time. But that left the stream midway, and the process was never released to the pool. I had to add a bit of code to call into the stream one more time.

crates/turbo-tasks-bytes/src/stream.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl<T: Clone> Default for Stream<T> {
+    fn default() -> Self {
+        Self::new_closed(vec![])
+    }
+}
+
+impl<T: Clone + PartialEq> PartialEq for Stream<T> {
+    // A Stream is equal if it's the same internal pointer, or both streams are
+    // closed with equivalent values.
```
sokra, line 100: This has an interesting behavior where a === b can change after the streams have been read to the end.

sokra: Also a === b might change after serialization...

sokra: Do we need Eq?

jridgewell: > This has an interesting behavior where a === b can change after the streams have been read to the end.

Yes, but it can only go from false -> true.

> Also a === b might change after serialization...

Not quite, it already needs to be closed before serialization, so it's the same "has the source been fully read" case.

> Do we need Eq?

Isn't that a requirement for use in a Vc?

sokra: > Yes, but it can only go from false -> true.

hmm... technically this would be fine for Vcs, but it still feels a bit weird...

> Isn't that a requirement for use in a Vc?

You could opt-out of the eq check on celling...

sokra (approved):
Sounds good...
If you want to merge it make sure the next.js PR passes CI, merge this one, release a tag, update the next.js with the tag, and merge that.

PR #4261, opened 2023-03-20
docs: added backend tab for internal pkgs

rhenriquez28:
### Description
I had a very difficult time trying to integrate Nest.js apps and internal packages in Turborepo. This is because I didn't understand that in order to use the internal package I needed to bundle it with my app and tell Webpack to allow my internal package as a `node_module` that needed to be bundled. I eventually figured it out by reading the Next.js tabs of the Internal Packages docs and connecting the dots.

However, I feel like the Internal Packages docs only put emphasis on the frontend experience and doesn't mention anything about how people could use internal packages if they had a backend app.

This PR adds a "Webpack (Backend)" tab to the "Try running the app" and "Configuring your app" which lets the user know what they need to do in order to use an internal package in a backend app.

### DIFF
--- a/docs/pages/repo/docs/handbook/sharing-code/internal-packages.mdx
+++ b/docs/pages/repo/docs/handbook/sharing-code/internal-packages.mdx
@@ -191,7 +191,7 @@ The fix is simple - we need to tell Next.js to bundle the files from certain pac
   <Tab>
 If you want to use `math-helpers` with a backend app, you'll need a bundler like [Webpack](https://webpack.js.org/), [Parcel](https://parceljs.org/) or [Rollup](https://rollupjs.org/) to bundle your internal package with your app. For this example, we'll use Webpack.
 
-You'll likely want not bundle any `node_modules` or node imports like `path`. In order to exclude all these from your bundle, we can use something like [`webpack-node-externals`](https://www.npmjs.com/package/webpack-node-externals) or the equivalent in your bundler of choice.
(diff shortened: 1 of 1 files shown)


sinclairnick:
I agree. Turborepo definitely leaves a lot to be figured out. Of course taking on the complexities and annoyances of the JS/TS language are difficult, but I think for Turborepo to achieve it's goal, being able to produce backends is a very standard goal that's unfortunately brushed over in the docs.

tknickman (changes requested):
Hey @rhenriquez28, thanks for the PR! 

I think the way that this is positioned as being for "backend apps" could lead to more confusion, as it depends heavily on your particular setup, node version, and if you're using esm or cjs etc. 

I would recommend generalizing this for bundling internal apps using webpack, either for the client or the server (the instructions should be largely the same).

Also, another alternative here is to just build the package (without publishing) - then consumers can use it without needing to care about bundling it (this can be cached by turbo, which will result in faster build times as you won't need to rebuild the package unless it has changed) - which we do already have a section addressing bundling here https://turbo.build/repo/docs/handbook/publishing-packages/bundling, where we reccomend using tsup (esbuild) for simplicity.

anthonyshew:
Hey, y'all! I'm going through doing some docs maintenance and found this PR.

I really appreciate the spirit of this! I pulled it down, took a look, and had some feedback - but it was about to be almost exactly the same as Tom's feedback! 😄 There's a simpler, more generalized path to this bundling that we'll be able to recommend to folks.

I'm going to close this PR for now since it hasn't seen activity for longer than a month. If you'd like to pick it back up and incorporate some of those new thoughts, please feel free to open it back up! Thank you!

Issue #4260, opened 2023-03-20
[turbopack] rust-analyzer broken due to missing Cargo.toml in #4188


cdaringe:
### What version of Turbopack are you using?

main

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

`rust-analyzer` server cannot initialize in the project root

### Expected Behavior

`rust-analyzer` initializes in the project root.

#4188 adds an entity called `nextpack`, and adds an analyzer reference to `Cargo.toml`, but the referenced Cargo.toml is not present:
https://github.com/vercel/turbo/pull/4188/files#diff-2f2bd68e0e094d4ab6d9eeb62674d92deaa12525ed3e363d76d5afb814767a78

### To Reproduce

- open vscode in repo root
- install rust-analyzer

observe rust-analyzer extension server output denote it's boot failure

### Reproduction Repo

n/a

cdaringe:
per reading #4188, one is supposed to generate the cargo.yaml through a bit of manual & (subjectively slow) process. Should a _default_ Cargo.yaml be placed there, and nextpack dev _overriddes/updates_ it?

minimally, an patch to `CONTRIBUTING.md` is warranted

cdaringe:
blarg, the `--sync` script fails:

```
Synchronizing nextpack Cargo.toml with Turbo and Next.js…
conflicting dependency:
  next:  httpmock = { default-features = false, version = "0.6.7" }
  turbo: httpmock = "0.6.7"
conflicting dependency:
  next:  turbo-tasks-fetch = { default-features = false }
  turbo: turbo-tasks-fetch = {}
Synchronizing nextpack Cargo.lock with Turbo and Next.js…
confliction lockfile entry dependencies:
  next:  twox-hash 1.6.3 -> cfg-if = 0.1.10
  turbo: twox-hash 1.6.3 -> cfg-if = 1.0.0

### Merged with conflicts: 2 conflicts in Cargo.toml, 1 conflicts in Cargo.lock ###
```

i think it's worth doing a rethink on how this integration works, and if `turborepos` rust IDE toolchain need be coupled to an external project's cargo config

cool-little-fish:
I encounter the same problem with you, comment the configuration `rust-analyzer.linkedProjects` for vscode works for me. Current solution that develop nextpack inside current workpace seems problematic with rust-analyzer.

jridgewell:
We're going to be removing `nextpack` entirely.

PR #4266, opened 2023-03-20
Remove nextpack

jridgewell:
Nextpack had various problems:
- It failed VSCode if nextpack wasn't initialized
- The dir hierarchy didn't play well with Vim's COC
- The `.gitignore` prevented ripgrep from finding files (both grepping and fuzzy file path opening)

We're instead switching to the internal nextpack repo for development on both the next.js and turbo repos.

Closes https://github.com/vercel/turbo/issues/4260


alexkirsz (approved):
LGTM 🚀

PR #4268, opened 2023-03-20
Fix getting workspace ignores in case of non-monorepo

### LINKED ISSUE #4140: [turborepo] Configuring globalDependencies in a non-monorepo

stefanofa:
### Description

Hey @mehulkar I tried to fix #4140 .
This is my first time reading or writing in go so it's likely I may have done something wrong.

The way I did it is a bit redundant, as in the case of a monorepo the call to `getWorkspaceGlobs` is made 2 times and so is not the most efficient. Unfortunately I think I lack a lot of context to try and come up with a better solution, but I felt like trying :)

### DIFF
--- a/cli/internal/packagemanager/yarn.go
+++ b/cli/internal/packagemanager/yarn.go
@@ -12,6 +12,12 @@ import (
 	"github.com/vercel/turbo/cli/internal/turbopath"
 )
 
+type NoWorkspacesFoundError struct {}
+
+func (e *NoWorkspacesFoundError) Error() string {
+	return "package.json: no workspaces found. Turborepo requires Yarn workspaces to be defined in the root package.json"
+}
+
 var nodejsYarn = PackageManager{
 	Name:         "nodejs-yarn",
 	Slug:         "yarn",
@@ -27,7 +33,7 @@ var nodejsYarn = PackageManager{
 			return nil, fmt.Errorf("package.json: %w", err)
 		}
 		if len(pkg.Workspaces) == 0 {
-			return nil, fmt.Errorf("package.json: no workspaces found. Turborepo requires Yarn workspaces to be defined in the root package.json")
+			return nil, &NoWorkspacesFoundError{}
 		}
 		return pkg.Workspaces, nil
 	},
@@ -40,20 +46,15 @@ var nodejsYarn = PackageManager{
 		// For example: `apps/*/node_modules/**/+(package.json|yarn.json)`
 		// The `extglob` `+(package.json|yarn.json)` (from micromatch) after node_modules/** is redundant.
 
-		// In case of a non-monorepo the workspaces field is empty and getWorkspaceGlobs would fail,
-		// and should therefore be handled separately.
-		pkg, err := fs.ReadPackageJSON(rootpath.UntypedJoin("package.json"))
-		if err != nil {
-			return nil, fmt.Errorf("package.json: %w", err)
-		}
-		if len(pkg.Workspaces) == 0 {
-			ignores := make([]string, 1)
-			ignores[0] = "node_modules/**"
-			return ignores, nil
-		}
-
 		globs, err := pm.getWorkspaceGlobs(rootpath)
 		if err != nil {
+			// In case of a non-monorepo, the workspaces field is empty and only node_modules in the root should be ignored
+			_, ok := err.(*NoWorkspacesFoundError)
+			if ok {
+				ignores := make([]string, 1)
+				ignores[0] = "node_modules/**"
+				return ignores, nil
+			}
 			return nil, err
 		}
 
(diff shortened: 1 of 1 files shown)


mehulkar (commented):
Hey! thanks for taking a stab at this!

> The way I did it is a bit redundant, as in the case of a monorepo the call to `getWorkspaceGlobs` is made 2 times and so is not the most efficient. 

Just making sure, do you mean there's a duplicate call to `fs.ReadPackageJSON`?

Instead of `fs.ReadPackageJSON` twice, how about returning a custom error from `getWorkspaceGlobs` (instead of `fmt.Errorf()`), and returning the `ignores: ["node_modules/**]` in that case, but bubbling up all other errors?

Happy to give you another pointer to custom error handling if you want! 

PS, I know @tknickman also said he might have time to take a look at this, in case he had a different solution in mind.

stefanofa:
> Just making sure, do you mean there's a duplicate call to fs.ReadPackageJSON?

Yes, that's what I meant.

By the way, I tried the refactor you suggested!

cli/internal/packagemanager/yarn.go
```diff
@@ -42,6 +48,13 @@ var nodejsYarn = PackageManager{
 
 		globs, err := pm.getWorkspaceGlobs(rootpath)
 		if err != nil {
+			// In case of a non-monorepo, the workspaces field is empty and only node_modules in the root should be ignored
+			_, ok := err.(*NoWorkspacesFoundError)
+			if ok {
```
mehulkar, line 53: you can use `errors.As()` here instead of coercing the type. Example from another part of the codebase:
https://github.com/vercel/turbo/blob/6bc8b5e0bdf920e799ca693bf697acdf0796640b/cli/internal/core/engine.go#L183-L193

cli/internal/packagemanager/yarn.go
```diff
@@ -42,6 +48,13 @@ var nodejsYarn = PackageManager{
...
+				ignores := make([]string, 1)
+				ignores[0] = "node_modules/**"
+				return ignores, nil
```
mehulkar, line 56:
```suggestion
				return []string{"node_modules/**"}, nil
```

(I'm not a Go expert here, but this reads a bit cleaner to me. I'm not sure if the size allocation is implicit when declaring with literal values, or if your `make` with size is better. cc @chris-olszewski

mehulkar (commented):
Couple more questions:

- What happens in a real monorepo when the workspace config is missing?
- In `npm` and `pnpm` implementations of `getWorkspaceIgnores`, the default return value is `**/node_modules/**`. Should we use that as the default here also? Or can we trust the user that if the workspaces config is missing, then we really only want to ignore the root `./node_modules`?

chris-olszewski:
> What happens in a real monorepo when the workspace config is missing?

This should still error as we call `packageManager.GetWorkspaces` in the `BuildPackageGraph` context constructor.

>     * In `npm` and `pnpm` implementations of `getWorkspaceIgnores`, the default return value is `**/node_modules/**`. Should we use that as the default here also? Or can we trust the user that if the workspaces config is missing, then we really only want to ignore the root `./node_modules`?


I think from my groking that `node_modules/**` matches what yarn does by default: https://github.com/yarnpkg/yarn/blob/3119382885ea373d3c13d6a846de743eca8c914b/src/config.js#L816

mehulkar:
Let's merge this into my branch then!

PR #4271, opened 2023-03-20
chore(eslint): bump versions

tknickman:
### Description

Publish workspaces, and the eslint config and plugin as part of the standard publish flow.

### DIFF
--- a/packages/eslint-config-turbo/package.json
+++ b/packages/eslint-config-turbo/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eslint-config-turbo",
-  "version": "0.0.9",
+  "version": "0.0.10",
   "description": "ESLint config for Turborepo",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


packages/eslint-config-turbo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-config-turbo",
-  "version": "0.0.9",
+  "version": "0.0.10",
```
tknickman, line 3: These were published manually this AM - hopefully for the last time

packages/eslint-plugin-turbo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-turbo",
-  "version": "0.0.9",
+  "version": "0.0.10",
```
tknickman, line 3: Same

mehulkar (approved):
Do we log somewhere which packages are being published? And can we add a guide somewhere on how to add new packages so we don't have any thing that gets omitted accidentally?

tknickman:
It's in the output here, but could definitely be improved. Yea good suggestion! I'll throw a docs update up

PR #4284, opened 2023-03-21
improve source mapping of console output and errors

sokra:
### Description

* source map node.js process output
* fix race condition in console output
* decode magic identifier
* add source context
* add some coloring to the console
* improve readablility of sources in the devtools

image (3)
image (4)

fixes WEB-547
fixes WEB-431
fixes WEB-385
fixes WEB-753


crates/turbopack-node/src/source_map/mod.rs
```diff
@@ -1,7 +1,371 @@
...
+    };
+    match resolved {
+        Err(err) => {
+            // There was an error resolving the source map
+            write!(writable, "\n    at {}", original_frame)?;
+            if *first_error {
+                write!(writable, "\n    (error resolving source map: {})", err)?;
+                *first_error = false;
+            } else {
+                write!(writable, "    (error resolving source map)")?;
```
alexkirsz, line 93: No new line at the beginning here?

crates/turbopack-node/src/source_map/mod.rs
```diff
@@ -1,7 +1,371 @@
...
+    };
+    let highlight = if ansi_colors {
+        Style::new().bold().underline()
+    } else {
+        Style::new()
+    };
+    match resolved {
+        Err(err) => {
+            // There was an error resolving the source map
+            write!(writable, "\n    at {}", original_frame)?;
```
alexkirsz, line 88: Could we move the padding into a constant?

crates/turbopack-node/src/source_map/mod.rs
```diff
@@ -1,7 +1,371 @@
...
+                    writable,
+                    "\n    at {} {}",
+                    frame.with_path(&project_path.path).style(highlight),
+                    format_args!("[{}]", original_frame.with_name(None)).style(lowlight)
+                )?;
+            }
+            let (line, column) = frame.get_pos().unwrap_or((0, 0));
+            if let FileLinesContent::Lines(lines) = &*lines {
+                let lines = lines.iter().map(|l| l.content.as_str());
+                let ctx = get_source_context(lines, line - 1, column - 1, line - 1, column - 1);
```
alexkirsz, line 140: Can't this underflow if we're in the `frame.get_pos() = None` case above and we set line and column to 0?

crates/turbopack-node/src/source_map/mod.rs
```diff
@@ -1,7 +1,371 @@
...
+                } else {
+                    write!(writable, "\n{}", ctx)?;
+                }
+            }
+        }
+    }
+    Ok(())
+}
+
+fn format_source_context_lines(ctx: &SourceContextLines, f: &mut impl Write) {
```
alexkirsz, line 153: This is similar to the code in `crates/turbopack-cli-utils/src/issue.rs`. Can we share some code between the two?

My comments in `src/issue.rs` also apply here.

crates/turbopack-node/src/source_map/mod.rs
```diff
@@ -1,7 +1,371 @@
...
+    };
+    let Some(sm) = *generate_source_map.generate_source_map().await? else {
+        return Ok(ResolvedSourceMapping::NoSourceMap);
+    };
+    let trace = SourceMapTraceVc::new(sm, line, column, name.map(|s| s.to_string()))
+        .trace()
+        .await?;
+    match &*trace {
+        TraceResult::Found(frame) => {
+            if let Some(project_path) = frame.file.strip_prefix("/turbopack/[project]/") {
```
alexkirsz, line 292: The `"/turbopack/"` prefix is duplicated in three different places. Move it to a constant.

Same comment for `"[project]"`, let's make sure we can't break this code by updating either of these.

crates/turbopack-core/src/chunk/chunk_in_group.rs
```diff
@@ -77,6 +78,17 @@ impl Asset for ChunkInGroup {
     }
 }
 
+#[turbo_tasks::value_impl]
+impl GenerateSourceMap for ChunkInGroup {
+    #[turbo_tasks::function]
+    async fn generate_source_map(&self) -> Result<OptionSourceMapVc> {
+        let Some(inner) = GenerateSourceMapVc::resolve_from(self.inner).await? else {
+            return Ok(OptionSourceMapVc::cell(None))
```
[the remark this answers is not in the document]
sokra: No, we want to know if something has source maps. For an empty_map we can't differ between unmapped and no source map.

Eventually we probably want to allow the user to inspect code without source map in some way, while unmapped code is generated code that the user shouldn't see.

crates/turbopack-node/src/pool.rs
ForsakenHarmony, line 1: have you tried using the node.js sourcemap support?

crates/turbopack-test-utils/Cargo.toml
```diff
@@ -11,7 +11,7 @@ bench = false
 
 [dependencies]
 anyhow = "1.0.47"
-once_cell = "1.13.0"
+once_cell = { workspace = true }
```
ForsakenHarmony, line 14: I think everything except for `similar` could be taken from the workspace here

alexkirsz (approved):

PR #4287, opened 2023-03-21
ci(workflow): resurrect daily next.js test

kwonoj:
### Description

Part 1 for WEB-762.

This is an attempt to simple-reenablement to previous next.js integration test.

Please note the general test strategy will be still revisited: i.e there was a discussion for making next.js upstream test to turbopack-aware, and some other approaches as well. Meanwhile, this'll allow to resurrect old dashboard / slack status as quick as possible. Once we have stable ground for the whole test, this can go away then.


.github/workflows/on-nextjs-release-publish.yml
```diff
@@ -0,0 +1,71 @@
+# A workflow supposed to run when there is a new release of Next.js.
+# This relies on next.js upstream's `repository_dispatch` workflow.
+name: Next.js Release Publish
```
mehulkar, line 3:
```suggestion
name: Next.js Release
```

.github/workflows/on-nextjs-release-publish.yml
mehulkar, line 1: release-publish was confusing to me, just "release" or just "publish" for the filename and the `name` field would be good

.github/workflows/on-nextjs-release-publish.yml
```diff
@@ -0,0 +1,71 @@
...
+  upload_test_results:
+    name: Upload test results
+    # Don't upload results when this workflow is run manually via workflow_dispatch
+    if: ${{ github.event_name != 'workflow_dispatch' }}
+    needs: [next_js_integration]
+    uses: ./.github/workflows/upload-nextjs-integration-test-results.yml
+    with:
+      is_main_branch: false
+
+  post_to_slack:
```
mehulkar, line 48: Can we make this a more general action? Turborepo team could like to use it to run turbo against examples at a cadence also

.github/workflows/on-nextjs-release-publish.yml
```diff
@@ -0,0 +1,71 @@
...
+      version: ${{ inputs.version || github.event.client_payload.version }}
+      diff_base: "release"
+
+  # Upload test results to branch.
+  upload_test_results:
+    name: Upload test results
+    # Don't upload results when this workflow is run manually via workflow_dispatch
+    if: ${{ github.event_name != 'workflow_dispatch' }}
+    needs: [next_js_integration]
+    uses: ./.github/workflows/upload-nextjs-integration-test-results.yml
```
mehulkar, line 44: Could this also be a bit more generalized so we can use it in turborepo?

.github/workflows/upload-nextjs-integration-test-results.yml
```diff
@@ -0,0 +1,63 @@
+# Reusable workflow to upload next.js integration test result to specific branch `nextjs-integration-test-data`
```
mehulkar, line 1: Is using a branch for this what we want long term? Would it be useful to make a dashboad in Tinybird or Preset or some other thing we use internally to track metrics over time?

mehulkar (commented):
leaving some thoughts since I got tagged for review, but nothing blocking.

.github/workflows/on-nextjs-release-publish.yml
kwonoj: Currently this action's output format is strict to test only, so not sure how much useful it'll be. But I also want this to be more generic - I'd like to let @wbinnssmith chime in here.

.github/workflows/on-nextjs-release-publish.yml
kwonoj: Any specific usecase you have in mind? I can try to generalize, but honestly I'd like to switch this to some other storage if possible.

.github/workflows/upload-nextjs-integration-test-results.yml
kwonoj: I think https://github.com/vercel/turbo/pull/4287#discussion_r1144030161 answers, I wish to upload this to somewhere else, like honeycomb or some sort of obsevability so we can consolidate all of the data into single place with more query-able. Uploading it into this branch was immediate stopgap following turbopack's benchmark practices to unblock us immediately.

.github/workflows/on-nextjs-release-publish.yml
kwonoj: Yeah, good point.

jridgewell (approved):

PR #4288, opened 2023-03-21
Pass `context` through to `ImportMappingReplacement::result`

wbinnssmith:
This passes through the importing context to `ImportMap::result`, typically the directory of the importing file.

This will be used by an upcoming PR to Next.js that implements resolving font files relative to the importing file in `next/font/local`.


jridgewell (commented):
Did we resolve the issue with this sometimes being a dir and sometimes a file?

wbinnssmith:
> Could the font resolving be implemented as a resolve plug-in?

Yeah, let's do that. Filed as WEB-767.

> Did we resolve the issue with this sometimes being a dir and sometimes a file?

It's consistent now. The only issue is there's a bug where if the import map replacement fails for some reason, it gets retried up the file tree. But I think that's a separate issue?

jridgewell (approved):

PR #4292, opened 2023-03-21
Handle deleted files

gsoltis:
### Description

Previously, we used `fs_util::canonicalize` as a way of converting from slashes in paths. That returns an error if the underlying file doesn't exist, which in turn breaks our ability to handle deleted files.

This PR:
 - renames several paths to be more explicit
 - dedupes some path calculation
 - uses a path_slash library to implement the equivalent from Go's `FromSlash` (NOTE: this is probably not the right location for this code. I would be happy to have suggestions as to how to incorporate this into the path library cc @nathanhammond @NicholasLYang @chris-olszewski )
 - adds a test for deleted files that fails on `main`

### Testing Instructions

Added new test for deleted files.

### DIFF
--- a/crates/turborepo-scm/src/git.rs
+++ b/crates/turborepo-scm/src/git.rs
@@ -5,7 +5,11 @@ use std::{
 
 use anyhow::anyhow;
 use git2::{DiffFormat, DiffOptions, Repository};
-use turborepo_paths::{fs_util, project::ProjectRoot, project_relative_path::ProjectRelativePath};
+use path_slash::PathBufExt;
+use turborepo_paths::{
+    fs_util, project::ProjectRoot, project_relative_path::ProjectRelativePath,
+    AbsoluteNormalizedPath, AbsoluteNormalizedPathBuf,
+};
 
 use crate::Error;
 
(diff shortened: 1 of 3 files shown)


crates/turborepo-scm/src/git.rs
```diff
@@ -64,50 +68,53 @@ pub fn changed_files(
...
     repo_root: &ProjectRoot,
     file_path: &Path,
-    monorepo_root: &ProjectRelativePath,
+    turborepo_root: &AbsoluteNormalizedPath,
 ) -> Result<PathBuf, Error> {
     // We know the path is relative to the repo root so we can convert it to a
     // ProjectRelativePath
     let project_relative_file_path = ProjectRelativePath::new(file_path)?;
     // Which we then resolve to an absolute path
     let absolute_file_path = repo_root.resolve(project_relative_file_path);
```
chris-olszewski, line 77: Reading through more of `turborepo_path`, I'm surprised that we need to use `Path::from_slash`: https://github.com/vercel/turbo/blob/main/crates/turborepo-paths/src/absolute_normalized_path.rs#L131

chris-olszewski (approved):

crates/turborepo-scm/src/git.rs
```diff
@@ -64,50 +68,53 @@ pub fn changed_files(
...
     // Then we call canonicalize to get a system path instead of a Unix style path
```
NicholasLYang, line 78: Should probably delete this comment

crates/turborepo-scm/src/git.rs
```diff
@@ -64,50 +68,53 @@ pub fn changed_files(
 fn get_stripped_system_file_path(
```
NicholasLYang, line 68: The comment above needs to be updated

NicholasLYang (approved):

gsoltis:
No longer necessary, covered by other work by @NicholasLYang

PR #4299, opened 2023-03-22
turbopath.rs

nathanhammond:
This is an alternative implementation of `AbsoluteSystemPath`, intended to replace the vendored buck2 implementation.

It's not _quite_ complete (still has a few more `impl`s to write) but wanted to go ahead and get the PR up in public.


crates/turbopath/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "turbopath"
+version = "0.1.0"
+edition = "2021"
+license = "MPL-2.0"
```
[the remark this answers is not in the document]
nathanhammond: The diff from this PR (ported from everybody else) has rounds-to-zero original code and/or the implementation is entirely constrained by the API such that there was actually no choice in implementation.

MPL-2.0 is the preferred license of the Turboverse for our own work and it is arguable as to where to draw the line on this one for "owned work." So I punted: first commit will be Apache/MIT. Further work on this will begin to introduce clearly owned code, and should be licensed under MPL-2.0.

crates/turbopath/src/lib.rs
```diff
@@ -0,0 +1,696 @@
...
+
+    delegate! {
+        to self.0 {
+            pub fn as_path(&self) -> &Path;
+            pub fn capacity(&self) -> usize;
+            // INVALID pub fn clear(&mut self);
+            pub fn into_boxed_path(self) -> Box<Path>;
+            pub fn into_os_string(self) -> OsString;
+            pub fn pop(&mut self) -> bool;
+            // FIXME: this must accept only system components.
```
nathanhammond, line 64: This (and the other `FIXME`s) are blocked on `Anchored(Unix|System)Path`.

crates/turbopath/src/lib.rs
```diff
@@ -0,0 +1,696 @@
...
+// impl_cmp_os_str!(Path, OsStr);
+// impl_cmp_os_str!(Path, &'a OsStr);
+// impl_cmp_os_str!(Path, Cow<'a, OsStr>);
+// impl_cmp_os_str!(Path, OsString);
+// impl_cmp_os_str!(&'a Path, OsStr);
+// impl_cmp_os_str!(&'a Path, Cow<'b, OsStr>);
+// impl_cmp_os_str!(&'a Path, OsString);
+// impl_cmp_os_str!(Cow<'a, Path>, OsStr);
+// impl_cmp_os_str!(Cow<'a, Path>, &'b OsStr);
+// impl_cmp_os_str!(Cow<'a, Path>, OsString);
```
nathanhammond, line 696: Need to port these out of upstream.

nathanhammond (commented):
State of the world:

PR #4305, opened 2023-03-22
Add an integration test for current error behavior:

gsoltis:
### Description

No new functionality. 
 - Adds an integration test to capture our current output when a script fails

### Testing Instructions

Run new integration test


mehulkar (commented):
I think we already have this captured here? https://github.com/vercel/turbo/blob/main/cli/integration_tests/monorepo_one_script_error/run.t

gsoltis:
Oh yeah, nevermind, don't need this then.

PR #4306, opened 2023-03-22
Daemon server rust port

arlyon:
### Description

This ports the daemon server to rust, leaving only run an prune in go-land. This needs some polish but I would like to get some feedback now. I have also added a flag for running the grpc server over http instead, which is useful for debugging. It has support for reflection, so can be used via postman or grpcurl.

For code reviews, it should be ok to go in commit-order.

Outstanding parts:

- [ ] incomplete tests
- [ ] final windows support (UDS)
- [ ] a number of unwraps

 Out of scope (for now):

- startup performance pass

### Testing Instructions

This is a drop-in replacement for the go server, so `turbo run` in any repo should work identically.


crates/globwatch/examples/cancel.rs
```diff
@@ -0,0 +1,43 @@
...
+        let span = info_span!("config_fut");
+        let _ = span.enter();
+        for x in 0..5 {
+            info!(parent: &span, "iteration {}", x);
+            config
+                .include("/Users/arlyon/Programming/globwatch/src/**".to_string())
+                .await;
+            tokio::time::sleep(Duration::from_secs(1)).await;
+            config
+                .exclude("/Users/arlyon/Programming/globwatch/src/**".to_string())
```
mehulkar, line 33: We are all arlyon now!

crates/globwatch/readme.md
```diff
@@ -0,0 +1,27 @@
+# Globwatch
+
+> Watch a set of globs efficiently
+
+This library provides an async interface over notify and glob-match to
+efficiently watch a list of globs. Where possible it attempts to minimize the
```
mehulkar, line 6:
```suggestion
watch a list of globs. Where possible it attempts to minimize the
```

nit: redundant. I don't think anyone would ever claim the opposite :). And the explanation of how it's efficient is already way better than the claim!

crates/globwatch/src/lib.rs
```diff
@@ -0,0 +1,361 @@
+//! A wrapper around notify that allows for glob-based watching.
+//!
+//! ## What is flushing?
```
mehulkar, line 3: 🚽

crates/globwatch/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+            .map_err(|_| ConfigError)?;
+        rx.await.map_err(|_| ConfigError)
+    }
+}
+
+/// Gets the minimum set of paths that can be watched for a given glob,
+/// specified in minimatch glob syntax.
+///
+/// note: it is currently extremely conservative, handling only **, braces, and
+/// ?. any other case watches the entire directory.
```
mehulkar, line 275:
```suggestion
/// note: it is currently extremely conservative, handling only **, braces, and ?.
/// Any other case watches the entire directory.
```

`?` was lonely on its own line

crates/globwatch/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+    #[test_case("foo/**", vec!["foo"])]
+    #[test_case("foo/{a,b}", vec!["foo/a", "foo/b"])]
+    #[test_case("foo/*/bar", vec!["foo"])]
+    #[test_case("foo/[a-d]/bar", vec!["foo"])]
+    #[test_case("foo/a?/bar", vec!["foo"])]
+    #[test_case("foo/ab?/bar", vec!["foo/a", "foo/ab"])]
+    #[test_case("foo/ab?c?", vec!["foo/a", "foo/ab", "foo/abc", "foo/ac"])]
+    // todo: this should be ["foo/a/a", "foo/a/ab", "foo/b/a", "foo/b/ab"]
+    #[test_case("foo/{a,b}/ab?", vec!["foo/a", "foo/b"])]
+    fn test_handles_doublestar(glob: &str, paths_exp: Vec<&str>) {
```
mehulkar, line 353: Is there a use case for watching negation patterns with `!`?

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -269,7 +269,7 @@ fn enable_caching(url: &str) -> Result<()> {
     println!("Visit {} in your browser to enable Remote Caching", url);
 
     // We return an error no matter what
-    return Err(anyhow!("link after enabling caching"));
+    Err(anyhow!("link after enabling caching"))
```
mehulkar, line 272: Is this related to this PR?

mehulkar (commented):
I've done a first pass at a review, but I'm definitely not the right person for the job at the moment. I haven't spent much time with the daemon in Go and also haven't spent any time learning Rust yet 🙈 🦀. I can do another pass / run the code if you would like a deeper dive though!

crates/turborepo-lib/src/globwatcher/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+use log::{trace, warn};
+use notify::RecommendedWatcher;
+
+/// Tracks changes for a given hash. A hash is a unique identifier for a set of
+/// files. Given a hash and a set of globs to track, this will watch for file
+/// changes and allow the user to query for changes.
+#[derive(Clone)]
+pub struct HashGlobWatcher<T: Watcher> {
+    hash_globs: Arc<Mutex<HashMap<String, Glob>>>,
+    glob_status: Arc<Mutex<HashMap<String, HashSet<String>>>>,
```
chris-olszewski, line 19: Should the keys to the maps be `Rc<String>` since they're duplicated between `hash_globs` and `glob_status`?

crates/globwatch/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+}
+
+impl GlobWatcher<notify::RecommendedWatcher> {
+    /// Create a new watcher, using the given flush directory as a temporary
+    /// storage when flushing file events. For more information on flushing,
+    /// see the module-level documentation.
+    #[tracing::instrument]
+    pub fn new(flush_dir: PathBuf) -> Result<(Self, GlobSender), Error> {
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
+        let (tconf, rconf) = tokio::sync::mpsc::channel(10);
```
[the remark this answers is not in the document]
arlyon: No real reason, I may make it unbounded.

crates/turborepo-lib/src/globwatcher/mod.rs
arlyon: I like it.

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -1,10 +1,22 @@
...
 /// Runs the daemon command.
-pub async fn main(command: &DaemonCommand, base: &CommandBase) -> anyhow::Result<()> {
+pub async fn main(
+    command: &Option<DaemonCommand>,
+    base: &CommandBase,
+    idle_time: &Option<String>,
+) -> anyhow::Result<()> {
+    let command = match command {
+        Some(command) => command,
+        None => return run_daemon(base, idle_time).await.map_err(Into::into),
```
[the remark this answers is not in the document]
arlyon: Result is not smart enough to be able to map the err variant.

crates/globwatch/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        if chunk.starts_with('{') && chunk.ends_with('}') {
+            return chunk[1..chunk.len() - 1]
+                .split(',')
+                .map(|c| chunks.iter().chain(std::iter::once(&c)).collect())
+                .collect();
+        }
+
+        // a question mark in the first character is invalid
+        if chunk.starts_with('?') {
```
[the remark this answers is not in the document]
arlyon: Copying a below comment which addresses this:

> I have thought about this a fair bit. I think in globwatch we can just be conservative and watch the parent folder, as the HashGlobWatcher goes and validates all the changes against the raw globs. I think it is best summarized as 'globwatch does a best effort'.

crates/turborepo-lib/src/globwatcher/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+                let mut hash_globs = self.hash_globs.lock().expect("ok");
+
+                for ((glob, hash_status), path) in glob_status
+                    .iter()
+                    .cartesian_product(repo_relative_paths_iter)
+                    .filter(|((glob, _), path)| {
+                        glob_match::glob_match(glob, path.to_str().unwrap())
+                    })
+                {
+                    for hash in hash_status.iter() {
```
NicholasLYang, line 90: Some of this might be nice to factor out into separate functions. It's a little hard to keep a mental model between all the glob/globs/glob_status/hash_globs.

NicholasLYang (changes requested):

crates/turborepo-lib/src/daemon/endpoint.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Gets a stream of incoming connections from a Unix socket.
+/// On windows, this will use the `uds_windows` crate, and
+/// poll the result in another thread.
+pub async fn open_socket(
+    path: turborepo_paths::AbsoluteNormalizedPathBuf,
+) -> Result<
+    (
+        pidlock::Pidlock,
+        impl Stream<Item = Result<impl Connected + AsyncWrite + AsyncRead, std::io::Error>>,
```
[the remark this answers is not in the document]
arlyon: Its an ugly type, but a where clause would require two generic parameters introduced on the function signature which I don't think is very nice either..

crates/turborepo-lib/src/daemon/bump_timeout.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub fn deadline(&self) -> Instant {
+        self.0 + self.duration()
+    }
+
+    pub fn elapsed(&self) -> Duration {
+        self.0.elapsed()
+    }
+
+    /// Resets the deadline to the current time plus the given duration.
+    pub fn reset(&self, duration: Duration) {
```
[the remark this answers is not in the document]
arlyon: Do you mind elaborating a little bit. We don't do any read-then-writes, so maybe I am missing something.

crates/turborepo-lib/src/commands/link.rs
mehulkar: no worries, just wanted to make sure i understand the diff

crates/turborepo-lib/src/daemon/bump_timeout.rs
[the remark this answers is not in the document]
NicholasLYang: Let's say we have three threads. Thread A passes 5s and thread B passes 1s. Thread C calls wait. This sequence of events could happen (I think)

A -> sets duration to 5s + elapsed
B -> sets duration to 1s + elapsed
A -> stores duration as 5s + elapsed
C -> calls wait reads duration as 5s + elapsed, sleeps until then
B -> stores duration as 1s + elapsed

I'm not sure this is a big issue because really it just means a thread may sleep for a little too long.

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -59,6 +62,43 @@ pub async fn main(command: &DaemonCommand, base: &CommandBase) -> anyhow::Result
...
+        turborepo_paths::AbsoluteNormalizedPathBuf::new(base.repo_root.clone()).expect("absolute");
+
+    let timeout = idle_time
+        .as_ref()
+        .map(|s| {
+            go_parse_duration::parse_duration(s).map_err(|_| DaemonError::InvalidTimeout(s.clone()))
+        })
+        .transpose()?
+        .map(|d| Duration::from_nanos(d as u64))
+        .unwrap_or_else(|| Duration::from_secs(60 * 60 * 4));
```
chris-olszewski, line 94: This duration should probably be extracted as a constant since `Duration::from_secs` is now a const function.

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -59,6 +62,43 @@ pub async fn main(command: &DaemonCommand, base: &CommandBase) -> anyhow::Result
...
+        let hash = format!("{}-turbo.log", base.repo_hash());
+
+        let logs = turborepo_paths::ForwardRelativePath::new("logs").expect("forward relative");
+        let file = turborepo_paths::ForwardRelativePath::new(&hash).expect("forward relative");
+
+        folder.join(logs).join(file)
+    };
+
+    let repo_root =
+        turborepo_paths::AbsoluteNormalizedPathBuf::new(base.repo_root.clone()).expect("absolute");
```
chris-olszewski, line 85: Not necessarily related to your change, but we should probably be storing the repo root as an `AbsoluteNormalizedPathBuf`

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -0,0 +1,313 @@
...
+            changed_output_globs: changed.into_iter().collect(),
+        }))
+    }
+}
+
+impl<T: Watcher> NamedService for DaemonServer<T> {
+    const NAME: &'static str = "turborepo.Daemon";
+}
+
+#[cfg(test)]
```
chris-olszewski, line 239: I notice we dropped the unit tests for signal handling and cleanup on panic. Unsure if we get this functionality for free with tonic, but if not we should probably try to port them just to make sure there aren't any regressions.

arlyon: Yeah mainly because I wasn't sure about the best way to do those. I was hoping to replicate your integration tests but since they are flakey I'm not sure what we should do here. Putting a note to discus later.

crates/turborepo-lib/src/commands/daemon.rs
arlyon: Opted instead to use the clap default_value_t flag to we can document this in the cli.

crates/turborepo-lib/src/commands/daemon.rs
arlyon: Lets discuss today, want to get the semantics right here.

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+}
+
+#[tonic::async_trait]
+impl<T: Watcher + Send + 'static> proto::turbod_server::Turbod for DaemonServer<T> {
+    async fn hello(
+        &self,
+        request: tonic::Request<proto::HelloRequest>,
+    ) -> Result<tonic::Response<proto::HelloResponse>, tonic::Status> {
+        self.bump_timeout();
```
gsoltis, line 168: Can we use an [interceptor](https://docs.rs/tonic/latest/tonic/service/trait.Interceptor.html) ([example](https://github.com/hyperium/tonic/blob/master/examples/src/interceptor/server.rs#L37)) like in the Go implementation so that we don't have to remember to `bump_timeout` on every call of every service?

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -0,0 +1,313 @@
...
+        } else {
+            Ok(tonic::Response::new(proto::HelloResponse {}))
+        }
+    }
+
+    async fn shutdown(
+        &self,
+        _request: tonic::Request<proto::ShutdownRequest>,
+    ) -> Result<tonic::Response<proto::ShutdownResponse>, tonic::Status> {
+        self.shutdown.lock().await.take().map(|s| s.send(()));
```
gsoltis, line 181: Can you confirm that this triggers an asynchronous shutdown? Otherwise there is a deadlock for a graceful shutdown waiting for this shutdown rpc to finish.

gsoltis (changes requested):
I started looking at the server implementation, but I haven't gotten to the filewatching portion just yet.

crates/globwatch/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+        rx.await.map_err(|_| ConfigError)
+    }
+}
+
+/// Gets the minimum set of paths that can be watched for a given glob,
+/// specified in minimatch glob syntax.
+///
+/// note: it is currently extremely conservative, handling only **, braces, and
+/// ?. any other case watches the entire directory.
+fn glob_to_paths(glob: &str) -> Vec<PathBuf> {
```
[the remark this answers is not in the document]
gsoltis: I agree with @NicholasLYang I think we ought to see if we can do some pre-parsing of these into a more useful data format.

crates/globwatch/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+                .timeout_at(token),
+        )
+    }
+
+    #[tracing::instrument(skip(watcher))]
+    fn handle_config_change(watcher: &mut T, config: WatcherChange) {
+        match config {
+            WatcherChange::Include(glob, _) => {
+                for p in glob_to_paths(&glob) {
+                    if let Err(e) = watcher.watch(&p, notify::RecursiveMode::Recursive) {
```
gsoltis, line 178: What happens if we fail to watch something? What is the fallback if we later ask "did any of these files change"?

crates/turborepo-lib/src/globwatcher/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+use globwatch::{GlobSender, GlobWatcher, StopToken, Watcher};
+use itertools::Itertools;
+use log::{trace, warn};
+use notify::RecommendedWatcher;
+
+/// Tracks changes for a given hash. A hash is a unique identifier for a set of
+/// files. Given a hash and a set of globs to track, this will watch for file
+/// changes and allow the user to query for changes.
+#[derive(Clone)]
+pub struct HashGlobWatcher<T: Watcher> {
```
gsoltis, line 17: Can you document what the different hashmaps are?

crates/turborepo-lib/src/globwatcher/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+            Some(watcher) => watcher.into_stream(token),
+            None => {
+                warn!("watcher already consumed");
+                return;
+            }
+        };
+
+        // watch all the globs currently in the map
+        for glob in start_globs {
+            self.config.include(glob.to_string()).await.unwrap();
```
gsoltis, line 64: Can this fail? What happens if we fail to start watching a glob?

crates/turborepo-lib/src/globwatcher/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+        &self,
+        hash: &str,
+        mut candidates: HashSet<String>,
+    ) -> HashSet<String> {
+        self.config.flush().await.unwrap();
+
+        let globs = self.hash_globs.lock().unwrap();
+        match globs.get(&Arc::new(hash.to_string())) {
+            Some(glob) => {
+                candidates.retain(|c| glob.include.contains(c));
```
gsoltis, line 137: Is this backwards? If there's some new glob we don't know about in `candidates` it will get pruned here, but we don't actually know if anything changed because we (maybe) weren't watching it. 

IIUC, `hash_globs` contains the globs that _haven't_ changed.

gsoltis (changes requested):
Re: filewatching

Some meta comments: 
`globwatcher.go` has a fair amount of comments, since it's complicated code. I would love to see the same here.

We also need to port the globwatcher tests from `globwatcher_test.go`.

In the rust implementation of `globwatcher`, it's not clear to me what happens when there are filewatching errors. 

It looks to me like we often call `.unwrap()` on things like cookie files, which can definitely fail. On that same note, we should have a timeout waiting for them along with fallback behavior.

crates/turborepo-lib/src/daemon/server.rs
arlyon: I have reworked this into a `tower` layer which is compatible with any hyper-based server (not just grpc servers).

crates/turborepo-lib/src/daemon/server.rs
arlyon: This function does not yield after the signal is sent so we know the response is sent before the server can close.

I believe that this would still not be an issue if it did though. I will however move the shutdown handler since as it is current requests are just dropped midflight rather than gracefully allowing them to finish.

crates/globwatch/src/lib.rs
arlyon: It depends per-watcher. For FsEvents on macOS.

Although you're right we will need to notify the caller about failed watches..

crates/globwatch/src/lib.rs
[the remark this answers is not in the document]
arlyon: All done, with handling unicode in the process.

crates/turborepo-lib/src/globwatcher/mod.rs
arlyon: Yes, you are correct. We want to retain the items in candidate that are _not_ in glob.

crates/globwatch/src/lib.rs
arlyon: @gsoltis I am currently returning a grpc error if watching failed but perhaps the correct option is to just silently return an 'ok' here?

crates/turborepo-lib/src/globwatcher/mod.rs
arlyon: I have reworked it to bubble up

crates/globwatch/src/lib.rs
gsoltis: Returning an error if watching failed (`NotifyOutputsWritten`) is fine. The important part is when we later ask if anything has changed (`GetChangedOutputs`). If any errors have occurred, including failing to start watching, we _must_ return that everything changed, because we don't know any better.

crates/turborepo-lib/src/globwatcher/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+            };
+
+            // we can stop tracking that glob
+            globs.include.remove(glob);
+            if globs.include.is_empty() {
+                hash_globs.remove(hash);
+            }
+
+            // store the hash and glob so we can remove it from the glob_status
+            exclude_globs.push(glob.to_owned());
```
gsoltis, line 236: Two comments on this line:
 1. I think `exclude_globs` will have duplicates, we could have multiple hashes per glob
 2. I _think_ we don't want to stop watching the glob if we hit the exclusion case above. 

Maybe we try something like `let mut stop_watching = true` before the loop over hashes, and we set `stop_watching = false` if we hit an exclusion. Then we can add to `exclude_globs` once if `stop_watching` is still true after the loop over hashes?

gsoltis (changes requested):
This is getting there. The server part looks pretty solid.

For the filewatching, I think we need the globwatcher tests ported from `globwatcher_test.go`. That will help a lot to build some confidence that this will work as intended.

crates/globwatch/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+    Exclude(String, Option<Id>),
+}
+
+/// A sender for watcher configuration changes.
+#[derive(Debug, Clone)]
+pub struct GlobSender(tokio::sync::mpsc::Sender<WatcherCommand>);
+
+/// The server is no longer running.
+#[derive(Debug, Copy, Clone)]
+pub struct ConfigError;
```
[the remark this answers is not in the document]
arlyon: I have reworked ConfigError to have a few different cases so I am dismissing this.

crates/globwatch/src/lib.rs
arlyon: I believe this has been handled properly now, by capturing globs with failed watch notices and immediately removing them again from the hashmaps.

gsoltis (approved):
Globwatching LGTM.

Let's get the tests running and passing and then let's get it merged!

crates/glob-match/src/lib.rs
```diff
@@ -24,13 +24,13 @@ struct Wildcard {
 
 type Capture = Range<usize>;
 
-pub fn glob_match(glob: &str, path: &str) -> bool {
+pub fn glob_match(glob: &str, path: &str) -> Option<bool> {
```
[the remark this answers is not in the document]
arlyon: The option is to signify invalid globs. If the glob is invalid, you get a None since matching / not matching doesn't really make sense.

PR #4308, opened 2023-03-22
Consolidate task outputs

gsoltis:
### Description

 - Moves all task output into `TaskCache`. This is unlikely to be its final home, but it at least consolidates it.
 - Makes some slight behavior changes: turbo errors that happen during a task now have the task prefix in front of them. In general, any output related to a particular task will run through the `TaskCache` instance for that task in preparation for saving all of it together.

### Testing Instructions

Existing integration tests, including the new one for task failures

### DIFF
--- a/cli/internal/run/real_run.go
+++ b/cli/internal/run/real_run.go
@@ -10,7 +10,6 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


anthonyshew:
Just cleaning up some stale stuff!

PR #4309, opened 2023-03-22
Tests for Vercel Run uploads

mehulkar:


mehulkar:
@NicholasLYang I updated this to a node server instead of python and am seeing better results with the issue where it was having trouble consistently killing the server. I think I want to defer using the Rust crate. It'll need a bit more work that I'm not prepared to do right now:

- keep track of and expose the pid to kill afterwards
- write port to a file so test can use it and point TURBO_API to it
- customize response somehow (i want a runID in my test that is used later in the test, so hardcoding in the general mock server doesn't make as much sense to me)

I made some progress on these in another branch but then had trouble making the server run from the integration test context and decided to timebox myself. I'll put these changes in another branch if you can spot what I'm doing wrong

turborepo-tests/integration/tests/run_summary/spaces.t
```diff
@@ -0,0 +1,89 @@
...
+
+Kill what's running on port 8000 first, but also return 0 exit code if nothing is running on 8000
+  $ PID=$(lsof -t -i:8000 2>/dev/null) && [[ -n $PID ]] && kill $PID || true
+
+Start mock server. Note if anything fails in the test run after this,
+the cleanup step won't run at the end so we have to be careful
+send stdout and stderr to /dev/null and also background the server
+  $ node "${TESTDIR}/mock-api.js" --port 8000 &
+
+Run turbo. Since we have a --token (Note: remote caching will be turned on, and there isn't a way to turn it off for this test)
```
arlyon, line 13: > Since we have a --token

Incomplete sentence?

arlyon (approved):
Approving as a fix-ship

PR #4310, opened 2023-03-22
Add svelte app into kitchen sink example

mehulkar:


mehulkar:
I haven't had a chance to finish this, and considering I'll be out for a couple months, I think it's better to close this down right now

PR #4312, opened 2023-03-22
Don't hash env var when the value is an empty string

mehulkar:


cli/internal/env/env.go
```diff
@@ -68,8 +68,12 @@ func (evm EnvironmentVariableMap) mapToPair(transformer func(k string, v string)
...
 func (evm EnvironmentVariableMap) ToSecretHashable() EnvironmentVariablePairs {
 	return evm.mapToPair(func(k, v string) string {
-		hashedValue := sha256.Sum256([]byte(v))
-		return fmt.Sprintf("%v=%x", k, hashedValue)
+		if v != "" {
+			hashedValue := sha256.Sum256([]byte(v))
+			return fmt.Sprintf("%v=%x", k, hashedValue)
+		}
+
+		return fmt.Sprintf("%v=%s", k, "")
```
tknickman, line 76: 👍🏼

tknickman (approved):

PR #4313, opened 2023-03-23
fix: use signal handler in shim

### LINKED ISSUE #3711: [turborepo] regression on handling SIGINT

1.7.4

npm

Linux

### Describe the Bug

Referring to #444 and its implementation on #663, it looks like there was a behavioural regression introduced in `1.7.0` where SIGINT is no longer handled properly. The last working version was `1.6.3`.

The bug is related to the fact that when interrupting a program (`^C`), turbo doesn't wait for the program to handle the SIGINT command and immediately terminates the process.

_It is worth noting that the SIGINT instruction continues to run in the background, and the terminal becomes interactive again. This poses a problem when running turbo from a Docker container, for example, while Docker waits for the SIGINT instruction to be handled and then kills the container process. Since turbo doesn't wait for SIGINT to complete, docker simply kills the process prematurely. The expected behaviour here would be that of `1.6.3`_.

### Expected Behavior

Turbo should wait for the program to handle the `SIGINT` instruction and only then exit the running process.

### To Reproduce

__./apps/bar/package.json__
```jsonc
{
   "name": "@foo/bar",
   "scripts": {
       "start": "trap 'echo wait…; sleep 3; echo done!; exit 0' SIGINT; sleep 10",
   }
}
```

__turbo@1.6.3__ _(expected behaviour)_
```sh

$ npx turbo run start --filter=@foo/bar --only
• Packages in scope: @foo/bar
• Running start in 1 packages
• Remote caching disabled
@foo/bar:start: cache miss, executing a0c8f8ecc5ad4cad
@foo/bar:start:
@foo/bar:start: > start
@foo/bar:start: > trap 'echo wait…; sleep 3; echo done!; exit 0' SIGINT; sleep 10
@foo/bar:start:
^C@foo/bar:start: wait…  # <<< SIGINT
@foo/bar:start: done!
$ echo interactive again....
```

__turbo@1.7.0__ _(unexpected behaviour)_
```sh
$ npx turbo run start --filter=@foo/bar --only
• Packages in scope: @foo/bar
• Running start in 1 packages
• Remote caching disabled
@foo/bar:start: cache miss, executing 52f6ea6626d6482f
@foo/bar:start:
@foo/bar:start: > start
@foo/bar:start: > trap 'echo wait…; sleep 3; echo done!; exit 0' SIGINT; sleep 10
@foo/bar:start:
^C@foo/bar:start: wait… # <<< SIGINT
$ echo interactive again
interactive again
$ @foo/bar:start: done! # <<< bg process
```

### Reproduction Repo

_No response_

chris-olszewski:
### Description
I believe this fixes #3711

#4196 added graceful shutting down when calling `go-turbo`, this does the same but for when we need to invoke the correct Rust binary in `shim.rs`. I extracted the logic added in #4196 to a function that can be used throughout the codebase.

### Testing Instructions

Currently only doing manual via the reproduction provided in #3711.
Edit: I originally wanted to add integration test to cover these cases, this has proven to be a challenge. Will still look at trying to orchestrate a test for this, but considering how much traction the related issues have, I don't want to block on getting an integration test.


arlyon:
Looks solid so far! I am going to give this a whirl and see if I can break it. Looks like we just had to properly handle all the possible entry points.

arlyon:
After some experimentation, I can confirm this fixes the global shim issue at least on my machine :)

raspo:
Is it possible to add a unit test to prevent this kind of regression in the future?
In any case, thank you for the fix  🙏

arlyon (approved):
I agree a few integration tests for the various entry points may be useful at this point. Approving as a fixship.

steveluscher:
Thank you for this!

Issue #4289, opened 2023-03-21
[turborepo] turbo prune --docker puts patches folder in full subfolder -- would json subfolder be more appropriate?


shawnmcknight:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

When running `turbo prune --scope=<pkg> --docker` and pnpm `patchedDependencies` exists, the prune process will copy the `patches` folder to `out/full/patches`.  Based on turborepo's [documentation for deploying with docker](https://turbo.build/repo/docs/handbook/deploying-with-docker), the recommended approach is to first copy content from the `out/json` folder and then install dependencies with the package manager.  Subsequently, the remaining content from `out/full` is copied in.  This separation of content between the `out/json` folder and the `out/full` folder provide a nice separation of what's needed to install dependencies versus what's needed in the final build.

However, with the `patches` folder being copied to `out/full`, content from `full` must be copied prior to installing dependencies to ensure that the patches get applied properly.  This seems contrary to the intent of the `json` folder being used for dependency bootstrapping and the `full` folder being used after dependency installation.

### Describe the solution you'd like

Place the pnpm `patches` folder in `out/json/patches` rather than `out/full/patches`.  This will allow the documented step of `COPY --from=builder /app/out/json/ .` to continue to copy in the content necessary for performing a `pnpm install` as it would bring the pnpm patches through

### Describe alternatives you've considered

I've currently adjusted my Dockerfile to have the following steps to ensure patches are available prior to installing:
```Dockerfile
# Install dependencies
COPY --link .gitignore .gitignore
COPY --link .npmrc .npmrc
COPY --link --from=pruner tmp/app/out/json/ .
COPY --link --from=pruner tmp/app/out/pnpm-lock.yaml ./pnpm-lock.yaml
COPY --link --from=pruner tmp/app/out/full/patches/ ./patches

RUN pnpm install --frozen-lockfile
```
Note the extra step when copying patches which would not be needed if it was present in the `out/json` folder.

chris-olszewski:
Thanks for the suggestion. This was definitely an oversight on my part.

shawnmcknight:
I just tested out the current version and this fix worked well.  Thanks!

Resolved by #4318.

PR #4318, opened 2023-03-23
feat(place patches in json folder when pruning for docker)

chris-olszewski:
### Description

Implements #4289

Using the `--docker` flag with prune produces a `full` and `json` directory where the `json` directory only includes everything required to install dependencies. If a dependency is patched it makes sense to include the patch in the `json` directory as both the `package.json` and the lockfile reference the patch file.

### Testing Instructions

Added an integration test to verify that patches end up in the `json` output folder.

### DIFF
--- a/cli/internal/prune/prune.go
+++ b/cli/internal/prune/prune.go
@@ -281,6 +281,15 @@ func (p *prune) prune(opts *turbostate.PrunePayload) error {
 			); err != nil {
 				return errors.Wrap(err, "Failed copying patch file")
 			}
+			if opts.Docker {
+				jsonDir := outDir.Join(turbopath.RelativeSystemPath("json"))
+				if err := fs.CopyFile(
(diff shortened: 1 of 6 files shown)


cli/internal/prune/prune.go
```diff
@@ -281,6 +281,15 @@ func (p *prune) prune(opts *turbostate.PrunePayload) error {
 			); err != nil {
 				return errors.Wrap(err, "Failed copying patch file")
 			}
+			if opts.Docker {
```
NicholasLYang, line 284: This appears to copy to both the full dir and the json dir, while the original suggestion was to copy to json instead of full. Is that on purpose?

NicholasLYang (approved):
One question, otherwise lgtm

cli/internal/prune/prune.go
chris-olszewski: Yes, json should just be a subset of full. `json` is just what is needed to install and `full` is what is needed to use the scoped workspace

Issue #4307, opened 2023-03-22
[turborepo] Next.js13 and tailwind do not have proper loader


hamidre13:
### What version of Turborepo are you using?

1.8

### What package manager are you using / does the bug impact?

Yarn v1

### What operating system are you using?

Mac

### Describe the Bug

I have a turborepo project and upgraded to next.js13. I do have tailwind as my CSS library. After creating app directory and adding the tailwind CSS import in the root layout.tsx I am getting this error:
`web:dev: error - ./styles/tailwind.scss
web:dev: Module parse failed: Unexpected character '@' (1:0)
web:dev: You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
web:dev: > @tailwind base;
web:dev: | @tailwind components;
web:dev: | @tailwind utilities;`

### Expected Behavior

application compiles and run without loader error compiles

### To Reproduce

simply add tailwind css to nextjs 13 and turborepo in app directory

chris-olszewski:
Hi @hamidre13, the error you linked could be due to a Turbo configuration problem, but it's hard to debug without the exact configuration. Could you provide a repo with a reproduction?

engrwaqasali:
facing the same loader issue when starting a new project with 

`npx degit vercel/turbo/examples/with-yarn with-yarn`

and simply cannot run the project with yarn dev

i reported the issue cannot start new project
[4316](https://github.com/vercel/turbo/issues/4316)

hamidre13:
Thanks @chris-olszewski , 

Here is my web tsconfig web file:
```{
  "extends": "tsconfig/nextjs.json",
  "include": [
    "next-env.d.ts",
    "**/*.ts",
    "**/*.tsx",
    ".next/types/**/*.ts"
  ],
  "exclude": [
    "node_modules"
  ],
  "compilerOptions": {
    "baseUrl": ".",
    "plugins": [
      {
        "name": "next"
      }
    ]
  }
}
```

Here is the next tsconfig:
```
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "display": "Next.js",
  "extends": "./base.json",
  "compilerOptions": {
    "target": "es5",
    "lib": ["dom", "dom.iterable", "esnext"],
    "allowJs": true,
    "skipLibCheck": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "incremental": true,
    "esModuleInterop": true,
[...]

PR #4321, opened 2023-03-23
fix: update example lockfiles

### LINKED ISSUE #4316: [turborepo] could not start a new project

1.8.5

Yarn v1

Mac

### Describe the Bug

could not start a new project from the example, i am new to turbo and i want to setup a monorepo, two dashboards,
```
waqasali@waqass-MBP new % npx degit vercel/turbo/examples/with-yarn with-yarn
Need to install the following packages:
  degit@2.8.4
Ok to proceed? (y) y
> cloned vercel/turbo#HEAD to with-yarn
waqasali@waqass-MBP new % cd with-yarn 
waqasali@waqass-MBP with-yarn % yarn
yarn install v1.22.19
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 🔨  Building fresh packages...
✨  Done in 3.84s.
waqasali@waqass-MBP with-yarn % git init
Initialized empty Git repository in /Users/waqasali/doerz/eduApply/monorepo/new/with-yarn/.git/
waqasali@waqass-MBP with-yarn % git add .
```
on yarn dev
it gives me the error
```
docs:dev: error - ../../packages/ui/Button.tsx
docs:dev: Module parse failed: Unexpected token (4:9)
docs:dev: You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
docs:dev: | 
docs:dev: | export const Button = () => {
docs:dev: >   return <button>Boop</button>;
docs:dev: | };
docs:dev: |

```

same error for another app

### other info
node version: 16.9.0
i am on mac m2

### My global installed packages
```
waqasali@waqass-MBP with-yarn % npm list -g --depth=0
/Users/waqasali/.nvm/versions/node/v16.19.0/lib
├── corepack@0.15.1
├── npm@8.19.3
├── turbo@1.8.5
└── yarn@1.22.19
```

### Expected Behavior

on yarn dev both the apps should start

### To Reproduce

run new project from example with-yarn
`npx degit vercel/turbo/examples/with-yarn with-yarn`

### Reproduction Repo

_No response_

chris-olszewski:
### Description

Update various lockfiles so that users who copy examples directly don't face unexpected breakages due to outdated lockfiles.

Fixes #4316 (and I believe #4307)

### Testing Instructions

As part of this PR, I changed the example tests so we run them using the lockfiles that are checked in. This should help us catch breakages due to old versions. I believe it will become increasingly important if we roll out @tknickman's new version of `create-turbo` that allows users to specify an example to use as a template as the template includes these lockfiles.


cli/integration_tests/setup_git.sh
```diff
@@ -1,11 +1,15 @@
...
 git init ${TARGET_DIR} --quiet --initial-branch=main
 GIT_ARGS="--git-dir=${TARGET_DIR}/.git --work-tree=${TARGET_DIR}"
 git ${GIT_ARGS} config user.email "turbo-test@example.com"
 git ${GIT_ARGS} config user.name "Turbo Test"
 echo "script-shell=$(which bash)" > ${TARGET_DIR}/.npmrc
-npm --prefix=${TARGET_DIR} install --silent
+# Some examples don't use npm
+if [[ "$SKIP_INSTALL" != "--skip-install" ]]; then
+  npm --prefix=${TARGET_DIR} install --silent
+fi
```
mehulkar, line 13: similar thing in #4190 (which i still haven't gotten passing), fyi https://github.com/vercel/turbo/pull/4190/files#diff-b7727a8fd178a7b4189de2fd392fa25a3f07140042eafbb2e339fa3407c63e01R14-R16

we can use this, it's fine

mehulkar (commented):
i think this makes sense, but
- I'm not sure if keeping lockfiles in all examples is what we want. (i'm good with it, but not sure why it's not the case now)
- the test script also updates to `turbo@canary`, so you might need 2 install steps (first with lockfile, then after the turbo@canary update if there was one.

examples_tests/setup.sh
```diff
@@ -30,7 +25,7 @@ function set_package_manager() {
 }
 
 # Set the packageManger version
-NPM_PACKAGE_MANAGER_VALUE="npm@8.1.2"
+NPM_PACKAGE_MANAGER_VALUE="npm@8.12.1"
```
chris-olszewski, line 28: Updating this to match the npm value that we use in `with-npm`

chris-olszewski:
> the test script also updates to turbo@canary, so you might need 2 install steps (first with lockfile, then after the turbo@canary update if there was one.

If I'm reading `setup.sh` correctly, I think that installs only happen after we've bumped the turbo version.

mehulkar:
> I think that installs only happen after we've bumped the turbo version.

yeah, but I think in a _safe_ npm world, you want `npm ci` to get `node_modules/` populated as "before the fact", and _then_ make updates? If you do edit package.json and then `npm i`, I don't think you would get the same result, not sure. Also very possible this is FUD, but I feel like I've been burned by this before and have since determined that the only way to get deterministic behavior is to do `npm i` on an "established" source of truth (meaning lockfile is up to date and `node_modules/` already exists)

examples_tests/yarn-with-npm.t
```diff
@@ -1,4 +1,5 @@
-  $ . ${TESTDIR}/setup.sh with-npm yarn
+Ignore Yarn warning that it found a package-lock.json
```
mehulkar, line 1: you could also delete the packagelock and do a commit like this: 
https://github.com/vercel/turbo/blob/1b383bbb9dcb5b0a203ee55d8fc6510a6123a142/cli/integration_tests/single_package/run-yarn.t#L4-L7

mehulkar (approved):
The `pnpm` tests are failing, but once those are resolved, I'm good with the general idea of this PR!

PR #4324, opened 2023-03-23
fix error handling and hanging problems

sokra:
### Description

* fix hanging tasks when recomputing cells
* fix inactive tasks when using streams
* fix fatal error handling for node.js evaluation

We need to ensure that Vc reads and connected Tasks are correctly handled in when producing streams. It's important that the Tasks are connected to something that is active when consumers read from the stream. Also Vc reads need to invalidate the stream and cause the production of a new stream. That was not handled correctly before.

Now we use a child task to produce the stream that is active as part of the function that returns the stream. We also use a little hack to update the stream in the parent task when the child task is invalidated.

It's a bit hacky, but eventually this kind of logic should probably move into turbo_tasks as part of some streaming concept.

### Testing Instructions

Placing `process.exit(0)` in `next.config.js` or `postcss.config.js`

### DIFF
--- a/crates/turbo-tasks-bytes/src/stream.rs
+++ b/crates/turbo-tasks-bytes/src/stream.rs
@@ -37,7 +37,7 @@ impl<T: Clone> Stream<T> {
         }
     }
 
-    /// Crates a new Stream, which will lazily pull from the source stream.
+    /// Creates a new Stream, which will lazily pull from the source stream.
     pub fn new_open(
         pulled: Vec<T>,
         source: Box<dyn StreamTrait<Item = T> + Send + 'static>,


crates/turbopack-node/src/evaluate.rs
```diff
@@ -242,80 +254,133 @@ pub async fn evaluate(
...
+    // Note the following code uses some hacks to create a child task that produces
+    // a stream that is returned by this task.
 
-    // The evaluation sent an initial intermediate value without completing. We'll
-    // need to spawn a new thread to continually pull data out of the process,
-    // and ferry that along.
-    let stream = JavaScriptStream::new_open(
+    // We create a new cell in this task, which will be updated from the
+    // [compute_evaluate_stream] task.
+    let cell = turbo_tasks::macro_helpers::find_cell_by_type(*JAVASCRIPTEVALUATION_VALUE_TYPE_ID);
```
alexkirsz, line 264: This needs to go into a `turbo-tasks` primitive. We shouldn't expose cell APIs or private macro generated identifiers outside of turbo tasks.

crates/turbopack-node/src/pool.rs
```diff
@@ -493,16 +493,17 @@ impl RunningNodeJsPoolProcess {
...
             recv_future,
             self.stdout_handler.handle_operation(),
             self.stderr_handler.handle_operation(),
         );
-        stdout?;
-        stderr?;
-        result
+        let result = result?;
+        stdout.context("unable to handle stdout from the node.js process in a structured way")?;
+        stderr.context("unable to handle stderr from the node.js process in a structured way")?;
```
alexkirsz, line 505:
```suggestion
        stdout.context("unable to handle stdout from the Node.js process in a structured way")?;
        stderr.context("unable to handle stderr from the Node.js process in a structured way")?;
```

alexkirsz (approved):

PR #4329, opened 2023-03-24
Implement Streaming Static/Proxy Rendering

jridgewell:
### Description

Implements streaming rendering for static/proxy renders (I don't know why they're called static or proxy…) using an initial `Headers` message, N `BodyChunk` messages, and a final `BodyEnd` message.

Updating proxy rendering was very easy. It already implemented separate messages for header and body, but it buffered the full body before sending the full contents. But static rendering was a little more difficult, it has multiple paths that are immediately ready with a full body that I didn't want to break apart. So I kept the `Response` message for a full response that transforms into a `ContentSourceContent::Static`, and implemented a new enum type for streaming responses into a `ContentSourceContent::HttpProxy`.

These is some weirdness with different data types on the `headers` field of a `ContentSourceContent::Static`'s `StaticContent` and a `ContentSourceContent::HttpProxy`'s `ProxyResultVc`. One is a `HeaderListVc` and the other is a `Vec<(String, String)>` (and really should be a `HeadersList`), but I'm leaving this for a follow up.

### Testing Instructions

Paired Next.js PR: https://github.com/vercel/next.js/pull/47476

Re: WEB-27

### DIFF
--- a/crates/turbopack-dev-server/src/source/mod.rs
+++ b/crates/turbopack-dev-server/src/source/mod.rs
@@ -52,6 +52,26 @@ impl BodyError {
     }
 }
 
+impl From<&str> for BodyError {
+    fn from(err: &str) -> Self {
+        BodyError {
+            err: err.to_string(),
+        }
+    }
+}
+
+impl From<String> for BodyError {
+    fn from(err: String) -> Self {
+        BodyError { err }
+    }
+}
+
+impl From<anyhow::Error> for BodyError {
+    fn from(value: anyhow::Error) -> Self {
+        value.to_string().into()
+    }
+}
+
 /// The return value of a content source when getting a path. A specificity is
 /// attached and when combining results this specificity should be used to order
 /// results.
@@ -246,29 +266,29 @@ pub struct ContentSourceData {
     pub cache_buster: u64,
 }
 
-type Chunk = Result<Bytes, BodyError>;
+pub type BodyChunk = Result<Bytes, BodyError>;
 /// A request body.
 #[turbo_tasks::value(shared)]
 #[derive(Default, Clone, Debug)]
 pub struct Body {
     #[turbo_tasks(trace_ignore)]
-    chunks: Stream<Chunk>,
+    chunks: Stream<BodyChunk>,
 }
 
 impl Body {
     /// Creates a new body from a list of chunks.
-    pub fn new(chunks: Vec<Chunk>) -> Self {
+    pub fn new(chunks: Vec<BodyChunk>) -> Self {
         Self {
             chunks: Stream::new_closed(chunks),
         }
     }
 
     /// Returns an iterator over the body's chunks.
-    pub fn read(&self) -> StreamRead<Chunk> {
+    pub fn read(&self) -> StreamRead<BodyChunk> {
         self.chunks.read()
     }
 
-    pub fn from_stream<T: StreamTrait<Item = Chunk> + Send + Sync + Unpin + 'static>(
+    pub fn from_stream<T: StreamTrait<Item = BodyChunk> + Send + Unpin + 'static>(
         source: T,
     ) -> Self {
         Self {
--- a/crates/turbopack-node/src/render/mod.rs
+++ b/crates/turbopack-node/src/render/mod.rs
@@ -1,15 +1,7 @@
-use async_stream::stream as generator;
 use indexmap::IndexMap;
 use serde::{Deserialize, Serialize};
-use turbo_tasks_bytes::Stream;
-use turbo_tasks_fs::FileSystemPathVc;
-use turbopack_core::asset::AssetVc;
-use turbopack_dev_server::source::Body;
 
-use crate::{
-    pool::NodeJsOperation, route_matcher::Param, source_map::trace_stack, ResponseHeaders,
-    StructuredError,
-};
+use crate::{route_matcher::Param, ResponseHeaders, StructuredError};
 
 pub(crate) mod error_page;
 pub mod issue;
@@ -52,7 +44,7 @@ enum RenderProxyIncomingMessage {
     Error(StructuredError),
 }
 
-#[derive(Deserialize)]
+#[derive(Deserialize, Debug)]
 #[serde(tag = "type", rename_all = "camelCase")]
 enum RenderStaticIncomingMessage {
     #[serde(rename_all = "camelCase")]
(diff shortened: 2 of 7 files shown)


sokra (changes requested):

crates/turbopack-node/src/render/mod.rs
```diff
@@ -51,8 +58,61 @@ enum RenderStaticIncomingMessage {
...
+
+pub(crate) fn stream_body_chunks(
+    mut operation: NodeJsOperation,
+    intermediate_asset: AssetVc,
+    intermediate_output_path: FileSystemPathVc,
+    project_dir: FileSystemPathVc,
+) -> Body {
+    let chunks = Stream::new_open(
+        vec![],
+        Box::pin(generator! {
```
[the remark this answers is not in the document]
jridgewell: This is a bit more difficult because of the 2 paths that have different return types. I was attempting to reuse code, and because the stream doesn't need to read any Vcs, I thought it would be ok (the node eval streaming did need to read Vcs).

For proxy results which only have a streamed result, it'd be ok to use the same manual cell approach. But for static rendering, I would need to completely change the return type and wouldn't be able to share code.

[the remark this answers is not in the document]
sokra: It read VCs, e.g. recv() applies source maps which reads VCs, also structured error print uses Vcs

jridgewell (commented):
Ok, this is now fully updated. I really dislike the amount of duplication that this requires for anything that streams. Because of the different response types (proxy always streams a response, but static can return a rewrite, full response, or stream response) I can't share code between the two.

Streaming itself is cumbersome. The incoming stream needs defined types, and the outgoing stream needs a new set of defined types with "prepared" values. This all happens in a pulling function, which is 2 steps removed from the actual place I want to use the stream in the main function. In the main function, I can only read the prepared values and return them. This might be simpler if we can macro over the boilerplate.

Oh, and the `generator! {}` block doesn't format, so that's ugly.

crates/turbopack-node/src/render/render_proxy.rs
```diff
@@ -44,105 +52,55 @@ pub async fn render_proxy(
...
-                    error,
-                    intermediate_asset,
-                    intermediate_output_path,
-                    project_dir
-                )
-                .await?
-            )
-        }
+    let mut stream = render.read();
+    let first = match stream.next().await {
```
sokra, line 69:
```suggestion
    let first = match stream.try_next().await? {
```

To propagate potential errors

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -246,29 +266,29 @@ pub struct ContentSourceData {
     pub cache_buster: u64,
 }
 
-type Chunk = Result<Bytes, BodyError>;
+pub type BodyChunk = Result<Bytes, BodyError>;
```
sokra, line 269:
```suggestion
pub type BodyChunk = Result<Bytes, SharedError>;
```

BodyError coverts everything to a string loosing all the structure. We want to avoid that.

crates/turbopack-node/src/render/render_proxy.rs
```diff
@@ -44,105 +52,55 @@ pub async fn render_proxy(
...
-        status,
-        headers,
-        body: body.into(),
-    })
+    let body = Body::from_stream(stream.map(|item| match item {
+        Ok(RenderItem::BodyChunk(b)) => Ok(b),
+        Ok(v) => Err(BodyError::new(format!("unexpected render item: {:#?}", v))),
+        Err(e) => Err(BodyError::new(format!(
+            "error streaming proxied contents: {}",
+            e
```
sokra, line 87: Use `.context()` instead of format!
We want to avoid converting to string

crates/turbopack-node/src/render/render_proxy.rs
```diff
@@ -172,13 +131,167 @@ async fn proxy_error(
...
+    intermediate_output_path: FileSystemPathVc,
+    project_dir: FileSystemPathVc,
+    error_path: FileSystemPathVc,
+) -> RenderStreamVc {
+    // Note the following code uses some hacks to create a child task that produces
+    // a stream that is returned by this task.
+
+    // We create a new cell in this task, which will be updated from the
+    // [render_stream_internal] task.
+    let cell = turbo_tasks::macro_helpers::find_cell_by_type(*RENDERSTREAM_VALUE_TYPE_ID);
```
sokra, line 170: We probably should move these hacks into turbo-tasks-bytes

crates/turbopack-node/src/render/render_static.rs
```diff
@@ -77,84 +93,48 @@ pub async fn render_static(
...
-                    trace_stack(
-                        error,
-                        intermediate_asset,
-                        intermediate_output_path,
-                        project_dir
-                    )
-                    .await?
-                )
+    let mut stream = render.read();
+    let first = match stream.next().await {
```
sokra, line 109: try_next

crates/turbopack-node/src/render/render_static.rs
```diff
@@ -201,3 +181,176 @@ async fn static_error(
...
+    let Ok(sender) = sender.await else {
+        // Impossible to handle the error in a good way.
+        return;
+    };
+
+    let stream = generator! {
+        let data = data.await?;
+        // Read this strongly consistent, since we don't want to run inconsistent
+        // node.js code.
+        let pool = pool.strongly_consistent().await?;
```
sokra, line 277: This strongly_consistent will no longer work since the pool is an argument and not a function call.
We need to move the get_pool into this method.
Probably the same for the evaluate pool

sokra (changes requested):
Look good in general. Reviewed from mobile. Looking forward to see this in action

crates/turbopack-dev-server/src/source/mod.rs
jridgewell: This requires a slightly larger change, and it's complicated by `ShareError` not being serializable. That would prevent any body response from being cachable, which I think is bad for our end goal? Anyways, I'd like to defer this to a follow up PR for the time being.

crates/turbopack-node/src/render/render_proxy.rs
jridgewell: Done.

crates/turbopack-node/src/render/render_proxy.rs
jridgewell: I agree, but I'm deferring to a follow up PR.

crates/turbopack-node/src/render/render_proxy.rs
jridgewell: I actually think we should move `Stream<T>` into `turbo-tasks` proper, and design a macro interface as you suggested.

crates/turbopack-node/src/render/render_static.rs
jridgewell: Done.

crates/turbopack-node/src/render/render_static.rs
jridgewell: Done.

crates/turbopack-dev-server/src/source/mod.rs
sokra: > That would prevent any body response from being cachable, which I think is bad for our end goal?

It would still be cacheable in memory.

While I like to have static rendering to be cacheable, next.js doesn't cache rendering at all in dev and always server-renders a fresh page.

sokra (approved):
Approving that, but I need some cleanup in follow-up PRs. Also make sure to document public methods

PR #4330, opened 2023-03-24
Revert "feat: port npm lockfile to Rust"

nathanhammond:
Reverts vercel/turbo#3632

There is nothing materially wrong with the work in #3632 but our current build pipeline can't handle it. Reverting it out temporarily to not lose the ability to release. Once we discover the root cause we'll reintroduce this change.


nathanhammond:
This does not appear to be the root cause.

PR #4331, opened 2023-03-24
Bump rustc version.

nathanhammond:
The [build issues](https://github.com/vercel/turbo/actions/workflows/turborepo-release-step-2.yml) appear to be related to the `rustc` version. Rust is in the process of attempting to land `llvm@16` which was released on March 18 (depending on where you stand in the world).

I've bisected the release versions and confirmed that `nightly-2023-03-21` is the last working release— _coincidentally_ the version from the day we ran the 1.8.5 release. For now I'm hardcoding that into our release process; I have not done any further investigation of the actual root cause which will enable us to update past `nightly-2023-03-21`.

We will need to do that at some point in order to make sure we don't lose the ability to release.


arlyon (approved):
We may need to check with turbopack / swc to make sure this bump is ok, I think that they have some limitations around which nightly releases are ok to use, and rust-toolchain covers both projects.

Still, approving this for now for the -repo side

alexkirsz (changes requested):
I'm surprised changing the toolchain did not trigger a full re-build of Turbopack. We're using the deprecated box syntax, so I know this at least should fail with a newer nightly. We'll have to update to fix this.

alexkirsz (approved):

nathanhammond:
Confirmed that the 2023-03-09 nightly works on Turborepo: https://github.com/vercel/turbo/actions/runs/4510036351/jobs/7940505843

Merging.

PR #4336, opened 2023-03-24
allow to mark tasks as finished 

sokra:
### Description

this allows to exclude them from strongly consistency

mark evaluate stream as finished to allow to return early from strongly consitent reads

### DIFF
--- a/crates/turbopack-node/src/evaluate.rs
+++ b/crates/turbopack-node/src/evaluate.rs
@@ -12,6 +12,7 @@ use parking_lot::Mutex;
 use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
 use turbo_tasks::{
+    mark_finished,
     primitives::{JsonValueVc, StringVc},
     util::SharedError,
     CompletionVc, RawVc, TryJoinIterExt, Value, ValueToString,
@@ -311,6 +312,7 @@ async fn compute_evaluate_stream(
     args: Vec<JsonValueVc>,
     sender: JavaScriptStreamSenderVc,
 ) {
+    mark_finished();
     let Ok(sender) = sender.await else {
         // Impossible to handle the error in a good way.
         return;
--- a/crates/turbopack-node/src/evaluate.rs
+++ b/crates/turbopack-node/src/evaluate.rs
@@ -380,6 +380,9 @@ async fn compute_evaluate_stream(
         if sender.send(value).await.is_err() {
             return;
         }
+        if sender.flush().await.is_err() {
+            return;
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


jridgewell (commented):
I think this is still buggy. If you have a stream that reads from a Vc, the Vc could be unfinished and racy? The stream itself could recompute several time before settling. I think the solution has be that we don't allow Vc inside streams at all?

jridgewell (approved):

crates/turbopack-node/src/evaluate.rs
```diff
@@ -380,6 +382,9 @@ async fn compute_evaluate_stream(
         if sender.send(value).await.is_err() {
             return;
         }
+        if sender.flush().await.is_err() {
```
jridgewell, line 385: Can this flush happen after the loop?

jridgewell (approved):

jridgewell:
jridgewell
So compute_evaluate_stream is immediately marked finished, which means that the request starts pulling a few chunks from the stream
But, the stream at some point reads a Vc, and that Vc isn’t finished yet
The first time it reads, it sends down some more data based on that result (edited) 
But then the Vc recomputes and settles
Now the stream’s cell is updated
We’ve already sent down data, so the client has something written into DOM (edited) 
But then the thing reading our StreamVc restarts, because the stream invalidated
And starts sending down data from the beginning again
So all those initial bytes are resent into the DOM
And possibly doubled


sokra
even when invalidated a function will run to completion before starting again
so the first compute_evaluate_stream will run to completion, producing the full stream.
The dev server run get_content and that will either return the first stream or the second one (depending on when the invalidation happens, in your example the first stream). Once it start sending to the client, the stream has left turbo tasks and it won't be affected by changes anymore.
Basically once get().strongly_consistent().await? returns the result is fixed for the dev server. It can't change the stream instance in the middle of serving.
We don't restart a stream, we create a new stream with different content. All streams are complete on their own


jridgewell
Ok, I was misunderstood the “even when invalidated a function will run to completion before starting again” part
The stream itself won’t stop pulling results from the process, and whoever’s reading the stream will have that stream reference until it closes
I thought turbo functions would stop polling into the future when an invalidation happens, and would restart with a new future


sokra
So the following would happen for an invalidate in the middle of streaming:
1. dev server calls get()
1. something calls evaluate()
1. evaluate() calls compute_evaluate_stream()
1. evaluate().await? returns a stream
1. get().await? returns a stream
1. dev server passes stream to hyper
1. compute_evaluate_stream() produces a stream element
1. hyper sends that to the client
1. something invalidates compute_evaluate_stream(), it enters InProgressDirty state
1. compute_evaluate_stream() produces more content
1. hyper sends more content to the client
1. compute_evaluate_stream() finished and restarts
1. compute_evaluate_stream() changes the cell in evaluate()
1. that invalidates the "something" that awaited evaluate()
1. "something" and compute_evaluate_stream() start
1. ...

PR #4337, opened 2023-03-24
Minimum Node Version

MANTENN:
### Description

SCR-20230324-g3z
Module resolution fails on Node 16
### Testing Instructions
1. Switch to node 16 and run `npm run dev`
2. Upgrade to node 18 and the error goes away


MANTENN:
Workspaces is introduced in npm 7; its not related to nvm

PR #4341, opened 2023-03-24
Add `Rope::to_bytes()`

wbinnssmith:
This implements `to_bytes()` on Rope.

Test Plan: Added a unit test

### DIFF
--- a/crates/turbo-tasks-fs/src/rope.rs
+++ b/crates/turbo-tasks-fs/src/rope.rs
@@ -109,6 +109,11 @@ impl Rope {
     pub fn to_str(&self) -> Result<Cow<'_, str>> {
         self.data.to_str()
     }
+
+    /// Returns a slice of all bytes
+    pub fn to_bytes(&self) -> Result<Cow<'_, [u8]>> {
+        self.data.to_bytes()
+    }
 }
 
 impl<T: Into<Bytes>> From<T> for Rope {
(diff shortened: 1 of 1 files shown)


crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -451,6 +456,21 @@ impl InnerRope {
             }
         }
     }
+
+    /// Returns a slice of all bytes.
+    pub fn to_bytes(&self) -> Result<Cow<'_, [u8]>> {
+        match &self[..] {
+            [] => Ok(Cow::Owned(vec![])),
+            [Shared(inner)] => inner.to_bytes(),
+            [Local(bytes)] => Ok(Cow::Owned(bytes.to_vec())),
```
jridgewell, line 465: Borrow instead of cloning

crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -451,6 +456,21 @@ impl InnerRope {
             }
         }
     }
+
+    /// Returns a slice of all bytes.
+    pub fn to_bytes(&self) -> Result<Cow<'_, [u8]>> {
+        match &self[..] {
+            [] => Ok(Cow::Owned(vec![])),
```
jridgewell, line 463: Borrow `EMPTY_BUF`

jridgewell (changes requested):

crates/turbo-tasks-fs/src/rope.rs
```diff
@@ -952,4 +975,11 @@ mod test {
             ]
         );
     }
+
+    #[test]
+    fn test_to_bytes() -> Result<()> {
+        let rope = Rope::from("abc");
+        assert_eq!(rope.to_bytes()?, Cow::Owned::<[u8]>(vec![0x61, 0x62, 0x63]));
```
jridgewell, line 982: Should be borrowed? Owned will only happen if there are multiple sections.

PR #4344, opened 2023-03-25
Record exit codes per task in a run

mehulkar:
This adds exitCodes per task when we know them. Currently, without the `--continue` flag, when a task fails, turbo exits without closing out the Run Summary. We will address this separately, but this lays the baseline for gathering successful and error codes when possible.


cli/internal/run/real_run.go
```diff
@@ -233,7 +233,8 @@ func (ec *execContext) exec(ctx gocontext.Context, packageTask *nodes.PackageTas
 		prefixedUI.Error(fmt.Sprintf("error fetching from cache: %s", err))
 	} else if hit {
 		ec.taskHashTracker.SetExpandedOutputs(packageTask.TaskID, taskCache.ExpandedOutputs)
-		tracer(runsummary.TargetCached, nil)
+		// We only cache successful builds, so we can assume this is a successCode exit.
```
mehulkar, line 236:
```suggestion
		// We only cache successful executions, so we can assume this is a successCode exit.
```

PR #4349, opened 2023-03-27
fix nested dynamic imports

sokra:
### Description

In the case when nesting `import()` where the second `import()` references an already available module, it tries to access an non-existing chunk loader module.

After this change it will generate `Promise.resolve().then(() => __turbopack_import__("...")` instead.

A little bit of refactoring was needed to gain access to the AvailabilityInfo in code generation.

fix WEB-789 ([link](https://linear.app/vercel/issue/WEB-789))

### DIFF
--- a/crates/turbopack-ecmascript/src/code_gen.rs
+++ b/crates/turbopack-ecmascript/src/code_gen.rs
@@ -1,5 +1,6 @@
+use serde::{Deserialize, Serialize};
 use swc_core::ecma::visit::{AstParentKind, VisitMut};
-use turbo_tasks::Value;
+use turbo_tasks::{debug::ValueDebugFormat, trace::TraceRawVcs, Value};
 use turbopack_core::chunk::availability_info::AvailabilityInfo;
 
 use crate::chunk::EcmascriptChunkingContextVc;
@@ -37,8 +38,14 @@ pub trait CodeGenerateableWithAvailabilityInfo {
     ) -> CodeGenerationVc;
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, Serialize, Deserialize, TraceRawVcs, ValueDebugFormat)]
+pub enum CodeGen {
+    CodeGenerateable(CodeGenerateableVc),
+    CodeGenerateableWithAvailabilityInfo(CodeGenerateableWithAvailabilityInfoVc),
+}
+
 #[turbo_tasks::value(transparent)]
-pub struct CodeGenerateables(Vec<CodeGenerateableVc>);
+pub struct CodeGenerateables(Vec<CodeGen>);
 
 pub fn path_to(
     path: &[AstParentKind],
--- a/crates/turbopack-ecmascript/src/lib.rs
+++ b/crates/turbopack-ecmascript/src/lib.rs
@@ -72,7 +72,9 @@ use self::{
         placeable::EcmascriptExportsReadRef, EcmascriptChunkItemContent,
         EcmascriptChunkItemContentVc, EcmascriptChunkItemOptions, EcmascriptExportsVc,
     },
-    code_gen::{CodeGenerateableWithAvailabilityInfo, CodeGenerateableWithAvailabilityInfoVc},
+    code_gen::{
+        CodeGen, CodeGenerateableWithAvailabilityInfo, CodeGenerateableWithAvailabilityInfoVc,
+    },
     parse::ParseResultVc,
 };
 use crate::{
@@ -411,7 +413,7 @@ impl EcmascriptChunkItem for ModuleChunkItem {
         availability_info: Value<AvailabilityInfo>,
     ) -> Result<EcmascriptChunkItemContentVc> {
         let this = self_vc.await?;
-        if *this.module.analyze().need_availability_info().await? {
+        if *this.module.analyze().needs_availability_info().await? {
             availability_info
         } else {
             Value::new(AvailabilityInfo::Untracked)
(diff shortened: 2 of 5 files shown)


crates/turbopack-ecmascript/src/references/esm/dynamic.rs
```diff
@@ -71,10 +75,22 @@ impl ChunkableAssetReference for EsmAsyncAssetReference {
...
 #[turbo_tasks::value_impl]
 impl CodeGenerateable for EsmAsyncAssetReference {
+    #[turbo_tasks::function]
+    async fn code_generation(
+        &self,
+        _context: EcmascriptChunkingContextVc,
+    ) -> Result<CodeGenerationVc> {
+        bail!("Code generation for EsmAsyncAssetReference requires availability info")
+    }
+}
```
alexkirsz, line 85: Is it still necessary for this to implement `CodeGenerateable`?

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -120,6 +120,35 @@ pub struct AnalyzeEcmascriptModuleResult {
     pub successful: bool,
 }
 
+#[turbo_tasks::value_impl]
+impl AnalyzeEcmascriptModuleResultVc {
+    #[turbo_tasks::function]
+    pub async fn need_availability_info(self) -> Result<BoolVc> {
```
alexkirsz, line 126:
```suggestion
    pub async fn needs_availability_info(self) -> Result<BoolVc> {
```

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -428,13 +446,20 @@ async fn gen_content(
...
+        if let Some(code_gen) = CodeGenerateableWithAvailabilityInfoVc::resolve_from(r).await? {
+            code_gens.push(code_gen.code_generation(context, availability_info));
+        } else if let Some(code_gen) = CodeGenerateableVc::resolve_from(r).await? {
             code_gens.push(code_gen.code_generation(context));
         }
     }
     for c in code_generation.await?.iter() {
         let c = c.resolve().await?;
-        code_gens.push(c.code_generation(context));
+        if let Some(code_gen) = CodeGenerateableWithAvailabilityInfoVc::resolve_from(c).await? {
```
alexkirsz, line 458: It's weird that we have `CodeGenerateable` implementors that will actually never have their `code_generation` implementation called. Perhaps `code_generation` should be an enum of `Default(CodeGenerateableVc), WithAvailabilityInfo(CodeGenerateableWithAvailabilityInfoVc)`.

crates/turbopack-ecmascript/src/tree_shake/chunk_item.rs
```diff
@@ -27,16 +28,32 @@ pub struct EcmascriptModulePartChunkItem {
...
+        self_vc.content_with_availability_info(Value::new(AvailabilityInfo::Untracked))
+    }
+
+    #[turbo_tasks::function]
+    async fn content_with_availability_info(
+        self_vc: EcmascriptModulePartChunkItemVc,
+        availability_info: Value<AvailabilityInfo>,
+    ) -> Result<EcmascriptChunkItemContentVc> {
+        let this = self_vc.await?;
+        let availability_info = if *this.module.analyze().need_availability_info().await? {
```
jridgewell, line 41: Is there harm in just always passing down the availability info?

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -400,15 +401,31 @@ impl EcmascriptChunkItem for ModuleChunkItem {
...
+        self_vc.content_with_availability_info(Value::new(AvailabilityInfo::Untracked))
+    }
+
+    #[turbo_tasks::function]
+    async fn content_with_availability_info(
+        self_vc: ModuleChunkItemVc,
+        availability_info: Value<AvailabilityInfo>,
+    ) -> Result<EcmascriptChunkItemContentVc> {
+        let this = self_vc.await?;
+        if *this.module.analyze().need_availability_info().await? {
```
jridgewell, line 414: Ditto.

crates/turbopack-ecmascript/src/code_gen.rs
```diff
@@ -26,6 +28,15 @@ pub trait CodeGenerateable {
     fn code_generation(&self, context: EcmascriptChunkingContextVc) -> CodeGenerationVc;
 }
 
+#[turbo_tasks::value_trait]
+pub trait CodeGenerateableWithAvailabilityInfo {
```
jridgewell, line 32: If there's no harm, could we eliminate this and use a default method impl in `CodeGenerateable` trait?

jridgewell (approved):

crates/turbopack-ecmascript/src/tree_shake/chunk_item.rs
sokra: cache key of turbo functions are the arguments. So if it would pass down the availability info even when not needed that would duplicate the work. If not passing down availability info it can share the cache. This wrapper function will make sure we can leverage more caching when availability info isn't needed.

crates/turbopack-ecmascript/src/lib.rs
sokra: Ok sounds good. That allows to save one resolve operation.

crates/turbopack-ecmascript/src/references/esm/dynamic.rs
sokra: removed

PR #4354, opened 2023-03-27
fix: error swallowing when using continue

chris-olszewski:
### Description

Fixes regression added in #4254 where we swallowed exit codes when `--continue` was used to invoke turbo.

Don't love this fix, but it was the most straightforward fix I could see. @mehulkar unsure if I'm handling the case where an execution summary doesn't have an exit code correctly.

### Testing Instructions

See added integration test and the changed run summary test which now has a 1 exit code.

### DIFF
--- a/cli/internal/run/real_run.go
+++ b/cli/internal/run/real_run.go
@@ -150,6 +150,22 @@ func RealRun(
 		base.UI.Error(err.Error())
 	}
 
+	// When continue on error is enabled don't register failed tasks as errors
+	// and instead must inspect the task summaries.
+	if ec.rs.Opts.runOpts.continueOnError {
+		for _, summary := range runSummary.RunSummary.Tasks {
+			if childExit := summary.Execution.ExitCode(); childExit != nil {
+				childExit := *childExit
+				if childExit < 0 {
+					childExit = -childExit
+				}
+				if childExit > exitCode {
+					exitCode = childExit
+				}
+			}
+		}
+	}
+
 	runSummary.Close(exitCode, base.RepoRoot)
 
 	if exitCode != 0 {
--- a/cli/internal/runsummary/execution_summary.go
+++ b/cli/internal/runsummary/execution_summary.go
@@ -89,6 +89,16 @@ func (ts *TaskExecutionSummary) MarshalJSON() ([]byte, error) {
 	return json.Marshal(&serializable)
 }
 
+// ExitCode access exit code nil means no exit code was received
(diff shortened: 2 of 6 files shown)


tknickman (approved):

cli/internal/run/real_run.go
```diff
@@ -150,6 +150,22 @@ func RealRun(
...
+				childExit := *childExit
+				if childExit < 0 {
+					childExit = -childExit
+				}
+				if childExit > exitCode {
+					exitCode = childExit
+				}
+			}
+		}
+	}
```
mehulkar, line 167: Hmm, when we use `--continue` does `errs := engine.Execute(visitorFn, execOpts)` not return the ChildExitErr? I feel like it would be a better fix to always return the ChildExit from the task and then choose what to do about it based on the `--continue` flag. (we don't have to block on this if it's urgent to get out)

cli/internal/runsummary/execution_summary.go
```diff
@@ -89,6 +89,16 @@ func (ts *TaskExecutionSummary) MarshalJSON() ([]byte, error) {
...
 
+// ExitCode access exit code nil means no exit code was received
+func (ts *TaskExecutionSummary) ExitCode() *int {
+	var exitCode int
+	if ts.exitCode == nil {
+		return nil
+	}
+	exitCode = *ts.exitCode
+	return &exitCode
+}
```
mehulkar, line 100: Is this different from just  `return ts.exitCode?` (in which case do you just want to make the field public?)

mehulkar (approved):
The test cases look good, and it's ok to use TaskSummary for this purpose, but I imagine that we're missing a better primitive return value from `engine.execute()`. I can rubber stamp this if we need to get something out though

cli/internal/runsummary/execution_summary.go
chris-olszewski: No, it should be the same. I was just trying to change as few lines as possible with this PR.

cli/internal/run/real_run.go
chris-olszewski: > Hmm, when we use --continue does errs := engine.Execute(visitorFn, execOpts) not return the ChildExitErr?

Correct, we swallow the error and just log a warning along with populating the task execution summary.

> I feel like it would be a better fix to always return the ChildExit from the task and then choose what to do about it based on the --continue flag

So doing this involves exposing the run options to the engine which is a little more plumbing that I wanted to do since this bug is giving front issues.

PR #4362, opened 2023-03-28
Make build faster

kdy1:
### Description

 - I disabled LTO for release builds. (Related: https://github.com/rust-lang/cargo/issues/9171)
We may need to revert these patches from CI

### Testing Instructions

---

fix WEB-812 ([link](https://linear.app/vercel/issue/WEB-812))

### DIFF
--- a/.github/workflows/turborepo-release.yml
+++ b/.github/workflows/turborepo-release.yml
@@ -14,6 +14,7 @@
 name: Turborepo Release
 
 env:
+  CARGO_PROFILE_RELEASE_LTO: true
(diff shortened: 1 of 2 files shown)


arlyon (approved):
I am for this for incremental builds. Is it worth re-enabling this during release builds? I realise that we haven't explicitly turned it on but lto can have pretty significant perf gains (between a few percent and 20 percent, from memory).

sokra (approved):
LGTM, but since it mostly affects Turborepo someone from Turborepo should better review it

arlyon (approved):

PR #4364, opened 2023-03-28
move env var handling into rust and expose summarize env as a command

arlyon:
### Description

We don't currently really handle env vars in a uniform way. This sets up clap to handle this by adding the clap env feature and apply the appropriate args, as well as pull the detection logic out of go.


chris-olszewski (commented):
Rust stuff looks good, will defer to people more in tune with run summary stuff to give the stamp.

One suggestion I have is that we should be adding integration tests to make sure that all of these flags get respected when they come from env vars instead of the command line. The serialization of Rust args has been a source of bugs in the past.

NicholasLYang (commented):
Should we add a note in the [command line reference page](https://turbo.build/repo/docs/reference/command-line-reference) that you can configure some arguments with env variables?

cli/integration_tests/basic_monorepo/run_summary/enable.t
```diff
@@ -0,0 +1,70 @@
...
+| true    | true    | yes      |
+| true    | false   | no       |
+
+| false   | missing | no       |
+| false   | true    | yes      |
+| false   | false   | no       |
+
+| missing | missing | no       |
+| missing | true    | yes      |
+| missing | false   | no       |
```
mehulkar, line 21: I'm not sure for sure that this is the right result for each combination, but it's a starting point

cli/internal/run/run.go
```diff
@@ -136,18 +136,6 @@ func optsFromArgs(args *turbostate.ParsedArgsFromRust) (*Opts, error) {
 }
 
 func configureRun(base *cmdutil.CmdBase, opts *Opts, signalWatcher *signals.Watcher) *run {
-	if os.Getenv("TURBO_FORCE") == "true" {
-		opts.runcacheOpts.SkipReads = true
-	}
-
-	if os.Getenv("TURBO_REMOTE_ONLY") == "true" {
-		opts.cacheOpts.SkipFilesystem = true
-	}
```
mehulkar, line 145: I think we should omit this transfer in this PR, especially because it seems like there are some unknowns as to what takes precedence (env var or flag) shown by the test cases I added

mehulkar (commented):
@arlyon thanks so much for adding this!

- Can we back out the other env vars and only focus on run summary? 
- Could you help me make the test pass? Since this is a new flag/env var it doesn't matter what we pick, but it would be good to match how the other env var / flag combos work. If the `env` feature from `clap` doesn't Just Work™ (and match our existing precedence for other flags), we'll have to make a decision about how to proceed

mehulkar:
> Should we add a note in the [command line reference page](https://turbo.build/repo/docs/reference/command-line-reference) that you can configure some arguments with env variables?

yeah, let's tackle that separately though, we can do it once for all flags

mehulkar (approved):
Tests passing and I backported the thing. I'd like to bikeshed the `--summarize` name a little and get some another pair of eyes before merge

cli/integration_tests/no_args.t
```diff
@@ -45,7 +45,7 @@ Make sure exit code is 2 when no args are passed
         --dry-run [<DRY_RUN>]            [possible values: text, json]
         --single-package                 Run turbo in single-package mode
         --filter <FILTER>                Use the given selector to specify package(s) to act as entry points. The syntax mirrors pnpm's syntax, and additional documentation and examples can be found in turbo's documentation https://turbo.build/repo/docs/reference/command-line-reference#--filter
-        --force                          Ignore the existing cache (to force execution)
+        --force                          Ignore the existing cache (to force execution) [env: TURBO_FORCE=]
```
mehulkar, line 48: wait why is the test passing when this is still here

mehulkar: because i didn't push up another commit 🤦🏾

cli/integration_tests/basic_monorepo/run_summary/enable.t
chris-olszewski: All of these outcomes looks solid to me.

chris-olszewski (approved):

PR #4366, opened 2023-03-28
better `turbopack-node` rendering error page

ForsakenHarmony:


sokra:
Would this still show the error overlay with potential compilation errors?

Since the runtime error might be caused by previous compilation errors, which are only shown in the error overlay.

ForsakenHarmony:
@sokra yes, this is just what's underneath the error overlay

sokra (approved):

PR #4368, opened 2023-03-28
allow to alias free vars to modules

sokra:
### Description

Allows to e. g. alias `Buffer` to `import { Buffer } from "node:buffer"`

Basically implements the webpack `ProvidePlugin`

### Testing Instructions

Tests added in next.js PR


crates/turbopack-ecmascript/tests/analyzer/graph/peg/large
ForsakenHarmony, line 1: what is this file?

crates/turbopack-core/src/compile_time_info.rs
```diff
@@ -3,54 +3,65 @@ use std::collections::HashMap;
...
             let mut map = std::collections::HashMap::new();
-            $crate::compile_time_defines_internal!(map, $($more)+);
+            $crate::definable_name_map_internal!(map, $($more)+);
             $crate::compile_time_info::CompileTimeDefines(map)
         }
     };
 }
 
+#[macro_export]
+macro_rules! free_var_references {
```
ForsakenHarmony, line 55: This is not used yet, right?

crates/turbopack-ecmascript/tests/analyzer/graph/free-vars/resolved-effects.snapshot
```diff
@@ -0,0 +1,3 @@
+0 -> 3 member call = ???*0*["from"]("Hello World")
+- *0* FreeVar(Buffer)
+  ⚠️  unknown global
```
ForsakenHarmony, line 3: does this not get detected in this testcase?

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2869,6 +2830,7 @@ pub enum WellKnownObjectKind {
     NodePreGyp,
     NodeExpressApp,
     NodeProtobufLoader,
+    NodeBuffer,
```
ForsakenHarmony, line 2833: is this just for testing?

ForsakenHarmony (approved):

crates/turbopack-ecmascript/tests/analyzer/graph/peg/large
sokra: It's a marker to skip some snapshots that would produce very large snapshots

crates/turbopack-core/src/compile_time_info.rs
sokra: It's used by next.js

crates/turbopack-ecmascript/tests/analyzer/graph/free-vars/resolved-effects.snapshot
sokra: Indeed... hmm...

sokra: fixed

crates/turbopack-ecmascript/src/analyzer/mod.rs
sokra: Its also used here: https://github.com/vercel/turbo/blob/07bf57235cdda2cc973e35b646464ee1847d1403/crates/turbopack-ecmascript/src/references/mod.rs#L1832

but there is no special behavior for it defined yet. But it could static evaluate e. g. `Buffer.from("SGVsbG8gV29ybGQ=", "base64").toString("utf-8")` in future, but I don't see a need for that right now.

PR #4377, opened 2023-03-29
Minimum requirements

MANTENN:
### Description

I spent a lot of time trying to get turbo working and concluded turbo requires NPM V7 because thats when workspaces were introduced.

### Testing Instructions

1. Downgrade to NPM 6 and run `turbo build` after `npm i`

- errors about missing packages will appear

2. Upgrade and the errors go away

### DIFF
--- a/docs/pages/repo/docs/installing.mdx
+++ b/docs/pages/repo/docs/installing.mdx
@@ -15,7 +15,7 @@ import { Tabs, Tab } from '../../../components/Tabs'
 - Windows 64-bit, ARM 64-bit
 
 Minimum Requirements
-- Node 18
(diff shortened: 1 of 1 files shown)


nathanhammond:
Hey @MANTENN I appreciate the PR.

A couple of things:
1. We don't actually require `npm@7` or newer in order to work in single-package mode. In single-package mode we probably work all the way back to `npm@5`.
2. The last LTS that included `npm@6` by default is Node.js 14, which is EOL in 6 days.

I'm sorry it took a bit to figure out that your chosen package manager didn't support workspaces, but I'm not intending to update the documentation for software that is about to be unsupported by its provider.

PR #4378, opened 2023-03-29
Make attempted include tasks that started but didn't finish

mehulkar:
The purpose of this is to

- account for tasks that somehow don't hit the `tracer()` function and get marked as done by the time the whole run is done. This is unlikely, but 🤷🏾 
- account for new states we may want to add
- makes `attempted` more semantically correct. If we started, there was an attempted! It doesn't matter what state we ended up in.


cli/internal/runsummary/execution_summary.go
```diff
@@ -107,10 +110,10 @@ type executionSummary struct {
...
 
 	// These get serialized to JSON
-	success   int
-	failure   int
-	cached    int
-	attempted int
+	success   int // number of tasks that exited successfully (does not include cache hits)
+	failure   int // number of tasks that exited with failure
+	cached    int // number of tasks that had a cache hit
+	attempted int // number of tasks that started
```
tknickman, line 116: This is still a little confusing. I would expect:

success + failure = attempted. And cached is <= success

cli/internal/runsummary/format_execution_summary.go
```diff
@@ -14,7 +14,13 @@ func (rsm *Meta) printExecutionSummary() {
 	summary := rsm.RunSummary
 	ui := rsm.ui
 
-	if summary.ExecutionSummary.cached == summary.ExecutionSummary.attempted && summary.ExecutionSummary.attempted > 0 {
+	attempted := summary.ExecutionSummary.attempted
+	successful := summary.ExecutionSummary.cached + summary.ExecutionSummary.success
+	cached := summary.ExecutionSummary.cached
+	// TODO: can we use a method on ExecutionSummary here?
+	duration := time.Since(summary.ExecutionSummary.startedAt).Truncate(time.Millisecond)
```
tknickman, line 21: Nice

tknickman (approved):

cli/internal/runsummary/execution_summary.go
mehulkar: yeah agreed, but I don't want to break how `success` is used right now

PR #4388, opened 2023-03-29
refactor(turborepo): Factored out API Client crate

NicholasLYang:
### Description
Factored out the API client into a separate crate so that we can use the API client without importing all of turborepo-lib. Also makes turborepo-lib less monolithic.


crates/turborepo-api-client/src/lib.rs
```diff
@@ -264,31 +267,33 @@ impl APIClient {
...
+        version: &'static str,
+    ) -> Result<Self> {
         let client = match timeout {
             Some(timeout) => reqwest::Client::builder()
                 .timeout(std::time::Duration::from_secs(timeout))
                 .build()?,
             None => reqwest::Client::builder().build()?,
         };
 
+        let user_agent = format!(
```
NicholasLYang, line 282: We make user_agent a field in APIClient so that turborepo-lib can pass the version info instead of having turborepo-api-client have a dependency on get_version. That way we don't have to share get_version between turborepo-api-client and turborepo-lib.

arlyon: I think that is a good choice.

arlyon (approved):

.github/CODEOWNERS
```diff
@@ -8,7 +8,7 @@
 
 /cli @vercel/turbo-oss
 /benchmark @vercel/turbo-oss
-/buildcontainer @vercel/turbo-oss
+n/buildcontainer @vercel/turbo-oss
```
mehulkar, line 11:
```suggestion
/buildcontainer @vercel/turbo-oss
```

sokra (approved):

PR #4392, opened 2023-03-29
Use `SharedError` for body streaming

jridgewell:
Follow up to https://github.com/vercel/turbo/pull/4329, this removes `BodyError` (which was just a simple `String` and not an `Error`) and switches to `SharedError`.

I've implemented a basic `PartialEq` and `Serialization` for `SharedError` so that it doesn't infect everything that uses `Body`.

### DIFF
--- a/crates/turbo-tasks/src/util.rs
+++ b/crates/turbo-tasks/src/util.rs
@@ -8,7 +8,7 @@ use std::{
     time::Duration,
 };
 
-use anyhow::Error;
+use anyhow::{anyhow, Error};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 pub use super::{id_factory::IdFactory, no_move_vec::NoMoveVec, once_map::*};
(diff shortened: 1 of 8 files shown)


kwonoj (approved):

crates/turbo-tasks/src/util.rs
```diff
@@ -47,6 +49,41 @@ impl From<Error> for SharedError {
...
+        let mut e = match messages.pop() {
+            Some(e) => anyhow!(e),
+            None => return Err(Error::custom("expected at least 1 error message")),
+        };
+        while let Some(message) = messages.pop() {
+            e = e.context(message);
+        }
+        Ok(SharedError::new(e))
+    }
+}
```
sokra, line 85: Not sure if this is a good idea. It's not perfectly restoring error types. Maybe we should return an Serialization error when there is an non-anyhow error in the chain.

sokra (approved):

Issue #4389, opened 2023-03-29
[WEB-818] [turborepo] [Regression] cache no longer working


notaphplover:
### What version of Turborepo are you using?

1.8.7

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Linux

### Describe the Bug

Since version 1.8.7, `turbo` is no longer caching my tasks.

### Expected Behavior

I expect `turbo` to cache my tasks.

### To Reproduce

I made a reproduction repo. Simply follow the next steps:

1. Clone the repo.
2. Install dependencies.
3. Run the build root script.
4. Run again the build root script.

If you try again with `turbo@a.8.6`, the second time the `build` root script is called, it hits the cache

### Reproduction Repo

https://github.com/notaphplover/turbortepo-cache-issue

<sub>[WEB-818](https://linear.app/vercel/issue/WEB-818/[turborepo]-[regression]-cache-no-longer-working)</sub>

tknickman:
Validated, fixing - thank you @notaphplover

tknickman:
We've deprecated 1.8.7, and will be rolling out 1.8.8 shortly.

tknickman:
1.8.8 is now available.

PR #4393, opened 2023-03-29
Add cache state to task summaries on real runs (2nd try)

mehulkar:
- Original: #4225
- Bug: #4389
- Revert: #4390 

The second commit in this PR shows the fix for the bug. I'm not sure I can write a test to cover this


cli/internal/runcache/runcache.go
```diff
@@ -124,26 +125,35 @@ func (tc *TaskCache) RestoreOutputs(ctx context.Context, prefixedUI *cli.Prefixe
...
+			return cache.ItemStatus{Local: false, Remote: false}, nil
 		}
 
 		if err := tc.rc.outputWatcher.NotifyOutputsWritten(ctx, tc.hash, tc.repoRelativeGlobs); err != nil {
 			// Don't fail the whole operation just because we failed to watch the outputs
 			prefixedUI.Warn(ui.Dim(fmt.Sprintf("Failed to mark outputs as cached for %v: %v", tc.pt.TaskID, err)))
 		}
 	} else {
+		// If no outputs have changed, that means we have a local cache hit.
+		cacheStatus.Local = true
```
mehulkar, line 156: This was the missing piece the last time around

NicholasLYang (approved):
I think this makes sense, but I'm not sure if I can repro locally, should we canary and test?

mehulkar:
> but I'm not sure if I can repro locally, should we canary and test?

This is how I test locally at both commits:
```
~/build-turbo.sh && localturbo run build --skip-infer --force && localturbo run build --skip-infer && localturbo run build --skip-infer
```
- `build-turbo` is an alias to `cd && make` in my turbo repo directory
- `localturbo` is an alias to where the turbo binary is built on my system

NicholasLYang:
Yeah I tried something similar. It worked, but I couldn't reproduce the bug when I checked out the first commit. I'll try again and confirm that the bug occurs with my local turbo version and that the second commit fixes it.

NicholasLYang:
Hmm I still can't get a cache miss on the first commit. If you can repro on the first commit and it seems to be fixed on the second commit then I'm down to merge.

mehulkar:
@NicholasLYang I just had some trouble reproducing also, but I was able to get it by running 20 builds in a row with:

```
$ ~/build-turbo.sh && localturbo run build --skip-infer --force && for i in {1..20}; do localturbo run build --skip-infer --output-logs=new-only; done
```

Here's my local output from `v1.8.7`, the tip of this branch, and the tip of this branch minus one commit

https://gist.github.com/mehulkar/3c312e9df2ede8800f4992caaacf6811

mehulkar:
going to merge this as I feel pretty confident that we've identified and fixed the bug.

PR #4397, opened 2023-03-30
ChunkingContext refactor pt. 1

alexkirsz:
### Description

This is part one of a larger ChunkingContext refactor.

In this episode, we:
1. Move the runtime and evaluation code off of the entry JS chunk and into a dedicated chunk. This means that, within a chunk group, instead of having one chunk that contains modules + runtime code and chunks that only contain modules, we'll have chunks that only contain modules and one chunk that only contains runtime code.
2. Start a larger enterprise of differentiating between "intermediate" chunks (`ChunkVc`s), and "output" chunks (raw `AssetVc`s). Before, we used `EcmascriptChunkVc`s to represent both intermediate chunks that would never actually be created (optimized away) AND final chunks that would be served or written to disk. Now, `EcmascriptChunkVc`s are only intermediate assets, and `DevEcmascriptChunkVc`s are the final chunks.
3. Move a few modules from `turbopack-core` to `turbopack-dev`, as their existence (or the implementation) is only meaningful in a dev context (`ChunkList` and `Manifest*`, the latter of which depends on the chunk list for now).

Things left to figure out:
1. What does the `Chunk` trait represent here? We no longer use `.chunking_context()`, and `.path()` is redundant with `Asset::path`. However, it no longer makes sense for `Chunk`s to be `Asset`s as their content is determined by the chunking context. Should we have two traits, one for intermediate chunks and one for output chunks?

The `Chunk` trait could look like this:

```rust
trait Chunk {
  fn ident(&self) -> IdentVc;
  fn parallel_chunks(&self) -> ChunksVc;
  fn references(&self) -> AssetReferencesVc;
}
```

Things that will happen in the next part(s):
1. The `ChunkGroupVc::evaluated` logic will move into `turbopack-dev`, as it doesn't make sense in a build context. This will essentially make `ChunkGroupVc` obsolete: the only thing we need it to do is to compute all parallel chunks, but I'm thinking this should be a method on the `Chunk` trait instead.

Things I'll leave for future refactors:
* Figure out the whole Input, Intermediate, and Output asset story.

### Testing Instructions

Snapshots + Next integration tests


crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -135,41 +150,79 @@ impl ChunkGroupVc {
...
+                    Ok(evaluated_entry
+                        .as_root_chunk(chunking_context)
+                        .resolve()
+                        .await?)
+                }
+            })
+            .try_join()
+            .await?
+            .into_iter()
+            .collect();
```
alexkirsz, line 220: This is a big difference: before, we were putting evaluated entries inside of the main entries of the "evaluated chunk". Now, they are considered separate chunks to begin with, and it's up to the optimizer to merge them.

crates/turbopack-tests/tests/snapshot.rs
```diff
@@ -305,13 +311,17 @@ async fn walk_asset(
         return Ok(());
     }
 
-    diff(path, asset.content()).await?;
+    if path.await?.is_inside(&*output_path) {
```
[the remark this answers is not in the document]
alexkirsz: I think this is the reason for the large negative line count of this PR.

crates/turbopack-dev/src/ecmascript/chunk.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+
+#[turbo_tasks::value_impl]
+impl GenerateSourceMap for EcmascriptDevChunk {
+    #[turbo_tasks::function]
+    fn generate_source_map(self_vc: EcmascriptDevChunkVc) -> OptionSourceMapVc {
+        self_vc.own_content().generate_source_map()
+    }
+}
```
sokra, line 115: A custom Introspectable probably makes sense to allow inspecting the inner chunk.. See ChunkInGroup

crates/turbopack-core/src/chunk/evaluate.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use super::{ChunkVc, ChunkableAsset, ChunkableAssetVc, ChunkingContext, ChunkingContextVc};
+use crate::asset::{Asset, AssetVc, AssetsVc};
+
+/// Marker trait for the chunking context to accept evaluated entries.
+///
+/// The chunking context implementation will resolve the dynamic entry to a
+/// well-known value or trait object.
+#[turbo_tasks::value_trait]
+pub trait EvaluatedEntry: Asset + ChunkableAsset {}
```
sokra, line 11:
```suggestion
pub trait EvaluatableAsset: Asset + ChunkableAsset {}
```

sokra (approved):
Might require some manual tests for:

* Introspection
* SourceMaps
* HMR
* source maps in runtime error overlay

crates/turbopack-dev/src/ecmascript/evaluate/chunk.rs
```diff
@@ -0,0 +1,265 @@
...
+                    {
+                        Ok(Some(
+                            placeable
+                                .as_chunk_item(chunking_context.into())
+                                .id()
+                                .await?,
+                        ))
+                    } else {
+                        Ok(None)
+                    }
```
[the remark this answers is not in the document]
alexkirsz: The purpose of the `EvaluatableAsset` trait is mostly to act as a marker trait. It's up to the implementor to actually decide *where* it can be evaluated, by implementing specific traits such as `EcmascriptChunkPlaceable`. For instance, we could have Wasm evaluated entries, which we might want to handle differently.

Issue #2688, opened 2022-11-13
Add possibility to use community templates with create-turbo


manuschillerdev:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

Currently, `create-turbo` creates a useful default setup with nextjs and docs as apps, and a shared ui package as example for using shared codes in apps.

I'd love to be able to reference different templates with `create-turbo`, similar to https://remix.run/docs/en/v1/pages/stacks.

It would be an awesome opportunity to involve the community and make getting started with turborepo even simpler.

### Describe the solution you'd like

```
pnpx create-turbo --template {git-url|local-path}
```

I'm not too familiar with the interals of `create-turbo`, so I'm not sure what technical details would need to be considered for being able to use templates from the community instead of core templates (https://github.com/vercel/turbo/tree/main/packages/create-turbo/templates)

### Describe alternatives you've considered

honestly: none so far

tknickman:
Took a little bit, but this finally landed! It's available now in canary, and we would love any feedback:

npx create-turbo@canary -e {git repo | or example}

manuschillerdev:
awesome! thanks @tknickman !

Issue #664, opened 2022-02-02
Ability to create a repo with create-turbo from an example


bring-shrubbery:
### Describe the feature you'd like to request

When running `create-turbo` at the moment, you always get the same placeholder setup (next.js, docs, tsconfig, config, ui packages). Having a `--example` flag that you can use to provide an example from the `examples` folder of this repo would be very nice. Maybe similar to how `create-next-app` does it?

### Describe the solution you'd like

I would like to run this command to create a design system turborepo:

```bash
npx create-turbo@latest --example design-system
```

### Describe alternatives you've considered

Alternatively, an "Examples" section could be added to the docs, which could automatically render the README files from examples in this repo. This would be a useful and simpler solution, compared to changing the create-turbo CLI.

DefinitelyMaybe:
perhaps a specific example could be chosen within the cli instead of as a flag? either way this would be helpful.

bring-shrubbery:
@DefinitelyMaybe absolutely, I was just suggesting the same way as next.js has when creating repo from an example.

tknickman:
Took a little bit, but this finally landed! It's available now in canary, and we would love any feedback:

`npx create-turbo@canary -e {example}`

bring-shrubbery:
@tknickman Nice! I will definitely try this out in a near future.

Issue #313, opened 2021-12-16
Able to `create-turbo` without TypeScript


JeniaBR:
### Describe the feature you'd like to request

Have the ability explicitly tell the CLI to bootstrap the app using JavaScript (without TypeScript).

### Describe the solution you'd like

A new CLI flag that will opt-in to bootstrap the app without TypeScript.

**In case we want the TS be the default:**
```
npx create-turbo@latest --js
```

**In case we want the JS be the default:**
```
npx create-turbo@latest --ts
```
The later case is also aligned with `create-next-app`. 

### Describe alternatives you've considered

Do it manually.

SushritPasupuleti:
This would honestly bring a lot more people into the monorepo space. Almost all the examples you can find are TS based.

Ben888GitHub:
I just tried `npx create-turbo@latest --js` and it is still created in Typescript, what can I really do to create this turborepo app with JavaScript?

JeniaBR:
> I just tried `npx create-turbo@latest --js` and it is still created in Typescript, what can I really do to create this turborepo app with JavaScript?

It is not implemented by Turborepo team.

You can do it manually, without `npx create-turbo` command.

mryechkin:
Would love to see this implemented 🙏

PR #4398, opened 2023-03-30
feat(create-turbo): support examples

tknickman:
### Description

`create-turbo` with examples support! Same API as `create-next-app`

TODO:

- [x] Docs
- [x] Tests

### Issues

fixes https://github.com/vercel/turbo/issues/313
fixes https://github.com/vercel/turbo/issues/664
fixes https://github.com/vercel/turbo/issues/2688


packages/create-turbo/__tests__/isWritable.test copy.ts
```diff
@@ -0,0 +1,35 @@
+import path from "path";
```
arlyon, line 1: This file name probably needs changing

packages/create-turbo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "create-turbo",
-  "version": "1.8.8",
+  "version": "1.9.0-alpha.0",
```
mehulkar, line 3: we probably don't want to bump this here?

arlyon, line 3: Exciting!

packages/create-turbo/src/cli.ts
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env node
+
+import chalk from "chalk";
+import { Command } from "commander";
+import notifyUpdate from "./utils/notifyUpdate";
+import { turboGradient } from "./logger";
```
mehulkar, line 6: seems like a candidaate for turbo-utils package

packages/create-turbo/src/cli.ts
```diff
@@ -0,0 +1,60 @@
...
+  .option(
+    "-e, --example [name]|[github-url]",
+    `
+  An example to bootstrap the app with. You can use an example name
+  from the official Turborepo repo or a GitHub URL. The URL can use
+  any branch and/or subdirectory
+`
+  )
+  .option(
+    "--example-path <path-to-example>",
```
mehulkar, line 34: any short alias for this? maybe `-p`?

packages/create-turbo/src/commands/create/createProject.ts
```diff
@@ -0,0 +1,231 @@
...
+    repoUrl = new URL(example);
+  } catch (err: any) {
+    if (err.code !== "ERR_INVALID_URL") {
+      error(err);
+      process.exit(1);
+    }
+  }
+
+  if (repoUrl) {
+    if (repoUrl.origin !== "https://github.com") {
```
mehulkar, line 52: can we use optional chaining? (aside: are we bundling/transpiling/minifying this package before publish?)

packages/create-turbo/src/commands/create/createProject.ts
```diff
@@ -0,0 +1,231 @@
...
+    if (!found) {
+      error(
+        `Could not locate an example named ${chalk.red(
+          `"${example}"`
+        )}. It could be due to the following:\n`,
+        `1. Your spelling of example ${chalk.red(
+          `"${example}"`
+        )} might be incorrect.\n`,
+        `2. You might not be connected to the internet or you are behind a proxy.`
+      );
```
mehulkar, line 94: Two very different types of errors. Seems like something we could throw/catch different known errors instead of this? (maybe i'm getting too used to Golang 😃 )

packages/create-turbo/src/commands/create/createProject.ts
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+
+  const root = path.resolve(appPath);
+
+  if (!(await isWriteable(path.dirname(root)))) {
+    error(
+      "The application path is not writable, please check folder permissions and try again."
+    );
+    error("It is likely you do not have write permissions for this folder.");
```
mehulkar, line 105: also could be three errors: the directory doesnt' exist, you don't have permissions, or unknown error

packages/create-turbo/src/commands/create/createProject.ts
```diff
@@ -0,0 +1,231 @@
...
+  example: string;
+  examplePath?: string;
+}): Promise<{
+  cdPath: string;
+  hasPackageJson: boolean;
+  availableScripts: Array<string>;
+}> {
+  let repoInfo: RepoInfo | undefined;
+  let repoUrl: URL | undefined;
+  const isDefaultExample = example === "basic" || example === "default";
```
arlyon, line 40: Maybe an as const would be handy here

packages/create-turbo/src/commands/create/createProject.ts
arlyon, line 52: I'm wondering if we'll end up with anyone who omits the scheme...?

packages/create-turbo/src/commands/create/createProject.ts
```diff
@@ -0,0 +1,231 @@
...
+    error("Unable to create project directory");
+    console.error(err);
+    process.exit(1);
+  }
+  const { isEmpty, conflicts } = isFolderEmpty(root);
+  if (!isEmpty) {
+    error(
+      `${chalk.dim(root)} has ${conflicts.length} conflicting ${
+        conflicts.length === 1 ? "file" : "files"
+      } - please try a different location`
```
mehulkar, line 122: typically not a fan of this pattern. Phrasing so you can just have one way of saying things is preferable. E.g.

```
Please try a different location. Number of conflicting files found: <number>
```

It's not that much less friendly of an error, but reduces complexity (not to mention it's more machine readable, and easier to parse for people who don't speak english and just want to see the number than parse grammar)

packages/create-turbo/src/commands/create/createProject.ts
```diff
@@ -0,0 +1,231 @@
...
+  const originalDirectory = process.cwd();
+  process.chdir(root);
+
+  /**
+   * clone the example repository
+   */
+  const loader = turboLoader("Downloading files...");
+  try {
+    if (repoInfo) {
+      const repoInfo2 = repoInfo;
```
arlyon, line 136: What is the motivation here? They are references to the same object no?

packages/create-turbo/src/commands/create/createProject.ts
```diff
@@ -0,0 +1,231 @@
...
+        // ignore
+      }
+
+      // read the scripts from the package.json
+      availableScripts.push(...Object.keys(packageJsonContent.scripts || {}));
+    }
+  }
+
+  // Copy `.gitignore` if the application did not provide one
+  addGitIgnore(ignorePath);
```
mehulkar, line 223: hmm, why?  We should respect the template as-is, IMO.

packages/create-turbo/src/commands/create/createProject.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  const rootPackageJsonPath = path.join(root, "package.json");
+  const rootMetaJsonPath = path.join(root, "meta.json");
+  const ignorePath = path.join(root, ".gitignore");
+  const hasPackageJson = fs.existsSync(rootPackageJsonPath);
+  const isTurboExample =
+    repoInfo && repoInfo.username === "vercel" && repoInfo.name === "turbo";
+  const availableScripts = [];
+
+  // remove the meta file from turbo examples
```
mehulkar, line 180: can we model this as a "transformer" that gets applied to known repos? it will be easier to rip it out if we decide we don't want our examples to be fancy and have "compile" time logic to build.

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -0,0 +1,181 @@
...
+      if (idx === 0 || group !== lastGroup) {
+        console.log(chalk.cyan(group));
+      }
+      console.log(
+        ` - ${chalk.bold(title)}${description ? `: ${description}` : ""}`
+      );
+      lastGroup = group;
+    });
+  } else {
+    console.log(chalk.cyan("apps"));
```
arlyon, line 113: I would have thought that in the absence of workspaces we'd want to treat the package as a pseudo-workspace with one item in it. I think itd clean out code up a little.

packages/create-turbo/src/commands/create/createProject.ts
```diff
@@ -0,0 +1,231 @@
...
+    if (packageJsonContent) {
+      if (isDefaultExample) {
+        packageJsonContent.name = projectName;
+      }
+
+      // if we're using a pre-release version of create-turbo, install turbo canary instead of latest
+      const shouldUsePreRelease = semverPrerelease(cliPkgJson.version) !== null;
+      if (shouldUsePreRelease && packageJsonContent?.devDependencies?.turbo) {
+        packageJsonContent.devDependencies.turbo = "canary";
+      }
```
mehulkar, line 207: Also something that can go in a "transformer", so we can make it optional more easily. Personally think that if an example has or does not have turbo in package.json, we should respect it as-is. We can apply special logic to vercel/turbo examples if we need it as a starting point

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -0,0 +1,181 @@
...
+        interactive: false,
+      },
+    });
+    loader.stop();
+  }
+
+  // once we're done moving things around, init a new repo
+  tryGitInit(root);
+
+  if (projectDirIsCurrentDir) {
```
arlyon, line 137: Love this attention to detail.

packages/create-turbo/src/utils/examples.ts
```diff
@@ -0,0 +1,135 @@
...
+export async function isUrlOk(url: string): Promise<boolean> {
+  try {
+    const res = await got.head(url);
+    return res.statusCode === 200;
+  } catch (err) {
+    return false;
+  }
+}
+
+export async function getRepoInfo(
```
mehulkar, line 27: The github url parsing logic here seems like something we can find from an existing package 🤔 . I've personally implemented this so many times in so many languages and I always feel like it's an easy place to save some intenral LOC. It's not a lot of code and sunk cost etc etc, but just a suggestion

packages/create-turbo/src/utils/examples.ts
```diff
@@ -0,0 +1,135 @@
...
+    (t === "" && _branch === undefined)
+  ) {
+    try {
+      const infoResponse = await got(
+        `https://api.github.com/repos/${username}/${name}`
+      );
+      const info = JSON.parse(infoResponse.body);
+      return { username, name, branch: info["default_branch"], filePath };
+    } catch (err) {
+      return;
```
arlyon, line 52: Above, if this fails, we error out that the url is invalid. We should probably change that.

packages/create-turbo/src/utils/examples.ts
```diff
@@ -0,0 +1,135 @@
...
+): Promise<RepoInfo | undefined> {
+  const [, username, name, t, _branch, ...file] = url.pathname.split("/");
+  const filePath = examplePath
+    ? examplePath.replace(/^\//, "")
+    : file.join("/");
+
+  if (
+    // Support repos whose entire purpose is to be a Turborepo example, e.g.
+    // https://github.com/:username/:my-cool-turborepo-example-repo-name.
+    t === undefined ||
```
arlyon, line 39: I don't like this var name

packages/create-turbo/src/utils/isFolderEmpty.ts
```diff
@@ -0,0 +1,37 @@
...
+  "docs",
+  "mkdocs.yml",
+  "npm-debug.log",
+  "yarn-debug.log",
+  "yarn-error.log",
+  "yarnrc.yml",
+  ".yarn",
+];
+
+export function isFolderEmpty(root: string): {
```
mehulkar, line 26: I missed where this is used, but the `VALID_FILES` seems like a heuristic that's just waiting to break

packages/create-turbo/src/utils/examples.ts
```diff
@@ -0,0 +1,135 @@
...
+  return tempFile;
+}
+
+export async function downloadAndExtractRepo(
+  root: string,
+  { username, name, branch, filePath }: RepoInfo
+) {
+  const tempFile = await downloadTar(
+    `https://codeload.github.com/${username}/${name}/tar.gz/${branch}`,
+    `turbo-ct-example`
```
arlyon, line 103: Is this correct?

packages/create-turbo/src/utils/examples.ts
```diff
@@ -0,0 +1,135 @@
...
+      ),
+  });
+
+  await fs.unlink(tempFile);
+}
+
+export async function downloadAndExtractExample(root: string, name: string) {
+  const tempFile = await downloadTar(
+    `https://codeload.github.com/vercel/turbo/tar.gz/main`,
+    `turbo-ct-example`
```
arlyon, line 124: Same again.

packages/create-turbo/src/utils/git.ts
```diff
@@ -0,0 +1,88 @@
+import fs from "fs-extra";
+import { execSync } from "child_process";
+import path from "path";
+import rimraf from "rimraf";
+
+export const DEFAULT_IGNORE = `
```
arlyon, line 6: It would be nice if this was somewhere else, idk how we'd bundle it though... just handy to have for others. There is a repo that hosts .gitignore files maybe we can consider opening a pr there and just downloading it (or a comment with a link to it) so we have a canonical version.

mehulkar (commented):
Left a first pass review. My main blocking comment is that if `npx create-turbo -e <something>` should respect `<something>` as-is. 

If we _absolutely have_ to transform it somehow, I suggest two things:

- Ensure that all transformations are additional git commits after the first (if we're initializing a git repo in the first place)
- Model transformation code so that we can make it optional (we can do this in follow up work, but I think it would be better to do first time around, even if it's just moving the logic to a new function in a separate file). 

As an aside: I somewhat like `degit`'s approach of using templates. Do we want to support using `create-turbo` without initializing git repo?

arlyon (dismissed):
A few comments and questions. I will say generally attention to detail / ux is exceptional 😍

packages/create-turbo/package.json
tknickman: This is technically the "latest" published version, and it will get overridden post merge and deploy of the next version so it wouldn't be breaking, but there's also no harm in removing it either, will revert.

packages/create-turbo/src/cli.ts
tknickman: Yea agreed, I have on my list to go through and consolidate the logging from the different packages

packages/create-turbo/src/cli.ts
tknickman: I didn't add one so we would match `create-next-app` exactly, but not opposed to an alias here.

packages/create-turbo/src/commands/create/createProject.ts
tknickman: We are yes

packages/create-turbo/src/commands/create/createProject.ts
tknickman: This is straight from Next, I'll look into improving it

packages/create-turbo/src/commands/create/createProject.ts
tknickman: Missing `.turbo` in `.gitignore` is a huge pain point, I think it's very reasonable to include this if one is missing. `create-next-app` also does this.

packages/create-turbo/src/commands/create/createProject.ts
tknickman: We could, but I'm not sure if that's worth the time right now when we would only have one transformer. I'm usually a fan of optimizing later when the patterns emerge. This is a pretty straight-forward case. The example is either one of ours (in which case we have a few special rules) or it's not (we don't do anything)

packages/create-turbo/src/commands/create/createProject.ts
tknickman: I can restrict this to our examples only, but this is also existing behavior that we rely on for testing.

packages/create-turbo/src/utils/examples.ts
tknickman: This is also from create-next-app

packages/create-turbo/src/utils/isFolderEmpty.ts
tknickman: This is also from create-next-app

packages/create-turbo/__tests__/isWritable.test copy.ts
tknickman: Gooood catch, thanks!

packages/create-turbo/src/commands/create/createProject.ts
tknickman: Hmm good call, we could support both

packages/create-turbo/src/commands/create/createProject.ts
tknickman: This was lifted right from create-next-app, but I'll clean it up a bit

packages/create-turbo/src/utils/examples.ts
tknickman: I agree 😄

packages/create-turbo/src/utils/git.ts
tknickman: Yes, what we really care about here is .turbo, but this is a good call, maybe we can augment another shared one.

tknickman:
@mehulkar good call RE the transformations, maybe we can support a `--skip-transform` option or something that doesn't prompt you for the package manager and skips the rest of the transforms as well

packages/create-turbo/src/commands/create/index.ts
tknickman: Yea I was doing that, but since we do support non-monorepos as first party users of turbo I didn't want to make them also look like monorepos

packages/create-turbo/src/commands/create/createProject.ts
tknickman: If we've gotten this far we know the directory exists (that's validated on input)

packages/create-turbo/src/commands/create/createProject.ts
tknickman: 👍🏼 moved this into a function to keep a single source of truth

packages/create-turbo/src/utils/git.ts
tknickman: I think I'm going to punt this one for now, and possible come back to it when we have support for dynamic transforms that could take care of this in a more repeatable way.

packages/create-turbo/src/utils/examples.ts
tknickman: 👍🏼 updated the error

tknickman:
@mehulkar we know have lite transformers. 

This added a little more complexity because it introduces the case when you have specified skip-transform, but the example doesn't use a package manager that you have installed. So we have to handle the downstream effects of that (disabling install, etc)

packages/create-turbo/src/cli.ts
```diff
@@ -0,0 +1,65 @@
...
+import notifyUpdate from "./utils/notifyUpdate";
+import { turboGradient } from "./logger";
+
+import { create } from "./commands";
+import cliPkg from "../package.json";
+
+const createTurboCli = new Command();
+
+// create
+createTurboCli
```
mehulkar, line 14: Do we want a `--skip-git` option?

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -0,0 +1,231 @@
...
+  // selected package manager can be undefined if the user chooses to skip transforms
+  const selectedPackageManagerDetails = await prompts.packageManager({
+    packageManager,
+    skipTransforms,
+  });
+
+  if (packageManager && opts.skipTransforms) {
+    warn(
+      '"--skip-transforms" flag conflicts with the package manager argument. The package manager argument will be ignored.'
+    );
```
mehulkar, line 67: You can use the `.conflicts()` method in commander instead of this https://github.com/tj/commander.js/blob/HEAD/examples/options-conflicts.js

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -0,0 +1,231 @@
...
+        }
+      } catch (err) {
+        handleWorkspaceErrors(err);
+      }
+    }
+  }
+
+  if (successfulTransforms > 0) {
+    // create a second commit after running transforms
+    tryGitCommit("feat(create-turbo): apply transforms");
```
mehulkar, line 116: 💯 , but a commit after every transform would be preferable, IMO

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -0,0 +1,231 @@
...
+      warn(
+        `Try running without "--skip-transforms" to convert "${exampleName}" to a package manager that is available on your system.`
+      );
+      console.log();
+    } else if (selectedPackageManagerDetails) {
+      console.log("Installing packages. This might take a couple of minutes.");
+      console.log();
+
+      const loader = turboLoader("Installing dependencies...").start();
+      await install({
```
mehulkar, line 172: Does installing dependencies create a lockfile or is there already one? In either case, I think this is also a transform. (Especially if we every want to think about `turbo install` as a feature)

packages/create-turbo/src/transforms/index.ts
```diff
@@ -0,0 +1,13 @@
+import { transform as packageManagerTransform } from "./package-manager";
+import { transform as internalTransform } from "./internal-example";
+import { transform as gitIgnoreTransform } from "./git-ignore";
+import type { TransformInput, TransformResult } from "./types";
+
+/**
+ * In the future, we may want to support sourcing additional transforms from the templates themselves.
+ */
+export const transforms: Array<(args: TransformInput) => TransformResult> = [
+  internalTransform,
```
mehulkar, line 10: How about `coreExample`? "internal is a bit unclear to me" since it's not internal, we're intending external usage

packages/create-turbo/src/transforms/internal-example.ts
```diff
@@ -0,0 +1,62 @@
+import path from "path";
+import fs from "fs-extra";
+import semverPrerelease from "semver/functions/prerelease";
+import cliPkgJson from "../../package.json";
+import { isDefaultExample } from "../utils/isDefaultExample";
+import { TransformInput, TransformResult } from "./types";
+
+// applied to "local" examples (those hosted within vercel/turbo/examples)
```
mehulkar, line 8:
```suggestion
// applied to "core" examples (those hosted within vercel/turbo/examples)
```

packages/create-turbo/src/transforms/package-manager.ts
```diff
@@ -0,0 +1,20 @@
+import { convert } from "@turbo/workspaces";
+import { TransformInput, TransformResult } from "./types";
+
+export async function transform(args: TransformInput): TransformResult {
+  const { project, prompts } = args;
+  const { root, packageManager } = prompts;
+
+  if (packageManager && project.packageManager !== packageManager.name) {
+    await convert({
```
mehulkar, line 9: 👏🏾 nice

packages/create-turbo/src/utils/examples.ts
mehulkar: can we publish it as a library and use it instead of copy pasting?

packages/create-turbo/src/utils/isDefaultExample.ts
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_EXAMPLES = new Set(["basic", "default"]);
+
+export function isDefaultExample(example: string): boolean {
+  return DEFAULT_EXAMPLES.has(example);
+}
```
mehulkar, line 5: Seems very specific to the "internal-transform", so putting it in a general utility seems like an unnecessary indirection

packages/create-turbo/src/commands/create/index.ts
tknickman: I did look at that originally, but package manager is an argument, and skip-transform is an option, and you can't specify conflicts between the two.

packages/create-turbo/src/commands/create/index.ts
tknickman: Yea I did think about this, and It does produce a lock if one doesn't exist. However, I would argue that's not a transform, it's just a by-product of what you asked for via arguments. There's already another option to enable / disable this behavior (`skip-install`)

packages/create-turbo/src/transforms/index.ts
tknickman: internal means internal to the repo. It's an example that comes from `vercel/turbo`. I like internal a bit more than core because it carries with it some information about the example itself (I don't know what a core example is just by looking at the name), but I can try and come up with something more formal.

packages/create-turbo/src/utils/examples.ts
tknickman: Yep, i'm planning to talk to the next team about this to see about a core-create or similar we could all utilize and open source.

packages/create-turbo/src/utils/isDefaultExample.ts
tknickman: Yea good point, I'll clean this up a bit more

PR #4399, opened 2023-03-30
support require.context

ForsakenHarmony:
fix WEB-535


crates/turbopack-ecmascript/src/references/require_context.rs
```diff
@@ -0,0 +1,267 @@
...
+            DirectoryContent::Entries(entries) => entries,
+            DirectoryContent::NotFound => continue,
+        };
+
+        for (_, entry) in entries {
+            match entry {
+                DirectoryEntry::File(path) => {
+                    if let Some(relative_path) = root.get_relative_path_to(&*path.await?) {
+                        if filter.is_match(&relative_path) {
+                            list.push((relative_path, *path));
```
sokra, line 67: It need to apply some resolve options here. e. g. `extensions`.

e.g. a file `./abc/index.ts` should have multiple keys: `./abc/index.ts` `./abc/index` `./abc/` or `./abc`

a package.json with `main` or `exports` would also influence that. (webpack doesn't handle that case)

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2852,6 +2873,97 @@ impl WellKnownObjectKind {
...
+        }
+    }
+
+    let regex = if !applied_flags.is_empty() {
+        format!("(?{}){}", applied_flags, pattern)
+    } else {
+        pattern
+    };
+
+    Regex::new(&regex).map_err(|_| "could not convert javascript regex to rust regex")
```
alexkirsz, line 2912:
```suggestion
    Regex::new(&regex).map_err(|_| "could not convert ECMAScript regex to Rust regex")
```

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2852,6 +2873,97 @@ impl WellKnownObjectKind {
...
 
+#[derive(Debug, Clone)]
+pub struct RequireContextOptions {
+    pub dir: String,
+    pub include_subdirs: bool,
+    /// this is a regex (pattern, flags)
+    pub filter: Regex,
+}
+
+pub fn regex_from_js(pattern: &str, flags: &str) -> Result<Regex, &'static str> {
```
alexkirsz, line 2884: Would this always error if a JS regex is not representable as a Rust regex, or are there some inconsistencies between the two? e.g. some syntax that means something as a JS regex but a different thing as a rust regex.

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2852,6 +2873,97 @@ impl WellKnownObjectKind {
...
+    } else {
+        pattern
+    };
+
+    Regex::new(&regex).map_err(|_| "could not convert javascript regex to rust regex")
+}
+
+pub fn parse_require_context(args: &Vec<JsValue>) -> Result<RequireContextOptions, &'static str> {
+    if !(1..=3).contains(&args.len()) {
+        return Err("only 1-3 arguments are supported (mode is not supported)");
```
alexkirsz, line 2917: Do we have a task to support "mode"? If so, link it here.

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2852,6 +2873,97 @@ impl WellKnownObjectKind {
...
+    let filter = if let Some(filter) = args.get(2) {
+        if let JsValue::Constant(ConstantValue::Regex(pattern, flags)) = filter {
+            regex_from_js(pattern, flags)?
+        } else {
+            return Err("filter needs to be a regex");
+        }
+    } else {
+        // https://webpack.js.org/api/module-methods/#requirecontext
+        // > optional, default /^\.\/.*$/, any file
+        Regex::new("^\\./.*$").expect("valid static regex should always compile")
```
alexkirsz, line 2943: Make this a lazy static so we don't recompile it on every call.

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -725,6 +730,52 @@ pub(crate) async fn analyze_ecmascript_module(
...
+                        if (1..=3).contains(&args.len()) {
+                            let options = match parse_require_context(&args) {
+                                Ok(options) => options,
+                                Err(reason) => {
+                                    let (args, hints) = explain_args(&args);
+                                    handler.span_err_with_code(
+                                        span,
+                                        &format!(
+                                            "require.context({args}) is not statically \
+                                             analyse-able: {reason}{hints}",
```
sokra, line 744: When `reason` is an anyhow Error you can use `PrettyPrintError` here

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2852,6 +2873,97 @@ impl WellKnownObjectKind {
...
+        filter,
+    })
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RequireContextValue {
+    pub(crate) map: IndexMap<String, String>,
+}
+
+impl Hash for RequireContextValue {
```
alexkirsz, line 2958: Can't be derived?

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2852,6 +2873,97 @@ impl WellKnownObjectKind {
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        for (k, v) in self.map.iter() {
```
alexkirsz, line 2960: Should hash something even in the case it's empty (e.g. the length) so consecutive values don't hash the same [{}, { k: 1 }], [{ k: 1 }, {}].

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2852,6 +2873,97 @@ impl WellKnownObjectKind {
...
+    let regex = if !applied_flags.is_empty() {
+        format!("(?{}){}", applied_flags, pattern)
+    } else {
+        pattern
+    };
+
+    Regex::new(&regex).map_err(|_| "could not convert javascript regex to rust regex")
+}
+
+pub fn parse_require_context(args: &Vec<JsValue>) -> Result<RequireContextOptions, &'static str> {
```
alexkirsz, line 2915: Comment this public item.

crates/turbopack-ecmascript/src/analyzer/mod.rs
alexkirsz, line 2884: Comment this public item.

crates/turbopack-ecmascript/src/analyzer/well_known.rs
```diff
@@ -337,6 +345,99 @@ pub fn require(args: Vec<JsValue>) -> JsValue {
     }
 }
 
+pub fn require_context_require(val: RequireContextValue, args: Vec<JsValue>) -> JsValue {
```
alexkirsz, line 348: Comment this public item.

crates/turbopack-ecmascript/src/analyzer/well_known.rs
```diff
@@ -337,6 +345,99 @@ pub fn require(args: Vec<JsValue>) -> JsValue {
...
+    if args.len() < 1 {
```
alexkirsz, line 349:
```suggestion
    if args.is_empty() {
```

or

```suggestion
    if args != 1 {
```

considering your comment below.

crates/turbopack-ecmascript/src/analyzer/well_known.rs
```diff
@@ -337,6 +345,99 @@ pub fn require(args: Vec<JsValue>) -> JsValue {
...
+        );
+    }
+
+    let Some(s) = args[0].as_str() else {
+        return JsValue::Unknown(
+            Some(Arc::new(JsValue::call(
+                box JsValue::WellKnownFunction(WellKnownFunctionKind::RequireContextRequire(val)),
+                args,
+            ))),
+            "only constant argument is supported",
```
alexkirsz, line 365:
```suggestion
            "require.context(...).require() only accepts a single, constant string argument",
```

crates/turbopack-ecmascript/src/analyzer/well_known.rs
```diff
@@ -337,6 +345,99 @@ pub fn require(args: Vec<JsValue>) -> JsValue {
...
+    };
+
+    JsValue::Module(ModuleValue {
+        module: m.to_string().into(),
+        annotations: ImportAnnotations::default(),
+    })
+}
+
+pub fn require_context_require_keys(val: RequireContextValue, args: Vec<JsValue>) -> JsValue {
+    if args.len() == 0 {
```
alexkirsz, line 386:
```suggestion
    if args.is_empty() {
```

I'm surprised Clippy did not complain?

crates/turbopack-ecmascript/src/analyzer/well_known.rs
```diff
@@ -337,6 +345,99 @@ pub fn require(args: Vec<JsValue>) -> JsValue {
...
+        JsValue::array(val.map.keys().cloned().map(|k| k.into()).collect())
+    } else {
+        JsValue::Unknown(
+            Some(Arc::new(JsValue::call(
+                box JsValue::WellKnownFunction(WellKnownFunctionKind::RequireContextRequireKeys(
+                    val,
+                )),
+                args,
+            ))),
+            "no arguments are supported",
```
alexkirsz, line 396:
```suggestion
            "require.context(...).keys() does not accept arguments",
```

crates/turbopack-ecmascript/src/analyzer/well_known.rs
```diff
@@ -337,6 +345,99 @@ pub fn require(args: Vec<JsValue>) -> JsValue {
...
 }
 
+pub fn require_context_require(val: RequireContextValue, args: Vec<JsValue>) -> JsValue {
+    if args.len() < 1 {
+        return JsValue::Unknown(
+            Some(Arc::new(JsValue::call(
+                box JsValue::WellKnownFunction(WellKnownFunctionKind::RequireContextRequire(val)),
+                args,
+            ))),
+            "only a single argument is supported",
```
alexkirsz, line 355:
```suggestion
            "require.context(...).require() only accepts a single, constant string argument",
```

crates/turbopack-ecmascript/src/analyzer/well_known.rs
```diff
@@ -337,6 +345,99 @@ pub fn require(args: Vec<JsValue>) -> JsValue {
...
+                    val,
+                )),
+                args,
+            ))),
+            "no arguments are supported",
+        )
+    }
+}
+
+pub fn require_context_require_resolve(val: RequireContextValue, args: Vec<JsValue>) -> JsValue {
```
alexkirsz, line 401: Comment this public item.

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -1870,53 +1921,12 @@ async fn value_visitor_inner(
...
-                    "only a single argument is supported",
-                )
-            }
-        }
+        ) => require_resolve_visitor(origin, args, in_try).await?,
+        JsValue::Call(
+            _,
+            box JsValue::WellKnownFunction(WellKnownFunctionKind::RequireContext),
+            args,
+        ) => require_context_visitor(origin, args, in_try).await?,
```
sokra, line 1929: static evaluation for require.context... You are really (over)killing it...

webpack doesn't even support that.

So I can do something fancy like:

```js
const context = require.context("...");
const resolved = context.resolve("./a");
const imported = require(resolved); // <- usually this would be too dynamic
```

crates/turbopack-ecmascript/src/analyzer/well_known.rs
```diff
@@ -337,6 +345,99 @@ pub fn require(args: Vec<JsValue>) -> JsValue {
...
+    if args.len() != 1 {
+        return JsValue::Unknown(
+            Some(Arc::new(JsValue::call(
+                box JsValue::WellKnownFunction(
+                    WellKnownFunctionKind::RequireContextRequireResolve(val),
+                ),
+                args,
+            ))),
+            "only a single argument is supported",
```
alexkirsz, line 410:
```suggestion
            "require.context(...).resolve() only accepts a single, constant string argument",
```

sokra (approved):
The approach is a bit inefficient from incremental compilation point of view, but we can move that to tech dept:

* adding/removing a file in the root directory with walk the whole directory again (list_dir is recomputed)
    * This could be solved by using a tree reprensentation instead of `FlatDirList`
* The module using `require.context` need to be re-analyzed once a file is added/removed from the context
    * This could be solved by moving the context map into a separate module
    * And computing the require.context for static analysis only when a key is accessed from it. And making this a dedicated tt function

crates/turbopack-ecmascript/src/analyzer/well_known.rs
```diff
@@ -337,6 +345,99 @@ pub fn require(args: Vec<JsValue>) -> JsValue {
...
+
+    let Some(s) = args[0].as_str() else {
+        return JsValue::Unknown(
+            Some(Arc::new(JsValue::call(
+                box JsValue::WellKnownFunction(
+                    WellKnownFunctionKind::RequireContextRequireResolve(val),
+                ),
+                args,
+            ))),
+            "only constant argument is supported",
```
alexkirsz, line 422:
```suggestion
            "require.context(...).resolve() only accepts a single, constant string argument",
```

crates/turbopack-ecmascript/src/references/mod.rs
alexkirsz, line 744: Unrelated to this PR: we should switch to american english: https://linear.app/vercel/issue/WEB-890/convert-turbopack-errors-and-messaging-to-american-english

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -725,6 +730,52 @@ pub(crate) async fn analyze_ecmascript_module(
                         )
                     }
 
+                    JsValue::WellKnownFunction(WellKnownFunctionKind::RequireContext) => {
+                        let args = linked_args(args).await?;
+                        if (1..=3).contains(&args.len()) {
```
alexkirsz, line 735: Why do we validate this again here if it's also done in `parse_require_context`?

Could we catch the error and show it with `span_err_with_code`?

crates/turbopack-ecmascript/src/references/require_context.rs
```diff
@@ -0,0 +1,267 @@
...
+
+#[turbo_tasks::value(transparent)]
+pub(crate) struct FlatDirList(Vec<(String, FileSystemPathVc)>);
+
+#[turbo_tasks::function]
+pub(crate) async fn list_dir(
+    dir: FileSystemPathVc,
+    recursive: bool,
+    filter: RegexVc,
+) -> Result<FlatDirListVc> {
```
alexkirsz, line 46: Can this be a static method on `FlatDirListVc`? `FlatDirListVc::new`

crates/turbopack-ecmascript/src/references/require_context.rs
```diff
@@ -0,0 +1,267 @@
...
+    let root = &*dir.await?;
+    let filter = &*filter.await?;
+
+    let mut queue = VecDeque::with_capacity(1);
+    queue.push_back(dir);
```
alexkirsz, line 51: Nit:
```suggestion
    let mut queue: VecDeque<_> = [dir].collect();
```

crates/turbopack-ecmascript/src/references/require_context.rs
```diff
@@ -0,0 +1,267 @@
...
+                    }
+                }
+                DirectoryEntry::Directory(path) if recursive => {
+                    queue.push_back(*path);
+                }
+                // ignore everything else
+                _ => {}
+            }
+        }
+    }
```
alexkirsz, line 78: I believe `read_dir` does not return entries in a consistent ordering, so we might have to sort them here instead (or at the `read_dir` callsite). You could also use `GraphTraversal` here but it's likely overkill.

crates/turbopack-ecmascript/src/references/require_context.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub origin_relative: String,
+    pub request: RequestVc,
+    pub result: ResolveResultVc,
+}
+
+#[turbo_tasks::value(transparent)]
+pub(crate) struct RequireContextMap(IndexMap<String, RequireContextMapEntry>);
+
+#[turbo_tasks::function]
+pub(crate) async fn generate_require_context_map(
```
alexkirsz, line 95: This could be a static method on `RequireContextMapVc`

crates/turbopack-ecmascript/src/references/require_context.rs
```diff
@@ -0,0 +1,267 @@
...
+            map.insert(
+                context_relative.clone(),
+                RequireContextMapEntry {
+                    origin_relative,
+                    request,
+                    result,
+                },
+            );
+        } else {
+            // TODO: does this need to throw an error?
```
alexkirsz, line 123: Is this even possible? If it's theoretically impossible, we should throw an invariant error here at least.

crates/turbopack-ecmascript/src/references/require_context.rs
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+
+    Ok(RequireContextMapVc::cell(map))
+}
+
+#[turbo_tasks::value]
+#[derive(Hash, Debug)]
+pub struct CjsRequireContextAssetReference {
```
alexkirsz, line 132: Comment this public item.

alexkirsz (dismissed):

crates/turbopack-ecmascript/src/references/mod.rs
ForsakenHarmony: yea, the static evaluation was the reason I didn't put the context into a separate asset/module

crates/turbopack-ecmascript/src/analyzer/mod.rs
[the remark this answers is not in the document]
ForsakenHarmony: this is also because of the static analysis because `JsValue::Unknown` only supports `&'static str`

crates/turbopack-dev/js/src/runtime.js
```diff
@@ -187,6 +189,61 @@ function commonJsRequire(sourceModule, id) {
   return module.exports;
 }
 
+/**
+ * @param {Module} sourceModule
+ * @param {RequireContextMap} map
+ * @returns {RequireContext}
+ */
+function webpackRequireContext(sourceModule, map) {
```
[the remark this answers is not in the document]
ForsakenHarmony: that was because it's a webpack feature, but I'll remove it

crates/turbopack-ecmascript/src/references/require_context.rs
ForsakenHarmony: How would the `package.json` influence this?
The API seems to be file path based, so webpack's behavior kinda makes sense to me

crates/turbopack-ecmascript/src/analyzer/mod.rs
ForsakenHarmony: no, `IndexMap` doesn't implement `Hash`

crates/turbopack-ecmascript/src/analyzer/well_known.rs
ForsakenHarmony: The reason it's not `!= 1` is that a common use-case seems to be
```js
function importAll(r) {
  return r.keys().map(r);
}

const items = importAll(require.context("./"));
```
Where it's passed directly into the `.map` which would add more arguments (the index and array)

crates/turbopack-ecmascript/src/references/require_context.rs
ForsakenHarmony: I don't like that at all, but `VecDeque::from([dir])` seems to work

crates/turbopack-ecmascript/src/analyzer/mod.rs
ForsakenHarmony: I think the "base" regex features are all the same and if anything there might be something that's supported only in rust or only in js

crates/turbopack-ecmascript/src/analyzer/mod.rs
ForsakenHarmony: created one

crates/turbopack-ecmascript/src/analyzer/mod.rs
ForsakenHarmony: will do

crates/turbopack-ecmascript/src/analyzer/mod.rs
ForsakenHarmony: hmm, I can add the length, but I don't think it would prevent your example from hashing differently

crates/turbopack-ecmascript/src/analyzer/mod.rs
ForsakenHarmony: done

crates/turbopack-ecmascript/src/references/mod.rs
ForsakenHarmony: good point, fixed

crates/turbopack-ecmascript/src/references/require_context.rs
ForsakenHarmony: ah yes, I wanted to sort, but forgot about it

crates/turbopack-ecmascript/src/references/require_context.rs
ForsakenHarmony: I guess, but I'm not sure if it improves the API 🤔

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -369,7 +417,7 @@ pub enum JsValue {
     WellKnownFunction(WellKnownFunctionKind),
     /// Not-analyzable value. Might contain the original value for additional
     /// info. Has a reason string for explanation.
-    Unknown(Option<Arc<JsValue>>, &'static str),
+    Unknown(Option<Arc<JsValue>>, JsValueUnknownReason),
```
sokra, line 420: Could this be a `Cow<'static, str>`

PR #4400, opened 2023-03-30
ci(workflow): run next-swc check with latest turbopack

kwonoj:
### Description

This PR enables our next-swc check runs against with PR's change so we can verify if PR's change will break next-swc or not.

Part of WEB-763.


NicholasLYang (approved):
Looks good on my end

padmaia (approved):

PR #4403, opened 2023-03-30
Refactor Dry Run so that output is printed by Run Summaries

mehulkar:
This is pre-work for https://github.com/vercel/turbo/pull/4394. 

The goal is to make `RunSummary` responsible for all the JSON and text and saved versions of what happened in a Run through the `.Close()` interface. This PR incorporates DryRun output into this mold. 

Reviewing one commit at a time will hopefully make the entire diff a bit easier to understand, but generally:

1. Give `RunSummary` all of `runOpts`

    we were passing in a lot of pieces already, and I wanted yet another flag for dry run.
    To avoid circular imports, I had to move `runOpts` into the `util` package.
1. Use `.Close()` to print dry run output
1. Some refactor around making DryRun execution look more like Real Run execution. 

One other benefit of making it work this way is that we may soon be in a good spot to move RunSummaries into Rust.

### DIFF
--- a/cli/internal/run/dry_run.go
+++ b/cli/internal/run/dry_run.go
@@ -35,7 +35,7 @@ func DryRun(
 	taskSummaries := []*runsummary.TaskSummary{}
 
 	mu := sync.Mutex{}
-	dryRunExecFunc := func(ctx gocontext.Context, packageTask *nodes.PackageTask, taskSummary *runsummary.TaskSummary) error {
+	execFunc := func(ctx gocontext.Context, packageTask *nodes.PackageTask, taskSummary *runsummary.TaskSummary) error {
 		// Assign some fallbacks if they were missing
 		if taskSummary.Command == "" {
 			taskSummary.Command = runsummary.MissingTaskLabel
@@ -61,7 +61,7 @@ func DryRun(
 		return rs.ArgsForTask(taskID)
 	}
 
-	visitorFn := g.GetPackageTaskVisitor(ctx, engine.TaskGraph, getArgs, base.Logger, dryRunExecFunc)
+	visitorFn := g.GetPackageTaskVisitor(ctx, engine.TaskGraph, getArgs, base.Logger, execFunc)
 	execOpts := core.EngineExecutionOptions{
 		Concurrency: 1,
 		Parallel:    false,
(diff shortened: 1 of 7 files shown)


cli/internal/run/dry_run.go
```diff
@@ -44,8 +45,11 @@ func DryRun(
 			taskSummary.Framework = runsummary.MissingFrameworkLabel
 		}
 
+		// This mutex is not _really_ required, since we are using Concurrency: 1 as an execution
+		// option, but we add it here to match the shape of RealRuns execFunc.
```
chris-olszewski, line 47: Not a strong feeling, but this feels unnecessary. I think a comment above `Concurrency: 1` would be just fine.

cli/internal/run/dry_run.go
```diff
@@ -80,7 +80,8 @@ func DryRun(
 	// Assign the Task Summaries to the main summary
 	summary.RunSummary.Tasks = taskSummaries
 
-	return summary.CloseDryRun(g.WorkspaceInfos, rs.Opts.runOpts.dryRunJSON)
+	// This is terrible, but pass a 0 exitCode into Close()
+	return summary.Close(0, g.WorkspaceInfos)
```
chris-olszewski, line 84: Alternative would be to have `Close` take a `*int` so we could pass `nil` here. After recent exit code shenanigans, I'd prefer if we don't have fake successful exit codes getting created (even if they shouldn't ever end up getting out). I think bare minimum we should choose a more obviously fake exit code like -1.

cli/internal/run/real_run.go
```diff
@@ -163,7 +163,9 @@ func RealRun(
 		}
 	}
 
-	runSummary.Close(exitCode)
+	if err := runSummary.Close(exitCode, g.WorkspaceInfos); err != nil {
+		// Ignore the error. All errors are swallowed for real runs
```
chris-olszewski: Would it be worth warning if there was an error flushing the summary?

chris-olszewski (approved):
LGTM!

NicholasLYang (approved):
Makes sense to me

cli/internal/run/dry_run.go
mehulkar: my reasoning is:

- the code isn't actually being used downstream (it's not being used by the method, and if it were, the `executionSummary` will not be part of the dry run output)
- even if it was being used, it really _is_ a successful exit of the parent task, so we can see how it's ok. I could use `the `successCode` 

After thinking about it, I think it's the comment I should change/remove rather than the code

PR #4413, opened 2023-03-31
Account for canary versions in integration test for run summary

mehulkar:


cli/integration_tests/single_package/dry-run.t
```diff
@@ -33,7 +33,7 @@ Check
   {
     "id": "[a-zA-Z0-9]+", (re)
     "version": "0",
-    "turboVersion": "[0-9\.]+", (re)
+    "turboVersion": "[a-z0-9\.]+", (re)
```
NicholasLYang, line 36: Does this regex also accept the `-` in the version?

chris-olszewski (approved):

cli/integration_tests/single_package/dry-run.t
mehulkar: blah, I should use what we already have and is working 
https://github.com/vercel/turbo/blob/a5427a7904b41e523125db3a3b7bbcddcd469c39/cli/integration_tests/basic_monorepo/dry_run_json.t#L8-L10

mehulkar: https://regex101.com/r/TnLKKL/1

PR #4414, opened 2023-03-31
mark resolve issues inside a try block as warnings

ForsakenHarmony:
### Description

I also added source information to the resolve issues where possible


padmaia:
Should it be a warning or should it just be accepted?

crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -1478,6 +1464,42 @@ impl VisitAstPath for Analyzer<'_> {
...
+
+        if let Some(finalizer) = &stmt.finalizer {
+            ast_path.with(
+                AstParentNodeRef::TryStmt(stmt, TryStmtField::Finalizer),
+                |ast_path| {
+                    finalizer.visit_with_path(self, ast_path);
+                },
+            )
+        }
+    }
```
sokra, line 1501: You might also compute the `in_try` flag from `ast_path`.

Should we handle function declarations/expressions, etcs. in try blocks:

``` js
try {
  // in try here
  function x() {
    // not in try here?
  }
  // in try here
} ...
```

crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -115,13 +115,15 @@ pub enum Effect {
         /// The ast path to the condition.
         ast_path: Vec<AstParentKind>,
         span: Span,
+        in_try: bool,
```
sokra, line 118: I guess that doesn't need the flag, but nevermind...

crates/turbopack-ecmascript/tests/analyzer/graph/try/input.js
```diff
@@ -0,0 +1,9 @@
+let pkg;
+
+try {
+  pkg = require("packages/not-found");
+} catch (e) {
+  pkg = require("packages/found");
+}
+
+pkg.fn();
```
sokra, line 9: Probably need some snapshots and updates to the existing snapshots

sokra (approved):

crates/turbopack-ecmascript/src/analyzer/graph.rs
ForsakenHarmony: Didn't think of that, changing it to derive it from the `ast_path`

crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -618,6 +616,17 @@ pub fn as_parent_path_with(
...
 
+pub fn is_in_try(ast_path: &AstNodePath<AstParentNodeRef<'_>>) -> bool {
+    ast_path
+        .iter()
+        .fold(false, |in_try, ast_ref| match ast_ref.kind() {
+            AstParentKind::ArrowExpr(ArrowExprField::Body) => false,
+            AstParentKind::Function(FunctionField::Body) => false,
+            AstParentKind::TryStmt(TryStmtField::Block) => true,
+            _ => in_try,
+        })
```
sokra, line 627: `rev().find_map()` might also work well and might be a bit faster

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map

crates/turbopack-ecmascript/src/analyzer/graph.rs
ForsakenHarmony: Unfortunately, it is actually used

PR #4415, opened 2023-03-31
remove process, make env a source asset

sokra:
### Description

To provide a full process polyfill:

* We need to remove the process hack
* We need to make the ProcessEnvAsset a Source Asset instead of a Module Asset (we want to allow injecting the process polyfill.
* We need to pass source assets instead of module assets as runtime_entries and process them with the AssetContext


crates/turbopack-env/src/asset.rs
```diff
@@ -49,84 +36,14 @@ impl Asset for ProcessEnvAsset {
...
-        let env = asset.env.read_all().await?;
+    async fn content(&self) -> Result<AssetContentVc> {
+        let env = self.env.read_all().await?;
 
         // TODO: In SSR, we use the native process.env, which can only contain string
         // values. We need to inject literal values (to emulate webpack's
         // DefinePlugin), so create a new regular object out of the old env.
-        let mut code = "const env = process.env = {...process.env};\n\n".to_string();
+        let mut code = RopeBuilder::default();
+        code.push_static_bytes(b"const env = process.env = {...process.env};\n\n");
```
jridgewell, line 46: Nit:
```suggestion
        code += "const env = process.env = {...process.env};\n\n";
```

jridgewell (approved):
Snapshot updates?

PR #4423, opened 2023-04-01
Add a debugging section in our docs

mehulkar:


docs/pages/repo/docs/_meta.json
```diff
@@ -4,6 +4,7 @@
   },
   "installing": "Installing Turborepo",
   "getting-started": "Getting Started",
+  "debugging": "Debugging Turborepo",
```
tknickman, line 7: Should we combine with troubleshooting?

arlyon: I think there is SoC but I'm not sure what the naming should be. "Debugging Builds" may be better or, since most of this is around caching simply 'Debugging Caching'/

mehulkar: Good call, I think troubleshooting topics. Of the troubleshooting topics

These two should go somewhere in the Monorepo handbook
- My dependency isn't being built correctly
- My types are not being found

These are relevant to the content here. I'll use the existing trouble shooting page for this content and then move monorepo troubleshooting into the handhook
- I'm not seeing any cache hits
- I'm seeing cache hits, but my build is broken
- My build is caching the wrong environment variables

docs/pages/repo/docs/troubleshooting.mdx
```diff
@@ -5,37 +5,152 @@ description: This guide aims to help you debug issues with your Turborepo builds
...
+One of the [core concepts][2] behind Turbo is that when a declared input
+changes, the cached outputs for that task are invalidated. As part of running any task,
+Turborepo creates the following directories:
+
+- A `.turbo` at the root of your repo
+- A `.turbo` directory in each workspace if your project is a monorepo (e.g. `apps/my-app/.turbo/`)
+- A `turbo` directory in `node_modules/.cache`
 
-## I'm not seeing any cache hits
+Becuase the first two directories are not git-ignored by default, you may see an
```
arlyon, line 82: typo
```suggestion
Because the first two directories are not git-ignored by default, you may see an
```

docs/pages/repo/docs/troubleshooting.mdx
```diff
@@ -5,37 +5,152 @@ description: This guide aims to help you debug issues with your Turborepo builds
...
+issue where you run the same task twice and get a cache missg, even though you
```
arlyon, line 83:
```suggestion
issue where you run the same task twice and get a cache missing, even though you
```

docs/pages/repo/docs/troubleshooting.mdx
```diff
@@ -5,37 +5,152 @@ description: This guide aims to help you debug issues with your Turborepo builds
...
+- Are [cache outputs correctly specified][d-config-outputs] in your Turborepo [pipeline][d-def-pipeline]?
+
+  Pipeline settings are not inherited or merged, so they need to be
+  re-specified in [workspace-specific tasks][d-workspace-tasks] (e.g. `web#build` does
+  **not** inherit pipeline settings from `build`).
+
+- [Are relevant inlined environment variables accounted for?][12]
+
+  To verify, run `turbo` in verbose mode by adding `-vvv` to
+  `turbo run <task>` and look at which environment variables are included in
```
arlyon, line 106: As we move more stuff into rust, -vvv is going to get very very very verbose (about 12 log lines for each grpc request, for example). We may want to reconsider what goes in debug vs trace.

arlyon (approved):
Content looks solid, so consider this a fixship :)

docs/pages/repo/docs/troubleshooting.mdx
```diff
@@ -5,37 +5,152 @@ description: This guide aims to help you debug issues with your Turborepo builds
...
+- A `.turbo` directory in each workspace if your project is a monorepo (e.g. `apps/my-app/.turbo/`)
+- A `turbo` directory in `node_modules/.cache`
 
-## I'm not seeing any cache hits
+Because the first two directories are not git-ignored by default, you may see an
+issue where you run the same task twice and get a cache missing, even though you
+didn't change anything, because the generated `.turbo` directories are getting included as
+the task _inputs_, and invalidating cache. To avoid this problem, you can limit your
+[`inputs` configuration][r-inputs-config], or add `.turbo/` to your `.gitignore` file. We recommend
+the latter.
```
mehulkar, line 87:
```suggestion
the task _inputs_, and invalidating cache. To avoid this problem, add `.turbo` to your
`.gitignore` file. Alternatively, you can also limit your [`inputs` configuration][r-inputs-config]
so that `.turbo` is not included in the cache inputs.
```

docs/pages/repo/docs/troubleshooting.mdx
mehulkar: yeah agreed, i'm not sure what the levels are currently, but breaking them down into more levels makes sense to me.

PR #4424, opened 2023-04-01
ci(workflow): apply latest turbopack when build next-swc

kwonoj:
### Description

- closes WEB-763.

This PR expands current daily integration setup to apply local patch to the turbopack when build `next-swc`. This'll allow to run integartion against latest TOT where possible.

Since we expect there are breaking changes makes incompatible build with next-* codes in next.js to the turbopack, workflow falls back to building normal next-swc with latest published turbopack if build fails for any reason. We do not have fine grained way to determine if build fails due to incompatiblity or other reason - simply falls back as soon as build fails.

### DIFF
--- a/.github/workflows/daily-nextjs-integration-test.yml
+++ b/.github/workflows/daily-nextjs-integration-test.yml
@@ -26,7 +26,7 @@ jobs:
     uses: ./.github/workflows/nextjs-integration-test.yml
     with:
       diff_base: "none"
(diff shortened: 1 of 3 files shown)


kwonoj:
This job is an example for failing build of next-swc falls back

https://github.com/vercel/turbo/actions/runs/4580249232/jobs/8088807766

jridgewell (approved):

PR #4427, opened 2023-04-01
Fix plex.tv showcase url

cursecodes:
### Description

Was looking at the showcase page on the docs and found that the plex link led to some random manufacturing platform (https://plex.com). This fixes it to redirect to https://plex.tv.

### DIFF
--- a/docs/components/clients/users.ts
+++ b/docs/components/clients/users.ts
@@ -241,7 +241,7 @@ export const users: Array<TurboUser> = [
   {
     caption: "Plex",
     image: "/images/logos/plex.svg",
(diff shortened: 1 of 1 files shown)


mehulkar (changes requested):
The associated logo looks more like plex.com rather than tv. Would need to look at git history and ask around which company is intended to be here

cursecodes:
The logo is an outdated plex.tv logo - not sure if the logo or url was a mistake.

mehulkar (approved):

mehulkar:
Thanks @cursecodes

PR #4433, opened 2023-04-03
port glob-based filesystem walking to rust

arlyon:
### Description

This ports glob-based file walking to rust. It uses the walkdir crate to traverse the filesystem, and eagerly skips directories that are known not to match.

We could probably speed this up a little bit with a PR to the glob-match crate for partial matches, rather than the naive pop-then-retry approach being used here. It is assumed that the time saved from this outweighs the cost of traversing into directories that are likely to be completely ignored later on, saving a very large number of glob evaluations later on.

### Testing Instructions

This is used in prune and run so should perform similarly in rust as in go (with some minor perf overhead from ffi). 

### Symlinks

- **test::glob_walk::_broken_symlink_none_1_1_expects**
  How `walkdir` treats symlinks is different from how the go impl did. In go, broken symlinks are yielded whereas in rust these become IO errors. We will likely need to attempt to parse the io error to recover and yield those.
- **test::glob_walk::single_star_match**
  Same as above, the broken symlink does not appear in the list of files, leading to a test failure.
- **test::glob_walk::_working_symlink_c_none_1_1_expects**
  Another implementation difference. We do not enter the symlink, so the glob doesn't make sense. It could be that the go implementation will yield a partially-matching symlink if a glob crosses it. I propose we just yield the symlink if something inside it could match.
- **test::glob_walk::a_followed_by_double_star_match**
  I am not sure why this is incorrect. Thoughts?

### Doublestar crossing directory boundary

- **test::glob_walk::abc_then_doublestar**
  The go implementation will yield `abc` as a result for the glob `abc/**`.
- **test::glob_walk_files::no_trailing_slash_basepath_works**
  Same as above.
- **test::glob_walk_files::output_globbing_grabs_the_desired_content**
  Same as above.
- **test::glob_walk_files::passing_captures_all_children**
  Same as above.
- **test::glob_walk_files::passing_just_a_directory_to_exclude_prevents_capture_of_children**
  Same as above.
- **test::glob_walk_files::passing_to_exclude_prevents_capture_of_children**
  Same as above.
- **test::glob_walk_files::traversal_works_within_base_path**
  Same as above.
- **test::glob_walk_files::self_references_work**
  Same as above.


cli/internal/run/global_hash.go
```diff
@@ -90,7 +90,7 @@ func calculateGlobalHash(
 			return GlobalHashable{}, err
 		}
 
-		f, err := globby.GlobFiles(rootpath.ToStringDuringMigration(), globalFileDependencies, ignores)
+		f, err := ffi.Glob(rootpath.ToStringDuringMigration(), globalFileDependencies, ignores, true)
```
mehulkar, line 93: what's the extra argument `filesOnly`?

mehulkar (commented):
In Run Summary, we eventually want to map the retunred files from a glob to the source glob. We don't have to add it here, but something to keep that in the back of your mind as you do the glob walk in Rust. 

E.g. given `["**/turbo.json.*", "**/.env"]`, we should be able to get back

```
{
   "**/turbo.json.*": [...]
  "**/.env": [...]
}
```

arlyon:
Ty for the context :) I'll keep that in mind.

chris-olszewski (commented):
Looks promising, most important part of this PR will be porting all of the unit tests.

crates/globwalk/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+        .into_iter()
+        // we need to eagerly filter folders, preventing traversal into
+        // folders that aren't included or are excluded. we do this by
+        // evaluating 'potential matches' for the include list meaning
+        // 'it is possible that a subfolder may match', as well as definite
+        // matches for the exclude list
+        .filter_entry(|e| {
+            include
+                .iter()
+                .any(|p| potential_match(e.path().to_string_lossy().as_ref(), p))
```
NicholasLYang, line 51: Maybe we should ignore or error on non-Unicode paths

crates/globwalk/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+
+pub enum WalkType {
+    Files,
+    Folders,
+    All,
+}
+
+#[derive(Debug)]
+pub enum WalkError {
+    Error(walkdir::Error),
```
NicholasLYang, line 14: Can we do `#[from]` with thiserror? Alternatively, is there any issue with returning the `walkdir::Error` type itself?

crates/globwalk/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/// Checks if a path is a partial match for a glob, meaning that a
+/// subfolder could match.
+fn potential_match(path: &str, glob: &str) -> bool {
+    let matches = glob_match(glob, path);
+
+    // pop last chunk from glob and try again
+    if !matches {
+        if let Some((prefix, _)) = glob.rsplit_once('/') {
```
NicholasLYang, line 91: nit, you could do `Option::map_or` here.

crates/globwalk/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+                    .try_into()
+                    .expect("these are absolute paths")
+            })
+            .map_err(WalkError::Error)
+        })
+        .filter(move |res: &Result<AbsolutePathBuf, _>| match res {
+            Ok(path) => match walk_type {
+                // if we encounter a file, then we need to check if it matches
+                WalkType::Files | WalkType::All if path.is_file() => {
+                    include
```
NicholasLYang, line 68: This could be factored out into a separate function since we do it a couple times and this iterator chain is a bit of a beast.

NicholasLYang (commented):
Overall makes sense to me. Just have a couple nits

crates/globwalk/src/lib.rs
arlyon: Going to bubble up these issues instead of lossy consumption.

crates/globwalk/src/lib.rs
arlyon: I think for one variant we can probably forgo that. It is an enum in case we need to add additional stuff later. I usually have an enum as the top level error for non-leaf libraries.

gsoltis (changes requested):
I think this needs to be rebased and have the tests from [globby_test.go](https://github.com/vercel/turbo/blob/main/cli/internal/globby/globby_test.go) ported. Those tests will help us to determine if there is a behavior change

crates/globwalk/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+    path: &str,
+    include: &'a [String],
+    exclude: &'a [String],
+    is_directory: bool,
+) -> Result<MatchType, &'a String> {
+    if include.is_empty() {
+        return Ok(MatchType::Match);
+    }
+
+    println!("matching: {:?}", path);
```
gsoltis, line 161: Drop `println`

arlyon:
Closing in favour of #5168

PR #4439, opened 2023-04-03
port(turborepo): New Turbopath

NicholasLYang:
### Description

This is meant as a very very simple port of turbopath to Rust. It's basically the Go code but the Rust version.

The goal here is to make something that works, and that we can extend with a little ChatGPT and copilot help.

### Testing Instructions

There are doc tests on relevant functions and some regular tests too.


crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -0,0 +1,175 @@
...
+        self.0.starts_with(base.as_ref())
+    }
+
+    pub fn ends_with<P: AsRef<Path>>(&self, child: P) -> bool {
+        self.0.ends_with(child.as_ref())
+    }
+
+    pub fn join<P: AsRef<Path>>(&self, path: P) -> AbsoluteSystemPathBuf {
+        AbsoluteSystemPathBuf(self.0.join(path))
+    }
```
chris-olszewski, line 134: Do we want to expose this method? It has (IMO) non-obvious behavior e.g. `AbsoluteSystemPathBuf::new("/tmp/foo").unwrap().join("/etc") == "/etc"`

crates/turbopath/src/anchored_system_path_buf.rs
```diff
@@ -0,0 +1,51 @@
...
+        AnchoredSystemPathBuf(path.into())
+    }
+
+    pub fn as_path(&self) -> &Path {
+        self.0.as_path()
+    }
+
+    pub fn into_path_buf(self) -> PathBuf {
+        self.0
+    }
```
chris-olszewski, line 46: Is there a reason for preferring this method instead `impl Into<PathBuf> for AnchoredSystemPathBuf`?

crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -0,0 +1,175 @@
...
+        self.0.file_name()
+    }
+
+    pub fn exists(&self) -> bool {
+        self.0.exists()
+    }
+
+    pub fn into_path_buf(self) -> PathBuf {
+        self.0
+    }
```
chris-olszewski, line 158: `Into<PathBuf> for AbsoluteSystemPathBuf`?

crates/turbopath/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+impl IntoSystem for &Path {
+    fn into_system(self) -> Result<PathBuf, PathValidationError> {
+        let path_str = self.to_str().ok_or(PathValidationError::InvalidUnicode)?;
+
+        Ok(PathBuf::from_slash(path_str))
+    }
+}
+
+impl IntoUnix for &Path {
+    /// NOTE: `into_unix` *only* converts Windows paths to Unix paths *on* a
```
gsoltis, line 45: FWIW, `turbo` doesn't really have the notion of a "windows path". There's "system path" which is a path native to whatever OS you're currently running on, and "Unix path" which is treated as a portable format (and as such should always be relative).

If you're on a non-windows system, you shouldn't really have to worry about windows paths. They shouldn't occur in a situation where you are expected to manipulate them, and if they do, it's likely a bug.

crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -0,0 +1,175 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::path::{Path, PathBuf};
+    /// use turbopath::AbsoluteSystemPathBuf;
+    /// let path = PathBuf::from("/Users/user");
+    /// let absolute_path = AbsoluteSystemPathBuf::new(path).unwrap();
+    /// #[cfg(windows)]
+    /// assert_eq!(absolute_path.as_path(), Path::new("\\Users\\user"));
```
chris-olszewski, line 35: I think this will fail on windows since a drive is required for a path to be absolute: https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.is_absolute

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -0,0 +1,130 @@
...
+    ///
+    /// * `path`:
+    ///
+    /// returns: Result<RelativeUnixPathBuf, PathValidationError>
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// ```
+    pub fn new(path: impl Into<PathBuf>) -> Result<Self, PathValidationError> {
```
chris-olszewski, line 30: Should we only be compiling this for Windows to avoid misuse?

crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -0,0 +1,175 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `path`: The path to be converted to an `AbsoluteSystemPathBuf`.
+    ///
+    /// returns: AbsoluteSystemPathBuf
+    pub fn new_unchecked(path: impl Into<PathBuf>) -> Self {
+        let path = path.into();
+        AbsoluteSystemPathBuf(path)
+    }
```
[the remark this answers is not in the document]
gsoltis: We have an equivalent on the Go side that is "FromUpstream", for use when consuming an API that we have verified to produce an absolute system path. But, it is very much the kind of thing we can add if/when we need it, and don't need to include in the initial implementation if we don't have a motivation for it.

crates/turbopath/src/absolute_system_path_buf.rs
gsoltis: Eh, that behavior definitely doesn't seem great. I think there is likely a place for `join_literal` or `join_unix_path` or `join_tokens` that takes a slice or something. But none of those should clobber the underlying absolute path.

crates/turbopath/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// Windows system. Do not pass a Windows path on a Unix system and
+    /// assume it'll be converted.
+    fn into_unix(self) -> Result<PathBuf, PathValidationError> {
```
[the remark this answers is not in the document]
gsoltis: I think we expect this to be called, but be a no-op, on unix systems. This is an explicit step for humans to understand that we are transitioning to a path type suitable for sharing and no longer using the system representation, even if the underlying bytes are identical.

crates/turbopath/src/relative_unix_path_buf.rs
gsoltis: As above, we expect this to be called on unix systems, but to effectively be a no-op.

gsoltis (changes requested):

crates/turbopath/src/absolute_system_path_buf.rs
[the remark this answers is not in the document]
arlyon: Proposal to make this fn `unsafe`. I know it doesn't propose any _language_ correctness issues, but it does potentially cause build correctness issues...

crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -0,0 +1,175 @@
...
+    ///
+    /// ```
+    /// use std::path::Path;
+    /// use turbopath::{AbsoluteSystemPathBuf, AnchoredSystemPathBuf};
+    /// let base = AbsoluteSystemPathBuf::new("/Users/user").unwrap();
+    /// let anchored_path = AbsoluteSystemPathBuf::new("/Users/user/Documents").unwrap();
+    /// let anchored_path = base.anchor(&anchored_path).unwrap();
+    /// assert_eq!(anchored_path.as_path(), Path::new("Documents"));
+    /// ```
+    pub fn anchor(
```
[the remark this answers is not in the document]
arlyon: If we have `AnchoredSystemPath` we should probably keep the naming the same, although I prefer 'relative path' over 'anchored path' unless the anchored type keeps track of the anchor point.

crates/turbopath/src/anchored_system_path_buf.rs
```diff
@@ -0,0 +1,51 @@
...
+        if path.is_absolute() {
+            return Err(PathValidationError::NotRelative(path.to_path_buf()));
+        }
+
+        Ok(AnchoredSystemPathBuf(path.into_system()?))
+    }
+}
+
+impl AnchoredSystemPathBuf {
+    pub fn strip_root(
```
arlyon, line 23: Can this not just be `root.anchor(path)`?

crates/turbopath/src/anchored_system_path_buf.rs
```diff
@@ -0,0 +1,51 @@
...
+            .strip_prefix(root.as_path())
+            .map_err(|_| PathValidationError::NotParent(root.to_string(), path.to_string()))?
+            .to_path_buf();
+
+        Ok(AnchoredSystemPathBuf(stripped_path))
+    }
+
+    pub fn new_unchecked(path: impl Into<PathBuf>) -> Self {
+        AnchoredSystemPathBuf(path.into())
+    }
```
arlyon, line 38: Same here I think marking this unsafe might be an idea

crates/turbopath/src/absolute_system_path_buf.rs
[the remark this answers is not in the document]
NicholasLYang: I think marking it unsafe is reasonable, but agreed that we should wait until it's needed

crates/turbopath/src/absolute_system_path_buf.rs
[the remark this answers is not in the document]
NicholasLYang: I wasn't a huge fan of `RelativeTo` as a name. This feels a little more clear that you're using `self` as an anchor for the path.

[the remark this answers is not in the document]
NicholasLYang: I thought about keeping track of the anchor point btw. We could do that although currently that'd require some copying

chris-olszewski (changes requested):
We need to add `turbopath` to the turborepo rust tests so they get run in CI

gsoltis (approved):
I'm leaving a fix-ship to ensure that the tests are running and to update the PR description to include a useful testing description.

sokra (approved):

ForsakenHarmony:
I just saw this, and I'm very confused, this seems like it's completely unnecessary in rust, it's basically just `std::path` in worse?

PR #4442, opened 2023-04-04
feat(turbopack): initial mdxRs next.config.js support

kwonoj:
### Description

Related with WEB-488.

This PR enables `mdxRs` configuration option from next.config.js's `experimental`. When it is enabled, it allows to transform `.md` and `.mdx` with mdx compiler with inheriting any existing jsx transform options to align with other jsx components transform options.

Currently while there are mdx / mdxRs config, both shares same transform (mdxrs). It may need to be either 1. consolidate options in next.config.js or 2. support js-based mdx correctly in turbopack, but something to consider later.

Lastly, with `Runtime::Classic` unlike existing test cases in next.js this transform will fail with `React is not defined` runtime error. Manually adding import to react works, however it's unclear if turbopack supposed to auto-inject those importsource or not. For now, make minimal piece working first.

### DIFF
--- a/crates/turbopack-mdx/src/lib.rs
+++ b/crates/turbopack-mdx/src/lib.rs
@@ -1,8 +1,11 @@
 #![feature(min_specialization)]
 
 use anyhow::{anyhow, Result};
-use mdxjs::compile;
-use turbo_tasks::{primitives::StringVc, Value};
+use mdxjs::{compile, Options};
+use turbo_tasks::{
+    primitives::{OptionStringVc, StringVc},
+    Value,
+};
 use turbo_tasks_fs::{rope::Rope, File, FileContent, FileSystemPathVc};
 use turbopack_core::{
     asset::{Asset, AssetContent, AssetContentVc, AssetVc},
@@ -31,12 +34,37 @@ fn modifier() -> StringVc {
     StringVc::cell("mdx".to_string())
 }
 
+/// Subset of mdxjs::Options to allow to inherit turbopack's jsx-related configs
+/// into mdxjs.
+#[turbo_tasks::value(serialization = "auto_for_input")]
+#[derive(PartialOrd, Ord, Hash, Debug, Copy, Clone)]
+pub struct MdxTransformOptions {
+    pub development: bool,
+    pub preserve_jsx: bool,
+    pub jsx_runtime: OptionStringVc,
+    pub jsx_import_source: OptionStringVc,
+    pub filepath: Option<FileSystemPathVc>,
+}
+
+impl Default for MdxTransformOptions {
+    fn default() -> Self {
+        Self {
+            development: true,
+            preserve_jsx: false,
+            jsx_runtime: OptionStringVc::cell(None),
+            jsx_import_source: OptionStringVc::cell(None),
+            filepath: None,
+        }
+    }
+}
+
 #[turbo_tasks::value]
 #[derive(Clone, Copy)]
 pub struct MdxModuleAsset {
     source: AssetVc,
     context: AssetContextVc,
     transforms: EcmascriptInputTransformsVc,
+    options: MdxTransformOptions,
 }
 
 /// MDX components should be treated as normal j|tsx components to analyze
@@ -49,6 +77,7 @@ async fn into_ecmascript_module_asset(
 ) -> Result<EcmascriptModuleAssetVc> {
     let content = current_context.content();
     let this = current_context.await?;
+    let transform_options = this.options;
 
     let AssetContent::File(file) = &*content.await? else {
         anyhow::bail!("Unexpected mdx asset content");
@@ -58,9 +87,35 @@ async fn into_ecmascript_module_asset(
         anyhow::bail!("Not able to read mdx file content");
     };
 
+    let jsx_runtime = if let Some(runtime) = &*transform_options.jsx_runtime.await? {
+        match runtime.as_str() {
+            "automatic" => Some(mdxjs::JsxRuntime::Automatic),
+            "classic" => Some(mdxjs::JsxRuntime::Classic),
+            _ => None,
+        }
+    } else {
+        None
+    };
+
+    let filepath = if let Some(filepath) = transform_options.filepath {
+        Some(filepath.await?.path.clone())
+    } else {
+        None
+    };
+
+    let options = Options {
+        development: transform_options.development,
+        jsx: transform_options.preserve_jsx, // true means 'preserve' jsx syntax.
+        jsx_runtime,
+        jsx_import_source: transform_options.jsx_import_source.await?.clone_value(),
+        filepath,
+        ..Default::default()
+    };
     // TODO: upstream mdx currently bubbles error as string
     let mdx_jsx_component =
-        compile(&file.content().to_str()?, &Default::default()).map_err(|e| anyhow!("{}", e))?;
+        compile(&file.content().to_str()?, &options).map_err(|e| anyhow!("{}", e))?;
+
+    println!("MDX JSX: {:#?} {}", options, mdx_jsx_component);
 
     let source = VirtualAssetVc::new_with_ident(
         this.source.ident(),
@@ -83,11 +138,13 @@ impl MdxModuleAssetVc {
         source: AssetVc,
         context: AssetContextVc,
         transforms: EcmascriptInputTransformsVc,
+        options: Value<MdxTransformOptions>,
     ) -> Self {
         Self::cell(MdxModuleAsset {
             source,
             context,
             transforms,
+            options: options.into_value(),
         })
     }
 
--- a/crates/turbopack-mdx/src/lib.rs
+++ b/crates/turbopack-mdx/src/lib.rs
@@ -2,10 +2,7 @@
 
 use anyhow::{anyhow, Result};
 use mdxjs::{compile, Options};
-use turbo_tasks::{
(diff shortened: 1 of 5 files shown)


kwonoj:
Next.js side counterpart: https://github.com/vercel/next.js/pull/47886

crates/turbopack-mdx/src/lib.rs
```diff
@@ -58,9 +87,35 @@ async fn into_ecmascript_module_asset(
         anyhow::bail!("Not able to read mdx file content");
     };
 
+    let jsx_runtime = if let Some(runtime) = &*transform_options.jsx_runtime.await? {
+        match runtime.as_str() {
+            "automatic" => Some(mdxjs::JsxRuntime::Automatic),
+            "classic" => Some(mdxjs::JsxRuntime::Classic),
+            _ => None,
```
arlyon, line 94: Perhaps we should error here if an unexpected string is provided (or make jsx_runtime an enum to start with?)

arlyon (approved):
Giving the turborepo stamp, though I expect pack to have more to say. Left one comment.

crates/turbopack-mdx/src/lib.rs
kwonoj: There are a couple of different sources of existing runtime enum for the runtime so yes, this might need some update later.

crates/turbopack-mdx/src/lib.rs
```diff
@@ -31,12 +34,37 @@ fn modifier() -> StringVc {
...
 }
 
+/// Subset of mdxjs::Options to allow to inherit turbopack's jsx-related configs
+/// into mdxjs.
+#[turbo_tasks::value(serialization = "auto_for_input")]
+#[derive(PartialOrd, Ord, Hash, Debug, Copy, Clone)]
+pub struct MdxTransformOptions {
+    pub development: bool,
+    pub preserve_jsx: bool,
+    pub jsx_runtime: OptionStringVc,
```
jridgewell, line 44: Nit: does this need to be an Vc?

crates/turbopack-mdx/src/lib.rs
```diff
@@ -31,12 +34,37 @@ fn modifier() -> StringVc {
...
+    pub jsx_import_source: OptionStringVc,
```
jridgewell, line 45: Ditto.

crates/turbopack-mdx/src/lib.rs
```diff
@@ -31,12 +34,37 @@ fn modifier() -> StringVc {
...
+    }
+}
+
 #[turbo_tasks::value]
 #[derive(Clone, Copy)]
 pub struct MdxModuleAsset {
     source: AssetVc,
     context: AssetContextVc,
     transforms: EcmascriptInputTransformsVc,
+    options: MdxTransformOptions,
```
jridgewell, line 67: Instead of having a struct with Vcs, we could have a Vc of values here, which would help the propagation of data without invalidating everything.

crates/turbopack-mdx/src/lib.rs
```diff
@@ -58,9 +87,35 @@ async fn into_ecmascript_module_asset(
...
+            "automatic" => Some(mdxjs::JsxRuntime::Automatic),
+            "classic" => Some(mdxjs::JsxRuntime::Classic),
+            _ => None,
+        }
+    } else {
+        None
+    };
+
+    let filepath = if let Some(filepath) = transform_options.filepath {
+        Some(filepath.await?.path.clone())
```
jridgewell, line 101: Does this need to be a system path? We should use `to_sys_path` instead.

crates/turbopack-mdx/src/lib.rs
```diff
@@ -58,9 +87,35 @@ async fn into_ecmascript_module_asset(
...
+        jsx_import_source: transform_options.jsx_import_source.await?.clone_value(),
+        filepath,
+        ..Default::default()
+    };
     // TODO: upstream mdx currently bubbles error as string
     let mdx_jsx_component =
-        compile(&file.content().to_str()?, &Default::default()).map_err(|e| anyhow!("{}", e))?;
+        compile(&file.content().to_str()?, &options).map_err(|e| anyhow!("{}", e))?;
+
+    println!("MDX JSX: {:#?} {}", options, mdx_jsx_component);
```
jridgewell, line 118: Log?

crates/turbopack-mdx/src/lib.rs
```diff
@@ -83,11 +138,13 @@ impl MdxModuleAssetVc {
         source: AssetVc,
         context: AssetContextVc,
         transforms: EcmascriptInputTransformsVc,
+        options: Value<MdxTransformOptions>,
```
jridgewell, line 141: Does this need a `Value` wrapper?

crates/turbopack/src/module_options/mod.rs
```diff
@@ -318,12 +321,34 @@ impl ModuleOptionsVc {
...
+                let jsx = enable_jsx.await?;
+                (
+                    OptionStringVc::cell(jsx.runtime.clone()),
+                    OptionStringVc::cell(jsx.import_source.clone()),
+                )
+            } else {
+                (OptionStringVc::cell(None), OptionStringVc::cell(None))
+            };
+
+            let mdx_transform_options = MdxTransformOptions {
```
jridgewell, line 335: Let's turn this into a Vc and propagate it from here.

crates/turbopack-mdx/src/lib.rs
kwonoj: ohno

crates/turbopack-mdx/src/lib.rs
kwonoj: it's mostly identifier so not a huge matter afaik.

cli/cmd/turbo/version.go
```diff
@@ -1,3 +1,3 @@
 package main
 
-const turboVersion = "1.8.9-canary.1"
+const turboVersion = "1.9.0-canary.1"
```
sokra, line 3: I don't think you should update that

crates/turbopack-mdx/src/lib.rs
```diff
@@ -31,12 +34,51 @@ fn modifier() -> StringVc {
...
+/// Subset of mdxjs::Options to allow to inherit turbopack's jsx-related configs
+/// into mdxjs.
+#[turbo_tasks::value(shared)]
+#[derive(PartialOrd, Ord, Hash, Debug, Clone)]
+pub struct MdxTransformOptions {
+    pub development: bool,
+    pub preserve_jsx: bool,
+    pub jsx_runtime: Option<String>,
+    pub jsx_import_source: Option<String>,
+    pub filepath: Option<FileSystemPathVc>,
```
sokra, line 46:
```suggestion
```

Why would be pass the filepath when we already have this info from `source.ident().path()`

crates/turbopack-mdx/src/lib.rs
```diff
@@ -58,9 +101,36 @@ async fn into_ecmascript_module_asset(
...
+
+    let options = Options {
+        development: transform_options.development,
+        jsx: transform_options.preserve_jsx, // true means 'preserve' jsx syntax.
+        jsx_runtime,
+        jsx_import_source: transform_options
+            .jsx_import_source
+            .as_ref()
+            .map(|s| s.into()),
+        filepath,
```
sokra, line 128:
```suggestion
        filepath: self.source.ident().to_string().await?.clone_value(),
```

sokra (changes requested):

sokra:
This was incorrectly rebased/pulled. Need a rebase.

crates/turbopack/src/lib.rs
```diff
@@ -173,9 +173,10 @@ async fn apply_module_type(
             ModuleCssModuleAssetVc::new(source, context.into(), *transforms).into()
         }
         ModuleType::Static => StaticModuleAssetVc::new(source, context.into()).into(),
-        ModuleType::Mdx(transforms) => {
-            MdxModuleAssetVc::new(source, context.into(), *transforms).into()
-        }
+        ModuleType::Mdx {
+            transforms,
+            options,
+        } => MdxModuleAssetVc::new(source, context.into(), *transforms, (*options).into()).into(),
```
jridgewell, line 179: Is the into necessary?
```suggestion
        } => MdxModuleAssetVc::new(source, context.into(), *transforms, *options).into(),
```

crates/turbopack/src/module_options/mod.rs
```diff
@@ -318,12 +321,30 @@ impl ModuleOptionsVc {
...
-        if enable_mdx {
+        if enable_mdx || enable_mdx_rs {
+            let (jsx_runtime, jsx_import_source) = if let Some(enable_jsx) = enable_jsx {
+                let jsx = enable_jsx.await?;
+                (jsx.runtime.clone(), jsx.import_source.clone())
+            } else {
+                (None, None)
+            };
+
+            let mdx_transform_options = MdxTransformOptionsVc::cell(MdxTransformOptions {
```
jridgewell, line 332: Nit: you should be able to do `MdxTransformOptions {}.cell()` instead of `MdxTransformOptionsVc::cell(MdxTransformOptions {})`

jridgewell (approved):

sokra (approved):

PR #4446, opened 2023-04-04
fix HMR for streaming responses

sokra:
### Description

this adds HMR support for streaming responses

### DIFF
--- a/crates/turbopack-core/src/version.rs
+++ b/crates/turbopack-core/src/version.rs
@@ -1,10 +1,9 @@
+use std::sync::Arc;
+
 use anyhow::{anyhow, Result};
 use serde::{Deserialize, Serialize};
 use turbo_tasks::{
-    debug::ValueDebugFormat,
-    primitives::{JsonValueVc, StringVc},
-    trace::TraceRawVcs,
-    IntoTraitRef,
+    debug::ValueDebugFormat, primitives::StringVc, trace::TraceRawVcs, IntoTraitRef, TraitRef,
 };
 use turbo_tasks_fs::{FileContent, FileContentReadRef, LinkType};
 use turbo_tasks_hash::{encode_hex, hash_xxh3_hash64};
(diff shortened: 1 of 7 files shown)


crates/turbopack-dev-server/src/update/server.rs
```diff
@@ -72,8 +72,17 @@ impl<P: SourceProvider + Clone + Send + Sync> UpdateServer<P> {
...
                                 }
                             };
-                            let stream = UpdateStream::new(TransientInstance::new(Box::new(get_content))).await?;
-                            streams.insert(resource, stream);
+                            match UpdateStream::new(TransientInstance::new(Box::new(get_content))).await {
+                                Ok(stream) => {
+                                    streams.insert(resource, stream);
+                                }
+                                Err(err) => {
+                                    println!("Failed to create update stream for {resource}: {}", PrettyPrintError(&err));
```
alexkirsz, line 80:
```suggestion
                                    eprintln!("Failed to create update stream for {resource}: {}", PrettyPrintError(&err));
```

crates/turbopack-core/src/version.rs
```diff
@@ -186,17 +185,20 @@ pub enum Update {
 #[derive(PartialEq, Eq, Debug, Clone, TraceRawVcs, ValueDebugFormat, Serialize, Deserialize)]
 pub struct TotalUpdate {
     /// The version this update will bring the object to.
-    pub to: VersionVc,
+    #[turbo_tasks(trace_ignore)]
+    pub to: TraitRef<VersionVc>,
```
alexkirsz, line 189: Why is this necessary?

sokra: `compute_update_stream` is only executed the `UpdateStreamItem` changes (since it only this is awaited). But `VersionVc` might be equal while the underlying content can change. So when trying to send 2 `TotalUpdate` with `to` pointing to the same Vc (but with different content) it won't send that.

alexkirsz (approved):

PR #4448, opened 2023-04-04
make sure Node.js process closes when server disappears

sokra:
### Description

avoid zombie node.js processes


crates/turbopack-node/js/src/ipc/index.ts
```diff
@@ -83,6 +83,9 @@ function createIpc<TIncoming, TOutgoing>(
       }
     });
   });
+  socket.once("close", () => {
+    process.exit(0);
```
alexkirsz, line 87: Add a comment to explain how this can happen, and that this avoids zombie processes.

alexkirsz (approved):

PR #4449, opened 2023-04-04
Strict Environment Handling

nathanhammond:
To provide a way to reduce information leakage into the task execution environment, add a strict mode for environment variable handling. This mode does not propagate non-enumerated variables to the environment of the task being executed.

Things that changed to make sure to review which are harder to do in code, pop into the deploy:
- `turbo.json` schema.
- Docs for CLI and Configuration.

## Open Design Option

This partitions the global cache key based solely upon existence of `--env=strict` vs. `--env=loose`. It could instead do that in tandem with existence of `globalPassthroughEnv` if and only if `--env=strict` information is moved into the task hashes which have `passthroughEnv` specified.

That will increase possible granularity, but I suspect that the use case for this feature is going to be people _wanting_ to lock it down.


nathanhammond (commented):
Reviewer's guide.

crates/turborepo-lib/src/cli.rs
```diff
@@ -309,6 +322,11 @@ pub struct RunArgs {
     /// .html). Outputs dot graph to stdout when if no filename is provided
     #[clap(long, num_args = 0..=1, default_missing_value = "")]
     pub graph: Option<String>,
+    /// Environment variable mode.
+    /// Loose passes the entire environment.
+    /// Strict uses an allowlist specified in turbo.json.
+    #[clap(long = "env", default_value = "infer", num_args = 0..=1, default_missing_value = "infer")]
```
NicholasLYang, line 328: Smallest nit, but do we want to consider naming this `--env-mode`? I feel like `--env` could be used to forward environment variables to tasks, like in Docker.

NicholasLYang (approved):
Mostly needs some tests but otherwise LGTM

crates/turborepo-lib/src/cli.rs
nathanhammond: Naming is precisely the bikesheds I want to address in this review, actually!

cli/integration_tests/strict_env_vars/global_passthrough_env.t
```diff
@@ -0,0 +1,69 @@
...
+  [-0-9:.TWZ+]+ \[DEBUG] turborepo_lib::cli: pkg_inference_root set to "" (re)
+  [-0-9:.TWZ+]+ \[DEBUG] turbo: Found go binary at "[\-\w\/]+" (re)
+  [-0-9:.TWZ+]+ \[DEBUG] turbo: build tag: (go|rust) (re)
+  [-0-9:.TWZ+]+ \[INFO]  turbo: skipping turbod since we appear to be in a non-interactive context (re)
+  [-0-9:.TWZ+]+ \[DEBUG] turbo: global hash env vars: vars=\["VERCEL_ANALYTICS_ID"] (re)
+  [-0-9:.TWZ+]+ \[DEBUG] turbo: new global hash: value=b7e2e8ed3383c982 (re)
+  [-0-9:.TWZ+]+ |[DEBUG] turbo: local cache folder: path="" (re)
+  No tasks were executed as part of this run.
+
+Loose + ["BONUS_VAR"]
```
nathanhammond, line 53: This has the same hash as the above one, since loose explicitly means "don't pay attention to it."

cli/integration_tests/strict_env_vars/strict_env_vars.t
```diff
@@ -0,0 +1,67 @@
+Setup
+  $ . ${TESTDIR}/../setup.sh
+  $ . ${TESTDIR}/setup.sh $(pwd) monorepo
+
+Strict with no passthroughEnv opts you into the new hashing behavior.
```
nathanhammond, line 5: This is an optional decision. We could make it use the old hashing behavior here. ("You specified a nonsense combination of options.")

If we do that, though, we need to move the EnvMode flag into each individual task hash.

cli/internal/run/run.go
```diff
@@ -253,18 +256,29 @@ func (r *run) run(ctx gocontext.Context, targets []string) error {
...
 	)
 
 	if err != nil {
 		return fmt.Errorf("failed to collect global hash inputs: %v", err)
 	}
 
-	if globalHash, err := fs.HashObject(getGlobalHashable(globalHashable)); err == nil {
-		r.base.Logger.Debug("global hash", "value", globalHash)
-		g.GlobalHash = globalHash
+	if globalHashable.envMode == util.Infer && globalHashable.envVarPassthroughs == nil {
```
nathanhammond, line 268: Opt into the old global hash behavior.

nathanhammond (commented):
Some decisions about what causes hash changes.

cli/internal/run/real_run.go
```diff
@@ -271,8 +292,29 @@ func (ec *execContext) exec(ctx gocontext.Context, packageTask *nodes.PackageTas
...
 	cmd := exec.Command(ec.packageManager.Command, argsactual...)
 	cmd.Dir = packageTask.Pkg.Dir.ToSystemPath().RestoreAnchor(ec.repoRoot).ToString()
-	envs := fmt.Sprintf("TURBO_HASH=%v", hash)
-	cmd.Env = append(os.Environ(), envs)
+
+	currentState := env.GetEnvMap()
+	passthroughEnv := env.EnvironmentVariableMap{}
+
+	if strictEnv {
+		defaultPassthrough := []string{
```
gsoltis, line 300: I think Windows will also require `SYSTEMROOT`. Go adds it by default if it's missing, but let's be explicit that we're including it.

crates/turborepo-lib/src/cli.rs
mehulkar: Update this to `--experimental-env-mode` for now and hid it from help output. We can bikeshed before real release

cli/internal/run/run.go
mehulkar: I'm not clear on whether we need this switch anymore. I think the reasoning was that since we're already busting global cache, we might as well drop the `getOldGlobalHashable`, but I think we can avoid busting global cache altogether, so we should keep the `getOldGlobalHashable` all around

cli/integration_tests/prune/produces_valid_turbo_json.t
```diff
@@ -19,6 +19,7 @@ Make sure we prune tasks that reference a pruned workspace
         "dependsOn": [],
         "inputs": [],
         "outputMode": "full",
+        "passthroughEnv": [],
```
mehulkar, line 22: This shouldn't be added, since it's an experimental field 🤔 , need to figure out why it's happening

cli/integration_tests/bad_flag.t
```diff
@@ -19,7 +19,7 @@ Bad flag with an implied run command should display run flags
...
   
-  Usage: turbo <--cache-dir <CACHE_DIR>|--cache-workers <CACHE_WORKERS>|--concurrency <CONCURRENCY>|--continue|--dry-run [<DRY_RUN>]|--single-package|--filter <FILTER>|--force|--global-deps <GLOBAL_DEPS>|--graph [<GRAPH>]|--ignore <IGNORE>|--include-dependencies|--no-cache|--no-daemon|--no-deps|--output-logs <OUTPUT_LOGS>|--only|--parallel|--pkg-inference-root <PKG_INFERENCE_ROOT>|--profile <PROFILE>|--remote-only|--scope <SCOPE>|--since <SINCE>|--summarize [<SUMMARIZE>]|--log-prefix <LOG_PREFIX>|TASKS|PASS_THROUGH_ARGS|--experimental-space-id <EXPERIMENTAL_SPACE_ID>>
+  Usage: turbo <--cache-dir <CACHE_DIR>|--cache-workers <CACHE_WORKERS>|--concurrency <CONCURRENCY>|--continue|--dry-run [<DRY_RUN>]|--single-package|--filter <FILTER>|--force|--global-deps <GLOBAL_DEPS>|--graph [<GRAPH>]|--experimental-env-mode [<ENV_MODE>]|--ignore <IGNORE>|--include-dependencies|--no-cache|--no-daemon|--no-deps|--output-logs <OUTPUT_LOGS>|--only|--parallel|--pkg-inference-root <PKG_INFERENCE_ROOT>|--profile <PROFILE>|--remote-only|--scope <SCOPE>|--since <SINCE>|--summarize [<SUMMARIZE>]|--log-prefix <LOG_PREFIX>|TASKS|PASS_THROUGH_ARGS|--experimental-space-id <EXPERIMENTAL_SPACE_ID>>
```
mehulkar, line 22: It isn't possible to hide flags from the usage message

mehulkar:
OK, this PR is ready for a review. I added experimental prefixes to the flag and the config, fixed the global hash changes, and added a bunch more tests. I didn't add any tests for hash changes when adding `experimentalPassThroughEnv` to each task, because adding config to the pipeline busts the global cache. (We _could_ exercise this case with a Workspace Config, but I wanted to wrap up the PR for 1.9 release. Since it's experimental, we can add that test case if we need to later)

docs/pages/repo/docs/reference/command-line-reference.mdx
```diff
@@ -178,6 +178,27 @@ Task details include:
...
 
+#### `--experimental-env-mode`
+
+<Callout type="info">
+  **Warning**: This in experimental flag, so its name and behavior can change.
+</Callout>
+
+`type: string`
+
+Controls the available variables to your tasks based upon the values specified in the `globalEnvPassthrough` and `envPassthrough` fields in `turbo.json`.
```
gsoltis, line 189: Here and elsewhere on this page `globalEnvPassthrough` -> `experimentalGlobalEnvPassthrough`, etc.

docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -176,6 +187,27 @@ The list of environment variables a task depends on.
   caching](/repo/docs/core-concepts/caching#automatic-environment-variable-inclusion).
 </Callout>
 
+## `experimentalPassthroughEnv`
+
+`type: string[]`
+
+An allowlist of environment variables that should be made available to this task but should not contribute to the task's cache key. Using this key opts this task into strict environment variable mode.
```
gsoltis, line 194: Maybe add a note that the _names_ contribute to the hash key, but the _values_ don't. So updating this list will bust the cache, but changing the values of vars in the list won't

gsoltis (changes requested):
Unless I missed it, I don't think we define what "strict" or "loose" modes are in the docs anywhere. We discuss how to control which one is picked, but not what they are or what they do.

Let's also make sure that we are including `SYSTEMROOT` in the default list of passthrough vars, it's required on Windows. Go will do it by default, but there should be no harm in us being explicit and that will ensure it doesn't get lost in the porting effort.

mehulkar:
@gsoltis added `SYSTEMROOT` (with tests) and moved docs to https://github.com/vercel/turbo/pull/4490, since we want to wait till release on those anyway.

mehulkar (approved):

gsoltis (approved):

PR #4458, opened 2023-04-04
feat(ecmascript): support configurable emotion options

kwonoj:
### Description

Related with WEB-669.

PR implements initial path to configuring emotion transforms. It doesn't fully implement all of the options values though -`auto-label` have different values between next.config.js to the swc's actual transform options, as well as `importMap`. These will be gradually follow up since it is not an immediate blocker for the feature itself.


crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -28,7 +28,12 @@ pub enum EcmascriptInputTransform {
     ClientDirective(StringVc),
     CommonJs,
     Custom(CustomTransformVc),
-    Emotion,
+    Emotion {
+        #[serde(default)]
+        sourcemap: bool,
```
jridgewell, line 33: Can we just always enable the sourcemap?

crates/turbopack/src/module_options/module_options_context.rs
```diff
@@ -89,6 +89,38 @@ impl Default for TypescriptTransformOptionsVc {
...
+    Always,
+    Never,
+}
+
+//[TODO]: need to support importmap, there are type mismatch between
+//[TODO]: next.config.js to swc's emotion options
+#[turbo_tasks::value(shared)]
+#[derive(Default, Clone, Debug)]
+pub struct EmotionTransformOptions {
+    pub enabled: bool,
```
jridgewell, line 104: Do we need an `enabled` if the `enable_emotion` if this is only used inside an `Option` wrapper? Shouldn't the `Some(EmotionTransformOptions)` imply that it is enabled?

crates/turbopack-ecmascript/src/transform/mod.rs
kwonoj: for now, this inherits what next.config.js allow to configure. I hope to simplify / consolidate but thinking about feature parity to existing next.js configs we may need to expose.

crates/turbopack/src/module_options/module_options_context.rs
kwonoj: That's how I started, and then ends up this was somewhat more ergonomics when try to read next.config.js, we have couple of conditions

1. if compiler option non-exists - None
2. if exists, somehow user configures `emotion: false` or `emotion: {enabled :false}`

and having Option<T> was easier to gracefully read & create config value among those conditions. It is possible to argue who would do 2 (if you specifiy compiler option, explicitly disabling it is non common) but that's certainly possible case if someone relies on some dynamic evaluation based on env / fn.

jridgewell (approved):

PR #4459, opened 2023-04-04
fix(lockfile): pnpm

tknickman:
### Description

In https://github.com/vercel/turbo/pull/4398 I committed a lockfile from `pnpm@7.30.5`, which is (somehow) incompatible with 7.18.2 which we install in CI. This broke our release scripts. 

This PR commits the lockfile from running `pnpm install` with `7.18.2`


tknickman:
We could also be extra safe if we want and revert create-turbo PR, update the lockfile in that branch and resubmit

tknickman:
Closing this and going with the less destructive approach

PR #4460, opened 2023-04-04
Redirect all test output to /dev/null

mehulkar:


examples_tests/npm-non-monorepo.t
```diff
@@ -1,5 +1,5 @@
   $ . ${TESTDIR}/setup.sh non-monorepo npm
 # run twice and make sure it works
-  $ npx turbo build lint > /dev/null
-  $ npx turbo build lint > /dev/null
+  $ npx turbo build lint > /dev/null 2>&1
```
NicholasLYang, line 3: Do we need to check the exit code to confirm that this ran correctly?

mehulkar: if it's non-zero, the test will fail as prysk expects you to add a `[1]` or whatever in the line below

PR #4464, opened 2023-04-04
Better dotenv error messages

jridgewell:
The issue message emitted by `TryDotenvProcessEnv` was pretty bad: `Execution of TryDotenvProcessEnv::read_all failed`. This error message comes from #3550, which attaches the calling function as the context for every failed turbo function.

This PR accomplishes 2 main things:
1. Expose an explicit method determining whether a the `DotenvEnvProcess`'s prior or current env failed
2. Detects a current env failure and uses the error's root cause for the issue message.

Screen Shot 2023-04-04 at 7 29 52 PM

Fixes WEB-851
Tests: https://github.com/vercel/next.js/pull/47937


crates/turbo-tasks-env/src/dotenv.rs
```diff
@@ -16,22 +16,34 @@ pub struct DotenvProcessEnv {
...
-            Some(p.read_all().await?)
-        } else {
-            None
+impl DotenvProcessEnv {
+    /// Attempts to assemble the EnvMapVc for our dotenv file. If either the
+    /// prior fails to read, or the current dotenv can't be parsed, an
+    /// appropriate Ok(DotenvReadResult) will be returned. If an unexpected
+    /// error (like disk reading or remote cache access) fails, then a regular
+    /// Err() will be returned.
+    pub async fn try_read_all(&self) -> Result<DotenvReadResult> {
```
sokra, line 40: You could avoid all this complicate logic by making in a two step process:

```
#[turbo_tasks::function]
pub fn read_all_prior(self) -> Result<EnvMapVc> { ... }

#[turbo_tasks::function]
pub fn read_all_with_prior(self, prior: EnvMapVc) -> Result<EnvMapVc> { ... }
```

crates/turbopack-env/src/try_env.rs
```diff
@@ -1,62 +1,61 @@
...
+                    // file, which we don't really care about (we report the filepath as the Issue
+                    // context, not the description). So extract the real error.
+                    description: StringVc::cell(e.root_cause().to_string()),
+                }
+                .cell()
+                .as_issue()
+                .emit();
+                Ok(self.prior.read_all())
+            }
+        }
```
sokra, line 59: With the two step process you can handle it like:

```
let prior = dotenv.read_all_prior();
let prior = match prior.resolve().await {
  Ok(p) => p,
  Err(err) => { /* handle prior error */ }
}
match dotenv.read_all_with_prior(prior).await {
  Ok(p) => { ... }
  Err(err) => { /* handle current error */ }
}
```

sokra (changes requested):

jridgewell (commented):
Oh wow, I don't know why I didn't see the potential to split the methods. Done.

sokra (approved):

PR #4470, opened 2023-04-05
Update swc_core to `v0.75.22`

### LINKED ISSUE #2825: [turbopack] esm interrupt error when transform to lower browserslist_query

1.6.3

npm

Mac

### Describe the Bug

import a cjs module in esm module with a lower browserslist_query, will throw `Uncaught ReferenceError: module is not defined`.

### Expected Behavior

should not have the ReferenceError.

### To Reproduce

Change the browserslist_query:
Screen Shot 2022-11-24 at 14 08 48

```js
// index.js
import lib from './lib.js';
console.log(lib());
```

```js
// lib.js
module.exports = async function() {
  return 1;
}
```

### Reproduction Repo

_No response_

### LINKED ISSUE #4092: [turbopack] css source code failed: filter: alpha(opacity=100)

Next 13.2.4-canary.4

Yarn v1

Windows

### Describe the Bug

Turbopack is not able to compile some CSS which contains an alpha filter with an `opacity=N` setting. This seems to be an older filter for IE 5~9, but seems to be valid CSS regardless. I am using a node package which uses this CSS, resulting in an error when navigating to "/":

```
error - [parse] [project-with-next]/node_modules/react-responsive-carousel/lib/styles/carousel.css  [project-with-next]/node_modules/react-responsive-carousel/lib/styles/carousel.css:229:29  Parsing css source code failed
     225 |       display: inline-block;
     226 |       margin: 0 8px; }
     227 |       .carousel .control-dots .dot.selected, .carousel .control-dots .dot:hover {
     228 |         opacity: 1;
         |                              v
     229 +         filter: alpha(opacity=100); }
         |                              ^
     230 |   .carousel .carousel-status {
     231 |     position: absolute;
     232 |     top: 0;
     233 |     right: 0;
  
  Expected Declaration value
  ```

### Expected Behavior

Compiling without errors

### To Reproduce

`yarn && yarn run dev`

### Reproduction Repo

https://github.com/daniel-koudouna/turbo-css-opacity-parse

### PR DESCRIPTION
**Pull request by kdy1**
### Description

This update includes an improvement of `@swc/helpers`, so we need to update `@swc/helpers` too when updating next-swc.

### Testing Instructions

Fix WEB-744
Fix WEB-859

---

 - Fixes #2825.
 - Fixes #4092.

### CODE REVIEW

**Review by padmaia — APPROVED** [review]

**Review by alexkirsz — APPROVED** [review]

File: crates/turbopack-ecmascript/src/transform/mod.rs [hunk 1]
```diff
@@ -215,9 +222,15 @@ impl EcmascriptInputTransform {
                     ..Default::default()
                 };
                 let p = std::mem::replace(program, Program::Module(Module::dummy()));
+                let hash = {
+                    let mut hasher = DefaultHasher::new();
+                    p.hash(&mut hasher);
```
sokra, line 227: That sounds pretty expensive. Do we need to hash the whole AST?

sokra (approved):

crates/turbopack-ecmascript/src/transform/mod.rs
kdy1: What I actually need is the hash of the source code. Would it be better?

sokra (approved):

crates/turbopack-ecmascript/src/transform/mod.rs
sokra: We can pass that via TransformContext

sokra: But how is the transform using that? Would a comment change have effects on the transform?

kdy1: It's used to generate a hash for the file.

Related: https://github.com/swc-project/plugins/pull/163

PR #4472, opened 2023-04-05
fix: typos

albertothedev:


albertothedev:
> nice, thank you!

Yesterday I tried updating the branch so it would be merged but the tests take long enough enough that there were other commits by the time it was ready so I left it for today. I have just updated it but it's asking for an approving review before merging:

FireShot Capture 005 - fix_ typos by albertothedev · Pull Request #4472 · vercel_turbo - github com

PR #4476, opened 2023-04-05
handle module evaluation errors in the runtime correctly

sokra:
### Description

Caches the module evaluation error

```
try {
  // when throwing on the first try
  require("module")
} catch {
  // it should throw on the second try too
  require("module")
}
```

### DIFF
--- a/crates/turbopack-dev/js/src/runtime.js
+++ b/crates/turbopack-dev/js/src/runtime.js
@@ -166,6 +166,7 @@ function interopEsm(raw, ns, allowExportDefault) {
  */
 function esmImport(sourceModule, id, allowExportDefault) {
   const module = getOrInstantiateModuleFromParent(id, sourceModule);
+  if (module.error) throw module.error;
   const raw = module.exports;
   if (raw.__esModule) return raw;
   if (module.interopNamespace) return module.interopNamespace;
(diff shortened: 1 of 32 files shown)


crates/turbopack-dev/js/src/runtime.js
```diff
@@ -276,6 +279,7 @@ function instantiateModule(id, source) {
   /** @type {Module} */
   const module = {
     exports: {},
+    error: undefined,
```
alexkirsz, line 282: We shouldn't expose this on the module object itself. Let's store in a hash map instead.

ForsakenHarmony (approved):

alexkirsz (changes requested):

crates/turbopack-dev/js/src/runtime.js
sokra: hmm... That would make receiving the error really expensive.

```diff
-  return getOrInstantiateModuleFromParent(id, sourceModule).exports;
+  const module = getOrInstantiateModuleFromParent(id, sourceModule);
+  const internalState = internalState.get(module); // <- expensive
+  if (internalState.error) throw module.error;
+  return module.exports;
```

So I don't think it's worth the isolation here.

Also webpack is exposing this via `module.error` too, so it's probably even more compatible when exposing it.

PR #4477, opened 2023-04-05
add ServerDirective transform which reports unsupported

sokra:
### Description

"use server" isn't supported yet, but we want to have an error message


jridgewell (commented):
LGTM

PR #4478, opened 2023-04-05
chore: Remove release flag from turborepo-ffi build

NicholasLYang:
### Description

We build turborepo-ffi on release mode currently for our development. This slows down compilation speeds, so I switched us to dev mode.

### Testing Instructions

Existing tests and CI should work.

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -42,7 +42,7 @@ turborepo-ffi-install: turborepo-ffi turborepo-ffi-copy-bindings
 
 .PHONY: turborepo-ffi
 turborepo-ffi:
-	cd ../crates/turborepo-ffi && cargo build --release --target-dir ./target
+	cd ../crates/turborepo-ffi && cargo build --target-dir ./target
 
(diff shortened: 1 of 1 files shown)


mehulkar (approved):

cli/Makefile
```diff
@@ -42,7 +42,7 @@ turborepo-ffi-install: turborepo-ffi turborepo-ffi-copy-bindings
 
 .PHONY: turborepo-ffi
 turborepo-ffi:
-	cd ../crates/turborepo-ffi && cargo build --release --target-dir ./target
+	cd ../crates/turborepo-ffi && cargo build --target-dir ./target
```
mehulkar, line 45: how/where is release mode configured?

arlyon: There are other tasks that are used during cross-compile that do use release mode. See https://github.com/vercel/turbo/blob/2d979d2f42e30322695a333373b35b5908196df5/cli/Makefile#L79-L97

arlyon (approved):

PR #4479, opened 2023-04-05
chore: Add profiles to cargo using aliases

NicholasLYang:
### Description

Some of the cargo commands in test.yml were getting unwieldy. Plus it's useful to have a simple command to run for testing turborepo code versus turbopack code.

### Testing Instructions

Existing test.yml code should verify that it works.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,6 +6,13 @@ linker = "rust-lld"
 
 [alias]
 xtask = "run --package xtask --"
+tr-build = "build -p turbo"
+tr-run = "run -p turbo"
+tr-test = "test -p turborepo-lib -p turborepo-scm -p turborepo-lockfiles -p turbopath -p turborepo-api-client --features rustls-tls"
+tr-check = "check -p turbo"
+# Builds all test code to check for compiler errors before running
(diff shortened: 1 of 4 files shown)


.cargo/config.toml
```diff
@@ -6,6 +6,13 @@ linker = "rust-lld"
 
 [alias]
 xtask = "run --package xtask --"
+tr-build = "build -p turbo"
+tr-run = "run -p turbo"
+tr-test = "test -p turborepo-lib -p turborepo-scm -p turborepo-lockfiles -p turborepo-paths --features rustls-tls"
+tr-check = "check -p turbo"
+# Builds all test code to check for compiler errors before running
+tp-pre-test = "nextest run --no-run --workspace --release --exclude turbo --exclude turborepo-ffi --exclude turborepo-lib --exclude turborepo-scm --exclude turborepo-paths --exclude turborepo-lockfiles"
+tp-test = "nextest run --workspace --release --no-fail-fast --exclude turbo --exclude turborepo-ffi --exclude turborepo-lib --exclude turborepo-scm --exclude turborepo-paths --exclude turborepo-lockfiles"
```
mehulkar, line 15: is this essentially the equivalent of package.json `scripts`?

mehulkar (approved):

.cargo/config.toml
NicholasLYang: Kinda, although these are far less used because generally you're recommended to use a regular build command like `cargo build` and then put secondary build script logic in `build.rs`

PR #4488, opened 2023-04-06
add included metadata to chunks to allow deduplicate chunk loads at runtime

sokra:
### Description

Chunk loading is now more than just a single chunk filepath. It might consist of additional metadata which tells the chunk loading logic the entry modules included in this chunk. This allows the runtime logic to avoid loading chunks when all entry modules are already loaded.

This means loading the same chunk from multiple sources (with different available modules) will only trigger a single load and so avoids transferring more code than needed.

This is especially relevant in app dir where every client component is considers as separate chunk loading.

next.js: https://github.com/vercel/next.js/pull/48025


crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -116,6 +117,19 @@ pub trait Chunk: Asset {
     }
 }
 
+#[turbo_tasks::value(shared)]
+#[derive(Default)]
+pub struct OutputChunkRuntimeInfo {
```
alexkirsz, line 122: Add a comment to explain why this exists/how this struct is used.

crates/turbopack-dev/js/src/runtime.dom.js
```diff
@@ -17,28 +17,38 @@ let BACKEND;
...
+            : otherChunkData.path;
         if (otherChunkPath.endsWith(".css")) {
           // Mark all CSS chunks within the same chunk group as this chunk as loaded.
+          // They are just injected as <link> tag and have to way to communicate completion.
           const cssResolver = getOrCreateResolver(otherChunkPath);
           cssResolver.resolve();
         } else if (otherChunkPath.endsWith(".js")) {
-          // Only wait for JS chunks to load.
-          chunksToWaitFor.push(otherChunkPath);
+          // Chunk might started loading, so we want to avoid triggering another load.
```
alexkirsz, line 31:
```suggestion
          // Chunk might have started loading, so we want to avoid triggering another load.
```

crates/turbopack-dev/js/src/runtime.dom.js
```diff
@@ -17,28 +17,38 @@ let BACKEND;
...
+          getOrCreateResolver(otherChunkPath);
         }
       }
 
-      if (params.runtimeModuleIds.length > 0) {
-        await waitForChunksToLoad(chunksToWaitFor);
+      // This wait for chunks to be loaded, but also marks included items as available.
```
alexkirsz, line 36:
```suggestion
      // This waits for chunks to be loaded, but also marks included items as available.
```

crates/turbopack-dev/js/src/runtime.dom.js
```diff
@@ -17,28 +17,38 @@ let BACKEND;
       }
 
       const chunksToWaitFor = [];
```
alexkirsz, line 19:
```suggestion
```

crates/turbopack-dev/js/src/runtime.js
```diff
@@ -209,12 +210,42 @@ externalRequire.resolve = (name, opt) => {
...
+      return availableModules.get(included);
+    });
+    if (promises.length > 0 && promises.every((p) => p)) {
+      // When all included items are already loaded or loading, we can skip loading ourselves
+      return Promise.all(promises);
+    }
+    const promise = loadChunkPath(source, chunkData.path);
+    for (const included of includedList) {
+      if (!availableModules.has(included))
+        availableModules.set(included, promise);
```
alexkirsz, line 237: The two promises should race.

1. Load chunk A with modules 1, 2
2. Load chunk B with modules 2, 3
3. Chunk B loads 
4. Load chunk C with module 2
5. Chunk A loads

In the current situation, chunk `C` will only be marked as loaded once `A` loads, but it should actually be marked as loaded instantly.

crates/turbopack-dev/js/src/runtime.nodejs.js
```diff
@@ -13,11 +13,16 @@ let BACKEND;
...
-        for (const otherChunkPath of params.otherChunks) {
-          loadChunk(otherChunkPath, {
-            type: SourceTypeRuntime,
-            chunkPath,
-          });
+        for (const otherChunkData of params.otherChunks) {
+          loadChunk(
+            typeof otherChunkData === "string"
+              ? otherChunkData
+              : otherChunkData.path,
```
alexkirsz, line 20: Should be a function `getChunkPath(chunkData)`, we do this very often.

crates/turbopack-dev/src/ecmascript/chunk_data.rs
```diff
@@ -0,0 +1,72 @@
...
+use turbo_tasks_fs::FileSystemPath;
+use turbopack_core::{
+    asset::{Asset, AssetVc},
+    chunk::{ModuleIdReadRef, OutputChunk, OutputChunkRuntimeInfo, OutputChunkVc},
+};
+
+#[derive(Serialize, Deserialize, Hash, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum ChunkData {
+    Simple(String),
```
alexkirsz, line 15:
```suggestion
    Simple {
        #[serde(flatten)]
        path: String,
    },
```

crates/turbopack-dev/src/ecmascript/chunk_data.rs
```diff
@@ -0,0 +1,72 @@
...
+        // The difference is a chunk is a file served by the dev server, and an
+        // item is one of several that are contained in that chunk file.
+        let Some(path) = output_root.get_path_to(&*path) else {
+        return Ok(None);
+    };
+        let path = path.to_string();
+
+        let Some(output_chunk) = OutputChunkVc::resolve_from(chunk).await? else {
+        return Ok(Some(ChunkData::Simple(path)));
+    };
```
alexkirsz, line 41:
```suggestion
        let path = chunk.ident().path().await?;
        // The "path" in this case is the chunk's path, not the chunk item's path.
        // The difference is a chunk is a file served by the dev server, and an
        // item is one of several that are contained in that chunk file.
        let Some(path) = output_root.get_path_to(&*path) else {
            return Ok(None);
        };
        let path = path.to_string();

        let Some(output_chunk) = OutputChunkVc::resolve_from(chunk).await? else {
            return Ok(Some(ChunkData::Simple(path)));
        };
```

crates/turbopack-dev/src/ecmascript/manifest/chunk_item.rs
```diff
@@ -39,17 +39,13 @@ impl EcmascriptChunkItem for DevManifestChunkItem {
...
-            if let Some(path) = output_root.get_path_to(chunk_path) {
-                chunk_server_paths.insert(path.to_string());
+        let mut chunks_data = IndexSet::new();
+        for &chunk in chunks.iter() {
+            // The chunk data is necessary to load the chunk from the server.
+            if let Some(chunk_data) = ChunkData::from_asset(&output_root, chunk).await? {
+                chunks_data.insert(chunk_data);
             } else {
-                // ignore all chunks that are not in the output root
+                // ignore all chunks that doesn't have chunk data
```
alexkirsz, line 48:
```suggestion
                // ignore all chunks that don't have chunk data
```

crates/turbopack-dev/src/ecmascript/manifest/loader_item.rs
```diff
@@ -88,23 +89,23 @@ impl EcmascriptChunkItem for DevManifestLoaderItem {
...
-                chunk.ident().path().to_string().await?,
-                manifest.chunking_context.output_root().to_string().await?
-            );
-        };
+        let chunk_server_data =
+            if let Some(data) = ChunkData::from_asset(&output_root, chunk).await? {
+                data
+            } else {
+                bail!(
+                    "chunk path {} is not in output root {}",
```
alexkirsz, line 104: This error is no longer correct.

alexkirsz (changes requested):

crates/turbopack-dev/js/src/runtime.nodejs.js
sokra: It's only used once in this file

crates/turbopack-dev/src/ecmascript/chunk_data.rs
sokra: serde bug preventing that

alexkirsz (approved):

PR #4491, opened 2023-04-06
Allow passing arbitrary data to static/proxy renders

jridgewell:
This adds an arbitrary `serde_json::Value` data field to our generic rendering code, allowing us to pass down the `NextConfig.output` field (and anything else we want) to the Node.js renderer.

Re: WEB-842

### DIFF
--- a/crates/turbo-tasks/src/read_ref.rs
+++ b/crates/turbo-tasks/src/read_ref.rs
@@ -4,6 +4,7 @@ use std::{
     fmt::{Debug, Display},
     hash::Hash,
     marker::PhantomData,
+    mem::transmute,
     sync::Arc,
 };
 
@@ -37,7 +38,7 @@ impl<T, U> std::ops::Deref for ReadRef<T, U> {
 
     fn deref(&self) -> &Self::Target {
         let inner: &T = &self.0;
-        unsafe { std::mem::transmute(inner) }
+        unsafe { transmute(inner) }
     }
 }
 
@@ -130,17 +131,18 @@ impl<T, U: Serialize> Serialize for ReadRef<T, U> {
     }
 }
 
-impl<'de, T: Deserialize<'de>> Deserialize<'de> for ReadRef<T, T> {
+impl<'de, T, U: Deserialize<'de>> Deserialize<'de> for ReadRef<T, U> {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: serde::Deserializer<'de>,
     {
-        let value = T::deserialize(deserializer)?;
-        Ok(Self(Arc::new(value), PhantomData))
+        let value = Arc::new(U::deserialize(deserializer)?);
+        let inner = unsafe { transmute(value) };
+        Ok(Self(inner, PhantomData))
     }
 }
 
-impl<T> ReadRef<T, T> {
+impl<T, U> ReadRef<T, U> {
     pub fn new(arc: Arc<T>) -> Self {
         Self(arc, PhantomData)
     }
--- a/crates/turbo-tasks/src/read_ref.rs
+++ b/crates/turbo-tasks/src/read_ref.rs
@@ -122,12 +122,12 @@ where
     }
 }
 
-impl<T, U: Serialize> Serialize for ReadRef<T, U> {
+impl<T: Serialize, U> Serialize for ReadRef<T, U> {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: serde::Serializer,
     {
-        (**self).serialize(serializer)
+        T::serialize(&self.0, serializer)
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/turbopack-node/src/render/mod.rs
```diff
@@ -19,6 +20,7 @@ pub struct RenderData {
     raw_query: String,
     raw_headers: Vec<(String, String)>,
     path: String,
+    data: Option<JsonValue>,
```
sokra, line 23:
```suggestion
    data: Option<turbo_tasks::JsonValueReadRef>,
```

That should allow to serialize it without cloning

crates/turbopack-node/src/render/node_api_source.rs
```diff
@@ -148,6 +156,7 @@ impl GetContentSourceContent for NodeApiGetContentResult {
                 raw_query: raw_query.clone(),
                 raw_headers: raw_headers.clone(),
                 path: format!("/{}", self.path),
+                data: Some(self.render_data.await?.clone_value()),
```
sokra, line 159:
```suggestion
                data: Some(self.render_data.await?),
```

crates/turbopack-node/src/render/rendered_source.rs
```diff
@@ -219,6 +227,7 @@ impl GetContentSourceContent for NodeRenderGetContentResult {
                 raw_query: raw_query.clone(),
                 raw_headers: raw_headers.clone(),
                 path: format!("/{}", source.pathname.await?),
+                data: Some(self.render_data.await?.clone_value()),
```
sokra, line 230:
```suggestion
                data: Some(self.render_data.await?),
```

sokra (approved):

crates/turbopack-node/src/render/mod.rs
jridgewell: Done, but this required me to update the `Deserialize` impl for `ReadRef<T>` to `ReadRef<T, U>`. Please verify if I did that correctly.

crates/turbo-tasks/src/read_ref.rs
```diff
@@ -130,17 +131,18 @@ impl<T, U: Serialize> Serialize for ReadRef<T, U> {
...
-        Ok(Self(Arc::new(value), PhantomData))
+        let value = Arc::new(U::deserialize(deserializer)?);
+        let inner = unsafe { transmute(value) };
+        Ok(Self(inner, PhantomData))
     }
 }
 
-impl<T> ReadRef<T, T> {
+impl<T, U> ReadRef<T, U> {
     pub fn new(arc: Arc<T>) -> Self {
```
sokra, line 146: Changing that makes `new` unsafe.

crates/turbo-tasks/src/read_ref.rs
```diff
@@ -130,17 +131,18 @@ impl<T, U: Serialize> Serialize for ReadRef<T, U> {
     }
 }
 
-impl<'de, T: Deserialize<'de>> Deserialize<'de> for ReadRef<T, T> {
+impl<'de, T, U: Deserialize<'de>> Deserialize<'de> for ReadRef<T, U> {
```
sokra, line 134:
```suggestion
impl<'de, T: Deserialize<'de>, U> Deserialize<'de> for ReadRef<T, U> {
```

You can deserialize the T instead of U to avoid the unsafe

crates/turbo-tasks/src/read_ref.rs
```diff
@@ -130,17 +131,18 @@ impl<T, U: Serialize> Serialize for ReadRef<T, U> {
...
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: serde::Deserializer<'de>,
     {
-        let value = T::deserialize(deserializer)?;
-        Ok(Self(Arc::new(value), PhantomData))
+        let value = Arc::new(U::deserialize(deserializer)?);
+        let inner = unsafe { transmute(value) };
+        Ok(Self(inner, PhantomData))
```
sokra, line 141:
```suggestion
        let value = Arc::new(T::deserialize(deserializer)?);
        Ok(Self(value, PhantomData))
```

crates/turbo-tasks/src/read_ref.rs
sokra, line 134: Technically this is not sound since it allows you to construct any `ReadRef<T, U>` from deserialization, even where it's not guaranteed that T and U are binary identical. But I don't know a better way, so meh...

crates/turbo-tasks/src/read_ref.rs
```diff
@@ -130,17 +131,18 @@ impl<T, U: Serialize> Serialize for ReadRef<T, U> {
...
     }
 }
 
-impl<T> ReadRef<T, T> {
+impl<T, U> ReadRef<T, U> {
```
sokra, line 145:
```suggestion
impl<T> ReadRef<T, T> {
```

sokra (changes requested):

crates/turbo-tasks/src/read_ref.rs
jridgewell: Verified that this works, because transparent Values have a trivial tuple deserializer to convert from U (the inner type) -> T (tuple type), and non-transparent Values are `T = U`.

sokra (approved):

PR #4495, opened 2023-04-06
Update data sent to vercel for runs

mehulkar:
- Add context and command when starting a Run
- Update data sent for Tasks
- Fix PATCH endpoint to mark Run as done
- Improve error propagation


gsoltis (commented):
I think the command piece is not going to work as implemented. I'm going to make a PR that attempts to synthesize something that will be runnable by the user and we can figure out how to integrate it, along with the directory it will need to be run from.

cli/internal/runsummary/vercel.go
```diff
@@ -22,28 +26,91 @@ type vercelRunPayload struct {
...
 }
+
+func newVercelTaskPayload(taskSummary *TaskSummary) *vercelTask {
+	hit := taskSummary.CacheState.Local || taskSummary.CacheState.Remote
+	status := "MISS"
+	var source string
+	if hit {
+		source = "REMOTE"
+		status = "HIT"
+	}
```
gaspar09, line 96: source is not set on cache misses.

gaspar09 (dismissed):

cli/internal/runsummary/vercel.go
```diff
@@ -22,28 +26,91 @@ type vercelRunPayload struct {
...
+		Name:      taskSummary.Task,
+		Workspace: taskSummary.Package,
+		Hash:      taskSummary.Hash,
+		StartTime: startTime,
+		EndTime:   endTime,
+		Cache: vercelCacheStatus{
+			Status: status,
+			Source: source,
+		},
+		ExitCode:     *taskSummary.Execution.exitCode,
```
gaspar09, line 112: I'm finding some exitCodes for tasks are not being set.

mehulkar: they won't be set for failed tasks. that will be covered in #4524.

gsoltis (approved):

PR #4499, opened 2023-04-07
Synthesize a command for a Run

gsoltis:
- Switches `pkgInferencePath` to be a `RelativeSystemPath` (note "" token for not set)
 - Records `repositoryPath` in Run Payload
 - Switches timestamps over to `int64` since they milliseconds
 - Synthesizes a plausible command to produce an identical set of tasks and packages to execute

Note merge base is #4495 

### Testing Instructions

Added a new test for command syntheses

### DIFF
--- a/cli/internal/runsummary/vercel.go
+++ b/cli/internal/runsummary/vercel.go
@@ -13,10 +13,10 @@ type vercelRunPayload struct {
 	ID string `json:"vercelId,omitempty"`
 
 	// StartTime is when this run was started
-	StartTime int `json:"startTime,omitempty"`
+	StartTime int64 `json:"startTime,omitempty"`
 
 	// EndTime is when this run ended. We will never be submitting start and endtime at the same time.
-	EndTime int `json:"endTime,omitempty"`
+	EndTime int64 `json:"endTime,omitempty"`
 
 	// Status is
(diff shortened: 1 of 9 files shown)


gaspar09 (approved):

cli/internal/runsummary/vercel.go
```diff
@@ -49,31 +53,32 @@ type vercelTask struct {
...
 	Cache        vercelCacheStatus `json:"cache,omitempty"`
 	ExitCode     int               `json:"exitCode,omitempty"`
 	Dependencies []string          `json:"dependencies,omitempty"`
 	Dependents   []string          `json:"dependents,omitempty"`
 }
 
-func newVercelRunCreatePayload(runsummary *RunSummary) *vercelRunPayload {
-	startTime := int(runsummary.ExecutionSummary.startedAt.UnixMilli())
-	var context = "LOCAL"
+func (rsm *Meta) newVercelRunCreatePayload() *vercelRunPayload {
```
mehulkar, line 64: If it's going into the spaces payload, it should be part of the Run Summary. It seems like good information regardless. If you're good with that, we can merge this in and I'll update in my PR

gsoltis: Added to execution summary

PR #4516, opened 2023-04-10
use sorted data structure for npm lockfile

### LINKED ISSUE #4510: [turborepo] prune package-lock.json not stable

1.8.8

npm

Mac, Windows, docker

### Describe the Bug

Running `turbo prune --scope=<workspace> --docker` multiple times always generates a different package-lock.json.

This causes the docker build to never use the cache to install dependencies.

The keys in the generated JSON have a different order, see for example this screenshot:
image

This occurs on MacOS, Windows (WSL) and inside the docker build.

### Expected Behavior

The generated package-lock.json should be stable to allow docker to cache the build step.

### To Reproduce

1. Checkout the reproduction repo
2. Run `npx turbo prune --scope=worker --docker` to create pruned output
3. Copy `out/package-lock.json` somewhere…
4. Run `npx turbo prune --scope=worker --docker` again
5. Compare the newly created package-lock.json with the one from step 2.

The reproduction repo is kinda large, maybe I will create a smaller repro in the coming days.

### Reproduction Repo

https://github.com/GW2Treasures/gw2treasures.com/tree/9c86b97ea829b25ee89cc2b28a07ec36b1cd8a0d

chris-olszewski:
### Description
Fixes #4510 by using `BTreeMap` instead of a `HashMap` so we have stable ordering of the package entries.

### Testing Instructions

Setup a monorepo using NPM as a package manager. Run `turbo prune --scope=web --out-dir=a && turbo prune --scope=web --out-dir=b && diff a/package-lock.json b/package-lock.json` and expect a zero exit code.

### DIFF
--- a/crates/turborepo-lockfiles/src/npm.rs
+++ b/crates/turborepo-lockfiles/src/npm.rs
@@ -5,6 +5,8 @@ use serde_json::Value;
 
 use super::{Error, Lockfile, Package};
 
+type Map<K, V> = std::collections::BTreeMap<K, V>;
+
 // we change graph traversal now
 // resolve_package should only be used now for converting initial contents
 // of workspace package.json into a set of node ids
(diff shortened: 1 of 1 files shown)


mehulkar:
Guessing code owners need to be updates since turborepo team didn't get asked for review

mehulkar (approved):
Nice. Does this need to happen n the yarn/berry/pnpm Go implementations also? It is possible to write a test?

chris-olszewski:
> Does this need to happen n the yarn/berry/pnpm Go implementations also? 

No, the Go JSON library sorts maps automatically whereas serde doesn't make assumptions so if you want serialization to be ordered then you either need to change the underlying data structure to be ordered or a custom serializer. 

> It is possible to write a test?

Done

mehulkar (approved):

crates/turborepo-lockfiles/src/npm.rs
```diff
@@ -5,22 +5,24 @@ use serde_json::Value;
 
 use super::{Error, Lockfile, Package};
 
+type Map<K, V> = std::collections::BTreeMap<K, V>;
```
sokra, line 8: You might want to consider IndexMap here. It's a hash map that keeps insertion order. That's more performant than sorting. Doesn't matter in this case, but just in general.

PR #4520, opened 2023-04-10
port(turborepo): Config

NicholasLYang:
### Description

Creates an `ExecutionState` struct that gets serialized to the Go code. The goal is to build up more and more of the info, such as config and graph state in this struct, so that we can do more work on the Rust side.


crates/turborepo-lib/src/config/repo.rs
```diff
@@ -21,50 +21,57 @@ pub struct RepoConfig {
 
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq, Default)]
 struct RepoConfigValue {
-    apiurl: Option<String>,
-    loginurl: Option<String>,
-    teamslug: Option<String>,
-    teamid: Option<String>,
+    #[serde(rename = "apiurl")]
```
mehulkar, line 24: what are all the renames for?

mehulkar (approved):
The code makes sense to me (to the extent that I can pick out the nuances of Rust). I'd be a little wary of merging this without some manual testing though if you haven't already because we have no automated checks for remote cache config and it's a high risk feature.

crates/turborepo-lib/src/config/repo.rs
NicholasLYang: `loginurl` is not idiomatic Rust naming, so I figured it's better to have `login_url` in the Rust code then rename it to `loginurl` via Serde

mehulkar: that makes sense, but I wasn't sure where the non-idiomatic versions (e.g. `loginurl`) are used and why we can't replace that?

cli/internal/run/run.go
```diff
@@ -43,18 +43,18 @@ Arguments passed after '--' will be passed through to the named tasks.
...
+func ExecuteRun(ctx gocontext.Context, helper *cmdutil.Helper, signalWatcher *signals.Watcher, executionState *turbostate.ExecutionState) error {
+	base, err := helper.GetCmdBase(executionState)
 	LogTag(base.Logger)
 	if err != nil {
 		return err
 	}
-	tasks := args.Command.Run.Tasks
-	passThroughArgs := args.Command.Run.PassThroughArgs
+	tasks := executionState.CLIArgs.Command.Run.Tasks
+	passThroughArgs := executionState.CLIArgs.Command.Run.PassThroughArgs
```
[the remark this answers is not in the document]
NicholasLYang: Hmm well CLIArgs is really just `ParsedArgsFromRust` in turbostate. If you want an example of what's actually in it, you could probably log that out on the Go side in main.go

crates/turborepo-lib/src/config/repo.rs
NicholasLYang: It's used in the config, and we want to avoid accidentally signing people out/unlinking them when we move from the Go to the Rust version.

crates/turborepo-lib/src/config/user.rs
```diff
@@ -157,6 +158,18 @@ mod test {
...
+        writeln!(&mut config_file, "{{\"token\": \"baz\"}}")?;
+        let env = {
+            let mut map = HashMap::new();
+            map.insert("VERCEL_ARTIFACTS_TOKEN".into(), "bat".into());
+            map
+        };
+        let config = UserConfigLoader::new(config_file.path().to_path_buf())
+            .with_environment(Some(env))
+            .load()?;
+        assert_eq!(config.token(), Some("bat"));
```
chris-olszewski, line 172: Nit, but this seems like we should either refactor this into a function to be shared across tests or make this a table test instead of copy-pasta for each new environment override

crates/turborepo-lib/src/config/user.rs
```diff
@@ -92,7 +92,8 @@ impl UserConfigLoader {
 
         let config = Config::builder()
             .add_source(raw_disk_config.clone())
-            .add_source(Environment::with_prefix("turbo").source(environment))
+            .add_source(Environment::with_prefix("TURBO").source(environment.clone()))
+            .add_source(Environment::with_prefix("VERCEL_ARTIFACTS").source(environment))
```
chris-olszewski, line 96: Not important, but these prefixes all get made lowercase: https://github.com/mehcode/config-rs/blob/master/src/env.rs#L247

crates/turborepo-lib/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "MPL-2.0"
 [features]
 # Allows configuring a specific tls backend for reqwest.
 # See top level Cargo.toml for more details.
+default = ["rustls-tls"]
```
chris-olszewski, line 10: Just verifying that this is meant to be part of this PR and didn't sneak in from dev workflow.

chris-olszewski (approved):

cli/internal/client/client.go
```diff
@@ -47,14 +48,6 @@ func (c *APIClient) SetToken(token string) {
 	c.token = token
 }
 
-// RemoteConfig holds the authentication and endpoint details for the API client
-type RemoteConfig struct {
-	Token    string
-	TeamID   string
-	TeamSlug string
-	APIURL   string
-}
```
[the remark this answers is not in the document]
NicholasLYang: Yeah I'm gonna block this PR on #4309

crates/turborepo-lib/src/config/user.rs
NicholasLYang: yeah I mostly made them uppercase because I tend to see environment variables in uppercase and it was confusing me to see the prefixes as lowercase.

crates/turborepo-lib/Cargo.toml
NicholasLYang: I'll make this a separate PR but I got sick of getting the feature error message. If turbopack is cool with this, I'm just gonna turn it on here

cli/internal/cmdutil/cmdutil.go
```diff
@@ -155,15 +155,7 @@ func (h *Helper) GetCmdBase(cliConfig *turbostate.ParsedArgsFromRust) (*CmdBase,
...
 	}
-	repoConfig, err := config.ReadRepoConfigFile(config.GetRepoConfigPath(repoRoot), cliConfig)
-	if err != nil {
-		return nil, err
-	}
-	userConfig, err := config.ReadUserConfigFile(h.UserConfigPath, cliConfig)
-	if err != nil {
-		return nil, err
-	}
-	remoteConfig := repoConfig.GetRemoteConfig(userConfig.Token())
```
mehulkar, line 166: `userConfig.Token()` is used for more than just the remoteCache, so collapsing it down to just `remoteConfig` mayyy be confusing. Luckily, we will probably need a different token for remote cache and for other API endpoints soon, so this works in favor of that.

crates/turborepo-lib/src/config/user.rs
```diff
@@ -144,19 +145,36 @@ mod test {
...
+                let mut map = HashMap::new();
+                map.insert(env_var.into(), env_var_value.clone());
+                map
+            };
+            let config = UserConfigLoader::new(config_file.path().to_path_buf())
+                .with_environment(Some(env))
+                .load()?;
+
+            assert_eq!(config.token(), Some(&env_var_value));
+        }
```
mehulkar, line 176: Would be helpful to have a test showing which env var takes precedence. From what I understand this tests that either one works. I was just trying to answer what happens when both are set.

turborepo-tests/integration/tests/config.t
mehulkar, line 1: Looks like this is testing _everything_ that comes in from Rust to Go. Can we call this something to reflect that? E.g. `rust-to-go-payload.t` or `arg-passing.t` or something?

turborepo-tests/integration/tests/config.t
```diff
@@ -0,0 +1,399 @@
...
+          ],
+          "pass_through_args": [],
+          "experimental_space_id": null
+        }
+      }
+    }
+  }
+
+Run test run with team overloaded from both env and flag (flag should take precedence)
+  $ TURBO_TEAM=vercel ${TURBO} run build --__test-run --team turbo
```
mehulkar, line 270: One more test for both `TURBO_TOKEN` and `--token` and `VERCEL_ARTIFACTS_TOKEN` would be awesome

turborepo-tests/integration/tests/config.t
```diff
@@ -0,0 +1,399 @@
...
+          ],
+          "pass_through_args": [],
+          "experimental_space_id": null
+        }
+      }
+    }
+  }
+
+Run test run with token overloaded
+  $ ${TURBO} run build --__test-run --token 1234567890
```
mehulkar, line 138: can we ` | jq .remoteConfig` on all these test since that's all that is being tested? Would make it easier to scan

cli/internal/cmdutil/cmdutil.go
NicholasLYang: Yes, but all of those uses as far as I can tell are now ported to Rust

mehulkar: Moving to rust is fine, collapsing it to a remote config token is the issue. Eg the token is used to post runs to spaces

cli/internal/cmdutil/cmdutil.go
```diff
@@ -74,12 +71,12 @@ func (h *Helper) Cleanup(cliConfig *turbostate.ParsedArgsFromRust) {
 	}
 }
 
-func (h *Helper) getUI(cliConfig *turbostate.ParsedArgsFromRust) cli.Ui {
+func (h *Helper) getUI(cliArgs *turbostate.ParsedArgsFromRust) cli.Ui {
 	colorMode := ui.GetColorModeFromEnv()
-	if cliConfig.GetNoColor() {
+	if cliArgs.NoColor {
```
mehulkar, line 76: For a different day, but wish this was just combined with `.Color`

cli/internal/cmdutil/cmdutil.go
mehulkar: yeah I like it.

cli/internal/turbostate/turbostate.go
```diff
@@ -97,45 +88,18 @@ type ParsedArgsFromRust struct {
...
-	}
-	return 0, fmt.Errorf("no remote cache timeout provided")
+// APIClientConfig holds the authentication and endpoint details for the API client
+type APIClientConfig struct {
+	Token        string `json:"token"`
+	TeamID       string `json:"team_id"`
+	TeamSlug     string `json:"team_slug"`
+	APIURL       string `json:"api_url"`
+	UsePreflight bool   `json:"use_preflight"`
+	Timeout      uint64 `json:"timeout"`
```
mehulkar, line 104: We can do this on the rust side, but it would be good to document where these values can come from  (flag, xdg config, `./.turbo/config.json`, env var, etc). I always forget that `teamId` and `token` are in separate places

turborepo-tests/integration/tests/remote-config.t
```diff
@@ -0,0 +1,42 @@
...
+    "team_id": null,
+    "team_slug": null,
+    "api_url": "https://vercel.com/api",
+    "use_preflight": false,
+    "timeout": 20
+  }
+
+Run test run with api overloaded
+  $ ${TURBO} run build --__test-run --api http://localhost:8000 | jq .remote_config.api
+  null
```
mehulkar, line 18: wait why is this null? shouldn't it be localhost:8000?

mehulkar (approved):
One question on a test case that looks unexpected, but otherwise looks good!

turborepo-tests/integration/tests/remote-config.t
NicholasLYang: Ah good catch, had the wrong field (`api` instead of `api_url`)

cli/internal/turbostate/turbostate.go
mehulkar: yeah but there are 2 different config files. hard to remember which one has what

NicholasLYang: Gotcha. Added a comment to make it clearer

PR #4521, opened 2023-04-10
feat(contextcondition): support InPath contextcondition

kwonoj:
### Description

Attempt to close WEB-862.

Currently turbopack in next-swc applies all of moduleoptionscontext into any kind of internal assets. Current available contextcondition (`ContextCondition::InDirectory`) does not gives enough context to determine where it comes from, makes non trivial to determine if given context need to be treated differently. 

~I'm trying to workaround by supplying addtional context for the filesystemkind, and caller can use it for creating a new contextcondition. In case of turbopack, I assume any embedded assets in the binary won't be an external / user's. Next-swc will assign a new condition like https://github.com/vercel/next.js/compare/canary...client-context-embedded-fs#diff-92086db0c6bc192f76dab5d612a562df8fddffea05244c1c00d58e7288879e11R213-R214.~

~Still bit unsure if this is a legit apporach or would be better way to go. PR is for the POC + further improvement based on the feedback.~

Per review suggestion, this PR now supports `InPath(FileSystemVc)` as new condition instead.

### DIFF
--- a/crates/turbopack/src/condition.rs
+++ b/crates/turbopack/src/condition.rs
@@ -1,6 +1,8 @@
+use async_recursion::async_recursion;
+use futures::{stream, StreamExt};
 use serde::{Deserialize, Serialize};
 use turbo_tasks::trace::TraceRawVcs;
-use turbo_tasks_fs::FileSystemPath;
+use turbo_tasks_fs::{FileSystem, FileSystemPath, FileSystemPathVc};
 
 #[derive(Debug, Clone, Serialize, Deserialize, TraceRawVcs, PartialEq, Eq)]
 pub enum ContextCondition {
@@ -8,6 +10,7 @@ pub enum ContextCondition {
     Any(Vec<ContextCondition>),
     Not(Box<ContextCondition>),
     InDirectory(String),
+    InPath(FileSystemPathVc),
 }
 
 impl ContextCondition {
@@ -27,12 +30,28 @@ impl ContextCondition {
         ContextCondition::Not(Box::new(condition))
     }
 
+    #[async_recursion]
     /// Returns true if the condition matches the context.
-    pub fn matches(&self, context: &FileSystemPath) -> bool {
+    pub async fn matches(&self, context: &FileSystemPath) -> bool {
         match self {
-            ContextCondition::All(conditions) => conditions.iter().all(|c| c.matches(context)),
-            ContextCondition::Any(conditions) => conditions.iter().any(|c| c.matches(context)),
-            ContextCondition::Not(condition) => !condition.matches(context),
+            ContextCondition::All(conditions) => {
+                stream::iter(conditions)
+                    .all(|c| async move { c.matches(context).await })
+                    .await
+            }
+            ContextCondition::Any(conditions) => {
+                stream::iter(conditions)
+                    .any(|c| async move { c.matches(context).await })
+                    .await
+            }
+            ContextCondition::Not(condition) => !condition.matches(context).await,
+            ContextCondition::InPath(path) => {
+                if let Ok(path) = path.await {
+                    context.is_inside(&path)
+                } else {
+                    false
+                }
+            }
             ContextCondition::InDirectory(dir) => {
                 context.path.starts_with(&format!("{dir}/"))
                     || context.path.contains(&format!("/{dir}/"))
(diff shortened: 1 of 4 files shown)


kdy1 (approved):

crates/turbopack/src/condition.rs
```diff
@@ -27,12 +30,22 @@ impl ContextCondition {
         ContextCondition::Not(Box::new(condition))
     }
 
+    #[async_recursion]
     /// Returns true if the condition matches the context.
-    pub fn matches(&self, context: &FileSystemPath) -> bool {
+    pub async fn matches(&self, context: &FileSystemPath) -> bool {
```
sokra, line 35: I think we tried to convert this to async before and saw a performance issue with that. Do you remember @alexkirsz ?

crates/turbopack/src/condition.rs
```diff
@@ -1,13 +1,16 @@
...
-use turbo_tasks_fs::FileSystemPath;
+use turbo_tasks_fs::{FileSystem, FileSystemKind, FileSystemPath};
 
 #[derive(Debug, Clone, Serialize, Deserialize, TraceRawVcs, PartialEq, Eq)]
 pub enum ContextCondition {
     All(Vec<ContextCondition>),
     Any(Vec<ContextCondition>),
     Not(Box<ContextCondition>),
     InDirectory(String),
+    Kind(FileSystemKind),
```
sokra, line 13: I would rather use something like `InPath(FileSystemPathVc)` here since the fs type is more an implementation detail. It should be transparent to the user how the fs is implementated.

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -70,6 +70,21 @@ use crate::{
     rope::{Rope, RopeReadRef, RopeReader},
 };
 
+#[turbo_tasks::value(shared)]
+#[derive(Debug, Clone, Copy)]
+pub enum FileSystemKind {
```
sokra, line 75: Some of the variants are implementation Details of other crates. It's also problematic to define more filesystems in other crates

sokra (changes requested):

crates/turbopack/src/condition.rs
kwonoj: If we change impl to `use something like InPath(FileSystemPathVc)`, how does match would work if we can't use async fn?

crates/turbopack/src/condition.rs
kwonoj: The part I am not sure is how to distinguish internal path via given `FileSystemPathVc`. I think I may miss something, but path read by filesystempathvc doesn't give context if it is internal to the turbopack or if it's user's. Mind elaborate what did I miss?

crates/turbopack/src/condition.rs
sokra: Benchmark seem to be happy, so nevermind

crates/turbopack/src/condition.rs
```diff
@@ -27,12 +30,31 @@ impl ContextCondition {
...
+                    .any(|c| async move { c.matches(context).await })
+                    .await
+            }
+            ContextCondition::Not(condition) => !condition.matches(context).await,
+            ContextCondition::InPath(path) => {
+                if let Ok(path) = path.await {
+                    path.fs
+                        .root()
+                        .await
+                        .map_or(false, |root| context.is_inside(&root))
```
sokra, line 53: This ignores the error here, but it's better to propagate the error upwards

crates/turbopack/src/condition.rs
```diff
@@ -27,12 +30,31 @@ impl ContextCondition {
...
+            ContextCondition::Any(conditions) => {
+                stream::iter(conditions)
+                    .any(|c| async move { c.matches(context).await })
+                    .await
+            }
+            ContextCondition::Not(condition) => !condition.matches(context).await,
+            ContextCondition::InPath(path) => {
+                if let Ok(path) = path.await {
+                    path.fs
+                        .root()
```
sokra, line 51:
```suggestion
```

crates/turbopack/src/condition.rs
kwonoj: So change `fn matches(...) -> bool` signature to `fn matches(..) -> Result<bool>`?

sokra (approved):

PR #4524, opened 2023-04-10
Gather Run Summary even when tasks fail

mehulkar:


cli/integration_tests/run_summary/errortask.t
```diff
@@ -0,0 +1,112 @@
...
+      "inferred": [],
+      "global": [
+        "SOME_ENV_VAR=",
+        "VERCEL_ANALYTICS_ID="
+      ]
+    },
+    "execution": {
+      "startTime": [0-9]+, (re)
+      "endTime": [0-9]+, (re)
+      "error": {},
```
mehulkar, line 57: this should probably be the error, not sure why it's empty (or maybe test fixture is wrong)

cli/internal/run/real_run.go
```diff
@@ -117,6 +114,16 @@ func RealRun(
 			mu.Unlock()
 		}
 
+		// On error, we'll close out all the other tasks
+		// but then continue on so we can close out the task summaries.
+		if err != nil {
+			if !ec.rs.Opts.runOpts.ContinueOnError {
+				ec.logger.Info("%s errored, canceling other tasks", packageTask.TaskID)
+				ec.processes.Close()
```
gsoltis, line 122: I'm not sure why this needs to move. It seems odd to have this error handling so far from where the error happened.

gsoltis (changes requested):
I don't think we should move the handling of the script error unless we really need to.

cli/internal/run/real_run.go
mehulkar: revert it. 

it made more sense to me that a child `exec()` doesn't reach up to the parent and tell it what to do; I prefer the child to just return the error and the parent decides what to do, but it's not _actually_ what my change did, it was still inside the Walk(). And regardless, it's not necessary for this change.

gsoltis (approved):

PR #4527, opened 2023-04-11
No env var inference in strict mode.

nathanhammond:
We presently use framework inference to determine if we should automatically include certain environment variables into the hash. In strict environment variable mode we should not automatically include any env vars into the hash or the execution environment.

You must instead:
- Specify all environment variables that you want available in your build environment.
- Specify whether you want those considered for the hash (`env`) or merely passed through (`passthroughEnv`).

We will also apply this to "inferred strict" on a per-task basis. If you specify `passthroughEnv` we infer that your _task_ is supposed to be in `strict` mode.

### DIFF
--- a/crates/turborepo-lib/src/cli.rs
+++ b/crates/turborepo-lib/src/cli.rs
@@ -56,6 +56,7 @@ pub enum DryRunMode {
 pub enum EnvMode {
     Infer,
     Loose,
+    StrictIncludeFrameworkVars,
     Strict,
 }
 
@@ -726,6 +727,26 @@ mod test {
                 ..Args::default()
             },
             "env_mode: specified strict"
+        );
+
+        assert_eq!(
+            Args::try_parse_from([
+                "turbo",
+                "run",
+                "build",
+                "--experimental-env-mode",
+                "strict-include-framework-vars"
+            ])
+            .unwrap(),
+            Args {
+                command: Some(Command::Run(Box::new(RunArgs {
+                    tasks: vec!["build".to_string()],
+                    env_mode: EnvMode::StrictIncludeFrameworkVars,
+                    ..get_default_run_args()
+                }))),
+                ..Args::default()
+            },
+            "env_mode: specified strict-include-framework-vars"
         );
 
         assert_eq!(
--- a/cli/internal/graph/graph.go
+++ b/cli/internal/graph/graph.go
@@ -83,7 +83,7 @@ func (g *CompleteGraph) GetPackageTaskVisitor(
 		useOldTaskHashable := false
 		if taskEnvMode == util.Infer {
 			if taskDefinition.PassthroughEnv != nil {
-				taskEnvMode = util.Strict
+				taskEnvMode = util.StrictIncludeFrameworkVars
 			} else {
 				// If we're in infer mode we have just detected non-usage of strict env vars.
 				// Since we haven't stabilized this we don't want to break their cache.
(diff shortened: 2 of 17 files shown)


mehulkar (commented):
this change should probably be paired with[^1] adding `envMode` and `experimentalGlobalPassThroughEnv` into `RunSummary` so it's obvious why env var changed between runs

[^1]: I'd prefer a separate PR, but not going a big deal

gsoltis:
I don't think we should skip inferred vars in strict mode. The point is to allow users to discover if they are relying on an env var that doesn't make it into the hash. If the env var does make it into the hash, for whatever reason, it should be fine to rely on it.

Is there perhaps another way that users could get the information about what was inferred? I think it's in the Run Summary, right?

crates/turborepo-lib/src/cli.rs
```diff
@@ -56,6 +56,7 @@ pub enum DryRunMode {
 pub enum EnvMode {
     Infer,
     Loose,
+    StrictIncludeFrameworkVars,
```
nathanhammond, line 59: The name here is a bikeshed. Note that the name introducing friction for using this is intentional.

Happy to entertain other options but this is already less aggressive than my original. 😅

nathanhammond (commented):
Reviewer's guide! Review commit-by-commit.

1. Adds a strict mode check.
2. Makes strict mode stricter, but creates a level underneath it that includes framework vars.
3. Is a docs patch for naming issues.

crates/turborepo-lib/src/cli.rs
gsoltis: I think we should go with something a bit shorter, at least for the user-facing portions. We're intending to communicate that the env vars available will match the set of env vars that `turbo` considers to be dependencies, and that means we're going infer some vars. Some ideas:
 - inferred
 - hashed
 - hashed-only
 - auto
 - detected

I'm least wild about the "hashed" variants since we don't want to force users to think about hashing. For better or worse they already need to think about inference at least a little, so I think that's less bad. "auto" and "detected" are pretty general. After writing this out, my vote goes to "auto". It suggests that `turbo` is doing something beyond what the user has specified, and I think it also contrasts nicely with "strict": "do your best guess at what will work" vs "don't do anything, I'm going to specify everything". I'm open to other [short] suggestions though.

nathanhammond: After implementation I don't think we can drop `infer` so we're going to be four states:
- `--infer`: we figure it out per-task based on `turbo.json`. (collapses to `loose` or `strict-include-framework-vars`)
- `--loose`: you get everything.
- `--strict`: we include only what you tell us.
- `--strict-include-framework-vars`: we include what you tell us and the framework vars.

This primary way to get this is: "you defined `passThroughEnv` but _don't_ pass in a command line flag". This is what gets you `strict-include-framework-vars`. In general the user should _rarely_ want or need to specify this.

The primary use case for manually specifying this flag is "I want strict mode for all my tasks, and for tasks that are Next.js I want to include `NEXT_PUBLIC_*`." Packages basically _never_ change frameworks, so using this is really just a feature request for "globbing over env vars" in `turbo.json`."

The problem here is that "auto", "detect", and "infer" are all synonyms and are extremely difficult to differentiate so we need two names for things which will basically never be specified on the command line, one of which is the "wrong" solution to the problem:
- "we pull it from turbo.json"
- "strict + framework vars"

nathanhammond: A lot of words to say:
- I think the clarity of length is valuable. This clarity also _improves_ the run summary.
- I believe that users will rarely, if ever, actually specify it.
- The friction will push people toward doing it the way we want them to, which is more-easily shareable: in `turbo.json`, in `env` and `passThroughEnv`, using a yet-unimplemented globbing feature.

nathanhammond:
Closing in favor of #4788.

PR #4529, opened 2023-04-11
Env var run summary data generation

nathanhammond:
This hoists the identification of environment mode much earlier in the process and captures the passthrough environment variables (hashed values included!) for run summary and `--dry=json`.


cli/internal/env/env.go
```diff
@@ -56,6 +56,10 @@ type EnvironmentVariablePairs []string
 // mapToPair returns a deterministically sorted set of EnvironmentVariablePairs from an EnvironmentVariableMap
 // It takes a transformer value to operate on each key-value pair and return a string
 func (evm EnvironmentVariableMap) mapToPair(transformer func(k string, v string) string) EnvironmentVariablePairs {
+	if evm == nil {
+		return nil
+	}
```
nathanhammond, line 61: Updates this to respect `nil` as a distinct value from `[]`.

cli/internal/graph/graph.go
```diff
@@ -111,6 +120,14 @@ func (g *CompleteGraph) GetPackageTaskVisitor(
 		packageTask.LogFile = logFile
 		packageTask.Command = command
 
+		var passthrough env.EnvironmentVariablePairs
+		if taskDefinition.PassthroughEnv != nil {
+			passthroughEnvDetailedMap, err := env.GetHashableEnvVars(taskDefinition.PassthroughEnv, nil, "")
+			if err == nil {
+				passthrough = passthroughEnvDetailedMap.All.ToSecretHashable()
+			}
+		}
```
nathanhammond: We need to capture the `ToSecretHashable()` value of these to make them extra-useful in run summary instead of them just being a list of env var names.

cli/internal/graph/graph.go
```diff
@@ -111,6 +120,13 @@ func (g *CompleteGraph) GetPackageTaskVisitor(
 		packageTask.LogFile = logFile
 		packageTask.Command = command
 
+		var envVarPassthroughMap env.EnvironmentVariableMap
+		if taskDefinition.PassthroughEnv != nil {
+			if envVarPassthroughDetailedMap, err := env.GetHashableEnvVars(taskDefinition.PassthroughEnv, nil, ""); err == nil {
+				envVarPassthroughMap = envVarPassthroughDetailedMap.BySource.Explicit
+			}
+		}
```
nathanhammond, line 128: This is the per-task population of passthrough vars.

cli/internal/taskhash/taskhash.go
```diff
@@ -277,18 +277,68 @@ func (th *Tracker) CalculateFileHashes(
...
+	task                 string
+	outputs              fs.TaskOutputs
+	passThruArgs         []string
+	hashableEnvPairs     []string
+	globalHash           string
+	taskDependencyHashes []string
+}
+
+// calculateTaskHashFromHashable returns a hash string from the taskHashable
+func calculateTaskHashFromHashable(full *taskHashable) (string, error) {
```
nathanhammond, line 307: This mirrors the `global` variant 1:1.

cli/internal/runsummary/task_summary.go
```diff
@@ -30,15 +30,18 @@ type TaskSummary struct {
...
 // TaskEnvVarSummary contains the environment variables that impacted a task's hash
 type TaskEnvVarSummary struct {
-	Configured []string `json:"configured"`
-	Inferred   []string `json:"inferred"`
-	Global     []string `json:"global"`
+	Configured        []string `json:"configured"`
+	Inferred          []string `json:"inferred"`
+	Global            []string `json:"global"`
+	Passthrough       []string `json:"passthrough"`
+	GlobalPassthrough []string `json:"globalPassthrough"`
```
nathanhammond, line 44: Please bikeshed naming.

nathanhammond (commented):
Reviewer's guide.

This is mostly a refactoring so that we capture the information early enough to get it into the run summary.

gsoltis (commented):
Behavior change LGTM, with a minor casing nit. 

Going to defer to @mehulkar for run summary changes.

cli/integration_tests/dry_json/monorepo.t
```diff
@@ -120,13 +121,16 @@ Setup
     },
     "expandedOutputs": [],
     "framework": "<NO FRAMEWORK DETECTED>",
+    "envMode": "Infer",
```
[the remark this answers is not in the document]
nathanhammond: https://github.com/vercel/turbo/pull/4529/commits/0a3f14c5601973926160970be2d6a9006c164440

cli/integration_tests/dry_json/monorepo.t
```diff
@@ -120,13 +121,16 @@ Setup
...
     "environmentVariables": {
       "configured": [],
       "inferred": [],
       "global": [
         "SOME_ENV_VAR=",
         "VERCEL_ANALYTICS_ID="
-      ]
+      ],
+      "passthrough": null,
+      "globalPassthrough": null
```
mehulkar, line 133: not to self, this is null because it's not included in the config. if the config sets it to an empty array that means something. so this is good 👍🏾

cli/internal/env/env.go
mehulkar, line 61: For my understanding, how would this ever be nil? Do the tests fail without this check?

cli/internal/graph/graph.go
```diff
@@ -76,12 +78,19 @@ func (g *CompleteGraph) GetPackageTaskVisitor(
...
+			taskEnvMode = util.Strict
+		}
+
 		// TODO: maybe we can remove this PackageTask struct at some point
 		packageTask := &nodes.PackageTask{
 			TaskID:          taskID,
 			Task:            taskName,
 			PackageName:     packageName,
 			Pkg:             pkg,
+			EnvMode:         taskEnvMode,
```
mehulkar, line 93: Is this used off `PackageTask` anywhere? If it's just for dry json / run summary, we might not need it here

cli/internal/graph/graph.go
```diff
@@ -76,12 +78,19 @@ func (g *CompleteGraph) GetPackageTaskVisitor(
 			return fmt.Errorf("Could not find definition for task")
 		}
 
+		// Task env mode is only independent when global env mode is `infer`.
+		taskEnvMode := globalEnvMode
+		if taskEnvMode == util.Infer && taskDefinition.PassthroughEnv != nil {
+			taskEnvMode = util.Strict
+		}
```
mehulkar, line 85: Hmm, this seems like new logic? In the docs, we wrote:

> If strict mode is specified or inferred, all tasks are run in strict mode, regardless of their configuration.
[source](https://turbo.build/repo/docs/reference/command-line-reference#--experimental-env-mode)

Is this not true or intentional? (I thought I got this bit from either the code or a comment in Slack (I didn't come up with it))

cli/internal/run/global_hash.go
```diff
@@ -33,52 +33,46 @@ type GlobalHashable struct {
...
-	}{
-		globalFileHashMap:    named.globalFileHashMap,
-		rootExternalDepsHash: named.rootExternalDepsHash,
-		hashedSortedEnvPairs: named.envVars.All.ToHashable(),
-		globalCacheKey:       named.globalCacheKey,
-		pipeline:             named.pipeline,
+}
+
+// calculateGlobalHashFromHashable returns a hash string from the globalHashable
+func calculateGlobalHashFromHashable(full GlobalHashable) (string, error) {
```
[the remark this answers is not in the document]
mehulkar: is this the cleaned up version or do we need a merge/rebase here?

cli/internal/run/real_run.go
```diff
@@ -72,6 +72,11 @@ func RealRun(
 
 	runCache := runcache.New(turboCache, base.RepoRoot, rs.Opts.runcacheOpts, colorCache)
 
+	globalEnvMode := rs.Opts.runOpts.EnvMode
+	if globalEnvMode == util.Infer && turboJSON.GlobalPassthroughEnv != nil {
+		globalEnvMode = util.Strict
+	}
```
mehulkar, line 78: I'd really like not to have any duplication between run and dry run (Took great pains to remove all the duplications that we had before). Is it possible to avoid thiS?

cli/internal/runsummary/task_summary.go
mehulkar: How about 
- `allowList` and `globalAllowList`? 
- `allowed` and `globalAllowed` (which follows the pattern of the configired/inferred)

cli/internal/taskhash/taskhash.go
```diff
@@ -277,18 +277,68 @@ func (th *Tracker) CalculateFileHashes(
...
+	switch full.envMode {
+	case util.Infer:
+		if full.passthroughEnv != nil {
+			// In infer mode, if there is any passThru config (even if it is an empty array)
+			// we'll hash the whole object, so we can detect changes to that config
+			// Further, resolve the envMode to the concrete value.
+			full.envMode = util.Strict
```
mehulkar, line 314: I see that we're "resolving" the mode to strict. Will this show up in Run Summary as `strict` or `infer`? We should do the same behavior in both places, or add `envMode` and `resolvedEnvMode`. Don't have a strong preference, but this is leading me to thing we need a `runOptions` key in Run Summary that shows the raw values typed by the user, and then the rest of the data is resolved.

cli/internal/taskhash/taskhash.go
```diff
@@ -277,18 +277,68 @@ func (th *Tracker) CalculateFileHashes(
...
+		})
+	case util.Loose:
+		// Remove the passthroughs from hash consideration if we're explicitly loose.
+		full.passthroughEnv = nil
+		return fs.HashObject(full)
+	default:
+		// When we aren't in infer or loose mode we can hash the whole object as is.
+		return fs.HashObject(full)
+	}
+}
```
mehulkar, line 340: Is this necessary? I can understand avoiding busting a global hash between releases, but seems overkill to avoid busting individual task hashes too. CAn we just `fs.HashObject(full)` for all these modes?

cli/internal/util/run_opts.go
```diff
@@ -12,6 +14,11 @@ const (
 	Strict EnvMode = "Strict"
 )
 
+// MarshalText implements TextMarshaler for the struct.
+func (s EnvMode) MarshalText() (text []byte, err error) {
+	return []byte(strings.ToLower(string(s))), nil
+}
```
mehulkar, line 20: Niice, I was wondering if there was a cheap way to do this. `JSON.Marshal()` works with this?

cli/internal/env/env.go
nathanhammond: You can manually create a nil `EnvironmentVariableMap`. This prevents transformation of `nil` to `[]`. And yes, tests fail without this change.

cli/internal/graph/graph.go
nathanhammond: It's not new logic, it starts by inheriting `globalEnvMode`'s value which will be `strict` in the scenario described in the docs. The only thing it does is collapse the `infer` state for the task when it has not already been locked.

cli/internal/graph/graph.go
nathanhammond: We need it for the task hash itself in case globalEnvMode is `infer`: https://github.com/vercel/turbo/pull/4529/files#diff-7790aa2d704a01d8951c2bdc87cdf80f13b234ba1b4db29cd1502e7b84e4d3e4R287

cli/internal/run/global_hash.go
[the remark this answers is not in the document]
nathanhammond: This is already rebased after the 4528 change.

cli/internal/run/real_run.go
nathanhammond: This can be avoided by hoisting the calculation earlier.

Aside: the way in which those were all collapsed into `GlobalHashable` is creating lots of pain because it now tightly couples real_run, dry_run, globalhash, and runsummary. Particularly, `envVars` is in an "interim state" and the rest are "resolved state". `envVars` is an `env.DetailedMap` and presently getting hashed as is when it should instead be `.All.ToHashable()`.

cli/internal/taskhash/taskhash.go
nathanhammond: Global: can present infer (meaning per-task), strict, or loose.
Task: can present only strict or loose.

I'll double check to make sure this is correct and there are tests for it.

cli/internal/taskhash/taskhash.go
nathanhammond: Hashing `full` here is effectively a global hash change since it applies to _all_ tasks by changing the object shape.

nathanhammond: Also, this is still "experimental" so if we start hashing it in we're committing to that hash shape for everybody before stabilizing it.

cli/internal/util/run_opts.go
nathanhammond: Yup. This is how we're going to make the global hash portable to Rust.

cli/internal/runsummary/task_summary.go
nathanhammond: We use "passthrough" in the config key, so I believe that continuing to do so is valuable.

cli/internal/graph/graph.go
nathanhammond: I've added a new commit that collapses this further so that the output for a task is always `loose` or `strict`.

cli/internal/taskhash/taskhash.go
nathanhammond: The updated commit ensures that this invariant is respected.

cli/internal/run/real_run.go
nathanhammond: Hoisted to `run`.

cli/internal/run/global_hash.go
```diff
@@ -47,28 +47,40 @@ type oldGlobalHashable struct {
...
+		}
 
-	// In infer mode, if there is any passThru config (even if it is an empty array)
-	// we'll hash the whole object, so we can detect changes to that config
-	if named.envVarPassthroughs != nil {
-		return fs.HashObject(named)
+		// If we're in infer mode, and there is no global pass through config,
+		// we use the old struct layout. this will be true for everyone not using the strict env
+		// feature, and we don't want to break their cache.
+		return fs.HashObject(oldGlobalHashable{
```
arlyon, line 64: I know I missed the boat here but is my assumption that `oldGlobalHashable` == 'non-strict mode' correct? If that is the case, then maybe renaming to reflect that would be an idea.

cli/internal/run/global_hash.go
```diff
@@ -47,28 +47,40 @@ type oldGlobalHashable struct {
...
+		full.envVarPassthroughs = nil
+		return fs.HashObject(full)
+	case util.Strict:
+		// Collapse `nil` and `[]` in strict mode.
+		if full.envVarPassthroughs == nil {
+			full.envVarPassthroughs = make([]string, 0)
+		}
+		return fs.HashObject(full)
+	default:
+		panic("unimplemented environment mode")
```
arlyon, line 82: Do we have lints set up to statically infer exhaustiveness (https://pkg.go.dev/github.com/nishanths/exhaustive@v0.10.0)? I don't particularly like this but if we don't have it it's probably not worth setting up now.

arlyon (approved):

cli/internal/run/global_hash.go
nathanhammond: We don't have exhaustiveness lints. I looked for it, it exists, and this is the only enum in the codebase that it would check.

I yearn for Rust's approach here.

cli/internal/run/global_hash.go
nathanhammond: There have been iterations on this where it used to be not "just env mode" ... we got here by path dependency and the external opinion is a good thought.

cli/integration_tests/run_summary/fixture-configs/all.json
```diff
@@ -0,0 +1,12 @@
...
+  "pipeline": {
+    "build": {
+      "outputs": ["dist/**"],
+      "experimentalPassThroughEnv": ["LOCAL_VAR_PT"],
+      "env": ["LOCAL_VAR_DEP"]
+    }
+  },
+  "experimentalGlobalPassThroughEnv": ["GLOBAL_VAR_PT"],
+  "globalEnv": ["GLOBAL_VAR_DEP"]
+}
```
mehulkar, line 12: Are these the exact same as the fixture configs in `strict_env_vars`? Don't love the copy paste, but I guess it doesn't have to block this getting out

cli/integration_tests/run_summary/strict_env.t
```diff
@@ -0,0 +1,206 @@
...
+  $ export GLOBAL_VAR_DEP=higlobaldep
+  $ export LOCAL_VAR_PT=hilocalpt
+  $ export LOCAL_VAR_DEP=hilocaldep
+  $ export OTHER_VAR=hiother
+  $ export SYSTEMROOT=hisysroot
+
+Run as `infer`
+  $ rm -rf .turbo/runs
+  $ ${TURBO} run build --summarize > /dev/null
+  $ cat .turbo/runs/*.json | jq -r '.envMode'
```
mehulkar, line 16: I understand that it's unnecessary if you're `rm -rf`'ing right before., but all the other tests explicitly get the first one with `head -n1`.

cli/integration_tests/run_summary/strict_env.t
```diff
@@ -0,0 +1,206 @@
...
+  loose
+  $ cat .turbo/runs/*.json | jq -r '.tasks[0].envMode'
+  loose
+  $ cat .turbo/runs/*.json | jq -r '.tasks[0].environmentVariables | {passthrough, globalPassthrough}'
+  {
+    "passthrough": null,
+    "globalPassthrough": [
+      "GLOBAL_VAR_PT=cecd31fff1e723588eac8fe1edff89a6d2ec072f5c3bd884a98297487670b1f0"
+    ]
```
mehulkar, line 145: Fine since we're still experimental, but I was under the impression that the _value_ of pass through vars is not part of the hash. If that's still the case, the run summary could also probably omit the values.

cli/internal/run/global_hash.go
mehulkar, line 82: - What's the benefit of using `panic` over bubbling up an error?
- Is it even possible to hit this case, if we're limiting the CLI flag to a set of known values?

mehulkar (approved):
Left some comments, but 🚢

cli/integration_tests/run_summary/fixture-configs/all.json
nathanhammond: I feel like I edited them, but I'll check.

cli/integration_tests/run_summary/strict_env.t
nathanhammond: You mentioned that this didn't guarantee ordering in some other test so I took this strategy.

Why was the ksuid ordering not guaranteed?

cli/integration_tests/run_summary/strict_env.t
nathanhammond: I think that your concern is unnecessarily conflating "is part of run summary" with "is part of hash". There are _lots_ of things in the output that aren't (intentionally) included in the hash so this doesn't strike me as a problem. We can address this in any UI: "highlight the hash inputs".

I explicitly went through to add this because being able to unmask that value in the future if you can guess the input to the hash is super-valuable to discover why the run outputs are what they are.

cli/internal/run/global_hash.go
nathanhammond: Bubbling up an error to the user implies that they have an ability to do something about this issue. They don't.

`panic` is for early notice of code authoring errors.

In general this pattern is the primitive version of exhaustiveness checking. "We blow up if you don't account for all your inputs." Rust has compiler-enforced exhaustiveness checking while Go only has a third-party linter.

This is presently impossible to hit but is designed to blow up if you add a new mode and don't implement it in the switch.

nathanhammond: In other words: our tests should hit this and then blow up if you or I failed to write this code.

cli/integration_tests/run_summary/strict_env.t
mehulkar: > I think that your concern is unnecessarily conflating "is part of run summary" with "is part of hash"

I didn't intend it as that. My view is that the summary is a good way to understand what the tool _did_.  If it's not doing anything with a thing, it doesn't need to be reflected back to the user. In this case, my earlier comment is promoted by the fact that we do _something_ with the env var keys, but nothing with their values.

cli/integration_tests/run_summary/strict_env.t
mehulkar: I don't remember the details to dig it up again, but I'm 99% sure that was an incorrect statement and I solved the issue I was seeing a different way. I think it's still safe to rely on ksuid ordering. (That said, it's not a requirement since `rm && cat` already does what we need).

cli/integration_tests/run_summary/strict_env.t
nathanhammond: But we _are_ doing something with it! We're passing that KV env pair into the environment! And it's going to be different on different people's boxes, but the hash will _still be the same_!

Of _all_ the places where we decide whether or not to capture this information I believe this is the _most_ important one to capture.

cli/integration_tests/run_summary/fixture-configs/all.json
nathanhammond: I didn't; I unified this.

cli/internal/util/run_opts.go
nathanhammond: Reader: this was _not_ how that was going to happen.

cli/integration_tests/run_summary/strict_env.t
nathanhammond: That's really my distinction too: `turbo` may not care other than shuffling the correct things into the environment, but the user definitely cares what actually ended up there.

We should be doing things for the _user_.

PR #4530, opened 2023-04-11
feat(summary): add taskId to single package task

tknickman:
### Description

`taskId` should always exist.


mehulkar (approved):
I don't have a strong objection with this, but:
- `taskId` all over the codebase is `pkg#task`, i'm concerned this exception is confusing
- this duplicates the task name in dry json. (it was originally implemented without that for this reason, I think?)

Other than that, I'm ok with this

cli/internal/runsummary/task_summary.go
```diff
@@ -51,11 +51,12 @@ func (ts *TaskSummary) cleanForSinglePackage() {
 	for i, dependent := range ts.Dependents {
 		dependents[i] = util.StripPackageName(dependent)
 	}
+	taskID := util.StripPackageName(ts.TaskID)
```
mehulkar, line 54: Can you keep this as `task` though, it's not a taskId anymore

```suggestion
	task := util.StripPackageName(ts.TaskID)
```

cli/internal/runsummary/task_summary.go
```diff
@@ -51,11 +51,12 @@ func (ts *TaskSummary) cleanForSinglePackage() {
...
 
-	ts.Task = util.RootTaskTaskName(ts.TaskID)
+	ts.TaskID = taskID
```
mehulkar, line 56:
```suggestion
	ts.TaskID = task
```

cli/internal/runsummary/task_summary.go
```diff
@@ -51,11 +51,12 @@ func (ts *TaskSummary) cleanForSinglePackage() {
...
+	ts.Task = taskID
```
mehulkar, line 57:
```suggestion
	ts.Task = task
```

tknickman:
> taskId all over the codebase is pkg#task, i'm concerned this exception is confusing

The task ID is a string ID unique to a single task in the array - it shouldn't tell you what the string contains, nor should assumptions be made about it. We provide task, and now package as separate keys which should be sufficient. 

> this duplicates the task name in dry json. (it was originally implemented without that for this reason, I think?)

I'm fine with this, since they have a different meaning (your task name vs the unique ID for the task)

mehulkar:
> The task ID is a string ID unique to a single task in the array

Hmm I don't think that's true, but we can say that now. If you look at some of the ways it's used, it's always assumed to be in the pkg#task format. Eg. look at `GetTaskId()` or any of the other helpers here: https://github.com/vercel/turbo/blob/main/cli/internal/util/task_id.go

mehulkar:
Uniqueness is a key property of `taskID` yes, but I'm just saying it's not the only key property of how taskId is used. As long as this change is restricted to the tail end of a run, we're ok, but if you removed the `//#` part of a `taskId` before the graph is constructed or tasks are executed, I'm pretty sure things would fail

PR #4531, opened 2023-04-11
fix(docs): context menu wide screen

tknickman:
### Description

Missed a class on the new context menu


mehulkar (approved):
curious, what does adding this fix?

tknickman:
The context menu is positioned absolutely, but that's relative to the nearest relative parent. Because I forgot to make the logo relative, if the screen is wide enough it's positioned relative to the window:


CleanShot 2023-04-11 at 13 56 02@2x

PR #4533, opened 2023-04-11
test(turborepo): Vercel API Mock Crate

NicholasLYang:
### Description

Consolidates the ad-hoc Vercel API mocks that we have in a single crate that spins up a test server. Users are responsible for finding their own open port.

### Testing Instructions

Existing link and login tests should cover this


mehulkar (commented):
Can I use this from prysk tests?

NicholasLYang:
Yeah I'll make it a binary that you can run

NicholasLYang:
Added a binary, so you can run `cargo run -p vercel-api-mock` and get the mock server. It currently allocates the port itself and prints it to stdout, but I can definitely make it so you can set the port too

crates/turborepo-vercel-api-mock/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+        );
+    let addr = SocketAddr::from(([127, 0, 0, 1], port));
+    // We print the port so integration tests can use it
+    println!("{}", port);
+    axum_server::bind(addr)
+        .serve(app.into_make_service())
+        .await?;
+
+    Ok(())
+}
```
mehulkar, line 80: I'd like to be able to inspect the request & response easily, is this possible? Instead of adding every endpoint, can we just make it accept any GET/POST request and be able to inspect it after the fact?

NicholasLYang: Hmm I feel like a catch-all endpoint might be better with a different server than this. This is mostly for situations where you need a specific endpoint mocked. Unless I'm misunderstanding?

crates/turbo-tasks-build/src/lib.rs
```diff
@@ -43,7 +43,7 @@ pub fn generate_register() {
     if lib_entry.exists() {
         entries.push(("register.rs".to_string(), lib_entry));
     } else {
-        let bin_entry = src_dir.join("main.rs");
+        let bin_entry = src_dir.join("lib");
```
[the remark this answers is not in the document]
NicholasLYang: Weird. No idea how that got in. Probably an IDE glitch

sokra (approved):

PR #4536, opened 2023-04-11
feat(docs): improved examples cards

tknickman:
### Description

Improved examples cards. Supports:

1. Link to github
2. Link to vercel to deploy (if available)
3. Copy create-turbo command


mehulkar (commented):
Nice. I'm not sure, but is there a need to add handling of prefers-reduced-motion media query? there isn't a lot of motion, but i  wasn't sure if the animation should be turned off also

mehulkar (approved):

tknickman:
Yea good suggestion, we do support this on the docs home - I can include it here as well just to be thorough. I'm going to tweak this a little more still as well.

PR #4539, opened 2023-04-12
Include TimeSaved metric in Run Summaries

mehulkar:


cli/internal/runsummary/task_summary.go
```diff
@@ -1,12 +1,17 @@
...
 	"github.com/vercel/turbo/cli/internal/fs"
 	"github.com/vercel/turbo/cli/internal/turbopath"
 	"github.com/vercel/turbo/cli/internal/util"
 )
 
+type TaskCacheSummary struct {
+	Local     bool `json:"local"`
+	Remote    bool `json:"remote"`
+	TimeSaved int  `json:"timeSaved"`
+}
```
mehulkar, line 13: I would like to just use `spacesCacheStatus` here, but that would change the `--dry=json` `cache` output.

cli/internal/taskhash/taskhash.go
```diff
@@ -426,19 +426,30 @@ func (th *Tracker) SetExpandedOutputs(taskID string, outputs []turbopath.Anchore
...
-		return cache.ItemStatus{Local: false, Remote: false}
+		return CacheDetails{
+			ItemStatus: cache.ItemStatus{Local: false, Remote: false},
+			TimeSaved:  0,
+		}
 	}
 	return status
 }
+
+type CacheDetails struct {
```
mehulkar, line 452: better name for this?

cli/integration_tests/dry_json/single_package_no_config.t
```diff
@@ -39,7 +39,9 @@ Setup
         "hashOfExternalDependencies": "",
         "cache": {
           "local": false,
-          "remote": false
+          "remote": false,
+          "status": "MISS",
```
mehulkar, line 43: This is a bit weird. I really don't want to customize for dry runs, but it also doesn't make too much sense to see "HIT" or "MISS" here since it didn't run. Should it say "WILL HIT" and "WILL MISS" instead?

cli/internal/runsummary/task_summary.go
```diff
@@ -7,6 +7,45 @@ import (
...
+type TaskCacheSummary struct {
+	Local     bool   `json:"local"`            // Deprecated, but keeping around for --dry=json
+	Remote    bool   `json:"remote"`           // Deprecated, but keeping around for --dry=json
+	Status    string `json:"status"`           // should always be there
+	Source    string `json:"source,omitempty"` // can be empty on status:miss
+	TimeSaved int    `json:"timeSaved"`        // always include, but can be 0
+}
+
+// NewTaskCacheSummary decorates a cache.ItemStatus into a TaskCacheSummary
+// Importantly, it adds the
```
mehulkar, line 21:
```suggestion
// Importantly, it adds the derived keys of `source` and `status` based on
// the local/remote booleans. It would be nice if these were just included
// from upstream, but that is a more invasive change.
```

cli/integration_tests/dry_json/single_package_no_config.t
mehulkar: or maybe it's fine, since local:false, remote:false also have the same problem

arlyon: yeah I think MISS or HIT is fine here. maybe 'outcome' is a better name than 'status'?

edit: I can see status has been around longer than this pr. ignore.

cli/internal/runcache/runcache.go
```diff
@@ -109,12 +109,12 @@ type TaskCache struct {
 
 // RestoreOutputs attempts to restore output for the corresponding task from the cache.
 // Returns true if successful.
```
arlyon, line 111: A comment here explaining the new return type would be nice. Error is self-explanatory.

arlyon (commented):
just one comment around a doc string, otherwise looks ok. only commenting, since I'm not super familiar w/ cache

cli/internal/cache/cache_fs.go
```diff
@@ -88,9 +88,9 @@ func (f *fsCache) Exists(hash string) ItemStatus {
...
 	var event string
 	if hit {
-		event = cacheEventHit
+		event = CacheEventHit
 	} else {
-		event = cacheEventMiss
+		event = CacheEventMiss
 	}
 	payload := &CacheEvent{
 		Source:   "LOCAL",
```
gsoltis, line 96: Maybe this should be using the new `CacheSourceFS` constant? Ditto for http cache / `CacheSourceRemote`

cli/internal/run/real_run.go
```diff
@@ -271,8 +271,11 @@ func (ec *execContext) exec(ctx gocontext.Context, packageTask *nodes.PackageTas
 		ErrorPrefix:  prettyPrefix,
 		WarnPrefix:   prettyPrefix,
 	}
-	cacheStatus, err := taskCache.RestoreOutputs(ctx, prefixedUI, progressLogger)
-	ec.taskHashTracker.SetCacheStatus(packageTask.TaskID, cacheStatus)
+	cacheStatus, timeSaved, err := taskCache.RestoreOutputs(ctx, prefixedUI, progressLogger)
+	ec.taskHashTracker.SetCacheStatus(
```
gsoltis, line 275: Need to check `err` before using return values, or document why it's safe anyways

cli/internal/runcache/runcache.go
gsoltis, line 111: Update comment to account for new return value. Actually, the return documentation is already out of date and needs updating either way.

cli/internal/runcache/runcache.go
```diff
@@ -172,7 +173,9 @@ func (tc *TaskCache) RestoreOutputs(ctx context.Context, prefixedUI *cli.Prefixe
 	default:
 		// NoLogs, do not output anything
 	}
-	return cacheStatus, nil
+	// TODO: timeSaved could be part of cacheStatus, so we don't have to make a new struct
+	// downstream, but this would be a more invasive change right now.
+	return cacheStatus, timeSaved, nil
```
gsoltis, line 178: I don't know how to fix this, but it's a bummer that we lose the time saved value in the most efficient case: no restoration necessary.

Maybe we put it on the daemon someday along with how long it took to generate those outputs? cc @arlyon

gsoltis (approved):

cli/integration_tests/run_summary/single-pacakge.t
```diff
@@ -82,5 +82,7 @@ Check
   $ echo $TASK_SUMMARY | jq '.cache'
```
chris-olszewski, line 82: Not related to this change, but `git mv single-pacakge.t single-package.t`

cli/integration_tests/dry_json/single_package_no_config.t
chris-olszewski: My two cents is that it's having `HIT`/`MISS` is fine as I feel like the "will be a" is implied by the dry run.

chris-olszewski (approved):
LGTM

cli/internal/runcache/runcache.go
mehulkar: Yeah, our timings could use a bit more sophistication in general. `timeSaved` doesn't account for the time it took to restore the cache at all, it's just how long the task took when it populated the cache.

PR #4545, opened 2023-04-12
Exclude Pipeline Definition from Global Hash

nathanhammond:
The global hash does not need to include the pipeline so long as all material elements to the pipeline are present in the task hash. After #4529 that invariant became true.

This PR does two things:
- Removes `pipeline` from the global hash key inputs.
- Adjusts the dry run JSON schema to move `rootPipeline` from `globalHashInputs` and instead attach it to a new root key, `pipeline`. This is a possibly-breaking change so I incremented the `runSummarySchemaVersion`.

🚨🚨🚨
This PR triggers a universal cache invalidation! The payoff is no longer breaking the cache any time `turbo.json` is updated.


gsoltis:
Let's hold this to batch with other global-hash busting changes.

cli/internal/fs/turbo_json.go
```diff
@@ -277,14 +277,9 @@ type TaskOutputs struct {
...
-	var inclusions []string
-	var exclusions []string
-	copy(inclusions, to.Inclusions)
-	copy(exclusions, to.Exclusions)
-	sort.Strings(inclusions)
-	sort.Strings(exclusions)
-	return TaskOutputs{Inclusions: inclusions, Exclusions: exclusions}
+func (to *TaskOutputs) Sort() {
+	sort.Strings(to.Inclusions)
+	sort.Strings(to.Exclusions)
```
mehulkar, line 282: Making sure my Go understanding is correct. Since `to` is not a pointer, this sort stays local to this function, and the `copy()` is unnecessary. Is that correct?

cli/internal/fs/turbo_json_test.go
```diff
@@ -212,10 +212,11 @@ func Test_TaskOutputsSort(t *testing.T) {
...
 	taskOutputs := TaskOutputs{Inclusions: inclusions, Exclusions: exclusions}
-	sortedOutputs := taskOutputs.Sort()
-	assertIsSorted(t, sortedOutputs.Inclusions, "Inclusions")
-	assertIsSorted(t, sortedOutputs.Exclusions, "Exclusions")
-	assert.False(t, cmp.DeepEqual(taskOutputs, sortedOutputs)().Success())
+	taskOutputs.Sort()
+	assertIsSorted(t, taskOutputs.Inclusions, "Inclusions")
+	assertIsSorted(t, taskOutputs.Exclusions, "Exclusions")
+
+	assert.True(t, cmp.DeepEqual(taskOutputs, TaskOutputs{Inclusions: []string{"bar", "foo/**"}, Exclusions: []string{".hidden/**", "special-file"}})().Success())
```
mehulkar, line 219: we need a prettier for Golang

cli/internal/run/global_hash.go
```diff
@@ -30,7 +30,6 @@ type GlobalHashable struct {
 	rootExternalDepsHash string
 	envVars              env.DetailedMap
 	globalCacheKey       string
-	pipeline             fs.PristinePipeline
```
mehulkar, line 33: 👏🏾 👏🏾

cli/internal/runsummary/format_json.go
```diff
@@ -55,12 +56,13 @@ func (rsm *Meta) normalize() {
...
-	Packages          []string           `json:"-"`
-	EnvMode           util.EnvMode       `json:"envMode"`
-	ExecutionSummary  *executionSummary  `json:"execution,omitempty"`
-	Tasks             []*TaskSummary     `json:"tasks"`
+	ID                ksuid.KSUID         `json:"id"`
+	Version           string              `json:"version"`
+	TurboVersion      string              `json:"turboVersion"`
+	GlobalHashSummary *GlobalHashSummary  `json:"globalCacheInputs"`
+	Packages          []string            `json:"-"`
+	Pipeline          fs.PristinePipeline `json:"pipeline"`
```
mehulkar, line 64: We don't need this here

cli/internal/runsummary/format_text.go
```diff
@@ -42,7 +42,7 @@ func (rsm Meta) FormatAndPrintText(workspaceInfos workspace.Catalog) error {
 	fmt.Fprintln(w1, util.Sprintf("  ${GREY}Global Files\t=\t%d${RESET}", fileCount))
 	fmt.Fprintln(w1, util.Sprintf("  ${GREY}External Dependencies Hash\t=\t%s${RESET}", summary.GlobalHashSummary.RootExternalDepsHash))
 	fmt.Fprintln(w1, util.Sprintf("  ${GREY}Global Cache Key\t=\t%s${RESET}", summary.GlobalHashSummary.GlobalCacheKey))
-	if bytes, err := json.Marshal(summary.GlobalHashSummary.Pipeline); err == nil {
+	if bytes, err := json.Marshal(summary.Pipeline); err == nil {
```
mehulkar, line 45: We can just remove this, it was purely there for showing global cache inputs. since it's not a global hash input anymore, we don't need to print it. (source: https://github.com/vercel/turbo/pull/3934)

cli/internal/runsummary/run_summary.go
```diff
@@ -54,14 +55,15 @@ type Meta struct {
...
-	Packages          []string           `json:"packages"`
-	EnvMode           util.EnvMode       `json:"envMode"`
-	ExecutionSummary  *executionSummary  `json:"execution,omitempty"`
-	Tasks             []*TaskSummary     `json:"tasks"`
+	ID                ksuid.KSUID         `json:"id"`
+	Version           string              `json:"version"`
+	TurboVersion      string              `json:"turboVersion"`
+	GlobalHashSummary *GlobalHashSummary  `json:"globalCacheInputs"`
+	Packages          []string            `json:"packages"`
+	Pipeline          fs.PristinePipeline `json:"pipeline"`
```
mehulkar, line 63: We don't need this here if it's just for printing in dry run

cli/integration_tests/edit_turbo_json/global.t
```diff
@@ -0,0 +1,32 @@
+Setup
+  $ . ${TESTDIR}/../_helpers/setup.sh
+  $ . ${TESTDIR}/../_helpers/setup_monorepo.sh $(pwd)
+
+Baseline global hash
+  $ cp "$TESTDIR/fixture-configs/1-baseline.json" "$(pwd)/turbo.json" && git commit -am "no comment" --quiet
+  $ BASELINE=$(${TURBO} build -vv 2>&1 | "$TESTDIR/../_helpers/get-global-hash.sh")
+
+Update pipeline: global hash remains stable.
+  $ cp "$TESTDIR/fixture-configs/2-update-pipeline.json" "$(pwd)/turbo.json" && git commit -am "no comment" --quiet
```
mehulkar, line 10: It's pretty hard to see the diff between these configs. Could you model them as git patches or do it inline with `jq` so it's easieer to see what's changing and then be able to validate that the test is doing the right thing also?

cli/integration_tests/edit_turbo_json/fixture-configs/1-baseline.json
```diff
@@ -0,0 +1,19 @@
...
+      "outputs": []
+    },
+    "my-app#build": {
+      "outputs": ["banana.txt", "apple.json"]
+    },
+
+    "something": {},
+    "//#something": {},
+
+    "maybefails": {}
```
mehulkar, line 17: I don't think we need all these task definitions for the tests. Only `build` and `my-app#build` are maybe needed. Can we delete the extras?

mehulkar (commented):
A couple things that I think we should change, comments in line. 
Other than that, I think we just need to decide _when_ to do this? 1.10? Or a patch is ok?

cli/internal/fs/turbo_json_test.go
nathanhammond: It's called `gofmt`, and we use it, but it decided that this didn't need to change. 🤷‍♂️

cli/internal/runsummary/format_text.go
nathanhammond: I'm okay with dropping it; but I think that having the pipeline as defined in the run summary is super-useful so you don't have to go back to N `turbo.json`s (including workspace configs) to figure out what happened.

That's why I simply promoted it up a level.

cli/internal/fs/turbo_json_test.go
nathanhammond: (rustfmt is a _lot_ nicer)

cli/internal/runsummary/format_json.go
nathanhammond: Collapsing this conversation over into: https://github.com/vercel/turbo/pull/4545#discussion_r1178611355

cli/internal/runsummary/run_summary.go
nathanhammond: Collapsing this conversation over into: https://github.com/vercel/turbo/pull/4545#discussion_r1178611355

cli/internal/runsummary/format_text.go
mehulkar: But these are not resolved configs are they? They are just the root turbo.json. The resolved configs are already in the output for each task and I agree that is super useful

arlyon (approved):

cli/integration_tests/edit_turbo_json/global.t
nathanhammond: I just attempted this in two different ways. Neither of them are a better overall experience.

***

### Either:
Can introduce transitive failures in testing. After my first attempt with diffs things just _didn't work_. Where did this problem get introduced? But since it's not on disk at the point at which we're debugging the experience is miserable.

### Diffs:
The test authoring experience for this becomes "resolve all the files again at each step" and then recalculate the diffs. That's not good.

### Programmatic:
We're in a shell environment; `jq` to set multiple values on a change results in basically inlining a bunch of assignments.

For example:
`jq -r '.pipeline.build.env = ["CC"] | .passThroughEnv="SOMETHING"'` and this keeps getting longer and longer the more changes that need to be made.

***

Open them in your editor and toggle between them if you need to compare.

nathanhammond: Easy peasy:
comparing two fixture files

nathanhammond:
>  I think we just need to decide when to do this? 1.10? Or a patch is ok?

1.10 at earliest. We used to do it _whenever_ but I think aligning to minors is a good answer.

tknickman (approved):

PR #4546, opened 2023-04-12
Chunking Context Refactor pt. 3: Address PR comments from pt. 2

alexkirsz:
### Description

This PR does two things, addressing related comments from https://github.com/vercel/turbo/pull/4450:
1. Replace `ChunkReferenceVc` and `ParallelChunkReferenceVc` with `Chunk::parallel_chunks`.
2. Moves all non-generic optimization logic out of `turbopack-core` and into `turbopack-dev`. These optimization methods are very specific to a development environment, and `turbopack-build` will have different ones. 

### Testing Instructions

Snapshots
link WEB-891


crates/turbopack-dev/src/chunking_context.rs
```diff
@@ -453,54 +456,47 @@ where
...
-/// Computes the list of all chunk children of a given chunk.
-async fn get_chunk_children(parent: ChunkVc) -> Result<impl Iterator<Item = ChunkVc> + Send> {
-    Ok(parent
-        .references()
-        .await?
+    let ecmascript_chunks =
+        optimize_ecmascript_chunks(EcmascriptChunksVc::cell(ecmascript_chunks)).await?;
+    let css_chunks = optimize_css_chunks(CssChunksVc::cell(css_chunks)).await?;
+
+    let chunks = ecmascript_chunks
```
alexkirsz, line 493: This creates an explicit order between JS and CSS chunks in the output asset list, which causes the changes in snapshots.

crates/turbopack-core/src/chunk/optimize.rs
```diff
@@ -3,310 +3,30 @@
...
-#[turbo_tasks::value_trait]
-pub trait ChunkOptimizer {
-    fn optimize(&self, chunks: ChunksVc) -> ChunksVc;
-}
-
-/// Trait to mark a chunk as optimizable.
-#[turbo_tasks::value_trait]
-pub trait OptimizableChunk: Chunk + Asset {
-    fn get_optimizer(&self) -> ChunkOptimizerVc;
-}
```
alexkirsz, line 29: This is no longer relevant, as the chunking context has to know about all the different chunk types it can handle anyway. So instead of the chunks themselves implementing this trait, it's the chunking context which resolves chunks to particular types, and optimizes them accordingly.

crates/turbopack-core/src/chunk/containment_tree.rs
```diff
@@ -0,0 +1,220 @@
...
+                        None
+                    } else {
+                        Some(tree)
+                    }
+                })
+                .collect::<Vec<_>>()
+        }
+
+        fn skip_unnessary_nodes<T>(trees: &mut IndexMap<FileSystemPathVc, Rc<RefCell<Node<T>>>>) {
+            for tree in trees.values_mut() {
```
kdy1, line 139: `values_mut` is a bit confusing.

kdy1: I was checking if `let mut tree = tree.get_mut()` will work at the line below

sokra (approved):

PR #4550, opened 2023-04-12
chore(turbo): consistent internal package names

tknickman:
### Description

This PR is not urgent nor required, it's a vanity change for pretty, more consistent package names for our internal packages. 

I named most of these originally, so I'm trying to right my past wrong 😉 

None of these are published, so it's a simple atomic change.


mehulkar (approved):
Can you update codeowners too? https://github.com/vercel/turbo/blob/main/.github/CODEOWNERS

tknickman:
There's no actual renames file / directory renames here @mehulkar, these are all just package.json "name" changes.

PR #4555, opened 2023-04-12
fix: better support for pnpm aliases

chris-olszewski:
### Description

This is a follow up to #4541 that fixes the underlying issue. We now will attempt to return the extracted version instead of the full key when we encounter one.

A quick explanation for how this happened is that when [pnpm aliases](https://pnpm.io/aliases) are used the resolved version in the lockfile will be the full lockfile key instead of a version since the name won't match the specifier (e.g. `"foo: npm:bar@^1.0.0"` will have a version of `/bar/1.0.0` instead of just `1.0.0`). This is all fine unless an external dependency also depends on `bar@1.0.0` then we get a dependency of `Package{Key: "/bar/1.0.0", Version: "1.0.0"}` when going through the external dependency where we would get `Package{Key: "/bar/1.0.0", Version: "/bar/1.0.0"}` going through the workspace with the alias.

### Testing Instructions

Updated the unit test added in the initial PR to reflect that we're no longer getting a package with a version that's actually a lockfile key. This change also changes the test for pnpm overrides, but this change is more correct as we're now only returning the version. I also tested that the changes don't trigger the repro that was provided in #3638

### DIFF
--- a/cli/internal/lockfile/lockfile_test.go
+++ b/cli/internal/lockfile/lockfile_test.go
@@ -1,0 +1,25 @@
+package lockfile
+
+import (
+	"sort"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func Test_ByKeySortIsStable(t *testing.T) {
+	packagesA := []Package{
+		{"/foo/1.2.3", "1.2.3", true},
+		{"/baz/1.0.9", "/baz/1.0.9", true},
+		{"/bar/1.2.3", "1.2.3", true},
+		{"/foo/1.2.3", "/foo/1.2.3", true},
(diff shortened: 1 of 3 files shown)


mehulkar (commented):
Shuold we also revert the ByKey change if it's not necessary anymore? Or at least test that the tests pass without it?

arlyon (approved):

mehulkar:
> >Or at least test that the tests pass without it?
>
> Which tests?

I meant all of them, but especially the one you added in #4542. but not sure if that unit test would actually exercise this. in either case, ✅

mehulkar (approved):

PR #4565, opened 2023-04-13
Bind to IPv6 and IPv4

sokra:
### Description

The std api doesn't allow to control the `IPV6_ONLY` flag for sockets and it defaults to true.
But we want to bind to both IPv6 and IPv4.


alexkirsz (approved):

alexkirsz:
Missing Cargo.lock update?

PR #4571, opened 2023-04-13
this vendors glob-match with unicode support

arlyon:
### Description

I have a PR open with glob-match to add unicode grapheme support. This PR vendors the dependency until that lands (https://github.com/devongovett/glob-match/pull/6)

### Testing Instructions

`cargo test`


crates/glob-match/src/lib.rs
```diff
@@ -0,0 +1,2113 @@
...
+  !negated
+}
+
+/// gets a slice to a unicode grapheme at the given index
+/// respecting potential escaped characters
+#[cfg(feature = "unic-segment")]
+fn get_char_slice<'a>(glob: &str, glob_bytes: &'a [u8], index: &mut usize) -> Option<&'a [u8]> {
+  use unic_segment::GraphemeCursor;
+  let mut cur = GraphemeCursor::new(*index, glob.len());
+  let end = cur.next_boundary(glob, 0).unwrap().unwrap();
```
chris-olszewski, line 331: Traced through and I believe that these unwraps will never get triggered, but just in case could we change these to a helpful panic message?
```suggestion
  let end = cur.next_boundary(glob, 0).unwrap_or_else(|e| panic!("Invalid boundary {:?}", e)).unwrap();
```

chris-olszewski (approved):
These tests are beautiful

ForsakenHarmony (commented):
why not a git dependency?
https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

PR #4573, opened 2023-04-13
docs: Link to Vercel from the Nav context menu

jaredpalmer:

### DIFF
--- a/docs/components/LogoContext/items.tsx
+++ b/docs/components/LogoContext/items.tsx
@@ -63,6 +63,14 @@ export const PRODUCT_MENU_ITEMS = ({
   site,
 }: ContextList): Array<ContextItem> => [
   {
+    name: "vercel",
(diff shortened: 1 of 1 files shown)


docs/components/LogoContext/items.tsx
```diff
@@ -62,6 +62,14 @@ export const PLATFORM_MENU_ITEMS = ({
 export const PRODUCT_MENU_ITEMS = ({
   site,
 }: ContextList): Array<ContextItem> => [
+  {
+    name: "vercel",
+    "aria-label": "Open Vercel Home in New Tab",
+    children: "Next.js",
```
mehulkar, line 68:
```suggestion
    children: "Vercel",
```

tknickman (approved):
Nice, I originally left this out because you can just click on the same logo to go to Vercel already. but I'm into this, much easier to discover (especially with both).

PR #4575, opened 2023-04-13
port(turborepo): Cache Item

NicholasLYang:
### Description

Ports cache item to Rust. Does not hook up the code to anything yet.

### Reviewer Guide

~~This PR is still being cleaned up, so bear with me. Currently the functionality is all there and the restore tests are ported. I'm working on porting the check_name tests and so on.~~ PR is ready to go with all tests ported. Currently working on a stacked PR that hooks up cache item to HTTP cache

The main code to review is `turborepo-cache/src/cache-archive`, which is the ported cache item code. The code should be a 1-1 match with the Go code, with the same file names and function names.

~~I did *not* port the `cachedDirTree` struct as it was a little confusing and not very idiomatic in Rust. I'll explore a more idiomatic way of memoizing the lookups after this is completed.~~ Cached dir tree is ported

### Testing Instructions

Ports restore_test.go, which includes fairly comprehensive tests on restoration and symlinks.


crates/turborepo-ffi/src/cache.rs
```diff
@@ -0,0 +1,88 @@
+use turbopath::AbsoluteSystemPathBuf;
+use turborepo_api_client::APIClient;
+use turborepo_cache::{http::HttpCache, signature_authentication::ArtifactSignatureAuthenticator};
+
+use crate::{proto, Buffer};
+
+#[no_mangle]
+#[tokio::main]
```
NicholasLYang, line 8: lets us call this async function as sync and have `tokio` spin up a runtime

crates/turborepo-ffi/src/cache.rs
```diff
@@ -0,0 +1,88 @@
...
+            };
+            return resp.into();
+        }
+    };
+
+    let http_cache_config = req.http_cache.expect("http_cache is required field");
+
+    let api_client_config = http_cache_config
+        .api_client
+        .expect("api_client is required field");
```
NicholasLYang, line 24: protobuf makes non-primitive fields optional for some reason

gsoltis: Fun fact: current best practice for protobuf is that all fields are intended to be optional. it's for forward and backward compatibility between different versions of services

cli/internal/run/global_hash.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/vercel/turbo/cli/internal/util"
 )
 
-const _globalCacheKey = "Buffalo buffalo Buffalo buffalo buffalo buffalo Buffalo buffalo"
+const _globalCacheKey = "You don't understand! I coulda had class. I coulda been a contender. I could've been somebody, instead of a bum, which is what I am."
```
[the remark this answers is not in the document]
gsoltis: If this does need to change, we'll need to coordinate this change with a few other things. I think @nathanhammond has a change that needs a global cache key change as well.

gsoltis (changes requested):
Reviewed Go portions

cli/internal/ffi/ffi.go
```diff
@@ -7,12 +7,12 @@ package ffi
 
 // #include "bindings.h"
 //
-// #cgo darwin,arm64 LDFLAGS:  -L${SRCDIR} -lturborepo_ffi_darwin_arm64  -lz -liconv
-// #cgo darwin,amd64 LDFLAGS:  -L${SRCDIR} -lturborepo_ffi_darwin_amd64  -lz -liconv
+// #cgo darwin,arm64 LDFLAGS:  -L${SRCDIR} -lturborepo_ffi_darwin_arm64  -lz -liconv -framework CoreFoundation
```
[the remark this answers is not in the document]
NicholasLYang: In the code? They're basically random C dependencies that show up depending on what Rust packages we use. If we were building a Rust binary they would be invisible to the user

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        // Throw an error if trying to create a cache that contains a type we don't
+        // support.
+        if !matches!(
+            header.entry_type(),
+            EntryType::Regular | EntryType::Directory | EntryType::Symlink
+        ) {
+            return Err(CacheError::UnsupportedFileType(
+                header.entry_type(),
+                Backtrace::capture(),
```
nathanhammond, line 102: This backtrace is basically unnecessary in this instance.

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        // Consistent creation
+        header.set_uid(0);
+        header.set_gid(0);
+        header.as_gnu_mut().unwrap().set_atime(0);
+        header.set_mtime(0);
+        header.as_gnu_mut().unwrap().set_ctime(0);
+
+        if matches!(header.entry_type(), EntryType::Regular) && header.size()? > 0 {
```
nathanhammond, line 113: Weird library; no way that `header.size()?` can actually error here. I guess they're reusing this struct on the read side as well.

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,156 @@
...
+            let file = OpenOptions::new().read(true).open(source_path.as_path())?;
+            self.writer.append(&header, file)?;
+        } else {
+            self.writer.append(&header, &mut std::io::empty())?;
```
nathanhammond, line 117: Interesting that you can't write the header without specifying a body. (Underlying distinction: the `tar` format has to pad blocks, the `WriteHeader` method in Go does that for you, and `Close` writes the last one.)

crates/turborepo-paths/src/anchored_unix_path_buf.rs
```diff
@@ -0,0 +1,40 @@
...
+        self.0
+    }
+
+    pub fn make_canonical_for_tar(&mut self, is_dir: bool) {
+        if is_dir {
+            if !self.0.ends_with(b"/") {
+                self.0.push(b'/');
+            }
+        }
+    }
```
nathanhammond, line 19: After this happens it's not _really_ an `AnchoredUnixPathBuf`. It's an `AnchoredUnixTarPathBuf`. I'd prefer creating a new object here.

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,156 @@
...
+    ) -> Result<Header, CacheError> {
+        let mut header = Header::new_gnu();
+
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::MetadataExt;
+            header.set_mode(file_info.mode());
+        }
+        path.make_canonical_for_tar(file_info.is_dir());
+        header.set_path(path.as_str()?)?;
```
nathanhammond, line 135: This isn't a viable method of setting the header path; it blows up at 255 bytes. Also, it accepts a `Path`, so we should not introduce utf8 lossiness here.

https://docs.rs/tar/latest/tar/struct.Header.html#method.set_path

Can we get a test for longer names added as well to make sure?

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,156 @@
...
+        &mut self,
+        anchor: &AbsoluteSystemPathBuf,
+        file_path: &AnchoredSystemPathBuf,
+    ) -> Result<(), CacheError> {
+        let source_path = anchor.resolve(file_path);
+
+        let file_info = fs::symlink_metadata(source_path.as_path())?;
+        let cache_destination_name = RelativeUnixPathBuf::new(file_path.to_str()?.as_bytes())?;
+
+        let mut header = Self::create_header(cache_destination_name.into(), &file_info)?;
```
nathanhammond, line 88: This is a weird abstraction. Let's just inline it, or extract all of it?

nathanhammond (commented):
Part 1: Create

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
```diff
@@ -0,0 +1,133 @@
...
+    //
+    let cleaned_linkname = clean(linkname);
+
+    // 1. Check to see if the link target is absolute _on the current platform_.
+    // If it is an absolute path it's canonical by rule.
+    if cleaned_linkname.is_absolute() {
+        return Ok(cleaned_linkname);
+    }
+
+    let cleaned_linkname = AnchoredSystemPathBuf::try_from(cleaned_linkname.as_path())?;
```
nathanhammond, line 118: I think this should be unchecked.

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    _ = symlink_from.remove();
+
+    let symlink_to = header.link_name()?.expect("have linkname");
+
+    if symlink_to.is_dir() {
+        symlink_from.symlink_to_file(symlink_to)?;
+    } else {
+        symlink_from.symlink_to_dir(symlink_to)?;
+    }
```
nathanhammond, line 66: Ugh. Windows.

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    #[cfg(unix)]
```
nathanhammond, line 68: I know this to be Darwin-only. Does this play nicely on Linux?

nathanhammond (commented):
Part 2: Restore

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,156 @@
...
+use turbopath::{
+    AbsoluteSystemPathBuf, AnchoredSystemPathBuf, AnchoredUnixPathBuf, RelativeUnixPathBuf,
+};
+
+use crate::CacheError;
+
+enum CacheWriter<'a> {
+    Compressed(tar::Builder<zstd::Encoder<'a, BufWriter<File>>>),
+    Uncompressed(tar::Builder<BufWriter<File>>),
+}
```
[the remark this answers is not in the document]
NicholasLYang: Unfortunately no, since we can't assign to a variable two different generic implementations, i.e. we can't do:

```
let writer = if compressed {
  tar::Builder::new(zstd::Encoder::new(file)); 
} else {
  tar::Builder::new(file)
};
```

In fact, I believe I threw in the towel in a later PR and just made this a `Box<dyn Write>` which is a lot better tbh, at the cost of some small performance impact.

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -0,0 +1,1099 @@
...
+
+            graph.add_edge(source_node, link_node, ());
+
+            header_lookup.insert(processed_sourcename, entry.header().clone());
+        }
+
+        let nodes = petgraph::algo::toposort(&graph, None)
+            .map_err(|_| CacheError::CycleDetected(Backtrace::capture()))?;
+
+        for node in nodes {
```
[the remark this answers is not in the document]
NicholasLYang: I know right! petgraph is surprising me with its ergonomics

crates/turborepo-cache/src/cache_archive/create.rs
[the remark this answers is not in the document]
nathanhammond: I saw that this one already had `Box<dyn Read>` and I'm generally in favor of this. 👍

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -187,6 +187,29 @@ impl AbsoluteSystemPath {
...
+            use std::os::windows::fs::OpenOptionsExt;
+
+            options.custom_flags(FILE_FLAG_SEQUENTIAL_SCAN);
+        }
+
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::OpenOptionsExt;
+
+            options.mode(0o644);
```
nathanhammond, line 208: Does this need to be set on open?

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -0,0 +1,1099 @@
...
+            return Err(CacheError::WindowsUnsafeName(
+                name.to_string(),
+                Backtrace::capture(),
+            ));
+        }
+    }
+
+    // There's no easier way to remove trailing slashes in Rust
+    // because `OsString`s are really just `Vec<u8>`s.
+    let no_trailing_slash: PathBuf = name.components().collect();
```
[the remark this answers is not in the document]
NicholasLYang: There's no easy way to truncate an `OsString` as far as I know

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -0,0 +1,1099 @@
...
+    if name.as_os_str().is_empty() {
+        return PathValidation {
+            well_formed: false,
+            windows_safe: false,
+        };
+    }
+
+    let mut well_formed = true;
+    let mut windows_safe = true;
+    let name = name.to_string_lossy();
```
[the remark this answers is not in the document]
NicholasLYang: We can't turn an `OsString` into a `Vec<u8>` easily because on Windows `OsString` is modeled as `Vec<u16>` whereas on Unix it's `Vec<u8>` (it's modeled as such, but actually it's stored as a custom superset of UTF-8 internally). We could in theory use a bit of unsafe to transmute an `OsString` to a `Vec<u8>` but that's relying on an internal, unstable API, which is probably not the greatest plan.

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -103,8 +105,8 @@ impl AbsoluteSystemPath {
         }
     }
 
-    pub(crate) fn new_unchecked(path: &Path) -> &Self {
-        unsafe { &*(path as *const Path as *const Self) }
+    pub unsafe fn new_unchecked(path: &Path) -> &Self {
```
gsoltis, line 108: Can you use `AbsoluteSystemPathBuf::new().unwrap()` instead? It looks like this is only used in tests

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use tar::{EntryType, Header};
+use turbopath::{AbsoluteSystemPath, AnchoredSystemPath, AnchoredUnixPathBuf, RelativeUnixPathBuf};
+
+use crate::CacheError;
+
+// We use an enum to get around Rust's generic restrictions
+// i.e. you can't have a function that can return two different
+// versions of a generic type like Vec<u32> and Vec<u64>
+enum CacheArchive<'a> {
```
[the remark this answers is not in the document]
NicholasLYang: Not sure honestly. This was in the Go code so I figured this should do it too.

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -336,7 +337,7 @@ mod test {
         .unwrap();
 
         let mut base = CommandBase {
-            repo_root: Default::default(),
+            repo_root: AbsoluteSystemPathBuf::new(tempdir().unwrap().path()).unwrap(),
```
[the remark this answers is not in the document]
NicholasLYang: It does, but I have no clue why, changing it

crates/turborepo-paths/src/anchored_unix_tar_path_buf.rs
```diff
@@ -0,0 +1,15 @@
+use std::path::Path;
+
+use bstr::{BString, ByteSlice};
+
+use crate::PathError;
+
+/// A path that is anchored, unix style, and always ends in '/'
+/// when pointing to a directory
+pub struct AnchoredUnixTarPathBuf(pub(crate) BString);
```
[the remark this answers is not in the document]
NicholasLYang: To do that, we'd need to move `make_canonical_for_tar` too, but that requires editing the path using push. I'm not sure we want to expose that as an API in turbopath.

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,174 @@
...
+        let file = options
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(path.as_path())?;
+
+        // Flush to disk in 1mb chunks.
+        let file_buffer = BufWriter::with_capacity(2usize.pow(20), file);
+
+        let is_compressed = path.as_path().ends_with(".zst");
```
gsoltis, line 51: Instead of `as_path`, let's add support for `.ends_with` to `AbsoluteSystemPath`.

As a general comment, we should aim for not importing `fs` and `path` and friends. It may not always be possible, but that's the ideal.

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,174 @@
...
+    // Wires up the chain of writers:
+    // tar::Builder -> zstd::Encoder (optional) -> BufWriter -> File
+    fn create(path: &AbsoluteSystemPath) -> Result<Self, CacheError> {
+        let mut options = OpenOptions::new();
+
+        let file = options
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(path.as_path())?;
```
gsoltis, line 46: Use `open` on `AbsoluteSystemPath`

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn add_file(
+        &mut self,
+        anchor: &AbsoluteSystemPath,
+        file_path: &AnchoredSystemPath,
+    ) -> Result<(), CacheError> {
+        // Resolve the fully-qualified path to the file to read it.
+        let source_path = anchor.resolve(file_path);
+
+        // Grab the file info to construct the header.
+        let file_info = fs::symlink_metadata(source_path.as_path())?;
```
gsoltis, line 72: Use `symlink_metadata` on `AbsoluteSystemPath`

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        // Normalize the path within the cache
+        let file_path = RelativeUnixPathBuf::new(file_path.to_str()?.as_bytes())?;
```
gsoltis, line 75: This seems like a conversion that should be internal to the `turbopath` library

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,174 @@
...
+        let canonical_file_path = file_path.make_canonical_for_tar(file_info.is_dir());
```
gsoltis, line 76: This one is probably a judgment call. The extra type is not _really_ buying us much here, since we immediately convert it to a `Path`. In this case it might be ok to have `file_path.to_canonical_tar_path()` or something like that return a `Path` directly. We're only doing this because the tar library's API requires a path, it then goes right back and does all of the work to turn it back into unix path bytes.

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn create_header(
+        source_path: &AbsoluteSystemPath,
+        file_info: &fs::Metadata,
+    ) -> Result<Header, CacheError> {
+        let mut header = Header::new_gnu();
+
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::MetadataExt;
+            header.set_mode(file_info.mode());
```
gsoltis, line 103: Is there a default mode that we should set on windows? What happens if we restore a windows-made tar on unix?

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        // Do we need to populate the additional linkname field in Header?
+        if file_info.is_symlink() {
+            let link = fs::read_link(source_path.as_path())?;
```
gsoltis, line 108: move `read_link` to `AbsoluteSystemPath`.

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -0,0 +1,1088 @@
...
+    }
+
+    // There's no easier way to remove trailing slashes in Rust
+    // because `OsString`s are really just `Vec<u8>`s.
+    let no_trailing_slash: PathBuf = name.components().collect();
+
+    // We know this is indeed anchored because of `check_name`,
+    // and it is indeed system because we just split and combined with the
+    // system path separator above
+    Ok(unsafe { AnchoredSystemPathBuf::new_unchecked(no_trailing_slash) })
```
gsoltis, line 205: See below for `check_name`, but I bet we can skip the need for an `unsafe` if we operate on a `RelativeUnixPath` and make the conversion internal to the `turbopath` library.

gsoltis (changes requested):
I got through some of that `cache_archive` code. Will get to more of it tomorrow

crates/turborepo-cache/src/cache_archive/create.rs
NicholasLYang: Does this set the correct permissions?

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -0,0 +1,1088 @@
...
+}
+
+#[derive(Debug, PartialEq)]
+struct PathValidation {
+    well_formed: bool,
+    windows_safe: bool,
+}
+
+fn check_name(name: &Path) -> PathValidation {
```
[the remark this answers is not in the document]
NicholasLYang: Yeah I can do that, although if/when we do the UTF-8 conversion, we can just use normal string methods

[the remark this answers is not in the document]
NicholasLYang: Actually...hmm I'm not sure it's a good idea to wrap this in a `RelativeUnixPath`, because the whole point of this function is to detect whether it's actually a `RelativeUnixPath` or something else.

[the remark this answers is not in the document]
NicholasLYang: Yeah this might be the simplest way to do this until we move to UTF-8, because `canonicalize_name` needs to return an `AnchoredSystemPathBuf`, which is really a `PathBuf` underneath the hood, and the only way to get a `PathBuf` from `&[u8]` is via a `PathBuf`.

crates/turborepo-cache/src/cache_archive/restore.rs
gsoltis: Even if we don't wrap this in `RelativeUnixPathBuf`, this conversion (`components` + `collect`) seems like it should be in the path library, and definitely not leaking an `unsafe` out to a consumer.

crates/turborepo-cache/src/cache_archive/restore_regular.rs
```diff
@@ -0,0 +1,57 @@
...
+    Ok(processed_name)
+}
+
+impl CachedDirTree {
+    pub fn safe_mkdir_file(
+        &mut self,
+        anchor: &AbsoluteSystemPath,
+        processed_name: &AnchoredSystemPath,
+    ) -> Result<(), CacheError> {
+        let is_root_file = processed_name.as_path().parent() == Some(Path::new("."));
```
gsoltis, line 49: Drop `as_path()`.

Have you verified that `parent()` returns `"."`?

The docs say "This means it returns `Some("")` for relative paths with one component"

crates/turborepo-cache/src/cache_archive/restore_directory.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    // Given a path, checks the dir cache to determine where we actually need
+    // to start restoring, i.e. which directories we can skip over because
+    // we've already created them.
+    // Returns the anchor at the depth where we need to start restoring, and
+    // the index into the path components where we need to start restoring.
+    fn get_starting_point(&mut self, path: &AnchoredSystemPath) -> (AbsoluteSystemPathBuf, usize) {
+        let mut i = 0;
+        for (idx, (path_component, prefix_component)) in path
+            .as_path()
```
gsoltis, line 49: Drop `.as_path()`

crates/turborepo-cache/src/cache_archive/restore_directory.rs
```diff
@@ -0,0 +1,174 @@
...
+    ) -> Result<(), CacheError> {
+        // Iterate through path segments by os.Separator, appending them onto
+        // current_path. Check to see if that path segment is a symlink
+        // with a target outside of anchor.
+        let (mut calculated_anchor, start_idx) = self.get_starting_point(processed_name);
+        for component in processed_name.as_path().components().skip(start_idx) {
+            calculated_anchor = check_path(
+                anchor,
+                &calculated_anchor,
+                &AnchoredSystemPath::new(Path::new(component.as_os_str()))?,
```
gsoltis, line 86: This constructor could probably match `AbsoluteSystemPath`'s so that we don't need `Path::new`

crates/turborepo-cache/src/cache_archive/restore_directory.rs
```diff
@@ -0,0 +1,174 @@
...
+                component.as_os_str().to_os_string(),
+            );
+        }
+
+        // If we have made it here we know that it is safe to call fs::create_dir_all
+        // on the join of anchor and processed_name.
+        //
+        // This could _still_ error, but we don't care.
+        let resolved_name = anchor.resolve(processed_name);
+        fs::create_dir_all(&resolved_name)?;
```
gsoltis, line 100: Use `AbsoluteSystemPath` method

crates/turborepo-cache/src/cache_archive/restore_directory.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+
+            let metadata = fs::metadata(&resolved_name)?;
```
gsoltis, line 106: Use `AbsoluteSystemPath` method. And do you mean `symlink_metadata` or `metadata`? Can you leave a comment?

crates/turborepo-cache/src/cache_archive/restore_directory.rs
```diff
@@ -0,0 +1,174 @@
...
+
+fn check_path(
+    original_anchor: &AbsoluteSystemPath,
+    accumulated_anchor: &AbsoluteSystemPath,
+    segment: &AnchoredSystemPath,
+) -> Result<AbsoluteSystemPathBuf, CacheError> {
+    // Check if the segment itself is sneakily an absolute path...
+    // (looking at you, Windows. CON, AUX...)
+    if segment
+        .as_path()
```
gsoltis, line 124: Drop `.as_path()`

crates/turborepo-cache/src/cache_archive/restore_directory.rs
```diff
@@ -0,0 +1,174 @@
...
+        .any(|c| matches!(c, Component::Prefix(_) | Component::RootDir))
+    {
+        return Err(CacheError::LinkOutsideOfDirectory(
+            segment.to_string(),
+            Backtrace::capture(),
+        ));
+    }
+
+    let combined_path = accumulated_anchor.resolve(segment);
+    let Ok(file_info) = fs::symlink_metadata(combined_path.as_path()) else {
```
gsoltis, line 135: Use `AbsoluteSystemPath` method

crates/turborepo-cache/src/cache_archive/restore_directory.rs
```diff
@@ -0,0 +1,174 @@
...
+    if !file_info.is_symlink() {
+        return Ok(combined_path);
+    }
+
+    // Check to see if the symlink targets outside of the originalAnchor.
+    // We don't do eval symlinks because we could find ourself in a totally
+    // different place.
+
+    // 1. Get the target.
+    let link_target = fs::read_link(combined_path.as_path())?;
```
gsoltis, line 151: Use `AbsoluteSystemPath` method

gsoltis (changes requested):
I made it through a bit of restore. As a general comment, a lot of code is opting out of the path library, at which point we lose the benefits of having it.

Can you re-request review once this is using the path library?

chris-olszewski (commented):
Still going through, but I think rebasing will be necessary before approving since I think there's a good amount of conflicting changes between main and this branch when it comes to the paths library.

chris-olszewski (commented):
Took another read through and overall looks good. Two things that feel necessary before stamping:
 - Getting CI to run on this PR, unsure why they didn't run
 - Not exposing the `new_unchecked` outside of the crate

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,182 @@
...
+        header.as_gnu_mut().unwrap().set_atime(0);
+        header.set_mtime(0);
+        header.as_gnu_mut().unwrap().set_ctime(0);
+
+        Ok(header)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
chris-olszewski: Sorry mistyped, I meant `cacheitem/create_test.go`

crates/turborepo-paths/src/anchored_system_path.rs
```diff
@@ -0,0 +1,84 @@
...
+        let path_str = path_ref
+            .to_str()
+            .ok_or_else(|| PathError::InvalidUnicode(path_ref.to_string_lossy().to_string()))?;
+
+        let system_path = Cow::from_slash(path_str);
+        if let Cow::Owned(path) = system_path {
+            Err(PathError::NotSystem(path.to_string_lossy().to_string()))
+        } else {
+            Ok(unsafe { &*(path_ref as *const Path as *const Self) })
+        }
```
chris-olszewski, line 62: Slight nit on this check is a noop on non-Windows and and on Windows can be simplified to a simple string check
```suggestion
        #[cfg(windows)]
        if path_str.contains('/') {
            return Err(PathError::NotSystem(path.to_string_lossy().to_string()))
        }
        Ok(unsafe { &*(path_ref as *const Path as *const Self) })
```

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -0,0 +1,1074 @@
...
+            name.to_string_lossy().to_string(),
+            Backtrace::capture(),
+        ));
+    }
+
+    #[cfg(windows)]
+    {
+        if !windows_safe {
+            return Err(CacheError::WindowsUnsafeName(
+                name.to_string(),
```
chris-olszewski, line 185:
```suggestion
                name.to_string_lossy().to_string(),
```

chris-olszewski (changes requested):
Only blocking issue for me is exposing the unsafe constructor outside of the crate

crates/turborepo-paths/src/anchored_system_path.rs
NicholasLYang: Yeah the UTF-8 PR overhauls this entire setup and removes path-slash as a dependency

crates/turborepo-cache/src/cache_archive/create.rs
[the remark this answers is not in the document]
NicholasLYang: Could we do this in a follow up PR? There's a few things I need to do around getting a SHA that will expand this PR

crates/turborepo-paths/src/anchored_unix_tar_path_buf.rs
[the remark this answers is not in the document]
gsoltis: I'm still not convinced we want this as part of the path library. We construct it, then immediately convert it to a `Path` and throw the wrapper away. Maybe in this case we should construct the Path directly? Or if we're going to have a wrapper, that wrapper should be part of the tar code?

turborepo-tests/integration/tests/command-version.t
```diff
@@ -4,5 +4,7 @@ Setup
 Test version matches that of version.txt
   $ diff <(head -n 1 ${VERSION}) <(${TURBO} --version)
 
+
 TODO: resolve ambiguity
-$ ${TURBO} -v
+  $ ${TURBO} --version
+  1.10.4-canary.3
```
gsoltis, line 10: This looks like it shouldn't be here

turborepo-tests/integration/tests/single_package/dry-run.t
```diff
@@ -19,7 +19,7 @@ Check
   Tasks to Run
   build
     Task                           = build                                                                                                                                           
-    Hash                           = d2295def33764d46                                                                                                                                
+    Hash                           = d2295def33764d46
```
gsoltis, line 22: I think this is a whitespace diff? Any idea what's going on here?

crates/turborepo-cache/src/cache_archive/create.rs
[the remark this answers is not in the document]
gsoltis: I think it would be good to have those available. You could do it in a separate stacked PR if you don't want to pollute this one, but that's the real assurance that this code is correct.

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -165,8 +178,8 @@ impl AbsoluteSystemPath {
         Ok(())
     }
 
-    pub fn resolve(&self, path: &AnchoredSystemPathBuf) -> AbsoluteSystemPathBuf {
-        let path = self.0.join(path.as_path());
+    pub fn resolve(&self, path: impl AsRef<AnchoredSystemPath>) -> AbsoluteSystemPathBuf {
```
gsoltis, line 181: Nit: I think this can just be `&AnchoredSystemPath`

gsoltis (changes requested):

turborepo-tests/integration/tests/single_package/dry-run.t
NicholasLYang: No clue, but it was causing some test failures

turborepo-tests/integration/tests/command-version.t
NicholasLYang: Any idea what the original test was aiming to do?

crates/turborepo-cache/src/cache_archive/create.rs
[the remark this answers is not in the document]
NicholasLYang: I can open a stacked PR, but it'd be nice to merge this one before opening another PR

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
```diff
@@ -0,0 +1,150 @@
...
+    // - Relative Unix Path
+    // - Relative Windows Path
+    //
+    // At this point we simply assume that it's a relative path—no matter
+    // which separators appear in it and where they appear,  We can't do
+    // anything else because the OS will also treat it like that when it is
+    // a link target.
+    //
+    let source = anchor.resolve(processed_name);
+    let canonicalized = source.parent().unwrap_or(anchor).resolve(&cleaned_linkname);
```
gsoltis, line 142: What's the scenario where `parent()` doesn't exist? Is `anchor` actually an ok replacement?

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
```diff
@@ -0,0 +1,150 @@
...
+    // should be treated as targets for that platform. The author may be
+    // generating an artifact that should work on Windows on a Unix device.
+    //
+    // Given all of that, our best option is to restore link targets _verbatim_.
+    // No modification, no slash conversion.
+    //
+    // In order to DAG sort them, however, we do need to canonicalize them.
+    // We canonicalize them as if we're restoring them verbatim.
+    //
+    let cleaned_linkname = Utf8PathBuf::try_from(clean(linkname)).map_err(|_| {
```
gsoltis, line 117: Are symlinks required to be utf8? Or are we only supporting symlinks that are utf8?

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
```diff
@@ -0,0 +1,150 @@
...
+        CacheError::PathError(
+            PathError::InvalidUnicode(linkname.to_string_lossy().to_string()),
+            Backtrace::capture(),
+        )
+    })?;
+
+    // 1. Check to see if the link target is absolute _on the current platform_.
+    // If it is an absolute path it's canonical by rule.
+    if cleaned_linkname.is_absolute() {
```
gsoltis, line 126: How about `turbopath::categorize(Utf8Path) -> Either<AnchoredSystemPathBuf, AbsoluteSystemPathBuf>`?

Then we aren't operating directly on `Utf8Path` instances (which we should likely treat like `Path` and try to avoid outside the path library), and we don't need to expose `AnchoredSystemPathBuf::new_unchecked`?

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
```diff
@@ -0,0 +1,150 @@
...
+    Ok(processed_name)
+}
+
+// canonicalizeLinkname determines (lexically) what the resolved path on the
+// system will be when linkname is restored verbatim.
+pub fn canonicalize_linkname(
+    anchor: &AbsoluteSystemPath,
+    processed_name: &AnchoredSystemPathBuf,
+    linkname: &Path,
+) -> Result<Utf8PathBuf, CacheError> {
```
gsoltis, line 94: I think we need to be careful with this function. We're introducing a failure mode that doesn't exist on the Go side. `canonicalizeLinkname` _always_ returns a `string`. Do we actually need the restriction that symlink targets (which may be garbage!) are valid utf8 strings? It looks like if we error here, we fail the whole restoration, rather than just deferring symlink creation?

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
```diff
@@ -0,0 +1,150 @@
...
+    dir_cache: &mut CachedDirTree,
+    anchor: &AbsoluteSystemPath,
+    header: &tar::Header,
+) -> Result<AnchoredSystemPathBuf, CacheError> {
+    let processed_name = canonicalize_name(&header.path()?)?;
+
+    let processed_linkname = canonicalize_linkname(
+        anchor,
+        &processed_name,
+        &header.link_name()?.expect("has linkname"),
```
gsoltis, line 22: Might be good to pull this out once so we don't have `.expect` more than we need to.

gsoltis: In fact, rather than crash, we probably ought to report a malformed tar. I suspect it's possible to have a the bit set for a symlink without having a link name present.

crates/turborepo-paths/src/anchored_system_path_buf.rs
```diff
@@ -115,6 +137,29 @@ impl AnchoredSystemPathBuf {
...
 
+    pub fn new_unchecked(path: impl Into<Utf8PathBuf>) -> Self {
+        AnchoredSystemPathBuf(path.into())
+    }
+
+    // Takes in a path that has already been validated as anchored
+    // via `check_name` in `turborepo-cache` and constructs an
+    // `AnchoredSystemPathBuf` with no trailing slashes.
+    pub fn from_validated_tar_path(path: &str) -> Self {
+        let path = Utf8Path::new(path);
```
gsoltis, line 148: How about a constructor that does the right thing, instead of `check_name`? This is essentially `new_unchecked`. 

`AnchoredSystemPathBuf::from_cleaned(raw) -> Result<(Self, bool), PathError>` to handle the extra windows-safe bit

gsoltis (changes requested):
I think we're getting there, but there's still a bit of path manipulation that I think ought to be in the path library.

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
NicholasLYang: Yeah we are. Since all paths are required to be utf-8, symlinks aren’t any different

crates/turborepo-cache/src/cache_archive/restore_symlink.rs
NicholasLYang: I think we should keep the requirement that all paths are UTF-8. If I remember correctly, we don't restore symlinks outside of the repo directory, so any symlink target is inside the repo. Having any path in the repo be valid UTF-8 is a reasonable rule IMO.

gsoltis:
We should restore links to outside of the repo, we just shouldn't traverse
them (or any link)

crates/turborepo-paths/src/anchored_system_path_buf.rs
NicholasLYang: I wanna keep as little tar-specific code in turbopath as possible. Otherwise we have an awkward split of some cache code in turbopath and some in turborepo-cache

NicholasLYang:
Sure, but regardless symlinks should be valid Unicode. Unless there's a good example of when we'd want to handle non-Unicode symlinks (while also having Unicode paths in every other place), I wouldn't consider this a blocking issue for this PR. Since this PR is not connected to anything, if we discover a usecase of non-Unicode symlinks, we can follow up in another PR fixing this.

gsoltis (changes requested):

crates/turborepo-paths/src/lib.rs
```diff
@@ -86,6 +91,14 @@ pub trait IntoUnix {
     fn into_unix(self) -> Utf8PathBuf;
 }
 
+fn validate_system(path: impl AsRef<str>) -> Result<(), PathError> {
+    if is_not_system(&path) {
```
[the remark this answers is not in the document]
NicholasLYang: Sure, but this is out of scope of this PR. This was just a light refactor to existing logic.

crates/turborepo-paths/src/relative_unix_path_buf.rs
```diff
@@ -1,35 +1,50 @@
...
 impl RelativeUnixPathBuf {
     pub fn new(path: impl Into<String>) -> Result<Self, PathError> {
         let path_string = path.into();
-        if path_string.starts_with('/') {
+        if path_string.starts_with('/') || Utf8Path::new(&path_string).is_absolute() {
             return Err(PathError::NotRelative(path_string));
         }
-        Ok(Self(path_string))
+
+        let unix_path = path_string.into_unix();
```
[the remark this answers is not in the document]
NicholasLYang: Sure. I added this because I realized that the new function was doing zero conversion or validation.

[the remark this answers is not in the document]
NicholasLYang: Should we error here instead?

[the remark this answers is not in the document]
gsoltis: what's the validation that you'd like to do? But in general, yes, if it's not a relative unix path, we should error. If the caller wants to do a conversion, we have explicit calls for that.

gsoltis (approved):
Awesome. Let's get this merged.

gsoltis (approved):
Awesome. Let's get this merged.

gsoltis (approved):
Awesome. Let's get this merged.

gsoltis (approved):
Awesome. Let's get this merged.

PR #4582, opened 2023-04-14
ignore internal and server-relative url() in CSS

sokra:
### Description

* ignore internal urls like `url(#noiseFilter)`
* ignore server-relative urls like `url(/foo)`

### DIFF
--- a/crates/turbopack-css/src/references/mod.rs
+++ b/crates/turbopack-css/src/references/mod.rs
@@ -154,20 +154,24 @@ impl<'a> VisitAstPath for AssetReferencesVisitor<'a> {
 
         let src = url_string(u);
 
-        let issue_span = u.span;
-        self.references.push(
-            UrlAssetReferenceVc::new(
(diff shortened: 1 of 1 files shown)


crates/turbopack-css/src/references/mod.rs
```diff
@@ -154,20 +154,24 @@ impl<'a> VisitAstPath for AssetReferencesVisitor<'a> {
...
-                IssueSourceVc::from_byte_offset(
-                    self.source,
-                    issue_span.lo.to_usize(),
-                    issue_span.hi.to_usize(),
-                ),
-            )
-            .into(),
-        );
+        // ignore internal urls like `url(#noiseFilter)`
+        // ignore server-relative urls like `url(/foo)`
```
alexkirsz, line 158: Server-relative URLs don't have any special behavior?

What about absolute URLs?

alexkirsz (approved):

PR #4583, opened 2023-04-14
remove panic since this might happen due to eventual consistency

sokra:
### Description

remove panic since this might happen due to eventual consistency

### DIFF
--- a/crates/turbopack-ecmascript/src/references/esm/module_item.rs
+++ b/crates/turbopack-ecmascript/src/references/esm/module_item.rs
@@ -3,12 +3,9 @@ use std::mem::replace;
 use anyhow::Result;
 use swc_core::{
     common::DUMMY_SP,
-    ecma::{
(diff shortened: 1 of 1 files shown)


crates/turbopack-ecmascript/src/references/esm/module_item.rs
```diff
@@ -47,10 +47,6 @@ impl CodeGenerateable for EsmModuleItem {
         let mut visitors = Vec::new();
 
         let path = &self.path.await?;
-        assert!(
-            matches!(path.last(), Some(AstParentKind::ModuleDecl(_))),
-            "EsmModuleItem was created with a path that points to a unexpected ast node"
-        );
```
alexkirsz, line 53: We should have a comment that explains the eventual consistency issue here.

alexkirsz (approved):

PR #4587, opened 2023-04-14
Vc<T> and Turbo Engine type system improvements

alexkirsz:
This PR changes our Turbo Engine code generation from generating additional `TypeVc` types to allowing the `Vc<Type>` notation. It also brings other improvements to the Turbo Engine type system, like more type-safe downcasting and upcasting, better support for primitives, the possibility to accept `&self` in `#[value_impl]` implementations everywhere, and a bunch of other changes.

link WEB-379


crates/turbo-tasks/src/vc/mod.rs
```diff
@@ -0,0 +1,524 @@
...
+/// let some_ref: ReadRef<T> = some_vc.await?;
+/// some_ref.some_method_on_t();
+/// ```
+pub struct Vc<T>
+where
+    T: ?Sized,
+{
+    // TODO(alexkirsz) Should be private (or undocumented), but turbo-tasks-memory needs it to be
+    // accessible.
+    pub node: RawVc,
```
sokra, line 43: at least add `#[doc(hidden)]`

sokra (approved):

PR #3338, opened 2023-01-17
feat(turbopack): Add tree shaking

kdy1:
x-ref: https://vercel.slack.com/archives/C03EWR7LGEN/p1673561577048649

This Is still WIP.

 - [ ] Analyzer

The basic logic is implemented, but it requires more checks for live/initial.

 - [ ] Splitter
 - [ ] Merger

PR #4596, opened 2023-04-17
refactor: Move transforms and compile time info into `EcmascriptOption`

kdy1:
### Description

This PR reflects feedback from #3338. (https://github.com/vercel/turbo/pull/3338#discussion_r1138234264)

 - Closes WEB-726
 - Next.js counterpart: https://github.com/vercel/next.js/pull/48512

### DIFF
--- a/crates/turbopack-ecmascript/src/lib.rs
+++ b/crates/turbopack-ecmascript/src/lib.rs
@@ -130,9 +130,7 @@ pub struct EcmascriptModuleAsset {
     pub source: AssetVc,
     pub context: AssetContextVc,
     pub ty: EcmascriptModuleAssetType,
-    pub transforms: EcmascriptInputTransformsVc,
     pub options: EcmascriptOptions,
-    pub compile_time_info: CompileTimeInfoVc,
     pub inner_assets: Option<InnerAssetsVc>,
     #[turbo_tasks(debug_ignore)]
     #[serde(skip)]
@@ -150,17 +148,13 @@ impl EcmascriptModuleAssetVc {
         source: AssetVc,
         context: AssetContextVc,
         ty: Value<EcmascriptModuleAssetType>,
-        transforms: EcmascriptInputTransformsVc,
         options: Value<EcmascriptOptions>,
-        compile_time_info: CompileTimeInfoVc,
     ) -> Self {
         Self::cell(EcmascriptModuleAsset {
             source,
             context,
             ty: ty.into_value(),
-            transforms,
             options: options.into_value(),
-            compile_time_info,
             inner_assets: None,
             last_successful_analysis: Default::default(),
         })
@@ -171,18 +165,14 @@ impl EcmascriptModuleAssetVc {
         source: AssetVc,
         context: AssetContextVc,
         ty: Value<EcmascriptModuleAssetType>,
-        transforms: EcmascriptInputTransformsVc,
         options: Value<EcmascriptOptions>,
-        compile_time_info: CompileTimeInfoVc,
         inner_assets: InnerAssetsVc,
     ) -> Self {
         Self::cell(EcmascriptModuleAsset {
             source,
             context,
             ty: ty.into_value(),
-            transforms,
             options: options.into_value(),
-            compile_time_info,
             inner_assets: Some(inner_assets),
             last_successful_analysis: Default::default(),
         })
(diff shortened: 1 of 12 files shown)


crates/turbopack/src/module_options/mod.rs
```diff
@@ -231,6 +226,13 @@ impl ModuleOptionsVc {
             .collect(),
         );
 
+        let ecmascript_options = EcmascriptOptions {
+            split_into_parts: enable_tree_shaking,
+            import_parts: enable_tree_shaking,
+            transforms: app_transforms,
+            compile_time_info,
```
kdy1, line 233: @sokra Do you have any idea how can I get `compile_time_info` from here?

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -84,13 +84,17 @@ use crate::{
 };
 
 #[turbo_tasks::value(serialization = "auto_for_input")]
-#[derive(PartialOrd, Ord, Hash, Debug, Default, Copy, Clone)]
+#[derive(PartialOrd, Ord, Hash, Debug, Copy, Clone)]
 pub struct EcmascriptOptions {
```
sokra, line 88: We probably want to use a `EcmascriptOptionsBuilder` here to allow extending this in future (since it has no longer Default)

crates/turbopack/src/module_options/mod.rs
sokra: ~Move `compile_time_info` from AssetContext/ModuleAssetContext to `ModuleOptionsContext`.~

That's not that easy.

* Add `compile_time_info` to `ModuleOptionsContext`
* Add a TODO that `compile_time_info` should be removed from `AssetContext`

crates/turbopack/src/lib.rs
```diff
@@ -122,9 +122,7 @@ async fn apply_module_type(
                 source,
                 context.into(),
                 Value::new(EcmascriptModuleAssetType::Ecmascript),
-                *transforms,
                 Value::new(*options),
-                context.compile_time_info(),
```
sokra, line 127: ModuleType need to change to something like:

```
pub enum ModuleType {
    Ecmascript {
        options: EcmascriptOptions,
    },
    Typescript {
        options: EcmascriptOptions,
    },
    TypescriptWithTypes {
        options: EcmascriptOptions,
    },
    TypescriptDeclaration {
        options: EcmascriptOptions,
    },
```

You also need to derive TraceRawVcs for `EcmascriptOptions`

sokra (changes requested):
This also requires changes in next.js, we need a corresponding PR there

crates/turbopack/src/module_options/module_options_context.rs
```diff
@@ -239,6 +242,9 @@ pub struct ModuleOptionsContext {
     pub placeholder_for_future_extensions: (),
     #[serde(default)]
     pub enable_tree_shaking: bool,
+
+    #[serde(default)]
+    pub compile_time_info: Option<CompileTimeInfoVc>,
```
kdy1, line 247: @sokra Is `Option<CompileTimeInfoVc>` fine here?

kdy1: As this type derives `Default`, I used `Option` but I'm not sure

sokra: I guess yes

kdy1: If so, is it fine to use `Option<CompileTimeInfoVc>` for `EcmascriptOptions`?

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -84,13 +84,17 @@ use crate::{
...
     /// module is split into smaller module parts which can be selectively
     /// imported
     pub split_into_parts: bool,
     /// imports will import parts of modules
     pub import_parts: bool,
+
+    pub transforms: EcmascriptInputTransformsVc,
+
+    pub compile_time_info: CompileTimeInfoVc,
```
kdy1, line 97: I'm not sure if this can be changed to `Option`, but there's one problem if I don't

crates/turbopack/src/module_options/mod.rs
kdy1: At here, I need a way to get `CompileTimeInfoVc`, not `Option<CompileTimeInfoVc>`

kdy1:
Closing as stale, and I'll retry this in future

PR #4598, opened 2023-04-17
More fluent GraphTraversal API

alexkirsz:
### Description

This PR refactors the API of `GraphTraversal` to be more fluent, and less magical. It should make it easier to use and understand. Had this in store from trying to debug a lifetime issue in another branch.

### Testing Instructions

Surface-level change.

Next.js PR: https://github.com/vercel/turbo/pulls/alexkirsz

### DIFF
--- a/crates/turbo-tasks/src/graph/graph_store.rs
+++ b/crates/turbo-tasks/src/graph/graph_store.rs
@@ -2,7 +2,8 @@ use std::collections::HashSet;
 
 /// A graph store is a data structure that will be built up during a graph
 /// traversal. It is used to store the results of the traversal.
-pub trait GraphStore<Node>: Default {
+pub trait GraphStore {
+    type Node;
     type Handle: Clone;
 
     // TODO(alexkirsz) An `entry(from_handle) -> Entry` API would be more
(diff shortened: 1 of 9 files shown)


crates/turbopack-core/src/changed.rs
```diff
@@ -20,16 +20,15 @@ async fn get_referenced_assets(parent: AssetVc) -> Result<impl Iterator<Item = A
...
-    )
-    .await
-    .completed()?
-    .into_inner()
-    .into_iter()
-    .map(content_changed)
-    .collect();
+    let completions = NonDeterministic::new()
+        .skip_duplicates()
+        .visit([root], get_referenced_assets)
```
alexkirsz, line 25: This is the difference in interface: no more type wrangling to create the traversal.

crates/turbo-tasks/src/graph/graph_store.rs
```diff
@@ -66,9 +90,9 @@ where
     }
 }
```
sokra, line 91: `SkipDuplicates` could implement `skip_duplicates` and return self

sokra (approved):

crates/turbopack-node/src/lib.rs
```diff
@@ -170,13 +170,12 @@ async fn separate_assets(
...
     };
 
-    let graph = GraphTraversal::<SkipDuplicates<ReverseTopological<Type>, _>>::visit(
-        once(Type::Internal(intermediate_asset)),
-        get_asset_children,
-    )
-    .await
-    .completed()?
-    .into_inner();
+    let graph = ReverseTopological::<Type>::new()
```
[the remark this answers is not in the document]
alexkirsz: I'll try!

crates/turbo-tasks/src/graph/graph_store.rs
alexkirsz: For now, with the way the traits are set up, this would require specialization, so I'm going to skip this for now.

PR #4609, opened 2023-04-17
fix(ecmascript): eval assignop to the ident

kwonoj:
### Description

- closes WEB-889

When there is a variable with `+=` assignop only to the given ident

```
var css = '';
for (var i = 0; i < length; i++) {
        var partRule = this.rules[i];

        if (typeof partRule === 'string') {
          css += partRule;
        } else if (partRule) {
         css += partString;
        }
      }

      if (css) {
       ....
      }
```

evaluated result becomes

```
var css = '';
for(var i = 0; i < length; i++){
    var partRule = this.rules[i];
    if (typeof partRule === 'string') {
        css += partRule;
    } else if (partRule) {
        css += partString;
    }
}
if ("TURBOPACK compile-time falsy", 0) {
    "TURBOPACK unreachable";
}
```

so `if (css)` condition never executed even though it can be executed depends on the runtime assignop condition. 

PR attempts to evaluate if assignop is for the ident to prevent this.

### DIFF
--- a/crates/turbopack-ecmascript/src/analyzer/graph.rs
+++ b/crates/turbopack-ecmascript/src/analyzer/graph.rs
@@ -941,7 +941,6 @@ impl VisitAstPath for Analyzer<'_> {
         n: &'ast AssignExpr,
         ast_path: &mut AstNodePath<AstParentNodeRef<'r>>,
     ) {
-        println!("visit_assign_expr: {:#?}", n);
         ast_path.with(
             AstParentNodeRef::AssignExpr(n, AssignExprField::Left),
             |ast_path| match &n.left {


sokra (changes requested):
Could you add a test case to tests/analyzer/graph?

The fix doesn't look right to me...

kwonoj:
@sokra Not sure how to translate test outputs to confirm if this is incorrect approach: https://github.com/vercel/turbo/pull/4609/commits/6dba17d19e2d524ff2323a144942f21ce2093d39

crates/turbopack-ecmascript/tests/analyzer/graph/assign/input.js
```diff
@@ -0,0 +1,40 @@
...
+          this.componentId,
+          name,
+          stylis(css, `.${name}`, undefined, this.componentId)
+        );
+      }
+
+      names = joinStrings(names, name);
+    }
+  }
+}
```
sokra, line 40:
```
let a = "";
a += "x";

let b = 0;
b -= 1;

let c = 0;
c++;

let d = 0;
d ||= 1;

let e = 1;
e &&= 2;
```

should resolve as:

```
a = "" | a + "x"
b = 0 | b - 1
c = 0 | c + 1
d = 0 | 1
e = 1 | 2
```

kdy1 (approved):

crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -952,6 +952,11 @@ impl VisitAstPath for Analyzer<'_> {
                             self.visit_expr(expr, ast_path);
                         },
                     );
+
+                    if let Some(key) = expr.as_ident() {
+                        let value = self.eval_context.eval(&n.right);
+                        self.add_value(key.to_id(), value);
```
kdy1, line 958: This should not be `add_value`, I think, because for

```ts
a = 1
a += 1 
```

`a` is 2, not 1

sokra (approved):

PR #4611, opened 2023-04-17
Improve task cache performance

sokra:
### Description

Performance improvements:

* use thread local task cache to avoid accessing dash map too often
* use prehash to avoid rehashing the key in the task cache


anthonyshew:

PR #4615, opened 2023-04-17
chore(turbo): stricter node engines

tknickman:
### Description

Support stricter node versions and package manager version to avoid things like https://github.com/vercel/turbo/pull/4461

This also adds a .node-version file to support automatic node version switching for those using https://github.com/Schniz/fnm (cc @Schniz!)


mehulkar (approved):
Seems good, but we should update to node 18 pretty soon

tknickman:
Yep, one step at a time 😄

PR #4617, opened 2023-04-18
Move Turborepo integration tests to their own workspace 

mehulkar:
Moving into their own workspace so we can run them independently via Turbo (not Makefile), and potentially finding ways to test against examples, etc. Instead of:

Before:
```sh
cd cli && make integration-tests
````

After:

```sh
cd turborepo-integration-tests && pnpm test
```

Alternatively:

```bash
# global turbo
turbo run test --filter=turborepo-integration-tests
# via turbo
pnpm test -- --filter=turborepo-integration-tests
```


package.json
```diff
@@ -16,7 +16,8 @@
     "turbo-prebuilt": "node turbow.js",
     "docs": "pnpm -- turbo run dev --filter=docs --no-cache",
     "run-example": "./scripts/run-example.sh",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "test": "pnpm -- turbo run test"
```
mehulkar, line 20: Following the pattern, but if we had global turbo in Github CI, we wouldn't need to use `pnpm -- turbo`

turborepo-integration-tests/turbo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": ["//"],
+  "pipeline": {
+    "test": {
+      "dependsOn": ["cli#build"],
```
mehulkar, line 5: This is here for completeness, but in the general case, we're building a local turbo before we run the `test` task, so turbo is already built. I can remove this if people have strong opinions.

arlyon (approved):

mehulkar:
Merging this as all relevant checkes have passed. The remaining checks are Turbopack Rust testing and Turbopack benchmarking, which are very slow and also unaffected by this PR. They are triggered because `test.yml` changed

PR #4630, opened 2023-04-18
Move examples tests to their own workspace

mehulkar:
This PR:
- Makes `examples_tests` an official workspace in `turborepo-tests/examples`
- Extracts some helpers from `turborepo-tests/integration` into their own workspace
  - This will cause examples tests to run when those helpers change, avoiding issues like #4626.
- Make `examples/` a workspace, and sets up a package.json dependency, so that when examples change, examples_tests are also invalidated. (Note that this doesn't affect when CI runs since that's configured in Github Actions workflow configs)


turborepo-tests/examples/turbo.json
```diff
@@ -0,0 +1,26 @@
...
+        "./run-example.sh",
+
+        // TODO: model ../../examples as a workspace so we can wire up
+        // a workspace "dependency" and use that to trigger changes instead
+        // of relative paths.
+        "../../examples/**/*.ts",
+        "../../examples/**/*.tsx",
+        "../../examples/**/*.json",
+        "../../examples/**/*.js",
+        "../../examples/**/*.yaml",
```
mehulkar, line 17: Unless I got these paths wrong (which I don't _think_ I did??) this doesn't actually work. I'm not seeing these in `expandedInputs` in the Run Summary.

turborepo-tests/helpers/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "turborepo-tests-helpers",
+  "scripts": {
+    "build": "echo passthrough"
```
mehulkar, line 4: Is this required for the topological dependency to work?

turborepo-tests/helpers/package.json
```diff
@@ -0,0 +1,6 @@
...
+  }
+}
```
nathanhammond, line 6:
```suggestion
{
  "name": "turborepo-tests-helpers"
}
```

turborepo-tests/integration/package.json
```diff
@@ -8,5 +8,8 @@
     "test:interactive": ".cram_env/bin/prysk -i --shell=`which bash` tests/**/*.t",
     "test:parallel": ".cram_env/bin/pytest -n auto tests/**/*.t --prysk-shell=`which bash`",
     "pretest:parallel": ".cram_env/bin/pip3 install --quiet pytest \"prysk[pytest-plugin]\" pytest-xdist"
+  },
+  "devDependencies": {
```
nathanhammond, line 12:
```suggestion
  "dependencies": {
```

turborepo-tests/examples/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "turborepo-examples-tests",
+  "scripts": {
+    "test": "./run-example.sh"
+  },
+  "devDependencies": {
```
nathanhammond, line 6: Doesn't really matter; but since this thing _doesn't work_ without the dependency we should mark it as such.

It's a hard dependency for `turborepo-examples-tests`.

```suggestion
  "dependencies": {
```

turborepo-tests/examples/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "turborepo-examples-tests",
```
nathanhammond, line 2: Match file system layout?

```suggestion
  "name": "turborepo-tests-examples",
```

examples/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "turborepo-examples",
+  "scripts": {
+    "build": "echo passing"
+  }
+}
```
nathanhammond, line 6:
```suggestion
{
  "name": "turborepo-examples"
}
```

turborepo-tests/integration/turbo.json
```diff
@@ -2,7 +2,7 @@
   "extends": ["//"],
   "pipeline": {
     "test": {
-      "dependsOn": ["cli#build"],
+      "dependsOn": ["cli#build", "^build"],
```
nathanhammond, line 5: This should be handled by adding a synthetic task:

```
  "pipeline": {
    "topo": {
      "dependsOn": ["^topo"]
    },
    "test": {
      "dependsOn": ["cli#build", "^topo"],
      "env": ["GO_TAG"]
    }
```

This pattern also enables scoping down the `inputs` for any part of the synthetic graph.

turborepo-tests/integration/tests/link.t
```diff
@@ -1,5 +1,5 @@
 Setup
-  $ . ${TESTDIR}/_helpers/setup.sh
+  $ . ${TESTDIR}/../../helpers/setup.sh
```
nathanhammond, line 2: I'd really like for this to be:
`node -r path -e 'console.log(path.dirname(require.resolve("turborepo-tests-helpers/package")));'`

...but we'd need to make that "run-once" instead of "run a billion times".

turborepo-tests/examples/turbo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": ["//"],
+  "pipeline": {
+    "test": {
+      "dependsOn": ["^build", "cli#build"],
+      "outputs": []
+    }
+  }
```
nathanhammond, line 8:
```suggestion
  "pipeline": {
    "topo": {
      "dependsOn": ["^topo"]
    },
    "test": {
      "dependsOn": ["cli#build", "^topo"]
    }
  }
```

turborepo-tests/examples/README.md
```diff
@@ -0,0 +1,8 @@
+# `turborepo-tests/examples`
+
+The tests in this directory exercise the `examples/` directory. They validate
+that when someone uses those examples (likely via `npx create-turbo -e <example>`),
+the example works.
+
+These tests do _not_ use a local turbo build, they use the version installed in each
```
nathanhammond, line 7:
```suggestion
These tests do _not_ use a local `turbo` build, they use the version installed in each
```

pnpm-workspace.yaml
```diff
@@ -4,10 +4,11 @@ packages:
...
   - "benchmark"
-  - "turborepo-tests/integration"
+  - "turborepo-tests/*"
   # Intentionally exclude the `npm` `turbo` package from the workspaces.
   - "!packages/turbo"
   - "packages/turbo-tracing-next-plugin/test/with-mongodb-mongoose"
   - "crates/next-dev-tests/test-harness"
   - "crates/next-dev-tests/tests"
   - "crates/*/js"
+  - examples/
```
nathanhammond, line 14:
```suggestion
  - examples
```

nathanhammond (changes requested):
First pass.

turborepo-tests/integration/tests/link.t
mehulkar: yeah, we could at least set `MONOREPO_ROOT_DIR`, instead of relying on prysk's TESTDIR directly, here, but honestly, don't want to do another round on this PR right now heh. It would just be an aesthetic change

nathanhammond: Yeah. This isn't blocking.

turborepo-tests/integration/turbo.json
mehulkar: done

nathanhammond (approved):

PR #4632, opened 2023-04-18
feat(turbo): add spaces link

tknickman:
### Description

Extend link/unlink to support target.

### DIFF
--- a/crates/turborepo-api-client/src/lib.rs
+++ b/crates/turborepo-api-client/src/lib.rs
@@ -77,8 +77,19 @@ impl Team {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Space {
+    pub id: String,
+    pub name: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct TeamsResponse {
     pub teams: Vec<Team>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SpacesResponse {
+    pub spaces: Vec<Space>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -206,6 +217,37 @@ impl APIClient {
         response.json().await.map_err(|err| {
             anyhow!(
                 "Error getting caching status: {}",
+                err.status()
+                    .and_then(|status| status.canonical_reason())
+                    .unwrap_or(&err.to_string())
+            )
+        })
+    }
+
+    pub async fn get_spaces(&self, token: &str, team_id: Option<&str>) -> Result<SpacesResponse> {
+        // create url with teamId if provided
+        let endpoint = match team_id {
+            Some(team_id) => format!("/v0/spaces?limit=100&teamId={}", team_id),
+            None => "/v0/spaces?limit=100".to_string(),
+        };
+
+        let response = self
+            .make_retryable_request(|| {
+                let request_builder = self
+                    .client
+                    .get(self.make_url(endpoint.as_str()))
+                    .header("User-Agent", self.user_agent.clone())
+                    .header("Content-Type", "application/json")
+                    .header("Authorization", format!("Bearer {}", token));
+
+                request_builder.send()
+            })
+            .await?
+            .error_for_status()?;
+
+        response.json().await.map_err(|err| {
+            anyhow!(
+                "Error getting spaces: {}",
                 err.status()
                     .and_then(|status| status.canonical_reason())
                     .unwrap_or(&err.to_string())
--- a/crates/turborepo-lib/src/cli.rs
+++ b/crates/turborepo-lib/src/cli.rs
@@ -177,6 +177,12 @@ pub enum DaemonCommand {
     Stop,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Serialize, ValueEnum)]
+pub enum LinkTarget {
+    RemoteCache,
+    Spaces,
+}
+
 impl Args {
     pub fn new() -> Result<Self> {
         let mut clap_args = match Args::try_parse() {
@@ -250,6 +256,10 @@ pub enum Command {
         /// Do not create or modify .gitignore (default false)
         #[clap(long)]
         no_gitignore: bool,
+
+        /// Specify what should be linked (default "remote cache")
+        #[clap(long, value_enum)]
+        target: Option<LinkTarget>,
     },
     /// Login to your Vercel account
     Login {
@@ -280,7 +290,11 @@ pub enum Command {
     Run(Box<RunArgs>),
     /// Unlink the current directory from your Vercel organization and disable
     /// Remote Caching
-    Unlink {},
+    Unlink {
+        /// Specify what should be unlinked (default "remote cache")
+        #[clap(long, value_enum)]
+        target: Option<LinkTarget>,
+    },
 }
 
 #[derive(Parser, Clone, Debug, Default, Serialize, PartialEq)]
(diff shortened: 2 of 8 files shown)


crates/turborepo-lib/src/commands/link.rs
```diff
@@ -127,56 +126,125 @@ pub async fn link(base: &mut CommandBase, modify_gitignore: bool) -> Result<()>
...
+                        base.ui.rainbow(">>> Success!"),
+                        base.ui.apply(BOLD.apply_to(&repo_root_with_tilde)),
+                        base.ui.apply(BOLD.apply_to(&space.name)),
+                        GREY.apply_to(
+                            "To remove Spaces integration, run `npx turbo unlink --target spaces`"
+                        )
+                    );
+                }
+                Err(e) => {
+                    return Err(anyhow!("Could not persist space id to turbo.json: {}", e));
```
gsoltis, line 241: Might be good to render what we expect them to add here. If we fail to do it automatically, maybe they can do it manually.

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -127,56 +126,125 @@ pub async fn link(base: &mut CommandBase, modify_gitignore: bool) -> Result<()>
         )
     })?;
 
-    let teams_response = api_client
-        .get_teams(token)
-        .await
-        .context("could not get team information")?;
+    match target {
+        LinkTarget::RemoteCache => {
```
gsoltis, line 130: These `match` arms are pretty big. May be a good idea to split them into separate functions?

gsoltis (approved):
LGTM with a couple of nits. 

I would recommend getting a Rust expert to go over it before merging though.

crates/turborepo-lib/src/commands/unlink.rs
```diff
@@ -13,3 +25,67 @@ pub fn unlink(base: &mut CommandBase) -> Result<()> {
...
+        }
+    }
+    Ok(())
+}
+
+fn remove_spaces_from_turbo_json(base: &CommandBase) -> Result<UnlinkSpacesResult> {
+    let turbo_json_path = base.repo_root.join("turbo.json");
+
+    if !turbo_json_path.exists() {
+        return Err(anyhow!("turbo.json not found."));
```
NicholasLYang, line 66: Should we not add this as context to the error on `File::open`?

crates/turborepo-lib/src/commands/unlink.rs
```diff
@@ -13,3 +25,67 @@ pub fn unlink(base: &mut CommandBase) -> Result<()> {
...
+    }
+
+    let turbo_json_file = File::open(&turbo_json_path)?;
+    let mut turbo_json: TurboJson = serde_json::from_reader(turbo_json_file)?;
+    let has_spaces_id = turbo_json
+        .experimental_spaces
+        .unwrap_or(SpacesJson {
```
NicholasLYang, line 73: We could derive `Default` on `SpacesJson` and then do `unwrap_or_default`

crates/turborepo-lib/src/cli.rs
```diff
@@ -280,7 +290,11 @@ pub enum Command {
     Run(Box<RunArgs>),
     /// Unlink the current directory from your Vercel organization and disable
     /// Remote Caching
-    Unlink {},
+    Unlink {
+        /// Specify what should be unlinked (default "remote cache")
+        #[clap(long, value_enum)]
```
NicholasLYang, line 295: We can avoid the `Option` here by using `default_value_t`

NicholasLYang (commented):
Looks pretty great! Just a couple notes

crates/turborepo-lib/src/commands/link.rs
tknickman: Good call - will add

crates/turborepo-lib/src/commands/unlink.rs
tknickman: Love it

crates/turborepo-lib/src/cli.rs
tknickman: Nice!

crates/turborepo-api-client/src/lib.rs
```diff
@@ -213,6 +224,37 @@ impl APIClient {
         })
     }
 
+    pub async fn get_spaces(&self, token: &str, team_id: Option<&str>) -> Result<SpacesResponse> {
+        // create url with teamId if provided
+        let endpoint = match team_id {
+            Some(team_id) => format!("/v0/spaces?limit=100&teamId={}", team_id),
+            None => "/v0/spaces?limit=100".to_string(),
```
mehulkar, line 231: Do we need to limit=100? Shouldn't there be only 1?

crates/turborepo-api-client/src/lib.rs
```diff
@@ -213,6 +224,37 @@ impl APIClient {
...
+        };
+
+        let response = self
+            .make_retryable_request(|| {
+                let request_builder = self
+                    .client
+                    .get(self.make_url(endpoint.as_str()))
+                    .header("User-Agent", self.user_agent.clone())
+                    .header("Content-Type", "application/json")
+                    .header("Authorization", format!("Bearer {}", token));
```
mehulkar, line 241: Have we already required a `token` to be set or could it be `nil` here?

crates/turborepo-lib/src/cli.rs
```diff
@@ -250,6 +256,10 @@ pub enum Command {
         /// Do not create or modify .gitignore (default false)
         #[clap(long)]
         no_gitignore: bool,
+
+        /// Specify what should be linked (default "remote cache")
+        #[clap(long, value_enum)]
+        target: Option<LinkTarget>,
```
mehulkar, line 262: Why not do both links in the same step instead of having two separate paths?

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -31,13 +33,19 @@ pub(crate) enum SelectedTeam<'a> {
...
+}
+
 pub(crate) const REMOTE_CACHING_INFO: &str = "  Remote Caching shares your cached Turborepo task \
                                               outputs and logs across
   all your team’s Vercel projects. It also can share outputs
   with other services that enable Remote Caching, like CI/CD systems.
   This results in faster build times and deployments for your team.";
 pub(crate) const REMOTE_CACHING_URL: &str =
     "https://turbo.build/repo/docs/core-concepts/remote-caching";
+pub(crate) const SPACES_URL: &str = "https://vercel.com/docs/spaces";
```
mehulkar, line 48: I think this should be a turbo.build URL, not vercel. The feature to post to a space belongs to turbo

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -127,56 +126,125 @@ pub async fn link(base: &mut CommandBase, modify_gitignore: bool) -> Result<()>
...
+                        "
+    {} {} linked to {}
+
+    {}
+        ",
+                        base.ui.rainbow(">>> Success!"),
+                        base.ui.apply(BOLD.apply_to(&repo_root_with_tilde)),
+                        base.ui.apply(BOLD.apply_to(&space.name)),
+                        GREY.apply_to(
+                            "To remove Spaces integration, run `npx turbo unlink --target spaces`"
```
mehulkar, line 236: if we know the user invoked global turbo, we should change this to a global turbo command `turbo unlink`. (not sure if we have any affordances for this, and not blocking)

mehulkar (approved):
Looks good to me too. COuple things:

- What happens if you run `turbo link --target spaces` from within  a workspace directory that has a `turbo.json`?  I think it will still go to root turbo.json, but can we error and show a message if needed?
- Not for this PR, but we'll need to coordinate how this is _used_ with the `--experimental-space-id` flag, and we also need it to be configurable with an env var. Just something to keep in mind.

crates/turborepo-lib/src/commands/unlink.rs
tknickman: Hmm - these are two different error types though right? If the file doesn't exist, we can just bail early and not even try and open. 

I did add more error context on the `File::open` call - but if there's a better way to consolidate this let me know!

crates/turborepo-api-client/src/lib.rs
tknickman: We have already required it

crates/turborepo-lib/src/commands/link.rs
tknickman: Yep, this is a placeholder for now

crates/turborepo-lib/src/cli.rs
tknickman: Because you may want to link / unlink without always affecting both. Also during rollout this is easier option. 

We may want to change this in the future once we're fully rolled out

crates/turborepo-api-client/src/lib.rs
tknickman: There will be as many as you currently belong to! In the future we can filter by repo url

tknickman:
@mehulkar yes this will only change root for now, and actually using this val will be a follow up

PR #4635, opened 2023-04-19
Fix misleading GitLab CI documentation

will3942:
### Description

The example `.gitlab-ci.yml` code leads you to believe you need to add the `variables` section to enable remote caching. If you actually uncomment those lines and add that section you end up with environment variables with the *literal* values `TURBO_TEAM=$TURBO_TEAM` and `TURBO_TOKEN=$TURBO_TOKEN`.

The `$` variables do **not** get expanded and so when you then add the variables in the UI these have no effect. The correct behaviour is to not add this variables section in the `.gitlab-ci.yml` and just add the variables in the UI.

A further improvement would be for the turbo CLI to print out an error message when the variables are incorrect / when authentication to the caching server fails.

I've also made an improvement to the suggested cache setup for pnpm. By caching based on the contents of the lockfile we don't have to reinstall packages between branches, this also corresponds with pnpm's [recommended setup](https://pnpm.io/continuous-integration#gitlab-ci).

### Testing Instructions

1. Setup an example turborepo repo on GitLab
2. Follow the old documentation to setup a `.gitlab-ci.yml` file and observe that the cache is always missed
3. Remove the variables section from the `.gitlab-ci.yml` file and observe that the cache is hit


anthonyshew (approved):
Thanks for this! Much appreciated!

PR #4636, opened 2023-04-19
Add support for logging events and intervals in the macOS profiler

alexkirsz:
### Description

Surprisingly, I couldn't find an adequate out-of-the-box crate for adding signposts to the macOS profiler. I had to dig in to semi-private macOS APIs to add this capability.

Here's what signposts look like in the profiler view:

image

### Testing Instructions

N/A

link WEB-931


crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+// Safety: unclear
+unsafe impl Sync for Signpost {}
+// Safety: unclear
+unsafe impl Send for Signpost {}
+
+// NOTE(alexkirsz) Not sure that's correct, but it's what the Dart SDK does and
+// it seems to work.
+// See https://github.com/dart-lang/sdk/blob/3e2d3bc77fa8bb5139b869e9b3a5357b5487df18/runtime/vm/timeline_macos.cc#L34-L35
+static FORMAT_BUFFER: [u8; 64] = [0; 64];
+static FORMAT_BUFFER_LEN: u32 = (FORMAT_BUFFER.len() * std::mem::size_of::<u8>()) as u32;
```
kdy1, line 98: I think this can be a constant, but I'm not sure if it matters

crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+enum SignpostType {
+    Event = os_signpost_type_t_OS_SIGNPOST_EVENT,
+    IntervalBegin = os_signpost_type_t_OS_SIGNPOST_INTERVAL_BEGIN,
+    IntervalEnd = os_signpost_type_t_OS_SIGNPOST_INTERVAL_END,
+}
+
+impl Signpost {
+    /// Emits a signpost event with the given name.
+    pub fn event(&self, name: &str) {
+        let name = CString::new(name).unwrap();
```
kdy1, line 110: We don't need to allocate here.

kdy1: [`CString::new`](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new) and [`CStr::from_bytes_with_nul`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_with_nul) has same error condition

crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        self.emit(name.as_c_str(), None, SignpostType::Event)
+    }
+
+    /// Emits a signpost event with the given name and message.
+    pub fn event_with_message(&self, name: &str, message: &str) {
+        let name = CString::new(name).unwrap();
```
kdy1, line 116: Same here

crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        let message = CString::new(escape_message(message)).unwrap();
```
kdy1, line 117: Not sure, but I think it's same here

crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        self.emit(
+            name.as_c_str(),
+            Some(message.as_c_str()),
+            SignpostType::Event,
+        )
+    }
+
+    /// Starts a new interval signpost with the given name.
+    pub fn interval(&self, name: &str) -> SignpostInterval {
+        let name = CString::new(name).unwrap();
```
kdy1, line 127: Same here

crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        SignpostInterval {
+            signpost: *self,
+            name,
+        }
+    }
+
+    /// Starts a new interval signpost with the given name and message.
+    pub fn interval_with_message(&self, name: &str, message: &str) -> SignpostInterval {
+        let name = CString::new(name).unwrap();
```
kdy1, line 139: Same here

crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        let message = CString::new(escape_message(message)).unwrap();
```
kdy1, line 140: Same here (not sure)

crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+impl SignpostInterval {
+    /// Ends the interval signpost. Calling this is equivalent to dropping the
+    /// interval.
+    pub fn end(self) {}
+
+    /// Ends the interval signpost with the given message.
+    pub fn end_with_message(self, message: &str) {
+        let message = CString::new(escape_message(message)).unwrap();
```
kdy1, line 183: Same here

crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn drop(&mut self) {
+        self.signpost
+            .emit(self.name.as_c_str(), None, SignpostType::IntervalEnd);
+    }
+}
+
+/// Escapes a message for use with signposts, which will try to parse it as a
+/// format string.
+fn escape_message(message: &str) -> String {
+    message.replace('%', "%%")
```
kdy1, line 203: This can be `Cow` with check for existence of `%`

crates/utils/signposter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        Log {
+            os_log: unsafe { &_os_log_default as *const _ as *mut _ },
+        }
+    }
+}
+
+impl Log {
+    /// Create a new log with the given subsystem and category.
+    pub fn new_with_subsystem_and_category(subsystem: &str, category: LogCategory) -> Self {
+        let subsystem = CString::new(subsystem).unwrap();
```
kdy1, line 65: Same here, the content of `subsystem` is not dropped while the function returns

sokra:
Is it possible use the https://crates.io/crates/tracing and make signposts a provider for that?

crates/utils/signposter/src/lib.rs
alexkirsz: I can always hardcode it to 64 and use it in `FORMAT_BUFFER`'s initialization.

crates/utils/signposter/src/lib.rs
alexkirsz: Rust's `&str` aren't null-terminated. We need to copy the string and append `\0`, which is what `CString::new` does here (and everywhere below).

crates/utils/signposter/src/lib.rs
alexkirsz: Can you explain what you mean here?

alexkirsz:
@sokra Yes. Right now this implementation is very minimal as the goal was to get it working first and foremost to get some insights fast, but eventually we definitely should integrate it with tracing.

alexkirsz:
> Failed to execute xcrun: Os { code: 2, kind: NotFound, message: "No such file or directory" }

https://github.com/vercel/turbo/actions/runs/4745099662/jobs/8426941687?pr=4636

I need to make the whole crate a noop for anything else than macOS.

alexkirsz:
I've added a tracing layer which is now used in turbopack-dev-server instead of using `signposter` directly.

Be aware that tracing is currently force-disabled in `nextpack` because some next-swc crates enable  `max_level_off`/`max_level_*` features, which are applied on the whole workspace. cc @kdy1 could we disable these by default somehow and only enable them for actual releases in CI?

kdy1:
Yes, it's possible and it can be done using `sed`. Note that `sed` behaves bit differently on mac OS.

sokra (approved):

PR #4639, opened 2023-04-19
restrict snapshot path for windows path length limit

sokra:
### Description

windows...


crates/turbopack-test-utils/src/snapshot.rs
```diff
@@ -31,19 +31,21 @@ pub async fn snapshot_issues<
     let expected_issues = expected(issues_path).await?;
     let mut seen = HashSet::new();
     for (plain_issue, debug_string) in captured_issues.into_iter() {
+        let title = plain_issue
+            .title
+            .replace('/', "__")
+            // We replace "*", "?", and '"' because they're not allowed in filenames on Windows.
```
kwonoj, line 37: just in case worth to add `:` as well?

kwonoj (approved):

PR #4640, opened 2023-04-19
Limit the cache inputs for the docs rss and schema tasks

mehulkar:


docs/turbo.json
```diff
@@ -14,12 +14,17 @@
       "dependsOn": ["^build", "rss", "schema"]
     },
     "dev": {
-      "dependsOn": ["schema"]
+      "dependsOn": ["schema"],
+      "persistent": true,
+      "cache": false
     },
     "schema": {
+      "inputs": [], // None of the files in this workspace matter
```
mehulkar, line 22: This empty array setting _should_ work, but I'm seeing that in `--dry=json` and `--summarize` it still takes the full workspace. Not a blocker for this PR, but it's a bug we should fix

docs/turbo.json
```diff
@@ -14,12 +14,17 @@
...
+      "dependsOn": ["@turbo/types#topo"],
```
tknickman, line 23: We _really_ need a better way to do these. It's the most confusing thing

tknickman (approved):

PR #4643, opened 2023-04-19
Fix path

mehulkar:


gsoltis (approved):
Thanks!

mehulkar:
removed automerge label since i'm already using GH's automerge feature (and prefer it, because I can customize the commit message). don't want the tools to compete with each other

PR #4644, opened 2023-04-19
Disable cargo-sweep install when setting up turborepo environment

mehulkar:
This is very slow and affects turborepo test suites


kdy1 (approved):

ForsakenHarmony:
hmm, can you show a run where this is slow?

PR #4649, opened 2023-04-20
Image processing improvements

sokra:
### Description

* refactor image processing
* fallback guess format for extension
* expose pathname_to_url for wrapped source usage

see also https://github.com/vercel/next.js/pull/48622

### DIFF
--- a/crates/turbopack-image/src/process.rs
+++ b/crates/turbopack-image/src/process.rs
@@ -8,13 +8,13 @@ use image::{
         png::{CompressionType, PngEncoder},
     },
     imageops::FilterType,
-    GenericImageView, ImageEncoder, ImageFormat,
+    DynamicImage, GenericImageView, ImageEncoder, ImageFormat,
 };
 use mime::Mime;
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DisplayFromStr};
 use turbo_tasks::{debug::ValueDebugFormat, primitives::StringVc, trace::TraceRawVcs};
-use turbo_tasks_fs::{FileContent, FileContentVc, FileSystemPathVc};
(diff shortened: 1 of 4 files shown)


crates/turbopack-image/src/process.rs
```diff
@@ -109,32 +172,45 @@ fn compute_blur_data(
...
         }
         #[cfg(feature = "avif")]
         ImageFormat::Avif => {
             use image::codecs::avif::AvifEncoder;
-            AvifEncoder::new_with_speed_quality(&mut buf, 6, options.quality).write_image(
-                small_image.as_bytes(),
-                blur_width,
-                blur_height,
-                small_image.color(),
+            AvifEncoder::new_with_speed_quality(&mut buf, 6, quality - 15).write_image(
```
alexkirsz, line 180: These should go into constants that explain how we came up with these numbers. Why `- 15` here?

crates/turbopack-image/src/process.rs
```diff
@@ -233,6 +281,45 @@ pub async fn get_meta_data(
...
+    }
+    #[cfg(not(feature = "webp"))]
+    if matches!(format, Some(ImageFormat::WebP)) {
+        format = Some(ImageFormat::Jpeg);
+    }
+    let format = if let Some(format) = format {
+        format
+    } else {
+        ImageFormat::Jpeg
+    };
```
alexkirsz, line 317:
```suggestion
    let format = format.unwrap_or(ImageFormat::Jpeg);
```

alexkirsz (approved):

crates/turbopack-image/src/process.rs
sokra: removed it... I copied that from next.js but there seem to be no reason for that...

PR #4662, opened 2023-04-21
feat(ecma-plugins): extract custom emotion transform

kwonoj:
### Description

First step for WEB-940.

Context: https://vercel.slack.com/archives/C03EWR7LGEN/p1681789689115509

Currently all of the ecma transforms are explicitly listed under EcmaInputTransform in turbopack-ecmascript. This makes enum verbose, we have to manually expand it each time adding new transform, as well as turbopack-ecmascript gets larger to contain all of the 3rd party transforms by default.

PR extracts non-core transforms into a new crate, named as ecmascript-plugins then utilize EcmaInputTransform::Custom to invoke transforms instead. `EcmaInputTransform::Custom` is renamed to `EcmaInputTransform::Plugin` as well. Goal is extracting all of 3rd party / non-core transforms. This also reduces multiple steps to construct option value between caller (next-*) to actual transform (swcOptions).

https://github.com/vercel/next.js/pull/48671 have corresponding next.js changes.

### DIFF
--- a/crates/turbopack/src/module_options/mod.rs
+++ b/crates/turbopack/src/module_options/mod.rs
@@ -15,8 +15,9 @@ use turbopack_core::{
 };
 use turbopack_css::{CssInputTransform, CssInputTransformsVc};
 use turbopack_ecmascript::{
-    EcmascriptInputTransform, EcmascriptInputTransformsVc, EcmascriptOptions,
+    CustomTransformVc, EcmascriptInputTransform, EcmascriptInputTransformsVc, EcmascriptOptions,
 };
+use turbopack_ecmascript_plugins::transform::emotion::build_emotion_transformer;
 use turbopack_mdx::MdxTransformOptions;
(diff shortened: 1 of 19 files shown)


crates/turbopack/src/module_options/mod.rs
```diff
@@ -99,23 +100,13 @@ impl ModuleOptionsVc {
...
-                } else {
-                    None
-                },
-            });
+
+        if let Some(transformer) = build_emotion_transformer(enable_emotion).await? {
+            transforms.push(EcmascriptInputTransform::Plugin(TransformPluginVc::cell(
+                transformer,
+            )));
         }
```
sokra, line 108: (Eventually) `enable_emotion` should be removed from the ModuleOptionsContext and passed as `custom_ecmascript_transforms`

sokra (approved):

crates/turbopack/src/module_options/mod.rs
kwonoj: Yes, that makes more sense.

PR #4663, opened 2023-04-21
test: Improve testing system for tree shaker

kdy1:
### Description

We need a testing system that supports enabling only _some_ of exports. This PR implements it by using `config.json`.

Fix WEB-917.

### Testing Instructions

Look at output.mdx


sokra (approved):

alexkirsz (approved):
🚀

PR #4664, opened 2023-04-21
chore: specify rust crates as inputs to building turborepo

mehulkar:


cli/turbo.json
```diff
@@ -3,7 +3,34 @@
   "extends": ["//"],
   "pipeline": {
     "build": {
-      "outputs": ["turbo", "turbo.exe"]
+      "outputs": [
+        // TODO: this gets moved over to ../target, can we omit this?
+        "turbo",
```
NicholasLYang, line 8: yeah I'm not sure why this is here. We build `go-turbo` in `cli` but not `turbo` anymore

cli/turbo.json
```diff
@@ -3,7 +3,34 @@
...
+        // TODO: does this get moved over to ../target also? Can we omit this too?
+        "turbo.exe",
+        // TODO: This is 7GB, do we want to ever even cache the debug build?
+        "../target/debug/**",
+        // TODO: Do we need the whole directory? go-turbo and turbo are ~45mb, the whole directory is 800mb
+        "../target/release/**"
```
NicholasLYang, line 14: Probably not without lockfile analysis. If we can be smart and cache rust dependencies and restore them even without a full cache hit, then caching would make sense.

cli/turbo.json
```diff
@@ -3,7 +3,34 @@
...
+      ],
+
+      "inputs": [
+        "**/*.go",
+        "**/*_test.go",
+
+        // Rust crates that should trigger a rebuild
+        // TODO: this is not the full list (e.g. pidlock is missing here)
+        // what else do we need here? And should turbo be crawling / using cargo
```
NicholasLYang, line 23: One thing I've been meaning to do is rename all of the turborepo crates to be prefixed with `turborepo`, as agreed upon in conversations with web-tooling. That'll make it easy to filter

cli/turbo.json
```diff
@@ -3,7 +3,34 @@
...
+        // to figure this out automatically? Or should we wire up a package.json dependency graph?
+        "../crates/turborepo/**",
+        "../crates/turbopath/**",
+        "../crates/turborepo-*/**",
+
+        // Exclude target directories of everything specified above
+        "!../crates/turborepo/target/**",
```
NicholasLYang, line 30: These shouldn't be here unless explicitly specified as an output directory (like in the `turborepo-ffi` rule in the Makefile). Otherwise the default output is `target/` at the root of the repo.

cli/turbo.json
mehulkar: That would be nice, but I assume there will still be crates that both teams wil use (hopefully that's the purpose of monorepo'ing both). I'm not sure how we'll account for that without crawling deps (maybe this is what you meant by "Rust support" in the team meeting?)

cli/turbo.json
mehulkar: Just making sure you see that these are `!` prefixed?  (which means that it's configured to "shouldn't be here") as you say?

cli/turbo.json
mehulkar: probably because it was there before rust

cli/turbo.json
NicholasLYang: I think that's a bridge we'll cross when we come to it. We can always come up with another prefix like `shared-`

cli/turbo.json
NicholasLYang: Yeah, so it's nbd either way, but mostly point out that these directories won't exist in the first place

mehulkar: oh, I only added them because they did exist 🤔 , otherwise wouldn't have known to exclude them. I'll double check

cli/turbo.json
```diff
@@ -3,8 +3,18 @@
...
+        "../target/debug/go-turbo",
+        "../target/debug/turbo",
+        "../target/release/go-turbo",
+        "../target/release/turbo"
+      ],
+
+      "inputs": [
+        "**/*.go",
+        "**/*_test.go",
+        "../crates/turborepo-*/**" // Rust crates
```
mehulkar, line 16: @NicholasLYang going to block this on #4853, so it's easier to specify the dependencies. Let me know if there's a better way to do this too

PR #4666, opened 2023-04-21
Remove smoke test before publish

mehulkar:
I'm not sure there's much value in running unit tests before a release and it takes about 10 minutes to run this step (e.g. https://github.com/vercel/turbo/actions/runs/4761979875) 

If we really think they can catch bugs before release, but have two arguments against:

1. We are actively writing fewer unit tests in Go as we port to Rust, so the value of these diminishes over time
2. From other parts of our team process, we expect that "`main` is always shippable". If this is true and we have the right checks in place to make sure this is true, then a smoke test before release seems unnecessary,.

The argument for keeping _some_ kind of smoke test is that multiple merges to main that are independently tested may end up breaking each other. I think this is valid reasoning, but at a cost of 10 minutes, I'm not sure it's worth the cost. 

#### Alternatives

- Keep smoke tests for patch/minor/major releases, but omit them for canary
- Instead of running the test suite, just run `make` and see if the code compiles
- Run integration tests instead of unit tests before release
- Reduce babysitting for releases, so an extra 10 minutes does not feel expensive (we are effectively waiting for full Rust port before we do this)

### DIFF
--- a/.github/workflows/turborepo-release-step-2.yml
+++ b/.github/workflows/turborepo-release-step-2.yml
@@ -10,22 +10,7 @@ on:
         description: "Staging branch to run release from"
 
 jobs:
-  smoke-test:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          ref: ${{ inputs.release_branch }}
-      - name: Build turborepo CLI from source
-        uses: ./.github/actions/setup-turborepo-environment
-        with:
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
-          target: ${{ matrix.os.name }}
-      - name: Run Integration Tests
-        run: turbo run test --filter=turborepo-tests-integration
-
   darwin:
-    needs: [smoke-test]
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v3
@@ -59,7 +44,6 @@ jobs:
 
   # compiles linux and windows in a container
   cross:
-    needs: [smoke-test]
     runs-on: ubuntu-latest
     container:
       image: docker://ghcr.io/vercel/turbo-cross:v1.18.5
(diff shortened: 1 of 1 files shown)


nathanhammond:
I'm pretty strongly opposed to this as it is an infrequently run task and is a nice insurance policy. If we turn the entire release process into "press one button" then we basically don't care about this until we decide that we need to get the time spent going from "release button-press to user's hands" shorter.

We stop caring after we press the button.

mehulkar:
> is a nice insurance policy.

are they though? maybe if they were integration, e2e tests, or if we ran examples tests with a local build. are the unit tests really providing any security?

> We stop caring after we press the button.
I'd still care. Even if we add nightly canaries, it is very likely that we'll want mid-day canary releases for any number of reasons. If I'm pressing the button, shaving 10m off the time before I'm able to use it is really great.

arlyon:
I have no strong opinions about this but if I were to pick between removing this or putting everything into a one-step click and forget I would choose the latter. One thing worth verifying is if the smoke tests actually trigger rust tests, so I would almost argue that this should be to add rust (integration) tests to this, than to remove the go unit tests.

mehulkar:
> One thing worth verifying is if the smoke tests actually trigger rust tests

at the moment they don't: https://github.com/vercel/turbo/blob/ad4bc412ac5629f5786f3bd3682e7af2e8f6c51a/.github/workflows/turborepo-release-step-2.yml#L24-L25

tknickman (commented):
I also agree with @nathanhammond here, we're going to be saving a lot of time consolidating to one release workflow that I don't think we're going to mind keeping a smoke test step in here. As far as usefulness we can always revisit what we're running, but I think we should have _something_

mehulkar:
Sounds good, I"ll close this down for now.

PR #4670, opened 2023-04-21
dev-server content might change any time, we can't cache them

sokra:
### Description

Browser cache might lead to inconsistent state


crates/turbopack-dev-server/src/http.rs
```diff
@@ -144,6 +144,13 @@ pub async fn process_request_with_content_source(
                     )?);
                 }
 
+                if !header_map.contains_key("cache-control") {
```
alexkirsz, line 147:
```suggestion
                if !header_map.contains_key("cache-control") {
                    // The dev server contents might change at any time, we can't cache them.
```

PR #4672, opened 2023-04-21
Use global turbo install for smoke test

mehulkar:
Smaller change from the one in #4666. This avoids building turbo multiple times before running unit tests before a release

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -234,7 +234,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: ./.github/actions/setup-turborepo-environment
+      - name: Build turborepo CLI from source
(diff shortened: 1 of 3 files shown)


tknickman (approved):

.github/workflows/test.yml
```diff
@@ -1105,8 +1109,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-
-      - uses: ./.github/actions/setup-turborepo-environment
+      - name: Build turborepo CLI from source
```
NicholasLYang, line 1112: Does `setup-turborepo-environment` build it from source? I thought it just set up the environment

mehulkar: It does now, I added it in https://github.com/vercel/turbo/pull/4647

PR #4673, opened 2023-04-21
Remove misleading github action name

mehulkar:


chris-olszewski (approved):
Do we want to change the name to something along the lines of "setup turborepo build tools" instead of removing the name completely?

mehulkar:
@chris-olszewski I couldn't think of a good name. "setup turborepo build tools" also doesn't account for the global installation

Issue #4469, opened 2023-04-05
[turborepo] migrate-env-var-dependencies codemod fails with error: TypeError: Za(...).map is not a function


graup:
### What version of Turborepo are you using?

1.8.8

### What package manager are you using / does the bug impact?

Yarn v2/v3 (node_modules linker only)

### What operating system are you using?

Mac

### Describe the Bug

```
npx @turbo/codemod migrate-env-var-dependencies
? Where is the root of the repo where the transform should run? .
 INFO  Migrating environment variable dependencies from "globalDependencies" and "dependsOn" to "env" in "turbo.json"...

Unexpected error. Please report it as a bug:
TypeError: Za(...).map is not a function
    at za (/Users/paulgrau/.npm/_npx/b5f7626c653117ba/node_modules/@turbo/codemod/dist/transforms/migrate-env-var-dependencies.js:52:1169)
    at Object.ni [as transformer] (/Users/paulgrau/.npm/_npx/b5f7626c653117ba/node_modules/@turbo/codemod/dist/transforms/migrate-env-var-dependencies.js:56:17017)
    at Command.L (/Users/paulgrau/.npm/_npx/b5f7626c653117ba/node_modules/@turbo/codemod/dist/cli.js:6:29)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Command.parseAsync (/Users/paulgrau/.npm/_npx/b5f7626c653117ba/node_modules/commander/lib/command.js:935:5)
```

### Expected Behavior

No error

### To Reproduce

Our turbo.json looks like this:

```
{
  "baseBranch": "origin/main",
  "pipeline": {
    "build": {
      "dependsOn": ["^build"],
      "outputs": ["dist/**", ".next/**"]
    },
    "lint": {
      "dependsOn": ["^build"],
      "outputs": []
    },
    "format": {
      "dependsOn": [],
      "outputs": []
    },
    "test": {
      "dependsOn": ["^build"],
      "outputs": []
    },
    "dev": {
      "dependsOn": ["^build"],
      "cache": false
    },
    "storybook": {
      "cache": false
    },
    "product-test#build": {
      "dependsOn": ["^build"],
      "outputs": [".next/**"]
    },
    "product#build": {
      "dependsOn": [
        "^build",
        "$APP_ENV",
        "$VERCEL_ENV",
        "$VERCEL_GIT_COMMIT_REF",
      ],
      "outputs": [".next/**"]
    }
  }
}

```

[...]

Issue #4085, opened 2023-03-06
[eslint-config-turbo] 0.0.9 throws Ga(...).map is not a function error


mikecousins:
### What version of Turborepo are you using?

0.0.9

### What package manager are you using / does the bug impact?

Yarn v2/v3 (node_modules linker only)

### What operating system are you using?

Mac

### Describe the Bug

Updating from 0.0.8 to 0.0.9 throws the following error:

`TypeError: Failed to load plugin 'turbo' declared in '../../.eslintrc.js » eslint-config-pm » eslint-config-turbo': Ga(...).map is not a function`

We're using Typescript, we have a shared package for our eslint config.

Package config:

```
module.exports = {
  env: {
    browser: true,
    node: true,
  },
  parser: '@typescript-eslint/parser',
  plugins: ['@typescript-eslint'],
  settings: {
    react: {
      version: 'detect',
    },
  },
  extends: [
    'eslint:recommended',
    'plugin:react/recommended',
    'plugin:react-hooks/recommended',
    'plugin:@typescript-eslint/recommended',
    'plugin:prettier/recommended',
    'plugin:cypress/recommended',
    'plugin:jsx-a11y/recommended',
    'turbo',
  ],
  rules: {
    'react/button-has-type': 'error',
    'react/react-in-jsx-scope': 'off', // need this for vite support
    'jsx-a11y/label-has-associated-control': [
      1,
      {
        controlComponents: ['ReadonlyText', 'Field'],
      },
    ],
    // Automatically flag env vars missing from turbo.json when process.env is used
    'turbo/no-undeclared-env-vars': 'error',
    '@typescript-eslint/no-unused-vars': [
      'error',
      {
        argsIgnorePattern: '^_',
        destructuredArrayIgnorePattern: '^_',
        ignoreRestSiblings: true,
      },
    ],
  },
};
```

Root config:
```
module.exports = {
  // This tells ESLint to load the config from the workspace `eslint-config-pm`
  extends: ['pm'],
  root: true,
};
```

While debugging this I've tried removing all other plugins, all the rule overrides, the env node, and the settings node without success.

### Expected Behavior

Should run lint successfully.

### To Reproduce

Update to 0.0.9 of eslint-config-turbo and attempt to run lint.

tknickman:
Will take a look, thanks for the report, can you try running with debug logging enabled:
`DEBUG=eslint:cli-engine yarn lint` 

and share any more output?

[...]

PR #4682, opened 2023-04-24
fix(turbo-utils): support old workspace format

tknickman:
### Description

Support old workspace format (workspaces.packages) in our find config util used by codemod and the eslint plugin 

Fixes https://github.com/vercel/turbo/issues/4469 and https://github.com/vercel/turbo/issues/4085


packages/turbo-utils/__fixtures__/common/old-workspace-config/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "private": true,
+  "workspaces": {
+    "packages": [
+      "apps/*",
+      "packages/*"
+    ]
```
mehulkar, line 7: Doesn't look like you _need_ 3 workspaces for this test and this use case. I'd try to keep fixtures as small as possible so it's easier to understand what part of them matters

mehulkar (approved):

Issue #2791, opened 2022-11-21
[turborepo] [yarn 3] pruned lockfile missing resolution entries


erj826:
### What version of Turborepo are you using?

1.6.3

### What package manager are you using / does the bug impact?

Yarn v2/v3 (node_modules linker only)

### What operating system are you using?

Mac

### Describe the Bug

To optimize our Docker builds we are attempting to use turbo prune to create a lighter weight image. After running prune we copy our pruned lockfile and .json files into a directory and then attempt to run an immutable install.

We have observed that when the root package.json includes yarn resolutions we are unable to run `yarn --immutable` successfully in the _pruned_ monorepo.

### For example

In this reproduction `ajv` is a dependency of `eslint` so it is a necessary dependency in the pruned monorepo.

```
ericjacobson@Erics-MacBook-Air pruned % yarn why ajv
├─ @eslint/eslintrc@npm:0.4.3
│  └─ ajv@npm:6.12.6 (via npm:^6.12.4)
│
├─ eslint@npm:7.32.0
│  └─ ajv@npm:6.12.6 (via npm:^6.10.0)
│
└─ table@npm:6.8.1
   └─ ajv@npm:8.11.2 (via npm:^8.0.1)
```

When the resolution **is not** included in the root package.json the generated `out/yarn.lock` is correct, and includes an `ajv` entry.

When the resolution **is** included in the root package.json then generated `out/yarn.lock` is **incorrect**, and does not include any `ajv` entries.

### Expected Behavior

The pruned lockfile should include dependencies that have a corresponding yarn resolution.

### To Reproduce

See the README.md in the reproduction repo for more details.

### Reproduction Repo

https://github.com/erj826/turbo-prune-bug-repro-11-21

quinnturner:
This was first mentioned in my ticket in #2049. I will close my version as this ticket is more scoped. CC: @chris-olszewski, since you're assigned to that ticket, perhaps you intend to assign yourself to this ticket?

alexbchr:
Any update on this? I don't see much activity happening on this issue but I have the exact same problem in our app and this is quite blocking.

erj826:
Looks like this work might be in progress. https://github.com/vercel/turbo/pull/4437

@chris-olszewski do you still expect this to be fixed in the `1.9` release?

chris-olszewski:
It should be released as part of a canary next week.

chris-olszewski:
[...]

PR #4684, opened 2023-04-24
feat: Use Rust Berry lockfile impl

chris-olszewski:
### Description

Built on top of #4629

Overview of changes:
 - ~~Change `subgraph` so we calculate workspace's closure in parallel in a similar manner to Go~~ This had to get reverted, it appears that rayon was causing [compilation issues on linux](https://github.com/vercel/turbo/actions/runs/4791974785/jobs/8522988391#step:4:648)
 - `npm_subgraph` -> `subgraph` makes this FFI call generic
 - Addition of an optional `resolutions` map to the `subgraph` and `transitive_closure`, this is only used by Berry. In the ideal world this would be all of the root `package.json`, but serializing all of that via protobuf is overkill since berry is the only lockfile that needs that info and we only use that field.
 - Addition of `patches` FFI function, this wasn't needed before as NPM doesn't have patch files encoded in the lockfile
 - Addition of `global_changes` FFI function, previously for NPM we were just doing this in Go since there were just two fields we needed to check and parsing JSON is cheap.
 - Switches lockfile usage of berry from the Go implementation to the Rust one introduced in #4589 

Reviewer notes:
 - The passing of the optional resolutions map is very icky and one off, but this is the only lockfile the requires additional information to properly parse. Once the package graph is squarely in Rust land this should go back to being more generic.

This will finally close out #2791

### Testing Instructions
Added a new integration test that uses resolutions. This integration test invokes the new `patches` FFI call.

For a full test of the new resolution feature, clone and follow the instructions in the [repro](https://github.com/erj826/turbo-prune-bug-repro-11-21)

The `lockfile_aware_caching` integration tests all hit the new `global_changes` FFI callsite and provide test coverage.


NicholasLYang:
Pretty sure you can use rayon if you add `-lm` to the linker flags. I ran into this recently too

crates/turborepo-ffi/src/lockfile.rs
```diff
@@ -4,7 +4,9 @@ use std::{
 };
 
 use thiserror::Error;
-use turborepo_lockfiles::{self, npm_subgraph as real_npm_subgraph, NpmLockfile, Package};
+use turborepo_lockfiles::{
+    self, npm_subgraph as real_npm_subgraph, BerryLockfile, LockfileData, NpmLockfile, Package,
```
NicholasLYang, line 8: Can we rename this back to just `npm_subgraph` now?

crates/turborepo-ffi/src/lockfile.rs
```diff
@@ -56,38 +64,122 @@ fn npm_transitive_closure_inner(
...
+    let proto::SubgraphRequest {
+        contents,
+        workspaces,
+        packages,
+        resolutions,
+        ..
+    } = request;
+    let contents = match package_manager {
+        proto::PackageManager::Npm => real_npm_subgraph(&contents, &workspaces, &packages)?,
+        proto::PackageManager::Berry => turborepo_lockfiles::berry_subgraph(
```
NicholasLYang, line 118: nit: could we use the same style of usage for both here and npm? Like either imported or with the `turborepo_lockfiles::` specified?

NicholasLYang (approved):
Makes sense to me

PR #4686, opened 2023-04-24
refactor(turborepo): cache signature authentication

NicholasLYang:
### Description

Refactors the cache signature authentication code to no longer rely on Go implementation details.

This requires a cache bump as it is a **breaking** change to cache signatures.

### Testing Instructions

Existing cache signature authentication now calls the Rust implementation and confirms that it produces the same tag. There are also new tests that randomly generate test cases and confirm that the Rust and Go implementation agree on the tag output.


cli/internal/cache/cache_signature_authentication.go
```diff
@@ -44,21 +47,12 @@ func (asa *ArtifactSignatureAuthentication) generateTag(hash string, artifactBod
...
-	}{
-		Hash:   hash,
-		TeamId: teamId,
-	}
-	metadata, err := json.Marshal(artifactMetadata)
+	secret, err := asa.getSecretKey()
 	if err != nil {
 		return nil, err
 	}
+	metadata := []byte(hash)
```
NicholasLYang, line 55: Are there any cases where a plain byte slice could be different in Rust and Go?

cli/internal/cache/cache_signature_authentication_test.go
```diff
@@ -7,14 +7,16 @@ import (
...
+	"math/rand"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/vercel/turbo/cli/internal/ffi"
 )
 
 func Test_SecretKeySuccess(t *testing.T) {
-	teamId := "team_someid"
+	teamID := "team_someid"
```
NicholasLYang, line 19: Go linter complained about `teamId`

cli/internal/cache/cache_signature_authentication.go
gsoltis: As long as the encodings match for the strings, I believe they should be identical.

cli/internal/cache/cache_signature_authentication_test.go
```diff
@@ -93,8 +95,51 @@ func Test_SecretKeyErrors(t *testing.T) {
...
 }
 
+func generateRandomBytes() []byte {
+	length := rand.Intn(250)
+	b := make([]byte, length)
+	rand.Read(b)
+	return b
+}
+
+var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
```
chris-olszewski, line 105: I don't love the arbitrary nature of this restriction of runes we're considering as valid letters. Especially since it's missing 0-9 which will appear in a lot of hashes.

If we are only going to be fuzzing a subset of possible values of `string` I feel like `generateTag` should accept some newtype that validates at construction that the string only contains expected runes. I don't think we'd need to actually check this as I believe we're getting all of our hashes from `hex.Encode` which outputs `[0-f]`. I realize that requires quite a bit of plumbing, but it gives us a compile time guarantee that we're not passing in anything unexpected as `hash`.

crates/turborepo-ffi/messages.proto
```diff
@@ -89,3 +89,18 @@ message SubgraphResponse {
     string error = 2;
   }
 }
+
+message VerifySignatureRequest {
+  string hash = 1;
+  bytes artifact_body = 2;
+  string team_id = 3;
```
chris-olszewski, line 96: `hash` and `team_id` being strings worry me since a Go string "holds arbitrary bytes" unlike Rust. 
I feel that we should either
 - Make this field bytes
 - Do verification on the Go side for `team_id` and error for a user if they try to use a non-utf8 team id, as mentioned previously I think we can verify `hash` by using a newtype for `PackageTask.Hash`

chris-olszewski (changes requested):
I *think* this is correct, but I'm not comfortable with the fact that the fuzz testing we're doing isn't testing a wide range of inputs that *could* make into into the signature.

cli/internal/cache/cache_signature_authentication_test.go
NicholasLYang: Yeah for hash I'm now creating a random sequence of bytes and hashing it to get our random hash

crates/turborepo-ffi/messages.proto
NicholasLYang: I converted team_id to bytes. For hash, I feel like if we're now testing the accurate domain, i.e. random hashes, then it's okay to keep it as a string.

chris-olszewski (approved):
Fuzzing ❤️

jridgewell (approved):

PR #4691, opened 2023-04-25
Allow for dropped updates in the benchmarks

alexkirsz:
### Description

This improves the reliability of HMR benchmarks for bundlers that can drop updates (Vite, Turbopack). Before, an update being dropped meant the whole benchmark was canceled.

### Testing Instructions

Running the benchmarks

### DIFF
--- a/crates/turbopack-bench/src/lib.rs
+++ b/crates/turbopack-bench/src/lib.rs
@@ -33,7 +33,14 @@ use crate::{bundlers::Bundler, util::PageGuard};
 pub mod bundlers;
 pub mod util;
 
-const MAX_UPDATE_TIMEOUT: Duration = Duration::from_secs(60);
+/// The maximum amount of time to wait for HMR during the setup and warmup
+/// phase.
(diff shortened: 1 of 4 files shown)


crates/turbopack-bench/src/lib.rs
```diff
@@ -33,7 +33,14 @@ use crate::{bundlers::Bundler, util::PageGuard};
...
 
-const MAX_UPDATE_TIMEOUT: Duration = Duration::from_secs(60);
+/// The maximum amount of time to wait for HMR during the setup and warmup
+/// phase.
+const MAX_INIT_UPDATE_TIMEOUT: Duration = Duration::from_secs(60);
+/// The maximum amount of time to wait for HMR during the actual benchmark. This
+/// is a lot shorter than the init timeout because we expect updates to
+/// generally happen quickly, and we don't want to wait for a long time when an
+/// update is dropped.
+const MAX_UPDATE_TIMEOUT: Duration = Duration::from_secs(5);
```
sokra, line 43: Isn't that a bit too short for 30k modules with e. g. webpack?

PR #4692, opened 2023-04-25
add FixedStaticAsset and TextContentSourceAsset

sokra:
### Description

needed for metadata support in next.js

### DIFF
--- a/crates/turbopack-ecmascript/src/text/mod.rs
+++ b/crates/turbopack-ecmascript/src/text/mod.rs
@@ -1,0 +1,50 @@
+use anyhow::Result;
+use turbo_tasks::primitives::StringVc;
+use turbo_tasks_fs::FileContent;
+use turbopack_core::{
+    asset::{Asset, AssetContentVc, AssetVc},
+    ident::AssetIdentVc,
+};
+
+use crate::utils::StringifyJs;
+
+#[turbo_tasks::function]
+fn modifier() -> StringVc {
+    StringVc::cell("text content".to_string())
+}
+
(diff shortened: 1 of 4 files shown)


crates/turbopack-ecmascript/src/text/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    #[turbo_tasks::function]
+    async fn content(&self) -> Result<AssetContentVc> {
+        let source = self.source.content().file_content();
+        let FileContent::Content(content) = &*source.await? else {
+            return Ok(FileContent::NotFound.cell().into());
+        };
+        let text = content.content().to_str()?;
+        let text = text.trim_end().replace("\r\n", "\n");
```
alexkirsz, line 45: It seems arbitrary that we should apply these transforms to the source file. Does `asset/source` in webpack also do this?

crates/turbopack-static/src/fixed.rs
```diff
@@ -0,0 +1,39 @@
+use anyhow::Result;
+use turbo_tasks_fs::FileSystemPathVc;
+use turbopack_core::{
+    asset::{Asset, AssetContentVc, AssetVc},
+    ident::AssetIdentVc,
+};
+
+/// A static asset that is served that a fixed output path. It won't use
```
alexkirsz, line 8:
```suggestion
/// A static asset that is served at a fixed output path. It won't use
```

alexkirsz (approved):

ForsakenHarmony (approved):

PR #4694, opened 2023-04-25
swap out log for tracing and add the TurboFormatter

arlyon:
This allows us to switch out facades when needed and sets up a default facade for stdout that follows the current formatting for turbo which is to simply ignore all event fields except the message.

The output format is identical.


crates/turborepo-lib/src/formatter.rs
```diff
@@ -0,0 +1,103 @@
...
+            .and_then(|_| write_message::<Yellow, Default>(writer, event)),
+            &Level::INFO => write_message::<Default, Default>(writer, event),
+            // trace and debug use the same style
+            _ => {
+                let now = Local::now();
+                write!(
+                    writer,
+                    "{} [{}] {}: ",
+                    // build our own timestamp to match the hashicorp/go-hclog format used by the
+                    // go                 // binary
```
tknickman, line 61: Nice. Make sure these match the go binary when running with `-vvv` (or similar)!

Also, weird comment format here (trailing `// binary`)

tknickman (commented):
Nice, looks good to me - I'll let a rust pro check it out but nothing blocking from my end

crates/turborepo-lib/src/shim.rs
```diff
@@ -677,70 +678,21 @@ fn is_turbo_binary_path_set() -> bool {
...
-                now.format("%Y-%m-%dT%H:%M:%S.%3f%z"),
-                record.level(),
-                record.target(),
-                record.args()
-            )
-        }
-    });
-
-    builder.init();
+    // respect TURBO_LOG_VERBOSITY env var
```
mehulkar, line 689: is this new? i didn't know this was a thing

mehulkar (approved):
What did you mean by "facade" in the PR description?  Are we planning to invoke Rust to log things from Go land?

crates/turborepo-lib/src/shim.rs
arlyon: All I mean is that tracing lets you output logs to a bunch of different places; stdout, files, [opentelemetry](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/) and even [chrome tracing](https://docs.rs/tracing-chrome/latest/tracing_chrome/) through the same logging / span infrastructure :)

arlyon: Our default 'subscriber' is one that prints to stdout following the same format as our go code.

PR #4699, opened 2023-04-25
Combine release process into a single step

mehulkar:
Single step deployment so we don't have to babysit releases. This is _mostly_ a copy paste of the existing workflows into a single workflow, but a few changes:

- The Rust and Go artifacts are now built in parallel. This was not a necessary optimization for this PR, but requested by @tknickman 
- In steps 2 and 3 `inputs.release_branch` was provided. Now it's set as an `output` from the `staging` step and then read into the `env` at the beginning of each step. Initially I chose this approach because the diff was easier to see:

    ```diff
    - {{ inputs.release_branch }}
    + {{ env.release_branch }}
    ```

    But the diff is now large enough that this isn't really that easy to see anymore. I can use another approach (like using the `outputs` directly instead of setting an environment variable, but would need to test it again.
- In order to support the above, I updated each job to `needs: [staging]` (along with whatever it already needed). I don't know if this is strictly necessary if the job is done after `staging` through other nodes, but adding the "dependency" relationship made sense to me, because these steps use `outputs` from `staging`. 
- Switch `dawidd6/action-download-artifact@v2` to `actions/download-artifact@v3` to download Go artifacts from the cache. We were using the nonstandard action only because `actions/download-artifact` does not support download artifacts from another workflow. Since we no longer have multiple workflows, we don't need it. The change is also now aligned to how we upload/download the Rust artifacts. 
- `build-native` is now called `build-rust`, because the former name didn't make sense to me, especially when it's in the same workflow as building other native things now too (i.e. Go)


arlyon (approved):
This is cool with me. I think we are reasonably confident in the release process and dry builds are still in tact so we can test this.

tknickman (approved):
Looks good, can you share some links to test builds?

mehulkar:
@tknickman seems like new actions don't show up in the list until they're merged into main, so I can't test them? Unless you know of another way https://github.com/vercel/turbo/actions

mehulkar:
Tested in personal fork here: https://github.com/mehulkar/turbo/actions/runs/4878717121/jobs/8704879555. Had to setup some secrets, etc, but it looks like its working now. Updated the PR description with the set of salient changes as well

gsoltis (approved):

PR #4707, opened 2023-04-26
error when concurrently writing files

sokra:
### Description

Go into an error state when a file is written from multiple sources

### DIFF
--- a/crates/turbo-tasks-fs/src/lib.rs
+++ b/crates/turbo-tasks-fs/src/lib.rs
@@ -36,7 +36,7 @@ use std::{
 };
 
 use anyhow::{anyhow, bail, Context, Result};
-use auto_hash_map::AutoMap;
+use auto_hash_map::{map::Entry, AutoMap};
 use bitflags::bitflags;
 use dunce::simplified;
 use glob::GlobVc;
@@ -53,7 +53,7 @@ use tokio::{
     io::{AsyncBufReadExt, AsyncReadExt, BufReader},
 };
 use turbo_tasks::{
-    mark_stateful,
+    get_invalidator, mark_stateful,
     primitives::{BoolVc, StringReadRef, StringVc},
     spawn_thread,
     trace::TraceRawVcs,
@@ -143,6 +143,12 @@ impl DiskWatcher {
     }
 }
 
+#[derive(Serialize, Deserialize)]
+enum WriterState {
+    Authorative,
+    InConflict,
+}
+
 #[turbo_tasks::value(cell = "new", eq = "manual")]
 pub struct DiskFileSystem {
     pub name: String,
@@ -154,6 +160,8 @@ pub struct DiskFileSystem {
     invalidator_map: Arc<InvalidatorMap>,
     #[turbo_tasks(debug_ignore, trace_ignore)]
     dir_invalidator_map: Arc<InvalidatorMap>,
+    #[turbo_tasks(debug_ignore, trace_ignore)]
+    authorative_write_map: Arc<Mutex<HashMap<PathBuf, AutoMap<Invalidator, WriterState>>>>,
     #[turbo_tasks(debug_ignore, trace_ignore)]
     #[serde(skip)]
     watcher: Arc<DiskWatcher>,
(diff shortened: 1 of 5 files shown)


crates/turbo-tasks-memory/src/task.rs
```diff
@@ -2819,6 +2819,20 @@ impl Task {
...
             .get_read_collectibles_task(trait_id, create_new)
     }
+
+    pub fn is_active(&self, backend: &MemoryBackend) -> bool {
+        if let TaskMetaStateReadGuard::Full(state) = self.state() {
+            let mut active = false;
+            for scope in state.scopes.iter() {
+                backend.with_scope(scope, |scope| {
+                    active = active || scope.state.lock().is_active();
+                })
```
alexkirsz, line 2829: I don't think we need to call 

```suggestion
                if backend.with_scope(scope, |scope| { scope.state.lock().is_active() }) {
                    return true;
                }
```

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -678,6 +687,51 @@ impl FileSystem for DiskFileSystem {
...
+            let mut has_own = false;
+            // Remove all inactive invalidators.
+            for (invalidator, state) in take(inner_map).into_iter() {
+                if invalidator == own_invalidator {
+                    has_own = true;
+                    new_map.insert(invalidator, state);
+                } else if invalidator.is_active() {
+                    new_map.insert(invalidator, state);
+                } else {
+                    invalidator.invalidate();
```
jridgewell, line 705: Do we need to invalidate? If this is inactive, couldn't we just drop it?

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -678,6 +687,51 @@ impl FileSystem for DiskFileSystem {
 
         let _lock = self.mutex_map.lock(full_path.clone()).await;
 
+        // Make sure that there is only one authorative writer at a time.
+        let conflict = {
+            let own_invalidator = get_invalidator();
+            let mut authorative_write_map = self.authorative_write_map.lock().unwrap();
+            let inner_map = authorative_write_map.entry(full_path.clone()).or_default();
+            let mut new_map = AutoMap::new();
```
jridgewell, line 695: I don't see a path for ever inserting a value into `new_map`. Yes, we're inserting values from the `inner_map`, but that's just the previous value of `new_map`, so we could never have inserted something into the `inner_map` to begin with. There's also a path if there's a conflict, but the map can never start with a conflict, so we don't ever enter that code either.

I think we're missing a `new_map.insert(own_invalidator, …)` right after this line, which means we don't really need to track `has_own` at all (it'll always have own).

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -155,6 +161,8 @@ pub struct DiskFileSystem {
     #[turbo_tasks(debug_ignore, trace_ignore)]
     dir_invalidator_map: Arc<InvalidatorMap>,
     #[turbo_tasks(debug_ignore, trace_ignore)]
+    authorative_write_map: Arc<Mutex<HashMap<PathBuf, AutoMap<Invalidator, WriterState>>>>,
```
jridgewell, line 164: Why does this need an inner mapping? Shouldn't this be achievable with just a `Invalidator`? Either you're the first writer, you're the first writer after the previous writer became inactive, or you're in conflict and we can early exit.

jridgewell (changes requested):

crates/turbo-tasks-fs/src/lib.rs
sokra: It might become active later and we want to recompute it in this case to make the conflict check again.

crates/turbo-tasks-fs/src/lib.rs
sokra: yeah, that's missing...

crates/turbo-tasks-fs/src/lib.rs
sokra: We want all write operations to report the conflict, including the first one. So in case of a conflict we need to recompute all write operations to report the conflict. But we only want to invalidate write operations that have not reported a conflict yet, since we would end up in a loop otherwise. So that's why we store the state for each operation and only invalidate the authoritive ones.

anthonyshew:

PR #4708, opened 2023-04-26
Make sure that we only hash the env pairs.

nathanhammond:
This corrects the hashing behavior for the global hash if there is an EnvMode specified. The struct now known as `GlobalHashableInputs` contains additional data that is not required to be present in the hash; particularly that `envVars` should be `env.EnvironmentVariablePairs`, not `env.DetailedMap`.

This didn't result in user-facing bugs, but it results in unnecessarily partitioned caches when using env modes.

We can't hash what is now known as `GlobalHashableInputs` directly, we must grab the components of it that we need, which this does for the (currently) two available hash formats.

### DIFF
--- a/cli/internal/run/global_hash.go
+++ b/cli/internal/run/global_hash.go
@@ -24,8 +24,8 @@ var _defaultEnvVars = []string{
 	"VERCEL_ANALYTICS_ID",
 }
 
-// GlobalHashable represents all the things that we use to create the global hash
-type GlobalHashable struct {
+// GlobalHashableInputs represents all the things that we use to create the global hash
+type GlobalHashableInputs struct {
 	globalFileHashMap    map[turbopath.AnchoredUnixPath]string
 	rootExternalDepsHash string
 	envVars              env.DetailedMap
@@ -35,9 +35,31 @@ type GlobalHashable struct {
 	envMode              util.EnvMode
 }
 
-// This exists because the global hash used to have different fields. Changing
-// to a new struct layout changes the global hash. We can remove this converter
-// when we are going to have to update the global hash for something else.
+type newGlobalHashable struct {
+	globalFileHashMap    map[turbopath.AnchoredUnixPath]string
+	rootExternalDepsHash string
+	envVars              env.EnvironmentVariablePairs
+	globalCacheKey       string
+	pipeline             fs.PristinePipeline
+	envVarPassthroughs   []string
+	envMode              util.EnvMode
+}
+
+// newGlobalHash is a transformation of GlobalHashableInputs.
+// It's used for the situations where we have an `EnvMode` specified
+// as that is not compatible with existing global hashes.
+func newGlobalHash(full GlobalHashableInputs) (string, error) {
+	return fs.HashObject(newGlobalHashable{
+		globalFileHashMap:    full.globalFileHashMap,
+		rootExternalDepsHash: full.rootExternalDepsHash,
+		envVars:              full.envVars.All.ToHashable(),
+		globalCacheKey:       full.globalCacheKey,
+		pipeline:             full.pipeline,
+		envVarPassthroughs:   full.envVarPassthroughs,
+		envMode:              full.envMode,
+	})
+}
+
 type oldGlobalHashable struct {
 	globalFileHashMap    map[turbopath.AnchoredUnixPath]string
 	rootExternalDepsHash string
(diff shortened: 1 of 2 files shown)


cli/internal/run/global_hash.go
```diff
@@ -24,8 +24,8 @@ var _defaultEnvVars = []string{
 	"VERCEL_ANALYTICS_ID",
 }
 
-// GlobalHashable represents all the things that we use to create the global hash
-type GlobalHashable struct {
+// GlobalHashableData represents all the things that we use to create the global hash
+type GlobalHashableData struct {
 	globalFileHashMap    map[turbopath.AnchoredUnixPath]string
 	rootExternalDepsHash string
 	envVars              env.DetailedMap
```
nathanhammond, line 31: This right here is actually the cause of the problem.

cli/internal/run/global_hash.go
```diff
@@ -46,38 +59,60 @@ type oldGlobalHashable struct {
...
 }
 
+// newGlobalHash is a transformation of GlobalHashableData.
+// It's used for the situations where we have an EnvMode.
+// We will collapse "old" and "new" at strict mode stabilization.
+func newGlobalHash(full GlobalHashableData) (string, error) {
+	return fs.HashObject(newGlobalHashable{
+		globalFileHashMap:    full.globalFileHashMap,
+		rootExternalDepsHash: full.rootExternalDepsHash,
+		envVars:              full.envVars.All.ToHashable(),
```
nathanhammond, line 69: And this line is the fix. Everything else is just plumbing.

nathanhammond (commented):
Reviewer's Guide!

nathanhammond:
@mehulkar

> Makes sense to me to limit the data needed for hashing, but why does including `DetailedMap` partition the cache unnecessarily?
> 
> `DetailedMap` is made of `All` and `BySource`. `BySource` is essentially a filtered view of `All`. I would expect them both to change together.

1. `All` and `BySource` are covariant, they will change together. But `BySource`'s `Explicit` and `Matching` are independently variant with regards to `All`. Adding a key to `Explicit` that already appears in `Matching` results in a different hash.

Since for globalHash we only have `.*THASH.*` as a matcher here is an example of an unnecessary invalidation for somebody going through a deprecation workflow right now:
```json
{}
```

```json
{
  "globalEnv": ["SOME_THASH"]
}
```

2. Generally, serializing a random struct into the hash and hoping that it remains stable is a bad idea. It's a refactoring hazard. We would break the hash right now simply by adding any field to `DetailedMap` or `BySource`.

cli/internal/run/global_hash.go
```diff
@@ -35,6 +35,19 @@ type GlobalHashable struct {
 	envMode              util.EnvMode
 }
 
+// newGlobalHashable is a transformation of GlobalHashableData.
+// It's used for the situations where we have an EnvMode.
+// We will collapse "old" and "new" at strict mode stabilization.
```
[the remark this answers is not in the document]
nathanhammond: It really means "when we decide to stop branching". And that's the name of the project internally.

cli/internal/run/global_hash.go
```diff
@@ -35,6 +35,19 @@ type GlobalHashable struct {
 	envMode              util.EnvMode
 }
 
+// newGlobalHashable is a transformation of GlobalHashableData.
+// It's used for the situations where we have an EnvMode.
```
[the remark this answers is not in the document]
nathanhammond: (I changed all of these, take a look.)

cli/internal/run/global_hash.go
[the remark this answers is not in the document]
nathanhammond: (I changed all of these, take a look.)

cli/internal/run/global_hash.go
```diff
@@ -46,44 +68,52 @@ type oldGlobalHashable struct {
 	pipeline             fs.PristinePipeline
 }
 
-// calculateGlobalHashFromHashable returns a hash string from the globalHashable
-func calculateGlobalHashFromHashable(full GlobalHashable) (string, error) {
+// oldGlobalHash is a transformation of GlobalHashableInputs.
+// This exists because the existing global hashes are still usable
+// in some configurations that do not include a specified `EnvMode`.
+// We can remove this whenever we want to migrate users.
```
mehulkar, line 74: It might be useful to add a searchable sigil like "TODO: Invalidate global hash"  or whatever that's searrchabl. We're tracking this already in linear, but maybe in the codebase would be more useful? (not a blocker)

nathanhammond: The PR for this is already up where we drop the `pipeline`.

PR #4709, opened 2023-04-26
Allow the dev server socket to be reused immediately

alexkirsz:
### Description

This avoids running into "Address already in use (os error 48)" when restarting the dev server in quick succession.

### DIFF
--- a/crates/turbopack-dev-server/src/lib.rs
+++ b/crates/turbopack-dev-server/src/lib.rs
@@ -110,6 +110,8 @@ impl DevServer {
         // Allow the socket to be reused immediately after closing. This ensures that
         // the dev server can be restarted on the same address without a buffer time for
         // the OS to release the socket.
+        // https://docs.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse
+        #[cfg(not(windows))]
         let _ = socket.set_reuse_address(true);
         if matches!(addr, SocketAddr::V6(_)) {
             // When possible bind to v4 and v6, otherwise ignore the error
--- a/crates/turbopack-dev-server/src/lib.rs
+++ b/crates/turbopack-dev-server/src/lib.rs
@@ -107,6 +107,10 @@ impl DevServer {
         // real TCP listener, see if it bound, and get its bound address.
         let socket = Socket::new(Domain::for_address(addr), Type::STREAM, Some(Protocol::TCP))
             .context("unable to create socket")?;
+        // Allow the socket to be reused immediately after closing. This ensures that
+        // the dev server can be restarted on the same address without a buffer time for
+        // the OS to release the socket.
+        let _ = socket.set_reuse_address(true);
         if matches!(addr, SocketAddr::V6(_)) {
             // When possible bind to v4 and v6, otherwise ignore the error
             let _ = socket.set_only_v6(false);


crates/turbopack-dev-server/src/lib.rs
```diff
@@ -107,6 +107,10 @@ impl DevServer {
         // real TCP listener, see if it bound, and get its bound address.
         let socket = Socket::new(Domain::for_address(addr), Type::STREAM, Some(Protocol::TCP))
             .context("unable to create socket")?;
+        // Allow the socket to be reused immediately after closing. This ensures that
+        // the dev server can be restarted on the same address without a buffer time for
+        // the OS to release the socket.
+        let _ = socket.set_reuse_address(true);
```
ForsakenHarmony, line 113: This will also allow binding to the same port twice, no?

ForsakenHarmony: pretty sure this will break our `find_port` function
https://github.com/vercel/next.js/blob/3bdaed01ba9e1a86a16fd243e5bdb5e4894ed5aa/packages/next-swc/crates/next-dev/src/lib.rs#L173-L173

alexkirsz: That's `SO_REUSEPORT`.

ForsakenHarmony: > This indicates that futher calls to bind may allow reuse of local addresses. For IPv4 sockets this means that a socket may bind even when there's a socket already listening on this port

from the rustdoc

ForsakenHarmony: https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse#using-so_reuseaddr

also windows docs

ForsakenHarmony: it also seems like `SO_REUSEADDR` on systems that make the differentiation between it and `SO_REUSEPORT` only makes it possible to reuse the port on a different address (i.e. if you have 2 network interfaces)

https://stackoverflow.com/a/14388707

alexkirsz: Let's enable this only for linux and macos then?

alexkirsz: Or also add `SO_EXCLUSIVEADDRUSE` on Windows. But I don't think `socket2` implements that.

ForsakenHarmony: But even on Linux/macOS, it doesn't do what you want (or only does it implicitly)

alexkirsz: GPT-4 seems to agree?

> 
> Using SO_REUSEADDR on Windows can potentially introduce a security issue due to the way the option is implemented on the platform. The primary difference between the Windows and Unix/Linux implementations is that on Windows, SO_REUSEADDR allows any socket, even from a different process, to bind to the same address and port while the original socket is still active.
> 
> This behavior could lead to a security issue called a "port hijacking" attack. An attacker could create a new socket, set the SO_REUSEADDR option, and bind it to the same address and port as an existing service. This would enable the attacker to intercept incoming connections meant for the original service, potentially exposing sensitive data or allowing unauthorized access.
> 
> To mitigate this risk, you can use the SO_EXCLUSIVEADDRUSE socket option on Windows instead of SO_REUSEADDR. This option prevents any other socket from binding to the same address and port, even if the original socket is closed and in the TIME_WAIT state.
> 

> On Unix-based platforms (e.g., Linux, macOS), using SO_REUSEADDR generally does not pose the same security issues as on Windows. This is because on these platforms, the SO_REUSEADDR option allows binding to an address and port only if the original socket is in the TIME_WAIT state, i.e., after it has been closed. In contrast, on Windows, SO_REUSEADDR permits binding to an address and port even if the original socket is still active.
> 
> However, there are still potential security concerns associated with using SO_REUSEADDR on Unix-based platforms, mainly related to the TIME_WAIT state:
> 
> TIME_WAIT state hijacking: If an attacker can predict the sequence numbers and forge packets, they might be able to inject malicious data into a new connection that reuses the same address and port as a recently closed connection. This risk is low, as modern operating systems implement random sequence number generation and other security measures to prevent such attacks.
> 
> Resource exhaustion: An attacker could create a large number of connections to a server that uses SO_REUSEADDR and then close them, causing the server to use up resources handling the TIME_WAIT states. This could lead to a denial-of-service attack. To mitigate this risk, servers should implement rate limiting and other security measures.
> 
> It is essential to carefully consider the risks associated with using SO_REUSEADDR or other socket options, and to implement appropriate security measures to protect your application. For some applications, using SO_REUSEPORT might be a better alternative, as it allows multiple sockets to bind to the same address and port simultaneously without introducing the same security concerns as SO_REUSEADDR. Additionally, SO_REUSEPORT can provide load balancing and fault tolerance benefits.

ForsakenHarmony: I mean, we don't really care about security when running a dev server on some port

But if something else starts a dev server on a specific IP instead of `0.0.0.0` (i.e. 127.0.0.1 to not expose it to the local network) we will not get an error that the port is in use which is a bit unexpected

alexkirsz: GPT-4:

> On Linux and macOS, using SO_REUSEADDR does not allow binding to the same port on a different address (e.g., 127.0.0.1 instead of 0.0.0.0) if another process is already bound to that port. The primary purpose of SO_REUSEADDR on Unix-based platforms is to allow the reuse of an address and port that is in the TIME_WAIT state, i.e., after a socket has been closed but is still waiting for any lingering packets to be delivered.

That sounds like exactly what we want?

ForsakenHarmony: @alexkirsz can I get a proper source?

alexkirsz: I think the benefits of being able to restart the dev server quickly trumps the drawback of not getting an error in the case that there already is a different local address bound on a given port.

Do we agree that this should not break `find_port`, except on Windows?

ForsakenHarmony: and it also says linux now matches this behavior

I'm fine with merging this for macOS/Linux only, but we might have to update the port finding code, we've already had a bunch of bug reports because of it not reporting ports in use correctly / not finding a new port as expected in the past and this might make it worse

ForsakenHarmony: this may very well break `find_port` on macos and linux

sokra (approved):

crates/turbopack-dev-server/src/lib.rs
jridgewell: This seems to match the old behavior from before https://github.com/vercel/turbo/pull/4565: https://github.com/rust-lang/rust/blob/9c044d77a334609513f3b615e0763a40cc02424d/library/std/src/sys_common/net.rs#L399-L407

jridgewell (approved):

PR #4719, opened 2023-04-26
Port RecursiveCopy to Rust

gsoltis:
### Description

 - Ports `RecursiveCopy` to rust and wires it up.
 - Ports `CopyFile` to rust but only uses it internally within `recursive_copy`. It is available for later though.

### Testing Instructions

Ported `copy_file_test.go` to Rust, which includes the test for `RecursiveCopy`. To Go test also continues to pass against the rust implementation.
Existing `prune` integration tests pass

### Review Notes 
The commits are:
 1. Tweaks to Go to ensure we always have absolute system paths
 2. Rust implementation
 3. FFI wiring

It's possible I've missed specifying a new crate in some location, please let me know.

Note the new methods on `AbsoluteSystemPathBuf`. These are an example of the kinds of things that will let us spot and flag uses of things like `fs::Open` and `fs::Metadata` in other pieces of code. 

Also note that our existing `RecursiveCopy` has some weird behavior: symlinks to directories get turned into empty folders and broken symlinks are not copied at all. I've reproduced these quirks so that we can fix them on purpose rather than bury them in the porting effort.

### DIFF
--- a/crates/turbopath/src/absolute_system_path_buf.rs
+++ b/crates/turbopath/src/absolute_system_path_buf.rs
@@ -1,13 +1,23 @@
+#[cfg(not(windows))]
+use std::os::unix::fs::symlink as symlink_dir;
+#[cfg(not(windows))]
+use std::os::unix::fs::symlink as symlink_file;
+#[cfg(windows)]
+use std::os::windows::fs::{symlink_dir, symlink_file};
 use std::{
     borrow::Cow,
     ffi::OsStr,
     fmt,
+    fs::{self, Metadata},
+    io::{self, Write},
     path::{Components, Path, PathBuf},
 };
 
 use serde::Serialize;
 
-use crate::{AnchoredSystemPathBuf, IntoSystem, PathValidationError, RelativeSystemPathBuf};
+use crate::{
+    AnchoredSystemPathBuf, IntoSystem, PathError, PathValidationError, RelativeSystemPathBuf,
+};
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Default, Serialize)]
 pub struct AbsoluteSystemPathBuf(PathBuf);
@@ -143,6 +153,57 @@ impl AbsoluteSystemPathBuf {
         AbsoluteSystemPathBuf(self.0.join(path.as_path()))
     }
 
+    pub fn join_literal(&self, segment: &str) -> Self {
+        AbsoluteSystemPathBuf(self.0.join(Path::new(segment)))
+    }
+
+    pub fn ensure_dir(&self) -> Result<(), io::Error> {
+        if let Some(parent) = self.0.parent() {
+            fs::create_dir_all(parent)
+        } else {
+            Ok(())
+        }
+    }
+
+    pub fn remove(&self) -> Result<(), io::Error> {
+        fs::remove_file(self.0.as_path())
+    }
+
+    pub fn set_readonly(&self) -> Result<(), PathError> {
+        let mut perms = self.metadata()?.permissions();
+        perms.set_readonly(true);
+        fs::set_permissions(self.0.as_path(), perms)?;
+        Ok(())
+    }
+
+    pub fn is_readonly(&self) -> Result<bool, PathError> {
+        Ok(self.metadata()?.permissions().readonly())
+    }
+
+    pub fn symlink_to_file<P: AsRef<Path>>(&self, to: P) -> Result<(), PathError> {
+        let system_path = to.as_ref();
+        let system_path = system_path.into_system()?;
+        symlink_file(&system_path, &self.0.as_path())?;
+        Ok(())
+    }
+
+    pub fn symlink_to_dir<P: AsRef<Path>>(&self, to: P) -> Result<(), PathError> {
+        let system_path = to.as_ref();
+        let system_path = system_path.into_system()?;
+        symlink_dir(&system_path, &self.0.as_path())?;
+        Ok(())
+    }
+
+    pub fn read_symlink(&self) -> Result<PathBuf, io::Error> {
+        fs::read_link(self.0.as_path())
+    }
+
+    pub fn create_with_contents(&self, contents: &str) -> Result<(), io::Error> {
+        let mut f = fs::File::create(self.0.as_path())?;
+        write!(f, "{}", contents)?;
+        Ok(())
+    }
+
     pub fn to_str(&self) -> Result<&str, PathValidationError> {
         self.0
             .to_str()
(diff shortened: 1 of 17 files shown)


gsoltis:
Marked as draft while I sort out windows issues

crates/turborepo-fs/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+    builder.create(dir.as_path())?;
+    Ok(())
+}
+
+pub fn copy_file(from: &AbsoluteSystemPathBuf, to: &AbsoluteSystemPathBuf) -> Result<()> {
+    let metadata = from.metadata()?;
+    copy_file_with_type(from, metadata.file_type(), to)
+}
+
+fn copy_file_with_type(
```
arlyon, line 78: As long as we are going this primarily via FFI I think doing this strictly blocking is OK, but once we begin to use this on the rust side, it would be nice if we make the copy calls async so that we can yield the thread.

arlyon: That said, more of our files are probably very small, so maybe we can just leave it sync and throw it in a thread if needs-be.

arlyon (commented):
Some questions.

crates/turborepo-fs/src/lib.rs
gsoltis: I'm happy to move this to async post-port. For now I'm trying to maintain the existing behavior.

crates/turborepo-fs/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+    let src_metadata = src.metadata()?;
+    if src_metadata.is_dir() {
+        let walker = WalkDir::new(src.as_path()).follow_links(false);
+        for entry in walker.into_iter() {
+            match entry {
+                Err(e) => {
+                    if let Some(_) = e.io_error() {
+                        // Matches go behavior where we translate path errors
+                        // into skipping the path we're currently walking
+                        continue;
```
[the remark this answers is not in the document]
gsoltis: Oh, yes. Good idea!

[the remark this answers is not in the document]
gsoltis: Hmm, assuming you're referring to [filter_entry](https://docs.rs/walkdir/latest/walkdir/struct.FilterEntry.html#method.filter_entry), I don't think it works for this case. It only gets called with non-error entries, whereas we want to filter entries that encountered IO errors

crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -143,6 +151,65 @@ impl AbsoluteSystemPathBuf {
...
+
+        #[cfg(not(windows))]
+        return Ok(std::os::unix::fs::symlink(&system_path, self.0.as_path())?);
+    }
+
+    pub fn symlink_to_dir<P: AsRef<Path>>(&self, to: P) -> Result<(), PathError> {
+        let system_path = to.as_ref();
+        let system_path = system_path.into_system()?;
+        #[cfg(windows)]
+        return Ok(std::os::windows::fs::symlink_dir(to)?);
```
chris-olszewski, line 197: I think this is missing an argument: https://doc.rust-lang.org/std/os/windows/fs/fn.symlink_dir.html

crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -143,6 +151,65 @@ impl AbsoluteSystemPathBuf {
...
+        let system_path = to.as_ref();
+        let system_path = system_path.into_system()?;
+        #[cfg(windows)]
+        return Ok(std::os::windows::fs::symlink_file(
+            &system_path,
+            self.0.as_path(),
+        )?);
+
+        #[cfg(not(windows))]
+        return Ok(std::os::unix::fs::symlink(&system_path, self.0.as_path())?);
```
chris-olszewski, line 190: We could use named imports to hide the platform differences in this function.
```suggestion
        // This could be pulled to the module level
        #[cfg(windows)]
        use std::os::windows::fs::symlink_file as symlink_file;
        #[cfg(not(windows))]
        use std::os::unix::fs::symlink as symlink_file;
        
        return Ok(symlink_file(
            &system_path,
            self.0.as_path(),
        )?);
```

crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -143,6 +151,65 @@ impl AbsoluteSystemPathBuf {
...
+        let system_path = to.as_ref();
+        let system_path = system_path.into_system()?;
+        #[cfg(windows)]
+        return Ok(std::os::windows::fs::symlink_dir(to)?);
+
+        #[cfg(not(windows))]
+        return Ok(std::os::unix::fs::symlink(&system_path, self.0.as_path())?);
+    }
+
+    pub fn read_symlink(&self) -> Result<PathBuf, io::Error> {
```
chris-olszewski, line 203: Super nit, but some places in this file we use `Result<T, io::Error>` and in others we use `io::Result<T>`. It would be nice to standardize how we express this type. (I have a slight preference for the first option, but this a complete bikeshed discussion)

crates/turbopath/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+anyhow = { workspace = true }
```
chris-olszewski, line 10:
```suggestion
```

crates/turbopath/src/absolute_system_path_buf.rs
arlyon: Also prefer first type (didn't know io::Result existed!)

arlyon (approved):

gsoltis:
Rebased onto `main`

crates/turborepo-ffi/src/lib.rs
```diff
@@ -121,3 +122,26 @@ pub extern "C" fn previous_content(buffer: Buffer) -> Buffer {
...
+        Err(err) => {
+            let resp = proto::RecursiveCopyResponse {
+                error: Some(err.to_string()),
+            };
+            return resp.into();
+        }
+    };
+    let response = match turborepo_fs::recursive_copy(
+        &AbsoluteSystemPathBuf::new_unchecked(req.src),
+        &AbsoluteSystemPathBuf::new_unchecked(req.dst),
```
chris-olszewski, line 139: Since `RecursiveCopy` is only used in prune and that isn't exactly performance sensitive, I'd rather skip the addition of `new_unchecked` and either unwrap the `Result` type from `new` or something like `map_err(|e| e.to_string()`. I'm just worried  of possible `new_unchecked`

chris-olszewski (approved):

crates/turborepo-ffi/src/lib.rs
gsoltis: I'm making the change for this and removing `new_unchecked`. It's possible that it will be back at some point, in the Go version we have `AnchoredSystemPathFromUpstream` for places where we know some API returns an absolute path. But the FFI code is (hopefully) short-lived and (hopefully) not in a hot loop, so I think it's fine to do the check.

jridgewell (approved):

PR #4721, opened 2023-04-26
feat(turbopack-ecmascript): accept transform plugin in moduleoptioncontext

kwonoj:
### Description

- Relates with WEB-955, WEB-848.

This PR is a part of implementing WEB-848, supporting relay graphql transform in next.js. Historically we've been expanding `ModuleOptionContext` to have each own config (`enable_mdx`, `enable_emotion`...) but with new CustomTransformPlugin we can pass actual transforms instead, what WEB-955 aims.

PR introduces new config option `custom_ecma_transform_plugins` to the ModuleOptionContext to support it. `custom_ecma_transform_plugins` is a struct have two fields, `before` / `after` which allows to specify order of custom transforms if it should be invoked _before_ any core transform kicks in, or otherwise.

This effectively deprecates existing options, notably `custom_ecmascript_transforms` and `custom_ecmascript_app_transforms`. It is currently not actively being used (next-swc have 2 places utilizing `custom_ecmascript_transforms`), we can consolidate to the new option instead.

One another notable change is customtransformer's transform is now async fn. It is due to `TransformContext` have couple of `Vc` properties, that cannot be accessed in a sync transform fn. Refer counterpart next.js PR (https://github.com/vercel/next.js/pull/48899/files#diff-c0df3b5a390436575f35365774b078cb5303ef395b064b8992dfdd7dba8f3691) as ref.

### DIFF
--- a/crates/turbopack-ecmascript-plugins/src/transform/relay.rs
+++ b/crates/turbopack-ecmascript-plugins/src/transform/relay.rs
@@ -1,0 +1,72 @@
+use std::path::PathBuf;
+
+use async_trait::async_trait;
+use swc_core::{
+    common::{util::take::Take, FileName},
+    ecma::{
+        ast::{Module, Program},
+        visit::FoldWith,
+    },
+};
+use turbo_tasks_fs::FileSystem;
+use turbopack_ecmascript::{CustomTransformer, TransformContext};
+
+#[derive(Debug)]
+pub struct RelayTransformer {
+    config: swc_relay::Config,
+}
+
+impl RelayTransformer {
+    pub fn new(config: swc_relay::Config) -> Self {
+        Self { config }
+    }
+}
+
+#[async_trait]
+impl CustomTransformer for RelayTransformer {
+    async fn transform(
+        &self,
+        program: &mut Program,
+        ctx: &TransformContext<'_>,
+    ) -> Option<Program> {
+        let root = if let Ok(file_path) = ctx.file_path.await {
+            file_path
+                .fs
+                .root()
+                .await
+                .map_or(PathBuf::new(), |v| PathBuf::from(v.path.to_string()))
+        } else {
+            PathBuf::new()
+        };
+
+        let (root, config) = if self.config.artifact_directory.is_some() {
+            (root, None)
+        } else {
+            let config = swc_relay::Config {
+                artifact_directory: Some(PathBuf::from("__generated__")),
+                ..self.config
+            };
+            (
+                if root.as_os_str().is_empty() {
+                    PathBuf::from(".")
+                } else {
+                    root
+                },
+                Some(config),
+            )
+        };
+
+        let p = std::mem::replace(program, Program::Module(Module::dummy()));
+        *program = p.fold_with(&mut swc_relay::relay(
+            config.as_ref().unwrap_or_else(|| &self.config),
+            FileName::Real(PathBuf::from(ctx.file_name_str)),
+            root,
+            // [TODO]: pages_dir comes through next-swc-loader
+            // https://github.com/vercel/next.js/blob/ea472e8058faea8ebdab2ef6d3aab257a1f0d11c/packages/next/src/build/webpack-config.ts#L792
+            None,
+            Some(ctx.unresolved_mark),
+        ));
+
+        None
+    }
+}
--- a/crates/turbopack-ecmascript-plugins/src/transform/relay.rs
+++ b/crates/turbopack-ecmascript-plugins/src/transform/relay.rs
@@ -1,5 +1,6 @@
 use std::path::PathBuf;
 
+use anyhow::Result;
 use async_trait::async_trait;
 use swc_core::{
     common::{util::take::Take, FileName},
@@ -28,16 +29,14 @@ impl CustomTransformer for RelayTransformer {
         &self,
         program: &mut Program,
         ctx: &TransformContext<'_>,
-    ) -> Option<Program> {
-        let root = if let Ok(file_path) = ctx.file_path.await {
-            file_path
-                .fs
-                .root()
-                .await
-                .map_or(PathBuf::new(), |v| PathBuf::from(v.path.to_string()))
-        } else {
-            PathBuf::new()
-        };
+    ) -> Result<Option<Program>> {
+        let root = ctx
+            .file_path
+            .await?
+            .fs
+            .root()
+            .await
+            .map_or(PathBuf::new(), |v| PathBuf::from(v.path.to_string()));
 
         let (root, config) = if self.config.artifact_directory.is_some() {
             (root, None)
@@ -67,6 +66,6 @@ impl CustomTransformer for RelayTransformer {
             Some(ctx.unresolved_mark),
         ));
 
-        None
+        Ok(None)
     }
 }
--- a/crates/turbopack-ecmascript-plugins/src/transform/relay.rs
+++ b/crates/turbopack-ecmascript-plugins/src/transform/relay.rs
@@ -9,7 +9,6 @@ use swc_core::{
         visit::FoldWith,
     },
 };
-use turbo_tasks_fs::FileSystem;
 use turbopack_ecmascript::{CustomTransformer, TransformContext};
 
 #[derive(Debug)]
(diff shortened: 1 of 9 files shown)


crates/turbopack-ecmascript-plugins/src/transform/relay.rs
```diff
@@ -0,0 +1,65 @@
...
+    ecma::{
+        ast::{Module, Program},
+        visit::FoldWith,
+    },
+};
+use turbo_tasks_fs::FileSystem;
+use turbopack_ecmascript::{CustomTransformer, TransformContext};
+
+#[derive(Debug)]
+pub struct RelayTransformer {
```
kwonoj, line 15: Decide to move this transform into turbopack side, as it is possible turbopack standalone might need this transform as well.

crates/turbopack-ecmascript-plugins/src/transform/relay.rs
```diff
@@ -0,0 +1,72 @@
...
+    ) -> Option<Program> {
+        let root = if let Ok(file_path) = ctx.file_path.await {
+            file_path
+                .fs
+                .root()
+                .await
+                .map_or(PathBuf::new(), |v| PathBuf::from(v.path.to_string()))
+        } else {
+            PathBuf::new()
+        };
```
sokra, line 40: `transform` should return a `Result` and errors should be propagated instead of ignored.

crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -77,8 +78,10 @@ pub enum EcmascriptInputTransform {
 
 /// The CustomTransformer trait allows you to implement your own custom SWC
 /// transformer to run over all ECMAScript files imported in the graph.
+#[async_trait]
 pub trait CustomTransformer: Debug {
-    fn transform(&self, program: &mut Program, ctx: &TransformContext<'_>) -> Option<Program>;
+    async fn transform(&self, program: &mut Program, ctx: &TransformContext<'_>)
+        -> Option<Program>;
```
sokra, line 84: Why does it return an `Option<Program>` when it already has a `&mut Program`?

It looks like that `Option<Program>` is not really needed.

crates/turbopack/src/module_options/module_options_context.rs
```diff
@@ -156,6 +156,18 @@ impl Default for StyledComponentsTransformConfigVc {
...
+/// Configuration options for the custom ecma transform to be applied.
+#[turbo_tasks::value(shared)]
+#[derive(Default, Clone)]
+pub struct CustomEcmaTransformPlugins {
+    /// List of plugins to be applied before the main transform.
+    /// Transform will be applied in the order of the list.
+    pub before: Vec<TransformPluginVc>,
+    /// List of plugins to be applied after the main transform.
+    /// Transform will be applied in the order of the list.
+    pub after: Vec<TransformPluginVc>,
```
sokra, line 168: before and after is a bit vague.

```suggestion
    /// List of plugins to be applied before the main transform.
    /// Transform will be applied in the order of the list.
    pub source_transforms: Vec<TransformPluginVc>,
    /// List of plugins to be applied after the main transform.
    /// Transform will be applied in the order of the list.
    pub output_transforms: Vec<TransformPluginVc>,
```

crates/turbopack/src/module_options/module_options_context.rs
```diff
@@ -156,6 +156,18 @@ impl Default for StyledComponentsTransformConfigVc {
     }
 }
 
+/// Configuration options for the custom ecma transform to be applied.
+#[turbo_tasks::value(shared)]
+#[derive(Default, Clone)]
+pub struct CustomEcmaTransformPlugins {
```
sokra, line 162:
```suggestion
pub struct CustomEcmascriptTransformPlugins {
```

We use `Ecmascript` instead of `Ecma` everywhere

crates/turbopack-ecmascript-plugins/src/transform/relay.rs
```diff
@@ -0,0 +1,72 @@
...
+        &self,
+        program: &mut Program,
+        ctx: &TransformContext<'_>,
+    ) -> Option<Program> {
+        let root = if let Ok(file_path) = ctx.file_path.await {
+            file_path
+                .fs
+                .root()
+                .await
+                .map_or(PathBuf::new(), |v| PathBuf::from(v.path.to_string()))
```
sokra, line 37: That's always an empty PathBuf. The root of the fs has always `path = ""`

sokra (changes requested):

crates/turbopack-ecmascript-plugins/src/transform/relay.rs
kwonoj: Huh, didn't know. Is there a way to get an asbolute root in that case?

kwonoj: if we can get a full absolute root, https://github.com/vercel/turbo/pull/4721#discussion_r1182870967 can be done as well.

kwonoj: Updated https://github.com/vercel/turbo/pull/4721/files#diff-236047cd15903b7c05afa8b8857302452a4c02acb690a5ca98e5f372c4898aa3R32

crates/turbopack/src/module_options/mod.rs
```diff
@@ -93,7 +94,30 @@ impl ModuleOptionsVc {
...
+
+        let (before_transform_plugins, after_transform_plugins) =
+            if let Some(transform_plugins) = custom_ecma_transform_plugins {
+                let transform_plugins = transform_plugins.await?;
+                (
+                    transform_plugins
+                        .source_transforms
+                        .iter()
+                        .cloned()
+                        .map(|plugin| EcmascriptInputTransform::Plugin(plugin))
```
jridgewell, line 106: Nit:
```suggestion
                        .map(EcmascriptInputTransform::Plugin)
```

crates/turbopack/src/module_options/mod.rs
```diff
@@ -93,7 +94,30 @@ impl ModuleOptionsVc {
...
+                        .collect(),
+                    transform_plugins
+                        .output_transforms
+                        .iter()
+                        .cloned()
+                        .map(|plugin| EcmascriptInputTransform::Plugin(plugin))
```
jridgewell, line 112: Nit:
```suggestion
                        .map(EcmascriptInputTransform::Plugin)
```

crates/turbopack/src/module_options/module_options_context.rs
```diff
@@ -187,11 +199,15 @@ pub struct ModuleOptionsContext {
...
     #[serde(default)]
     pub preset_env_versions: Option<EnvironmentVc>,
+    #[deprecated(note = "use custom_ecma_transform_plugins instead")]
     #[serde(default)]
     pub custom_ecmascript_app_transforms: Vec<EcmascriptInputTransform>,
+    #[deprecated(note = "use custom_ecma_transform_plugins instead")]
     #[serde(default)]
     pub custom_ecmascript_transforms: Vec<EcmascriptInputTransform>,
     #[serde(default)]
+    pub custom_ecma_transform_plugins: Option<CustomEcmascriptTransformPluginsVc>,
```
jridgewell, line 209: Nit: Just call these `custom_transform_plugins`?

crates/turbopack-ecmascript-plugins/src/transform/relay.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+#[async_trait]
+impl CustomTransformer for RelayTransformer {
+    async fn transform(
+        &self,
+        program: &mut Program,
+        ctx: &TransformContext<'_>,
+    ) -> Result<Option<Program>> {
+        // If user supplied artifactory_directory, it should be resolvable already.
```
jridgewell, line 32:
```suggestion
        // If user supplied artifact_directory, it should be resolvable already.
```

jridgewell (approved):

crates/turbopack/src/module_options/module_options_context.rs
kwonoj: I expect we'll have things like css_transform later, based on experiences with swc's transform plugin.

PR #4723, opened 2023-04-27
Add ability to run a single integration test

mehulkar:


turborepo-tests/integration/turbo.json
```diff
@@ -7,8 +7,12 @@
       "dependsOn": ["^topo"]
     },
     "test": {
-      "dependsOn": ["cli#build", "topo"],
-      "env": ["GO_TAG"]
+      "dependsOn": ["cli#build", "topo", "test:setup"]
```
chris-olszewski, line 10: Is the dropping of `"env": ["GO_TAG"]` intentional?

chris-olszewski (approved):
Thanks for this!

turborepo-tests/integration/turbo.json
mehulkar: I thought it belonged in `cli#build` instead, and now that this dependency exist, that would cover our case 
https://github.com/vercel/turbo/blob/main/cli/turbo.json#L5-L6

mehulkar:
Failed unit tests were due to time out, merging

PR #4724, opened 2023-04-27
Remove `THASH`

nathanhammond:
After 1.9 as a deprecation period (and having heard no user feedback as to being an integral part of their workflow) this removes `THASH`.

Related: #4526


mehulkar (approved):
Do we need to wait for 1.10 to remove? cc @gsoltis

nathanhammond:
@mehulkar Absolutely. Removing features in `patch` releases is not good.

PR #4730, opened 2023-04-27
Delete unused git_go file

mehulkar:


chris-olszewski (approved):
We need to change `git_rust.go` to no longer have build constraints at the top so we still get a `go` package even when `GO_TAG=go`

PR #4731, opened 2023-04-27
Revert daemon port

gsoltis:
### Description

Reverts #4306 while we debug issues found via canary. Unfortunately the PR does not revert cleanly.


gsoltis:
The Go lint check failure is for restoring old code that previously did not pass lint. I'm inclined to leave it rather than edit code that's being restored

arlyon (approved):

gsoltis:
Merging. Outstanding failure is Go lint, (see above), and the benchmark of non-turbopack bundlers is still running, but unlikely to be affected here.

PR #4734, opened 2023-04-27
Add gitbranch and sha into spaces payload

mehulkar:
This is meant to be a pretty naive implementation for dogfooding. It doesn't affect _most_ current users of Turbo, so I haven't implemented any corner cases (e.g. dirty state and non-git repos, etc). I will incrementally add that in as the need arises.


cli/internal/scm/scm.go
```diff
@@ -51,3 +53,23 @@ func FromInRepo(repoRoot turbopath.AbsoluteSystemPath) (SCM, error) {
 	}
 	return newFallback(dotGitDir.Dir())
 }
+
+// GetCurrentBranch returns the current branch
+func GetCurrentBranch() string {
+	command := []string{"rev-parse", "--abbrev-ref", "HEAD"}
+	out, err := exec.Command("git", command...).CombinedOutput()
```
nathanhammond, line 60: What's cwd?

cli/internal/scm/scm.go
```diff
@@ -51,3 +53,23 @@ func FromInRepo(repoRoot turbopath.AbsoluteSystemPath) (SCM, error) {
 	}
 	return newFallback(dotGitDir.Dir())
 }
+
+// GetCurrentBranch returns the current branch
+func GetCurrentBranch() string {
+	command := []string{"rev-parse", "--abbrev-ref", "HEAD"}
```
nathanhammond, line 59: This can pull a lot of surprising information with it.

Detached HEAD:
```
cd turbo
git checkout 2b0654e0e19da62547f968fb95bb2d0fe982ca1c
git rev-parse --abbrev-ref HEAD
# HEAD
```

Here is a branch and a tag with the same name:
```
~/repos/triage/4734 (master)$ git log
commit 91ae59e50f922accfca0b052db6446ab54e3a051 (HEAD -> master, tag: master, tag: foo)
Author: Nathan Hammond <nathan.hammond@vercel.com>
Date:   Fri Apr 28 15:18:20 2023 +0800

    initial commit
~/repos/triage/4734 (master)$ git rev-parse --abbrev-ref HEAD
# heads/master
```

You may want to look into `--symbolic-full-name` to get consistent behavior.

Bonus points:
- Does this work with worktrees?
- What about submodules?

nathanhammond (commented):
Mostly a bunch of comments about `git rev-parse`. We need to be defensive of its output.

cli/internal/runsummary/spaces.go
```diff
@@ -94,3 +102,34 @@ func newSpacesTaskPayload(taskSummary *TaskSummary) *spacesTask {
 		Logs:         string(taskSummary.GetLogs()),
 	}
 }
+
+// getStateOfRepo returns the sha and branch when in a git repo
+// Otherwise it should return empty strings right now.
+// We my add handling of other scms and non-git tracking in the future.
+func getStateOfRepo() (string, string) {
```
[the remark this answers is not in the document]
NicholasLYang: I really wish Rust had named return types. Returning tuples is such an easy way to introduce a bug

cli/internal/runsummary/spaces.go
```diff
@@ -94,3 +102,34 @@ func newSpacesTaskPayload(taskSummary *TaskSummary) *spacesTask {
...
+	var sha string
+	var branch string
+
+	// If we're in CI, try to get the values we need from environment variables
+	if ci.IsCi() {
+		vendor := ci.Info()
+		branchVarName := vendor.BranchEnvVar
+		shaVarName := vendor.ShaEnvVar
+		// Get the values of the vars
+		vars := env.FromKeys(allEnvVars, []string{shaVarName, branchVarName})
```
NicholasLYang, line 121: Why are we doing this instead of indexing `allEnvVars`?

cli/internal/scm/scm.go
mehulkar: Switched to `git branch --show-current`, any concerns with that?

cli/internal/runsummary/spaces.go
```diff
@@ -94,3 +102,34 @@ func newSpacesTaskPayload(taskSummary *TaskSummary) *spacesTask {
...
+func getStateOfRepo() (string, string) {
+	allEnvVars := env.GetEnvMap()
+
+	var sha string
+	var branch string
+
+	// If we're in CI, try to get the values we need from environment variables
+	if ci.IsCi() {
+		vendor := ci.Info()
+		branchVarName := vendor.BranchEnvVar
```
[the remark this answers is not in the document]
mehulkar: agreed, i'm not sure how well this will go. I think the idea is to match vercel preview branches, but they're working with fundamentally different data (github) than us (git), so....

cli/internal/runsummary/spaces.go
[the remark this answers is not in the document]
mehulkar: hah well `getStateOfRepo` certainly conflicts with `RepoState` (which is now ExecutionState?) from rust land :)

I didn't want to say `git`, because I wanted to convey that we want this to handle non-git and other vcs scenarios also. But maybe that's YAGNI, i'll take your suggestion

cli/internal/runsummary/spaces.go
mehulkar: i like to make things more complicated than they are, apparently. thanks for calling me out 😃

cli/internal/scm/scm.go
nathanhammond: Nope: https://github.com/git/git/blob/master/builtin/branch.c#L500-L513

cli/internal/runsummary/run_summary.go
```diff
@@ -105,6 +106,7 @@ func NewRunSummary(
 			EnvMode:           globalEnvMode,
 			Tasks:             []*TaskSummary{},
 			GlobalHashSummary: globalHashSummary,
+			GitState:          getGitState(repoRoot),
```
mehulkar, line 109: one call out here is that this happens before runs. i chose to do it at the start, in case there are some git state modifying behaviors during the run (unlikely and likely unintentional), but we should try to capture up front)

cli/internal/scm/scm.go
```diff
@@ -51,3 +54,27 @@ func FromInRepo(repoRoot turbopath.AbsoluteSystemPath) (SCM, error) {
 	}
 	return newFallback(dotGitDir.Dir())
 }
+
+// GetCurrentBranch returns the current branch
+func GetCurrentBranch(dir string) string {
```
nathanhammond, line 59:
```suggestion
func GetCurrentBranch(dir turbopath.AbsoluteSystemPath) string {
```

cli/internal/scm/scm.go
```diff
@@ -51,3 +54,27 @@ func FromInRepo(repoRoot turbopath.AbsoluteSystemPath) (SCM, error) {
...
+
+	out, err := cmd.Output()
+	if err != nil {
+		return ""
+	}
+	return strings.TrimRight(string(out), "\n")
+}
+
+// GetCurrentSha returns the current SHA
+func GetCurrentSha(dir string) string {
```
nathanhammond, line 71:
```suggestion
func GetCurrentSha(dir turbopath.AbsoluteSystemPath) string {
```

cli/internal/scm/scm.go
```diff
@@ -51,3 +54,27 @@ func FromInRepo(repoRoot turbopath.AbsoluteSystemPath) (SCM, error) {
...
+	cmd := exec.Command("git", []string{"rev-parse", "HEAD"}...)
+	cmd.Dir = dir
```
nathanhammond, line 73:
```suggestion
	cmd.Dir = dir.ToString()
```

cli/internal/scm/scm.go
```diff
@@ -51,3 +54,27 @@ func FromInRepo(repoRoot turbopath.AbsoluteSystemPath) (SCM, error) {
...
+
+	out, err := cmd.Output()
+	if err != nil {
+		return ""
+	}
+	return strings.TrimRight(string(out), "\n")
```
nathanhammond, line 79:
```suggestion
	return strings.TrimRight(string(out), "\r\n")
```

cli/internal/runsummary/spaces.go
```diff
@@ -19,11 +19,11 @@ type spacesRunPayload struct {
 	Command        string `json:"command,omitempty"`        // the thing that kicked off the turbo run
 	RepositoryPath string `json:"repositoryPath,omitempty"` // where the command was invoked from
 	Context        string `json:"context,omitempty"`        // the host on which this Run was executed (e.g. Github Action, Vercel, etc)
+	GitBranch      string `json:"gitBranch"`
+	GitSha         string `json:"gitSha"`
```
nathanhammond, line 23: The prefix here says that we should maybe consider pushing this into an object with properties. That could make it forward-compatible:

```js
{
  scm: {
    type: "git",
    branch: "foo",
    sha: "C0FFEE"
  }
}
```

We do it this way in run summary, so it's a _bit_ weird to depart from that.

cli/internal/runsummary/repo_summary.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	dirString := dir.ToString()
+	// Otherwise fallback to using `git`
+	if gitstate.Branch == "" {
+		gitstate.Branch = scm.GetCurrentBranch(dirString)
+	}
+	if gitstate.Sha == "" {
+		gitstate.Sha = scm.GetCurrentSha(dirString)
+	}
```
nathanhammond, line 37: I guess it is theoretically possible that we:

1. match CI,
2. these values are `""`
3. we ask `git` and it gives us a different answer.

But, since these are used to drive the actual CI behavior I don't find that likely. I think we should probably wrap this in an `else`.

```suggestion
	var vendor Vendor;

	if ci.IsCi() {
		vendor := ci.Info()
	}

	if vendor != nil {
		gitstate.Sha = allEnvVars[vendor.ShaEnvVar]
		gitstate.Branch = allEnvVars[vendor.BranchEnvVar]
	} else {
		// Otherwise fallback to using `git`
		gitstate.Branch = scm.GetCurrentBranch(dirString)
		gitstate.Sha = scm.GetCurrentSha(dirString)		
	}
```

(The question is "which is likely to produce a worse outcome in a failure mode" and this is my hypothesis.)

cli/internal/runsummary/run_summary.go
nathanhammond: This doesn't really bother me a ton.

nathanhammond (changes requested):

cli/internal/runsummary/spaces.go
mehulkar: This is just where the API is right now. I agree we could change it but there are other departures from current run summary structure also, so I don't want to tug on this thread right now

cli/internal/runsummary/repo_summary.go
mehulkar: I made the fallback in CI for a couple reasons

- our vendor definitions may have typos or incomplete (it is incomplete now)
- local runs which are trying to emulate a CI system but only partially setting the env vars

I agree the failure mode of env vars being set to `""` in a real CI system is unlikely.

I can still change this if you feel strongly about it, but I think I'm still leaning towards leaving this as is

cli/internal/scm/scm.go
```diff
@@ -51,3 +54,27 @@ func FromInRepo(repoRoot turbopath.AbsoluteSystemPath) (SCM, error) {
...
+// GetCurrentBranch returns the current branch
+func GetCurrentBranch(dir string) string {
+	cmd := exec.Command("git", []string{"branch", "--show-current"}...)
+	cmd.Dir = dir
+
+	out, err := cmd.Output()
+	if err != nil {
+		return ""
+	}
+	return strings.TrimRight(string(out), "\n")
```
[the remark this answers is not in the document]
mehulkar: Added a unit test (because those are running in windows already), and it passes

cli/internal/scm/scm.go
mehulkar: Will continue this thread here https://github.com/vercel/turbo/pull/4734#discussion_r1182523783

cli/internal/scm/scm.go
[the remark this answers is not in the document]
nathanhammond: ahh, this eventually calls stdio `puts` which is just `\n`—even in MSVC! `git` being C it has as _lot_ of Unix behavior even on Windows.

cli/internal/scm/testdata/myrepo/foo
nathanhammond, line 1: Maybe `rm`? Can't tell how it is used. Alternatively rename?

nathanhammond (approved):

cli/internal/scm/scm_test.go
```diff
@@ -15,8 +15,8 @@ func TestGetCurrentBranchMain(t *testing.T) {
 	testDir := getTestDir(t, "myrepo")
 
 	// Setup git
-	gitCommand(t, testDir, []string{"config", "--global", "user.email", "turbo@vercel.com"})
-	gitCommand(t, testDir, []string{"config", "--global", "user.name", "Turbobot"})
+	gitCommand(t, testDir, []string{"config", "--local", "user.email", "turbo@vercel.com"})
```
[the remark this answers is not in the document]
nathanhammond: Oh, they're new. 🤦‍♂️

cli/internal/runsummary/repo_summary.go
nathanhammond: I don't have a strong enough opinion here to push it either direction. This can go either way.

cli/internal/scm/scm_test.go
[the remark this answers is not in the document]
mehulkar: I started with `--global`, because on Windows and Linux CI matrix, `git commit` spouted a bit message asking me to set the user and recommended this. But then when I ran the tests locally, it (obviously) also set _my_ global config, ruining all future commits of my real work. 

So I opted to make it local instead. Interestingly:

- the `--local` flag did not work in CI (on the same windows/linux CI matrix boxes), so I had to drop the flag altogether
- the macOS CI job doesn't require setting the user/email. Not sure why that is

cli/internal/scm/testdata/myrepo/foo
mehulkar: I could do a `.keep` file to stay with convention, but I think I needed something in order to have an initial commit

PR #4738, opened 2023-04-28
improve commonjs esm interop and node ESM

sokra:
### Description

* generate a default export when import faux modules
* detect specified module type and follow that
* emit errors/warnings when using non-matching syntax

next.js test case: https://github.com/vercel/next.js/pull/48940

### DIFF
--- a/crates/turbopack-ecmascript/src/references/esm/export.rs
+++ b/crates/turbopack-ecmascript/src/references/esm/export.rs
@@ -69,8 +69,9 @@ async fn expand_star_exports(root_asset: EcmascriptChunkPlaceableVc) -> Result<E
                 category: StringVc::cell("analyze".to_string()),
                 message: StringVc::cell(format!(
                     "export * used with module {} which has no exports\nTypescript only: Did you \
-                     want to export only types with `export type {{ ... }} from \"...\"`?",
-                    // TODO recommend export type * from "..." once https://github.com/microsoft/TypeScript/issues/37238 is implemented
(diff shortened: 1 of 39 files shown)


crates/turbopack-ecmascript/src/references/esm/export.rs
```diff
@@ -69,8 +69,9 @@ async fn expand_star_exports(root_asset: EcmascriptChunkPlaceableVc) -> Result<E
                 category: StringVc::cell("analyze".to_string()),
                 message: StringVc::cell(format!(
                     "export * used with module {} which has no exports\nTypescript only: Did you \
-                     want to export only types with `export type {{ ... }} from \"...\"`?",
-                    // TODO recommend export type * from "..." once https://github.com/microsoft/TypeScript/issues/37238 is implemented
+                     want to export only types with `export type * from \"...\"`?\nNote: Using \
+                     `export type` is more efficient than `export *` as it won't emit any runtime \
+                     code.",
```
alexkirsz, line 74: We should have a way to add suggestions to issues. This `Note` could be formatted in such a way that it's clearly a suggestion.

alexkirsz (approved):

crates/turbopack-ecmascript/src/references/esm/export.rs
sokra: Yep. Initially I planned to add this as `sub_issue` with `IssueSeverity::Suggestion`

PR #4739, opened 2023-04-28
Stabilize --env-mode

nathanhammond:
This PR updates the documentation and adds a codemod for stabilizing environment variables.


mehulkar (commented):
This has been in my queue for a while, but leaving a comment and requesting that you request a re-review when it's rebased/ready (after #4788 is merged presumably). Wanted to make sure I communicated that I'm aware of this since it's been sitting open and in my queue forever

mehulkar (approved):

PR #4752, opened 2023-04-28
port(turborepo): Run stub

NicholasLYang:
### Description

This is more a proof of concept than an actual implementation. Attempts to write the beginning of run in what's hopefully semi-idiomatic Rust, with all of the helper functions stubbed out.

### Testing Instructions

The stub is gated behind a rust feature flag `run-stub`, so to test the code add `--features run-stub` to your cargo command


crates/turborepo-lib/src/manager.rs
```diff
@@ -0,0 +1,10 @@
+#[derive(Debug)]
+pub struct Manager {
```
gsoltis, line 2: Is this the child process manager? Can you add a comment?

crates/turborepo-lib/src/opts.rs
```diff
@@ -0,0 +1,154 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    cli::{Command, DryRunMode, EnvMode, LogPrefix, RunArgs},
+    daemon::{DaemonClient, DaemonConnector},
+    Args,
+};
+
+#[derive(Debug)]
+pub struct Opts<'a> {
```
gsoltis, line 11: Do we care about the opts having a lifetime? Would it make it easier if they owned their data? It's not a ton of data...

crates/turborepo-lib/src/run/context.rs
```diff
@@ -0,0 +1,41 @@
+use std::rc::Rc;
+
+use anyhow::Result;
+use turbopath::AbsoluteSystemPathBuf;
+
+use crate::{package_json::PackageJson, run::graph::WorkspaceCatalog};
+
+pub struct Context {
```
gsoltis, line 8: Can we name this `PackageGraph`? Or something like that? The `Context` name from Go is not great.

crates/turborepo-lib/src/run/graph.rs
```diff
@@ -0,0 +1,59 @@
...
+        &self,
+        _workspace_name: &str,
+        _is_single_package: bool,
+    ) -> Result<TurboJson> {
+        // TODO
+        Ok(TurboJson::default())
+    }
+}
+
+#[derive(Default)]
```
gsoltis, line 55: Here, and elsewhere, I'm assuming we intend to remove the `Default` derivations in the future?

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+
+#[derive(Debug)]
+pub struct Run<'a> {
+    base: &'a CommandBase,
+    opts: Opts<'a>,
+    processes: Manager,
+}
+
+impl<'a> Run<'a> {
+    pub fn new(base: &'a CommandBase, opts: Opts<'a>) -> Self {
```
gsoltis, line 28: Would it make more sense for the `Run` to own the `CommandBase`? Do we continue using it after executing a command?

crates/turborepo-lib/src/run/pipeline.rs
```diff
@@ -0,0 +1,100 @@
+use std::collections::{HashMap, HashSet};
+
+use serde::{Deserialize, Serialize};
+
+pub type Pipeline = HashMap<String, BookkeepingTaskDefinition>;
+
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct BookkeepingTaskDefinition {
```
gsoltis, line 8: cc @mehulkar 

Is there a path to dropping the `Bookkeeping` in a rust world? Post-port? Or something we can do as part of the porting effort?

crates/turborepo-lib/src/run/task_id.rs
```diff
@@ -0,0 +1,47 @@
+pub const TASK_DELIMITER: &str = "#";
+pub const ROOT_PKG_NAME: &str = "//";
+
+pub fn get_task_id(pkg_name: impl std::fmt::Display, target: &str) -> String {
```
gsoltis, line 4: I'm 99% certain that task id doesn't need to be a string. Can we try a simple struct?

gsoltis (changes requested):
This is a great start, and I think is already highlighting some areas for improvement. 

Let's do the renames, and maybe the task id bit, then merge.

Can you also add a stub test that just "executes" a run?

gsoltis (changes requested):
This is a great start, and I think is already highlighting some areas for improvement. 

Let's do the renames, and maybe the task id bit, then merge.

Can you also add a stub test that just "executes" a run?

crates/turborepo-lib/src/run/pipeline.rs
mehulkar: It's main goal was to assist in merging fields. If Rust has a better way to distinguish between when a field is missing vs when it's the "empty value", we should be able to drop it. The [integration tests covering this behavior][1] are pretty good, so I'd just say "try it" 

[1]: https://github.com/vercel/turbo/tree/71cb29adf6b4b05a0a08ccb8fd1cf561a97f2670/turborepo-tests/integration/tests/composable_config

crates/turborepo-lib/src/run/graph.rs
NicholasLYang: yeah definitely. Mostly for boilerplate

crates/turborepo-lib/src/run/mod.rs
NicholasLYang: I've thought about having all the commands be methods of CommandBase, that way we can have all of the relevant resources be owned by CommandBase

crates/turborepo-lib/src/run/pipeline.rs
NicholasLYang: Yep! We can do `Option<T>` for those

crates/turborepo-lib/src/run/task_id.rs
NicholasLYang: What do you mean by simple struct here?

crates/turborepo-lib/src/opts.rs
NicholasLYang: Right now it can go either way. I think in the future I'd like a single `'run` lifetime that's tied to either `CommandBase` or a `Run` struct. And that can own stuff like Args that then in turn get transformed to `Opts`. For now I don't think we have to make that call until this is used.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+        cli::{Command, RunArgs},
+        commands::CommandBase,
+        get_version,
+        run::Run,
+        ui::UI,
+        Args,
+    };
+
+    #[tokio::test]
+    async fn test_run() -> Result<()> {
```
NicholasLYang, line 130: This is not an integration test because I'd need some way to plumb rust features through integration tests.

crates/turborepo-lib/src/run/task_id.rs
gsoltis: Thought about this a bit, and I think "simple" might not be the right word. And we maybe shouldn't block on this. I'm thinking of something like:
```rust
pub enum TaskId {
  SinglePackage { task: String },
  Monorepo {
    package: Option<String>,
    task: String
  }
}
```

gsoltis (approved):
Can you add a comment re: `Manager` and then let's merge + iterate

gsoltis:
Merged as #5099

PR #4754, opened 2023-04-28
Match TS's extends resolution algorithm

### LINKED ISSUE #2682: [WEB-963] [turbopack] tsconfig extends field doesn't resolve

1.6.3

Yarn v1

Windows

### Describe the Bug

Turbopack fails on a nextjs app when there is an `extends` field in the `tsconfig.json`. Error says `extends field doesn't resolve correctly`

### Expected Behavior

Be able to extend local tsconfig files.

### To Reproduce

create a package that exports tsconfig files. import it on a next 13 app and in the tsconfig extend it. run `next dev --turbo`

### Reproduction Repo

_No response_

<sub>[WEB-963](https://linear.app/vercel/issue/WEB-963/[turbopack]-tsconfig-extends-field-doesnt-resolve)</sub>

jridgewell:
### Description

Implements TS's `extends` [resolution algorithm](https://github.com/microsoft/TypeScript/blob/611a912d/src/compiler/commandLineParser.ts#L3294-L3326). In particular, there are a few cases that need to be properly handled:

- Relative paths: if the file doesn't exist and the path doesn't end in `.json`, try again with `.json` extension
- Windows paths: same thing
- Posix paths: same thing
- `.` and `..` (but not `./` nor `../` because they're relative paths): join with `/tsconfig.json` before looking up
- `foo` module import: join with `/tsconfig.json` before looking up

### Testing Instructions

Snapshot tests

Fixes WEB-963
fix #2682


crates/turbopack-ecmascript/src/typescript/resolve.rs
```diff
@@ -74,45 +73,98 @@ pub async fn read_tsconfigs(
...
+    extends: &str,
+    resolve_options: ResolveOptionsVc,
+) -> Result<AssetOptionVc> {
+    let context = tsconfig.ident().path().parent();
+    let request = RequestVc::parse_string(extends.to_string());
+
+    // TS's resolution is weird, and has special behavior for different import
+    // types.
+    let extends = match &*request.await? {
+        // TS has special behvaior for "rooted" paths (absolute paths):
```
alexkirsz, line 118:
```suggestion
        // TS has special behavior for "rooted" paths (absolute paths):
```

crates/turbopack-ecmascript/src/typescript/resolve.rs
```diff
@@ -74,45 +73,98 @@ pub async fn read_tsconfigs(
...
+        // https://github.com/microsoft/TypeScript/blob/611a912d/src/compiler/commandLineParser.ts#L3303-L3313
+        Request::Windows { path: Pattern::Constant(path) } |
+        // Server relative is treated as absolute
+        Request::ServerRelative { path: Pattern::Constant(path) } => {
+            return resolve_extends_rooted_or_relative(context, request, resolve_options, path).await;
+        }
+
+        // TS has special behvaior for (explicitly) './' and '../', but not '.' nor '..':
```
alexkirsz, line 126:
```suggestion
        // TS has special behavior for (explicitly) './' and '../', but not '.' nor '..':
```

alexkirsz (approved):

crates/turbopack-ecmascript/src/typescript/resolve.rs
```diff
@@ -74,45 +73,98 @@ pub async fn read_tsconfigs(
...
+        }
+        // All other types are treated as module imports, and joined with
+        // "tsconfig.json". This includes "relative" imports like '.' and '..'.
+        _ => {
+            format!("{extends}/tsconfig.json")
+        }
+    };
+
+    let request = RequestVc::parse_string(extends);
+    Ok(resolve(context, request, resolve_options).first_asset())
```
sokra, line 147:
```suggestion
    // There might be multiple alternatives like
    // "some/path/node_modules/xyz/abc.json" and "some/node_modules/xyz/abc.json".
    // We only want to use the first one.
    Ok(resolve(context, request, resolve_options).first_asset())
```

keeping the comment from the old version

sokra (approved):

Issue #2891, opened 2022-12-01
Execution of Build order in development mode


devGitAbhishek:
### What version of Turborepo are you using?

1.5.6

### What package manager are you using / does the bug impact?

Yarn v1

### What operating system are you using?

Mac

### Describe the Bug

- Consider the following repo setup:

**apps**
    UI [ depends on packages A,B,C,D,E ] _(This has been made with create react app)_
**packages** _(all packages have webpack config file)_
     A [ depends on E ]
     B [ depends on E ]
     C [ depends on E ]
     D [ depends on E & F ]
     E
     F

- Actual run order in development mode: yarn dev ( turbo run dev --continue )

1. package E
2. package F
3. package A,B,C,D

- For the turbo.json file we have included the following things in pipeline: 

 "dev": {
      "outputs": [".dist/**"]
    } 
In the above if we include depends on ["^dev"], then only package build is done which is E and after that it stops, there is no other script that gets executed.

- Note: 

1. yarn dev executes all dev jobs for app and packages.
2. For app: dev: **_react-scripts start_**
3. For all packages: **_webpack --watch_**

### Expected Behavior

- Required run order for dev:

package E
package F
package A,B,C,D
app UI

- When we execute production build where following is included in pipeline it work fine.
  "build": {
      "dependsOn": ["^build"],
      "outputs": [".dist/**"],
      "cache": true
    }

### To Reproduce

1. Make setup of turborepo, with all packages having webpack and babel setup.
2. Under apps have a app of CRA.
3. In package.json of app and packages define a dev command under script which actually does job of starting up app and building packages.
4. Run yarn dev from root of project

milovangudelj:
As the docs state you can't have task `A` depend on a long running task `B` such as a dev script executed in watch mode (e.g. with `--watch` flag). That's because task `A` will wait until task `B`, on which it depends, has finished and since dev tasks don't exit `A` will never run.

> Label a task as `persistent` if it is a long-running process, such as a dev server or `--watch` mode. Turbo will prevent other tasks from depending on persistent tasks. Without setting this config, if any other task depends on `dev`, it will never run, because `dev` never exits. With this option, `turbo` can warn you about an invalid configuration.

So this is not a bug but rather a missing feature. Btw, a watch mode is being considered as you can read in this issue (#986). See the last couple of comments made in Nov by Nathan Hammond.

devGitAbhishek:
[...]

PR #4755, opened 2023-04-28
fix: allow for workplace descriptors without a protocol

chris-olszewski:
### Description

Fixes the issue found by @quinnturner in [#2891](https://github.com/vercel/turbo/issues/2791#issuecomment-1526477128).

Crux of the issue was we were being too restrictive when tracking dependency protocols to properly infer them for later construction. We now allow for at most two descriptors in the protocol resolver: one without a protocol and one with a protocol. Another change was no longer adding all descriptors for a workspace and instead only including descriptors that come from other workspaces.

### Testing Instructions

New testing fixture contains a lockfile where one workspace depends on another via `*` and the other depends on it via `workspace:*`.


NicholasLYang (commented):
Just for my curiosity, why are we storing both `with` and `without` instead of normalizing to a single format?

chris-olszewski:
> Just for my curiosity, why are we storing both `with` and `without` instead of normalizing to a single format?

We need to differentiate them since Yarn encodes these descriptors into the lockfile (e.g. `interal-dep: workspace:*` and `internal-dep: *` reference the same workspace, but we need to track how the workspace is referenced). We could store just the protocol/lack of, but then we would need to reconstruct the range string which either requires an allocation or being clever.

crates/turborepo-lockfiles/src/berry/mod.rs
```diff
@@ -266,11 +266,10 @@ impl<'a> BerryLockfile<'a> {
...
 
-                if let Some(descriptors) = reverse_lookup.get(locator) {
-                    for descriptor in descriptors {
-                        resolutions.insert((*descriptor).clone(), locator.clone());
-                    }
-                }
+                // Included workspaces will always have their locator listed as a descriptor.
+                // All other descriptors should show up in the other workspace package
+                // dependencies.
+                resolutions.insert(Descriptor::from(locator.clone()), locator.clone());
```
mehulkar, line 272:
```
"c@*, c@workspace:*, c@workspace:packages/c":
```

Am I understanding correctly that this whole line is the "descriptor" and the "locator" is the `packages/c` part?

mehulkar (approved):

NicholasLYang (approved):

crates/turborepo-lockfiles/src/berry/mod.rs
chris-olszewski: Kinda: `c@*`, `c@workspace:*`, and `c@workspace:packages/c` are all descriptors and then they all get concat'd to be a key in the YAML lockfile. `c@workspace:packages/c` is *also* a locator in this lockfile which appears in the `resolution` field of the lockfile entry.

PR #4766, opened 2023-05-01
no longer opt-in to kqueue on macos

arlyon:
kqueue performance is pretty awful, and fsevent is the default, so this PR removes that feature and uses fsevent on macOS instead.

This does not totally resolve the perf issues, but does at least prevent it from using 100% cpu for multiple minutes while it registers handlers. Smarter glob-watching and request timeouts are coming in a separate PR.


mehulkar (approved):
Fine with me as a drive-by review without having done any benchmarking myself!

gsoltis (approved):

gsoltis:
Force merging, since this is into a feature branch.

PR #4767, opened 2023-05-01
Stripped down unix path

gsoltis:
### Description

 - Experiment with a very stripped down `RelativeUnixPath` implementation
 - Test out using it in `git.rs`

### Testing Instructions

Existing tests in `git.rs`

### DIFF
--- a/crates/turbopath/src/relative_unix_path.rs
+++ b/crates/turbopath/src/relative_unix_path.rs
@@ -1,0 +1,25 @@
+use std::path::Path;
+
+use crate::{IntoSystem, PathError, PathValidationError, RelativeSystemPathBuf};
+
+pub struct RelativeUnixPath {
+    inner: Path,
+}
+
+impl RelativeUnixPath {
+    pub fn new<P: AsRef<Path>>(value: &P) -> Result<&Self, PathError> {
+        let path = value.as_ref();
+        if path.is_absolute() {
+            return Err(PathValidationError::NotRelative(path.to_owned()).into());
+        }
+        // copied from stdlib path.rs: relies on the representation of
+        // RelativeUnixPath being just a Path, the same way Path relies on
+        // just being an OsStr
+        Ok(unsafe { &*(path as *const Path as *const Self) })
+    }
+
+    pub fn to_system_path(&self) -> Result<RelativeSystemPathBuf, PathError> {
+        let system_path = self.inner.into_system()?;
+        Ok(RelativeSystemPathBuf::new_unchecked(system_path))
+    }
+}
--- a/crates/turbopath/src/relative_unix_path_buf.rs
+++ b/crates/turbopath/src/relative_unix_path_buf.rs
@@ -44,6 +44,8 @@ impl RelativeUnixPathBuf {
 mod tests {
     use super::*;
 
+    use std::path::Path;
+
     #[test]
     fn test_relative_unix_path_buf() {
         let path = RelativeUnixPathBuf::new(PathBuf::from("foo/bar")).unwrap();
@@ -68,6 +70,6 @@ mod tests {
     #[test]
     fn test_convert_from_windows_path() {
         let path = RelativeUnixPathBuf::new(PathBuf::from("foo\\bar")).unwrap();
-        assert_eq!(path.as_path(), Path::new("foo/bar"));
+        assert_eq!(path.0.as_path(), Path::new("foo/bar"));
     }
 }
--- a/crates/turbopath/src/relative_unix_path_buf.rs
+++ b/crates/turbopath/src/relative_unix_path_buf.rs
@@ -42,9 +42,9 @@ impl RelativeUnixPathBuf {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
-
     use std::path::Path;
+
+    use super::*;
 
     #[test]
     fn test_relative_unix_path_buf() {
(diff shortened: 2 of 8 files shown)


crates/turbopath/src/relative_unix_path.rs
```diff
@@ -0,0 +1,21 @@
+use std::path::PathBuf;
+
+use path_slash::PathBufExt;
+
+use crate::RelativeSystemPathBuf;
+
+pub struct RelativeUnixPath<'a> {
```
NicholasLYang, line 7: So the reason we only have `RelativeSystemPathBuf` and not `RelativeSystemPath` in turbopath is because in the process of creating the path, there is a chance we need to convert the slashes, which requires allocation. Since the `-Path` types are unsized and therefore not an allocation, we can't create them without throwing out automatic conversion.

To sum it up, if we want a `RelativeUnixPath` type, we either need to validate but not convert on construction, or we need to make this only constructible via a `RelativeUnixPathBuf`

crates/turbopath/src/relative_unix_path.rs
```diff
@@ -0,0 +1,21 @@
...
+    inner: &'a str,
+}
+
+impl<'a> From<&'a str> for RelativeUnixPath<'a> {
```
NicholasLYang, line 11: Yeah this is basically an unchecked construction, which IMO should be explicit and not the default behavior

crates/turborepo-scm/src/git.rs
```diff
@@ -102,9 +97,9 @@ fn add_files_from_stdout(
 ) {
     let stdout = String::from_utf8(stdout).unwrap();
     for line in stdout.lines() {
-        let path = Path::new(line);
+        let path: RelativeUnixPath = line.into();
```
NicholasLYang, line 100: Do we know that git2 always outputs unix paths? Even so, this should probably be explicit as an unsafe conversion.

NicholasLYang (changes requested):

crates/turbopath/src/relative_unix_path.rs
gsoltis: Agreed. Thinking about it more, we should not have this impl and it should be explicit.

crates/turborepo-scm/src/git.rs
gsoltis: I don't know about `git2`. The `git` cli always uses unix paths.

crates/turbopath/src/relative_unix_path.rs
NicholasLYang: Yeah we'll probably want both at some point, also because `&RelativeSystemPathBuf` is a little inefficient (similar to `&String` versus `&str`). If we want the full infrastructure, we'd have `RelativeSystemPathBuf::new` return a `Cow` that is either a `RelativeSystemPath` or a `RelativeSystemPathBuf` depending on whether we allocate or not. That'd be the most idiomatic Rust version. But I think for now we should stick with `RelativeSystemPathBuf`

gsoltis: I reworked `RelativeUnixPath` to be based on the `Path` implementation in the stdlib. Seems ok so far.

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -36,49 +33,11 @@ impl RelativeUnixPathBuf {
...
-
-    pub fn starts_with<P: AsRef<Path>>(&self, base: P) -> bool {
-        self.0.starts_with(base.as_ref())
-    }
-
-    pub fn ends_with<P: AsRef<Path>>(&self, child: P) -> bool {
-        self.0.ends_with(child.as_ref())
-    }
-
-    pub fn join<P: AsRef<Path>>(&self, path: P) -> RelativeUnixPathBuf {
```
NicholasLYang, line 61: What's the reason why we're deleting these?

NicholasLYang (approved):
Looks good to me. Did you mean to rename `PathValidationError` to `PathError`? Seems like the enum declaration isn't renamed

gsoltis:
Yeah, i need to fix the windows tests. PathError covers path validation,
IO, and the error from to_slash. It's not a rename, it's over the top of
the existing one.

gsoltis:
On Wed, May 3, 2023, 3:50 PM Nicholas Yang ***@***.***> wrote:

> ***@***.**** approved this pull request.
>
> Looks good to me. Did you mean to rename PathValidationError to PathError?
> Seems like the enum declaration isn't renamed
> ------------------------------
>
> In crates/turbopath/src/relative_unix_path_buf.rs
> <https://github.com/vercel/turbo/pull/4767#discussion_r1184378115>:
>
> > -
> -    pub fn parent(&self) -> Option<Self> {
> -        self.0
> -            .parent()
> -            .map(|p| RelativeUnixPathBuf(p.to_path_buf()))
> -    }
> -
> -    pub fn starts_with<P: AsRef<Path>>(&self, base: P) -> bool {
> -        self.0.starts_with(base.as_ref())
> -    }
> -
> -    pub fn ends_with<P: AsRef<Path>>(&self, child: P) -> bool {
> -        self.0.ends_with(child.as_ref())
> -    }
> -
> -    pub fn join<P: AsRef<Path>>(&self, path: P) -> RelativeUnixPathBuf {
>
> What's the reason why we're deleting these?
>
We're not using them. Especially with this type, i want us to be very sure
we need the pieces we implement. If we need these implementations we can
resurrect them.

PR #4772, opened 2023-05-02
Fix TSConfig extends pointing to node module

jridgewell:
In #4754, I screwed up the resolution logic for configs that reference another config inside a node module. In it, I incorrectly joined `/tsconfig.json` before performing any lookup, because I thought the joining was happening in the `combinePaths` in https://github.com/microsoft/TypeScript/blob/611a912d/src/compiler/commandLineParser.ts#L3315. But, that combine has no real affect, it immediately gets stripped in a call to `getDirectoryPath` in https://github.com/microsoft/TypeScript/blob/611a912d/src/compiler/moduleNameResolver.ts#LL1703C93-L1703.

Instead, it performs a basic `node_module` lookup using the input module name, and even tries to load the file referenced at that location (eg, it's a `extends: 'foo/tsconfig.json'`). If that fails, then it combines with an implied `/tsconfig` the same way a `/index` is appended when doing `require('./directory')`.

Fixes WEB-974

### DIFF
--- a/crates/turbopack-tests/tests/snapshot/typescript/tsconfig-extends-module-full-path/input/tsconfig.json
+++ b/crates/turbopack-tests/tests/snapshot/typescript/tsconfig-extends-module-full-path/input/tsconfig.json
@@ -1,0 +1,3 @@
+{
+  "extends": "tsconfig-mod/tsconfig.json"
+ }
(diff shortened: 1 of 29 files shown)


crates/turbopack-tests/tests/snapshot/typescript/tsconfig-extends-module-full-path/input/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "tsconfig-mod/tsconfig.json"
```
sokra, line 2: I think the test case could be improve by also testing omitting the `.json` extension and naming the file different than tsconfig

```suggestion
  "extends": "tsconfig-mod/config"
```

sokra (approved):

PR #4778, opened 2023-05-02
Add log rotation to the daemon server

arlyon:
### Description

This PR expands our tracing setup to allow the daemon to opt-in to rotated log files on top of the regular stdout logging. This output is non-blocking and uses the standard tracing format, respecting the env vars set by the command that runs the daemon.

### Testing Instructions

Run `turbo daemon` with `TURBO_LOG_VERBOSITY` set to `trace`, and open the path it prints for the log file to see that the logs are also being output there.


crates/turborepo-lib/src/tracing.rs
```diff
@@ -1,16 +1,107 @@
...
+
+impl TurboSubscriber {
+    /// Sets up the tracing subscriber, with a default stdout layer using the
+    /// TurboFormatter.
+    ///
+    /// ## Logging behaviour:
+    /// - If stdout is a terminal, we use ansi colors. Otherwise, we do not.
+    /// - If the `TURBO_LOG_VERBOSITY` env var is set, it will be used to set
+    ///   the verbosity level. Otherwise, the default is `WARN`. See the
+    ///   documentation on the RUST_LOG env var for syntax.
```
arlyon, line 55: This is critically a superset of debug, trace, warn, etc but also allows fun things like `TURBO_LOG_VERBOSITY=turborepo-lib=trace,tonic=debug,error`

crates/turborepo-lib/src/tracing.rs
```diff
@@ -1,16 +1,107 @@
...
+    /// - If the verbosity argument (usually detemined by a flag) is provided,
+    ///   it overrides the default global log level. This means it overrides the
+    ///   `TURBO_LOG_VERBOSITY` global setting, but not per-module settings.
```
arlyon, line 58: In the above example, if you had `TURBO_LOG_VERBOSITY=turborepo-lib=trace,tonic=debug,error` and `-vv` if would be equivalent to `TURBO_LOG_VERBOSITY=turborepo-lib=trace,tonic=debug,info`

crates/turborepo-lib/src/tracing.rs
```diff
@@ -1,16 +1,107 @@
...
+    ///
+    /// Returns a `reload::Handle` that can be used to reload the subscriber.
+    /// This allows us to register additional layers after setup, for example
+    /// when configuring logrotation in the daemon.
+    pub fn new_with_verbosity(verbosity: usize) -> Self {
```
arlyon, line 63: Generally, I will add some tests for this once we agree that this is the behaviour we want.

crates/turborepo-lib/src/cli.rs
```diff
@@ -552,7 +553,7 @@ pub async fn run(repo_state: Option<RepoState>) -> Result<Payload> {
 
             match command {
                 Some(command) => daemon::daemon_client(command, &base).await,
-                None => daemon::daemon_server(&base, idle_time).await,
+                None => daemon::daemon_server(&base, idle_time, logger).await,
```
NicholasLYang, line 556: I'm wondering if we should put `logger` into `CommandBase`, since we already have a UI object there.

crates/turborepo-lib/src/tracing.rs
```diff
@@ -1,16 +1,107 @@
...
+        let max_level = match verbosity {
+            0 => LevelFilter::WARN,
+            1 => LevelFilter::INFO,
+            2 => LevelFilter::DEBUG,
+            _ => LevelFilter::TRACE,
+        };
+
+        let stdout = fmt::layer()
+            .event_format(TurboFormatter::new_with_ansi(
+                std::io::stdout().is_terminal(),
```
NicholasLYang, line 73: We might want to adapt the existing logic in `Args::ui` so that we obey the --color/--no-color flags

arlyon: 👍

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -64,8 +65,12 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
...
 }
 
-pub async fn daemon_server(base: &CommandBase, idle_time: &String) -> Result<(), DaemonError> {
-    let log_file = {
+pub async fn daemon_server(
+    base: &CommandBase,
+    idle_time: &String,
+    logging: &TurboSubscriber,
+) -> Result<(), DaemonError> {
+    let (log_folder, log_file) = {
```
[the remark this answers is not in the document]
arlyon: No, I just didn't want the intermediary vars polluting the parent scope.

crates/turborepo-lib/src/tracing.rs
arlyon: I have opted to moving Args::ui to ShimArgs::ui instead, since it is a reasonable expectation that the shim should respect these flags.

NicholasLYang (approved):
One last thing, it might be good to document `TURBO_LOG_VERBOSITY` in `CONTRIBUTING.md`. Otherwise looks good to me!

mehulkar:
This is in my queue to review, but I think it needs to merge main in, the diff looks like it includes the daemon implementation as well?

PR #4788, opened 2023-05-03
Optional framework inference

nathanhammond:
This makes it possible to opt out of framework inference by specifying `--framework-inference=false` on the command line. This is designed to interact gracefully with `--env-mode=strict` as a separate-but-related knob to turn for control over environment variables.


nathanhammond (commented):
Reviewer's Guide! Use `?w=1`, commit-by-commit is valid.

cli/internal/runsummary/format_json.go
```diff
@@ -55,12 +55,13 @@ func (rsm *Meta) normalize() {
...
-	EnvMode           util.EnvMode       `json:"envMode"`
-	ExecutionSummary  *executionSummary  `json:"execution,omitempty"`
-	Tasks             []*TaskSummary     `json:"tasks"`
+	ID                 ksuid.KSUID        `json:"id"`
+	Version            string             `json:"version"`
+	TurboVersion       string             `json:"turboVersion"`
+	GlobalHashSummary  *GlobalHashSummary `json:"globalCacheInputs"`
+	Packages           []string           `json:"-"`
+	EnvMode            util.EnvMode       `json:"envMode"`
+	FrameworkInference bool               `json:"frameworkInference"`
```
mehulkar, line 64: Starting to think we're eventually just going to want something like:

```golang
	Config runOpts               `json:"config"`
```

cli/internal/runsummary/run_summary.go
```diff
@@ -56,14 +59,15 @@ type Meta struct {
...
-	EnvMode           util.EnvMode       `json:"envMode"`
-	ExecutionSummary  *executionSummary  `json:"execution,omitempty"`
-	Tasks             []*TaskSummary     `json:"tasks"`
+	ID                 ksuid.KSUID        `json:"id"`
+	Version            string             `json:"version"`
+	TurboVersion       string             `json:"turboVersion"`
+	GlobalHashSummary  *GlobalHashSummary `json:"globalCacheInputs"`
+	Packages           []string           `json:"packages"`
+	EnvMode            util.EnvMode       `json:"envMode"`
+	FrameworkInference bool               `json:"frameworkInference"`
```
mehulkar, line 68: Is there any chance we'll add more inference? If so, might makes sense to nest this as `inference.framework`?

cli/internal/util/run_opts.go
```diff
@@ -27,6 +27,8 @@ type RunOpts struct {
 	Parallel bool
 
 	EnvMode EnvMode
+	// Whether or not to perform framework inference.
```
mehulkar, line 30: nit, not blocking (take it or leave it)
```suggestion
	// Whether or not to infer the framework for each workspace.
```

docs/pages/repo/docs/reference/command-line-reference.mdx
```diff
@@ -293,6 +293,16 @@ turbo run build --global-deps=".env.*" --global-deps=".eslintrc" --global-deps="
...
 
+#### `--framework-inference`
+
+`type: bool`
+
+Specify whether or not to do framework inference for tasks. Defaults to `true`, can be configured to be `false` which skips framework inference for tasks. This disables [automatic environment variable inclusion](../core-concepts/caching#automatic-environment-variable-inclusion).
+
+```sh
+turbo run build --framework-inference=false
+```
```
mehulkar, line 304: Typically want this in a separate PR with a [docs: next release](https://github.com/vercel/turbo/labels/docs%3A%20next%20release) label

turborepo-tests/integration/tests/_fixtures/framework_inference/.gitignore
mehulkar, line 1: Kinda wish we could use one of the examples for this test instead of a pseudo next-app fixture.

cli/internal/run/dry_run.go
```diff
@@ -44,7 +44,11 @@ func DryRun(
 		}
 
 		if taskSummary.Framework == "" {
-			taskSummary.Framework = runsummary.MissingFrameworkLabel
+			if rs.Opts.runOpts.FrameworkInference {
+				taskSummary.Framework = runsummary.NoFrameworkDetected
+			} else {
+				taskSummary.Framework = runsummary.FrameworkDetectionSkipped
```
mehulkar, line 50: The only reason `MissingFrameworkLabel` is _only_ in dry run, is because it was already there and I didn't want to put it in Real Run summaries. I can't remember why I made that decision though. Maybe I thought that `framework: ""` was just as clear as a placeholder, so I didn't want a placeholder in a new place?

For *skipped* inference, maybe we _do_ want it to say "skipped" even in real run summaries, to differentiate from `framework: ""`?

Or you could make the case that since we have the `frameworkInference: false` at the top level of the summary, you could... uhhh... *infer* the meaning of `framework: ""` in each task? 

I guess I don't have a strong enough opinion, because while leaving this comment, I'm leaning towards adding the Missing _and_ skipped labels into real run summaries, whereas I explicitly decided _not_ to add it before. What are your thoughts?

mehulkar (commented):
Wanted a conversation on adding the skipped label into real run summaries before ✅ , but code looks good to me other than that.

cli/internal/runsummary/format_json.go
nathanhammond: Yeah, I've got another one coming in `dotEnv`. I'm not excited about making a backward-incompatible change to the object shape right now, though. That should come with a minor at least.

cli/internal/runsummary/run_summary.go
nathanhammond: After how this went I'm expecting it to be extremely limited, and likely opted-into using improved `extends` instead.

docs/pages/repo/docs/reference/command-line-reference.mdx
nathanhammond: This would go out in the next patch so I elected short-term disagreement to reduce process overhead. The commit is separate, though, for exactly this reason.

turborepo-tests/integration/tests/_fixtures/framework_inference/.gitignore
nathanhammond: Examples aren't intended to be static for testing certain code paths, though. They exist to solve user-facing problems and are expected to change independently from the tests. I see this more as a unit test written in a way that works for both Go and Rust.

We can and should do more "e2e" type testing against the examples, but nowhere else in `integration` do we have anything reach out to `../examples`.

cli/internal/run/dry_run.go
nathanhammond: There are two ways to derive this information:
- `.tasks[].framework`
- `.frameworkInference`

And the consequences of the configuration show up in run summary as:
- `.environmentVariables.inferred`

We don't _need_ it; it's derivable, and having an arbitrary string slug as part of an API feels bad. I added this to the PR _primarily_ because I didn't want people getting `NoFrameworkDetected` when it was actually skipped.

The other legit option is to _remove_ it from dry run, but that's an API change and should be held to at least 1.10.

nathanhammond: (In other words: rather than adding it to run summary, I'd pitch removing it from dry run.)

nathanhammond: In a post-Go world we could do `Option<Framework>` which would also easily serialize to `null` or `slug`, though since both `""` and `null` are falsey it's not a big difference. (though it may communicate more-clearly at a glance)

cli/internal/runsummary/format_json.go
mehulkar: It could be backwards compatible by duplicating whatever else you're thinking of in two places. There are other things that fall into that boat. That said, I'm not even sure we want all of runOpts, or in the form they exist today. So this is not a change we need.

docs/pages/repo/docs/reference/command-line-reference.mdx
mehulkar: Fine with me, but you and me both know that "short-term" is more than often a lie we tell ourselves 😛

cli/internal/run/dry_run.go
mehulkar: You can `omitempty` to differentiate too. 

I'm ok with the direction of removing from dry run, and in that case it's fine to only add this differentiation there.

mehulkar (approved):

cli/internal/run/dry_run.go
nathanhammond: I want the field to always be present though! lots of things blow up on missing keys.

docs/pages/repo/docs/reference/command-line-reference.mdx
nathanhammond: The right answer here is "release branches" but that's a lot of work.

cli/internal/runsummary/run_summary.go
nathanhammond: (We're already discussing switching the default on this flag in the next major, and then I'm going to propose removing the flag + feature in the major after that. But better `extends` is a prerequisite for that.)

PR #4792, opened 2023-05-03
chore(docs): update nextra

tknickman:
### Description

Needed to fix the missing copy code button on the docs, ended up updating Nextra and Next while I was in there. 

### Testing Instructions

Checkout the preview!


arlyon (approved):

.npmrc
```diff
@@ -1,2 +1,2 @@
 auto-install-peers = true
-hoist = false
+hoist-pattern[]=vfile-message
```
tknickman, line 2: `rehype-katex` was added to `nextra` in 2.0.2 to support LaTeX which brings in `hast-util-from-html` which has a dependency issue. 

I sent them a PR here: https://github.com/syntax-tree/hast-util-from-html/pull/4, but need to patch it on our end for now.

docs/next.config.js
```diff
@@ -1,9 +1,8 @@
 const { withSentryConfig } = require("@sentry/nextjs");
 const withNextra = require("nextra")({
   theme: "nextra-theme-docs",
-  themeConfig: "./theme.config.js",
-  unstable_flexsearch: true,
-  unstable_staticImage: true,
+  themeConfig: "./theme.config.tsx",
+  defaultShowCopyCode: true,
```
tknickman, line 5: I don't know why this isn't set by default, have to check on that one (cc @shuding)

docs/pages/blog/_meta.json
```diff
@@ -1,6 +1,5 @@
 {
   "*": {
-    "//": "Hide all the links from the sidebar or navbar, and disable ToC and Sidebar for these pages.",
```
tknickman, line 3: New nextra isn't a fan of this key in meta (jsonc support wen)

docs/theme.config.tsx
```diff
@@ -9,12 +9,9 @@ import { Discord, Github } from "./components/Social";
 
 const SITE_ROOT = "https://turbo.build";
 
-/**
- * @type {import('nextra-theme-docs').DocsThemeConfig}
- */
-const theme = {
+const config: DocsThemeConfig = {
```
tknickman, line 12: Thank god we have real types here now!

docs/next.config.js
shuding: Will make that the default!

PR #4800, opened 2023-05-03
chore: add underlying error to lockfile error messages

chris-olszewski:
### Description

Noticed that I didn't include the underlying error in some of the lockfile error variants. This PR adds underlying errors to the error message to all variants where applicable.

### Testing Instructions

`rustc`

### DIFF
--- a/crates/turborepo-ffi/src/lockfile.rs
+++ b/crates/turborepo-ffi/src/lockfile.rs
@@ -23,7 +23,7 @@ impl From<Package> for proto::LockfilePackage {
 enum Error {
     #[error("error performing lockfile operation: {0}")]
     Lockfile(#[from] turborepo_lockfiles::Error),
-    #[error("error decoding protobuf")]
+    #[error("error decoding protobuf: {0}")]
     Protobuf(#[from] prost::DecodeError),
     #[error(transparent)]
     BerryParse(#[from] turborepo_lockfiles::BerryError),
--- a/crates/turborepo-lockfiles/src/berry/mod.rs
+++ b/crates/turborepo-lockfiles/src/berry/mod.rs
@@ -21,13 +21,13 @@ use super::Lockfile;
 
 #[derive(Debug, Error)]
 pub enum Error {
-    #[error("unable to parse")]
+    #[error("unable to parse yaml: {0}")]
     Parse(#[from] serde_yaml::Error),
-    #[error("unable to parse")]
+    #[error("unable to parse identifier: {0}")]
     Identifiers(#[from] identifiers::Error),
     #[error("unable to find original package in patch locator {0}")]
     PatchMissingOriginalLocator(Locator<'static>),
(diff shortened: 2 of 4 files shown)


tknickman (approved):

crates/turborepo-ffi/src/lockfile.rs
```diff
@@ -23,7 +23,7 @@ impl From<Package> for proto::LockfilePackage {
 enum Error {
     #[error("error performing lockfile operation: {0}")]
     Lockfile(#[from] turborepo_lockfiles::Error),
-    #[error("error decoding protobuf")]
+    #[error("error decoding protobuf: {0}")]
```
mehulkar, line 26: Where would I begin to look up how this is working? I'd expect this diff to contain the actual value in addition to (what I assume is) the placeholder `{0}`. I'm guessing that means `[error(...)]` (macro??) is doing something fancy under the hood. But I don't know how to search for `error`  🤣 

(Trying to learn Rust passively alongside reading the book)

mehulkar (approved):
dumb learning question, but otherwise 🚢

crates/turborepo-ffi/src/lockfile.rs
chris-olszewski: So it already included the value! The `Error` enum can be viewed as just:
```
enum Error {
    Lockfile(turborepo_lockfiles::Error),
    Protobuf(prost::DecodeError),
    BerryParse(turborepo_lockfiles::BerryError),
    UnsupportedPackageManager(proto::PackageManager),
}
```
All of the `#[error(...)]` and `#[from]` macros generate some code for us that makes this enum more ergonomic.
 - The `#[error("error decoding protobuf: {0}")]` line is used as a formatting string when deriving a [`Display`](https://doc.rust-lang.org/std/fmt/trait.Display.html) implementation for the enum. To be exact it generates:
`write!( __formatter, "error decoding protobuf: {field__0}", field__0 = 0.as_display())`
- The `#[from]` macro generates a [`From`](https://doc.rust-lang.org/std/convert/trait.From.html) implementation for us that converts the protobuf error into our enum. This is especially useful when combined with the [`?` syntax](https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator) as it automatically attempts to convert to the correct error type.
```
impl std::convert::From<prost::DecodeError> for Error {
    #[allow(deprecated)]
    fn from(source: prost::DecodeError) -> Self {
        Error::Protobuf { 0: source }
    }
}
```

The easiest way to actually understand what a macro does is to open some code up that uses it and then use rust-analyzer's [expand-macro-recursively](https://rust-analyzer.github.io/manual.html#expand-macro-recursively) command to see exactly what it's doing. (You can also try reading the [macro source](https://docs.rs/thiserror-impl/1.0.40/src/thiserror_impl/expand.rs.html#195), but this takes way more time IMO)

PR #4816, opened 2023-05-04
Prefactor package hashing

nathanhammond:
We currently do file hashing for packages (and fallbacks) in _lots_ of different places, making it _very_ difficult to trace through or (more-importantly) modify the set of files whose hashes we need to include.

This PR makes no material changes, each commit passes all of our tests. At the end of it all of the hashing logic ends up in just one spot.


cli/internal/taskhash/taskhash.go
```diff
@@ -62,44 +62,13 @@ func NewTracker(rootNode string, globalHash string, pipeline fs.Pipeline) *Track
...
-	}
-}
-
-// packageFileHashKey is a hashable representation of a packageFileSpec.
-type packageFileHashKey string
-
-// hashes the inputs for a packageTask
-func (pfs packageFileSpec) ToKey() packageFileHashKey {
-	sort.Strings(pfs.inputs)
-	return packageFileHashKey(fmt.Sprintf("%v#%v", pfs.pkg, strings.Join(pfs.inputs, "!")))
```
nathanhammond, line 86: This _probably_ can't go wrong, but we already have a unique identifier, the taskID. This was to enable multiple tasks in the same package with the same specified inputs to not have to rerun file hashing.

It's an optimization, but not actually the ideal one. Now that everything calls into `GetFileHashes` that could just maintain a set of all files it has seen and apply it across all packages, all inputs.

I do not add this feature in this PR.

cli/internal/taskhash/taskhash.go
```diff
@@ -62,44 +62,13 @@ func NewTracker(rootNode string, globalHash string, pipeline fs.Pipeline) *Track
...
-}
-
-func (pfs *packageFileSpec) getHashObject(pkg *fs.PackageJSON, repoRoot turbopath.AbsoluteSystemPath) (map[turbopath.AnchoredUnixPath]string, error) {
-	return hashing.GetPackageFileHashes(repoRoot, pkg.Dir, pfs.inputs)
-}
-
-func (pfs *packageFileSpec) hash(hashObject map[turbopath.AnchoredUnixPath]string) (string, error) {
-	hashOfFiles, otherErr := fs.HashObject(hashObject)
```
nathanhammond, line 114: `fs.HashObject` is actually infallible, in spite of the signature.

From `xxhash.Write`:

> `Write` adds more data to `d`. It always returns `len(b), nil`.

cli/internal/taskhash/taskhash.go
```diff
@@ -119,8 +88,9 @@ func (th *Tracker) CalculateFileHashes(
 		if taskID == th.rootNode {
 			continue
 		}
-		pkgName, _ := util.GetPackageTaskFromId(taskID)
-		if pkgName == th.rootNode {
+
+		packageName, _ := util.GetPackageTaskFromId(taskID)
```
nathanhammond, line 92: We're not character constrained, so clear names that match structs are good.

cli/internal/taskhash/taskhash.go
```diff
@@ -129,45 +99,51 @@ func (th *Tracker) CalculateFileHashes(
...
+			for packageFileHashInputs := range hashQueue {
+				pkg, ok := workspaceInfos.PackageJSONs[packageFileHashInputs.packageName]
 				if !ok {
-					return fmt.Errorf("cannot find package %v", packageFileSpec.pkg)
+					return fmt.Errorf("cannot find package %v", packageFileHashInputs.packageName)
 				}
-				hashObject, err := packageFileSpec.getHashObject(pkg, repoRoot)
+
+				// Get the hashes of each file, keyed by the path.
+				hashObject, err := hashing.GetPackageFileHashes(repoRoot, pkg.Dir, packageFileHashInputs.taskDefinition.Inputs)
```
nathanhammond, line 125: Intentionally keeps the same signature that everywhere else does and doesn't redirect it so that we don't accidentally end up in the "N ways to do things" situation again.

cli/internal/taskhash/taskhash.go
```diff
@@ -129,45 +99,51 @@ func (th *Tracker) CalculateFileHashes(
...
 					return err
 				}
+
+				// Save off the hash information, keyed by package task.
 				th.mu.Lock()
-				pfsKey := packageFileSpec.ToKey()
-				hashes[pfsKey] = hash
-				hashObjects[pfsKey] = hashObject
+				hashes[packageFileHashInputs.taskID] = hash
+				hashObjects[packageFileHashInputs.taskID] = hashObject
```
nathanhammond, line 139: `taskID` is a _much_ better key, making the whole thing understandable.

cli/internal/taskhash/taskhash.go
```diff
@@ -129,45 +99,51 @@ func (th *Tracker) CalculateFileHashes(
...
+
+				// Get the hashes of each file, keyed by the path.
+				hashObject, err := hashing.GetPackageFileHashes(repoRoot, pkg.Dir, packageFileHashInputs.taskDefinition.Inputs)
 				if err != nil {
 					return err
 				}
-				hash, err := packageFileSpec.hash(hashObject)
+
+				// Get the combined hash of all the files.
+				hash, err := fs.HashObject(hashObject)
```
nathanhammond, line 131: No indirection where behavior can be modified spookily at a distance. Optimizing for clarity in hash construction.

cli/internal/taskhash/taskhash.go
```diff
@@ -272,12 +248,9 @@ func (th *Tracker) calculateDependencyHashes(dependencySet dag.Set) ([]string, e
...
 // first.
 func (th *Tracker) CalculateTaskHash(packageTask *nodes.PackageTask, dependencySet dag.Set, logger hclog.Logger, args []string, useOldTaskHashable bool) (string, error) {
-	pfs := specFromPackageTask(packageTask)
-	pkgFileHashKey := pfs.ToKey()
-
-	hashOfFiles, ok := th.packageInputsHashes[pkgFileHashKey]
+	hashOfFiles, ok := th.packageInputsHashes[packageTask.TaskID]
 	if !ok {
-		return "", fmt.Errorf("cannot find package-file hash for %v", pkgFileHashKey)
+		return "", fmt.Errorf("cannot find package-file hash for %v", packageTask.TaskID)
```
nathanhammond, line 253: This error message is now infinitely better.

nathanhammond (commented):
Reviewer's guide!

There were two goals here:
1. Force all file hashing into a single straight-through code path.
2. Have the `packageFileHashInputs` available for modification based upon `TaskDefinition` more-generally.

Along the way this fixed two bugs; one of which entirely broke manual hashing in a couple scenarios.

Review it commit-by-commit.

cli/internal/taskhash/taskhash.go
mehulkar: yesss. this tripped me up a lot when i was working on this code, thank you

cli/internal/taskhash/taskhash.go
mehulkar: I'm surprised you didn't say "rounds-to-infinity" in true nathan style 😉 😛

mehulkar (approved):
This is great, thanks for the step by step and the review notes on each commit. Couple comments:

1. The last few commits tripped me up a little because the PR says there are "no material changes", but the error propagation _does_ seem to have changed, (in that it fixes a real bug where a fallback hashing strategy was not being used?)
2. There's [a bug](https://linear.app/vercel/issue/TURBO-1021/manual-hashing-inputs-does-not-include-relative-paths) where the manual hashing fallback doesn't seem to work on relative paths, which I ran into because [`git hash-object` fails on a large amount of files](https://linear.app/vercel/issue/TURBO-1022/git-hashing-fails-with-large-amounts-of-files). Neither of these tickets are relevant to this PR except to say: I feel more confident about fixing those bugs after this PR, so thank you!

cli/internal/taskhash/taskhash.go
gsoltis: I think this is going to result in rehashing the same package multiple times if there is more than one task in that package.

For instance, if `test` depends on `build`, and neither specifies `inputs`, this will double the amount of hashing that we do.

gsoltis: Ok, I doubled checked what I wrote and it looks like there's a bug where it's not doing what we expect. This change _would_ result in rehashing the same package multiple times if we were already properly deduping sets of files to hash, but we are not. 

The reason we are not currently properly deduping is because we are adding the _addresses_ of `packageFileSpec` instances to a set, which are all going to be unique, even if the actual values are equivalent. 

The original author (me) probably should've noticed this given the effort to produce a key for a map later on, and knowing that the set is implemented via a map under the hood. I'm going to file an issue to fix this post-port, since it will be much easier to do the struct value equality in Rust.

gsoltis (approved):

cli/internal/hashing/package_deps_hash.go
```diff
@@ -224,12 +224,28 @@ func getPackageFileHashesFromInputs(rootPath turbopath.AbsoluteSystemPath, packa
...
-// GetPackageDeps Builds an object containing git hashes for the files under the specified `packagePath` folder.
-func GetPackageDeps(rootPath turbopath.AbsoluteSystemPath, p *PackageDepsOptions) (map[turbopath.AnchoredUnixPath]string, error) {
-	if len(p.InputPatterns) == 0 {
-		return getPackageFileHashesFromGitIndex(rootPath, p.PackagePath)
+// GetPackageFileHashes Builds an object containing git hashes for the files under the specified `packagePath` folder.
+func GetPackageFileHashes(rootPath turbopath.AbsoluteSystemPath, packagePath turbopath.AnchoredSystemPath, inputs []string) map[turbopath.AnchoredUnixPath]string {
+	if len(inputs) == 0 {
+		result, err := getPackageFileHashesFromGitIndex(rootPath, packagePath)
+		if err != nil {
+			result, err := getPackageFileHashesFromProcessingGitIgnore(rootPath, packagePath, nil)
```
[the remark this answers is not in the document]
nathanhammond: I don't know yet! I copy/pasted the implementations. Those three code paths should probably be just one.

cli/internal/hashing/package_deps_hash.go
```diff
@@ -253,7 +253,7 @@ func GetPackageFileHashes(rootPath turbopath.AbsoluteSystemPath, packagePath tur
 func manuallyHashFiles(rootPath turbopath.AbsoluteSystemPath, files []turbopath.AnchoredSystemPath) (map[turbopath.AnchoredUnixPath]string, error) {
 	hashObject := make(map[turbopath.AnchoredUnixPath]string)
 	for _, file := range files {
-		hash, err := fs.GitLikeHashFile(file.ToString())
+		hash, err := fs.GitLikeHashFile(file.RestoreAnchor(rootPath))
```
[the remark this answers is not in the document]
nathanhammond: Yep, this bug is fully addressed.

cli/internal/taskhash/taskhash.go
nathanhammond: The actual deduping I want to do is to push all files into a set, do a set intersection, and only hash files that don't already have a hash. There's no particular reason to tie it to a package or task.

PR #4820, opened 2023-05-04
First pass at file hashing for a package

gsoltis:
### Description

Ports pieces of `package_deps_hash.go`:
 - Hashing a package that does not use `inputs`
 - Calling various `git` commands: `hash-object`, `ls-tree`, `status`
 - Wires up FFI for `pub fn get_package_file_hashes_from_git_index`

### Testing Instructions

Relevant tests ported. Note that some tests are not necessary anymore (validating irrelevant fields from `git`), and some are implicit in other tests (`find_git_root`). In this PR, the existing Go tests also run over the FFI boundary against the rust implementation.

Tests for hashing specific filesets not ported yet, since the Rust functionality doesn't exist yet.


crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -62,6 +62,10 @@ impl AbsoluteSystemPathBuf {
         Ok(AbsoluteSystemPathBuf(system_path))
     }
 
+    pub fn new_unchecked(raw: impl Into<PathBuf>) -> Self {
```
NicholasLYang, line 65: I believe @arlyon recommended making these methods unsafe. Should we start doing that?

crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -171,6 +183,10 @@ impl AbsoluteSystemPathBuf {
         }
     }
 
+    pub fn create_dir(&self) -> Result<(), io::Error> {
```
NicholasLYang, line 186: Can we keep the naming consistent as `create_dir_all`?

crates/turbopath/src/absolute_system_path_buf.rs
```diff
@@ -245,6 +261,11 @@ impl AbsoluteSystemPathBuf {
     pub fn open(&self) -> Result<fs::File, PathError> {
         Ok(fs::File::open(&self.0)?)
     }
+
+    pub fn to_realpath(&self) -> Result<Self, PathError> {
```
NicholasLYang, line 265: What does realpath mean in this context?

crates/turbopath/src/lib.rs
```diff
@@ -43,11 +47,18 @@ pub enum PathValidationError {
...
+    NotRelative(String),
     #[error("Path {0} is not parent of {1}")]
     NotParent(String, String),
     #[error("Path {0} is not a unix path")]
     NotUnix(String),
+    #[error("{0} is not a prefix for {1}")]
+    PrefixError(String, String),
+}
+
+pub(crate) fn not_relative_error(bytes: &[u8]) -> PathValidationError {
```
NicholasLYang, line 59: We should put this inside an `impl PathValidationError`

crates/turbopath/src/relative_unix_path.rs
```diff
@@ -1,25 +1,48 @@
...
+    inner: BStr,
 }
 
 impl RelativeUnixPath {
-    pub fn new<P: AsRef<Path>>(value: &P) -> Result<&Self, PathError> {
+    pub fn new<P: AsRef<BStr>>(value: &P) -> Result<&Self, PathError> {
         let path = value.as_ref();
-        if path.is_absolute() {
-            return Err(PathValidationError::NotRelative(path.to_owned()).into());
+        if path[0] == b'/' {
```
NicholasLYang, line 15: Will this panic on empty paths?

crates/turbopath/src/relative_unix_path.rs
```diff
@@ -1,25 +1,48 @@
...
+        #[cfg(windows)]
+        {
+            let system_path_bytes = self
+                .inner
+                .iter()
+                .map(|byte| if *byte == b'/' { b'\\' } else { *byte })
+                .collect::<Vec<u8>>();
+            // Is this safe to do? We think we have utf8 bytes or bytes that roundtrip
+            // through utf8
+            let system_path_string = unsafe { String::from_utf8_unchecked(system_path_bytes) };
```
NicholasLYang, line 43: Yeah I'm not sure if this is safe. Not a unicode expert, but is there a chance that `b'/'` could mean something else inside a unicode string? Like if it's part of a multi-code point character?

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -1,75 +1,163 @@
...
-    ///
-    /// ```
-    /// ```
-    pub fn new(path: impl Into<PathBuf>) -> Result<Self, PathValidationError> {
-        let path = path.into();
-        if path.is_absolute() {
-            return Err(PathValidationError::NotRelative(path));
+    pub fn new(path: impl Into<Vec<u8>>) -> Result<Self, PathError> {
+        let bytes: Vec<u8> = path.into();
+        if !bytes.is_empty() && bytes[0] == b'/' {
```
NicholasLYang, line 13: nit: you could also do `bytes.first() == Some(b'/')`

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -1,75 +1,163 @@
...
+    }
+
+    // write_escaped_bytes writes this path to the given writer in the form
+    // "<escaped path>", where escaped_path is the path with '"' and '\n'
+    // characters escaped with '\'.
+    pub fn write_escapted_bytes<W: Write>(&self, writer: &mut W) -> Result<(), PathError> {
+        writer.write_all(&[b'\"'])?;
+        let mut i: usize = 0;
+        while i < self.0.len() {
+            if let Some(mut index) = self.0[i..]
```
NicholasLYang, line 31: If we're doing this in a loop, isn't this O(n^2)? Could we loop through the slice and upon encountering `\"`, we write a `\\` ahead of it?

NicholasLYang (changes requested):
Focusing on the path implementation. Do we know that if we encounter the byte `b'\\'` it's always going to refer to the character `\`?

crates/turbopath/src/absolute_system_path_buf.rs
gsoltis: This one appears to have been a bad rebase and isn't used. I think if and when it reappears, we can reassess.

crates/turbopath/src/absolute_system_path_buf.rs
gsoltis: `man realpath` -> "realpath – returns the canonicalized absolute pathname"

It's a standard unix tool that resolves all symlinks and gives an absolute path.

crates/turbopath/src/lib.rs
gsoltis: Done.

crates/turbopath/src/relative_unix_path_buf.rs
gsoltis: `bytes.first() == Some(&b'/')`, but yeah, I like that better.

crates/turbopath/src/relative_unix_path.rs
gsoltis: Switched to `path.first()`

crates/turbopath/src/relative_unix_path_buf.rs
gsoltis: `i` is the index into `self.0`. Each time we encounter a byte that needs escaping (@ `index`), we write out everything from `i` to `index`, then  `\\` and the byte-to-be-escaped, then increment `i` by one + `index`, moving us past the byte we just escaped. I'll add a comment.

chris-olszewski (commented):
First pass

NicholasLYang (changes requested):

crates/turborepo-scm/src/hash_object.rs
```diff
@@ -0,0 +1,194 @@
...
+    pkg_prefix: &RelativeUnixPathBuf,
+    hashes: &mut GitHashes,
+) -> Result<()> {
+    thread::scope(move |scope| -> Result<()> {
+        let write_thread = scope.spawn(move || -> Result<()> {
+            let mut writer = BufWriter::new(writer);
+            for path in to_hash {
+                path.write_escapted_bytes(&mut writer)?;
+                writer.write_all(&[b'\n'])?;
+                writer.flush()?;
```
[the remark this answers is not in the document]
gsoltis: I'm not sure. Each flush writes to stdin for `git hash-object`.

This function could eventually have 1000s of files passed in, which I don't think we want to flush all at once, but we don't have to flush every single file individually.

From some research on the Go side, it's currently unbuffered writes, which we could also do.

crates/turborepo-scm/src/hash_object.rs
```diff
@@ -0,0 +1,194 @@
...
+            }
+            // writer is dropped here, closing stdin
+            Ok(())
+        });
+        let mut i: usize = 0;
+        let mut buffer: [u8; 41] = [0; 41];
```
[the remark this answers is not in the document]
gsoltis: Extracted to a constant.

crates/turborepo-scm/src/hash_object.rs
```diff
@@ -0,0 +1,194 @@
...
+        loop {
```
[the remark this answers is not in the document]
gsoltis: Yup, good idea.

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -1,75 +1,171 @@
...
+    pub fn write_escapted_bytes<W: Write>(&self, writer: &mut W) -> Result<(), PathError> {
+        writer.write_all(&[b'\"'])?;
+        // i is our pointer into self.0, and to_escape_index is a pointer to the next
+        // byte to be escaped. Each time we find a byte to be escaped, we write
+        // out everything from i to to_escape_index, then the escape byte, '\\',
+        // then the byte-to-be-escaped. Finally we set i to 1 + to_escape_index
+        // to move our pointer past the byte we just escaped.
+        let mut i: usize = 0;
+        while i < self.0.len() {
+            if let Some(mut to_escape_index) = self.0[i..]
```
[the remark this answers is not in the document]
gsoltis: I think the `.enumerate()` call is in the wrong spot, it will enumerate the bytes to be escaped, not the underlying bytes. Do you know if fixing that will cause a tuple to be allocated and passed through the filter function?

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -1,75 +1,171 @@
...
+                .iter()
+                .position(|byte| *byte == b'\"' || *byte == b'\n')
+            {
+                // renormalize the index into the byte vector
+                to_escape_index += i;
+                writer.write_all(&self.0[i..to_escape_index])?;
+                let byte = self.0[to_escape_index];
+                if byte == b'\"' {
```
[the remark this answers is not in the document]
gsoltis: Nope, good spot.

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -1,75 +1,171 @@
...
+                i = self.0.len();
+            }
+        }
+        writer.write_all(&[b'\"'])?;
+        Ok(())
+    }
+
+    pub fn strip_prefix(&self, prefix: &RelativeUnixPathBuf) -> Result<Self, PathError> {
+        let prefix_len = prefix.0.len();
+        if prefix_len == 0 {
```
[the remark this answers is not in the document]
gsoltis: It should be.

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -1,75 +1,171 @@
...
+            return Ok(self.clone());
+        }
+        if !self.0.starts_with(&prefix.0) {
+            return Err(PathError::PathValidationError(
+                PathValidationError::NotParent(prefix.0.to_string(), self.0.to_string()),
+            ));
+        }
+        if self.0[prefix_len] != b'/' {
```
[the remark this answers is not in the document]
gsoltis: Yup. Fixed.

crates/turbopath/src/relative_unix_path_buf.rs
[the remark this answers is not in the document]
NicholasLYang: Ah yeah the enumerate should be before. I wouldn't worry about the tuple allocation because tuples are value types so they'd be on the stack, and Rust is probably smart enough to split that into two registers anyways.

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -1,75 +1,171 @@
...
             .to_str()
-            .ok_or_else(|| PathValidationError::InvalidUnicode(self.0.clone()))
+            .or_else(|_| Err(PathError::Utf8Error(self.0.as_bytes().to_owned())))?;
+        Ok(s)
+    }
+
+    // write_escaped_bytes writes this path to the given writer in the form
+    // "<escaped path>", where escaped_path is the path with '"' and '\n'
+    // characters escaped with '\'.
+    pub fn write_escapted_bytes<W: Write>(&self, writer: &mut W) -> Result<(), PathError> {
```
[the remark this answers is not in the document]
gsoltis: Done.

crates/turbopath/src/relative_unix_path_buf.rs
```diff
@@ -1,75 +1,171 @@
...
+            return Ok(self.clone());
+        }
+        if !self.0.starts_with(&prefix.0) {
+            return Err(PathError::PathValidationError(
+                PathValidationError::NotParent(prefix.0.to_string(), self.0.to_string()),
+            ));
+        }
+        if self.0[prefix_len] != b'/' {
+            let prefix_str = prefix.as_str().unwrap_or("invalid utf8").to_string();
+            let this = self.as_str().unwrap_or("invalid utf8").to_string();
```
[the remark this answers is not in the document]
gsoltis: Yup, fixed.

crates/turbopath/src/relative_unix_path_buf.rs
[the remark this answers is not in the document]
gsoltis: Implemented a version of this.

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -0,0 +1,148 @@
...
+        (tmp_dir, dir)
+    }
+
+    fn require_git_cmd(repo_root: &AbsoluteSystemPathBuf, args: &[&str]) {
+        let mut cmd = Command::new("git");
+        cmd.args(args).current_dir(repo_root);
+        assert_eq!(cmd.output().unwrap().status.success(), true);
+    }
+
+    fn setup_repository(repo_root: &AbsoluteSystemPathBuf) {
```
NicholasLYang, line 57: Should we reuse the `setup_repository` code from `git.rs`?

NicholasLYang (approved):
Just one or two nits but lgtm otherwise!

crates/turborepo-lib/src/daemon/file_descriptor_set.bin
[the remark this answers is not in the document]
gsoltis: I added the `.gitignore` change onto this PR.

crates/turborepo-scm/src/package_deps.rs
gsoltis: We can. We'd need to find a natural place for that function to live, right?

NicholasLYang: Yeah we can have a test_utils.rs or something

crates/turborepo-scm/src/hash_object.rs
```diff
@@ -32,65 +30,66 @@ pub(crate) fn hash_objects(
...
                 writer.write_all(&[b'\n'])?;
                 writer.flush()?;
             }
             // writer is dropped here, closing stdin
             Ok(())
         });
-        let mut i: usize = 0;
-        let mut buffer: [u8; 41] = [0; 41];
-        loop {
+        //let mut i: usize = 0;
```
chris-olszewski, line 79:
```suggestion
```

chris-olszewski (approved):

crates/turborepo-ffi/src/lib.rs
```diff
@@ -163,3 +163,87 @@ pub extern "C" fn recursive_copy(buffer: Buffer) -> Buffer {
     };
     response.into()
 }
+
+#[no_mangle]
+pub extern "C" fn get_package_file_hashes_from_git_index(buffer: Buffer) -> Buffer {
+    let req: proto::GetPackageFileHashesFromGitIndexRequest = match buffer.into_proto() {
```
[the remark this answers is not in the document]
nathanhammond: I mean, this whole thing is supposed to go away soon. Doing much more than the minimum here is low ROI.

crates/turborepo-scm/src/hash_object.rs
```diff
@@ -0,0 +1,192 @@
...
+                writer.write_all(&[b'\n'])?;
+                writer.flush()?;
+            }
+            // writer is dropped here, closing stdin
+            Ok(())
+        });
+        // Buffer size is HASH_LEN + 1 to account for the trailing \n
+        let mut buffer: [u8; HASH_LEN + 1] = [0; HASH_LEN + 1];
+        for (i, filename) in to_hash.iter().enumerate() {
+            if i == to_hash.len() {
```
nathanhammond, line 82: I'm pretty sure that this doesn't do a safety check on the reader being entirely consumed at the same time as the file list in `to_hash` is consumed.

`write_escaped_bytes` _should_ prevent this from being an issue, but the test for newlines didn't get ported to Rust so I don't know if it works:
https://github.com/vercel/turbo/blob/main/cli/internal/hashing/package_deps_hash_test.go#L45-L115

I'm less certain about the behavior of a broken pipe on inputs happening on the last path in the list.

crates/turborepo-scm/src/hash_object.rs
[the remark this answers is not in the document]
nathanhammond: I decided that it didn't matter in the first pass, I still don't think it matters. There might be a minor perf win here, but it's not the primary limiter.

PR #4824, opened 2023-05-04
fix: no longer include peer dependencies in transitive closures

### LINKED ISSUE #4757: [turborepo] "error performing lockfile operation" on 1.9.3

1.9.3

npm

Mac

### Describe the Bug

After upgrading to 1.9.3. from 1.8, I'm seeing the following warning when running any turbo command:

```
Issues occurred when constructing package graph. Turbo will function, but some features may not be available:  4 errors occurred:
        * error performing lockfile operation
        * error performing lockfile operation
        * error performing lockfile operation
        * error performing lockfile operation
```

I also see it on the canary builds. Turborepo operates the way it used to (from what I can tell) but I can't find a way to determine the actual nature of these errors. Increasing verbosity doesn't appear to help.

### Expected Behavior

Expected to see a description of the actual lockfile error(s) so that they can be corrected.

### To Reproduce

This happens whenever I run `turbo <command>`. Unable to share a repo link, as it's a private repository.

### Reproduction Repo

_No response_

chris-olszewski:
### Description

Fixes #4757

Underlying issue was I was including `peerDependencies` in the dependency crawl. This resulted in an error since we expect to only be looking at external dependencies in the dependency crawl.

### Testing Instructions

Added a failing unit test where there's a workspace that's used as a peer dependency that conflicts with a package in the registry.


mehulkar (approved):

jgoz:
Thanks for following through on this @chris-olszewski. Really appreciate your effort here.

PR #4828, opened 2023-05-05
Implement TaskInput derive macro

alexkirsz:
### Description

This PR implements a `TaskInput` derive macro that allows users to define structures and enums that can be directly passed as arguments to `tt::function`s, provided they are composed exclusively of other `TaskInput` implementors (i.e. primitives for which we've manually implemented `TaskInput`, `Vc`s, and other structs that derive `TaskInput`).

There are a lot of cases in Turbopack where we'd like to pass more advanced structures to `tt::function`s, but since there's no existing way to create composite structures and enums from existing `Vc`s and primitives, we often end up:

1. with functions that have 6+ arguments;
2. passing multiple `Option` arguments when they only make sense alongside another such argument;
3. or wrapping our arguments into `Value` (which has many other requirements).

For 2., this might look something like:

```rust
#[turbo_tasks::function]
fn do_something(value1: SomeVc, value2: Option<OtherVc>, value3: Option<AnotherVc>) {
  // ...
}
```

Where `value2` might only make sense when `value3` is also provided. An enum would be better in this case:

```rust
#[derive(Clone, TaskInput)
enum CustomEnum {
  NoValues,
  Values {
    value2: OtherVc,
    value3: AnotherVc,
  }
}

#[turbo_tasks::function]
fn do_something(value1: SomeVc, values: CustomEnum) {
  // ...
}
```

For 3., this might look like the following:

```rust
#[derive(PartialOrd, Ord, Debug, Hash, Clone, Copy)]
pub enum SomeArgument {
  A,
  B,
  ...
}

// and then later

some_function(Value::new(SomeArgument::B))
```

which can now be replaced by

```rust
#[derive(TaskInput, Clone)]
pub enum SomeArgument {
  A,
  B,
  ...
}

// and then later

some_function(SomeArgument::B)
```

### Testing Instructions

I added a series of tests in turbo-tasks-memory.


crates/turbo-tasks-macros/src/derive/task_input_macro.rs
```diff
@@ -0,0 +1,267 @@
...
+
+            let variants_discriminants: Vec<_> = (0..variants_idents.len())
+                .map(|i| Literal::usize_unsuffixed(i))
+                .collect();
+
+            (
+                quote! {
+                    match value {
+                        turbo_tasks::TaskInput::List(value) => {
+                            let mut iter = value.iter();
```
jridgewell, line 103: Nit: make this name more distinct, since it's declaration and usage are split between two different token streams.

crates/turbo-tasks-memory/tests/task_input.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+#[derive(Clone, TaskInput)]
+struct OneUnnamedField(u32);
+
+#[turbo_tasks::function]
+async fn one_unnamed_field(input: OneUnnamedField) -> CompletionVc {
+    assert_eq!(input.0, 42);
+    CompletionVc::immutable()
+}
```
jridgewell, line 24: Nit: I think this test file would be much more straight forward if we didn't depend on turbo at all. Something like:

```suggestion
#[derive(TaskInput)]
struct OneUnnamedField(u32);

#[test]
fn test_one_unnamed_field() -> Result<()> {
    let input = OneUnnamedField(42);
    let actual: OneUnnamedField = FromTaskInput::try_from(&TaskInput::from(input))?;
    assert_eq!(actual.0, 42);
    Ok(())
}
```

crates/turbo-tasks-macros/src/derive/task_input_macro.rs
```diff
@@ -0,0 +1,267 @@
...
+                            })
+                        },
+                        _ => Err(anyhow::anyhow!("invalid task input type, expected list")),
+                    }
+                },
+                quote! {
+                    match value {
+                        #(
+                            #ident::#variants_idents #variants_fields_destructuring => {
+                                let mut list = Vec::with_capacity(1 + #variants_fields_len);
```
jridgewell, line 125: Ditto distinct naming.

jridgewell (approved):

crates/turbo-tasks-memory/tests/task_input.rs
alexkirsz: Fair enough. I'll move these tests to turbo-tasks. I'll just keep one here as an integration test between tt::function and derive(TaskInput).

sokra (approved):

crates/turbo-tasks-macros-shared/src/expand.rs
```diff
@@ -103,15 +105,22 @@ pub fn expand_fields<
...
-    fields: impl Iterator<Item = &'a Field>,
-    ignore_field: &I,
+    fields: impl Iterator<Item = &'a Field> + ExactSizeIterator,
+    filter_field: &I,
 ) -> (TokenStream, Vec<TokenStream>) {
+    let fields_len = fields.len();
     let (captures, fields_idents): (Vec<_>, Vec<_>) = fields
+        .filter(|field| filter_field(field))
         .enumerate()
-        .filter(|(_i, field)| !ignore_field(field))
```
jridgewell, line 113: I think this is a bug. The enumerate is done first so that we can get the index of the item in input. If we filter first, then index 1 could refer to the third field.

PR #4833, opened 2023-05-05
Combine release workflow into a single step

mehulkar:
- implemented in #4699
- reverted in #4830

The issue is that this doesn't work in Windows:

```bash
echo "STAGE_BRANCH=${{ needs.stage.outputs.stage-branch }}" >> $GITHUB_ENV
```

so the correct commit doesn't get checked out to build the Windows binary.

I had noted in the PR description in #4699 that this approach isn't vital to the process and it came from a different motivation to keep the diff small. In the second commit in this PR, I change the approach to just use the `outputs` of the stage branch directly where it is used. 

Tested in https://github.com/mehulkar/turbo/actions/runs/4895855848. You can see that Windows job builds the same ref sa the other jobs:
- Windows commit https://github.com/mehulkar/turbo/actions/runs/4895855848/jobs/8742252520#step:3:490
- Ubuntu: https://github.com/mehulkar/turbo/actions/runs/4895855848/jobs/8742252419#step:3:477


NicholasLYang (approved):
Honestly so so glad this is a single step now!

PR #4837, opened 2023-05-05
fix: sort tasks in run summary output by task id

chris-olszewski:
### Description

Was using `--summarize` to debug a run and got annoyed that tasks were getting reordered.

### Testing Instructions

Inspect output of a summary json and verify that tasks no longer have a random order.


tknickman (approved):
lgtm, to be honest I don't remember why we weren't doing this originally (@mehulkar do you remember?)

mehulkar (approved):

PR #4839, opened 2023-05-05
Update basic example for App Router.

anthonyshew:


examples/basic/packages/ui/Header.tsx
```diff
@@ -0,0 +1,5 @@
+import * as React from "react";
+
+export const Header = ({ text }: { text: string }) => {
+  return <h1>{text}aaaaa</h1>;
```
mehulkar, line 4: something better here?

examples/basic/packages/ui/Button.tsx
```diff
@@ -1,5 +1,7 @@
+"use client";
+
 import * as React from "react";
 
 export const Button = () => {
-  return <button>Boop</button>;
+  return <button onClick={() => console.log("boop")}>Boop</button>;
```
mehulkar, line 6:
```suggestion
  return <button onClick={() => alert("boop")}>Boop</button>;
```

mehulkar (approved):
🚢

examples/basic/packages/ui/Header.tsx
anthonyshew: Ha, forgot I left that there when I was debugging that caching question.

PR #4840, opened 2023-05-05
Update kitchen-sink with App Router.

anthonyshew:


tknickman (changes requested):
One minor change request regarding the new tsup config.

examples/kitchen-sink/packages/ui/tsup.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  entry: {
+    ".": "src/index.tsx",
+  },
+  banner: {
+    js: "'use client'",
+  },
```
tknickman, line 9: This works, but it's a little confusing as it's now dictating the type of components we can put in this package. And it makes the decision between bundling / using an internal package more nuanced. Not a change request, just a comment!

examples/kitchen-sink/packages/ui/tsup.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
```
tknickman, line 3: Also, now that we have a config here you can remove the args from `package.json` `build` and `dev` scripts and just call 'tsup and `tsup --watch` respectively. You will also need to update this config to support overrides from the cli args to make that watch option work!

```js

export default defineConfig((options: Options) => ({
  [defaults]
  ...options,
}));
```

tknickman (changes requested):
One minor change request regarding the new tsup config.

examples/kitchen-sink/packages/ui/tsup.config.ts
anthonyshew: It's definitely weird but I poked around the ecosystem and it doesn't look like there is an idiomatic way that people handle this yet. It's also notable that it's generally an okay approach as we're all learning how to migrate together. 😬

tknickman (approved):

PR #4841, opened 2023-05-05
feature(turborepo): AbsoluteSystemPath

NicholasLYang:
### Description

In the process of porting cache stuff, I started needing an `AbsoluteSystemPath` so I wouldn't have to clone paths too much.

This PR creates that type and implements a conversion from it to `AbsoluteSystemPathBuf` as `ToOwned` and vice versa for `Borrowed`. It also changes any usage of `&AbsoluteSystemPathBuf` to `&AbsoluteSystemPath`.

There's also some renaming and moving of functions to match `std::path`

### Testing Instructions

I added a test for creating an `AbsoluteSystemPath`

### DIFF
--- a/crates/turbopath/src/absolute_system_path.rs
+++ b/crates/turbopath/src/absolute_system_path.rs
@@ -1,0 +1,199 @@
+#[cfg(not(windows))]
+use std::os::unix::fs::symlink as symlink_file;
+#[cfg(not(windows))]
+use std::os::unix::fs::symlink as symlink_dir;
+#[cfg(windows)]
+use std::os::windows::fs::{symlink_dir, symlink_file};
+use std::{
+    borrow::Cow,
+    fmt, fs,
+    fs::Metadata,
+    io,
+    path::{Path, PathBuf},
+};
+
+use path_slash::CowExt;
+
+use crate::{
+    AbsoluteSystemPathBuf, AnchoredSystemPathBuf, IntoSystem, PathError, PathValidationError,
+    RelativeSystemPathBuf, RelativeUnixPath,
+};
+
+pub struct AbsoluteSystemPath(Path);
+
+impl ToOwned for AbsoluteSystemPath {
+    type Owned = AbsoluteSystemPathBuf;
+
+    fn to_owned(&self) -> Self::Owned {
+        AbsoluteSystemPathBuf(self.0.to_owned())
+    }
+}
+
+impl AsRef<AbsoluteSystemPath> for AbsoluteSystemPath {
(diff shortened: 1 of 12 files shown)


crates/turbopath/src/absolute_system_path.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+impl AbsoluteSystemPath {
+    pub fn new<P: AsRef<Path> + ?Sized>(value: &P) -> Result<Cow<Self>, PathError> {
+        let path = value.as_ref();
+        if path.is_relative() {
+            return Err(PathValidationError::NotAbsolute(path.to_owned()).into());
+        }
+
+        let system_path = path.to_slash().ok_or_else(|| {
```
gsoltis, line 57: I don't think this is right. `to_slash` converts to unix-style, not system.

Separately, I'm not sure we want `new` to do the conversion. I think having that be explicit would be helpful for readers following the code that uses this type.

gsoltis (changes requested):
Are there instances where you are relying on `new` to do a conversion?

crates/turbopath/src/absolute_system_path.rs
NicholasLYang: Do we want to have an unchecked new, or just allow creating an `AbsoluteSystemPath` from an `AbsoluteSystemPathBuf`?

NicholasLYang: I suspect this is worthing having because most paths that we encounter will be system paths, so they won't require any allocation. Right now we do a copy every time we wrap something in `AbsoluteSystemPathBuf`

gsoltis: I think a `pub(crate) fn unchecked_new(...)` would be appropriate for places inside `turbopath` where we already know we have a system path.

gsoltis: And if you find yourself needing to convert to a system path from _outside_ of `turbopath`, I think an explicit call to a conversion function is the way to do it, even if it returns a `Cow` of some sort.

NicholasLYang: Isn't this an explicit call to a conversion function? I'm not clear on how that's different than what we have here

gsoltis: Discussed in person, but to document: we should have a constructor where we are validating but not converting for instances where we expect to already have an absolute system path, and a constructor that explicitly does the conversion if we have places in the code where we need to do a conversion.

gsoltis (approved):

PR #4843, opened 2023-05-05
Update with-changesets with App Router.

anthonyshew:


examples/with-changesets/package.json
```diff
@@ -15,7 +15,7 @@
     "eslint": "^7.32.0",
     "eslint-config-acme": "workspace:*",
     "prettier": "^2.5.1",
-    "turbo": "latest"
+    "turbo": "^1.9.3"
```
mehulkar, line 18: Didn't catch this in the other PRs, btu for turbo we _do_ want `latest`, especially since these examples are used to test published versions

mehulkar (changes requested):

mehulkar (approved):

examples/with-changesets/package.json
mehulkar: never mind, we have lockfiles now so this is misleading

PR #4844, opened 2023-05-05
Update with-docker with App Router.

anthonyshew:


tknickman (approved):
Validated this one locally since it's built with standalone and can be a little weird. But looks awesome!

PR #4847, opened 2023-05-05
Update with-react-native-web with App Router.

anthonyshew:


examples/with-react-native-web/packages/ui/src/index.tsx
```diff
@@ -1 +1 @@
-export { Button, type ButtonProps } from "./button";
+export { Button, type ButtonProps } from "./Button";
```
tknickman, line 1: This looks like it's causing the build to fail.

examples/with-react-native-web/packages/ui/tsup.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
```
tknickman, line 3: Same comments here from https://github.com/vercel/turbo/pull/4840#discussion_r1188640745

examples/with-react-native-web/packages/ui/src/index.tsx
anthonyshew: Classic "works on my machine."

anthonyshew:
Okay, think we're good now. 😁

tknickman (approved):
Looks good to me, was expo looking good for you locally? Sometimes I've had some issues there, but as long as all looks good we can :sendit:

PR #4856, opened 2023-05-05
fix: relax clippy regex

chris-olszewski:
### Description

I think the addition of [cargo clippy --fix](https://doc.rust-lang.org/nightly/clippy/usage.html#automatically-applying-clippy-suggestions) suggestion to the clippy output made the sed regex no longer match.

### Testing Instructions

CI should break on this PR as it will now find all of the clippy warnings on main.


.github/workflows/test.yml
```diff
@@ -536,6 +530,21 @@ jobs:
             fi
           fi
 
+      - name: Stage Clippy warnings for updating expected warnings on main
+        if: github.ref == 'refs/heads/main'
+        uses: actions/cache/save@v3
```
sokra, line 535:
```suggestion
```

Not sure, but should this be removed?

anthonyshew:
Just closing up some stales!

PR #4868, opened 2023-05-08
vdbg! macro

alexkirsz:
### Description

The `vdbg!` macro works the same way as the `dbg!` macro, with the additional benefit of automatically calling `ValueDebugFormat` implementations.

This means that you can do:

```rust
vdbg!(42, asset_ident_vc);
vdbg!(42, asset_ident_vc; depth = 1);
```

and the macro will print out

```rust
[src/main.rs:10] 42 = 42
[src/main.rs:10] some_vc = AssetIdent {
    path: FileSystemPath {
        fs: FileSystem(DiskFileSystem {
            name: "project",
            root: "...",
        }),
        path: "...",
    },
    query: None,
    fragment: None,
    assets: [],
    modifiers: [],
    part: None,
}
```

```rust
[src/main.rs:10] 42 = 42
[src/main.rs:10] asset_ident_vc = AssetIdent {
    path: FileSystemPath,
    query: core::option::Option<turbo_tasks::primitives::StringVc>,
    fragment: core::option::Option<turbo_tasks::primitives::StringVc>,
    assets: alloc::vec::Vec<(turbo_tasks::primitives::StringVc, turbopack_core::ident::AssetIdentVc)>,
    modifiers: alloc::vec::Vec<turbo_tasks::primitives::StringVc>,
    part: core::option::Option<turbopack_core::resolve::ModulePartVc>,
}
```

This is an ergonomic improvement over having to call `.dbg()` or `.dbg_depth()` manually in four major ways:
1. No need to be in an async context and call `.await?`: the macro will spawn a task on its own. This means that it is no longer necessary to convert a `turbo_tasks::function` to async just to debug one of its arguments.
2. No need to import the `turbo_tasks::debug::ValueDebug` trait. You can call `turbo_tasks::vdbg!(...)` directly, of `vdbg!(...)` if you have it imported already.
3. No need to call `eprintln!` or `dbg!`: `vdbg!` calls `eprintln!` directly under the hood, with the same format as `dbg!`.
4. It's just one macro, with support for the `depth` argument baked in.

### Testing Instructions

N/A
link WEB-520

### DIFF
--- a/crates/turbo-tasks/src/manager.rs
+++ b/crates/turbo-tasks/src/manager.rs
@@ -118,6 +118,12 @@ pub trait TurboTasksApi: TurboTasksCallApi + Sync + Send {
     fn mark_own_task_as_finished(&self, task: TaskId);
 
     fn connect_task(&self, task: TaskId);
+
+    /// Wraps the given future in the current task.
+    fn detached(
+        &self,
+        f: Pin<Box<dyn Future<Output = Result<()>> + Send + 'static>>,
+    ) -> Pin<Box<dyn Future<Output = Result<()>> + Send + 'static>>;
 }
 
 /// The type of stats reporting.
@@ -979,6 +985,22 @@ impl<B: Backend + 'static> TurboTasksApi for TurboTasks<B> {
     fn mark_own_task_as_finished(&self, task: TaskId) {
         self.backend.mark_own_task_as_finished(task, self);
     }
+
+    fn detached(
+        &self,
+        f: Pin<Box<dyn Future<Output = Result<()>> + Send + 'static>>,
+    ) -> Pin<Box<dyn Future<Output = Result<()>> + Send + 'static>> {
+        Box::pin(
+            TURBO_TASKS.scope(
+                turbo_tasks(),
+                CURRENT_TASK_ID.scope(
+                    CURRENT_TASK_ID.with(|id| *id),
+                    self.backend
+                        .execution_scope(CURRENT_TASK_ID.with(|id| *id), f),
+                ),
+            ),
+        )
+    }
 }
 
 impl<B: Backend + 'static> TurboTasksBackendApi<B> for TurboTasks<B> {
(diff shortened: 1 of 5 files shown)


crates/turbo-tasks/src/manager.rs
```diff
@@ -979,6 +985,22 @@ impl<B: Backend + 'static> TurboTasksApi for TurboTasks<B> {
...
+
+    fn detached(
+        &self,
+        f: Pin<Box<dyn Future<Output = Result<()>> + Send + 'static>>,
+    ) -> Pin<Box<dyn Future<Output = Result<()>> + Send + 'static>> {
+        Box::pin(
+            TURBO_TASKS.scope(
+                turbo_tasks(),
+                CURRENT_TASK_ID.scope(
+                    CURRENT_TASK_ID.with(|id| *id),
```
jridgewell, line 997: Nit:
```suggestion
                    CURRENT_TASK_ID.get(),
```

crates/turbo-tasks/src/manager.rs
```diff
@@ -979,6 +985,22 @@ impl<B: Backend + 'static> TurboTasksApi for TurboTasks<B> {
...
+                    self.backend
+                        .execution_scope(CURRENT_TASK_ID.with(|id| *id), f),
```
jridgewell, line 999: Nit: is it slow to double access? Could we `let current_task = CURRENT_TASK_ID.get()` and reuse it in both places?

jridgewell (approved):
Can we add a default impl for items which implement `Debug` but don't implement `ValueDebugFormat` specifically?

ForsakenHarmony:
@jridgewell isn't this already the case?
https://github.com/vercel/turbo/blob/main/crates/turbo-tasks/src/debug/mod.rs#L75

jridgewell:
Ohh, I didn't read the comments and thought we still needed an `impl<T> ValueDebugFormat for T` (not `for &T`).

PR #4870, opened 2023-05-08
Add `dotEnv` to `turbo.json`

nathanhammond:
Enable configuring `.env` files inside of `turbo.json`. These are addition-only to file hashes, and enables specifying load order for cache distinctions.


cli/internal/fs/turbo_json.go
```diff
@@ -556,6 +568,23 @@ func (btd *BookkeepingTaskDefinition) UnmarshalJSON(data []byte) error {
...
 
+	if task.DotEnv != nil {
+		btd.definedFields.Add("DotEnv")
+
+		// Going to _at least_ be an empty array.
+		btd.TaskDefinition.DotEnv = make(turbopath.AnchoredUnixPathArray, 0, len(task.DotEnv))
+
+		// Port the raw globalDotEnv values in.
+		for _, dotEnvPath := range task.DotEnv {
+			if filepath.IsAbs(dotEnvPath) {
```
NicholasLYang, line 579: Is this the case for only task dotenvs or also the global dot env?

cli/internal/hashing/package_deps_hash.go
```diff
@@ -371,10 +380,13 @@ func gitHashObject(anchor turbopath.AbsoluteSystemPath, filesToHash []turbopath.
 	return output, nil
 }
 
-func manuallyHashFiles(rootPath turbopath.AbsoluteSystemPath, files []turbopath.AnchoredSystemPath) (map[turbopath.AnchoredUnixPath]string, error) {
-	hashObject := make(map[turbopath.AnchoredUnixPath]string)
+func manuallyHashFiles(rootPath turbopath.AbsoluteSystemPath, files []turbopath.AnchoredSystemPath, existing bool) (map[turbopath.AnchoredUnixPath]string, error) {
+	hashObject := make(map[turbopath.AnchoredUnixPath]string, len(files))
 	for _, file := range files {
 		hash, err := fs.GitLikeHashFile(file.RestoreAnchor(rootPath))
+		if errors.Is(err, &os.PathError{}) {
```
NicholasLYang, line 387: Is there supposed to be an `existing &&` check here?

mehulkar (commented):
- Is there any difference between using this key and the `inputs` key? 
- do we want to throw an error if it's specified in both places?
- What happens if you add `inputs: [!.env]`?

cli/internal/fs/turbo_json_test.go
```diff
@@ -27,6 +27,150 @@ func assertIsSorted(t *testing.T, arr []string, msg string) {
...
+			},
+		},
+	}
+
+	assert.Equal(t, pipelineExpected, turboJSON.Pipeline)
+
+	// Snapshot test of serialization.
+	bytes, _ := turboJSON.MarshalJSON()
+	assert.Equal(t, "{\"globalDotEnv\":[\"z\",\"y\",\"x\"],\"pipeline\":{\"build\":{\"outputs\":[],\"cache\":true,\"dependsOn\":[],\"inputs\":[],\"outputMode\":\"full\",\"dotEnv\":[\"3\",\"2\",\"1\"],\"env\":[],\"persistent\":false}},\"remoteCache\":{}}", string(bytes))
+}
```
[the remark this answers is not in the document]
nathanhammond: Funny story: I have integration tests for `passThroughEnv` that didn't catch the issues that this caught and I'm adding bonus tests to the `EnvMode` stabilization PR right now and fixing the behavior too.

cli/internal/fs/turbo_json.go
```diff
@@ -556,6 +568,23 @@ func (btd *BookkeepingTaskDefinition) UnmarshalJSON(data []byte) error {
...
+				log.Printf("[WARNING] Using an absolute path in \"dotEnv\" (%v) will not work and will be an error in a future version", dotEnvPath)
```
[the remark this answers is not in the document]
nathanhammond: I'm not actually convinced that there is anything wrong with absolute paths here. I aligned to the current behavior but I believe we should actually remove this warning and allow absolute paths.

nathanhammond:
@mehulkar 

> Is there any difference between using this key and the `inputs` key? 

Yes, specifying the first glob in `inputs` toggles off the "include everything in the git index inside this package's folder" default. (caveats apply, but they're bugs: [see Slack thread](https://vercel.slack.com/archives/C02HEJASXGD/p1683777237028369?thread_ts=1683753757.861519&cid=C02HEJASXGD))

Having an alternative method to specify this does a few things:
- By simple existence causes people to actively consider their `.env` files. Sometimes simply naming a thing or pattern makes the UX better. (For context: you could do 100% of what Ember Services could do when introduced using vanilla Ember things if you knew the incantation; same applies here.)
- Creates a configuration location that does not toggle off the default behavior.
- Gives us some rails to ride on for future "help me not screw up `.env` files" behavior.

> do we want to throw an error if it's specified in both places?

No, because it doesn't make any material difference if you include it in both places (they get combined into the same list of hashed files).

> What happens if you add `inputs: [!.env]`?

Given this configuration:
```
{
  "inputs": ["**", "!.env*"],
  "dotEnv": [".env.local", ".env"]
}
```

The file hash will include `.env` and `.env.local` (if present on disk). This is a conservative default; it results in over-partitioning in the event of misconfiguration.

In one proposed possible future variant `turbo --dotenv` will load environment variables from those locations and _not_ merge the files into the task hash; instead encouraging you to specify your environment variable dependencies using `env` (and `passthroughEnv`). This would make the configuration you describe _appear_ sensible, but would already be configured for you by turning all of `dotEnv` to `excludes` when that flag is present.

(Including the entire file and the entire fallback chain is _already_ over-partitioning; this would enable precise targeting.)

cli/internal/fs/turbo_json.go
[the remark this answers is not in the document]
nathanhammond: Follow along here: https://vercel.slack.com/archives/C02HEJASXGD/p1683801269758379

cli/internal/fs/turbo_json.go
nathanhammond: Follow along here: https://vercel.slack.com/archives/C02HEJASXGD/p1683801269758379

cli/internal/hashing/package_deps_hash.go
nathanhammond: Nah, `nil` can't ever be the thing; and it's type-constrained to `error`. Therefore `errors.Is` already covers the existence check.

cli/internal/graph/graph.go
```diff
@@ -85,15 +85,22 @@ func (g *CompleteGraph) GetPackageTaskVisitor(
...
 				// But our old behavior's actual meaning of this state is `loose`.
 				taskEnvMode = util.Loose
 			}
 		}
 
+		// Detect usage of strict env vars.
+		usingStrictEnvVars := globalEnvMode != util.Infer || taskDefinition.PassthroughEnv != nil
+
+		// Detect usage of dotEnv.
+		usingDotEnv := taskDefinition.DotEnv != nil
```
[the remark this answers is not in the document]
nathanhammond: This is going away before landing since it's going to trigger a hash change.

cli/internal/fs/turbo_json.go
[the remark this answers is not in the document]
nathanhammond: In the mean time I made it error for absolute paths; we can always relax errors in a backwards-compatible way.

nathanhammond (commented):
Reviewer's Guide!

cli/internal/fs/turbo_json_test.go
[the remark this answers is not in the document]
nathanhammond: (It was this serialization test that actually caught issues since it's only used in `prune`.)

cli/internal/hashing/package_deps_hash.go
NicholasLYang: Gotcha. But are we using `existing` anywhere in this function then?

nathanhammond: Oh, also no! `GitLikeHashFile` calls `Open` as the first step, which blows up with an `os.PathError` if it fails.

So we simply try to read it and if that fails we inspect the error to make sure it's the one that you get if it doesn't exist.

A separate existence check would be an unnecessary bonus round trip to the FS.

nathanhammond: OH OH OH; yes there is a bug here. Test incoming. Thank you for your detailed review!

cli/internal/fs/turbo_json.go
nathanhammond: (I locked it down to non-absolute for now; we can always relax an error in the future.)

mehulkar (approved):
The code looks good, but I'm having trouble getting on board with special casing a certain kind of file. Have we considered any alternative ways of doing this? For example an `$auto` option came up a while ago, maybe a similar `inputs: ['$dotenv']` could work? This is not as discoverable as what you have here, but thinking out loud.

Specifically, I'm concerned about:

- Two affordances for specifying file inputs
- Other env var declaration tools than dotenv (now or in the future)
- Other ways of specifying dotenv config that don't match these patterns
- Abusing this feature to "always include a file" that has nothing to do with dotenv
- API expansion in general

I'm also thinking of the auto inclusion of `package.json` and `turbo.json` when inputs are specified. That's another exception to `inputs` that applies under "some but not all" circumstances, and is a totally different mechanism than the one for dotenv files.

In any case, I don't know of any conversations or RFCs where we decided on this implementation, so I don't love having this conversation after the code is already done. I'll assume that you've already done due diligence, but wanted to give a "real" review.

PR #4871, opened 2023-05-08
chore(create-turbo): extract common utils

tknickman:
### Description

Reshuffling utils from create-turbo to `@turbo/utils` to make them available to `@turbo/gen`

- Also fixes a few bugs with tests that don't have fixtures


packages/create-turbo/src/commands/create/prompts.ts
```diff
@@ -98,17 +55,13 @@ export async function packageManager({
...
       !packageManager ||
-      (packageManager as PackageManager) !== "npm" ||
-      !Object.keys(availablePackageManagers).includes(packageManager),
+      !availablePackageManagers?.[packageManager as PackageManager]?.available,
     choices: ["npm", "pnpm", "yarn"].map((p) => ({
       name: p,
       value: p,
-      disabled:
-        // npm should always be available
-        p === "npm" ||
```
mehulkar, line 108: did we lose this logic?

mehulkar (approved):

packages/create-turbo/src/commands/create/prompts.ts
tknickman: Yep, but it was intentional. This was added in an attempt to patch another bug. That bug has [since been fixed](https://github.com/vercel/turbo/pull/4603/files) so we no longer need to allow list here.

PR #4872, opened 2023-05-08
feat(turbo-gen): add @turbo/gen

tknickman:
### Description

`@turbo/gen`

Stacked on https://github.com/vercel/turbo/pull/4871 and should be reviewed / merged prior to this PR


packages/turbo-gen/README.md
```diff
@@ -0,0 +1,66 @@
...
+```bash
+@turbo/gen generate [generator-name]
+```
+
+### Writing Generators
+
+`@turbo/gen` will search the root of your monorepo, and every workspace for generators defined at:
+
+```bash
+generators/config.js
```
mehulkar, line 63: should this live at a turbo-namespaced directory?

packages/turbo-gen/src/utils/getWorkspaceStructure.ts
```diff
@@ -0,0 +1,56 @@
...
+export function getWorkspaceStructure({
+  project,
+}: {
+  project: Project;
+}): WorkspaceStructure {
+  // get the workspace roots first, any assumptions we make
+  // should at least be based around configured workspaces
+  const roots = getWorkspaceRoots({ project });
+  const hasRootApps = roots.includes("apps");
+  const hasRootPackages = roots.includes("packages");
```
mehulkar, line 33: maybe pull these "known" directories into a constant somewhere so it's a little easier to track down/ customize if we need to. E.g. if someone has `sites/` for their apps

packages/turbo-gen/src/utils/plop.ts
```diff
@@ -0,0 +1,222 @@
...
+// TODO: Support a TS config file
+const TURBO_GENERATOR_CONFIG = path.join("generators", "config.js");
+
+// support root plopfile so that users with existing configurations can use them immediately
+const DEFAULT_CONFIG_LOCATIONS = [
+  TURBO_GENERATOR_CONFIG,
+  "plopfile.js",
+  "plopfile.cjs",
+  "plopfile.mjs",
+];
```
mehulkar, line 17: Should these additional config files be in the readme also?

packages/turbo-gen/src/utils/workspaceRoots.ts
```diff
@@ -0,0 +1,32 @@
...
+      return;
+    } else {
+      const globParts = glob.split("/");
+      const globRoot = globParts[0];
+      workspaceRoots.add(globRoot);
+    }
+  });
+
+  return Array.from(workspaceRoots);
+}
```
mehulkar, line 32: Is this logic something that could go in `@turbo/workspaces`?

mehulkar (approved):

packages/turbo-gen/README.md
```diff
@@ -0,0 +1,66 @@
+# `@turbo/gen`
+
+> This package is currently in **beta**. Please report any issues you encounter, and give us feedback about your experience using it!
+
+Easily extend your Turborepo with new apps, and packages. Create new empty workspaces, copy existing workspaces, add workspaces from remote sources (just like `create-turbo`!).or run custom generators defined using [Plop](https://plopjs.com/) configurations.
```
NicholasLYang, line 5:
```suggestion
Easily extend your Turborepo with new apps, and packages. Create new empty workspaces, copy existing workspaces, add workspaces from remote sources (just like `create-turbo`!) or run custom generators defined using [Plop](https://plopjs.com/) configurations.
```

packages/turbo-gen/src/utils/gatherAddRequirements.ts
```diff
@@ -0,0 +1,51 @@
...
+  opts,
+}: TurboGeneratorArguments) {
+  let source: Workspace | undefined = undefined;
+  const { answer: what } = await prompts.what({ override: opts.what });
+
+  // suggestion for the name based on the (optional) example path
+  const suggestion =
+    opts.examplePath?.split("/").pop() ||
+    opts.example?.split("/").pop() ||
+    undefined;
```
NicholasLYang, line 17: What's the reason for this extra `undefined`?

packages/turbo-gen/src/utils/getWorkspaceList.ts
```diff
@@ -0,0 +1,42 @@
...
+import inquirer from "inquirer";
+import {
+  getWorkspaceStructure,
+  getGroupFromWorkspace,
+} from "./getWorkspaceStructure";
+import { WorkspaceType } from "../generators/types";
+
+export function getWorkspaceList({
+  project,
+  what,
```
NicholasLYang, line 11: Not a huge fan of `what` as a variable name. Why not `workspaceType`?

NicholasLYang (approved):
I do wonder if at some point we'd consider porting this to Rust, but for now LGTM!

packages/turbo-gen/src/utils/getWorkspaceList.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (what === "app" && structure.hasRootApps) {
+      workspacesForDisplay = structure.workspacesByGroup.apps;
+    } else if (what === "package" && structure.nonAppWorkspaces.length > 0) {
+      workspacesForDisplay = structure.nonAppWorkspaces;
+    }
+  }
+
+  // build final list with separators between groups
+  let lastGroup: string | undefined;
+  workspacesForDisplay.forEach((workspace) => {
```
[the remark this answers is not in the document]
tknickman: Yea this is fair, I need to get in the habit of using `for...of`. I generally reach for `forEach` when I need to create a collection from an `iterable`, but not every item of the iterable will be included. Because the alternative is `.map(...).filter(...)` which is two iterations instead of the one from the `forEach`.

tknickman:
@NicholasLYang yea if we want to eventually do our own templating engine that is definitely the way to go - since we're building off of plop for this first crack at it we kind of need a node component so we can extend their API directly.

packages/turbo-gen/src/utils/workspaceRoots.ts
tknickman: Yea, this, and "categorize workspaces" (bucket workspaces by their parent directory) is something I find myself doing a ton so I think we should move it there.

packages/turbo-gen/src/utils/plop.ts
tknickman: Yes will definitely add these to docs

packages/turbo-gen/README.md
tknickman: Moved to turbo/generators/config.js

PR #4875, opened 2023-05-08
Submit originatingUser with Run

mehulkar:
This sets up `originationUser` to come from environment variables in Github and Vercel. Locally, we fallback to the `user.name` config from git. There are a few gaps here:

- `git config user.name` may or may not match anything in Vercel. I think that's ok for local Runs posted to Spaces. Since you have a direct link
- In the future, we likely want a personal token to post to Vercel. Currently, you just get a team token with `turbo link`. Once we've resolved that we can stop falling back to git, and then things will work. Incidentally, personal token is also how `vc deploy` associates the deploy with a Vercel user (as I learned from @TooTallNate)

Some additional open questions for follow up / after dogfooding: 

- We should submit "author of latest commit" and "submitter of Run" to be separate concepts. The UI has no affordance for this right now, so I'm not trying to handle that case. For runs sent to a Space that is linked to a git repo, the author of the commit can be looked up on the backend too, so this may be a larger design concern about non-git-connect Spaces.
- @gaspar09 mentioned the `originationUser` key name isn't great. we can change it later


cli/internal/runsummary/run_summary.go
```diff
@@ -68,6 +71,7 @@ type RunSummary struct {
 	FrameworkInference bool               `json:"frameworkInference"`
 	ExecutionSummary   *executionSummary  `json:"execution,omitempty"`
 	Tasks              []*TaskSummary     `json:"tasks"`
+	User               string             `json:"user"`
```
chris-olszewski, line 74: Do we want `"user": ""` in the JSON?

cli/internal/runsummary/spaces.go
```diff
@@ -28,9 +28,7 @@ type spacesRunPayload struct {
 	Client         spacesClientSummary `json:"client"`                   // Details about the turbo client
 	GitBranch      string              `json:"gitBranch"`
 	GitSha         string              `json:"gitSha"`
-
-	// TODO: we need to add these in
-	// originationUser string
+	User           string              `json:"originationUser"`
```
chris-olszewski, line 31: Same as above, do we want `"originationUser": ""` in the payload?

chris-olszewski (approved):
Looks good to me, just making sure that we want to be sending empty strings instead of undefined or null.

cli/internal/runsummary/run_summary.go
mehulkar: This only shows up in dry run json and run summary. I think empty string is ok and will make issues more obvious

cli/internal/runsummary/spaces.go
mehulkar: Miiight be a good candidate for omitempty her actually. @gaspar09 what do you think? I think if we omit it, the backend will use the token to associate with a person, which seems good.

PR #4880, opened 2023-05-09
refactor webpack loader execution

sokra:
### Description

* [fix error propagation in ContextCondition](https://github.com/vercel/turbo/commit/e0a8a54ad384698331717b423360531a6d52b6f5)
* [improve JSON deserialization error message](https://github.com/vercel/turbo/commit/b03233005c86ba39275eaf82fbd1c4d38c79bad4)
* [fix glob matching](https://github.com/vercel/turbo/commit/d868be41c5cef78551b7f1c6e369a4242151ef1b)
* allow to rename assets and apply pipeline on renamed asset

https://github.com/vercel/next.js/pull/49535

### DIFF
--- a/crates/turbo-tasks-fs/src/glob.rs
+++ b/crates/turbo-tasks-fs/src/glob.rs
@@ -50,8 +50,7 @@ impl Glob {
     pub fn execute(&self, path: &str) -> bool {
         let match_partial = path.ends_with('/');
         self.iter_matches(path, true, match_partial)
-            .next()
-            .is_some()
+            .any(|result| matches!(result, ("", _)))
     }
 
     fn iter_matches<'a>(
@@ -415,5 +414,15 @@ mod tests {
         println!("{glob:?} {path}");
 
         assert!(glob.execute(path));
+    }
+
+    #[rstest]
+    #[case::early_end("*.raw", "hello.raw.js")]
+    fn glob_not_matching(#[case] glob: &str, #[case] path: &str) {
+        let glob = Glob::parse(glob).unwrap();
+
+        println!("{glob:?} {path}");
+
+        assert!(!glob.execute(path));
     }
 }
--- a/crates/turbopack-core/src/ident.rs
+++ b/crates/turbopack-core/src/ident.rs
@@ -31,6 +31,16 @@ impl AssetIdent {
     pub fn add_asset(&mut self, key: StringVc, asset: AssetIdentVc) {
         self.assets.push((key, asset));
     }
+
+    pub async fn rename_as(&mut self, pattern: &str) -> Result<()> {
+        let root = self.path.root();
+        let path = self.path.await?;
+        self.path = root
+            .join(&pattern.replace("*", &path.path))
+            .resolve()
+            .await?;
+        Ok(())
+    }
 }
 
 #[turbo_tasks::value_impl]
@@ -92,6 +102,13 @@ impl AssetIdentVc {
     pub async fn with_part(self, part: ModulePartVc) -> Result<Self> {
         let mut this = self.await?.clone_value();
         this.part = Some(part);
+        Ok(Self::new(Value::new(this)))
+    }
+
+    #[turbo_tasks::function]
+    pub async fn rename_as(self, pattern: &str) -> Result<Self> {
+        let mut this = self.await?.clone_value();
+        this.rename_as(pattern).await?;
         Ok(Self::new(Value::new(this)))
     }
 
--- a/crates/turbopack-node/js/src/ipc/evaluate.ts
+++ b/crates/turbopack-node/js/src/ipc/evaluate.ts
@@ -49,7 +49,8 @@ export const run = async (
           const value = await getValue(ipc, ...msg.args);
           await ipc.send({
             type: "end",
-            data: value === undefined ? undefined : JSON.stringify(value),
+            data:
+              value === undefined ? undefined : JSON.stringify(value, null, 2),
           });
         } catch (e) {
           await ipc.sendError(e as Error);
--- a/crates/turbopack/src/lib.rs
+++ b/crates/turbopack/src/lib.rs
@@ -259,113 +259,129 @@ impl ModuleAssetContextVc {
     }
 
     #[turbo_tasks::function]
-    async fn process_default(
+    fn process_default(
         self_vc: ModuleAssetContextVc,
         source: AssetVc,
         reference_type: Value<ReferenceType>,
-    ) -> Result<AssetVc> {
-        let ident = source.ident().resolve().await?;
-        let options = ModuleOptionsVc::new(ident.path().parent(), self_vc.module_options_context());
+    ) -> AssetVc {
+        process_default(self_vc, source, reference_type, Vec::new())
+    }
+}
 
-        let reference_type = reference_type.into_value();
-        let part = match &reference_type {
-            ReferenceType::EcmaScriptModules(EcmaScriptModulesReferenceSubType::ImportPart(
-                part,
-            )) => Some(*part),
-            _ => None,
-        };
-        let mut current_source = source;
-        let mut current_module_type = None;
-        for rule in options.await?.rules.iter() {
-            if rule
-                .matches(source, &*ident.path().await?, &reference_type)
-                .await?
-            {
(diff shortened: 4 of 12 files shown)


crates/turbopack/src/lib.rs
```diff
@@ -259,113 +259,129 @@ impl ModuleAssetContextVc {
...
+            .await?
+        {
+            for effect in rule.effects() {
+                match effect {
+                    ModuleRuleEffect::SourceTransforms(transforms) => {
+                        current_source = transforms.transform(current_source);
+                        if current_source.ident().resolve().await? != ident {
+                            // The ident has been changed, so we need to apply new rules.
+                            let mut processed_rules = processed_rules.clone();
+                            processed_rules.push(i);
```
jridgewell, line 305: Can you leave a comment explaining this? I think this is to avoid an infinite loop for `foo.js` -> `foo.transformed.js` -> `foo.transformed.transformed.js`?

If we only want to exclude the transformer that _just_ ran, we could just use an index instead of a `Vec`.

crates/turbopack/src/module_options/mod.rs
```diff
@@ -469,10 +470,17 @@ impl ModuleOptionsVc {
...
                     ModuleRuleCondition::All(vec![
-                        ModuleRuleCondition::ResourcePathEndsWith(ext.to_string()),
+                        if !glob.contains("/") {
+                            ModuleRuleCondition::ResourceBasePathGlob(GlobVc::new(glob).await?)
+                        } else {
+                            ModuleRuleCondition::ResourcePathGlob {
+                                base: execution_context.project_path().await?,
+                                glob: GlobVc::new(glob).await?,
+                            }
+                        },
```
jridgewell, line 483: Nit: to avoid a negated negate.
```suggestion
                        if glob.contains("/") {
                            ModuleRuleCondition::ResourcePathGlob {
                                base: execution_context.project_path().await?,
                                glob: GlobVc::new(glob).await?,
                            }
                        } else {
                            ModuleRuleCondition::ResourceBasePathGlob(GlobVc::new(glob).await?)
                        },
```

crates/turbopack/src/module_options/rule_condition.rs
```diff
@@ -20,6 +20,18 @@ pub enum ModuleRuleCondition {
     ResourcePathInDirectory(String),
     ResourcePathInExactDirectory(FileSystemPathReadRef),
     ResourcePathRegex(#[turbo_tasks(trace_ignore)] Regex),
+    /// For paths that are within the same filesystem as the `base`, it need to
+    /// match the relative path from base to resource. This includes `./` or
+    /// `../` prefix. For paths in a different filesystem, it need to match
+    /// the resource path in that filesystem without any prefix. This means
+    /// any glob starting with `./` or `../` will only match paths in the
+    /// project. Globs starting with `**` can match any path.
```
jridgewell, line 28: Because we're forcing the path to be relative with a leading `./` or `../` with a `get_relative_path_to` call, I think this breaks a lot of cases? Like, I don't write my globs with `./foo/**/*.js`, I just write `foo/**/*.js`.

jridgewell (changes requested):

crates/turbopack-core/src/ident.rs
```diff
@@ -31,6 +31,16 @@ impl AssetIdent {
     pub fn add_asset(&mut self, key: StringVc, asset: AssetIdentVc) {
         self.assets.push((key, asset));
     }
+
+    pub async fn rename_as(&mut self, pattern: &str) -> Result<()> {
+        let root = self.path.root();
+        let path = self.path.await?;
+        self.path = root
+            .join(&pattern.replace("*", &path.path))
```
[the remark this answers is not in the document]
sokra: I think `*` would be easier to understand for the user:

```
"*.image.js": {
  loaders: ["image-generator-loader"],
  as: "*.png"
}
```

vs

```
"*.image.js": {
  loaders: ["image-generator-loader"],
  as: "{}.png"
}
```

crates/turbopack-node/js/src/ipc/evaluate.ts
```diff
@@ -49,7 +49,8 @@ export const run = async (
           const value = await getValue(ipc, ...msg.args);
           await ipc.send({
             type: "end",
-            data: value === undefined ? undefined : JSON.stringify(value),
+            data:
+              value === undefined ? undefined : JSON.stringify(value, null, 2),
```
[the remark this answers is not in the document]
sokra: We pass the string to `serde_json` which includes the source frame (of the JSON). It looks nicer when the source frame is actually readable, instead of being a single line of JSON.

crates/turbopack-core/src/ident.rs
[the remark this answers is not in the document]
jridgewell: Ahh, ok. Seeing this as a real config makes it more clear. Never mind.

crates/turbopack/src/lib.rs
sokra: You could build a cycle with multiple items:

```
"*.a": { as: "*.b" },
"*.b": { as: "*.c" },
"*.c": { as: "*.a" },
```

But it also handles valid cases like:

```
"*.optimize.*": { loader: "image-optimizer", as: "*" },
"*.minimize.*": { loader: "minimize-loader", as: "*" },
"*.svg": { loader: "svg-loader" }
```

`file.optimize.minimize.svg` need to have multiple `processed_rules`

crates/turbopack/src/module_options/rule_condition.rs
sokra: Yeah, I was also thinking this might be a source of errors.

Do you have a better idea for that?

These are the use cases I had in mind:

```
A: *.png - matching any file with that extension
B: ./*.png - matching png files in the project directory
C: ../other-package/**/*.png - matching png files in a different package
D: ./sub/dir/*.png - matching png files in a sub directory
E: **/dir/*.png - matching png files that are in "dir" anywhere
```

If we omit the prefix A and B are conflicting.

jridgewell: I think you're cases are correct, except for `D`. I would have written that `sub/dir/*.png`. We could make the globs relative if they contain a `/` and don't begin with `.`?

sokra: yes possible. Might be a bit unexpected in this edge case `{./,dir}/sub/*.png`, but I can live with that

crates/turbo-tasks-fs/src/glob.rs
```diff
@@ -84,6 +84,11 @@ impl Glob {
     }
 }
 
+struct GlobMatchResult<'a> {
```
jridgewell, line 87: Docs please, particularly for what `is_path_separator_equivalent` does.

crates/turbopack/src/module_options/mod.rs
```diff
@@ -462,13 +462,16 @@ impl ModuleOptionsVc {
...
                                 base: execution_context.project_path().await?,
                                 glob: GlobVc::new(glob).await?,
                             }
                         } else {
-                            ModuleRuleCondition::ResourceBasePathGlob(GlobVc::new(glob).await?)
+                            ModuleRuleCondition::ResourcePathGlob {
+                                base: execution_context.project_path().await?,
+                                glob: GlobVc::new(&format!("**/{glob}")).await?,
+                            }
                         },
```
jridgewell: Sorry, I think this is still incorrect. The examples being:

1. `./foo/bar.png`
2. `foo/bar.png`
3. `**/foo/bar.png`

1\. is definitely rooted in the current directory. 3\. can match any `foo/bar.png` in any directory. Currently 2\. matches 3\.'s behavior, but I think it should match 1\.'s behavior.

So we'd need 3 branches:

```rust
if glob.starts_with("./") || glob.starts_with("../") {
  // rooted
} else if glob.contains("/") {
  // prepend `./` to make it rooted
} else {
  // prepend `**/` to make it matcy any
}
```

sokra: yeah, you are right. I also want to handle starting with `**/` to not prepend `./` in that case

jridgewell (approved):

PR #4884, opened 2023-05-10
refactor(ecmascript): remove unwrap_module ast cloning

kwonoj:
### Description

Closing WEB-1024 on turbopack end.

### DIFF
--- a/crates/turbopack-ecmascript/src/transform/mod.rs
+++ b/crates/turbopack-ecmascript/src/transform/mod.rs
@@ -9,7 +9,7 @@ use swc_core::{
     base::SwcComments,
     common::{chain, util::take::Take, FileName, Mark, SourceMap},
     ecma::{
-        ast::{Module, ModuleItem, Program},
+        ast::{Module, ModuleItem, Program, Script},
         atoms::JsWord,
         preset_env::{self, Targets},
         transforms::{
@@ -221,7 +221,22 @@ impl EcmascriptInputTransform {
                     ..Default::default()
                 };
 
-                let module_program = unwrap_module_program(program);
+                let module_program = std::mem::replace(program, Program::Module(Module::dummy()));
+
+                let module_program = if let Program::Script(Script {
+                    span,
+                    mut body,
+                    shebang,
+                }) = module_program
+                {
+                    Program::Module(Module {
+                        span,
+                        body: body.drain(..).map(|stmt| ModuleItem::Stmt(stmt)).collect(),
+                        shebang,
+                    })
+                } else {
+                    module_program
+                };
 
                 *program = module_program.fold_with(&mut chain!(
                     preset_env::preset_env(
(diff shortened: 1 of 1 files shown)


crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -363,21 +363,6 @@ pub fn remove_shebang(program: &mut Program) {
...
-        Program::Module(module) => Program::Module(module.take()),
-        Program::Script(s) => Program::Module(Module {
-            span: s.span,
-            body: s
-                .body
-                .iter()
-                .map(|stmt| ModuleItem::Stmt(stmt.clone()))
-                .collect(),
-            shebang: s.shebang.clone(),
-        }),
```
sokra, line 377: Just noticing that the behavior is not the same. `unwrap_module_program` converts `script` to `module`. We need to keep that behavior

sokra (changes requested):

crates/turbopack-ecmascript/src/transform/mod.rs
kwonoj: Ok, converted to use visitMut instead: 

https://github.com/vercel/turbo/pull/4884/files#diff-b1b0731c098d856a121581436124b8fd726411b872ed875aacce400c825b0fbfR375

jridgewell (approved):

crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -221,7 +221,11 @@ impl EcmascriptInputTransform {
                     ..Default::default()
                 };
 
-                let module_program = unwrap_module_program(program);
+                let mut module_program =
+                    std::mem::replace(program, Program::Module(Module::dummy()));
+
+                let mut unwrap_visitor = UnwrapModuleProgram;
+                module_program.visit_mut_with(&mut unwrap_visitor);
```
sokra, line 228:
```suggestion
                let mut program =
                    std::mem::replace(program, Program::Module(Module::dummy()));

                let module_program = if let Program::Script(s) = program {
                    Program::Module(Module {
                        span: s.span,
                        body: s
                            .body
                            .iter()
                            .map(|stmt| ModuleItem::Stmt(stmt.clone()))
                            .collect(),
                        shebang: s.shebang.clone(),
                    })
                } else {
                    program
                }
```

crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -355,18 +359,23 @@ pub fn remove_shebang(program: &mut Program) {
...
+
+    // If given program is a script, replace it with a module.
+    fn visit_mut_program(&mut self, program: &mut Program) {
+        if let Program::Script(s) = program {
+            *program = Program::Module(Module {
+                span: s.span,
+                body: s
+                    .body
+                    .iter()
+                    .map(|stmt| ModuleItem::Stmt(stmt.clone()))
```
sokra, line 374: It's still cloning...

sokra (changes requested):
You don't need a Visitor

sokra (approved):

PR #4888, opened 2023-05-10
make turbopack respect the order of extensions from ResolveOptions

ForsakenHarmony:
### Description

It would probably be a bit more efficient to split it up in `read_matches`, but getting that to work requires a bit more work to sort the results because it recursively calls itself with the same pattern
fix WEB-925


jridgewell (commented):
Can you explain the difference? This looks like exactly what the [`Alternatives` branch](https://github.com/vercel/turbo/blob/705b3851c3702206561b4ccc1c6159d0a408a9fa/crates/turbopack-core/src/resolve/mod.rs#L714-L720) in `resolve_internal` does

ForsakenHarmony:
> Can you explain the difference? This looks like exactly what the [`Alternatives` branch](https://github.com/vercel/turbo/blob/705b3851c3702206561b4ccc1c6159d0a408a9fa/crates/turbopack-core/src/resolve/mod.rs#L714-L720) in `resolve_internal` does

@jridgewell yes, I can probably reuse that

jridgewell (commented):
What I mean is that the old code used the `Alternatives` path. So it looks like you've just manually reimplemented the same code? Is there a difference that I'm not seeing?

ForsakenHarmony:
> 



> What I mean is that the old code used the `Alternatives` path. So it looks like you've just manually reimplemented the same code? Is there a difference that I'm not seeing?
@jridgewell the old code is using `Pattern::alternatives`, not `Request::alternatives`

jridgewell (approved):
Ohhhh. Strange that they would behave differently? What was the bug?

ForsakenHarmony:
@jridgewell `Pattern::alternatives` get handled in `read_matches`, but that doesn't respect the order

The bug I was looking into was `posts.json` being resolved instead of `posts.tsx` because it just listed the directory alphabetically

jridgewell:
I think it might be the `read_dir` in `read_matches` returning alphabetical order? Should we fix that code to iterate the pattern alternatives instead of the dir entries?

ForsakenHarmony:
@jridgewell my initial plan was to fix it, but preserving the order with the recursive calls is a bit annoying

Issue #4885, opened 2023-05-10
[turborepo] Hashing error: cannot find package-file hash for package


abrisco:
### What version of Turborepo are you using?

1.9.3

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

When I build incrementally, I run into the following error:
```
Hashing error: cannot find package-file hash for package#!.next/!!coverage/**!!dist/**!!next-env.d.ts!!node_modules/**!**/*.js*!**/*.js*!**/*.jsx!**/*.ts!**/*.tsx!.env*!prisma/schema.prisma!tsconfig*.json
```

It's triggered for seemingly random packages in my monorepo but that could just be the execution order of tasks being different and I'm running into the same subset of broken tasks just in different orders. Nuking the `node_modules/.cache` directory gets me building again. This is reproducible without using remote caching

### Expected Behavior

I expect to be able to run tasks incrementally without issue.

### To Reproduce

It seems simply running incrementally leads to this failure. Below is my `turbo.json` file. My `build` tasks are just `"build": "tsc"`, `test` is `"test": "jest"` and `image` is a docker build script which runs `turbo prune` to gather files that should go in the docker image being built. My theory is that running turbo recursively is what's leading to the issue.

```json
{
  "$schema": "https://turbo.build/schema.json",
  "pipeline": {
    "//#coverage": {
      "dependsOn": ["//#package-deps", "@example/lcov-merger#build"],
      "inputs": ["!**/node_modules/**", "**/coverage/lcov.info"],
      "outputMode": "new-only",
      "outputs": ["coverage/_coverage_report.dat"]
    },
    "//#package-deps": {
      "inputs": [
        ".npmignore",
        ".npmrc",
        ".nvmrc",
        "apps/*/*/package.json",
        "apps/*/package.json",
        "packages/*/*/package.json",
        "packages/*/*/prisma/schema.prisma",
        "pnpm-lock.yaml",
        "pnpm-workspace.yaml",
        "tools/*/package.json"
      ],
      "outputMode": "new-only",
      "outputs": ["pnpm-lock.yaml"]
    },
    "build": {
      "dependsOn": ["//#package-deps", "^build"],
      "inputs": [
        "!.next/**",
        "!.vercel/**",
        "!coverage/**",
        "!dist/**",
        "!next-env.d.ts",
        "!node_modules/**",
        ".env*",
        "**/*.js",
        "**/*.jsx",
        "**/*.ts",
[...]

PR #4891, opened 2023-05-10
fix: check if user passes absolute path to prune

chris-olszewski:
### Description

As discovered in #4885, we forcibly join `--out-dir` to the repo root even if it's absolute. This PR adds a check that skips the join if the user passes an absolute system path.

### Testing Instructions

Added failing integration test


mehulkar (approved):

abrisco:
Thank you so much for putting together this PR! Any idea when a release would go out that includes this change? I'm unfamiliar with the release cadence of Turborepo.

chris-olszewski:
@abrisco If this doesn't make this weeks patch release I'll make sure to release a canary early next week so you can use it.

PR #4902, opened 2023-05-11
use more accurate function return values for static analysis

cool-little-fish:
### Description

  During current static analysis process, turbopack will give all function a default `JsValue::Constant(ConstantValue::Undefined)` as one of possible return value of function, but function always has a return value when the last statement of function body is ReturnStatement.

Although there maybe be dead loop or throw error syntax inside of function, but that case shouldn't affect static analysis process. 

This PR add a check before setting the initial return values for function, which will make the static analysis process more accurate.
### Testing Instructions
check the diff result of snapshot.


sokra (approved):

sokra:
Thanks

PR #4905, opened 2023-05-11
Handle leading doublestars inside braces

gsoltis:
### Description

The check for leading doublestar capture groups was not brace-aware. Add an index tracking the start of the current option we're tracking inside braces, and use that, rather than the start of the glob for doublestar checks.

### Testing Instructions

Added new test cases for this behavior.


arlyon (approved):
Happy to merge this despite potentially switching to a different glob impl

PR #4910, opened 2023-05-11
Build turbo once to build JS packages before publishing

mehulkar:
In the latest release, the `Publish to NPM` step took 12 mins (out of total 49 minutes)[^1]. I think we can shave off some of that time by not building turbo for every package that needs to build.

We may not really even need to use turbo to build these package for publishing, but I don't want to mess with this too much.

[^1]: https://github.com/vercel/turbo/actions/runs/4952890439/jobs/8860263831#step:13:1


nathanhammond (approved):
I'm on board with this, but we should really get this whole thing into the dependency graph.

mehulkar:
@nathanhammond I'll keep chipping away at it :)

PR #4911, opened 2023-05-11
Run Rust and JS package tests before publish

mehulkar:
If we're going to run smoke tests before, we should go all the way. Go unit tests won't catch issues with the JS packages, etc

### DIFF
--- a/.github/workflows/turborepo-release.yml
+++ b/.github/workflows/turborepo-release.yml
@@ -113,7 +113,7 @@ jobs:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
           target: ${{ matrix.os.name }}
       - name: Run JS Package Tests
-        run: turbo run check-types test --filter=packages/ --color
+        run: turbo run check-types test --filter="./packages/*" --filter="\!@vercel/*" --color
 
   build-go-darwin:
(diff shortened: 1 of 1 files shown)


tknickman (approved):
I'm here for it

mehulkar:
@nathanhammond what do you mean? Opening a PR with the staging branch so it runs the PR triggered tests? Or do you mean just relying on the tests that run during PRs?

I'm ok with either conversation, [I'm not sure of the value of smoke tests](https://github.com/vercel/turbo/pull/4666) in general, but I do feel like _just_ running Go Unit tests by themselves doesn't seem that helpful.

Also, not sure what you meant by "normal PR workflow", but note that our PR Github workflows also only run on certain file changes, so they may not always be triggered when we want them to be.

mehulkar:
@nathanhammond will wait on your reply before merging (feel free to merge for me in your work hours if you end up not having any objections)

mehulkar:
Talked with @nathanhammond and there aren't any blocking concerns. There is _probably_ something we can do to use the CI workflow we have instead of a maintaining a separate smoke-test Github Workflow, but I'm not sure what exactly that is. We can circle back to it

PR #4913, opened 2023-05-12
Send task summaries as tasks finish

mehulkar:
This PR attempts to do three things:

- Make Spaces requests as soon as possible
- Ensure that first request to create the Run blocks all other requests (so we can get the runID)
- Not block task execution

### Reviewer notes

In order to accomplish these goals, I've extracted out a `spacesClient`, that will make/queue the requests.


cli/internal/runsummary/run_summary.go
```diff
@@ -124,10 +116,16 @@ func NewRunSummary(
...
 		singlePackage:      singlePackage,
 		shouldSave:         shouldSave,
-		apiClient:          apiClient,
-		spaceID:            spaceID,
 		synthesizedCommand: synthesizedCommand,
 	}
+
+	// Note: this sets up a bidirectional relationship
+	// rsm has a reference to the spacesClient, and spacesClient has a reference to rsm
+	rsm.spacesClient = newSpacesClient(spaceID, apiClient, ui, &rsm)
```
gsoltis, line 124: FWIW it looks like you don't need to create the circular relationship. You can pass `rsm` in in the place you're using it.

cli/internal/runsummary/spaces.go
```diff
@@ -1,13 +1,235 @@
...
+// This first request is the POST /run request. We need to block on it because
+// the response contains the run ID from the server, which we need to construct the
+// URLs of subsequent requests.
+func (c *spacesClient) start() {
+	if !c.enabled {
+		return
+	}
+
+	mu := sync.Mutex{}
+	firstReqDone := false
```
gsoltis, line 75: It might be easier to use an atomic [CompareAndSwap](https://pkg.go.dev/sync/atomic#CompareAndSwapInt32) here. Then you don't need the mutex logic, whoever manages to do the swap goes on to do the first request. Just note that due to our Go version we don't have the atomic types, just the atomic functions.

cli/internal/runsummary/spaces.go
```diff
@@ -1,13 +1,235 @@
...
+						c.makeRequest(r)
+						close(c.run.created) // close this channel to signal that other requests can proceed
+					}(req)
+
+				} else {
+					mu.Unlock()
+
+					// If this is not the first request, wait for the run to be created
+					<-c.run.created
+					go c.makeRequest(req)
```
gsoltis, line 97: I'm not I understand the intention here. Are you intending to limit to 8 concurrent requests? This line starts another goroutine to do the request, meaning you will have 8 goroutines reading from the requests channel and then spinning up a goroutine for each request. 

A goroutine-per-request is a totally reasonable Go thing to do, it's just weird to have 8 goroutines to kick them off.

If you intend to have 8 concurrent requests, drop the `go` statement here. 
If instead you intend to have a goroutine per request, I would have a single goroutine read from the channel. If the first request hasn't finished yet, buffer the requests. Once the first request finishes, fire off everything in the buffer and directly fire off anything else that comes from the channel.

cli/internal/runsummary/spaces.go
```diff
@@ -1,13 +1,235 @@
...
+}
+
+func newSpacesClient(spaceID string, api *client.APIClient, ui cli.Ui, rsm *Meta) *spacesClient {
+	return &spacesClient{
+		api:      api,
+		ui:       ui,
+		rsm:      rsm,
+		spaceID:  spaceID,
+		enabled:  spaceID != "",
+		requests: make(chan *spaceRequest), // TODO: give this a size based on tasks
```
gsoltis, line 55: Without a buffer here, you will block execution once you have request in flight equal to the number of processing goroutines you have. That might be ok, depending on what you are doing in the processing goroutines (See below)

cli/internal/runsummary/spaces.go
```diff
@@ -1,13 +1,235 @@
...
+		}
+	}
+
+	if c.spaceID == "" {
+		c.errors = append(c.errors, req.error("No spaceID found"))
+		return
+	}
+
+	if !c.api.IsLinked() {
+		c.errors = append(c.errors, req.error("Repo is not linked to a Space. Run `turbo link --target=spaces` first"))
```
gsoltis, line 131: Are we going to end up with a flood of these errors?

gsoltis (changes requested):
I think the goroutine setup needs some work. I don't think it matches what you describe in the comment for `start()`

cli/internal/runsummary/spaces.go
mehulkar: I started with this pattern:
https://github.com/vercel/turbo/blob/c1f3e666db38f466ae5d0f2008f76b3618260069/cli/internal/run/dry_run.go#L109-L111
and
https://github.com/vercel/turbo/blob/c1f3e666db38f466ae5d0f2008f76b3618260069/cli/internal/cache/async_cache.go#L36-L38

but after setting up 8 workers to process requests from the channel, I observed that once they are unblocked by the first request, they seem to become blocking code and block task execution. So I added another goroutine-per-request. 

Conversely, if I only made a goroutine per-request, I also observed some congestion because each request blocked the next one.

It's entirely possible I did something in each of these approaches, but this double go-routine approach is how I was able to land with the total execution time of my test request to actually be the same as without making these requests in the first place.

cli/internal/runsummary/spaces.go
mehulkar: No, I have a janky check that prevents the flood of errors, but I think it's a bit brittle, since it relies on an internal detail of how the request url is calculated for requests after the first one.

cli/internal/runsummary/spaces.go
mehulkar: @gsoltis I took another stab at this here, mind taking another look? https://github.com/vercel/turbo/pull/4924

cli/internal/runsummary/spaces.go
```diff
@@ -1,13 +1,258 @@
...
+	// Create a labeled statement so we can break out of the for loop more easily
+
+	// Setup a for loop that goes infinitely until we break out of it
+FirstRequest:
+	for {
+		// A select statement that can listen for messages from multiple channels
+		select {
+		// listen for new requests coming in
+		case req := <-c.requests:
+			if req == nil {
```
gsoltis, line 85: You shouldn't need this. What is probably happening is that you are closing `c.requests` before the first request has finished, causing you to continuously read the zero value from the channel. You can detect this via `case req, isOpen := <-c.requests`. In the case where `isOpen` is false, switch to something that waits for the first request, sends the pending requests, and then you're done.

cli/internal/runsummary/spaces.go
```diff
@@ -1,13 +1,258 @@
...
+
+func newSpacesClient(spaceID string, api *client.APIClient, ui cli.Ui) *spacesClient {
+	return &spacesClient{
+		api:      api,
+		ui:       ui,
+		spaceID:  spaceID,
+		enabled:  spaceID != "",
+		requests: make(chan *spaceRequest), // TODO: give this a size based on tasks
+		// Set a default, empty one here, so we'll have something downstream and not a segfault
+		run: &spaceRun{
```
gsoltis, line 55: This, combined with the `Unmarshal` call below, makes me a bit nervous. It appears to be mixing resources and data (channel and `ID`, `URL`), as well as potentially using uninitialized zero-values. It might be better to split the struct into resource management that needs lifecycle control (creating, closing the channel) and plain data (`ID`, `URL`). You shouldn't need extra concern about segfaulting because it should be the same logic as making sure you have valid values before using them. I _think_ in this case it's the same as not using the data until `created` is closed?

gsoltis (changes requested):

cli/internal/runsummary/spaces.go
mehulkar: Ahh, thanks for the tip, WaitGroup and channel close were in the wrong order 

(although I still don't get why it's reading the 0 value continuously... I thought it would read from the channel on when something writes to it)

cli/internal/runsummary/spaces.go
mehulkar: separated the two

cli/internal/runsummary/spaces.go
```diff
@@ -1,15 +1,259 @@
...
+
+	// Create a labeled statement so we can break out of the for loop more easily
+
+	// Setup a for loop that goes infinitely until we break out of it
+FirstRequest:
+	for {
+		// A select statement that can listen for messages from multiple channels
+		select {
+		// listen for new requests coming in
+		case req := <-c.requests:
```
mehulkar, line 82: handle the case where the only thing we get here was closed without any other requests being sent to the channel. something like this:

```suggestion
		case req, ok := <-c.requests:
		if !ok {
		  return // exit or whatever
		}
```

cli/internal/runsummary/spaces.go
```diff
@@ -1,15 +1,259 @@
...
+
+func (c *spacesClient) postTask(task *TaskSummary) {
+	if !c.enabled {
+		return
+	}
+
+	c.queueRequest(&spaceRequest{
+		method: "POST",
+		makeURL: func(self *spaceRequest, run *spaceRun) error {
+			if run.ID == "" {
```
gsoltis, line 208: Are you ever sending this request when there's no `run.ID`? I guess that would happen if the first request failed? 

There might be a better way to handle that scenario (checking `c.run.ID` after `c.runCreated` closes). Then, by the time you are calling `dequeueRequest`, you already know whether or not to expect an id to be present, and can short-circuit there.

cli/internal/runsummary/spaces.go
```diff
@@ -1,15 +1,259 @@
...
+	}
+
+	// and then continue listening for more requests as they come in until the channel is closed
+	for req := range c.requests {
+		go c.dequeueRequest(req)
+	}
+}
+
+func (c *spacesClient) makeRequest(req *spaceRequest) {
+	if !c.enabled {
```
gsoltis, line 111: Does this need to be here? How would you get this far?

If you're concerned, another option is to have a different, empty struct implement the same methods and use that instead of `spacesClient` if spaces is not enabled. Similar to how the [NoOpOutputWatcher](https://github.com/vercel/turbo/blob/3b22f1cf42ec31857c4924bff01ed3ed1fc61502/cli/internal/runcache/output_watcher.go#L18) works.

cli/internal/runsummary/spaces.go
```diff
@@ -1,15 +1,259 @@
...
+			return
+		}
+	}
+
+	if c.spaceID == "" {
+		c.errors = append(c.errors, req.error("No spaceID found"))
+		return
+	}
+
+	if !c.api.IsLinked() {
```
gsoltis, line 138: Maybe this ought to be checked earlier in the process? Like, maybe consider spaces disabled if the api client is not linked?

cli/internal/runsummary/spaces.go
```diff
@@ -1,15 +1,259 @@
...
+	// for every request that fails. On the other hand, if that POST /run request fails, and N
+	// requests fail after that as a consequence, it is ok to print all of those errors.
+	if req.makeURL != nil {
+		if err := req.makeURL(req, c.run); err != nil {
+			c.errors = append(c.errors, err)
+			return
+		}
+	}
+
+	if c.spaceID == "" {
```
gsoltis, line 133: I don't think this can happen? We check enabled, and enabled is driven by having a space ID

gsoltis (approved):
Shipfix: Assuming we don't switch whether or not spaces is enabled or disabled in the middle of a run, I think this could benefit from checking once, as early as possible, and then either short-circuiting or handling requests as appropriate without continually checking. It would probably cut down on the error handling as well.

cli/internal/runsummary/spaces.go
mehulkar: I can do that, but I'd need that additional bookkeeping, and it would be "far" enough away from this makeURL function that I'd want this check here anyway.

PR #4918, opened 2023-05-12
Re-throw module.error when instantiating runtime modules

alexkirsz:
### Description

Runtime modules can be executed multiple times within the same context. However, when such a module encounters a runtime error during execution, this error is cached in module.error and not properly re-thrown when the module is required again from `getOrInstantiateRuntimeModule`, which means that on later runs we’ll silently fail to instantiate the module, and run into any errors that expects a global side-effect from the module.

link WEB-1045

### DIFF
--- a/crates/turbopack-dev/js/src/runtime.js
+++ b/crates/turbopack-dev/js/src/runtime.js
@@ -482,8 +482,9 @@ function instantiateModule(id, source) {
       break;
   }
 
-  runModuleExecutionHooks(module, (refresh) => {
-    try {
+  // NOTE(alexkirsz) This can fail when the module encounters a runtime error.
+  try {
+    runModuleExecutionHooks(module, (refresh) => {
       moduleFactory.call(module.exports, {
         e: module.exports,
         r: commonJsRequire.bind(null, module),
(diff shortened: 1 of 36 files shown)


crates/turbopack-dev/js/src/runtime.js
```diff
@@ -530,21 +531,26 @@ function runModuleExecutionHooks(module, executeModule) {
...
+      registerExportsAndSetupBoundaryForReactRefresh(
+        module,
+        globalThis.$RefreshHelpers$
+      );
+    }
+  } catch (e) {
+    throw e;
+  } finally {
+    // Always cleanup the intercept, even if module execution failed.
+    cleanupReactRefreshIntercept();
```
alexkirsz, line 552: This change isn't related to the issue, but since we don't expect the runtime to completely crash and burn after a module instantiation error occurs, we should still clean up any side-effects.

crates/turbopack-dev/js/src/runtime.js
```diff
@@ -1418,6 +1424,9 @@ function instantiateRuntimeModule(moduleId, chunkPath) {
 function getOrInstantiateRuntimeModule(moduleId, chunkPath) {
   const module = moduleCache[moduleId];
   if (module) {
+    if (module.error) {
+      throw module.error;
+    }
```
alexkirsz, line 1429: This is the actual fix.

ForsakenHarmony (approved):

PR #4634, opened 2023-04-19
Use cacheitem to restore HTTP cache.

nathanhammond:
This change causes `http_cache` to use `cacheitem` to restore tar files.

PR #4919, opened 2023-05-12
Move http cache writing to cacheitem.

nathanhammond:
The HTTP cache creates its own tar files that aren't `cacheitem`. This migrates them to use `cacheitem`, same as filesystem cache.

### DIFF
--- a/cli/internal/cache/cache_http.go
+++ b/cli/internal/cache/cache_http.go
@@ -4,24 +4,16 @@
 package cache
 
 import (
-	"archive/tar"
 	"bytes"
 	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
-	log "log"
 	"net/http"
-	"os"
-	"path/filepath"
 	"strconv"
-	"time"
-
-	"github.com/DataDog/zstd"
 
 	"github.com/vercel/turbo/cli/internal/analytics"
 	"github.com/vercel/turbo/cli/internal/cacheitem"
-	"github.com/vercel/turbo/cli/internal/tarpatch"
 	"github.com/vercel/turbo/cli/internal/turbopath"
 )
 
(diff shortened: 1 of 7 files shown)


cli/internal/cacheitem/cacheitem.go
```diff
@@ -31,7 +31,7 @@ type CacheItem struct {
 	tw         *tar.Writer
 	zw         io.WriteCloser
 	fileBuffer *bufio.Writer
-	handle     io.Reader
+	handle     interface{}
```
nathanhammond, line 34: Caches are `Reader | Writer | Closer` but you can't express that in Go. So this goes to `interface{}` and each place that needs it does a type assertion.

#4634 made this partially generic, this finishes the work.

cli/internal/cache/cache_http.go
```diff
@@ -79,69 +67,28 @@ func (cache *httpCache) Put(_ turbopath.AbsoluteSystemPath, hash string, duratio
...
-	defer func() { _ = w.Close() }()
-	zw := zstd.NewWriter(w)
-	defer func() { _ = zw.Close() }()
-	tw := tar.NewWriter(zw)
-	defer func() { _ = tw.Close() }()
-	for _, file := range files {
-		// log.Printf("caching file %v", file)
-		if err := cache.storeFile(tw, file); err != nil {
-			log.Printf("[ERROR] Error uploading artifact %s to HTTP cache due to: %s", file, err)
-			// TODO(jaredpalmer): How can we cancel the request at this point?
```
nathanhammond, line 97: This is fixed by using the channel. Previously failing here didn't actually blow up the artifact.

cli/internal/cache/cache_http.go
```diff
@@ -79,69 +67,28 @@ func (cache *httpCache) Put(_ turbopath.AbsoluteSystemPath, hash string, duratio
...
-		}
-	}
-}
+func (cache *httpCache) write(w io.WriteCloser, anchor turbopath.AbsoluteSystemPath, files []turbopath.AnchoredSystemPath, cacheErrorChan chan error) {
```
nathanhammond, line 80: This is now 1:1 with `cache_fs`'s implementation.

nathanhammond (commented):
Reviewer's guide!

gsoltis (approved):

PR #4920, opened 2023-05-12
feat(docs): add generators for release posts

tknickman:
### Description

Use our own generators for release blog posts

### DIFF
--- a/docs/turbo/generators/utils.js
+++ b/docs/turbo/generators/utils.js
@@ -1,0 +1,51 @@
+const axios = require("axios");
+
+const PUBLIC_TB_TOKEN =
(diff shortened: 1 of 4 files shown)


docs/turbo/generators/utils.js
```diff
@@ -0,0 +1,51 @@
+const axios = require("axios");
```
mehulkar, line 1: can we stop using axios for new code and just use native fetch or undici? 😢  (we need to upgrade to node 18 too, so maybe undici till that happens)

mehulkar (approved):
Can you add something in CONTRIBUTING.md or maybe docs/README.md on how to use the generator? I know I should know the API, but I probably won't remember :D

docs/turbo/generators/utils.js
tknickman: Yea fair we do need to rip it all out soon

tknickman:
Docs added!

PR #4921, opened 2023-05-12
fix(turbo/gen): fix generate name arg

tknickman:
### Description

Fix a bug with passing a generator name arg directly. Also add a better error message and fallback when a generate name is passed directly and it doesn't exist.

### DIFF
--- a/packages/turbo-gen/src/commands/generate/prompts.ts
+++ b/packages/turbo-gen/src/commands/generate/prompts.ts
@@ -1,5 +1,6 @@
 import inquirer from "inquirer";
 import type { Generator } from "../../utils/plop";
+import { logger } from "@turbo/utils";
 
 export async function customGenerators({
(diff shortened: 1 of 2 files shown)


packages/turbo-gen/src/commands/generate/prompts.ts
```diff
@@ -8,24 +9,26 @@ export async function customGenerators({
...
+      generators.find(
+        (g) => !(g instanceof inquirer.Separator) && g.name === generator
+      )
+    ) {
+      return {
+        selectedGenerator: generator,
+      };
+    }
+
+    logger.warn(`Generator "${generator}" not found`);
```
tknickman, line 23: If you passed a generator name that is invalid, we warn here, and then send you to the picker.

mehulkar (approved):

PR #4923, opened 2023-05-12
fix(cli): example should imply copy

tknickman:
### Description

Example should imply copy.


mehulkar (approved):

tknickman:
I hunted for a way to do this in clap directly, but couldn't find anything. @NicholasLYang if you know of a better way I'm happy to do it that way!

PR #4924, opened 2023-05-12
Simplify goroutine setup for async task requests

mehulkar:
This change works, but I see that the total execution time jumps from ~500ms to ~2s though (both FULL TURBO). I can share the test I'm using (in front). 

The execution time shouldn't change by much, because execution is closed off _before_ the `wg.Wait()`, so I don't think this diff is the best either

### DIFF
--- a/cli/internal/runsummary/spaces.go
+++ b/cli/internal/runsummary/spaces.go
@@ -191,10 +191,6 @@ func (c *spacesClient) createRun(rsm *Meta) {
 		// handler for when the request finishes. We set the response into a struct on the client
 		// because we need the run ID and URL from the server later.
 		onDone: func(req *spaceRequest, response []byte) {
-			if response == nil {
-				return
-			}
-
 			if err := json.Unmarshal(response, c.run); err != nil {
 				c.errors = append(c.errors, req.error(fmt.Sprintf("Error unmarshaling response: %s", err)))
 			}
--- a/cli/internal/runsummary/spaces.go
+++ b/cli/internal/runsummary/spaces.go
@@ -191,9 +191,6 @@ func (c *spacesClient) createRun(rsm *Meta) {
 		// handler for when the request finishes. We set the response into a struct on the client
 		// because we need the run ID and URL from the server later.
 		onDone: func(req *spaceRequest, response []byte) {
-			// close the run.created channel, because all other requests are blocked on it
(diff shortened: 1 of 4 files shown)


gsoltis:
Not quite what I meant. You're seeing a slowdown because the channel is unbuffered: writers must wait for an available reader. While the first request is pending, no one can write to the channel. So, the goal is to read from the channel as fast as possible and not block reading on IO. Pseudocode of one way to do it:

```go
func start() {
  firstRequestDone := make(chan interface{})
	go func() {
		pending := []*request{}
		startedFirstRequest := false
	FirstRequest:
		for {
			select {
			case <-firstRequestDone:
				for _, req := range pending {
					go doReq(req)
				}
				break FirstRequest
			case req <- requestsChannel:
				if startedFirstRequest {
					pending = append(pending, req)
				} else {
					startedFirstRequest = true
					doFirstReq(req)
				}
			}
		}
		for _, req := range requestsChannel {
			go doReq(req)
		}
	}()
}
```

In this case, you have a single reader that never blocks. When it gets the first request, it starts it. While the first request is in flight, it buffers subsequent requests. When the first request completes, it fires off the buffered requests and switches to reading and immediately firing off requests

cli/internal/runsummary/spaces.go
```diff
@@ -69,36 +69,47 @@ func (c *spacesClient) start() {
...
+
+	// Setup a for loop that goes infinitely until we break out of it
+FirstRequest:
+	for {
+		// A select statement that can listen for messages from multiple channels
+		select {
+		// listen for new requests coming in
+		case req := <-c.requests:
+			if req == nil {
+				continue
```
mehulkar, line 86: I added this check to prevent a bunch of `nil`s getting added to the `pending` slice, but I'm not sure why/how nil messages are being passed to this channel. My guess is it has something to do with the `for` loop, but I'm not sure.

mehulkar:
@gsoltis thanks! this mostly worked. Only thing I don't understand he requests channel receives a ton of `nil` messages and not sure where they're coming from.

I also don't quite get the `for`+`select` combo pattern, but I see it's a common way to listen for messages, so I'll read up on that.

mehulkar:
Going to merge this into the base PR for a holistic look and put in any questions there.

PR #4928, opened 2023-05-13
Remove fallback to configured git user

mehulkar:
When we don't have a known CI environment variable we can rely on the token that is used to submit the Run to associate the user. The backend already handles this case


nathanhammond (approved):
Thanks for circling back to this!

PR #4929, opened 2023-05-13
ci: parallelize Examples tests with turbo

mehulkar:
This is an alternative approach to #4190. There is more boilerplate, but it models each test of an example as its own workspace, so turbo can parallelize its execution. If when/we turn on remote caching in CI that will help even more.

## Reviewer notes (please read 😄 )

- Even though we have turbo.json configuration for what contributes to the cache key, we are still using the Github Actions to limit when this runs to changes in `examples/` and `examples-tests/**`. Since these tests do not exercise a locally built turbo, this is fine. In the future we may want to run examples tests against CLI changes as well, but we will cross that bridge when we get there.
- Because the `examples/` directory itself is a single workspace, any change to any example will cause a cache miss on _all_ examples-tests. This is not ideal, but I'm not attempting to solve that here.
- There are some funky tests that don't really make sense to me (e.g. `npm` with the `with-yarn`). I'm not planning to change/delete those in this PR, unless I've translated the matrix wrong.
- We are now validating the output of turbo. I had to do this mostly for debugging issues with this PR, but ended up keeping it. The disadvantage is that if a new version of turbo is published and changes this output, we won't know until the next time these tests are triggered. This exacerbates an existing problem, so I'm noting here, but I think it's actually a good problem to surface when it happens. We could consider running these examples tests on publishing a new version.


examples-tests/pnpm-gatsby/test.t
```diff
@@ -0,0 +1,29 @@
...
+  
+  \> with-gatsby@0.0.0 build (.*)/test.t (re)
+  \> turbo build "lint" "--output-logs=none" (re)
+  
+  • Packages in scope: docs, eslint-config-custom, tsconfig, ui, web (esc)
+  • Running build, lint in 5 packages (esc)
+  • Remote caching disabled (esc)
+  
+   Tasks:    5 successful, 5 total
+  Cached:    3 cached, 5 total
```
mehulkar, line 26: Not going to try to fix this here, but interesting that `with-gatsby` is not getting Full Turbo on the second run

mehulkar:
Closing in favor of #5011, because of so much noise from automation in this PR

PR #4941, opened 2023-05-15
create-turbo: automatic git configuration.

nathanhammond:
Not all environments where `create-turbo` is run will have a preconfigured `git` config. This ensures that we always have an author config for the initial commit.

It also reduces the number of processes executed to reduce the chances of intermittent failures.

### DIFF
--- a/packages/create-turbo/src/utils/git.ts
+++ b/packages/create-turbo/src/utils/git.ts
@@ -51,7 +51,6 @@ export function isInMercurialRepository(): boolean {
 export function tryGitInit(root: string, message: string): boolean {
   let didInit = false;
   try {
-    execSync("git --version", { stdio: "ignore" });
     if (isInGitRepository() || isInMercurialRepository()) {
       return false;
     }
--- a/packages/create-turbo/src/utils/git.ts
+++ b/packages/create-turbo/src/utils/git.ts
@@ -60,10 +60,7 @@ export function tryGitInit(root: string, message: string): boolean {
 
     execSync("git checkout -b main", { stdio: "ignore" });
 
-    execSync("git add -A", { stdio: "ignore" });
-    execSync(`git commit -m "${message}"`, {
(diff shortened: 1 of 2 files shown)


packages/create-turbo/src/utils/git.ts
```diff
@@ -51,7 +51,6 @@ export function isInMercurialRepository(): boolean {
 export function tryGitInit(root: string, message: string): boolean {
   let didInit = false;
   try {
-    execSync("git --version", { stdio: "ignore" });
```
nathanhammond, line 54: This is unnecessary; just a bonus place where weird errors can creep in.

packages/create-turbo/src/utils/git.ts
```diff
@@ -78,12 +74,18 @@ export function tryGitInit(root: string, message: string): boolean {
...
 }
+
+function gitCommit(message: string) {
+  execSync(
+    `git commit --author="Turbobot <turbobot@vercel.com>" -am "${message}"`,
+    {
+      stdio: "ignore",
+    }
+  );
+}
```
nathanhammond, line 91: Factored out into a separate method so you get either the `try` behavior, or the `boolean` behavior as you see fit.

packages/create-turbo/src/utils/git.ts
```diff
@@ -78,12 +74,18 @@ export function tryGitInit(root: string, message: string): boolean {
...
+    gitCommit(message);
     return true;
   } catch (err) {
     return false;
   }
 }
+
+function gitCommit(message: string) {
+  execSync(
+    `git commit --author="Turbobot <turbobot@vercel.com>" -am "${message}"`,
```
nathanhammond, line 86: Hardcodes the automated commits to be from us: `Turbobot <turbobot@vercel.com>`

tknickman (approved):
Needs test updates, but lgtm

PR #4944, opened 2023-05-15
expose mdx `providerImportSource` option

ForsakenHarmony:
### Testing Instructions

fix WEB-1012

### DIFF
--- a/crates/turbopack/src/module_options/mod.rs
+++ b/crates/turbopack/src/module_options/mod.rs
@@ -70,6 +70,7 @@ impl ModuleOptionsVc {
             ref decorators,
             enable_mdx,
             enable_mdx_rs,
+            ref mdx_provider_import_source,
             ref enable_postcss_transform,
             ref enable_webpack_loaders,
             preset_env_versions,
@@ -427,6 +428,7 @@ impl ModuleOptionsVc {
                 preserve_jsx: false,
                 jsx_runtime,
                 jsx_import_source,
+                provider_import_source: mdx_provider_import_source.clone(),
             })
             .cell();
 
--- a/crates/turbopack/src/module_options/mod.rs
+++ b/crates/turbopack/src/module_options/mod.rs
@@ -70,7 +70,6 @@ impl ModuleOptionsVc {
             ref decorators,
             enable_mdx,
             enable_mdx_rs,
-            ref mdx_provider_import_source,
             ref enable_postcss_transform,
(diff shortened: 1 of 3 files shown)


crates/turbopack/src/module_options/mod.rs
```diff
@@ -71,6 +71,7 @@ impl ModuleOptionsVc {
             ref decorators,
             enable_mdx,
             enable_mdx_rs,
+            ref mdx_provider_import_source,
```
kwonoj, line 74: The other options like `enable_jsx`, we pass a struct to the options. Maybe we should do the same by accepting `MdxTransformOptions` instead of having `enable_mdx_*` and its options separately?

kwonoj: Yes, it was me that didn't expose way to configure `MdxTransformOptions`, since tried to model next.config.js's `mdxRs` boolean flag doesn't have other options. But since we have cases to configure options now, I think it'd better to make `enable_mdx_rs` to accept option itself instead. 

Other options works similar, caller can do 

```
MdxTransformOptions {
 import_source: 'somecustomvalue'
 ...Default::default()
}
```

ForsakenHarmony: the only good way to do it I see right now would be to create another struct with just 

```rs
struct MdxTransformOptions {
    import_source: Option<String>,
}
```

because there should be no way to set the other options as we still want to set those later in this function

kwonoj: Yes, for now I think it's ok to expose whole `MdxTransformOptions`, then manually overriding it like

```
if enable_mdx {
 MdxTransformOptions {
  ..other custom value
  // only allow to override provider_import_source for now
   provider_import_source: user_options.provider_import_source
 }  
}
```

I'm currently working on the moduleoptioncontext and one of the plan is extracting out this mdx setup to the caller side, since this is really up to next.js or other bundlers instead of turbopack sets up non-configurable default for it. 

Exposing the option itself will allow making those happen more easier I think.

ForsakenHarmony: Added a separate struct now 🙃

kwonoj (approved):

crates/turbopack/src/module_options/mod.rs
kwonoj: Thanks!

PR #4950, opened 2023-05-15
feat(turbo-gen): support TS config files

tknickman:
### Description

- Add support for TS generator config files
- Converts all of our generator configs to TS


docs/turbo/generators/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+  "p.eyJ1IjogIjAzYzA0Y2MyLTM1YTAtNDhhNC05ZTZjLThhMWE0NGNhNjhkZiIsICJpZCI6ICJmOWIzMTU5Yi0wOTVjLTQyM2UtOWIwNS04ZDZlNzIyNjEwNzIifQ.A3TOPdm3Lhmn-1x5m6jNvulCQbbgUeQfAIO3IaaAt5k";
+
+export async function releasePostStats(answers: Answers): Promise<string> {
+  const [starsResponse, downloadsResponse, timeSavedResponse] =
+    await Promise.all([
+      fetch("https://api.github.com/repos/vercel/turbo"),
+      fetch("https://api.npmjs.org/versions/turbo/last-week"),
+      fetch(
+        `https://api.us-east.tinybird.co/v0/pipes/turborepo_time_saved_ticker.json?token=${PUBLIC_TB_TOKEN}`
+      ),
```
tknickman, line 19: Just for you @mehulkar 😉

packages/turbo-gen/src/utils/plop.ts
```diff
@@ -1,27 +1,36 @@
...
 
 export type Generator = PlopGenerator & {
   basePath: string;
   name: string;
 };
 
+// init ts-node for plop to support ts configs
+register({
+  transpileOnly: true,
+});
```
tknickman, line 32: Plop does support TS configs, but it relies on ts-node being configured https://github.com/plopjs/plop/issues/192#issuecomment-567129666

docs/turbo/generators/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+          local_cache_minutes_saved: number;
+        }
+      ];
+    },
+  ]);
+
+  const totalMinutesSaved: number =
+    timeSavedData.data[0].remote_cache_minutes_saved +
+    timeSavedData.data[0].local_cache_minutes_saved;
+  const totalYearsSaved: number = Math.floor(totalMinutesSaved / 60 / 24 / 365);
```
chris-olszewski, line 40: I find grouping the divisors a little more readable. Maybe we should just have a `MINUTES_IN_YEAR` constant
```suggestion
  const totalYearsSaved: number = Math.floor(totalMinutesSaved / (60 * 24 * 365));
```

chris-olszewski (approved):

docs/turbo/generators/utils.ts
tknickman: 👍🏼 good call, will add in a follow up so I don't have to suffer through the rust rebuild from the lockfile change 😭

docs/turbo/generators/utils.ts
mehulkar: YESSSSSSSSS. Next up we drop `jest` for `node:test`

PR #4951, opened 2023-05-15
feat(generators): support starting from a template

tknickman:
### Description

If the custom generator command is run and no generator config exists, prompt the user to create one from a set of templates.

### DIFF
--- a/docs/turbo/generators/utils.ts
+++ b/docs/turbo/generators/utils.ts
@@ -6,6 +6,8 @@ interface Answers extends Object {
   turboYearsSaved: string;
 }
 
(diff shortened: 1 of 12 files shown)


docs/turbo/generators/utils.ts
```diff
@@ -6,6 +6,8 @@ interface Answers extends Object {
   turboYearsSaved: string;
 }
 
+const MINUTES_IN_YEAR = 60 * 24 * 365;
```
tknickman, line 9: cc @chris-olszewski added here from your prev comment!

mehulkar (approved):

Issue #4759, opened 2023-04-30
More clear language for `outputs have not changed since previous run`


maschwenk:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

This message is a bit confusing and there's no references to it in the documentation. My best understanding of it is something like this:

A depends on B. A & B have `build` tasks defined in the standard way in `turbo.json`

```json
"build": {
      "dependsOn": ["^build"],
      "outputs": ["dist/**"]
    },
```

If I change the README of B, and that does not result in a new output in `dist` (or more specifically the hash of the output) , then we don't bother running `build` on A. The presumption being if A depends on B's `outputs` and the `outputs` don't change, we don't need to rebuild A.

### Describe the solution you'd like

If the above is accurate, I wonder how we could change the language to make it clearer, or if it would suffice to link to documentation that explains it. If it isn't accurate, then that might suggest we should change the language anyway 😄 .

If my description is accurate, perhaps it could say something like:

```Output hash for <package name>#build is unchanged, dependent task(s) [list of dependent tasks] will not run```
```Output hash for package_a#build is unchanged, dependent task(s) package_a#build will not run```

but maybe that's untrue?

### Describe alternatives you've considered

Read the source code.

mehulkar:
This was confusing to me too, open to new language. This message means two things:

- There was a cache hit (i.e. inputs did not change)
- There is no need to _restore_ cache from the local `.turbo` folder (or `--cache-dir) or the remote cache, because the required outputs are already there. 

This is good because you save additional time from not having to unzip a tarball, and restore it in the place where you've configured `outputs` for the task. 

Does that make sense? If so, feel free to propose a different message or a place in the docs where this information can go!

mehulkar:
I'm updating this language in https://github.com/vercel/turbo/pull/4952, feel free to provide any feedback there.

mehulkar:
This was done in https://github.com/vercel/turbo/pull/5127

maschwenk:
@mehulkar nice! The change regarding this issue as well as the other subtle string changes make a lot of sense. Thank you! Sorry for dropping the thread here!

PR #4952, opened 2023-05-15
Include timeSaved metric when skipping cache check

mehulkar:
After this change, the `timeSaved` metric will be included in:

- `--dry=json` output
- `--summarize` file when the daemon does not need to restore cache
  - Data posted to Spaces

## Testing
I used this script to get the cache status from the `--summarize` flag for each real run:
```
FILE=$(/bin/ls .turbo/runs/*.json | tail -n1) && echo $FILE && cat $FILE | jq '.tasks | map(.taskId, .cache)'
```

```bash
# 1. confirm cache miss with 0 timeSaved
$ localturbo build --filter=<my-app> --summarize --skip-infer --force 

# 2. with both local and remote caches enabled, confirmed local cache hit with timeSaved
# this is because fs cache has a higher priority, so it should always be preferred
$ localturbo build --filter=<my-app> --summarize --skip-infer 

# 3. use --remote-only to skip fs cache, and confirmed remote cache hit with timeSaved
$ localturbo build --filter=<my-app> --summarize --skip-infer --remote-only

# Now that we've confirmed that both local and remote caches are populated

# 4. Remove token, so we can confirm remote cache hit created a local cache hit with timeSaved value
$ localturbo build --filter=<my-app> --summarize --skip-infer  --token=""

# 5. Turn of fs and remote caching with --token and remote-only flags, and confirmed cache miss with 0 timeSaved
$ localturbo build --filter=<my-app> --summarize --skip-infer --token="" --remote-only 
```

<details>
<summary>Internal test instuctions</summary>

Some copy-pasteable commands for internal `front` repo, since it has remote cache enabled and we can test there.
This won't work for anyone who doesn't have access to `front` repo (i.e. OSS readers)

```bash
FILE=$(/bin/ls .turbo/runs/*.json | tail -n1) && echo $FILE && cat $FILE | jq '.tasks | map(select(.taskId == "@pyra/promise-utils#build")) | .[0].cache'
```

```bash
# cache miss + 0 timeSaved
localturbo build --filter="@pyra/promise-utils" --summarize --skip-infer --force
# remote cache hit a non-zero timeSaved
localturbo build --filter="@pyra/promise-utils" --summarize --skip-infer --remote-only
# local cache hit a non-zero timeSaved
localturbo build --filter="@pyra/promise-utils" --summarize --skip-infer --token=""
# no cache enabled, but cache hit a non-zero timeSaved (cache restoration)
localturbo build --filter="@pyra/promise-utils" --summarize --skip-infer --token="" --remote-only
git clean -fdx && pnpm i
# cache miss + 0 timeSaved
localturbo build --filter="@pyra/promise-utils" --summarize --skip-infer --token="" --remote-only
```
</details>


cli/internal/cache/cache.go
```diff
@@ -271,31 +300,34 @@ func (mplex *cacheMultiplexer) Fetch(anchor turbopath.AbsoluteSystemPath, key st
...
+			_ = mplex.storeUntil(anchor, key, itemStatus.TimeSaved, actualFiles, i)
 
 			// If another cache had already set this to true, we don't need to set it again from this cache
-			combinedCacheState.Local = combinedCacheState.Local || itemStatus.Local
-			combinedCacheState.Remote = combinedCacheState.Remote || itemStatus.Remote
-			return combinedCacheState, actualFiles, duration, err
+			combinedCacheState.Hit = itemStatus.Hit
+			combinedCacheState.Source = itemStatus.Source
+			combinedCacheState.TimeSaved = itemStatus.TimeSaved
+			return combinedCacheState, actualFiles, nil
```
mehulkar, line 314: This logic change is a bit concerning. Previously, the `||` logic resulted in "first cache hit wins". Now it's "last one wins". But because there's a return statement, it's still "first one wins". So I _think_ we're effectively at the same result, but I could use another pair of eyes to confirm that I'm thinking about this right.

cli/internal/run/dry_run.go
```diff
@@ -113,7 +113,7 @@ func populateCacheState(turboCache cache.Cache, taskSummaries []*runsummary.Task
 			for index := range queue {
 				task := taskSummaries[index]
 				itemStatus := turboCache.Exists(task.Hash)
-				task.CacheSummary = runsummary.NewTaskCacheSummary(itemStatus, nil)
```
mehulkar, line 116: We lose this `nil` behavior. But now since now timeSaved is available everywhere (in theory), we don't need to distinguish from `0`.

cli/internal/runcache/runcache.go
```diff
@@ -155,8 +153,9 @@ func (tc *TaskCache) RestoreOutputs(ctx context.Context, prefixedUI *cli.Prefixe
 		}
 	} else {
 		// If no outputs have changed, that means we have a local cache hit.
-		cacheStatus.Local = true
-		prefixedUI.Warn(fmt.Sprintf("Skipping cache check for %v, outputs have not changed since previous run.", tc.pt.TaskID))
+		// We will still check to see whether the cache exists, so we can get the duration.
+		cacheStatus = tc.rc.cache.Exists(tc.hash)
```
mehulkar, line 157: This is the purpose of this PR. One thing that's weird here is before we were _assuming_ that `hasChangedOutputs == false` means cache hit. In order to get the timeSaved, we have to check the cache. That means:

1. the thing we end up returning is no longer an assumption, we're actually checking the existence of cache
2. the assumption can be wrong / different from what the actual cache status is (e.g. if the `hasChangedOutputs` implementation is wrong)

I considered doing this instead:

```suggestion
		cacheStatus = newFSTaskCacheStatus(true, 0)
		cacheStatus.TimeSaved = tc.rc.cache.Exists(tc.hash).TimeSaved
```

but I feel like that's an equally weird implementation, since we're combining two sources of truth.

Open to suggestions!

gsoltis: Can we store the time saved with the daemon alongside the globs? Then you either get "no change, here's time saved", check fs cache, or check http cache

cli/internal/cache/cache.go
```diff
@@ -32,15 +32,44 @@ type Cache interface {
...
+	}
 }
 
 const (
 	// CacheSourceFS is a constant to indicate local cache hit
 	CacheSourceFS = "LOCAL"
 	// CacheSourceRemote is a constant to indicate remote cache hit
 	CacheSourceRemote = "REMOTE"
+	// CacheSourceNone is an empty string because there is no source for a cache miss
+	CacheSourceNone = ""
```
NicholasLYang, line 72: Do we want to make this `MISS` just in case we log out ItemStatus?

cli/internal/runcache/runcache.go
```diff
@@ -110,12 +110,13 @@ type TaskCache struct {
...
 // Returns the cacheStatus, the timeSaved, and error values, so the consumer can understand
 // what happened in here.
-func (tc *TaskCache) RestoreOutputs(ctx context.Context, prefixedUI *cli.PrefixedUi, progressLogger hclog.Logger) (cache.ItemStatus, int, error) {
+func (tc *TaskCache) RestoreOutputs(ctx context.Context, prefixedUI *cli.PrefixedUi, progressLogger hclog.Logger) (cache.ItemStatus, error) {
 	if tc.cachingDisabled || tc.rc.readsDisabled {
 		if tc.taskOutputMode != util.NoTaskOutput && tc.taskOutputMode != util.ErrorTaskOutput {
 			prefixedUI.Output(fmt.Sprintf("cache bypass, force executing %s", ui.Dim(tc.hash)))
 		}
-		return cache.ItemStatus{Local: false, Remote: false}, 0, nil
+		// TODO: do we want a more speicic bypass status here?
```
NicholasLYang, line 118:
```suggestion
		// TODO: do we want a more specific bypass status here?
```

NicholasLYang (approved):

cli/internal/runcache/runcache.go
```diff
@@ -155,8 +153,9 @@ func (tc *TaskCache) RestoreOutputs(ctx context.Context, prefixedUI *cli.Prefixe
 		}
 	} else {
 		// If no outputs have changed, that means we have a local cache hit.
-		cacheStatus.Local = true
-		prefixedUI.Warn(fmt.Sprintf("Skipping cache check for %v, outputs have not changed since previous run.", tc.pt.TaskID))
+		// We will still check to see whether the cache exists, so we can get the duration.
+		cacheStatus = tc.rc.cache.Exists(tc.hash)
+		prefixedUI.Warn(fmt.Sprintf("Configured outputs are already in place for %s, skipping cache check", tc.pt.TaskID))
```
mehulkar, line 158: https://github.com/vercel/turbo/issues/4759

maschwenk, line 158:
```suggestion
		prefixedUI.Warn(fmt.Sprintf("Cache hit for %s; Configured outputs are already in place, skipping expanding cache into output directories", tc.pt.TaskID))
```
Pretty wordy, but based on your message in that channel, seems more communicative. It just feels weird to me to say "skipping cache check", because if this message is supposed to communicate there was a cache hit...why would we check a cache in that case?

mehulkar: "Skipping cache check" might be revealing the internal implementation, but it _is_ more correct than what you've proposed here. It's true that expanding the cache is skipped, but it's also true that we didn't even check the cache to see if there is a hit or not. If you deleted the cache, you'd still get this message.

I'll think about your suggestion for a bit though, thanks!

maschwenk: > "Skipping cache check" might be revealing the internal implementation, but it _is_ more correct than what you've proposed here. It's true that expanding the cache is skipped, but it's also true that we didn't even check the cache to see if there is a hit or not. If you deleted the cache, you'd still get this message.
> 
> I'll think about your suggestion for a bit though, thanks!

😆 I think that's where the terminology gets confusing. "Cache hit" as an outside dev to me could mean:

1. The inputs haven't changed and therefore the hit/miss is immaterial, we don't even need to look at the cache
2. The inputs haven't changed and therefore if we were to check the cache, we'd definitely get a hit

If I'm understanding, you are saying 1 is really what we want to communicate.

mehulkar: it may not be what we _want_ to communicate to an outside dev, but what's happening is two things:

1. you ran this task before, and we know you wrote outputs XYZ. We've been watching those outputs since then, and we _know_ they haven't changed.
2. we also know your _inputs_ didn't change, because we are starting with the same task hash.
3. so we know we can skip the cache check and the cache restoration

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -255,6 +258,9 @@ impl<T: Watcher + Send + 'static> proto::turbod_server::Turbod for DaemonServer<
     ) -> Result<tonic::Response<proto::NotifyOutputsWrittenResponse>, tonic::Status> {
         let inner = request.into_inner();
 
+        self.times_saved
+            .insert(Arc::new(inner.hash).to_string(), inner.time_saved);
```
chris-olszewski, line 262: I think this is what you want
```suggestion
            .insert(inner.hash.clone(), inner.time_saved);
```

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -58,6 +58,8 @@ pub struct DaemonServer<T: Watcher> {
     shutdown_rx: Option<Receiver<()>>,
 
     running: Arc<AtomicBool>,
+
+    times_saved: HashMap<String, u64>,
```
chris-olszewski, line 62:
```suggestion
    times_saved: Arc<Mutex<HashMap<String, u64>>>,
```

mehulkar:
Moving the daemon work into https://github.com/vercel/turbo/pull/5101

cli/internal/runcache/runcache.go
mehulkar: This is done now

cli/internal/runcache/runcache.go
mehulkar: There's enough to talk about that I'm moving it to its own PR

tknickman (approved):

mehulkar:
I've backed out some of the refactor work and other changes. Notably this introduces a couple limitations:

- On _remote_ cache hits, the cache source is still noted as local. This is a race condition issue and an issue with the merging of cache hit data structures, I think. I am pretty sure it's out of scope for this PR and it doesn't affect the timeSaved value
- It removes the `timeSaved` value from dry runs (will add back in separate PR)

I've added more test instructions above of how I tested this and I'm seeing values from the daemon when skipping cache restoration. 

Since I've gotten a couple approvals already and the diff is now smaller, no new code is added, I'm going to merge this!

PR #4966, opened 2023-05-16
Initial tracing implementation

sokra:
### Description

* adds a raw_trace subscriber that will emit the raw trace events into `.turbopack/trace.log` (resp. `.next/trace.log` for next.js)
* adds a CLI script which converts the raw trace into a chrome trace json file -> https://ui.perfetto.dev/
* adds a `TURBOPACK_TRACING` (resp. `NEXT_TURBOPACK_TRACING`) env var to enable tracing
* adds some presets e. g. `turbopack` or `next` to enable tracing for certain things.
* add tracing for invalidations

There are three different visualization modes:

#### `--single`

Shows all cpu time as it would look like when a single cpu would execute the workload.

(10 concurrent tasks that take 1s are shown as 10 tasks that take 1s with total time of 10s)

Pro:
* Concurrency is visualized by bar filling (dark filled bars -> too few concurrency)
* It injects pseudo bars with "cpus idle" for low concurrency (with `--idle`)

Con:
* Total time won't be represented correctly, since a single CPU would take longer

Use Case: Gives a good overview of slow tasks in a build.

#### `--merged`

Shows all cpu time scaled by the concurrency.

(10 concurrent tasks that take 1s are shown as 10 tasks that take 0.1s with total time of 1s)

Pro:
* Total time is represented correctly
* Low concurrent tasks are bigger
* Concurrency is visualized by bar filling (dark filled bars -> too few concurrency)
* It injects pseudo bars with "cpus idle" for low concurrency (with `--idle`)

Con:
* Individual tasks time won't be represented correctly.

Use Case: Gives a good overview why a build is slow overall.

#### `--threads`

Shows cpu time distributed on infinite virtual cpus/threads.

(10 concurrent tasks that take 1s are shown as 10 concurrent tasks that take 1s with total time of 1s)

Pro:
* Concurrency is shown via multiple CPU
* Most realistic visualization

Con:
* Hard to read


crates/turbo-tasks-macros/src/func.rs
```diff
@@ -154,9 +154,25 @@ pub fn gen_native_function_code(
...
-        quote! { #original_function(#(#input_arguments),*).await }
+        quote! {
+            turbo_tasks::macro_helpers::tracing::Instrument::instrument(
+                async {
+                    let turbo_tasks_result = #original_function(#(#input_arguments),*).await;
+                    turbo_tasks::macro_helpers::notify_scheduled_tasks();
+                    turbo_tasks_result
+                },
+                turbo_tasks::macro_helpers::tracing::info_span!(#name_code)
+            ).await
```
alexkirsz, line 165: Can we expose a turbo tasks API for this instead of doing this in the macro itself? We should strive to keep the macro output minimal.

e.g. `turbo_tasks::macro_helpers::instrument!(#name_code, #original_function(#(#input_arguments),*))`

crates/turbo-tasks-macros/src/func.rs
```diff
@@ -154,9 +154,25 @@ pub fn gen_native_function_code(
...
+        }
     } else {
-        quote! { #original_function(#(#input_arguments),*) }
+        quote! {
+            {
+                let _turbo_tasks_function_guard = turbo_tasks::macro_helpers::tracing::info_span!(#name_code).entered();
+                let turbo_tasks_result = #original_function(#(#input_arguments),*);
+                turbo_tasks::macro_helpers::notify_scheduled_tasks();
+                turbo_tasks_result
+            }
```
alexkirsz, line 174: Same comment.

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -802,13 +806,16 @@ impl Job {
                     .await;
             }
             Job::RemoveFromScopeQueue(queue, id) => {
+                let _guard = info_span!("RemoveFromScopeQueue::AddToScopeQueue").entered();
                 run_remove_from_scope_queue(queue, id, backend, turbo_tasks);
                 backend.scope_add_remove_priority.finish_high();
             }
             Job::UnloadRootScope(id) => {
+                let span = info_span!("RemoveFromScopeQueue::UnloadRootScope");
```
alexkirsz, line 814:
```suggestion
                let span = info_span!("Job::UnloadRootScope");
```

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -802,13 +806,16 @@ impl Job {
                     .await;
             }
             Job::RemoveFromScopeQueue(queue, id) => {
+                let _guard = info_span!("RemoveFromScopeQueue::AddToScopeQueue").entered();
```
alexkirsz, line 809:
```suggestion
                let _guard = info_span!("Job::RemoveFromScopeQueue").entered();
```

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -802,13 +806,16 @@ impl Job {
                     .await;
```
alexkirsz, line 806: Is this one missing a span?

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -817,6 +824,7 @@ impl Job {
                 }
             }
             Job::GarbageCollection => {
+                let _guard = info_span!("RemoveFromScopeQueue::GarbageCollection").entered();
```
alexkirsz, line 827:
```suggestion
                let _guard = info_span!("Job::GarbageCollection").entered();
```

crates/turbo-tasks/src/graph/visit.rs
```diff
@@ -63,10 +64,60 @@ where
...
-        VisitControlFlow::Continue(edge)
+        VisitControlFlow::Continue(edge, Span::current())
     }
 
     fn edges(&mut self, node: &Node) -> Self::EdgesFuture {
         (self)(node.clone())
     }
 }
+
+pub struct WithSpan<Node, Abort, Impl, VisitImpl, F>
```
alexkirsz, line 75: Comment this item.

crates/turbo-tasks/src/graph/with_future.rs
```diff
@@ -29,18 +32,23 @@ impl<T, H> Future for With<T, H>
...
     ) -> std::task::Poll<Self::Output> {
         let this = self.project();
+        let guard = this.span.enter();
         match this.future.poll(cx) {
-            std::task::Poll::Ready(result) => std::task::Poll::Ready((
-                this.handle.take().expect("polled after completion"),
-                result,
-            )),
+            std::task::Poll::Ready(result) => {
+                drop(guard);
```
alexkirsz, line 45: Isn't that equivalent to `instrument`ing the future directly? Why do we need to keep the span as part of this struct instead?

crates/turbo-tasks/src/graph/control_flow.rs
```diff
@@ -1,7 +1,9 @@
+use tracing::Span;
+
 /// The control flow of visiting an edge during a graph traversal.
 pub enum VisitControlFlow<Node, Abort = !> {
     /// The traversal should continue on the outgoing edges of the given node.
-    Continue(Node),
+    Continue(Node, Span),
```
alexkirsz, line 6: Can we have the knowledge of `Span` live at the visitor level instead? Where `Span` would be a property of the `Node`, and the visitor would require some `Node: WithSpan` bound.

Something like the following:

```
ReverseTopological::new().skip_duplicates().visit(
    [node],
    |node| {
        async move { ... }
    }.instrument()
)
```

crates/turbo-tasks/src/util.rs
```diff
@@ -220,3 +224,28 @@ impl<T: ?Sized + Debug + 'static> Debug for StaticOrArc<T> {
         (**self).fmt(f)
     }
 }
+
+pin_project! {
+    pub struct WrapFuture<F, W> {
```
alexkirsz, line 229: Comment this item.

crates/turbopack-node/src/evaluate.rs
```diff
@@ -415,6 +421,12 @@ async fn pull_operation(
     let mut file_dependencies = Vec::new();
     let mut dir_dependencies = Vec::new();
 
+    let node_execution_start = Instant::now();
+    let end_execution = || {
+        let duration = node_execution_start.elapsed().as_micros() as u64;
+        info!(name = "node.js evaluation", duration);
```
alexkirsz, line 427:
```suggestion
        info!(name = "Node.js evaluation", duration);
```

crates/turbopack-node/src/render/render_proxy.rs
```diff
@@ -258,9 +262,17 @@ async fn render_stream_internal(
         }
         operation.send(RenderProxyOutgoingMessage::BodyEnd).await?;
 
+        let entry = module.ident().to_string().await?;
+        let node_execution_start = Instant::now();
+        let end_execution = || {
+            let duration = node_execution_start.elapsed().as_micros() as u64;
+            info!(name = "node.js rendering", entry = display(entry), duration);
```
alexkirsz, line 269:
```suggestion
            info!(name = "Node.js rendering", entry = display(entry), duration);
```

crates/turbopack-node/src/render/render_proxy.rs
```diff
@@ -258,9 +262,17 @@ async fn render_stream_internal(
         }
         operation.send(RenderProxyOutgoingMessage::BodyEnd).await?;
 
+        let entry = module.ident().to_string().await?;
+        let node_execution_start = Instant::now();
+        let end_execution = || {
```
alexkirsz, line 267: This would be more elegantly represented by a span guard. ```let node_execution_span = duration_span!(Level::TRACE, "Node.js rendering");``` to be `drop`ped later.

You're reusing this pattern multiple times above and below, so it makes sense to abstract it.

crates/turbopack-cli-utils/src/exit.rs
```diff
@@ -0,0 +1,24 @@
+use std::sync::{Arc, Mutex};
+
+use anyhow::{Context, Result};
+
+pub struct CloseGuard<T>(Arc<Mutex<Option<T>>>);
```
alexkirsz, line 5: Comment this item.

crates/turbopack-cli-utils/src/raw_trace.rs
```diff
@@ -0,0 +1,173 @@
...
+use serde_json::Value;
+use tracing::{
+    field::{display, Visit},
+    span, Subscriber,
+};
+use tracing_subscriber::{fmt::MakeWriter, registry::LookupSpan, Layer};
+
+use crate::tracing::{FullTraceRow, TraceRow};
+
+pub struct RawTraceLayer<
```
alexkirsz, line 12: Comment this item.

crates/turbopack-cli-utils/src/raw_trace.rs
```diff
@@ -0,0 +1,173 @@
...
+            let mut buf = match borrow {
+                Ok(buf) => {
+                    a = buf;
+                    &mut *a
+                }
+                _ => {
+                    b = Vec::new();
+                    &mut b
+                }
+            };
```
alexkirsz, line 47: Add a comment to explain what you're doing here and why.

crates/turbopack-cli-utils/src/raw_trace.rs
```diff
@@ -0,0 +1,173 @@
...
+        ctx: tracing_subscriber::layer::Context<'_, S>,
+    ) {
+        let mut values = ValuesVisitor::new();
+        attrs.values().record(&mut values);
+        self.write(TraceRow::Start {
+            id: id.into_u64(),
+            parent: if attrs.is_contextual() {
+                ctx.current_span().id().map(|p| p.into_u64())
+            } else {
+                attrs.parent().map(|p| p.into_u64())
```
alexkirsz, line 80: The `.map` could probably move out of the `if` here.

crates/turbopack-cli-utils/src/raw_trace.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    fn on_event(&self, event: &tracing::Event<'_>, ctx: tracing_subscriber::layer::Context<'_, S>) {
+        let mut values = ValuesVisitor::new();
+        event.record(&mut values);
+        self.write(TraceRow::Event {
+            parent: if event.is_contextual() {
+                ctx.current_span().id().map(|p| p.into_u64())
+            } else {
+                event.parent().map(|p| p.into_u64())
+            },
```
alexkirsz, line 112: Same comment ^

crates/turbopack-cli-utils/src/tracing.rs
```diff
@@ -0,0 +1,43 @@
+use serde::{Deserialize, Serialize};
+use serde_json::Map;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct FullTraceRow<'a> {
```
alexkirsz, line 5: Comment these items.

crates/turbopack-convert-trace/src/main.rs
alexkirsz, line 1: Would appreciate a high level overview of what this utility does, perhaps as a mod-level comment here.

alexkirsz (changes requested):
Change the level of all spans and events here from `Level::Info` to `Level::Trace`.

crates/turbopack-node/src/source_map/mod.rs
```diff
@@ -311,6 +312,7 @@ impl StructuredError {
     }
 }
 
+#[instrument(skip_all)]
```
ForsakenHarmony, line 315: Shouldn't the whole tracing impl be behind a feature?

crates/turbopack-cli/src/main.rs
```diff
@@ -1,28 +1,77 @@
...
+            }
+            "turbo-tasks" => {
+                trace = "turbopack=trace,turbopack_core=trace,turbopack_ecmascript=trace,\
+                         turbopack_css=trace,turbopack_dev=trace,turbopack_image=trace,\
+                         turbopack_dev_server=trace,turbopack_json=trace,turbopack_mdx=trace,\
+                         turbopack_node=trace,turbopack_static=trace,turbopack_cli_utils=trace,\
+                         turbopack_cli=trace,turbopack_ecmascript=trace,turbo_tasks=trace,\
+                         turbo_tasks_viz=trace,turbo_tasks_memory=trace,turbo_tasks_fs=trace"
+                    .to_string();
+            }
```
ForsakenHarmony, line 45: I feel like constructing these strings from a list would be a lot more readable, and it's only done once anyway?

crates/turbopack-node/src/source_map/mod.rs
ForsakenHarmony: mostly concerned about performance, but according to the benchmarks it's fine, so all good

alexkirsz (commented):
Forgot to submit these comments last week.

EDIT: They're outdated, marking them as resolved.

crates/turbopack-cli-utils/src/trace_writer.rs
```diff
@@ -0,0 +1,98 @@
+// This is inspired by tracing-appender non_blocking, but allows writing a owned
+// Vec<u8> instead of a reference, and uses a unbounded channel to avoid slowing
+// down the application.
```
alexkirsz, line 3:
```suggestion
// This is inspired by tracing-appender non_blocking, but allows writing an owned
// Vec<u8> instead of a reference, and uses an unbounded channel to avoid slowing
// down the application.
```

alexkirsz: This could be a doc comment on the struct directly.

crates/turbopack-cli-utils/src/tracing.rs
```diff
@@ -0,0 +1,77 @@
...
+    borrow::Cow,
+    fmt::{Display, Formatter},
+};
+
+use serde::{Deserialize, Serialize};
+
+/// A raw trace line.
+#[derive(Debug, Serialize, Deserialize)]
+pub enum TraceRow<'a> {
+    Start {
```
alexkirsz, line 11: Please comment each of these properties, or link to the corresponding docs.

crates/turbopack-cli-utils/src/tracing.rs
```diff
@@ -0,0 +1,77 @@
...
+        ts: u64,
+        id: u64,
+        parent: Option<u64>,
+        #[serde()]
```
alexkirsz, line 15: What's up with these attributes?

crates/turbopack-cli/src/arguments.rs
```diff
@@ -9,6 +12,14 @@ pub enum Arguments {
     Dev(DevArguments),
 }
 
+impl Arguments {
+    pub fn dir(&self) -> Option<&Path> {
```
alexkirsz, line 16: Comment this public item.

crates/turbopack-cli/src/main.rs
```diff
@@ -1,28 +1,101 @@
...
+            "turbopack_dev=trace",
+            "turbopack_image=trace",
+            "turbopack_dev_server=trace",
+            "turbopack_json=trace",
+            "turbopack_mdx=trace",
+            "turbopack_node=trace",
+            "turbopack_static=trace",
+            "turbopack_cli_utils=trace",
+            "turbopack_cli=trace",
+            "turbopack_ecmascript=trace",
```
alexkirsz, line 35: Seems like forgetting to add a new crate to this list is an easy oversight. There should either be a comment in `Cargo.toml` to remind people to do this, or we could read directly from `Cargo.toml` at compile-time. Or there might be some other way to do this automatically for all `turbopack*` crates?

crates/turbopack-cli/src/main.rs
```diff
@@ -1,28 +1,101 @@
...
+    .concat()
+});
+static TRACING_TURBO_TASKS_TARGETS: Lazy<Vec<&str>> = Lazy::new(|| {
+    [
+        &TRACING_TURBOPACK_TARGETS[..],
+        &[
+            "turbo_tasks=trace",
+            "turbo_tasks_viz=trace",
+            "turbo_tasks_memory=trace",
+            "turbo_tasks_fs=trace",
```
alexkirsz, line 47: Same comment for `turbo_tasks` crates.

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -281,7 +282,7 @@ where
 #[derive(Eq, PartialEq, Clone, Hash)]
 enum ChunkContentGraphNode<I> {
     // Chunk items that are placed into the current chunk
-    ChunkItem(I),
+    ChunkItem(I, StringReadRef),
```
alexkirsz, line 285:
```suggestion
    ChunkItem {
        item: I,
        ident: StringReadRef,
    },
```

alexkirsz (approved):

crates/turbo-tasks/src/manager.rs
```diff
@@ -857,16 +871,18 @@ impl<B: Backend + 'static> TurboTasksCallApi for TurboTasks<B> {
 }
 
 impl<B: Backend + 'static> TurboTasksApi for TurboTasks<B> {
+    #[instrument(level = Level::INFO, skip_all, name = "invalidate")]
     fn invalidate(&self, task: TaskId) {
         self.backend.invalidate_task(task, self);
     }
 
+    #[instrument(level = Level::INFO, skip_all, name = "invalidate", fields(name = display(&reason)))]
```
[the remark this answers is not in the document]
sokra: It allows to trace these high-level invalidate spans, without tracing all of turbo-tasks.

PR #4967, opened 2023-05-16
fix(turborepo): rust implementation of file hashing via git index

gsoltis:
### Description

This reverts commit f51b85b.

- Moves git index-based file hashing to rust again
- Use a clean path instead of `realpath`. We don't want to resolve symlinks, just directory traversals
- Errors when our calls to `git` are not successful. Attempts to read `stderr` in error cases

### Testing Instructions

Added a test for a symlinked repo and directory traversal

### DIFF
--- a/crates/turborepo-scm/src/hash_object.rs
+++ b/crates/turborepo-scm/src/hash_object.rs
@@ -1,0 +1,192 @@
+use std::{
+    io::{BufWriter, Read, Write},
+    panic,
+    process::{Command, Stdio},
+    thread,
+};
+
+use nom::{Finish, IResult};
+use turbopath::{AbsoluteSystemPathBuf, RelativeUnixPathBuf};
+
+use crate::{package_deps::GitHashes, Error};
+
+pub(crate) fn hash_objects(
+    pkg_path: &AbsoluteSystemPathBuf,
+    to_hash: Vec<RelativeUnixPathBuf>,
+    pkg_prefix: &RelativeUnixPathBuf,
+    hashes: &mut GitHashes,
+) -> Result<(), Error> {
+    if to_hash.is_empty() {
+        return Ok(());
+    }
+    let mut git = Command::new("git")
+        .args(["hash-object", "--stdin-paths"])
+        .current_dir(pkg_path)
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .stdin(Stdio::piped())
+        .spawn()?;
+    {
+        let stdout = git
+            .stdout
+            .as_mut()
+            .ok_or_else(|| Error::git_error("failed to get stdout for git hash-object"))?;
+        // We take, rather than borrow, stdin so that we can drop it and force the
+        // underlying file descriptor to close, signalling the end of input.
+        let stdin: std::process::ChildStdin = git
+            .stdin
+            .take()
+            .ok_or_else(|| Error::git_error("failed to get stdin for git hash-object"))?;
+        let mut stderr = git
+            .stderr
+            .take()
+            .ok_or_else(|| Error::git_error("failed to get stderr for git hash-object"))?;
+        let result = read_object_hashes(stdout, stdin, &to_hash, pkg_prefix, hashes);
+        if let Err(err) = result {
+            let mut buf = String::new();
+            let bytes_read = stderr.read_to_string(&mut buf)?;
+            if bytes_read > 0 {
+                // something failed with git, report that error
+                return Err(Error::git_error(buf));
+            }
+            return Err(err);
+        }
+    }
+    git.wait()?;
+    Ok(())
+}
+
+const HASH_LEN: usize = 40;
+
+fn read_object_hashes<R: Read, W: Write + Send>(
+    mut reader: R,
+    writer: W,
+    to_hash: &Vec<RelativeUnixPathBuf>,
+    pkg_prefix: &RelativeUnixPathBuf,
+    hashes: &mut GitHashes,
+) -> Result<(), Error> {
+    thread::scope(move |scope| -> Result<(), Error> {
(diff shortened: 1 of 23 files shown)


gsoltis:
Note that this PR makes an explicit call to `.clean()`. In the future, I think this is likely to be moved into a `.join()` implementation.

crates/turborepo-scm/src/hash_object.rs
```diff
@@ -0,0 +1,188 @@
...
+        .take()
+        .ok_or_else(|| Error::git_error("failed to get stdin for git hash-object"))?;
+    let mut stderr = git
+        .stderr
+        .take()
+        .ok_or_else(|| Error::git_error("failed to get stderr for git hash-object"))?;
+    let result = read_object_hashes(stdout, stdin, &to_hash, pkg_prefix, hashes);
+    if let Err(err) = result {
+        return Err(read_git_error(&mut stderr).unwrap_or(err));
+    }
```
NicholasLYang, line 47: nit: `map_err` might work better here

```suggestion
    let result = read_object_hashes(stdout, stdin, &to_hash, pkg_prefix, hashes).map_err(|err| {
      read_git_error(&mut stderr).unwrap_or(err)
    })?;
```

crates/turborepo-scm/src/ls_tree.rs
```diff
@@ -0,0 +1,143 @@
...
+    Ok(())
+}
+
+struct LsTreeEntry<'a> {
+    filename: &'a [u8],
+    hash: &'a [u8],
+}
+
+fn parse_ls_tree(i: &[u8]) -> Result<LsTreeEntry<'_>, Error> {
+    match nom_parse_ls_tree(i).finish() {
```
NicholasLYang, line 63: Do we care if there's text left over? Generally with combinators we want to ensure that there's nothing left

crates/turborepo-scm/src/ls_tree.rs
```diff
@@ -0,0 +1,143 @@
...
+    let mut reader = BufReader::new(reader);
+    let mut buffer = Vec::new();
+    loop {
+        buffer.clear();
+        {
+            let bytes_read = reader.read_until(b'\0', &mut buffer)?;
+            if bytes_read == 0 {
+                break;
+            }
+            {
```
NicholasLYang, line 46: What's the reason for this being in a block?

crates/turborepo-scm/src/ls_tree.rs
```diff
@@ -0,0 +1,143 @@
...
+                let entry = parse_ls_tree(&buffer)?;
+                let hash = String::from_utf8(entry.hash.to_vec())?;
+                let path = RelativeUnixPathBuf::new(entry.filename)?;
+                hashes.insert(path, hash);
+            }
+        }
+    }
```
NicholasLYang, line 53: Could we do this? 
```suggestion
    while reader.read_until(b'\0', &mut buffer) != 0 {
            let entry = parse_ls_tree(&buffer)?;
            let hash = String::from_utf8(entry.hash.to_vec())?;
            let path = RelativeUnixPathBuf::new(entry.filename)?;
            hashes.insert(path, hash);
            buffer.clear();
    }
```

crates/turborepo-scm/src/hash_object.rs
```diff
@@ -0,0 +1,188 @@
...
+                writer.write_all(&[b'\n'])?;
+                writer.flush()?;
+            }
+            // writer is dropped here, closing stdin
+            Ok(())
+        });
+        // Buffer size is HASH_LEN + 1 to account for the trailing \n
+        let mut buffer: [u8; HASH_LEN + 1] = [0; HASH_LEN + 1];
+        for (i, filename) in to_hash.iter().enumerate() {
+            if i == to_hash.len() {
```
NicholasLYang, line 75: I'm confused, when does this happen?

crates/turborepo-scm/src/ls_tree.rs
gsoltis: I think an artifact from an earlier iteration. Removed.

crates/turborepo-scm/src/ls_tree.rs
gsoltis: Yup, that works.

crates/turborepo-scm/src/ls_tree.rs
gsoltis: Added some `all_consuming` calls.

crates/turborepo-scm/src/hash_object.rs
gsoltis: Dropped. It shouldn't.

crates/turborepo-scm/src/lib.rs
```diff
@@ -21,4 +29,57 @@ pub enum Error {
...
+    if let Ok(bytes_read) = stderr.read_to_string(&mut buf) {
+        if bytes_read > 0 {
+            // something failed with git, report that error
+            Some(Error::git_error(buf))
+        } else {
+            None
+        }
+    } else {
+        None
+    }
```
NicholasLYang, line 68: nit: this could be
```suggestion
    let bytes_read = stderr.read_to_string(&mut buf).ok()?;
    if bytes_read > 0 {
        // something failed with git, report that error
        Some(Error::git_error(buf))
    } else {
        None
    }
```

NicholasLYang: If you *really* want to code golf you can do:

```suggestion
    let bytes_read = stderr.read_to_string(&mut buf).ok()?;
    (bytes_read > 0).then(|| Error::git_error(buf))
```

But that's a little unnecessarily fancy tbh

NicholasLYang (approved):

PR #4969, opened 2023-05-16
ci: Remove Rust/Go toolchain for examples tests

mehulkar:
Examples tests do not use a local build of Turborepo, so we do not need the toolchain for that.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -361,11 +361,18 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      # Note: the setup-node action inside this is somehow required for
+      # the node setup below this step. Not sure why.
       - uses: ./.github/actions/setup-turborepo-environment
         with:
           target: ${{ matrix.os.name }}
           github-token: "${{ secrets.GITHUB_TOKEN }}"
+          setup-rust: false
+          setup-go: false
 
+      # TODO: .github/actions/setup-node is pretty similar to the following
+      # steps (Setup Pnpm, pnpm cache, corepack, setup node), but it didn't
+      # work out of box. We should take another stab at it later.
       - name: Setup Pnpm
         uses: pnpm/action-setup@v2.2.4
         with:
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -69,11 +69,11 @@ runs:
         save-if: ${{ github.ref == 'refs/heads/main' && inputs.save-cache || 'false' }}
 
     - name: "Install cargo-sweep"
-      if: ${{ inputs.install-cargo-sweep }} == "true"
+      if: ${{ inputs.install-cargo-sweep == 'true' }}
       uses: baptiste0928/cargo-install@v1
       with:
         crate: cargo-sweep
 
     - name: "Run cargo-sweep"
-      if: ${{ inputs.install-cargo-sweep }} == "true"
+      if: ${{ inputs.install-cargo-sweep == 'true' }}
       uses: ./.github/actions/cargo-sweep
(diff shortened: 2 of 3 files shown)


.github/actions/setup-node/action.yml
```diff
@@ -22,6 +26,12 @@ runs:
       with:
         version: 7.12.1
 
+    - name: Make sure pnpm always has a cache
+      if: ${{ inputs.enable-pnpm-cache == 'true' }}
+      shell: bash
+      run: |
+        mkdir -p `pnpm store path`
```
mehulkar, line 33: I'm faithfully adding this in here to keep parity with what the workflow job does now, but not sure if it works, and if it should be applied unviersally to all other invocations of setup-node. Not _really_ trying to go down that path in this PR.

nathanhammond: There was an issue with the pnpm or setup-node action that would throw an error at this step in _some_ cases, I think around the pnpm 6=>7 transition.

I haven't gone back to check on it; this was the official workaround at that point in time.

This is safe to do and is at most a no-op.

.github/workflows/test.yml
```diff
@@ -379,31 +379,16 @@ jobs:
...
+      - name: "Setup Node"
+        uses: ./.github/actions/setup-node
         with:
-          version: 7.2.1
+          enable-corepack: "false"
+          package-install: "false"
+          enable-pnpm-cache: "true"
 
-      - name: Make sure pnpm always has a cache
+      - name: Install Turbo globally
```
nathanhammond, line 389:
```suggestion
      - name: Install Turborepo globally
```

.github/workflows/test.yml
```diff
@@ -379,31 +379,16 @@ jobs:
...
-        shell: bash
-        run: corepack disable
-
-      - name: Setup Node.js
-        uses: actions/setup-node@v3
-        with:
-          node-version: 16
-          cache: ${{ matrix.manager }}
-          cache-dependency-path: package.json
+        run: npm i -g turbo@canary
```
nathanhammond, line 391: I agree that our existing setup is weird and this is basically no different from our existing setup unless we make changes in the global=>local delegation.

But why would we choose to do this instead of `--skip-infer` and use the `HEAD` version directly for executing the examples?

mehulkar: examples tests are testing the _examples_, not the cli. They don't even run when the cli source changes. They're meant to test that an out-of-box example works. This is noted in the README too: https://github.com/vercel/turbo/blob/main/turborepo-tests/examples/README.md

I agree it would be nice to utilize examples to test `HEAD` of cli as well, but this PR is not the thing that will get us there. (nudge: read the rfc I wrote months ago and help move it forward 😃)

tknickman (approved):

.github/workflows/test.yml
```diff
@@ -361,31 +361,15 @@ jobs:
...
         with:
-          version: 7.2.1
+          enable-corepack: "false"
+          package-install: "false"
 
-      - name: Make sure pnpm always has a cache
+      - name: Install Turborepo globally
         shell: bash
-        run: |
-          mkdir -p `pnpm store path`
```
mehulkar, line 377: We lose this, but it looks like conditionally making this happen causes issues with the setup-node Post cleanup code: https://github.com/vercel/turbo/actions/runs/5006261750/jobs/8971325789?pr=4929

> Error: Cache folder path is retrieved for pnpm but doesn't exist on disk: /home/runner/setup-pnpm/node_modules/.bin/store/v3

I _think_ we can setup pnpm cache correctly in a separate pass, it's mostly incidental that the Examples job has this line and the others jobs don't.

mehulkar: reverting this piece, something isn't working in CI and I don't want to spend more time on this right now!

PR #4973, opened 2023-05-16
fix(create-turbo): Add --fix-lockfile for pnpm install

nathanhammond:
If:
- `create-turbo` is run in an environment detected as CI,
- using `pnpm`, and
- the `pnpm` version is newer than the lockfile that we ship.

The installation step will fail because `pnpm` makes `--frozen-lockfile` the default on CI.

To address this, `create-turbo` should call `pnpm install --fix-lockfile`, which will succeed in a CI environment.


tknickman (approved):
Works for me

PR #4978, opened 2023-05-16
feat(docs): add code gen page

tknickman:
### Description

Docs for code gen


weyert:
@tknickman Is this supposed to be compatible with Plop? When I add a line `import globby from 'globby'`-line to `config.ts` it all stops working for me. Stops finding custom generators when I remove this line it can find the custom generators again

tknickman:
Thanks for taking a look @weyert, I would expect that to work - I'll take a look before release. 

Thanks!

weyert:
@tknickman I have made a small repro: https://github.com/weyert/turbo-gen-repro

`turbo generate adr` doesn't work until I remove this line:
https://github.com/weyert/turbo-gen-repro/blob/main/turbo/generators/config.ts#L37

tknickman:
@weyert thanks for the repro - this is happening because we're bundling our utility package to CJS, and globby is ESM. 

After we get this initial version out I'll look into supporting ESM dependencies as well! 

I'm also going to update this now to output a meaningful error instead of nothing at all in the meantime.

docs/pages/repo/docs/core-concepts/monorepos/_meta.json
```diff
@@ -2,5 +2,6 @@
   "running-tasks": "Running Tasks",
   "task-dependencies": "Task Dependencies",
   "filtering": "Filtering Workspaces",
+  "code-generation": "Code Generation",
```
mehulkar, line 5: How about "Generating new workspaces"? We aren't generating code so much as places to put code

docs/pages/repo/docs/core-concepts/monorepos/code-generation.mdx
```diff
@@ -0,0 +1,221 @@
...
+title: Code Generation
+description: Extend your Turborepo with new apps, and packages. Create new empty workspaces, copy existing workspaces, add workspaces from remote sources (just like `create-turbo`!) or run custom generators defined using Plop configurations.
+---
+
+import Callout from "../../../../../components/Callout";
+
+# Code Generation
+
+Splitting your monorepo into individual workspaces is a great way to organize your code, speed up tasks, and improve the local development
+experience. One of the benefits of this approach is that it's easy to generate new source code for packages, modules,
```
mehulkar, line 11: This sentence reads kind of weird. Using a monorepo doesn't make it easy to generate new source code. The tool we're introducing does.

mehulkar (commented):
Left my 2c, not blocking

docs/pages/repo/docs/core-concepts/monorepos/_meta.json
tknickman: I checked out how this is described in the docs of a few other tools (rails, nx etc) and they all refer to it as code generation so I'm in favor of leaving as is for the cross shoppers. 

And you can generate whatever you want with the custom generators!

docs/pages/repo/docs/core-concepts/monorepos/code-generation.mdx
tknickman: I agree, this was originally how I had positioned it but this update was suggested from @anthonyshew - but I might have missed the mark on it a bit. 

@anthonyshew do you want to tweak this?

docs/pages/repo/docs/core-concepts/monorepos/_meta.json
anthonyshew: To be fair, you can indeed generate code with this feature, all the way down to UI components.

anthonyshew (approved):

chris-olszewski (approved):

PR #4980, opened 2023-05-16
refactor(turborepo): Signature Authentication (2nd try)

NicholasLYang:
### Description

Refactors the signature authentication to use a language-independent algorithm, as the previous implementation depended on Go's JSON serialization.

NOTE: This is a global hash bump.

### Testing Instructions

Testing is using fuzzing across the Go-Rust boundary to check that signatures are correctly generated on both sides.


crates/turborepo-cache/src/signature_authentication.rs
```diff
@@ -0,0 +1,267 @@
...
+
+impl ArtifactSignatureAuthenticator {
+    pub fn new(team_id: Vec<u8>, secret_key_override: Option<Vec<u8>>) -> Self {
+        Self {
+            team_id,
+            secret_key_override,
+        }
+    }
+
+    fn secret_key(&self) -> Result<Vec<u8>, SignatureError> {
```
arlyon, line 39: I believe that HMAC_SHA256 doesn't have key length requirements but a comment that verifies that would be handy.

crates/turborepo-cache/src/signature_authentication.rs
```diff
@@ -0,0 +1,267 @@
...
+        Ok(BASE64_STANDARD.encode(hmac_output))
+    }
+
+    pub fn validate(
+        &self,
+        hash: &[u8],
+        artifact_body: &[u8],
+        expected_tag: &str,
+    ) -> Result<bool, SignatureError> {
+        let secret_key = hmac::Key::new(HMAC_SHA256, &self.secret_key()?);
```
arlyon, line 96: I also think we could pull HMAC_SHA256 into a constant

arlyon (approved):
Approved with some minor comments. I think we should announce this in the turbo channel / turboverse before merging.

crates/turborepo-cache/src/signature_authentication.rs
```diff
@@ -0,0 +1,271 @@
...
+         TURBO_REMOTE_CACHE_SIGNATURE_KEY environment variable"
+    )]
+    NoSignatureSecretKey,
+    #[error("serialization error: {0}")]
+    SerializationError(#[from] serde_json::Error),
+    #[error("base64 encoding error: {0}")]
+    Base64EncodingError(#[from] base64::DecodeError),
+}
+
+static TURBO_HMAC_ALGORITHM: Algorithm = HMAC_SHA256;
```
NicholasLYang, line 24: this is static because apparently you can't have a const that refers to another const

PR #4984, opened 2023-05-16
feat(cli): rework generator api

tknickman:
### Description

Make the following changes:
- alias generate to gen / g
- make run the default

Default commands are a little annoying in clap, so reworked the entire structure of the arguments.

### DIFF
--- a/crates/turborepo-lib/src/cli.rs
+++ b/crates/turborepo-lib/src/cli.rs
@@ -265,12 +265,27 @@ pub enum Command {
         target: LinkTarget,
     },
     /// Generate a new app / package
+    #[clap(aliases = ["g", "gen"])]
     Generate {
+        #[serde(skip)]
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/cli.rs
```diff
@@ -265,12 +265,27 @@ pub enum Command {
         target: LinkTarget,
     },
     /// Generate a new app / package
+    #[clap(aliases = ["g", "gen"])]
     Generate {
```
tknickman, line 269: Default commands in clap are a little tricky, so Generate _becomes_ the default command (run) and then the subcommand is optional.

gsoltis (approved):

mehulkar (approved):
Do we want to add any integration tests for this command?

tknickman:
@mehulkar yes - I have a task to add these as a follow up!

PR #4985, opened 2023-05-16
feat(turbopack): basic sass-loader support

kwonoj:
### Description

WEB-654.

This is pairing PR from next.js https://github.com/vercel/next.js/pull/49882 (not a breaking change though) to have necessary support to run sass-loader. Mainly, it adds near-dummy context in `webpack-loaders` as well as assigning specific module types for the scss / sass. 

I'm not entirely in favor of having internal custom module type logic in webpack loaders setup by extension - however next.config.js itself doesn't have enough information other than extension + loader to determine module type. Something would like to address.


crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -41,6 +41,66 @@ const toPath = (file: string) => {
...
+
+const loaderFlag = "LOADER_EXECUTION";
+
+const cutOffByFlag = (stack, flag) => {
+  const errorStack = stack.split("\n");
+  for (let i = 0; i < errorStack.length; i++) {
+    if (errorStack[i].includes(flag)) {
+      errorStack.length = i;
+    }
+  }
```
jridgewell, line 74: Is this missing a break statement?

crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -41,6 +41,66 @@ const toPath = (file: string) => {
...
+class DummySpan {
+  traceChild() {
+    return new DummySpan();
+  }
+
+  traceFn<T>(fn: (span: DummySpan) => T): T {
+    try {
+      return fn(this);
+    } finally {
+    }
```
jridgewell, line 93:
```suggestion
    return fn(this);
```

crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -41,6 +41,66 @@ const toPath = (file: string) => {
...
+  }
+
+  async traceAsyncFn<T>(fn: (span: DummySpan) => T | Promise<T>): Promise<T> {
+    try {
+      return await fn(this);
+    } finally {
+    }
```
jridgewell, line 100:
```suggestion
    return await fn(this);
```

crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -41,6 +41,66 @@ const toPath = (file: string) => {
...
+  }
+  return errorStack.join("\n");
+};
+
+/**
+ * @param {string} stack stack trace
+ * @returns {string} stack trace without the loader execution flag included
+ */
+const cutOffLoaderExecution = (stack) => cutOffByFlag(stack, loaderFlag);
```
jridgewell, line 82: Can you explain why this is needed?

kwonoj: These are copy-paste from what webpack does for the logging with its levels. We could not do it, but tried to mimic its behavior as much.

kwonoj: ref: https://github.com/webpack/webpack/blob/81c8fbde2e0e4b8d2e30a5109dedc082ef8c09dc/lib/Compilation.js#L1166

crates/turbopack-node/js/src/transforms/webpack-loaders.ts
kwonoj: Same goes for https://github.com/vercel/turbo/pull/4985/files#r1197370422, could cut it but thought it'd better to preserve these logics as much as same.

crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -59,16 +113,41 @@ const transform = (
...
+          currentTraceSpan: new DummySpan(),
           rootContext: contextDir,
           getOptions() {
             const entry = this.loaders[this.loaderIndex];
             return entry.options && typeof entry.options === "object"
               ? entry.options
               : {};
           },
+          getResolve: () => ({
+            // [TODO] this is incomplete
```
jridgewell, line 125: We should have a `__turbopack_require_context__.resolve()` which is our module runtime's resolver.

jridgewell (approved):

PR #4989, opened 2023-05-17
Exclude turborepo-tests from JS tests

mehulkar:
Not sure why, but these tests are getting caught in the filters as they are now

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -455,7 +455,7 @@ jobs:
 
       - name: Run tests
         run: |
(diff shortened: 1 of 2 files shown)


nathanhammond (approved):
🚀

For docs we conventionally always use the long form of flag names; I've also stuck to that convention in our CI, but (unlike the docs situation) we haven't had a conversation about that. I have a mild preference for CI to use the long form of flags and always include the `run` since I see our codebase as a second source of documentation.

PR #4990, opened 2023-05-17
ci: Fix create-turbo usage in smoke tests

mehulkar:


mehulkar:
I think this should work, but when I tried testing this workflow from this branch, it just has the same error again. Makes me think that "Run this workflow from this branch" does not do what I think it does
https://github.com/vercel/turbo/actions/runs/4999115622

nathanhammond (approved):

PR #4993, opened 2023-05-17
ci: Remove go/rust toolchain from JS package tests

mehulkar:


.github/workflows/test.yml
```diff
@@ -447,11 +447,9 @@ jobs:
...
           fetch-depth: ${{ steps.fetch-depth.outputs.depth  }}
 
-      - name: Build turborepo CLI from source
-        uses: ./.github/actions/setup-turborepo-environment
-        with:
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
-          target: ${{ matrix.os.name }}
+      - name: Install Turbo globally
+        shell: bash
+        run: npm i -g turbo@canary
```
nathanhammond, line 452: - hard block: `canary` isn't always the most-recently-published build.
- negotiable: I feel like we should be running against the `HEAD` version as much as possible.

nathanhammond (changes requested):

.github/workflows/test.yml
mehulkar: I don't think this needs to be a hard block, we're using `turbo` to exercise the test suite, it's not relevant what's _in_ the test suite. I can use `@latest` just as well. In fact `setup-turborepo-environment` itself installs `@canary`, so all this PR is doing is removing a bunch of other stuff that action does.

chris-olszewski: I think this is a good point, if we're not on `HEAD` we should at least be using the closest build to `HEAD`. We ideally want the latest published version of turbo to be used: default to `canary`, but if we have a newer patch/major we should use that instead.

mehulkar: Why does it matter? This test is not for testing turbo source code. Am I missing something?

mehulkar: In any case, can we agree this PR is not changing or related to this discussion?

mehulkar: i've edited the diff to make it more clear what's actually happening here

PR #4998, opened 2023-05-17
Node.js production runtime POC

alexkirsz:
### Description

This implements a Node.js production runtime for Turbopack, to first be used for `next build --turbo`, but that can easily be extended to support more production use cases (e.g. ncc).

The runtime works differently from the dev runtime we're currently using for `next dev`: instead of having each chunk register itself against a global registry when evaluated, chunks export their module factories as a CommonJS default export. The runtime itself lives in a separate chunk, and then an "exported chunk" is generated that instantiates runtime entries and exports a given entry module.

Corresponding Next.js PR: ~~https://github.com/vercel/next.js/pull/49942~~ https://github.com/vercel/next.js/pull/50375


crates/turbopack-build/js/src/runtime.js
```diff
@@ -0,0 +1,359 @@
+/* eslint-disable @next/next/no-assign-module-variable */
```
alexkirsz, line 1: This is a slimmed down version of our dev runtime that doesn't include HMR.

It's currently missing some more recent updates to our dev runtime. 

Eventually, this runtime should be optimized for conciseness and performance. But for now, readability is privileged.

crates/turbopack-build/src/ecmascript/node/content.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        let mut code = CodeBuilder::default();
+
+        // When a chunk is executed, it will either register itself with the current
+        // instance of the runtime, or it will push itself onto the list of pending
+        // chunks (`self.TURBOPACK`).
+        //
+        // When the runtime executes (see the `evaluate` module), it will pick up and
+        // register all pending chunks, and replace the list of pending chunks
+        // with itself so later chunks can register directly with it.
```
alexkirsz, line 61:
```suggestion
```

Not relevant to the way this particular runtime works.

crates/turbopack-build/src/ecmascript/node/manifest/chunk_asset.rs
```diff
@@ -0,0 +1,152 @@
+use anyhow::{bail, Result};
```
alexkirsz, line 1: This is mostly the same as the dev chunking context implementation, but doesn't support chunk data (yet?). This is because chunk data didn't exist at the time I implemented this POC.

crates/turbopack-dev/src/chunking_context.rs
```diff
@@ -180,22 +180,26 @@ impl ChunkingContext for DevChunkingContext {
 
     #[turbo_tasks::function]
     async fn chunk_path(&self, ident: AssetIdentVc, extension: &str) -> Result<FileSystemPathVc> {
-        fn clean(s: &str) -> String {
+        fn clean_separators(s: &str) -> String {
             s.replace('/', "_")
         }
+        fn clean_additional_extensions(s: &str) -> String {
+            s.replace('.', "_")
+        }
```
alexkirsz, line 188: `next start` won't serve files that start with a `.`, hence this particular change (which is duplicated on the `turbopack-build` side).

TODO: write a comment to explain this.

alexkirsz:
The Socket comment has nothing to do with this PR, and relates to test fixtures. cc @mehulkar might need to update these fixtures?

crates/turbopack-ecmascript/src/manifest/chunk_asset.rs
alexkirsz, line 1: I originally moved the manifest logic into `turbopack-dev` because it contained dev-specific logic in the form of the chunk list. I've since refactored the chunk list logic so the manifest does not need to know about it.

mehulkar:
@alexkirsz no idea. It looks like turbopack jobs are failing and I can't tell by scrolling the logs for it which fixtures they're using. Ping me on slack if there's something I'm missing?

crates/turbopack-ecmascript/js/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@vercel/turbopack",
```
sokra, line 2:
```suggestion
  "name": "@vercel/turbopack-ecmascript",
```

sokra (changes requested):
I quite dislike the duplication of the runtime code. Could we move the shared part to `turbopack-ecmascript`? (I want to avoid dev to prod differences as this could lead to production only problems)

We originally had the the logic for `import()` in turbopack-dev as it uses an extra indirection (manifest chunk) to allow lazy compiling the graph behind the `import()`. This won't be needed in prod. But I think it's ok to have it in `turbopack-ecmascript` and maybe use some kind of ChunkingContext trait method to control if the manifest module should be in a separate chunk or not.

crates/turbopack-build/js/src/runtime.js
```diff
@@ -0,0 +1,359 @@
...
+  if (!hOP.call(obj, name)) Object.defineProperty(obj, name, options);
+}
+
+/**
+ * Adds the getters to the exports object
+ *
+ * @param {Exports} exports
+ * @param {Record<string, () => any>} getters
+ */
+function esm(exports, getters) {
```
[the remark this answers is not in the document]
alexkirsz: Yep, this could probably be shared as part of `turbopack-ecmascript`. The other solutions are:
* to create a crate for shared utils between `turbopack-dev` and `turbopack-build`.
* to consolidate both crates into a single crate `turbopack-ecmascript-runtime`.

crates/turbopack-core/src/chunk/chunking_context.rs
```diff
@@ -52,4 +52,7 @@ pub trait ChunkingContext {
         entry: ChunkVc,
         evaluatable_assets: EvaluatableAssetsVc,
     ) -> AssetsVc;
+
+    /// Generate an output chunk asset for the given chunk.
+    fn generate_chunk(&self, chunk: ChunkVc) -> AssetVc;
```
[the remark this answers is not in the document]
alexkirsz: It is getting called externally from `ManifestChunkAssetVc` in order to generate the manifest chunk output asset. In this case, we only want to generate a single chunk, while `chunk_group` will return more than one assets (and, in development, add its own chunks like the chunk list, which are unnecessary for the manifest chunk).

An alternative, which is closer to what I was doing before, is to move this method to `EcmascriptChunkingContext` to generate a single ES chunk, that the `ManifestLoaderItem` would then call.

crates/turbopack-core/src/ident.rs
```diff
@@ -99,4 +100,136 @@ impl AssetIdentVc {
     pub async fn path(self) -> Result<FileSystemPathVc> {
         Ok(self.await?.path)
     }
+
+    /// Computes a unique output asset name for the given asset identifier.
+    #[turbo_tasks::function]
+    pub async fn output_name(
```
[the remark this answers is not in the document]
alexkirsz: Fair enough. I'll simplify the turbopack-build version to use a content hash.

alexkirsz:
Turning this back into a draft while I address https://github.com/vercel/turbo/pull/4998#discussion_r1201614127

.github/workflows/test.yml
```diff
@@ -107,6 +107,13 @@ jobs:
             crates/turbopack-bench/**
             !*.md
 
+      - name: Turbopack TypeScript related changes
+        id: turbopack_typescript
```
sokra, line 111: It needs an entry in `output` in this job too

crates/turbopack-core/src/chunk/chunking_context.rs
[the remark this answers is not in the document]
sokra: Could we allow the ManifestChunkAsset to also use `chunk_group` and potentially loading multiple chunks?

crates/turbopack-core/src/ident.rs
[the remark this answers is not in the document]
sokra: You can delay that for future PRs. There are extra challenges when using content hashes

crates/turbopack-build/src/ecmascript/node/evaluate/runtime.rs
```diff
@@ -0,0 +1,178 @@
...
+fn runtime_fs() -> FileSystemVc {
+    VirtualFileSystemVc::new().into()
+}
+
+#[turbo_tasks::value_impl]
+impl Asset for EcmascriptBuildNodeRuntimeChunk {
+    #[turbo_tasks::function]
+    fn ident(&self) -> AssetIdentVc {
+        let ident = AssetIdentVc::new(Value::new(AssetIdent {
+            path: runtime_fs().root().join("runtime.js"),
```
sokra, line 96: Not a huge fan of having an extra filesystem for that. Could we just put that into the embedded filesystem. No worries if the file doesn't exist there. We already do that a bunch of times in other places.

It's more clear when every turbopack file is inside of `[turbopack]/`

sokra (commented):
No way I can read all that code... It's probably fine...

.github/workflows/test.yml
```diff
@@ -448,6 +456,36 @@ jobs:
         run: |
           turbo run check-types test --filter=...[${{ github.event.pull_request.base.sha || 'HEAD^1' }}] --filter="./packages/*" --color
 
+  turbopack_typescript:
```
sokra, line 459: You also need to add this job to the `done` and `ok` jobs a bunch of times.

crates/turbopack-build/src/ecmascript/node/evaluate/chunk.rs
```diff
@@ -0,0 +1,245 @@
...
+        )?;
+
+        let other_chunks = this.other_chunks.await?;
+        for other_chunk in &*other_chunks {
+            let other_chunk_path = &*other_chunk.ident().path().await?;
+            if let Some(other_chunk_public_path) = output_root.get_path_to(other_chunk_path) {
+                writedoc!(
+                    code,
+                    r#"
+                        runtime.loadChunk({});
```
sokra, line 110:
```suggestion
                        runtime.loadChunk(require({}));
```

To make the bundle statically analyse-able, e. g. by nft, it would be great to avoid dynamic require() calls and inject the require call with the chunk path.

alexkirsz: How would that work with dynamic imports on the server-side? We transform the import into a `__turbopack_load_chunk__` call. Would that also need to be replaced with a direct call to `require` instead?

sokra: Yeah that would be perfect

alexkirsz: Created https://linear.app/vercel/issue/WEB-1112/make-the-production-nodejs-runtime-statically-analyzeable to track this.

sokra (approved):
Wasn't able to fully review that, but I trust the test suite 🤞

PR #4233, opened 2023-03-16
feat(examples): add with-gatsby

tknickman:
### Description

Create `with-gatsby` template

### Testing Instructions

Pull example, `pnpm install`, `pnpm build`

PR #4999, opened 2023-05-17
chore: make node version less strict for gatsby example

mehulkar:
node 18 was added as a requirement in #4233 when this example was created,
but our CI runs examples tests with node 16 at the moment, which adds an
extra output warning. We can easily workaround that, but in this case, the [gatsby
starter template][1] does not have this requirement and none of our other examples do
either, although [`with-docker` requires >14][2], and the [`kitchen-sink` remix app requires > 14][3]).

This makes our examples at least consistent with _our_ CI. We have a task to generally update everything
to Node 18 where we _could_ add these requirements uniform. 

An alternative is to just remove this `engines`, but it's probably good practice to recommend a minimum
of Node 16.

[1]: https://github.com/gatsbyjs/gatsby-starter-hello-world/blob/master/package.json
[2]: https://github.com/vercel/turbo/blob/717e8c335ce5d277ac6841ae74b0b6051af1c743/examples/with-docker/package.json#L25
[3]: https://github.com/vercel/turbo/blob/717e8c335ce5d277ac6841ae74b0b6051af1c743/examples/kitchen-sink/apps/blog/package.json#L30

### DIFF
--- a/examples/with-gatsby/package.json
+++ b/examples/with-gatsby/package.json
@@ -18,7 +18,7 @@
     "turbo": "latest"
   },
   "engines": {
-    "node": ">=18.0.0"
+    "node": ">=16.0.0"
   },
   "dependencies": {},
   "packageManager": "pnpm@7.29.1"


mehulkar:
On a side note... the only reason I ran into this is because I accidentally added a `pnpm-gatsby` example test in https://github.com/vercel/turbo/pull/4929. I'm going to delete that and we can take this PR back to draft. I imagine what we should do is add that test and resolve any issues like this as part of that.

tknickman (approved):
Looks good, If I remember correctly one of the plugins we had on this example (possibly the pnpm one?) required 18 at the time (I definitely remember I added it for some reason haha). But the build looks good so :shipit:

PR #5006, opened 2023-05-17
feat(basic): add generators to basic example

tknickman:
### Description

Add a simple generator to the basic example


tknickman:
Requires https://github.com/vercel/turbo/pull/4978 (because of the link in the comment)

arlyon (approved):
Nice!

PR #5007, opened 2023-05-17
feat(turborepo): Scripts to build a debug version on windows

gsoltis:
### Description

 - Removes the `rerun-if-changed` guard from `build.rs`. It's not properly scoped, and most of the compilation should be cached, so prefer to just always run it. Note that for prod builds we don't do anything extra, so don't incur the cost.
 - Update `build.rs` to call `build_go.bat` instead of `make` on windows
 - Add a `build_go.bat` script. Handles building `turborepo-ffi`, moving the library + header to the correct spot, and building `go_turbo.exe`
 - Add a `build_windows.bat` script. Builds `turborepo` (`turbo.exe`), which in turn builds `go-turbo.exe` via `build.rs`. Copies resulting binaries into `target/debug/`

Known limitation:
 - If the daemon is running, the copy of `turbo.exe` will fail because the existing binary is in use

### Testing Instructions

Tested a few times on parallels / windows-on-arm.


cli/build_go.bat
```diff
@@ -0,0 +1,14 @@
+cd ..\crates\turborepo-ffi
+cargo build --target x86_64-pc-windows-gnu --target-dir .\target
+if %errorlevel% neq 0 exit /b %errorlevel%
+copy .\target\x86_64-pc-windows-gnu\debug\libturborepo_ffi.a ..\..\cli\internal\ffi\libturborepo_ffi_windows_amd64.a
```
NicholasLYang, line 4: Unless there's an existing `libturborepo_ffi.a` I think rust produces a `turborepo_ffi.lib`

chris-olszewski (approved):
Tested this on my Windows machine and everything built correctly with cmd and git bash

PR #5011, opened 2023-05-17
ci(examples): make every examples test a workspace

mehulkar:
### Summary
Parallelize examples tests with turbo instead of Github Actions.

(This is an alternative approach to parallelizing with pytest+prysk in #4190)

### Why
- Reduce the number of Github Workflow runs  triggered when an example
changes. This should be faster and easier to read in a PR (2 checks instead of 16).
- More turbo dogfooding

### Caveats
All examples tests are still run when any of the examples change, because each
example itself is not a workspace, so we cannot wire up individual dependencies.
It might make sense to put this test _inside_ each `examples/`, but since people
download these examples (via `npx create-turbo`), we'd need to be careful about
not including the tests into the example.

### Reviewer notes

- `turbo.json` configures `inputs` for caching, but we still using the Github Actions
   config to to limit these runs to changes in `examples/` and `examples-tests/**`.
   We could opt out of this and rely on turbo's caching, but I'm deferring that for now.

- Some tests don't really make sense to me (e.g. `npm` with the `with-yarn` example)
   but I'm not planning to change/delete those in this PR, unless I've translated the
   matrix wrong.

- The `pnpm-gatsby` test was not part of the matrix before. Moving to this setup revealed
   that it does not pass in CI (because of mismatched node version). I'm keeping the test,
   but filtering it out in CI, and will take a separate pass at making it pass. It already has another
   issue where it doesn't get "FULL TURBO".

Recent work in this area:
- https://github.com/vercel/turbo/pull/5242
- https://github.com/vercel/turbo/pull/5240


tknickman:
> There are some funky tests that don't really make sense to me
(e.g. npm with the with-yarn). I'm not planning to change/delete those
in this PR, unless I've translated the matrix wrong.

We don't even really _need_  the `with-yarn` or `with-npm` examples anymore since create-turbo always uses basic and just codemods between the package managers. The only reason I left them was for those who may prefer browsing an example before getting started. But I could be convinced to yank them out.

mehulkar:
@tknickman yeah I'm ok to add/remove as well. Let's figure that out outside this PR

gsoltis (approved):

PR #5018, opened 2023-05-18
Add ServerInfo struct for node rendering

jridgewell:
### Description

This adds a new serializable helper struct for Turbopack's server information (`hostname`, `ip`, `port`, `protocol`).

### Testing Instructions

None

Re: WEB-1064


kwonoj (approved):
Curious if there are existing struct around one of our deps 🤔

PR #5019, opened 2023-05-18
feat(prune) allow pruning of projects using Yarn PnP 

chris-olszewski:
### Description

We have had Yarn2+ support* for quite awhile now. This PR is one of the final steps to remove that asterisks and allow Yarn PnP to be used.

A note is that we currently don't copy over the `.pnp.cjs` file or the `.yarn/cache` which are required for [zero installs](https://yarnpkg.com/features/zero-installs). We skip this a naive copy would cause Docker cache misses in the same way a naive copy of the lockfile would. Future work can include correctly pruning this file and directory to allow for the `yarn install` on a cache miss to be even faster.

### Testing Instructions

Manual testing that `yarn install --immutable` works for PnP projects.

Added an integration test that verifies we no longer throw when trying to prune a PnP project and that we don't copy over files that would bust the cache.


arlyon (approved):
Nice!

PR #5020, opened 2023-05-18
feat(turborepo): remove RelativeSystemPathBuf, clean paths on join

gsoltis:
### Description
Commits are:
 - switch all uses of `join_relative` to `join_literal(s)`
 - drop `relative_system_path_buf.rs`
 - add `.clean()` to the `.join_*()` methods on `AbsoluteSystemPath`, defer join methods on `AbsoluteSystemPathBuf` to `AbsoluteSystemPath`
 - Some cleanup to lock down access to path methods that we don't currently use in production code

### Testing Instructions

Existing test suite, plus some debug_assert calls for joining literals


NicholasLYang (approved):
Always nice to delete code

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -93,11 +89,8 @@ pub async fn daemon_server(
 
         let hash = format!("{}-turbo.log", base.repo_hash());
 
-        let logs = RelativeSystemPathBuf::new("logs").expect("forward relative");
-        let file = RelativeSystemPathBuf::new(hash).expect("forward relative");
-
-        let log_folder = folder.join_relative(logs);
-        let log_file = log_folder.join_relative(file);
+        let log_folder = folder.join_literal("logs");
+        let log_file = log_folder.join_literal(hash.as_str());
```
chris-olszewski, line 93: Can we put the `format` call in here directly? I got worried that we weren't using something literal-ish i.e. user input.
```suggestion
        let log_file = log_folder.join_literal(&format!("{}-turbo.log", base.repo_hash()));
```

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -107,21 +108,30 @@ impl AbsoluteSystemPath {
         &self.0
     }
 
-    pub fn join_relative(&self, path: &RelativeSystemPathBuf) -> AbsoluteSystemPathBuf {
-        let path = self.0.join(path.as_path());
-        AbsoluteSystemPathBuf(path)
+    // intended for joining literals or obviously single-token strings
+    pub fn join_literal(&self, segment: &str) -> AbsoluteSystemPathBuf {
```
chris-olszewski, line 112: I have a slight issue with the naming here since we allow non-literals and we make use of this behavior. I feel like `join_path_component` is slightly more honest.

chris-olszewski (approved):
Just two minor nits

crates/turborepo-paths/src/absolute_system_path.rs
gsoltis: went with `join_component`

crates/turborepo-lib/src/commands/daemon.rs
gsoltis: Done.

PR #5024, opened 2023-05-19
ci: simplify filter for running JS tests

mehulkar:
Using ! syntax seems to create a union set between the filters and catches many more tests than we want to run in this case. Using `./packages` works better, because even though it includes packages that the turborepo team does not own, the filter by changed workspaces should exclude them. Even if it doesn't, it's ok to run all JS package tests without accounting for code ownership.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -444,7 +444,7 @@ jobs:
 
       - name: Run tests
         run: |
-          turbo run check-types test --filter=...[${{ github.event.pull_request.base.sha || 'HEAD^1' }}] --filter="./packages/*" --filter="!@vercel/*" --filter="!turborepo-tests*" --color
+          turbo run check-types test --filter=...[${{ github.event.pull_request.base.sha || 'HEAD^1' }}] --filter="./packages/*" --color
 
   rust_prepare:
     name: Check rust crates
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -444,7 +444,7 @@ jobs:
 
       - name: Run tests
         run: |
-          turbo run check-types test --filter=...[${{ github.event.pull_request.base.sha || 'HEAD^1' }}] --filter="./packages/*" --filter="!@vercel/*" --filter="!turborepo-tests*" --color
+          turbo run check-types test --filter=...[${{ github.event.pull_request.base.sha || 'HEAD^1' }}] --filter="./packages/*" --color
```
chris-olszewski, line 447: Was it the use of multiple exclusions that caused the overfit or maybe the exclusion of a pattern? [We document](https://turbo.build/repo/docs/core-concepts/monorepos/filtering#excluding-workspaces) this usage so I would like a better understanding of what we did wrong here. (Also noticed we have 0 test coverage for exclusions 😢 )

chris-olszewski (approved):
Approving as this overfit seems okay, but we should probably have a follow up issue to figure/document what went wrong so we can advise users on correct usage of exclusion.

PR #5035, opened 2023-05-19
feat(turbo/gen): improved onboarding

tknickman:
### Description

Improves the experience of running `turbo gen` without existing configuration

### DIFF
--- a/packages/turbo-gen/src/commands/run/prompts.ts
+++ b/packages/turbo-gen/src/commands/run/prompts.ts
@@ -47,18 +47,18 @@ export async function customGenerators({
 }
 
 export async function chooseGeneratorTemplate() {
-  return await inquirer.prompt<{ answer: "ts" | "js" }>({
+  return inquirer.prompt<{ answer: "ts" | "js" }>({
     type: "list",
     name: "answer",
     message: "Should the generator config be created with TS or JS?",
     default: "ts",
     choices: [
       {
-        name: "js",
+        name: "JavaScript",
         value: "js",
       },
       {
-        name: "ts",
+        name: "TypeScript",
         value: "ts",
       },
     ],
@@ -66,7 +66,7 @@ export async function chooseGeneratorTemplate() {
 }
 
 export async function confirm({ message }: { message: string }) {
-  return await inquirer.prompt<{ answer: boolean }>({
+  return inquirer.prompt<{ answer: boolean }>({
     type: "confirm",
     name: "answer",
     message,
--- a/packages/turbo-gen/src/generators/custom.ts
+++ b/packages/turbo-gen/src/generators/custom.ts
@@ -11,16 +11,18 @@ export async function generate({
   project,
   opts,
 }: CustomGeneratorArguments) {
+  let isOnboarding = false;
   let generators = getCustomGenerators({ project, configPath: opts.config });
   if (!generators.length) {
-    logger.error(`No custom generators found.`);
+    logger.error(`No generators found.`);
     console.log();
 
     const { answer } = await prompts.confirm({
-      message: `Would you like to add generators to ${project.name}?`,
+      message: `Would you like to add an example generator to ${project.name}?`,
     });
 
     if (answer) {
+      isOnboarding = true;
       const { answer: template } = await prompts.chooseGeneratorTemplate();
       try {
         await setupFromTemplate({ project, template });
@@ -32,9 +34,25 @@ export async function generate({
         throw err;
       }
 
+      // make it obvious that we're done creating a generator, and now we're running it
+      console.log();
+      logger.info(`Example generator config successfully created!`);
+      logger.info(`Loading generator config...`);
+      console.log();
+
       // fetch generators again, and continue to selection prompt
       generators = getCustomGenerators({ project, configPath: opts.config });
+
+      // something went wrong and we weren't able to find our new demo generator
+      if (!generators.length) {
+        logger.error(`Error loading generator.`);
+        return;
+      }
     } else {
+      console.log();
+      logger.dimmed(
+        "Learn more about generators: https://turbo.hotwire.dev/reference/generators"
+      );
       return;
     }
   }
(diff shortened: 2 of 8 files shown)


packages/turbo-gen/src/commands/run/prompts.ts
```diff
@@ -47,26 +47,26 @@ export async function customGenerators({
 }
 
 export async function chooseGeneratorTemplate() {
-  return await inquirer.prompt<{ answer: "ts" | "js" }>({
```
mehulkar, line 50: why do the awaits go away?

packages/turbo-gen/src/generators/custom.ts
```diff
@@ -11,16 +11,18 @@ export async function generate({
...
+  let isOnboarding = false;
   let generators = getCustomGenerators({ project, configPath: opts.config });
   if (!generators.length) {
-    logger.error(`No custom generators found.`);
+    logger.error(`No generators found.`);
     console.log();
 
     const { answer } = await prompts.confirm({
-      message: `Would you like to add generators to ${project.name}?`,
+      message: `Would you like to add an example generator to ${project.name}?`,
```
mehulkar, line 21:
```suggestion
      message: `Would you like to add an custom generator to ${project.name}?`,
```

If we're going with "custom" generators, let's do that everywhere?

packages/turbo-gen/src/generators/custom.ts
```diff
@@ -32,9 +34,25 @@ export async function generate({
         throw err;
       }
 
+      // make it obvious that we're done creating a generator, and now we're running it
+      console.log();
+      logger.info(`Example generator config successfully created!`);
```
mehulkar, line 39:
```suggestion
      logger.info(`Custom generator config successfully created!`);
```

packages/turbo-gen/src/generators/custom.ts
```diff
@@ -32,9 +34,25 @@ export async function generate({
...
+      logger.info(`Loading generator config...`);
+      console.log();
+
       // fetch generators again, and continue to selection prompt
       generators = getCustomGenerators({ project, configPath: opts.config });
+
+      // something went wrong and we weren't able to find our new demo generator
```
mehulkar, line 46:
```suggestion
      // something went wrong and we weren't able to find our new custom generator
```

packages/turbo-gen/src/generators/custom.ts
```diff
@@ -32,9 +34,25 @@ export async function generate({
...
+      if (!generators.length) {
+        logger.error(`Error loading generator.`);
```
mehulkar, line 48: Go linter tells us not to use `.` at the end of error messages, so we don't do that anywhere in the turbo CLI. Just calling out, not a big deal

packages/turbo-gen/src/generators/custom.ts
```diff
@@ -32,9 +34,25 @@ export async function generate({
...
+        return;
+      }
     } else {
+      console.log();
+      logger.dimmed(
+        "Learn more about generators: https://turbo.hotwire.dev/reference/generators"
```
mehulkar, line 54: lol what is this URL? did copilot generate this for you!?

packages/turbo-gen/src/generators/custom.ts
```diff
@@ -57,14 +75,22 @@ export async function generate({
...
 
     throw new GeneratorError(message, {
       type: "plop_error_running_generator",
     });
+  } finally {
+    if (isOnboarding) {
+      console.log();
+      logger.info(`Congrats! You just ran your first Turborepo generator`);
+      logger.dimmed(
+        "Learn more about Turborepo generators at https://turbo.hotwire.dev/reference/generators"
```
mehulkar, line 91: same here

packages/turbo-gen/src/templates/simple-ts/config.ts
```diff
@@ -8,19 +8,31 @@ export default function generator(plop: PlopTypes.NodePlopAPI): void {
...
+          if (input.includes(".")) {
+            return "file name cannot include an extension";
+          }
+          if (input.includes(" ")) {
+            return "file name cannot include spaces";
+          }
+          if (!input) {
+            return "file name is required";
+          }
+          return true;
```
mehulkar, line 22:
```suggestion
          // allows alphanumeric chars, underscores, hyphens, and plus signs
          return /^[\w\-+]+$/test(input)
```
and:
- add any other special chars you might want to allow
- hoist that regex out of the function so it's not recreated on every call

mehulkar (approved):
Some comments, the URL is the only thing to confirm (or maybe it's because the generator is a demo/joke thing that i'm not understanding??)

packages/turbo-gen/src/commands/run/prompts.ts
tknickman: They're redundant here since we don't need the result in this function, we can just return the promise and let the caller handle it. 

`const foo = async () => await somethingAsync()` 
is the same as
`const foo = async () => somethingAsync()`

Either way the caller will have to:
`const bar = await foo()`

packages/turbo-gen/src/generators/custom.ts
tknickman: Yea copilot was convinced this is what the URL should be for all the docs. I've been fighting this for the entire project and these two must have slipped through 😂 

I am on the chat beta, so maybe there is some wonkiness there. 

Thanks for the catch!

packages/turbo-gen/src/generators/custom.ts
tknickman: Good call

packages/turbo-gen/src/templates/simple-ts/config.ts
tknickman: I actually broke this out intentionally so the error message that the user sees could be more specific instead of just "must match this regex". The performance different isn't noticeable, and I think the more specific error messages are worth the verbosity.

PR #5036, opened 2023-05-19
ci: Remove go/rust toolchain from JS package tests (#4993)

mehulkar:
This reapplies #4993 after it was reverted in #5033. The real issue is that JS package tests shuoldn't be running the `cli#test` command (which of course will fail if Go/Rust isn't setup). I've targeted this PR to the #5024 branch since that will fix the real issue first.


mehulkar:
Tested in https://github.com/vercel/turbo/pull/5063

PR #5038, opened 2023-05-19
fix(daemon): kill daemon when root is removed

### LINKED ISSUE #5025: [turborepo] Newly-cloned git cloned monorepo reports erroneous "cache hit" for a common monorepo lib package

1.9.8

npm

Mac

### Describe the Bug

I have one package that is a shared utility library, which is imported by a number of other packages. This package is declared as a `devDependency` to get the NPM Monorepo link, but to keep it out of our consumer's `node_modules` when they run `npm install`.

Here is how one of the **consuming** packages that uses this shared utility library is declaring the monorepo dep:
```json
"devDependencies": {
    "@internal/devspaces-flowify-js-utils": "^0.0.0",
    "app-root-path": "^3.1.0",
    "open": "^7"
  },
```

And here is the shared utility library's `package.json` file:
```json
{
  "private": true,
  "name": "@internal/devspaces-flowify-js-utils",
  "description": "A collection of shared utilities that are useful for any VSCode plugin or its compiled core",
  "main": "dist/index.js",
  "typings": "dist/index.d.ts",
  "types": "dist/index.d.ts",
  "version": "0.0.0",
  "scripts": {
    "build": "run-s clean typecheck bundle",
    "bundle": "node ./scripts/bundle.js",
    "clean": "rm -rf ./dist",
    "lint": "eslint './src/**/*.{tsx,ts,jsx,js}' --no-error-on-unmatched-pattern",
    "lint:fix": "eslint './src/**/*.{tsx,ts,jsx,js}' --no-error-on-unmatched-pattern --fix",
    "prepublishOnly": "npm run --prefix ../.. build -- --filter @flowify/devspaces-flowify-js-utils",
    "typecheck": "tsc --project ./tsconfig.json --noEmit"
  },
  "devDependencies": {
    "@types/command-exists": "^1.2.0",
    "@types/fs-extra": "^11.0.1",
    "@types/unzipper": "^0.10.6",
    "@types/tar": "^6.1.5"
  },
  "dependencies": {
    "@antongolub/git-root": "^1.5.10",
    "chalk": "^4",
    "command-exists": "^1.2.9",
    "fast-glob": "^3.2.12",
    "fs-extra": "^11.1.1",
    "got": "^11",
    "unzipper": "^0.10.14",
    "tar": "^6.1.15"
  }
}
```

This utility package keeps reporting a cache hit, even after I've blasted everything away in the repo with`git clean -xdf`, which should wiping out the Turbo cache (see images below for behavior):

### Top-level project folder structure
top-level-structure

### Monorepo packages folder structure
package-structure

### Erroneous Cache Hit with no `node_modules` folder
CleanShot 2023-05-18 at 17 31 45

Even after a fresh repository clone, the package reports a cache hit, which leads me to believe that there's either a bug with the Caching mechanism, or Turbo's Cache location is erroneously reported as `node_modules/.cache/turbo` in the [official documentation](https://turbo.build/repo/docs/reference/command-line-reference#--cache-dir).

Downgrading clear back to `1.6` has corrected this issue, so I am unblocked for now, but I am concerned there's a lingering bug that is causing this issue.

### Expected Behavior

I would expect there to not be a cache hit, since I blasted away all non-SCM files. I would also not expect a cache hit if I re-cloned the git repository, even if the repository was cloned at some point in the past, but was removed before re-cloning.

### To Reproduce

Please refer to screenshots in the bug description. The shared utility library is compiled to plain JS (with `*.d.ts` files) using `tsup`.

### Reproduction Repo

An internal, enterprise repository

chris-olszewski:
### Description

Fixes #5025

In Go code we would shut down the daemon whenever the repository root was removed. This PR adds that logic back by watching the root and exiting the daemon if the root is removed.

### Testing Instructions

Added unit test. Follow instructions in reproduction repository in `uncurated-tests/turborepo-netflix-cache-repro`


crates/turborepo-globwatch/src/lib.rs
```diff
@@ -287,6 +289,16 @@ impl<T: Watcher> WatchConfig<T> {
             .map_err(ConfigError::WatchError)
     }
 
+    /// Register a single path to be included by the watcher.
+    pub async fn include_path(&self, path: &Path) -> Result<(), ConfigError> {
+        trace!("watching {:?}", path);
+        self.watcher
+            .lock()
+            .expect("watcher lock poisoned")
+            .watch(path, notify::RecursiveMode::NonRecursive)
```
chris-olszewski, line 298: I changed this to be `NonRecursive` as I don't think we want to recursively watch the repo root: @arlyon

arlyon (approved):

arlyon:
Nice catch, LGTM

NicholasLYang (approved):

crates/turborepo-lib/src/globwatcher/mod.rs
```diff
@@ -66,7 +66,7 @@ impl<T: Watcher> HashGlobWatcher<T> {
     /// Watches a given path, using the flush_folder as temporary storage to
     /// make sure that file events are handled in the appropriate order.
     #[tracing::instrument(skip(self, token))]
-    pub async fn watch(&self, token: StopToken) {
+    pub async fn watch(&self, token: StopToken) -> Result<(), ConfigError> {
```
gsoltis, line 69: Do other callers need to handle this error as well?

chris-olszewski: Good catch, the only caller happens in a macro that ended up suppressing the lint.

gsoltis (approved):

PR #5050, opened 2023-05-22
feat(ecmascript-plugins): support swc ecmatransform plugins

kwonoj:
### Description

Part 1 for WEB-998.

This PR implements a new turbopack-ecmascript plugin for the swc's ecmatransform plugin. New transform `SwcEcmaTransformPluginsTransformer` accepts array of plugin module contains `CompiledPluginModuleBytes` from swc's wasm plugin then executes it. It doesn't utilize SWC's internal cache mechanism to keep compiled / serialized wasm module (`wasmer::Module`), instead expect turbopack manages it. For those reason, we only accepts `CompiledPluginModuleBytes` from swc as it is a serialized copy does not requires re-compilation, while swc have few other different representation to the module bytes.

It is exposed as a separate feature instead of being default: there are some of platforms / targets we can't support plugins yet, and making it default will makes `next-swc` to fail to build.

### DIFF
--- a/crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
+++ b/crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
@@ -1,0 +1,193 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use swc_core::ecma::ast::Program;
+use turbo_tasks::primitives::StringVc;
+use turbo_tasks_fs::FileSystemPathVc;
+use turbopack_core::issue::{Issue, IssueSeverity, IssueSeverityVc, IssueVc};
+use turbopack_ecmascript::{CustomTransformer, TransformContext};
+
+#[turbo_tasks::value(transparent)]
+pub struct PluginModule(
+    #[turbo_tasks(trace_ignore)]
+    #[cfg(feature = "swc_ecma_transform_plugin")]
+    pub swc_core::plugin_runner::plugin_module_bytes::CompiledPluginModuleBytes,
+    // Dummy field to avoid turbo_tasks macro complains about empty struct.
+    // This is due to we can't import CompiledPluginModuleBytes by default, it should be only
+    // available for the target / platforms can support swc plugins (which can build wasmer)
+    #[cfg(not(feature = "swc_ecma_transform_plugin"))] pub Option<()>,
+);
+
+#[turbo_tasks::value(shared)]
+struct UnsupportedSwcEcmaTransformPluginsIssue {
+    pub context: FileSystemPathVc,
+}
+
+#[turbo_tasks::value_impl]
+impl Issue for UnsupportedSwcEcmaTransformPluginsIssue {
+    #[turbo_tasks::function]
+    fn severity(&self) -> IssueSeverityVc {
+        IssueSeverity::Warning.into()
+    }
+
+    #[turbo_tasks::function]
+    fn category(&self) -> StringVc {
+        StringVc::cell("transform".to_string())
+    }
+
+    #[turbo_tasks::function]
+    async fn title(&self) -> Result<StringVc> {
+        Ok(StringVc::cell(format!(
+            "Unsupported SWC Ecma transform plugins on this platform."
+        )))
+    }
+
+    #[turbo_tasks::function]
+    fn context(&self) -> FileSystemPathVc {
+        self.context
+    }
+
+    #[turbo_tasks::function]
+    fn description(&self) -> StringVc {
+        StringVc::cell(
+            "Turbopack does not yet support running SWC ecma transform plugins on this platform."
+                .to_string(),
+        )
+    }
+}
+
+#[derive(Debug)]
+pub struct SwcEcmaTransformPluginsTransformer {
+    #[cfg(feature = "swc_ecma_transform_plugin")]
+    plugins: Vec<(PluginModuleVc, serde_json::Value)>,
+}
+
+impl SwcEcmaTransformPluginsTransformer {
+    #[cfg(feature = "swc_ecma_transform_plugin")]
+    pub fn new(plugins: Vec<(PluginModuleVc, serde_json::Value)>) -> Self {
+        Self { plugins }
+    }
+
+    #[cfg(not(feature = "swc_ecma_transform_plugin"))]
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+#[async_trait]
+impl CustomTransformer for SwcEcmaTransformPluginsTransformer {
+    #[cfg_attr(not(feature = "swc_ecma_transform_plugin"), allow(unused))]
+    async fn transform(&self, program: &mut Program, ctx: &TransformContext<'_>) -> Result<()> {
+        #[cfg(feature = "swc_ecma_transform_plugin")]
+        {
+            use std::{cell::RefCell, rc::Rc, sync::Arc};
+
+            use swc_core::{
+                common::{
+                    comments::SingleThreadedComments,
+                    plugin::{
+                        metadata::TransformPluginMetadataContext, serialized::PluginSerializedBytes,
+                    },
+                    util::take::Take,
+                },
+                ecma::ast::Module,
+                plugin::proxies::{HostCommentsStorage, COMMENTS},
+                plugin_runner::plugin_module_bytes::PluginModuleBytes,
+            };
+
+            let mut plugins = vec![];
+            for (plugin_module, config) in &self.plugins {
+                let plugin_module = &plugin_module.await?.0;
+
+                plugins.push((
+                    plugin_module.get_module_name().to_string(),
+                    config.clone(),
+                    Box::new(plugin_module.clone()),
+                ));
+            }
+
+            let should_enable_comments_proxy =
+                !ctx.comments.leading.is_empty() && !ctx.comments.trailing.is_empty();
+
+            //[TODO]: as same as swc/core does, we should set should_enable_comments_proxy
+            // depends on the src's comments availability. For now, check naively if leading
+            // / trailing comments are empty.
+            let comments = if should_enable_comments_proxy {
+                // Plugin only able to accept singlethreaded comments, interop from
+                // multithreaded comments.
+                let mut leading =
+                    swc_core::common::comments::SingleThreadedCommentsMapInner::default();
+                ctx.comments.leading.as_ref().into_iter().for_each(|c| {
+                    leading.insert(c.key().clone(), c.value().clone());
+                });
+
+                let mut trailing =
+                    swc_core::common::comments::SingleThreadedCommentsMapInner::default();
+                ctx.comments.trailing.as_ref().into_iter().for_each(|c| {
+                    trailing.insert(c.key().clone(), c.value().clone());
+                });
+
+                Some(SingleThreadedComments::from_leading_and_trailing(
+                    Rc::new(RefCell::new(leading)),
+                    Rc::new(RefCell::new(trailing)),
+                ))
+            } else {
+                None
+            };
+
+            let transformed_program =
+                COMMENTS.set(&HostCommentsStorage { inner: comments }, || {
+                    let module_program =
+                        std::mem::replace(program, Program::Module(Module::dummy()));
+                    let module_program =
+                        swc_core::common::plugin::serialized::VersionedSerializable::new(
+                            module_program,
+                        );
+                    let mut serialized_program =
+                        PluginSerializedBytes::try_serialize(&module_program)?;
+
+                    let transform_metadata_context = Arc::new(TransformPluginMetadataContext::new(
+                        Some(ctx.file_name_str.to_string()),
+                        //[TODO]: Support env-related variable injection, i.e process.env.NODE_ENV
+                        "development".to_string(),
+                        None,
+                    ));
+
+                    // Run plugin transformation against current program.
+                    // We do not serialize / deserialize between each plugin execution but
+                    // copies raw transformed bytes directly into plugin's memory space.
+                    // Note: This doesn't mean plugin won't perform any se/deserialization: it
+                    // still have to construct from raw bytes internally to perform actual
+                    // transform.
+                    for (_plugin_name, plugin_config, plugin_module) in plugins.drain(..) {
+                        let mut transform_plugin_executor =
+                            swc_core::plugin_runner::create_plugin_transform_executor(
+                                &ctx.source_map,
+                                &ctx.unresolved_mark,
+                                &transform_metadata_context,
+                                plugin_module,
+                                Some(plugin_config),
+                            );
+
+                        serialized_program = transform_plugin_executor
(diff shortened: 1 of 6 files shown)


kwonoj:
Looks like some of test snapshots failing with sourcemap differences, may comes with swc_core updates.

kdy1:
CI failure is not caused by swc_core update. See https://github.com/vercel/turbo/pull/5049
It does not have any failure.

kwonoj:
@kdy1 hmm, any idea sourcemap would make differences then? The changes in this PR is not even enabled since it is behind compile time flag, also it doesn't affect normal transform's sourcemap.

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,193 @@
...
+
+#[turbo_tasks::value(transparent)]
+pub struct PluginModule(
+    #[turbo_tasks(trace_ignore)]
+    #[cfg(feature = "swc_ecma_transform_plugin")]
+    pub swc_core::plugin_runner::plugin_module_bytes::CompiledPluginModuleBytes,
+    // Dummy field to avoid turbo_tasks macro complains about empty struct.
+    // This is due to we can't import CompiledPluginModuleBytes by default, it should be only
+    // available for the target / platforms can support swc plugins (which can build wasmer)
+    #[cfg(not(feature = "swc_ecma_transform_plugin"))] pub Option<()>,
```
sokra, line 17:
```suggestion
    #[cfg(not(feature = "swc_ecma_transform_plugin"))] pub (),
```

Why does it need to be an Option?

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,193 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use swc_core::ecma::ast::Program;
+use turbo_tasks::primitives::StringVc;
+use turbo_tasks_fs::FileSystemPathVc;
+use turbopack_core::issue::{Issue, IssueSeverity, IssueSeverityVc, IssueVc};
+use turbopack_ecmascript::{CustomTransformer, TransformContext};
+
+#[turbo_tasks::value(transparent)]
+pub struct PluginModule(
```
sokra, line 10: Documentation

sokra, line 10:
```suggestion
pub struct CompiledSwcPluginModule(
```

I would be a bit more specific in the naming

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,193 @@
...
+    #[turbo_tasks(trace_ignore)]
+    #[cfg(feature = "swc_ecma_transform_plugin")]
+    pub swc_core::plugin_runner::plugin_module_bytes::CompiledPluginModuleBytes,
+    // Dummy field to avoid turbo_tasks macro complains about empty struct.
+    // This is due to we can't import CompiledPluginModuleBytes by default, it should be only
+    // available for the target / platforms can support swc plugins (which can build wasmer)
+    #[cfg(not(feature = "swc_ecma_transform_plugin"))] pub Option<()>,
+);
+
```
sokra, line 19: I would expect a turbo_tasks::function `CompiledSwcPluginModuleVc::compile(FileSystemPathVc)` constructor here, which allows to compile a WASM file into a compiled module. This would utilize turbo tasks caching and invalidation logic, so updating the WASM will recompile the files...

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
kwonoj: No need, my oversight.

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
kwonoj: `CompiledPluginModuleBytes` is a struct points to compiled / serialized swc plugin. Compilation itself would occur when read wasm files, and constructing it into CompiledPluginModuleBytes itself immediately. My thought was if a module is created by `PluginModuleVc`, it is managed by turbo which corresponds to the specific compiled binary.

kwonoj: Thinking this bit more again, probably we could expose `new` to construct plugin module still.

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
kwonoj: I wanted to not to leak detail `compiled` part, revised struct name bit.

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,223 @@
...
+    into = "new",
+    cell = "new"
+)]
+pub struct SwcPluginModule(
+    #[turbo_tasks(trace_ignore)]
+    #[cfg(feature = "swc_ecma_transform_plugin")]
+    pub swc_core::plugin_runner::plugin_module_bytes::CompiledPluginModuleBytes,
+    // Dummy field to avoid turbo_tasks macro complains about empty struct.
+    // This is due to we can't import CompiledPluginModuleBytes by default, it should be only
+    // available for the target / platforms can support swc plugins (which can build wasmer)
```
alexkirsz, line 27:
```suggestion
    // Dummy field to avoid turbo_tasks macro complaining about empty struct.
    // This is because we can't import CompiledPluginModuleBytes by default, it should be only
    // available for the target / platforms that support swc plugins (which can build wasmer)
```

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,223 @@
...
+use turbo_tasks::primitives::StringVc;
+use turbo_tasks_fs::FileSystemPathVc;
+use turbopack_core::issue::{Issue, IssueSeverity, IssueSeverityVc, IssueVc};
+use turbopack_ecmascript::{CustomTransformer, TransformContext};
+
+/// A wrapper around a SWC's ecma transform wasm plugin module bytes, allowing
+/// it to operate with the turbo_task caching requirements.
+/// Internally this contains a `CompiledPluginModuleBytes`, which points to the
+/// compiled, serialized wasmer::Module instead of raw file bytes to reduce the
+/// cost of the compilation.
```
alexkirsz, line 13:
```suggestion
/// A wrapper around an SWC's ecma transform wasm plugin module bytes, allowing
/// it to operate with the turbo_tasks caching requirements.
/// Internally this contains a `CompiledPluginModuleBytes`, which points to the
/// compiled, serialized wasmer::Module instead of raw file bytes to reduce the
/// cost of the compilation.
```

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    #[turbo_tasks::function]
+    fn category(&self) -> StringVc {
+        StringVc::cell("transform".to_string())
+    }
+
+    #[turbo_tasks::function]
+    async fn title(&self) -> Result<StringVc> {
+        Ok(StringVc::cell(format!(
+            "Unsupported SWC Ecma transform plugins on this platform."
```
alexkirsz, line 70:
```suggestion
            "Unsupported SWC EcmaScript transform plugins on this platform."
```

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    #[turbo_tasks::function]
+    fn context(&self) -> FileSystemPathVc {
+        self.context
+    }
+
+    #[turbo_tasks::function]
+    fn description(&self) -> StringVc {
+        StringVc::cell(
+            "Turbopack does not yet support running SWC ecma transform plugins on this platform."
```
alexkirsz, line 82:
```suggestion
            "Turbopack does not yet support running SWC EcmaScript transform plugins on this platform."
```

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,223 @@
...
+                    plugin_module.get_module_name().to_string(),
+                    config.clone(),
+                    Box::new(plugin_module.clone()),
+                ));
+            }
+
+            let should_enable_comments_proxy =
+                !ctx.comments.leading.is_empty() && !ctx.comments.trailing.is_empty();
+
+            //[TODO]: as same as swc/core does, we should set should_enable_comments_proxy
```
alexkirsz, line 141: Do we have an issue tracking this? If so, add the ticket number here.

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,223 @@
...
+    fn description(&self) -> StringVc {
+        StringVc::cell(
+            "Turbopack does not yet support running SWC ecma transform plugins on this platform."
+                .to_string(),
+        )
+    }
+}
+
+#[derive(Debug)]
+pub struct SwcEcmaTransformPluginsTransformer {
```
alexkirsz, line 89: Comment this public item.

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,223 @@
...
+    #[cfg(feature = "swc_ecma_transform_plugin")]
+    plugins: Vec<(SwcPluginModuleVc, serde_json::Value)>,
+}
+
+impl SwcEcmaTransformPluginsTransformer {
+    #[cfg(feature = "swc_ecma_transform_plugin")]
+    pub fn new(plugins: Vec<(SwcPluginModuleVc, serde_json::Value)>) -> Self {
```
alexkirsz, line 96: To make the API clearer, this should either accept:
```suggestion
    pub fn new(plugins: Vec<SwcEcmaTransformPlugin>) -> Self {
```
where
```rust
pub struct SwcEcmaTransformPlugin {
    module: SwcPluginModuleVc,
    config: serde_json::Value
}
```

or

```suggestion
    pub fn new(plugins: Vec<(SwcPluginModuleVc, SwcPluginConfig)>) -> Self {
```

where `SwcPluginConfig` is a newtype:

```rust
pub struct SwcPluginConfig(serde_json::Value);
```

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,223 @@
...
+                    let module_program =
+                        swc_core::common::plugin::serialized::VersionedSerializable::new(
+                            module_program,
+                        );
+                    let mut serialized_program =
+                        PluginSerializedBytes::try_serialize(&module_program)?;
+
+                    let transform_metadata_context = Arc::new(TransformPluginMetadataContext::new(
+                        Some(ctx.file_name_str.to_string()),
+                        //[TODO]: Support env-related variable injection, i.e process.env.NODE_ENV
```
alexkirsz, line 180: Do we have an issue tracking this?

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
```diff
@@ -0,0 +1,223 @@
...
+impl SwcEcmaTransformPluginsTransformer {
+    #[cfg(feature = "swc_ecma_transform_plugin")]
+    pub fn new(plugins: Vec<(SwcPluginModuleVc, serde_json::Value)>) -> Self {
+        Self { plugins }
+    }
+
+    #[cfg(not(feature = "swc_ecma_transform_plugin"))]
+    pub fn new() -> Self {
+        Self {}
+    }
```
alexkirsz, line 103: Why allow creating this object at all if that feature is not enabled? This whole module could be gated under `#[cfg(feature = "swc_ecma_transform_plugin")]`

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
kwonoj: no, this is actually more close to swc upstream issue.

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
kwonoj: There isn't. Also it need some audit if we're going to make transformcontext to pass Option<Comment> to allow this or not.

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
kwonoj: This is sort of intentional: currently upstream swc's config itself is not finalized for the plugins, so I'd rather take it out separately for now and then visit everything at once with changes. It'll be breaking changes, but swc's plugin is experimental anyway.

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
kwonoj: If this whole module is behind of flag, turbo_tasks's `register` fn panics when build. I tried to make it work but couldn't.

alexkirsz: Ah yes, I've run into this before. This is because the macro will still pick up the structures. I've created [WEB-1102](https://linear.app/vercel/issue/WEB-1102/cant-use-rust-features-with-turbo-tasks) to track this. Could you add a short comment to refer to this issue?

arlyon (approved):

crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs
kwonoj: Yup, added.

jridgewell (approved):

PR #5054, opened 2023-05-22
Add codemod to transform literals to wildcards

nathanhammond:
In 1.10 all existing literal env var names become wildcards. Users with literals env vars that could be confused as wildcards need to be escaped. This automates that process.


docs/pages/repo/docs/reference/codemods.mdx
```diff
@@ -327,3 +328,57 @@ Run the codemod:
 ```sh
 npx @turbo/codemod stabilize-env-mode
 ```
+
+### `transform-literals-to-wildcards`
```
mehulkar, line 332:
```suggestion
### `transform-env-literals-to-wildcards`
```

packages/turbo-codemod/src/commands/migrate/steps/getTransformsForMigration.ts
```diff
@@ -13,13 +13,24 @@ function getTransformsForMigration({
...
+
+  transforms.sort((a, b) => {
+    if (gt(a.introducedIn, b.introducedIn)) {
+      return 1;
+    } else if (eq(a.introducedIn, b.introducedIn)) {
+      return 0;
+    } else {
+      return -1;
+    }
+  });
```
mehulkar, line 31: I can never remember what positive and negative means. Is this sorting so more recent transforms run first or vice versa? a comment would be useful if that's why we're sorting in the first place

packages/turbo-codemod/src/transforms/transform-literals-to-wildcards.ts
```diff
@@ -0,0 +1,124 @@
...
+  if (Array.isArray(globalPassThroughEnv)) {
+    config.globalPassThroughEnv = globalPassThroughEnv.map(transformEnvVarName);
+  }
+
+  return migrateTaskConfigs(config);
+}
+
+function migrateTaskConfigs(config: TurboJsonSchema) {
+  for (const [_, taskDef] of Object.entries(config.pipeline)) {
+    let { env, passThroughEnv } = taskDef;
```
mehulkar, line 47: Could pull this list of known env var fields

```js
for const field of knownEnvFields {
  taskDef[field] = taskDef[field].map(transformEnvVarName)
}
```

packages/turbo-codemod/src/transforms/transform-literals-to-wildcards.ts
```diff
@@ -0,0 +1,124 @@
...
+  try {
+    packageJSON = fs.readJSONSync(packageJsonPath);
+  } catch (e) {
+    // readJSONSync probably failed because the file doesn't exist
+  }
+
+  if ("turbo" in packageJSON) {
+    return runner.abortTransform({
+      reason:
+        '"turbo" key detected in package.json. Run `npx @turbo/codemod transform create-turbo-config` first',
```
mehulkar, line 84: When can we just stop handling this use case across the board?

packages/turbo-types/src/types/config.ts
```diff
@@ -270,3 +270,4 @@ export type OutputMode =
   | "none";
 
 export type AnchoredUnixPath = string;
+export type EnvWildcard = string;
```
mehulkar, line 273: does this get documented somehow?

mehulkar (approved):
lgtm! some suggestions inline, nothing blocking

packages/turbo-types/src/types/config.ts
nathanhammond: I've got that over in the wildcard PR; I did this to minimize conflicts. This ends up documented in the schema, but never gets presented to the user in VS Code. 😕

packages/turbo-codemod/src/transforms/transform-literals-to-wildcards.ts
nathanhammond: Probably now; the transforms weren't sorted previously resulting in this being a requirement to not have everything blow up.

packages/turbo-codemod/src/transforms/transform-literals-to-wildcards.ts
nathanhammond: This was a "keep with the pattern" since we do exactly this in four other codemods. 🤷‍♂️

packages/turbo-codemod/src/commands/migrate/steps/getTransformsForMigration.ts
nathanhammond: This sorts ascending; transforms run from oldest to newest, allowing us to rely on the previous transforms having already been completed.

As for remembering:
- JS sort is default ascending.
- 1 > 0, so return 1 when it is greater.
- 0 = 0, so map 0 to equal.
- -1 < 0, so -1 maps to smaller.

If you pass negative zero in, well, you get what was coming. (I don't know if JS cares or not.)

tknickman: 😬 good catch!

packages/turbo-codemod/src/transforms/transform-literals-to-wildcards.ts
tknickman: We also still support running these as one offs, so it's possible someone will get the order wrong themselves. So I would still be in favor of keeping this check around for the nice error - but it is definitely a little painful

tknickman (approved):

packages/turbo-codemod/src/transforms/transform-literals-to-wildcards.ts
tknickman: Since the purpose of these codemods are intended for migration, I think we should leave them in. 

I think these have a different contract than the core CLI, where it's totally fine to remove the features / logs / checks etc. after deprecation.

mehulkar: For all new codemods going forward too?

PR #5062, opened 2023-05-22
ci(turbo): allow publishing from non-main

tknickman:
### Description

In preparation for release branches, allow publishing from non-main branches


cli/Makefile
```diff
@@ -228,7 +228,6 @@ stage-release: cmd/turbo/version.go
 	cat $(CLI_DIR)/../version.txt
 	git diff -- $(CLI_DIR)/../version.txt
 	git status
-	@test main = "`git rev-parse --abbrev-ref HEAD`" || (echo "Refusing to publish from non-main branch `git rev-parse --abbrev-ref HEAD`" && false)
```
tknickman, line 231: We should probably still echo a very loud warning just to make sure you're sure

mehulkar: - could add it to the workflow: https://github.com/vercel/turbo/blob/main/.github/workflows/turborepo-release.yml
- prevent publishes when not in CI (`CI` env is present or something)

mehulkar (approved):

cli/Makefile
tknickman: Yea those are both great suggestions 👍🏼 will add

PR #5065, opened 2023-05-22
feat(turborepo): Port HTTP Cache

NicholasLYang:
### Description

Ports HTTP cache using `cache_archive` 

### Testing Instructions

Implemented a round-trip test using the `vercel-api-mock` crate


crates/turborepo-vercel-api-mock/src/lib.rs
```diff
@@ -82,7 +94,69 @@ pub async fn start_test_server(port: u16) -> Result<()> {
...
+        )
+        .route(
+            "/v8/artifacts/:hash",
+            head(|Path(hash): Path<String>| async move {
+                let duration = durations3.lock().await.get(&hash).cloned().unwrap_or(0);
+                let mut headers = HeaderMap::new();
+
+                headers.insert(
+                    "x-artifact-duration",
+                    HeaderValue::from_str(&duration.to_string()).unwrap(),
```
nathanhammond, line 153: A bit surprised that this much ceremony is required here.

> You require MOAR TRAITS.

crates/turborepo-api-client/src/lib.rs
```diff
@@ -240,7 +255,162 @@ impl APIClient {
...
+        token: &str,
+        team_id: &str,
+        team_slug: Option<&str>,
+        use_preflight: bool,
+    ) -> Result<Response> {
+        self.get_artifact(hash, token, team_id, team_slug, use_preflight, Method::GET)
+            .await
+    }
+
+    pub async fn artifact_exists(
```
nathanhammond, line 320: Want to rename this to `head_artifact` so badly. (not worth it)

crates/turborepo-api-client/src/lib.rs
```diff
@@ -240,7 +255,162 @@ impl APIClient {
...
+        let mut request_url = self.make_url(&format!("/v8/artifacts/{}", hash));
+        let mut allow_auth = true;
+
+        if self.use_preflight {
+            let preflight_response = self
+                .do_preflight(
+                    token,
+                    &request_url,
+                    "PUT",
+                    "Content-Type, x-artifact-duration, Authorization, User-Agent, x-artifact-tag",
```
nathanhammond, line 275: Sort?

crates/turborepo-api-client/src/lib.rs
```diff
@@ -240,7 +255,162 @@ impl APIClient {
         })
     }
 
-    pub fn new(base_url: impl AsRef<str>, timeout: u64, version: &str) -> Result<Self> {
+    pub async fn put_artifact(
+        &self,
+        hash: &str,
+        artifact_body: &[u8],
+        duration: u32,
```
nathanhammond, line 262: We use 32 and 64 in different places, align 'em?

crates/turborepo-api-client/src/lib.rs
```diff
@@ -240,7 +255,162 @@ impl APIClient {
...
+            Url::parse(location)?
+        } else {
+            response.url().clone()
+        };
+
+        let allowed_headers = headers
+            .get("Access-Control-Allow-Headers")
+            .map_or("", |h| h.to_str().unwrap_or(""));
+
+        let allow_auth = allowed_headers.to_lowercase().contains("authorization");
```
nathanhammond, line 400: `x-authorization-foo` also matches this. Needs boundary checking.

Let's get a test for this too.

crates/turborepo-vercel-api-mock/src/lib.rs
```diff
@@ -82,7 +94,69 @@ pub async fn start_test_server(port: u16) -> Result<()> {
...
+                    let mut file = OpenOptions::new()
+                        .append(true)
+                        .create(true)
+                        .open(&file_path)
+                        .unwrap();
+
+                    let duration = headers
+                        .get("x-artifact-duration")
+                        .and_then(|header_value| header_value.to_str().ok())
+                        .and_then(|duration| duration.parse::<u64>().ok())
```
nathanhammond, line 113: 64 is here.

crates/turborepo-api-client/src/lib.rs
```diff
@@ -111,10 +120,16 @@ pub struct UserResponse {
     pub user: User,
 }
 
+pub struct PreflightResponse {
+    location: Url,
+    allow_auth: bool,
```
nathanhammond, line 125: I prefer longer variable names. 🤷‍♂️

```suggestion
    allow_authorization_header: bool,
```

crates/turborepo-api-client/src/lib.rs
```diff
@@ -240,7 +255,162 @@ impl APIClient {
...
+            .header("Access-Control-Request-Method", request_method)
+            .header("Access-Control-Request-Headers", request_headers)
+            .header("Authorization", format!("Bearer {}", token));
+
+        let response = retry::make_retryable_request(request_builder).await?;
+
+        let headers = response.headers();
+        let location = if let Some(location) = headers.get("Location") {
+            let location = location.to_str()?;
+            Url::parse(location)?
```
nathanhammond, line 391: This parses the location as a fully-qualified domain. I suspect we should be using `join` here.

One test, please!

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -114,6 +127,7 @@ impl CacheWriter {
             header.set_entry_type(EntryType::Directory);
         } else if file_info.is_file() {
             header.set_entry_type(EntryType::Regular);
+            header.set_size(file_info.len());
```
nathanhammond, line 130: whomp whomp. Good catch.

crates/turborepo-cache/src/http.rs
nathanhammond, line 1: Have not yet reviewed this file, submitting review part one first in case I get distracted.

nathanhammond (commented):
Review part one, all but `http.rs`

crates/turborepo-cache/src/http.rs
```diff
@@ -0,0 +1,253 @@
...
+                CacheError::ApiClientError(
+                    turborepo_api_client::Error::ReqwestError(e),
+                    Backtrace::capture(),
+                )
+            })?
+        };
+
+        let files = Self::restore_tar(&self.repo_root, &body)?;
+
+        Ok((files, duration))
```
nathanhammond, line 151: Can we pretty please not return a tuple here?

crates/turborepo-cache/src/http.rs
```diff
@@ -0,0 +1,253 @@
...
+             path: AnchoredSystemPathBuf::from_raw("package-lock.json").unwrap(),
+             contents: "Badlands"
+        },
+        TestFile {
+            path: AnchoredSystemPathBuf::from_raw("src/main.js").unwrap(),
+            contents: "Tree of Life"
+        }
+    ], 12845, "The Gleaners and I")]
+    #[tokio::test]
+    async fn test_round_trip(files: Vec<TestFile>, duration: u32, hash: &str) -> Result<()> {
```
nathanhammond, line 210: Teach me! How does this get invoked?

nathanhammond (commented):
Review part 2!

crates/turborepo-vercel-api-mock/src/lib.rs
NicholasLYang: Yeah...this is the unfortunate reality of rust crates. Either very trait heavy or type heavy, either way it can get confusing.

crates/turborepo-api-client/src/lib.rs
NicholasLYang: This is the Go behavior, not sure if we should fix it in the port or after

crates/turborepo-vercel-api-mock/src/lib.rs
NicholasLYang: Damn good catch

crates/turborepo-api-client/src/lib.rs
NicholasLYang: This is also the same behavior as the Go code, should we change or keep the same?

crates/turborepo-cache/src/http.rs
```diff
@@ -0,0 +1,253 @@
...
+        duration: u32,
+        token: &str,
+    ) -> Result<(), CacheError> {
+        let mut artifact_body = Vec::new();
+        self.write(&mut artifact_body, anchor, files).await?;
+
+        let tag = self
+            .signer_verifier
+            .as_ref()
+            .map(|signer| signer.generate_tag(hash.as_bytes(), &artifact_body))
```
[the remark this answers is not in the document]
NicholasLYang: I'm actually confused about this. We call `ioutil.ReadAll` to get the artifact body which we then use to create the artifact signature. Does that block on `cache.write` finishing? I

crates/turborepo-cache/src/http.rs
NicholasLYang: Would you prefer a struct? IMO a tuple is just fine here since they are separate types

crates/turborepo-api-client/src/lib.rs
nathanhammond: That's a bug, we should fix it. Since I think it is just `Url::join` vs. `Url::parse` let's bundle it.

crates/turborepo-api-client/src/lib.rs
nathanhammond: It's a bug, and it's a straightforward regex:

```
let authorization_regex = Regex::new(r"(?i)(?:^|,) *authorization *(?:,|$)").unwrap();

authorization_regex.is_match(allowed_header);
```

Exploded:
```
(?i) - case insensitive
(?: - non-capturing alternation
  ^ - start of string
  | - or
  , - comma reset
)
 * - optional whitespace
authorization - literal string
 * - optional whitespace
(?: - non-capturing alternation
  , - comma
  | - or
  $ - end of string
)

***

Could be stricter on the whitespace if we wanted.

crates/turborepo-cache/src/http.rs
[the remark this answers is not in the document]
nathanhammond: Oooo, [it doesn't do it this way in the current version](https://github.com/vercelturbo/blob/main/cli/internal/cacheitem/cacheitem.go#L81). I had a a thing where the writer would create the sha simultaneously at one point, by splitting it.

I may have deleted it when I was trying to figure out why the whole thing slowed down and the eventual `gzip` => `zstd` migration.

crates/turborepo-cache/src/http.rs
nathanhammond: I basically always prefer structs because on numerous occasions I have screwed up tuples in a way where having the name would have rescued me. Yeah, they're different types now, but code changes over time and we end up with the ridiculous return values we have in the Go version.

nathanhammond: Actually, this feedback comment doesn't follow my rule regarding code style linting (enforced automatically with a linter or GTFO), Clippy doesn't have this lint to enable, so I'm marking this as resolved.

I still hold this opinion, but it's not blocking unless I make it automated.

nathanhammond (approved):
Fix-ship with the authorization header tweak + test and the location header tweak + test.

The `CacheResponse` stuff looks good.

PR #5066, opened 2023-05-22
fix(daemon): add short sleep to repo root removal

chris-olszewski:
### Description

Add a short delay to our repo root removal to ensure that we are watching the config flush directory before we attempt to flush it. This should be a super short lived hack

### Testing Instructions

CI


gsoltis (approved):
Some of the GH Actions runners have a _lot_ of variance...

PR #5070, opened 2023-05-23
fix(daemon): plumb through flush watch errors

chris-olszewski:
### Description

First pass as plumbing the status of watching the flush directory to the glob watcher and watcher configs.
Notable changes:
 - Before we start to process flush commands we wait until we get confirmation that we are watching the flush directory so we can properly handle flush commands
 - Failing to watch the flush directory will cause the watch future held by the server to exit. The daemon server won't shut down, but will respond with errors to all glob watching related RPC requests

Reviewer notes:
 - The current timeout logic means that if erroring on setting up the flush directory takes over 500ms we skip the flush and proceed as usual
 - We now return errors for `get_changed_outputs` RPC call
 - I'm unsure of the changes to the `serve` function. I've manually tested them, but they feel off. In the future when we possibly have multiple long running futures this won't scale.

### Testing Instructions

Unit test should now consistently pass and not depend on watching the flush directory succeeding before the watcher is converted into a stream.

Did some manual testing to make sure that we don't exit if glob watching fails and instead report an error to clients:
From the daemon logs:
```
2023-05-23T15:36:56.894-0700 [TRACE] globwatch: watching "/private/tmp/turborepo-netflix-cache-repro"              
2023-05-23T15:36:56.898-0700 [TRACE] globwatch: watching flush dir: "/private/var/folders/3m/rxkycvgs5jgfvs0k9xcgp6
km0000gn/T/turbod/3e1692e405755b5a/flush"                                                                          
ERROR faked failure for globawatch                                                                                 
ERROR Globwatch config error: ServerFailedToStart                                                                  
2023-05-23T15:37:21.582-0700 [TRACE] log: registering event source with poller: token=Token(1), interests=READABLE 
| WRITABLE
```
From the turbo output:
```
web:build: Failed to check if we can skip restoring outputs for web#build: rpc error: code = Internal desc = failed to watch flush directory. Proceeding to check cache
```

### DIFF
--- a/crates/turborepo-globwatch/src/lib.rs
+++ b/crates/turborepo-globwatch/src/lib.rs
@@ -34,7 +34,10 @@ use itertools::Itertools;
 use merge_streams::MergeStreams;
 pub use notify::{Error, Event, Watcher};
 pub use stop_token::{stream::StreamExt, StopSource, StopToken, TimedOutError};
-use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
+use tokio::sync::{
+    mpsc::{UnboundedReceiver, UnboundedSender},
+    watch,
+};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tracing::{event, span, trace, warn, Level};
 
@@ -45,6 +48,7 @@ pub struct GlobWatcher {
     flush_dir: PathBuf,
 
     config: UnboundedReceiver<WatcherCommand>,
+    setup_handle: tokio::task::JoinHandle<Result<(), Error>>,
 }
 
 impl GlobWatcher {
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -281,9 +281,15 @@ impl<T: Watcher + Send + 'static> proto::turbod_server::Turbod for DaemonServer<
...
 
-        Ok(tonic::Response::new(proto::GetChangedOutputsResponse {
-            changed_output_globs: changed.into_iter().collect(),
-        }))
+        match changed {
+            Ok(changed) => Ok(tonic::Response::new(proto::GetChangedOutputsResponse {
+                changed_output_globs: changed.into_iter().collect(),
+            })),
+            Err(e) => {
+                error!("failed to watch flush directory: {:?}", e);
```
arlyon, line 289: Is this failing to register a watcher or not having the flush itself fail?

crates/turborepo-globwatch/src/lib.rs
```diff
@@ -124,15 +139,29 @@ impl GlobWatcher {
...
-        let flush_dir = Arc::new(self.flush_dir);
+        let flush_dir = Arc::new(flush_dir);
         let flush = Arc::new(Mutex::new(HashMap::<u64, oneshot::Sender<()>>::new()));
 
-        Box::pin(
+        // Wait until we have watched the flush directory, before we start processing
+        // events
+        setup_handle
+            .await
+            .expect("globwatch setup thread panicked")?;
```
arlyon, line 162: Great, so we can queue up flush requests and guarantee that the dir is watched so they don't get lost. And it looks like the time-to-first-request for startup is not affected either. :)

arlyon: Something like this:

```rust
        let setup_stream = setup_handle.into_stream().map(|_| WatcherCommand::Start);
        (
            UnboundedReceiverStream::new(self.stream).map(Either::Left),
            setup_stream
                .chain(UnboundedReceiverStream::new(self.config))
                .map(Either::Right),
        )
```

crates/turborepo-lib/src/daemon/server.rs
chris-olszewski: Can be both so made the message more generic.

arlyon (approved):

PR #5072, opened 2023-05-23
improve asset path

sokra:
### Description

* include basename in static path
* only use 8 chars for hash

This aligns with next.js, which seem to be a reasonable default.

Production mode can omit the basename and only include the hash.

### DIFF
--- a/crates/turbopack-dev/src/chunking_context.rs
+++ b/crates/turbopack-dev/src/chunking_context.rs
@@ -332,11 +332,25 @@ impl ChunkingContext for DevChunkingContext {
     }
 
     #[turbo_tasks::function]
-    fn asset_path(&self, content_hash: &str, basename: &str, extension: &str) -> FileSystemPathVc {
-        self.asset_root_path.join(&format!(
-            "{basename}.{content_hash}.{extension}",
-            content_hash = &content_hash[..8]
-        ))
+    async fn asset_path(
+        &self,
+        content_hash: &str,
+        original_asset_ident: AssetIdentVc,
+    ) -> Result<FileSystemPathVc> {
+        let source_path = original_asset_ident.path().await?;
+        let basename = source_path.file_name();
+        let asset_path = match source_path.extension() {
+            Some(ext) => format!(
+                "{basename}.{content_hash}.{ext}",
+                basename = &basename[..basename.len() - ext.len() - 1],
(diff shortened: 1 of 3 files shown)


crates/turbopack-dev/src/chunking_context.rs
```diff
@@ -332,9 +332,11 @@ impl ChunkingContext for DevChunkingContext {
     }
 
     #[turbo_tasks::function]
-    fn asset_path(&self, content_hash: &str, extension: &str) -> FileSystemPathVc {
-        self.asset_root_path
-            .join(&format!("{content_hash}.{extension}"))
+    fn asset_path(&self, content_hash: &str, basename: &str, extension: &str) -> FileSystemPathVc {
```
alexkirsz, line 335: Nit: could we have the basename come before the content_hash in the order of the arguments, to match what's actually getting written to disk?

```suggestion
    fn asset_path(&self, basename: &str, content_hash: &str, extension: &str) -> FileSystemPathVc {
```

crates/turbopack-static/src/lib.rs
```diff
@@ -150,9 +150,15 @@ impl Asset for StaticAsset {
...
-            Some(ext) => self.context.asset_path(&content_hash_b16, ext),
-            None => self.context.asset_path(&content_hash_b16, "bin"),
+        let source_path = source_path.await?;
+        let basename = source_path.file_name();
+        let asset_path = match source_path.extension() {
+            Some(ext) => self.context.asset_path(
+                &content_hash_b16,
+                &basename[..basename.len() - ext.len() - 1],
+                ext,
+            ),
```
alexkirsz, line 160:
```suggestion
            Some(ext) => self.context.asset_path(
                &basename[..basename.len() - ext.len() - 1],
                &content_hash_b16,
                ext,
            ),
```

alexkirsz (approved):

crates/turbopack-dev/src/chunking_context.rs
sokra: The format is an implementation detail of turbopack-dev.

Production would probably use a different format (and won't use `basename` at all).

I guess we should probably pass `original_ident: AssetIdentVc` instead and do all the extension & basename stuff in turbopack-dev.

Issue #4932, opened 2023-05-13
Add ability to cache docker images


jakeleventhal:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

Turbo should be able to cache docker images and use those with cache hits.

### Describe the solution you'd like

Implement caching with docker images.

### Describe alternatives you've considered

N/A

chris-olszewski:
Hi Jake, could you elaborate a little bit more on what you mean by having Turborepo cache docker images? Docker itself already has fairly advanced image caching.

jakeleventhal:
sure, what im referring to is being able to cache the output of a docker build much in the same way that i would cache any other output in turbo. For instance, i should be able to run a script "image" via `turbo image` which builds a docker image, and specify the image as an output (similar to how bazel works). then in ci, my remote cache should be able to recognize a cache hit, and download the docker artifact as an output

mehulkar:
@jakeleventhal what happens when you do that today? I imagine the artifact size is pretty large, so writing to cache (especially remote cache) could take a while, but turbo shouldn't have an opinion over what the outputs are. If you're running into issues, can you provide a reproduction repo?

mehulkar:
sorry, it's been ~a decade since I used docker, does it not create the image on your local disk first? Is it no longer a local directory or file on your filesystem when you create the image?

jakeleventhal:
They get saved to a ~/.docker directory

mehulkar:
Is it possible to customize that location such that you could put it in the repo location? (I assume that even if that is possible it may interfere with other docker operations). That said, it seems like there are a couple general purpose features here:

- `outputs` outside the repo root (which is already possible, I think)
- or maybe a `--cache-dir` per task, so in your case the `image` task could specify the location

Will bring this up with the team!

jakeleventhal:
There isn't a way to specify this unfortunately.

PR #5078, opened 2023-05-23
Greatly expand the content for caching information.

nathanhammond:
As part of documenting all of the new environment variable features, I exploded apart the caching content into multiple pages. The environment variables section is _by far_ the longest and most-complex portion of our interface.


tknickman (approved):
A few comments, but overall a massive improvement. Also we should add redirects for the old routes.

docs/pages/repo/docs/core-concepts/caching.mdx
```diff
@@ -44,84 +42,23 @@ Let's say that you run the task again without changing any of its inputs:
...
-    }
-  }
-}
-```
+Restoring files and logs from the cache happens near-instantaneously. This can reduce your build times from minutes or hours down to seconds or milliseconds. Although specific results will vary depending on the shape and granularity of your codebase's dependency graph, most teams find that they can reduce their overall monthly build time by around 40-85% with Turborepo's caching.
 
 ## Turn off caching
 
-Sometimes you really don't want to write the cache output (e.g. when you're using [`next dev`](https://nextjs.org/docs/api-reference/cli#development) or `react-scripts start` for live reloading). To disable cache writes, append `--no-cache` to any command:
+In some environments you don't want to write the cache output. To disable cache writes, append `--no-cache` to any command. This will run `dev` (and all tasks that it dependsOn) in all workspaces, but it won't cache the output.
```
gsoltis, line 53: The context for the last sentence comes in the commandline example, which is a little awkward.

Maybe consider moving the last sentence after the commandline example? Or ending it with ":"? As written it's not immediately clear what "this" is, or where `dev` came from.

docs/pages/repo/docs/core-concepts/caching.mdx
```diff
@@ -44,84 +42,23 @@ Let's say that you run the task again without changing any of its inputs:
...
 ```shell
-# Run `dev` npm script in all workspaces in parallel,
-# but don't cache the output
 turbo run dev --no-cache
 ```
 
 Note that `--no-cache` disables cache writes but does not disable cache reads. If you want to disable cache reads, use the `--force` flag.
 
-You can also disable caching on specific tasks by setting the [`pipeline.<task>.cache`](/repo/docs/reference/configuration#cache) configuration to `false`:
+You can also permanently disable caching on specific tasks by setting the [`pipeline.<task>.cache`](/repo/docs/reference/configuration#cache) configuration to `false`:
```
gsoltis, line 61: I don't know that "permanently" belongs here. I mean, yes, for some definitions of permanent, but if I remove the config setting, it's no longer disabled, which is not permanent. 

Maybe there's another way to phrase it? "You can configure specific tasks to skip writing to the cache by ..." or something like that?

gsoltis (commented):
I haven't finished going through everything yet. But I didn't want to block on getting through all of it.

docs/pages/repo/docs/core-concepts/caching/environment-variable-inputs.mdx
```diff
@@ -0,0 +1,279 @@
...
+
+- A `*` anywhere in the pattern will match **zero or more** characters.
+- A leading `!` means that the entire pattern will be negated.
+- A `!` in any position other than the first position is a literal `!`.
+- Characters with special meaning (leading `!` and `*`) may be escaped by placing a `\` in front of them to get a literal value.
+
+### Examples
+
+- `"*"`: matches every environment variable.
+- `"!*"`: excludes from the match set all environment variables.
```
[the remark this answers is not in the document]
nathanhammond: Yeah, this one doesn't actually need a caveat.

docs/pages/repo/docs/core-concepts/caching/environment-variable-inputs.mdx
```diff
@@ -0,0 +1,279 @@
...
+- `"FOO*"`: matches `FOO`, `FOOD`, and `FOO_FIGHTERS`.
+- `"FOO\*"`: because this is specified in JSON, this resolves to `"FOO*"` and matches `FOO`, `FOOD`, and `FOO_FIGHTERS`.
+- `"FOO\\*"`: matches a single environment variable named `FOO*`.
+- `"!FOO*"`: excludes from the match set all environment variables that start with `FOO`.
```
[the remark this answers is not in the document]
nathanhammond: My goal was to make sure people didn't think they could provide a list of "only excludes" and have that do something. I'll take a swing at explaining that.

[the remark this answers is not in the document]
nathanhammond: Oh, I already mention it above. I'll just delete it here, the examples are non-normative anyway.

docs/pages/repo/docs/core-concepts/caching/environment-variable-inputs.mdx
```diff
@@ -0,0 +1,279 @@
...
+
+## Framework Inference
+
+<Callout type="info">
+  This feature can be disabled by passing [`--framework-inference=false`](../reference/command-line-reference#--framework-inference) to your `turbo` command.
+</Callout>
+
+By default, Turborepo attempts to detect the framework for workspaces inside of your turborepo and uses this to help ensure that all of the default environment variables are properly considered for task hashes.
+
+If Turborepo successfully detects your framework you do not need to manually specify certain framework-specific environment variables inside of `turbo.json`'s pipeline configuration. The supported frameworks and the environment variables that Turborepo will detect and the env wildcards that will be include into the task's `env` key:
```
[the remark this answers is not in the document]
nathanhammond: It shows up in run summaries, and dry runs too; I think that is what we should tell people to look at.

docs/pages/repo/docs/core-concepts/caching/to-cache-or-not-to-cache.mdx
```diff
@@ -0,0 +1,51 @@
...
+  "$schema": "https://turbo.build/schema.json",
+  "pipeline": {
+    "lint": {
+      "cache": true
+    }
+  }
+}
+```
+
+Turborepo actually defaults to `"cache": true`, so you don't actually need to specify this value, meaning that this is an identical configuration:
```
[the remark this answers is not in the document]
nathanhammond: well, actually...

docs/pages/repo/docs/core-concepts/caching.mdx
nathanhammond: persistently?

docs/pages/repo/docs/core-concepts/caching/to-cache-or-not-to-cache.mdx
```diff
@@ -0,0 +1,51 @@
+---
+title: To Cache or Not to Cache
+description: Whether 'tis more performant to cache.
+---
+
+import Callout from "../../../../../components/Callout";
+
+# To Cache or Not To Cache
```
gsoltis, line 8: What do you think about starting off with a sentence recommending caching by default? A user who wants a quick answer to "should I cache or not" should probably get a "yes, cache" and then revisit for more details if something goes wrong.

docs/pages/repo/docs/core-concepts/caching/what-to-cache.mdx
```diff
@@ -0,0 +1,85 @@
+---
+title: What to Cache?
+description: The configuration of Turborepo `outputs`
+---
+
+import Callout from "../../../../../components/Callout";
+
+# What to Cache?
+
+By default, the console output (also known as `stdout`) from all tasks is _always_ captured by Turborepo. If your task does not emit any files (e.g. unit tests with Jest) you can omit `outputs`. Beyond that, you will generally need to specify which files on disk you need to capture (and restore!) in order to be able to run subsequent dependent tasks.
```
gsoltis, line 10: As above, we cache `stdout` _and_ `stderr`. You can probably just drop the part in parenthesis. "console output" covers both.

gsoltis, line 10: "capture (and restore!)" makes sense to me because I know what it's referring to, but the user is not the one doing the capturing or restoring. The capturing and restoring is also not guaranteed to happen (they could just stay unchanged where the task wrote them), so might be the wrong level of detail for this point. Maybe there's another way to word this? 
"...specify which output files are required to be able to run ..."
"specify which of your task's output files are used by dependent tasks"
something else?

docs/pages/repo/docs/core-concepts/caching/what-to-cache.mdx
```diff
@@ -0,0 +1,85 @@
...
+  "$schema": "https://turbo.build/schema.json",
+  "pipeline": {
+    "build": {
+      "outputs": ["dist/**"]
+    }
+  }
+}
+```
+
+This idea is simple but can grow in complexity depending upon the tools you are using in your pipeline.
```
gsoltis, line 27: I would probably drop this sentence. We don't need to scare them.

docs/pages/repo/docs/core-concepts/caching/what-to-cache.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+### Inclusions
+
+In the above task definition for `build` it should be noted that it applies to _all_ of the `build` tasks in the entire Turborepo. It's quite likely that each of those individual tasks have slight differences in where their outputs are located, especially as the repository grows.
```
gsoltis, line 31: This is a good discussion, but I think is better located in the workspace configurations docs. The link to get there is 3ish paragraphs and one example in, and probably should be up front. Something like "<Some note that this applies to _all_ `build` tasks>, to handle different workspaces that produce task outputs in different locations, we recommend Workspace Configurations" then move on.

"How to specify outputs" and "how to customize individual package-task configurations" are different (but related) topics, and I think this page would benefit with focusing on the former with links to the latter. I'm on board with improving our documentation of "how to customize individual package-task configurations" too.

docs/pages/repo/docs/core-concepts/caching/file-inputs.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+By default, a workspace is considered to have been updated when any of the files in that workspace have changed. However, not every file change inside of the entire repo should result in rerunning a task. Turborepo gives you the tools to make sure that you can define every file that matters.
+
+## Global Files
+
+Almost every repository has some configuration that appears at the root of the repository. These files have the potential to impact the behavior of child workspaces.
+
+### `package.json` and `turbo.json`
+
+By default, Turborepo includes the root `package.json`, any lockfile that may be present, and the root `turbo.json` file into the global hash. You do not need to specify them separately.
```
gsoltis, line 18: We don't include the lockfile anymore. We include a hash of resolved external root deps, but lockfile changes should now only invalidate for workspaces that depend on whatever changed.

docs/pages/repo/docs/core-concepts/caching/file-inputs.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+In this example every single task in the turborepo would be rerun when the root `tsconfig.json` file changes.
+
+## Workspace Files
+
+In addition to global files, which appear outside of any workspace, you may also specify which files to include on a per-workspace basis.
+
+### `package.json` and `turbo.json`
+
+The contents of `package.json` and `turbo.json` are automatically considered for workspace hashes. This behavior is not configurable.
```
gsoltis, line 46: Can you add something to this sentence to indicate that these are workspace-level `package.json` and `turbo.json`? I know it's in the "Workspace Files" section, but it might help clarify, especially since some users will only have one `turbo.json`. Maybe "The contents of each workspace's `package.json` and `turbo.json`, if it exists, are automatically...." or something like that.

docs/pages/repo/docs/core-concepts/caching/file-inputs.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+### Turborepo's Default Behavior
+
+The default behavior of Turborepo is to inspect the repository and identify which files should be included in the hash calculation. Turborepo typically uses Git to identify which files to consider:
+
+- All of the files which appear in the Git index which are children of the workspace directory.
+- All of the untracked files which are children of the workspace directory and are not `.gitignore`d.
```
gsoltis, line 53: I think "git index" and references to tracking is a little too deep into the `git` weeds. Maybe we can say anything not ignored by `git`?

docs/pages/repo/docs/core-concepts/caching/file-inputs.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+If either Git or the Git index (typically in `.git/index`) is unavailable, we attempt to approximate the contents which would appear in the Git index using the following strategy:
```
gsoltis, line 55: I would drop the reference to `.git/index`. Something along the lines of "If this is not a git repository" or "If git is not available" I think conveys the same idea without inviting users to inspect their `.git` directory.

docs/pages/repo/docs/core-concepts/caching/file-inputs.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+<Callout type="info">
+  These two approaches do not always produce identical results! The scenarios where this does not produce identical results:
+
+- `.gitignore` files at other places in the directory hierarchy.
+- Files added to the Git index which are also specified inside of a `.gitignore`.
+
+You should aim to have both Git and the Git index of your repository available for maximum correctness.
+
+</Callout>
```
gsoltis, line 68: I would drop this. The fact that they produce different hashes is most likely a bug, and we should fix it.

Specifically the first case will be fixed by switching to the `ignore` crate in Rust, which takes into account the whole hierarchy of `.gitignore` files.

docs/pages/repo/docs/core-concepts/caching/environment-variable-inputs.mdx
```diff
@@ -0,0 +1,280 @@
...
+      "env": ["NEXT_PUBLIC_*", "!NEXT_PUBLIC_GIT_*"]
+    }
+  }
+}
+```
+
+Exclusion patterns apply only to the set of variables matched via inclusions.
+
+<Callout>
+  Be careful about the combination of wildcards and CI environments. Some CI environments set additional environment variables such as `NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA` which will prevent ever getting a cache hit. After configuring wildcards be sure to carefully review a run summary from all of your environments in order to make sure that it is only including variables you care about.
```
gsoltis, line 59: Link to Run Summary docs / flag

gsoltis (commented):
Lots of great upgrades here. I still haven't made it through everything, but I'm 85% of the way through. 

Nothing blocking so far, mostly just recommended edits for clarity / level of detail.

docs/pages/repo/docs/core-concepts/caching/file-inputs.mdx
nathanhammond: Because you can `git add` a `.gitignore`d file it is impossible for the two approaches to be identical. We can resolve the first issue, but not the second.

```
~/repos/git-add-gitignore$ git ls-tree HEAD .
100644 blob 8e8874f882daafd328a3d63da96c5112d62de9cc	.gitignore
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391	ignored-file.txt
```

```
~/repos/git-add-gitignore$ cat .gitignore 
ignored-file.txt
```

More generally, I believe that we should document the world as it is, not as we want it to be. This is how we function right now and people need to understand how this works.

docs/pages/repo/docs/core-concepts/caching/file-inputs.mdx
nathanhammond: There was an earlier iteration that specified that we grab relevant bits from the lockfile but I deleted it because it was way too much detail for a section on file inputs.

docs/pages/repo/docs/core-concepts/caching/file-inputs.mdx
nathanhammond: I intentionally made that distinction because the `.git/index` and `.gitignore` are functionally distinct. The index can contain arbitrary things that `.gitignore` would by default exclude.

In other words: I chose "accuracy" + "precision" here.

docs/pages/repo/docs/core-concepts/caching/environment-variable-inputs.mdx
```diff
@@ -0,0 +1,280 @@
...
+    "build": "source .env && next build"
+  }
+}
+```
+
+```sh filename=".env"
+export NEXT_PUBLIC_GA_ID=UA-00000000-0
+```
+
+`turbo`, having calculated a task hash prior to executing the `build` script, will not discover the `NEXT_PUBLIC_GA_ID` environment variable's value, and thus be unable to partition the cache based on its value. Be careful to ensure that all of your environment variables are loaded into the environment prior to invoking `turbo`!
```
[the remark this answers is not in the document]
nathanhammond: This corrects the overly-reduced example that previously existed (which accidentally reduced to "actually, this just works."). In this case I'm trying to demonstrate a particular failure mode while avoiding talking about other features.

I don't feel strongly here, This is _buried_ at the bottom, so if you're looking at it you've already passed the `dotEnv` section.

docs/pages/repo/docs/core-concepts/caching/what-to-cache.mdx
nathanhammond: We've had multiple issues recently that resolve to "I have a root `build` task, each of my workspaces had a different output folder, none of which I specify in `outputs`, why isn't it working?"

Conceptually:
- I want to try hoisting this up for a bit. We probably lucky if we get 5 minutes of docs-reading per-user. I want to try to get this above the line for that five minutes and see how that ends up trading off with user success rate.
- I'm generally a fan of repetitiveness in docs.

docs/pages/repo/docs/core-concepts/caching/what-to-cache.mdx
nathanhammond: Yeah, this paragraph needs some love. It's mostly 1:1 with the existing state.

docs/pages/repo/docs/core-concepts/caching/to-cache-or-not-to-cache.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+In building a monorepo you need to decide on a per-task basis if you are able to skip the task based upon having run it previously. This behavior is controlled via `pipeline.<task>.cache` inside of `turbo.json`:
```
mehulkar, line 10: Do we want this to be focused on monorepo? It's outside the monorepo section in core concepts, and caching is not a monorepo-specific thing

mehulkar, line 10: link to the `pipeline.<task>.cache` config docs here?

docs/pages/repo/docs/core-concepts/caching/to-cache-or-not-to-cache.mdx
```diff
@@ -0,0 +1,51 @@
...
+```
+
+In these examples we have told Turborepo to cache the `lint` tasks. But, since we've only specified the `cache` key, the only thing that is being cached is contents of `stdout`. For some types of tasks, such as `lint` or `test` this type of minimal configuration is likely correct!
+
+However, for most tasks you will also need to specify [_what_ to cache](./what-to-cache), as well as the [files](./file-configuration) and [environment variables](./environment-variables) that must match in order for the cache to be valid.
+
+## When Not To Cache
+
+<Callout type="info">
+  `"cache": false` does not mean "always run!" It simply means "if this task is to be executed it will not be restored from cache." Do not rely on `"cache": false` to trigger side effects like deployments.
```
mehulkar, line 41:
```suggestion
  `"cache": false` does not mean "always run!" It means "if this task is to be executed it will not be restored from cache." Do not rely on `"cache": false` to trigger side effects like deployments.
```

docs/pages/repo/docs/core-concepts/caching/to-cache-or-not-to-cache.mdx
```diff
@@ -0,0 +1,51 @@
...
+</Callout>
+
+Since caching is the default behavior, and is ideal for _most_ situations, knowing when to opt out of that behavior is important.
+
+1. **Tasks which execute extremely fast.** If you intend to use the remote cache and the task can be executed in less than the time of a network round trip, say, 100 milliseconds, you should consider not caching the task.
```
mehulkar, line 46: nit:
```suggestion
1. **Tasks that execute extremely fast.** If you intend to use the remote cache and the task can be executed in less than the time of a network round trip, say, 100 milliseconds, you should consider not caching the task.
```

docs/pages/repo/docs/core-concepts/caching/to-cache-or-not-to-cache.mdx
```diff
@@ -0,0 +1,51 @@
...
+2. **Tasks whose output assets are enormous.** If the consequence of running a task is the production of Docker Container it is possible that the time spent creating a cache artifact, uploading it, and downloading it, will exceed the amount of time to regenerate it.
```
mehulkar, line 47: tiny sidenote fyi, we got an ask to help cache docker images recently https://github.com/vercel/turbo/issues/4932

mehulkar (approved):

arlyon (approved):
Discussion is good. Approving since I have no criticisms that haven't been expressed.

PR #5095, opened 2023-05-25
allow to create internal modules via AssetContext

sokra:
### Description

This allows to create an Module with inner asset via AssetContext.

This will also respect transitions.

This change allows use to use `context.process(source, ReferenceType::Internal(inner_assets)` instead of `EcmascriptModuleAssetVc::new_with_inner_assets(..., ..., ..., ..., ..., inner_assets)`

next.js PR: https://github.com/vercel/next.js/pull/50338

### DIFF
--- a/crates/turbopack-core/src/reference_type.rs
+++ b/crates/turbopack-core/src/reference_type.rs
@@ -1,6 +1,11 @@
 use std::fmt::Display;
 
-use crate::resolve::ModulePartVc;
+use indexmap::IndexMap;
+
+use crate::{asset::AssetVc, resolve::ModulePartVc};
+
+#[turbo_tasks::value(transparent)]
+pub struct InnerAssets(IndexMap<String, AssetVc>);
 
 // These enums list well-known types, which we use internally. Plugins might add
 // custom types too.
@@ -72,6 +77,7 @@ pub enum ReferenceType {
     Url(UrlReferenceSubType),
     TypeScript(TypeScriptReferenceSubType),
     Entry(EntryReferenceSubType),
+    Internal(InnerAssetsVc),
     Custom(u8),
     Undefined,
 }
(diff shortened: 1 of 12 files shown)


alexkirsz:
Is there any reason for us to keep using `EcmascriptModuleAssetVc::new_with_inner_assets` directly, or should we make that method private and always use `context.process` instead?

sokra:
> Is there any reason for us to keep using `EcmascriptModuleAssetVc::new_with_inner_assets` directly, or should we make that method private and always use `context.process` instead?

We can't make them private as it's used from `turbopack` crate, where the `ModuleAssetContext` is implemented

alexkirsz (approved):

crates/turbopack-core/src/reference_type.rs
```diff
@@ -1,6 +1,19 @@
 use std::fmt::Display;
 
-use crate::resolve::ModulePartVc;
+use indexmap::IndexMap;
+
+use crate::{asset::AssetVc, resolve::ModulePartVc};
+
+#[turbo_tasks::value(transparent)]
+pub struct InnerAssets(IndexMap<String, AssetVc>);
```
alexkirsz, line 8: Add comments to these public structs, explaining why they exist.

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -133,6 +132,62 @@ struct MemoizedSuccessfulAnalysis {
     exports: EcmascriptExportsReadRef,
 }
 
+pub struct EcmascriptModuleAssetBuilder {
```
alexkirsz, line 135: Comment this public item and what each argument represents.

crates/turbopack-core/src/reference_type.rs
alexkirsz, line 1: `ReferenceType` and `*ReferenceSubType` could `#[derive(TaskInput)]`. We never use them as `Vc` anyway.

alexkirsz (approved):

Issue #4878, opened 2023-05-09
Add `schema.json` to the `turbo` package


scottwillmoore:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

I noticed that [Rome](https://rome.tools/) doesn't use a traditional hosted JSON schema and instead they publish their JSON schema in their package so that it can then be referenced with a local path:

```json
{
  "$schema": "./node_modules/rome/configuration_schema.json",
}
```

I see three advantages to this approach...

1. The JSON schema will match the installed version of the package.
2. The JSON schema will work without internet access, provided that `node_modules` has been installed.
3. Any changes to the JSON schema in the development can be used straight away in examples, etc.

### Describe the solution you'd like

Consider publishing the JSON schema for the `turbo.json` configuration with the `turbo` package. You could still leave the hosted JSON schema (which resolves to the latest version of `turbo as an option, but also provide a local JSON schema. I'm happy to work on a PR 😇.

### Describe alternatives you've considered

Continue to use the hosted JSON schema.

PR #5100, opened 2023-05-25
feat: add turbo.json schema to turbo package

chris-olszewski:
### Description

Implements #4878

This PR copies the schema.json from the docs and includes it in the `turbo` package.

I went with a naive copy to keep this simple and avoid any weird issues symlinks might introduce. Based on the fact that the schema doesn't change much I also instrumented tests against drift instead of automatically keeping them in sync.

### Testing Instructions

Verify that `schema.json` gets included in the tar we send to npm
```
$ cd packages/turbo && pnpm pack --pack-destination=/tmp
$ cd /tmp
$ tar xzf turbo-1.10.0-canary.0.tgz
$ ls packages
LICENSE                 bin                     node-platform.js        schema.json
README.md               install.js              package.json
```


packages/turbo/schema.json
```diff
@@ -0,0 +1 @@
...ssing a signature.","default":false}},"additionalProperties":false},"WorkspaceSchema":{"type":"object","properties":{"$schema":{"type":"string","default":"https://turbo.build/schema.json"},"pipeline":{"type":"object","additionalProperties":{"$ref":"#/definitions/Pipeline","description":"The name of a task that can be executed by turbo. If turbo finds a workspace package with a package.json scripts object with a matching key, it will apply the pipeline task configuration to that npm script during execution."},"description":"An object representing the task dependency graph of your project. turbo interprets these conventions to schedule, execute, and cache the outputs of tasks in your project.\n\nDocumentation: https://turbo.build/repo/docs/reference/configuration#pipeline","default":{}},"extends":{"type":"array","items":{"type":"string"},"description":"This key is only available in Workspace Configs and cannot be used in your root turbo.json.\n\nTells turbo to extend your root `turbo.json` and overrides with the keys provided in your Workspace Configs.\n\nCurrently, only the \"//\" value is allowed.","default":["//"]}},"required":["extends","pipeline"],"additionalProperties":false}}}
```
chris-olszewski, line 1: The lack of formatting was copied from `docs/public/schema.json` unsure if this is intentional or not.

mehulkar (approved):
Seems fine as a quick fix, but you could also set it up the way `docs/` does:

1. Add `"@turbo/types": "workspace:*"` as a dependency
2. [Add a `schema` task dependency on `build`](https://github.com/vercel/turbo/blob/4e035e61eae802cd472dda40779e141d2e896659/docs/turbo.json#L14)
3. [Add the `topo` relationship in turbo.json](https://github.com/vercel/turbo/blob/4e035e61eae802cd472dda40779e141d2e896659/docs/turbo.json#L23)
4. [Add the `schema` script in package.json that generates where you want it](https://github.com/vercel/turbo/blob/4e035e61eae802cd472dda40779e141d2e896659/docs/package.json#L12)

anthonyshew:
We've moved this file now but still think this would be useful for folks. Worth an update?

chris-olszewski:
Closing this to re-implement in a better way.

PR #5101, opened 2023-05-25
Store and retrieve timeSaved value from daemon

mehulkar:


mehulkar:
Currently stuck on:

```rs
error[E0596]: cannot borrow `__self.times_saved` as mutable, as it is behind a `&` reference
   --> crates/turborepo-lib/src/daemon/server.rs:261:9
    |
256 |           &self,
    |            ---- consider changing this binding's type to be: `&mut DaemonServer<T>`
...
261 | /         self.times_saved
262 | |             .insert(inner.hash.clone(), inner.time_saved);
    | |_________________________________________________________^ `__self` is a `&` reference, so the data it refers to cannot be borrowed as mutable

error[E0382]: borrow of moved value: `inner.hash`
   --> crates/turborepo-lib/src/daemon/server.rs:295:52
    |
289 |                 &Arc::new(inner.hash),
    |                           ---------- value moved here
...
295 |         if let Some(value) = self.times_saved.get(&inner.hash.clone()) {
    |                                                    ^^^^^^^^^^^^^^^^^^ value borrowed here after move
```

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -58,6 +58,8 @@ pub struct DaemonServer<T: Watcher> {
     shutdown_rx: Option<Receiver<()>>,
 
     running: Arc<AtomicBool>,
+
+    times_saved: HashMap<String, u64>,
```
chris-olszewski, line 62: We want to wrap the times saved map in `Arc<Mutex<T>>` to allow for us to mutate this from different requests. 
Breaking it down:
 - `Arc` is "atomically reference counted" extremely similar to `Rc` which stands for reference counted which changes a value from being fully dropped whenever it goes out of scope to just decrementing the count and having the value get dropped once it hits 0. [`Arc` is thread safe unlike `Rc`](https://doc.rust-lang.org/std/sync/struct.Arc.html#thread-safety)
 - Mutex: just a normal mutex
```suggestion
    times_saved: Arc<std::sync::Mutex<HashMap<String, u64>>>,
```

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -98,6 +100,7 @@ impl DaemonServer<notify::RecommendedWatcher> {
             shutdown_rx: Some(recv_shutdown),
 
             running: Arc::new(AtomicBool::new(true)),
+            times_saved: HashMap::new(),
```
chris-olszewski, line 103:
```suggestion
            times_saved: Arc::new(std::sync::Mutex(HashMap::new())),
```

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -255,6 +258,9 @@ impl<T: Watcher + Send + 'static> proto::turbod_server::Turbod for DaemonServer<
     ) -> Result<tonic::Response<proto::NotifyOutputsWrittenResponse>, tonic::Status> {
         let inner = request.into_inner();
 
+        self.times_saved
+            .insert(inner.hash.clone(), inner.time_saved);
+
```
chris-olszewski, line 263: Now the we have this behind a mutex we need to lock it before we can access the data. We introduce a block so that `times_saved` drops as soon as possible. `times_saved` isn't actually the `HashMap` type, but a `MutexGuard` which will release the lock as soon as it gets dropped freeing up the resource for others.
One question I have about this code is what happens if we already have a time saved for an output hash? Do we always want to overwrite the old one?
```suggestion
        {
            let mut times_saved = self.times_saved.lock().expect("times saved lock poisoned");
            times_saved.insert(inner.hash.clone(), inner.time_saved);
        }
```

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -285,9 +291,15 @@ impl<T: Watcher + Send + 'static> proto::turbod_server::Turbod for DaemonServer<
             )
             .await;
 
+        let mut timeSaved = 0;
+        if let Some(value) = self.times_saved.get(&inner.hash.clone()) {
+            timeSaved = *value;
+        }
```
chris-olszewski, line 297: I can't add a code suggestion, but above `let changed = self` we should add a
`let hash = Arc::new(inner.hash);` and change the first argument from `&Arc::new(inner.hash);` to `&hash`.

Breaking down the other changes:
 - `copied`: This is a helper method that changes an `Option<&T>` to an `Option<T>` if `T` implements `Copy`
 - `unwrap_or_default`: converts an `Option<T>` to `T`, if the value is present it will be used, but if `None` than we use `T`'s default

Questions I have here: do we want to paper over missing a time saved? 
```suggestion
        let timeSaved = {
            let times_saved = self.times_saved.lock().expect("times saved lock poisoned");
            times_saved.get(hash.as_str()).copied().unwrap_or_default()
        };
```

crates/turborepo-lib/src/daemon/server.rs
mehulkar: yes we'd always want to overwrite since every time the cache is written and `notifyOutputsWritten` is called the duration is saved in the cache metadata also https://github.com/vercel/turbo/blob/d51ad5a9c775d3d0721e3a560b81f3829e447be1/cli/internal/runcache/runcache.go#L279-L282

mehulkar:
Thanks for the suggestions and explanations @chris-olszewski , that helped a lot! Merging this into #4952 to get a final review, it looks like its working well and I can pass values back and forth from the daemon now

PR #5107, opened 2023-05-26
ci(workflow): update turbopack test results to datadog

kwonoj:
### Description

This PR makes turbopack's test result to be uploaded to datadog for the observability. Datadog have a dedicated test / pipeline features for the CI, we'll going to utilize it.

Once uploaded, individual CI results can be displayed like

image

image

and we'll setup a dashboard to refine data what we want to see.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,17 @@
+[profile.tp-test-linux.junit]
+path = "junit.xml"
+report-name = "Turbopack tests (Linux)"
+store-success-output = true
+store-failure-output = true
+
+[profile.tp-test-mac.junit]
+path = "junit.xml"
+report-name = "Turbopack tests (Mac)"
+store-success-output = true
+store-failure-output = true
+
+[profile.tp-test-win.junit]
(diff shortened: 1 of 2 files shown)


kwonoj:
Note this won't work currently until we get the proper api key, however CI shouldn't fail.

.config/nextest.toml
```diff
@@ -0,0 +1,17 @@
+[profile.tp-test-linux.junit]
```
kwonoj, line 1: By using profile local test run won't create redundant reports.

.github/workflows/test.yml
```diff
@@ -1320,6 +1341,44 @@ jobs:
         if: steps.info.outputs.success == 'true'
         run: echo Ok
 
+  # Upload Turbopack's test results into datadog.
+  upload_test_results:
+    name: Upload Test results
+    needs:
+      - turbopack_rust_test1
```
kwonoj, line 1348: As a starting point, only uploads test results. We may expand this to the benchmark results.

sokra (approved):

PR #5119, opened 2023-05-29
Noting multiple global turbo installations.

anthonyshew:
### Description

This docs update notes how it's possible to install `turbo` in multiple places on your machine - and how to fix it.

### DIFF
--- a/docs/pages/repo/docs/faq.mdx
+++ b/docs/pages/repo/docs/faq.mdx
@@ -4,6 +4,7 @@ description: Frequently asked questions about Turborepo.
 ---
 
 import Callout from '../../../components/Callout'
+import { Tabs, Tab } from '../../../components/Tabs'
 
 # Frequently Asked Questions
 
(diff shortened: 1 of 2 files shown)


docs/pages/repo/docs/faq.mdx
```diff
@@ -16,6 +17,35 @@ You have two options when working with Turborepo:
 
 We recommend installing the `turbo` CLI globally. This gives you a smooth, ergonomic experience for running tasks.
 
+### Why isn't my global `turbo` working as expected?
```
tknickman, line 20: Another issue I've seen here is caused by the order that the global bin directory is referenced in your PATH. For example if yarn is included before pnpm, and you update turbo via pnpm, your version isn't going to change because turbo is going to resolve to the yarn version first.

tknickman (approved):

docs/pages/repo/docs/faq.mdx
arlyon: This is a dev-only issue but I often end up with turbo in my `.cargo/bin` as well...

arlyon (approved):

PR #5125, opened 2023-05-30
fix(blog): correct docs link copy

tknickman:
### Description

We love learning more.


chris-olszewski (approved):
CLI so good you need to learn it twice

PR #5127, opened 2023-05-30
Update log message when cache restoration is skipped

mehulkar:


anthonyshew (approved):

cli/internal/runcache/runcache.go
```diff
@@ -158,18 +158,24 @@ func (tc *TaskCache) RestoreOutputs(ctx context.Context, prefixedUI *cli.Prefixe
 			Source:    cache.CacheSourceFS,
 			TimeSaved: timeSavedFromDaemon,
 		}
-		prefixedUI.Warn(fmt.Sprintf("Skipping cache check for %v, outputs have not changed since previous run.", tc.pt.TaskID))
+	}
+
+	// Some more context to add into the cache hit messages.
+	// This isn't the cleanest way to updat ethe log message, so we should revisit during Rust port.
```
tknickman, line 164:
```suggestion
	// This isn't the cleanest way to update the log message, so we should revisit during Rust port.
```

tknickman (approved):

PR #5135, opened 2023-05-31
Merge timesaved and cache status data structures

mehulkar:
- Instead of Local/Remote booleans, use a Source field.
This ensure that we are more truthful about where the cache
was restored from, rather than which cache happened to win
the race condition.
- Adds timeSaved value from the same cache source where restoration
happened from.

### DIFF
--- a/cli/internal/cache/cache.go
+++ b/cli/internal/cache/cache.go
@@ -20,7 +20,7 @@ import (
 type Cache interface {
 	// Fetch returns true if there is a cache it. It is expected to move files
 	// into their correct position as a side effect
-	Fetch(anchor turbopath.AbsoluteSystemPath, hash string, files []string) (ItemStatus, []turbopath.AnchoredSystemPath, int, error)
+	Fetch(anchor turbopath.AbsoluteSystemPath, hash string, files []string) (ItemStatus, []turbopath.AnchoredSystemPath, error)
 	Exists(hash string) ItemStatus
 	// Put caches files for a given hash
 	Put(anchor turbopath.AbsoluteSystemPath, hash string, duration int, files []turbopath.AnchoredSystemPath) error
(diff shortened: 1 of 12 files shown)


cli/internal/cache/cache.go
```diff
@@ -271,31 +293,31 @@ func (mplex *cacheMultiplexer) Fetch(anchor turbopath.AbsoluteSystemPath, key st
...
 			// result is a success at fetching. Storing in lower-priority caches is an optimization.
-			_ = mplex.storeUntil(anchor, key, duration, actualFiles, i)
+			_ = mplex.storeUntil(anchor, key, itemStatus.TimeSaved, actualFiles, i)
 
-			// If another cache had already set this to true, we don't need to set it again from this cache
-			combinedCacheState.Local = combinedCacheState.Local || itemStatus.Local
-			combinedCacheState.Remote = combinedCacheState.Remote || itemStatus.Remote
-			return combinedCacheState, actualFiles, duration, err
+			// Return this cache, and exit the for loop, since we don't need to keep looking.
+			return itemStatus, actualFiles, nil
```
mehulkar, line 304: This is an important change. Instead of combining results from multiple caches, we're returning the first cache hit. One benefit of this is that it deterministically says which cache the `actualFiles` and `timeSaved` came from. By combining, we lose that information, which can make debugging harder.

cli/internal/cache/cache.go
```diff
@@ -271,31 +293,31 @@ func (mplex *cacheMultiplexer) Fetch(anchor turbopath.AbsoluteSystemPath, key st
...
+// Exists check each cache sequentially and return the first one that has a cache hit
 func (mplex *cacheMultiplexer) Exists(target string) ItemStatus {
-	syncCacheState := ItemStatus{}
 	for _, cache := range mplex.caches {
 		itemStatus := cache.Exists(target)
-		syncCacheState.Local = syncCacheState.Local || itemStatus.Local
-		syncCacheState.Remote = syncCacheState.Remote || itemStatus.Remote
+		if itemStatus.Hit {
+			return itemStatus
+		}
```
mehulkar, line 317: This is also an important change. Remove the combining for the same reason noted in https://github.com/vercel/turbo/pull/5135/files#r1213564982

NicholasLYang (approved):

PR #5152, opened 2023-05-31
feat(daemon): clean

tknickman:
### Description

Adds clean command to turbo daemon.

### DIFF
--- a/crates/turborepo-lib/src/commands/daemon.rs
+++ b/crates/turborepo-lib/src/commands/daemon.rs
@@ -67,24 +67,24 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
         }
         DaemonCommand::Clean => {
             // try to connect and shutdown the daemon
-            let client = connector.clone().connect().await;
+            let client = connector.connect().await;
             match client {
                 Ok(client) => match client.stop().await {
                     Ok(_) => {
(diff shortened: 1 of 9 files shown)


crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -60,6 +65,51 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
...
+                if let Err(e) = result {
+                    println!("Failed to remove socket file: {}", e);
+                    println!("Please remove manually: {}", sock_file);
+                    success = false;
+                }
+            }
+
+            if success {
+                println!("Success!");
+                println!("Restart the daemon with `turbo daemon start`");
```
chris-olszewski, line 110: Do we want to advise the user manually start the daemon instead of just having it get started as part of run?

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -51,7 +51,7 @@ pub enum ForkError {
     Spawn(#[from] std::io::Error),
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
chris-olszewski, line 54: Are we using the new clone functionality anywhere? I don't have an issue with the addition of the trait, just want to make sure I'm not missing anything.

chris-olszewski (approved):

PR #5153, opened 2023-05-31
chore(turborepo): Port piece of kill_live_server test for version mismatch

gsoltis:
### Description

While not the greatest test since it isn't testing actual server logic, it does assert that a client receiving a `failed_precondition` should translate that as a `VersionMismatch`. This matches the behavior in `connector_test.go`.

### Testing Instructions

New test functional added


gsoltis:
Follow on from #5151

tknickman (approved):

PR #5155, opened 2023-05-31
feat(create-turbo): use latest turbo by default

tknickman:
### Description

Use latest turbo by default (unless using create-turbo@canary). 

Also support new option `--turbo-version <version>`

### DIFF
--- a/packages/create-turbo/src/cli.ts
+++ b/packages/create-turbo/src/cli.ts
@@ -28,6 +28,10 @@ createTurboCli
     false
   )
   .option(
+    "--turbo-version <version>",
+    "Use a specific version of turbo (default: latest)"
+  )
+  .option(
     "-e, --example [name]|[github-url]",
     `
   An example to bootstrap the app with. You can use an example name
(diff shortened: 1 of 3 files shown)


packages/create-turbo/src/cli.ts
```diff
@@ -44,8 +48,8 @@ createTurboCli
   --example-path foo/bar
 `
   )
-  .version(cliPkg.version, "-v, --version", "output the current version")
-  .helpOption()
+  .version(cliPkg.version, "-v, --version", "Output the current version")
+  .helpOption("-h, --help", "Display help for command")
```
tknickman, line 52: This just fixes a minor casing inconsistency in the help message that has bothered me for a while

packages/create-turbo/src/commands/create/types.ts
```diff
@@ -1,8 +1,9 @@
-export type CreateCommandArgument = "string" | undefined;
+export type CreateCommandArgument = string | undefined;
```
tknickman, line 1: :bruh:

chris-olszewski (approved):

PR #5156, opened 2023-05-31
Support installing tip-of-canary next for benchmarks

jridgewell:
### Description

A small change to support installing a local `next@file:path/to/next.js/packages/next` next package. This fixes an issue where we run tip-of-canary `next-dev` vs the last published `next` package, and allows us to benchmark breaking changes.

The [`--install-links`](https://docs.npmjs.com/cli/v8/commands/npm-install#install-links) option is a little oxymoronic.  When set to true, it **doesn't** use symlinks, so we need to set it to false to avoid copying a bunch of files.

### DIFF
--- a/crates/turbopack-bench/src/util/npm.rs
+++ b/crates/turbopack-bench/src/util/npm.rs
@@ -42,7 +42,12 @@ pub fn install(install_dir: &Path, packages: &[NpmPackage<'_>]) -> Result<()> {
             .write_all(format!("{:#}", package_json).as_bytes())?;
     }
 
-    let mut args = vec!["install".to_owned(), "--force".to_owned()];
+    let mut args = vec![
+        "install".to_owned(),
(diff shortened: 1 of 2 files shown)


crates/turbopack-bench/src/util/npm.rs
```diff
@@ -42,7 +42,12 @@ pub fn install(install_dir: &Path, packages: &[NpmPackage<'_>]) -> Result<()> {
             .write_all(format!("{:#}", package_json).as_bytes())?;
     }
 
-    let mut args = vec!["install".to_owned(), "--force".to_owned()];
+    let mut args = vec![
+        "install".to_owned(),
+        "--force".to_owned(),
+        "--install-links".to_owned(),
```
kwonoj, line 48: Guess worth to leave comments as pr bod describes?

kwonoj (approved):

PR #5161, opened 2023-06-01
make HMR re-execute modules that self invalidate

ForsakenHarmony:
### Description
This can happen when a module that previously exported a React component no longer does, the next react refresh integration will then call `module.hot.invalidate()`

fixes the `acceptance/ReactRefreshRequire.test.ts` next.js integration test


crates/turbopack-ecmascript-runtime/js/src/dev/runtime/base/runtime-base.ts
```diff
@@ -15,7 +15,7 @@
 
 // This file must not use `import` and `export` statements. Otherwise, it
 // becomes impossible to augment interfaces declared in `<reference>`d files
-// (e.g. `Module`). Hence the need for `import()` here.
+// (e.g. `Module`). Hence, the need for `import()` here.
```
ForsakenHarmony, line 18: my IDE made a bunch of grammar suggestions 🙃

jridgewell (approved):

crates/turbopack-ecmascript-runtime/js/src/dev/runtime/base/runtime-base.ts
alexkirsz: TIL!

crates/turbopack-ecmascript-runtime/js/src/dev/runtime/base/runtime-base.ts
```diff
@@ -1116,8 +1155,8 @@ function getFirstModuleChunk(moduleId: ModuleId) {
 }
 
 /**
- * Removes a module from a chunk. Returns true there are no remaining chunks
- * including this module.
+ * Removes a module from a chunk.
+ * Returns `true` if there are no remaining chunks, including this module.
```
alexkirsz, line 1159: This no longer means the same thing. It should be:

```
Returns `true` if there are no remaining chunks including this module.
```

ForsakenHarmony: you are right, will fix in my next PR

PR #5171, opened 2023-06-01
Add timeSaved value into dry run output

mehulkar:


cli/internal/cache/cache_http.go
```diff
@@ -159,15 +160,9 @@ func (cache *httpCache) retrieve(hash string) (bool, []turbopath.AnchoredSystemP
 		b, _ := ioutil.ReadAll(resp.Body)
 		return false, nil, 0, fmt.Errorf("%s", string(b))
 	}
-	// If present, extract the duration from the response.
-	duration := 0
-	if resp.Header.Get("x-artifact-duration") != "" {
-		intVar, err := strconv.Atoi(resp.Header.Get("x-artifact-duration"))
-		if err != nil {
-			return false, nil, 0, fmt.Errorf("invalid x-artifact-duration header: %w", err)
```
mehulkar, line 167: We lose this error condition here, but it doesn't seem worth failing an entire cache fetch for, so I didn't preserve it.

tknickman: Can we at least debug log it so it doesn't get fully burried?

tknickman (approved):

PR #5172, opened 2023-06-01
docs: add separate pages for each CLI command

mehulkar:


tknickman (changes requested):
Can you search for links to these and make sure they're updated as well? There are links to commands _all_ over the docs so it might be a little annoying...

mehulkar:
@tknickman good call, I think I got them all (searched VS Code for `/command-line-reference` and updated 45 matches (some of the matches were the ones introduced in this PR, so not quite 45)

tknickman (approved):
Nice, 🚢

PR #5181, opened 2023-06-02
fix(turborepo): Restructure reading from stderr, fix parsing of ls-tree

gsoltis:
### Description

 - Ensure we read `stderr` after calling `wait()` on the child process
 - Fix parsing of `git ls-tree` to not truncate filenames at spaces

### Testing Instructions

Added a test of a filename with spaces

### DIFF
--- a/crates/turborepo-scm/src/lib.rs
+++ b/crates/turborepo-scm/src/lib.rs
@@ -69,7 +69,7 @@ pub(crate) fn wait_for_success<R: Read, T>(
     let stderr_output = read_git_error_to_string(stderr);
     let stderr_text = stderr_output
         .map(|stderr| format!(" stderr: {}", stderr))
-        .unwrap_or("".to_string());
+        .unwrap_or_default();
     let exit_text = if exit_status.success() {
         "".to_string()
     } else {
--- a/crates/turborepo-scm/src/lib.rs
+++ b/crates/turborepo-scm/src/lib.rs
@@ -61,10 +61,8 @@ pub(crate) fn wait_for_success<R: Read, T>(
     parse_result: Result<T, Error>,
 ) -> Result<T, Error> {
     let exit_status = child.wait()?;
-    if exit_status.success() {
-        if let Ok(result) = parse_result {
-            return Ok(result);
-        }
+    if exit_status.success() && parse_result.is_ok() {
+        return parse_result;
     }
     let stderr_output = read_git_error_to_string(stderr);
     let stderr_text = stderr_output
--- a/crates/turborepo-scm/src/lib.rs
+++ b/crates/turborepo-scm/src/lib.rs
@@ -40,43 +40,54 @@ impl Error {
     pub(crate) fn git_error(s: impl Into<String>) -> Self {
         Error::Git(s.into(), Backtrace::capture())
     }
-
-    fn from_git_exit_code(cmd: &str, path: &AbsoluteSystemPath, exit_code: Option<i32>) -> Error {
-        let s = format!(
-            "'{}' in {} exited with status code {}",
-            cmd,
-            path.to_string(),
-            exit_code
-                .map(|code| code.to_string())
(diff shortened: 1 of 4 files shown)


crates/turborepo-scm/src/lib.rs
```diff
@@ -40,43 +40,54 @@ impl Error {
...
-        Ok(())
+    if exit_status.success() {
+        if let Ok(result) = parse_result {
+            return Ok(result);
+        }
     }
+    let stderr_output = read_git_error_to_string(stderr);
+    let stderr_text = stderr_output
+        .map(|stderr| format!(" stderr: {}", stderr))
+        .unwrap_or("".to_string());
```
NicholasLYang, line 72:
```suggestion
        .unwrap_or_default();
```

crates/turborepo-scm/src/lib.rs
```diff
@@ -40,43 +40,54 @@ impl Error {
...
+            .unwrap_or("unknown".to_string());
+        format!(" exited with code {}", code)
+    };
+    let parse_error_text = if let Err(parse_error) = parse_result {
+        format!(" had a parse error {}", parse_error)
+    } else {
+        "".to_string()
+    };
+    let path_text = root_path.as_ref();
+    let err_text = format!(
```
NicholasLYang, line 88: Should we do this formatting in the error type itself? Like have the constituent pieces like `command`, `path_text`, etc. be fields in `Error::Git` and let `thiserror` actually format it?

crates/turborepo-scm/src/lib.rs
```diff
@@ -40,43 +40,54 @@ impl Error {
...
-    if !exit_status.success() {
-        Err(read_git_error(stderr).unwrap_or_else(|| {
-            Error::from_git_exit_code(command, root_path.as_ref(), exit_status.code())
-        }))
-    } else {
-        Ok(())
+    if exit_status.success() {
+        if let Ok(result) = parse_result {
+            return Ok(result);
+        }
```
chris-olszewski, line 67: Removes some indentation
```suggestion
    if exit_status.success() && parse_result.is_ok() {
        return parse_result;
```

chris-olszewski (approved):
LGTM

crates/turborepo-scm/src/lib.rs
```diff
@@ -40,43 +40,54 @@ impl Error {
...
 }
 
-pub(crate) fn wait_for_success<R: Read>(
+pub(crate) fn wait_for_success<R: Read, T>(
     mut child: Child,
     stderr: &mut R,
     command: &str,
     root_path: impl AsRef<AbsoluteSystemPath>,
-) -> Result<(), Error> {
+    parse_result: Result<T, Error>,
```
NicholasLYang, line 61: I'm not in love with this API, cause it feels like this function is doing two things now: waiting for success and handling parse errors. Could we factor out the stderr reading and git error construction into a different function, then have the parse error case and the process error case both call it?

NicholasLYang (changes requested):

crates/turborepo-scm/src/lib.rs
gsoltis: We use this error type in a few places with different arbitrary strings (ex: failing to open `stdout` for a child process). We may at some point wish to get more specific, but I don't think we're there yet. More `git` usage is coming, which may have its own new failure modes.

crates/turborepo-scm/src/lib.rs
gsoltis: That was the one of the problems this PR is attempting to solve: trying to read `stderr` too early due to a parse error, via the factored out `read_git_error` caused a deadlock because we had not yet called `.wait()`. So now they are intentionally coupled. A variety of things can go wrong, `wait_for_success` tries to provide as much information in an error as is available.

NicholasLYang (approved):

PR #5184, opened 2023-06-02
Add codemod for glob syntax issues.

arlyon:
### Description

We discovered a few classes of glob that are soon-to-be invalid with the tightening on the syntax from `wax`. This codemod handles the most common cases. Out of 720 unique globs encountered, only about 10 of them were invalid, and of those there were 3 general classes of syntax problem which are being addressed here, along with 3 just completely invalid ones. This codemod should as a result fix around 80% of the syntax issues.

### Testing Instructions

```
cd packages/turbo-codemod
pnpm test clean-globs
```


arlyon:
Adding @tknickman as the codemod boss. As an aside, I couldn't seem to get plop working with our config. It doesn't seem to like the config file being ts... am I doing something wrong or do we need to fix that?

packages/turbo-codemod/src/transforms/clean-globs.ts
```diff
@@ -0,0 +1,106 @@
...
+import path from "path";
+import fs from "fs-extra";
+import getTransformerHelpers from "../utils/getTransformerHelpers";
+import { getTurboConfigs } from "@turbo/utils";
+import Logger from "../utils/logger";
+
+// transformer details
+const TRANSFORMER = "clean-globs";
+const DESCRIPTION =
+  "Automatically clean up invalid globs from your &#x60;turbo.json&#x60; file";
```
chris-olszewski, line 13: Is there a reason for using an accent over just a `'`? If we continue going with the accent I'd much rather have it in it's own constant and use it in the construction of `DESCRIPTION`

packages/turbo-codemod/__tests__/clean-globs.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      transformer: "test",
+      rootPath: ".",
+      options: { force: false, dry: false, print: false },
+    });
+
+    // Now let's test the function
+    badGlobPatterns.forEach(([input, output]) => {
+      expect(fixGlobPattern(input, log)).toBe(output);
+    });
+  });
```
chris-olszewski, line 132: Just because all of these transforms could happen in the same glob, I'd like to see some tests for truly horrible globs that hit all of the conditions e.g. `!foo**/**/**bar`

packages/turbo-codemod/src/transforms/clean-globs.ts
```diff
@@ -0,0 +1,106 @@
...
+  let newPattern = pattern.replace(/\*\*\/\*\*/g, "**");
+  while (newPattern !== pattern) {
+    log.modified(`${pattern} to ${newPattern}`);
+    pattern = newPattern;
+    newPattern = pattern.replace(/\*\*\/\*\*/g, "**");
+  }
+
+  // For '!**/dist' and '!**/node_modules'
+  // Change '!**/dist' and '!**/node_modules' to '**/[!dist]' and '**/[!node_modules]'
+  newPattern = pattern.replace(/^!\*\*\/([a-z_\/]+)$/g, "**/[!$1]");
```
chris-olszewski, line 70: Shouldn't the charset should include more than just the lowercase letters and `_/`?

packages/turbo-codemod/src/transforms/clean-globs.ts
```diff
@@ -0,0 +1,106 @@
...
+  if (newPattern !== pattern) {
+    log.modified(`${pattern} to ${newPattern}`);
+    log.info("please make the previous transform is correct");
```
chris-olszewski, line 73: Is there any reason to why this is the only change we ask the user to verify?
```suggestion
    log.info("please make sure the previous transform is correct");
```

packages/turbo-codemod/src/transforms/clean-globs.ts
```diff
@@ -0,0 +1,106 @@
...
+    pattern = newPattern;
+  }
+
+  // For 'cypress/integration/**.test.ts', 'scripts/**.mjs', 'scripts/**.js', 'src/types/generated/**.ts'
+  // Change '**.ext' to '**/*.ext' where 'ext' is 'test.ts', 'mjs', 'js', 'ts'
+  newPattern = pattern.replace(/(\*\*)([a-z\.]+)/g, "$1/*$2");
```
chris-olszewski, line 79: Again this charset seems really restrictive

chris-olszewski (changes requested):
Main issue is with the charset in the transform patterns being too restrictive, we won't catch stuff like `**-foo`. Would also like to see just a few more tests where we apply multiple transforms to the same glob.

packages/turbo-codemod/src/transforms/clean-globs.ts
arlyon: This was the default output from `plop`. Maybe you can weigh in @tknickman ?

packages/turbo-codemod/src/transforms/clean-globs.ts
arlyon: After some consideration, this is a stupid correction. `wax` of course rejects this, but we pre-process the globs and pass the negations into our exclude list. I will remove this case altogether 🥴

packages/turbo-codemod/__tests__/clean-globs.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    expect(fixGlobPattern("a/b/c/*", log)).toBe("a/b/c/*");
+  });
+
+  test("transforms '!**/folder' to '**/[!folder]'", () => {
+    let badGlobPatterns = [
+      ["!**/dist", "**/[!dist]"],
+      ["!**/node_modules", "**/[!node_modules]"],
+      ["!**/foo/bar", "**/[!foo/bar]"],
+      ["!**/foo/bar/baz", "**/[!foo/bar/baz]"],
+      ["!**/foo/bar/baz/qux", "**/[!foo/bar/baz/qux]"],
```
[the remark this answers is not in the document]
arlyon: I did not find one matching that in my exploration, but I am hoping to get a better answer when github codesearch's api is released.

arlyon:
All feedback addressed :)

packages/turbo-codemod/src/transforms/clean-globs.ts
tknickman: Hmmm it's probably an issue with how I wrote the template. I might need an escaping helper when writing the output. I would just use the `'`

packages/turbo-codemod/src/transforms/clean-globs.ts
```diff
@@ -0,0 +1,98 @@
...
+  return config;
+}
+
+export function fixGlobPattern(pattern: string, log: Logger): string {
+  let oldPattern = pattern;
+  // For '../../app-store/**/**' and '**/**/result.json'
+  // Collapse back-to-back doublestars '**/**' to a single doublestar '**'
+  let newPattern = pattern.replace(/\*\*\/\*\*/g, "**");
+  while (newPattern !== pattern) {
+    log.modified(`${pattern} to ${newPattern}`);
```
tknickman, line 63: I would remove these manual logs, the transform runner will handle all of the logging for you!

packages/turbo-codemod/src/transforms/clean-globs.ts
```diff
@@ -0,0 +1,98 @@
...
+    pattern = newPattern;
+    newPattern = pattern.replace(/\*\*\/\*\*/g, "**");
+  }
+
+  // For '**.ext' change to '**/*.ext'
+  // 'ext' is a filename or extension and can contain almost any character except '*' and '/'
+  newPattern = pattern.replace(/(\*\*)([^*/]+)/g, "$1/*$2");
+  if (newPattern !== pattern) {
+    log.modified(`${pattern} to ${newPattern}`);
```
tknickman, line 72: same

packages/turbo-codemod/src/transforms/clean-globs.ts
```diff
@@ -0,0 +1,98 @@
...
+    pattern = newPattern;
+  }
+
+  // For 'prefix**' change to 'prefix*/**'
+  // 'prefix' is a folder name and can contain almost any character except '*' and '/'
+  newPattern = pattern.replace(/([^*/]+)(\*\*)/g, "$1*/$2");
+  if (newPattern !== pattern) {
+    log.modified(`${pattern} to ${newPattern}`);
```
tknickman, line 80: remove

packages/turbo-codemod/src/transforms/clean-globs.ts
```diff
@@ -0,0 +1,98 @@
...
+    pattern = newPattern;
+  }
+
+  if (oldPattern === pattern) {
+    log.unchanged(pattern);
```
tknickman, line 85: can remove

chris-olszewski (approved):

PR #5189, opened 2023-06-02
Error handling fixes for `turbopack-node` and the HMR runtime

ForsakenHarmony:
### Description
- errors during module init are no longer ignored
- non-Errors thrown inside node.js now display correctly

Bonus:
- you can now debug the spawned node.js processes without recompiling via `TURBOPACK_DEBUG_JS=operation`


crates/turbopack-node/src/debug.rs
```diff
@@ -0,0 +1,11 @@
+use std::env;
+
+const DEBUG_JS_VAR: &str = "TURBOPACK_DEBUG_JS";
+
+pub fn should_debug(operation: &str) -> bool {
+    let Ok(val) = env::var(DEBUG_JS_VAR) else {
+		return false;
+	};
```
jridgewell, line 8:
```suggestion
          return false;
    };
```

jridgewell (approved):

crates/turbopack-node/src/debug.rs
ForsakenHarmony: right, `cargo fmt` doesn't work on `let else` yet

crates/turbopack-ecmascript-runtime/js/src/dev/runtime/base/runtime-base.ts
```diff
@@ -823,12 +830,22 @@ function applyInternal(
     disposedModules
   );
 
+  let error: any;
+  var reportError = function (err: any) {
```
alexkirsz, line 834: Nit: (maybe we should start linting these files on top of TS?)

```suggestion
  const reportError = (err: any) => {
```

crates/turbopack-node/js/src/ipc/error.ts
```diff
@@ -0,0 +1,58 @@
...
+    return err
+  }
+
+  if (process.env.NODE_ENV === 'development') {
+    // provide better error for case where `throw undefined`
+    // is called in development
+    if (typeof err === 'undefined') {
+      return new Error(
+        'An undefined error was thrown, ' +
+        'see here for more info: https://nextjs.org/docs/messages/threw-undefined'
```
alexkirsz, line 22: Should these link to turbopack docs instead?

crates/turbopack-node/js/src/ipc/index.ts
```diff
@@ -133,9 +136,11 @@ function createIpc<TIncoming, TOutgoing>(
           ...structuredError(error),
         });
       } catch (err) {
+        console.error("failed to send error back to rust:", err);
```
alexkirsz, line 139: This should only be logged in some kind of verbose/tracing mode, as I don't think it's very actionable from a user's perspective.

crates/turbopack-node/src/debug.rs
```diff
@@ -0,0 +1,11 @@
+use std::env;
+
+const DEBUG_JS_VAR: &str = "TURBOPACK_DEBUG_JS";
+
+pub fn should_debug(operation: &str) -> bool {
```
alexkirsz, line 5: Comment this public item.

crates/turbopack-node/src/transforms/postcss.rs
```diff
@@ -233,7 +234,7 @@ impl PostCssTransformedAssetVc {
                 JsonValueVc::cell(css_path.into()),
             ],
             extra_configs_changed,
-            /* debug */ false,
+            should_debug("postcss_transform"),
```
alexkirsz, line 237: Could this be a trace level instead?

crates/turbopack-node/src/render/node_api_source.rs
```diff
@@ -64,6 +66,7 @@ pub struct NodeApiContentSource {
     route_match: RouteMatcherVc,
     entry: NodeEntryVc,
     render_data: JsonValueVc,
+    debug: bool,
```
alexkirsz, line 69: Could this be integrated to tracing instead?

crates/turbopack-node/js/src/ipc/error.ts
ForsakenHarmony: hmm, ideally yes, maybe I'll just remove the doc link

crates/turbopack-node/src/render/node_api_source.rs
ForsakenHarmony: yeah it's for starting node.js in debug mode

crates/turbopack-node/js/src/ipc/index.ts
ForsakenHarmony: I added this because previously the error outputted was something like "unexpected EOF" so while it might not be actionable I thought it might be easier to understand and if we receive feedback easier to debug, can remove it again though

crates/turbopack-ecmascript-runtime/js/src/dev/runtime/base/runtime-base.ts
ForsakenHarmony: They should actually be linted already, not sure why eslint is not complaining

ForsakenHarmony: nvm it was not linted, we didn't have `no-var` (which is broken because it includes `declare var`)

Issue #5117, opened 2023-05-29
[Turborepo] Add supporting to pnpm-lock.yaml v.6.1


Dwlad90:
### What version of Turborepo are you using?

1.9.9

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

Turbo prune not support to pnpm-lock.yaml v6.1.

When run the next command:
`turbo prune --docker --scope=<workspace>`

turbo throw exception:
`Turbo error: Failed to encode pruned lockfile: Unable to generate pnpm-lock.yaml with lockfileVersion: 6.1. Supported lockfile versions are [5.3 5.4 6.0]`

### Expected Behavior

Prune workspace successfully

### To Reproduce

1. Install pnpm@8.6.0
2. run `pnpm install`
3. run `turbo prune --docker --scope=<workspace>`

weyert:
Yeah, some settings are getting stored in the `pnpm` file now. I think it's safe to ignore them

ChiuMungZitAlexander:
I think it is working now.

edouardr:
> I think it is working now.

nothing is mentioned in the changelog of [v.1.10.0](https://github.com/vercel/turbo/releases/tag/v1.10.0).
I have also tested for the sake of it, and it is still an issue.

piyushchauhan2011:
still an issue, just tried with v1.10.1
```
#0 0.484  ERROR  Failed to encode pruned lockfile: Unable to generate pnpm-lock.yaml with lockfileVersion: 6.1. Supported lockfile versions are [5.3 5.4 6.0]
#0 0.484 Turbo error: Failed to encode pruned lockfile: Unable to generate pnpm-lock.yaml with lockfileVersion: 6.1. Supported lockfile versions are [5.3 5.4 6.0]
```

rayepps:
Can confirm that manually changing the version `6.1` to `6.0` in the lockfile will get it to work. We know we're playing with fire doing this but for our use case we don't have any better choices. In CI/CD we use this `sed` just before pruning.

```
sed -i -e "s/lockfileVersion: '6.1'/lockfileVersion: '6.0'/g" pnpm-lock.yaml
```

*cred to @manas-vessel

piyushchauhan2011:
[...]

PR #5195, opened 2023-06-02
feat(lockfiles): add support pnpm lockfile version 6.1

chris-olszewski:
### Description

Adds parse support for the new lockfile version added in https://github.com/pnpm/pnpm/pull/6557 fixing #5117 

The behavioral change is pretty small for us. The new information was added to allow for `pnpm install --frozen-lockfile` to produce an identical lockfile even if the `.npmrc` that was present during the lockfile's creation is no longer present. This does lead to a better UX for prune users as they will no longer need to copy `.npmrc` in their `Dockerfile`.

### Testing Instructions

Existing unit tests pass and added new fixture with the settings field that correctly parses.

### DIFF
--- a/crates/turborepo-lockfiles/fixtures/pnpm-v6.1.yaml
+++ b/crates/turborepo-lockfiles/fixtures/pnpm-v6.1.yaml
@@ -1,0 +1,22 @@
+lockfileVersion: "6.1"
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+  .: {}
+
+  packages/a:
+    dependencies:
+      c:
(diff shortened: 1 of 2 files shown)


weyert:
Yeah, correct change. Made a similar change for the vulnerability scanner Gemnasium :)

mehulkar (approved):

tknickman (approved):

gajus:
This is a pretty major issue. Any chance of releasing as a hotfix?

gsoltis:
@gajus we just released `turbo@1.10.2-canary.2` that includes this change.

gajus:
thank you!

PR #5196, opened 2023-06-02
feat(turborepo): Remove anyhow from package manager detection

gsoltis:
### Description

 - Removes usage of `anyhow` from `package_manager` so that we have enumerated errors
 - moves `shim.rs` to use package manager detection
 - use `lazy_regex` for the package manager regex

Notable other changes
 - Package manager detection no longer requires a `CommandBase` instance, but the resulting error, in one case, no longer has a URL underlined. I've preserved the underline version in a currently-unused method on that error. TBD on the best way to plumb that out to the user.
 - I've added an implementation that allows `AnchorSystemPathBuf` to be fed directly to `wax`. I'm not yet sure we instead want to move that to `RelativeUnixPathBuf`. This needs some windows testing 

### Testing Instructions

Existing test suite

### DIFF
--- a/crates/turborepo-lib/src/package_manager/mod.rs
+++ b/crates/turborepo-lib/src/package_manager/mod.rs
@@ -6,11 +6,10 @@ use std::{
     backtrace,
     fmt::{self, Display},
     fs,
-    path::PathBuf,
 };
 
-use anyhow::{anyhow, Result as AnyhowResult};
 use itertools::{Either, Itertools};
+use lazy_regex::{lazy_regex, Lazy};
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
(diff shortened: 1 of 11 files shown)


crates/turborepo-lib/src/package_manager/mod.rs
```diff
@@ -127,43 +128,127 @@ impl Globs {
...
+
+impl Display for MissingWorkspaceError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let err = match self.package_manager {
+            PackageManager::Pnpm | PackageManager::Pnpm6 => {
+                "pnpm-workspace.yaml: no packages found. Turborepo requires pnpm workspaces and \
+                 thus packages to be defined in the root pnpm-workspace.yaml"
+            }
+            PackageManager::Yarn | PackageManager::Berry => {
+                "package.json: no workspaces found. Turborepo requires Yarn workspaces to be \
```
NicholasLYang, line 183:
```suggestion
                "package.json: no workspaces found. Turborepo requires yarn workspaces to be \
```

crates/turborepo-lib/src/package_manager/mod.rs
```diff
@@ -127,43 +128,127 @@ impl Globs {
...
+            PackageManager::Npm => {
+                "package.json: no workspaces found. Turborepo requires npm workspaces to be \
+                 defined in the root package.json"
+            }
+        };
+        write!(f, "{}", err)
+    }
+}
+
+impl From<&PackageManager> for MissingWorkspaceError {
```
NicholasLYang, line 195: Minor detail, but we could definitely derive `Copy` for `PackageManager` since it's essentially the size of a u8

NicholasLYang (approved):

crates/turborepo-lib/src/package_manager/mod.rs
gsoltis: I gave this a quick try, and it's less helpful than one might hope. The method where this conversion is used is currently defined on `&self`, meaning that the conversion needs `*self` if you want to take advantage of the copy. Given the size of the type, we could define methods on `self` instead, but I'm not sure that communicates what we want.

Issue #5075, opened 2023-05-23
[turborepo] ERROR Unable to run generate - missing requirements (npx)


jpgilchrist:
### What version of Turborepo are you using?

1.9.8

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Windows

### Describe the Bug

```
$ npm install turbo@latest -g
$ npx create-turbo@latest
$ cd {new-repo-path}
$ turbo gen workspace
```

outputs

```
ERROR Unable to run generate - missing requirements (npx)
```

### Expected Behavior

to create a new workspace and/or prompt with options

https://turbo.build/repo/docs/core-concepts/monorepos/code-generation

### To Reproduce

see description

kashan-ahmad:
Facing the same issue regarding both a custom generator as well as with workspaces.

tknickman:
Hey thanks for the report, can you try this with 1.10, and also, let me know the result of `which npx`?

Also, @kashan-ahmad are you on Windows as well?

Thank you!

business-door:
Yep, I'm on windows and I got turbo@1.9.7, seems like windows is gonna cost me a few bucks since I miss bash already.

ngopas:
same issue here

turbo version  - 1.10.1
npx version  9.6.4

tknickman:
@ngopas, are you using windows as well?

sofshant:
I'm using windows and also running into this issue

tknickman:
Got it, I can reproduce on windows. Thanks for the report! 

I'll get a fix up asap

tknickman:
This will release in canary shortly and we'll get a patch out hopefully tomorrow!

tknickman:
Should be fixed in `1.10.2`

sofshant:
Can confirm its fixed thanks!

PR #5208, opened 2023-06-05
fix(gen): swap to which for windows

tknickman:
### Description

Fixes https://github.com/vercel/turbo/issues/5075

Verified this works via parallels:
CleanShot 2023-06-05 at 13 59 01@2x

### DIFF
--- a/crates/turborepo-lib/src/commands/generate.rs
+++ b/crates/turborepo-lib/src/commands/generate.rs
@@ -2,6 +2,7 @@ use std::process::{Command, Stdio};
 
 use anyhow::Result;
 use tracing::debug;
+use which::which;
 
 use crate::{
     child::spawn_child,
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/commands/generate.rs
```diff
@@ -2,22 +2,31 @@ use std::process::{Command, Stdio};
...
 };
 
 fn verify_requirements() -> Result<()> {
-    let output = Command::new("npx")
-        .arg("--version")
-        .stdout(Stdio::null())
-        .stderr(Stdio::null())
-        .status();
+    // find npx path
+    let npx_path = which("npx");
```
chris-olszewski, line 14: Minor thing, I think we can make use of [`context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#example) here and still expose the underlying error and reduce nesting by removing the `match npx_path` statement:
```suggestion
    use anyhow::Context;
    let npx_path = which("npx").context("Unable to run generate - missing requirements (npx)")?;
```

chris-olszewski (approved):

crates/turborepo-lib/src/commands/generate.rs
tknickman: Good call, updated!

NicholasLYang (approved):

PR #5210, opened 2023-06-05
Prepare StaticAssetsContentSourceVc for basePath

jridgewell:
### Description

This is part 1 of WEB-993 `basePath` support. The static assets (things served in the `/public` directory) can only serve requests that include the `basePath`, so we'll need to pass that in. In order to not completely invalidate the [source](https://github.com/vercel/next.js/blob/2b1f0d9351610b04d01638efed19252ca81d0023/packages/next-swc/crates/next-dev/src/lib.rs#L379-L380) function (and all the things that depend on cells created there, ie everything) every time the user edits the `next.config.js`, we'll need to accept a `StringVc`. This sets us up without breaking backwards compat, making this multi-PR adventure a little easier.

### DIFF
--- a/crates/turbopack-dev-server/src/source/static_assets.rs
+++ b/crates/turbopack-dev-server/src/source/static_assets.rs
@@ -78,32 +78,29 @@ impl Introspectable for StaticAssetsContentSource {
     #[turbo_tasks::function]
     async fn children(&self) -> Result<IntrospectableChildrenVc> {
         let dir = self.dir.read_dir().await?;
-        let children = match &*dir {
-            DirectoryContent::NotFound => Default::default(),
-            DirectoryContent::Entries(entries) => {
-                let prefix = self.prefix.await?;
-                entries
-                    .iter()
-                    .map(|(name, entry)| {
-                        let child = match entry {
-                            DirectoryEntry::File(path) | DirectoryEntry::Symlink(path) => {
-                                IntrospectableAssetVc::new(SourceAssetVc::new(*path).as_asset())
-                            }
-                            DirectoryEntry::Directory(path) => {
-                                StaticAssetsContentSourceVc::with_prefix(
-                                    StringVc::cell(format!("{}{name}/", &*prefix)),
-                                    *path,
-                                )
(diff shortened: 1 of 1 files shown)


crates/turbopack-dev-server/src/source/static_assets.rs
```diff
@@ -16,19 +16,27 @@ use super::{
...
 pub struct StaticAssetsContentSource {
-    pub prefix: String,
+    pub prefix: StringVc,
     pub dir: FileSystemPathVc,
 }
 
 #[turbo_tasks::value_impl]
 impl StaticAssetsContentSourceVc {
     #[turbo_tasks::function]
     pub fn new(prefix: String, dir: FileSystemPathVc) -> StaticAssetsContentSourceVc {
```
ForsakenHarmony, line 26: Is the plan to remove this later?
The different here only being `String` vs `StringVc` make the naming confusing

crates/turbopack-dev-server/src/source/static_assets.rs
```diff
@@ -71,24 +80,29 @@ impl Introspectable for StaticAssetsContentSource {
...
-                            *path,
-                        )
-                        .into(),
-                        DirectoryEntry::Other(_) => todo!("what's DirectoryContent::Other?"),
-                        DirectoryEntry::Error => todo!(),
-                    };
-                    (StringVc::cell(name.clone()), child)
-                })
-                .collect(),
+            DirectoryContent::Entries(entries) => {
```
ForsakenHarmony, line 83: can we use an early return for `NotFound` to reduce the indent here?

ForsakenHarmony (approved):

crates/turbopack-dev-server/src/source/static_assets.rs
jridgewell: Done.

crates/turbopack-dev-server/src/source/static_assets.rs
jridgewell: Yup.

PR #5215, opened 2023-06-06
ci(workflow): upload benchmark results to datadog

kwonoj:
### Description

Initial support for uploading benchmark results into datadog. It currently iterates over the raw benchmark data, and uploda `mean` and `std_dev`'s `point_estimate`.

As noted in the comment random 400 returned by datadog, which needs some investigations. For now accept those flaky as it won't block CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1052,6 +1052,30 @@ jobs:
         if: always()
         run: critcmp --export base > raw.json
 
+      - name: Upload results to datadog
+        if: always()
+        continue-on-error: true
+        env:
+          DATADOG_API_KEY: ${{ secrets.DD_KEY_TURBOPACK }}
+        run: |
+          npm install -g @datadog/datadog-ci
+          # Query raw benchmark output, create key:value pairs for each benchmark entries.
+          # The generated key name is compact format the path of the benchmark entry, i.e
+          # `base.hmr_to_commit.CSR.1000.mean`
+          # [TODO]: datadog-ci sometimes return 400 without detail, need to investigate further. For now accept the flaky uploads.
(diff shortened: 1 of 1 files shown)


tknickman (approved):

.github/workflows/test.yml
```diff
@@ -1027,6 +1027,30 @@ jobs:
...
+                echo ${item} | base64 --decode | jq -r ${1}
+              }
+
+              export METRICS_MEAN+=" --metrics $(_jq ".name" | sed -e 's/\/bench_/\//g' | sed -e 's/\/Turbopack /\//g' | sed -e 's/ modules//g'| sed -e 's/ /./g' | sed -e 's/\//./g').mean:$(printf "%0.2f" $(_jq ".mean"))"
+              export METRICS_STD+=" --metrics $(_jq ".name" | sed -e 's/\/bench_/\//g' | sed -e 's/\/Turbopack /\//g' | sed -e 's/ modules//g'| sed -e 's/ /./g' | sed -e 's/\//./g').std:$(printf "%0.2f" $(_jq ".std"))"
+          done
+          echo "Sending metrics $METRICS_MEAN"
+          datadog-ci metric --level pipeline --no-fail $METRICS_MEAN
+          echo "Sending metrics $METRICS_STD"
+          datadog-ci metric --level pipeline --no-fail $METRICS_STD
```
ForsakenHarmony, line 1052: Can we maybe do this in a `.js` script outside this yaml?

kwonoj: Technically yes, for now I didn't feel it requires additional script externally to the ci. I guess we'll probably need to iterate this over for few times, and that may bring the need to have a wrapper script instead.

wbinnssmith (approved):

PR #5218, opened 2023-06-06
Prepare RouterContentSource for basePath

jridgewell:
### Description

This is part 2 of [WEB-993](https://linear.app/vercel/issue/WEB-993) basePath support. A few of our next-specific content sources will need to be scoped under the `basePath` (like `_next/image` and `__nextjs_original-stack-frame`). These are currently served with a `RouterContentSource`, but it didn't have support for arbitrary prefixes.

We _could_ have changed the subpath for these sources to include the `basePath`, but that would require reading the `next_config.base_path()` in the [source](https://github.com/vercel/next.js/blob/2b1f0d9351610b04d01638efed19252ca81d0023/packages/next-swc/crates/next-dev/src/lib.rs#L413-L423) method, and it would invalidate our entire call graph whenever the `next.config.js` changed. Not a good choice.

### DIFF
--- a/crates/turbopack-dev-server/src/source/static_assets.rs
+++ b/crates/turbopack-dev-server/src/source/static_assets.rs
@@ -22,6 +22,7 @@ pub struct StaticAssetsContentSource {
 
 #[turbo_tasks::value_impl]
 impl StaticAssetsContentSourceVc {
+    // TODO: Remove this method and migrate users to `with_prefix`.
     #[turbo_tasks::function]
     pub fn new(prefix: String, dir: FileSystemPathVc) -> StaticAssetsContentSourceVc {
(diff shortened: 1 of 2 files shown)


crates/turbopack-dev-server/src/source/static_assets.rs
```diff
@@ -22,6 +22,7 @@ pub struct StaticAssetsContentSource {
 
 #[turbo_tasks::value_impl]
 impl StaticAssetsContentSourceVc {
+    // TODO: Remove this method and migrate users to `with_prefix`.
```
alexkirsz, line 25: Can you link to an issue here? `TODO(WEB-XXX)`

alexkirsz (approved):

PR #5228, opened 2023-06-06
feat(turborepo): Root turbo.json parsing

NicholasLYang:
### Description

Implements root turbo json parsing with conversion from a `RawTurboJson` as deserialized from JSON to a `TurboJson` with a `BookkeepingTaskDefinition`. Includes synthesis from package.json file.

### Testing Instructions

Adds testing for deserialization and synthesis.


crates/turborepo-lib/src/run/mod.rs
```diff
@@ -75,31 +79,7 @@ impl Run {
...
-        if filtered_pkgs.len() == pkg_dep_graph.len() {
-            for target in targets {
-                let key = task_id::root_task_id(target);
-                if pipeline.contains_key(&key) {
-                    filtered_pkgs.insert(task_id::ROOT_PKG_NAME.to_string());
-                    break;
-                }
-            }
-        }
+        let _filtered_pkgs = scope::resolve_packages(&opts.scope_opts, &self.base, &pkg_dep_graph)?;
```
NicholasLYang, line 82: This is getting deleted mostly because the lifetimes weren't worth resolving here and the filter/scope code will be done in a different PR.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -76,34 +80,11 @@ impl Run {
...
-
-        let pipeline = &turbo_json.pipeline;
-
         let scm = SCM::new(&self.base.repo_root);
 
-        let mut filtered_pkgs =
+        let _filtered_pkgs =
             scope::resolve_packages(&opts.scope_opts, &self.base, &pkg_dep_graph, &scm)?;
 
-        if filtered_pkgs.len() == pkg_dep_graph.len() {
```
NicholasLYang, line 97: Deleted because it's causing some lifetime issues and I figure this code will be handled by the actual filter/scope logic.

crates/turborepo-lib/src/run/graph.rs
NicholasLYang, line 1: Deleted because we don't need the `CompleteGraph` abstraction

crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -26,6 +28,7 @@ pub struct PackageGraph {
 pub struct Entry {
     package_json: PackageJson,
     package_json_path: AnchoredSystemPathBuf,
+    turbo_json: TurboJson,
```
chris-olszewski, line 31: I feel like this colocation is unnecessary and restricts us a bit as it means that we need to eagerly parse all `turbo.json` in order to construct a package graph and for some operations this isn't necessary.

crates/turborepo-lib/src/run/mod.rs
chris-olszewski: Maybe leave this commented or a TODO to make sure it doesn't get forgotten since in Go land this logic is in `run` and not the filter/scope module.

chris-olszewski (changes requested):
Mostly looks good. I'm primarily concerned with bundling the `TurboJson` so tightly with the package graph when it could easily be constructed from the package graph. In Go we populate each of the workspace's `TurboJson` when constructing the task graph, but this would force us to move that up to when we construct the package graph.

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -426,7 +426,7 @@ fn add_space_id_to_turbo_json(base: &CommandBase, space_id: &str) -> Result<()>
     }
 
     let turbo_json_file = File::open(&turbo_json_path)?;
-    let mut turbo_json: TurboJson = serde_json::from_reader(turbo_json_file)?;
+    let mut turbo_json: RawTurboJSON = serde_json::from_reader(turbo_json_file)?;
```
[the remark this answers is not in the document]
NicholasLYang: Huh TIL. I would have thought `from_reader` was faster if anything

crates/turborepo-lib/src/commands/info.rs
```diff
@@ -14,7 +16,15 @@ pub fn run(base: &mut CommandBase, workspace: Option<&str>) -> Result<()> {
...
         PackageManager::get_package_manager(&base.repo_root, Some(&root_package_json))?;
 
-    let package_graph = PackageGraph::builder(&base.repo_root, root_package_json)
+    let opts: Opts = base.args().try_into()?;
+
+    let root_turbo_json = TurboJson::load(
+        &base.repo_root,
+        &root_package_json,
+        opts.run_opts.single_package,
+    )?;
```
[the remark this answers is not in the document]
NicholasLYang: Yeah it was a dumb reason, I just didn't want to clone the `root_package_json` since `PackageGraph` takes ownership. I'll clone it for now, and maybe we can add a way to get a reference from package_graph sometime.

chris-olszewski (approved):
Overall looks good to me, only concern is that we might crash on parsing a legacy turbo config in `package.json`.

crates/turborepo-lib/src/execution_state.rs
```diff
@@ -28,11 +28,10 @@ impl<'a> TryFrom<&'a CommandBase> for ExecutionState<'a> {
     type Error = anyhow::Error;
 
     fn try_from(base: &'a CommandBase) -> Result<Self, Self::Error> {
-        let root_package_json =
-            PackageJson::load(&base.repo_root.join_component("package.json")).ok();
+        let root_package_json = PackageJson::load(&base.repo_root.join_component("package.json"))?;
```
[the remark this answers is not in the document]
NicholasLYang: Hmm I don't remember why I made this change. Will revert, although I do think we need to rethink our package manager inference a little post-port because silently failing on package.json parsing can cause some issues.

PR #5231, opened 2023-06-06
refactor: update tsconfig in basic example

PaulRBerg:
There's no need for re-declaring the `plugins` in each app's `tsconfig.json` file.

### DIFF
--- a/examples/basic/apps/docs/tsconfig.json
+++ b/examples/basic/apps/docs/tsconfig.json
@@ -1,8 +1,5 @@
 {
   "extends": "tsconfig/nextjs.json",
-  "compilerOptions": {
-    "plugins": [{ "name": "next" }]
-  },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
   "exclude": ["node_modules"]
 }
--- a/examples/basic/apps/web/tsconfig.json
+++ b/examples/basic/apps/web/tsconfig.json
@@ -1,8 +1,5 @@
 {
   "extends": "tsconfig/nextjs.json",
-  "compilerOptions": {
-    "plugins": [{ "name": "next" }]
-  },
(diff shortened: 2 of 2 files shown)


tknickman:
I believe Next adds these automatically if they're missing - cc @anthonyshew. We will chat with them to make sure this can support tsconfigs with extends.

tknickman (commented):
Agree with the change, however I believe Next.js adds these automatically if they're missing. We will chat with them to make sure this can support tsconfigs with extends.

cc @anthonyshew

PaulRBerg:
I don't think Next.js adds the `plugins` automatically - running `pnpm run build` in `apps/web` on my branch produces this console log:

> Your tsconfig.json extends another configuration, which means we cannot add the Next.js TypeScript plugin automatically. To improve your development experience, we recommend adding the Next.js plugin (`"plugins": [{ "name": "next" }]`) manually to your TypeScript configuration. Learn more: https://beta.nextjs.org/docs/configuring/typescript#using-the-typescript-plugin

anthonyshew:
Hi!

The root issue here is that (I think) Next.js parses the `tsconfig` to look for the plugin directly in its own file rather than reading the AST and any configurations being extended. So this puts monorepo-ing with the plugin between a rock and a hard place. You either have to accept the log line or you duplicate the configuration.

While I do agree de-deduplication would be ideal, the log that gets produced that you've mentioned, @PaulRBerg, would probably earn us an issue looking something like "Turborepo not working with Next.js because of missing plugin" or similar. 

We get a lot of issues/questions around misconfiguration of other tooling where folks perceive that Turborepo is the culprit (when it isn't) and this would end up being another one of those cases. We've found that the best way to keep these sorts of issue reports and pull requests down is, when we have the choice, to keep the logs clean, so I'm going to err on that side of things and close this.

I definitely respect the intent of your PR! Ideally, I would be hitting the big green button to merge this in - but we're going to have to work with what we have today and let the duplication stand. I appreciate you wanting to do some clean up here! We'll talk to the Next.js team about this and see what we can do.

Thank you!

PR #5237, opened 2023-06-07
feat(turborepo): Port Manual hashing

gsoltis:
### Description

 - ports `getPackageFileHashesFromProcessingGitIgnore` to Rust
 - wires up Go code to use FFI bridge for manual file hashing
 - adds `Deref` coercion for `RelativeUnixPathBuf`

### Testing Instructions

 - Existing test suite
 - Ports manual hashing test

### DIFF
--- a/crates/turborepo-scm/src/manual.rs
+++ b/crates/turborepo-scm/src/manual.rs
@@ -33,13 +33,6 @@ fn get_package_file_hashes_from_processing_gitignore(
     let mut walker_builder = WalkBuilder::new(&full_package_path);
     let mut includes = Vec::new();
     let mut excludes = Vec::new();
-    // for pattern in inputs {
-    //     if pattern.starts_with("!") {
-    //         excludes.push(full_package_path.join_literal(&pattern[1..]).
-    // to_string());     } else {
-    //         includes.push(full_package_path.join_literal(pattern).to_string());
-    //     }
-    // }
     for pattern in inputs {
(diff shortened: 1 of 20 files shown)


crates/turborepo-scm/src/manual.rs
```diff
@@ -0,0 +1,210 @@
...
+        let tmp_dir = tempfile::tempdir().unwrap();
+        let dir = AbsoluteSystemPathBuf::new(tmp_dir.path().to_path_buf())
+            .unwrap()
+            .to_realpath()
+            .unwrap();
+        (tmp_dir, dir)
+    }
+
+    #[test]
+    fn test_get_package_file_hashes_from_processing_gitignore() {
```
NicholasLYang, line 117: not a big deal but it might be nice to use something like [insta](https://insta.rs) or [test_case](https://crates.io/crates/test-case) for these kinds of tests

NicholasLYang (approved):

crates/turborepo-scm/src/manual.rs
gsoltis: I'm kinda on the fence. I like test_case, Alex has been using it I think, but it interacts a little poorly with debugging shortcuts in VSCode once you have more than a couple cases.

NicholasLYang: yeah and it doesn't interact with rustfmt that well. I gave up on it for some more involved test cases in turborepo-cache

PR #5238, opened 2023-06-08
ci: use latest published version of turbo in CI instead of latest canary

mehulkar:


mehulkar:
The failing examples tests here are fixed in #5240

tknickman (approved):

PR #5241, opened 2023-06-08
feat(run summary): Add whether turbo detected monorepo or not

mehulkar:
While debugging another issue, I came across a confusing behavior where it seemed like turbo was operating in single package mode. I added this bit of information in to get some visibility in, and figured it was a good feature for everyone.


tknickman (approved):

NicholasLYang:
Not a huge deal, but maybe we should have this as a mode field so it's more extensible?

mehulkar:
@NicholasLYang that feels to general, but let me know if you have a proposal

PR #5242, opened 2023-06-08
ci: update examples tests to diff full output 

mehulkar:
This will help debug examples tests when they fail.

(This is lifted from https://github.com/vercel/turbo/pull/5011)


turborepo-tests/examples/tests/pnpm-gatsby.t
```diff
@@ -1,5 +1,30 @@
...
+  
+  > with-gatsby@0.0.0 build /private/var/folders/vg/sr4krlws0k12g21phhjwy4z40000gn/T/prysk-tests-b73vp9vw/pnpm-gatsby.t
+  > turbo build "lint" "--output-logs=none"
+  
+  \xe2\x80\xa2 Packages in scope: docs, eslint-config-custom, tsconfig, ui, web (esc)
+  \xe2\x80\xa2 Running build, lint in 5 packages (esc)
+  \xe2\x80\xa2 Remote caching disabled (esc)
+  
+   Tasks:    5 successful, 5 total
+  Cached:    3 cached, 5 total
```
mehulkar, line 28: Not going to try to fix it here, but it's weird that the gatsby example doesn't get full turbo on the second run. might be expected

mehulkar: (This test isn't even running in CI, because this combo isn't part of the matrix)

gsoltis (approved):

PR #5255, opened 2023-06-08
feat(lockfiles): add rust implementation for yarn1

chris-olszewski:
### Description

Ports the Yarn v1 lockfile logic to Rust

Reviewer notes:
 - Since the v1 lockfile format is very straightforward and forgiving, a majority of this PR is around parsing SYML. The parser is built using nom based on the [PEG grammar](https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-parsers/sources/grammars/syml.pegjs) that Berry provides. This isn't an exact copy of that grammar as the Berry provided grammar is overly permissive compared to the format the Yarn v1 uses. We drop this support to match the parser that Yarn 1 uses.
 - This PR doesn't do any cleanup that having all lockfiles in Rust allows us to do. This will be done in a follow up PR

### Testing Instructions

Existing integration tests, ported unit tests. Manual checking that we correctly prune repos created via `npx create-turbo`


crates/turborepo-lockfiles/src/yarn1/de.rs
```diff
@@ -0,0 +1,320 @@
...
+    sequence::{delimited, pair, preceded, separated_pair, terminated, tuple},
+    IResult,
+};
+use regex::Regex;
+use serde_json::Value;
+
+// regex for trimming spaces from start and end
+fn pseudostring_replace() -> &'static Regex {
+    static RE: OnceLock<Regex> = OnceLock::new();
+    RE.get_or_init(|| Regex::new(r"^ *| *$").unwrap())
```
gsoltis, line 18: Not blocking, but maybe consider the `lazy_regex` crate to encapsulate this pattern?

crates/turborepo-lockfiles/src/yarn1/syml.pest
```diff
@@ -0,0 +1,57 @@
+grammar = { SOI ~ property_statements ~ EOI }
```
gsoltis, line 1: Are we using this file?

gsoltis (approved):

NicholasLYang (approved):

PR #5271, opened 2023-06-12
feat(ci): conditionally set jest verbosity

tknickman:
### Description

Only run jest in verbose mode when we have also set the github action to run in verbose mode.


tknickman:
Yea just saw that as well, just added silent as well.

PR #5272, opened 2023-06-12
chore(turborepo): Use cargo-groups

NicholasLYang:
### Description

Use [cargo-groups](https://github.com/nicholaslyang/cargo-groups) to group turborepo and turbopack crates in the workspace.

Also rearranged the test.yml workflow so that we have explicit turborepo/turbopack clippy jobs so we're not running clippy on everything all the time, and renamed a couple workflows accordingly.

### DIFF
--- a/crates/turborepo-globwalk/src/lib.rs
+++ b/crates/turborepo-globwalk/src/lib.rs
@@ -365,6 +365,7 @@ mod test {
         collapse_path, empty_glob::InclusiveEmptyAny, globwalk, MatchType, WalkError, WalkType,
     };
 
+    #[cfg(unix)]
     #[test_case("a/./././b", "a/b", 1 ; "test path with dot segments")]
     #[test_case("a/../b", "b", 0 ; "test path with dotdot segments")]
     #[test_case("a/./../b", "b", 0 ; "test path with mixed dot and dotdot segments")]
@@ -386,6 +387,7 @@ mod test {
         assert_eq!(segment, earliest_collapsed_segement);
     }
 
+    #[cfg(unix)]
     #[test_case("../a/b" ; "test path starting with ../ segment should return None")]
     #[test_case("/../a" ; "test path with leading dotdotdot segment should return None")]
     fn test_collapse_path_not(glob: &str) {
@@ -466,6 +468,7 @@ mod test {
         );
     }
 
+    #[cfg(unix)]
     #[test]
     fn do_match_empty_include() {
         let patterns: [&str; 0] = [];
@@ -535,6 +538,7 @@ mod test {
         tmp
     }
 
+    #[cfg(unix)]
     #[test_case("abc", 1, 1 => matches None ; "exact match")]
     #[test_case("*", 19, 15 => matches None ; "single star match")]
     #[test_case("*c", 2, 2 => matches None ; "single star suffix match")]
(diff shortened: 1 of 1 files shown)


crates/turborepo-ffi/src/lib.rs
```diff
@@ -444,14 +444,14 @@ pub extern "C" fn get_package_file_hashes_from_inputs(buffer: Buffer) -> Buffer
                 to_return.insert(filename, hash);
             }
             let file_hashes = proto::FileHashes { hashes: to_return };
-            let resp = proto::GetPackageFileHashesFromInputsResponse {
```
alexkirsz, line 447: Did you mean to include this change?

crates/turborepo-env/src/lib.rs
```diff
@@ -224,7 +224,8 @@ pub fn get_global_hashable_env_vars(
 ) -> Result<DetailedMap, Error> {
     let default_env_var_map = env_at_execution_start.from_wildcards(&DEFAULT_ENV_VARS[..])?;
 
-    let user_env_var_set = env_at_execution_start.from_wildcards_unresolved(global_env)?;
+    let user_env_var_set =
+        env_at_execution_start.wildcard_map_from_wildcards_unresolved(global_env)?;
```
alexkirsz, line 228: Did you mean to include this change?

.github/workflows/test.yml
```diff
@@ -525,64 +501,41 @@ jobs:
...
+      - name: Run cargo check
         run: |
-          cargo check --workspace --all-targets --release --features rustls-tls
+          cargo groups check turborepo-libraries --features rustls-tls
 
       - name: Run cargo clippy
         run: |
-          cargo clippy --workspace --all-targets --features rustls-tls
-
-      - name: Count clippy warnings
```
alexkirsz, line 536: Do we no longer have any clippy warnings?

This might still be useful when clippy adds warnings in the future.

crates/turborepo-ffi/src/lib.rs
NicholasLYang: Yes, since I changed the clippy code to error on any warnings, I had to fix this one

.github/workflows/test.yml
NicholasLYang: The clippy code that checks if our warnings are decreasing was very buggy and lead to a bunch of warnings sneaking in. I'm of the view that we should avoid any warnings in the first place, and pin the Rust version. When we bump the Rust version, we should fix any new warnings too. Of course, now that the clippy checks are split between turbopack/turborepo, turbopack is welcome to do something different.

alexkirsz (approved):

chris-olszewski (approved):

.github/workflows/test.yml
```diff
@@ -454,35 +454,11 @@ jobs:
...
-      - name: Run cargo check
-        uses: actions-rs/cargo@v1
-        with:
-          command: check
-          args: --workspace --all-targets --locked --features rustls-tls
-
-  rust_lint:
-    needs: [determine_jobs, rust_prepare]
+  # Very quick checks that we want to run first before all the long-running checks.
+  rust_pre_check:
```
ForsakenHarmony, line 458: merging these now means there's an extra 5 minute delay while it's checking licenses :/

ForsakenHarmony: And that format errors now block tests from running

PR #5273, opened 2023-06-12
fix(turborepo): avoid globbing directories due to ancestor truncation

gsoltis:
### Description
WIP

 - uses `wax::walk` to do globwalking in a loop over globs (`Any` is not supported by `wax`)
 - fixes an issue in `wax` where globwalking doesn't work for absolute path-based globs in unix
 - trades better glob-prefix-matching for the possibility of returning duplicate entries (test is currently failing)

### Testing Instructions

 - existing tests, new test added for exploration but will be removed

### DIFF
--- a/crates/turborepo-globwalk/src/lib.rs
+++ b/crates/turborepo-globwalk/src/lib.rs
@@ -11,8 +11,8 @@ use std::{
 use empty_glob::InclusiveEmptyAny;
 use itertools::Itertools;
 use path_slash::PathExt;
-use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
-use wax::{Any, BuildError, Glob, Pattern};
+use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf, PathError};
+use wax::{Any, BuildError, Glob, Pattern, WalkBehavior, WalkEntry};
 
 #[derive(Debug, PartialEq, Clone, Copy)]
 pub enum WalkType {
@@ -50,6 +50,10 @@ pub enum WalkError {
     InvalidPath,
     #[error("walk error: {0}")]
     WalkError(#[from] walkdir::Error),
+    #[error(transparent)]
+    Path(#[from] PathError),
+    #[error(transparent)]
+    WaxWalk(#[from] wax::WalkError),
 }
 
 /// Performs a glob walk, yielding paths that _are_ included in the include list
@@ -62,7 +66,7 @@ pub enum WalkError {
 ///       - collapse the path, and calculate the new base_path, which defined as
 ///         the longest common prefix of all the includes
 ///       - traversing above the root of the base_path is not allowed
-pub fn globwalk(
+pub fn _globwalk(
     base_path: &AbsoluteSystemPath,
     include: &[String],
     exclude: &[String],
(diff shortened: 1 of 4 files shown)


chris-olszewski (commented):
Some initial comments

crates/turborepo-globwalk/src/lib.rs
```diff
@@ -277,6 +281,79 @@ fn collapse_path(path: &str) -> Option<(Cow<str>, usize)> {
...
+    base_path: &AbsoluteSystemPath,
+    include: &[String],
+    exclude: &[String],
+    walk_type: WalkType,
+) -> Result<Vec<Result<AbsoluteSystemPathBuf, WalkError>>, WalkError> {
+    let (base_path_new, include_paths, exclude_paths) =
+        preprocess_paths_and_globs(base_path, include, exclude)?;
+    let inc_patterns = include_paths
+        .iter()
+        .map(|g| Glob::new(g.as_str()).map(|g| g.into_owned()))
```
chris-olszewski, line 294:
```suggestion
        .map(|g| Glob::new(g.as_str()))
```

crates/turborepo-globwalk/src/lib.rs
```diff
@@ -277,6 +281,79 @@ fn collapse_path(path: &str) -> Option<(Cow<str>, usize)> {
...
+        .collect::<Result<Vec<_>, BuildError>>()?;
+    let ex_patterns = exclude_paths
+        .iter()
+        .map(|g| Glob::new(g.as_str()).map(|g| g.into_owned()))
```
chris-olszewski, line 298:
```suggestion
        .map(|g| Glob::new(g.as_str()))
```

crates/turborepo-globwalk/src/lib.rs
```diff
@@ -277,6 +281,79 @@ fn collapse_path(path: &str) -> Option<(Cow<str>, usize)> {
...
+                }
+            } else {
+                glob.walk(&base_path_new)
+                    .not(ex_patterns.iter().map(|g| g.to_owned()))
+                    // Per docs, only fails if exclusion list is too large, since we're using
+                    // pre-compiled globs
+                    .expect(&format!(
+                        "Failed to compile exclusion globs: {:?}",
+                        ex_patterns
+                    ))
```
chris-olszewski, line 339: Avoid the allocation unless necessary
```suggestion
                    .unwrap_or_else(|e| panic!(
                        "Failed to compile exclusion globs: {:?}: {}",
                        ex_patterns,
                        e,
                    ))
```

crates/turborepo-globwalk/src/lib.rs
```diff
@@ -277,6 +281,79 @@ fn collapse_path(path: &str) -> Option<(Cow<str>, usize)> {
...
+                        if walk_type == WalkType::Files && md.is_dir() {
+                            vec![]
+                        } else {
+                            vec![Ok(path)]
+                        }
+                    }
+                }
+            } else {
+                glob.walk(&base_path_new)
+                    .not(ex_patterns.iter().map(|g| g.to_owned()))
```
chris-olszewski, line 333:
```suggestion
                    .not(ex_patterns.iter().cloned())
```

crates/turborepo-globwalk/src/lib.rs
```diff
@@ -277,6 +281,79 @@ fn collapse_path(path: &str) -> Option<(Cow<str>, usize)> {
...
+        .map(|g| Glob::new(g.as_str()).map(|g| g.into_owned()))
+        .collect::<Result<Vec<_>, _>>()?;
+
+    let result = inc_patterns
+        .into_iter()
+        .flat_map(|glob| {
+            // Check if the glob specifies an exact filename with no meta characters.
+            if let Some(prefix) = glob.variance().path() {
+                // We expect all of our globs to be absolute paths
+                debug_assert!(glob.has_root());
```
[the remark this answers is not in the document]
gsoltis: Heh, I think this is actually a pointer at a different error. This particular assert is for an internal error, `preprocess_paths_and_globs` should only return globs that are absolute paths (they are joined to `base_path`). User input into the function should not be able to trigger it.

However, from a quick read, it looks like we instead will just skip invalid globs in `collapse_paths`, called by `preprocess_paths_and_globs`, which should instead be an error.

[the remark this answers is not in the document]
gsoltis: Left the debug assert, but added the explicit error case above as well.

crates/turborepo-globwalk/src/lib.rs
```diff
@@ -277,6 +284,86 @@ fn collapse_path(path: &str) -> Option<(Cow<str>, usize)> {
...
+            Glob::new(g.as_str()).map_err(|e| e.into()).and_then(|g| {
+                if g.has_root() {
+                    Ok(g)
+                } else {
+                    Err(WalkError::InternalError {
+                        glob: g.to_string(),
+                        error: "expected glob to have a root".to_string(),
+                    })
+                }
+            })
```
NicholasLYang, line 307:
```suggestion
            match Glob::new(g.as_str()) {
                Err(e) => e.into(),
                Ok(g) if g.has_root() => Ok(g),
                Ok(g) => Err(WalkError::InternalError {
                    glob: g.to_string(),
                    error: "expected glob to have a root".to_string(),
                })
            }
```

crates/turborepo-globwalk/src/lib.rs
```diff
@@ -277,6 +284,86 @@ fn collapse_path(path: &str) -> Option<(Cow<str>, usize)> {
...
+                        )
+                    })
+                    .filter_map(|entry| match entry {
+                        Ok(entry) => {
+                            if walk_type == WalkType::Files && entry.file_type().is_dir() {
+                                None
+                            } else {
+                                Some(AbsoluteSystemPathBuf::new(entry.path()).map_err(|e| e.into()))
+                            }
+                        }
```
NicholasLYang, line 357:
```suggestion
                        Ok(entry) if walk_type == WalkType::Files && entry.file_type().is_dir() => None, 
                        Ok(entry) => Some(AbsoluteSystemPathBuf::new(entry.path()).map_err(|e| e.into()))
```

crates/turborepo-wax/src/walk.rs
```diff
@@ -570,9 +578,22 @@ pub struct Walk<'g> {
     root: PathBuf,
     prefix: PathBuf,
     walk: walkdir::IntoIter,
+    // This is a hack to express an empty iterator
+    is_empty: bool,
```
NicholasLYang, line 582: This isn't blocking, but I feel like we could either make this into an `Option` or have `walk::walk` return a generic `impl Iterator` and then return an empty iterator instead.

gsoltis: Agreed, this is definitely not the optimal way of doing this.

crates/turborepo-globwalk/src/lib.rs
gsoltis: This is sadly irrelevant in the latest revision, we can't rely on globs being rooted, as they don't count as rooted on windows.

NicholasLYang (approved):

chris-olszewski (approved):

PR #5274, opened 2023-06-12
fix(turborepo): Tighten build globs

gsoltis:
### Description

 - scopes `cli#build` inputs globs to fewer files. Previous version captured ~25k files, this is ~300

### Testing Instructions

 - verified shorter time to glob for integration tests

### DIFF
--- a/cli/turbo.json
+++ b/cli/turbo.json
@@ -14,9 +14,11 @@
       ],
 
       "inputs": [
-        "**/*.go",
-        "**/*_test.go",
-        "../crates/turborepo*/**" // Rust crates
+        "{internal,cmd}/**/*.go",
+        "!**/*_test.go",
+        "../crates/turborepo*/**/*.rs", // Rust crates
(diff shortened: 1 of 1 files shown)


NicholasLYang (approved):

cli/turbo.json
```diff
@@ -14,9 +14,11 @@
       ],
 
       "inputs": [
-        "**/*.go",
-        "**/*_test.go",
-        "../crates/turborepo*/**" // Rust crates
+        "{internal,cmd}/**/*.go",
+        "!**/*_test.go",
```
mehulkar, line 18: we don't want to build again when unit tests change?

mehulkar (approved):

cli/turbo.json
gsoltis: I don't think so...they don't influence the build. If they uncovered a bug, there would have to be a corresponding source code change for a fix, which would trigger a build.

Issue #4009, opened 2023-02-28
[turborepo] `npx @turbo/codemod update` throws an error


hymair:
### What version of Turborepo are you using?

1.8.1

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

after running `npx @turbo/codemod@latest update --force` :
``` 
WARNING: Git directory is not clean. Forcibly continuing...
? Where is the root of the repo to migrate? .
No codemods required to migrate from 1.8.1 to 1.8.3

Upgrading turbo from 1.8.1 to 1.8.3 (no codemods required)

Upgrading turbo with pnpm install turbo@latest --save-dev -w


Unexpected error. Please report it as a bug:
Error: Command failed: pnpm install turbo@latest --save-dev -w
    at checkExecSyncError (node:child_process:871:11)
    at execSync (node:child_process:943:15)
    at Command.N (/Users/i/.npm/_npx/36146fcc85b26196/node_modules/@turbo/codemod/dist/cli.js:4:8282)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Command.parseAsync (/Users/i/.npm/_npx/36146fcc85b26196/node_modules/commander/lib/command.js:935:5) {
  status: 1,
  signal: null,
  output: [
    null,
    <Buffer e2 80 89 45 52 52 5f 50 4e 50 4d 5f 48 4f 49 53 54 5f 50 41 54 54 45 52 4e 5f 44 49 46 46 e2 80 89 20 54 68 69 73 20 6d 6f 64 75 6c 65 73 20 64 69 72 ... 112 more bytes>,
    <Buffer >
  ],
  pid: 23255,
  stdout: <Buffer e2 80 89 45 52 52 5f 50 4e 50 4d 5f 48 4f 49 53 54 5f 50 41 54 54 45 52 4e 5f 44 49 46 46 e2 80 89 20 54 68 69 73 20 6d 6f 64 75 6c 65 73 20 64 69 72 ... 112 more bytes>,
  stderr: <Buffer >
}
```

### Expected Behavior

no error

### To Reproduce

Try the same command I suppose? `npx @turbo/codemod@latest update --force`

tknickman:
Thanks for the report, will take a look. 

What error do you get from running the same command?
`pnpm install turbo@latest --save-dev -w`

We'll make some improvements on that error output as well to be a bit more clear.

hymair:
[...]

PR #5279, opened 2023-06-13
feat(codemod): upgrade refactor

tknickman:
### Description

`@turbo/codemod migrate` refactor

1. Removes the one-off utils that codemod used in favor of our shared utils (`@turbo/utils` and `@turbo/workspaces`)
	- Transforms needed to be made `async` to support this
	- Tests needed to be updated to mock the new utils
1. `migrate` will attempt to shutdown the daemon before upgrading
1. `migrate` now uses the correct command pnpm (`add` instead of `install` ([docs](https://pnpm.io/cli/add)))

This also moves / refactors a few other utils. 
1. The return type of `getAvailablePackageManagers` has been changed from:
	- `Record<PackageManager, { available: boolean, version: string }>`
	to
	- `Record<PackageManager, string | undefined>`

	The previous version stored redundant info and caused TS headaches. 
1. The `getPackageManagersBinPaths` has been moved to `@turbo/utils` for sharing between packages

Fixes https://github.com/vercel/turbo/issues/4009


tknickman (commented):
If you want to test manually. Run:

`turbo build` in the packages/turbo-codemod` directory.

And then you can execute `./dist/cli.js` directly. 

Then you can create a new repo with create-turbo, move the version of turbo back manually, and then execute the local codemod against it. 

I have also been `pnpm pack`ing this, and mounting it inside a docker container with different combinations of package managers installed to make sure it works as expected. 

For example:
```
docker run -it -v /Users/knickman/Developer/vercel/development/turbo/packages/turbo-codemod/turbo-codemod-1.10.4-canary.0.tgz:/home/turbo-codemod-1.10.4-canary.0.tgz node /bin/bash
```

chandlervdw:
@tknickman will this support using `yarn`? I can run `npx @turbo/codemod update` just fine but I still get the upgrade warning. Is there an available `yarn dlx ...` command?

packages/turbo-codemod/src/commands/migrate/steps/getCurrentVersion.ts
```diff
@@ -15,31 +14,23 @@ function getCurrentVersion(
...
+  if (packageManager === "yarn") {
+    return exec(`yarn turbo --version`, { cwd: project.paths.root });
+  }
+  if (packageManager === "pnpm") {
+    return exec(`pnpm turbo --version`, { cwd: project.paths.root });
   }
 
-  // unable to determine local version,
-  return undefined;
+  return exec(`npm exec -c 'turbo --version'`, { cwd: project.paths.root });
```
arlyon, line 33: I would prefer we contained this in an if statement and errored if package manager was not one of the three known cases (or returned undefined).

arlyon (approved):

PR #5282, opened 2023-06-13
chore(build): break up cli#build

mehulkar:
- to see how long each part of the process takes
- more granular caching for each part of the process

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -5,6 +5,9 @@
   "scripts": {
     "clean": "make clean",
     "build": "make",
+    "build:go": "make go-turbo",
+    "build:rust": "cargo build -p turbo",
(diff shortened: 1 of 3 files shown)


cli/package.json
```diff
@@ -5,6 +5,9 @@
   "scripts": {
     "clean": "make clean",
     "build": "make",
+    "build:go": "make go-turbo",
+    "build:rust": "cargo build -p turbo",
+    "build:go-copy": "echo $PWD && ls ../target/debug && mv go-turbo ../target/debug/go-turbo",
```
mehulkar, line 10: We should not merge this as is. It does not include Windows handling or debug/release profile handling

mehulkar:
Just a proof of concept, no need to actually do this right now, it's better to focus on rust port. closing down

PR #5284, opened 2023-06-13
fix(spaces): read logs correctly when invoked from subdirectory

mehulkar:
When running `turbo <task>` from a sub directory, the `logFile` is not sufficient to actual read the logs, because it is a relative path. We need to anchor it from the root of the repo.

One way this comes up is for monorepos on Vercel, where the Root Directory is set, which causes the os.read to fail and no logs are sent to spaces.

### DIFF
--- a/cli/internal/graph/graph.go
+++ b/cli/internal/graph/graph.go
@@ -239,8 +239,7 @@ func (g *CompleteGraph) GetPackageJSONFromWorkspace(workspaceName string) (*fs.P
 // taskLogFile returns the path to the log file for this task execution as an absolute path
 func taskLogFile(root turbopath.AbsoluteSystemPath, dir turbopath.AnchoredSystemPath, taskName string) turbopath.AbsoluteSystemPath {
 	pkgDir := dir.RestoreAnchor(root)
-	filename := fmt.Sprintf("turbo-%v.log", taskName)
-	return pkgDir.Join(".turbo", turbopath.RelativeSystemPath(filename))
+	return pkgDir.UntypedJoin(".turbo", fmt.Sprintf("turbo-%v.log", taskName))
 }
 
 // getTaskGraphAncestors gets all the ancestors for a given task in the graph.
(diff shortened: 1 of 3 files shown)


cli/internal/runsummary/task_summary.go
```diff
@@ -80,9 +81,9 @@ type TaskSummary struct {
 	Execution              *TaskExecutionSummary                 `json:"execution,omitempty"` // omit when it's not set
 }
 
-// GetLogs reads the Logfile and returns the data
+// GetLogs reads the log file and returns the data
 func (ts *TaskSummary) GetLogs() []byte {
-	bytes, err := os.ReadFile(ts.LogFile)
+	bytes, err := os.ReadFile(ts.LogFileAbsolutePath.ToString())
```
gsoltis, line 86: `AbsoluteSystemPath` has a `ReadFile` method so you don't need to convert to a string here

cli/internal/graph/graph.go
```diff
@@ -229,10 +236,11 @@ func (g *CompleteGraph) GetPackageJSONFromWorkspace(workspaceName string) (*fs.P
...
 
-// repoRelativeLogFile returns the path to the log file for this task execution as a
-// relative path from the root of the monorepo.
-func repoRelativeLogFile(dir turbopath.AnchoredSystemPath, taskName string) string {
-	return filepath.Join(dir.ToStringDuringMigration(), ".turbo", fmt.Sprintf("turbo-%v.log", taskName))
+// taskLogFile returns the path to the log file for this task execution as an absolute path
+func taskLogFile(root turbopath.AbsoluteSystemPath, dir turbopath.AnchoredSystemPath, taskName string) turbopath.AbsoluteSystemPath {
+	pkgDir := dir.RestoreAnchor(root)
+	filename := fmt.Sprintf("turbo-%v.log", taskName)
+	return pkgDir.Join(".turbo", turbopath.RelativeSystemPath(filename))
```
gsoltis, line 243: `UntypedJoin` is probably a better choice here since the two segments are clearly from the immediately preceding code.

gsoltis (approved):

PR #5294, opened 2023-06-14
fix(resolve): do not recursively apply alias mapping

kwonoj:
### Description

Attempt to fix WEB-1178, when there's an alias like ` "*" : [ "somepath/* "]` creates infinite recursion.

### DIFF
--- a/crates/turbopack-core/src/resolve/mod.rs
+++ b/crates/turbopack-core/src/resolve/mod.rs
@@ -731,8 +731,16 @@ async fn resolve_internal(
         let result_ref = import_map.lookup(context, request).await?;
         let result = &*result_ref;
         if !matches!(result, ImportMapResult::NoEntry) {
+            // Prevent alias recursively applied infinitely, i.e
+            // paths: { "*": ["node_modules/*"] }
+            let resolve_op = ResolveOptions {
+                import_map: None,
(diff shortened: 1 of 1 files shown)


jridgewell (commented):
Can we verify this against front or similar app?

kwonoj:
I tried to use jsconfig test app which have some known patterns of alias. Any specific pattern would like to test more?

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -731,8 +731,16 @@ async fn resolve_internal(
         let result_ref = import_map.lookup(context, request).await?;
         let result = &*result_ref;
         if !matches!(result, ImportMapResult::NoEntry) {
+            // Prevent alias recursively applied infinitely, i.e
+            // paths: { "*": ["node_modules/*"] }
+            let resolve_op = ResolveOptions {
+                import_map: None,
```
sokra, line 737: This should also prevent transitive aliases, which should be fine.

```
paths: {
  "a": ["b"],
  "b": ["c"]
}

kwonoj: ugh. any suggestions?

kwonoj:
We'll need other approach probably.

PR #5297, opened 2023-06-14
refactor(turborepo): cli::run and args processing

NicholasLYang:
### Description

The `cli::run` function is pretty confusing and had a bug where `--skip-infer` would accidentally pass a relative path to the Go code and cause an error. I refactored this by moving all of the mutation to the end of the function so we can validate the different code paths a little easier. In the future we should do a scheme where `Args` is parsed into a separate struct with all of the fields properly validated and normalized.

This also moves `LogTag` so that if we encounter an error with constructing a `CommandBase`, we error instead of segfaulting.


chris-olszewski (changes requested):
It looks like this broke root interference. Would it be possible to add a unit test to prevent a regression in the `--cwd` behavior?

chris-olszewski (approved):

PR #5030, opened 2023-05-19
chore(docs): Update turbo.build to use Node 18

mehulkar:
- Node 18 is the LTS version
- [Node 16 reaches EOL in September](https://github.com/nodejs/release#release-schedule)
- Node 18 became available on Vercel [in November](https://vercel.com/changelog/node-js-18-lts-is-now-available)
- [ ] update the setting in [vercel project settings](https://vercel.com/vercel/turbo-site/settings)

PR #5300, opened 2023-06-15
docs(examples): update examples and docs to use node 18

mehulkar:
There are no code changes here, so it should be a pretty easy ✅


mehulkar:
This was previously approved in https://github.com/vercel/turbo/pull/5030 (before I reduced that PR to just updating docs to actually build with node 18), so I'm going to merge this when CI passes!

PR #5301, opened 2023-06-15
chore(ci): use node 18

mehulkar:


mehulkar:
This was approved in https://github.com/vercel/turbo/pull/5030, so I'm going to merge when it passes

mehulkar:
`Turbopack Rust testing on Windows` job failed, but it's not due to this PR (@kwonoj confirmed that my understanding is correct). 

The Node version is set to 16 here: https://github.com/vercel/turbo/blob/6b0614feea8b08c4da708c19364abf13f5b9ca41/.github/workflows/test.yml#L764-L785. 

Also, since this particular job is part of a matrix, and the other job (`Turbopack Rust testing on macos) passed, I feel confident that it is not related to the Node version

PR #4637, opened 2023-04-19
Move e2e tests to their own workspace

mehulkar:
Following #4617 and #4630, this PR moves e2e tests into turborepo-tests. This lets us run `turbo test` to pick these up instead of having a bespoke command.

PR #5302, opened 2023-06-15
ci(e2e): turn on remote caching for e2e tests

mehulkar:
This should get us one step closer to running all our tests in a single job and allow turbo caching to save us time, rather than skipping with github ignore patterns

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -183,7 +183,7 @@ jobs:
       turborepo_rust: ${{ steps.ci.outputs.diff != '' || steps.turborepo_rust.outputs.diff != '' }}
       turbopack_bench: ${{ steps.ci.outputs.diff != '' || steps.turbopack_bench.outputs.diff != '' }}
       go: ${{ steps.ci.outputs.diff != '' || steps.turborepo_go.outputs.diff != '' }}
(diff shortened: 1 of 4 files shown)


cli/turbo.json
```diff
@@ -20,13 +20,6 @@
         "../crates/turborepo*/Cargo.toml",
         "!../crates/**/target"
       ]
-    },
-    "e2e": {
-      "outputs": [],
-      "inputs": ["**/*.go", "go.mod", "go.sum", "scripts/e2e/e2e.ts"]
-    },
-    "e2e-prebuilt": {
-      "inputs": ["**/*.go", "go.mod", "go.sum", "scripts/e2e/e2e.ts"]
```
mehulkar, line 29: Unused since #4637

.github/workflows/test.yml
```diff
@@ -304,7 +304,7 @@ jobs:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
 
       - name: E2E Tests
-        run: turbo run test --filter=turborepo-tests-e2e
+        run: turbo run test --filter=turborepo-tests-e2e --remote-only --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
```
mehulkar, line 307: similar to how example tests do it

NicholasLYang (approved):

PR #5309, opened 2023-06-15
feat(turborepo): Hash with libgit2

gsoltis:
### Description

 - Use libgit2 rather than `git hash-object` to hash files

### Testing Instructions

Existing test suite

### DIFF
--- a/crates/turborepo-paths/src/relative_unix_path.rs
+++ b/crates/turborepo-paths/src/relative_unix_path.rs
@@ -19,7 +19,7 @@ impl RelativeUnixPath {
         Ok(unsafe { &*(path as *const BStr as *const Self) })
     }
 
-    pub(crate) fn to_system_path_buf(&self) -> Result<PathBuf, PathError> {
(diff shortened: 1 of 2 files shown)


crates/turborepo-paths/src/relative_unix_path.rs
```diff
@@ -19,7 +19,7 @@ impl RelativeUnixPath {
         Ok(unsafe { &*(path as *const BStr as *const Self) })
     }
 
-    pub(crate) fn to_system_path_buf(&self) -> Result<PathBuf, PathError> {
+    pub fn to_system_path_buf(&self) -> Result<PathBuf, PathError> {
```
NicholasLYang, line 22: Where are we using this?

NicholasLYang (approved):

crates/turborepo-paths/src/relative_unix_path.rs
gsoltis: Good catch. An earlier version was using it.

PR #5311, opened 2023-06-15
test(filetrace): skip failing tests

kwonoj:
### Description

These tests fails 100% on the windows CI, disabling until we fix it.

### DIFF
--- a/crates/turbopack/tests/node-file-trace.rs
+++ b/crates/turbopack/tests/node-file-trace.rs
@@ -206,9 +206,10 @@ static ALLOC: turbo_tasks_malloc::TurboMalloc = turbo_tasks_malloc::TurboMalloc;
 )]
 #[cfg_attr(
     not(feature = "bench_against_node_nft"),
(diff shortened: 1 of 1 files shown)


jridgewell (approved):

crates/turbopack/tests/node-file-trace.rs
```diff
@@ -244,6 +245,13 @@ static ALLOC: turbo_tasks_malloc::TurboMalloc = turbo_tasks_malloc::TurboMalloc;
             .expected_stderr("Cannot find module 'sub/@' or its corresponding type declarations")
     ),
 )]
+#[cfg_attr(
+    //[TODO]: WEB-1188 reenable windows once fixed.
+    not(all(feature = "bench_against_node_nft", target_os = "windows")),
```
sokra, line 250:
```suggestion
    not(any(feature = "bench_against_node_nft", target_os = "windows")),
```

this should be any, shouldn't it?

sokra (changes requested):

PR #5312, opened 2023-06-15
feat(turborepo): A few cleanup nits around log grouping

gsoltis:
### Description

 - iterate over channel rather than relying on a count
 - close the channel to exit output loop

### Testing Instructions

Existing test suite


mehulkar (approved):
Looks good. Was the missing closing of a channel a bug? or Was it not necessary because of looping over the task count?

gsoltis:
Not a bug previously due to keeping count, but required if you want to loop over the channel instead.

PR #5315, opened 2023-06-15
ci: replace sponge usage with jq

mehulkar:
Not _required_ per se, but submitting as a pre-requisite to #5314, one less dependency is good!


scripts/check-examples.sh
mehulkar, line 1: deleted this script instead of replacing sponge usage here, because i don't think anyone uses this anymore

chris-olszewski (approved):

PR #5323, opened 2023-06-16
feat(turborepo): Add context to glob errors

gsoltis:
### Description

I ran into an obscure glob error doing some work on another feature, so I'd like to plumb through the source glob. I expect we'll eventually do this in a more standardized way, but while we're in an ffi world it's kinda ad hoc.

 - fixes an error logging bug in Go
 - adds source globs to globwalking errors

### Testing Instructions

 - updated integration tests to remove baked-in bug

### DIFF
--- a/crates/turborepo-globwalk/src/lib.rs
+++ b/crates/turborepo-globwalk/src/lib.rs
@@ -45,8 +45,8 @@ pub use walkdir::Error as WalkDirError;
 #[derive(Debug, thiserror::Error)]
 pub enum WalkError {
     // note: wax 0.5 has a lifetime in the BuildError, so we can't use it here
-    #[error("bad pattern: {0}")]
-    BadPattern(#[from] BuildError),
+    #[error("bad pattern {0}: {1}")]
+    BadPattern(String, BuildError),
     #[error("invalid path")]
     InvalidPath,
     #[error("walk error: {0}")]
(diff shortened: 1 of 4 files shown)


crates/turborepo-globwalk/src/lib.rs
```diff
@@ -284,6 +282,23 @@ fn collapse_path(path: &str) -> Option<(Cow<str>, usize)> {
...
+pub(crate) fn any_with_contextual_error(
+    precompiled: Vec<Glob<'static>>,
+    text: Vec<String>,
+) -> Result<wax::Any<'static>, WalkError> {
+    wax::any(precompiled).map_err(|e| {
+        let text = text.iter().join(",");
+        WalkError::BadPattern(text, e)
+    })
+}
+
```
chris-olszewski, line 301: So this is totally up to you, but if you wanted to avoid these helper functions here's a way to achieve something like anyhow's `with_context` method:
```
trait GlobContext<T>: Sized {
    fn result(self) -> Result<T, BuildError>;

    fn with_context<F>(self, f: F) -> Result<T, WalkError>
    where F: FnOnce() -> String
    {
        self.result().map_err(|e| WalkError::BadPattern(f(), e))
    }
}

impl<T> GlobContext<T> for Result<T, BuildError> {
    fn result(self) -> Result<T, BuildError> {
        self
    }
}
```

You can then replace the uses of `any_with_contextual_error(pattern, text)` with `wax::any(pattern).with_context(|| text.join(","))` (as long as `GlobContext` is within scope)

`glob_with_contextual_error` uses go back to `Glob::new(raw).with_context(|| raw.to_string())`

chris-olszewski (approved):

crates/turborepo-globwalk/src/lib.rs
gsoltis: I'm inclined to leave as-is and then do this for real when we're not in FFI-land. Ideally this would tie all the way back to the `turbo.json` line where the glob came from (or workspace glob or whatever). I understand people use [miette](https://docs.rs/miette/latest/miette/) for this kind of thing.

PR #5324, opened 2023-06-16
ci: use --env-mode=strict for tests that invoke turbo

mehulkar:
Description
-----------

integration, e2e, and examples tests invoke their own instances of turbo internally. By using `--env-mode=strict`, we can ensure that the environment variables set for _our_ top level turbo does not affect the internal turbo invocation.

Testing Instructions
-------------------

Verify that CI still passes


turbo.json
```diff
@@ -1,5 +1,19 @@
...
+    "LOCALAPPDATA",
+    "TMPDIR",
+    "HOME",
+    "TMP",
+    "GOROOT",
+    "USERPROFILE"
+  ],
+  // This is probably overkill, but we do things per-platform often enough that we should segment
+  // everything by OS
+  "globalEnv": ["OS"],
```
mehulkar, line 16: did you mean`RUNNER_OS` here? We have that set in some tasks also)
https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

mehulkar (commented):
🚢

turbo.json
gsoltis: Hmm. I meant OS, since some tools read that, but we might actually want both. That will segment the cache between "runs on GH Actions" and running anywhere else.

gsoltis:
Test failures at least partially due to https://github.com/vercel/turbo/pull/5386

gsoltis:
Pretty sure the Turbopack test failure is unrelated

PR #5326, opened 2023-06-16
ci(integration): add colors to turbo run invocation

mehulkar:
We can add --color flag for the turbo invocation. We previously did not do this, because we were doing it with an environment variable and it was filtering down to the tests themselves. but using a flag does not do that.


anthonyshew (approved):

mehulkar:
Merging without waiting for benchmarks to finish

PR #5328, opened 2023-06-17
fix: image optimization typo

iojcde:
fix typo: Ecoding => Encoding

### Description

Fixed error description of uncompiled image formats

&#x20; I ran into this error while trying to load the an image in the GIF format, which doesn't seem to be supported yet.&#x20;

### Testing Instructions

Load an image with next/image with an unsupported format, such as GIF


arlyon:
Thank you!

PR #5318, opened 2023-06-15
feat(logging): group logging in github by task id

mehulkar:
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines

PR #5329, opened 2023-06-17
fix(turbo): print help message when turbo called with flags/env vars but no commands

mehulkar:
This was a kind of random bug that was getting in the way of tests passing in https://github.com/vercel/turbo/pull/5318

### DIFF
--- a/crates/turborepo-lib/src/cli.rs
+++ b/crates/turborepo-lib/src/cli.rs
@@ -588,6 +588,14 @@ pub async fn run(
         command
     } else {
         let run_args = mem::take(&mut cli_args.run_args).ok_or(anyhow!("No command specified"))?;
+        if run_args.tasks.is_empty() {
+            let mut cmd = <Args as CommandFactory>::command();
+            match cmd.print_help() {
+                Ok(_) => process::exit(1),
+                Err(_) => process::exit(1),
+            }
+        }
+
         Command::Run(Box::new(run_args))
     };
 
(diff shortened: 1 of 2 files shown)


turborepo-tests/integration/tests/no_args.t
```diff
@@ -65,6 +65,12 @@ Make sure exit code is 2 when no args are passed
...
   [1]
+
   $ ${TURBO} run
   ERROR at least one task must be specified
   [1]
+
+Run again with an environment variable that corresponds to a run argument and assert that
+we get the full help output.
+  $ TURBO_LOG_ORDER=stream ${TURBO} 2>&1 | head -n1
+  The build system that makes ship happen
```
mehulkar, line 76: Currently, I get the same error as if I had run `${TURBO} run`

crates/turborepo-lib/src/cli.rs
```diff
@@ -588,6 +588,14 @@ pub async fn run(
         command
     } else {
         let run_args = mem::take(&mut cli_args.run_args).ok_or(anyhow!("No command specified"))?;
+        if run_args.tasks.is_empty() {
+            let mut cmd = <Args as CommandFactory>::command();
+            match cmd.print_help() {
+                Ok(_) => process::exit(1),
+                Err(_) => process::exit(1),
+            }
```
workingjubilee, line 596: Since you are just ignoring this and dying regardless, you can write this as
```suggestion
            let _ = cmd.print_help();
            process::exit(1);
```

Alternatively, if you really like the current match style, use an "or" pattern:
```suggestion
            match cmd.print_help() {
                Ok(_) | Err(_) => process::exit(1),
            }
```

arlyon (approved):
LGMT!

PR #5339, opened 2023-06-20
fix(ci): Turbopack Clippy CI

NicholasLYang:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -519,11 +519,11 @@ jobs:
 
       - name: Run cargo check release
         run: |
-          cargo groups check turbopack --features rustls-tls --release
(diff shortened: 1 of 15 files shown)


.github/workflows/test.yml
```diff
@@ -491,11 +491,11 @@ jobs:
 
       - name: Run cargo check
         run: |
-          RUSTFLAGS="-D warnings" cargo groups check turborepo-libraries --features rustls-tls
+          cargo groups check turborepo-libraries --features rustls-tls
```
sokra, line 495: What's the reason for that change?

alexkirsz: cc @NicholasLYang

PR #5343, opened 2023-06-20
Turbopack: Don't decode url-encoded characters in request paths

wbinnssmith:
This prevents the Turbopack dev server from decoding characters like %2F in request paths when handling requests. Next.js expects this not to happen when handling routes, for example.

Test Plan: CI


kwonoj (approved):

jridgewell (changes requested):
All remaining uses of `urlencoding` in turbopack can be removed. I checked, and they're all to handle the fact that we decode here.

sokra (approved):

PR #5345, opened 2023-06-21
feat: port package graph

chris-olszewski:
### Description

Adds a rust port of the package graph contained in `context.go`

This PR is primarily concerned with the building of the package graph, more getter methods and graph traversal logic will need to be added for for prune/run.

Reviewer notes:
 - The package graph is configured by the `PackageGraphBuilder` object. The goal of this object is to allow for the creation of package graph without touching the file system
 - The builder gets converted to a `BuildState<T>` where `T` goes through `ResolvedPackageManager`, `ResolvedWorkspaces`, and `ResolvedLockfile`. The goal of these separate types is to avoid using methods out of order.
 - As of writing this doesn't yet hook up lockfile parsing to the package graph construction

### Testing Instructions
Added testing for the primary parts of the graph construction where things can go wrong:
 - duplicate workspace names
 - connecting internal dependencies
 - traversing the lockfile

There can be more tests written around lockfile construction as we already have surpassed the amount of unit tests on the Go side.


arlyon (approved):
No complaints, will probably end up re-scrutinizing this when something actually starts using it

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -221,11 +221,54 @@ impl AbsoluteSystemPath {
     pub fn components(&self) -> Utf8Components<'_> {
         self.0.components()
     }
+
+    pub fn collapse(&self) -> AbsoluteSystemPathBuf {
```
gsoltis, line 225: Do we actually need this? I thought we were keeping these paths "clean" already...

gsoltis (approved):

crates/turborepo-paths/src/absolute_system_path.rs
NicholasLYang: Not everywhere apparently. And specifically with constructing an `AbsoluteSystemPath`, you can't clean on construction because that might require allocation

NicholasLYang: But I don't know whether we need this specifically versus just using path-clean

gsoltis: > Not everywhere apparently.
That sounds like a bug. Can you file a ticket for it?

PR #5350, opened 2023-06-21
feat(turborepo) Add timeout to api client json methods

gsoltis:
### Description

 - Add a timeout param to JSONPost and JSONPatch
 - Set a default timeout of 10 seconds for analytics and spaces

### Testing Instructions

 - Add a test for request timeout


mknichel (approved):
Spaces code looks good. The 10 second timeout should be sufficient.

arlyon (approved):

PR #5354, opened 2023-06-21
feature(turborepo): Turbo Info Command

NicholasLYang:
### Description

Wrote a simple turbo info command. It's hidden because it's not meant for public use just yet (gotta write a proposal and do some devex work before that). Mostly meant to exercise and test package graph

### Testing Instructions

Added some integration tests for `turbo info`

### DIFF
--- a/crates/turborepo-lib/src/package_graph/builder.rs
+++ b/crates/turborepo-lib/src/package_graph/builder.rs
@@ -329,12 +329,13 @@ impl<'a> BuildState<'a, ResolvedWorkspaces> {
 
         let lockfile = match self.populate_lockfile() {
             Ok(lockfile) => Some(lockfile),
-            Err(e) => {
-                warn!(
-                    "Issues occurred when constructing package graph. Turbo will function, but \
-                     some features may not be available: {}",
-                    e
-                );
+            Err(_) => {
+                // TODO: Re-enable this warning once we have lockfile parsing hooked up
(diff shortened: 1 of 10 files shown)


crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -63,7 +69,7 @@ impl PackageGraph {
     }
 
     pub fn len(&self) -> usize {
-        self.workspace_graph.node_count()
+        self.workspaces.len()
```
gsoltis, line 72: This might be worth a comment as to whether or not the root is included

crates/turborepo-lib/src/package_graph/builder.rs
```diff
@@ -329,12 +329,13 @@ impl<'a> BuildState<'a, ResolvedWorkspaces> {
...
         let lockfile = match self.populate_lockfile() {
             Ok(lockfile) => Some(lockfile),
-            Err(e) => {
-                warn!(
-                    "Issues occurred when constructing package graph. Turbo will function, but \
-                     some features may not be available: {}",
-                    e
-                );
+            Err(_) => {
+                // TODO: Re-enable this warning once we have lockfile parsing hooked up
```
gsoltis, line 333: What problem was this causing?

gsoltis (approved):

PR #5360, opened 2023-06-22
Refactor ContentSources to RouteTree for more efficient routing

sokra:
### Description

Routing is inefficient with many pages

next.js PR: https://github.com/vercel/next.js/pull/51660


crates/turbopack-cli/src/dev/turbo_tasks_viz.rs
```diff
@@ -30,12 +32,51 @@ const INVALIDATION_INTERVAL: Duration = Duration::from_secs(3);
 
 #[turbo_tasks::value_impl]
 impl ContentSource for TurboTasksSource {
+    #[turbo_tasks::function]
+    fn get_routes(self_vc: TurboTasksSourceVc) -> RouteTreeVc {
+        RouteTreesVc::cell(vec![
```
alexkirsz, line 37: Do we need a `RouteTreesVc`? Could we do with just `RouteTreeVc::merge` directly instead?

crates/turbopack-convert-trace/src/main.rs
```diff
@@ -355,6 +353,9 @@ fn main() {
         eprint!("Emitting span tree...");
 
         let get_concurrency = |range: Range<u64>| {
+            if range.end <= range.start {
+                return 100;
```
alexkirsz, line 357: Move that value to a constant.

crates/turbopack-dev-server/src/http.rs
```diff
@@ -43,7 +43,10 @@ async fn get_from_source(
     source: ContentSourceVc,
     request: TransientInstance<SourceRequest>,
 ) -> Result<GetFromSourceResultVc> {
-    Ok(match &*resolve_source_request(source, request).await? {
+    Ok(match &*resolve_source_request(source, request)
+        .strongly_consistent()
```
alexkirsz, line 47: Can you add a comment that explains why we want this to be strongly consistent?

crates/turbopack-dev-server/src/source/asset_graph.rs
```diff
@@ -148,28 +150,69 @@ async fn expand(
     Ok(map)
 }
 
+#[turbo_tasks::value(serialization = "auto_for_input")]
+#[derive(Hash, PartialOrd, Ord, Debug, Clone)]
+struct UnresolvedAsset(AssetVc);
```
alexkirsz, line 155:
```suggestion
#[derive(Hash, PartialOrd, Ord, Debug, Clone, TaskInput)]
struct UnresolvedAsset(AssetVc);
```

Then you don't need to wrap it in a `Value`.

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -545,20 +491,34 @@ impl RewriteBuilder {
...
         }
     }
 
-    /// Sets the [ContentSource] from which to restart the lookup process.
-    /// Without a source, the lookup will restart from the original root
-    /// ContentSource.
-    pub fn content_source(mut self, source: ContentSourceVc) -> Self {
-        self.rewrite.source = Some(source);
-        self
+    pub fn new_source(source: ContentSourceVc, path_and_query: String) -> Self {
```
alexkirsz, line 501:
```suggestion
    pub fn new_source_with_path_and_query(source: ContentSourceVc, path_and_query: String) -> Self {
```

crates/turbopack-dev-server/src/source/route_tree.rs
```diff
@@ -0,0 +1,372 @@
...
+        let mut merged = iter.next().unwrap().clone();
+        merged.flat_merge(iter).await?;
+
+        Ok(merged.cell())
+    }
+}
+
+#[turbo_tasks::value]
+#[derive(Default, Clone, Debug)]
+pub struct RouteTree {
```
alexkirsz, line 91: Add comments to all public items in this file.

alexkirsz, line 91: I there a point to making all these `pub`?

crates/turbopack-dev-server/src/source/route_tree.rs
alexkirsz, line 1: Add comments to all public items in this file.

crates/turbopack-dev-server/src/source/route_tree.rs
```diff
@@ -0,0 +1,372 @@
...
+            return Ok(*trees.iter().next().unwrap());
+        }
+
+        // Find common base
+        let mut tree_values = trees.iter().try_join().await?;
+        let mut common_base = 0;
+        let last_tree = tree_values.pop().unwrap();
+        'outer: while common_base < last_tree.base.len() {
+            for tree in tree_values.iter() {
+                if tree.base.len() <= common_base {
```
alexkirsz, line 55: Shouldn't the `<` case be an invariant at this point? Maybe a `debug_assert`?

crates/turbopack-dev-server/src/source/route_tree.rs
```diff
@@ -0,0 +1,372 @@
...
+                    break 'outer;
+                }
+                if tree.base[common_base] != last_tree.base[common_base] {
+                    break 'outer;
+                }
+            }
+            common_base += 1;
+        }
+        tree_values.push(last_tree);
```
alexkirsz, line 64: Did we need to pop it in the first place?

crates/turbopack-node/src/render/node_api_source.rs
```diff
@@ -164,7 +149,7 @@ impl GetContentSourceContent for NodeApiGetContentResult {
                 original_url: original_url.clone(),
                 raw_query: raw_query.clone(),
                 raw_headers: raw_headers.clone(),
-                path: format!("/{}", self.path),
+                path: format!("/{}", 4),
```
alexkirsz, line 152: What happened here?

alexkirsz (approved):

crates/turbopack-cli/src/dev/turbo_tasks_viz.rs
sokra: Yes, we need that. I had `RouteTreeVc::merge` before, but this has weird effects.

`Vec<RouteTreeVc>` as argument creates a new instance for every different list of RouteTrees.
`RouteTreesVc::cell(...).merge()` stores the list of routes in a cell and the merge function is always the same instance, so it the resulting RouteTreeVc.
So "celling" the Vec of RouteTrees can reuse the tasks and might end the merging early when the resulting RouteTree is equal (flat equal!).

alexkirsz: Can you add a comment to the `merge` function to explain this?

Since this is a more turbo-tasks specific pattern, it would be helpful to add this case to some "recommended patterns" document.

crates/turbopack-dev-server/src/source/asset_graph.rs
sokra: Sadly that doesn't work, as TaskInput would resolve the contained RawVc, which we don't want.

I would love to have the ability to add `#[turbo_tasks::unresolved]` to arguments (and properties within TaskInput).

crates/turbopack-dev-server/src/source/route_tree.rs
sokra: Technically not, but without that we would need to skip the last element of an iterator, which has more code.

crates/turbopack-node/src/render/node_api_source.rs
sokra: ups

crates/turbopack-dev-server/src/source/asset_graph.rs
alexkirsz: Can you add a comment that explains this?

crates/turbopack-node/src/source_map/content_source.rs
```diff
@@ -74,30 +89,23 @@ impl ContentSource for NextSourceMapTraceContentSource {
...
+            line,
+            column,
+            frame.name.map(|c| c.to_string()),
+        )
+        .into();
+        let sources = sources
+            .await?
+            .iter()
+            .map(|s| WrappedGetContentSourceContentVc::new(*s, processor).into())
+            .collect();
```
jridgewell, line 104: Isn't this considerably more expensive, since we need to iterate a long list of catchalls, dynamics, fallbacks, etc?

crates/turbopack-dev-server/src/source/asset_graph.rs
```diff
@@ -148,28 +150,70 @@ async fn expand(
     Ok(map)
 }
 
+// This must not be a TaskInput since this would resolve the embedded asset
```
jridgewell, line 153: Please explain why we don't want to resolve the embedded asset.

crates/turbopack-dev-server/src/source/resolve.rs
```diff
@@ -83,26 +72,55 @@ pub async fn resolve_source_request(
...
+                                let new_asset_path =
+                                    urlencoding::decode(&new_uri.path()[1..])?.into_owned();
+                                request_overwrites.uri = new_uri;
+                                current_asset_path = new_asset_path;
+                                continue 'routes;
+                            }
+                            RewriteType::ContentSource {
+                                source,
+                                path_and_query,
+                            } => {
```
jridgewell, line 88: I this this rewrite logic creates a lot of inconsistencies. Before we were rewriting the current sources's URL only, but now we're changing the URL for the current route _and all routes that follow_. I think it has to only rewrite the the URL for the current route tree, right?

crates/turbopack-dev-server/src/source/resolve.rs
```diff
@@ -83,26 +72,55 @@ pub async fn resolve_source_request(
...
+                                route_tree =
+                                    source.get_routes().resolve_strongly_consistent().await?;
+                                continue 'routes;
+                            }
+                            RewriteType::Sources {
+                                sources: new_sources,
+                            } => {
+                                sources = *new_sources;
+                                continue 'sources;
+                            }
```
jridgewell, line 103: And this creates similar inconsistencies. There could be more sources after the current source that would give a valid result, but we'll never reach those. I think we need to only `sources = new_sources.concat(remaining_sources)` here

crates/turbopack-dev-server/src/source/router.rs
```diff
@@ -109,14 +78,42 @@ async fn get_introspection_children(
...
-        get(&self.routes, &self.fallback, "", path, data).await
+    async fn get_routes(&self) -> Result<RouteTreeVc> {
+        let prefix = &*self.prefix.await?;
+        if cfg!(debug_assertions) {
+            debug_assert!(prefix.is_empty() || prefix.ends_with('/'));
+            debug_assert!(!prefix.starts_with('/'));
+        }
+        let prefix = prefix
+            .is_empty()
+            .then(|| BaseSegment::from_static_pathname(prefix.as_str()).collect())
```
jridgewell, line 91: This is currently creating a `vec['']` for an empty prefix, and a `vec[]` for a non-empty one.
```suggestion
        let prefix = (!prefix.is_empty())
            .then(|| BaseSegment::from_static_pathname(prefix.as_str()).collect())
```

crates/turbopack-dev-server/src/source/router.rs
```diff
@@ -125,34 +122,54 @@ impl ContentSource for RouterContentSource {
     }
 }
 
+#[turbo_tasks::value]
+struct PrefixedRouterContentSourceMapper {
+    prefix: StringVc,
```
jridgewell, line 127: Seems unused? Unless you meant mean to pass the `prefix` `StringVc` to `PrefixedRouterGetContentSourceContent` instead of the `PrefixedRouterContentSourceMapperVc`?

jridgewell (changes requested):

crates/turbopack-node/src/source_map/content_source.rs
sokra: Yes, that's a problem. Luckily the list is usually short. But we can further optimize that in future. Eagerly returning the full list of `GetContentSourceContentVc`s isn't needed. Some kind of lazy iterator would be better.

crates/turbopack-dev-server/src/source/resolve.rs
sokra: If referring to the next route in `sources`, on `Rewrite` we do no longer apply following routes. The `continue 'routes;` will skip all remaining routes and starts applying a new route tree with the new URL.

The Rewrite is basically counted as route match.

crates/turbopack-dev-server/src/source/resolve.rs
sokra: Same, the Rewrite response is counted as route match and no following source is applied.

e. g. for `/some/path` and `/some/[dynamic]` returned by `get_routes`, once `/some/path` returns a Rewrite, `/some/[dynamic]` will no longer be applied

crates/turbopack-dev-server/src/source/route_tree.rs
```diff
@@ -0,0 +1,386 @@
...
+/// The prefix tree of routes. Also handling dynamic and catch all segments.
+#[turbo_tasks::value]
+#[derive(Default, Clone, Debug)]
+pub struct RouteTree {
+    base: Vec<BaseSegment>,
+    sources: Vec<GetContentSourceContentVc>,
+    static_segments: IndexMap<String, RouteTreeVc>,
+    dynamic_segments: Vec<RouteTreeVc>,
+    catch_all_sources: Vec<GetContentSourceContentVc>,
+    fallback_sources: Vec<GetContentSourceContentVc>,
```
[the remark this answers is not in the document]
sokra: Fallback is currently unused. I just migrated it from `Specificity::Fallback`.

crates/turbopack-dev-server/src/source/static_assets.rs
```diff
@@ -42,31 +44,62 @@ impl StaticAssetsContentSourceVc {
...
+        .flat_map(|(name, entry)| match entry {
+            DirectoryEntry::File(path) | DirectoryEntry::Symlink(path) => {
+                Some(RouteTreeVc::new_route(
+                    vec![BaseSegment::Static(name.clone())],
+                    None,
+                    StaticAssetsContentSourceItemVc::new(*path).into(),
+                ))
+            }
+            DirectoryEntry::Directory(path) => Some(
+                get_routes_from_directory(*path)
```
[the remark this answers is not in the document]
sokra: Yes we should make that more lazy. That's something for a follow-up work.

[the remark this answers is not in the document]
sokra: I leave that for the future. It only take a few 10s of milliseconds, so fine for now I guess

crates/turbopack-dev-server/src/source/mod.rs
```diff
@@ -91,6 +93,14 @@ pub enum ContentSourceContent {
     Rewrite(RewriteVc),
 }
 
+/// This trait can be emitted as collectible and will be applied after the
+/// request is handled and it's ensured that it finishes before the next request
+/// is handled.
```
jridgewell, line 98: Can you explain why you would want to do emit side effect vcs? What does it accomplish?

crates/turbopack-dev-server/src/source/asset_graph.rs
```diff
@@ -150,7 +150,10 @@ async fn expand(
     Ok(map)
 }
 
-// This must not be a TaskInput since this would resolve the embedded asset
+/// A unresolve asset. We need to have a unresolve Asset here as we need to
+/// lookup the Vc identity in the expanded set.
```
jridgewell, line 154: Sorry, I still don't understand. What is it about a resolved asset that prevents vc identity lookup? How does an unresolved asset solve that?

crates/turbopack-dev-server/src/source/route_tree.rs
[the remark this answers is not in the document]
jridgewell: What I'm saying is that it's unnecessary with this scheme, every fallback could be a catchall instead, right?

crates/turbopack-node/src/source_map/content_source.rs
jridgewell: Leave a TODO with linear task, please.

crates/turbopack-dev-server/src/source/static_assets.rs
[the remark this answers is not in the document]
jridgewell: Add a TODO with linear task, please.

jridgewell (approved):

crates/turbopack-dev-server/src/source/mod.rs
sokra: It moves the invalidation to a time after the request is sent to the user. This delays all computation for the state change to some time after this request and before the next request.

If we would just change it inline, the strongly consistent read from the request handling would wait for this invalidation to be completed as it might have effects on the current request (but it doesn't have these).

So for expensive invalidations this allows to respond ealier.

sokra: Technically it's also possible to return a list of SideEffectVcs, but this requires adding a list to all intermediate functions, which is quite a large change. Since we have to cool system of collectibles we can use that. It allows to emit a side effect from anywhere and the server will make sure that it's handled at the right time.

crates/turbopack-dev-server/src/source/static_assets.rs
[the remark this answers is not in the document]
sokra: done

PR #5365, opened 2023-06-22
fix(turborepo): Handle empty workspace tasks

gsoltis:
### Description

 - Don't crash if the given workspace doesn't have any tasks defined in its own `turbo.json`

### Testing Instructions

 - Verified manually on the `pnpm` `kitchen-sink` example. This is a step on the path to getting the examples running properly, which will in turn then validate this change in an ongoing fashion


NicholasLYang (approved):

gsoltis:
Closing in favor of https://github.com/vercel/turbo/pull/5370

PR #5367, opened 2023-06-23
fix(turborepo): Restructure spaces client and test error handling

gsoltis:
### Description

 - Restructure timeouts to more properly thread contexts
 - Refactor JSON posting methods in api client
 - Refactor spaces client to be a bit more testable
 - Refactor spaces client to specially handle the create run call, since it's the only one that needs a response and unlocks the remaining api calls
 - Handle failure to create a run by skipping subsequent requests

### Testing Instructions

 - added a new test for failing to create a run, ensuring that we can still call `Close` on a run summary

### DIFF
--- a/cli/internal/analytics/analytics.go
+++ b/cli/internal/analytics/analytics.go
@@ -26,12 +26,12 @@ type Client interface {
 }
 
 type Sink interface {
-	RecordAnalyticsEvents(events Events, timeout time.Duration) error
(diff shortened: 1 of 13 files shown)


arlyon (approved):

gsoltis:
e2e tests are not yet expected to pass. Verified experimentally that errors don't cause the process to hang

gsoltis:
Force merging due to aforementioned error source

PR #5371, opened 2023-06-23
test(turborepo): Create Cache Tests

NicholasLYang:
### Description

Tests for creating a cache. These do have different hashes than the Go ones, but I think that's fine since there's no expectation that tar construction or zstd compression is necessarily deterministic.


arlyon (approved):
I opened this and realised I had a pending comment which seems to be outdated.

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -0,0 +1,425 @@
...
+        } else {
+            // Throw an error if trying to create a cache that contains a type we don't
+            // support.
+            return Err(CacheError::CreateUnsupportedFileType(Backtrace::capture()));
+        }
+
+        // Consistent creation
+        header.set_uid(0);
+        header.set_gid(0);
+        header.as_gnu_mut().unwrap().set_atime(0);
```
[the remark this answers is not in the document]
arlyon: Please ignore, it is outdates.

crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -178,4 +402,21 @@ mod tests {
...
         Ok(())
     }
+
+    #[test]
+    fn test_compression() -> Result<()> {
+        let mut buffer = Vec::new();
+        let mut encoder = zstd::Encoder::new(&mut buffer, 0)?.auto_finish();
+        encoder.write(b"hello world")?;
+        // Should finish encoding on drop
+        drop(encoder);
```
arlyon, line 412: That feels odd to me but it makes sense. Coming from the hash / digest api I am surprised but I guess since the digest apis spit out data when you finalize rather than (I assume) write it as you go it makes sense to have 'implicit finalize' on drop.

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -33,8 +34,9 @@ impl CacheReader {
 
     pub fn open(path: &AbsoluteSystemPathBuf) -> Result<Self, CacheError> {
         let file = path.open()?;
+        let is_compressed = path.extension() == Some("zst");
```
arlyon, line 37: Probably worth discussing here whether we want to check for the zstd magic number as well / instead / as a fallback? It'll be slower in the general case but then we are not dependent on the file name causing issues

crates/turborepo-cache/src/cache_archive/create.rs
NicholasLYang: Yeah we could require an explicit finalize, but that's too easy to miss imo, so it's probably better to just finalize on drop

crates/turborepo-cache/src/cache_archive/restore.rs
NicholasLYang: I'll defer to @gsoltis on that, but we should probably keep the same Go behavior until we're done with the port

PR #5372, opened 2023-06-23
Restructure spaces client and test error handling

gsoltis:


gsoltis:
We don't need this, we directly use the correct path

PR #5382, opened 2023-06-26
chore: replace  `default_features` with `default-features`

mgrachev:
https://doc.rust-lang.org/cargo/reference/features.html#dependency-features


NicholasLYang (approved):
thanks!

PR #5385, opened 2023-06-26
feat(turborepo): Use Github log prefixes when running on Github Actions

gsoltis:
### Description

 - use grouped logging and github-style prefixes when we detect we're running in github actions

### Testing Instructions

 - new integration test

### DIFF
--- a/cli/internal/run/run.go
+++ b/cli/internal/run/run.go
@@ -129,11 +129,10 @@ func optsFromArgs(args *turbostate.ParsedArgsFromRust) (*Opts, error) {
 }
 
 func configureRun(base *cmdutil.CmdBase, opts *Opts, signalWatcher *signals.Watcher) *run {
-	if opts.runOpts.LogOrder == "auto" {
-		if name := ci.Constant(); name == "GITHUB_ACTIONS" {
-			opts.runOpts.LogOrder = "grouped"
-			opts.runOpts.IsGithubActions = true
-		}
+	if opts.runOpts.LogOrder == "auto" && ci.Constant() == "GITHUB_ACTIONS" {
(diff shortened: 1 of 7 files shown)


cli/internal/run/run.go
```diff
@@ -128,6 +129,12 @@ func optsFromArgs(args *turbostate.ParsedArgsFromRust) (*Opts, error) {
 }
 
 func configureRun(base *cmdutil.CmdBase, opts *Opts, signalWatcher *signals.Watcher) *run {
+	if opts.runOpts.LogOrder == "auto" && ci.Constant() == "GITHUB_ACTIONS" {
+		opts.runOpts.LogOrder = "grouped"
+		opts.runOpts.LogPrefix = "none"
+		opts.runOpts.IsGithubActions = true
+	}
```
chris-olszewski, line 136: This feels like this should happen in the `optsFromArgs` function

chris-olszewski (approved):

boennemann:
Quick feedback here: While this makes a lot of sense, this caused massive head-scratching on our side, because failed tasks do not expand by default.
We couldn't figure out for a while why the CI failed without meaningful output.

mehulkar:
@boennemann can you file an issue for that? I'm not sure, but I _think_ this is a solvable problem since the logs will be buffered (for grouping), so we should be able to _not_ add the Github Actions start/end markers for failed tasks

gsoltis:
@mehulkar @boennemann https://github.com/vercel/turbo/pull/5435 should move the error inside the correct group and tag it, which I think will cause GH Actions to highlight it

ilyasmez:
I've just noticed this new feature, and I love it!!! Thank you @gsoltis!

PR #5387, opened 2023-06-26
Mark Vcs as `must-use`

wbinnssmith:
This uses the `#[must-use]` macro [0] to mark Vcs as needing to be used when created. The compiler creates warnings for each case.

Test Plan: Create a Vc in code and verify a warning is shown when Turbopack is built.

[0] https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute

### DIFF
--- a/crates/turbo-tasks-macros/src/func.rs
+++ b/crates/turbo-tasks-macros/src/func.rs
@@ -212,12 +212,10 @@ pub fn split_signature(sig: &Signature) -> (Signature, Signature, Type, TokenStr
     if is_empty_type(raw_output_type) {
         // Can't emit a diagnostic on OutputType alone as it can be omitted from the
         // text for `()`. Use the whole signature span.
-        sig.span()
-            .unwrap()
-            .error(
-                "Cannot return `()` from a turbo_tasks function. Return Result<NothingVc> instead.",
-            )
-            .emit();
+        abort!(
+            sig.span(),
+            "Cannot return `()` from a turbo_tasks function. Return a NothingVc instead.",
+        )
     }
 
     let inline_ident = get_internal_function_ident(&sig.ident);
--- a/crates/turbo-tasks-macros/src/func.rs
+++ b/crates/turbo-tasks-macros/src/func.rs
@@ -208,6 +208,18 @@ pub fn gen_native_function_code(
 
 pub fn split_signature(sig: &Signature) -> (Signature, Signature, Type, TokenStream2) {
     let output_type = get_return_type(&sig.output);
+    let (raw_output_type, _) = unwrap_result_type(&output_type);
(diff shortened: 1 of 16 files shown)


ForsakenHarmony (approved):

jridgewell (changes requested):
I think we need this on TraitVcs and possibly on TT function return values?
- https://github.com/vercel/turbo/blob/wbinnssmith/3a6d5fce/crates/turbo-tasks-macros/src/value_trait_macro.rs#L247-L250
- https://github.com/vercel/turbo/blob/wbinnssmith/3a6d5fce/crates/turbo-tasks-macros/src/function_macro.rs#L42-L46

crates/turbo-tasks-memory/tests/collectibles.rs
```diff
@@ -124,11 +124,11 @@ async fn my_collecting_function_indirect() -> Result<ThingVc> {
 }
 
 #[turbo_tasks::function]
-fn my_multi_emitting_function() -> ThingVc {
-    my_transitive_emitting_function("", "a");
-    my_transitive_emitting_function("", "b");
+async fn my_multi_emitting_function() -> Result<ThingVc> {
+    my_transitive_emitting_function("", "a").await?;
```
sokra, line 128: Better use `let _ =` for these two

crates/turbopack/src/lib.rs
```diff
@@ -435,7 +435,7 @@ impl AssetContext for ModuleAssetContext {
                 request,
             );
 
-            result.add_reference(types_reference.into());
+            result.add_reference(types_reference.into()).await?;
```
sokra, line 438: This probably returns a new result. Maybe rename the function to `with_reference`

```suggestion
            result = result.add_reference(types_reference.into());
```

crates/turbo-tasks-macros/src/func.rs
```diff
@@ -208,26 +208,31 @@ pub fn gen_native_function_code(
...
 pub fn split_signature(sig: &Signature) -> (Signature, Signature, Type, TokenStream2) {
     let output_type = get_return_type(&sig.output);
+    let (raw_output_type, _) = unwrap_result_type(&output_type);
+    if is_empty_type(raw_output_type) {
+        // Can't emit a diagnostic on OutputType alone as it can be omitted from the
+        // text for `()`. Use the whole signature span.
+        sig.span()
+            .unwrap()
+            .error(
+                "Cannot return `()` from a turbo_tasks function. Return Result<NothingVc> instead.",
```
jridgewell, line 218:
```suggestion
                "Cannot return `()` from a turbo_tasks function. Return a NothingVc instead.",
```

jridgewell (approved):

PR #5388, opened 2023-06-26
Plainer explanation for synthetic tasks.

anthonyshew:
Iterating on the doc for dependencies outside of a task!


arlyon (approved):
Great explanation! :)

PR #5390, opened 2023-06-27
prototype

gsoltis:


gsoltis:
We don't need this, we're going w/ capnproto

PR #5392, opened 2023-06-27
feat(ecma-analyzer): support super() call in class ctor

kwonoj:
### Description

Attempt to close WEB-1089.

PR tries to support import resolustion inside of `super` call, if given args are referencing esm imports.

Since super() is sort of expr_call, I originally tried to reuse JsValue::Call as-is but not successfully. Still that might be a simple mistake and new addition to JsValue::Super may not be required.

### DIFF
--- a/crates/turbopack-ecmascript/src/analyzer/graph.rs
+++ b/crates/turbopack-ecmascript/src/analyzer/graph.rs
@@ -519,6 +519,21 @@ impl EvalContext {
             }
 
             Expr::Call(CallExpr {
+                callee: Callee::Super(_),
+                args,
+                ..
+            }) => {
+                // We currently do not handle spreads.
+                if args.iter().any(|arg| arg.spread.is_some()) {
+                    return JsValue::unknown_empty("spread in function calls is not supported");
+                }
+
+                let args = args.iter().map(|arg| self.eval(&arg.expr)).collect();
+
+                JsValue::super_call(args)
+            }
+
+            Expr::Call(CallExpr {
                 callee: Callee::Import(_),
                 args,
                 ..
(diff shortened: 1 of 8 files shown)


kdy1 (approved):

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -421,6 +421,9 @@ pub enum JsValue {
     /// A function call without a this context.
     /// `(total_node_count, callee, args)`
     Call(usize, Box<JsValue>, Vec<JsValue>),
+    /// A super call to the parent constructor.
+    /// `(total_node_count, args)`
+    Super(usize, Vec<JsValue>),
```
sokra, line 426:
```suggestion
    SuperCall(usize, Vec<JsValue>),
```

crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -898,7 +913,13 @@ impl Analyzer<'_> {
                     });
                 }
             }
-            _ => {}
+            Callee::Super(_) => self.add_effect(Effect::Call {
+                func: JsValue::FreeVar(js_word!("super")),
```
sokra, line 917:
```suggestion
                func: JsValue::Super,
``

super is not a free var. It should have a special token, or stay Unknown

File: crates/turbopack-ecmascript/src/analyzer/mod.rs [hunk 3]
```diff
@@ -811,6 +823,10 @@ impl JsValue {
         Self::Call(1 + f.total_nodes() + total_nodes(&args), f, args)
     }
 
+    pub fn callee_super(args: Vec<JsValue>) -> Self {
```
sokra, line 826:
```suggestion
    pub fn super_call(args: Vec<JsValue>) -> Self {
```

sokra (changes requested):
add a test case

jridgewell (approved):

PR #5395, opened 2023-06-27
New FileSystemPath methods: with_extension and file_stem

alexkirsz:


alexkirsz:
* main
  * **PR #5395** <a href="https://app.graphite.dev/github/pr/vercel/turbo/5395" target="_blank">Graphite</a>  👈
    * **PR #5396** <a href="https://app.graphite.dev/github/pr/vercel/turbo/5396" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/5395?utm_source=stack-comment).

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1089,6 +1089,41 @@ impl FileSystemPathVc {
     pub async fn is_inside_or_equal(self, other: FileSystemPathVc) -> Result<BoolVc> {
         Ok(BoolVc::cell(self.await?.is_inside_or_equal(&*other.await?)))
     }
+
+    #[turbo_tasks::function]
+    pub async fn with_extension(self, extension: &str) -> Result<FileSystemPathVc> {
+        let this = self.await?;
+        let Some((prefix_path, file_name)) = this.path.rsplit_once('/') else {
+            return Ok(self);
```
sokra, line 1097: It should still work when the file is on root.

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1089,6 +1089,41 @@ impl FileSystemPathVc {
...
+        };
+        let Some((file_stem, _old_extension)) = file_name.rsplit_once('.') else {
+            return Ok(self);
```
sokra, line 1100: It should add an extension even if there is none

sokra (changes requested):

crates/turbo-tasks-fs/src/virtual_fs.rs
```diff
@@ -0,0 +1,68 @@
...
+    DirectoryContentVc, FileContentVc, FileMetaVc, FileSystem, FileSystemPathVc, FileSystemVc,
+    LinkContentVc,
+};
+
+#[turbo_tasks::value]
+pub struct VirtualFileSystem;
+
+#[turbo_tasks::value_impl]
+impl VirtualFileSystemVc {
+    #[turbo_tasks::function]
```
sokra, line 14:
```suggestion
```

to force fresh identity on every call of new()

crates/turbo-tasks-fs/src/virtual_fs.rs
sokra, line 1: remove virtual_fs for turbopack_core

sokra (changes requested):

crates/turbo-tasks-fs/src/virtual_fs.rs
jridgewell: Why is that useful?

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1089,6 +1090,62 @@ impl FileSystemPathVc {
...
+    /// * The entire file name if the file name begins with `.` and has no other
+    ///   `.`s within;
+    /// * Otherwise, the portion of the file name before the final `.`
+    #[turbo_tasks::function]
+    pub async fn file_stem(self) -> Result<OptionStringVc> {
+        let this = self.await?;
+        if this.path.is_empty() {
+            return Ok(OptionStringVc::cell(None));
+        }
+        let file_name = if let Some((_prefix_path, file_name)) = this.path.rsplit_once('/') {
```
jridgewell, line 1135: Nit: A lot of methods deal with manipulating the `dirname`, `stem`, and `ext`, and have to rework this `rsplit('.')` (but only if there's no `/` in the extension). Can we rework this to a reusable method that returns a `(dirname, stem, ext)` tuple?

jridgewell (approved):
LGTM after @sokra's comments are taken care of.

crates/turbo-tasks-fs/src/virtual_fs.rs
alexkirsz: We want virtual file systems to have a different identity. For instance, if you use virtual file systems to represent the Node and Client FS in a test, you'd want them to map to different FS instances for safety, as we have some `FileSystemPathVc` methods that check FS identity in order to determine whether a path operation is correct.

crates/turbo-tasks-fs/src/virtual_fs.rs
alexkirsz: I've removed it but kept the export to avoid breaking Next.js when merging this PR. To be removed once the import changes on the Next.js side.

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1089,6 +1127,45 @@ impl FileSystemPathVc {
...
+
+    /// Creates a new [`FileSystemPathVc`] like `self` but with the given
+    /// extension.
+    #[turbo_tasks::function]
+    pub async fn with_extension(self, extension: &str) -> Result<FileSystemPathVc> {
+        let this = self.await?;
+        let path_without_extension = match this.split_extension() {
+            (path_without_extension, Some(_)) => path_without_extension,
+            (path_without_extension, None) => path_without_extension,
+        };
```
sokra, line 1139:
```suggestion
        let (path_without_extension, _) = this.split_extension();
```

sokra (approved):

PR #5396, opened 2023-06-27
Fix rebase error with ESM interop

alexkirsz:

PR #5397, opened 2023-06-27
Add support for CSS client references

alexkirsz:
### Description

This PR changes CSS assets to support the CSS client references use case of Next.js, where .css assets in the RSC graph should be replaced with a client reference assets.

CSS assets are now processed in two steps:
1. Processing the CSS source asset into either a Global or a Module CSS asset.
2. Inside of Global/Module assets, processing the CSS source asset into a CSS asset.

By default, (2.) will create a leaf CSS asset. But in Next.js RSC, this step will be replaced with creating a CSS client reference instead. This client reference can then be detected when walking the RSC graph, and a client CSS chunk can be produced from there.

### Testing Instructions

N/A


alexkirsz:
* main
  * **PR #5395** <a href="https://app.graphite.dev/github/pr/vercel/turbo/5395" target="_blank">Graphite</a> 
    * **PR #5396** <a href="https://app.graphite.dev/github/pr/vercel/turbo/5396" target="_blank">Graphite</a> 
      * **PR #5397** <a href="https://app.graphite.dev/github/pr/vercel/turbo/5397" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/5397?utm_source=stack-comment).

crates/turbopack-css/src/module_asset.rs
```diff
@@ -54,41 +47,47 @@ fn modifier() -> StringVc {
...
     #[turbo_tasks::function]
-    pub fn new(source: AssetVc, context: AssetContextVc, transforms: CssInputTransformsVc) -> Self {
-        Self::cell(ModuleCssModuleAsset {
-            inner: CssModuleAssetVc::new_module(source, context, transforms),
-        })
+    pub async fn new(source: AssetVc, context: AssetContextVc) -> Result<Self> {
+        let inner = context.process(
+            source,
+            Value::new(ReferenceType::Css(CssReferenceSubType::Internal)),
+        );
```
sokra, line 62: store the `source` and add a `inner()` method doing that that lazily

crates/turbopack-css/src/module_asset.rs
```diff
@@ -54,41 +47,47 @@ fn modifier() -> StringVc {
...
     }
 }
 
 #[turbo_tasks::value_impl]
-impl Asset for ModuleCssModuleAsset {
+impl Asset for ModuleCssAsset {
     #[turbo_tasks::function]
     fn ident(&self) -> AssetIdentVc {
-        self.inner.source_ident().with_modifier(modifier())
+        self.inner.ident().with_modifier(modifier())
```
sokra, line 72:
```suggestion
        self.source.ident().with_modifier(modifier())
```

crates/turbopack-css/src/references/internal.rs
```diff
@@ -0,0 +1,46 @@
...
+    chunk::{ChunkableAssetReference, ChunkableAssetReferenceVc},
+    reference::{AssetReference, AssetReferenceVc},
+    resolve::{ResolveResult, ResolveResultVc},
+};
+
+/// A reference to an internal CSS asset.
+#[turbo_tasks::value]
+#[derive(Hash, Debug)]
+pub struct InternalCssAssetReference {
+    asset: AssetVc,
```
sokra, line 14:
```suggestion
    source: AssetVc,
    context: AssetContextVc,
```

This should lazily process the source asset

crates/turbopack/src/module_options/mod.rs
```diff
@@ -249,19 +256,44 @@ impl ModuleOptionsVc {
...
                 .collect(),
             ),
             ModuleRule::new(
-                ModuleRuleCondition::ResourcePathEndsWith(".module.css".to_string()),
-                vec![ModuleRuleEffect::ModuleType(ModuleType::CssModule(
-                    css_transforms,
-                ))],
+                ModuleRuleCondition::all(vec![
+                    ModuleRuleCondition::ResourcePathEndsWith(".module.css".to_string()),
+                    ModuleRuleCondition::not(ModuleRuleCondition::ReferenceType(
```
sokra, line 268: This is pretty weird. We should not confront the user with internal modules by default.

Rules shouldn't match internal references by default. Rule should need to opt-in to matching internal references.

To do that, add `ReferenceType::is_internal()` that returns true for all `Internal` references.

Also add `ModuleRuleCondition::Internal`. When `is_internal()` returns true, a rule will only match if `ModuleRuleCondition::Internal` is specified.

This way it doesn't match rules for internal references by default, only when opt-in to that. That also applies to EcmascriptReferenceType::Internal.

crates/turbopack/src/module_options/module_rule.rs
```diff
@@ -74,8 +73,12 @@ pub enum ModuleType {
         transforms: EcmascriptInputTransformsVc,
         options: MdxTransformOptionsVc,
     },
-    Css(CssInputTransformsVc),
-    CssModule(CssInputTransformsVc),
+    CssGlobal,
+    CssModule,
+    Css {
```
sokra, line 78: Should we rename to `CssInner`

sokra (changes requested):

crates/turbopack-css/src/references/internal.rs
alexkirsz: The asset will already have been processed at that point, as we need to process it to compute other references.

crates/turbopack/src/module_options/module_rule.rs
alexkirsz: I think `CssGlobal` and `CssModule` are explicit enough that `Css` on its own is OK.

sokra (approved):

PR #5401, opened 2023-06-27
refactor(turborepo): Enumerate the things we hash

gsoltis:
### Description

 - no functional change
 - create explicit hash functions for each top-level kind of thing that we hash

### Testing Instructions

 - existing integration tests verify that hashes have not changed

### DIFF
--- a/cli/internal/fs/hash.go
+++ b/cli/internal/fs/hash.go
@@ -7,11 +7,75 @@ import (
 	"io"
 	"strconv"
 
+	"github.com/vercel/turbo/cli/internal/env"
+	"github.com/vercel/turbo/cli/internal/lockfile"
 	"github.com/vercel/turbo/cli/internal/turbopath"
+	"github.com/vercel/turbo/cli/internal/util"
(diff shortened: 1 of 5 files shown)


arlyon (approved):

cli/internal/fs/hash.go
```diff
@@ -7,11 +7,75 @@ import (
...
+
+// FileHashes is a hashable map of files to the hash of their contents
+type FileHashes map[turbopath.AnchoredUnixPath]string
+
+// TaskHashable is a hashable representation of a task to be run
+type TaskHashable struct {
+	GlobalHash           string
+	TaskDependencyHashes []string
+	PackageDir           turbopath.AnchoredUnixPath
+	HashOfFiles          string
```
arlyon, line 28: Other 'hash' has the word at the end, maybe rename?

gsoltis:
@arlyon I'm leaving the naming as-is since I'm just copying from existing structs (and making things public). I'm not intending to refactor these.

PR #5405, opened 2023-06-28
feat(turborepo): refactor scm and thread into scope calculation

gsoltis:
### Description

 - move scm-related functions into a `Git` struct
 - rename `Hasher` to `SCM`
 - use turbopath more throughout
 - create an `SCM` instance in run outline and pass it to scope

### Testing Instructions

 Existing test suite

### DIFF
--- a/crates/turborepo-scm/src/git.rs
+++ b/crates/turborepo-scm/src/git.rs
@@ -198,30 +198,7 @@ pub fn previous_content(
     // exactly
     let absolute_file_path = AbsoluteSystemPathBuf::from_unknown(&git_root, &file_path);
 
-    // let anchored_file_path = if file_path.is_absolute() {
-    //     let absolute_file_path = AbsoluteSystemPathBuf::try_from(file_path)?;
-    //     git_root.anchor(absolute_file_path)?
-    // } else {
-    //     file_path.as_path().try_into()?
-    // };
-
     scm.previous_content(from_commit, &absolute_file_path)
-    // let git_binary = which("git")?;
-    // let mut command = Command::new(git_binary);
-    // let command = command
-    //     .arg("show")
-    //     .arg(format!("{}:{}", from_commit, anchored_file_path.as_str()))
-    //     .current_dir(&git_root);
-
-    // let output = command.output()?;
-    // if output.status.success() {
-    //     Ok(output.stdout)
-    // } else {
-    //     Err(Error::Git(
-    //         String::from_utf8_lossy(&output.stderr).to_string(),
-    //         Backtrace::capture(),
-    //     ))
-    // }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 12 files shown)


crates/turborepo-scm/src/lib.rs
```diff
@@ -106,3 +112,115 @@ pub(crate) fn wait_for_success<R: Read, T>(
...
+
+#[derive(Debug)]
+pub struct Git {
+    root: AbsoluteSystemPathBuf,
+    bin: AbsoluteSystemPathBuf,
+}
+
+impl Git {
+    pub(crate) fn find(path_in_repo: &AbsoluteSystemPath) -> Option<Self> {
+        let bin = which::which("git").ok()?;
```
NicholasLYang, line 124: Should we log in case of errors here? Any case where we're silently turning a Result into an Option should probably be logged.

crates/turborepo-scm/src/git.rs
```diff
@@ -139,51 +186,53 @@ fn reanchor_path_from_git_root_to_turbo_root(
...
-        Err(Error::Git(
-            String::from_utf8_lossy(&output.stderr).to_string(),
-            Backtrace::capture(),
-        ))
-    }
+    // FIXME: this is probably wrong. We should know the path to the lockfile
+    // exactly
+    let absolute_file_path = AbsoluteSystemPathBuf::from_unknown(&git_root, &file_path);
+
+    // let anchored_file_path = if file_path.is_absolute() {
```
NicholasLYang, line 201: Should we delete this instead of commenting out?

crates/turborepo-scm/src/git.rs
```diff
@@ -33,97 +57,120 @@ pub fn changed_files(
...
-        let anchored_to_turbo_root_file_path =
-            reanchor_path_from_git_root_to_turbo_root(git_root, turbo_root, path).unwrap();
-        files.insert(anchored_to_turbo_root_file_path.to_string());
+    fn reanchor_path_from_git_root_to_turbo_root(
+        &self,
+        turbo_root: &AbsoluteSystemPath,
+        path: &RelativeUnixPath,
+    ) -> Result<AnchoredSystemPathBuf, Error> {
+        let absolute_file_path = self.root.join_unix_path(path)?;
+        let anchored_to_turbo_root_file_path = turbo_root.anchor(absolute_file_path.borrow())?;
```
NicholasLYang, line 148: do we need the `borrow()` here?

crates/turborepo-scm/src/lib.rs
gsoltis: Logging is a good idea. We don't in the current implementation, but we should.

crates/turborepo-scm/src/git.rs
gsoltis: Yup

crates/turborepo-scm/src/git.rs
gsoltis: nope

NicholasLYang (approved):

PR #5406, opened 2023-06-28
feat(turborepo): Calculate package inference

gsoltis:
### Description

- move scope to its own module
- port package inference calculation
- re-enable run test

### Testing Instructions

Run test runs now, and calls the package inference code


gsoltis:
Note that more comprehensive tests will arrive when the full scope/filter test suite is ported over

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -0,0 +1,48 @@
...
+
+use crate::package_graph;
+
+pub struct PackageInference {
+    package_name: Option<String>,
+    directory_root: AnchoredSystemPathBuf,
+}
+
+impl PackageInference {
+    pub fn calculate(
```
NicholasLYang, line 12: I kinda want a doc comment here explaining what exactly the package inference algorithm should do.

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -0,0 +1,48 @@
...
+        pkg_graph: &package_graph::PackageGraph,
+    ) -> Self {
+        debug!(
+            "Using {} as a basis for selecting pacakges",
+            pkg_inference_path
+        );
+        let full_inference_path = turbo_root.resolve(pkg_inference_path);
+        for (workspace_name, workspace_entry) in pkg_graph.workspaces() {
+            let pkg_path = turbo_root.resolve(&workspace_entry.package_json_path());
+            let inferred_path_is_below = pkg_path.contains(&full_inference_path);
```
NicholasLYang, line 24: this is the same as saying the inferred path is the child of the package path, right?

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -0,0 +1,48 @@
...
+            // We skip over the root package as the inferred path will always be below it
+            if inferred_path_is_below && (&pkg_path as &AbsoluteSystemPath) != turbo_root {
+                // set both. The user might have set a parent directory filter,
```
NicholasLYang, line 27: This comment always confused me because it makes it sound like this is an error case, but really it's arguably the most common situation, no? Like you're inside a package and you run `turbo` to only run tasks in said package.

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -0,0 +1,48 @@
...
+                // in which case we *should* fail to find any packages, but we should
+                // do so in a consistent manner
+                return Self {
+                    package_name: Some(workspace_name.to_string()),
+                    directory_root: workspace_entry.package_json_path().clone(),
+                };
+            }
+            let inferred_path_is_between_root_and_pkg = full_inference_path.contains(&pkg_path);
```
NicholasLYang, line 35: So if `turbo` is executed in say, `apps/` where there are `apps/docs` and `apps/web` workspaces, it'll only run tasks in these workspaces? Do we have this documented?

NicholasLYang (commented):
The PR looks reasonable as a port, but I'm honestly a little confused about this feature and how it works.

gsoltis:
@NicholasLYang I'll add some additional comments, but this is the code to power [this](https://turbo.build/repo/docs/core-concepts/monorepos/filtering#via-global-turbo) feature. We synthesize pieces of filter expressions based on the relation of the invocation directory to the turbo root and workspace roots.

crates/turborepo-lib/src/run/scope/filter.rs
gsoltis: No, this is the case where you invoke `turbo` in some directory, but _also_ specify a filter with a directory. We'll end up concatenating the two directories. That may or may not select anything, but it's a deterministic calculation.

crates/turborepo-lib/src/run/scope/filter.rs
gsoltis: Yes

crates/turborepo-lib/src/run/scope/filter.rs
gsoltis: [This](https://turbo.build/repo/docs/core-concepts/monorepos/filtering#via-global-turbo) is what we document. But yes, running in `apps/` will pick up `apps/docs` and `apps/web`. It is equivalent to setting `--filter={./apps/*}`

arlyon (approved):
I have no additional comments beyond what @NicholasLYang added, so I'm gonna approve conditional on resolving those

crates/turborepo-lib/src/run/scope/filter.rs
NicholasLYang: I'm confused, reading the comment, it says "the user might have set a parent directory filter", which made me think this doesn't always happen. So in the case where it doesn't happen, we're just inside a package, right? Like there's no logic in this code that specifically handles the parent directory filter situation?

NicholasLYang (approved):

crates/turborepo-lib/src/run/scope/filter.rs
gsoltis: This code is used to calculate the inference values that are used, but is not the code that actually uses them.

The default is the user not setting a directory filter. The only time there is a directory filter is when they pass `--filter={<some/dir>}`. In the case where there is no directory filter set, they may have set some other filter or no filter.

In the case where they have done that, and we calculate an inferred directory, we will append their filter to the end of our inferred directory.

PR #5410, opened 2023-06-28
make sure cargo only changes run tests

ForsakenHarmony:
### Description
Also simplify conditions, anything listed in `needs` is already a requirement, so conditions from there are "inherited"

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -176,6 +176,7 @@ jobs:
 
     outputs:
       rust: ${{ steps.ci.outputs.diff != '' || steps.rust.outputs.diff != '' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -440,11 +441,10 @@ jobs:
         working-directory: crates/turbopack-ecmascript-runtime/js
         run: pnpm run check
 
-  # Very quick checks that we want to run first before all the long-running checks.
-  rust_pre_check:
+  rust_lint:
```
ForsakenHarmony, line 444: This job just checks formatting and licences (and takes 5 minutes)

It shouldn't be a requirement for running checks

sokra (approved):

PR #5415, opened 2023-06-28
Turbopack: Execution tests in node.js

wbinnssmith:
This implements execution tests in turbopack and moves webpack and turbopack test suites from the Next.js repo [0] to the turbo repo. These test core Turbopack functionality.

Right now only Node.js targets are supported for execution tests. This allows them to run quickly, and more reliably (a previously flakey test now always runs). Support for issue snapshots is also implemented.

`node_modules` for snapshots and executions tests now live in their respective directories instead of the crate tests directory.

Test Plan:
- Updated snapshot tests with `UPDATE=1 cargo test -p turbopack-tests --test snapshot -- --nocapture`
- Updated execution tests with `UPDATE=1 cargo test -p turbopack-tests --test execution -- --nocapture`
- Looked over updated snapshot tests and checked only node_modules path changed

[0] https://github.com/vercel/next.js/tree/e7b9b9a8161c88b85942be70a5218f2a554595d9/packages/next-swc/crates/next-dev-tests/tests/integration


crates/turbopack-tests/tests/execution.rs
```diff
@@ -0,0 +1,345 @@
...
+// node_modules is stored as a sibling of the test fixtures. Don't run
+// it as a test.
+//
+// "Skip" directories named `__skipped__`, which include test directories to
+// skip.
+#[testing::fixture("tests/execution/*/*/*", exclude("node_modules|__skipped__"))]
+fn test(resource: PathBuf) {
+    if resource.ends_with("__skipped__") {
+        return;
+    }
```
jridgewell, line 91: Unnecessary with the exclude?

crates/turbopack-tests/js/jest-entry.ts
```diff
@@ -0,0 +1,23 @@
...
+const uncaughtExceptions: string[] = [];
+const unhandledRejections: string[] = [];
+
+process.on("uncaughtException", (e) => {
+  uncaughtExceptions.push(String(e));
+});
+
+process.on("unhandledRejection", (e) => {
+  unhandledRejections.push(String(e));
+});
```
jridgewell, line 13: Can you verify that these actually capture in jest? I think it actually unregisters both when running tests to install their own listeners, because Jest loves to screw everything up.

crates/turbopack-tests/tests/execution.rs
```diff
@@ -0,0 +1,345 @@
...
+    tt.run_once(async move {
+        let resource_str = resource.to_str().unwrap();
+        Ok(match snapshots {
+            IssueSnapshotMode::Snapshots => {
+                (*run_test_and_snapshot_issues(resource_str).await.unwrap()).clone()
+            }
+            IssueSnapshotMode::NoSnapshots => {
+                (*run_test(resource_str).await?.js_result.await?).clone()
+            }
+        })
```
jridgewell, line 172: Nit (pseudo code, please fix)
```suggestion
        let run_result = run_test(resource);
        if matches!(snapshots, IssueSnapshotMode::Snapshots) {
          // snapshots
        }
```

crates/turbopack-tests/tests/execution.rs
```diff
@@ -0,0 +1,345 @@
...
+                ContextCondition::InDirectory("node_modules".to_string()),
+                ModuleOptionsContext {
+                    ..Default::default()
+                }
+                .cell(),
+            )],
+            custom_ecma_transform_plugins: Some(CustomEcmascriptTransformPluginsVc::cell(
+                CustomEcmascriptTransformPlugins {
+                    source_transforms: vec![TransformPluginVc::cell(
+                        Box::<StyledJsxTransformer>::default(),
```
jridgewell, line 262: Do we need Styled for node execution tests?

crates/turbopack-tests/tests/execution.rs
```diff
@@ -0,0 +1,345 @@
...
+    let entry = context.process(
+        SourceAssetVc::new(jest_entry_asset).into(),
+        Value::new(ReferenceType::Entry(EntryReferenceSubType::Undefined)),
+    );
+
+    let res = evaluate(
+        entry,
+        chunk_root_path,
+        CommandLineProcessEnvVc::new().into(),
+        modules.first().unwrap().ident(),
```
jridgewell, line 324: Nit: can't we get this from the `entry`? Or the `jest_runtime_asset` directly (which is what `modules.first().unwrap()` gives us)

crates/turbopack-tests/tests/execution.rs
```diff
@@ -0,0 +1,345 @@
...
+        CompletionVc::immutable(),
+        false,
+    )
+    .await?;
+
+    let mut read = res.read();
+    let bytes = read
+        .next()
+        .await
+        .context("test node result did not emit anything")??;
```
jridgewell, line 338: You need a second pull, else the generator will remain in memory and never close. That's why we added the `try_into_single` method.

jridgewell (changes requested):
Sorry, I should have done a full review earlier.

crates/turbopack-tests/tests/execution.rs
```diff
@@ -156,77 +152,48 @@ fn get_messages(js_results: JsResult) -> Vec<String> {
...
-            IssueSnapshotMode::Snapshots => {
-                (*run_test_and_snapshot_issues(resource_str).await.unwrap()).clone()
-            }
-            IssueSnapshotMode::NoSnapshots => {
-                (*run_test(resource_str).await?.js_result.await?).clone()
-            }
-        })
+        let run_result = run_test(resource_str);
+        if matches!(snapshot_mode, IssueSnapshotMode::Snapshots) {
+            snapshot_issues(run_result);
```
jridgewell, line 163:
```suggestion
            snapshot_issues(run_result).await?;
```

crates/turbopack-tests/js/jest-entry.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export default async function run() {
+  const jestResult = await jest.run();
+
+  // Wait a full tick for unhandledRejection handlers to run -- a microtask is not sufficient.
+  await new Promise((resolve) => setTimeout(resolve, 0));
```
jridgewell, line 19: Technically, this isn't sufficient either. We'd need something like a `FinalizationRegistry` paired with an `AsyncLocalStorage` to detect that all async processes have finished.

crates/turbopack-tests/js/jest-entry.ts
jridgewell: Has this been verified?

jridgewell (approved):

crates/turbopack-tests/js/jest-entry.ts
wbinnssmith: yeah, via an unhandled Promise.reject in test code.

ForsakenHarmony (approved):

PR #5416, opened 2023-06-28
feat(turbopack-ecmascript): detect top level await after parsing

ForsakenHarmony:
### Description
For now, just add a field to the parse result

Closes WEB-1244


crates/turbopack-ecmascript/src/parse/mod.rs
```diff
@@ -330,6 +334,10 @@ async fn parse_content(
                 is_typescript,
             ));
 
+            let mut tla_visitor = TopLevelAwaitVisitor::default();
+
+            parsed_program.visit_with(&mut tla_visitor);
```
ForsakenHarmony, line 339: Wasn't sure when to check for it, so I've done it before it gets transformed in case preset env gets rid of it 🤔

crates/turbopack-ecmascript/src/parse/top_level_await.rs
```diff
@@ -0,0 +1,17 @@
...
+pub(super) struct TopLevelAwaitVisitor {
+    pub(super) has_top_level_await: bool,
+}
+
+impl Visit for TopLevelAwaitVisitor {
+    fn visit_await_expr(&mut self, _: &AwaitExpr) {
+        self.has_top_level_await = true;
+    }
+
+    // prevent non top level items from visiting their children
```
ForsakenHarmony, line 13: are there any other items that make it not top level?

crates/turbopack-ecmascript/src/parse/top_level_await.rs
```diff
@@ -0,0 +1,17 @@
...
+    fn visit_arrow_expr(&mut self, _: &ArrowExpr) {}
+    fn visit_class(&mut self, _: &Class) {}
```
jridgewell, line 15: Class can have it in a few positions, like the `extends (await foo)`, `static x = await foo`, `[await foo] = 1;`. You need to skip the class instance property value, and method bodies.

crates/turbopack-ecmascript/src/parse/top_level_await.rs
```diff
@@ -0,0 +1,17 @@
...
+    fn visit_function(&mut self, _: &Function) {}
```
jridgewell, line 16: I don't think `function` is an alias the way Babel uses it. Instead, we'd need more like  https://github.com/swc-project/swc/blob/main/crates/swc_ecma_transforms_compat/src/es2022/class_properties/used_name.rs#L21-L31

jridgewell (changes requested):

crates/turbopack-ecmascript/src/parse/top_level_await.rs
ForsakenHarmony: Thought class would cover everything, but I forgot about objects and we need some parts of the class now anyways

ForsakenHarmony: yeah I kinda forgot about object getters and setters

crates/turbopack-ecmascript/src/parse/mod.rs
sokra: It must be after transforming. Transforms might add new TLAs, so we want to see them.

I would put it into `AnalyzeEcmascriptModuleResult` instead of `ParseResult`

sokra (approved):

PR #5419, opened 2023-06-29
remove ChunkingType::Separate

sokra:
### Description

We don't need that.
That simplifies code

Also renames SeparateAsync to Async to simplify the name a bit

### Testing Instructions

CI


crates/turbopack-css/src/chunk/single_item_chunk/reference.rs
```diff
@@ -47,11 +44,3 @@ impl ValueToString for SingleItemCssChunkReference {
...
     }
 }
-
-#[turbo_tasks::value_impl]
-impl ChunkableAssetReference for SingleItemCssChunkReference {
-    #[turbo_tasks::function]
-    fn chunking_type(&self) -> Result<ChunkingTypeOptionVc> {
-        Ok(ChunkingTypeOptionVc::cell(Some(ChunkingType::Separate)))
-    }
-}
```
alexkirsz, line 57: Have you verified that this still produces a single CSS chunk for each source CSS asset?

sokra: I haven't, I guess the test do that.

But the `SingleItemCssChunkReference` is used in references from `CssChunk`, which is in the output graph, where `ChunkableAssetReference` is not used at all. `ChunkableAssetReference` is only used in the module graph while determining the chunking.

alexkirsz (approved):

PR #5436, opened 2023-06-30
feat(lockfile) hide berry lockfile lifetime

chris-olszewski:
### Description

In order to let the package graph hold a lockfile we need to avoid exposing a lifetime in the berry implementation. This PR changes the internal representations to all be completely owned.

Alternatives considered:
 - Continue with the self-referential struct, but have the lockfile pinned and then use `mem::transmute` to mutate the required lifetimes to be `static`. Attempted this, but ran into enough issues that it wasn't looking like a quick refactor.
 - Use [ouroboros](https://docs.rs/ouroboros/0.17.0/ouroboros/attr.self_referencing.html) to allow for the self-referential struct via the macro. In the future this might be worth a look, but I didn't want to block on verifying that we currently follow and in the future will follow all of the restrictions this macro imposes.

### Testing Instructions

This shouldn't change any behavior, just makes it so doesn't require any additional data.

`cargo instruments --example berry_resolutions -t time` shows that execution is `68ms` which is still very reasonable. These changes do increase memory pressure somewhat, but it's not unreasonable at all.


arlyon (approved):
Looks good. It seems like a minimal set of changes to where we could potentially make it unowned again in the future so I am for it.

PR #5439, opened 2023-07-01
Add support for `"workspaces"` config objects in yarn

eykrehbein:
### Description

To be able to use the [`nohoist` config](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/) in yarn workspaces, an object with the keys `nohoist` and `packages` needs to be provided for the `"workspaces"` entry in package.json. 

The proposed changes add support for a non-array `workspaces` entry in `package.json` when using yarn. Before that, only arrays were supported which let to the following exception inside the `expandWorkspaces` function when using an object:

```
$ yarn turbo gen <generator>
>>> Unexpected error. Please report it as a bug:
r.flatMap is not a function
```

A function (`parseWorkspacePackages`) was introduced to dynamically parse the `"workspaces"` config by reusing the logic that was already introduced to parse `pnpm` configs.

### Testing Instructions

The given test for yarn was adapted to use a `workspaces` object instead of an array


arlyon:
Hm, odd behaviour from yarn to say the least but I am open to including this. @tknickman is the generator master but he is away right now so I will take over. Going to run CI first then have a look, thanks!

eykrehbein:
@arlyon Is there anything I can do from my side right now? 🙂

arlyon:
I am merging in main to this branch to see if it helps with CI. This 'symmetric difference' error pops up every once in a while.

packages/turbo-workspaces/src/utils.ts
```diff
@@ -90,12 +108,8 @@ function getPnpmWorkspaces({
...
       const workspaceConfig = yaml.load(fs.readFileSync(workspaceFile, "utf8"));
       // validate it's the type we expect
-      if (
-        workspaceConfig instanceof Object &&
-        "packages" in workspaceConfig &&
-        Array.isArray(workspaceConfig.packages)
-      ) {
-        return workspaceConfig.packages as Array<string>;
+      if (workspaceConfig instanceof Object) {
+        return parseWorkspacePackages({ workspaceConfig });
```
chris-olszewski, line 112: `pnpm` doesn't need to share this codepath as `pnpm-workspace.yaml` has it's own schema.

packages/turbo-workspaces/src/managers/npm.ts
```diff
@@ -57,9 +57,9 @@ async function read(args: ReadArgs): Promise<Project> {
       lockFile: "package-lock.json",
     }),
     workspaceData: {
-      globs: packageJson.workspaces || [],
+      globs: (packageJson.workspaces as Array<string> | undefined) || [],
       workspaces: expandWorkspaces({
-        workspaceGlobs: packageJson.workspaces,
+        workspaceGlobs: packageJson.workspaces as Array<string> | undefined,
```
chris-olszewski, line 62: `npm` also supports declaring workspaces via `"workspaces": { "packages": ["package/a"] }` so we should use the `parseWorkspacePackages` method you added. [source](https://github.com/npm/map-workspaces/blob/main/lib/index.js#L29-L40)

chris-olszewski (changes requested):
Thanks for adding support for this! Only two minor changes and I think this is ready to go. Sorry for the delay on the review.

PR #5450, opened 2023-07-03
feat: async modules / top level await

ForsakenHarmony:
### Description
For now this is just for top level await, in the future we want to be able to support external esm modules

### Testing Instructions

Closes WEB-434
Closes WEB-987


crates/turbopack-ecmascript-runtime/js/src/shared/runtime-utils.ts
```diff
@@ -197,3 +232,172 @@ function requireContext(
...
+
+  return {
+    promise,
+    resolve: resolve!,
+    reject: reject!,
+  };
+}
+
+// everything below is just stolen from webpack
+const turbopackQueues = Symbol("turbopack queues");
```
ForsakenHarmony, line 268: I wanted to use a const string as the alternative, but typescript doesn't like it (TS1170)
```ts
const turbopackQueues = typeof Symbol !== "undefined" ? Symbol("turbopack queues") : "__turbopack_queues__" as const;
```

[Typescript Playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBFCuAnARiADgQ2AawIrwFNCIYBeOATzQJADMYBlCgW1QBsYBCU8gInjAATArQCWYAoN4wA-IxbsAFLwQp0WbDACOhYrwCUMAFwxeAfTOrUmHGZ1ECEC9IwlQkKAG4AUAHpfMIEwAHoy3t5QVAQwAIIQFGDA+A4AYmBkMIqKhqQAfDAAbiCigoYAZDAA3tq6BADCIAJQJmDwrASIMAC+PpHUsfGJyYQZcQlJtWkA2gC6MBXViI4gbAWSJqgrBBjpPeF90WOJALIggvBsBACiAB6w5JXegVNW6jjDjjMmirRg3-aEExHCYOHL5IolMGFYqCHx7IA)

crates/turbopack-ecmascript-runtime/js/src/shared/runtime-utils.ts
```diff
@@ -168,11 +181,37 @@ function interopEsm(
...
+      get [turbopackQueues]() {
+        return raw[turbopackQueues];
+      },
+      get [turbopackError]() {
+        return raw[turbopackError];
+      },
+    } satisfies AsyncModuleExt);
+
+    return module.namespaceObject;
+  }
```
sokra, line 213: A commonjs module exporting a Promise under `module.export` should not alter the Promise in any way.
Only Promises returned by ESM should be altered.

See interop test: https://sokra.github.io/interop-test/#single-promise-object-with-default-export
All bundlers and node agree on that here.

crates/turbopack-ecmascript/src/analyzer/top_level_await.rs
```diff
@@ -1,19 +1,24 @@
...
     let mut visitor = TopLevelAwaitVisitor::default();
 
     m.visit_with(&mut visitor);
 
-    visitor.has_top_level_await
+    visitor.top_level_await_span
 }
 
 #[derive(Default)]
 struct TopLevelAwaitVisitor {
```
sokra, line 20: It might be possible to merge this visitor into `detect_dynamic_export` to avoid walking the ast multiple times

sokra (changes requested):
Looks good in general.

I'm not sure if this hangs when there are cycles in the module graph... need to verify, maybe with a test

It would also be great to have some tests that actually evaluate the code. Wills PR would help here

crates/turbopack-ecmascript-runtime/js/src/shared/runtime-utils.ts
ForsakenHarmony: That should work now, I wonder how webpack handles that when you also have async modules

sokra (changes requested):
Ok there are a bunch of new interactions that didn't exist with webpack.

1. We expose 2 types of exports `module.exports` (used for require) and `module.namespaceObject` (use for import)
2. For an non-ESM (where `module.namespaceObject` doesn't exist), we use interopEsm to create a interop namespace object.
3. For ESM `module.exports` and `module.namespaceObject` must be equal
4. We need to proxify the exports for `dynamicExport`. (Both `module.exports` and `module.namespaceObject`)
5. We need to promisify the exports for async modules.

4. and 5. are quite new.

First thing: It seems like `dynamicExport` is a bit wrong (even on `main`) as it doesn't proxify `module.exports`. We need to add that.

Second thing: When combining 4. and 5. we need to proxify the resolved value of the promise. But we also don't want to change the Promise instance since it could already be imported (and awaited) by some other modules.

So we want to be able to change the resolved value.

So the `dynamicImport` method probably wants to detect async modules and modify the `module.namespaceObject[turbopackExports]` value. And the async modules implementation wants to avoid using it's local scoped `exports` object and instead resolve to the current value of `promise[turbopackExports]`, which might be proxified.

ForsakenHarmony:
* main
  * **PR #5506** <a href="https://app.graphite.dev/github/pr/vercel/turbo/5506" target="_blank">Graphite</a> 
    * **PR #5450** <a href="https://app.graphite.dev/github/pr/vercel/turbo/5450" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/5450?utm_source=stack-comment).

crates/turbo-tasks-build/src/lib.rs
```diff
@@ -162,7 +162,6 @@ pub fn rerun_if_glob(globs: &str, root: &str) {
     let cwd = env::current_dir().unwrap();
     let globs = cwd.join(globs.replace('/', PATH_SEP.to_string().as_str()));
     let root = cwd.join(root.replace('/', PATH_SEP.to_string().as_str()));
-    println!("cargo:rerun-if-changed={}", root.display());
```
sokra, line 165: This is important to detect a new directory added

ForsakenHarmony: But then we only need this and not the others because it's recursive according to the docs?

crates/turbo-tasks-bytes/src/stream.rs
```diff
@@ -99,6 +99,16 @@ pub enum SingleValue<T> {
...
 
+impl<T: fmt::Debug> fmt::Debug for SingleValue<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            SingleValue::None => f.debug_struct("SingleValue::None").finish(),
+            SingleValue::Multiple => f.debug_struct("SingleValue::Multiple").finish(),
+            SingleValue::Single(v) => f.debug_tuple("SingleValue::Single").field(v).finish(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
ForsakenHarmony: Unfortunately not because of the trait bound

crates/turbopack-ecmascript-runtime/js/src/shared/runtime-utils.ts
```diff
@@ -115,6 +131,17 @@ function dynamicExport(module: Module, object: Record<string, any>) {
...
       },
     });
+
+    // If this is inside an async module `module.namespaceObject` is a promise,
+    // so we need to replace it with a new promise.
+    if (isPromise(module.namespaceObject)) {
+      module.exports = module.namespaceObject = maybeWrapAsyncModulePromise(
+        module.namespaceObject,
+        () => namespaceObject
+      );
```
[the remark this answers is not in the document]
ForsakenHarmony: Done, but async modules now always have a proxy as exports

sokra (approved):
I'll merge it, but there are some outstanding follow-up tasks:

* Add documentation to all public items. Especially in the async_module.rs file. Mention what `Scc` stands for or maybe even write out the full name.
* Add a test case of circular dependency graphs.

PR #5452, opened 2023-07-04
improve dynamic reexporting

sokra:
### Description

This fixes some problems when `export * from "client component"`, as `"use client"` creates a non-enumerable Proxy object, so we need to really redirect property access.

* `__turbopack_export_namespace__` counts as dynamic
* use Proxy to correctly reexport non-enumerable or changing objects


crates/turbopack-ecmascript-runtime/js/src/shared/runtime-utils.ts
```diff
@@ -81,12 +84,38 @@ function esmExport(module: Module, getters: Record<string, () => any>) {
...
+    module.namespaceObject = new Proxy(module.exports, {
+      get(target, prop) {
+        if (
+          hasOwnProperty.call(target, prop) ||
+          prop === "default" ||
+          prop === "__esModule"
+        ) {
+          return Reflect.get(target, prop);
+        }
+        for (const obj of module[REEXPORTED_OBJECTS]!) {
```
alexkirsz, line 101:
```suggestion
        for (const obj of reexportedObjects) {
```

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -2553,18 +2574,27 @@ pub struct AstPath(#[turbo_tasks(trace_ignore)] Vec<AstParentKind>);
...
 pub fn is_turbopack_helper_import(import: &ImportDecl) -> bool {
-    import.asserts.as_ref().map_or(true, |asserts| {
+    import.asserts.as_ref().map_or(false, |asserts| {
         asserts.props.iter().any(|assert| {
             assert
                 .as_prop()
                 .and_then(|prop| prop.as_key_value())
                 .and_then(|kv| kv.key.as_ident())
-                .map_or(true, |ident| &*ident.sym != TURBOPACK_HELPER)
+                .map_or(false, |ident| &*ident.sym == TURBOPACK_HELPER)
```
alexkirsz, line 2583: Was this broken before, or was it used some other way?

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -2576,22 +2606,31 @@ fn has_cjs_export(p: &Program) -> bool {
...
 
     impl Visit for Visitor {
         visit_obj_and_computed!();
 
         fn visit_ident(&mut self, i: &Ident) {
-            if &*i.sym == "module"
-                || &*i.sym == "exports"
-                || &*i.sym == "__turbopack_export_value__"
-            {
+            if &*i.sym == "module" || &*i.sym == "exports" {
```
alexkirsz, line 2624: Does this only work if the ident is unresolved? e.g. will it false positive `const module = {}`?

crates/turbopack-ecmascript/src/references/mod.rs
sokra: It was broken before

crates/turbopack-ecmascript/src/references/mod.rs
sokra: Yes that's broken in that case. It's only a difference between no exports and commonjs exports, and doesn't matter at runtime. It generates a bit less code for "no exports", but nevermind...

alexkirsz: Can you add a comment explaining this, so we know that this is broken but it doesn't actually matter?

crates/turbopack-ecmascript-runtime/js/src/shared/runtime-utils.ts
```diff
@@ -99,15 +99,15 @@ function dynamicExport(module: Module, object: Record<string, any>) {
         ) {
           return Reflect.get(target, prop);
         }
-        for (const obj of reexportedObjects) {
```
alexkirsz: Didn't think you'd need this with `reexportedObjects` being clearly set to a defined value above.

alexkirsz (approved):

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -588,27 +588,48 @@ pub(crate) async fn analyze_ecmascript_module(
...
+                        star_exports: Default::default(),
+                    }
+                    .cell(),
+                )
             }
-            .cell()
-            .as_issue()
-            .emit();
+            DetectedDynamicExportType::Namespace => EcmascriptExports::DynamicNamespace,
+            DetectedDynamicExportType::Value => EcmascriptExports::Value,
```
ForsakenHarmony, line 609: hmm, I think this unfortunately breaks my async module implementation 🙃

PR #5461, opened 2023-07-05
use a lock to ensure atomic invalidation from file changes

sokra:
### Description

fixes "app_dir must be a directory" error

### DIFF
--- a/crates/turbo-tasks-fs/src/lib.rs
+++ b/crates/turbo-tasks-fs/src/lib.rs
@@ -51,6 +51,7 @@ use serde_json::Value;
 use tokio::{
     fs,
     io::{AsyncBufReadExt, AsyncReadExt, BufReader},
+    sync::RwLock,
 };
 use tracing::{instrument, Level};
 use turbo_tasks::{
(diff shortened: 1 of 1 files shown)


crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -156,6 +157,11 @@ pub struct DiskFileSystem {
     invalidator_map: Arc<InvalidatorMap>,
     #[turbo_tasks(debug_ignore, trace_ignore)]
     dir_invalidator_map: Arc<InvalidatorMap>,
+    /// Lock that makes invalidation atomic. It will keep a write lock during
+    /// watcher invalidation and a read lock during other operations.
+    #[turbo_tasks(debug_ignore, trace_ignore)]
+    #[serde(skip)]
+    invalidation_lock: Arc<RwLock<()>>,
```
alexkirsz, line 164: It feels wrong that this should be a lock on `()`. Can we use the type system to ensure we can't do this wrong instead? e.g. by having this lock condition access to the other locks?

jridgewell: I think you're suggesting that we have `RwLock<InvalidatorMap>`? Which I was going to suggest, too.

alexkirsz (approved):

PR #5470, opened 2023-07-06
[do not merge] Replace hashing algorithm with a cross-platform solution w/ capnp

arlyon:
### Description

This PR changes the hash function in turbo to use one based on capnproto and implements it in rust and go, as well as adding a few basic tests to ensure the hash output is the same across languages.

Capnproto is sensitive to ordering (since it does not serialize, it just sends intra-segment pointers as-is over the wire) so the code must be carefully compared between platforms.

### Testing Instructions

See tests. I may consider setting up a fuzzing harness to ensure they spit out the same value.

### Notes

The go implementation requires a 'stdlib' which I have just copied as-is into the repo. Similarly, it relies on running commands when the schema changes which also has not been set up to run automatically. This is due to the imminent scrapping of all that infrastructure once the go code is finally made obsolete.


arlyon:
Superseded

PR #5472, opened 2023-07-06
chore: suggest git upgrade

chris-olszewski:
### Description

We rely certain git flags to properly function and we can use the exit code `129` to identify when git doesn't recognize a flag we passed it. Currently suggesting `2.18` as `--no-renames` was added in that version and I believe that's the newest flag we rely on.

### Testing Instructions
Eyes


NicholasLYang (approved):

arlyon:
Because I was interested, ubuntu 16.04 LTS is on 2.7.4, but current 'in-life' LTS of both ubuntu and debian are on at least 2.20

PR #5479, opened 2023-07-07
chore: Windows ARM Compilation

NicholasLYang:
### Description

Removed ring as a required dependency, so that we can build Turbo natively on Windows ARM with the `aarch64-pc-windows-msvc` target. This does *not* let us distribute a native binary because we still have the MinGW/MSVC incompatibility between Go and Rust. However, it does allow us to run Rust tests on Windows ARM, and even build a full, albeit buggy `turbo` binary.

### Testing Instructions

Try it out on Windows ARM! Will require a 64 bit C toolchain and the following incantation `cargo build -p turbo --no-default features --features native-tls`


arlyon (approved):
Great!

PR #5487, opened 2023-07-10
turbopack-cli: modularize code to support turbopack build

wbinnssmith:
This work prepares for `turbopack build` by extracting code common between dev and build, such as asset context.

This also makes the cli accept a variadic list of entries rather than an unnamed directory parameter. The project directory is now available under `--dir` or `-d`.

Test Plan: `cargo run -p turbopack-cli dev` with a `src/index.js` present and `cargo run -p turbopack-cli dev src/entry.js` with `src/entry.js` present.

PR #5488, opened 2023-07-10
turbopack-cli: implement `turbopack build`

wbinnssmith:
Depends on #5487

This implements a basic version of `turbopack build`, only targeting browser targets. In the future, we could accept a cli or configuration option to target Node.

Test Plan: `cargo run -p turbopack-cli build` with a `src/index.js` present and `cargo run -p turbopack-cli build src/entry.js` with `src/entry.js` present.


crates/turbopack-cli/src/build/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+                ConsoleUiVc::new(TransientInstance::new(LogOptions {
+                    project_dir: PathBuf::from(self.project_dir),
+                    current_dir: current_dir().unwrap(),
+                    show_all: self.show_all,
+                    log_detail: self.log_detail,
+                    log_level: self.log_level,
+                }))
+                .into();
+
+            handle_issues(build_result, issue_reporter, &None, &None).await?;
```
jridgewell, line 127: Should we fail the build process if there are issues?

jridgewell (approved):

crates/turbopack-cli/src/build/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        .try_join()
+        .await?;
+
+    for chunk_group in entry_chunk_groups {
+        for entry in &*chunk_group.await? {
+            for asset in &*all_assets_from_entry(entry.to_owned()).await? {
+                asset.content().write(asset.ident().path()).await?;
+            }
+        }
+    }
```
sokra, line 289: That's pretty inefficient since it's generating and writing all assets sequentially. You could generate and write them all in parallel.

sokra (approved):

PR #5492, opened 2023-07-11
Use Workspace Configurations.

anthonyshew:
### Description

We don't have any examples to show folks for [Workspace Configurations](https://turbo.build/repo/docs/core-concepts/monorepos/configuring-workspaces). This example is a good one for doing so because of the various output directories involved.

### DIFF
--- a/examples/kitchen-sink/README.md
+++ b/examples/kitchen-sink/README.md
@@ -2,6 +2,8 @@
 
 This is an official starter Turborepo with multiple meta-frameworks all working in harmony and sharing packages.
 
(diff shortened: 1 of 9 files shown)


arlyon (approved):
Do we want this to be the 'gold standard' (minimally invasive config) or should we perhaps have a new example that includes this? I think given kitchen sink is supposed to be the complicated one so my vote is for the latter.

anthonyshew:
@arlyon All the rest of the examples use a root `turbo.json` so I wanted to specifically show off how you'd iteratively use the API differently as things grow on this one. 

There also might be a world where Workspace Configurations become the more preferable way to handle a Turborepo so helping folks get acquainted in an example seems like a good call.

PR #5503, opened 2023-07-12
feat(turbopack_core): define trait for diagnostics

kwonoj:
### Description

Extracting trait definition from https://github.com/vercel/next.js/pull/52356, renamed it for more generic usage that turbopack can emit any kind of arbitary metadata, but next-swc can use it as telemetry for its subset.

### DIFF
--- a/crates/turbopack-core/src/diagnostics/mod.rs
+++ b/crates/turbopack-core/src/diagnostics/mod.rs
@@ -1,0 +1,84 @@
+use std::collections::HashMap;
+
+use anyhow::Result;
+use async_trait::async_trait;
+use turbo_tasks::{emit, CollectiblesSource, Upcast, Vc};
+
+#[turbo_tasks::value(serialization = "none")]
+#[derive(Clone, Debug)]
+pub struct PlainDiagnostic {
+    pub category: String,
+    pub name: String,
+    pub payload: HashMap<String, String>,
+}
+
+#[turbo_tasks::value(transparent)]
+pub struct DiagnosticPayload(pub HashMap<String, String>);
+
+/// An arbitrary payload can be used to analyze, diagnose
+/// Turbopack's behavior.
+#[turbo_tasks::value_trait]
+pub trait Diagnostic {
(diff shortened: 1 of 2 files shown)


crates/turbopack-core/src/diagnostics/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+{
+    async fn peek_diagnostics_with_path(self) -> Result<Vc<CapturedDiagnostics>>;
+}
+
+#[async_trait]
+impl<T> DiagnosticContextExt for T
+where
+    T: CollectiblesSource + Copy + Send,
+{
+    async fn peek_diagnostics_with_path(self) -> Result<Vc<CapturedDiagnostics>> {
```
jridgewell, line 42:
```suggestion
    async fn peek_diagnostics(self) -> Result<Vc<CapturedDiagnostics>> {
```

crates/turbopack-core/src/diagnostics/mod.rs
```diff
@@ -0,0 +1,55 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use turbo_tasks::{emit, CollectiblesSource, Upcast, Vc};
+
+/// An arbitrary payload can be used to analyze, diagnose
+/// Turbopack's behavior.
+#[turbo_tasks::value_trait]
+pub trait Diagnostic {
+    // [TODO]: These are subject to change; not finalized yet.
```
jridgewell, line 9: Can we add docs now, so that we don't forget?

kwonoj: Mind elaborate? The comment explains as-is; these interface might change over time while we visiting more telemetry structures (and other diagnostics).

jridgewell (approved):

crates/turbopack-core/src/diagnostics/mod.rs
kwonoj: Oh I see

PR #5505, opened 2023-07-12
Implement hashing fallback

gsoltis:
### Description

 - add manual file hashing as a fallback for _any_ error from git-based hashing

### Testing Instructions

Added a new test that verifies we can still hash when git errors

### DIFF
--- a/crates/turborepo-paths/src/absolute_system_path_buf.rs
+++ b/crates/turborepo-paths/src/absolute_system_path_buf.rs
@@ -202,6 +202,10 @@ impl AbsoluteSystemPathBuf {
         fs::remove_file(self.0.as_path())
     }
 
+    pub fn remove_dir_all(&self) -> Result<(), io::Error> {
+        fs::remove_dir_all(self.0.as_path())
+    }
+
     pub fn set_readonly(&self) -> Result<(), PathError> {
         let metadata = fs::symlink_metadata(self)?;
         let mut perms = metadata.permissions();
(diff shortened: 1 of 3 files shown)


crates/turborepo-paths/src/absolute_system_path_buf.rs
```diff
@@ -202,6 +202,10 @@ impl AbsoluteSystemPathBuf {
         fs::remove_file(self.0.as_path())
     }
 
+    pub fn remove_dir_all(&self) -> Result<(), io::Error> {
```
chris-olszewski, line 205: This should be implemented on `AbsoluteSystemPath` instead of the owned type

chris-olszewski (approved):

crates/turborepo-paths/src/absolute_system_path_buf.rs
gsoltis: Good call. We should probably clean that up in general, there's currently a mix

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -20,7 +21,19 @@ impl SCM {
...
             ),
-            SCM::Git(git) => git.get_package_file_hashes(turbo_root, package_path, inputs),
+            SCM::Git(git) => git
+                .get_package_file_hashes(turbo_root, package_path, inputs)
+                .or_else(|e| {
+                    debug!(
+                        "failed to use git to hash files: {}. Falling back to manual",
+                        e
+                    );
+                    crate::manual::get_package_file_hashes_from_processing_gitignore(
```
nathanhammond, line 31: I carefully moved a lot of this fallback into "application" code inside of Go because doing the fallback lower-level caused the "we have three different ways that we hash files which are only mostly-compatible" situation.

I'd much prefer to hoist this back into application code out of `scm`.

PR #5508, opened 2023-07-12
fix(turborepo): Copy bytes from stdout and stderr before saving them

gsoltis:
### Description

 - `io.Writer` states that we cannot retain the bytes we are passed, so we need to make a copy of them before we save them

### Testing Instructions

Verified manually running vercel-site typechecking and observing the output is now correct


chris-olszewski (approved):
Gotta love Go slices

PR #5509, opened 2023-07-13
AST-based modification of turbo.json

nathanhammond:
When we run `link` and `unlink` we want those commands to create the smallest set of changes possible in the output while also accounting for edge cases like multiple definition.

This PR adds an AST-aware JSON-rewriter that modifies the file contents rather than attempting to deserialize and reserialize the entire document (which would otherwise be lossy unless deserialized to a concrete syntax tree).

The only implementation of a concrete syntax tree for JSON in the Rust ecosystem appears to be https://github.com/rome/tools/tree/main/crates/rome_json_parser which presently lexes `jsonc`, but does not link it into the rest of the parser. There is an open, abandoned PR for that effort: https://github.com/rome/tools/pull/4382

After review, I decided that the additional complexity and size of pulling in the entire Rome parsing toolchain was not going to be worthwhile at this time for a task this small in scope.

I investigated `pest` as an alternative, but grammar correctness edge cases for JSON are not great, and the whitespace and comment macros in the `pest` grammar would have to be hand-implemented. I elected against this for low confidence in a toy parser.

Eventually I chose to use `jsonc_parser` which doesn't actually create all of the necessary tokens, but using conservative patterns on top of the limited information I can ensure that it doesn't _fail_. However, this greatly limits the formatting abilities that we can accomplish. I decided this was an acceptable tradeoff for complexity, total effort, dep size, scope of use (just `link` and `unlink` setting one property).

Additional output formatting improvements are out-of-scope for this PR. This generates assuredly parseable content, but makes no guarantees as to what the output format looks like.

### DIFF
--- a/crates/turborepo-lib/src/commands/unlink.rs
+++ b/crates/turborepo-lib/src/commands/unlink.rs
@@ -1,9 +1,9 @@
-use std::{fs, fs::File};
+use std::fs;
 
 use anyhow::{Context, Result};
 use turborepo_ui::GREY;
 
-use crate::{cli::LinkTarget, commands::CommandBase, config::RawTurboJSON};
+use crate::{cli::LinkTarget, commands::CommandBase, config::RawTurboJSON, rewrite_json};
 
 enum UnlinkSpacesResult {
     Unlinked,
@@ -55,25 +55,13 @@ pub fn unlink(base: &mut CommandBase, target: LinkTarget) -> Result<()> {
 
 fn remove_spaces_from_turbo_json(base: &CommandBase) -> Result<UnlinkSpacesResult> {
     let turbo_json_path = base.repo_root.join_component("turbo.json");
+    let turbo_json = fs::read_to_string(&turbo_json_path)?;
 
-    let turbo_json_contents =
-        fs::read_to_string(&turbo_json_path).context("unable to open turbo.json file")?;
-    let mut turbo_json: RawTurboJSON = serde_json::from_str(&turbo_json_contents)?;
-    let has_spaces_id = turbo_json
-        .experimental_spaces
-        .unwrap_or_default()
-        .id
-        .is_some();
-    // remove the spaces config
-    // TODO: in the future unlink should possible just remove the spaces id
-    turbo_json.experimental_spaces = None;
-
-    // write turbo_json back to file
-    let config_file = File::create(&turbo_json_path)?;
-    serde_json::to_writer_pretty(&config_file, &turbo_json)?;
-
-    match has_spaces_id {
-        true => Ok(UnlinkSpacesResult::Unlinked),
-        false => Ok(UnlinkSpacesResult::NoSpacesFound),
+    let output = rewrite_json::unset_path(&turbo_json, &["experimentalSpaces", "id"])?;
+    if let Some(output) = output {
+        fs::write(turbo_json_path, &output)?;
+        Ok(UnlinkSpacesResult::Unlinked)
+    } else {
+        Ok(UnlinkSpacesResult::NoSpacesFound)
     }
 }
--- a/crates/turborepo-lib/src/commands/unlink.rs
+++ b/crates/turborepo-lib/src/commands/unlink.rs
@@ -59,7 +59,7 @@ fn remove_spaces_from_turbo_json(base: &CommandBase) -> Result<UnlinkSpacesResul
 
     let output = rewrite_json::unset_path(&turbo_json, &["experimentalSpaces", "id"])?;
     if let Some(output) = output {
-        fs::write(turbo_json_path, &output)?;
+        fs::write(turbo_json_path, output)?;
         Ok(UnlinkSpacesResult::Unlinked)
     } else {
         Ok(UnlinkSpacesResult::NoSpacesFound)
--- a/crates/turborepo-lib/src/rewrite_json.rs
+++ b/crates/turborepo-lib/src/rewrite_json.rs
@@ -48,7 +48,7 @@ pub fn set_path(
 
     // Find the token we'll be modifying and its path from the root.
     let current_path = &mut vec![];
-    let (closest_path, closest_node) = get_closest_node(&root, &path, current_path);
+    let (closest_path, closest_node) = get_closest_node(&root, path, current_path);
 
     // Pull the token metadata off of the token.
     let (property_count, range): (usize, jsonc_parser::common::Range) = match closest_node {
@@ -210,7 +210,7 @@ pub fn unset_path(
 
     // The key path can appear multiple times. This a vec that contains each time it
     // occurs.
-    let path_ranges = find_all_paths(&root, &path);
+    let path_ranges = find_all_paths(&root, path);
 
     if path_ranges.is_empty() {
         return Ok(None);
--- a/crates/turborepo-lib/src/rewrite_json.rs
+++ b/crates/turborepo-lib/src/rewrite_json.rs
@@ -1,0 +1,425 @@
+use jsonc_parser::{errors::ParseError, parse_to_ast};
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum RewriteError {
+    #[error("The JSON document failed to parse. {0}")]
+    ParseError(ParseError),
+    #[error("The JSON document contains no root object.")]
+    NoRoot,
+}
+/**
+ * When generating replacement content it's one of two things.
+ */
+enum GenerateType {
+    Object,
+    Member,
+}
+
+/**
+ * Range stores the primitive AST token details required for document
+ * mutation.
+ */
+struct Range {
+    pub start: usize,
+    pub end: usize,
+    pub replacement_char: String,
+}
+
+/**
+ * Given a JSONC document, an object traversal path, and a _pre-serialized_
+ * JSON value `set_path` will return a minimally-mutated JSONC document with
+ * the path specified set to the JSON value.
+ *
+ * - If the path exists, it will clobber the existing contents.
+ * - If the path does not exist, it will synthesize object members and
+ *   objects to ensure it does exist.
+ *
+ * In the event that the key appears multiple times in the document the last
+ * instance of the key will be updated and earlier instances will be
+ * disregarded.
+ */
+pub fn set_path(
+    json_document_string: &str,
+    path: &[&str],
+    json_value: &str,
+) -> Result<String, RewriteError> {
+    let root = get_root(json_document_string)?;
+
+    // Find the token we'll be modifying and its path from the root.
+    let current_path = &mut vec![];
+    let (closest_path, closest_node) = get_closest_node(&root, &path, current_path);
+
+    // Pull the token metadata off of the token.
+    let (property_count, range): (usize, jsonc_parser::common::Range) = match closest_node {
+        jsonc_parser::ast::Value::Object(literal) => (literal.properties.len(), literal.range),
+        jsonc_parser::ast::Value::StringLit(literal) => (0, literal.range),
+        jsonc_parser::ast::Value::NumberLit(literal) => (0, literal.range),
+        jsonc_parser::ast::Value::BooleanLit(literal) => (0, literal.range),
+        jsonc_parser::ast::Value::Array(literal) => (0, literal.range),
+        jsonc_parser::ast::Value::NullKeyword(literal) => (0, literal.range),
+    };
+
+    // Figure out what we should be generating:
+    // - An object to be assigned to an existing member. ("object")
+    // - A member to add to an existing object. ("member")
+    let generate_type: GenerateType = if !closest_path.is_empty() {
+        GenerateType::Object
+    } else {
+        GenerateType::Member
+    };
+
+    // Identify the token replacement metadata: start, end, and possible trailing
+    // join character
+    let (start, end, separator) = match generate_type {
+        GenerateType::Object => {
+            let start = range.start;
+            let end = range.end;
+            let separator = "";
+            (start, end, separator)
+        }
+        GenerateType::Member => {
+            let start = range.start + 1;
+            let end = range.start + 1;
+            let separator = if property_count > 0 { "," } else { "" };
+            (start, end, separator)
+        }
+    };
+
+    // Generate the serialized JSON to insert into the document.
+    // We synthesize objects for missing path segments.
+    let missing_path_segments = &path[closest_path.len()..];
+    let computed_object = match generate_type {
+        GenerateType::Object => generate_object(missing_path_segments, json_value),
+        GenerateType::Member => generate_member(missing_path_segments, json_value, separator),
+    };
+
+    // Generate a new document!
+    let mut output: String = json_document_string.to_owned();
+    output.replace_range(start..end, &computed_object);
+
+    Ok(output)
+}
+
+/**
+ * get_root returns the document root, or information on the error
+ * encountered with the input json_document_string.
+ */
+fn get_root(json_document_string: &str) -> Result<jsonc_parser::ast::Value, RewriteError> {
+    let parse_result_result = parse_to_ast(
+        json_document_string,
+        &Default::default(),
+        &Default::default(),
+    );
+
+    match parse_result_result {
+        Ok(parse_result) => match parse_result.value {
+            Some(root) => Ok(root),
+            None => Err(RewriteError::NoRoot),
+        },
+        Err(parse_error) => Err(RewriteError::ParseError(parse_error)),
+    }
+}
+
+/**
+ * get_closest_node traverses the JSON document via recursive tail calls to
+ * find the last node that exists with in this JSON key path.
+ */
+fn get_closest_node<'a>(
+    current_node: &'a jsonc_parser::ast::Value<'a>,
+    target_path: &[&str],
+    current_path: &'a mut Vec<&'a str>,
+) -> (&'a Vec<&'a str>, &'a jsonc_parser::ast::Value<'a>) {
+    // No target_path? We've arrived.
+    if target_path.is_empty() {
+        return (current_path, current_node);
+    }
+
+    match current_node {
+        // Only objects can have key paths.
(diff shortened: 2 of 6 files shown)


crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+
+                        let replacement_char = if is_first || is_last { "" } else { "," };
+
+                        ranges.push(Range {
+                            start,
+                            end,
+                            replacement_char: replacement_char.to_owned(),
+                        });
+                    } else {
+                        // We must recurse.
```
NicholasLYang, line 310: lol

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+ */
+fn find_all_paths<'a>(
+    current_node: &'a jsonc_parser::ast::Value<'a>,
+    target_path: &Vec<&'a str>,
+    current_path: Vec<&'a str>,
+) -> Vec<Range> {
+    let mut ranges: Vec<Range> = vec![];
+
+    // Early exit when it's impossible to have matching ranges.
+    if target_path.len() == 0 {
```
NicholasLYang, line 267:
```suggestion
    if target_path.is_empty() {
```

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+        jsonc_parser::ast::Value::Object(literal) => (literal.properties.len(), literal.range),
+        jsonc_parser::ast::Value::StringLit(literal) => (0, literal.range),
+        jsonc_parser::ast::Value::NumberLit(literal) => (0, literal.range),
+        jsonc_parser::ast::Value::BooleanLit(literal) => (0, literal.range),
+        jsonc_parser::ast::Value::Array(literal) => (0, literal.range),
+        jsonc_parser::ast::Value::NullKeyword(literal) => (0, literal.range),
+    };
+
+    // Figure out what we should be generating:
+    // - An object to be assigned to an existing member. ("object")
```
NicholasLYang, line 63: I'm not sure these names are really clear. Could it be called `Append` or `Mutate`?

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+    let missing_path_segments = path[closest_path.len()..].to_vec();
+    let computed_object = match generate_type {
+        GenerateType::Object => generate_object(missing_path_segments, json_value),
+        GenerateType::Path => generate_path(missing_path_segments, json_value, separator),
+    };
+
+    // Generate a new document!
+    let mut output: String = json_document_string.to_owned();
+    output.replace_range(start..end, &computed_object);
+    return Ok(output);
```
NicholasLYang, line 106:
```suggestion
    Ok(output)
```

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+ * get_closest_node traverses the JSON document via recursive tail calls to
+ * find the last node that exists with in this JSON key path.
+ */
+fn get_closest_node<'a>(
+    current_node: &'a jsonc_parser::ast::Value<'a>,
+    target_path: &Vec<&'a str>,
+    current_path: Vec<&'a str>,
+) -> (Vec<&'a str>, &'a jsonc_parser::ast::Value<'a>) {
+    // No target_path? We've arrived.
+    if target_path.len() == 0 {
```
NicholasLYang, line 139:
```suggestion
    if target_path.is_empty() {
```

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+        return (current_path, current_node);
+    }
+
+    match current_node {
+        // Only objects can have key paths.
+        jsonc_parser::ast::Value::Object(obj) => {
+            let object_property = obj.properties.iter().rev().find(|property| {
+                let current_property_name = property.name.as_str();
+                return target_path[0] == current_property_name;
```
NicholasLYang, line 148:
```suggestion
                target_path[0] == current_property_name
```

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    return output;
+}
+
+/**
+ * Given path segments, generate a JSON object member with an optional
+ * trailing separator.
+ */
+fn generate_path(path_segments: Vec<&str>, value: &str, separator: &str) -> String {
```
NicholasLYang, line 197: Do we escape stuff like `"` when writing back the JSON?

NicholasLYang (commented):
First pass of review. Mostly Rust style nits and a question around escaping

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+        &Default::default(),
+        &Default::default(),
+    );
+
+    return match parse_result_result {
+        Ok(parse_result) => match parse_result.value {
+            Some(root) => Ok(root),
+            None => Err(RewriteError::NoRoot),
+        },
+        Err(parse_error) => Err(RewriteError::ParseError(parse_error)),
```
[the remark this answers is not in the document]
nathanhammond: I actually tried to golf this but it lost so much clarity I bailed back to this more-verbose version.

crates/turborepo-lib/src/rewrite_json.rs
nathanhammond: I was trying to choose names based upon the JSON spec and be consistent to that, but I didn't do a great job.

- `{ "foo": "bar" }`: object
- `"foo": value`: member where value is any valid, pre-serialized JSON thing.

`jsonc_parser` calls `members` `properties`, which I also used when adjacent to objects they provide. 🤷‍♂️

crates/turborepo-lib/src/rewrite_json.rs
nathanhammond: Nope, I tried to make that clear in the comments:

> Given a JSONC document, an object traversal path, and a _pre-serialized_ JSON value `set_path` will return a minimally-mutated JSONC document...

You'll also see that I format the spaceid string in `link.rs`. I consider this out of scope mostly because YAGNI. If we actually do need it I'm going to be finishing a PR to Rome.

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+            }
+
+            start = range.start + 1;
+            end = range.start + 1;
+        }
+    }
+
+    // Generate the serialized JSON to insert into the document.
+    // We synthesize objects for missing path segments.
+    let missing_path_segments = path[closest_path.len()..].to_vec();
```
chris-olszewski, line 97: Unnecessary allocation
```suggestion
    let missing_path_segments = &path[closest_path.len()..];
```

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+        }
+        // All other node types are complete.
+        _ => (current_path, current_node),
+    }
+}
+
+/**
+ * Given path segments, generate a JSON object.
+ */
+fn generate_object(path_segments: Vec<&str>, value: &str) -> String {
```
chris-olszewski, line 177:
```suggestion
fn generate_object(path_segments: &[&str], value: &str) -> String {
```

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    // We can only find paths on objects.
+    if let jsonc_parser::ast::Value::Object(obj) = current_node {
+        // We need a reference to the previous and next property to identify if we're
+        // looking at the first or last node which need special handling.
+        let properties: Vec<_> = obj.properties.iter().collect();
+
+        for index in 0..properties.len() {
+            let property = properties[index];
```
chris-olszewski, line 281:
```suggestion
        for (index, property) in properties.iter().copied().enumerate() {
```

chris-olszewski (commented):
This looks really cool! Still going to do another read through, but figured I'd get this up before EOD

crates/turborepo-lib/src/commands/unlink.rs
```diff
@@ -54,25 +54,16 @@ pub fn unlink(base: &mut CommandBase, target: LinkTarget) -> Result<()> {
...
-    // write turbo_json back to file
-    let config_file = File::create(&turbo_json_path)?;
-    serde_json::to_writer_pretty(&config_file, &turbo_json)?;
+    fs::write(turbo_json_path, &output)?;
 
-    match has_spaces_id {
-        true => Ok(UnlinkSpacesResult::Unlinked),
-        false => Ok(UnlinkSpacesResult::NoSpacesFound),
+    // Detect removal by comparing length.
+    if output.len() < turbo_json.len() {
```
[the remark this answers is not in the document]
nathanhammond: Yeah, it was a hack. The right answer is to make `unset_path` report on what it did.

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+ * - If the path does not exist, it will synthesize object members and
+ *   objects to ensure it does exist.
+ *
+ * In the event that the key appears multiple times in the document the last
+ * instance of the key will be updated and earlier instances will be
+ * disregarded.
+ */
+pub fn set_path(
+    json_document_string: &str,
+    path: Vec<&str>,
```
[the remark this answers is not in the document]
nathanhammond: One of the things that happened with `Vec` was it kept infecting things. This worked the other direction. Pulling the thread cause all the `Vec`s to disappear.

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+            if property_count > 0 {
+                separator = ",";
+            } else {
+                separator = ""
+            }
+
+            start = range.start + 1;
+            end = range.start + 1;
+        }
+    }
```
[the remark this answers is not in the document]
nathanhammond: I think it's maybe a bit less-vulnerable to mistakes to do it this way, but less-clear:

```rust
    let (separator, start, end) = match generate_type {
        GenerateType::Object => {
            ("", range.start, range.end)
        }
        GenerateType::Member => {
            let separator = if property_count > 0 {
                ","
            } else {
                ""
            };

            (separator, range.start + 1, range.start + 1)
        }
    };
```

I wouldn't attempt the original variant without Rust's "not-definitely-assigned" compiler help, but I have also made numerous mistakes with tuples recently and since Rust doesn't give me "named positions" in a tuple.

I've gone with a more-verbose tuple-assignment to protect from the user error:

```rust
    let (separator, start, end) = match generate_type {
        GenerateType::Object => {
            let separator = "";
            let start = range.start;
            let end = range.end;
            (separator, start, end)
        }
        GenerateType::Member => {
            let separator = if property_count > 0 {
                ","
            } else {
                ""
            };
            let start = range.start + 1;
            let end = range.start + 1;
            (separator, start, end)
        }
    };
```

crates/turborepo-lib/src/rewrite_json.rs
nathanhammond: I did a full refactor of this to function without indices and avoiding a `collect` call. What I really want is something that is `tuple_windows` but looks like this:

```rust
while let (previous: Option<&Thing>, current: &Thing, next: Option<&Thing>) = some_iterator.next() {
  let is_first = previous.is_none();
  let is_last = next.is_none();

  // Operate on `current` with context about previous and next.
}
```

`.peekable()` gives me half of this with a bit of ceremony, but nothing I could find offers `.previous()`.

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,411 @@
...
+                let current_property_name = property.name.as_str();
+                target_path[0] == current_property_name
+            });
+
+            // See if we found a matching key. If so, recurse.
+            match object_property {
+                Some(property) => {
+                    let next_current_node = &property.value;
+                    let next_property_name = property.name.as_str();
+                    let mut next_current_path = current_path.clone();
```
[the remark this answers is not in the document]
nathanhammond: I started by renaming the variable, and so that led me to a local maxima of `.clone()`. I really just wanted to alias the existing everything from that line forward.

Is `&mut *current_path` a reasonable pattern here?

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -149,7 +150,7 @@ fn get_closest_node<'a>(
                 Some(property) => {
                     let next_current_node = &property.value;
                     let next_property_name = property.name.as_str();
-                    let mut next_current_path = current_path.clone();
+                    let next_current_path = &mut *current_path;
```
nathanhammond, line 153: I want comments on this; I'm basically "renaming a variable for clarity."

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,420 @@
...
+ * specified path removed.
+ */
+pub fn unset_path(
+    json_document_string: &str,
+    path: &[&str],
+) -> Result<(bool, String), RewriteError> {
+    let root = get_root(json_document_string)?;
+
+    // The key path can appear multiple times. This a vec that contains each time it
+    // occurs.
```
chris-olszewski, line 212: Could you add a test for this case? I'm having trouble understanding when we'll have multiple key paths.

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -0,0 +1,420 @@
...
+}
+
+/**
+ * find_all_paths returns the list of ranges which define the specified
+ * token.
+ */
+fn find_all_paths<'a>(
+    current_node: &'a jsonc_parser::ast::Value<'a>,
+    target_path: &[&'a str],
+    current_path: &mut Vec<&'a str>,
```
chris-olszewski, line 260: Not sure if I'm missing something, but this argument seems to never be read? I ripped it out and all the tests continued to pass once I updated the callsites.

chris-olszewski (approved):
Second pass and this still looks really solid. Only concern I have is with the "multiple key path" case since that's where the string replacement can go sideways really fast. I'd appreciate some tests that verify that ranges always come out sorted as expected so that the string replacement works as intended.

crates/turborepo-lib/src/rewrite_json.rs
nathanhammond: There's already a test for it (every single test case includes duplicated keys); but this is basically for the the situation where:

```json
{
  "thing": "hello, world",
  "thing": "also, hello."
}
```

If you try to unset `thing` and only get one of those you didn't actually succeed at unsetting it when deserialized. Deserialize, mutate (unset), serialize ensures that you get only one, but if you don't do deserialization you have to account for garbage-in.

crates/turborepo-lib/src/rewrite_json.rs
nathanhammond: Ahh, vestigial code from when I copy/pasted the method signature from `get_closest_node`. 😜

nathanhammond: (Fantastic catch.)

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -133,9 +127,9 @@ fn get_root(json_document_string: &str) -> Result<jsonc_parser::ast::Value, Rewr
  */
 fn get_closest_node<'a>(
     current_node: &'a jsonc_parser::ast::Value<'a>,
-    target_path: &Vec<&'a str>,
-    current_path: Vec<&'a str>,
-) -> (Vec<&'a str>, &'a jsonc_parser::ast::Value<'a>) {
+    target_path: &[&'a str],
```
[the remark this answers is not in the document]
nathanhammond: Since none of the lifetimes ended up in the public API I didn't pay super-close attention to them. 😅

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -133,9 +127,9 @@ fn get_root(json_document_string: &str) -> Result<jsonc_parser::ast::Value, Rewr
...
+    current_path: &'a mut Vec<&'a str>,
+) -> (&'a mut Vec<&'a str>, &'a jsonc_parser::ast::Value<'a>) {
```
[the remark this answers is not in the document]
nathanhammond: The return definitely doesn't need to be mutable. That's a mistake.

Current state:
```
// "temporary value dropped while borrowed"
// I really wish this could get folded into the next line.
let current_path = &mut vec![];

// closest_path and current_path are actually the same vector.
let (closest_path, closest_node) = get_closest_node(&root, &path, current_path);
```

I don't love this:
```
let closest_path = &mut vec![];
let closest_node = get_closest_node(&root, &path, closest_path);

// closest_path is mutated inside of the method call, out-of-sight, out-of-mind.
```

The mutability here was to avoid cloning the Vec per-recurse:
https://github.com/vercel/turbo/pull/5509/commits/84a406ff108bb089b40ce6d3a9a17bb6b82e0dbf

***

I also don't know if it's actually achieving tail-call-optimization, but that was a, "I bet it has better odds if I do it as an argument."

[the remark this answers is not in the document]
chris-olszewski: Ah, yeah hadn't thought about TCO. Not sure if anything has changed, but I believe there's no guarantee and if you really want to achieve TCO you should probably write a loop. There are some crates that claim to allow TCO with macros, but I'm not sure if it's worth the effort considering the depth of our `turbo.json` schema.

PR #5511, opened 2023-07-13
fix(turborepo): Ensure .gitignore appended .turbo has preceding newline

nathanhammond:
When linking, if `.gitignore` does not have a trailing newline, we need to insert it.

This also adds a check for the identical `.turbo/` ignore directive.

### DIFF
--- a/crates/turborepo-lib/src/commands/link.rs
+++ b/crates/turborepo-lib/src/commands/link.rs
@@ -3,8 +3,7 @@ use std::os::unix::fs::PermissionsExt;
 use std::{
     fs,
     fs::{File, OpenOptions},
-    io,
-    io::{BufRead, Write},
+    io::{self, BufRead, Write},
 };
 
 use anyhow::{anyhow, Context, Result};
@@ -16,6 +15,8 @@ use dialoguer::{theme::ColorfulTheme, Confirm};
 use dirs_next::home_dir;
 #[cfg(test)]
 use rand::Rng;
+use regex::Regex;
+use turbopath::AbsoluteSystemPathBuf;
 use turborepo_api_client::{APIClient, CachingStatus, Space, Team};
 
 #[cfg(not(test))]
@@ -37,7 +38,6 @@ pub(crate) enum SelectedTeam<'a> {
 pub(crate) enum SelectedSpace<'a> {
     Space(&'a Space),
 }
-use itertools::Itertools;
 pub(crate) const REMOTE_CACHING_INFO: &str = "  Remote Caching shares your cached Turborepo task \
                                               outputs and logs across
   all your team’s Vercel projects. It also can share outputs
@@ -177,7 +177,7 @@ pub async fn link(
             };
 
             if modify_gitignore {
-                add_turbo_to_gitignore(base)?;
+                add_turbo_to_gitignore(&base.repo_root)?;
             }
 
             println!(
@@ -394,8 +394,9 @@ fn enable_caching(url: &str) -> Result<()> {
     Err(anyhow!("link after enabling caching"))
 }
 
-fn add_turbo_to_gitignore(base: &CommandBase) -> Result<()> {
-    let gitignore_path = base.repo_root.join_component(".gitignore");
+fn add_turbo_to_gitignore(repo_root: &AbsoluteSystemPathBuf) -> Result<()> {
+    let turbo_test_regex = Regex::new("^.turbo/? *$").unwrap();
+    let gitignore_path = repo_root.join_component(".gitignore");
 
     if !gitignore_path.exists() {
         let mut gitignore = File::create(gitignore_path)?;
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/commands/link.rs
```diff
@@ -404,15 +404,35 @@ fn add_turbo_to_gitignore(base: &CommandBase) -> Result<()> {
...
-        let mut lines = io::BufReader::new(gitignore).lines();
-        let has_turbo = lines.any(|line| line.map_or(false, |line| line.trim() == ".turbo"));
-        if !has_turbo {
-            let mut gitignore = OpenOptions::new()
-                .read(true)
-                .append(true)
-                .open(&gitignore_path)?;
-            writeln!(gitignore, ".turbo")?;
-        }
+        let lines = io::BufReader::new(&gitignore).lines();
```
NicholasLYang, line 407: tbh I don't know why I used a BufReader here. We should just load the entire file into memory. check if there's already a turbo line, and append accordingly

nathanhammond: I mean, a buffered reader is technically the correct thing to use here, so I don't see it as a bad choice, just overly-defensive.

> The default is currently 8 KB

...so in spite of it being technically correct we probably read basically everybody's `.gitignore` in full the first time. We get to choose our edge case:
- everybody pays a small bit of ceremony tax for `BufReader`.
- somebody with a massive `.gitignore` file gets memory pressure.

... neither of which are going to show up in any perf trace ever because you don't call `link` except once.

So I really only care about the correctness here.

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -394,8 +394,9 @@ fn enable_caching(url: &str) -> Result<()> {
     Err(anyhow!("link after enabling caching"))
 }
 
-fn add_turbo_to_gitignore(base: &CommandBase) -> Result<()> {
-    let gitignore_path = base.repo_root.join_component(".gitignore");
+fn add_turbo_to_gitignore(repo_root: &AbsoluteSystemPathBuf) -> Result<()> {
+    let turbo_test_regex = Regex::new("^.turbo/? *$").unwrap();
```
nathanhammond, line 398: - No user input, `unwrap` is safe.
- Uses the precise definition of whether or not `.turbo` is ignored.

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -177,7 +177,7 @@ pub async fn link(
             };
 
             if modify_gitignore {
-                add_turbo_to_gitignore(base)?;
+                add_turbo_to_gitignore(&base.repo_root)?;
```
nathanhammond, line 180: No need to send in the entire thing; just the path works.

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -404,14 +405,19 @@ fn add_turbo_to_gitignore(base: &CommandBase) -> Result<()> {
...
+
         if !has_turbo {
             let mut gitignore = OpenOptions::new()
                 .read(true)
                 .append(true)
                 .open(&gitignore_path)?;
-            writeln!(gitignore, ".turbo")?;
+
+            // Defensively ensure that it has a preceding newline.
+            writeln!(gitignore, "\n.turbo")?;
```
nathanhammond, line 420: Always precede with a newline. No need to be clever and check for existing newlines.

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -579,4 +585,61 @@ mod test {
...
+                let _ = add_turbo_to_gitignore(&repo_root);
+
+                let output = fs::read_to_string(&gitignore_path).unwrap();
+                assert_eq!(expected, output);
+            }
+        )*
+        }
+    }
+
+    gitignore_append_tests! {
```
nathanhammond, line 611: This are basically testing that `regex` + `lines` iterator isn't messing with us.

nathanhammond (commented):
Reviewer's guide!

crates/turborepo-lib/src/commands/link.rs
nathanhammond: Also, if somebody complains about a bonus newline, my "fix" to this is:

```
writeln!(gitignore, "\n# Ignore the Turborepo configuration directory.\n.turbo")?;
```

crates/turborepo-lib/src/commands/link.rs
gsoltis: Drive-by: you might want to consider using `test_case` instead. Having a unit test entirely in a macro makes it hard to set breakpoints inside of it.

Lots of examples [here](https://github.com/vercel/turbo/blob/d24075fc96d0109b7dc5b016d59ad59774134cfd/crates/turborepo-globwalk/src/lib.rs#L450)

NicholasLYang (approved):

crates/turborepo-lib/src/commands/link.rs
arlyon: Can we use expect instead? even just 'known regex' is probably enough.

arlyon (changes requested):

PR #5516, opened 2023-07-13
add filewatching tests to document behaviour regarding file add/delete

arlyon:
### Description

This adds some basic tests to validate the behaviour of the filewatching implementation across platforms. The differences are mainly due to operating system specific behaviors.

Here's a summary of the differences:

1. **TestFileWatchingSubfolderDeletion**: This test checks the recursive watching ability when a directory is deleted. It passes on all three operating systems.

2. **TestFileWatchingRootDeletion**: This test checks if file watching continues when the root is deleted and when the root is recreated. It passes on macOS, but fails on Linux and Windows due to no event being emitted when the root is recreated.

3. **TestFileWatchingSubfolderRename**: This test checks the rename event when a parent folder is renamed. It fails on all three operating systems. On macOS, rename events are not sent. On Linux, renaming generates file creation events of the new folder (and all the sub-contents), not a rename. On Windows, you cannot rename a watched folder.

4. **TestFileWatchingRootRename**: This test checks if file watching continues when the root is renamed back to its original name. It passes on macOS, but fails on Linux and Windows due to varying reasons.

5. **TestFileWatchSymlinkCreate**: This test checks the file creation event when a symlink is created. It passes on macOS and Windows (with admin permissions) but fails on Linux, because symlinks do not produce events.

6. **TestFileWatchSymlinkDelete**: This test checks if any events are raised when a symlink is deleted. It passes on all three operating systems.

7. **TestFileWatchSymlinkRename**: This test checks if any events are raised when a symlink is renamed. It passes on macOS but fails on Linux and Windows as they both raise an event for creating the file.

8. **TestFileWatchRootParentRename**: This test checks if file watching continues to report events when the parent directory of the root is renamed and renamed back. It passes on all three operating systems.

9. **TestFileWatchRootParentDelete**: This test checks if file watching resumes reporting events when the parent directory of the root is deleted and then recreated. It passes on macOS but fails on Linux and Windows, as no create event is emitted in both cases. 

### Testing Instructions

You can run this using make go-test. This works ok on macOS but running our go test suite on other platforms is complicated.

### DIFF
--- a/cli/internal/filewatcher/filewatcher_test.go
+++ b/cli/internal/filewatcher/filewatcher_test.go
@@ -152,12 +152,12 @@ func TestFileWatching(t *testing.T) {
 	expectNoFilesystemEvent(t, ch)
 }
 
-// TestFileWatchingParentDeletion tests that when a directory is deleted,
+// TestFileWatchingParentDeletion tests that when a repo subfolder is deleted,
 // recursive watching will still work for new folders
 //
 // ✅ macOS
-// ? Linux
-// ? Windows
+// ✅ Linux
+// ✅ Windows
 func TestFileWatchingSubfolderDeletion(t *testing.T) {
 	logger := hclog.Default()
 	logger.SetLevel(hclog.Debug)
@@ -234,8 +234,8 @@ func TestFileWatchingSubfolderDeletion(t *testing.T) {
 // will continue, and an add event will be sent for the re-created root.
 //
 // ✅ macOS
-// ? Linux
-// ? Windows
+// ❌ Linux - we do not get an event when the root is recreated L287
+// ❌ Windows - we do not get an event when the root is recreated L287
 func TestFileWatchingRootDeletion(t *testing.T) {
 	logger := hclog.Default()
 	logger.SetLevel(hclog.Debug)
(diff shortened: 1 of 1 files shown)


nathanhammond:
The main thing that I see here is that this is basically a cross-product of options kind of thing.

***

Position:
- Above root, in-path (parent)
- Above root, out-of-path (ensure irrelevance)
- At root
- Below root (child)

Root starting state:
- Hardlink
- Junction
- Symlink
- Directory

Action:
- Create
- Update
- Delete

Filetype:
- Normal
- Directory
- Symlink
- Hardlink
- Junction (Windows)

Symlink/Hardlink/Junction State:
- Broken
- Target out of root
- Target within the root

Directory State:
- Empty
- 2^16 files (this is a perf test, not correctness)

Directory, Symlink, Hardlink, Junction
- "Above root, in path" or "At root"
  - Points to/is a different place with an identical folder structure.

***

I'd use this framework like:

- Above the root, in the path
- where root is a directory
- create
- a symlink
- that points to a different place with an identical folder structure.
- which is outside of the current root folder.

(This example is probably one of the hardest.)

Many of these collapse to the same behavior, but going through this list would get us the enumeration of things we could possibly care about. It also means that we error on certain behaviors if we just detect which of these states we're in. "We don't support that, rebuilding your index now."

nathanhammond (commented):
These tests specify the behavior; but I _think_ we should instead try to create tests to constrain the state of the file index after each change. Which messages are going to be sent is going to differ based upon the platform/backend and I don't believe that it is worthwhile to attempt to align that.

Instead, given a starting point (e.g. FSEvents on macOS), if we treat "file watching messages" as if they were an input stream that needed parsing (state machine):

- what information from the event stream(s)
- anchored at where
- (and with whatever out-of-band lookups for additional information)

...is necessary in order to determine how to rebuild the file index?

The first-order bit here is "state of the file index." Each of our tests should be asserting against the outcome. Asserting against the events is simply specifying the behavior. It's useful for understanding while authoring the library but not material to users.

We need an eventually-consistent file index; we don't actually care at all what happens along the way. The ideal test behavior is modeling "user did X to their file system", our test asserts that after X we end up with a consistent file index. (This could mean "no changes" or "has changes" depending upon the action.)

gsoltis:
Covered in #5580

PR #5525, opened 2023-07-14
Make turbo work

nathanhammond:
Reverts https://github.com/vercel/turbo/pull/5007 which appears to have broken this.


gsoltis (changes requested):
Let's fix forward instead: https://github.com/vercel/turbo/pull/5530

anthonyshew:
Looks like we fixed forward instead!

PR #5528, opened 2023-07-14
move Asset::ident to more specific traits

sokra:
### Description

* `ident()` is no longer on Asset, but on `Module`, `Source`, `OutputAsset` or `Chunk`
* On the way, more AssetVc types needed to be switched to more specific traits

next.js PR: https://github.com/vercel/next.js/pull/52683

### DIFF
--- a/crates/turbo-tasks-memory/src/memory_backend.rs
+++ b/crates/turbo-tasks-memory/src/memory_backend.rs
@@ -443,11 +443,17 @@ impl Backend for MemoryBackend {
 
     fn task_execution_result(
         &self,
-        task: TaskId,
+        task_id: TaskId,
         result: Result<Result<RawVc>, Option<Cow<'static, str>>>,
         turbo_tasks: &dyn TurboTasksBackendApi<MemoryBackend>,
     ) {
-        self.with_task(task, |task| {
+        self.with_task(task_id, |task| {
+            #[cfg(debug_assertions)]
+            if let Ok(Ok(RawVc::TaskOutput(result))) = result.as_ref() {
+                if *result == task_id {
+                    panic!("Task {} returned itself as output", task.get_description());
+                }
+            }
             task.execution_result(result, self, turbo_tasks);
         })
     }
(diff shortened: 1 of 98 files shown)


crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -444,11 +444,17 @@ impl Backend for MemoryBackend {
...
         turbo_tasks: &dyn TurboTasksBackendApi<MemoryBackend>,
     ) {
-        self.with_task(task, |task| {
+        self.with_task(task_id, |task| {
+            #[cfg(debug_assertions)]
+            if let Ok(Ok(RawVc::TaskOutput(result))) = result.as_ref() {
+                if *result == task_id {
+                    panic!("Task {} returned itself as output", task.get_description());
+                }
+            }
```
alexkirsz, line 457: Should preferably be in a different PR.

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -338,6 +339,10 @@ where
     let mut graph_nodes = vec![];
 
     for asset in assets {
+        let asset = asset.resolve().await?;
+        let Some(asset) = ModuleVc::resolve_from(asset).await? else {
+            continue;
+        };
```
alexkirsz, line 345: Eventually we won’t need this, right? Can you add a TODO here?

crates/turbopack-core/src/virtual_output.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use crate::{
+    asset::{Asset, AssetContentVc, AssetVc},
+    ident::AssetIdentVc,
+    output::{OutputAsset, OutputAssetVc},
+};
+
+/// An [Asset] that is created from some passed source code.
+#[turbo_tasks::value]
+pub struct VirtualOutputAsset {
```
alexkirsz, line 12: I’d argue that all output assets are virtual, so might not be the best naming.

alexkirsz (approved):

PR #5531, opened 2023-07-14
feat: port prune to rust

chris-olszewski:
### Description

Port prune from Go to Rust and remove the old Go implementation. To achieve this the following was done:
 - Parsing the lockfile during package graph construction
 - Porting utility functions
 - Some minor changes in `turborepo_paths`, these were primarily moving methods from their owned to borrowed counterparts
 - Expanding `package.json` parsing to grab some of the information pruning requires
 - Porting of the prune command itself. I added some additional structure compared to the Go version, but not enough that comparing it to the Go version should be difficult

Notes for reviewers:
I apologize that this PR ended up touching as much as it did. Reviewing the PR by commit should at least make all of the changes and their impacts obvious. Commits before `2d4154c` are already on main and can be skipped.

### Testing Instructions

Existing unit tests and integration tests for file copying and package graph traversal. Actual lockfile behavior is mostly covered by unit tests that were ported when the lockfile were ported.

Also did various manual testing with pruning monorepos.


crates/turborepo-lib/src/commands/prune.rs
```diff
@@ -0,0 +1,267 @@
...
+use super::CommandBase;
+use crate::{
+    config::RawTurboJSON,
+    package_graph::{PackageGraph, WorkspaceName, WorkspaceNode},
+    package_json::PackageJson,
+    ui::BOLD,
+};
+
+// Files that should be copied from root and if they're required for install
+// All paths should be given as relative unix paths
```
NicholasLYang, line 20: Could we make this lazy_static and make the type RelativeUnixPathBuf?

NicholasLYang (commented):
I'm probably gonna need another pass to really grok the important parts, but the PR looks pretty solid!

crates/turborepo-paths/src/relative_unix_path.rs
```diff
@@ -29,18 +29,18 @@ impl RelativeUnixPath {
         Ok(unsafe { &*(path as *const str as *const Self) })
     }
 
-    pub(crate) fn to_system_path_buf(&self) -> Result<Utf8PathBuf, PathError> {
+    pub(crate) fn to_system_path_buf(&self) -> Utf8PathBuf {
```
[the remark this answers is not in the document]
gsoltis: Yes, we almost certainly should not need to do this conversion. If we're doing something with the filesystem, we should be converting to an absolute system path first.

crates/turborepo-lib/src/commands/prune.rs
```diff
@@ -0,0 +1,317 @@
...
+            turborepo_fs::copy_file(&from_path, docker_to)?;
+        }
+        Ok(())
+    }
+
+    fn copy_workspace(&self, package_json_path: &AnchoredSystemPathBuf) -> Result<()> {
+        let package_json_path = self.root.resolve(package_json_path);
+        let original_dir = package_json_path
+            .parent()
+            .ok_or_else(|| anyhow!("turbo doesn't support workspaces at file system root"))?;
```
NicholasLYang, line 255: IMO we should avoid writing new code with anyhow. Could we use thiserror here?

NicholasLYang (approved):
Logic looks good (although I'd be lying if I said I fully understood the complete nuances of lock files). I'd mostly love a reference, ideally in each package manager module, that gives a brief overview of the lockfile architecture and terminology for that specific package manager, i.e. berry has locators and descriptors.

Issue #5462, opened 2023-07-05
[Turborepo] Add a with-docker-pnpm example


jvzaniolo:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

`pnpm` is the package manager of the other examples so it's weird that the `with-docker` example uses `yarn`.
It would be nice to have a `with-docker-pnpm` example too.

### Describe the solution you'd like

The solution should use [`pnpm fetch`](https://pnpm.io/cli/fetch) which is the recommended way to use `pnpm` with docker 

### Describe alternatives you've considered

I was able to create a version with `pnpm` but it doesn't use `pnpm fetch` so I don't think it's the best way to use it

jvzaniolo:
Here's my working version with `pnpm`. I just copied the example with `yarn` and replaced it with `pnpm`

```Dockerfile
FROM node:18-alpine AS base

FROM base AS builder
RUN apk add --no-cache libc6-compat
RUN apk update
# Set working directory
WORKDIR /app
RUN npm install -g turbo
COPY . .
RUN turbo prune --scope=docs --docker

# Add lockfile and package.json's of isolated subworkspace
FROM base AS installer
RUN apk add --no-cache libc6-compat
RUN apk update
WORKDIR /app

# First install the dependencies (as they change less often)
COPY --from=builder /app/out/json/ .
COPY --from=builder /app/out/pnpm-lock.yaml ./pnpm-lock.yaml
RUN corepack enable
RUN pnpm install --frozen-lockfile

# Build the project
COPY --from=builder /app/out/full/ .
RUN pnpm dlx turbo run build --filter=docs

FROM base AS runner
WORKDIR /app

# Don't run production as root
RUN addgroup --system --gid 1001 nodejs
RUN adduser --system --uid 1001 nextjs
USER nextjs

COPY --from=installer /app/apps/docs/next.config.js .
COPY --from=installer /app/apps/docs/package.json .

# Automatically leverage output traces to reduce image size
# https://nextjs.org/docs/advanced-features/output-file-tracing
COPY --from=installer --chown=nextjs:nodejs /app/apps/docs/.next/standalone ./
COPY --from=installer --chown=nextjs:nodejs /app/apps/docs/.next/static ./apps/docs/.next/static
COPY --from=installer --chown=nextjs:nodejs /app/apps/docs/public ./apps/docs/public

CMD node apps/docs/server.js
```

edouardr:
[...]

Issue #534, opened 2022-01-10
Implementing turbo prune command for pnpm workspaces 


Aarush-Goyal:
### What version of Turborepo are you using?

1.0.24

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Linux

### Describe the Bug

I was running the command `turbo prune --scope=web` for a pnpm workspace using turborepo

And it returns this error message:

```
 ERROR  this command is not yet implemented for nodejs-pnpm
```

### Expected Behavior

It should have trimmed the workspace with relation to the scope declared which would make it easier for me to build the container image for that particular app.

### To Reproduce

You can create a simple turborepo that uses pnpm as the package manager like the pnpm example https://github.com/vercel/turborepo/tree/main/examples/with-pnpm

and try to run the command:
```
turbo prune --scope=web
```
I am running the pnpm version `6.24.4` node version `v17.3.0` turbo version `1.0.24`

AlonMiz:
can be based on this solution
https://www.npmjs.com/package/pnpm-isolate-workspace

Aarush-Goyal:
Thanks @AlonMiz and @ivoilic for your contributions.

iiian:
Can't wait to have this feature.

Aarush-Goyal:
@jaredpalmer can you please put this issue at a higher priority? It would make working with pnpm and turbo very easy.

iiian:
DISCLAIMER: This might be terrible advice.

In the meantime while we wait for `pnpm`-based `turbo prune`, you can use `pnpm` locally for blazing fast package management at dev-time while still using `turbo prune` for docker builds, if that is something you're interested in doing. Steps I'm taking to achieve this:

1. Fork your root `package.json` file. Maybe something like `package.docker-build.json`.
2. Update your `"packageManager"` property in `package.json` to target your `pnpm@x.y.z` (eg 6.32.4).
3. Create a `pnpm-workspace.yaml` as [explained in the pnpm documentation](https://pnpm.io/pnpm-workspace_yaml).
4. Update your `.dockerignore`, ymmv. Make sure you add `package.json` and `package.docker-build.json`, because the next step 👇 .
5. In your `Dockerfile`, `COPY package.docker-build.json package.json` at the proper moment in your context transfer.

Potential Drawbacks:
[...]

PR #5536, opened 2023-07-15
Add `examples/with-docker-pnpm` sample of using pnpm and docker

nicu-chiciuc:
### Description
Based on:
* https://github.com/vercel/turbo/issues/5462 
* https://github.com/vercel/turbo/issues/534

This PR adds a new example of using docker pnpm.

I've tried to keep the project code and structure similar to the existing `examples/with-docker` project with some updated goals:
* The `api` app is **strongly typed**, but uses standard ESM instead of relying on a build step. Additional type information can be added using JsDoc. 
* The Dockerfile might not make the best performance decision (I assume), but I've kept it simple enough so that I could understand it. Initially I didn't even use the node alpine version, but switched to it afterwards.
* I've removed most next config configurations and left just these:
```
export default {
  reactStrictMode: true,
  experimental: {
    appDir: true,
  },
};
```

I've also installed `syncpack` to check if there are any mismatches. I've added a package.json script for it, and it can be ran like this:
```
pnpm run syncpack
```
Unfortunately it also show errors for the local dependencies: https://github.com/JamieMason/syncpack/issues/121

but it will also show other mismatches, so I could pretty easily fix those.

### Testing Instructions
I've personally tested this by
1. cloning the repo
2. cd into `./examples/with-docker-pnpm`
3. Follow the instruction in the readme (`pnpm install`)

The previous step may not work by default (it didn't for me) because the root package.json has this:
```
  "engines": {
    "node": "18.x",
    "pnpm": "7"
  }
```
And my local pnpm was `8.6.5`.
While writing this PR I just removed that requirement for pnpm.

This doesn't seem like a nice solution.

A simpler solution would be to copy the `with-docker-pnpm` folder into a different folder and follow the instruction in the README. This seems like a better solution since this is how a potential user would use the example.

After copying the folder, run:
```
pnpm install
pnpm run dev
```

and/or

```
docker compose build
docker compose up
```

 both methods should work.


EugeneJeonME:
I'm building this way too, and it's a bit frustrating.
Does you know how to build with the 'standalone' option?

JamieMason:
> Unfortunately it also show errors for the local dependencies: [JamieMason/syncpack#121](https://github.com/JamieMason/syncpack/issues/121)

@nicu-chiciuc there's a good chance you'll need to add some configuration to get what you need. I added an [example in this comment](https://github.com/JamieMason/syncpack/issues/121#issuecomment-1641898993) but your scenario might be slightly different, I don't know yet, feel free to open an issue where I can try and help work it out with you.

nicu-chiciuc:
I will close this branch for now. I think it may serve as a good starter example for someone, even if it's not merged in the official package.

PR #5538, opened 2023-07-15
fix(turborepo): Corepack may have concurrency problems

gsoltis:
### Description

 - set examples to run with concurrency=1 so as not to hit corepack concurrency issues. If this proves to fix the problem, we can investigate more efficient fixes

### Testing Instructions

The examples workflow


gsoltis:
Alternate approach: https://github.com/vercel/turbo/pull/5539

gsoltis:
Closing in favor of #5539

PR #5553, opened 2023-07-17
feat(turborepo): Port CI detection

NicholasLYang:
### Description

Necessary for run summary/spaces porting.


crates/turborepo-ci/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+mod vendors;
+
+use std::env;
+
+use lazy_static::lazy_static;
+
+pub use crate::vendors::Vendor;
+use crate::vendors::VENDORS;
+
+lazy_static! {
```
arlyon, line 12: Can we use once_cell here instead? Saves a dep. Could use a function to get or initialize the value. In the long term we could use lazy_cell but we can wait until stabilized.

arlyon (changes requested):

crates/turborepo-ci/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        !build_id.is_empty()
+            || !build_number.is_empty()
+            || !ci.is_empty()
+            || !ci_app_id.is_empty()
+            || !ci_build_id.is_empty()
+            || !ci_build_number.is_empty()
+            || !ci_name.is_empty()
+            || !continuous_integration.is_empty()
+            || !run_id.is_empty()
+            || !teamcity_version.is_empty()
```
chris-olszewski, line 37: We can keep the short circuiting behavior of the Go code if we hold off on querying all of the env vars we care about
```suggestion

const CI_ENV_VARS: &[&str] = [
    "BUILD_ID",
    "BUILD_NUMBER",
    "CI",
    "CI_APP_ID",
    "CI_BUILD_ID",
    "CI_BUILD_NUMBER",
    "CI_NAME",
    "CONTINUOUS_INTEGRATION",
    "RUN_ID",
    "TEAMCITY_VERSION",
]
.as_slice();

pub fn is_ci() -> bool {
    *IS_CI.get_or_init(|| {
        // We purposefully don't do `is_err()` because the Go version
        // returns false for both an unset env variable
        // and an env variable set to the empty string.
        CI_ENV_VARS
            .iter()
            .any(|env_var| !env::var(env_var).unwrap_or_default().is_empty())
```

crates/turborepo-ci/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+                    }
+                }
+            } else if !env.env.all.is_empty() {
+                let mut all = true;
+                for env_var in &env.env.all {
+                    if env::var(env_var).unwrap_or_default().is_empty() {
+                        all = false;
+                        break;
+                    }
+                }
```
chris-olszewski, line 64:
```suggestion
                let all = env.env.all.iter().all(|env_var| {
                    !env::var(env_var).unwrap_or_default().is_empty()
                });
```

crates/turborepo-ci/src/vendors.rs
```diff
@@ -0,0 +1,582 @@
...
+    pub(crate) env: VendorEnvs,
+    pub(crate) eval_env: Option<HashMap<&'static str, &'static str>>,
+    pub(crate) sha_env_var: Option<&'static str>,
+    pub(crate) branch_env_var: Option<&'static str>,
+    pub(crate) username_env_var: Option<&'static str>,
+}
+
+static VENDORS: OnceLock<[Vendor; 45]> = OnceLock::new();
+
+pub(crate) fn get_vendors() -> [Vendor; 45] {
```
chris-olszewski, line 23: I think we can avoid the clones here if we return static references instead.
```suggestion
pub(crate) fn get_vendors() -> &'static [Vendor] {
```

crates/turborepo-ci/src/vendors.rs
```diff
@@ -0,0 +1,582 @@
...
+                        all: vec![],
+                    },
+                    eval_env: None,
+                    sha_env_var: None,
+                    branch_env_var: None,
+                    username_env_var: None,
+                },
+            ]
+        })
+        .clone()
```
chris-olszewski, line 581:
```suggestion
        .as_slice()
```

crates/turborepo-ci/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+            || !ci_name.is_empty()
+            || !continuous_integration.is_empty()
+            || !run_id.is_empty()
+            || !teamcity_version.is_empty()
+    })
+}
+
+impl Vendor {
+    // Returns info about a CI vendor
+    pub fn get_info() -> Option<Vendor> {
```
chris-olszewski, line 43:
```suggestion
    pub fn get_info() -> Option<&'static Vendor> {
```

crates/turborepo-ci/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        for env in get_vendors() {
+            if let Some(eval_env) = &env.eval_env {
+                for (name, expected_value) in eval_env {
+                    if matches!(env::var(name), Ok(env_value) if *expected_value == env_value) {
+                        return Some(env.clone());
```
chris-olszewski, line 48:
```suggestion
                        return Some(env);
```

crates/turborepo-ci/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+                    }
+                }
+            } else if !env.env.any.is_empty() {
+                for env_var in &env.env.any {
+                    if matches!(env::var(env_var), Ok(v) if !v.is_empty()) {
+                        return Some(env.clone());
```
chris-olszewski, line 54:
```suggestion
                        return Some(env);
```

crates/turborepo-ci/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+                let mut all = true;
+                for env_var in &env.env.all {
+                    if env::var(env_var).unwrap_or_default().is_empty() {
+                        all = false;
+                        break;
+                    }
+                }
+
+                if all {
+                    return Some(env.clone());
```
chris-olszewski, line 67:
```suggestion
                    return Some(env);
```

crates/turborepo-ci/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+            };
+
+            for env in set_env.iter() {
+                let mut env_parts = env.split('=');
+                let key = env_parts.next().unwrap();
+                let val = env_parts.next().unwrap_or("some value");
+                env::set_var(key, val);
+            }
+
+            assert_eq!(Vendor::get_info(), want);
```
chris-olszewski, line 179:
```suggestion
            assert_eq!(Vendor::get_info(), want.as_ref());
```

chris-olszewski (approved):
Logic looks good to me, only issues are with some of the unnecessary allocations which should be straightforward to fix.

arlyon (approved):

PR #5557, opened 2023-07-18
Ensure output assets reference only output assets

sokra:
### Description

Ensure output assets reference only output assets

next.js PR: https://github.com/vercel/next.js/pull/52832


alexkirsz (approved):
image
nice

PR #5561, opened 2023-07-19
Write the team ID to the repository config when a repo is linked to a Space

mknichel:
### Description

This PR writes the team ID to the repository configuration file when a user links the repository to a Space so that that team ID will be used in future requests.

### Testing Instructions

`npx turbo login`
`npx turbo link --target=spaces`
`less .turbo/config.json`

### DIFF
--- a/crates/turborepo-lib/src/commands/link.rs
+++ b/crates/turborepo-lib/src/commands/link.rs
@@ -247,6 +247,11 @@ pub async fn link(
                 )
             })?;
 
+            fs::create_dir_all(base.repo_root.join_component(".turbo"))
+                .context("could not create .turbo directory")?;
+            base.repo_config_mut()?
+                .set_team_id(Some(team_id.to_string()))?;
+
             println!(
                 "
     {} {} linked to {}


crates/turborepo-lib/src/commands/link.rs
```diff
@@ -247,6 +247,11 @@ pub async fn link(
                 )
             })?;
 
+            fs::create_dir_all(base.repo_root.join_component(".turbo"))
```
mknichel, line 250: @nathanhammond @gsoltis While testing, I discovered that `link --target=spaces` wasn't writing the selected team ID to `.turbo/config.json`. I ran into a situation where I had linked the repo for Remote Caching to another team and when I tried to enable Spaces, it wasn't generating any Runs data because it was trying to use the wrong team ID on subsequent requests. 

I'm wondering what you recommend to do. We could ask the user if they want to override the team ID if the user tries to select a team that is different than the one that is already configured for the repository, we could show an error if the team ID that is already configured is different than the selected one, or we could store the team IDs for remote cache and spaces in separate fields. Thoughts?

mknichel: Ping @nathanhammond @gsoltis

PR #5566, opened 2023-07-19
feat: port package graph validation

chris-olszewski:
### Description

Porting our cycle detection from Go to Rust

### Testing Instructions

Added new unit tests. Compare to [ValidateGraph](https://github.com/vercel/turborepo/blob/261aea9cb4f30204bef8d3f638326f9b2dd9e85a/cli/internal/util/graph.go#L12) in Go


crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -63,8 +64,44 @@ impl PackageGraph {
         PackageGraphBuilder::new(repo_root, root_package_json)
     }
 
-    pub fn validate(&self) -> Result<()> {
-        // TODO
+    pub fn validate(&self) -> Result<(), builder::Error> {
+        // This is equivalent to AcyclicGraph.Cycles from Go's dag library
+        let cycles = petgraph::algo::tarjan_scc(&self.workspace_graph)
```
NicholasLYang, line 69: Aren't strongly connected components a subset of cycles?

chris-olszewski: ["A directed graph is acyclic if and only if it has no strongly connected subgraphs with more than one vertex"](https://en.wikipedia.org/wiki/Strongly_connected_component#Definitions). This is also exactly what we do in Go. `AcyclicGraph.Cycles` uses a Go implementation of Tarjan's SCC algorithm [source](https://github.com/pyr-sh/dag/blob/master/tarjan.go#L7)

crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -63,8 +64,44 @@ impl PackageGraph {
...
+            .into_iter()
+            .filter(|cycle| cycle.len() > 1)
+            .collect::<Vec<_>>();
```
arlyon, line 72: We can probably just map here and use itertools::join? Same cost for the golden path, and one fewer alloc for the error case.

https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.join

```rust
let cycles_output = petgraph::algo::tarjan_scc(&self.workspace_graph)
            .into_iter()
            .filter(|cycle| cycle.len() > 1)
            .map(|cycle| {
                // format string
            })
            .join("\n");

if !cycles.is_empty() { // checking against a string instead
    return Err(builder::Error::CyclicDependencies(cycles_output))
}
```

arlyon (approved):
Minor perf item otherwise good!

PR #5568, opened 2023-07-19
feature(turbo): Global Hash

NicholasLYang:
### Description

Implements global hash. Stacked on top of #5824 

### Testing Instructions

Manually tested on a bunch of repos to confirm that hash matches between Rust and Go.

### DIFF
--- a/crates/turborepo-lib/src/run/global_hash.rs
+++ b/crates/turborepo-lib/src/run/global_hash.rs
@@ -12,7 +12,7 @@ use turborepo_scm::SCM;
 use crate::{
     cli::EnvMode,
     hash::{GlobalHashable, TurboHash},
-    package_json::PackageJson,
+    package_graph::WorkspaceInfo,
     package_manager::PackageManager,
 };
 
@@ -41,9 +41,8 @@ pub struct GlobalHashableInputs {
 
 #[allow(clippy::too_many_arguments)]
 pub fn get_global_hash_inputs<L: ?Sized + Lockfile>(
-    root_external_dependencies: HashSet<&turborepo_lockfiles::Package>,
+    root_workspace: &WorkspaceInfo,
     root_path: &AbsoluteSystemPath,
-    root_package_json: &PackageJson,
     package_manager: &PackageManager,
     lockfile: Option<&L>,
     global_file_dependencies: Vec<String>,
@@ -107,8 +106,7 @@ pub fn get_global_hash_inputs<L: ?Sized + Lockfile>(
         }
     }
 
-    let root_external_dependencies_hash =
-        root_package_json.get_external_deps_hash(root_external_dependencies);
+    let root_external_dependencies_hash = root_workspace.get_external_deps_hash();
 
     debug!(
         "rust external deps hash: {}",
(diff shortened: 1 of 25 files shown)


crates/turborepo-scm/src/package_deps.rs
```diff
@@ -2,13 +2,28 @@ use std::collections::HashMap;
...
 
 use crate::{hash_object::hash_objects, Error, Git, SCM};
 
 pub type GitHashes = HashMap<RelativeUnixPathBuf, String>;
 
 impl SCM {
+    pub fn get_hashes_for_files(
+        &self,
+        turbo_root: &AbsoluteSystemPath,
+        files: &[impl AsRef<AnchoredSystemPath>],
```
NicholasLYang, line 17: Made this AsRef because we have a slice of `AnchoredSystemPathBuf`'s in global hash

crates/turborepo-lib/src/run/global_hash.rs
```diff
@@ -1,72 +1,172 @@
...
+                    .map_or(false, |env| !env.is_empty()) =>
+            {
+                self.env_mode = EnvMode::Strict;
+            }
+            EnvMode::Loose => {
+                self.pass_through_env = None;
+            }
+            // Collapse `None` and `Some([])` to `Some([])` in strict mode
+            // to match Go behavior
+            EnvMode::Strict if self.pass_through_env.is_none() => {
```
chris-olszewski, line 145: I'm pretty sure this is just an artifact of the Go version, but I don't think this branch is ever taken. The only time I can find `GlobalHashInputs` being constructed is in `get_global_hash_inputs` and that always populates `pass_through_env` with a `Some`

crates/turborepo-lib/src/run/global_hash.rs
```diff
@@ -1,72 +1,172 @@
...
+        let global_hashable = GlobalHashable {
+            global_cache_key: self.global_cache_key,
+            global_file_hash_map: self.global_file_hash_map,
+            root_external_dependencies_hash: self.root_external_dependencies_hash,
+            env: self.env,
+            resolved_env_vars: self
+                .resolved_env_vars
+                .map(|evm| evm.all.to_hashable())
+                .unwrap_or_default(),
+            pass_through_env: self.pass_through_env.unwrap_or_default(),
```
chris-olszewski, line 164: I've very confused about the purpose of `calculate_global_hash_from_inputs` is if we just collapse `None` and `Some(vec![])` into the same value. I think we want this field to be `Option<Vec<String>>` as well.

chris-olszewski (changes requested):
The only real blocker I see is the coercion of `None` and `Some(vec![])` with `pass_through_env` which seems to be different from Go

crates/turborepo-lib/src/run/global_hash.rs
NicholasLYang: yeah looking at the capnproto code I think it's okay if we just make it a `Vec<String>` because we collapse it anyways with the `unwrap_or_default`

crates/turborepo-lib/src/run/global_hash.rs
NicholasLYang: Actually I suspect it's okay to make it all `Vec<String>` because the Go hashing code will treat a nil slice as an empty slice when serializing to capnproto.

NicholasLYang: This was all left over from how we did hashing previously, where nil vs empty slice did matter

chris-olszewski (approved):

PR #5569, opened 2023-07-19
fix: pnpm alias workspace deps

### LINKED ISSUE #5441: [turborepo] The `^` char in `dependsOn` does not work when using `alias package name`

1.10.7

pnpm

Linux

### Describe the Bug

The `^` char in `dependsOn` does not work when using `alias package name`

My config
```json
    "build": {
      "dependsOn": ["^build"],
      "outputs": [".next/**", "!.next/cache/**", "dist/**"]
    },
```

My app dependency

```json
    "dependencies": {
       "ui": "workspace:@resolvex/ui@^",
    }
```

When I run `build` both the app and package run concurrently

But if I not using an alias it work correctly

### Expected Behavior

App wait util package done for `build`

### To Reproduce

Install dependency by using an alias

### Reproduction Repo

_No response_

chris-olszewski:
### Description

Fixes #5441 

Adds support for [referencing workspaces through aliases](https://pnpm.io/workspaces#referencing-workspace-packages-through-aliases) by properly resolving them to the correct workspace. Before we would mark a package as being an external dependency (or if the alias was a valid workspace depend on the incorrect one).

This PR now recognizes when `workspace:` dependency references a different package than the name that's used in the `package.json`.

Note for reviewers:
This probably isn't the cleanest solution in either Rust or Go, but while we need to maintain two codepaths this keeps the code roughly equivalent.

### Testing Instructions

Added unit tests on the Go side.

Tested manually with a repository where `web` specified it's dependency on `@scope/ui` as `"ui": "workspace:@scope/ui@*" and verified that:
 - `turbo run build`: `@scope/ui` finished building before building `web`, this hits the Go impl
 - `turbo prune --scope=web`: `@scope/ui` was included in the pruned repository, this hits the Rust impl

### DIFF
--- a/cli/internal/context/context.go
+++ b/cli/internal/context/context.go
@@ -247,6 +247,17 @@ func (c *Context) resolveWorkspaceRootDeps(rootPackageJSON *fs.PackageJSON, warn
 	return nil
 }
 
+type dependencySplitter struct {
+	workspaces map[string]*fs.PackageJSON
+	pkgDir     string
+	rootPath   string
+}
+
+func (d *dependencySplitter) isInternal(name, version string) bool {
+	item, ok := d.workspaces[name]
+	return ok && isWorkspaceReference(item.Version, version, d.pkgDir, d.rootPath)
+}
+
 // populateWorkspaceGraphForPackageJSON fills in the edges for the dependencies of the given package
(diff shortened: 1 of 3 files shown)


gsoltis (approved):
Thanks for tackling this.

PR #5574, opened 2023-07-20
chore: improve turbopack test snapshots and error recovery

ForsakenHarmony:
### Description
Issue snapshots now get saved as human-readable text instead of the `.dbg()` output

Execution eval errors get handled properly

### DIFF
--- a/crates/turbopack-tests/tests/execution.rs
+++ b/crates/turbopack-tests/tests/execution.rs
@@ -10,7 +10,7 @@ use std::{
 };
 
 use anyhow::{Context, Result};
-use turbo_tasks::{debug::ValueDebug, unit, Completion, TryJoinIterExt, TurboTasks, Value, Vc};
+use turbo_tasks::{unit, Completion, TryJoinIterExt, TurboTasks, Value, Vc};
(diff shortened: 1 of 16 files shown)


crates/turbopack-tests/tests/execution.rs
```diff
@@ -263,16 +263,28 @@ async fn run_test(resource: String) -> Result<Vc<RunTestResult>> {
...
+    let single = res
         .try_into_single()
         .await
-        .context("test node result did not emit anything")?
-    else {
-        panic!("Evaluation stream must yield SingleValue.");
+        .context("test node result did not emit anything")?;
+
+    let SingleValue::Single(bytes) = single else {
+        return Ok(RunTestResult {
```
jridgewell, line 276: Should this be returning an error instead, since we can't actually test the output?

jridgewell (approved):

sokra (approved):

crates/turbopack-tests/tests/execution.rs
ForsakenHarmony: The code calling this function checks if the results are empty, so it's not necessary.

We also can't collect issues if we return an error unfortunately.

PR #5582, opened 2023-07-20
Snapshot tests: canonicalize with dunce

wbinnssmith:
This uses dunce [0] to consistently canonicalize paths used in Turbopack's tests. Previously, `REPO_ROOT` used Windows UNC paths while the test path in snapshot tests did not. This caused an issue when using `string.strip_prefix()` to determine relative paths.

[0] https://crates.io/crates/dunce


ForsakenHarmony (approved):
was thinking it's probably this, but didn't have time to test

PR #5588, opened 2023-07-21
chore: add new types for task id and task name

chris-olszewski:
### Description

In Go we used strings to represent both task ids and task names. Since these are so similar and could easily get mixed up I'm changing these to be two distinct types to avoid mixups.

- `TaskName`: A task as it appears in the `pipeline` object in `turbo.json`. It can be either a normal task name (`build`) or a workspace specific task name (`app#build`)
- `TaskId`: A task as the task graph understands it. Any task names that aren't workspace specific will get a workspace associated with them: e.g. `build` will become `app#build` and `ui#build`

### Testing Instructions

Added some new unit tests. Existing unit tests.

### DIFF
--- a/crates/turborepo-lib/src/config/turbo.rs
+++ b/crates/turborepo-lib/src/config/turbo.rs
@@ -12,7 +12,7 @@ use crate::{
     config::Error,
     opts::RemoteCacheOpts,
     package_json::PackageJson,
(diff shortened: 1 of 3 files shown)


arlyon (approved):
There are a few places similar to this where we could improve here, so this is great.

PR #5597, opened 2023-07-24
WEB-1278: Turbopack build: Implement minification with swc minify

wbinnssmith:
This implements build minification for turbopack-cli build and on the node build chunking context, using swc minify. It required adapting swc's public minification code to produce `SourceMap`s to avoid generating and re-parsing source maps.

Test Plan:

Input: `src/index.js`:

```
let myVar = 3;
console.log("foo", myVar);
```

Output:

```
module.exports={"[output]/src/index.js (ecmascript)":function({r:__turbopack_require__,f:__turbopack_require_context__,i:__turbopack_import__,s:__turbopack_esm__,v:__turbopack_export_value__,n:__turbopack_export_namespace__,c:__turbopack_cache__,l:__turbopack_load__,j:__turbopack_dynamic__,g:global,__dirname,m:module,e:exports}){function(){console.log("foo",3)}.call(this)}};

//# sourceMappingURL=[output]_src_index_e254c5.js.map
```

- Verified source map matched input tokens.


crates/turbopack-build/src/chunking_context.rs
```diff
@@ -22,12 +24,37 @@ use crate::ecmascript::node::{
...
+    Copy,
+    Hash,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    TraceRawVcs,
+    Default,
+)]
+pub enum MinifyType {
```
jridgewell, line 41: Should this just be a boolean? Are we expecting more variants?

crates/turbopack-build/src/minify.rs
```diff
@@ -0,0 +1,145 @@
...
+use turbopack_ecmascript::ParseResultSourceMap;
+
+#[turbo_tasks::function]
+pub async fn minify(path: Vc<FileSystemPath>, code: Vc<Code>) -> Result<Vc<Code>> {
+    let original_map = *code.generate_source_map().await?;
+    let minified_code = perform_minify(path, code);
+
+    let merged = match (original_map, *minified_code.generate_source_map().await?) {
+        (Some(original_map), Some(minify_map)) => Some(Vc::upcast(
+            TracedSourceMap(original_map.trace(minify_map)).cell(),
```
jridgewell, line 26: Nit: Could `trace` just return a `TracedSourceMapVc`?

crates/turbopack-build/src/minify.rs
```diff
@@ -0,0 +1,145 @@
...
+        Some(*Box::new(Vc::upcast(
+            ParseResultSourceMap::new(cm, src_map_buf).cell(),
+        ))),
+    );
+
+    Ok(builder.build().cell())
+}
+
+// From https://github.com/swc-project/swc/blob/11efd4e7c5e8081f8af141099d3459c3534c1e1d/crates/swc/src/lib.rs#L523-L560
+fn emit(cm: Arc<SwcSourceMap>, program: Program) -> Result<(String, Vec<(BytePos, LineCol)>)> {
```
jridgewell, line 95: Nit: This is usually called `print`, and I was initially confused why we were emitting collectibles into the task graph.

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -99,6 +122,34 @@ impl<'a> From<sourcemap::Token<'a>> for Token {
...
+                dst_col: t.generated_column as u32,
+                dst_line: t.generated_line as u32,
+                name_id: !0,
+                src_col: !0,
+                src_line: !0,
+                src_id: !0,
+            },
+        })
+    }
+}
```
jridgewell, line 151: Leave comments explaining the `!0`

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -190,6 +241,16 @@ impl SourceMap {
...
     }
 
+    #[turbo_tasks::function]
+    pub async fn tokens(self: Vc<Self>) -> Result<Vc<Tokens>> {
+        let crate_map = match &*self.await? {
+            Self::Regular(m) => m.deref().deref().deref().to_owned(),
+            Self::Sectioned(m) => m.flatten().await?.deref().deref().deref().to_owned(),
+        };
+
+        Ok(Vc::cell(crate_map.tokens().map(|t| t.into()).collect()))
```
jridgewell, line 251: This is 2 clones of the tokens (one of the full map, then another as we convert into `Token`. Do we need to clone anything here? Can't we just refer to the original's tokens?

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -199,20 +260,20 @@ impl SourceMap {
...
+                    .map(Token::from)
+            }
 
             SourceMap::Sectioned(map) => {
                 let len = map.sections.len();
                 let mut low = 0;
                 let mut high = len;
                 let pos = SourcePos { line, column };
 
-                // A "greatest lower bound" binary search. We're looking for the closest section
```
jridgewell, line 215: Please readd this.

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -245,10 +306,78 @@ impl SourceMap {
         };
         Ok(OptionToken(token).cell())
     }
+
+    #[turbo_tasks::function]
+    pub async fn trace(self: Vc<Self>, other: Vc<SourceMap>) -> Result<Vc<SourceMap>> {
+        let own_map = match &*self.await? {
+            Self::Regular(m) => m.clone(),
+            Self::Sectioned(m) => m.flatten().await?,
```
jridgewell, line 314: Do we need to clone?

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -245,10 +306,78 @@ impl SourceMap {
...
+        };
+
+        let mut builder = sourcemap::SourceMapBuilder::new(own_map.get_file());
+        let tokens = other
+            .tokens()
+            .await?
+            .iter()
+            .map(|other_token| async move {
+                let other_token = match other_token {
+                    Token::Synthetic(_) => panic!("Unexpected synthetic token"),
```
jridgewell, line 324: It's not unusual for a map to contain synthetic tokens, and we should preserve them.

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -245,10 +306,78 @@ impl SourceMap {
...
+                    Token::Original(t) => t,
+                };
+
+                Ok((
+                    (*self
+                        .lookup_token(other_token.original_line, other_token.original_column)
+                        .await?)
```
jridgewell, line 331: Can we avoid the VCs for this?

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -245,10 +306,78 @@ impl SourceMap {
...
+                        .clone(),
+                    other_token.clone(),
+                ))
+            })
+            .try_join()
+            .await?;
+
+        let mut source_to_src_id = IndexMap::new();
+        for (original_token, other_token) in tokens {
```
jridgewell, line 340:
```suggestion
        for (traced_token, other_token) in tokens {
```

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -245,10 +306,78 @@ impl SourceMap {
...
+            if let Some(original_token) = original_token {
+                match original_token {
+                    Token::Original(original_token) => {
+                        let token = builder.add(
```
jridgewell, line 344: You should be able to add raw tokens to the builder, right?

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -320,6 +449,41 @@ impl SectionedSourceMap {
...
+        for section in &self.sections {
+            let map = match &*section.map.await? {
+                SourceMap::Regular(m) => (***m).clone(),
+                SourceMap::Sectioned(m) => (***m.flatten().await?).clone(),
+            };
+
+            for token in map.tokens() {
+                let raw = builder.add(
+                    token.get_dst_line() + section.offset.line as u32,
+                    token.get_dst_col() + section.offset.column as u32,
```
jridgewell, line 467: `offset.column` only applies if `get_dst_line() === 0`

jridgewell (changes requested):

crates/turbopack-build/src/chunking_context.rs
wbinnssmith: afaik bool wouldn't allow a `true` default value, and this is more self-documenting.

crates/turbopack-core/src/source_map/mod.rs
wbinnssmith: This is part of the public turbopack SourceMap interface, so I figured it should return those tokens.

crates/turbopack-build/Cargo.toml
```diff
@@ -22,6 +22,8 @@ indoc = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 serde_qs = { workspace = true }
+sourcemap = "6.0.2"
+swc_core = { workspace = true, features = ["__parser", "ecma_minifier"] }
```
kdy1, line 26: Please add `ecma_minifier_concurrent` to the list of features.

https://github.com/swc-project/swc/blob/9daa0e2c6a6d31e8694d180408362d5f6883403e/crates/swc_core/Cargo.toml#L48-L49

kdy1 (changes requested):

crates/turbopack-build/src/chunking_context.rs
alexkirsz: Agreed, use enums instead of booleans where it makes more sense.

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -320,6 +478,63 @@ impl SectionedSourceMap {
...
         Self { sections }
     }
+
+    pub async fn source_contents(&self) -> Result<Vec<(String, Option<String>)>> {
+        let mut sources = IndexMap::new();
+
+        for section in &self.sections {
+            let map = section.map.await?;
+            for (source, contents) in map.source_contents().await? {
+                match sources.entry(source) {
```
jridgewell, line 488: This is just `sources.entry(source).or_insert(contents)`, or really just `sources.insert(source, contents)`?

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -320,6 +478,63 @@ impl SectionedSourceMap {
...
+                // Derived from https://github.com/getsentry/rust-sourcemap/blob/f1b758a251a5edddc0767f58f8391912c062c889/src/types.rs#L946
+                let generated_line = token.generated_line() + section.offset.line;
+                let generated_column = token.generated_column()
+                    + if token.generated_line() == 0 {
+                        section.offset.column
+                    } else {
+                        0
+                    };
+
+                tokens.push(match token {
```
jridgewell, line 519: We need to know if this token is out-of-range (it crosses into the next section), and we can skip processing the rest of the tokens in this section if so. The correct implementation is https://github.com/jridgewell/trace-mapping/blob/1abee9ca32e18f075da802a015019118fcfe9484/src/any-map.ts#L48-L155, or an optimal implementation at https://github.com/jridgewell/trace-mapping/blob/8af9820e129906ca7ef29e58a7fe196e1d671c6d/src/any-map.ts#L49-L158

jridgewell (commented):
`RegularSourceMap` still needs to capture `sources_content` during it's `::new` so that we don't hit a panic if the struct is moved.

wbinnssmith:
> `RegularSourceMap` still needs to capture `sources_content` during it's `::new` so that we don't hit a panic if the struct is moved.

It should as of 817a435

jridgewell (approved):

arlyon (approved):
Approved on behalf of repo

PR #5615, opened 2023-07-26
fix(docs): remove non-existent options

QuiiBz:
### Description

This PR removes non-existent options in `link` and `login` commands. I've also noticed that `link` has some un-documented options (`--no-gitignore` and `--target`). Is this intended because unstable, or should I add the missing docs?

https://github.com/vercel/turbo/blob/32b6e5b3b8e0f460828a691f2fdff89fe82f005a/crates/turborepo-lib/src/cli.rs#L310

### Testing Instructions

Nothing to test

### DIFF
--- a/docs/pages/repo/docs/reference/command-line-reference/link.mdx
+++ b/docs/pages/repo/docs/reference/command-line-reference/link.mdx
@@ -7,11 +7,3 @@ description: Turborepo CLI Reference for `link` command
 
 Link the current directory to Remote Cache scope. The selected owner (either a user or and organization) will be able to share [cache artifacts](/repo/docs/core-concepts/caching) through [Remote Caching](/repo/docs/core-concepts/remote-caching).
 You should run this command from the root of your monorepo.
-
-### Options
-
-#### `--api`
-
-`type: string`
-
-Defaults to `https://api.vercel.com`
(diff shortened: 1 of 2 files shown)


docs/pages/repo/docs/reference/command-line-reference/login.mdx
```diff
@@ -7,20 +7,6 @@ description: Turborepo CLI Reference for `login` command
...
-
-`type: string`
-
-Defaults to `https://vercel.com/`.
-
-#### `--api`
-
-`type: string`
-
-Defaults to `https://vercel.com/api`.
```
mehulkar, line 22: `--api` should be a real option, but I just tried and it doesn't seem to do anything 🤔

docs/pages/repo/docs/reference/command-line-reference/link.mdx
```diff
@@ -7,11 +7,3 @@ description: Turborepo CLI Reference for `link` command
 
 Link the current directory to Remote Cache scope. The selected owner (either a user or and organization) will be able to share [cache artifacts](/repo/docs/core-concepts/caching) through [Remote Caching](/repo/docs/core-concepts/remote-caching).
 You should run this command from the root of your monorepo.
-
-### Options
-
-#### `--api`
```
mehulkar, line 13: I believe this is intended to be a real option

mehulkar (changes requested):
I'm not 100% sure, but I think these options are how you implement a custom remote cache, so I think they're intended to be real.

docs/pages/repo/docs/reference/command-line-reference/login.mdx
QuiiBz: Agree that it's also something that I'm expecting to work, but I wasn't able to make to work either when I created this PR.

PR #5624, opened 2023-07-27
fix tracing span names

sokra:
### Description

The Vc update broke some tracing span names

### DIFF
--- a/crates/turbopack-convert-trace/src/main.rs
+++ b/crates/turbopack-convert-trace/src/main.rs
@@ -30,7 +30,6 @@ use std::{
 
 use indexmap::IndexMap;
 use intervaltree::{Element, IntervalTree};
-use itertools::Itertools;
 use turbopack_cli_utils::tracing::{TraceRow, TraceValue};
 
 macro_rules! pjson {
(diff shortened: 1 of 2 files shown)


crates/turbopack-convert-trace/src/main.rs
```diff
@@ -411,6 +430,95 @@ fn main() {
...
+                        ) {
+                            for item in items {
+                                match item {
+                                    SpanItem::SelfTime { .. } => {
+                                        self_items.push(item);
+                                    }
+                                    SpanItem::Child(id) => {
+                                        // SAFETY: We never mutate the `name` of the spans or the
+                                        // Vec itself. We only mutate the `items` Vec.
+                                        let key = unsafe { &*(&*spans[id].name as *const str) };
```
alexkirsz, line 455: Do we really need to use unsafe here? Is there a measurable performance benefit? Otherwise, let's keep it to safe Rust only.

alexkirsz (dismissed):

crates/turbopack-convert-trace/src/main.rs
sokra: Rust will complain about mutable borrow otherwise, since we `take(&mut spans[id].items)` later. So we can't keep a immutable borrow on `span[id].name` while mutating `spans[id].items`.

The alternative would be to clone the name.

alexkirsz: Is there a performance issue with cloning the name, or using an `Arc<String>` here?

sokra: I measured no difference when using clone instead. So it's now safe.

crates/turbopack-convert-trace/src/main.rs
```diff
@@ -411,6 +448,92 @@ fn main() {
...
+                            items: Vec<SpanItem>,
+                        ) {
+                            for item in items {
+                                match item {
+                                    SpanItem::SelfTime { .. } => {
+                                        self_items.push(item);
+                                    }
+                                    SpanItem::Child(id) => {
+                                        let key = spans[id].name.clone();
+                                        if key == parent_name {
```
jridgewell, line 472: Does this also imply `id == parent_id`? Or does each span have a unique id regardless of the called function?

jridgewell (approved):

crates/turbopack-convert-trace/src/main.rs
sokra: The same function can be called multiple times (different ids) and it would have the same name.

PR #5625, opened 2023-07-27
fix: capnp message size allocation

NicholasLYang:
### Description

Per [this issue](https://github.com/capnproto/capnproto-rust/issues/424), we should explicitly allocate the right amount of space for each message to avoid running out of space.

### Testing Instructions

Manually tested the global hash on multiple real world repos.


crates/turborepo-lib/src/hash/mod.rs
```diff
@@ -130,7 +130,9 @@ impl From<LockFilePackages> for Builder<HeapAllocator> {
             }
         }
 
-        let mut canon_builder = Builder::new(HeapAllocator::default());
+        let size = builder.total_size().unwrap().word_count + 1;
```
gsoltis, line 133: Can you handle the error or leave a comment as to why the `.unwrap()` is safe?

gsoltis (changes requested):

crates/turborepo-lib/src/hash/mod.rs
NicholasLYang: We already have multiple unwraps in this code. We could switch this to a `TryFrom`, but to be honest, if this code fails, it's a fatal programmer error, so I don't feel too bad about unwrapping here. I'll switch it to an `expect`.

gsoltis (approved):
Can you leave a comment about which scenarios can cause a failure here? Why would we be unable to calculate total size? Is there a pathological message type or construction method? Explain why we are sure we won't hit this.

NicholasLYang:
Sure, I guess I'm confused because there's multiple unwraps/expects in this code already. Why is this one special? In any case, if we can't calculate the global hash, there's no real recovery we can do.

PR #5626, opened 2023-07-28
remove require.cache clear from chunk loading

sokra:
### Description

This would memory leak, since not properly cleared from cache.
Multiple turbopack runtimes should be able to share require.cache of chunks

It the responsibility of the application that clear require.cache if files do change.


crates/turbopack-ecmascript-runtime/js/src/dev/runtime/nodejs/runtime-backend-nodejs.ts
```diff
@@ -85,7 +85,6 @@ let BACKEND: RuntimeBackend;
     const resolved = require.resolve(
       "./" + path.relative(path.dirname(fromChunkPath), chunkPath)
     );
-    delete require.cache[resolved];
```
alexkirsz, line 88: This might break `NextMode::DevServer` reloading in SSR.

alexkirsz (changes requested):

crates/turbopack-ecmascript-runtime/js/src/dev/runtime/nodejs/runtime-backend-nodejs.ts
sokra: It should be in a fresh process

alexkirsz (approved):

PR #5628, opened 2023-07-28
Port filter and scope to rust

arlyon:
### Description

This ports over filter and scope to rust, roughly matching the go implementation with some minor deviations where it makes sense for rustiness. All the tests are ported as-is and should pass. Note that this adds a print line to the run stub which prints out the filtered packages.

### Testing Instructions

Two options. Run `cargo test` in the `turborepo-lib` package, or, to test out the real SCM implementation, run it on real projects by building it and performing prunes.

### DIFF
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -144,9 +144,56 @@ fn parse_concurrency(concurrency_raw: &str) -> Result<u32> {
     }
 }
 
+// LegacyFilter holds the options in use before the filter syntax. They have
+// their own rules for how they are compiled into filter expressions.
+#[derive(Debug)]
+pub struct LegacyFilter {
+    // include_dependencies is whether to include pkg.dependencies in execution (defaults to false)
+    include_dependencies: bool,
+    // skip_dependents is whether to skip dependent impacted consumers in execution (defaults to
+    // false)
+    skip_dependents: bool,
+    // entrypoints is a list of package entrypoints
+    entrypoints: Vec<String>,
+    // since is the git ref used to calculate changed packages
+    since: Option<String>,
+}
+
+impl LegacyFilter {
+    pub fn as_filter_pattern(&self) -> Vec<String> {
+        let prefix = if self.skip_dependents { "" } else { "..." };
+        let suffix = if self.include_dependencies { "..." } else { "" };
+        if self.entrypoints.is_empty() {
+            if let Some(since) = self.since.as_ref() {
+                vec![format!("[{}{}{}]", prefix, since, suffix)]
+            } else {
+                Vec::new()
+            }
+        } else {
+            let since = if let Some(since) = self.since.as_ref() {
+                format!("...{}", since)
+            } else {
+                "".to_string()
+            };
+            self.entrypoints
+                .iter()
+                .map(|pattern| {
+                    if pattern.starts_with("!") {
+                        pattern.to_owned()
+                    } else {
+                        format!("{}{}{}{}", prefix, pattern, since, suffix)
+                    }
+                })
+                .collect()
+        }
+    }
+}
+
 #[derive(Debug)]
 pub struct ScopeOpts {
     pub pkg_inference_root: Option<AnchoredSystemPathBuf>,
+    pub legacy_filter: LegacyFilter,
+    pub filter_patterns: Vec<String>,
 }
 
 impl<'a> TryFrom<&'a RunArgs> for ScopeOpts {
@@ -158,7 +205,17 @@ impl<'a> TryFrom<&'a RunArgs> for ScopeOpts {
             .as_ref()
             .map(AnchoredSystemPathBuf::from_raw)
             .transpose()?;
-        Ok(Self { pkg_inference_root })
+        let legacy_filter = LegacyFilter {
+            include_dependencies: args.include_dependencies,
+            skip_dependents: args.no_deps,
+            entrypoints: args.scope.clone(),
+            since: args.since.clone(),
+        };
+        Ok(Self {
+            pkg_inference_root,
+            legacy_filter,
+            filter_patterns: args.filter.clone(),
+        })
     }
 }
 
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -165,7 +165,7 @@ impl LegacyFilter {
         let suffix = if self.include_dependencies { "..." } else { "" };
         if self.entrypoints.is_empty() {
             if let Some(since) = self.since.as_ref() {
-                vec![format!("[{}{}{}]", prefix, since, suffix)]
+                vec![format!("{}[{}]{}", prefix, since, suffix)]
             } else {
                 Vec::new()
             }
@@ -240,5 +240,40 @@ impl ScopeOpts {
             self.legacy_filter.as_filter_pattern(),
         ]
         .concat()
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use test_case::test_case;
+
+    use super::LegacyFilter;
+
+    #[test_case(LegacyFilter {
+            include_dependencies: true,
+            skip_dependents: false,
+            entrypoints: vec![],
+            since: Some("since".to_string()),
+        }, &["...[since]..."])]
+    #[test_case(LegacyFilter {
+            include_dependencies: false,
+            skip_dependents: true,
+            entrypoints: vec![],
+            since: Some("since".to_string()),
+        }, &["[since]"])]
+    #[test_case(LegacyFilter {
+            include_dependencies: false,
+            skip_dependents: true,
+            entrypoints: vec!["entry".to_string()],
+            since: Some("since".to_string()),
+        }, &["entry...since"])]
+    fn basic_legacy_filter_pattern(filter: LegacyFilter, expected: &[&str]) {
+        assert_eq!(
+            filter.as_filter_pattern(),
+            expected
+                .into_iter()
+                .map(|s| s.to_string())
+                .collect::<Vec<_>>()
+        )
     }
 }
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -146,7 +146,7 @@ fn parse_concurrency(concurrency_raw: &str) -> Result<u32> {
 
 // LegacyFilter holds the options in use before the filter syntax. They have
 // their own rules for how they are compiled into filter expressions.
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct LegacyFilter {
     // include_dependencies is whether to include pkg.dependencies in execution (defaults to false)
     include_dependencies: bool,
@@ -178,7 +178,7 @@ impl LegacyFilter {
             self.entrypoints
                 .iter()
                 .map(|pattern| {
-                    if pattern.starts_with("!") {
+                    if pattern.starts_with('!') {
                         pattern.to_owned()
                     } else {
                         format!("{}{}{}{}", prefix, pattern, since, suffix)
@@ -193,7 +193,9 @@ impl LegacyFilter {
 pub struct ScopeOpts {
     pub pkg_inference_root: Option<AnchoredSystemPathBuf>,
     pub legacy_filter: LegacyFilter,
+    pub global_deps: Vec<String>,
     pub filter_patterns: Vec<String>,
+    pub ignore_patterns: Vec<String>,
 }
 
 impl<'a> TryFrom<&'a RunArgs> for ScopeOpts {
@@ -212,9 +214,11 @@ impl<'a> TryFrom<&'a RunArgs> for ScopeOpts {
             since: args.since.clone(),
         };
         Ok(Self {
+            global_deps: args.global_deps.clone(),
             pkg_inference_root,
             legacy_filter,
             filter_patterns: args.filter.clone(),
+            ignore_patterns: args.ignore.clone(),
         })
     }
 }
@@ -227,5 +231,15 @@ impl<'a> From<&'a RunArgs> for CacheOpts<'a> {
             workers: run_args.cache_workers,
             ..CacheOpts::default()
         }
+    }
+}
+
+impl ScopeOpts {
+    pub fn get_filters(&self) -> Vec<String> {
+        [
+            self.filter_patterns.clone(),
+            self.legacy_filter.as_filter_pattern(),
+        ]
+        .concat()
     }
 }
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -270,10 +270,7 @@ mod test {
     fn basic_legacy_filter_pattern(filter: LegacyFilter, expected: &[&str]) {
         assert_eq!(
             filter.as_filter_pattern(),
-            expected
-                .into_iter()
-                .map(|s| s.to_string())
-                .collect::<Vec<_>>()
+            expected.iter().map(|s| s.to_string()).collect::<Vec<_>>()
         )
     }
 }
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -170,11 +170,10 @@ impl LegacyFilter {
                 Vec::new()
             }
         } else {
-            let since = if let Some(since) = self.since.as_ref() {
-                format!("...{}", since)
-            } else {
-                "".to_string()
-            };
+            let since = self
+                .since
+                .as_ref()
+                .map_or_else(String::new, |s| format!("...{}", s));
             self.entrypoints
                 .iter()
                 .map(|pattern| {
--- a/crates/turborepo-lib/src/package_graph/mod.rs
+++ b/crates/turborepo-lib/src/package_graph/mod.rs
@@ -5,10 +5,7 @@ use std::{
 
 use anyhow::Result;
 use itertools::Itertools;
-use petgraph::{
-    visit::{depth_first_search, EdgeRef, Reversed},
-    Directed,
-};
+use petgraph::visit::{depth_first_search, EdgeRef, Reversed};
 use turbopath::{AbsoluteSystemPath, AnchoredSystemPath, AnchoredSystemPathBuf};
 use turborepo_lockfiles::Lockfile;
 
@@ -189,6 +186,7 @@ impl PackageGraph {
     /// a -> b -> c
     ///
     /// immediate_ancestors(c) -> {b}
+    #[allow(dead_code)]
     pub fn immediate_ancestors(
         &self,
         workspace: &WorkspaceNode,
@@ -214,6 +212,7 @@ impl PackageGraph {
     /// a -> b -> c (external)
     ///
     /// dependencies(a) = {b, c}
+    #[allow(dead_code)]
     pub fn dependencies<'a>(&'a self, node: &WorkspaceNode) -> HashSet<&'a WorkspaceNode> {
         let mut dependencies =
             self.transitive_closure_inner(Some(node), petgraph::Direction::Outgoing);
@@ -305,7 +304,7 @@ impl PackageGraph {
 
         let external_deps = self
             .workspaces()
-            .filter_map(|(name, info)| {
+            .filter_map(|(_name, info)| {
                 info.unresolved_external_dependencies.as_ref().map(|dep| {
                     (
                         info.package_path().to_unix().to_string(),
@@ -329,12 +328,12 @@ impl PackageGraph {
         } else {
             self.workspaces
                 .iter()
-                .filter(|(name, info)| {
+                .filter(|(_name, info)| {
                     closures.get(info.package_path().as_str())
                         != info.transitive_dependencies.as_ref()
                 })
-                .map(|(name, info)| match name {
-                    n => Some(n.to_owned()),
+                .map(|(name, _info)| match name {
+                    WorkspaceName::Other(n) => Some(WorkspaceName::Other(n.to_owned())),
                     // if the root package has changed, then we should report `None`
                     // since all packages need to be revalidated
                     WorkspaceName::Root => None,
--- a/crates/turborepo-lib/src/package_graph/mod.rs
+++ b/crates/turborepo-lib/src/package_graph/mod.rs
@@ -248,27 +248,23 @@ impl PackageGraph {
 
         let global_change = current_key != previous_key;
 
-        let (changed, root_change) = self
-            .workspaces
-            .iter()
-            .filter(|(name, info)| {
-                closures.get(info.package_json_path().as_str())
-                    != info.transitive_dependencies.as_ref()
-            })
-            .fold((vec![], false), |(mut changed, all), (name, info)| {
-                if WorkspaceName::Root.eq(name) {
-                    (changed, true)
-                } else {
-                    changed.push(name.to_owned());
-                    (changed, all)
-                }
-            });
-
-        Ok(if global_change || root_change {
-            self.workspaces.iter().map(|w| w.0.to_owned()).collect()
+        let changed = if global_change {
+            None
         } else {
-            changed
-        })
+            self.workspaces
+                .iter()
+                .filter(|(name, info)| {
+                    closures.get(info.package_path().as_str())
+                        != info.transitive_dependencies.as_ref()
+                })
+                .map(|(name, info)| match name {
+                    n => Some(n.to_owned()),
+                    WorkspaceName::Root => None,
+                })
+                .collect::<Option<Vec<WorkspaceName>>>()
+        };
+
+        Ok(changed.unwrap_or_else(|| self.workspaces.keys().cloned().collect()))
     }
 }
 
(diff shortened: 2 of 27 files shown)


crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -50,4 +64,1082 @@ impl PackageInference {
...
+                    // ourselves, then we should add ourselves to the list of packages
+                    walked_dependencies.insert(package);
+                } else if !selector.include_dependencies && !selector.include_dependents {
+                    // if we are neither including dependents or dependencies, then
+                    // add  to the list of cherry picked packages
+                    cherry_picked_packages.insert(package);
+                }
+            }
+
+            if no_selectors {
```
gsoltis, line 307: Nit: I think `selector_packages.is_empty()` is actually more descriptive here than pulling it out into a local variable, especially since the definition of `no_selectors` is a little far away.

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -50,4 +64,1082 @@ impl PackageInference {
...
+    fn filter_nodes_with_selector(
+        &self,
+        selector: &TargetSelector,
+    ) -> Result<HashSet<String>, ResolutionError> {
+        let mut entry_packages = HashSet::new();
+        let mut selector_used = false;
+
+        let path = self
+            .turbo_root
+            .join_unix_path(selector.parent_dir.to_unix().unwrap())
```
gsoltis, line 419: Here, and elsewhere, can you either remove the `.unwrap()`, or change it to an `.expect()` with documentation as to why we don't expect to hit it?

gsoltis (commented):
I'm only part way through reviewing. I'll defer to other reviewers for approval, but wanted to get my few comments in.

crates/turborepo-lib/src/run/scope/filter.rs
arlyon: I moved the case up and put in a continue

crates/turborepo-lib/src/run/scope/filter.rs
arlyon: I have a handful of unwraps I need to fix. I will get round to those asap.

NicholasLYang (dismissed):

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -0,0 +1,177 @@
...
+            if !found {
+                // if the file is not in any package, it must be in the root package
+                changed_packages.insert(ROOT_PKG_NAME.to_string());
+            }
+        }
+
+        Ok(changed_packages)
+    }
+
+    fn is_file_in_package(file: &AnchoredSystemPath, package_path: &AnchoredSystemPath) -> bool {
```
chris-olszewski, line 146: We're passing in the path to a workspace's `package.json` so if there's a changed `packages/foo/index.js` then this will fail since we're checking component equality against `packages/foo/package.json`.

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -50,4 +65,1046 @@ impl PackageInference {
...
+    ) -> Result<HashSet<String>, ResolutionError> {
+        let (include_selectors, exclude_selectors) =
+            selectors.into_iter().partition::<Vec<_>, _>(|t| !t.exclude);
+
+        let mut include = if !include_selectors.is_empty() {
+            self.filter_graph_with_selectors(include_selectors)?
+        } else {
+            self.pkg_graph
+                .workspaces()
+                .filter(|(name, _)| !name.to_string().eq(ROOT_PKG_NAME)) // the root package has to be explicitly included
```
chris-olszewski, line 230:
```suggestion
                .filter(|(name, _)| *name != &WorkspaceName::Root) // the root package has to be explicitly included
```

chris-olszewski (changes requested):
First pass. Main things I would like would be ports of the rest of the tests from `scope_test.go` that test the lockfile and file changes.

crates/turborepo-paths/src/anchored_system_path_buf.rs
```diff
@@ -196,6 +199,14 @@ impl AnchoredSystemPathBuf {
     pub fn components(&self) -> Utf8Components {
         self.0.components()
     }
+
+    pub fn join(&self, other: &AnchoredSystemPathBuf) -> AnchoredSystemPathBuf {
```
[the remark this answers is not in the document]
arlyon: Yes

crates/turborepo-lib/src/opts.rs
```diff
@@ -155,9 +155,58 @@ fn parse_concurrency(concurrency_raw: &str) -> Result<u32> {
...
+    since: Option<String>,
+}
+
+impl LegacyFilter {
+    pub fn as_filter_pattern(&self) -> Vec<String> {
+        let prefix = if self.skip_dependents { "" } else { "..." };
+        let suffix = if self.include_dependencies { "..." } else { "" };
+        if self.entrypoints.is_empty() {
+            if let Some(since) = self.since.as_ref() {
+                vec![format!("[{}{}{}]", prefix, since, suffix)]
```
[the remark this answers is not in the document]
arlyon: Looks like you're correct, nice spot. Added a test

crates/turborepo-paths/src/anchored_system_path_buf.rs
```diff
@@ -196,6 +199,14 @@ impl AnchoredSystemPathBuf {
...
+        Self(self.0.join(&other.0))
+    }
+
+    pub fn restore_anchor(&self, path: &AbsoluteSystemPath) -> AbsoluteSystemPathBuf {
```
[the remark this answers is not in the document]
arlyon: I was trying to emulate the go code as best as possible, but this is silly. I have just added a direct join using AsRef Utf8Path and a test. TY!

crates/turbopack-tests/tests/execution/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "tests-execution",
```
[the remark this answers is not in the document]
arlyon: I have a new PR for that, this change leaked in

crates/turborepo-paths/src/anchored_system_path_buf.rs
```diff
@@ -196,6 +199,14 @@ impl AnchoredSystemPathBuf {
...
+        path.join_unix_path(self.to_unix().unwrap()).unwrap()
+    }
```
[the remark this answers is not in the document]
arlyon: Fixed

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -82,7 +82,9 @@ impl Run {
         let scm = SCM::new(&self.base.repo_root);
 
         let _filtered_pkgs =
-            scope::resolve_packages(&opts.scope_opts, &self.base, &pkg_dep_graph, &scm)?;
+            scope::resolve_packages(&opts.scope_opts, &self.base.repo_root, &pkg_dep_graph, &scm)?;
+
+        println!("filtered_pkgs: {:?}", _filtered_pkgs);
```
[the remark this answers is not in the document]
arlyon: I am doing that now. At the time there were pieces that were still in progress that I think were blocking.

crates/turborepo-lib/src/package_manager/mod.rs
```diff
@@ -69,6 +69,20 @@ pub enum PackageManager {
     Yarn,
 }
 
+impl PackageManager {
```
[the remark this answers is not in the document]
arlyon: Resolved

crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -127,24 +140,78 @@ impl PackageGraph {
...
-    ) -> HashSet<&WorkspaceNode> {
-        let indexes = nodes
+    ) -> HashSet<&'a WorkspaceNode> {
+        self.transitive_closure_inner(nodes, &self.workspace_graph, Direction::Normal)
+    }
+
+    fn transitive_closure_inner<'a, 'b, I: IntoIterator<Item = &'b WorkspaceNode>>(
+        &'a self,
+        nodes: I,
+        graph: &'a Graph<WorkspaceNode, (), Directed, u32>,
```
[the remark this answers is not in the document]
arlyon: Calculating dependencies vs dependents is as simple as flipping the graph using an adaptor, so transitive_closure_inner takes a (potentially adapted) graph but retains the interior logic.

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -0,0 +1,177 @@
...
+use turbopath::{AbsoluteSystemPath, AnchoredSystemPath, AnchoredSystemPathBuf};
+use turborepo_scm::SCM;
+use wax::Pattern;
+
+use crate::{
+    package_graph::{PackageGraph, WorkspaceName},
+    run::task_id::ROOT_PKG_NAME,
+};
+
+pub trait PackageChangeDetector {
```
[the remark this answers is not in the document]
arlyon: Yes, this is the rust equivalent of the go 'changed_packages' function pointer. It is also only used for testing in that impl. The test impl is at the bottom of filter.rs

crates/turborepo-lib/src/run/scope/mod.rs
```diff
@@ -1,23 +1,29 @@
...
 
 pub fn resolve_packages(
     opts: &ScopeOpts,
-    base: &CommandBase,
-    pkg_graph: &package_graph::PackageGraph,
-    _scm: &SCM,
+    turbo_root: &AbsoluteSystemPath,
+    pkg_graph: &PackageGraph,
+    scm: &SCM,
 ) -> Result<HashSet<String>> {
```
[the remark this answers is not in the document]
arlyon: Yes, I think the WorkspaceName struct changed while working on this but that is a good change. Resolved.

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -0,0 +1,177 @@
...
+        let matcher = wax::Glob::new(lockfile_path.as_str())?;
+
+        if !changed_files.iter().any(|f| matcher.is_match(f.as_path())) {
+            return Ok((vec![], false));
+        }
+
+        // todo: implement once lockfile parsing is supported
+        // let previous_file = self.scm.previous_content(from_ref,
+        // &lockfile_path).unwrap(); let previous_lockfile = parse_lockfile();
+        // let additional_packages = changed_packages(prev_lockfile);
```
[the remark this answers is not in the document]
arlyon: Done

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -346,6 +347,12 @@ impl AbsoluteSystemPath {
     }
 }
 
+impl<'a> From<&'a AbsoluteSystemPath> for CandidatePath<'a> {
+    fn from(value: &'a AbsoluteSystemPath) -> Self {
+        CandidatePath::from(value.0.as_std_path())
+    }
+}
```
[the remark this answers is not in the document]
arlyon: Lets come back to this later, there may be a better place (potentially the trash lmao) for this.

crates/turborepo-lib/src/run/scope/change_detector.rs
arlyon: The other function seems to do essentially the same thing.

arlyon: Nice catch, I think at the time I thought that was the dir, not the file itself.

crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -236,6 +239,71 @@ impl PackageGraph {
...
+
+        let (changed, root_change) = self
+            .workspaces
+            .iter()
+            .filter(|(name, info)| {
+                closures.get(info.package_json_path().as_str())
+                    != info.transitive_dependencies.as_ref()
+            })
+            .fold((vec![], false), |(mut changed, all), (name, info)| {
+                if WorkspaceName::Root.eq(name) {
```
chris-olszewski, line 285: Total bikeshed, I prefer `match`/`matches!` compared to using `==` when we are checking against a specific variant
```suggestion
                if matches!(name, WorkspaceName::Root) {
```

crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -236,6 +239,71 @@ impl PackageGraph {
...
+        let global_change = current_key != previous_key;
+
+        let (changed, root_change) = self
+            .workspaces
+            .iter()
+            .filter(|(name, info)| {
+                closures.get(info.package_json_path().as_str())
+                    != info.transitive_dependencies.as_ref()
+            })
+            .fold((vec![], false), |(mut changed, all), (name, info)| {
```
chris-olszewski, line 284: On the Go side we iteratively `globalChange` (for the root package changing) primarily because `contains` [doesn't exist in Go](https://stackoverflow.com/a/10485970) and secondarily for short circuiting. Since we're not short circuiting here, I think we could simplify this by just collecting a vector of workspace names that had deps change. Afterwards just have `let root_change = changed.contains(&WorkspaceName::Root);`

crates/turborepo-lib/src/package_manager/mod.rs
```diff
@@ -495,6 +495,18 @@ impl PackageManager {
...
+    pub fn lockfile_path(&self, turbo_root: &AbsoluteSystemPath) -> AbsoluteSystemPathBuf {
+        let file_name = match self {
+            PackageManager::Berry => yarn::LOCKFILE,
+            PackageManager::Npm => npm::LOCKFILE,
+            PackageManager::Pnpm => pnpm::LOCKFILE,
+            PackageManager::Pnpm6 => pnpm::LOCKFILE,
+            PackageManager::Yarn => yarn::LOCKFILE,
+        };
+
+        turbo_root.join_component(file_name)
```
chris-olszewski, line 508:
```suggestion
        turbo_root.join_component(self.lockfile_name())
```

crates/turborepo-lockfiles/src/npm.rs
```diff
@@ -135,6 +138,16 @@ impl Lockfile for NpmLockfile {
     fn encode(&self) -> Result<Vec<u8>, crate::Error> {
         Ok(serde_json::to_vec_pretty(&self)?)
     }
+
+    fn global_change_key(&self) -> Vec<u8> {
+        let mut buf = vec![b'n', b'p', b'm', 0];
```
chris-olszewski, line 143: Is the null byte important?

crates/turborepo-lockfiles/src/npm.rs
```diff
@@ -135,6 +138,16 @@ impl Lockfile for NpmLockfile {
     fn encode(&self) -> Result<Vec<u8>, crate::Error> {
         Ok(serde_json::to_vec_pretty(&self)?)
     }
+
+    fn global_change_key(&self) -> Vec<u8> {
```
chris-olszewski, line 142: This is something that I'm not super passionate about, but I think for readability it might make more sense to use something like `serde_json::Value` here as it's slightly more ergonomic to construct.
e.g.
```
json!({
    "type": "npm",
    "version": self.lockfile_version,
    "requires": self.other.get("requires").cloned().unwrap_or(serde_json::Value::Null)
})
```

crates/turborepo-lockfiles/src/npm.rs
```diff
@@ -135,6 +138,16 @@ impl Lockfile for NpmLockfile {
...
+        let mut buf = vec![b'n', b'p', b'm', 0];
+        {
+            let mut writer = BufWriter::new(&mut buf);
```
chris-olszewski, line 145: I don't think `BufWriter` gets us anything here since we aren't wrapping up any I/O:
> BufWriter<W> can improve the speed of programs that make small and repeated write calls to the same file or network socket. It does not help when writing very large amounts at once, or writing just one or a few times. It also provides no advantage when writing to a destination that is in memory, like a [Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html)<u8>.

chris-olszewski (changes requested):
Overall I think this looks like it's in a great place. Only blocker from my perspective is getting the correct keys for the closures used in the global change code.

crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -236,6 +239,71 @@ impl PackageGraph {
...
+        } else {
+            return Err(ChangedPackagesError::NoLockfile);
+        };
+
+        let external_deps = self
+            .workspaces()
+            .filter_map(|(a, b)| {
+                b.unresolved_external_dependencies.as_ref().map(|b| {
+                    (
+                        a.to_string(),
```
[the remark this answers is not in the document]
arlyon: I probably mis-understood you first time round but I understand now. TY

crates/turborepo-lib/src/package_graph/mod.rs
arlyon: Looks much better now, TY

crates/turborepo-lockfiles/src/npm.rs
arlyon: No, I had it do a null byte just to ensure that there are no clashes if the prefixes overlapped.

crates/turborepo-lockfiles/src/npm.rs
arlyon: I also considered pulling in bincode which may be a little faster. I know readability is not great. I don't really have a great intuition for how hot these functions are but I am thinking it's called only a handful of times during a run so I am going to just use he macro.

crates/turborepo-lockfiles/src/npm.rs
arlyon: very unlikely situation now that the rest of the data is structured json, but whatever

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -198,6 +199,10 @@ impl AbsoluteSystemPath {
         ))
     }
 
+    pub fn join_path(&self, path: &AnchoredSystemPath) -> AbsoluteSystemPathBuf {
```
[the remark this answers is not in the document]
arlyon: Oops I added this and then rebased, used the new api, and forgot to remove

crates/turborepo-lib/src/opts.rs
```diff
@@ -220,6 +219,17 @@ impl<'a> TryFrom<&'a RunArgs> for ScopeOpts {
     }
 }
 
+impl<'a> From<&'a RunArgs> for CacheOpts<'a> {
+    fn from(run_args: &'a RunArgs) -> Self {
+        CacheOpts {
+            override_dir: run_args.cache_dir.as_deref(),
+            skip_filesystem: run_args.remote_only,
+            workers: run_args.cache_workers,
+            ..CacheOpts::default()
```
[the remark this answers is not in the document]
arlyon: Bad rebase and this snuck in, I have fixed it now

crates/turborepo-lib/src/run/scope/target_selector.rs
```diff
@@ -187,34 +187,26 @@ pub enum InvalidSelectorError {
 pub fn is_selector_by_location(
     raw_selector: &str,
 ) -> Option<Result<AnchoredSystemPathBuf, InvalidSelectorError>> {
-    if !raw_selector.starts_with('.') {
-        return None;
-    }
+    let mut raw_selector_inner = raw_selector;
```
chris-olszewski, line 190: I will not lie, this feels way too complicated for checking if `raw_selector` starts with `./`, `.\`, `../`, `..\` or if it is exactly `.`/`..`. (If it does something else, we need more comments) The Go code is a direct port of [JS](https://github.com/pnpm/pnpm/blob/fe322b67810dd6016bdc9ca636661a8e78710b4b/workspace/filter-workspace-packages/src/parsePackageSelector.ts#L63) and this feels like we're now different enough from JS that we should just embrace the difference and aim for readability. (Or go back to being a direct port of JS)


```
    let EXACT_MATCHES = [".", ".."];
    let PREFIXES = ["./", ".\\", "../", "..\\"];
    if EXACT_MATCHES.contains(&raw_selector)
        || PREFIXES
            .iter()
            .any(|prefix| raw_selector.starts_with(prefix))
    {
        Some(
            AnchoredSystemPathBuf::try_from(raw_selector)
                .map_err(|_| InvalidSelectorError::InvalidAnchoredPath(raw_selector.to_string())),
        )
    } else {
        None
    }
```

Could we get away with

crates/turborepo-lockfiles/src/npm.rs
chris-olszewski: This will only get called at most once during a run and only if the git based filter is used and there's a lockfile change across those commits i.e. it's not hot at all.

chris-olszewski, line 142: Is there a commit on your local machine that changes this function?

chris-olszewski (approved):
Approving because I think this is in a place where fixing forward anything is better than having this go through another round of reviews.

 - I do feel semi-strongly that our `isSelectorByLocation` port has gotten too complicated for what it does and we can simplify if we're not just a straight JS port anymore.
 - I think the `global_key` code can be a little simplified. Some of your comments lead me to believing that there's some additional commits coming that address some earlier feedback.

cli/internal/scope/filter/filter.go
```diff
@@ -110,14 +110,12 @@ func (r *Resolver) getFilteredPackages(selectors []*TargetSelector) (*SelectedPa
...
-			if err != nil {
-				return nil, err
-			}
-			return selected, nil
+	if len(allPackageSelectors) > 0 {
+		selected, err := r.filterGraph(allPackageSelectors)
+		if err != nil {
+			return nil, err
 		}
+		return selected, nil
```
chris-olszewski, line 118: Is this change necessary for the PR, if not it should be taken out.

crates/turborepo-lockfiles/src/npm.rs
arlyon: Oops, had that stashed

crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -271,27 +277,23 @@ impl PackageGraph {
...
+                        != info.transitive_dependencies.as_ref()
+                })
+                .map(|(name, info)| match name {
+                    n => Some(n.to_owned()),
+                    WorkspaceName::Root => None,
+                })
+                .collect::<Option<Vec<WorkspaceName>>>()
+        };
+
+        Ok(changed.unwrap_or_else(|| self.workspaces.keys().cloned().collect()))
```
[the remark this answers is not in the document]
arlyon: That should be handled by https://github.com/vercel/turbo/pull/5628/files#diff-a5b0a844af5c2cb27acb9290a561c3de7474e61882977056af5d79307ca18f20R346

I will add a comment to that effect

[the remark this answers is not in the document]
arlyon: That is handled by the `None` case in the short circuit. I added a comment for it

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -50,4 +65,1088 @@ impl PackageInference {
...
+    //             exclude_self: false,
+    //             include_dependencies: true,
+    //             name_pattern: "project-1".to_string(),
+    //             ..Default::default()
+    //         }
+    //     ],
+    //     None,
+    //     &["project-1", "project-2", "project-4"] ;
+    //     "select package with dependencies"
+    // )]
```
chris-olszewski, line 729: Are these meant to be commented out?

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -918,7 +918,11 @@ mod test {
             packages,
             expected
                 .iter()
-                .map(|s| WorkspaceName::Other(s.to_string()))
+                .map(|s| if ROOT_PKG_NAME.eq(*s) {
+                    WorkspaceName::Root
+                } else {
+                    WorkspaceName::Other(s.to_string())
+                })
```
chris-olszewski, line 925: Not a big deal, but we can just use the conversion logic.
```suggestion
                WorkspaceName::from(*s)
```

crates/turborepo-lockfiles/src/berry/mod.rs
```diff
@@ -551,18 +552,15 @@ impl Lockfile for BerryLockfile {
...
-                self.data
-                    .metadata
-                    .cache_key
-                    .as_deref()
-                    .unwrap_or("")
-                    .as_bytes(),
-            );
-        }
+
+        serde_json::to_writer(
```
chris-olszewski, line 556: Probably worth addressing the unused result lint

crates/turborepo-lockfiles/src/npm.rs
```diff
@@ -141,11 +141,15 @@ impl Lockfile for NpmLockfile {
 
     fn global_change_key(&self) -> Vec<u8> {
         let mut buf = vec![b'n', b'p', b'm', 0];
-        {
-            let mut writer = BufWriter::new(&mut buf);
-            writer.write_all(&self.lockfile_version.to_be_bytes()[..]);
-            serde_json::to_writer(writer, &self.other.get("requires"));
-        }
+
+        serde_json::to_writer(
```
chris-olszewski, line 145: Again, we should probably `expect` the result

PR #5633, opened 2023-07-28
Turbopack: Sort env vars for compatibility

wbinnssmith:
Windows sorts the environment variables returned, breaking snapshot tests.


crates/turbopack-env/src/asset.rs
```diff
@@ -42,15 +42,18 @@ impl Source for ProcessEnvAsset {
 impl Asset for ProcessEnvAsset {
     #[turbo_tasks::function]
     async fn content(&self) -> Result<Vc<AssetContent>> {
-        let env = self.env.read_all().await?;
+        let mut env = (*self.env.read_all().await?).clone();
```
ForsakenHarmony, line 45: Sorting should probably happen inside the `read_all` as that's already an `IndexMap` which should have the correct order

jridgewell (approved):

PR #5618, opened 2023-07-26
feat: basic wasm support for turbopack

ForsakenHarmony:
### Description
Adds basic support for loading WebAssembly files imported from ECMAScript modules.

I've also added some docs to the async module implementation.

Fixes WEB-316

PR #5636, opened 2023-07-28
feat: wasm text format and import support

ForsakenHarmony:


ForsakenHarmony:
* main
  * **PR #5618** <a href="https://app.graphite.dev/github/pr/vercel/turbo/5618" target="_blank">Graphite</a> 
    * **PR #5636** <a href="https://app.graphite.dev/github/pr/vercel/turbo/5636" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/5636?utm_source=stack-comment).

crates/turbopack/src/module_options/mod.rs
```diff
@@ -342,8 +342,7 @@ impl ModuleOptions {
             ModuleRule::new(
                 ModuleRuleCondition::any(vec![
                     ModuleRuleCondition::ResourcePathEndsWith(".wasm".to_string()),
-                    // TODO(WEB-1316): add support for `wat` files
-                    // ModuleRuleCondition::ResourcePathEndsWith(".wat".to_string()),
+                    ModuleRuleCondition::ResourcePathEndsWith(".wat".to_string()),
```
sokra, line 345: I would have expected that `.wat` is implemented as `SourceTransform` on top of normal WebAssembly.

crates/turbopack-wasm/src/lib.rs
sokra, line 1: Split this into sub modules

sokra (commented):
Public item should be documented

crates/turbopack-wasm/src/analysis.rs
```diff
@@ -0,0 +1,77 @@
...
+                    let export = export?;
+
+                    analysis.exports.push(export.name.to_string());
+                }
+            }
+
+            // skip over code sections
+            Payload::CodeSectionStart { size, .. } => {
+                parser.skip_section();
+                bytes = &bytes[size as usize..];
```
jridgewell, line 68: Does this double skip? We're already advancing `consumed` bytes, does `consumed` not include the range of the code section?

jridgewell (approved):

crates/turbopack-wasm/src/analysis.rs
ForsakenHarmony: No, because we tell the parser to skip the next section

weird API tbh, I would expect to give it the slice of bytes once and it to handle a cursor internally

PR #5652, opened 2023-08-01
feat: port task graph construction

chris-olszewski:
### Description

This PR ports the `Engine` structure from Go and its construction. It doesn't port the execution of the task graph or the validation of persistent task dependencies.

Major changes:
 - Introduction of `TaskId` and `TaskName` (originally in #5588, but kept here since this PR actually exercises the types)
 - We now parse task names as `TaskName` where applicable instead of `String`
 - Synthesizing a single task definition from multiple `turbo.json`, this is ported from Go in a very straightforward way. Post port, there's some work that can be done to greatly simplify the types involved with task definitions.

Reviewer notes: Saving `EngineBuilder` for last makes sense as all of the other changes are smaller and mostly contained to the file they happen in. The construction logic is almost 1:1 with the Go's `Engine.Prepare` logic including some of the oddities that it contains.

### Testing Instructions

Tests mostly come from [`engine_test.go`](https://github.com/vercel/turbo/blob/b8c55756a68779b85f695501ed7fdd644a8e7487/cli/internal/core/engine_test.go) unit tests that were removed. Additional unit tests were added, but we'll be relying on hooking this up to the rest of run and running against integration tests to make sure behavior is the same.


crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+impl From<TaskId<'static>> for TaskNode {
+    fn from(value: TaskId<'static>) -> Self {
+        Self::Task(value)
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct Building;
+#[derive(Debug, Default)]
+pub struct Built;
```
chris-olszewski, line 28: Build state is slightly unnecessary, but I do want to discourage mutable use of the engine after construction to simplify when we execute tasks.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -131,6 +138,31 @@ impl Run {
...
+        )
+        .with_root_tasks(root_turbo_json.pipeline.keys().cloned())
+        .with_turbo_jsons(Some(
+            Some((WorkspaceName::Root, root_turbo_json.clone()))
+                .into_iter()
+                .collect(),
+        ))
+        .with_tasks_only(opts.run_opts.only)
+        .with_workspaces(
+            filtered_pkgs
```
chris-olszewski, line 154: Eventually, I'd like for `filtered_pkgs` to be a set of `WorkspaceName` instead of `Strings`.

crates/turborepo-lib/src/engine/builder.rs
```diff
@@ -0,0 +1,985 @@
...
+                });
+            }
+            let task_definition = Rc::new(TaskDefinition::from_iter(self.task_definition_chain(
+                &mut turbo_jsons,
+                &task_id,
+                &task_id.as_non_workspace_task_name(),
+            )?));
+
+            // Skip this iteration of the loop if we've already seen this taskID
+            if visited.contains(&task_id) {
```
chris-olszewski, line 173: This check happens here in the Go code. I believe it can happen earlier, but I didn't want to shuffle around control flow during the port.

crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[derive(Debug)]
+pub struct Engine<S = Built> {
+    marker: std::marker::PhantomData<S>,
+    task_graph: Graph<TaskNode, ()>,
+    root_index: petgraph::graph::NodeIndex,
+    task_lookup: HashMap<TaskId<'static>, petgraph::graph::NodeIndex>,
+    task_definitions: HashMap<TaskId<'static>, Rc<TaskDefinition>>,
```
chris-olszewski, line 36: We could wrap up these `TaskId` in an `Rc` to prevent multiple allocations, but this introduces the question of where do we dedupe these to avoid multiple `Rc` of the same `TaskId`. Leaving this optimization for future perf work.

NicholasLYang (commented):
This looks good as far as I can tell, I'll give it another pass momentarily

crates/turborepo-lib/src/engine/builder.rs
```diff
@@ -0,0 +1,985 @@
...
+            let mut deps = task_definition
+                .task_dependencies
+                .iter()
+                .collect::<HashSet<_>>();
+            let mut topo_deps = task_definition
+                .topological_dependencies
+                .iter()
+                .collect::<HashSet<_>>();
+
+            if self.tasks_only {
```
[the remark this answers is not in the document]
chris-olszewski: Yeah, tasks only is too short and makes it sound like there could be non-tasks in the task graph.

crates/turborepo-lib/src/engine/builder.rs
```diff
@@ -0,0 +1,985 @@
...
+                });
+            }
+
+            if task_id.package() != ROOT_PKG_NAME
+                && self
+                    .package_graph
+                    .package_json(&WorkspaceName::from(task_id.package()))
+                    .is_none()
+            {
+                // If we have a pkg it should be in WorkspaceInfos.
```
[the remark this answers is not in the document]
chris-olszewski: Ah yeah, I was just copying the comments over from Go

PR #5665, opened 2023-08-03
ci: Getting turbo build to be cached

NicholasLYang:
### Description

Added a `build_turborepo` job that builds turborepo before running any jobs that require it. Since we're using turbo, this binary is saved to the remote cache and reused across runs. That way we don't have to rebuild turborepo multiple times.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -194,6 +194,34 @@ jobs:
       main_push: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
       tag_push: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/node-file-trace') }}
 
+  build_turborepo:
+    name: Build Turborepo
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -252,7 +280,7 @@ jobs:
 
   go_integration:
     name: Go Integration Tests
-    needs: determine_jobs
+    needs: build_turborepo
     if: needs.determine_jobs.outputs.go_integration == 'true'
```
tknickman, line 284: `determine_jobs` is now a transitive dependency for these which works but _could_ be confusing. An array [is supported here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-requiring-successful-dependent-jobs) so maybe it's worth specifying both? 

```yml
needs: [determine_jobs, build_turborepo]
```

tknickman (approved):

PR #5678, opened 2023-08-04
Update eslint.mdx

Publicker:


tknickman:
This is not required, as eslint automatically assumes the prefix, allowing the names to be a bit more succinct! 

https://eslint.org/docs/latest/extend/shareable-configs#using-a-shareable-config

PR #5684, opened 2023-08-07
feature(turborepo): `turborepo-ui` crate

NicholasLYang:
### Description

I need some of the UI functionality in `turborepo-cache`, so I factored out the `turborepo-ui` crate to avoid circular dependencies.

I also added the log replayer code (which as far as I can tell does not need to be a function value since we don't use anything other than `defaultLogReplayer` in the Go code) and `PrefixedUI`.

### Testing Instructions

Wrote a simple test for log replayer.

### DIFF
--- a/crates/turborepo-lib/Cargo.toml
+++ b/crates/turborepo-lib/Cargo.toml
@@ -102,10 +102,10 @@ turbo-updater = { workspace = true }
 turbopath = { workspace = true }
 turborepo-api-client = { workspace = true }
 turborepo-cache = { workspace = true }
+turborepo-ci = { workspace = true }
 turborepo-env = { workspace = true }
 turborepo-lockfiles = { workspace = true }
 turborepo-scm = { workspace = true }
-turborepo-ci = { workspace = true }
 turborepo-ui = { workspace = true }
 wax = { workspace = true }
 webbrowser = { workspace = true }
--- a/crates/turborepo-lib/Cargo.toml
+++ b/crates/turborepo-lib/Cargo.toml
@@ -105,6 +105,7 @@ turborepo-cache = { workspace = true }
 turborepo-env = { workspace = true }
 turborepo-lockfiles = { workspace = true }
 turborepo-scm = { workspace = true }
+turborepo-ci = { workspace = true }
 turborepo-ui = { workspace = true }
 wax = { workspace = true }
 webbrowser = { workspace = true }
--- a/crates/turborepo-lib/Cargo.toml
+++ b/crates/turborepo-lib/Cargo.toml
@@ -105,6 +105,7 @@ turborepo-cache = { workspace = true }
 turborepo-env = { workspace = true }
 turborepo-lockfiles = { workspace = true }
 turborepo-scm = { workspace = true }
+turborepo-ui = { workspace = true }
 wax = { workspace = true }
 webbrowser = { workspace = true }
 which = { workspace = true }
--- a/crates/turborepo-lib/src/run/mod.rs
+++ b/crates/turborepo-lib/src/run/mod.rs
@@ -4,6 +4,8 @@ mod global_hash;
 mod scope;
 pub mod task_id;
 
+use std::io::IsTerminal;
+
 use anyhow::{Context as ErrorContext, Result};
 use tracing::{debug, info};
 use turborepo_cache::{http::APIAuth, AsyncCache};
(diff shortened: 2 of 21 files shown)


crates/turborepo-ui/src/lib.rs
```diff
@@ -50,6 +50,10 @@ impl UI {
         Self { should_strip_ansi }
     }
 
+    pub fn is_ci() -> bool {
+        turborepo_ci::is_ci() && std::io::stdout().is_terminal()
+    }
```
chris-olszewski: This function is misplaced. Looking at Go since this is only used by `run.go`, I'd much rather remove `UI::is_ci` and change the usage in `run` to explicitly check if stdout is a terminal.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -61,7 +62,7 @@ impl Run {
         }
 
         // There's some warning handling code in Go that I'm ignoring
-        if self.base.ui.is_ci() && !opts.run_opts.no_daemon {
+        if UI::is_ci() && !opts.run_opts.no_daemon {
```
chris-olszewski, line 65: Remove the `UI::is_ci` variant to discourage others from depending on it. If we think this behavior should be used more I'd suggest we add a new `is_ci_and_no_tty`
```suggestion
        if turborepo_ci::is_ci() && !std::io::stdout().is_terminal() && !opts.run_opts.no_daemon {
```

crates/turborepo-ui/src/lib.rs
```diff
@@ -39,9 +50,8 @@ impl UI {
         Self { should_strip_ansi }
     }
 
-    // STUB
-    pub fn is_ci(&self) -> bool {
-        env::var("CI").is_ok()
+    pub fn is_ci() -> bool {
+        turborepo_ci::is_ci() && std::io::stdout().is_terminal()
```
chris-olszewski, line 54: I'm pretty sure we want this to be true when we don't have a tty
```suggestion
        turborepo_ci::is_ci() && !std::io::stdout().is_terminal()
```

chris-olszewski (changes requested):
Quick throwing these comments your way while I review the replay port. Only thing that really needs to get changed is negating `std::io::stdout().is_terminal()`

crates/turborepo-ui/src/lib.rs
NicholasLYang: Also is there any reason this isn't included in the general CI check?

crates/turborepo-ui/src/log_replayer.rs
```diff
@@ -0,0 +1,93 @@
...
+#[allow(dead_code)]
+impl<D: Display + Clone, W: Write> PrefixedUI<D, W> {
+    pub fn new(ui: UI, prefix: StyledObject<D>, output: W) -> Self {
+        Self { ui, prefix, output }
+    }
+
+    pub fn output(&mut self, message: impl Display) -> Result<(), Error> {
+        writeln!(
+            self.output,
+            "{} {}",
```
chris-olszewski, line 28: Looking at the Go implementation I see a few differences here that I'm not sure if they were intentional:
 - Go just concats the prefix and the message if the message is non-empty
 - If an empty message is passed then only an empty line will be written to the ui/writer
 - If when replaying the logs there's an empty line then we write `"{prefix}\n"`. This implementation would write `"{prefix} \n"`.

I'd much rather keep this (weird) behavior until we have all the ui/log code ported to avoid our future selves forgetting these differences when we port code that calls these functions.

chris-olszewski (changes requested):

crates/turborepo-ui/src/lib.rs
chris-olszewski: Not to my knowledge, not sure why we'd want to behave differently on CI depending on if we detect tty.

crates/turborepo-ui/src/log_replayer.rs
NicholasLYang: With the second point, we have a workaround in the log replayer code to print empty lines with the prefix. Since I'm writing the prefixed UI implementation, I decided to just have it do what we want and avoid the workaround. Looking at the other uses of prefixed UI, we don't ever print an empty message, so I think this is fine.

As for the other points, I can remove the space

crates/turborepo-lib/Cargo.toml
```diff
@@ -105,6 +105,7 @@ turborepo-cache = { workspace = true }
 turborepo-env = { workspace = true }
 turborepo-lockfiles = { workspace = true }
 turborepo-scm = { workspace = true }
+turborepo-ui = { workspace = true }
```
chris-olszewski, line 108:
```suggestion
turborepo-ci = { workspace = true }
turborepo-ui = { workspace = true }
```

crates/turborepo-ui/src/log_replayer.rs
```diff
@@ -0,0 +1,93 @@
...
+    output: W,
+}
+
+#[allow(dead_code)]
+impl<D: Display + Clone, W: Write> PrefixedUI<D, W> {
+    pub fn new(ui: UI, prefix: StyledObject<D>, output: W) -> Self {
+        Self { ui, prefix, output }
+    }
+
+    pub fn output(&mut self, message: impl Display) -> Result<(), Error> {
```
chris-olszewski, line 25: Since this differs from `PrefixUI.Output` in Go we should add a warning comment that it will output the prefix is passed an empty message unlike Go.

PR #5693, opened 2023-08-09
Added Vontigo as a client that uses Turbo

Smilefounder:
We are an open source community based in Vietnam primarily focused on the development of open source that using Svelte/ Sveltekit and Vercel's platforms & libraries. It has been some time since we incorporated Turbo into our systems and the results have been astounding. We would be quite honored if you could have our logo on your showcase.


arlyon:
I have approved CI and will have a look at the changes. Thanks for taking the time.

anthonyshew:
Thank you for using `turbo`! At this time, we're not looking to add any new logos.

PR #5694, opened 2023-08-09
Added Mixcore Studio as a client that uses Turbo

Smilefounder:
We are a Digital Agency based in Vietnam primarily focused on
the development of website and application for the top brands in Vietnam and worldwide. It has
been some time since we incorporated Turbo into our development systems and process and the
results have been astounding. We would be quite honored if you could have our logo on your showcase.


anthonyshew:
Thank you for using `turbo`! At this time, we're not looking to add any new logos.

PR #5695, opened 2023-08-09
fix(turborepo): Rationalize the install and execution process.

nathanhammond:
Previously we had a lot of complexity in our install scripts inherited from `esbuild`. This change removes _all_ of the incidental complexity in favor of relying upon default ecosystem behaviors in most cases.

Instead of attempting correction of configuration at both installation and invocation, it handles everything at invocation. Early returns ensure that we perform the exact same amount of work as we did in the previous variation.

It can now also detect and provide instructions on how to repair a broken `package-lock.json` caused by https://github.com/npm/cli/issues/4828. (Note: if the "just try installing" [saving throw](https://en.wikipedia.org/wiki/Saving_throw) works, we won't provide this information to the user. I elected this ordering for startup time reasons.)

As an added bonus, `turbo` no longer has a `postinstall` script which is more end-user-friendly.

Material changes:
- Does not attempt to extract turbo from a Yarn PnP without `preferUnplugged` support. This works with _all_ stable releases of `yarn` (berry). This is a change for users on canary versions >2.0.0-rc.0 && <=2.0.0-rc.30 from September 2019 – March 2020—but we already don't support in other portions of our codebase so this only changes where the error is thrown from.
- In the event of `npm install` download failure it does not attempt to directly download a tarball. If a user misconfigured their setup _and_ we didn't succeed at `npm install` to bail them out? It's okay for you to get an error with lots of details on how to fix your configuration.


packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
 
+const child_process = require('child_process');
+const fs = require('fs');
+const path = require('path');
+
+// If we do not find the correct platform binary, should we attempt to install it?
+const SHOULD_INSTALL = true;
+
+// If we do not find the correct platform binary, should we trust calling an emulated variant?
+const SHOULD_ATTEMPT_EMULATED = true;
```
nathanhammond, line 16: I left these in as configuration variables; but we can actually hardcode the decisions here.

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+
+  // Erase "npm_config_global" so that "npm install --global turbo" works.
+  // Otherwise this nested "npm install" will also be global, and the install
+  // will deadlock waiting for the global installation lock.
+  const env = { ...process.env, npm_config_global: undefined };
+
+  child_process.execSync(
+    `npm install --loglevel=error --prefer-offline --no-audit --progress=false`,
+    { cwd: turboPath, stdio: "pipe", env }
+  );
```
nathanhammond, line 31: Previously this had to carefully track which package and version, clean up after itself, and more.

Now we just rely on `npm` to do the right thing when looking at the `package.json` for `turbo`, clean up in the event of failure, etc.

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+}
+
+// This provides logging messages as it progresses towards calculating the binary path.
+function getBinaryPath() {
+  // First we see if the user has configured a particular binary path.
+  const TURBO_BINARY_PATH = process.env.TURBO_BINARY_PATH;
```
nathanhammond, line 37: We still need to allow the user to specify a particular binary path.

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+  const resolvedArch = arch === 'x64' ? '64' : arch;
+  const ext = platform === 'windows' ? '.exe' : '';
+
+  // Try all places in order until we get a hit.
+
+  // 1. The package which contains the binary we _should_ be running.
+  const correctBinary = availablePlatforms.includes(platform) && availableArchs.includes(resolvedArch) ? `turbo-${platform}-${resolvedArch}/bin/turbo${ext}` : null;
+  if (correctBinary !== null) {
+    try {
+      return require.resolve(`${correctBinary}`);
```
nathanhammond, line 73: By rule we should be able to `require.resolve()` the platform-specific package from within the `turbo` directory, so we do that. (1:1 with the existing behavior.)

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+
+  // 2. Install the binary that they need just in time.
+  if (SHOULD_INSTALL && correctBinary !== null) {
+    console.warn('Turborepo did not find the correct binary for your platform.');
+    console.warn('We will attempt to install it now.');
+
+    try {
+      installUsingNPM();
+      const resolvedPath = require.resolve(`${correctBinary}`);
+      console.warn('Installation has succeeded.');
```
nathanhammond, line 85: We don't succeed unless we can resolve it, so this gets split across lines.

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+      return resolvedPath;
+      } catch (e) {
+      console.warn('Installation has failed.');
+    }
+  }
+
+  // 3. Both Windows and macOS ARM boxes can run x64 binaries. Attempt to run under emulation.
```
nathanhammond, line 92: There is a chance that this triggers an unfriendly error or even pops up a "install Rosetta" GUI window, but that should be a very low-frequency outcome.

The primary cause for hitting this state is a user who has emulation set up on their box accidentally switching between `arm64` and `x64` versions of Node during upgrade.

Given the high odds of it working I'm comfortable doing this.

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+  const alternateBinary = (arch === "arm64" && ['darwin', 'windows'].includes(platform)) ? `turbo-${platform}-64/bin/turbo${ext}` : null;
+  if (SHOULD_ATTEMPT_EMULATED && alternateBinary !== null) {
+    try {
+      const resolvedPath = require.resolve(`${alternateBinary}`);
+      console.warn(`Turborepo detected that you're running:\n${platform} ${resolvedArch}.`);
+      console.warn(`We were not able to find the binary at:\n${correctBinary}`);
+      console.warn(`We found a possibly-compatible binary at:\n${alternateBinary}`);
+      console.warn(`We will attempt to run that binary.`);
```
nathanhammond, line 100: We also provide detailed output about what we're doing so that things going wrong are at least understandable.

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+      return resolvedPath;
+    } catch (e) {}
+  }
+
+  // We are not going to run `turbo` this invocation.
```
nathanhammond, line 105: [You will not go to space today.](https://www.explainxkcd.com/wiki/index.php/1133:_Up_Goer_Five)

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+  // Let's give the best error message that we can.
+
+  // Possible error scenarios:
+  // - The user is on a platform/arch combination we do not support.
+  // - We somehow got detection wrong and never attempted to run the _actual_ `correctBinary` or `alternateBinary`.
+  // - The user doesn't have the correct packages installed for their platform.
+
+  // Explain our detection attempt:
```
nathanhammond, line 113:
```
Turborepo failed to start.

Turborepo detected that you are running:
darwin arm64

***

We were not able to find the binary at:
turbo-darwin-arm64/bin/turbo

We looked for it at:
/Users/nathanhammond/repos/triage/test-npm-lockfile/node_modules/turbo/bin/node_modules
/Users/nathanhammond/repos/triage/test-npm-lockfile/node_modules/turbo/node_modules
/Users/nathanhammond/repos/triage/test-npm-lockfile/node_modules
/Users/nathanhammond/repos/triage/node_modules
/Users/nathanhammond/repos/node_modules
/Users/nathanhammond/node_modules
/Users/node_modules
/node_modules
/Users/nathanhammond/.node_modules
/Users/nathanhammond/.node_libraries
/Users/nathanhammond/.nvm/versions/node/v18.15.0/lib/node

***

Your platform (darwin) can sometimes run x86 under emulation.
We did not find a possibly-compatible binary at:
turbo-darwin-64/bin/turbo

We looked for it at:
/Users/nathanhammond/repos/triage/test-npm-lockfile/node_modules/turbo/bin/node_modules
/Users/nathanhammond/repos/triage/test-npm-lockfile/node_modules/turbo/node_modules
/Users/nathanhammond/repos/triage/test-npm-lockfile/node_modules
/Users/nathanhammond/repos/triage/node_modules
/Users/nathanhammond/repos/node_modules
/Users/nathanhammond/node_modules
/Users/node_modules
/node_modules
/Users/nathanhammond/.node_modules
/Users/nathanhammond/.node_libraries
/Users/nathanhammond/.nvm/versions/node/v18.15.0/lib/node

***

We did not find any binaries on this system.
This can happen if you run installation with the --no-optional flag.

***

Turborepo detected that your lockfile (/Users/nathanhammond/repos/triage/test-npm-lockfile/package-lock.json) does not enumerate all available platforms.
This is likely a consequence of an npm issue: https://github.com/npm/cli/issues/4828.

To resolve this issue for your repository, run:
npm install turbo@^1.10.12 --package-lock-only --save-dev && npm install

You will need to commit the updated lockfile.

***

If you believe this is an error, please include this message in your report.
```

nathanhammond: There is one other branch not represented in that output:

```
Turborepo checked to see if binaries for another platform are installed.
This typically indicates an error in sharing of pre-resolved node_modules across platforms.
One common reason for this is copying files to Docker.

We found these unnecessary binaries:
turbo-windows-64/bin/turbo.exe
```

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+        // If we don't show up in the lockfile it's the wrong lockfile.
+        if (parsedLockfile?.dependencies?.turbo) {
+          // Check to see if all of `turbo-<PLATFORM>-<ARCH>` is included.
+          const hasAllPackages = Object.keys(parsedLockfile?.dependencies ?? {}).filter(dependency => availablePackages.includes(dependency)).length === availablePackages.length;
+          if (!hasAllPackages) {
+            console.error();
+            console.error('***');
+            console.error();
+            console.error(`Turborepo detected that your lockfile (${current}) does not enumerate all available platforms.`);
+            console.error('This is likely a consequence of an npm issue: https://github.com/npm/cli/issues/4828.');
```
nathanhammond, line 210: This only shows up when we detect the issue so I think it's worthwhile to include.

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+
+    for (let i = 0; previous !== current && i < MAX_LOOKUPS; i++) {
+      try {
+        const lockfile = fs.readFileSync(current);
+        const parsedLockfile = JSON.parse(lockfile);
+
+        // If we don't show up in the lockfile it's the wrong lockfile.
+        if (parsedLockfile?.dependencies?.turbo) {
+          // Check to see if all of `turbo-<PLATFORM>-<ARCH>` is included.
+          const hasAllPackages = Object.keys(parsedLockfile?.dependencies ?? {}).filter(dependency => availablePackages.includes(dependency)).length === availablePackages.length;
```
nathanhammond, line 204: @chris-olszewski to double-check me here.

I'm trying to _approximate_ "this is depended upon by the project." Are these JSON paths going to do what I want them to do?

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+            if (parsedLockfile?.packages[""]?.dependencies?.turbo) {
+              version = `@${parsedLockfile.packages[""].dependencies.turbo}`;
+              environment = ' --save-prod';
+            } else if (parsedLockfile?.packages[""]?.devDependencies?.turbo) {
+              version = `@${parsedLockfile.packages[""].devDependencies.turbo}`;
+              environment = ' --save-dev';
+            } else if (parsedLockfile?.packages[""]?.optionalDependencies?.turbo) {
+              version = `@${parsedLockfile.packages[""].optionalDependencies.turbo}`;
+              environment = ' --save-optional';
+            }
```
nathanhammond, line 224: @chris-olszewski same note here:

I'm trying to find the version of `turbo` in use, with the range qualifier. Are these JSON paths going to do what I want them to do?

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+
+            console.error();
+            console.error('To resolve this issue for your repository, run:');
+            console.error(`npm install turbo${version} --package-lock-only${environment} && npm install`);
```
nathanhammond, line 228: @chris-olszewski This is the magic command that fixes things without `rm -rf`.

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+            console.error();
+            console.error(`You will need to commit the updated lockfile.`);
+          }
+          break;
+        }
+        break;
+      } catch (e) {}
+
+      let next = path.join(current, '..', '..', 'package-lock.json');
```
nathanhammond, line 237: No-dependencies `find-up`.

packages/turbo/package.json
```diff
@@ -9,15 +9,12 @@
   "main": "./bin/turbo",
   "scripts": {
     "postversion": "node bump-version.js",
-    "postinstall": "node install.js"
```
nathanhammond, line 12: 🎉

nathanhammond (commented):
Reviewer's Guide!

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+      const resolvedPath = require.resolve(`${correctBinary}`);
+      console.warn('Installation has succeeded.');
+      return resolvedPath;
+      } catch (e) {
+      console.warn('Installation has failed.');
+    }
+  }
+
+  // 3. Both Windows and macOS ARM boxes can run x64 binaries. Attempt to run under emulation.
+  const alternateBinary = (arch === "arm64" && ['darwin', 'windows'].includes(platform)) ? `turbo-${platform}-64/bin/turbo${ext}` : null;
```
gsoltis, line 93: I don't think we want to warn on windows/arm64. It's expected, at least for now, that the binary might be amd64. If/when we ship an actual windows/arm64 binary, then we can put the warning back.

gsoltis (approved):
Just a couple of notes on errors / warnings, otherwise LGTM.

Are there any concerns about running this when the repo is using package managers besides npm?

packages/turbo/bin/turbo
chris-olszewski: The JSON paths you're using will work for npm 5-8.

The `dependencies` field is a [legacy field](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#dependencies) that is used to allow older version of npm to read newer lockfiles. As of 9.0.0 `npm` will no longer populate this field by default. If we want to support lockfile versions 1-3 then we'd need to read `packages` and fallback to `dependencies` if it isn't present.

The keys of `packages` are slightly different which makes our lives harder. As long as the user is only specifying one version of `turbo`across the monorepo then the keys will just be `node_modules/turbo` etc. If they have multiple conflicting versions of turbo then they can be nested in workspace paths e.g. `apps/web/node_modules/turbo` and `node_modules/turbo`. 

Checking for the existence of top level `node_modules/turbo*` packages should be sufficient to since if they created a bad lockfile any missing nested turbo packages would also show up in the top level.

packages/turbo/bin/turbo
chris-olszewski: Yup these will work for npm 7+ lockfiles. Lockfile version 1 won't have `packages` and we'd need to check `(dev|opt)?(d|D)ependencies` in the root `package.json` itself.

packages/turbo/bin/turbo
chris-olszewski: 😮

chris-olszewski (approved):
This is so much more straight forward!

I think we can get away with not building out this level of support for `package-lock.json` version 1. The last version of npm that used this version by default was 6 and node 16 (EOL 9/11) was the last one to ship with npm 6.

packages/turbo/bin/turbo
```diff
@@ -1,10 +1,257 @@
...
+  );
+}
+
+// This provides logging messages as it progresses towards calculating the binary path.
+function getBinaryPath() {
+  // First we see if the user has configured a particular binary path.
+  const TURBO_BINARY_PATH = process.env.TURBO_BINARY_PATH;
+  if (TURBO_BINARY_PATH) {
+    if (!fs.existsSync(TURBO_BINARY_PATH)) {
+      console.warn(
```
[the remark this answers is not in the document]
nathanhammond: This is one of the few bits of code that is directly copy pasted. I am also in favor of a hard error here, so I am going to change the behavior. This may break people, but I think we're okay with that.

packages/turbo/bin/turbo
nathanhammond: I believe this is still worth warning on. Even if you are running under emulation you don't by default get the differently-named binary. So getting the correctly-named binary is still important.

Phrased differently: if you run an install on windows arm, you should get the windows arm binary. If you don't have the windows arm binary it is indicative of an error in configuration (probably the npm issue) every time.

Also, to get this warning at all you have to fail the install saving throw where we would drop the correct binary on your box.

I'm going to keep this as is.

gsoltis:
Ah, right, the package name should match, even if the architecture doesn't.
Ok, sgtm

nathanhammond:
> Are there any concerns about running this when the repo is using package managers besides npm? - @gsoltis 

Nope! There are two things happening here that do `npm`-related nonsense:

1. An `npm install` inside of the `<PROJECT>/node_modules/turbo` folder.

Writing to this destination (no matter how it is implemented underneath the hood with linking) results in safe behavior—even if it pollutes the cache, even if that cache is later restored from a random cache, even if that cache comes from a different platform, even if the base project is a different package manager. It works _because_ it leverages `npm install` to handle the behavior, which on rerun always puts things into the expected state for the dependencies, and the `resolve` algorithm.

The material difference is that, _in a misconfigured repository_, if we restore from a polluted cache, on the same platform (worst-case scenario), the correct binary will be found inside the `node_modules/turbo/node_modules/turbo-<platform>-<arch>` folder instead of `node_modules/turbo-<platform>-arch`. This is a harmless distinction since those two things should be identical in order to work anyway.

2. Looking for `package-lock.json`.

This does not directly invoke or mutate `npm`, it just looks on the file system for evidence of the `package-lock.json` error and gives you a command to fix it. Even failing the command generation just gives you a slightly more-permissive `turbo` version.

packages/turbo/bin/turbo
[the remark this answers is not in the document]
nathanhammond:
```
Turborepo was unable to find the executable specified by TURBO_BINARY_PATH:
/this/does/not/exist

TURBO_BINARY_PATH is intended for development use-cases. You likely want to unset the environment variable.
```

packages/turbo/bin/turbo
nathanhammond: I've updated the check to do the fallback (it actually falls back to do the 9+ check). It even cleaned up the code nicely!

packages/turbo/bin/turbo
nathanhammond: Okay, I'm going to elect against perfect support for `npm@6`. They'll get a less-perfect repair command that will still work but may cause more changes in their lockfile than strictly necessary.

PR #5697, opened 2023-08-09
feat: add async graph walker

chris-olszewski:
### Description

Adds an async graph walker that respects dependencies and will only emit nodes once all of their dependents are finished.

This walker implementation emits which nodes are ready to be processed as opposed to the Go version which performs a given callback on the node when ready. We let the caller perform the processing since this being async would limit the callback to being `Send` which would greatly complicate/restrict the operations that could be done on the graph. This has the cost of the caller having to manually mark a node as processed, but the benefit is that we impose no restrictions on what processing is done.

### Testing Instructions

Added unit tests for all of the behavior that we use from the Go graph walker.


crates/turborepo-lib/src/graph/walker.rs
```diff
@@ -0,0 +1,257 @@
...
+};
+use tracing::log::trace;
+
+pub struct Walker<N> {
+    cancel: watch::Sender<bool>,
+    node_events: Option<mpsc::Receiver<(N, oneshot::Sender<()>)>>,
+    join_handles: Vec<JoinHandle<()>>,
+}
+
+pub type WalkMessage<N> = (N, oneshot::Sender<()>);
```
chris-olszewski, line 20: We could opt for making smart pointer that automatically sends the done signal upon dropping (think `MutexGuard`), but I opted to be explicit to avoid accidental done signals. I think it would be too easy to accidentally drop the guard type since the inner type implements `Copy` and any caller will probably want to resolve the node index to the node weight for processing.

crates/turborepo-lib/src/graph/walker.rs
```diff
@@ -0,0 +1,257 @@
...
+
+// These constraint might look very stiff, but since all of the petgraph graph
+// types use integers as node ids and GraphBase already constraints these types
+// to Copy + Eq so adding Hash + Send + 'static as bounds aren't unreasonable.
+impl<N: Eq + Hash + Copy + Send + 'static> Walker<N> {
+    /// Create a new graph walker for a DAG that emits nodes at a given
```
NicholasLYang, line 26: at a given?

crates/turborepo-lib/src/graph/walker.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// The graph should not be modified after a walker is created as the nodes
```
NicholasLYang, line 27: Is there a way to ensure that we have an exclusive reference to the graph?

crates/turborepo-lib/src/graph/walker.rs
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    /// Start walking the graph and return a channel that emits node
+    /// indices once all of the nodes dependencies have finished.
+    /// It is up to the caller to use the oneshot channel to mark
+    /// a node as being done.
+    pub fn walk(&mut self) -> mpsc::Receiver<WalkMessage<N>> {
+        self.node_events
+            .take()
+            .expect("walking graph with walker that has already been used")
```
NicholasLYang, line 146: Could we make this function take `mut self` (and a type state pattern) instead to enforce that this walker can only be used once?

NicholasLYang (changes requested):

crates/turborepo-lib/src/graph/walker.rs
arlyon: I think lifetimes could get hairy quite quickly but maybe we should have the Walker hold a reference to the graph for its entire lifetime to ensure that it cannot be mutably borrowed elsewhere.

crates/turborepo-lib/src/graph/walker.rs
```diff
@@ -0,0 +1,257 @@
...
+        for node in graph.node_identifiers() {
+            // Each node can finish at most once so we set the capacity to 1
+            let (tx, rx) = broadcast::channel::<()>(1);
+            txs.insert(node, tx);
+            rxs.insert(node, rx);
+        }
+        // We will be emitting at most txs.len() nodes so emitting a node should never
+        // block
+        let (node_tx, node_rx) = mpsc::channel(txs.len());
+        let mut join_handles = vec![];
```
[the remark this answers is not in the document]
arlyon: I think FuturesUnordered would probably be more efficient and I don't think we need to be strict about the order they are resolved in since we are using tasks and they will be running anyways regardless of polling or not.

arlyon (approved):

arlyon:
No additional comments beyond Nick's changes

crates/turborepo-lib/src/graph/walker.rs
chris-olszewski: I'm slightly opposed to adding this restriction here as is clashes with the approach that `petgraph` takes for graph traversals
> [Dfs](https://docs.rs/petgraph/latest/petgraph/visit/struct.Dfs.html), [Bfs](https://docs.rs/petgraph/latest/petgraph/visit/struct.Bfs.html), [DfsPostOrder](https://docs.rs/petgraph/latest/petgraph/visit/struct.DfsPostOrder.html) and [Topo](https://docs.rs/petgraph/latest/petgraph/visit/struct.Topo.html) are basic visitors and they use “walker” methods: the visitors don’t hold the graph as borrowed during traversal, only for the .next() call on the walker. They can be converted to iterators through the [Walker](https://docs.rs/petgraph/latest/petgraph/visit/trait.Walker.html) trait.

[source](https://docs.rs/petgraph/latest/petgraph/visit/index.html#graph-traversal)

Walking doesn't actually need to access the underlying graph once the struct has been built, it just will send node ids out in the order. It is up to the caller to use those node ids how they wish. If people feel strongly that we should disallow this, then I can look into holding some reference to the graph.

NicholasLYang: I think we can wait on this until we have more concrete use-cases and see if we want the restriction

NicholasLYang (approved):

PR #5700, opened 2023-08-10
docs(ci-github-actions): fix indent 

mparramont:
### Description

On the npm tab the indent looks right, but it seems to me that that wasn't the case for the yarn and pnpm tabs. I think this fixes it! :)

I found out while C&P'ing the yarn one to use in a project.

### Testing Instructions

Build the docs and check that the indent looks good now for the yarn and pnpm tabs.


chris-olszewski (approved):
Thanks for the fix!

PR #5703, opened 2023-08-10
feat: port task graph validation

chris-olszewski:
### Description

Port of the task graph validation that checks the configuration of persistent tasks.

### Testing Instructions

There are quite a few integration tests that test the various validation errors. We will rely on those to catch any behavior changes from the Go code rather than trying to convert those to unit tests. At the moment we aren't running these tests with the Rust run, but that should come soon.


crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -110,4 +114,110 @@ impl Engine<Built> {
...
+        // see if using rayon to parallelize would provide a speedup
+        let (persistent_count, mut validation_errors) = self
+            .task_graph
+            .node_indices()
+            .map(|node_index| {
+                let TaskNode::Task(task_id) = self
+                    .task_graph
+                    .node_weight(node_index)
+                    .expect("graph should contain weight for node index")
+                else {
```
arlyon, line 133: Didn't even know you could apply this syntax on enums like that. Neat.

crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -110,4 +114,110 @@ impl Engine<Built> {
...
+                    .map_or(false, |task_def| task_def.persistent);
+
+                for dep_index in self
+                    .task_graph
+                    .neighbors_directed(node_index, petgraph::Direction::Outgoing)
+                {
+                    let TaskNode::Task(dep_id) = self
+                        .task_graph
+                        .node_weight(dep_index)
+                        .expect("node index not in graph")
```
arlyon, line 149: I would like to see a messeage like 'dep_index comes from iterating the graph and must be there'

Seems like a very common thing to need, I'm surprised they don't have an access pattern that simplifies this...

crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -110,4 +114,110 @@ impl Engine<Built> {
...
+                    else {
+                        panic!("{task_id} depends on root task node");
+                    };
+
+                    let task_definition = self.task_definitions.get(dep_id).ok_or_else(|| {
+                        ValidateError::MissingTask {
+                            task_id: dep_id.to_string(),
+                            package_name: dep_id.package().to_string(),
+                        }
+                    })?;
```
arlyon, line 159: Does it make sense to return _all_ validation issues, rather than just the first? I feel like these rules are fairly recoverable and it would help UX

crates/turborepo-lib/src/engine/mod.rs
chris-olszewski: I too learned that this was possible from @NicholasLYang and I've been in love ever since.

crates/turborepo-lib/src/engine/mod.rs
chris-olszewski: I agree that it's surprising that there isn't an "unwrap-less" method for getting neighbor weights. There's [`IntoNodeReferences`](https://docs.rs/petgraph/latest/petgraph/visit/trait.IntoNodeReferences.html) which bundles the id with the ability to fetch the weight, but it isn't implemented for any of the neighbor iterators. Maybe on a DYF I'll go through and try to implement it for some more iterators.

crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -110,4 +114,110 @@ impl Engine<Built> {
...
+                match result {
+                    Ok(true) => count += 1,
+                    Ok(false) => (),
+                    Err(e) => errs.push(e),
+                }
+                (count, errs)
+            });
+
+        if persistent_count > concurrency {
+            validation_errors.push(ValidateError::PersistentTasksExceedConcurrency {
```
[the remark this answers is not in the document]
chris-olszewski: That's interesting, both in Go/Rust the task graph is constructed using only packages that are in scope. Didn't realize this was an issue. Either way this is meant to be the port of the Go behavior.

PR #5704, opened 2023-08-10
chore(turborepo): Refactor globwatcher setup

gsoltis:
### Description

 - Plumb through path library usage
 - Start categorizing errors, avoid exposing underlying `notify::Error` (cc @nathanhammond for errors data point, we'll want to do more in the future)
 - Added `canonicalize` to `AbsoluteSystemPath` 

### Testing Instructions

Tests updated to use path library, otherwise unchanged

### DIFF
--- a/crates/turborepo-globwatch/src/lib.rs
+++ b/crates/turborepo-globwatch/src/lib.rs
@@ -23,6 +23,7 @@
 use std::{
     collections::HashMap,
     fs::File,
+    io,
     path::{Path, PathBuf},
     sync::{
         atomic::{AtomicU64, Ordering},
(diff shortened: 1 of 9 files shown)


crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -235,6 +235,11 @@ impl AbsoluteSystemPath {
         Ok(AbsoluteSystemPathBuf(cleaned_path))
     }
 
+    pub fn canonicalize(&self) -> Result<AbsoluteSystemPathBuf, PathError> {
+        let canonicalized = self.0.canonicalize_utf8()?;
```
NicholasLYang, line 239: We probably can't use this cause we need to use [dunce](https://docs.rs/dunce/latest/dunce/) to avoid UNC paths on Windows.

crates/turborepo-globwatch/src/lib.rs
```diff
@@ -23,31 +23,49 @@
...
     watch,
 };
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tracing::{error, event, span, trace, warn, Level};
+use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf, PathError};
+
+/// WatchError wraps errors produced by GlobWatcher
+#[derive(Debug, Error)]
+pub enum WatchError {
+    // TODO: find a generic way to include the path in these errors
```
NicholasLYang, line 51: Could you convert the Path to a string and add it in the error type? Would require some manual conversion but not too bad.

NicholasLYang (changes requested):

crates/turborepo-paths/src/absolute_system_path.rs
gsoltis: Ah good call, I was looking for `to_realpath`. I'll update it.

crates/turborepo-globwatch/src/lib.rs
gsoltis: I'm going to defer to @nathanhammond on a standard way to include metadata along with path errors, rather than continuing to add to the ad-hoc setup we currently have.

chris-olszewski (approved):

NicholasLYang (approved):

PR #5707, opened 2023-08-11
feature(turbo): Port Run Cache

NicholasLYang:
### Description

Ports runcache along with the required dependencies like `PrefixWriter`, `LogWriter`, etc.

### Testing Instructions

Unfortunately there aren't many tests to port for runcache. I'll work on writing some tests for the dependencies


crates/turborepo-api-client/src/lib.rs
```diff
@@ -284,7 +284,7 @@ impl APIClient {
         &self,
         hash: &str,
         artifact_body: &[u8],
-        duration: u32,
```
NicholasLYang, line 287: Since the daemon accepts a `u64`, I decided to change it everywhere to `u64`

crates/turborepo-lib/src/config/turbo.rs
```diff
@@ -92,7 +91,7 @@ struct RawTaskDefinition {
     #[serde(skip_serializing_if = "Option::is_none")]
     outputs: Option<Vec<String>>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    output_mode: Option<TaskOutputMode>,
```
NicholasLYang, line 95: Realized that `TaskOutputMode` is redundant since we already have it in `cli` as `OutputLogsMode`

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -84,7 +87,7 @@ impl Run {
 
             let client = connector.connect().await?;
             debug!("running in daemon mode");
-            opts.runcache_opts.output_watcher = Some(client);
```
NicholasLYang, line 87: I removed the daemon from `Opts` because I want it to be just configuration.

crates/turborepo-lib/src/package_task.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub task_id: String,
+    task: String,
+    package_name: String,
+    pkg: PackageJson,
+    env_mode: EnvMode,
+    pub(crate) task_definition: TaskDefinition,
+    pub dir: AbsoluteSystemPathBuf,
+    command: String,
+    outputs: Vec<String>,
+    excluded_outputs: Vec<String>,
```
chris-olszewski, line 19: These can both be accessed from `task_definition.outputs.inclusions`

chris-olszewski (commented):
Some quick observations that stood out to me

crates/turborepo-lib/src/package_task.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub(crate) log_file: String,
+    hash: String,
```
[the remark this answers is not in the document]
NicholasLYang: Yeah a lot of this is likely going to change once the task hashing code is added. I'll try to make this struct as small as possible with the goal of only adding it once we've figured out where exactly it's being used.

arlyon (approved):
Nothing to add aside from those struct comments from chris

chris-olszewski (approved):
- `u32` -> `u64` in the cache tests
- I am concerned with the current color selector code as I'm pretty sure it won't work like we want it to. I'm not blocking since we can fix it down the road, but I think going for something like `Arc<Mutex<ColorSelector>>` might be a straightforward solution where we can skip having to worry about atomic operation orderings.

crates/turborepo-ui/src/color_selector.rs
```diff
@@ -0,0 +1,71 @@
...
+        ]
+    })
+}
+
+/// Selects colors for tasks and caches accordingly.
+/// Shared between tasks so allows for concurrent access.
+#[derive(Default)]
+pub struct ColorSelector {
+    idx: AtomicUsize,
+    cache: HashMap<String, Style>,
```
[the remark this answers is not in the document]
NicholasLYang: yeah tbh this implementation is a little sketchy until we actually plug it into the concurrent tasks. I suspect it'll end up a RWLock

PR #5710, opened 2023-08-11
fix(Turborepo): Update filewatching and tests to match Go implementation

gsoltis:
### Description
 
- Ports tests from #5580 (incomplete)

### Testing Instructions

Adds new tests


anthonyshew:
Looks to be stale!

PR #5716, opened 2023-08-14
feature(turbo): Task Hash

NicholasLYang:
### Description

This implements task hashing via the visitor.


cli/internal/run/run.go
```diff
@@ -429,6 +435,47 @@ func (r *run) run(ctx gocontext.Context, targets []string, executionState *turbo
 	)
 }
 
+func checkPackageInputsHashes(executionState *turbostate.ExecutionState, taskHashTracker *taskhash.Tracker) error {
```
chris-olszewski, line 438: Could we leverage [`cmp.Equal`](https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal) instead of the hand rolled equality check?

crates/turborepo-lib/src/task_hash.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+}
+
+/// Caches package-inputs hashes, and package-task hashes.
+struct TaskHasher {
+    package_inputs_hashes: PackageInputsHashes,
+    package_task_env_vars: HashMap<TaskId<'static>, DetailedMap>,
+    package_task_hashes: HashMap<TaskId<'static>, String>,
+    package_task_framework: HashMap<TaskId<'static>, String>,
+    package_task_outputs: HashMap<TaskId<'static>, Vec<AnchoredSystemPathBuf>>,
```
chris-olszewski, line 161: As part of this port can we also port this to be theadsafe like it is in Go with the RwLock? The tracker will need to be shared across tasks in Rust as well.

crates/turborepo-lib/src/task_hash.rs
```diff
@@ -0,0 +1,318 @@
...
+            }
+        }
+
+        let file_hashes = FileHashes(hash_object);
+        let hash = file_hashes.clone().hash();
+
+        Ok(Some(hash))
+    }
+
+    pub async fn calculate_task_hash(
```
chris-olszewski, line 151: This can also be made sync if the other changes suggested are made.
```suggestion
    pub fn calculate_task_hash(
```

crates/turborepo-lib/src/task_hash.rs
```diff
@@ -0,0 +1,318 @@
+use std::collections::{HashMap, HashSet};
+
+use thiserror::Error;
+use tokio::sync::Mutex;
```
chris-olszewski, line 4: We should only use the tokio `Mutex` if it needs to be held across an await [and we probably should avoid writing code that does that](https://tokio.rs/tokio/tutorial/shared-state#holding-a-mutexguard-across-an-await).
```suggestion
use std::sync::Mutex;
```

chris-olszewski (commented):
Quick notes on the async part of the task hash tracker

crates/turborepo-lib/src/run/task_id.rs
```diff
@@ -8,7 +8,7 @@ pub const TASK_DELIMITER: &str = "#";
 pub const ROOT_PKG_NAME: &str = "//";
 
 /// A task identifier as it will appear in the task graph
-#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize)]
```
chris-olszewski, line 11: I think that we would want to serialize by first going through `String`. I'm not sure where this is getting used in this PR, but I would expect we want `"web#build"` and not `{ "package": "web", "task": "build" }`

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -229,10 +230,26 @@ impl Run {
...
+            && !root_turbo_json.global_pass_through_env.is_empty()
+        {
+            global_env_mode = EnvMode::Strict;
+        }
+
         let pkg_dep_graph = Arc::new(pkg_dep_graph);
         let engine = Arc::new(engine);
-        let visitor = Visitor::new(pkg_dep_graph, &opts);
-        visitor.visit(engine).await?;
+        let visitor = Visitor::new(
```
chris-olszewski, line 242: Is there a reason why we're no longer constructing the task hash tracker and calculating file hashes before we start walking the graph?

crates/turborepo-lib/src/task_hash.rs
```diff
@@ -0,0 +1,318 @@
...
+        task_env_mode: ResolvedEnvMode,
+        workspace: &WorkspaceInfo,
+        dependency_set: HashSet<&TaskNode>,
+        repo_root: &AbsoluteSystemPath,
+    ) -> Result<String, Error> {
+        let do_framework_inference = self.opts.run_opts.framework_inference;
+        let is_monorepo = !self.opts.run_opts.single_package;
+
+        let hash_of_files = self
+            .calculate_package_inputs_hash(scm, task_id, task_definition, workspace, repo_root)
```
chris-olszewski, line 165: Is there a reason for the change from the Go behavior where we eagerly calculate this?

I think it would be beneficial to keep that behavior as it allows us to easily use something like rayon to parallelize the calculation of these inputs.

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -78,10 +109,40 @@ impl<'a> Visitor<'a> {
...
+                .calculate_task_hash(
+                    &self.scm,
+                    &info,
+                    &task_definition,
+                    task_env_mode,
+                    workspace_info,
+                    dependency_set,
+                    self.repo_root,
+                )
+                .await?;
```
chris-olszewski, line 142: This will end up blocking the processing of any other tasks by the visitor since it happens in the main thread and not via one of the `tokio::spawn` tasks.

chris-olszewski (changes requested):
I think avoiding blocking the visitor is something we need to avoid from the jump since I expect it could have implications of how this gets designed.

crates/turborepo-lib/src/task_hash.rs
NicholasLYang: I'm not sure rayon is the right tool here since the hashes are computed by shelling out to git which is probably not CPU bound. I didn't want to spin up multiple concurrency primitives here because it felt unnecessary when we could use the visitor for both

crates/turborepo-lib/src/hash/mod.rs
```diff
@@ -28,58 +32,55 @@ mod proto_capnp {
...
-    inclusions: Vec<String>,
-    exclusions: Vec<String>,
+    pub dot_env: &'a [turbopath::RelativeUnixPathBuf],
 }
 
 pub struct LockFilePackages(pub Vec<turborepo_lockfiles::Package>);
-struct FileHashes(HashMap<turbopath::RelativeUnixPathBuf, String>);
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(transparent)]
```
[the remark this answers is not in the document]
NicholasLYang: Oh I was doing some comparison checking between Go and Rust. Not needed now

chris-olszewski (approved):
This looks great. A few questions and I think one incorrect magic string.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -229,10 +232,37 @@ impl Run {
...
+        if matches!(global_env_mode, EnvMode::Infer)
+            && !root_turbo_json.global_pass_through_env.is_empty()
+        {
+            global_env_mode = EnvMode::Strict;
+        }
+
+        let workspaces = pkg_dep_graph.workspaces().collect();
+        let package_inputs_hashes = PackageInputsHashes::calculate_file_hashes(
+            scm,
+            engine.tasks().par_bridge(),
```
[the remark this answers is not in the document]
NicholasLYang: Hmm yeah will `concurrency` apply to all forms of threads in our execution? Because we have a few situations like task execution, hashing, caching, etc. that spin up threads.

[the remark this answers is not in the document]
chris-olszewski: On the Go side it was more of a "best effort" limit to the amount of goroutines we would spin up. In some places we would respect it like with package task hashing, but in other places we wouldn't share that limit. I think we should probably just limit concurrency for stuff where we spend a decent amount of time e.g. hashing, task execution, output saving (?).

PR #5719, opened 2023-08-15
Use shallow references in `RebasedAsset`

jridgewell:
### Description

Changes `RebasedAsset::references()` to rebase only the directly referenced modules, allowing further rebasing to happen when their references are requested.

### Testing Instructions

Closes WEB-1379


alexkirsz (approved):

jridgewell:
This is only used by node-file-trace (along with all the `AggregatedGraph` related code), and it causes a failure. It's not worth it to track down why, so I'm just going to close this.

PR #5727, opened 2023-08-15
fix: give name to package so that it works with rust codepath

arlyon:
### Description

The rust codepath for turbo doesn't like nameless packages. This fixes that. 

### Testing Instructions

- compile turbo with rust run enabled (instructions are complicated for now, but will be easier shortly)
- error that a package.json doesn't have a name

Closes WEB-1381

### DIFF
--- a/crates/turbopack-tests/tests/execution/package.json
+++ b/crates/turbopack-tests/tests/execution/package.json
@@ -1,5 +1,6 @@
 {
   "private": true,
+  "name": "execution-test",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


arlyon:
Seems like we have some messed up codeowners detection here, so sorry about the issue. LMK what you want me to do w/ linear.

tknickman (approved):

kdy1 (approved):

wbinnssmith (approved):

PR #5728, opened 2023-08-15
feat: port run graph flag

chris-olszewski:
### Description

This is a very literal port of the dot serializing logic from Go without any features that we don't use (displaying cycles). I'm doing this so we can get byte matching output between Rust and Go for our integration tests which will help during the port. Once the port is done we should look into just using the `Dot` support that `petgraph` has.

### Testing Instructions
Build turbo with `cargo build -p turbo --features run-stub` and then run `turbo run build --graph` on a project. Note that until scope/filter is ported #5628 the packages included in the task graph won't match what is expected in the integration tests.


crates/turborepo-lib/src/run/mod.rs
```diff
@@ -164,6 +167,22 @@ impl Run {
         )
         .build()?;
 
+        if opts.run_opts.graph_file.is_some() || opts.run_opts.graph_dot {
+            match opts.run_opts.graph_file {
+                Some(graph_file) => {
+                    let graph_file = AbsoluteSystemPathBuf::from_unknown(
+                        &AbsoluteSystemPathBuf::cwd()?,
```
NicholasLYang, line 174: We should probably take into account the `--cwd` flag here, no?

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -164,6 +167,22 @@ impl Run {
         )
         .build()?;
 
+        if opts.run_opts.graph_file.is_some() || opts.run_opts.graph_dot {
```
NicholasLYang, line 170: We should probably just reduce this down to a single field with an enum like `GraphOutput::Stdout, GraphOutput::File(PathBuf)`

NicholasLYang (approved):
Nothing blocking, just some notes

crates/turborepo-lib/src/engine/dot.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+const GRAPH_PRELUDE: &str = "digraph {\n\tcompound = \"true\"
+\tnewrank = \"true\"
+\tsubgraph \"root\" {
+";
+
+fn render_graph<N>(
+    graph: &Graph<N, ()>,
+    display_node: impl Fn(&N) -> String,
```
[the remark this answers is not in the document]
chris-olszewski: Realized that `Fn` was overly restrictive so went with `FnMut` as that allows greater flexibility if we ever need something more complex for node rendering.

PR #5730, opened 2023-08-15
ci: Adding mold as linker

NicholasLYang:
### Description

Just a minor perf improvement to use `mold` versus `ld` for linking.

### Testing Instructions

confirmed with `readelf` that we're using mold


arlyon (commented):
Well that was a small change lmao. I have also seen this lockfile change so i think its ok leaving it in. Probably an issue w/ another PR that leaked

arlyon (approved):

PR #5731, opened 2023-08-15
chore(labels): consolidate issue labels

tknickman:
### Description

_The last label related PR._ 

This makes sure we're using the same **team** labels for both issues and pull requests. 

It has always bothered me that labels get: `area: [team]` while pull requests would get `team: [team]`. 

This is now consolidated to only the `owned-by: [team]` label for both issues and pull requests. 

We also now include a `created-by: [team]` label.

### DIFF
--- a/.github/turbo-orchestrator.yml
+++ b/.github/turbo-orchestrator.yml
@@ -30,7 +30,7 @@
 
 labeler:
   labels:
-    # teams
+    # owned-by
     - label: "owned-by: turborepo"
       when:
         isAnyFileOwnedByMatch: '@vercel\/turbo-oss'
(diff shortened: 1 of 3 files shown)


.github/turbo-orchestrator.yml
```diff
@@ -39,10 +39,10 @@ labeler:
         isAnyFileOwnedByMatch: '@vercel\/web-tooling'
 
     # created-by
-    - label: "created-by: web-tooling team"
+    - label: "created-by: web-tooling"
```
tknickman, line 42: Remove the team to match the owned-by label format

arlyon (approved):

PR #5732, opened 2023-08-15
Implement reusable `Module`/`OutputAsset` graph helpers

jridgewell:
### Description

This does 2 things:
- Implements 2 new graph traversal iterator functions, which return perform the graph traversal but return iterators that can be further chained
- Removes `all_referenced_modules` (which traverses transitive modules despite its doc comment) because it's the same `all_modules`

I found myself reimplementing the same graph traversal to get referenced `Module`/`OutputAsset`s. I didn't want to use `all_modules`/`all_assets_from_entries` because they already collect the result into their own `Vec`. The new methods return lazy iterators, allowing you to `map`/`flat_map` and collect that result into your output.

I had to move a few functions into `node-file-trace` due to it failing for some reason.

### Testing Instructions

Closes WEB-1385

### DIFF
--- a/crates/node-file-trace/src/lib.rs
+++ b/crates/node-file-trace/src/lib.rs
@@ -638,9 +638,9 @@ async fn main_operation(
     Ok(Vc::cell(Vec::new()))
 }
 
-/// Aggregates all [Module]s referenced by an [Module]. [ModuleReference]
-/// This does not include transitively references [Module]s, but it includes
-/// primary and secondary [Module]s referenced.
+/// Aggregates all [Module]s referenced by a [Module] through [ModuleReference]s.
+/// This does not include transitively referenced [Module]s, but it includes
+/// primary and secondary referenced [Module]s.
 #[turbo_tasks::function]
 pub async fn all_referenced_modules(module: Vc<Box<dyn Module>>) -> Result<Vc<Modules>> {
     let references_set = module.references().await?;
@@ -661,7 +661,7 @@ pub async fn all_referenced_modules(module: Vc<Box<dyn Module>>) -> Result<Vc<Mo
     Ok(Vc::cell(modules))
 }
 
-/// Aggregates all [Module]s referenced by an [Module] including transitively
+/// Aggregates all [Module]s referenced by a [Module], including transitively
 /// referenced [Module]s. This basically gives all [Module]s in a subgraph
 /// starting from the passed [Module].
 #[turbo_tasks::function]
--- a/crates/node-file-trace/src/lib.rs
+++ b/crates/node-file-trace/src/lib.rs
@@ -5,7 +5,7 @@
 mod nft_json;
 
 use std::{
-    collections::{BTreeSet, HashMap},
+    collections::{BTreeSet, HashMap, HashSet, VecDeque},
     env::current_dir,
     fs,
     future::Future,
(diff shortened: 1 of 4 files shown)


crates/node-file-trace/src/lib.rs
```diff
@@ -638,6 +638,52 @@ async fn main_operation(
     Ok(Vc::cell(Vec::new()))
 }
 
+/// Aggregates all [Module]s referenced by an [Module]. [ModuleReference]
+/// This does not include transitively references [Module]s, but it includes
+/// primary and secondary [Module]s referenced.
```
alexkirsz, line 643:
```suggestion
/// Aggregates all [Module]s referenced by a [Module] through [ModuleReference]s.
/// This does not include transitively referenced [Module]s, but it includes
/// primary and secondary referenced [Module]s.
```

crates/node-file-trace/src/lib.rs
```diff
@@ -638,6 +638,52 @@ async fn main_operation(
...
+        for &reference in resolve_result.await?.get_references() {
+            queue.push_back(reference.resolve_reference());
+        }
+    }
+    Ok(Vc::cell(modules))
+}
+
+/// Aggregates all [Module]s referenced by an [Module] including transitively
+/// referenced [Module]s. This basically gives all [Module]s in a subgraph
+/// starting from the passed [Module].
```
alexkirsz, line 666:
```suggestion
/// Aggregates all [Module]s referenced by a [Module], including transitively
/// referenced [Module]s. This basically gives all [Module]s in a subgraph
/// starting from the passed [Module].
```

crates/node-file-trace/src/lib.rs
```diff
@@ -638,6 +638,52 @@ async fn main_operation(
...
+        let references = references
+            .issue_file_path(parent.ident().path(), "expanding references of asset")
+            .await?;
+        for asset in references.await?.iter() {
+            if assets.insert(*asset) {
+                queue.push_back((*asset, all_referenced_modules(*asset)));
+            }
+        }
+    }
+    Ok(Vc::cell(assets.into_iter().collect()))
```
alexkirsz, line 684: Can we use our graph traversal algorithm here instead? It can parallelize the traversal to some degree.

alexkirsz (approved):

crates/node-file-trace/src/lib.rs
jridgewell: I attempted to do that, but NFT is failing for some reason. So I copied the old functions here, but it's still failing. 😭

crates/turbopack/src/rebase/mod.rs
```diff
@@ -50,14 +50,10 @@ impl OutputAsset for RebasedAsset {
...
     async fn references(&self) -> Result<Vc<OutputAssets>> {
-        let mut references = Vec::new();
-        for &module in all_referenced_modules(self.source).await?.iter() {
-            references.push(Vc::upcast(RebasedAsset::new(
-                module,
-                self.input_dir,
-                self.output_dir,
-            )));
-        }
+        let references = all_modules_iter([self.source].into_iter())
```
sokra, line 53: This should only return the shallow references (not the whole graph). It was using `all_referenced_modules` which only returns the shallow references.

It should also include primary and secondary references.

This is used in NFT and that's why it's failing

crates/node-file-trace/src/lib.rs
sokra: NFT walk all primary and secondary references, while turbopack walks only the primary ones.

`all_referenced_modules` gives primary and secondary references, `primary_referenced_modules` gives only the primary ones.

sokra (changes requested):

anthonyshew:
Closing as stale!

PR #5734, opened 2023-08-15
feat(Turborepo): Start implementing filewatching tests

gsoltis:
### Description

 - Some light type aliasing for the filewatcher to make it more readable
 - Add `PartialEq` for `Path` and `PathBuf` to `AbsoluteSystemPath` to allow for comparisons
 - Add _some_ of the tests from Go's `filewatcher_test.go`. 
 - Add `add_root` to mirror Go's implementation. This will likely eventually take the place of `include` so that globs are moved up a layer.

Note that some significant restructuring of Rust filewatching is going to need to happen to get the rest of the Go tests moved over.

### Testing Instructions

Added new tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2873,6 +2873,7 @@ dependencies = [
  "notify-debouncer-mini",
  "pin-project",
  "stop-token",
+ "tempfile",
  "test-case",
  "thiserror",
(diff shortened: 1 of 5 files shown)


anthonyshew:
Looks to be stale!

PR #5746, opened 2023-08-17
feat(turborepo): Framework inference

NicholasLYang:
### Description

Implements framework inference for turborepo.

### Testing Instructions

Ports tests from `inference_test.go`

### DIFF
--- a/crates/turborepo-lib/src/framework.rs
+++ b/crates/turborepo-lib/src/framework.rs
@@ -154,13 +154,9 @@ impl Matcher {
 pub fn infer_framework(workspace: &WorkspaceInfo, is_monorepo: bool) -> Option<&'static Framework> {
     let frameworks = get_frameworks();
 
-    for framework in frameworks {
-        if framework.dependency_match.test(workspace, is_monorepo) {
-            return Some(framework);
-        }
-    }
-
-    None
+    frameworks
+        .iter()
+        .find(|framework| framework.dependency_match.test(workspace, is_monorepo))
 }
 
 #[cfg(test)]
--- a/crates/turborepo-lib/src/framework.rs
+++ b/crates/turborepo-lib/src/framework.rs
@@ -1,0 +1,290 @@
+use std::sync::OnceLock;
+
+use crate::package_graph::WorkspaceInfo;
+
+#[derive(Debug, PartialEq)]
+enum Strategy {
+    All,
+    Some,
+}
+
+#[derive(Debug, PartialEq)]
+struct Matcher {
+    strategy: Strategy,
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -38,10 +38,13 @@ impl WorkspaceInfo {
     }
 }
 
+type PackageName = String;
+type PackageVersion = String;
+
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
 pub struct Package {
```
chris-olszewski, line 45: A quick search makes it look like this was only used in external deps set. I vote we just drop it.

crates/turborepo-lib/src/framework.rs
```diff
@@ -0,0 +1,290 @@
...
+
+impl Matcher {
+    pub fn test(&self, workspace: &WorkspaceInfo, is_monorepo: bool) -> bool {
+        // In the case where we're not in a monorepo, i.e. single package mode
+        // `unresolved_external_dependencies` is not populated. In which
+        // case we should check `dependencies` instead.
+        let deps = if is_monorepo {
+            workspace.unresolved_external_dependencies.as_ref()
+        } else {
+            workspace.package_json.dependencies.as_ref()
```
chris-olszewski, line 137: I realize this is just a port of Go behavior, but do we want to also check the other places dependencies could appear? Primarily `devDependencies` and `optionalDependencies`

crates/turborepo-lib/src/framework.rs
```diff
@@ -0,0 +1,290 @@
...
+pub fn infer_framework(workspace: &WorkspaceInfo, is_monorepo: bool) -> Option<&'static Framework> {
+    let frameworks = get_frameworks();
+
+    for framework in frameworks {
+        if framework.dependency_match.test(workspace, is_monorepo) {
+            return Some(framework);
+        }
+    }
+
+    None
```
chris-olszewski, line 163:
```suggestion
    frameworks
        .iter()
        .find(|framework| framework.dependency_match.test(workspace, is_monorepo))
```

chris-olszewski (approved):

crates/turborepo-lib/src/package_graph/mod.rs
NicholasLYang: I added it mostly for my clarification because a `HashMap<String, String>` is pretty opaque

chris-olszewski: Oops, I meant the `Package` struct not the type aliases.

PR #5748, opened 2023-08-17
Add support to integration tests for experimental rust codepath

arlyon:
### Description

This PR adds some basic logging / flags to use the rust codepath. When the binary is compiled with the `run-stub` feature, the `--experimental-rust-codepath` flag (better used with the env var of the same name) will opt the run into using rust, rather than go.

This also adds commands to integration and e2e for this. Note that e2e tests (prysk) will probably fail. This is because we currently print a warning when the codepath is entered.

### Testing Instructions

```
cargo build --features run-stub
cd turborepo-tests/integration
pnpm i
pnpm run test:rust-codepath
```

Note: this is based on https://github.com/vercel/turbo/pull/5727 otherwise turbo fails during workspace enumeration.


gsoltis:
@arlyon #5727 is merged, can you update?

chris-olszewski (approved):

PR #5750, opened 2023-08-17
chore: remove use of rc for task definitions

chris-olszewski:
### Description

After reading through the code again, the `Rc` around task defintions is unnecessary if we just wait to insert it into the task graph until we're done processing it. (This is necessary for task graph execution where we share immutable access to the task graph across threads and need to wrap it in an `Arc`)

### Testing Instructions

Existing unit tests


NicholasLYang (approved):
If the compiler approves, I approve

PR #5751, opened 2023-08-18
(Needs discussion) Reworking `prisma` docs

maschwenk:
### Description

This is more of a conversation starter. I'm working on a large `pnpm` turbo repo with both:

1. Graphql client codegen
2. Prisma client codegen

And I'm trying to figure out how best to handle caching both. One thing I've learned is that 

```txt
  output        = "../node_modules/.prisma/client-package-a"
```

Can be powerful for making sure codegen for Prisma clients does not go in the workspace root `node_modules`.

But the more general question is, what is the best approach for creating a Turborepo config that handles `gitignore`'d codegen, and ensure:

1. The code is cacheable
2. The code can be painlessly imported

### Testing Instructions

Docs/discussion

### DIFF
--- a/docs/pages/repo/docs/handbook/tools/prisma.mdx
+++ b/docs/pages/repo/docs/handbook/tools/prisma.mdx
@@ -201,8 +201,12 @@ Check out the section on [running tasks](/repo/docs/core-concepts/monorepos/runn
 
 ### 7. Caching the results of `prisma generate`
 
-`prisma generate` outputs files to the filesystem, usually inside `node_modules`. In theory, it should be possible to cache the output of `prisma generate` with Turborepo to save a few seconds.
+By default, Prisma will output generated clients to `node_modules`. When using `pnpm`, this can cause the generated clients to be put root workspace's `node_modules`, which can mean multiple packages are writing to the same place.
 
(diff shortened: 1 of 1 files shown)


farreldarian:
I still get `[PrismaClientInitializationError]: Your schema.prisma could not be found, and we detected that you are using Next.js.` with the current approach when using Next.js.

After some further investigation, I've found a workaround that seems to resolve this problem. Specifically, I've had success by generating the output to `../node_modules/@prisma/client/<output-name>`, and simultaneously directing the turbo output to `../../node_modules/.pnpm/@prisma+client@5.3.1_prisma@5.3.1/node_modules/@prisma/client/<output-name>/**`.

It's important to note that the "../../" at the beginning of the turbo output path is because the package is located under `packages/db` in my monorepo structure, also I'm using pnpm in this setup. The `<output-name>` can be anything, but it's crucial to specify the Prisma version absolutely. This adjustment appears to address the issue I encountered in the Next.js monorepo environment.

anthonyshew:
Unfortunately, from my testing on the idea described in this PR, I'm not having any luck.

I'm going to close this is stale since Prisma has had at least two majors since the date on this PR. That's on us!

Happy to take an updated PR (or re-open this one) if this solution still works and I'm not realizing it.

PR #5754, opened 2023-08-18
chore(types): consolidate packagejson type

tknickman:
### Description

Consolidate to a single definition of `PackageJson` to ensure no more of the alternate `workspace` format bugs sneak in again.

This type is in `@turbo/utils` instead of `@turbo/types` to keep the latter lean and mean for sharing only the turbo config type.

Closes TURBO-1223


chris-olszewski (approved):
Beautiful

Issue #5756, opened 2023-08-19
[turborepo] UI package is using older react version


anmol242:
### What version of Turborepo are you using?

latest

### What package manager are you using / does the bug impact?

npm

### What operating system are you using?

Mac

### Describe the Bug

When installing turbo with latest code it is adding ` "react": "^17.0.2",` as a dependency in `ui` package while `web` and `docs` are using ` "react": "^18.2.0",`. So, I had to manually change the react version in `ui` package.

### Expected Behavior

`UI` package should use same version similar to docs and web aps

### To Reproduce

Run `npx create-turbo@latest` in your local and you will see ui package is using `"react": "^17.0.2",`

tknickman:
Fixed in https://github.com/vercel/turbo/pull/5757. Thanks @anmol242!

PR #5757, opened 2023-08-19
fix(example): basic example react version in ui package

anmol242:
### Description
This PR fixes the issue mentioned in https://github.com/vercel/turbo/issues/5756

Corrected version of react same as `web` and `docs` app so that we don't have to manually change react version while installing boiler plate on turbo

### Testing Instructions

- Install basic boiler plate of turbo app. UI package will use the same react version which is used by `web` and `docs` app


anmol242:
Hey @tknickman and @chris-olszewski can I get a review on this PR?

tknickman (approved):
Thanks!

PR #5763, opened 2023-08-21
fix(turborepo): Adopt std::cell::OnceCell

nathanhammond:
We're using Tokio's `OnceCell` for configuration loading, but `OnceCell` has been stabilized in core. Use that instead.

Closes TURBO-1224

### DIFF
--- a/crates/turborepo-lib/src/commands/mod.rs
+++ b/crates/turborepo-lib/src/commands/mod.rs
@@ -1,8 +1,7 @@
-use std::borrow::Borrow;
+use std::{borrow::Borrow, cell::OnceCell};
 
 use anyhow::Result;
 use sha2::{Digest, Sha256};
-use tokio::sync::OnceCell;
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
 use turborepo_api_client::APIClient;
 use turborepo_ui::UI;
@@ -10,7 +9,7 @@ use turborepo_ui::UI;
 use crate::{
     config::{
         default_user_config_path, get_repo_config_path, ClientConfig, ClientConfigLoader,
-        RepoConfig, RepoConfigLoader, UserConfig, UserConfigLoader,
+        Error as ConfigError, RepoConfig, RepoConfigLoader, UserConfig, UserConfigLoader,
     },
     Args,
 };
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -55,18 +54,26 @@ impl CommandBase {
...
             .with_login(self.args.login.clone())
             .with_team_slug(self.args.team.clone())
-            .load()?;
+            .load()
+    }
 
-        self.repo_config.set(repo_config)?;
+    pub fn repo_config(&self) -> Result<&RepoConfig, ConfigError> {
+        self.repo_config.get_or_try_init(|| self.repo_config_init())
+    }
```
nathanhammond, line 69: This is a rigid new pattern here across all of the config loading sources. It makes it super-clear what's happening each time.

crates/turborepo-lib/src/config/client.rs
```diff
@@ -60,17 +62,12 @@ impl ClientConfigLoader {
...
 
-        // This goes wrong when TURBO_REMOTE_CACHE_TIMEOUT can't be deserialized to u64
-        match config_attempt {
-            Err(_) => Ok(ClientConfig {
-                config: ClientConfigValue {
-                    remote_cache_timeout: DEFAULT_TIMEOUT,
-                },
-            }),
-            Ok(config) => Ok(ClientConfig { config }),
-        }
```
nathanhammond, line 73: This behavior does not match any of our other configuration loaders which fail hard in the event of garbage input.

We can come back and create "this error is only a warning" behavior if we want to do so later, but for now make the failure mode consistent.

nathanhammond (commented):
Reviewer's Guide!

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -55,18 +54,26 @@ impl CommandBase {
...
 
-        Ok(())
+    pub fn repo_config_mut(&mut self) -> Result<&mut RepoConfig, ConfigError> {
+        // Approximates `get_mut_or_try_init`
+        self.repo_config
+            .get_or_try_init(|| self.repo_config_init())?;
```
chris-olszewski, line 74: We could rely `repo_config` to make sure that the init function remains consistent. Not sure if that consistency is worth obscuring that initialization is happening.
```suggestion
        self.repo_config()?;
```

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -81,62 +88,41 @@ impl CommandBase {
...
     }
 
-    pub fn user_config_mut(&mut self) -> Result<&mut UserConfig> {
-        if self.user_config.get().is_none() {
-            self.create_user_config()?;
-        }
+    pub fn user_config_mut(&mut self) -> Result<&mut UserConfig, ConfigError> {
+        // Approximates `get_mut_or_try_init`
+        self.user_config
+            .get_or_try_init(|| self.user_config_init())?;
```
chris-olszewski, line 104: Same comment as above.
```suggestion
        self.user_config()?;
```

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -81,62 +88,41 @@ impl CommandBase {
...
     }
 
-    pub fn client_config(&self) -> Result<&ClientConfig> {
-        if self.client_config.get().is_none() {
-            self.create_client_config()?;
-        }
+    pub fn client_config_mut(&mut self) -> Result<&mut ClientConfig, ConfigError> {
+        // Approximates `get_mut_or_try_init`
+        self.client_config
+            .get_or_try_init(|| self.client_config_init())?;
```
chris-olszewski, line 123:
```suggestion
        self.client_config()?;
```

crates/turborepo-lib/src/config/client.rs
```diff
@@ -133,72 +132,91 @@ mod test {
     #[test]
     fn test_client_arg_env_variable() -> Result<()> {
         #[derive(Debug)]
-        struct TestCase {
+        struct TestCase<'a> {
             arg: Option<u64>,
             env: String,
             output: u64,
+            want_err: Option<&'a str>,
```
chris-olszewski, line 139: I think we'll only ever want to check against static error messages so this restriction would be fine.
```suggestion
        struct TestCase {
            arg: Option<u64>,
            env: String,
            output: u64,
            want_err: Option<&'static str>,
```

crates/turborepo-lib/src/commands/mod.rs
nathanhammond: I had this and chose to change it for explicitness reasons. I'll switch it up.

PR #5764, opened 2023-08-21
fix(turborepo): Better unlink Output

nathanhammond:
`turbo unlink` previously would delete the entire configuration file which is unnecessarily punishing, this switches to do the absolute minimum change to the file.

Closes TURBO-1225

### DIFF
--- a/crates/turborepo-lib/src/commands/unlink.rs
+++ b/crates/turborepo-lib/src/commands/unlink.rs
@@ -11,26 +11,48 @@ enum UnlinkSpacesResult {
 }
 
 fn unlink_remote_caching(base: &mut CommandBase) -> Result<()> {
-    base.delete_repo_config_file()
-        .context("could not unlink. Something went wrong")?;
+    let repo_config: &mut crate::config::RepoConfig = base.repo_config_mut()?;
+    let needs_disabling = repo_config.team_id().is_some() || repo_config.team_slug().is_some();
 
-    println!(
-        "{}",
-        base.ui.apply(GREY.apply_to("> Disabled Remote Caching"))
-    );
+    if needs_disabling {
+        base.repo_config_mut()?.set_team_id(None)?;
+
+        println!(
+            "{}",
+            base.ui.apply(GREY.apply_to("> Disabled Remote Caching"))
+        );
+    } else {
+        println!(
+            "{}",
+            base.ui
+                .apply(GREY.apply_to("> No Remote Caching config found"))
+        );
+    }
 
     Ok(())
 }
 
 fn unlink_spaces(base: &mut CommandBase) -> Result<()> {
+    let repo_config: &mut crate::config::RepoConfig = base.repo_config_mut()?;
+    let needs_disabling =
+        repo_config.space_team_id().is_some() || repo_config.space_team_slug().is_some();
+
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/commands/unlink.rs
```diff
@@ -11,26 +11,48 @@ enum UnlinkSpacesResult {
...
-        .context("could not unlink. Something went wrong")?;
+    let repo_config: &mut crate::config::RepoConfig = base.repo_config_mut()?;
+    let needs_disabling = repo_config.team_id().is_some() || repo_config.team_slug().is_some();
 
-    println!(
-        "{}",
-        base.ui.apply(GREY.apply_to("> Disabled Remote Caching"))
-    );
+    if needs_disabling {
+        base.repo_config_mut()?.set_team_id(None)?;
```
chris-olszewski, line 18: Can we not reuse `repo_config` here?
```suggestion
        repo_config.set_team_id(None)?;
```

crates/turborepo-lib/src/commands/unlink.rs
```diff
@@ -11,26 +11,48 @@ enum UnlinkSpacesResult {
 }
 
 fn unlink_remote_caching(base: &mut CommandBase) -> Result<()> {
-    base.delete_repo_config_file()
-        .context("could not unlink. Something went wrong")?;
+    let repo_config: &mut crate::config::RepoConfig = base.repo_config_mut()?;
+    let needs_disabling = repo_config.team_id().is_some() || repo_config.team_slug().is_some();
```
chris-olszewski, line 15: Would it be worth moving this logic into the config struct itself to something like `maybe_disable_remote_caching -> Result<bool, Error>`?

chris-olszewski (approved):

crates/turborepo-lib/src/commands/unlink.rs
nathanhammond: I _almost_ did this, but I want to remove `team_slug` entirely and change the shape of how this works by default. It felt premature.

crates/turborepo-lib/src/commands/unlink.rs
nathanhammond: I did this in isolation from the `*_config` handling change; this is a good catch.

crates/turborepo-lib/src/commands/unlink.rs
```diff
@@ -11,26 +11,48 @@ enum UnlinkSpacesResult {
...
+            "{}",
+            base.ui.apply(GREY.apply_to("> Disabled Remote Caching"))
+        );
+    } else {
+        println!(
+            "{}",
+            base.ui
+                .apply(GREY.apply_to("> No Remote Caching config found"))
+        );
+    }
```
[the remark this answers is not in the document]
nathanhammond: I decided to understand this as "the visual noise of the logging makes it hard to tell what is going on."

New pattern:
```rust
let output;
if needs_disabling {
    repo_config.set_team_id(None)?;
    output = "> Disabled Remote Caching";
} else {
    output = "> No Remote Caching config found";
}

println!(
    "{}",
    base.ui.apply(GREY.apply_to(output))
);
```

PR #5769, opened 2023-08-21
fix: copy root json to full directory

### LINKED ISSUE #5768: [turborepo] Prune is no longer copying package.json to `full` directory

1.10.10

Mac

### Describe the Bug

Running `turbo prune --scope=X --docker` properly pairs down the list of packages, however the root `package.json` doesn't seem to be copied into the `full` folder.

This causes `yarn` to think that there are 0 packages and it generates an empty lock file.

According to the docs, there should be a `package.json` in the `full` directory.  
https://turbo.build/repo/docs/reference/command-line-reference/prune#--docker

### Expected Behavior

Copy the `package.json` into the `full` directory as well as the `json` directory

### To Reproduce

```
turbo prune --scope=motion-dashboard --docker --out-dir ~/tmp/prune-test
```

`root` directory
CleanShot 2023-08-21 at 13 39 20

`json` directory
CleanShot 2023-08-21 at 13 40 01

`full` directory
CleanShot 2023-08-21 at 13 38 31

#### With 1.10.9

`root` directory
CleanShot 2023-08-21 at 13 42 55

`json` directory
CleanShot 2023-08-21 at 13 43 16

`full` directory
CleanShot 2023-08-21 at 13 43 37

### Reproduction Repo

_No response_

chris-olszewski:
### Description

Fixes #5768

We were copying the root `package.json` into `out` and not `out/full` as intended if the `--docker` flag was used. This behavior would only appear if the root `package.json` contained patches the needed to be pruned. This PR gets us back to the Go behavior.

### Testing Instructions

Started also tracking the contents of `out/full` in an integration test

Closes TURBO-1227

### DIFF
--- a/crates/turborepo-lib/src/commands/prune.rs
+++ b/crates/turborepo-lib/src/commands/prune.rs
@@ -187,7 +187,7 @@ pub fn prune(
             prune.copy_file(&patch.to_system_path(), Some(CopyDestination::Docker))?;
         }
     } else {
-        prune.copy_file(package_json(), Some(CopyDestination::All))?;
+        prune.copy_file(package_json(), Some(CopyDestination::Docker))?;
     }
 
     Ok(())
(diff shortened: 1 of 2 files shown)


tknickman (approved):
👍🏼

PR #5770, opened 2023-08-21
test: verify we can walk DAGs with multiple roots

chris-olszewski:
### Description

As per a comment from @nathanhammond, this is a PR that adds a test to display that we can currently walk graphs that we couldn't in Go.

### Testing Instructions
New test passes

Closes TURBO-1229


nathanhammond (approved):

crates/turborepo-lib/src/graph/walker.rs
```diff
@@ -282,4 +282,68 @@ mod test {
         walker.wait().await.unwrap();
         assert_eq!(visited.lock().unwrap().as_slice(), &[c, b, e, d, a]);
     }
+
+    // test that long running nodes block dependents, but others can continue
+    #[tokio::test]
+    async fn test_multiple_roots() {
+        // a -- b -- c
+        //   \
+        // d - e
```
nathanhammond, line 291: I'm mildly concerned that the merging branch is from root, but I'm ~sure it works anyway. This is the "slightly harder" variant.

        // a -- b -- c
        //          /
        // d -- e -

tknickman (approved):

PR #5774, opened 2023-08-22
fix(turborepo): Force all URLs to be normalized during login.

nathanhammond:
The login URL that we pass through via `webbrowser-rs` is not escaped. As a consequence, [in what appear to be extremely rare scenarios](https://github.com/amodm/webbrowser-rs/issues/75), some downstream parser is failing to correctly process the URL, resulting in broken behavior. (The linked upstream library does not appear to be the root cause of the problem, though it may be able to be defensive of this scenario.)

The URL we were using previously—though it isn't spec-compliant—isn't _truly_ ambiguous. The guilty parser should not have been tripped up by it.

This also takes the first steps toward aligning the login code paths with aims on dropping `--sso-team`.

Closes TURBO-1232

### DIFF
--- a/crates/turborepo-lib/src/commands/login.rs
+++ b/crates/turborepo-lib/src/commands/login.rs
@@ -8,7 +8,6 @@ use axum::{extract::Query, response::Redirect, routing::get, Router};
 use reqwest::Url;
 use serde::Deserialize;
 use tokio::sync::OnceCell;
-use tracing::debug;
 #[cfg(not(test))]
 use tracing::warn;
 use turborepo_ui::{start_spinner, BOLD, CYAN, GREY, UNDERLINE};
@@ -18,6 +17,7 @@ use crate::{
         link::{verify_caching_enabled, REMOTE_CACHING_INFO, REMOTE_CACHING_URL},
         CommandBase,
     },
+    config::Error,
     get_version,
 };
 
@@ -26,8 +26,18 @@ const DEFAULT_PORT: u16 = 9789;
 const DEFAULT_SSO_PROVIDER: &str = "SAML/OIDC Single Sign-On";
 
 pub async fn sso_login(base: &mut CommandBase, sso_team: &str) -> Result<()> {
+    let repo_config = base.repo_config()?;
     let redirect_url = format!("http://{DEFAULT_HOST_NAME}:{DEFAULT_PORT}");
-    let mut login_url = Url::parse(&format!("{}/api/auth/sso", base.repo_config()?.login_url()))?;
+    let login_url_configuration = repo_config.login_url();
+    let mut login_url = Url::parse(login_url_configuration)?;
+
+    login_url
+        .path_segments_mut()
+        .map_err(|_: ()| Error::LoginUrlCannotBeABase {
+            value: login_url_configuration.to_string(),
+        })?
+        .extend(["api", "auth", "sso"]);
+
     login_url
         .query_pairs_mut()
         .append_pair("teamId", sso_team)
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/commands/login.rs
```diff
@@ -111,16 +121,25 @@ fn make_token_name() -> Result<String> {
 
 pub async fn login(base: &mut CommandBase) -> Result<()> {
     let repo_config = base.repo_config()?;
-    let login_url_base = repo_config.login_url();
-    debug!("turbo v{}", get_version());
-    debug!("api url: {}", repo_config.api_url());
-    debug!("login url: {login_url_base}");
```
nathanhammond, line 117: Wrong place for this. If we want config gathering logs we should be doing it when we set the cell so that it is consistent and has _far_ more information.

We don't even directly use `api_url` in this method—transitive use only.

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -26,8 +26,18 @@ const DEFAULT_PORT: u16 = 9789;
 const DEFAULT_SSO_PROVIDER: &str = "SAML/OIDC Single Sign-On";
 
 pub async fn sso_login(base: &mut CommandBase, sso_team: &str) -> Result<()> {
+    let repo_config = base.repo_config()?;
     let redirect_url = format!("http://{DEFAULT_HOST_NAME}:{DEFAULT_PORT}");
-    let mut login_url = Url::parse(&format!("{}/api/auth/sso", base.repo_config()?.login_url()))?;
```
nathanhammond, line 30: In this corner we have string concatenation!

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -26,8 +26,18 @@ const DEFAULT_PORT: u16 = 9789;
...
+    login_url
+        .path_segments_mut()
+        .map_err(|_: ()| Error::LoginUrlCannotBeABase {
+            value: login_url_configuration.to_string(),
+        })?
+        .extend(["api", "auth", "sso"]);
+
     login_url
         .query_pairs_mut()
         .append_pair("teamId", sso_team)
```
nathanhammond, line 43: And yet we _also_ let you reuse your query parameters!

If your `LOGINURL` (née `loginUrl`) is `https://www.example.com/?myparam=1` you get this hilarious outcome:

`https://www.example.com/?myparam=1/api/auth/sso?myparam=1&teamId=vercel`.

Good luck figuring out how that happened as a user.

We should probably template this somehow instead, defining the data that we have. (Think: https://api.github.com/)

- `teamId`
- `redirect_url`

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -26,8 +26,18 @@ const DEFAULT_PORT: u16 = 9789;
...
-    let mut login_url = Url::parse(&format!("{}/api/auth/sso", base.repo_config()?.login_url()))?;
+    let login_url_configuration = repo_config.login_url();
+    let mut login_url = Url::parse(login_url_configuration)?;
+
+    login_url
+        .path_segments_mut()
+        .map_err(|_: ()| Error::LoginUrlCannotBeABase {
+            value: login_url_configuration.to_string(),
+        })?
+        .extend(["api", "auth", "sso"]);
```
nathanhammond, line 39: So now we just append the path segments. I believe this is the least-bad outcome, resulting in no new broken behavior.

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -111,16 +121,25 @@ fn make_token_name() -> Result<String> {
...
+    login_url
+        .path_segments_mut()
+        .map_err(|_: ()| Error::LoginUrlCannotBeABase {
+            value: login_url_configuration.to_string(),
+        })?
+        .extend(["turborepo", "token"]);
+
+    login_url
+        .query_pairs_mut()
+        .append_pair("redirect_uri", &redirect_url);
```
nathanhammond, line 137: Side-eye at the inconsistent query param use.

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -111,16 +121,25 @@ fn make_token_name() -> Result<String> {
...
-    let login_url = format!("{login_url_base}/turborepo/token?redirect_uri={redirect_url}");
+    let login_url_configuration = repo_config.login_url();
+    let mut login_url = Url::parse(login_url_configuration)?;
+
+    login_url
+        .path_segments_mut()
+        .map_err(|_: ()| Error::LoginUrlCannotBeABase {
+            value: login_url_configuration.to_string(),
+        })?
+        .extend(["turborepo", "token"]);
```
nathanhammond, line 133: We have a Turborepo login endpoint and a `vc` login endpoint—which is far more feature-filled. If we want branding we should do that _without_ bifurcating the API code.

crates/turborepo-lib/src/config/mod.rs
```diff
@@ -37,6 +37,11 @@ pub enum Error {
          one"
     )]
     NoTurboJSON,
+    #[error(
+        "loginUrl is configured to \"{value}\", but cannot be a base URL. This happens in \
+         situations like using a `data:` URL."
+    )]
+    LoginUrlCannotBeABase { value: String },
```
nathanhammond, line 44: Upstream throws an empty-tuple error here, so we take and turn that into something more-useful.

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -111,16 +121,25 @@ fn make_token_name() -> Result<String> {
...
+
+    login_url
+        .query_pairs_mut()
+        .append_pair("redirect_uri", &redirect_url);
+
     println!(">>> Opening browser to {login_url}");
-    direct_user_to_url(&login_url);
     let spinner = start_spinner("Waiting for your authorization...");
+    direct_user_to_url(login_url.as_str());
```
nathanhammond, line 141: (Aligns `login` with `login_sso`, prepping these to be side-by-side diff-comparable before collapsing.)

nathanhammond (commented):
Reviewer's guide!

This fixes the Firefox `view-source:` issue, while prepping for a better login experience.

chris-olszewski (approved):

PR #5776, opened 2023-08-22
fix(turborepo): Remove automatic linking.

nathanhammond:
The automatic `link` (which only occurs for `--sso-team`) does not account for the fact that `login` is a one-time, global operation, which can be run from _anywhere_ because of global `turbo`.

This means that, _wherever_ you run `turbo login --sso-team` you get a `.turbo/config.json` generated relative to `cwd`.

The preferred method for automatically handling linking should be storing the remote cache configuration inside of `turbo.json`—which makes this automatic `link` step irrelevant and enables us to test for cache connectivity during runs and know concretely whether or not the exhibited behavior is intentional.

First step: remove the clearly buggy `link` implementation.

Review in whitespace-only mode: https://github.com/vercel/turbo/pull/5776/files?w=1

Closes TURBO-1234

### DIFF
--- a/crates/turborepo-lib/src/commands/login.rs
+++ b/crates/turborepo-lib/src/commands/login.rs
@@ -10,16 +10,9 @@ use serde::Deserialize;
 use tokio::sync::OnceCell;
 #[cfg(not(test))]
 use tracing::warn;
-use turborepo_ui::{start_spinner, BOLD, CYAN, GREY, UNDERLINE};
+use turborepo_ui::{start_spinner, BOLD, CYAN};
 
-use crate::{
-    commands::{
-        link::{verify_caching_enabled, REMOTE_CACHING_INFO, REMOTE_CACHING_URL},
-        CommandBase,
-    },
-    config::Error,
-    get_version,
-};
+use crate::{commands::CommandBase, config::Error};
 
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/commands/login.rs
```diff
@@ -48,11 +41,11 @@ pub async fn sso_login(base: &mut CommandBase, sso_team: &str) -> Result<()> {
     let spinner = start_spinner("Waiting for your authorization...");
     direct_user_to_url(login_url.as_str());
 
-    let verification_token = Arc::new(OnceCell::new());
-    run_sso_one_shot_server(DEFAULT_PORT, verification_token.clone()).await?;
+    let token_cell = Arc::new(OnceCell::new());
```
chris-olszewski, line 44: I know this is out of the scope of this PR, but this should really be a [`oneshot` channel](https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html)

chris-olszewski (approved):
🔥

crates/turborepo-lib/src/commands/login.rs
nathanhammond: Yeah, this is "future-diff-reduction" cleanup tagging along. I'll get that into the part where I collapse the two functions. 👍

PR #5779, opened 2023-08-22
fix: take reference of run cache opts

chris-olszewski:
### Description

Currently constructing a run cache takes ownership of the run cache opts struct which partially moves `opts` making it unusable for the rest of run. Taking ownership is unnecessary since all of the types that get saved by `RunCache` implement `Copy` and can be easily copied.

### Testing Instructions

It compiles 😄 

Closes TURBO-1238


tknickman (approved):

tknickman:
> It compiles 😄

Sold.

arlyon (approved):

Issue #5780, opened 2023-08-22
[turborepo] Document behavior for how a failing task interacts with the cache


closesimple-wl:
### What version of Turborepo are you using?

1.10.12

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

The documentation needs to spell out how a failing task is not cached.  Not to mention, there's no API to force caching a failed result.

This might be implied, but it's definitely not clear nor explicit, and we shouldn't have to do mental gymnastics to figure out the reasoning behind the software's behavior.

### Expected Behavior

A clear explanation in the docs.  Only a few sentences is needed, really.

### To Reproduce

Run a task whose `scripts` definition is `exit 1`.

tknickman:
Closed via https://github.com/vercel/turbo/pull/5782

closesimple-wl:
That's better, but it still doesn't explain the reasoning.

PR #5782, opened 2023-08-23
docs: clairify successful tasks are cached

chris-olszewski:
### Description

Addresses #5780

If we want get explicit about non-zero exit codes, then I think this info belongs in an callout box.

### Testing Instructions

👁️

### DIFF
--- a/docs/pages/repo/docs/core-concepts/caching.mdx
+++ b/docs/pages/repo/docs/core-concepts/caching.mdx
@@ -28,7 +28,7 @@ Let's say you run a `build` task with Turborepo using `turbo run build`:
 
 3. If Turborepo doesn't find any matching artifacts for the calculated hash, Turborepo will then **execute the task**.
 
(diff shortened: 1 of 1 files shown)


nathanhammond:
There's not actually a good reason that we _don't_ cache failed tasks. The things I'm thinking about needing to do prior to adding:
- Control over behavior a la `--continue`. (`--force=all`, `--force=failed-only`)
- Maybe explicitly marking all downstream tasks as failed?
- Maybe a `failedOutputs` array.

In any case, this is PR looks fine to me, though OP probably wants far more information than this.

tknickman (approved):

chris-olszewski:
> Maybe explicitly marking all downstream tasks as failed

I think this makes sense? One thing that we'd probably need to work out would be how this would interact with `--continue`

> Maybe a `failedOutputs` array.

Would we want to split out failed outputs compared to just adding them to `outputs`?

> In any case, this is PR looks fine to me, though OP probably wants far more information than this.

Yeah, this isn't meant to fully address the issue, but at least add the notion of task success affecting caching.

PR #5784, opened 2023-08-23
make swc comments immutable

sokra:
### Description

This fixes a consistency problem. SWC Emitter is mutating the passed `Comments` while emitting (probably to avoid cloning).
This is bad for us as it mutates a value that should be immutable (as it's in the cache). It causes missing comments when generating a module again or with different configs.

This change changes it to force comments to stay immutable after being parsed.


crates/turbopack-ecmascript/src/swc_comments.rs
```diff
@@ -0,0 +1,122 @@
...
+    fn move_leading(&self, _from: swc_core::common::BytePos, _to: swc_core::common::BytePos) {
+        panic!("Comments are immutable after parsing")
+    }
+
+    fn take_leading(
+        &self,
+        pos: swc_core::common::BytePos,
+    ) -> Option<Vec<swc_core::common::comments::Comment>> {
+        self.get_leading(pos)
+    }
```
jridgewell, line 72: Shoudn't this throw, since the name sounds like it's taking the comment at that position?

crates/turbopack-ecmascript/src/swc_comments.rs
```diff
@@ -0,0 +1,122 @@
...
+    fn move_trailing(&self, _from: swc_core::common::BytePos, _to: swc_core::common::BytePos) {
+        panic!("Comments are immutable after parsing")
+    }
+
+    fn take_trailing(
+        &self,
+        pos: swc_core::common::BytePos,
+    ) -> Option<Vec<swc_core::common::comments::Comment>> {
+        self.get_trailing(pos)
+    }
```
jridgewell, line 110: Ditto.

crates/turbopack-ecmascript/src/swc_comments.rs
```diff
@@ -0,0 +1,122 @@
...
+                .filter_map(|mut r| {
+                    let c = take(r.value_mut());
+                    (!c.is_empty()).then_some((*r.key(), c))
+                })
+                .collect(),
+        }
+    }
+}
+
+impl Comments for ImmutableComments {
```
kdy1, line 42: I think this should implement `with_leading`.

kdy1: I don't remember if codegen uses `with_leading`, but some transforms uses it.

https://github.com/swc-project/swc/blob/662f236aa5c1375ea4b175dbb6d85ad974503b84/crates/swc_ecma_transforms_typescript/src/strip.rs#L313-L315

kdy1: And well, `with_leading` calls `take_leading` and `add_leading_comments` by default

kdy1 (changes requested):

crates/turbopack-ecmascript/src/swc_comments.rs
jridgewell: Transformation code will receive a mutable [SwcComments](https://github.com/vercel/turbo/blob/f068a992b79a6abc4b41308f13ecfddba29cb93f/crates/turbopack-ecmascript/src/parse.rs#L246). It's only transformed into this immutable comments after transformation, when we insert the `ParseResult::Ok` into our cached graph.

crates/turbopack-ecmascript/src/swc_comments.rs
sokra: yeah, sadly the `Emitter` calls this method on comment. I guess to avoid cloning the comment. So we need to implement it.

Technically cloning an `Comment` isn't that expensive (Copy 13 bytes and Clone and Arc), so I would prefer the `Emitter` using `get_leading` instead of `take_leading`, but it is as it is.

There is a comment on the struct explaining the `take` behavior

sokra: TBH it's not very obvious that the `Emitter` mutates the passed comments. This is how this bug sneaked into turbopack @kdy1

crates/turbopack-ecmascript/src/swc_comments.rs
sokra: implemented them. Seems like they are unused currently, but it doesn't hurt to have non-panicing variants for them.

jridgewell (approved):

PR #5786, opened 2023-08-23
perform invalidation directly on writes

sokra:
### Description

This changes the `write` function in the filesystem to invalidate the path directly instead of going through the watcher. This makes it easier to find write conflicts as they will directly hang the compilation. Tracing can be used to print out the cycle of writes.

Closes WEB-1427

### DIFF
--- a/crates/turbo-tasks-fs/src/lib.rs
+++ b/crates/turbo-tasks-fs/src/lib.rs
@@ -508,7 +508,7 @@ impl DiskFileSystem {
 
     fn invalidate_from_write(&self, full_path: &Path, invalidators: HashSet<Invalidator>) {
         if !invalidators.is_empty() {
-            if let Some(path) = format_absolute_fs_path(&full_path, &self.name, self.root_path()) {
+            if let Some(path) = format_absolute_fs_path(full_path, &self.name, self.root_path()) {
                 if invalidators.len() == 1 {
                     let invalidator = invalidators.into_iter().next().unwrap();
                     invalidator.invalidate_with_reason(Write { path });
--- a/crates/turbo-tasks-fs/src/lib.rs
+++ b/crates/turbo-tasks-fs/src/lib.rs
@@ -68,7 +68,11 @@ use turbo_tasks_hash::hash_xxh3_hash64;
 use util::{extract_disk_access, join_path, normalize_path, sys_to_unix, unix_to_sys};
 pub use virtual_fs::VirtualFileSystem;
 
-use self::{invalidation::WatchStart, json::UnparseableJson, mutex_map::MutexMap};
+use self::{
+    invalidation::{WatchStart, Write},
+    json::UnparseableJson,
+    mutex_map::MutexMap,
+};
 use crate::{
     attach::AttachedFileSystem,
     invalidation::WatchChange,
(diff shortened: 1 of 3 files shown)


crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -710,22 +747,29 @@ impl FileSystem for DiskFileSystem {
...
+        let old_invalidators = self.register_sole_invalidator(&full_path)?;
+
         // We perform an untracked comparison here, so that this write is not dependent
         // on a read's Vc<FileContent> (and the memory it holds). Our untracked read can
         // be freed immediately. Given this is an output file, it's unlikely any Turbo
         // code will need to read the file from disk into a Vc<FileContent>, so we're
         // not wasting cycles.
         let compare = content.streaming_compare(full_path.clone()).await?;
         if compare == FileComparison::Equal {
+            if !old_invalidators.is_empty() {
```
jridgewell, line 762: Nit: comment why we're preserving the old invalidators. Especially because it seems like we're trying to emit the same contents from multiple tasks, shouldn't that be an error, too?

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -487,14 +505,33 @@ impl DiskFileSystem {
             path.join(&*unix_to_sys(&fs_path.path))
         })
     }
+
+    fn invalidate_from_write(&self, full_path: &Path, invalidators: HashSet<Invalidator>) {
+        if !invalidators.is_empty() {
+            if let Some(path) = format_absolute_fs_path(full_path, &self.name, self.root_path()) {
+                if invalidators.len() == 1 {
+                    let invalidator = invalidators.into_iter().next().unwrap();
+                    invalidator.invalidate_with_reason(Write { path });
```
jridgewell, line 514: Nit: I don't think we need to optimize this path given that it's here just to prevent us from ever hitting this path.

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -200,6 +204,20 @@ impl DiskFileSystem {
         Ok(())
     }
 
+    /// registers the path as an invalidator for the current task,
+    /// has to be called within a turbo-tasks function. It removes and returns
+    /// the current list of invalidators.
+    fn register_sole_invalidator(&self, path: &Path) -> Result<HashSet<Invalidator>> {
```
jridgewell, line 210: Could this be merged with `register_invalidator`. The difference seems to be whether it returns the old value, which could just be discarded by the callers of the old `register_invalidator`.

crates/turbo-tasks-fs/src/lib.rs
sokra: old_invalidators can also contain invalidators from `read` methods. It's not `write`-only.

We don't want to invalidate read calls when writing identical content to a file.

crates/turbo-tasks-fs/src/lib.rs
sokra: old_invalidators can also contain invalidators from `read` methods. It's not `write`-only.

So having exactly 1 `read` invalidator for a file is pretty likely.

crates/turbo-tasks-fs/src/lib.rs
jridgewell: I don't see how we get two different behaviors out of these two pieces of code:

```rust
# inserts a new invalidator
self.invalidator_map.insert(path_to_key(path), invalidator);

# replaces old invalidators
let mut invalidator_map = self.invalidator_map.lock().unwrap();
let old_invalidators = invalidator_map.insert(path_to_key(path), [invalidator].into());
```

They're both calling `map.insert(key)`, so how does one append and one replaces?

crates/turbo-tasks-fs/src/lib.rs
jridgewell: > We don't want to invalidate read calls when writing identical content to a file.

But that means there are two owners of the file. Shouldn't that be an error, why aren't they reusing the same write cell?

crates/turbo-tasks-fs/src/lib.rs
sokra: These are different `insert` methods. `InvalidatorMap::insert` adds a single invalidator to the list.

`HashMap::insert` replaces the current list with a different list.

crates/turbo-tasks-fs/src/lib.rs
sokra: maybe, but `old_invalidators.is_empty()` is not related to that case, since the invalidator map stores read and write invalidators. Having `old_invalidators` is valid, since they might be invalidators from read methods.

jridgewell (approved):

PR #5791, opened 2023-08-23
feat: port task graph execution

chris-olszewski:
### Description

This PR:
 - Ports Engine.Execute which is a wrapper around the DAG walk, but it adds concurrency controls.
 - Ports a noop package task visitor that prints the task id and the command it would run.

Notes for reviewers:
As one can see the structure of the DAG walker, engine execution, and package task visitor is very similar. They all use a `mspc` & `oneshot` to achieve what in Go we used callbacks for. In the future I think we'll either:
 - add the concurrency controls into the DAG walker itself and get rid of the `Engine::execute` layer
 - capture this "callback" pattern into a generic abstraction to reduce a lot of the boilerplate that is shared between all of these structures

### Testing Instructions

Try running some tasks and verify dependencies are respected
```
[0 olszewski@chriss-mbp] /tmp/tom-test $ turbo_dev build --experimental-rust-codepath        
Executing tsconfig#build: no script def
Executing eslint-config-custom#build: no script def
Executing tailwind-config#build: no script def
Executing ui#build: tsup
Executing web#build: next build
Executing docs#build: next build
[0 olszewski@chriss-mbp] /tmp/tom-test $ turbo_dev build --experimental-rust-codepath --filter=web
Executing eslint-config-custom#build: no script def
Executing tailwind-config#build: no script def
Executing tsconfig#build: no script def
Executing ui#build: tsup
Executing web#build: next build
```

Closes TURBO-1250


crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -116,6 +119,12 @@ impl Engine<Built> {
         )
     }
 
+    // TODO get rid of static lifetime and figure out right way to tell compiler the
```
chris-olszewski, line 122: I think what we'll want to eventually do is make an owned/borrowed version of task id ala https://github.com/sunshowers-code/borrow-complex-key-example.

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -0,0 +1,108 @@
...
+            package_graph,
+            opts,
+        }
+    }
+
+    pub async fn visit(&self, engine: Arc<Engine>) -> Result<(), Error> {
+        let concurrency = self.opts.run_opts.concurrency as usize;
+        let (node_sender, mut node_stream) = mpsc::channel(concurrency);
+
+        let engine_handle = {
```
chris-olszewski, line 49: We need the engine to execute in another task so it can send over the tasks via the node stream.

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -0,0 +1,108 @@
...
+
+            let _task_def = engine
+                .task_definition(&info)
+                .ok_or(Error::MissingDefinition)?;
+
+            tasks.push(tokio::spawn(async move {
+                println!(
+                    "Executing {info}: {}",
+                    command.as_deref().unwrap_or("no script def")
+                );
```
chris-olszewski, line 89: Once we have the chance of task failures we'll hook up error handling. I'm thinking that the visitor will manage the collection of errors via a `mpsc` channel setup, but this might change when we get to implementing it.

gsoltis (approved):
I see what you mean about multiple visiting implementations. I think that's fine for now, we can refactor later.

PR #5793, opened 2023-08-24
Automatically derive ValueDefault for primitive value types

alexkirsz:
### Description

This PR automatically derives `ValueDefault` for primitive value types. This means that you can replace `Vc::<String>::empty()` with `Default::default()`. The same goes for all primitive value types (`bool`, `u8`, `()`, etc.). The underlying type's `Default` implementation will be used. 

### Testing Instructions

N/A

Closes WEB-1430


jridgewell (approved):

sokra (approved):
ship it

PR #5794, opened 2023-08-24
fix "any" context condition implementation

sokra:
### Description

* the any logic was messed up
* don't resolve typescript in node_modules for evaluation

### DIFF
--- a/crates/turbopack/src/evaluate_context.rs
+++ b/crates/turbopack/src/evaluate_context.rs
@@ -12,8 +12,9 @@ use turbopack_core::{
 use turbopack_node::execution_context::ExecutionContext;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


crates/turbopack/src/evaluate_context.rs
```diff
@@ -49,31 +50,36 @@ pub async fn node_evaluate_asset_context(
         } else {
             "development".to_string()
         };
+    let resolve_options_context = ResolveOptionsContext {
```
ForsakenHarmony, line 53: Can we add some newlines and maybe a comment to the rules block below, took me a second to figure out why this is split into 2 declarations

ForsakenHarmony (approved):

PR #5796, opened 2023-08-24
feat(turborepo): Layered Config

nathanhammond:
This PR enables existing configuration settings to be stored in any of "global", "local", or "shared" locations.
- Removes the config crate.
- Sets up configuration layering per our requirements.
- Does not make any behavior changes as to where link/unlink and login/logout store the team and token values.

The primary value enabled here is that we can set `teamId` inside of `turbo.json` and omit the `link` step, and make login aware of which team scope it needs to auth to, eliminating the need for `--sso-team` in most cases.

Closes TURBO-1252


crates/turborepo-lib/src/commands/logout.rs
```diff
@@ -1,15 +1,28 @@
...
+    })?;
+    let output = if let Some(after) = unset_path(&before, &["token"], true)? {
+        global_config_path.create_with_contents(after)?;
+        ">>> Logged out"
+    } else {
+        ">>> Not logged in"
+    };
 
-    println!("{}", base.ui.apply(GREY.apply_to(">>> Logged out")));
+    println!("{}", base.ui.apply(GREY.apply_to(output)));
```
NicholasLYang, line 26: Minor nit but we now have `cprintln!`in `turborepo-ui` that makes this a little cleaner

NicholasLYang, line 26:
```suggestion
    cprintln!(base.ui, GREY, "{}", output);
```

crates/turborepo-lib/src/config/turbo_config.rs
```diff
@@ -0,0 +1,581 @@
...
+    turbo_mapping.iter().for_each(|(k, mapped)| {
+        if let Some(value) = environment.get(k) {
+            if !value.is_empty() {
+                output_map.insert(mapped.to_string(), value.to_string());
+            }
+        }
+    });
+
+    // Process signature
+    let signature = if let Some(signature) = output_map.get("signature").cloned() {
```
NicholasLYang, line 173: I'm confused as to why we're cloning then borrowing as a str. Could we not call `as_deref` here?

crates/turborepo-lib/src/config/turbo_config.rs
```diff
@@ -0,0 +1,581 @@
...
+        timeout: None,
+    };
+
+    Ok(output)
+}
+
+impl TurborepoConfigBuilder {
+    pub fn new(base: &CommandBase) -> Self {
+        Self {
+            repo_root: base.repo_root.to_owned(),
```
NicholasLYang, line 286: tiny nit: this doesn't have to be owned, you can thread the lifetime through if you want

crates/turborepo-lib/src/config/turbo_config.rs
```diff
@@ -0,0 +1,581 @@
...
+                    &global_config_path,
+                    e
+                ))
+            }
+        })?;
+        if contents.is_empty() {
+            contents = String::from("{}");
+        }
+        let global_config: ConfigurationOptions =
+            serde_json::from_str(&contents).map_err(|e| ConfigError::SerdeJson(e))?;
```
NicholasLYang, line 345: Can we add `#[from]` on the `SerdeJson` variant instead?

NicholasLYang (changes requested):
First pass. I think the core logic looks okay, although I'm not a massive fan of it potentially making it possible to put any config in any location. Like if I'm reading this correctly, you can slip the entire config into the `remote_config` field in `turbo.json`

nathanhammond:
> I think the core logic looks okay, although I'm not a massive fan of it potentially making it possible to put any config in any location. Like if I'm reading this correctly, you can slip the entire config into the `remote_config` field in `turbo.json`

I've got a full enumeration over here with comments about my thoughts for the things in this PR.
https://www.notion.so/vercel/Turborepo-Configuration-a15dd51ca99c4f56983fc38d86784035

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -163,10 +167,36 @@ pub async fn link(
...
+                        e
+                    ))
+                }
+            })?;
+            let no_preexisting_id =
+                if let Some(too_permissive) = unset_path(&before, &["teamid"], false)? {
+                    too_permissive
+                } else {
+                    before
+                };
```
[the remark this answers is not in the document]
nathanhammond: I cleaned up this pattern and the related read in advance of it.

crates/turborepo-lib/src/commands/logout.rs
nathanhammond: Adoption of `cprintln` is separate from this; we can go back and make that change throughout the codebase all at once. (It's a mechanical change, unlike, say `turbopath` changes.)

crates/turborepo-lib/src/config/turbo_config.rs
nathanhammond: Cleaned this up.

crates/turborepo-lib/src/config/turbo_config.rs
nathanhammond: It ends up used in lots of places to the point that it makes sense for the config builder to have an owned version.

crates/turborepo-lib/src/config/turbo_config.rs
nathanhammond: It's already there, I just didn't know that worked. 😅

NicholasLYang (commented):
Only concern is this error case here. Otherwise very glad to be one dependency lighter!

crates/turborepo-lib/src/config/turbo_config.rs
```diff
@@ -0,0 +1,592 @@
...
+    // Process the VERCEL_ARTIFACTS_* next.
+    vercel_artifacts_mapping
+        .iter()
+        .try_for_each(|(k, mapped)| -> Result<(), ConfigError> {
+            if let Some(value) = environment.get(k) {
+                let converted = value.to_str().ok_or_else(|| {
+                    ConfigError::Anyhow(anyhow!(
+                        "{} is not UTF8.",
+                        k.to_ascii_uppercase().to_str().unwrap()
+                    ))
```
chris-olszewski, line 275: Reading this freaked me out because we were unwrapping a key. We don't need to store the valid UTF8 keys as `OsString` in order to query the env:
```suggestion
            if let Some(value) = environment.get::<std::ffi::OsStr>(k.as_ref()) {
                let converted = value.to_str().ok_or_else(|| {
                    ConfigError::Anyhow(anyhow!(
                        "{} is not UTF8.",
                        k.to_ascii_uppercase()
                    ))
```

crates/turborepo-lib/src/config/turbo_config.rs
```diff
@@ -0,0 +1,592 @@
...
+fn get_env_var_config(
+    environment: &HashMap<OsString, OsString>,
+) -> Result<ConfigurationOptions, ConfigError> {
+    let mut turbo_mapping = HashMap::new();
+    turbo_mapping.insert(OsString::from("turbo_api"), "api_url");
+    turbo_mapping.insert(OsString::from("turbo_login"), "login_url");
+    turbo_mapping.insert(OsString::from("turbo_team"), "team_slug");
+    turbo_mapping.insert(OsString::from("turbo_teamid"), "team_id");
+    turbo_mapping.insert(OsString::from("turbo_token"), "token");
+    turbo_mapping.insert(OsString::from("turbo_remote_cache_timeout"), "timeout");
```
chris-olszewski, line 154:
```suggestion
    turbo_mapping.insert(String::from("turbo_api"), "api_url");
    turbo_mapping.insert(String::from("turbo_login"), "login_url");
    turbo_mapping.insert(String::from("turbo_team"), "team_slug");
    turbo_mapping.insert(String::from("turbo_teamid"), "team_id");
    turbo_mapping.insert(String::from("turbo_token"), "token");
    turbo_mapping.insert(String::from("turbo_remote_cache_timeout"), "timeout");
```

crates/turborepo-lib/src/config/turbo_config.rs
```diff
@@ -0,0 +1,592 @@
...
+    turbo_mapping
+        .iter()
+        .try_for_each(|(k, mapped)| -> Result<(), ConfigError> {
+            if let Some(value) = environment.get(k) {
+                let converted = value.to_str().ok_or_else(|| {
+                    ConfigError::Anyhow(anyhow!(
+                        "{} is not UTF8.",
+                        k.to_ascii_uppercase().to_str().unwrap()
+                    ))
+                })?;
```
chris-olszewski, line 173:
```suggestion
            if let Some(value) = environment.get::<std::ffi::OsStr>(k.as_ref()) {
                let converted = value.to_str().ok_or_else(|| {
                    ConfigError::Anyhow(anyhow!(
                        "{} is not UTF8.",
                        k.to_ascii_uppercase()
                    ))
                })?;
```

chris-olszewski (commented):
Overall looks pretty good, mostly some minor things. Will finish review once I'm back at keys

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -29,9 +27,9 @@ pub(crate) mod unlink;
 pub struct CommandBase {
     pub repo_root: AbsoluteSystemPathBuf,
     pub ui: UI,
-    user_config: OnceCell<UserConfig>,
-    repo_config: OnceCell<RepoConfig>,
-    client_config: OnceCell<ClientConfig>,
+    #[cfg(test)]
+    pub global_config_path: Option<AbsoluteSystemPathBuf>,
+    pub config: OnceCell<ConfigurationOptions>,
```
chris-olszewski, line 32: I don't think this needs to be public
```suggestion
   config: OnceCell<ConfigurationOptions>,
```

chris-olszewski (approved):
I think this is a far easier way to model our configuration story. My issues are primarily around the `read_or_default` API design which I think isn't quite it.

crates/turborepo-lib/src/config/turbo_config.rs
```diff
@@ -0,0 +1,592 @@
...
+    pub(crate) login_url: Option<String>,
+    #[serde(alias = "teamslug")]
+    #[serde(alias = "TeamSlug")]
+    #[serde(alias = "TEAMSLUG")]
+    pub(crate) team_slug: Option<String>,
+    #[serde(alias = "teamid")]
+    #[serde(alias = "TeamId")]
+    #[serde(alias = "TEAMID")]
+    pub(crate) team_id: Option<String>,
+    pub(crate) token: Option<String>,
```
gsoltis, line 46: Can you verify that this won't read a token from `turbo.json`, which we expect to be checked in?

crates/turborepo-lib/src/config/turbo_config.rs
```diff
@@ -0,0 +1,592 @@
...
+    fn get_configuration_options(self) -> Result<ConfigurationOptions, ConfigError>;
+}
+
+impl ResolvedConfigurationOptions for PackageJson {
+    fn get_configuration_options(self) -> Result<ConfigurationOptions, ConfigError> {
+        match &self.legacy_turbo_config {
+            Some(legacy_turbo_config) => {
+                let synthetic_raw_turbo_json: RawTurboJSON =
+                    serde_json::from_value(legacy_turbo_config.clone())
+                        .map_err(|_| ConfigError::Anyhow(anyhow!("global_de")))?;
```
gsoltis, line 113: I think this error could use some description.

crates/turborepo-lib/src/config/turbo_config.rs
```diff
@@ -0,0 +1,592 @@
...
+                            acc.enabled = Some(enabled);
+                        }
+                        if let Some(preflight) = current_source_config.preflight {
+                            acc.preflight = Some(preflight);
+                        }
+                        if let Some(timeout) = current_source_config.timeout {
+                            acc.timeout = Some(timeout);
+                        }
+                    }
+                    Err(_) => todo!(),
```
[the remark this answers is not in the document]
nathanhammond: That needed a `try_fold`, but I used `iter` and didn't originally know about `into_iter`. This has now disappeared.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -139,11 +151,10 @@ impl<'a> Run<'a> {
 
         let env_at_execution_start = EnvironmentVariableMap::infer();
 
-        let repo_config = self.base.repo_config()?;
-        let team_id = repo_config.team_id();
-        let team_slug = repo_config.team_slug();
-
-        let token = self.base.user_config()?.token();
+        let config = self.base.config().expect("FIXME");
```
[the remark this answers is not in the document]
nathanhammond: Good catch, I hadn't done `Error` type plumbing when I authored that.

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -403,6 +403,16 @@ impl AbsoluteSystemPath {
...
 
+    /// Attempts to read a file, and:
+    /// If the file does not exist it returns the default value.
+    /// For all other scenarios passes through the `read_to_string` results.
+    pub fn read_or_default(&self, default: String) -> Result<String, io::Error> {
+        match fs::read_to_string(&self.0) {
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(default),
+            read_result => read_result,
+        }
+    }
```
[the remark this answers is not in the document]
nathanhammond: This was me cleaning up this mess: https://github.com/vercel/turbo/pull/5796/commits/48117f76720ea1b8b5f967df0f15dcce3b65e887

It's intended to be a convenience because I need a JSON string at the next step, no matter what. I can improve the API a bit with generics, but that doesn't prevent a stack allocation.

```rust
pub fn read_or_default<I>(&self, default_value: I) -> Result<String, io::Error>
where I: Into<String>,
{
    match fs::read_to_string(&self.0) {
        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(default_value.into()),
        read_result => read_result,
    }
}
```

crates/turborepo-lib/src/config/turbo_config.rs
nathanhammond: Just needed to be deleted. I'd threaded it in but missed deletion.

crates/turborepo-lib/src/config/turbo_config.rs
nathanhammond: Yeah, I'll make sure there is a test for it. Some thoughts about "checkin-viable tokens" over here:
https://www.notion.so/Turborepo-Configuration-a15dd51ca99c4f56983fc38d86784035?d=ab20442220954f4ab635ab98d7bbcf28&pvs=4#e953df8e99f448df8b282f361b959f8c

crates/turborepo-paths/src/absolute_system_path.rs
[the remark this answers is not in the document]
nathanhammond: Here is some incremental cleanup: https://github.com/vercel/turbo/pull/5796/commits/23dc6b755a92e56ef30948e3a908617f5f406de9

This happens twice per run, three times if link/unlink/login/logout ... I'm not super-inclined to hyper-optimize this.

Also, I _think_ that the generic trick should mean the compiler sees `"{}"` as `&'static str`?

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -205,12 +211,13 @@ fn generate_member(path_segments: &[&str], value: &str, separator: &str) -> Stri
 pub fn unset_path(
     json_document_string: &str,
     path: &[&str],
+    case_sensitivity: bool,
```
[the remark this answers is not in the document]
nathanhammond: Good callout, went with: `match_case_sensitive`.

crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -44,7 +44,13 @@ pub fn set_path(
     path: &[&str],
     json_value: &str,
 ) -> Result<String, RewriteError> {
-    let root = get_root(json_document_string)?;
+    let string_to_process = if json_document_string.is_empty() {
+        "{}"
```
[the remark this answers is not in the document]
nathanhammond: This was prior to `read_or_default`, it shouldn't be doing this anymore.

crates/turborepo-lib/src/config/turbo_config.rs
nathanhammond: Since it doesn't really matter at definition time if this is an `OsString` I chose that so we can call `std::env::vars_os()` which doesn't do an unnecessary UTF8 pass and (more-importantly) ergonomically doesn't require us to do any plumbing of errors we know won't occur.

The only thing that's weird here is that it calls unwrap on one of our own specified keys, which we can tell by inspection is safe (and we're not at risk of creating a mapping that isn't UTF8).

nathanhammond: https://github.com/vercel/turbo/pull/5796/commits/49da009ca2caa38f8b26627b59c22928f5407daf

crates/turborepo-lib/src/config/turbo_config.rs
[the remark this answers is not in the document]
nathanhammond: https://github.com/vercel/turbo/pull/5796/commits/d0e926b96b37224c9e63a42df493e12e40fd0dd5?w=1

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -163,10 +167,28 @@ pub async fn link(
...
+
+            let no_preexisting_id = unset_path(&before, &["teamid"], false)?.unwrap_or(before);
+            let no_preexisting_slug =
+                unset_path(&no_preexisting_id, &["teamslug"], false)?.unwrap_or(no_preexisting_id);
+
+            let after = set_path(
+                &no_preexisting_slug,
+                &["teamId"],
+                &format!("\"{}\"", team_id),
+            )?;
```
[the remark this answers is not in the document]
nathanhammond: I've kept the `read_or_default`, but changed the API shape to be `Into<String>` which I believe should delay heap allocation until "just-in-time". We'll still have a stack allocation to rodata, but that seems likely fine?

The problem with the `read_optional_string` is that this leaks a _whole_ lot of boilerplate into the application.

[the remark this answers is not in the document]
nathanhammond: https://github.com/vercel/turbo/pull/5796/commits/23dc6b755a92e56ef30948e3a908617f5f406de9

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -80,7 +80,19 @@ impl<'a> Run<'a> {
         let root_turbo_json =
             TurboJson::load(&self.base.repo_root, &root_package_json, is_single_package)?;
 
-        opts.cache_opts.remote_cache_opts = root_turbo_json.remote_cache_options.clone();
+        let team_id = if let Some(configuration_options) = &root_turbo_json.remote_cache {
+            configuration_options.team_id.clone().unwrap_or_default()
+        } else {
+            "".to_string()
+        };
```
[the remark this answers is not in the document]
nathanhammond: ced93858b5c0d16506dc0b913b01f1c148c14f6c

crates/turborepo-paths/src/absolute_system_path.rs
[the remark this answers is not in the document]
chris-olszewski: > Also, I think that the generic trick should mean the compiler sees "{}" as &'static str and dumps it in rodata?

Correct

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -406,9 +406,12 @@ impl AbsoluteSystemPath {
     /// Attempts to read a file, and:
     /// If the file does not exist it returns the default value.
     /// For all other scenarios passes through the `read_to_string` results.
-    pub fn read_or_default(&self, default: String) -> Result<String, io::Error> {
+    pub fn read_or_default<I>(&self, default_value: I) -> Result<String, io::Error>
```
chris-olszewski, line 409: This is in no way blocking and if there is a change it can easily happen down the road.

I still think what we want is `pub fn read_optional_string(&self) -> Result<Option<String>, io::Error>`* as it gives the caller insight into what path the call went down. It gives greater flexibility for future use where users might have a default that would be better expressed as a non-`String` e.g.
```
let config_contents = config_path.read_optional_string()?;
let config: MyConfig = config_contents.map(|s| serde_json::from_str(&s)).unwrap_or_else(|| Ok(MyConfig { lever: true, knob: 12 })?
```

*And to be honest I feel that it should be named `read_string(&self) -> Result<Option<String>, io::Error>` and the current `read_string` should get renamed to `read_required_string`

If we want to keep the current behavior. Then there are two changes that could be made to keep with Rust conventions, I believe that this should get changed to:
```
pub fn read_or_else(&self, f: impl FnOnce() -> String) -> Result<String, io::Error> 
```
- The change to `or_else` over `or_default` matches the stdlib convention where `or_default` means that `T::default()` is used in the error case e.g. [`Option::unwrap_or_default`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default)
- The change to taking a closure instead of an `Into<String>` gives greater flexibility to the caller where they might want to have a dynamic default: e.g. `config_path.read_or_else(|| format!("default_value: {foo}"))`

chris-olszewski (approved):
My comment on `read_or_default` is not blocking in any way, but I do think we can make it more flexible in the future.

crates/turborepo-paths/src/absolute_system_path.rs
nathanhammond: If we surface this as a `FnOnce` via something like `read_to_string_or_else` it's purpose-defeating. That mapping ends up looking exactly like the body of the helper function, and copied straight into a dozen places in the code base.

```rust
path
    .read_to_string()
    .or_else(|e| {
        if e.kind() == std::io::ErrorKind::NotFound {
            Ok("{}".into())
        } else {
            Err(e)
        }
    })
```

The other alternative signature that you're proposing, usage looks like this:

```rust
// Existing in this PR:
pub fn read_or_fallback(&self, default: Into<String>) -> Result<String, io:Error> {}
let contents = path.read_or_fallback("{}")?;

// Proposed:
pub fn read_existing_to_string(&self) -> Result<Option<String>, io:Error> {}

// Allocates if you're not careful.
let contents = path
    .read_existing_to_string()?
    .unwrap_or("{}".to_string());

// No allocation but a _lot_ of ceremony.
let contents = path
    .read_existing_to_string()?
    .unwrap_or_else(|| "{}".to_string());
```

I don't hate it, but it's not the same pit-of-success API. To make the method name look like Rust I've done this:
```rust
pub fn read_existing_to_string_or(&self, default: Option<Into<String>>) -> Result<String, io:Error> {}
let contents = path.read_existing_to_string_or(Ok("{}"))?;
```

nathanhammond: Assuming we had:
```rust
pub fn read_existing_to_string(&self) -> Result<Option<String>, io:Error> {}
```

How would you propose reducing the boilerplate across invocations at multiple files?
```rust
let contents = path
    .read_existing_to_string()?
    .unwrap_or_else(|| "{}".to_string());
```

I just want a nice-to-use API (`let contents = path.read_or_default("{}")?;`) at the callsite. Is there a less-verbose, more-Rust-like API for this that I'm just not familiar with? Is this a `base64` type of situation?

https://twitter.com/mycoliza/status/1708933310548935148

chris-olszewski: > I don't hate it, but it's not the same pit-of-success AP

Clippy is already good at yelling at you if your `unwrap_or` contains a function call: https://rust-lang.github.io/rust-clippy/v0.0.212/index.html#or_fun_call

> How would you propose reducing the boilerplate across invocations at multiple files?

```
pub fn read_json(path: &AbsoluteSystemPath) -> Result<String>, io::Error> {
    path
        .read_existing_to_string()?
        .unwrap_or_else(|| "{}".to_string())
}
```
Not wonderful, but it allows the API to be flexible for future callers which might not want to provide a default in the form of a `String`.

> Is this a base64 type of situation?

To a degree, yes. I've drunk the Rust kool-aid and I don't find `foo.unwrap_or_else(|| "{}".to_string())` to be an eyesore/verbose. I'm used to the Rust `std` and I tend to try to model most APIs to pertain to the conventions laid out by it. Upshot of following those conventions is that tools like `clippy` are built around proper usage of the `std` APIs and we can rely on `clippy` telling callers how to use our library correctly.

PR #5809, opened 2023-08-25
fix(lint): eslint-plugin-turbo

tknickman:
### Description

Linting our lint


chris-olszewski (approved):
> Linting our lint

🤯

PR #5810, opened 2023-08-25
fix(lint): @turbo/workspaces

tknickman:
### Description

@turbo/workspaces lint fixes

### DIFF
--- a/packages/turbo-workspaces/src/cli.ts
+++ b/packages/turbo-workspaces/src/cli.ts
@@ -2,9 +2,9 @@
 
 import chalk from "chalk";
 import { Command } from "commander";
-
-import { summary, convert } from "./commands";
+import { logger } from "@turbo/utils";
 import cliPkg from "../package.json";
+import { summary, convert } from "./commands";
 import { ConvertError } from "./errors";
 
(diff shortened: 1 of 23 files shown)


packages/turbo-workspaces/src/cli.ts
```diff
@@ -41,13 +40,18 @@ workspacesCli
   .action(summary);
 
 workspacesCli.parseAsync().catch((error) => {
+  // eslint-disable-next-line no-console
```
chris-olszewski, line 43: In #5832 we swap to use our wrapper around `console.log`, is it worthwhile to switch that here too?

chris-olszewski (approved):

packages/turbo-workspaces/src/cli.ts
tknickman: Yes definitely, I like that method better since we don't need disables everywhere and it consolidates where we're actually logging to the user in case we want to make changes there in the future.

tknickman:
Depends on https://github.com/vercel/turbo/pull/5827 to fix a bug in the useFixtures test hook that is causing intermittent failures here.

PR #5811, opened 2023-08-25
fix(lint): correct globs

tknickman:
### Description

Fix globs -  silly mistake from copy pasting from the first package that I converted. 

Closes TURBO-1270


chris-olszewski (approved):

tknickman:
Sending this one through as CI has passed, and it could create a situation where bad code sneaks in that's not being checked in CI

PR #5821, opened 2023-08-28
avoid cloning a large list of task ids

sokra:
### Description

avoids cloning a vec/set for invalidation

Closes WEB-1461

### DIFF
--- a/crates/turbo-tasks/src/manager.rs
+++ b/crates/turbo-tasks/src/manager.rs
@@ -806,7 +806,7 @@ impl<B: Backend + 'static> TurboTasks<B> {
             let tasks = take(tasks_to_notify);
             if !tasks.is_empty() {
                 let _guard = trace_span!("finish_current_task_state").entered();
-                self.backend.invalidate_tasks(tasks, self);
+                self.backend.invalidate_tasks(&tasks, self);
             }
             *stateful
         })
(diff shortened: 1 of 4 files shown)


crates/turbo-tasks/src/manager.rs
```diff
@@ -1096,7 +1096,7 @@ impl<B: Backend + 'static> TurboTasksBackendApi<B> for TurboTasks<B> {
         });
         if result.is_err() {
             let _guard = trace_span!("schedule_notify_tasks", count = tasks.len()).entered();
-            self.backend.invalidate_tasks(tasks.to_vec(), self);
+            self.backend.invalidate_tasks(tasks, self);
```
sokra, line 1099: avoids cloning

crates/turbo-tasks/src/manager.rs
```diff
@@ -1111,8 +1111,7 @@ impl<B: Backend + 'static> TurboTasksBackendApi<B> for TurboTasks<B> {
         });
         if result.is_err() {
             let _guard = trace_span!("schedule_notify_tasks_set", count = tasks.len()).entered();
-            self.backend
-                .invalidate_tasks(tasks.iter().copied().collect(), self);
+            self.backend.invalidate_tasks_set(tasks, self);
```
sokra, line 1114: avoids cloning

jridgewell (approved):

PR #5822, opened 2023-08-28
continue on failing test steps

sokra:
### Description

results should still be uploaded when a part of the tests fail or timeout


.github/workflows/daily-nextjs-integration-test.yml
```diff
@@ -34,7 +34,7 @@ jobs:
   upload_test_results:
     name: Upload test results
     needs: [next_js_integration]
-    if: ${{ github.event_name == 'schedule' }}
+    if: ${{ github.event_name == 'schedule' }} && always()
```
sokra, line 37: yes, this has an effect

ForsakenHarmony (approved):

PR #5824, opened 2023-08-28
feature(turborepo): Rust capnproto hashing

NicholasLYang:
### Description

Switches the Rust side to capnproto. We're only merging the Rust code so we can avoid a backlog of stacked PRs while also avoiding the global hash bump on the Go side.

### Testing Instructions

Closes TURBO-1272


.github/actions/setup-capnproto/action.yml
```diff
@@ -0,0 +1,24 @@
+name: "Setup Capnproto"
+description: "Sets up capnproto for the current platform"
+inputs:
+  target:
+    description: "Compilation target"
+    required: true
```
chris-olszewski, line 6: Instead of passing in the compilation target can we instead use [`runner.os`](https://docs.github.com/en/actions/learn-github-actions/variables#detecting-the-operating-system)? Also, I think that this is the compilation host and not target.

crates/turborepo-lib/src/hash/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+        let mut builder = message.init_root();
+
+        {
+            let mut packages_builder = builder.reborrow().init_packages(packages.len() as u32);
+            for (i, turborepo_lockfiles::Package { key, version }) in packages.iter().enumerate() {
+                let mut package = packages_builder.reborrow().get(i as u32);
+                package.set_key(key);
+                package.set_version(version);
+                // we don't track this in rust, set it to false
+                package.set_found(false);
```
chris-olszewski, line 129: I'm assuming we set this to false on the Go side as well before hashing, but when we pass the lockfile transitive closure from Rust to Go we [set found to true for everything](https://github.com/vercel/turbo/blob/main/crates/turborepo-ffi/src/lockfile.rs#L19)

chris-olszewski (approved):
Basically all of the comments are just questions I have to try to understand the `capnproto` stuff better.

crates/turborepo-lib/src/hash/mod.rs
NicholasLYang: Yeah in the [global hash PR](https://github.com/vercel/turbo/pull/5568/files#diff-0d8990c569c001c7e8cf9439823abd0e0893cd383612e213c4f910d39afa1566) we set it to true

crates/turborepo-lib/src/hash/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+            }
+        }
+
+        // We're okay to unwrap here because we haven't hit the nesting
+        // limit and the message will not have cycles.
+        let size = builder
+            .total_size()
+            .expect("unable to calculate total size")
+            .word_count
+            + 1;
```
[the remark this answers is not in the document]
NicholasLYang: Yeah...I don't remember either. I feel like this was a situation where I got an out of bounds error and fixed it with a + 1. I'll try to figure out why

crates/turborepo-lib/src/hash/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+}
+
+struct TaskOutputs {
+    inclusions: Vec<String>,
+    exclusions: Vec<String>,
+}
+
+struct LockFilePackages(pub Vec<turborepo_lockfiles::Package>);
+
+struct FileHashes(pub HashMap<turbopath::RelativeUnixPathBuf, String>);
```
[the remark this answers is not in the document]
NicholasLYang: Yeah huh that probably should be a clippy lint

PR #5825, opened 2023-08-28
feat: port task execution summary

chris-olszewski:
### Description

This is a port of the `execution.go` module in `runsummary` that is used by `execContext` via the `tracerFn` closure that gets passed back. In the Rust version we instead communicate to the shared state via a mpsc channel. This PR does not include the chrometracing functionality, but that can easily be added as a responsibility of the state management task.

Differences from Go:

I simplified the states that a task can go through compared to Go:
 - `TargetExecuted`: This was used to mark when task execution finished and was used for the time saved field in the cache. We will no longer report the time that it takes to save outputs as part of the task execution time in the run summary, but this can be added as an additional state if desired later.
 - `TargetInitialized`/`TargetBuilding`: initialized was only used for the initial insertion into the map and building was the state that was used once we confirmed that we had an actual task to run i.e. there was a command attached to the task
 - `TargetBuildStopped`: This state was never constructed in Go

In Rust there is now just:
 - create a tracker
 - "start" the tracker, this starts the timer
 - move the started tracker to one of the finished states:
   - canceled
   - build finished
   - build failed
   - cache hit

I also completely removed the internal tracking of the various task execution states, the tracker instead returns the state upon completion. In Go, we never actually used the shared state except for when trying to [find the partially constructed summary](https://github.com/vercel/turbo/blob/main/cli/internal/runsummary/execution_summary.go#L228) to add events to. We can completely avoid this since the tracker itself is now the partially constructed summary. 

### Testing Instructions

Unit tests

Closes TURBO-1273


crates/turborepo-lib/src/lib.rs
```diff
@@ -26,6 +26,7 @@ mod package_json;
 mod package_manager;
 mod rewrite_json;
 mod run;
+mod runsummary;
```
NicholasLYang, line 29: nit: Could we put this under the `run` module?

crates/turborepo-lib/src/runsummary/execution.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+    pub async fn finish(self) -> Result<SummaryState, tokio::task::JoinError> {
+        let Self {
+            state_thread,
+            sender,
+        } = self;
+        // We drop the sender so the channel closes once all trackers have finished
+        // we don't explicitly close as that would
```
NicholasLYang, line 122: that would?

NicholasLYang (approved):
Looks good to me

crates/turborepo-lib/src/lib.rs
NicholasLYang: i.e. `run/summary`, mostly cause that's what my PR is using

crates/turborepo-lib/src/runsummary/execution.rs
chris-olszewski: I like my comments to have a mystery about them

PR #5827, opened 2023-08-28
fix(lint): @turbo/test-utils

tknickman:
### Description

Lint for test utils package


.github/workflows/test.yml
```diff
@@ -1268,7 +1268,13 @@ jobs:
 
         # workspaces are added here as their lint issues are fixed
       - name: Lint check
-        run: turbo run lint --filter=create-turbo --filter=turbo-ignore --filter=@turbo/types --filter=eslint-config-turbo --filter=eslint-plugin-turbo
+        run: turbo run lint \
+          --filter=create-turbo \
+          --filter=eslint-config-turbo \
+          --filter=eslint-plugin-turbo \
+          --filter=turbo-ignore \
+          --filter=@turbo-test-utils
```
chris-olszewski, line 1276: Double checked and we do use `/` in the `package.json` name.
```suggestion
          --filter=@turbo/test-utils
```

chris-olszewski (approved):
I'm assuming test/formatting will pass with the updated name.

tknickman:
> I'm assuming test/formatting will pass with the updated name.

🤦🏼 good catch, thanks!

packages/turbo-test-utils/src/useFixtures.ts
```diff
@@ -12,37 +19,45 @@ interface SetupTextFixtures {
...
 
-export default function setupTestFixtures({
+export function setupTestFixtures({
   directory,
   test = "",
   options = {},
 }: SetupTextFixtures) {
   const fixtures: Array<string> = [];
-  const parentDirectory = path.join(directory, test ? test : "test-runs");
+  const parentDirectory = path.join(directory, test ? test : uuidv4());
```
tknickman, line 28: Fixes a bug that was causing multiple tests to sometimes collide when running in parallel. 

Multiple fixtures would be created under the same parent, and when one finished the entire directory would be removed. This fixes that.

packages/turbo-test-utils/src/useFixtures.ts
```diff
@@ -12,37 +19,45 @@ interface SetupTextFixtures {
...
 
-  afterEach(() => {
-    fixtures.forEach((fixture) => {
-      fs.rmSync(fixture, { recursive: true, force: true });
-    });
+  afterEach(async () => {
+    await Promise.all(
+      fixtures.map((fixture) =>
+        rimraf(fixture, {
```
tknickman, line 33: When running a lot of tests (workspaces runs like 500 in parallel) this can sometimes hit a busy signal. This makes that a bit more resilient.

PR #5839, opened 2023-08-30
fix(lint): @turbo/gen

tknickman:
### Description

Fix lint for `@turbo/gen`


packages/turbo-gen/src/commands/raw/index.ts
```diff
@@ -1,59 +1,73 @@
+import { convertCase, logger } from "@turbo/utils";
```
tknickman, line 1: This is the file responsible for the command that `turbo` core uses to pass arguments to this package.

This entire file was not type-safe at all. It's a tricky one since we could get total garbage into the JSON argument here, but this re-works things so that it's a bit more defensive and happy. 

Good news is we had great test coverage for this so it's easy to validate.

packages/turbo-gen/src/utils/plop.ts
```diff
@@ -139,29 +140,26 @@ export function getCustomGenerators({
...
 
   // add in separators to group by workspace
-  const gensWithSeparators: Array<Generator | inquirer.Separator> = [];
-  const lastGroup = undefined;
+  const gensWithSeparators: Array<Generator | Separator> = [];
   Object.keys(gensByWorkspace).forEach((group) => {
-    if (group !== lastGroup) {
-      gensWithSeparators.push(new inquirer.Separator(group));
-    }
+    gensWithSeparators.push(new Separator(group));
```
chris-olszewski, line 162: Just confirming: the removed conditional was just always true, correct?

chris-olszewski (approved):

packages/turbo-gen/src/utils/plop.ts
tknickman: That's correct, this was a bad if that ts linting pointed out

PR #5849, opened 2023-08-31
Child process manager

arlyon:
### Description

Adds a library for managing child processes. Spawning a process on the manager gives you a handle to the task, through which you can listen for its completion. You can also take down all tasks on the manager at once, and anyone listening to child tasks will be notified of exit codes etc.

### Testing Instructions

Go unit tests are ported (where it makes sense–due to some type-improvement-redesign a number of tests were unneeded). 

Closes TURBO-1287


chris-olszewski (changes requested):
In Go the process manager is shared across many goroutines via the [`execContext`](https://github.com/vercel/turbo/blob/main/cli/internal/run/real_run.go#L356) which is captured by the [`execFn` closure](https://github.com/vercel/turbo/blob/main/cli/internal/run/real_run.go#L237) that is used in the task graph traversal.

In Rust we'll also be taking this approach since the [visitor visits each task graph node in it's own task](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/task_graph/visitor.rs#L85).

I don't believe the current setup of having the process manager state encoded in the type will be feasible. We wouldn't be able to share the manager via `Arc<Mutex<ProcessManager<Open>>>` since closing it would require changing of the type. There might be some way to manage this state in it's own task and communicate to it via an `mpsc`, but that would be an additional layer of abstraction that I don't think is necessary.

While I'm a huge fan of encoding states via the type system, but I don't think in this case we can determine state at compile time.

crates/turborepo-lib/src/process/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+        let code = child.wait().await;
+        assert_eq!(code, Some(0));
+
+        manager.stop().await;
+
+        // this is idempotent, so calling it after the manager is stopped is ok
+        child.kill().await;
+
+        let code = child.wait().await;
+        assert_eq!(code, None);
```
chris-olszewski, line 175: I think would be useful to keep the behavior we had in Go and have a sentinel error for when the child was killed by us vs. killed by someone else (or at least we had the intent on killing it)

crates/turborepo-lib/src/process/shared_adaptor.rs
```diff
@@ -0,0 +1,116 @@
...
+        let mut command = super::super::child::Command::new("sleep");
+        command.arg("1");
+
+        manager.start();
+        let child = manager.spawn(command, Duration::from_secs(1));
+        assert!(child.is_some());
+        manager.wait().await;
+        assert!(!manager.is_started());
+        manager.stop().await;
+        assert!(!manager.is_started());
```
chris-olszewski, line 114: Can we do a more direct port of the [Go test](https://github.com/vercel/turbo/blob/main/cli/internal/process/manager_test.go#L36)? Starting multiple sleep children from different tasks before stopping them and then making sure that all of them didn't finish running.

crates/turborepo-lib/src/process/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub async fn stop(self) -> ProcessManager<Closed> {
+        let mut set = JoinSet::new();
+
+        for mut child in self.state.0.into_iter() {
+            set.spawn(async move { child.stop().await });
+        }
+
+        debug!("waiting for {} processes to exit", set.len());
+
+        while let Some(out) = set.join_next().await {
```
gsoltis, line 71: We should probably have a timeout on this, the Go code has a `killTimeout`

crates/turborepo-lib/src/process/child.rs
```diff
@@ -0,0 +1,473 @@
...
+        let gid = group.id();
+
+        let mut child = group.into_inner();
+
+        let _task = tokio::spawn(async move {
+            info!("waiting for task");
+            tokio::select! {
+                command = command_rx.recv() => {
+                    let state = match command {
+                        Some(ChildCommand::Stop) | None => {
```
gsoltis, line 202: What does `None` signify here?

crates/turborepo-lib/src/process/child.rs
```diff
@@ -0,0 +1,473 @@
...
+pub enum ChildCommand {
+    Stop,
+    Kill,
+}
+
+impl Child {
+    /// Start a child process, returning a handle that can be used to interact
+    /// with it. The command will be started immediately.
+    pub fn spawn(mut command: Command, shutdown_style: ShutdownStyle) -> Self {
+        let mut group = command.group().spawn().expect("failed to start child");
```
gsoltis, line 181: Am I correct in understanding that `group()` sets up a separate process group, like `Setpgid` from the Go side?

crates/turborepo-lib/src/process/child.rs
```diff
@@ -0,0 +1,473 @@
...
+
+                    {
+                        let mut task_state = task_state.write().await;
+                        *task_state = state;
+                    }
+                }
+                status = child.wait() => {
+                    // the child process exited
+                    let exit_code = status.ok().and_then(|s| s.code());
+                        {
```
gsoltis, line 221: Nit: I think the indentation here is weird? 

Also, in what scenarios can `child.wait()` fail? It looks like we're throwing away the error and just sending `None` on the channel...

crates/turborepo-lib/src/process/child.rs
```diff
@@ -0,0 +1,473 @@
...
+
+        let (command_tx, mut command_rx) = ChildCommandChannel::new();
+        let (exit_tx, exit_rx) = watch::channel(None);
+
+        let state = Arc::new(RwLock::new(ChildState::Running(command_tx)));
+        let task_state = state.clone();
+        let pid = group.inner().id();
+        let gid = group.id();
+
+        let mut child = group.into_inner();
```
gsoltis, line 195: Nit: we call `group.inner()` a bunch. Maybe we should grab the `gid`, then `into_inner()` and get everything else from the child?

crates/turborepo-lib/src/process/child.rs
```diff
@@ -0,0 +1,473 @@
...
+    pub fn stdout(&mut self) -> Option<tokio::process::ChildStdout> {
+        self.stdout.lock().unwrap().take()
+    }
+
+    pub fn stderr(&mut self) -> Option<tokio::process::ChildStderr> {
+        self.stderr.lock().unwrap().take()
+    }
+
+    /// Get a channel for interacting with the child process.
+    fn child_channel(state: &ChildState) -> Option<&ChildCommandChannel> {
```
gsoltis, line 301: Should this be a method on `ChildState` instead of `Child`?

crates/turborepo-lib/src/process/child.rs
```diff
@@ -0,0 +1,473 @@
...
+        match state {
+            ChildState::Running(child) => Some(child),
+            _ => None,
```
gsoltis, line 304: Nit: probably better to enumerate all of the variants so that we get an error if we ever add one.

gsoltis (changes requested):

crates/turborepo-lib/src/process/shared_adaptor.rs
```diff
@@ -0,0 +1,116 @@
...
+use super::{Closed, Open, ProcessManager};
+
+/// A simple adaptor over `ProcessManager` that allows it to be shared between
+/// threads.
+#[derive(Debug)]
+pub struct SharedProcessManager(Arc<Mutex<Either<ProcessManager<Open>, ProcessManager<Closed>>>>);
+
+impl SharedProcessManager {
+    pub fn new() -> Self {
+        Self(Arc::new(Mutex::new(Either::Right(
```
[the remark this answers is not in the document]
arlyon: It may be handy to close and reuse one so I have left start in, but defaulting to open makes sense.

crates/turborepo-lib/src/process/shared_adaptor.rs
```diff
@@ -0,0 +1,116 @@
...
+        let lock = self.0.lock().unwrap();
+        lock.is_left()
+    }
+
+    pub async fn wait(&self) {
+        let mut lock = self.0.lock().unwrap();
+        // we need ownership of the data in the lock, so replace it with a default value
+        let old = std::mem::replace(&mut *lock, Either::Right(ProcessManager::<Closed>::new()));
+        *lock = match old {
+            Either::Left(manager) => Either::Right(manager.wait().await),
```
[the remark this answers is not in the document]
arlyon: Yep, lock is held. If two SharedProccessManagers call wait or stop at the same time, the second one would essentially await until the first has completed their action, then fall through to the 'closed' state (a no-op)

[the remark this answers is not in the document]
arlyon: That means that as soon as the first item decides to wait, no other shared manager can call stop underneath or vice versa

crates/turborepo-lib/src/process/shared_adaptor.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    pub fn spawn(
+        &self,
+        command: super::child::Command,
+        timeout: std::time::Duration,
+    ) -> Option<super::child::Child> {
+        let mut lock = self.0.lock().unwrap();
+        lock.as_mut().map_left(|l| l.spawn(command, timeout)).left()
```
chris-olszewski, line 57: In the case that the process manager is in the process of stopping, then this function will block on all existing processes exiting before it can return `None`. If we are going to wait until closing finishes, I think we need to at least make this await to avoid blocking a thread just waiting for a lock to become free.

Ideally, we would have the same behavior as Go where we recognize that we're shutting down and immediately return `None`.

crates/turborepo-lib/src/process/child.rs
```diff
@@ -0,0 +1,473 @@
...
+// }
+
+// impl CommandBuilder {
+//     pub fn new(program: impl Into<CString>) -> Self {
+//         // let c = tokio::process::Command::new(program);
+//         // c.args(args)
+
+//         *CommandBuilder::default().program(program.into())
+//     }
+// }
```
chris-olszewski, line 46: I think this commented out code can get removed?

crates/turborepo-lib/src/process/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+        }
+        Some(child)
+    }
+
+    /// Stop the process manager, closing all child processes. On posix
+    /// systems this will send a SIGINT, and on windows it will just kill
+    /// the process immediately.
+    pub async fn stop(&self) {
+        self.is_closing
+            .store(true, std::sync::atomic::Ordering::Relaxed);
```
chris-olszewski, line 66: Do we need to be concerned with the behavior of multiple `stop` calls?

crates/turborepo-lib/src/process/child.rs
arlyon: Added comment

crates/turborepo-lib/src/process/child.rs
arlyon: Correct. This will be more useful when we set up windows signal handling

crates/turborepo-lib/src/process/child.rs
arlyon: I have folded that error into the return type

crates/turborepo-lib/src/process/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+        }
+    }
+}
+
+impl ProcessManager<Open> {
+    /// Spawn a new child process to run the given command.
+    ///
+    /// The handle of the child can be either waited or stopped by the caller,
+    /// as well as the entire process manager.
+    pub fn spawn(&mut self, command: child::Command, stop_timeout: Duration) -> child::Child {
```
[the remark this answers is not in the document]
arlyon: Ah, this param is not a timeout for the task, but for the graceful shutdown. Have renamed to reflect that.

crates/turborepo-lib/src/process/mod.rs
arlyon: Added

crates/turborepo-lib/src/process/mod.rs
arlyon: There is a timeout per-child which should provide an upper bound, although there is perhaps a situation where kill hangs _also_. I may set the timeout here to be whatever the kill timeout is plus one second.

crates/turborepo-lib/src/process/child.rs
```diff
@@ -0,0 +1,473 @@
...
+        let mut child = group.into_inner();
+
+        let _task = tokio::spawn(async move {
+            info!("waiting for task");
+            tokio::select! {
+                command = command_rx.recv() => {
+                    let state = match command {
+                        Some(ChildCommand::Stop) | None => {
+                            // we received a command to stop the child process
+                            shutdown_style.process(&mut child, Some(exit_tx)).await.unwrap()
```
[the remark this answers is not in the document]
arlyon: Removed the unwrap altogether with a refactor :)

crates/turborepo-lib/src/process/mod.rs
arlyon: No, this is intentional to support 'upgrading' the urgency in the event we started a wait but then decided to stop.

crates/turborepo-lib/src/process/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// Spawn a new child process to run the given command.
+    ///
+    /// The handle of the child can be either waited or stopped by the caller,
+    /// as well as the entire process manager.
+    ///
+    /// If spawn returns None,
+    pub fn spawn(&self, command: child::Command, stop_timeout: Duration) -> Option<child::Child> {
+        if self.is_closing.load(std::sync::atomic::Ordering::Relaxed) {
+            return None;
+        }
```
chris-olszewski, line 55: I'm think we might have a TOCTOU race condition here that could result in unmanaged children:
 - A call to `spawn` and `close` happen at the same time
 - The `spawn` call checks `is_closing`, but the setting of `true` hasn't happened from `close`
 - The `close` call then sets the `is_closing`, it then proceeds to lock children and send closing signals to all children. It then releases the mutex
 - The `spawn` call, waiting on the lock held by `close` can now proceed to add the newly spawned `child` into the children vector
 - After all of the children close `close` proceeds and clears the vector of children since it believes it has closed them all. This leaves us with no handle for the `Child` spawned by the `spawn` call.

crates/turborepo-lib/src/process/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        {
+            // clean up the vec and re-open
+            let mut lock = self.children.lock().expect("not poisoned");
+            // just allocate a new vec rather than clearing the old one
+            *lock = vec![];
+            self.is_closing
+                .store(false, std::sync::atomic::Ordering::Relaxed);
+        }
```
chris-olszewski, line 116: Just flagging this because I always have a bad time reasoning about atomic orderings. Do we want this to be relaxed?

> Relaxed accesses are the absolute weakest. They can be freely re-ordered and provide no happens-before relationship.

Does this mean compiler/hardware is free to re-order the "set true" and "set false" operations? If so, that's not what we want.

Reading through the [cpp spec 🤢 ](https://en.cppreference.com/w/cpp/atomic/memory_order#Relaxed_ordering) makes it seem that even on the same thread these can get reordered:

> In particular, this may occur if D is completed before C in thread 2, either due to compiler reordering or at runtime

crates/turborepo-lib/src/process/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+use self::child::{Child, ChildExit};
+
+/// A process manager that is responsible for spawning and managing child
+/// processes. When the manager is Open, new child processes can be spawned
+/// using `spawn`. When the manager is Closed, all currently-running children
+/// will be closed, and no new children can be spawned.
+#[derive(Debug, Clone)]
+pub struct ProcessManager {
+    is_closing: Arc<AtomicBool>,
+    children: Arc<Mutex<Vec<child::Child>>>,
```
chris-olszewski, line 33: I think to avoid to sidestep the complexity of splitting these two pieces of state into their own individual sync mechanisms, we should just bundle them together behind a shared lock. This is what we do in Go and it protects us from using one of these pieces of information without being synced with the other.
```suggestion
    inner: Arc<Mutex<ManagerInner>>,
```

```
struct ManagerInner {
    is_closing: bool,
    children: Vec<child::Child>,
}
```

chris-olszewski (changes requested):
This looks really close. I think the tests are all there, but we keep all of the state behind one lock to avoid the complexity of having them separate. This is what we do in Go and if the lock ends up having too much contention and cause perf issues we can always come back and fix it down the line.

crates/turborepo-lib/src/process/mod.rs
arlyon: Nice, I think I fixed it by requiring a lock of the child vec before checking if we are closing. I really want to keep this to support 'kill promotion' allowing us to stop / kill things at the process manager level. If we lock the whole thing, we can't send new commands while waiting for another to complete.

chris-olszewski (approved):

gsoltis:
Pushed a quick fix to the windows build

gsoltis (approved):

PR #5851, opened 2023-08-31
feature(turborepo): Go Hashing

NicholasLYang:
### Description

The code for implementing capnproto hashing in Go. Does a comparison between Go and Rust for task hashes and global hash.

### Testing Instructions

Closes TURBO-1288


.github/actions/setup-go/action.yml
```diff
@@ -11,7 +11,7 @@ runs:
     - name: Set up Go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18
+        go-version: 1.19
```
gsoltis, line 14: Why are we doing this?

cli/internal/fs/hash.go
```diff
@@ -20,56 +20,28 @@ type LockfilePackages []lockfile.Package
...
-	return hashObject(hashes)
+	return hash.HashFileHashes(hashes)
 }
 
 // HashTask produces the hash for a particular task
-func HashTask(task *TaskHashable) (string, error) {
-	return hashObject(task)
+func HashTask(task *hash.TaskHashable) (string, error) {
+	// s, _ := json.MarshalIndent(task, "", "  ")
+	// fmt.Printf("HashTask: %s\n", s)
```
gsoltis, line 36: Drop commented-out code

cli/internal/fs/hash.go
```diff
@@ -20,56 +20,28 @@ type LockfilePackages []lockfile.Package
...
+	return hash.HashTaskHashable(task)
 }
 
 // HashGlobal produces the global hash value to be incorporated in every task hash
-func HashGlobal(global GlobalHashable) (string, error) {
-	return hashObject(global)
+func HashGlobal(global hash.GlobalHashable) (string, error) {
+	// s, _ := json.MarshalIndent(global, "", "  ")
+	// fmt.Printf("HashGlobal: %s\n", s)
```
gsoltis, line 43: Drop commented-out code

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -271,4 +271,137 @@ impl<'a> Run<'a> {
 
         Ok(())
     }
+
+    #[tokio::main]
```
gsoltis, line 275: Why do we need this annotation? Is this a separate entrypoint?

go.work
```diff
@@ -0,0 +1,3 @@
+go 1.20
+
+use ./cli
```
gsoltis, line 3: We should not have this file.

gsoltis (changes requested):

.github/actions/setup-go/action.yml
NicholasLYang: Had some errors with capnproto that required a Go version bump

crates/turborepo-lib/src/run/mod.rs
NicholasLYang: yeah because we're not running the `run` function.

gsoltis (approved):

turborepo-tests/integration/tests/inference/no_workspaces.t.rej.orig
```diff
@@ -0,0 +1,31 @@
+@@ -53,18 +47,15 @@
```
chris-olszewski, line 1: Is this file meant to be committed?

crates/turborepo-lib/src/engine/builder.rs
```diff
@@ -80,7 +80,7 @@ impl<'a> EngineBuilder<'a> {
     pub fn with_root_tasks<I: IntoIterator<Item = TaskName<'static>>>(mut self, tasks: I) -> Self {
         self.root_enabled_tasks = tasks
             .into_iter()
-            .filter(|name| name.package() == Some(ROOT_PKG_NAME))
+            .filter(|name| name.package() == Some(ROOT_PKG_NAME) || name.package().is_none())
```
chris-olszewski, line 83: I don't think this is right, we're now considering all tasks that aren't workspace tasks as root tasks:
e.g. a pipeline with entries for just `build`, `test` and `//#test` we're now marking both `build` and `test` as root tasks when they aren't. 

This is bad because we now are listing `//#build` as a task even if the user hasn't specified it as a task which goes against our docs: 
> For example, if your pipeline declares a "build" task, and you want to include the build script defined in the monorepo's root package.json file with turbo run build, you must opt the root into it by declaring "//#build": {...} in your configuration

[link](https://turbo.build/repo/docs/core-concepts/monorepos/running-tasks#running-tasks-from-the-root)

crates/turborepo-lib/src/run/task_id.rs
```diff
@@ -20,8 +20,8 @@ pub struct TaskId<'a> {
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, Hash)]
 #[serde(try_from = "String", into = "String")]
 pub struct TaskName<'a> {
-    package: Option<Cow<'a, str>>,
-    task: Cow<'a, str>,
+    pub(crate) package: Option<Cow<'a, str>>,
+    pub(crate) task: Cow<'a, str>,
```
chris-olszewski, line 24: As my comment above displays, I really think we should keep these private and force callers to use the parsing logic in `From<&str> for TaskName` 

```suggestion
    package: Option<Cow<'a, str>>,
    task: Cow<'a, str>,
```

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -364,8 +366,30 @@ impl<'a> Run<'a> {
...
+                &pkg_dep_graph,
+                &scm,
+            )?;
+
+            if filtered_pkgs.len() != pkg_dep_graph.len() {
+                for target in self.targets() {
+                    let task_name = TaskName {
+                        package: Some(ROOT_PKG_NAME.into()),
+                        task: target.into(),
+                    };
```
chris-olszewski, line 382: I don't think this is correct, if a user passes `//#build` as a target then the task name this constructs is `TaskName { package: Some("//"), task: "//#build" }`

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -364,8 +366,30 @@ impl<'a> Run<'a> {
...
+
+                    if root_turbo_json.pipeline.contains_key(&task_name) {
```
chris-olszewski, line 384: `TurboJson` has a private method `has_task` that I think does what we want, I think making it public and using it here might make sense instead of trying to construct the right `TaskName` for lookup

chris-olszewski (changes requested):
I think the changes to the filtering belong in their own PR as they're not obvious and they have some serious behavior change implications.

crates/turborepo-lib/src/run/mod.rs
NicholasLYang: Can you even do that? I don't think `//#build` is a valid target:

crates/turborepo-lib/src/engine/builder.rs
NicholasLYang: Yeah you're right. I was confused as to what having `package` be `None` meant

crates/turborepo-lib/src/run/mod.rs
NicholasLYang: I'm not sure that's what we want. Say we have a target `build`, `has_task` will be true if we have a task definition for `build` in the pipeline. What we really want is for it to be true only if there is a task definition for `//#build` in the pipeline, since that will override the default behavior and include the root workspace in the filtered pkgs.

crates/turborepo-lib/src/run/mod.rs
chris-olszewski: Workspace tasks, including root tasks, are only valid if they're defined in `turbo.json` so `//#build` is only valid if it appears in the pipeline e.g. we can run `turbo run //#build:ts` in our monorepo since we define ` //#build:ts` pipeline, but we can't run a `//#build` since it isn't present (or `docs#build` for that matter).

crates/turborepo-lib/src/execution_state.rs
```diff
@@ -39,8 +42,23 @@ pub struct SpacesAPIClientConfig<'a> {
 
 impl<'a> TryFrom<&'a CommandBase> for ExecutionState<'a> {
     type Error = anyhow::Error;
-
     fn try_from(base: &'a CommandBase) -> Result<Self, Self::Error> {
+        let run = Run::new(base);
+
+        let global_hash;
+        let task_hash_tracker;
+        #[cfg(debug_assertions)]
```
arlyon, line 50: maybe a job for this one day 

https://github.com/rust-lang/rust/pull/115416

crates/turborepo-lib/src/run/mod.rs
arlyon: do we want to be more explicit here rather than just wrapping it in tokio::main? I am not sure where the function is called from but unless it is the 'main' entrypoint into a tokio app I think we should make this fn sync and spawn our own runtime

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -314,4 +326,157 @@ impl<'a> Run<'a> {
...
+        let pkg_dep_graph = PackageGraph::builder(&self.base.repo_root, root_package_json.clone())
+            .with_single_package_mode(opts.run_opts.single_package)
+            .build()?;
+
+        let root_turbo_json =
+            TurboJson::load(&self.base.repo_root, &root_package_json, is_single_package)?;
+
+        let root_workspace = pkg_dep_graph
+            .workspace_info(&WorkspaceName::Root)
+            .expect("must have root workspace");
```
arlyon, line 351: Not really a comment for this PR but an infallible API for the root if this is in fact the case would be nice.

crates/turborepo/src/main.rs
```diff
@@ -68,7 +68,10 @@ fn run_go_binary(execution_state: ExecutionState) -> Result<i32> {
 fn main() -> Result<()> {
     let exit_code = match turborepo_lib::main() {
         Payload::Rust(res) => res.unwrap_or(1),
-        Payload::Go(base) => run_go_binary((&*base).try_into()?)?,
+        Payload::Go(base) => {
+            let execution_state = (&*base).try_into()?;
+            run_go_binary(execution_state)?
```
arlyon, line 73: What is the motivation for assigning here?

crates/turborepo-lib/src/run/task_id.rs
```diff
@@ -9,7 +9,7 @@ pub const ROOT_PKG_NAME: &str = "//";
 
 /// A task identifier as it will appear in the task graph
 #[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize)]
-#[serde(from = "String")]
+#[serde(from = "String", into = "String")]
```
arlyon, line 12: I have not seen this directive and it's un-googleable. No need for a comment in the code but what does this do?

arlyon (approved):
Some minor comments but otherwise v happy :)

crates/turborepo-lib/src/run/task_id.rs
chris-olszewski: When serializing this type, `serde` will first convert it into the type specified by `into`: https://serde.rs/container-attrs.html#into

PR #5856, opened 2023-09-01
feat(example): update tailwind starter template

tknickman:
### Description

Updated official starter. 

`npx create-turbo@latest -e https://github.com/vercel/turbo/tree/update-starter/examples/with-tailwind`

Closes TURBO-1289


chris-olszewski (approved):
It looks so much better 😍

PR #5858, opened 2023-09-01
feat(Turborepo): add filesystem cookies on top of filewatching

gsoltis:
### Description

 - ports `cookie.go`
 - implements waiting for a roundtrip through the filesystem so filewatching queries can get consistent answers

### Testing Instructions

 - ports tests from `cookie_test.go`

Closes TURBO-1291

### DIFF
--- a/crates/turborepo-filewatch/src/cookie_jar.rs
+++ b/crates/turborepo-filewatch/src/cookie_jar.rs
@@ -1,0 +1,341 @@
+use std::{
+    collections::HashMap,
+    fs::OpenOptions,
+    path::PathBuf,
+    sync::{atomic::AtomicUsize, Arc},
+    time::Duration,
+};
+
+use notify::{Event, EventKind};
+use thiserror::Error;
+use tokio::{
+    sync::{broadcast, oneshot, Mutex},
+    time::error::Elapsed,
+};
+use tracing::debug;
+use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf, PathRelation};
+
+use crate::NotifyError;
+
+#[derive(Clone, Debug, Error)]
+pub enum WatchError {
+    #[error(transparent)]
+    RecvError(#[from] broadcast::error::RecvError),
+    #[error("filewatching encountered errors: {0}")]
+    NotifyError(#[from] NotifyError),
+    #[error("filewatching has closed, cannot watch cookies")]
+    Closed,
+}
+
+#[derive(Debug, Error)]
+pub enum CookieError {
+    #[error(transparent)]
+    Watch(#[from] WatchError),
+    #[error("cookie timeout expired")]
+    Timeout(#[from] Elapsed),
+    #[error("failed to receiver cookie notification: {0}")]
+    RecvError(#[from] oneshot::error::RecvError),
+    #[error("failed to write cookie file {0}")]
+    IO(#[from] std::io::Error),
+}
+
+type CookieResponse = Result<(), WatchError>;
+
+pub struct CookieJar {
+    root: AbsoluteSystemPathBuf,
+    serial: AtomicUsize,
+    timeout: Duration,
+    watches: Arc<Mutex<Watches>>,
+    // _exit_ch exists to trigger a close on the receiver when an instance
+    // of this struct is dropped. The task that is receiving events will exit,
+    // dropping the other sender for the broadcast channel, causing all receivers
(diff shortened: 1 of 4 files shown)


crates/turborepo-filewatch/src/cookie_jar.rs
```diff
@@ -0,0 +1,341 @@
...
+    fs::OpenOptions,
+    path::PathBuf,
+    sync::{atomic::AtomicUsize, Arc},
+    time::Duration,
+};
+
+use notify::{Event, EventKind};
+use thiserror::Error;
+use tokio::{
+    sync::{broadcast, oneshot, Mutex},
```
chris-olszewski, line 12: I think we should be using the standard mutex here.

[tokio mutex vs std mutex](https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex) and the [alternatives suggested](https://tokio.rs/tokio/tutorial/shared-state#holding-a-mutexguard-across-an-await)

TL;DR use std unless you need to hold the mutex across an await and even then, you should try to avoid doing that

crates/turborepo-filewatch/src/cookie_jar.rs
```diff
@@ -0,0 +1,341 @@
...
+
+// result flattening is an unstable feature, so add a manual helper to do so.
+// This version is for unwrapping events coming from filewatching
+fn unwrap_event(
+    event: Result<Result<Event, NotifyError>, broadcast::error::RecvError>,
+) -> Result<Event, WatchError> {
+    match event {
+        Ok(event) => event.map_err(|e| e.into()),
+        Err(e) => Err(e.into()),
+    }
```
chris-olszewski, line 179:
```suggestion
    Ok(event??)
```

crates/turborepo-filewatch/src/cookie_jar.rs
```diff
@@ -0,0 +1,341 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+// result flattening is an unstable feature, so add a manual helper to do so.
+// This version is for unwrapping events coming from filewatching
+fn unwrap_event(
```
chris-olszewski, line 173: Nit: I think `flatten_event` would be a lot better naming wise. `unwrap` make me think it'll panic if we encounter something that's not an event

crates/turborepo-filewatch/src/cookie_jar.rs
```diff
@@ -0,0 +1,341 @@
...
+                .insert(cookie_path.as_std_path().to_owned(), tx);
+        }
+        let mut opts = OpenOptions::new();
+        opts.truncate(true).create(true).write(true);
+        {
+            // dropping the resulting file closes the handle
+            _ = cookie_path.open_with_options(opts)?;
+        }
+        let resp = tokio::time::timeout(self.timeout, rx).await?;
+        unwrap_resp(resp)
```
chris-olszewski, line 107:
```suggestion
        Ok(resp??)
```

crates/turborepo-filewatch/src/cookie_jar.rs
```diff
@@ -0,0 +1,341 @@
...
+    match event {
+        Ok(event) => event.map_err(|e| e.into()),
+        Err(e) => Err(e.into()),
+    }
+}
+
+// result flattening is an unstable feature, so add a manual helper to do so.
+// This version is for flattening notifications coming back from the cookie
+// watching task
+fn unwrap_resp(
```
chris-olszewski, line 185: This can get dropped if we make use of `??`

chris-olszewski (commented):
First pass and it looks solid. I want to give the tests another look over with fresh eyes.

crates/turborepo-filewatch/src/cookie_jar.rs
gsoltis: Good reference. Updated to `std::sync::Mutex`

crates/turborepo-filewatch/src/cookie_jar.rs
gsoltis: `flatten_event` is definitely better, switched

crates/turborepo-filewatch/src/cookie_jar.rs
gsoltis: Dropped.

chris-olszewski (approved):

PR #5871, opened 2023-09-05
Add `Send` bound to `T` in `Vc<T>` to make type error less cryptic

kdy1:
### Description

Currently, the diagnostics for `T` in `Vc<T>` not implementing `Send` is too cryptic.

### Testing Instructions

Closes WEB-1498


wbinnssmith (approved):

sokra (approved):
If the rust compiler is happy, that's probably reviewed enough

PR #5882, opened 2023-09-05
fix: avoid panic during task cache construction

chris-olszewski:
### Description

When hooking up the task graph visitor to the task cache I hit a [failing debug assertion](https://github.com/vercel/turbo/blob/main/crates/turborepo-paths/src/absolute_system_path.rs#L179). This was triggered by the the [join we perform](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/run/cache.rs#L69) in the task graph construction.

This PR:
 - We join anchored workspace directory instead of the resolved absolute system path. [This now matches Go](https://github.com/vercel/turbo/blob/main/cli/internal/runcache/runcache.go#L313)
 - Moves this logic into it's own method so it can be tested without an entire `RunCache` instance
 - Removes an unnecessary cloning of the `TaskOutput` object

I'll note that this *probably* isn't what we want to do in the future as we're joining anchored system paths with globs that will [most definitely contain relative unix paths](https://github.com/vercel/turbo/blob/main/cli/internal/nodes/packagetask.go#L38). On windows this results in globs like `apps\web\.turbo/turbo-build.log`. I do want to fix this, but since that would be a behavior change in both Go&Rust, it belongs in its own PR.

### Testing Instructions

Moved code that panicked to it's own method and added a unit test that would exercise it.

### DIFF
--- a/crates/turborepo-lib/src/task_graph/mod.rs
+++ b/crates/turborepo-lib/src/task_graph/mod.rs
@@ -197,22 +197,25 @@ impl TaskDefinition {
         task_name: &TaskId,
         workspace_dir: &AnchoredSystemPath,
     ) -> TaskOutputs {
-        let mut repo_relative_globs = self.hashable_outputs(task_name);
-
-        let repo_relative_glob = |glob: &str| -> String {
+        let make_glob_repo_relative = |glob: &str| -> String {
             let mut repo_relative_glob = workspace_dir.to_string();
             repo_relative_glob.push(std::path::MAIN_SEPARATOR);
             repo_relative_glob.push_str(glob);
             repo_relative_glob
         };
 
+        // At this point repo_relative_globs are still workspace relative, but
+        // the processing in the rest of the function converts this to be repo
+        // relative.
+        let mut repo_relative_globs = self.hashable_outputs(task_name);
+
         for input in repo_relative_globs.inclusions.iter_mut() {
-            let relative_input = repo_relative_glob(input.as_str());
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/task_graph/mod.rs
```diff
@@ -192,6 +192,33 @@ impl TaskDefinition {
...
     }
 
+    pub fn repo_relative_hashable_outputs(
+        &self,
+        task_name: &TaskId,
+        workspace_dir: &AnchoredSystemPath,
+    ) -> TaskOutputs {
+        let mut repo_relative_globs = self.hashable_outputs(task_name);
+
+        let repo_relative_glob = |glob: &str| -> String {
```
NicholasLYang, line 202: nit: this could be a better name like `create_repo_relative_glob`

crates/turborepo-lib/src/task_graph/mod.rs
```diff
@@ -192,6 +192,33 @@ impl TaskDefinition {
         hashable
     }
 
+    pub fn repo_relative_hashable_outputs(
+        &self,
+        task_name: &TaskId,
+        workspace_dir: &AnchoredSystemPath,
+    ) -> TaskOutputs {
+        let mut repo_relative_globs = self.hashable_outputs(task_name);
```
NicholasLYang, line 200: Aren't these *not* repo relative globs? As in they're workspace relative globs

NicholasLYang (approved):
Just some naming nits

PR #5890, opened 2023-09-06
chore: disable clippy checks on fsevents

chris-olszewski:
### Description

Since we just vendor the `fsevent` module, I want to disable clippy from running on it. We could also fix up all of the lints, but that is slightly more invasive. 

I have my rust-analyzer check command set to `cargo clippy` so I get all clippy warnings and for whatever reason, one of the clippy warnings in `fsevent` prevented compiler errors from showing up in my editor.

### Testing Instructions

`cargo clippy` on `turborepo-lib` now doesn't hard fail.

Closes TURBO-1305


chris-olszewski:
Closing now that #5891

PR #5891, opened 2023-09-06
Fix clippy warnings / errors in filewatching code

gsoltis:
### Description

 - switch `if let Err(_) = ...` to `if ....is_err()`
 - drop unnecessary references
 - allow `bad_bit_mask` in `fsevents` because we want to be able to define a bitmask set to 0

### Testing Instructions

 - clippy

Closes TURBO-1306


chris-olszewski (approved):
I think we need to add a deny clippy warnings onto the crate itself [e.g.](https://github.com/vercel/turbo/pull/5890/files#diff-ad8a28a5b5f5d3b9cdd5860822364a25778bb3586b8281cc7fef10a2e8c4b5b5R1).

PR #5898, opened 2023-09-07
fix(release): correct ubuntu:xenial install

tknickman:
### Description

Fix capnproto action.

Validated in test action

Closes TURBO-1310

### DIFF
--- a/.github/actions/setup-capnproto/action.yml
+++ b/.github/actions/setup-capnproto/action.yml
@@ -7,7 +7,7 @@ runs:
     - name: "Setup capnproto for Linux"
       if: runner.os == 'Linux'
       shell: bash
-      run: sudo apt-get update && sudo apt-get install capnproto
+      run: sudo apt-get -y update && sudo apt-get install -y capnproto
 
     - name: "Setup capnproto for macos"
       if: runner.os == 'macOS'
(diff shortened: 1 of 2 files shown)


.github/workflows/turborepo-release.yml
```diff
@@ -217,7 +217,7 @@ jobs:
           - host: ubuntu-latest
             container: ubuntu:xenial
             container-options: "--platform=linux/amd64 --rm"
-            container-setup: "apt-get update && apt-get install -y curl musl-tools"
+            container-setup: "apt-get update && apt-get install -y curl musl-tools sudo"
```
tknickman, line 220: Required because the capnproto action needs `sudo`, which isn't installed in the `linux/amd64 xenial` container. We could remove the sudo from the capnproto action, but because it's used for all linux distros this was the easier option.

.github/actions/setup-capnproto/action.yml
```diff
@@ -7,7 +7,7 @@ runs:
     - name: "Setup capnproto for Linux"
       if: runner.os == 'Linux'
       shell: bash
-      run: sudo apt-get update && sudo apt-get install capnproto
+      run: sudo apt-get -y update && sudo apt-get install -y capnproto
```
tknickman, line 10: Auto accept install prompt.

arlyon (approved):
I don't really have a security concern with using sudo in CI, since it's not our infra

tknickman:
Nothing in CI is testing this, going to merge it to unblock release.

Issue #5868, opened 2023-09-05
[turborepo] Cannot create starter app with create-turbo-app and throw 503 error


amm834:
### What version of Turborepo are you using?

1.0.0

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Describe the Bug

```
➜  workspace pnpm dlx create-turbo@latest
../Library/pnpm/store/v3/tmp/dlx-4492    | +116 ++++++++++++
../Library/pnpm/store/v3/tmp/dlx-4492    | Progress: resolved 116, reused 116, downloaded 0, added 116, done

>>> TURBOREPO

>>> Welcome to Turborepo! Let's get you set up with a new codebase.

? Where would you like to create your turborepo? next-fdd
? Which package manager do you want to use? pnpm

Downloading files. This might take a moment.


>>> Unexpected error. Please report it as a bug:
av [Error]: Response code 503 (Service Unavailable)
    at Uo (/Users/admin/Library/pnpm/store/v3/tmp/dlx-4492/node_modules/.pnpm/create-turbo@1.10.13/node_modules/create-turbo/dist/cli.js:45:268)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Command.Nc (/Users/admin/Library/pnpm/store/v3/tmp/dlx-4492/node_modules/.pnpm/create-turbo@1.10.13/node_modules/create-turbo/dist/cli.js:218:615)
    at async Command.parseAsync (/Users/admin/Library/pnpm/store/v3/tmp/dlx-4492/node_modules/.pnpm/commander@10.0.1/node_modules/commander/lib/command.js:935:5)

```

### Expected Behavior

It should create a starter app.

### To Reproduce

```
pnpm dlx create-turbo@latest
```

### Reproduction Repo

Cannot provide cause it have not been created to use it.

bryanjhickey:
Some issue with me.

```bash
pnpm dlx create-turbo@latest


.../Library/pnpm/store/v3/tmp/dlx-90617  | +116 ++++++++++++
.../Library/pnpm/store/v3/tmp/dlx-90617  | Progress: resolved 116, reused 116, downloaded 0, added 116, done

>>> TURBOREPO

>>> Welcome to Turborepo! Let's get you set up with a new codebase.

? Where would you like to create your turborepo? ./my-turborepo
? Which package manager do you want to use? pnpm

Downloading files. This might take a moment.


>>> Unexpected error. Please report it as a bug:
[...]

PR #5900, opened 2023-09-07
feat(create-turbo): better error when GH is down

tknickman:
### Description

Show a better error message when Github is down

Fixes https://github.com/vercel/turbo/issues/5868

### DIFF
--- a/packages/create-turbo/src/commands/create/index.ts
+++ b/packages/create-turbo/src/commands/create/index.ts
@@ -10,6 +10,7 @@ import {
 import {
   getAvailablePackageManagers,
   createProject,
+  DownloadError,
   logger,
 } from "@turbo/utils";
 import { tryGitCommit, tryGitInit } from "../../utils/git";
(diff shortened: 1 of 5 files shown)


chris-olszewski (approved):

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -33,8 +34,15 @@ function handleErrors(err: unknown) {
   } else if (err instanceof ConvertError && err.type !== "unknown") {
     error(chalk.red(err.message));
     process.exit(1);
-    // handle unknown errors (no special handling, just re-throw to catch at root)
-  } else {
+    // handle download errors from @turbo/utils
+  } else if (err instanceof DownloadError) {
+    error(chalk.red("Unable to download template from Github"));
```
nathanhammond, line 39: GitHub

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -85,12 +93,20 @@ export async function create(
...
-  const { hasPackageJson, availableScripts, repoInfo } = await createProject({
-    appPath: root,
-    example: exampleName,
-    isDefaultExample: isDefaultExample(exampleName),
-    examplePath,
-  });
+
+  let projectData = {} as Awaited<ReturnType<typeof createProject>>;
+  try {
+    projectData = await createProject({
```
nathanhammond, line 99: This eventually hits `got` with no timeout: https://github.com/vercel/turbo/blob/main/packages/turbo-utils/src/examples.ts#L48C34-L50

The failure mode for the last outage didn't return a 500 unicorn, it just hung indefinitely.

This is definitely an improvement, but it doesn't address all failure modes.

nathanhammond (commented):
Sorry, this happened while I was asleep. ❤️

PR #5906, opened 2023-09-08
feat: port output control

chris-olszewski:
### Description

This is a port of what we do in Go for controlling writing to stdout/stderr. This abstraction will be used in conjunction with `PrefixedUI` and `LogWriter` to achieve the same behavior as Go.

In Go we achieve this behavior with a variety of wrapper structs around stdout and stderr that implement `io.Write`:
graph

In Rust we combine all the the nodes from "concurrent ui" on down and expose `OutputWriter` which implements `io::Write` which can then be wrapped in either of the other structs this crate provides. This roughly translates to [these lines in Go](https://github.com/vercel/turbo/blob/main/cli/internal/run/real_run.go#L212-L235)

Notes:
 - Currently this we wrap the underlying output writers in an `Arc<Mutex>` to allow them to be written to from multiple threads. This matches the behavior in Go, but in the future if we find we have too much contention with the mutex, we could try switching to a mpsc channel.
 - This PR also includes adding a separate writer for errors to `PrefixedUI` to match the Go implementation.

### Testing Instructions

Added unit tests for the new `output` module as well as an integration test that displays how `turbo` will use all of these modules together to achieve the same behavior as we have in Go. The integration test models how we use these components in Go: [prefix ui](https://github.com/vercel/turbo/blob/main/cli/internal/run/real_run.go#L412-L418) and [task writer](https://github.com/vercel/turbo/blob/main/cli/internal/run/real_run.go#L412-L418)

Closes TURBO-1312

### DIFF
--- a/crates/turborepo-ui/src/output.rs
+++ b/crates/turborepo-ui/src/output.rs
@@ -1,0 +1,324 @@
+use std::{
+    borrow::Cow,
+    cell::RefCell,
+    io::{self, Write},
+    sync::{Arc, Mutex},
+};
+
+/// OutputSink represent a sink for outputs that can be written to from multiple
+/// threads through the use of Loggers.
+pub struct OutputSink<W> {
+    writers: Arc<Mutex<SinkWriters<W>>>,
+}
+
+struct SinkWriters<W> {
+    out: W,
+    err: W,
+}
+
+/// OutputClient allows for multiple threads to write to the same OutputSink
+pub struct OutputClient<W> {
+    behavior: OutputClientBehavior,
+    buffer: Option<RefCell<Vec<SinkBytes<'static>>>>,
+    writers: Arc<Mutex<SinkWriters<W>>>,
+}
+
+pub struct OutputWriter<'a, W> {
+    logger: &'a OutputClient<W>,
+    destination: Destination,
+}
+
+/// Enum for controlling the behavior of the client
+#[derive(Debug, Clone, Copy)]
+pub enum OutputClientBehavior {
+    /// Every line sent to the client will get immediately sent to the sink
+    Passthrough,
+    /// Every line sent to the client will get immediately sent to the sink,
+    /// but a buffer will be built up as well and returned when finish is called
+    InMemoryBuffer,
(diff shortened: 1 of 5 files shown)


crates/turborepo-ui/src/output.rs
```diff
@@ -0,0 +1,324 @@
...
+        let writers = self.writers.clone();
+        OutputClient {
+            behavior,
+            buffer,
+            writers,
+        }
+    }
+}
+
+impl<W: Write> OutputClient<W> {
```
arlyon, line 85: Does it make sense for us to implement https://docs.rs/futures/latest/futures/sink/trait.Sink.html on this? That way we can fairly trivially take a stream (say stdin), map, filter, etc and route the stream directly to the Sink

arlyon (commented):
Looks good, one small API nicety :)

crates/turborepo-ui/src/output.rs
```diff
@@ -0,0 +1,324 @@
...
+            bytes
+        }))
+    }
+
+    fn handle_bytes(&self, bytes: SinkBytes) -> io::Result<usize> {
+        if matches!(
+            self.behavior,
+            OutputClientBehavior::InMemoryBuffer | OutputClientBehavior::Grouped
+        ) {
+            let bytes = SinkBytes {
```
NicholasLYang, line 150: I'm confused, what is this doing versus just passing `bytes` directly?

NicholasLYang: Is it to convert the buffer to owned?

NicholasLYang (approved):
Just one question

crates/turborepo-ui/src/output.rs
chris-olszewski: I think in the future if we start using the async IO trait variants (`AsyncWrite` etc) that will definitely be the route we want to go.

crates/turborepo-ui/src/output.rs
chris-olszewski: Yes, this is just to change `SinkBytes<'a>` to `SinkBytes<'static>` so we can store it in the in memory buffer.

PR #5915, opened 2023-09-08
chore: add error to prefixed UI

chris-olszewski:
### Description

In Go the prefix UI has output/warn/error methods that we use to write and prefix writes. This PR changes the prefixed UI in two external ways:
 - adds the error method
 - no longer require all prefixes to be provided, matching the behavior in Go

I'd suggest reviewing each commit on it's own as I tried to make sure each one was a contained change.

### Testing Instructions

Added new unit tests. The real testing will come when we use this in run and can check for Go parity.

Closes TURBO-1316


crates/turborepo-ui/src/prefixed.rs
```diff
@@ -14,58 +14,84 @@ use crate::UI;
...
         Self {
             ui,
-            output_prefix,
-            warn_prefix,
             out,
             err,
+            output_prefix: None,
+            warn_prefix: None,
+            error_prefix: None,
+            default_prefix: Style::new().apply_to(String::new()),
```
chris-olszewski, line 34: This is slightly undesired, but it's slightly cleaner than having a different codepath when a prefix isn't present. None of these calls perform an allocation so I think it's worth keeping this around.

crates/turborepo-ui/src/prefixed.rs
```diff
@@ -14,58 +14,84 @@ use crate::UI;
...
         }
     }
 
+    pub fn with_output_prefix(mut self, output_prefix: StyledObject<String>) -> Self {
+        self.output_prefix = Some(self.ui.apply(output_prefix));
+        self
+    }
```
chris-olszewski, line 41: These could easily be written using a macro, but for 3 copies of 4 lines, I'm not sure it's worth the effort. If we find we need more prefixes or this code becomes more convoluted, then we can consider doing all of this via a macro.

crates/turborepo-ui/src/prefixed.rs
```diff
@@ -14,58 +14,84 @@ use crate::UI;
...
             err,
+            output_prefix: None,
+            warn_prefix: None,
+            error_prefix: None,
+            default_prefix: Style::new().apply_to(String::new()),
         }
     }
 
+    pub fn with_output_prefix(mut self, output_prefix: StyledObject<String>) -> Self {
+        self.output_prefix = Some(self.ui.apply(output_prefix));
```
chris-olszewski, line 39: Note: we now eagerly apply the `ui` to the prefixes when they get passed in. Since we don't offer a way to change `ui` after construction we don't need to wait until the actual line writing to perform this operation.

chris-olszewski (commented):
Review comments

NicholasLYang (approved):

PR #5925, opened 2023-09-11
feat(examples): update gatsby

tknickman:
### Description

Gatsby updates and lint fixes


examples/with-gatsby/packages/eslint-config-custom/package.json
```diff
@@ -1,25 +1,10 @@
...
-    "eslint-config-prettier": "^8.3.0",
-    "eslint-config-turbo": "latest",
-    "eslint-plugin-gatsby": "^1.0.2",
-    "eslint-plugin-react": "7.31.8"
-  },
   "devDependencies": {
-    "@types/eslint": "^8.21.2",
-    "eslint-config-react-app": "^7.0.1",
-    "typescript": "^4.7.4"
+    "@vercel/style-guide": "^5.0.0",
```
mehulkar, line 7: is this supposed to be here? seems like a weird thing to ship to customers in an example

mehulkar (approved):

examples/with-gatsby/packages/eslint-config-custom/package.json
tknickman: @mehulkar it is. We're using this instead of writing all our own rules / including another preset (like airbnb etc)

PR #5930, opened 2023-09-11
feat: use prefixed ui in task graph visitor

chris-olszewski:
### Description

Various plumbing changes that allow us to construct a prefixed UI instance for each task. This instance is necessary for interaction with the task cache.

Reviewer Notes:
Each commit is a fairly simple addition/change to the visitor and can be reviewed on its own.

### Testing Instructions

Observe that colors are displayed when running tasks with the run stub still:
Screenshot 2023-09-11 at 2 25 08 PM

Closes TURBO-1322

### DIFF
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -71,7 +71,7 @@ pub struct RunOpts<'a> {
     pub(crate) no_daemon: bool,
     pub(crate) single_package: bool,
     pub log_prefix: LogPrefix,
-    pub log_order: LogOrder,
+    pub log_order: ResolvedLogOrder,
     summarize: Option<Option<bool>>,
     pub(crate) experimental_space_id: Option<String>,
     pub is_github_actions: bool,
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/opts.rs
```diff
@@ -70,8 +70,8 @@ pub struct RunOpts<'a> {
     pub graph: Option<GraphOpts<'a>>,
     pub(crate) no_daemon: bool,
     pub(crate) single_package: bool,
-    pub log_prefix: LogPrefix,
-    pub log_order: LogOrder,
+    pub log_prefix: ResolvedLogPrefix,
+    pub log_order: ResolvedLogOrder,
```
chris-olszewski, line 74: Post port we should change the CLI to take `Option<LogOrder>` and remove the `LogOrder::Auto` variant in favor of `None`. This removes the need for the duplicate `ResolvedLogOrder` definition.

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -172,6 +203,103 @@ impl<'a> Visitor<'a> {
 
         Ok(())
     }
+
+    fn sink(opts: &Opts) -> OutputSink<StdWriter> {
+        let err_writer = match matches!(
```
gsoltis, line 208: Nit: could this be an `if` statement? `if matches!(...) && opts.run_opts.is_github_actions { ... } else { ... }` ?

Also, future note: we should probably always force everything to `stdout` on GH actions, even if we aren't grouping task logs. It's the closest we'll get to an in-order representation of what happened.

gsoltis (approved):

PR #5931, opened 2023-09-11
fix: make `OutputWriter` references `Send`

chris-olszewski:
### Description

As I was hooking up the task cache I encountered the dreaded "future is not `Send` as this value is used across an await" due to a [`task_cache.restore_outputs(&mut prefixed_ui).await`](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/run/cache.rs#L149) call. The easy solution is to drop the `RefCell` and instead use a `RwLock`.

[Quick reading that explains *why* locals that are held across an await need to be `Send`](https://blog.rust-lang.org/inside-rust/2019/10/11/AsyncAwait-Not-Send-Error-Improvements.html#how-are-async-functions-implemented)

### Testing Instructions

Added assertion to make sure we'll be able to hold `&mut OutputWriter` across `.await` i.e. implements `Send`

Closes TURBO-1323


arlyon (approved):
LGTM

PR #5935, opened 2023-09-12
fix(examples-tests): correct gatsby prysk test

tknickman:
### Description

Fix gatsby examples test

Closes TURBO-1325


mehulkar (approved):
seems correct, but unclear if it's deterministic or not

PR #5939, opened 2023-09-13
Revert "Child process manager"

gsoltis:
Reverts vercel/turbo#5849

This was merged without running the Rust unit tests, which don't yet pass. Reverting until we can sort out the unit tests

Closes TURBO-1326


gsoltis:
Force-merging to revert failing unit tests

PR #5940, opened 2023-09-13
feat(turborepo): add basic child process manager package

gsoltis:
Previous PR #5849 was opened by @arlyon and it auto-merged without passing unit tests first. Re-opening from the same branch to sort out unit tests.

### Description

Adds a library for managing child processes. Spawning a process on the manager gives you a handle to the task, through which you can listen for its completion. You can also take down all tasks on the manager at once, and anyone listening to child tasks will be notified of exit codes etc.

### Testing Instructions

Go unit tests are ported (where it makes sense–due to some type-improvement-redesign a number of tests were unneeded). 

Closes TURBO-1287


arlyon:
Oh that is strange, I can see tests failing in the ci as well. It's a few windows things that I am going to work through this morning :)

nathanhammond (approved):
Stamping a pre-approved PR without direct review.

Issue #4762, opened 2023-05-01
Support bun (bun.sh)


patlux:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

If I set `"packageManager": "bun@0.5.9"` in my `package.json` and run turbo, I get:

```sh
$ turbo run lint
 ERROR  run failed: could not detect yarn version: exit status 1
Turbo error: could not detect yarn version: exit status 1
error: script "lint" exited with code 1 (SIGHUP)
```

### Describe the solution you'd like

It should support bun.

### Describe alternatives you've considered

None

mehulkar:
Hey @patlux thanks for the suggestion! I'll bring this up with the team. I imagine that (1) we're open to a contribution here or (2) we'll wait till after the Rust port to complete before attempting to add a new package manager.

mehulkar:
Hey @patlux we talked about this at our team meeting today. It looks like one of the major initial differences is that [bun generates a binary lockfile](https://github.com/oven-sh/bun/blob/c87d65856c06a03a2470fdc1bd48bc4042dadaec/docs/install/lockfile.md#L1). Other than just detecting that file to infer `bun` is the package manager, we also need to be able to parse that file to generate a workspace graph, which is a pretty significant effort. 

We are watching bun adoption and demand for this feature. If this issue gets a lot of interest we will consider adding support for it.

Thanks again for filing an issue!

patlux:
@mehulkar Thank you for sharing the info!

There is also the option to generate a `yarn.lock`-file when running `bun install`. So I think for the beginning it would be totally fine to use `bun install --yarn` and use the generated `yarn.lock`-file. What do you think?

paperdave:
this was mentioned today in buns discord
i want to note that if you run "bun ./bun.lockb", bun will print out a yarn yaml lockfile to stout, which can be parsed using the same code assuming yarn is supported.

RobSchilderr:
> this was mentioned today in buns discord
> i want to note that if you run "bun ./bun.lockb", bun will print out a yarn yaml lockfile to stout, which can be parsed using the same code assuming yarn is supported.

Was looking for this comment. Could there be more info on using turbo and bun?

izakfilmalter:
Any movement on this? Vercel now supports it for deploys: https://vercel.com/changelog/bun-install-is-now-supported-with-zero-configuration

nathanhammond:
[...]

PR #5943, opened 2023-09-13
feat(create-turbo): support bun (beta)

tknickman:
### Description

Support `bun` in `create-turbo` and `turbo/workspaces`. This will allow users to start from any example using bun as a package manager. 

Requires https://github.com/vercel/turbo/pull/5934

Tasks still TODO:
- [x] Bun specific tests
- [x] Validation when converting _to_ bun (bun only supports single level globs, so not all projects can be converted to bun workspaces, we need to check this)

Related to https://github.com/vercel/turbo/issues/4762


packages/create-turbo/src/commands/create/prompts.ts
```diff
@@ -52,10 +52,15 @@ export async function packageManager({
...
       !manager || !availablePackageManagers[manager as PackageManager],
-    choices: ["npm", "pnpm", "yarn"].map((p) => ({
-      name: p,
-      value: p,
-      disabled: availablePackageManagers[p as PackageManager]
+    choices: [
+      { pm: "npm", label: "npm workspaces" },
+      { pm: "pnpm", label: "pnpm workspaces" },
+      { pm: "yarn", label: "yarn workspaces" },
+      { pm: "bun", label: "bun workspaces (beta)" },
```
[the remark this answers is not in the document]
nathanhammond: I don't have strong opinions here. Seems fine to set expectations.

packages/turbo-workspaces/__tests__/managers.test.ts
```diff
@@ -272,6 +278,8 @@ describe("managers", () => {
           new RegExp(`^.*\/${directoryName}\/yarn.lock$`);
         } else if (fixtureManager === "npm") {
           new RegExp(`^.*\/${directoryName}\/package-lock.json$`);
+        } else if (fixtureManager === "bun") {
+          new RegExp(`^.*\/${directoryName}\/bun.lockb$`);
```
nathanhammond, line 282: I'm so happy that this is the actual behavior now.

packages/turbo-workspaces/src/commands/summary/index.ts
```diff
@@ -88,7 +88,7 @@ export async function summaryCommand(directory: SummaryCommandArgument) {
     Object.keys(workspacesByDirectory).forEach((dir, idx) => {
       renderDirectory({
         number: idx + 1,
-        workspaces: workspacesByDirectory[directory],
+        workspaces: workspacesByDirectory[dir],
```
nathanhammond, line 91: Unrelated discovery?

packages/turbo-workspaces/src/convert.ts
```diff
@@ -67,7 +67,7 @@ export async function convertProject({
 
   logger.mainStep("Installing dependencies");
   if (!options?.skipInstall) {
-    await MANAGERS[to.name].convertLock({ project, logger, options });
+    await MANAGERS[to.name].convertLock({ project, to, logger, options });
```
nathanhammond, line 70: Unrelated discovery?

packages/turbo-workspaces/src/install.ts
```diff
@@ -68,6 +68,18 @@ export const PACKAGE_MANAGERS: Record<
       semver: ">=2",
     },
   ],
+  bun: [
+    {
+      name: "bun1",
```
nathanhammond, line 73: I'm comfortable with this being just `bun` since it doesn't end up exposed to users or constrain us from renaming to `bun1` later.

packages/turbo-workspaces/src/install.ts
```diff
@@ -68,6 +68,18 @@ export const PACKAGE_MANAGERS: Record<
...
+      template: "bun",
+      command: "bun",
+      installArgs: ["install"],
+      version: "1.x",
```
nathanhammond, line 77: This should be `latest`, it's the registry tag.

```suggestion
      version: "latest",
```

packages/turbo-workspaces/src/install.ts
```diff
@@ -68,6 +68,18 @@ export const PACKAGE_MANAGERS: Record<
...
+      executable: "bunx",
+      semver: "<2",
```
nathanhammond, line 79: I think this is the syntax?
```suggestion
      semver: ">= 1.0.1 && <2",
```

We don't work less than 1.0.1.

packages/turbo-workspaces/src/updateDependencies.ts
```diff
@@ -37,6 +37,10 @@ function updateDependencyList({
   return { dependencyList, updated };
 }
 
+/**
+ * Convert workspace dependencies to the format that `to` requires. Only needed when pnpm is involved as
+ * it requires `workspace:*` and all the rest support `*`
```
nathanhammond, line 42: Did you characterize the edge cases for `bun`? Can you enumerate them?

packages/turbo-workspaces/src/managers/pnpm.ts
```diff
@@ -205,25 +205,33 @@ async function clean(args: CleanArgs): Promise<void> {
...
-    );
-    if (!options?.dry && existsSync(project.paths.lockfile)) {
-      try {
-        await execa("pnpm", ["import"], {
-          stdio: "ignore",
-          cwd: project.paths.root,
-        });
-      } finally {
-        rmSync(project.paths.lockfile, { force: true });
+    // pnpm does not support importing a bun lockfile
```
nathanhammond, line 211: Yes it does! By way of `yarn`! `bun bun.lockb` will output a `yarn.lock` file. You can execute that and then feed that into the transform.

packages/turbo-workspaces/src/managers/bun.ts
```diff
@@ -0,0 +1,227 @@
...
+  logger.rootHeader();
+
+  // package manager
+  logger.rootStep(
+    `adding "packageManager" field to ${path.relative(
+      project.paths.root,
+      project.paths.packageJson
+    )}`
+  );
+  // TODO: This technically isn't valid as part of the spec (yet)
```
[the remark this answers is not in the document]
nathanhammond: Riding on these rails will likely require people to disable `corepack`. That's not a great experience, in spite of us needing it specified. Not sure I have a good answer here.

packages/turbo-workspaces/src/utils.ts
```diff
@@ -205,6 +201,47 @@ function getMainStep({
...
 }
 
+/**
+ * At the time of writing, bun only support simple globs (can only end in /*) for workspaces. This means we can't convert all projects
+ * from other package manager workspaces to bun workspaces, we first have to validate that the globs are compatible.
+ *
+ * NOTE: It's possible a project could work with bun workspaces, but just not in the way it is currently defined. We will
+ * not change existing globs to make a project work with bun, we will only convert projects that are already compatible.
+ *
+ * This function matches the behavior of bun's glob validation: https://github.com/oven-sh/bun/blob/main/src/install/lockfile.zig#L2889
```
nathanhammond, line 211: Good idea to port this precisely. Use a GitHub permalink (`y` on your keyboard after going to that URL) for this. (This link is already broken.)

The other package managers arbitrarily switch between micromatch and minimatch, but fortunately that's basically compatible except for cases that are _far_ too complicated for workspace definition.

Personal opinion: these shouldn't have allowed globs to begin with.

packages/turbo-workspaces/src/utils.ts
```diff
@@ -205,6 +201,47 @@ function getMainStep({
...
+ */
+function isCompatibleWithBunWorkspace({
```
nathanhammond, line 213:
```suggestion
function isCompatibleWithBunWorkspaces({
```

packages/turbo-workspaces/src/utils.ts
```diff
@@ -205,6 +201,47 @@ function getMainStep({
...
+  const validator = (glob: string) => {
+    if (glob.includes("*")) {
+      // no multi level globs
+      if (glob.includes("**")) {
+        return false;
+      }
+
+      // no * in the middle of a path
+      const withoutLastPathSegment = glob
+        .split(path.sep)
```
nathanhammond, line 227: These are glob patterns, not paths. This should actually be `/`.

nathanhammond (changes requested):
Mostly small things!

packages/turbo-workspaces/src/install.ts
nathanhammond: `1.x` being a weird thing is because of (old) `corepack` mistakes.

packages/turbo-workspaces/src/managers/pnpm.ts
nathanhammond: This also means that we can do `bun` => `yarn`.

packages/turbo-workspaces/src/utils.ts
Jarred-Sumner: yes I suggest using a permalink. note that eventually we will implement a glob in zig and this limitation will go away. A bunch of companies are asking for it so we might use a [filesystem-only libc implementation](https://man7.org/linux/man-pages/man3/glob.3.html) in the worst case

packages/turbo-workspaces/src/commands/summary/index.ts
tknickman: Yea this was introduced (by me) as part of https://github.com/vercel/turbo/pull/5810.

I'm going to move this (and one other fix) out of this PR to another one so we can merge that ahead of this PR, incase we need a patch before bun support is fully ready.

packages/turbo-workspaces/src/utils.ts
tknickman: Good catch

packages/turbo-workspaces/src/commands/summary/index.ts
tknickman: Moved here: https://github.com/vercel/turbo/pull/5951

packages/turbo-workspaces/src/managers/pnpm.ts
tknickman: Done

packages/turbo-workspaces/src/updateDependencies.ts
tknickman: Most of them seem to resolve around not starting with a clean project (no node_modules anywhere)

mehulkar (commented):
Code looks good, but wasn't able to execute tested minimally with:

```
curl -fsSL https://bun.sh/install | bash
./cli.js
// answered path prompt with /Users/mehulkar/bun-test
// selected bun from prompt
cd ~/bun-prompt

$ ./node_modules/.bin/turbo run build
Error: We did not find a package manager specified in your root package.json. Please set the "packageManager" property in your root package.json (https://nodejs.org/api/packages.html#packagemanager) or run `npx @turbo/codemod add-package-manager` in the root of your monorepo.
$ ./node_modules/.bin/turbo --version
1.10.14-canary.1
```

Is @nathanhammond 's work required for this and it's not released yet or am I missing something else?

tknickman:
@mehulkar yep turbo core won't work with bun yet until @nathanhammond's PR lands (this PR does not include that one, but it is marked as required in the description)

nathanhammond (approved):

PR #5952, opened 2023-09-14
improve error handling of developer workflow tool

sokra:
### Description

improve error handling in the developer workflow tooling

Closes WEB-1563

### DIFF
--- a/packages/devlow-bench/src/describe.ts
+++ b/packages/devlow-bench/src/describe.ts
@@ -159,7 +159,7 @@ export async function reportMeasurement(
     if (!prev) {
       throw new Error(`No measurement named ${relativeTo} found`);
     }
-    if (prev.unit !== "ms") {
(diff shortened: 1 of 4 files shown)


packages/devlow-bench/src/describe.ts
```diff
@@ -131,14 +159,14 @@ export function reportMeasurement(
...
-    if (prev.unit !== "ms") {
+    if (prev.unit !== unit) {
       throw new Error(
         `Measurement ${relativeTo} is not a "${unit}" measurement`
       );
     }
     reportedValue -= prev.value;
   }
-  currentScenario.iface.measurement(
+  await currentScenario.iface.measurement(
```
sokra, line 169: `measurement` is actually async and to handle errors correctly we need to await it

jridgewell (approved):

PR #5953, opened 2023-09-14
Revert "Reduce the number of tasks for better performance"

sokra:
Reverts vercel/turbo#5950

Closes WEB-1569


jridgewell (approved):
Why revert?

sokra:
> Why revert?

It breaks the benchmark in CI and also next.js in https://github.com/vercel/next.js/pull/55404

PR #5958, opened 2023-09-15
fix: convert the lockfile path to a repo-root-relative path before use

arlyon:
### Description

In scope / filter I mistakenly used an absolute path instead of a repo-relative one in the matcher. This resolves that. It additionally makes more-clear a constraint that we have that lockfiles must be within a repo to support change detection, and adds error handling / tests for that.

It also removes the use of wax (globbing) from the initial impl, which was ported across verbatim from go, since I believe we are just comparing resolved paths at this point. (tagging @chris-olszewski since you worked in this area last)

https://github.com/vercel/turbo/blob/b916a874e8bd192145dbc75525750a0860b8070c/cli/internal/scope/scope.go#L258-L265

### Testing Instructions

3 new test cases

Closes TURBO-1330

### DIFF
--- a/crates/turborepo-lib/src/run/scope/change_detector.rs
+++ b/crates/turborepo-lib/src/run/scope/change_detector.rs
@@ -147,9 +147,8 @@ impl<'a> SCMChangeDetector<'a> {
             .all(|(a, b)| a == b)
     }
 
-    /// Get a list of changes from the lockfile.
-    ///
-    /// Returning Ok(None) here indicates
+    /// Get a list of package names that have changed due to changes in the
+    /// lockfile across the HEAD and the given ref.
     fn get_changes_from_lockfile(
         &self,
         changed_files: &HashSet<AnchoredSystemPathBuf>,
@@ -160,13 +159,8 @@ impl<'a> SCMChangeDetector<'a> {
             .package_manager()
             .lockfile_path(self.turbo_root);
 
-        let lockfile_path_relative = lockfile_path
-            .anchor(self.turbo_root)
-            .expect("was created above");
-
-        let matcher = wax::Glob::new(lockfile_path_relative.as_str())?;
-
-        if !changed_files.iter().any(|f| matcher.is_match(f.as_path())) {
+        // no changes to lockfile? then no changes because of lockfile
+        if !SCMChangeDetector::lockfile_changed(self.turbo_root, changed_files, &lockfile_path)? {
             return Ok(vec![]);
         }
 
@@ -181,6 +175,20 @@ impl<'a> SCMChangeDetector<'a> {
             .changed_packages(previous_lockfile.as_ref())?;
 
         Ok(additional_packages)
+    }
+
+    fn lockfile_changed(
+        turbo_root: &AbsoluteSystemPath,
+        changed_files: &HashSet<AnchoredSystemPathBuf>,
+        lockfile_path: &AbsoluteSystemPath,
+    ) -> Result<bool, ChangeDetectError> {
+        let lockfile_path_relative = turbo_root
+            .anchor(lockfile_path)
+            .expect("lockfile should be in repo");
+
+        let matcher = wax::Glob::new(lockfile_path_relative.as_str())?;
+
+        Ok(changed_files.iter().any(|f| matcher.is_match(f.as_path())))
     }
 }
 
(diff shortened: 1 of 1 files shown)


mehulkar (approved):

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -178,6 +176,18 @@ impl<'a> SCMChangeDetector<'a> {
...
+    fn lockfile_changed(
+        turbo_root: &AbsoluteSystemPath,
+        changed_files: &HashSet<AnchoredSystemPathBuf>,
+        lockfile_path: &AbsoluteSystemPath,
+    ) -> Result<bool, ChangeDetectError> {
+        let lockfile_path_relative = turbo_root
+            .anchor(lockfile_path)
+            .map_err(|_| ChangeDetectError::LockfileNotInRepo)?;
+
+        Ok(changed_files.iter().any(|f| f == &lockfile_path_relative))
```
chris-olszewski, line 189:
```suggestion
        Ok(changed_files.contains(&lockfile_path_relative))
```

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -202,3 +216,58 @@ impl From<ChangedPackagesError> for ChangeDetectError {
...
+        let changed_files = changed_files
+            .iter()
+            .map(|s| turbopath::AnchoredSystemPathBuf::from_raw(s).unwrap())
+            .collect();
+        let changes =
+            SCMChangeDetector::lockfile_changed(&turbo_root, &changed_files, &lockfile_path);
+
+        // we don't want to implement PartialEq on the error type,
+        // so simply compare the debug representations
+        assert_eq!(format!("{:?}", changes), format!("{:?}", expected));
```
chris-olszewski, line 247: I always feel like there should be a better way to handle comparing possible errors. This might be the cleanest way around it I've seen, but it still feels slightly off.

chris-olszewski (approved):

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -192,6 +202,10 @@ pub enum ChangeDetectError {
     NoLockfile,
     #[error("Lockfile error: {0}")]
     Lockfile(turborepo_lockfiles::Error),
+    /// This is a potential case that may arise in Bun. We may need to
```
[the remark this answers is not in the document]
chris-olszewski: Actually this might no longer be an issue: https://github.com/oven-sh/bun/pull/5335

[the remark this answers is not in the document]
nathanhammond: @chris-olszewski is correct! Bun had showstopping bugs with `turbo` in `bun@1.0.0`, and 5335 landed in `bun@1.0.1` along with the other fixes for `turbo`.

So: no version of `bun` that works with `turbo` supports non-local lockfiles! 🎉

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -192,6 +202,10 @@ pub enum ChangeDetectError {
...
+    /// relax / handle this case in the future
+    #[error("Lockfile not in repo")]
+    LockfileNotInRepo,
```
nathanhammond, line 208:
```suggestion
```

nathanhammond (approved):
We can remove the "outside of project" check!

crates/turborepo-lib/src/run/scope/change_detector.rs
arlyon: Yeah its a bit jank. I didn't want to just check that the enum variant was the same either...

crates/turborepo-lib/src/run/scope/change_detector.rs
arlyon: Sorted

crates/turborepo-lib/src/run/scope/change_detector.rs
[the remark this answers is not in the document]
arlyon: New iteration up now, TY

arlyon:
Enabling auto merge

PR #5959, opened 2023-09-15
reduce number of task in async_module

sokra:
### Testing Instructions

Closes WEB-1580

### DIFF
--- a/crates/turbopack-ecmascript/src/references/async_module.rs
+++ b/crates/turbopack-ecmascript/src/references/async_module.rs
@@ -6,7 +6,7 @@ use swc_core::{
     ecma::ast::{ArrayLit, ArrayPat, Expr, Ident, Pat, Program},
     quote,
 };
-use turbo_tasks::{primitives::Bools, trace::TraceRawVcs, TryFlatJoinIterExt, Value, Vc};
+use turbo_tasks::{trace::TraceRawVcs, TryFlatJoinIterExt, Value, Vc};
 use turbopack_core::chunk::availability_info::AvailabilityInfo;
 
 use super::esm::base::ReferencedAsset;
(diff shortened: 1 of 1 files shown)


ForsakenHarmony (approved):

crates/turbopack-ecmascript/src/references/async_module.rs
```diff
@@ -124,19 +113,23 @@ impl AsyncModuleScc {
...
         for scc in &*this.scope.get_scc_children(this.scc).await? {
             // Because we generated SCCs there can be no loops in the children, so calling
             // recursively is fine.
-            bools.push(AsyncModuleScc::new(*scc, this.scope).is_async());
+            if *AsyncModuleScc::new(*scc, this.scope).is_async().await? {
+                return Ok(Vc::cell(true));
+            }
         }
 
-        Ok(Vc::<Bools>::cell(bools).any())
```
jridgewell, line 139: Is this faster because we've implemented early exiting, or because we no longer need to build up this `Bools` vc just to await it? Does this have implications for the [just-approved](https://github.com/vercel/next.js/pull/55437) https://github.com/vercel/next.js/blob/35fd8bcb0388420618a6ffb08ad5e93f716016c1/packages/next-swc/crates/next-api/src/project.rs#L706

jridgewell (approved):

crates/turbopack-ecmascript/src/references/async_module.rs
sokra: early exist reduces the number of tasks. The Bools::any() also save a task, but it's only a single one.

sokra: Completions is technically similar, but no early exit is possible here. We could avoid 1 task by not using Completions::completed(), but it's only 1 task and it has effects on the invalidation: Without Completions::completed() it would need to rerun the current function when any completion fires. With Completions::completed() it only reruns Completions::completed() and doesn't need to rerun the current function.

There is a difference in how often these methods are called `is_async` is called many many times (for every module), while `any_output_changed` is only called once.

PR #5961, opened 2023-09-15
reduce resolve_internal and import map lookup tasks

sokra:

### DIFF
--- a/crates/turbopack-core/src/resolve/mod.rs
+++ b/crates/turbopack-core/src/resolve/mod.rs
@@ -1010,7 +1010,7 @@ pub async fn resolve(
     request: Vc<Request>,
     options: Vc<ResolveOptions>,
 ) -> Result<Vc<ResolveResult>> {
-    let raw_result = resolve_internal(lookup_path, request, options);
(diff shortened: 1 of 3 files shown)


ForsakenHarmony (approved):

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1080,7 +1080,15 @@ async fn handle_resolve_plugins(
...
-#[turbo_tasks::function]
+fn resolve_internal_boxed(
+    lookup_path: Vc<FileSystemPath>,
+    request: Vc<Request>,
+    options: Vc<ResolveOptions>,
+) -> Pin<Box<dyn Future<Output = Result<Vc<ResolveResult>>> + Send>> {
+    Box::pin(resolve_internal(lookup_path, request, options))
+}
+
+// Not a turbo-tasks function: Called too often
```
ForsakenHarmony, line 1091: I guess this might reduce caching, but I'm assuming you looked into that

sokra: Yeah, it also cases a stack overflow due to recursive calls....

The effect of caching is probably something I need to measure...

PR #5962, opened 2023-09-15
fix: pin thiserror to <1.0.45 which requires updating the toolchain

arlyon:
### Description

The latest thiserror uses certain apis that aren't available to us without a toolchain update.
 
https://github.com/dtolnay/thiserror/pull/246/files

Pin the version instead

### Testing Instructions

Should be covered by existing test suite

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -226,7 +226,8 @@ sourcemap = "6.0.2"
 syn = "1.0.107"
 tempfile = "3.3.0"
 test-case = "3.0.0"
-thiserror = "1.0.38"
+# pinned due to 1.0.45 requiring toolchain update
+thiserror = "<1.0.45"
 tiny-gradient = "0.1.0"
 tokio = "1.25.0"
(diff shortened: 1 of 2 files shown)


gsoltis (approved):

Cargo.toml
```diff
@@ -226,7 +226,8 @@ sourcemap = "6.0.2"
 syn = "1.0.107"
 tempfile = "3.3.0"
 test-case = "3.0.0"
-thiserror = "1.0.38"
+# pinned due to 1.0.45 requiring toolchain update
+thiserror = "<1.0.45"
```
ForsakenHarmony, line 230:
```suggestion
thiserror = "=1.0.44"
```
?

ForsakenHarmony (approved):

Cargo.toml
ForsakenHarmony: doesn't really matter though

gsoltis:
Disabled auto-merge in case we want to make @ForsakenHarmony 's suggested change

PR #5964, opened 2023-09-15
feat(Turborepo): Enable rust daemon

gsoltis:
### Description

 - enables Rust daemon
 - makes `turbo daemon start` work again
 - sets a blanket 100ms timeout on all RPCs to the gRPC server
 - turns down the logging from `TRACE` to `INFO`. Turborepo code is left at `DEBUG`.
 - _DELETES STALE PID FILES_ note this behavior change. If the pid file references a process that is not running, we will delete it and start the daemon. This technically opens the door to a race condition, but having two daemons running is not the end of the world, the socket will lead to one of them and the other will time out eventually. 

### Testing Instructions

Manual + existing test suite


crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -39,9 +39,12 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
...
             client.restart().await?;
         }
-        // connector.connect will have already started the daemon if needed,
-        // so this is a no-op
-        DaemonCommand::Start => {}
+        DaemonCommand::Start => {
+            // We don't care about the client, but we do care that we can connect
+            // which ensures that daemon is started if it wasn't already.
+            let _ = connector.connect().await?;
+            println!("Daemon is running");
```
arlyon, line 46: Nice

crates/turborepo-pidlock/src/lib.rs
```diff
@@ -171,9 +171,14 @@ impl Pidlock {
             Ok(pid) if process_exists(pid) => {
                 Some(pid.try_into().expect("if a pid exists it is a valid u32"))
             }
-            Ok(_) => {
+            Ok(pid) => {
                 warn!("stale pid file at {:?}", self.path);
-                None
+                if let Err(e) = fs::remove_file(&self.path) {
+                    warn!("failed to remove stale pid file: {}", e);
+                    Some(pid.try_into().expect("if a pid exists it is a valid u32"))
```
arlyon, line 178: Do we want to return the pid if the owning process is not running? I think this should return None unconditionally

arlyon (changes requested):
One question otherwise looks solid :)

crates/turborepo-pidlock/src/lib.rs
gsoltis: If we can't remove the pid file, the daemon is not going to start up, because it requires the file it writes to be a new file. If we return `None` here, we'll try to start the daemon, when at this point, we actually do need to kick an error out to the user (we can't remove a stale pid).

It may be possible that we want a `Result` instead of an `Option` as a return value...

cli/internal/daemonclient/daemonclient.go
```diff
@@ -33,9 +34,14 @@ func New(client *connector.Client) *DaemonClient {
 
 // GetChangedOutputs implements runcache.OutputWatcher.GetChangedOutputs
 func (d *DaemonClient) GetChangedOutputs(ctx context.Context, hash string, repoRelativeOutputGlobs []string) ([]string, int, error) {
```
chris-olszewski, line 36: Are the output globs being passed in using a `\` path separator on Windows?

crates/turborepo-filewatch/src/globwatcher.rs
```diff
@@ -24,26 +25,53 @@ pub struct GlobSet {
     exclude: Any<'static>,
 }
 
+#[derive(Debug, Error)]
+pub struct GlobError {
+    // Boxed to minimize error size
+    underlying: Box<wax::BuildError>,
+    raw: String,
```
chris-olszewski, line 32: Maybe `glob` or `input` would be slightly more descriptive than `raw`?

crates/turborepo-filewatch/src/globwatcher.rs
```diff
@@ -24,26 +25,53 @@ pub struct GlobSet {
...
+            .iter()
             .map(|raw_glob| {
-                let glob = Glob::from_str(&raw_glob)?.to_owned();
+                let glob = compile_glob(raw_glob)?;
                 Ok(glob)
             })
-            .collect::<Result<Vec<_>, wax::BuildError>>()?;
-        let exclude = wax::any(excludes)?.to_owned();
+            .collect::<Result<Vec<_>, GlobError>>()?;
+        let exclude = wax::any(excludes)
```
chris-olszewski, line 69: Reading the doc comment of `wax::any`, since we're passing `Glob`s I think we'll only fail if the combined excluded globs end up being [10 MB](https://docs.rs/regex/latest/src/regex/builders.rs.html#53). If we end up hitting that, we probably shouldn't print out that much data to the terminal and just say that the combined excludes are too large.

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -187,13 +188,14 @@ where
             ));
 
         Server::builder()
+            .timeout(Duration::from_millis(100))
```
chris-olszewski, line 191: Maybe pull out this duration into a constant labeled `REQUEST_TIMEOUT` or something. The `timeout` doesn't make it super clear what the duration controls.

crates/turborepo-pidlock/src/lib.rs
chris-olszewski: +1 on making this return `Result`

chris-olszewski (approved):
LGTM

.vscode/launch.json
```diff
@@ -137,7 +137,7 @@
       "request": "launch",
       "preLaunchTask": "prepare turbo",
       "program": "${workspaceRoot}/target/debug/turbo",
-      "args": ["daemon"],
+      "args": ["--skip-infer", "daemon"],
```
NicholasLYang, line 140: What's the reason for this change?

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -171,8 +174,20 @@ pub async fn daemon_server(
         }
         CloseReason::Interrupt
     });
-    // TODO: be more methodical about this choice:
-    let cookie_dir = base.repo_root.join_component(".git");
+    // We already store logs in .turbo and recommend it be gitignore'd.
```
NicholasLYang, line 177: I'm fine with this, but we should note that we only add `.turbo` to people's gitignore when they run `turbo link`, so if someone adds turbo and it starts the daemon, then they might end up accidentally committing this cookie dir.

gsoltis:
pidfile changes implemented here: https://github.com/vercel/turbo/pull/6025

Will rebase and update once that's merged

crates/turborepo-lib/src/commands/daemon.rs
gsoltis: We also write cached logs there: https://github.com/vercel/turbo/blob/2bb8c9a2477b5fb20b4711fe34412559bc85aab9/cli/internal/graph/graph.go#L241

.vscode/launch.json
gsoltis: It's for debugging, and it ensures that we run the daemon that we're intending to debug, rather than whatever the repo depends on at the time.

cli/internal/daemonclient/daemonclient.go
gsoltis: some parts of the glob have `\` separators, I think from filepaths that we concat with the globs.

crates/turborepo-filewatch/src/globwatcher.rs
gsoltis: changed to `raw_glob`

crates/turborepo-filewatch/src/globwatcher.rs
gsoltis: I went to switch this to something like "globs elided, combined excludes are too large", but decided to back out the change. I think in the (hopefully extremely rare) case when this happens, we still want to know what those globs are, as a pointer to where they came from, vs not having any idea.

If in the future we have some provenance information, like "the globs from line 5 of `turbo.json` are too big", I think it would be better to just do that, but while we're lacking that information, I think explicit is better here so that a user / we have a chance at tracking down where the globs came from.

arlyon (approved):

PR #5966, opened 2023-09-15
fix(turborepo): Run outline fixes

NicholasLYang:
### Description
Found a lot of bugs while testing #5851, so split it off into a separate PR.

### Testing Instructions

Closes TURBO-1334


crates/turborepo-lib/src/config/turbo.rs
```diff
@@ -13,7 +13,7 @@ use crate::{
     config::Error,
     package_json::PackageJson,
     run::task_id::{TaskId, TaskName, ROOT_PKG_NAME},
-    task_graph::{BookkeepingTaskDefinition, Pipeline, TaskDefinitionHashable, TaskOutputs},
```
NicholasLYang, line 16: This is renamed because it's unnecessarily confusing to have a `-Hashable` struct that is not actually used to hash anything

crates/turborepo-lib/src/engine/builder.rs
```diff
@@ -238,6 +248,8 @@ impl<'a> EngineBuilder<'a> {
             }
         }
 
+        graph::validate_graph(&engine.task_graph)?;
```
NicholasLYang, line 251: We need to validate against cycles or self referential tasks

crates/turborepo-lib/src/hash/mod.rs
```diff
@@ -41,15 +41,16 @@ mod proto_capnp {
...
 }
 
+#[derive(Debug)]
 pub struct TaskHashable<'a> {
     // hashes
     pub(crate) global_hash: &'a str,
     pub(crate) task_dependency_hashes: Vec<String>,
     pub(crate) hash_of_files: &'a str,
-    pub(crate) external_deps_hash: String,
+    pub(crate) external_deps_hash: Option<String>,
```
NicholasLYang, line 50: We need to wrap fields in `Option` to mimic how Go automatically omits string fields that are empty when hashing to capnproto (it can do this because all fields are optional in capnproto).

crates/turborepo-lib/src/package_graph/builder.rs
```diff
@@ -63,6 +67,7 @@ impl<'a> PackageGraphBuilder<'a> {
             package_manager: None,
             package_jsons: None,
             lockfile: None,
+            quiet: false,
```
NicholasLYang, line 70: Mostly for testing because we don't want the logging outputs to end up in our integration tests, since we're running the Rust hashing code silently before executing the Go code.

crates/turborepo-lib/src/run/global_hash.rs
```diff
@@ -64,12 +67,23 @@ pub fn get_global_hash_inputs<'a, L: ?Sized + Lockfile>(
     let mut global_deps = HashSet::new();
 
     if !global_file_dependencies.is_empty() {
-        let globs = package_manager.get_workspace_globs(root_path)?;
+        let exclusions = match package_manager.get_workspace_globs(root_path) {
```
NicholasLYang, line 70: Technically this is a little less efficient than the Go version because we're getting the inclusions even though we don't need them. But that's a minor cost since regardless we have to deserialize the workspaces, process the globs, etc.

crates/turborepo-lib/src/run/global_hash.rs
```diff
@@ -125,11 +142,15 @@ pub fn get_global_hash_inputs<'a, L: ?Sized + Lockfile>(
 impl<'a> GlobalHashableInputs<'a> {
     pub fn calculate_global_hash_from_inputs(mut self) -> String {
         match self.env_mode {
-            EnvMode::Infer if !self.pass_through_env.is_empty() => {
+            // In infer mode, if there is any pass_through config (even if it is an empty array)
+            // we'll hash the whole object, so we can detect changes to that config
+            // Further, resolve the envMode to the concrete value.
+            EnvMode::Infer if self.pass_through_env.is_some() => {
```
NicholasLYang, line 148: Classic `nil` is not empty bug

crates/turborepo-lib/src/run/global_hash.rs
```diff
@@ -29,26 +30,28 @@ enum GlobalHashError {}
...
+    is_monorepo: bool,
     root_workspace: &WorkspaceInfo,
     root_path: &AbsoluteSystemPath,
     package_manager: &PackageManager,
     lockfile: Option<&L>,
     global_file_dependencies: &'a [String],
     env_at_execution_start: &EnvironmentVariableMap,
     global_env: &'a [String],
-    global_pass_through_env: &'a [String],
+    global_pass_through_env: Option<&'a [String]>,
```
gsoltis, line 54: As above, what's the distinction vs an empty array?

crates/turborepo-lib/src/run/global_hash.rs
gsoltis: IIUC, there is a distinction between empty array and not set?

gsoltis (approved):
Not a blocker, just a little confused about the `Option<Vec<...>>`

crates/turborepo-lib/src/run/global_hash.rs
NicholasLYang: Yep, in this case due to our code but also capnproto makes that distinction too

crates/turborepo-lib/src/config/turbo.rs
```diff
@@ -31,7 +31,7 @@ pub struct TurboJson {
     pub(crate) global_deps: Vec<String>,
     pub(crate) global_dot_env: Vec<RelativeUnixPathBuf>,
     pub(crate) global_env: Vec<String>,
-    pub(crate) global_pass_through_env: Vec<String>,
+    pub(crate) global_pass_through_env: Option<Vec<String>>,
```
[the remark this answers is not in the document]
NicholasLYang: This is a quirk of Go. Specifically `[]string` is equivalent to `Option<Vec<String>>` in Rust because `[]string` can be `nil` or an empty slice. In the `nil` case, the field is not serialized at all, while in the empty case it is serialized as an empty array.

mehulkar (commented):
Seems like a pretty grab bag of changes, it would be helpful to have a description of what's actually being fixed (I couldn't glean anything useful out of the PR comments or the commit messages).

crates/turborepo-lib/src/run/global_hash.rs
```diff
@@ -102,11 +116,14 @@ pub fn get_global_hash_inputs<'a, L: ?Sized + Lockfile>(
...
-    let root_external_dependencies_hash = root_workspace.get_external_deps_hash();
+    let root_external_dependencies_hash =
+        is_monorepo.then(|| root_workspace.get_external_deps_hash());
 
     debug!(
-        "rust external deps hash: {}",
+        "external deps hash: {}",
         root_external_dependencies_hash
+            .as_deref()
+            .unwrap_or("no hash (single package)")
```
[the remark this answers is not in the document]
NicholasLYang: This isn't an error, just a log output. In single package mode we don't calculate the root external dependencies hash

crates/turborepo-lib/src/task_graph/mod.rs
```diff
@@ -35,12 +35,11 @@ pub struct TaskOutputs {
     pub exclusions: Vec<String>,
 }
 
-// taskDefinitionHashable exists as a definition for PristinePipeline, which is
-// used downstream for calculating the global hash. We want to exclude
-// experimental fields here because we don't want experimental fields to be part
-// of the global hash.
+// These are the stable fields of a TaskDefinition, versus the experimental ones
+// TODO: Consolidate this and experiments, because the split is an artifact of
+// the Go implementation
```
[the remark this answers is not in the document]
NicholasLYang: Yeah definitely a post-port refactor I'm looking forward to

PR #5970, opened 2023-09-16
fix(turbo-gen): bun workspace generated dependency range should be workspace:*

KaiSpencer:
### Description

A bun monorepo uses the `workspace:*` value for workspace dependencies as per pnpm
[Bun Workspace Docs](https://bun.sh/docs/install/workspaces)

Amend turbo-gen package to use `workspace:*` dependency range for bun and pnpm 

Previously generated:
```json
{
  "my-package": "*"
}
```

Correctly generated:
```json
{
  "my-package": "workspace:*"
}
```

### DIFF
--- a/packages/turbo-gen/src/commands/workspace/prompts.ts
+++ b/packages/turbo-gen/src/commands/workspace/prompts.ts
@@ -7,12 +7,23 @@ import type { Project, Workspace } from "@turbo/workspaces";
 import {
   validateDirectory,
   logger,
+  type PackageManager,
   type DependencyGroups,
   type PackageJson,
 } from "@turbo/utils";
 import { getWorkspaceStructure } from "../../utils/getWorkspaceStructure";
 import type { WorkspaceType } from "../../generators/types";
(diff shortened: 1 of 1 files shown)


nathanhammond:
We had a conversation with Jarred before selecting *. At this point we don't intend to change—but that doesn't mean we won't consider it in the future.

We're monitoring reported issues in Bun—and reporting many as well. The two approaches in Bun are not _intended_ to diverge, but if they do it's a bug. If one eventually has greater compatibility we may adopt it.

We do like workspace:* because it prevents accidental registry shadowing, but we are also choosing based upon people's familiarity.

Thank you for the PR! We do appreciate you paying attention to the details. Keep us up to date with how workspace:* goes in your projects?

KaiSpencer:
Thank you @nathanhammond for the response and explanation behind the intent which is mistook for a bug! 

Coming from pnpm I like workspace:*, and using bun it did actually bite me with some of the standard monorepo base packages like "eslint-config-custom", "ts-config". 

I'll raise an issue in future if I come across anything more compelling.

PR #5979, opened 2023-09-19
feat(create-turbo): support package manager opt

tknickman:
### Description

Support package manager as an option for `create-turbo`. Existing command is still available for backwards compatibility. If both are specified, option wins. 

Closes TURBO-1335

### DIFF
--- a/packages/create-turbo/src/cli.ts
+++ b/packages/create-turbo/src/cli.ts
@@ -3,7 +3,7 @@
 import http from "node:http";
 import https from "node:https";
 import chalk from "chalk";
-import { Command } from "commander";
+import { Command, Option } from "commander";
 import { logger } from "@turbo/utils";
 import { ProxyAgent } from "proxy-agent";
 import cliPkg from "../package.json";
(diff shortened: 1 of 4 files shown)


packages/create-turbo/src/cli.ts
```diff
@@ -21,9 +21,17 @@ const createTurboCli = new Command();
...
   .description("Create a new Turborepo")
-  .usage(`${chalk.bold("<project-directory> <package-manager>")} [options]`)
+  .usage(`${chalk.bold("<project-directory>")} [options]`)
   .argument("[project-directory]")
+  // TODO: argument is still provided (but removed from help)
+  // for backwards compatibility, remove this in the next major
   .argument("[package-manager]")
+  .addOption(
+    new Option(
+      "-m, --manager <package-manager>",
```
nathanhammond, line 31: Making this `--package-manager` for clarity. Short flag exists for frequent users.

nathanhammond, line 31: `p` being for `--example-path` is killing me. I just love ... path ... dependency.

nathanhammond (approved):

packages/create-turbo/src/cli.ts
tknickman: Yea, this was kept the same to match Next.js

PR #5980, opened 2023-09-19
fix:(create-turbo) Review changes for package manager flag.

nathanhammond:
#5979 was on automerge and I didn't notice. This contains the review changes.

Closes TURBO-1336

### DIFF
--- a/packages/create-turbo/src/commands/create/index.ts
+++ b/packages/create-turbo/src/commands/create/index.ts
@@ -59,7 +59,11 @@ export async function create(
   packageManagerCmd: CreateCommandArgument,
   opts: CreateCommandOptions
 ) {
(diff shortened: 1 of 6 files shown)


packages/create-turbo/src/commands/create/index.ts
```diff
@@ -59,7 +59,11 @@ export async function create(
   packageManagerCmd: CreateCommandArgument,
   opts: CreateCommandOptions
 ) {
-  const { manager: packageManagerOpt, skipInstall, skipTransforms } = opts;
+  const {
+    packageManager: packageManagerOpt,
```
nathanhammond, line 63: This will _eventually_ clean up nicely.

PR #5986, opened 2023-09-20
fix(turborepo): Make shim do complete package manager detection.

nathanhammond:
In 1.10.2 we introduced package manager detection into workspace existence detection. However, we did not include the `package.json` contents into the method call, which resulted in incomplete detection.

This updates that method call to include the `package.json` (if present) which gives us the correct behavior for package manager detection. This also pulls on a turbopath type thread to do some more migration.

The change that resulted in less-robust workspace detection:
https://github.com/vercel/turbo/commit/4e99d40db30d8584f140556a01986624b47dad60#diff-8459a232d027b507053b9df40feba72a5fdc6b1b08d5348140ab2ce9dd4d861e

Closes TURBO-1341

### DIFF
--- a/crates/turborepo-lib/src/shim.rs
+++ b/crates/turborepo-lib/src/shim.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use anyhow::{anyhow, Result};
-use camino::{Utf8Path, Utf8PathBuf};
+use camino::Utf8PathBuf;
 use const_format::formatcp;
 use dunce::canonicalize as fs_canonicalize;
 use semver::Version;
@@ -20,8 +20,8 @@ use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
 use turborepo_ui::UI;
 
 use crate::{
-    cli, get_version, package_manager::WorkspaceGlobs, spawn_child, tracing::TurboSubscriber,
-    PackageManager, Payload,
+    cli, get_version, package_json::PackageJson, package_manager::WorkspaceGlobs, spawn_child,
+    tracing::TurboSubscriber, PackageManager, Payload,
 };
 
 // all arguments that result in a stdout that much be directly parsable and
@@ -195,11 +195,6 @@ pub enum RepoMode {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-struct PackageJson {
-    version: String,
-}
-
-#[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 struct YarnRc {
     pnp_unplugged_folder: Utf8PathBuf,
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/shim.rs
```diff
@@ -403,20 +395,20 @@ impl LocalTurboState {
...
             #[allow(clippy::needless_borrow)]
-            let bin_path = root.join(&platform_package_executable_path);
+            let bin_path = root.join_components(&platform_package_executable_path_components);
             match fs_canonicalize(&bin_path) {
                 Ok(bin_path) => {
-                    let resolved_package_json_path = root.join(platform_package_json_path);
-                    let platform_package_json_string =
-                        fs::read_to_string(resolved_package_json_path).ok()?;
-                    let platform_package_json: PackageJson =
-                        serde_json::from_str(&platform_package_json_string).ok()?;
```
nathanhammond, line 413: As a side effect of using a unified `PackageJson` and `PackageJson::load` we remove the simplistic `serde_json` approach. Almost certainly a tiny bit slower but this gets called a maximum of one time and consistent parsing of `package.json` is better.

crates/turborepo-lib/src/shim.rs
```diff
@@ -467,17 +459,19 @@ impl RepoState {
         let potential_turbo_roots = reference_dir
             .ancestors()
             .filter_map(|path| {
-                let has_package_json = path.join_component("package.json").exists();
+                let package_json = PackageJson::load(&path.join_component("package.json")).ok();
```
nathanhammond, line 462: This is the reason I pulled the `turbopath` thread in this PR. The load needed an `AbsoluteSystemPath` so rather than do it at this call site I plumbed it through as far back as I could justify.

crates/turborepo-lib/src/shim.rs
```diff
@@ -467,17 +459,19 @@ impl RepoState {
...
                 if !has_package_json && !has_turbo_json {
                     return None;
                 }
 
                 // FIXME: We should save this package manager that we detected
-                let workspace_globs = PackageManager::get_package_manager(path, None)
-                    .and_then(|mgr| mgr.get_workspace_globs(path))
-                    .ok();
+                let workspace_globs =
+                    PackageManager::get_package_manager(path, package_json.as_ref())
```
nathanhammond, line 472: This (`None` => `package_json`) is the actual bugfix in this PR.

nathanhammond (commented):
Reviewer's guide!

crates/turborepo-lib/src/shim.rs
```diff
@@ -317,8 +317,14 @@ impl LocalTurboState {
     // - `pnpm install`
     // - `npm install --install-strategy=linked`
     fn generate_linked_path(root_path: &AbsoluteSystemPath) -> Option<AbsoluteSystemPathBuf> {
-        let canonical_path =
-            fs_canonicalize(root_path.join_components(&["node_modules", "turbo", ".."])).ok()?;
```
nathanhammond: Detailed tests caught that this was not materially the same behavior as the original.

crates/turborepo-lib/src/shim.rs
```diff
@@ -317,8 +317,14 @@ impl LocalTurboState {
...
+        let canonical_path = fs_canonicalize(
+            root_path
+                .as_path()
+                .join("node_modules")
+                .join("turbo")
+                .join(".."),
+        )
+        .ok()?;
```
nathanhammond, line 327: So now it uses `as_path` to expose the `Utf8Path` operations directly.

nathanhammond (commented):
Why did tests fail?

crates/turborepo-lib/src/shim.rs
```diff
@@ -306,48 +301,49 @@ impl LocalTurboState {
...
-    fn generate_linked_path(root_path: &Utf8Path) -> Option<Utf8PathBuf> {
-        let canonical_path =
-            fs_canonicalize(root_path.join("node_modules").join("turbo").join("..")).ok()?;
+    fn generate_linked_path(root_path: &AbsoluteSystemPath) -> Option<AbsoluteSystemPathBuf> {
+        let canonical_path = fs_canonicalize(
+            root_path
+                .as_path()
+                .join("node_modules")
+                .join("turbo")
+                .join(".."),
```
gsoltis, line 325:
```suggestion
root_path.join_component("node_modules").as_path()
```

gsoltis (approved):
I think we can probably go even a little bit further with `turbopath` here, but we don't need to do that to get the bugfix in.

crates/turborepo-lib/src/shim.rs
gsoltis: If the two behave differently, that's probably something we ought to fix.

crates/turborepo-lib/src/shim.rs
nathanhammond: I originally did the equivalent, but had to walk that back after our tests caught a limitation with that approach:

https://github.com/vercel/turbo/pull/5986/files/721c8bc947b4bef8904a10c009bd0c2828774753..e1252d6bd9ed7952c8968fa872a5770b1eda423e#diff-8459a232d027b507053b9df40feba72a5fdc6b1b08d5348140ab2ce9dd4d861e

nathanhammond: (`AbsoluteSystemPath` does a lexical clean, underlying system behavior just does a string concat. If we lexically clean prior to file-system-aware canonicalization we get the wrong output.)

PR #5993, opened 2023-09-21
add progress and top self time to convert trace tool

sokra:
### Description

add progress
add top 10 self time spans

Closes WEB-1624


ForsakenHarmony (approved):
Could use https://crates.io/crates/indicatif

PR #5994, opened 2023-09-21
Allow task log prefixing in Github Actions

Zertsov:
### Description

Add task log prefixing for Github Actions.

### Testing Instructions

1. `cd` into `turborepo-tests/integration`
2. run: `.cram_env/bin/prysk --shell=bash tests/ordered/github.t`


turborepo-tests/integration/tests/ordered/github.t
```diff
@@ -33,6 +33,36 @@ because otherwise prysk interprets them as multiline commands
   Cached:    0 cached, 2 total
     Time:\s*[\.0-9]+m?s  (re)
   
+# Build as if we are in Github Actions with a task log prefix.
+  $ export GITHUB_ACTIONS=1
+  $ ${TURBO} run build --force --log-prefix="task"
```
mehulkar, line 38:
```suggestion
GITHUB_ACTIONS=1 ${TURBO} run build --force --log-prefix="task" --filter=util
```

- not blocking, but i try to set env vars just for the command so it doesn't leak into other tests in the file
- added a filter flag so you don't run into non-deterministic ordering output. Based on the fact that `util#build` is logged _after_ `my-app#build` in this test right now, I'm guessing they are running concurrently

mehulkar (approved):
small thing with tests

turborepo-tests/integration/tests/ordered/github.t
Zertsov: Should I apply this to the preceding test? Also if I'm not mistaken don't these `.t` files need to start with a `$` in order for the command to be executed in a shell?

chris-olszewski: +1 to the `--filter`

chris-olszewski (approved):

turborepo-tests/integration/tests/ordered/github.t
```diff
@@ -33,6 +33,36 @@ because otherwise prysk interprets them as multiline commands
   Cached:    0 cached, 2 total
     Time:\s*[\.0-9]+m?s  (re)
   
+# Build as if we are in Github Actions with a task log prefix.
+  $ export GITHUB_ACTIONS=1
```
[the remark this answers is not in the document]
Zertsov: Probably my unfamiliarity with the `.t` files, but I thought that that export was just for the test it was running in. Can any single test do changes to subsequent tests (in this case, doing the `export` affecting all other tests)?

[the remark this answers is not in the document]
nathanhammond: Prysk runs each _file_ in a new shell, so they're not localized per-test. The way these comparisons work it:
- Ignores all unindented lines.
- Runs all of the `$`-prefixed ones.
- inlines stdout + stderr
- Grabs the output contents of the entire shell.
- Checks for pass fail based on diff of the input and the output.

If you ever get a failing test it pops out a `.t.err` file which you can just diff with the `.t` file. It's ... primitive? But has saved us enough times to be worth dealing with bad regex ergonomics, a Python toolchain, and other quirks.

turborepo-tests/integration/tests/ordered/github.t
mehulkar: > do we care to have two tasks or is it ok to just test that a single task has the prefix?
single task is fine IMO

> Should I apply this to the preceding test?
generally, I wouldn't change unrelated, existing code. That has a tendency to balloon the diff and make git history harder to follow (not to mention it might cause unexpected issues and sink your time). In this case, 🤷🏾

PR #6000, opened 2023-09-21
fix(turbopack-ecmascript): make free var replacements (compile time defines) work for member expressions and allow JSON values

ForsakenHarmony:
### Description

Closes WEB-1589


crates/turbopack-ecmascript/src/references/constant_value.rs
```diff
@@ -35,17 +35,20 @@ impl CodeGenerateable for ConstantValue {
...
-        .into();
+        let path = &self.path.await?;
 
-        Ok(CodeGeneration { visitors }.cell())
+        let visitor = create_visitor!(path, visit_mut_expr(expr: &mut Expr) {
+            *expr = match value {
+                CompileTimeDefineValue::Bool(true) => quote!("(\"TURBOPACK compile-time value\", true)" as Expr),
+                CompileTimeDefineValue::Bool(false) => quote!("(\"TURBOPACK compile-time value\", false)" as Expr),
+                CompileTimeDefineValue::String(ref s) => quote!("(\"TURBOPACK compile-time value\", $e)" as Expr, e: Expr = s.to_string().into()),
+                CompileTimeDefineValue::JSON(ref s) => quote!("(\"TURBOPACK compile-time value\", JSON.parse($e))" as Expr, e: Expr = s.to_string().into()),
```
jridgewell, line 45: Is this used somewhere? Can we hoist this value into a module-level constant and refer to it?

jridgewell (approved):

crates/turbopack-core/src/compile_time_info.rs
```diff
@@ -77,6 +77,7 @@ macro_rules! free_var_references {
 pub enum CompileTimeDefineValue {
     Bool(bool),
     String(String),
+    JSON(String),
```
sokra, line 80: I'm not sure about this unstructured form... storing a serde::Value would allow us to do more static evaluation, like `typeof` and could replace bool and string too.

sokra (approved):

crates/turbopack-core/src/compile_time_info.rs
ForsakenHarmony: yeah, I wanted to do that (actually I did implement it that way), but the `serialization = "auto_for_input"` requires this to implement `Hash` and `Ord`, but `serde_json::Value` doesn't implement either of those

crates/turbopack-ecmascript/src/references/constant_value.rs
ForsakenHarmony: what do you mean?

jridgewell: Currently, we have something like:

```js
function foo() {
  return JSON.parse('{"foo": true').foo
}
```

It'd be better if we hoisted that JSON constant:
```js
const __FOO__ = JSON.parse('{"foo": true');
function foo() {
  return __FOO__.foo
}
```

ForsakenHarmony: I guess it would be nice, but also kinda annoying to do with how our codegen stuff currently works

webpack also injects everything directly

ForsakenHarmony: the module level constant would still be mutable on the inside, which could break things as well 🤔

PR #6010, opened 2023-09-22
Signal when signal handler is registered

gsoltis:
### Description

- add a printout to signal that the signal handler is registered

### Testing Instructions

Existing test suite

Closes TURBO-1348


chris-olszewski (changes requested):
This should be combined with a revert of https://github.com/vercel/turbo/pull/6006

chris-olszewski (approved):
If it passes on CI, then :shipit:

PR #6013, opened 2023-09-22
chore: remove anyhow from task cache

chris-olszewski:
### Description

In general we're trying to move away from anyhow. On a more operational level since a cache miss is reported as an error, keeping our own error variant where it's easy to check if it is a `CacheError::CacheMiss` without downcasting is preferable.

### Testing Instructions

`cargo check`

### DIFF
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -16,6 +16,14 @@ use crate::{
     task_graph::{TaskDefinition, TaskOutputs},
 };
 
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Error replaying logs: {0}")]
+    Ui(#[from] turborepo_ui::Error),
+    #[error("Error accessing cache: {0}")]
+    Cache(#[from] turborepo_cache::CacheError),
+}
+
 pub struct RunCache {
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/run/cache.rs
```diff
@@ -16,6 +16,20 @@ use crate::{
...
+    Ui(#[from] turborepo_ui::Error),
+    #[error("Error accessing cache: {0}")]
+    Cache(#[from] turborepo_cache::CacheError),
+}
+
+impl Error {
+    pub fn is_cache_miss(&self) -> bool {
+        matches!(&self, Self::Cache(CacheError::CacheMiss))
+    }
+}
```
nathanhammond, line 31: Unused? (If you have a plan for use seems fine.)

nathanhammond (approved):

Zertsov (approved):

crates/turborepo-lib/src/run/cache.rs
chris-olszewski: Yes, currently unused but will be to differentiate between a cache miss and an error accessing the cache which is needed [in run](https://github.com/vercel/turbo/blob/main/cli/internal/run/real_run.go#L443)

PR #6018, opened 2023-09-25
add warning when multiple package managers may impact inference

arlyon:
### Description

Adds a small warning when multiple package managers are detected in the same repo. This interferes with package inference and may cause repositories to be reported as single-package when they are not.

### Testing Instructions

Create a repo with both a package-lock.json and a yarn.lock and observe a new warning log line.

Closes TURBO-1353

### DIFF
--- a/crates/turborepo-lib/src/shim.rs
+++ b/crates/turborepo-lib/src/shim.rs
@@ -499,7 +499,7 @@ fn spawn_local_turbo(
     let already_has_single_package_flag = shim_args
         .remaining_turbo_args
         .contains(&"--single-package".to_string());
-    let should_add_single_package_flag = repo_state.mode == RepoMode::SinglePackage
+    let should_add_single_package_flag = repo_state.get_mode() == RepoMode::SinglePackage
         && !already_has_single_package_flag
         && supports_skip_infer_and_single_package;
 
--- a/crates/turborepo-repository/src/inference.rs
+++ b/crates/turborepo-repository/src/inference.rs
@@ -6,7 +6,7 @@ use crate::{
     package_manager::{self, PackageManager, WorkspaceGlobs},
 };
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Copy, Clone)]
 pub enum RepoMode {
     SinglePackage,
     MultiPackage,
@@ -15,7 +15,7 @@ pub enum RepoMode {
 #[derive(Debug)]
 pub struct RepoState {
     pub root: AbsoluteSystemPathBuf,
-    pub mode: RepoMode,
+    mode: RepoMode,
     pub package_manager: Result<PackageManager, package_manager::Error>,
 }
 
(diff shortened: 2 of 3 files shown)


crates/turborepo-lib/src/shim.rs
```diff
@@ -474,10 +477,20 @@ impl RepoState {
                 }
 
                 // FIXME: We should save this package manager that we detected
-                let workspace_globs =
-                    PackageManager::get_package_manager(path, package_json.as_ref())
-                        .and_then(|mgr| mgr.get_workspace_globs(path))
-                        .ok();
+                let workspace_globs = match PackageManager::get_package_manager(path, None)
```
chris-olszewski, line 480: Do we not want to include the `package_json` here? We're now relying running detection even if they specified a package manager via `packageManager`

chris-olszewski (changes requested):
I think this needs is `use crate::package_json::PackageJson` and I do have questions about dropping the `package.json` to the package manager detection.

crates/turborepo-lib/src/shim.rs
arlyon: I think that was a leftover from trying to trace a fix. Definitely not supposed to be in the commit.

nathanhammond (approved):
Regarding @chris-olszewski's opinion: I feel like this should be hoisted up to the _first_ instance where it can happen.

This _happens_ to be the first instance, so I don't _really_ mind it being here.

Alternatively, we could `exit 1`. I think I'd be happy with that too.

crates/turborepo-lib/src/shim.rs
```diff
@@ -475,9 +479,20 @@ impl RepoState {
...
                 // FIXME: We should save this package manager that we detected
                 let workspace_globs =
-                    PackageManager::get_package_manager(path, package_json.as_ref())
+                    match PackageManager::get_package_manager(path, Some(package_json))
                         .and_then(|mgr| mgr.get_workspace_globs(path))
-                        .ok();
+                    {
+                        Ok(globs) => Some(globs),
+                        Err(Error::MultiplePackageManagers { managers }) => {
+                            warn!(
```
nathanhammond, line 487: This is actually the wrong place to print out the warning. This is generating _potential_ `turbo` roots, but we don't actually know if this is going to be the one selected. This could be _entirely_ below the pivot for the eventually-detected root, making the warning message wrong.

So, for this to work, we would need to turn `workspace_globs` into `detect_result` and effectively forward the `Result<PackageManager, Error>`.

Then, at consumption time in `process_potential_turbo_roots`, we know that we got an ambiguous detection. If `process_potential_turbo_roots` is materially impacted by the workspace glob being ambiguous, _then_ we print a warning.

nathanhammond (changes requested):

chris-olszewski (changes requested):
Can you update the testing instructions? I've been trying to test this following the test instructions and I haven't been able to get the warning to print.

crates/turborepo-repository/src/inference.rs
```diff
@@ -107,6 +107,23 @@ impl RepoState {
             .map(|root| root.into())
             .ok_or_else(|| Error::NotFound(reference_dir.to_owned()))
     }
+
+    pub fn get_mode(&self) -> RepoMode {
+        static WARN_INFERENCE: std::sync::OnceLock<()> = std::sync::OnceLock::new();
```
chris-olszewski, line 112: This side-effect concerns me a bit since it means that if we change when we query the mode for the first time the warning will appear there instead. Also, if this happens to get called in the shim we could end up printing the warning twice.

chris-olszewski (commented):
Just because I find the shim code very confusing and hard to follow, when what situation will the `cli.rs` warning trigger vs the `shim.rs` warning trigger?

crates/turborepo-repository/src/inference.rs
arlyon: Originally this PR would print at calculation time but I was asked by Nathan to move it to query time instead. Perhaps we should just have manual check instead so we can control when this happens.

PR #6022, opened 2023-09-25
feat(prune): add ability to specify scopes via argument instead of flag

chris-olszewski:
### Description

As per @nathanhammond's suggestion, `--scope` should be an arg since it is required.

UI change: 
We now have help text for the `SCOPE` argument, wording isn't great, but it matches the wording used for the overall prune command.

Clap will now complain about missing scopes for prune before our error message in `prune.rs` where we check if [`scopes` is empty](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/commands/prune.rs#L234C12-L234C12). All existing usages of `turbo prune` will continue working just a different error message:

Old:
```
[1 olszewski@chriss-mbp] /Users/olszewski/code/vercel/turborepo/turborepo-tests/integration $ turbo prune
Turbo error: at least one target must be specified
```

New:
```
[0 olszewski@chriss-mbp] /Users/olszewski/code/vercel/turborepo/turborepo-tests/integration $ turbo_dev prune
 ERROR  the following required arguments were not provided:
  <SCOPE>...

Usage: turbo prune <SCOPE>...

```

### Testing Instructions

Added new unit tests that use the arg variant and a new test to make sure users can't mix the two ways to specify scope.

Closes TURBO-1357

### DIFF
--- a/crates/turborepo-lib/src/cli.rs
+++ b/crates/turborepo-lib/src/cli.rs
@@ -323,8 +323,15 @@ pub enum Command {
     Logout {},
     /// Prepare a subset of your monorepo.
     Prune {
-        #[clap(long)]
-        scope: Vec<String>,
+        #[clap(hide = true, long)]
+        scope: Option<Vec<String>>,
+        /// Workspaces that should be included in the subset
+        #[clap(
+            required_unless_present("scope"),
+            conflicts_with("scope"),
+            value_name = "SCOPE"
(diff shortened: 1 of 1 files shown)


nathanhammond (approved):
Needs doc updates, but should in theory coincide with the next release:
https://github.com/vercel/turbo/pull/6028

PR #6027, opened 2023-09-26
feat(turborepo-auth): create new crate to handle authentication to backend

mehulkar:
This creates a `turborepo-auth` crate that contains the logic for logging
into a backend (Vercel by default). It only covers logging into non-SAML enforced
teams to get a proof of concept merged, sso_login and logout will be moved in
the future.

## Test with

```bash
turbo run build -F cli
./target/debug/turbo login --skip-infer
```

Closes TURBO-1360


chris-olszewski (approved):
LGTM. One of the immediate TODO's once SSO gets moved into this crate is to drop `anyhow` and offer up a singular `Error` type for the library.

crates/turborepo-auth/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+pub async fn login<F>(
+    api_client: APIClient,
+    ui: &UI,
+    mut set_token: F,
+    login_url_configuration: &str,
+) -> Result<()>
+where
+    F: FnMut(&str) -> Result<()>,
+{
+    println!("WHAT");
```
chris-olszewski, line 38:
```suggestion
```

crates/turborepo-auth/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+        let mut got_token = String::new();
+        let set_token = |t: &str| -> Result<(), anyhow::Error> {
+            got_token = t.to_string();
+            Ok(())
+        };
+
+        login(api_client, &ui, set_token, &url).await.unwrap();
+
+        api_server.abort();
+        println!("got_token: {}", got_token);
```
chris-olszewski, line 183: I think this can get dropped as well
```suggestion
```

chris-olszewski (approved):

crates/turborepo-auth/src/lib.rs
mehulkar: WHAT!

PR #6028, opened 2023-09-26
Update docs and tests.

nathanhammond:
@chris-olszewski I didn't want to scribble in your branch. But this is basically `s/--scope//`.

It should land right before the next time we do a point release.

Closes TURBO-1361


chris-olszewski (approved):
Thanks!

arlyon (approved):

PR #6034, opened 2023-09-26
Reuse existing login tokens

Zertsov:
### Description

Instead of fetching new tokens on every login attempt, look for an existing token to exit early.

### Testing Instructions

```
cd creates/turborepo-lib/src/commands
cargo test test_login -- --exact
cargo test test_sso_login -- --exact
```

### DIFF
--- a/crates/turborepo-lib/src/commands/login.rs
+++ b/crates/turborepo-lib/src/commands/login.rs
@@ -1,12 +1,50 @@
 use anyhow::Result;
 use turborepo_api_client::APIClient;
 use turborepo_auth::{login as auth_login, sso_login as auth_sso_login};
+use turborepo_ui::{BOLD, CYAN, UI};
 
 use crate::commands::CommandBase;
 
+fn print_cli_authorized(user: &str, ui: &UI) {
+    println!(
+        "
+{} Turborepo CLI authorized for {}
+{}
+{}
+",
+        ui.rainbow(">>> Success!"),
+        user,
+        ui.apply(
+            CYAN.apply_to("To connect to your Remote Cache, run the following in any turborepo:")
+        ),
+        ui.apply(BOLD.apply_to("  npx turbo link"))
+    );
+}
+
 pub async fn sso_login(base: &mut CommandBase, sso_team: &str) -> Result<()> {
-    let api_client: APIClient = base.api_client()?;
     let ui = base.ui;
+    // Check if token exists first. Must be there for the user and contain the team
+    // passed into this function.
+    if let Some(token) = base.user_config()?.token() {
+        let client_user: APIClient = base.api_client()?;
+        let client_teams: APIClient = base.api_client()?;
+
+        let (result_user, result_teams) =
+            tokio::join!(client_user.get_user(token), client_teams.get_teams(token));
+
+        if let (Ok(response_user), Ok(response_teams)) = (result_user, result_teams) {
+            if response_teams
+                .teams
+                .iter()
+                .any(|team| team.slug == sso_team)
+            {
+                println!("{}", ui.apply(BOLD.apply_to("Existing token found!")));
+                print_cli_authorized(&response_user.user.email, &ui);
+                return Ok(());
+            }
+        }
+    }
+    let api_client: APIClient = base.api_client()?;
     let login_url_config = base.repo_config()?.login_url().to_string();
 
     // We are passing a closure here, but it would be cleaner if we made a
@@ -21,6 +59,16 @@ pub async fn sso_login(base: &mut CommandBase, sso_team: &str) -> Result<()> {
 pub async fn login(base: &mut CommandBase) -> Result<()> {
     let api_client: APIClient = base.api_client()?;
     let ui = base.ui;
+    // Check if token exists first. Must be there for the user and contain the team
+    // passed into this function.
+    if let Some(token) = base.user_config()?.token() {
+        if let Ok(response) = api_client.get_user(token).await {
+            println!("{}", ui.apply(BOLD.apply_to("Existing token found!")));
+            print_cli_authorized(&response.user.email, &ui);
+            return Ok(());
+        }
+    }
+
     let login_url_config = base.repo_config()?.login_url().to_string();
 
     // We are passing a closure here, but it would be cleaner if we made a


crates/turborepo-lib/src/commands/login.rs
```diff
@@ -19,6 +19,21 @@ const DEFAULT_PORT: u16 = 9789;
 const DEFAULT_SSO_PROVIDER: &str = "SAML/OIDC Single Sign-On";
 
 pub async fn sso_login(base: &mut CommandBase, sso_team: &str) -> Result<()> {
+    if let Some(token) = base.user_config()?.token() {
+        // Try to fetch the user with the token we found to see if we get a good
+        // response back.
+        // TODO: Write response to disk close to / related to token.
+        match base.api_client()?.get_user(token).await {
```
mehulkar, line 26: `sso_login` can't just check `get_user()`, I think it needs to check `get_teams()` and match the passed `sso_team` arg to it to make sure the current token on disk actually has access to that team

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -91,7 +87,47 @@ fn make_token_name() -> Result<String> {
...
+    // If we've already had the user log in, check to see if
+    // the token we found is still valid.
+    //
+    // If the token we found is not valid, go through the process to get
+    // a new token issued.
+    if let Some(token) = user_config.token() {
+        // Try to fetch the user with the token we found to see if we get a good
+        // response back.
+        // TODO: Write response to disk close to / related to token.
+        match base.api_client()?.get_user(token).await {
```
mehulkar, line 120: `get_user` works here, but I'd just use `get_teams()` as with sso_token and pull that out into a `validate_token` shared function. 

One hiccup we'll have down the line is that if the user did `turbo login --sso-team=<xx>` the token will only be scoped to that team. So this check for `get_user` would be successful, but not actually grant access to other non-SSO teams. So `get_teams()` will be good for that as well (if just to print out which teams the user has access to)

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -346,14 +365,23 @@ mod test {
...
         handle.abort();
 
-        assert_eq!(
-            base.user_config().unwrap().token().unwrap(),
-            turborepo_vercel_api_mock::EXPECTED_TOKEN
-        );
+        // Save token we just got. Run the login again to make sure the token we got
+        // doesn't change.
+        assert_eq!(first_token, turborepo_vercel_api_mock::EXPECTED_TOKEN);
+        assert_eq!(first_token, second_token);
```
mehulkar, line 384: Is it possible to assert that we didn't hit the the `/turborepo/token` endpoint (or startup the one_shot_server)? The equality check is somewhat dependent on the server's behavior (which can mostly trust, but not _really_ the thing we need to validate here

mehulkar (changes requested):
Looks clean and readable even to a Rust newb like me, nice job! Comments inline

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -19,6 +19,21 @@ const DEFAULT_PORT: u16 = 9789;
...
+        // response back.
+        // TODO: Write response to disk close to / related to token.
+        match base.api_client()?.get_user(token).await {
+            Ok(response) => {
+                let ui = &base.ui;
+                println!("{}", ui.apply(BOLD.apply_to("Existing token found!")));
+                print_cli_authorized(&response.user.email, ui);
+                return Ok(());
+            }
+            Err(_) => {} // An error here means we got a bad response - assume we need a new token.
```
NicholasLYang, line 33: This is where I'd use an `if..let` as well.

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -19,6 +19,21 @@ const DEFAULT_PORT: u16 = 9789;
...
 
 pub async fn sso_login(base: &mut CommandBase, sso_team: &str) -> Result<()> {
+    if let Some(token) = base.user_config()?.token() {
+        // Try to fetch the user with the token we found to see if we get a good
+        // response back.
+        // TODO: Write response to disk close to / related to token.
+        match base.api_client()?.get_user(token).await {
+            Ok(response) => {
+                let ui = &base.ui;
+                println!("{}", ui.apply(BOLD.apply_to("Existing token found!")));
```
NicholasLYang, line 29: We have a handy macro `cprintln!` that can make this a little simpler:
```suggestion
                cprintln!(ui, BOLD, "Existing token found!");
```

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -19,6 +19,25 @@ const DEFAULT_PORT: u16 = 9789;
 const DEFAULT_SSO_PROVIDER: &str = "SAML/OIDC Single Sign-On";
 
 pub async fn sso_login(base: &mut CommandBase, sso_team: &str) -> Result<()> {
+    if let Some(token) = base.user_config()?.token() {
+        // Try to fetch the user with the token we found to see if we get a good
+        // response back.
+        // TODO: Write response to disk close to / related to token.
+        if let Ok(response) = base.api_client()?.get_user(token).await {
+            let teams_response = base.api_client()?.get_teams(token).await?;
```
mehulkar, line 27: can we do these in parallel?

Zertsov: we can sure try 🔥

crates/turborepo-lib/src/commands/login.rs
Zertsov: I admit defeat for now. We can revisit how we're doing this at a later time.

mehulkar:
@Zertsov I merged #6027 that conflicts with this. Since we're already here, you might as well wait for $6037 also? I can help get main merged into this branch/PR afterwards, sorry for blocking your second PR into this repo!

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -1,12 +1,50 @@
 use anyhow::Result;
 use turborepo_api_client::APIClient;
 use turborepo_auth::{login as auth_login, sso_login as auth_sso_login};
+use turborepo_ui::{BOLD, CYAN, UI};
 
 use crate::commands::CommandBase;
 
+fn print_cli_authorized(user: &str, ui: &UI) {
```
Zertsov, line 8: I added this here even though it's printed by the auth crate because if we exit early we still wanna print out the success message

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -1,12 +1,50 @@
...
+}
+
 pub async fn sso_login(base: &mut CommandBase, sso_team: &str) -> Result<()> {
-    let api_client: APIClient = base.api_client()?;
     let ui = base.ui;
+    // Check if token exists first. Must be there for the user and contain the team
+    // passed into this function.
+    if let Some(token) = base.user_config()?.token() {
+        let client_user: APIClient = base.api_client()?;
+        let client_teams: APIClient = base.api_client()?;
```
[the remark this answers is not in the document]
Zertsov: We can, I thought that this would allow for concurrent requests. If my understanding is wrong then we can just use a single client.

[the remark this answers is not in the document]
NicholasLYang: One client is probably best

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -5,29 +5,39 @@ use turborepo_auth::{login as auth_login, sso_login as auth_sso_login};
...
-    let api_client: APIClient = base.api_client()?;
     let ui = base.ui;
+    let api_client: APIClient = base.api_client()?;
     let login_url_config = base.repo_config()?.login_url().to_string();
+    let token_path = base.repo_config()?.get_path().to_owned();
 
     // We are passing a closure here, but it would be cleaner if we made a
     // turborepo-config crate and imported that into turborepo-auth.
     let set_token = |token: &str| -> Result<(), anyhow::Error> {
         Ok(base.user_config_mut()?.set_token(Some(token.to_string()))?)
```
mehulkar, line 16: Use `token_path` here too (or go all the way and remove this closure and use `token_path` in the crate)

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -5,29 +5,39 @@ use turborepo_auth::{login as auth_login, sso_login as auth_sso_login};
...
 pub async fn login(base: &mut CommandBase) -> Result<()> {
     let api_client: APIClient = base.api_client()?;
     let ui = base.ui;
     let login_url_config = base.repo_config()?.login_url().to_string();
+    let token_path = base.repo_config()?.get_path().to_owned();
 
     // We are passing a closure here, but it would be cleaner if we made a
     // turborepo-config crate and imported that into turborepo-auth.
     let set_token = |token: &str| -> Result<(), anyhow::Error> {
         Ok(base.user_config_mut()?.set_token(Some(token.to_string()))?)
```
mehulkar, line 39: Use `token_path` here too (or go all the way and remove this closure and use `token_path` in the crate)

crates/turborepo-lib/src/commands/login.rs
Zertsov: We talked in a call about this - there will be a follow up PR with cleanup efforts around this.

crates/turborepo-lib/src/commands/login.rs
Zertsov: We talked in a call about this - there will be a follow up PR with cleanup efforts around this.

mehulkar: missed a comment here, but totally nonblocking, you and me are WELL AWARE of this quirk and have talked about it enough 😅

mehulkar (approved):
I didn't read the tests and mock Client super closely, but our walkthrough from yesterday was enough for me. We also talked about the only blocking thing for me (using `token_path`) and I'm good with figuring this out in a follow up!

crates/turborepo-auth/src/lib.rs
```diff
@@ -44,15 +64,27 @@ where
...
 {
+    // Check if token exists first.
+    if let Ok(token) = std::fs::read_to_string(token_path) {
+        if let Ok(response) = api_client.get_user(&token).await {
+            println!("{}", ui.apply(BOLD.apply_to("Existing token found!")));
+            print_cli_authorized(&response.user.email, &ui);
+            return Ok(());
+        }
+    }
+
```
nathanhammond, line 87: Heya @Zertsov! You're reading a JSON file, and then passing the entire thing as the token. This will never work.

crates/turborepo-auth/src/lib.rs
```diff
@@ -177,16 +204,39 @@ struct SsoPayload {
...
+                .iter()
+                .any(|team| team.slug == sso_team)
+            {
+                println!("{}", ui.apply(BOLD.apply_to("Existing token found!")));
+                print_cli_authorized(&response_user.user.email, &ui);
+                return Ok(());
+            }
+        }
+    }
+
```
nathanhammond, line 239: This _also_ doesn't actually correctly load the token.

nathanhammond (commented):
This doesn't appear to actually do what it says on the tin because it fails to parse the inputs and instead treats the entire config file as a string.

We need this to actually attempt to load and parse the config, but as it stands it doesn't do that.

crates/turborepo-auth/src/lib.rs
Zertsov: 🤦

PR #6037, opened 2023-09-27
feat(turborepo-auth): move sso_login implementation to new crate

mehulkar:
This is a follow up of #6027 moving `sso_login` and associated functions
into the `turborepo-auth` crate.

Closes TURBO-1365


crates/turborepo-auth/Cargo.toml
```diff
@@ -5,9 +5,10 @@ edition = "2021"
 license = "MPL-2.0"
 
 [dependencies]
-anyhow.workspace = true
+anyhow = { workspace = true, features = ["backtrace"] }
```
mehulkar, line 8: Made this match how it's specified in `turborepo-lib`

NicholasLYang (approved):
LGTM

crates/turborepo-auth/Cargo.toml
chris-olszewski: Not that it is blocking in anyway, but Cargo resolves features in an additive way so when `turborepo-lib` is built any of it's dependencies will get a version of `anyhow` with the `backtrace` feature. [Feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification)

chris-olszewski (approved):

PR #6041, opened 2023-09-27
fix: Run outline bugs

NicholasLYang:
### Description

Fixes bug with root workspace inclusion logic.

### Testing Instructions

Closes TURBO-1367

### DIFF
--- a/crates/turborepo-lib/src/run/mod.rs
+++ b/crates/turborepo-lib/src/run/mod.rs
@@ -121,7 +121,11 @@ impl<'a> Run<'a> {
 
             if filtered_pkgs.len() != pkg_dep_graph.len() {
                 for target in self.targets() {
-                    let task_name = TaskName::from(target.as_str()).into_root_task();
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/run/mod.rs
```diff
@@ -121,7 +121,8 @@ impl<'a> Run<'a> {
 
             if filtered_pkgs.len() != pkg_dep_graph.len() {
                 for target in self.targets() {
-                    let task_name = TaskName::from(target.as_str());
+                    let task_name = TaskName::from(target.as_str()).into_root_task();
```
chris-olszewski, line 124: Do we need to worry about a user passing in a non-root workspace task here? e.g. `turbo run docs#lint --filter=docs` in a repo with a `//#lint` task definition would end up adding the root workspace to filtered packages and `//#lint` would end up in the task graph.

chris-olszewski (approved):

PR #6043, opened 2023-09-27
feat: hook up process manager

chris-olszewski:
### Description

Hook up the process manager to the task graph visitor and start executing tasks. 

The PR includes some tangentially related changes that only impact our terminal output during a run so it made sense to bundle in this PR.

### Testing Instructions

Run tasks using Rust! 
```
[0 olszewski@chriss-mbp] /tmp/pnpm-test $ turbo_dev lint --experimental-rust-codepath
web:lint: 
web:lint: > web@1.0.0 lint /private/tmp/pnpm-test/apps/web
web:lint: > next lint
web:lint: 
ui:lint: 
ui:lint: > ui@0.0.0 lint /private/tmp/pnpm-test/packages/ui
ui:lint: > eslint .
ui:lint: 
docs:lint: 
docs:lint: > docs@1.0.0 lint /private/tmp/pnpm-test/apps/docs
docs:lint: > next lint
docs:lint: 
docs:lint: ✔ No ESLint warnings or errors
web:lint: ✔ No ESLint warnings or errors
```

You can also attempt running error integration tests specifically the one that is useful is `tests/run/one-script-error.t` (via `EXPERIMENTAL_RUST_CODEPATH=true .cram_env/bin/prysk --shell=bash tests/run/one-script-error.t`) as it tests execution cancelling due to errors and `--continue` usage. Note that the test won't pass as the task cache and run summary isn't hooked up, but modulo those the logs are matching.

Closes TURBO-1369


crates/turborepo-lib/src/process/child.rs
```diff
@@ -177,6 +179,18 @@ impl Child {
...
+        let label = {
+            let cmd = command.as_std();
+            format!(
+                "({}) {} {}",
+                cmd.get_current_dir()
+                    .map(|dir| dir.to_string_lossy())
+                    .unwrap_or_default(),
+                cmd.get_program().to_string_lossy(),
+                cmd.get_args().map(|s| s.to_string_lossy()).join(" ")
+            )
```
chris-olszewski, line 191: This is a [port from Go](https://github.com/vercel/turbo/blob/main/cli/internal/process/child.go#L112) so we can match outputs. Note that Go includes the program in `Args` while Rust does not so we must add it explicitly

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -177,18 +190,119 @@ impl<'a> Visitor<'a> {
...
             tasks.push(tokio::spawn(async move {
+                let task_id = info;
                 let _task_cache = task_cache;
                 let mut prefixed_ui =
-                    Self::prefixed_ui(ui, is_github_actions, &output_client, pretty_prefix);
-                prefixed_ui.output(command);
+                    Self::prefixed_ui(ui, is_github_actions, &output_client, pretty_prefix.clone());
+
+                let mut cmd = Command::new(package_manager.to_string());
+                cmd.args(["run", task_id.task()]);
```
chris-olszewski, line 213: Note that in this PR, properly setting up the environment is not handled. Opting to do this later down the line in it's own PR to make this one more contained.

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -177,18 +190,119 @@ impl<'a> Visitor<'a> {
...
+                    Err(e) => {
+                        error!("unable to pipe outputs from command: {e}");
+                        callback.send(Err(StopExecution)).ok();
+                        manager.stop().await;
+                        return;
+                    }
+                    Ok(None) => {
+                        // TODO: how can this happen? we only update the
+                        // exit status with Some and it is only initialized with
+                        // None. Is it still running?
```
chris-olszewski, line 263: @arlyon Do you know if this can happen and if so what it means? I followed through the code and I don't think this is a possible return value, so I'm not sure how to handle it.

chris-olszewski (commented):
Reviewer's Guide

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -177,18 +190,119 @@ impl<'a> Visitor<'a> {
...
+
+                let mut cmd = Command::new(package_manager.to_string());
+                cmd.args(["run", task_id.task()]);
+                cmd.current_dir(workspace_directory.as_path());
+                cmd.stdout(Stdio::piped());
+                cmd.stderr(Stdio::piped());
+
+                // TODO: avoid passing Duration::MAX and instead allow processes to run without
+                // a timeout
+                let mut process = match manager.spawn(cmd, Duration::MAX) {
```
arlyon, line 220: The duration here actually indicates the time between requesting a graceful shutdown and when it is forcibly killed. There is currently no mechanism for having tasks themselves time out.

crates/turborepo-lib/src/task_graph/visitor.rs
arlyon: Yes, looks like you are correct. There was a time where if a process was killed by a signal we would mark that with a None exit code but since then we now return this ChildExit enum instead so the option is no longer needed. I can open a PR or you can handle it here up to you.

arlyon (approved):
One comment about the graceful stop timeout but not a blocker to merging.

crates/turborepo-lib/src/task_graph/visitor.rs
chris-olszewski: I'm more than happy to handle in a follow up. Just wanted to double check that I wasn't missing some behavior.

PR #6050, opened 2023-09-28
Tracing improvements

sokra:
### Description

* Add tracing for resolve tasks
* add per execution span summary

### Testing Instructions

Closes WEB-1678


crates/turbopack-convert-trace/src/main.rs
```diff
@@ -150,17 +150,24 @@ fn main() {
...
             if let Some(v) = values.get("name") {
                 return format!("{v} ({name})").into();
-            } else {
-                return name.into();
+            }
+        }
+        if matches!(
+            name,
+            "turbo_tasks::resolve_call" | "turbo_tasks::resolve_trait_call"
+        ) {
```
jridgewell, line 168: Nit: Should we just use a `match name {}` block?

jridgewell (approved):

PR #6051, opened 2023-09-28
Performance Improvements (1)

sokra:
### Description

* increase sharding to avoid lock contention
* fix availablity info typo
* add size hints to avoid multiple allocations
* improve AutoMap/Set thresholds
* uses non-zero usize to leverage Option<> optimization

### Testing Instructions

Closes WEB-1679

### DIFF
--- a/crates/turbo-tasks-testing/src/lib.rs
+++ b/crates/turbo-tasks-testing/src/lib.rs
@@ -73,7 +73,7 @@ impl TurboTasksCallApi for VcStorage {
                     .map_err(SharedError::new);
 
                 let mut tasks = this.tasks.lock().unwrap();
-                if let Task::Spawned(event) = replace(&mut tasks[i], Task::Finished(result)) {
+                if let Task::Spawned(event) = replace(&mut tasks[i - 1], Task::Finished(result)) {
                     event.notify(usize::MAX);
                 }
             },
--- a/crates/turbo-tasks-testing/src/lib.rs
+++ b/crates/turbo-tasks-testing/src/lib.rs
@@ -48,7 +48,7 @@ impl TurboTasksCallApi for VcStorage {
         let future = func();
         let i = {
             let mut tasks = self.tasks.lock().unwrap();
-            let i = tasks.len();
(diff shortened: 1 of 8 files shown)


crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -353,7 +353,7 @@ impl EcmascriptModuleAsset {
         availability_info: Value<AvailabilityInfo>,
     ) -> Result<Vc<EcmascriptModuleContent>> {
         let this = self.await?;
-        if *self.analyze().needs_availability_info().await? {
+        let availability_info = if *self.analyze().needs_availability_info().await? {
```
jridgewell, line 356: Uh, what does this bug fix affect? Is this part of the reason for our poor chunking?

crates/turbo-tasks-testing/src/lib.rs
```diff
@@ -48,7 +48,7 @@ impl TurboTasksCallApi for VcStorage {
         let future = func();
         let i = {
             let mut tasks = self.tasks.lock().unwrap();
-            let i = tasks.len();
+            let i = tasks.len() + 1;
```
jridgewell, line 51: Nit: Track `id` and `i` separately. We only need a `TaskId::from(i + 1)` once, instead of a `+ 1` and several `- 1`s

jridgewell (approved):
Does anything actually use `size_hint`?

crates/turbopack-ecmascript/src/lib.rs
sokra: It only affects caching. It idea was to not pass availability_info when it's not needed to leverage caching between mutliple calls (with different availability_info). The bug only makes caching worse, especially for client components (which have different availability_info each).

sokra:
> Does anything actually use `size_hint`?

I guess `collect` will use it for the initial capacity. Without it it starts with an empty buffer and grows it on every element by factor 2. So a correct `size_hint` will only cause one allocation instead of `log N` ones.

PR #6052, opened 2023-09-28
feat(turbo-ignore) add a --max-buffer option for dry runs that have large outputs

mehulkar:
Closes TURBO-1371

### DIFF
--- a/packages/turbo-ignore/src/cli.ts
+++ b/packages/turbo-ignore/src/cli.ts
@@ -30,6 +30,12 @@ turboIgnoreCli
       "The directory to run in (default: cwd)"
     )
   )
+  .addOption(
+    new Option(
+      "-b, --max-buffer <number>",
+      "maxBuffer for the child process in KB (default: 1024 * 1024)"
+    ).argParser((val) => parseInt(val, 10))
+  )
   .version(cliPkg.version, "-v, --version", "Output the current version")
   .helpOption("-h, --help", "Display help for command")
   .showHelpAfterError(false)
--- a/packages/turbo-ignore/src/cli.ts
+++ b/packages/turbo-ignore/src/cli.ts
@@ -33,8 +33,8 @@ turboIgnoreCli
   .addOption(
     new Option(
       "-b, --max-buffer <number>",
-      "maxBuffer for the child process in KB (default: 1024 * 1024)"
-    ).argParser((val) => parseInt(val, 10))
+      "maxBuffer for the child process in KB (default: 1024 KB)"
+    ).argParser((val) => parseInt(val, 10) * 1024)
   )
   .version(cliPkg.version, "-v, --version", "Output the current version")
   .helpOption("-h, --help", "Display help for command")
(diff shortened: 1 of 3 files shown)


packages/turbo-ignore/src/ignore.ts
```diff
@@ -95,50 +95,54 @@ export function turboIgnore(
...
+
+  const execOptions: { cwd: string; maxBuffer?: number } = {
+    cwd: root,
+  };
+
+  if (opts.maxBuffer) {
+    execOptions.maxBuffer = opts.maxBuffer;
+  }
+
+  info(`execOptions: ${JSON.stringify(execOptions, null, 2)}`);
```
chris-olszewski, line 107: I don't think we want to be logging this to all users

packages/turbo-ignore/src/cli.ts
```diff
@@ -30,6 +30,12 @@ turboIgnoreCli
       "The directory to run in (default: cwd)"
     )
   )
+  .addOption(
+    new Option(
+      "-b, --max-buffer <number>",
+      "maxBuffer for the child process in KB (default: 1024 * 1024)"
+    ).argParser((val) => parseInt(val, 10))
```
chris-olszewski, line 37: If I'm reading the docs correctly, I'm pretty sure this is in bytes:

>maxBuffer [<number>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Number_type) Largest amount of data in bytes allowed on stdout or stderr. If exceeded, the child process is terminated and any output is truncated. See caveat at [maxBuffer and Unicode](https://nodejs.org/api/child_process.html#maxbuffer-and-unicode). Default: 1024 * 1024

[source](https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback)

```suggestion
      "maxBuffer for the child process in KB (default: 1024 KB)"
    ).argParser((val) => parseInt(val, 10) * 1024)
```

packages/turbo-ignore/src/types.ts
```diff
@@ -23,4 +23,6 @@ export interface TurboIgnoreOptions {
   task?: string;
   // A ref/head to compare against if no previously deployed SHA is available
   fallback?: string;
+  // The maxBuffer for the child process in KB
```
chris-olszewski, line 26:
```suggestion
  // The maxBuffer for the child process in bytes
```

chris-olszewski (approved):
I think it makes sense to give users the ability to control the number of bytes in either KB or MB. I think MB is slightly nicer, but I realize for some it won't be granular enough.

packages/turbo-ignore/src/types.ts
```diff
@@ -23,4 +23,6 @@ export interface TurboIgnoreOptions {
   task?: string;
   // A ref/head to compare against if no previously deployed SHA is available
   fallback?: string;
+  // The maxBuffer for the child process in bytes
```
mehulkar, line 26:
```suggestion
  // The maxBuffer for the child process in KB
```

PR #6053, opened 2023-09-28
Performance Improvements (2)

sokra:
### Description

Use SmallVec to inline some elements into the field/stack

### Testing Instructions

Closes WEB-1680

### DIFF
--- a/crates/turbo-tasks-memory/src/aggregation_tree/bottom_tree.rs
+++ b/crates/turbo-tasks-memory/src/aggregation_tree/bottom_tree.rs
@@ -3,6 +3,7 @@ use std::{hash::Hash, ops::ControlFlow, sync::Arc};
 use nohash_hasher::{BuildNoHashHasher, IsEnabled};
 use parking_lot::{Mutex, MutexGuard};
 use ref_cast::RefCast;
+use smallvec::SmallVec;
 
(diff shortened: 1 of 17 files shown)


crates/turbo-tasks-memory/src/aggregation_tree/bottom_tree.rs
```diff
@@ -113,7 +114,7 @@ impl<T, I: Clone + Eq + Hash + IsEnabled> BottomTree<T, I> {
     ) where
         I: 'a,
     {
-        let mut following = Vec::new();
+        let mut following = SmallVec::default();
```
jridgewell, line 117: Does this require a turbofish cast?

jridgewell (approved):
We're now specializing the template implementations on the `const I: usize`, which I think can balloon our built byte size if we use different sizes with otherwise similar types. Should we be extracting reusable types that specify the size?

eg, `type Children<I> = SmallVec<[&I; 16]>` in `crates/turbo-tasks-memory/src/aggregation_tree/bottom_tree.rs` or `type Tasks = AutoSet<TaskId, BuildNoHashHasher<TaskId>, 2>` in `crates/turbo-tasks-memory/src/cell.rs`?

PR #6069, opened 2023-09-29
chore(ci): Use better named outputs for triggering turborepo_build job

mehulkar:
- Changes the name of the output that is used to trigger a build of turborepo. Both Rust and Go changes trigger this output to be set to true, so naming the output `outputs.go` didn't work
- Do not set `outputs.go` when Rust files change. After this PR, only Go linting and Unit tests rely on this output, so it is appropriate for them to only run when Go files have changed.

Closes TURBO-1380

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -183,7 +183,8 @@ jobs:
       turbopack_typescript: ${{ steps.ci.outputs.diff != '' || steps.turbopack_typescript.outputs.diff != '' }}
       turborepo_rust: ${{ steps.ci.outputs.diff != '' || steps.turborepo_rust.outputs.diff != '' }}
       turbopack_bench: ${{ steps.ci.outputs.diff != '' || steps.turbopack_bench.outputs.diff != '' }}
-      go: ${{ steps.ci.outputs.diff != '' || steps.turborepo_go.outputs.diff != '' || steps.turborepo_rust.outputs.diff != '' }}
+      go: ${{ steps.ci.outputs.diff != '' || steps.turborepo_go.outputs.diff != ''}}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -183,7 +183,8 @@ jobs:
       turbopack_typescript: ${{ steps.ci.outputs.diff != '' || steps.turbopack_typescript.outputs.diff != '' }}
       turborepo_rust: ${{ steps.ci.outputs.diff != '' || steps.turborepo_rust.outputs.diff != '' }}
       turbopack_bench: ${{ steps.ci.outputs.diff != '' || steps.turbopack_bench.outputs.diff != '' }}
-      go: ${{ steps.ci.outputs.diff != '' || steps.turborepo_go.outputs.diff != '' || steps.turborepo_rust.outputs.diff != '' }}
+      go: ${{ steps.ci.outputs.diff != '' || steps.turborepo_go.outputs.diff != ''}}
```
mehulkar, line 186: `go_lint` and `go_unit` run conditionally based on this output, so we don't need `turborepo_rust` changes to be included

chris-olszewski: I think we need to still run `go_unit` if either `go` or `turborepo_rust` is marked as change since Go depends on Rust code via FFI now and we have unit tests that exercise this.

chris-olszewski (changes requested):
I think the change to `go_unit` is the only blocking thing

mehulkar:
@chris-olszewski updated

.github/workflows/test.yml
mehulkar: fixed good call

chris-olszewski (approved):

mehulkar:
examples-test CI job failed because of pnpm-gatsby example. It doesn't seem related to this test

PR #6071, opened 2023-09-29
feat: add task cache save

chris-olszewski:
### Description

Add Rust port of `TaskCache.SaveOutputs`. 

As part of this I did do a slight rework of `AsyncCache` to avoid the need to mutate it on writing to the cache. An alternative would be to throw a `Mutex` around the reference to the `RunCache`, but that would require holding a guard across awaits [which tokio recommends against](https://tokio.rs/tokio/tutorial/shared-state#holding-a-mutexguard-across-an-await).

### Testing Instructions

For the refactors the existing unit tests. For the task cache method, I'm unsure of the best way to test it other than using it the task visitor. @NicholasLYang do you have any suggestions for testing the task cache?

Closes TURBO-1382

### DIFF
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -102,8 +102,6 @@ impl RunCache {
             ui: self.ui,
         }
     }
-
-    // pub fn notify
 }
 
 pub struct TaskCache {
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -1,4 +1,4 @@
-use std::{io::Write, sync::Arc};
+use std::{io::Write, sync::Arc, time::Duration};
 
 use console::StyledObject;
 use tracing::{debug, log::warn};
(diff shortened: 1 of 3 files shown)


NicholasLYang:
Unfortunately I haven't found a good way to test the task cache. It'll probably fall under e2e or integration tests once the entire pipeline is hooked up. Also, I really want to test whether the async cache is actually worth keeping. It could very well be that tokio's underlying primitives are good enough

crates/turborepo-lib/src/run/cache.rs
```diff
@@ -96,6 +102,8 @@ impl RunCache {
             ui: self.ui,
         }
     }
+
+    // pub fn notify
```
NicholasLYang, line 106:
```suggestion
```

crates/turborepo-lib/src/run/cache.rs
```diff
@@ -281,4 +289,65 @@ impl TaskCache {
...
+            Some(daemon_client) => daemon_client
+                .notify_outputs_written(
+                    self.hash.to_string(),
+                    self.repo_relative_globs.inclusions.clone(),
+                    self.repo_relative_globs.exclusions.clone(),
+                    duration.as_millis() as u64,
+                )
+                .await
+                .map_err(Error::from),
+            None => Err(Error::NoDaemon),
```
NicholasLYang, line 338: I think in the Go implementation we don't error if the daemon doesn't exist. We just return nil. However, this does seem to be the more sensible approach

NicholasLYang (approved):

PR #6073, opened 2023-10-01
fix(Turborepo): copy yarn berry config files when running `turbo prune`

### LINKED ISSUE #6063: [turborepo] `turbo prune --scope=<app>` doesn't include `.yarnrc.yml` in `out/` or `out/json` folder

1.10.14

Mac

### Describe the Bug

The `.yarnrc.yml` file and the `.yarn/releases`, `.yarn/patches`, `.yarn/plugins`, `.yarn/releases`, `.yarn/sdks` and `.yarn/versions` folders are not included in the out/` or the `out/json` folder when running `turbo prune --scope=<app>`.

So if you use the out folder in a standalone way (like a docker image), running `yarn install` inside the output folder will not have the same behaviour as running it in your monorepo root folder. You will even lose the `nodeLinker: node-modules` configuration if you set it in the `.yarnrc.yml` config file.

### Expected Behavior

The `.yarnrc.yml` file and the `.yarn/releases`, `.yarn/patches`, `.yarn/plugins`, `.yarn/releases`, `.yarn/sdks` and `.yarn/versions` folders are included in the out/` or the `out/json` folder when running `turbo prune --scope=<app>`.

### To Reproduce

We can use the turborepo folder example to reproduce the issue, run the following using `yarn` `v1.22.19`:

```sh
git clone git@github.com:vercel/turbo.git
cd turbo/examples/with-yarn
echo 'nodeLinker: node-modules' > .yarnrc.yml
yarn policies set-version 3.6.3 
yarn install
turbo prune --scope=web --out-dir=/some/folder/outside/monorepo/to/test/it/standalone
cd /some/folder/outside/monorepo/to/test/it/standalone
yarn install
```

You will see that the first `yarn install` was run with the selected version (3.6.3), and the second one with 1.22.19. 

---

We can try the same with the same version of Yarn, but we will have another problem. Run the following using `yarn` `v3.6.3`:

```sh
git clone git@github.com:vercel/turbo.git
cd turbo/examples/with-yarn
echo 'nodeLinker: node-modules' > .yarnrc.yml
yarn policies set-version 3.6.3 
yarn install
turbo prune --scope=web --out-dir=/some/folder/outside/monorepo/to/test/it/standalone
cd /some/folder/outside/monorepo/to/test/it/standalone
yarn install
```

You will see that the first `yarn install` was run with the `node-modules` node linker options (which is compatible with turbo), and the second one with the default one (pnp) which is not.

---

In both cases, I expect both `yarn install` to behave the same.

### Reproduction Repo

_No response_

<sub>[TURBO-1386](https://linear.app/vercel/issue/TURBO-1386/[turborepo]-turbo-prune-scope=app-doesnt-include-yarnrcyml-in-out-or)</sub>

artberri:
### Description

Fixes #6063 

- Copy `.yarnrc.yml` file when present during turbo prune to out and out/full w/docker
- Copy `.yarn/releases`, `.yarn/patches`, `.yarn/plugins`, `.yarn/releases`, `.yarn/sdks` and `.yarn/versions` folders when present during turbo prune to out and out/full w/docker.
- Add a new example for Yarn v2+ (berry)

These files and folders are required in Yarn V2+ with the `node-modules` node linker config to reproduce idempotent installations and should be included in the output folder. For more info about those folders and files:
https://yarnpkg.com/configuration/yarnrc
https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

### Testing Instructions

```sh
cd cli
make turbo
cd ../examples/with-berry
yarn install
turbo prune web # or turbo prune web --docker
```


crates/turborepo-lib/src/commands/prune.rs
```diff
@@ -52,6 +52,32 @@ lazy_static! {
...
+        (
+            RelativeUnixPath::new(".yarn/plugins").unwrap(),
+            Some(CopyDestination::Docker)
+        ),
+        (
+            RelativeUnixPath::new(".yarn/releases").unwrap(),
+            Some(CopyDestination::Docker)
+        ),
+        (
+            RelativeUnixPath::new(".yarn/sdks").unwrap(),
```
chris-olszewski, line 74: Reading the description of what is in `.yarn/sdks` I'm not sure if most Docker images want these files.

crates/turborepo-lib/src/commands/prune.rs
```diff
@@ -52,6 +52,32 @@ lazy_static! {
...
+        (
+            RelativeUnixPath::new(".yarnrc.yml").unwrap(),
+            Some(CopyDestination::Docker)
+        ),
+    ];
+    static ref ADDITIONAL_DIRECTORIES: Vec<(&'static RelativeUnixPath, Option<CopyDestination>)> = vec![
+        (
+            RelativeUnixPath::new(".yarn/patches").unwrap(),
+            Some(CopyDestination::Docker)
+        ),
```
chris-olszewski, line 64: We already copy the patches on line 196.
```suggestion
```

chris-olszewski (changes requested):
Thanks for the PR! I think including the `.yarnrc.yml` is a good fix along with `.yarn/releases` since that can get referenced from `.yarnrc.yml`, but I'm not convinced on the rest of the changes are desired. 

We don't want prune to copy files that will be unused on the Docker image and `.yarn/plugins`, `.yarn/sdks`, and `.yarn/versions` all seem like for most projects they wouldn't be necessary in order to run a server. If you have an example where these are necessary, I think it would be acceptable to hide this behavior behind a flag. Maybe `--include-full-config` or something along those lines.

Is that okay with you?

weyert:
@artberri  Isn't the prune command already doing that?

chris-olszewski (approved):
Your explanation makes sense. Again, thank you for the PR and the properly set up berry example. I'll try to get this merged in for the next patch release.

weyert:
Would this file to be the right place to add `pnpm` files too? E.g. `pnpm-workspace.yaml` and `.pnpmfile.cjs`?

chris-olszewski:
> Would this file to be the right place to add `pnpm` files too? E.g. `pnpm-workspace.yaml` and `.pnpmfile.cjs`?

Yes to `.pnpmfile.cjs`, we actually already copy `pnpm-workspace.yaml` [here](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/commands/prune.rs#L94)

PR #6074, opened 2023-10-01
feat(Turborepo): Repository JS API

gsoltis:
### Description

 - Adds a JS Package, `@turbo/repository`, that wraps some of Turborepo's repository analysis functionality
 - Initial feature set:
   - Detect repository root
   - Detect monorepo vs single-package repo
   - Detect package manager (name only, no specific version yet)
   - Enumerate repo-relative paths to workspaces
 - Memoizes detected package manager (or error) in `RepoState`. Not used in core Turborepo yet, just `@turbo/repository`
 - GH Actions and package infrastructure to build npm packages. 
   - Mirrors a lot of similar infrastructure from Next.js / `next-swc`
   - Does not yet actually publish, currently packages and uploads as an output of the workflow. Once this PR is merged, I'll submit another to flip from `npm pack` to `npm publish`
   - Does not yet tackle automated versioning

Note that this is a new package, at `0.0.1-canary.0`. This does not need to be v1.0 names or API, this is primarily to get the pipeline in place and allow a space for iteration.

### Testing Instructions

Tested via manual linking + embedding into Spaces Repository Index CLI and comparing results

Closes TURBO-1384


mehulkar (commented):
This is awesome. It's very close to what I had imagined. The main difference is that you have `crates/turborepo-napi` as a middle layer to export _all_ the JS code. I imagine `@turborepo/repository` will depend on `turborepo-napi` then? Or were you thinking of publishing `turborepo-napi` in all its glory?

Personally, I would prefer:
- Rust "shim" that imports `crates/*` to live in `packages/`
- Smaller/focused JS packages like `@turborepo/auth` and `@turborepo/repository` (which may or may not end up importing the same core Rust crates)

For example:

```
.
├── crates/turborepo-core-lib-1
├── crates/turborepo-core-lib-1
├── packages/turborepo-repository
│   ├── rust-imports
│   ├── src/index.js
│   ├── package.json
├── packages/turborepo-auth
│   ├── rust-imports
│   ├── src/index.js
│   ├── package.json
```

gsoltis:
@mehulkar IIUC I have taken your suggestion re: organization. There is now a small rust wrapper crate scoped just to repository analysis that depends on pieces of core Turborepo functionality. It is located alongside the js package that handles wrapping the native library.

.github/workflows/turborepo-library-release.yml
```diff
@@ -0,0 +1,113 @@
...
+          enable-corepack: false
+
+      - name: Setup toolchain
+        run: ${{ matrix.settings.setup }}
+        if: ${{ matrix.settings.setup }}
+
+      - name: Build native library
+        run: |
+          cd packages/turbo-repository
+          mkdir -p js/dist
```
mehulkar, line 59: shouldn't the build script create the build dir?

Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = [
   "crates/turbo-tasks*",
   "crates/turbopack*",
   "crates/turborepo*",
+  "packages/turbo-repository/rust/",
```
mehulkar, line 9: Any benefit in including this here? (I'm still a bit fuzzy on what this root Cargo.toml does and what's expected to be in here)

packages/turbo-repository/README.md
```diff
@@ -0,0 +1,12 @@
+## Repository Analysis JS Package
```
mehulkar, line 1:
```suggestion
# `@turbo/repository`
```

packages/turbo-repository/js/README.md
```diff
@@ -0,0 +1,10 @@
+## Repositories
+
+Provides JS repository analysis functionality, scoped to the JS ecosystem.
```
mehulkar, line 3:
```suggestion
Provides repository analysis functionality, scoped to the web ecosystem.
```

packages/turbo-repository/npm/darwin-arm64/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@turbo/repository-darwin-arm64",
```
mehulkar, line 2: how does these platform-speciifc packages get built and published for `turbo`? Could we eventually match that and not keep each of these in source control?

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+            name: package_manager.to_string(),
+            package_manager,
+        }
+    }
+}
+
+#[napi]
+impl Repository {
+    #[napi(factory, js_name = "detectJS")]
+    pub fn detect_js(path: Option<String>) -> Result<Self> {
```
mehulkar, line 35: I'd prefer to anchor on "web" rather than "JS" as our ecosystem (e.g. a Next.js app isn't a JS app, it's a _web_ app)

packages/turbo-repository/scripts/publish.mjs
```diff
@@ -0,0 +1,67 @@
...
+import { execa } from "execa";
+import { copy } from "fs-extra";
+import { readFile, writeFile } from "fs/promises";
+
+const cwd = process.cwd();
+let platform = process.argv[process.argv.length - 1];
+
+if (platform.endsWith("publish.mjs")) {
+  // No platform was passed, default to dev platform
+  platform = "darwin-arm64";
```
mehulkar, line 13:
```suggestion
  platform = `${process.platform}-${process.arch}`
```

not sure if this works across the board

packages/turbo-repository/js/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+Note that this is not yet a stable package, and functionality, API, and naming may all change.
```
mehulkar, line 5:
```suggestion
```

The version already says this, don't want to remember keep this note in sync with the version

packages/turbo-repository/rust/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Build:
+
+Build native library and js wrapper
+
+```
+pnpm exec napi build --platform
+```
+
+## Test
```
mehulkar, line 17: Do you want to add a simple test case here instead of manual tests? (hint hint this seems like a pretty ideal place to use the node test runner 😸 )

mehulkar (approved):
Left some comments in line, nothing too blocking. I like the setup of `packages/turbo-repository`, nice work!

How will we lint / test the Rust code in `packages/turbo-*/rust`? Does cargo automatically pick it up since it's in the root Cargo.toml?

packages/turbo-repository/js/README.md
gsoltis: I don't think this is correct. `turbo` today works with the javascript ecosystem, whether or not that javascript is building something for the web. CLIs, backend services, etc. are all fair game. The scope is really anything that is driven via `package.json`.

Cargo.toml
gsoltis: Yes, this is the equivalent of adding a workspaces entry but for Rust. Among other things, it allows the `turbo-repository` crate to depend on crates defined under `crates/`

packages/turbo-repository/npm/darwin-arm64/package.json
gsoltis: We could standardize at some point. We have a [script](https://github.com/vercel/turbo/blob/main/cli/scripts/npm-native-packages/npm-native-packages.js) for `turbo`, the method here is copied from [next-swc](https://github.com/vercel/next.js/tree/canary/packages/next-swc/crates/napi/npm) in the Next.js repo.

packages/turbo-repository/rust/src/lib.rs
gsoltis: Mentioned elsewhere, but I don't think that's correct. Our ecosystem is not the (only) web, but anything JS, including backend services and CLIs.

packages/turbo-repository/scripts/publish.mjs
gsoltis: It will not work across the board, linux in particular requires a libc specifier. However, for now at least, this is only used for local dev.

packages/turbo-repository/js/README.md
gsoltis: I think we'll need to do a significant revision of the README anyway when we commit to a more stable API. I'd like to keep this as a more explicit indication than just the version.

packages/turbo-repository/rust/README.md
gsoltis: At some point, yes. I'd like to end up with some tests to verify many of the example repos we already have.

.github/workflows/turborepo-library-release.yml
gsoltis: Yeah, the build script does create this folder now. I'll delete this line from the GH workflow.

packages/turbo-repository/README.md
gsoltis: Mentioned elsewhere, but the README is going to go through a revision when we're ready to commit a bit more to this.

gsoltis:
Merging, since I don't think this affects Turbopack tests

PR #6076, opened 2023-10-02
Performance Improvements (5)

sokra:
### Description

The `AsyncModule` code generation in every ESM flags it as "needs availability info" which causes code generation to be duplicated for each entrypoints.

This improves it a little bit by checking is the module is an async module before. It also fixes a bug where availability is not correct reduced and only passes the availability root to async module computation instead of the full availability info.

### Testing Instructions

Closes WEB-1697


crates/turbopack-ecmascript/src/references/async_module.rs
```diff
@@ -240,16 +248,34 @@ impl CodeGenerateableWithAvailabilityInfo for AsyncModule {
...
+                    add_async_dependency_handler(program, &async_idents);
+                }));
+            }
         }
 
         Ok(CodeGeneration { visitors }.into())
     }
+
+    #[turbo_tasks::function]
+    async fn get_availability_info_needs(
```
jridgewell, line 269: Doc comment please.

crates/turbopack-ecmascript/src/code_gen.rs
```diff
@@ -39,6 +39,13 @@ pub trait CodeGenerateableWithAvailabilityInfo {
         chunking_context: Vc<Box<dyn EcmascriptChunkingContext>>,
         availability_info: Value<AvailabilityInfo>,
     ) -> Vc<CodeGeneration>;
+
+    fn get_availability_info_needs(
```
jridgewell, line 43: Docs

crates/turbopack-ecmascript/src/chunk/item.rs
```diff
@@ -281,17 +281,20 @@ impl FromChunkableModule for Box<dyn EcmascriptChunkItem> {
...
+            AvailabilityInfo::Complete {
                 available_modules,
                 current_availability_root,
-            } => AvailabilityInfo::Inner {
+            } => AvailabilityInfo::Complete {
                 available_modules: available_modules.with_roots(vec![current_availability_root]),
                 current_availability_root: Vc::upcast(module),
             },
+            AvailabilityInfo::OnlyAvailableModules { .. } => {
+                panic!("OnlyAvailableModules should not appear here")
```
jridgewell, line 296:
```suggestion
                unreachable!("OnlyAvailableModules should not appear here")
```

crates/turbopack-core/src/chunk/availability_info.rs
```diff
@@ -23,20 +60,58 @@ impl AvailabilityInfo {
...
+            Self::Complete {
                 available_modules, ..
             } => Some(*available_modules),
+            Self::OnlyAvailableModules {
+                available_modules, ..
+            } => Some(*available_modules),
+        }
+    }
+
+    pub fn reduce_to_needs(self, needs: AvailabilityInfoNeeds) -> Self {
```
jridgewell, line 84: Docs

crates/turbopack-core/src/chunk/availability_info.rs
```diff
@@ -1,19 +1,56 @@
+use std::ops::{Add, AddAssign};
+
 use turbo_tasks::Vc;
 
 use super::available_modules::AvailableAssets;
 use crate::module::Module;
 
+#[turbo_tasks::value(shared)]
+#[derive(Copy, Clone, Debug)]
+pub enum AvailabilityInfoNeeds {
```
jridgewell, line 10: Docs

crates/turbopack-core/src/chunk/availability_info.rs
```diff
@@ -1,19 +1,56 @@
...
+
+impl AddAssign for AvailabilityInfoNeeds {
+    fn add_assign(&mut self, rhs: Self) {
+        *self = *self + rhs;
+    }
+}
+
 #[turbo_tasks::value(serialization = "auto_for_input")]
 #[derive(PartialOrd, Ord, Hash, Clone, Copy, Debug)]
 pub enum AvailabilityInfo {
```
jridgewell, line 42: Docs. What is the difference between `Complete` and `OnlyAvailableModules`?

jridgewell (approved):

crates/turbopack-ecmascript/src/chunk/item.rs
ForsakenHarmony: I don't think it's unreachable

crates/turbo-tasks/src/manager.rs
```diff
@@ -410,7 +410,7 @@ impl<B: Backend + 'static> TurboTasks<B> {
     /// Calls a native function with arguments. Resolves arguments when needed
     /// with a wrapper [Task].
     pub fn dynamic_call(&self, func: FunctionId, inputs: Vec<ConcreteTaskInput>) -> RawVc {
-        if inputs.iter().all(|i| i.is_resolved() && !i.is_nothing()) {
+        if inputs.iter().all(|i| i.is_resolved()) {
```
ForsakenHarmony, line 413: is this an unrelated bug?

crates/turbopack-core/src/chunk/availability_info.rs
ForsakenHarmony, line 10: Wouldn't this be better off as a struct with 2 boolean fields?

And then implement `BitOr` instead of `Add`

crates/turbopack-ecmascript/src/references/async_module.rs
```diff
@@ -124,7 +127,12 @@ impl AsyncModuleScc {
         for scc in &*this.scope.get_scc_children(this.scc).await? {
             // Because we generated SCCs there can be no loops in the children, so calling
             // recursively is fine.
-            if *AsyncModuleScc::new(*scc, this.scope).is_async().await? {
+            if *AsyncModuleScc::new(*scc, this.scope)
+                .resolve()
```
ForsakenHarmony, line 131: maybe we should start documenting why a `.resolve()` is necessary

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -507,13 +502,20 @@ impl EcmascriptChunkItem for ModuleChunkItem {
...
-        let content = this
-            .module
-            .module_content(this.chunking_context, availability_info);
         let async_module_options = this
             .module
             .get_async_module()
-            .module_options(availability_info);
+            .module_options(availability_info.current_availability_root());
+        let is_async_module = async_module_options.await?.is_some();
+        let availability_info_needs = *this
```
ForsakenHarmony, line 510: maybe we should document why we do this

ForsakenHarmony (changes requested):

crates/turbo-tasks/src/manager.rs
sokra: Not a bug, just unnecessary. Resolving `Nothing` leads to `Nothing`. We don't have to create a resolve task to do that...

crates/turbopack-core/src/chunk/availability_info.rs
sokra: yeah that makes sense...

crates/turbopack-core/src/chunk/availability_info.rs
sokra: `Complete` also contains the root module.

crates/turbopack-ecmascript/src/references/async_module.rs
sokra: Technically it's not necessary, but it will avoid the resolve task for `is_async` (`is_async` is called with unresolved `self` argument)

crates/turbopack-core/src/chunk/availability_info.rs
```diff
@@ -1,19 +1,89 @@
...
+            (Self::None, rhs) => rhs,
+            (lhs, Self::None) => lhs,
+            (Self::Complete, _) | (_, Self::Complete) => Self::Complete,
+            (Self::Root, Self::Root) => Self::Root,
+            (Self::AvailableModules, Self::AvailableModules) => Self::AvailableModules,
+            (Self::Root, Self::AvailableModules) | (Self::AvailableModules, Self::Root) => {
+                Self::Complete
+            }
+        }
+    }
```
jridgewell, line 46: Did you forget to delete this?

sokra: hmm... seems like I also forgot to compile at all...

crates/turbopack-ecmascript/src/references/async_module.rs
sokra: add added a comment

PR #6082, opened 2023-10-02
chore(ci): Only run examples CI on examples changes or version bump

NicholasLYang:
### Description

We currently run examples on all code changes, which is not necessary since examples run on the released turbo, not the current code. This PR changes the CI to only run when the examples code is changed, when CI changes, or when a new version of turborepo is released.

### Testing Instructions

Closes TURBO-1396

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,6 +68,13 @@ jobs:
             Cargo.*
             rust-toolchain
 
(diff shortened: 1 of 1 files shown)


mehulkar (approved):
The version one probably won’t help because it gets bumped after the release. But seems fine.

PR #6086, opened 2023-10-03
fix(turborepo): Process single-package independently.

nathanhammond:
#6021 Removed the global `--single-package` flag, which was _backwards_-compatible. But for forward-compatibility we need to make sure that we don't try to parse it as an argument unless we're in `run`.

Closes TURBO-1401

### DIFF
--- a/crates/turborepo-lib/src/cli.rs
+++ b/crates/turborepo-lib/src/cli.rs
@@ -205,13 +205,11 @@ impl Args {
     pub fn new() -> Result<Self> {
         // We always pass --single-package in from the shim.
         // We need to omit it, and then add it in for run.
-        let arg_separator_position = std::env::args_os()
-            .find_position(|input_token| input_token == "--")
-            .map(|(index, _)| index);
+        let arg_separator_position =
+            std::env::args_os().position(|input_token| input_token == "--");
 
-        let single_package_position = std::env::args_os()
-            .find_position(|input_token| input_token == "--single-package")
-            .map(|(index, _)| index);
+        let single_package_position =
+            std::env::args_os().position(|input_token| input_token == "--single-package");
 
         let is_single_package = match (arg_separator_position, single_package_position) {
             (_, None) => false,
--- a/crates/turborepo-lib/src/cli.rs
+++ b/crates/turborepo-lib/src/cli.rs
@@ -4,6 +4,7 @@ use anyhow::{anyhow, Result};
 use camino::Utf8PathBuf;
 use clap::{ArgAction, CommandFactory, Parser, Subcommand, ValueEnum};
 use clap_complete::{generate, Shell};
+use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use tracing::{debug, error};
 use turbopath::AbsoluteSystemPathBuf;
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/cli.rs
```diff
@@ -202,8 +203,44 @@ pub enum LinkTarget {
 
 impl Args {
     pub fn new() -> Result<Self> {
-        let mut clap_args = match Args::try_parse() {
-            Ok(args) => args,
+        // We always pass --single-package in from the shim.
+        // We need to omit it, and then add it in for run.
+        let arg_separator_position = std::env::args_os()
+            .find_position(|input_token| input_token == "--")
+            .map(|(index, _)| index);
```
NicholasLYang, line 210:
```suggestion
            .position(|input_token| input_token == "--");
```

crates/turborepo-lib/src/cli.rs
```diff
@@ -202,8 +203,44 @@ pub enum LinkTarget {
...
+
+        let single_package_position = std::env::args_os()
+            .find_position(|input_token| input_token == "--single-package")
+            .map(|(index, _)| index);
```
NicholasLYang, line 214:
```suggestion
            .position(|input_token| input_token == "--single-package");
```

NicholasLYang (approved):

crates/turborepo-lib/src/cli.rs
```diff
@@ -202,8 +203,44 @@ pub enum LinkTarget {
...
+                arg_separator_position
+                    .is_some_and(|arg_separator_position| index > &arg_separator_position)
+                    || input_token != "--single-package"
+            })
+            .map(|(_, input_token)| input_token);
+
+        let mut clap_args = match Args::try_parse_from(single_package_free) {
+            Ok(mut args) => {
+                // And then only add them back in when we're in `run`.
+                if let Some(ref mut run_args) = args.run_args {
```
NicholasLYang, line 238: Should we only add it if we've detected a single package flag?

NicholasLYang (changes requested):

crates/turborepo-lib/src/cli.rs
nathanhammond: That's what the next line does:
```
run_args.single_package = is_single_package
```

This is a computed value, based upon the presence of the flag.

https://github.com/vercel/turbo/pull/6086/files#diff-ff97a6d3bfd4f86fbdd5a7ef844ba82b0371fdd6e016919429b4ba63e725f7a3R216-R222

And, since we're looking for `Some(run_args)`, it ends up applied only to the `run` command.

crates/turborepo-lib/src/cli.rs
```diff
@@ -202,8 +203,44 @@ pub enum LinkTarget {
...
+                    run_args.single_package = is_single_package
+                }
```
nathanhammond, line 240: @NicholasLYang are you requesting this?

```suggestion
                if is_single_package {
                    if let Some(ref mut run_args) = args.run_args {
                        run_args.single_package = true
                    }
                }
```

It's not materially different from the existing code.

crates/turborepo-lib/src/cli.rs
nathanhammond: TIL: position

crates/turborepo-lib/src/cli.rs
NicholasLYang: Ah gotcha, nvm then

PR #6087, opened 2023-10-03
chore: Added logging for filter code

NicholasLYang:
### Description

Added more logging to filter code so we can debug it easier.

### Testing Instructions

Closes TURBO-1402


cli/internal/scope/filter/filter.go
```diff
@@ -110,6 +113,7 @@ func (r *Resolver) getFilteredPackages(selectors []*TargetSelector) (*SelectedPa
 			allPackageSelectors = append(allPackageSelectors, selector)
 		}
 	}
+	r.Logger.Debug("Filtering packages", "prodPackageSelectors", prodPackageSelectors, "allPackageSelectors", allPackageSelectors)
```
chris-olszewski, line 116: Nit, but the whole `prodPackageSelectors` controlled by `followProdDepsOnly` is dead code and will always be empty. (The behavior was ported from pnpm, but never actually hooked up)

chris-olszewski (approved):

cli/internal/scope/filter/filter.go
NicholasLYang: Yeah tbh I was wondering about that code

arlyon (approved):

PR #6090, opened 2023-10-03
fix: don't execute tasks when calculating task hashes

chris-olszewski:
### Description

Issues with spawning tasks were causing task hashes to diverge. This is an issue that should be addressed, but it should not effect the task hashing comparison.

Furthermore, we don't want to be executing the tasks twice once in Go and once in Rust.

### Testing Instructions

`pnpm test` in the e2e directory

Closes TURBO-1415


NicholasLYang:
Running all of your tasks twice would be a great anti-Turborepo: takes longer, does more work than it should

NicholasLYang (approved):

PR #6092, opened 2023-10-04
fix(turborepo): Size comes from the thing being read.

### LINKED ISSUE #6046: [turborepo] fails to hash the project without a clear reason

latest

pnpm

Mac

### Describe the Bug

I started to see this issue after updating the turbo to the latest version

```bash
➜ npx turbo run build
 ERROR  run failed: error hashing package files: io error: No such file or directory (os error 2)
Turbo error: error hashing package files: io error: No such file or directory (os error 2)
```

It seems like this issue was introduced on `1.10.7`. Did some testing and `1.10.6` works fine.
It would be awesome to have a clear error message though.

### Expected Behavior

turbo run works :)

### To Reproduce

1. clone the repo
2. `pnpm install`
3. `cd apps/server`
4. `npx turbo run build`

### Reproduction Repo

https://github.com/bamotf/bamotf

<sub>[TURBO-1399](https://linear.app/vercel/issue/TURBO-1399/[turborepo]-fails-to-hash-the-project-without-a-clear-reason)</sub>

nathanhammond:
By splitting into two separate sources the metadata for hashing does not remain in sync with the contents of the file.

Further, `fs::read_to_end` already calls metadata on the open file, so this is duplicated work.

As a side effect, this PR also restores 1.10.6 behavior for hashing of symlinks.

Fixes #6046

### DIFF
--- a/crates/turborepo-scm/src/manual.rs
+++ b/crates/turborepo-scm/src/manual.rs
@@ -1,4 +1,4 @@
-use std::{fs::Metadata, io::Read};
+use std::io::{ErrorKind, Read};
 
 use globwalk::fix_glob_pattern;
 use hex::ToHex;
@@ -9,13 +9,13 @@ use wax::{any, Glob, Pattern};
 
 use crate::{package_deps::GitHashes, Error};
 
-fn git_like_hash_file(path: &AbsoluteSystemPath, metadata: &Metadata) -> Result<String, Error> {
+fn git_like_hash_file(path: &AbsoluteSystemPath) -> Result<String, Error> {
     let mut hasher = Sha1::new();
     let mut f = path.open()?;
     let mut buffer = Vec::new();
-    f.read_to_end(&mut buffer)?;
+    let size = f.read_to_end(&mut buffer)?;
     hasher.update("blob ".as_bytes());
-    hasher.update(metadata.len().to_string().as_bytes());
+    hasher.update(size.to_string().as_bytes());
     hasher.update([b'\0']);
     hasher.update(buffer.as_slice());
     let result = hasher.finalize();
@@ -30,17 +30,19 @@ pub(crate) fn hash_files(
     let mut hashes = GitHashes::new();
     for file in files.into_iter() {
         let path = root_path.resolve(file.as_ref());
-        let metadata = match path.symlink_metadata() {
-            Ok(metadata) => metadata,
-            Err(e) => {
-                if allow_missing && e.is_io_error(std::io::ErrorKind::NotFound) {
-                    continue;
+        match git_like_hash_file(&path) {
+            Ok(hash) => hashes.insert(file.as_ref().to_unix(), hash),
+            Err(e) => match e {
+                Error::Io(ref io_error, _) => {
+                    if allow_missing && io_error.kind() == ErrorKind::NotFound {
+                        continue;
+                    } else {
+                        return Err(e);
+                    }
                 }
-                return Err(e.into());
-            }
+                _ => return Err(e),
+            },
         };
-        let hash = git_like_hash_file(&path, &metadata)?;
-        hashes.insert(file.as_ref().to_unix(), hash);
     }
     Ok(hashes)
 }
@@ -109,7 +111,7 @@ pub(crate) fn get_package_file_hashes_from_processing_gitignore<S: AsRef<str>>(
         if metadata.is_symlink() {
             continue;
         }
-        let hash = git_like_hash_file(path, &metadata)?;
+        let hash = git_like_hash_file(path)?;
         hashes.insert(relative_path, hash);
     }
     Ok(hashes)


crates/turborepo-scm/src/manual.rs
```diff
@@ -9,13 +9,13 @@ use wax::{any, Glob, Pattern};
 
 use crate::{package_deps::GitHashes, Error};
 
-fn git_like_hash_file(path: &AbsoluteSystemPath, metadata: &Metadata) -> Result<String, Error> {
+fn git_like_hash_file(path: &AbsoluteSystemPath) -> Result<String, Error> {
     let mut hasher = Sha1::new();
     let mut f = path.open()?;
     let mut buffer = Vec::new();
-    f.read_to_end(&mut buffer)?;
+    let size = f.read_to_end(&mut buffer)?;
```
nathanhammond, line 16: `read_to_end` already attempts to do this optimally in terms of syscall count, we don't really need to help out.

`read_to_end` is this:
```rust
// Reserves space in the buffer based on the file size when available.
fn read_to_end(&mut self, buf: &mut Vec<u8>) -> io::Result<usize> {
    let size = buffer_capacity_required(self);
    buf.reserve(size.unwrap_or(0));
    io::default_read_to_end(self, buf, size)
}
```

`buffer_capacity_required` is this:
```rust
fn buffer_capacity_required(mut file: &File) -> Option<usize> {
    let size = file.metadata().map(|m| m.len()).ok()?;
    let pos = file.stream_position().ok()?;
    // Don't worry about `usize` overflow because reading will fail regardless
    // in that case.
    Some(size.saturating_sub(pos) as usize)
}
```

crates/turborepo-scm/src/manual.rs
```diff
@@ -30,17 +30,19 @@ pub(crate) fn hash_files(
     let mut hashes = GitHashes::new();
     for file in files.into_iter() {
         let path = root_path.resolve(file.as_ref());
-        let metadata = match path.symlink_metadata() {
```
nathanhammond, line 33: Point of failure one: `lstat`.

crates/turborepo-scm/src/manual.rs
```diff
@@ -30,17 +30,19 @@ pub(crate) fn hash_files(
...
+                    } else {
+                        return Err(e);
+                    }
                 }
-                return Err(e.into());
-            }
+                _ => return Err(e),
+            },
         };
-        let hash = git_like_hash_file(&path, &metadata)?;
```
nathanhammond, line 42: Point of failure two: `fopen`. `fopen` also happens to follow symlinks, which means that our metadata call for returning size is not correlated the the actual number of bytes read in.

crates/turborepo-scm/src/manual.rs
```diff
@@ -30,17 +30,19 @@ pub(crate) fn hash_files(
     let mut hashes = GitHashes::new();
     for file in files.into_iter() {
         let path = root_path.resolve(file.as_ref());
-        let metadata = match path.symlink_metadata() {
-            Ok(metadata) => metadata,
-            Err(e) => {
-                if allow_missing && e.is_io_error(std::io::ErrorKind::NotFound) {
-                    continue;
+        match git_like_hash_file(&path) {
```
nathanhammond, line 33: So, now we just attempt to hash the file, and let `read_to_end` do its thing.

We can then investigate the returned error to see if it is `ENOENT`.

crates/turborepo-scm/src/manual.rs
```diff
@@ -109,7 +111,7 @@ pub(crate) fn get_package_file_hashes_from_processing_gitignore<S: AsRef<str>>(
         if metadata.is_symlink() {
             continue;
         }
```
nathanhammond, line 113: This line of code + `follow_links = false` means that if the symlink target is outside of the package or in the package but otherwise ignored we get a bad hash.

I believe that we should be following symlinks, and if it is a broken symlink, hash it the same way `git` does (by target).

nathanhammond (commented):
Reviewer's Guide!

crates/turborepo-scm/src/manual.rs
nathanhammond: (This would also means that if we `ls-tree` and have a symlink, we need to replace it with the target hash if the target exists, and, if the target is a directory, walk it.)

gsoltis: The intention with the change was to hash symlink files, not their targets. There's been a series of differently-incorrect behavior here:
 - As @arlyon notes, `ls-tree` does list a hash for the symlink. So without inputs, they are wholesale ignored.
 - _when using inputs_, and at least since switching to the encoder setup in Go, symlinks caused an error and we fell back to manual hashing. It's very possible that was the behavior before also, I haven't checked
 - manual hashing in Go calls `os.Open`, which reads the target.
 - Rust `ls-tree` has the same restriction as the Go implementation, it ignores symlinks
 - Rust git-based hashing, _with inputs_, was intended to switch to hashing the contents of the symlink, but there are two problems:
   1. We eventually get to `fopen` as @nathanhammond notes. This follows the link. The Go manual file hashing implementation has the same behavior. At least for the Rust implementation, that was not the intention, and I would consider it a bug at least insofar as it doesn't do what was intended.
   2. As noted in the slack thread, the _raw_ contents of the link is not portable, so probably not what we want to hash

In terms of what I think _should_ happen is the following:
 - Hash a portable representation of the contents of the symlink, _not_ the target. This saves us from having to care about file vs directory, broken vs working, and in-repo vs not, which in the presence of `inputs` is not even necessarily a clear distinction (do we follow in-repo, but out-of-inputs links?). Users will be required to ensure that the target is covered by their inputs if they want the target hashed. This is mirrors our behavior for caching: we don't cache the _target_ of symlinks. If you want it cached, ensure that the target is also covered by the `outputs` globs. 
 - Not for this PR, but eventually: determine how to handle symlinks when `inputs` is not specified.

gsoltis (changes requested):
Let's make sure we're all on the same page about where this is going before merging. 

I'll defer to @arlyon for Rust notes.

crates/turborepo-scm/src/manual.rs
gsoltis, line 16:
```suggestion
    // Note that read_to_end reads the target if f is a symlink. Currently, this can happen when we are hashing
    // a specific set of files, which in turn only happens for handling dotEnv files. It is likely that in the future we
    // will want to ensure that the target is better accounted for in the set of inputs to the task.
    // Manual hashing, as well as global deps and other places that support globs all ignore symlinks.
    let size = f.read_to_end(&mut buffer)?;
```

gsoltis (approved):
Documented `read_to_end` behavior re: symlinks, updating internal discussion on desired eventual symlink handling.

PR #6100, opened 2023-10-04
fix: cache restore including directories

chris-olszewski:
### Testing Instructions

Added unit test that uses directories.

### DIFF
--- a/crates/turborepo-cache/src/async_cache.rs
+++ b/crates/turborepo-cache/src/async_cache.rs
@@ -210,7 +210,11 @@ mod tests {
             .put(
                 repo_root_path.clone(),
                 hash.clone(),
-                test_case.files.iter().map(|f| f.path.clone()).collect(),
+                test_case
+                    .files
+                    .iter()
+                    .map(|f| f.path().to_owned())
+                    .collect(),
                 test_case.duration,
             )
             .await
@@ -281,7 +285,11 @@ mod tests {
             .put(
                 repo_root_path.clone(),
                 hash.clone(),
-                test_case.files.iter().map(|f| f.path.clone()).collect(),
+                test_case
+                    .files
+                    .iter()
+                    .map(|f| f.path().to_owned())
+                    .collect(),
                 test_case.duration,
             )
             .await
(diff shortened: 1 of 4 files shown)


crates/turborepo-cache/src/cache_archive/create.rs
```diff
@@ -288,10 +291,11 @@ mod tests {
                 file_type: FileType::File,
             },
         ],
-        "027346e0349f948c0a2e7e9badb67d27fcc8ff4d5eacff1e5dd6a09c23a54d6793bf7ef1f25c9ed6b8c74f49d86d7b87478b7a00e24ea72e2ed2cadc0286c761",
-        "027346e0349f948c0a2e7e9badb67d27fcc8ff4d5eacff1e5dd6a09c23a54d6793bf7ef1f25c9ed6b8c74f49d86d7b87478b7a00e24ea72e2ed2cadc0286c761",
+        "973c16d7e8662d3483ff0679b5337d7b9ba2001dbe863604fc8cc60254305750616312b9f988112db918d50fd087d89444d43a64beb4e8102109c5c628510131",
+        "973c16d7e8662d3483ff0679b5337d7b9ba2001dbe863604fc8cc60254305750616312b9f988112db918d50fd087d89444d43a64beb4e8102109c5c628510131",
         "1a2b32fe2b252ec622e5a15af21b274d702faa623d09c6fc51a44e7562cc84ac8b8c368d98d284dfb6666680ee252b071d5fbff44564a952ebaa12fe6f389e68",
```
chris-olszewski, line 296: I expect this will also need to change, but will rely on CI to tell me what it should change to.

nathanhammond (approved):
Red/Green for the win.

nathanhammond (commented):
Also, I have bad news: what you've just discovered is that existing caches don't have that field set. In order to ship cache restoration in Rust we either need to:
- Update the global hash key so we never try to restore caches without that busted size header.
- Somehow duck punch the tar lib to deal with that situation on restoration.

Arguably, this change, which sets this header field should be the one that bumps the global cache key, but it's not a requirement.

PR #6102, opened 2023-10-05
Set --single-package everywhere we can.

nathanhammond:
After parsing args the `run_args` can appear in different places in the struct depending upon how it was invoked. This sets all possible locations.

The distinction here is:
```
turbo run build --single-package
turbo build --single-package
```

Closes TURBO-1423


chris-olszewski (approved):
Good catch!

PR #6103, opened 2023-10-05
perf: Implement optimization for barrel files

kdy1:
### Description

https://vercel.slack.com/archives/C04KC8A53T7/p1696440227248069

### Testing Instructions

Not tested yet

Closes WEB-1710


crates/turbopack-ecmascript/src/tree_shake/pkg_json/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+        }
+        FileContent::NotFound => Ok(false.into()),
+    }
+}
+
+#[turbo_tasks::function]
+async fn read_package_json_for(filename: Vc<FileSystemPath>) -> Result<Vc<FileContent>> {
+    let mut f = filename;
+
+    while !f.await?.is_root() {
```
sokra, line 35: we have a `find_context_file` method which does that

crates/turbopack-ecmascript/src/tree_shake/pkg_json/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+    side_effects: bool,
+}
+
+fn true_by_default() -> bool {
+    true
+}
+
+#[turbo_tasks::function]
+pub async fn is_side_effect_free(filename: Vc<FileSystemPath>) -> Result<Vc<bool>> {
+    let package_json = read_package_json_for(filename.clone()).await?;
```
sokra, line 19: You probably want to use the existing `read_package_json` that we already have

kdy1: Thanks! I don't understand how I missed it.

PR #6108, opened 2023-10-05
chore(ci): Separate turbopack benchmarking jobs into new workflow

mehulkar:
This PR attempts to improve two issues I'm running into:

- changes to test.yml trigger these benchmarks and even though they're not blocking it's very noisy and slow to figure out what's going on
- test.yml is a large file and it's difficult (for me at least) to read through it all every time i want to debug/investigate something. Separating some things would be very helpful

### Cons
 Personally I think these are both ok, but listing in case people miss it in the diff

- Two different PR comments for benchmarking summary and tests.
- Some duplication of "determine jobs", etc steps

Closes TURBO-1425


.github/workflows/test.yml
```diff
@@ -1277,7 +944,6 @@ jobs:
           subjob ${{needs.turbopack_rust_test1.result}} "Turbopack Rust tests (linux)"
           subjob ${{needs.turbopack_typescript.result}} "Turbopack Typescript checks"
           subjob ${{needs.turborepo_rust_test.result}} "Turborepo Rust tests"
-          subjob ${{needs.turbopack_rust_test_bench1.result}} "Turbopack Rust benchmark tests (linux)"
```
ForsakenHarmony, line 1280: this makes it no longer required

.github/workflows/bench-turbopack.yml
```diff
@@ -0,0 +1,579 @@
...
+
+      - name: Setup Rust
+        uses: ./.github/actions/setup-rust
+        with:
+          save-cache: true
+
+      - name: Setup Go
+        uses: ./.github/actions/setup-go
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
```
mehulkar, line 235: Not sure why Go needs to be setup for this job, but not planning to change it in this PR since it's a copy paste of before. I don't want to break something that is working.

sokra (changes requested):
You can't update the PR comment from multiple workflows, that would break the comment

.github/workflows/test.yml
mehulkar: Isn't it still working the same way from the new workflow?  https://github.com/vercel/turbo/pull/6108/files#diff-9f350e5932ba25c06a424509fea40f34c62ac56cdbff4be5d8196202a99e497bR514-R515

ForsakenHarmony: oh, I guess, but we need to change the repo settings and this will create a second comment for the summary for each PR

mehulkar: > but we need to change the repo settings

I don't understand, can you clarify? Also I am not sure if your comment means this PR is breaking something or if it's a suggestion. Feel free to ❌ request changes if it's the former!

> this will create a second comment

yes I called this out in the PR description already. Do you consider this a good reason not to split?

> maybe instead of making it completely separate we can call it from this workflow like the cleanup job is called

I'm not opposed to this, but I've moved over 5 benchmarking jobs. Should all 5 of them be called from test.yml in your opinion? I feel like it's better to just be a separate workflow (also note that turborepo also has `bench-turborepo.yml`, so it would align well with that)

wbinnssmith (approved):

PR #6109, opened 2023-10-05
feat(turborepo): Spaces

NicholasLYang:
### Description

Implements spaces and run summary. Does not hook up task summaries to the actual execution

Notes:
- There are now two main groups of structs: Trackers and Summaries. 
- Trackers are created *before* the run and consist of mostly channels and join handles. 
- Summaries are constructed *after* the run and have all of the data in the dry run representation.
- I put a lot more data into `Visitor`, so it can then construct the summaries. This hopefully reduces the amount of functions with a lot of parameters. However, it does mean that constructing the summary moves the visitor. Debatable if that's good or not.
- Refactored `APIAuth` and `APIClient` to be a single instance that is `Arc`'d

### Testing Instructions

Added tests for rust codepath to integration tests

Closes TURBO-1426

### DIFF
--- a/crates/turborepo-api-client/src/spaces.rs
+++ b/crates/turborepo-api-client/src/spaces.rs
@@ -1,14 +1,9 @@
 use chrono::{DateTime, Local};
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 use turbopath::AnchoredSystemPath;
+use turborepo_vercel_api::SpaceRun;
 
 use crate::{retry, APIClient, Error};
-
-#[derive(Deserialize)]
-pub struct SpaceRun {
-    pub id: String,
-    pub url: String,
-}
 
 #[derive(Serialize)]
 #[serde(rename_all = "lowercase")]
@@ -121,10 +116,30 @@ impl APIClient {
     pub async fn create_space_run(
         &self,
         space_id: &str,
+        token: &str,
         payload: CreateSpaceRunPayload,
     ) -> Result<SpaceRun, Error> {
-        let url = self.make_url(&format!("/v0/spaces/{}/runs", space_id));
-        let request_builder = self.client.post(&url).json(&payload);
+        let mut url = self.make_url(&format!("/v0/spaces/{}/runs", space_id));
+        let mut allow_auth = true;
+
+        if self.use_preflight {
+            let preflight_response = self
+                .do_preflight(token, &url, "POST", "Authorization, User-Agent")
+                .await?;
+
+            allow_auth = preflight_response.allow_authorization_header;
+            url = preflight_response.location.to_string();
+        }
+
+        let mut request_builder = self.client.post(&url).json(&payload);
+
+        if allow_auth {
+            request_builder = request_builder.header("Authorization", format!("Bearer {}", token));
+        }
+
+        if let Some(constant) = turborepo_ci::Vendor::get_constant() {
+            request_builder = request_builder.header("x-artifact-client-ci", constant);
+        }
 
         let response = retry::make_retryable_request(request_builder)
             .await?
--- a/crates/turborepo-api-client/src/spaces.rs
+++ b/crates/turborepo-api-client/src/spaces.rs
@@ -1,6 +1,8 @@
+use chrono::{DateTime, Local};
 use serde::{Deserialize, Serialize};
+use turbopath::AnchoredSystemPath;
 
-use crate::{retry, APIClient, Client, Error};
+use crate::{retry, APIClient, Error};
 
 #[derive(Deserialize)]
 pub struct SpaceRun {
@@ -61,6 +63,41 @@ pub struct CreateSpaceRunPayload {
     #[serde(rename = "originationUser")]
     pub user: Option<String>,
     pub client: SpaceClientSummary,
+}
+
+impl CreateSpaceRunPayload {
+    pub fn new(
+        start_time: DateTime<Local>,
+        synthesized_command: &str,
+        package_inference_root: Option<&AnchoredSystemPath>,
+        git_branch: Option<String>,
+        git_sha: Option<String>,
+        version: String,
+        user: Option<String>,
+    ) -> Self {
+        let start_time = start_time.timestamp_millis();
+        let vendor = turborepo_ci::Vendor::infer();
+        let run_context = vendor.map(|v| v.constant).unwrap_or("LOCAL");
+
+        CreateSpaceRunPayload {
+            start_time,
+            status: RunStatus::Running,
+            command: synthesized_command.to_string(),
+            package_inference_root: package_inference_root
+                .map(|p| p.to_string())
+                .unwrap_or_default(),
+            ty: "TURBO",
+            run_context,
+            git_branch,
+            git_sha,
+            user,
+            client: SpaceClientSummary {
+                id: "turbo",
+                name: "Turbo",
+                version,
+            },
+        }
+    }
 }
 
 #[derive(Serialize)]
--- a/crates/turborepo-api-client/src/spaces.rs
+++ b/crates/turborepo-api-client/src/spaces.rs
@@ -20,14 +20,14 @@ pub struct SpaceClientSummary {
     pub version: String,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Default, Debug, Serialize)]
 struct SpacesCacheStatus {
     status: String,
     source: Option<String>,
     time_saved: u32,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Default, Debug, Serialize)]
 pub struct SpaceTaskSummary {
     key: String,
     name: String,
@@ -43,12 +43,18 @@ pub struct SpaceTaskSummary {
 }
 
 #[derive(Serialize)]
+#[serde(rename_all = "UPPERCASE")]
+pub enum SpaceRunType {
+    Turbo,
+}
+
+#[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CreateSpaceRunPayload {
     pub start_time: i64,
     pub status: RunStatus,
     #[serde(rename = "type")]
-    pub ty: &'static str, // Hardcoded to "TURBO"
+    pub ty: SpaceRunType, // Hardcoded to "TURBO"
     pub command: String,
     #[serde(rename = "repositoryPath")]
     pub package_inference_root: String,
@@ -82,7 +88,7 @@ impl CreateSpaceRunPayload {
             package_inference_root: package_inference_root
                 .map(|p| p.to_string())
                 .unwrap_or_default(),
-            ty: "TURBO",
+            ty: SpaceRunType::Turbo,
             run_context,
             git_branch,
             git_sha,
@@ -98,7 +104,7 @@ impl CreateSpaceRunPayload {
 
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
-struct FinishSpaceRunPayload {
+pub struct FinishSpaceRunPayload {
     status: RunStatus,
     end_time: i64,
     exit_code: u32,
--- a/crates/turborepo-api-client/src/spaces.rs
+++ b/crates/turborepo-api-client/src/spaces.rs
@@ -21,25 +21,25 @@ pub struct SpaceClientSummary {
 }
 
 #[derive(Default, Debug, Serialize)]
-struct SpacesCacheStatus {
-    status: String,
-    source: Option<String>,
-    time_saved: u32,
+pub struct SpacesCacheStatus {
+    pub status: String,
+    pub source: Option<String>,
+    pub time_saved: u32,
 }
 
 #[derive(Default, Debug, Serialize)]
 pub struct SpaceTaskSummary {
-    key: String,
-    name: String,
-    workspace: String,
-    hash: String,
-    start_time: i64,
-    end_time: i64,
-    cache: SpacesCacheStatus,
-    exit_code: u32,
-    dependencies: Vec<String>,
-    dependents: Vec<String>,
-    logs: String,
+    pub key: String,
+    pub name: String,
+    pub workspace: String,
+    pub hash: String,
+    pub start_time: i64,
+    pub end_time: i64,
+    pub cache: SpacesCacheStatus,
+    pub exit_code: u32,
+    pub dependencies: Vec<String>,
+    pub dependents: Vec<String>,
+    pub logs: String,
 }
 
 #[derive(Serialize)]
@@ -107,11 +107,11 @@ impl CreateSpaceRunPayload {
 pub struct FinishSpaceRunPayload {
     status: RunStatus,
     end_time: i64,
-    exit_code: u32,
+    exit_code: i32,
 }
 
 impl FinishSpaceRunPayload {
-    pub fn new(end_time: i64, exit_code: u32) -> Self {
+    pub fn new(end_time: i64, exit_code: i32) -> Self {
         Self {
             status: RunStatus::Completed,
             end_time,
@@ -213,14 +213,14 @@ impl APIClient {
         run_id: &str,
         api_auth: &APIAuth,
         end_time: i64,
-        exit_code: u32,
+        exit_code: i32,
     ) -> Result<(), Error> {
         let url = format!("/v0/spaces/{}/runs/{}", space_id, run_id);
 
         let payload = FinishSpaceRunPayload::new(end_time, exit_code);
 
         let request_builder = self
-            .create_request_builder(&url, &api_auth, Method::PATCH)
+            .create_request_builder(&url, api_auth, Method::PATCH)
             .await?
             .json(&payload);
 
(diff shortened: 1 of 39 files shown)


crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -107,10 +107,53 @@ impl CommandBase {
...
+        let team_slug = repo_config.team_slug();
+
+        let token = self.user_config()?.token();
+
+        Ok(team_id.zip(token).map(|(team_id, token)| APIAuth {
+            team_id: team_id.to_string(),
+            token: token.to_string(),
+            team_slug: team_slug.map(|s| s.to_string()),
+        }))
+    }
```
chris-olszewski, line 136: I'm confused as to why we have two identical functions.

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -107,10 +107,53 @@ impl CommandBase {
...
+
     pub fn args(&self) -> &Args {
         &self.args
     }
 
+    pub fn spaces_api_client(&self) -> Result<APIClient> {
```
chris-olszewski, line 142: Not sure if we expect the clients to diverge more in the future, but if we don't refactoring the constructors to make it clear that they only differ in `api_url`

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -236,6 +216,12 @@ impl<'a> Run<'a> {
 
         let color_selector = ColorSelector::default();
 
+        let api_auth = self.base.api_auth()?;
+        let api_client = self.base.api_client()?;
+
+        let async_cache =
+            AsyncCache::new(&opts.cache_opts, &self.base.repo_root, api_client, api_auth)?;
+
```
chris-olszewski, line 224: Any reason for the code shuffling?

turborepo-tests/integration/tests/run-summary/discovery.t
```diff
@@ -18,4 +18,23 @@ Setup
    Cached:    0 cached, 1 total
      Time:\s*[\.0-9]+m?s  (re)
   Summary:    .+\.turbo\/runs\/[a-zA-Z0-9]+.json (re)
-  
\ No newline at end of file
+  
+
+
+  $ ${TURBO} run build --summarize --filter=my-app --experimental-rust-codepath --no-daemon
```
chris-olszewski, line 24: I'm not sure I understand what this integration test is supposed to be testing.

crates/turborepo-paths/src/absolute_system_path.rs
```diff
@@ -30,7 +31,8 @@ pub enum PathRelation {
     Child,
 }
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Serialize)]
```
chris-olszewski, line 34: Do we want to be serializing absolute system paths?

crates/turborepo/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license = "MPL-2.0"
 # This is for the convenience of running daily dev workflows, i.e running
 # `cargo xxx` without explicitly specifying features, not that we want to
 # promote this as default backend.
-default = ["rustls-tls"]
+default = ["rustls-tls", "go-daemon", "run-stub"]
```
chris-olszewski, line 12: I'm guessing this is here by accident. If not, then it should be in it's own PR along with some testing/explanation.
```suggestion
default = ["rustls-tls"]
```

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -1,28 +1,168 @@
...
     state_thread: tokio::task::JoinHandle<SummaryState>,
     #[serde(skip)]
     sender: mpsc::Sender<Message>,
     command: String,
-    package_inference_path: Option<&'a AnchoredSystemPath>,
-    started_at: DateTime<Local>,
+    success: usize,
+    failed: usize,
+    cached: usize,
+    attempted: usize,
```
chris-olszewski, line 47: All of these are tracked already in the `SummaryState` method that gets returned by calling `finish`.
```suggestion
```

chris-olszewski (changes requested):
First pass that I want to get out before I continue.

I could be wrong, but I can't imagine that allowing the spaces client to send unstructured data is desired or unavoidable.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -323,19 +311,24 @@ impl<'a> Run<'a> {
 
         debug!("package file hashes: {:?}", package_file_hashes);
 
+        let spaces_client = self.base.spaces_api_client()?;
+        let spaces_api_auth = self.base.spaces_api_auth()?;
         let mut run_summary = RunSummary::new(
```
chris-olszewski, line 316: What's the expected way for the task graph visitor to communicate to the run summary code since the run summary construction happens after the task graph gets visited?

crates/turborepo-lib/src/run/summary/spaces.rs
```diff
@@ -62,10 +181,69 @@ impl SpacesClient {
...
+            }
+
+            SpacesClientResult {
+                errors: self.errors,
+                run: Some(run),
+            }
+        });
+
+        Ok(SpacesClientHandle { handle, tx })
+    }
```
chris-olszewski, line 217: I'm pretty sure this isn't what we want. Making this `async` means that upon creation we aren't moving forward with execution until that first spaces API request finishes which isn't what the Go code did. We want to be able to fire off the `create_run` request in the background and proceed to start execution ASAP.

crates/turborepo-lib/src/run/summary/spaces.rs
```diff
@@ -1,19 +1,71 @@
...
+    tx: Sender<SpaceRequest>,
+}
+
+impl Debug for SpacesClientHandle {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        // We can't print much more than that since handle/tx are both
+        // opaque
+        f.debug_struct("SpaceClientHandle").finish()
+    }
+}
```
chris-olszewski, line 43: In the past I've used `Derivate` for this sort of thing: https://mcarton.github.io/rust-derivative/latest/index.html

crates/turborepo-api-client/src/spaces.rs
```diff
@@ -1,8 +1,58 @@
...
+}
 
 impl APIClient {
-    #[allow(dead_code)]
-    async fn create_spaces_run() {
-        todo!()
+    pub async fn create_space_run(
+        &self,
+        space_id: &str,
+        payload: impl Serialize,
```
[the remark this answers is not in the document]
NicholasLYang: Yeah, fixed now. tbf, this is almost exactly what we did in Go too

crates/turborepo-lib/src/commands/mod.rs
NicholasLYang: Yeah...me too. There was some debate about what to do around spaces auth vs regular API auth (cc @Zertsov). I don't recall the exact details but I remember Nathan saying something about just duplicating the auth and the API clients for now.

chris-olszewski: Do we want them to be distinct types?

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -1,28 +1,164 @@
...
     command: String,
-    package_inference_path: Option<&'a AnchoredSystemPath>,
-    started_at: DateTime<Local>,
+    #[serde(rename = "repoPath", skip_serializing_if = "Option::is_none")]
+    package_inference_root: Option<&'a AnchoredSystemPath>,
+    #[serde(serialize_with = "serialize_datetime")]
+    pub(crate) start_time: DateTime<Local>,
+    #[serde(serialize_with = "serialize_optional_datetime")]
+    pub(crate) end_time: Option<DateTime<Local>>,
+    pub(crate) exit_code: Option<u32>,
```
chris-olszewski, line 50: I'm realizing I should have pushed back on this in the run summary PR. I don't think we want to be serializing this structure. This structure should be responsible on collecting the stats of individual runs. There should be another structure that matches the schema of the JSON/text we output that gets populated when it is requested. Basically rename this to`ExecutionSummaryTracker` and introduce a new `ExecutionSummary`

The Go code had poor boundaries between the display logic vs the collecting the data.
 - `end_time` and `exit_code` are only known once execution completes and will always be present so we could drop the wrapping `Option` if they got moved to the new `ExecutionSummary`
 - `package_inference_root` is only used for display
 - `start_time` gets passed in from the top only to be stored until the end for display

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -1,28 +1,164 @@
...
+                duration.num_seconds()
+            )
+        } else if duration.num_minutes() > 0 {
+            format!("{}m{}s", duration.num_minutes(), duration.num_seconds())
+        } else if duration.num_seconds() > 0 {
+            format!("{}s", duration.num_seconds())
+        } else {
+            format!("{}ms", duration.num_milliseconds())
+        }
+    }
```
chris-olszewski, line 75: I think this can get extracted to something like and it would decouple our display logic for time from the rest of the summary
```
struct TurboTime(OldDuration)

impl TurboTime {
    pub fn new(start: DateTime, end: DateTime) -> Self {
        Self(end_time.trunc_subsecs(3).singed_duration_since(start.truc_subset(3)))
    }
}

impl fmt::Display for TurboTime {
...
}
```

crates/turborepo-lib/src/commands/mod.rs
NicholasLYang: Fixed now. Seems like we consolidated config

crates/turborepo-lib/src/run/summary/execution.rs
NicholasLYang: Hmm yeah I'm starting to agree with that. I suspect what we should do is completely split out tracking and run summary. Because right now we're trying to do both at the same time (see `RunSummary` vs `RunSummaryInner`) and that just isn't working. We can make a bunch of tracking structs, then convert to summaries and print those.

chris-olszewski (changes requested):
Seems a few of the changes are causing tests to fail to compile.

Overall this looks really good, I'm excited to hook it up.

crates/turborepo-cache/src/async_cache.rs
```diff
@@ -31,8 +31,8 @@ impl AsyncCache {
     pub fn new(
         opts: &CacheOpts,
         repo_root: &AbsoluteSystemPath,
-        api_client: APIClient,
-        api_auth: Option<APIAuth>,
+        api_client: Arc<APIClient>,
+        api_auth: Option<Arc<APIAuth>>,
```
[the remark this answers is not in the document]
NicholasLYang: Yeah, with the new perf info, a couple clones won't hurt

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -71,16 +83,21 @@ impl<'a> Visitor<'a> {
...
         global_env_mode: EnvMode,
         ui: UI,
         silent: bool,
+        version: &'static str,
         manager: ProcessManager,
         repo_root: &'a AbsoluteSystemPath,
         global_env: EnvironmentVariableMap,
+        packages: HashSet<WorkspaceName>,
+        started_at: DateTime<Local>,
+        global_hash_inputs: GlobalHashableInputs<'a>,
```
chris-olszewski, line 100: I'm not sure I get why adding these into the `Visitor` is warranted.

The `RunTracker` constructor takes `version` and `started_at`, I think it's reasonable to hold onto these and use them for the `finish` call at that level. This avoids us accidentally passing in different values for the run start API call and the run finish API call.

`packages` and `global_hash_inputs` don't affect walking the task graph at all and can be provided in the call to `finish`

```suggestion
        manager: ProcessManager,
        repo_root: &'a AbsoluteSystemPath,
        global_env: EnvironmentVariableMap,
```

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -392,6 +414,40 @@ impl<'a> Visitor<'a> {
         Ok(errors)
     }
 
+    /// Finishes visiting the tasks, creates the run summary, and either
+    /// prints, saves, or sends it to spaces.
+    pub(crate) async fn finish(self) -> Result<(), Error> {
```
chris-olszewski, line 419: I think we want to take `exit_code` as a parameter here as the `visit` method returns and exit code and we can plumb it in.

crates/turborepo-lib/src/run/summary/mod.rs
```diff
@@ -122,50 +198,174 @@ impl<'a> RunSummary<'a> {
...
+            run_type,
+            spaces_client_handle: self.spaces_client_handle,
+        })
+    }
+
+    #[tracing::instrument(skip(pkg_dep_graph, ui))]
+    pub async fn finish<'a>(
+        self,
+        start_time: DateTime<Local>,
+        exit_code: u32,
```
chris-olszewski, line 230: Does the Spaces API require an unsigned int? [Rust](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code) and [Go](https://pkg.go.dev/os#ProcessState.ExitCode) produce signed ints if so we should probably port whatever conversion logic we had in Go

```suggestion
        exit_code: i32,
```

chris-olszewski (commented):
Just flushing some notes I have at the moment

crates/turborepo-lib/src/run/summary/mod.rs
```diff
@@ -48,67 +69,122 @@ enum RunType {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    execution: Option<ExecutionSummary<'a>>,
     packages: HashSet<WorkspaceName>,
     env_mode: EnvMode,
     framework_inference: bool,
-    execution_summary: Option<ExecutionSummary<'a>>,
     tasks: Vec<TaskSummary<'a>>,
-    user: Option<String>,
+    user: String,
+    pub scm: SCMState,
```
chris-olszewski, line 108: I'm not seeing us reference this field from outside this module
```suggestion
    scm: SCMState,
```

chris-olszewski (approved):
Looks good. Any remaining concerns I have are all very minor and can easily get addressed later down the road.

crates/turborepo-lib/src/run/global_hash.rs
```diff
@@ -34,23 +34,23 @@ pub struct GlobalHashableInputs<'a> {
     pub pass_through_env: Option<&'a [String]>,
     pub env_mode: EnvMode,
     pub framework_inference: bool,
-    pub dot_env: &'a [RelativeUnixPathBuf],
+    pub dot_env: Option<&'a [RelativeUnixPathBuf]>,
+    pub env_at_execution_start: &'a EnvironmentVariableMap,
```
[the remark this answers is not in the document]
NicholasLYang: It's used so that `GlobalHashableInputs` can generate `GlobalHashSummary`. It's a little awkward though, so we could change it.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -235,11 +213,13 @@ impl<'a> Run<'a> {
             .workspace_info(&WorkspaceName::Root)
             .expect("must have root workspace");
 
-        let root_external_dependencies_hash = root_workspace.get_external_deps_hash();
+        let is_monorepo = !opts.run_opts.single_package;
+
+        let root_external_dependencies_hash =
+            is_monorepo.then(|| root_workspace.get_external_deps_hash());
```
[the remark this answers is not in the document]
NicholasLYang: Yeah it's fairly recent. Almost a Ruby-esque feature

crates/turborepo-ci/src/lib.rs
```diff
@@ -36,7 +37,23 @@ pub fn is_ci() -> bool {
...
 impl Vendor {
     // Returns info about a CI vendor
-    pub fn get_info() -> Option<&'static Vendor> {
+    pub fn infer() -> Option<&'static Vendor> {
+        *VENDOR.get_or_init(Self::infer_inner)
+    }
+
+    /// Gets user from CI environment variables
+    /// We return an empty String instead of None because
+    /// the API expects some sort of string here.
```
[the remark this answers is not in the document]
NicholasLYang: Yeah I'll clarify that

crates/turborepo-lib/src/task_graph/visitor.rs
NicholasLYang: Yeah I ended up a little too dogmatic in the refactoring. I wanted trackers to *only* have join handles and receivers, and the finish call to take no arguments. I think that's a good middle ground

crates/turborepo-lib/src/run/summary/mod.rs
NicholasLYang: Just an assumption on my part. Changed

PR #6112, opened 2023-10-05
chore: fix clippy lints

chris-olszewski:
### Description

Fix some clippy lints that snuck in:
 - The clippy lints weren't blocking for `turborepo-auth` crate due to the lack of `#![deny(clippy:all)]`
 - Unsure why I get the clippy warnings locally, but they don't show up on CI. Clippy is running on this crate in CI, but the lints don't show up. Either way, the lints were correct as the `move` directive in the `async` block means that the redeclaration is unnecessary since they already are referenced in the block.

### Testing Instructions

CI

Closes TURBO-1429


NicholasLYang (approved):

crates/turborepo-auth/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(clippy::all)]
```
mehulkar, line 1: why don't we have this as the default behavior? If we're using a linter, shouldn't we go all in and require that it passes universally?

mehulkar (approved):

crates/turborepo-auth/src/lib.rs
chris-olszewski: Sadly due to a few constraints we can't just make this the default behavior:
 - Workspace level clippy configuration is [severely limited](https://doc.rust-lang.org/nightly/clippy/lint_configuration.html) to what we can configure at that level
 - Turborepo crates aren't default members of the workspace

We could pass it as part of the invocation during CI so at least CI would always catch it, but it appears that `cargo groups` doesn't support arg forwarding:
```
[2 olszewski@chriss-mbp] /Users/olszewski/code/vercel/turborepo $ cargo groups clippy turborepo-libraries --features rustls-tls -- --deny clippy::all   
error: unexpected argument '--deny' found

Usage: Usage: cargo groups clippy [OPTIONS] <GROUP>

```
@NicholasLYang if you're looking for something to do on your DYF 😉

PR #6116, opened 2023-10-05
Ensure process.env.TURBOPACK is always set

timneutkens:
### Description

Related to https://github.com/vercel/next.js/pull/56496. Server Actions currently shows an error as it seems to use the react-dom-webpack version instead of react-dom-turbopack. This seems to be caused by the changes in https://github.com/vercel/next.js/pull/56438 which checks for `process.env.TURBOPACK` which is not always set currently, unlike `process.turbopack`. This PR adds it in the other places where `process.turbopack` is set.

### DIFF
--- a/crates/turbopack-cli/src/contexts.rs
+++ b/crates/turbopack-cli/src/contexts.rs
@@ -177,6 +177,7 @@ pub fn get_client_asset_context(
 fn client_defines(node_env: &NodeEnv) -> Vc<CompileTimeDefines> {
     compile_time_defines!(
         process.turbopack = true,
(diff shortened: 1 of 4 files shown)


timneutkens:
* main
  * **PR #6116** <a href="https://app.graphite.dev/github/pr/vercel/turbo/6116" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/6116?utm_source=stack-comment).

jridgewell (approved):

sokra (approved):

PR #6141, opened 2023-10-10
Escape colons in globs before sending to the daemon

### LINKED ISSUE #6136: [turborepo] turbo outputs cache breaks on tasks with `:` in the name

1.10.15

pnpm

Mac

### Describe the Bug

For compat reasons with older workflows and to match other patterns in our monorepo we prefix our turbo commands with `turbo:` in our package.json scripts, eg we'd run `pnpm turbo turbo:build` to run our turbrepo-managed build scripts.

This worked fine on 1.10.14. After updating to 1.10.15 we're now seeing the following warn log:

```
@alto/eslint-plugin:turbo:build: Failed to mark outputs as cached for @alto/eslint-plugin#turbo:build: rpc error: code = InvalidArgument desc = failed to parse glob expression: libs/eslint-plugin/.turbo/turbo-turbo:build.log
```

the full output from the `pnpm turbo turbo:build` command is as follows:

```
❯ pnpm turbo turbo:build
• Packages in scope: @alto/eslint-plugin, [others redacted]
• Running turbo:build in 11 packages
• Remote caching disabled
@alto/eslint-plugin:turbo:build: Failed to mark outputs as cached for @alto/eslint-plugin#turbo:build: rpc error: code = InvalidArgument desc = failed to parse glob expression: libs/eslint-plugin/.turbo/turbo-turbo:build.log
@alto/eslint-plugin:turbo:build: cache hit, suppressing logs 7414b54f61b794e9
```

If we rename all of our commands to `turbo-build` instead of `turbo:build` then the warning goes away.

### Expected Behavior

Turbo can cache outputs for commands that include a `:` in the name.

### To Reproduce

1. Add a `:` to a turbo command name
2. see the log show up when you run the command

### Reproduction Repo

_No response_

<sub>[TURBO-1440](https://linear.app/vercel/issue/TURBO-1440/[turborepo]-turbo-outputs-cache-breaks-on-tasks-with-in-the-name)</sub>

gsoltis:
### Description

Fixes #6136 
 - Escaped `:` in globs sent to the daemon from the Go client from unix clients, and truncate filenames with `:` on windows due to [Alternate Data Stream](https://learn.microsoft.com/en-us/sysinternals/downloads/streams)
 - Dropped printed warning when we fail to use the daemon for something. It is entirely possible that filewatching will initially time out while getting set up. We don't need to warn on that, there's nothing the user can do, it may just take a while on a large repository.

Note for the future: this may not be the ideal location for converting the format of the globs. It should most likely happen earlier in the process and be tagged with glob source information.

### Testing Instructions

 - Adds a test of the formatting rules applied to globs

Closes TURBO-1441

### DIFF
--- a/cli/internal/daemonclient/daemonclient.go
+++ b/cli/internal/daemonclient/daemonclient.go
@@ -5,6 +5,7 @@ package daemonclient
 import (
 	"context"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/vercel/turbo/cli/internal/daemon/connector"
@@ -33,11 +34,24 @@ func New(client *connector.Client) *DaemonClient {
 	}
 }
 
-// Our globs are expected to be in unix format, and to have : escaped.
-// Important Note: these are _repository relative_ globs, and so will not
-// include Windows drive letters with their own ':' token.
+// formats a repo-relative glob to unix format with ':' characters handled.
+// On windows, ':' is an invalid path character, but you can, and Turborepo does,
+// read to and write from files that contain alternate data streams denoted by ':'.
+// In the case of windows and an alternate data stream, we want change notifications just
+// for the root file. Note that since ':' denotes a data stream for a _file_, it cannot
+// appear in a directory name. Thus, if we find one, we know it's in the filename.
+// See https://learn.microsoft.com/en-us/sysinternals/downloads/streams
 func formatRepoRelativeGlob(input string) string {
-	return strings.ReplaceAll(filepath.ToSlash(input), ":", "\\:")
+	unixInput := filepath.ToSlash(input)
+	if runtime.GOOS == "windows" {
+		colonIndex := strings.Index(input, ":")
+		if colonIndex > -1 {
+			// we found an alternate data stream
+			unixInput = unixInput[:colonIndex]
+		}
+		return unixInput
+	}
+	return strings.ReplaceAll(unixInput, ":", "\\:")
 }
 
 // GetChangedOutputs implements runcache.OutputWatcher.GetChangedOutputs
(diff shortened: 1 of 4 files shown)


cli/internal/daemonclient/daemonclient.go
```diff
@@ -32,12 +34,32 @@ func New(client *connector.Client) *DaemonClient {
 	}
 }
 
+// formats a repo-relative glob to unix format with ':' characters handled.
+// On windows, ':' is an invalid path character, but you can, and Turborepo does,
+// read to and write from files that contain alternate data streams denoted by ':'.
+// In the case of windows and an alternate data stream, we want change notifications just
+// for the root file. Note that since ':' denotes a data stream for a _file_, it cannot
+// appear in a directory name. Thus, if we find one, we know it's in the filename.
+// See https://learn.microsoft.com/en-us/sysinternals/downloads/streams
```
chris-olszewski, line 43: I'm hoping you can't delete data streams from a file individually. i.e. Do we need to worry about the situation where `build:schema` and `build:types` point to the same log file, but only the `types` stream gets deleted and we miss it's removal since the file is still there?

chris-olszewski (approved):
TIL files have streams on windows.

cli/internal/daemonclient/daemonclient.go
chris-olszewski: Based on the `-d` flag of [the CLI tool](https://learn.microsoft.com/en-us/sysinternals/downloads/streams#using-streams) I'm guessing the answer is yes. Obviously non-blocking, but fun.

gsoltis: You can delete individual streams, we get a modification event for the file, which is sufficient for our purposes.

PR #6142, opened 2023-10-10
fix(turborepo): Rust integration test fixes

NicholasLYang:
### Testing Instructions

Closes TURBO-1442

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -332,6 +332,35 @@ jobs:
       - name: Integration Tests
         run: turbo run test --filter=turborepo-tests-integration --color --env-mode=strict --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
 
+  turborepo_integration_rust:
+    name: Turborepo Integration Tests (Rust Codepath)
+    needs: [determine_jobs, build_turborepo]
+    if: needs.determine_jobs.outputs.turborepo_integration == 'true'
+    timeout-minutes: 30
+    runs-on: ${{ matrix.os.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - name: ubuntu
+            runner: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-turborepo-environment
+        with:
+          target: ${{ matrix.os.name }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -316,6 +316,35 @@ jobs:
...
+
+      - name: Cache Prysk
+        id: cache-prysk
+        uses: actions/cache@v3
+        with:
+          path: cli/.cram_env
+          key: prysk-venv-${{ matrix.os.name }}
+
+      - name: Integration Tests
+        run: turbo run test:rust-codepath --filter=turborepo-tests-integration --color --env-mode=strict --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
```
mehulkar, line 346: Can't we just use the existing task `test` and set the `EXPERIMENTAL_RUST_CODEPATH` env var in this job, and set it to pass through?

```suggestion
        run: turbo run test --filter=turborepo-tests-integration --color --env-mode=strict --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
```

NicholasLYang: Interestingly enough, no, because that will enable EXPERIMENTAL_RUST_CODEPATH for the turbo that's running the tests too

chris-olszewski (approved):

.github/workflows/test.yml
mehulkar: ah ok, that makes sense. #inception

.github/workflows/test.yml
```diff
@@ -332,6 +332,35 @@ jobs:
       - name: Integration Tests
         run: turbo run test --filter=turborepo-tests-integration --color --env-mode=strict --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
 
+  turborepo_integration_rust:
```
mehulkar, line 335: Can this be matrixed into the job above? It's 90% the same. An `if: {{matrix.codepath == "rust"}}` shuold do the trick to change the command that runs?

NicholasLYang: We don't want this to be blocking, so I think a second job is best

mehulkar: Why not make it blocking? Seems like we're close to the end and should make it blocking at some point

PR #6147, opened 2023-10-11
feat: add pprof flamegraphs behind a feature flag

arlyon:
### Description

This adds a basic integration for pprof in turbo. It's a little muddy because pprof doesn't really show usage over the threadpool very well, but a good starting point.

Closes TURBO-1445

### DIFF
--- a/crates/turborepo-lib/src/tracing.rs
+++ b/crates/turborepo-lib/src/tracing.rs
@@ -68,6 +68,9 @@ pub struct TurboSubscriber {
 
     chrome_update: Handle<Option<ChromeLog>, DaemonLogLayered>,
     chrome_guard: Mutex<Option<tracing_chrome::FlushGuard>>,
+
+    #[cfg(feature = "pprof")]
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/tracing.rs
```diff
@@ -129,6 +141,17 @@ impl TurboSubscriber {
     }
 }
 
+impl Drop for TurboSubscriber {
+    fn drop(&mut self) {
+        // drop the guard so that the non-blocking file writer stops
+        #[cfg(feature = "pprof")]
+        if let Ok(report) = self.pprof_guard.report().build() {
+            let file = std::fs::File::create("flamegraph.svg").unwrap();
+            report.flamegraph(file).unwrap();
```
chris-olszewski, line 150: I don't think we want to panic during a drop if we can't create a flamegraph

chris-olszewski (approved):

mehulkar (approved):

PR #6148, opened 2023-10-11
Improve performance by reworking calculate_file_hashes

arlyon:
### Description

A large amount of time prepping turbo for execution is spent in `calculate_file_hashes`. This is particularly bad on windows taking around 300ms a pop for each package. The cost of running this once at the root is much lower in comparison, so we pre-cache this instead, storing it in a hashmap.

Pre-optimisation: https://github.com/vercel/turbo/actions/runs/6725968656
Post-optimisation: https://github.com/vercel/turbo/actions/runs/6727286792

Please inspect the profiles to see the improvement. To do so, download the artifact at the bottom of the page and load them into ui.perfetto.dev. In the benches above we trade 110 calls to `git status` at 150ms a piece for one at 900ms. This gives us a heuristic for when this optimisation makes sense. Generally, calculating this per-core will be faster if your task count is less than `5 * NUM_CPU` (on this repository). If you exceed this then the cache approach works out faster.

Closes TURBO-1446

### DIFF
--- a/crates/turborepo-scm/src/package_deps.rs
+++ b/crates/turborepo-scm/src/package_deps.rs
@@ -1,12 +1,178 @@
 use std::collections::HashMap;
 
 use itertools::{Either, Itertools};
-use tracing::debug;
+use tracing::{debug, warn};
 use turbopath::{AbsoluteSystemPath, AnchoredSystemPath, PathError, RelativeUnixPathBuf};
 
 use crate::{hash_object::hash_objects, Error, Git, SCM};
 
 pub type GitHashes = HashMap<RelativeUnixPathBuf, String>;
+
+/// Precalculates workspace data where it makes sense to reduce overhead from
+/// lazily calculating it on demand.
+pub enum CachedPackageFileHasher<'a> {
+    Manual,
+    Git(GitCachedPackageFileHasher<'a>),
+}
+
+impl<'a> CachedPackageFileHasher<'a> {
+    /// When creating a new CachedPackageFileHasher with the git hashes, we
+    /// precalculate and store a datastructure that maps package paths to
+    /// hashes of the files in that package, and a list of files that
+    /// changed. This saves having to call out to git for each package.
+    pub fn new(
+        scm: &'a SCM,
+        repo_root: &AbsoluteSystemPath,
+        package_roots: impl Iterator<Item = &'a AnchoredSystemPath>,
+    ) -> Self {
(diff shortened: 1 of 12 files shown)


mehulkar (commented):
Is it possible to split up the instrumentation piece into another PR so we can potentially have before/after timings after the rework?

arlyon:
@mehulkar I dropped that commit and moved it into it's own PR

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -1,13 +1,209 @@
...
+    /// hashes of the files in that package, and a list of files that
+    /// changed. This saves having to call out to git for each package.
+    pub fn new(
+        scm: &'a SCM,
+        repo_root: &AbsoluteSystemPath,
+        package_roots: impl Iterator<Item = &'a AnchoredSystemPath>,
+    ) -> Self {
+        match scm {
+            SCM::Git(git) => {
+                let mut map: HashMap<&AnchoredSystemPath, (GitHashes, Vec<_>)> = Default::default();
```
chris-olszewski, line 30:
```suggestion
                let mut map: HashMap<&AnchoredSystemPath, (GitHashes, Vec<_>)> = HashMap::with_capacity({
                    let (lower, upper) = package_roots.size_hint();
                    upper.unwrap_or(lower)
                });
```

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -1,13 +1,209 @@
...
+                let span = tracing::debug_span!("constructing map");
+                let enter = span.enter();
+
+                // a subpackage will always have a longer length than its parent,
+                // so we sort by length to ensure that we always and drain subpackages
+                // before their parents
+                for package in package_roots {
+                    let package_str = package.as_str();
+
+                    let (hashes, files) = map.entry(package).or_default();
```
chris-olszewski, line 81: Just asking due to the use of `entry`, but do we ever expect a `package` to appear twice in `package_roots`? If so I believe `files` will contain two duplicate entries. Maybe just use an `map.insert` or dedupe `package_roots` before we iterate through them.

chris-olszewski (changes requested):
Overall the changes seem solid to me. The breakages from the e2e and integration tests seem valid and should be addressed.

arlyon:
This PR is being put on ice until we figure out a plan for `optimize`

PR #6150, opened 2023-10-11
Auth and API Client refactor

Zertsov:
Allows for an api client to be implementable via a trait and re-adds the existing token check.

To test: run the relative tests in the files changed.


.vscode/settings.json
```diff
@@ -21,5 +21,8 @@
   },
   "search.exclude": {
     "crates/turbopack-tests/tests/snapshot/**": true
-  }
+  },
+  "rust-analyzer.linkedProjects": [
```
Zertsov, line 25: whoops, I'll get rid of this

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -0,0 +1,334 @@
...
+
+    login_url
+        .query_pairs_mut()
+        .append_pair("redirect_uri", &redirect_url);
+
+    println!(">>> Opening browser to {login_url}");
+    let spinner = start_spinner("Waiting for your authorization...");
+
+    // Try to open browser for auth confirmation.
+    {
```
NicholasLYang, line 56: Why's this in a block?

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -0,0 +1,334 @@
...
+            login_url_configuration.to_string(),
+            token_cell.clone(),
+        )
+        .await?;
+
+    spinner.finish_and_clear();
+
+    let token = token_cell
+        .get()
+        .ok_or_else(|| anyhow!("Failed to get token"))?;
```
NicholasLYang, line 76: This could probably be an `ok_or` since there isn't any allocation, so a closure is more overhead than the anyhow call.

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -0,0 +1,334 @@
...
+    use turborepo_vercel_api::{
+        CachingStatusResponse, Membership, PreflightResponse, Role, SpacesResponse, Team,
+        TeamsResponse, User, UserResponse, VerifiedSsoUser,
+    };
+    use turborepo_vercel_api_mock::start_test_server;
+
+    use super::*;
+
+    // Used to track how many times the server was hit.
+    lazy_static::lazy_static! {
```
NicholasLYang, line 107: Could we store this in the `MockLoginServer` struct?

crates/turborepo-auth/src/server/login_server.rs
```diff
@@ -0,0 +1,59 @@
...
+        &self,
+        port: u16,
+        login_url_base: String,
+        login_token: Arc<OnceCell<String>>,
+    ) -> Result<()>;
+}
+
+/// TODO: Document this.
+pub struct DefaultLoginServer;
+impl DefaultLoginServer {
```
NicholasLYang, line 26: You can derive `Default` on this too. Although tbh you could just constrict it literally since that's even shorter.

NicholasLYang (changes requested):

crates/turborepo-auth/src/auth/login.rs
Zertsov: Yep, not a problem. I'll move it in there - that was the initial purpose anyway 🔥

crates/turborepo-auth/src/server/login_server.rs
Zertsov: Haven't seen / worked with `derive(Default)` yet, I'll look into that

Zertsov:
looks like I lost some stuff in my rebase, gonna take some time to sort that out

crates/turborepo-auth/src/auth/login.rs
chris-olszewski: Clippy will yell since this is a function call in an `ok_or`. Is there much overhead for a closure like this? I've always treated them as "zero cost" and eagerly used `_else` variants

crates/turborepo-api-client/src/lib.rs
```diff
@@ -364,20 +425,25 @@ impl APIClient {
...
+
+impl APIClient {
     pub fn new(
         base_url: impl AsRef<str>,
         timeout: u64,
         version: &str,
         use_preflight: bool,
-    ) -> Result<Self> {
-        let builder_result = if timeout != 0 {
+    ) -> Result<APIClient> {
```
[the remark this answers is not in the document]
Zertsov: fixed

crates/turborepo-auth/src/auth/login.rs
Zertsov: remnant from when I was trying to understand the borrow checker and/or get around it, should be gone now

crates/turborepo-auth/src/auth/logout.rs
```diff
@@ -0,0 +1,16 @@
+use anyhow::Result;
+use tracing::error;
+use turborepo_ui::{GREY, UI};
+
+pub fn logout<F>(ui: &UI, mut set_token: F) -> Result<()>
```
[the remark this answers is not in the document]
Zertsov: I get your point, but I feel like it keeps more consistency with how the whole crate is laid out. If I'm importing stuff from it, I'd expect a `use turborepo_auth::{login, logout, sso_login}`. Open to suggestions and opinions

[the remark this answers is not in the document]
mehulkar: @chris-olszewski the fact that we're passing in `set_token` right now is an artifact of "it's too hard to move the actual logic in". But eventually I think this crate should receive an optional path, and handle the deletion internally. We need to decouple this from all the other config stuff that's happening

crates/turborepo-auth/src/auth/login.rs
NicholasLYang: Closures are fat pointers if I remember correctly, so two words. If any type is under that, I'd just do `ok_or`

NicholasLYang: But really that's a nitpick

mehulkar (approved):

PR #6158, opened 2023-10-12
chore: add more instrumentation

arlyon:
### Description

This PR is just the additional span annotations from #6148 split into its own thing so we can more easily measure perf diff.

### Testing Instructions

N/A

Closes TURBO-1453

### DIFF
--- a/crates/turborepo-lib/src/task_graph/visitor.rs
+++ b/crates/turborepo-lib/src/task_graph/visitor.rs
@@ -10,7 +10,7 @@ use console::{Style, StyledObject};
 use futures::{stream::FuturesUnordered, StreamExt};
 use regex::Regex;
 use tokio::{process::Command, sync::mpsc};
-use tracing::{debug, error};
+use tracing::{debug, error, Span};
 use turbopath::AbsoluteSystemPath;
 use turborepo_env::{EnvironmentVariableMap, ResolvedEnvMode};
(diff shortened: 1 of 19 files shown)


crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -114,7 +115,11 @@ impl<'a> Visitor<'a> {
         let mut tasks = FuturesUnordered::new();
         let errors = Arc::new(Mutex::new(Vec::new()));
 
+        let span = Span::current();
+
         while let Some(message) = node_stream.recv().await {
+            let span = tracing::debug_span!(parent: &span, "queue_task", task = %message.info);
+            let _enter = span.enter();
```
chris-olszewski, line 122: Do we want to drop this here so we can measure how long we block the main thread before moving the work to a new task?

chris-olszewski (approved):
More tracing! Do think we might want to measure how long we block the main thread in the visitor, but aside from that it all looks good.

mehulkar (approved):

crates/turborepo-lib/src/task_graph/visitor.rs
arlyon: The amount of work done after spawning the future is so low that we can probably get away with not creating an additional span for it.

PR #6166, opened 2023-10-12
fix(turborepo): Only use sccache in Vercel repo

NicholasLYang:
### Description

We now use sccache with s3 in our builds, but that doesn't work on forks because they don't get access to our s3 credentials. This disables sccache on forks.

### Testing Instructions
This PR is from a fork so it should test the logic

Closes TURBO-1459

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -236,15 +236,6 @@ jobs:
         env:
           SCCACHE_BUCKET: turborepo-sccache
           SCCACHE_REGION: us-east-2
-          RUSTC_WRAPPER: "sccache"
-          CARGO_INCREMENTAL: 0
-          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-
-      - run: turbo run build --filter=cli --color --env-mode=strict --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
-        env:
-          SCCACHE_BUCKET: turborepo-sccache
-          SCCACHE_REGION: us-east-2
           # Only use sccache if we're in the Vercel repo.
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -241,9 +241,20 @@ jobs:
...
 
+      - run: turbo run build --filter=cli --color --env-mode=strict --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
+        env:
+          SCCACHE_BUCKET: turborepo-sccache
+          SCCACHE_REGION: us-east-2
+          # Only use sccache if we're in the Vercel repo.
+          RUSTC_WRAPPER: ${{ !github.event.pull_request.head.repo.fork && "sccache" || "" }}
+          CARGO_INCREMENTAL: 0
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
mehulkar, line 252: Is the whole block supposed to be duplicated or did you just mean to add the conditional to the block above? Seems like this will run the build twice

NicholasLYang: whoops yeah, removed

.github/workflows/test.yml
```diff
@@ -236,14 +236,16 @@ jobs:
         env:
           SCCACHE_BUCKET: turborepo-sccache
           SCCACHE_REGION: us-east-2
-          RUSTC_WRAPPER: "sccache"
+          # Only use sccache if we're in the Vercel repo.
+          RUSTC_WRAPPER: ${{ !github.event.pull_request.head.repo.fork && "sccache" || "" }}
```
chris-olszewski, line 240: GH Actions apparently doesn't like double quotes:
> Wrapping with double quotes (") will throw an error.


[source](https://docs.github.com/en/actions/learn-github-actions/expressions#literals)
```suggestion
          RUSTC_WRAPPER: ${{ !github.event.pull_request.head.repo.fork && 'sccache' || '' }}
```

mehulkar (approved):
✔️  other than the quoting issue chris mentioned

PR #6167, opened 2023-10-12
fix: support non UTF-8 logs

chris-olszewski:
### Description

This PR adds support for non UTF-8 output from tasks. Previously we would fail if a task produced UTF-8 when attempting to convert it into a `String` via either [`BufRead::lines`](https://doc.rust-lang.org/std/io/trait.BufRead.html#method.lines) or [`BufRead::read_line`](https://doc.rust-lang.org/std/io/trait.BufRead.html#method.read_line)

I'd suggest reviewing this by-commit as each commit should either:
 - add a failing test
 - fix the failing test that was added in the previous commit

### Testing Instructions

Added failing unit tests for dealing with bytes that aren't valid UTF-8.

Tested the behavior of how we handle non-utf8 output in `turbo`:
```
# Script that produces invalid utf8
[0 olszewski@chriss-mbp] /tmp/global $ cat packages/ui/nasty.sh
printf "%b" '\x00\x9f\x92\x96'
[0 olszewski@chriss-mbp] /tmp/global $ turbo_dev build --experimental-rust-codepath --filter=ui --force
ui:build: cache bypass, force executing c1a3e94329df105c
ui:build: 
ui:build: > ui@0.0.0 build /private/tmp/global/packages/ui
ui:build: > ./nasty.sh
ui:build: 
ui:build:
[0 olszewski@chriss-mbp] /tmp/global $ turbo_dev build --experimental-rust-codepath --filter=ui        
ui:build: cache hit, suppressing logs c1a3e94329df105c
ui:build: 
ui:build: > ui@0.0.0 build /private/tmp/global/packages/ui
ui:build: > ./nasty.sh
ui:build: 
ui:build:
```

Closes TURBO-1460

### DIFF
--- a/crates/turborepo-lib/src/process/child.rs
+++ b/crates/turborepo-lib/src/process/child.rs
@@ -760,7 +760,7 @@ mod test {
     }
 
     #[tokio::test]
-    async fn test_wait_with_non_utf8_single_outut() {
+    async fn test_wait_with_non_utf8_single_output() {
         let script = find_script_dir().join_component("hello_non_utf8.js");
         let mut cmd = Command::new("node");
         cmd.args([script.as_std_path()]);
@@ -775,7 +775,6 @@ mod test {
             .await
             .unwrap();
 
-        // There are no ordering guarantees so we accept either order of the logs
         assert_eq!(buffer, &[0, 159, 146, 150, b'\n']);
         assert_matches!(exit, Some(ChildExit::Finished(Some(0))));
     }
--- a/crates/turborepo-lib/src/process/child.rs
+++ b/crates/turborepo-lib/src/process/child.rs
@@ -27,7 +27,7 @@ use futures::future::{try_join3, try_join4};
 use itertools::Itertools;
 pub use tokio::process::Command;
 use tokio::{
-    io::{AsyncBufReadExt, AsyncRead, BufReader},
+    io::{AsyncRead, BufReader},
     join,
     sync::{mpsc, watch, RwLock},
 };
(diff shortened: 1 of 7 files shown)


crates/turborepo-lib/src/process/child.rs
```diff
@@ -733,4 +736,47 @@ mod test {
...
+            .await
+            .unwrap();
+
+        assert_eq!(out, &[0, 159, 146, 150, b'\n']);
+        assert!(err.is_empty());
+        assert_matches!(exit, Some(ChildExit::Finished(Some(0))));
+    }
+
+    #[tokio::test]
+    async fn test_wait_with_non_utf8_single_outut() {
```
NicholasLYang, line 763:
```suggestion
    async fn test_wait_with_non_utf8_single_output() {
```

crates/turborepo-lib/src/process/child.rs
```diff
@@ -733,4 +736,47 @@ mod test {
...
+        let mut child = Child::spawn(cmd, ShutdownStyle::Kill).unwrap();
+
+        let mut buffer = Vec::new();
+
+        let exit = child
+            .wait_with_single_piped_output(&mut buffer)
+            .await
+            .unwrap();
+
+        // There are no ordering guarantees so we accept either order of the logs
```
NicholasLYang, line 778: I'm confused, what are the possible orders?

NicholasLYang (approved):

crates/turborepo-lib/src/process/child.rs
chris-olszewski: Oops, got copy-pasta'd from another test.

PR #6169, opened 2023-10-12
chore(ci): Break up e2e tests

mehulkar:
This commit just moves some code around while I was exploring just
what our e2e tests cover. I want to combine these with our Prysk integration
tests (or vice-versa) eventually as there's some duplication and then figure out how to run
each against each of the binaries we create. Consolidation will also help focus our efforts
on running functional tests across more environments.

The main change worth noting is that before the `basicPrune` and `explicitDepPrune` pipeline fixtuers
were also running all the build/test/lint/file change tests. This was mostly unnecessary and just easier
to setup because of the for loop. I've removed these extra tests since they are not testing anything important.

Closes TURBO-1462


turborepo-tests/e2e/monorepo.ts
```diff
@@ -9,35 +9,55 @@ const turboPath = path.join(
...
+import type { PackageManager } from "./types";
 
-type NPMClient = "npm" | "pnpm6" | "pnpm" | "yarn" | "berry";
+const PACKAGE_MANAGER_VERSIONS: { [key in PackageManager]: string } = {
+  yarn: "yarn@1.22.17",
+  berry: "yarn@3.1.1",
+  pnpm6: "pnpm@6.22.2",
+  pnpm: "pnpm@7.2.1",
+  npm: "npm@8.3.0",
+};
```
mehulkar, line 20: I am interested in this piece. This is the current package managers we're validating with these tests, which is good to know. I'd be surprised if it matched the integration test suite

turborepo-tests/e2e/index.ts
```diff
@@ -1,878 +1,147 @@
...
-    excludePrune: ["c#build"],
-    includePrune: ["a#build"],
-  }, // expect c#build to be removed, since there is no dep between a -> c
-  {
-    npmClient: "yarn" as const,
-    pipeline: explicitPrunePipeline,
-    name: "explicitDepPrune",
-    excludePrune: ["c#build"],
-    includePrune: ["a#build", "b#build"],
-  }, // expect c#build to be included, since a depends on c
```
mehulkar, line 99: this comment was incorrect, it should have been
```
// expect b#build to be included, since a depends on b
```

turborepo-tests/e2e/tests/prune-docker.ts
```diff
@@ -0,0 +1,86 @@
...
+import * as uvu from "uvu";
+import {
+  getImmutableInstallForPackageManager,
+  getCommandOutputAsArray,
+  getLockfileForPackageManager,
+} from "../helpers";
+import { PackageManager } from "../types";
+import { Monorepo } from "../monorepo";
+
+export default function (
```
chris-olszewski, line 12: Not in the scope of this PR, but I think that the `prune/*.t` tests might cover all of the behavior that gets tested here.

chris-olszewski (approved):
Looks *amazing* to me. This makes the e2e test so much more legible.

turborepo-tests/e2e/monorepo.ts
mehulkar: We do have the _ability_ to use other package managers in our setup scripts:

https://github.com/vercel/turbo/blob/2b7ca782cce807231177b65c91ed00c61e796425/turborepo-tests/integration/README.md#L33-L37

But I don't see many examples. Here's one at least:
https://github.com/vercel/turbo/blob/2b7ca782cce807231177b65c91ed00c61e796425/turborepo-tests/integration/tests/single_package/run-yarn.t#L3

turborepo-tests/e2e/helpers.ts
```diff
@@ -0,0 +1,143 @@
...
+import execa from "execa";
+import { Readable } from "stream";
+// @ts-ignore-next-line
+import tar from "tar";
+// @ts-ignore-next-line
+import { ZstdCodec } from "zstd-codec";
+import { Monorepo } from "./monorepo";
+import type { DryRun, PackageManager } from "./types";
+
+export const matchTask =
```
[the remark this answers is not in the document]
mehulkar: Good thought, but I really wanna avoid another CI cycle 😅 , going to leave these as-is

PR #6173, opened 2023-10-13
chore: add a few diagnostic log lines for measuring time-to-execute

arlyon:
### Description

For collecting benches we are extracting data from chrome traces. We can use the following `jq` script to very easily pull out a summary of the data, but for that we need to log specific things out so that chrome tracing can pick them up.

```jq
map({
    platform: .args.platform, 
    name: "time-to-first-task", 
    startTimeUnixMs: (if .args.start_time then .args.start_time | tonumber else null end), 
    durationMs: (if .args.message == "running visitor" then .ts else null end), 
    turboVersion: .args.turbo_version,
    scm: (if .args.scm_manual == "true" then "manual" elif .args.scm_manual == "false" then "git" else null end),
}) 
| reduce .[] as $item ({}; . *= ($item | del(.. | nulls)))
```

### Testing Instructions

N/A (diagnostic changes only)

Closes TURBO-1464

### DIFF
--- a/crates/turborepo-lib/src/run/mod.rs
+++ b/crates/turborepo-lib/src/run/mod.rs
@@ -8,6 +8,7 @@ pub mod task_id;
 use std::{
     io::{BufWriter, IsTerminal, Write},
     sync::Arc,
+    time::SystemTime,
 };
 
 use anyhow::{anyhow, Context as ErrorContext, Result};
@@ -38,6 +39,7 @@ use crate::{
         global_hash::get_global_hash_inputs,
         summary::{GlobalHashSummary, RunSummary},
     },
+    shim::TurboState,
     task_graph::Visitor,
     task_hash::{PackageInputsHashes, TaskHashTrackerState},
 };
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/run/mod.rs
```diff
@@ -63,6 +65,14 @@ impl<'a> Run<'a> {
     }
 
     pub async fn run(&mut self) -> Result<i32> {
+        tracing::trace!(
+            platform = %TurboState::platform_name(),
+            start_time = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).expect("system time after epoch").as_micros(),
+            turbo_version = %TurboState::version(),
```
arlyon, line 71: Parameters supplied to the log using this syntax do not end up in the terminal, but they _do_ end up on the event in chrome tracing in the `args` object.

crates/turborepo-lib/src/shim.rs
```diff
@@ -256,10 +256,14 @@ impl TurboState {
             }
         };
 
-        formatcp!("turbo-{}-{}", OS, ARCH)
+        formatcp!("{}-{}", OS, ARCH)
     }
 
-    pub fn binary_name() -> &'static str {
+    pub const fn platform_package_name() -> &'static str {
```
arlyon, line 262: This was made const along with its parent since you can only call const functions in a const context.

mehulkar (approved):

PR #6176, opened 2023-10-13
feat(benchmark): add workflow to benchmark time to first task

mehulkar:
- [x] make data post successfully to TB
- [x] In the Github Action, save the trace as an output
- [x] Add the Run URL as a field to TB so we can look up the full trace later
- [x] Improve verbose logging situtation #6196
- [x] Make Github Action run on macos-latest, ubuntu-latest and windows-latest
- [x] Make Github Action run on push to main
- [x] Add commitSha and commitTimestamp to data

Closes TURBO-1467


benchmark/src/ttft-send.ts
```diff
@@ -0,0 +1,47 @@
...
+const token = process.env.TINYBIRD_TOKEN;
+
+if (!token) {
+  throw new Error("Missing TINYBIRD_TOKEN env variable");
+}
+
+const DATA_SOURCE_URL =
+  "https://api.us-east.tinybird.co/v0/events?name=turborepo_perf_ttft";
+
+async function foo() {
```
gsoltis, line 16: Rename?

benchmark/src/ttft.ts
```diff
@@ -0,0 +1,46 @@
...
+  };
+  console.log("Executing turbo build in child process with opts", opts);
+
+  try {
+    cp.execSync(
+      `${TURBO_BIN} run build --skip-infer --force --dry --profile ../profile.json`,
+      opts
+    );
+  } catch (e) {
+    // not sure why this is erroring
```
gsoltis, line 39: This seems suspicious. Maybe just drop the `try ... catch` and let it error if it's going to error?

gsoltis (changes requested):

benchmark/src/ttft.ts
mehulkar: removing the pipe did the trick. probably the same issue as we got from a customer recently where large output fills up the max buffer? I tried increasing it a bit (maybe not enough) but that didn't fix it, so just swallowing the output. we don't need it

mehulkar: I brought it back because it still failed in Windows for the same reason: https://github.com/vercel/turbo/actions/runs/6542776699/job/17766330242

I'd like to merge this and then come back to this after we're reduced the amount of output required for tracing

.github/workflows/bench-turborepo.yml
```diff
@@ -40,3 +47,61 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-node
+      - uses: ./.github/actions/setup-go
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      # setup Rust on windows, because apparently setup-turborepo-environment doesn't do this
+      - name: Add rustup for windows
+        if: ${{ matrix.os.runner == 'windows-latest'}}
```
gsoltis, line 79: There's a setup-rust action if that's helpful.

.github/workflows/bench-turborepo.yml
```diff
@@ -40,3 +47,61 @@ jobs:
...
+        run: |
+          rustup target add aarch64-unknown-linux-musl
+          rustup target add x86_64-pc-windows-gnu
+          rustup target add x86_64-unknown-linux-musl
```
gsoltis, line 83: Why do we need linux targets on windows?

benchmark/src/index.ts
```diff
@@ -179,8 +163,8 @@ function cachedBuildWithDependencyChange(): Timing[] {
 }
 
 class Benchmarks {
-  private readonly benchmarks: Benchmark[] = [];
-  private readonly tbirdEvents: TBirdEvent[] = [];
+  private readonly benchmarks: Array<Benchmark> = [];
```
gsoltis, line 166: Just curious, is there a reason for these changes? Like a lint or best practice or something?

gsoltis (approved):
Not blocking, but I have some questions.

.github/workflows/bench-turborepo.yml
mehulkar: Yeah.... it doesn't do the same thing as this, I didn't want to get too in the weeds of refactoring that also

.github/workflows/bench-turborepo.yml
mehulkar: we don't, I didn't want to experiment with it though, the feedback cycle on these is 30-45 mins long. This is copy paste from the release job which sets up for windows and linux https://github.com/vercel/turbo/blob/e28a1c072cfc76b4cb28b3b0256d5ce66336f335/.github/workflows/turborepo-release.yml#L180-L183

benchmark/src/index.ts
mehulkar: Since I was reusing some of the code from `src/index.ts`, I added tsc and eslint checking because I kept messing up easy things that tooling can catch. I believe some of these changes are from `eslint --fix`. I didn't spendt oo much time separating because it's not production code

PR #6179, opened 2023-10-16
fix double self time reporting

sokra:
### Description

fix profile for nested spans

### Testing Instructions

Closes WEB-1787

### DIFF
--- a/crates/turbopack-convert-trace/src/main.rs
+++ b/crates/turbopack-convert-trace/src/main.rs
@@ -236,34 +236,73 @@ fn main() {
                     span.end = ts;
                 }
             }
-            TraceRow::Enter {
-                ts,
-                id,
-                thread_id: _,
-            } => {
+            TraceRow::Enter { ts, id, thread_id } => {
                 let internal_id = ensure_span(&mut active_ids, &mut spans, id);
+                let mut parent_id = spans[internal_id].parent;
+                let mut in_parent = 0;
+                loop {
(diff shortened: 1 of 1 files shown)


crates/turbopack-convert-trace/src/main.rs
```diff
@@ -236,34 +236,73 @@ fn main() {
...
+                    );
+                    if in_parent > 0 {
+                        let parent_id = span.parent;
+                        let span = &mut spans[parent_id];
+                        if let Some(SelfTimeStarted {
+                            ts: ref mut parent_ts,
+                            ..
+                        }) = span.self_start
+                        {
+                            *parent_ts = max(ts, ts_start);
```
jridgewell, line 303: I don't understand this. We're setting the parent's starting timestamp to be later than originally reported?

sokra: Example:

Enter span A
1s
Enter span B
1s
Exit span B
1s
Exit span A


Before this change it collects 3s self time for span A and 1s self time of span B. But that is not correct. We need to subtract the 1s spend in span B from span A when they execute on the same thread.

So when we enter span B and there is already a parent span active on the same thread, we "exit" the parent span (A). Once we exit the inner span (B) again, we "enter" the parent span again. This "enter" is handled on that line. We enter the parent by setting the self start time to the current time.

jridgewell (approved):

PR #6182, opened 2023-10-16
chore(ci): fixes invalid workflow syntax

mehulkar:
Found these using the Github Actions VS Code extension: https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions

Closes TURBO-1470

### DIFF
--- a/.github/actions/setup-turborepo-environment/action.yml
+++ b/.github/actions/setup-turborepo-environment/action.yml
@@ -7,14 +7,6 @@ inputs:
   github-token:
     description: "GitHub token. You can pass secrets.GITHUB_TOKEN"
     required: true
-  setup-rust:
-    description: "Install Rust toolchain to build Turborepo"
-    required: false
-    default: "true"
-  setup-go:
-    description: "Install Go toolchain to build Turborepo"
-    required: false
-    default: "true"
 
 runs:
   using: "composite"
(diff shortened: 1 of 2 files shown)


.github/actions/setup-turborepo-environment/action.yml
```diff
@@ -7,16 +7,6 @@ inputs:
...
-  setup-rust:
-    description: "Install Rust toolchain to build Turborepo"
-    required: false
-    type: boolean
-    default: true
-  setup-go:
-    description: "Install Go toolchain to build Turborepo"
-    required: false
-    type: boolean
-    default: true
```
mehulkar, line 19: We aren't using these options anywhere

.github/workflows/turborepo-release.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         uses: ./.github/actions/setup-turborepo-environment
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
-          target: ${{ matrix.os.name }}
+          target: ubuntu-latest
```
chris-olszewski, line 79: Maybe change this to `linux` instead of `ubuntu-latest`? `ubuntu-latest` gives the impression that this target should be a GitHub Action Runner when that's incorrect.

Out of scope for this PR, but we might want to change this argument to just be `windows` boolean flag. Currently it only gets passed to `setup-rust` and it only has a behavior change if it is set to [`windows`](https://github.com/vercel/turbo/blob/main/.github/actions/setup-rust/action.yml#L51) (which isn't a [GH Runner string](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources))

chris-olszewski (approved):

PR #6187, opened 2023-10-17
Don't open browser when in test mode

Zertsov:
Browsers opening in tests is annoying, so maybe we don't do that.


chris-olszewski (approved):
There's another `webbrowser::open` call in `login.rs` that gets hit by `auth::login::tests::test_login` that can use the same treatment.

PR #6191, opened 2023-10-17
Deduplicate referenced_output_assets

jridgewell:
### Description

We're chunking and outputting a multiple times for some entry points, because `references_to_output_assets` doesn't return a stable reference for empty values. It might also matter for non-empty values, but I can easily reproduce the empty ones.

### Testing Instructions

Closes WEB-1790

### DIFF
--- a/crates/turbopack-core/src/chunk/chunk_group.rs
+++ b/crates/turbopack-core/src/chunk/chunk_group.rs
@@ -185,6 +185,6 @@ async fn references_to_output_assets(
     Ok(if output_assets.is_empty() {
         OutputAssets::empty()
     } else {
-        Vc::cell(output_assets)
+        OutputAssets::new(output_assets)
     })
 }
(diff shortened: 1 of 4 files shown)


crates/turbopack-core/src/chunk/chunk_group.rs
```diff
@@ -182,5 +182,9 @@ async fn references_to_output_assets(
         .copied()
         .filter(|&asset| set.insert(asset))
         .collect::<Vec<_>>();
-    Ok(output_assets)
+    Ok(if output_assets.is_empty() {
+        OutputAssets::empty()
+    } else {
+        OutputAssets::new(output_assets)
+    })
```
sokra, line 189: Maybe we should put this (empty check) logic into `OutputAssets::new`

sokra (approved):

PR #6194, opened 2023-10-17
Move OutputAssets length check into new method

jridgewell:
### Description

https://github.com/vercel/turbo/pull/6191#discussion_r1362057926

### Testing Instructions

Closes WEB-1793

### DIFF
--- a/crates/turbopack-core/src/output.rs
+++ b/crates/turbopack-core/src/output.rs
@@ -25,10 +25,10 @@ pub struct OutputAssets(Vec<Vc<Box<dyn OutputAsset>>>);
 impl OutputAssets {
     #[turbo_tasks::function]
     pub fn new(assets: Vec<Vc<Box<dyn OutputAsset>>>) -> Vc<Self> {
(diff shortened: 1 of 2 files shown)


crates/turbopack-core/src/output.rs
```diff
@@ -25,7 +25,11 @@ pub struct OutputAssets(Vec<Vc<Box<dyn OutputAsset>>>);
 impl OutputAssets {
     #[turbo_tasks::function]
     pub fn new(assets: Vec<Vc<Box<dyn OutputAsset>>>) -> Vc<Self> {
-        Vc::cell(assets)
+        if output_assets.is_empty() {
```
ForsakenHarmony, line 28: why do we even need this conditional?

ForsakenHarmony (approved):

crates/turbopack-core/src/output.rs
jridgewell: Good question, let me verify.

PR #6195, opened 2023-10-17
Remove `OutputAssets::new` if condition

jridgewell:
### Description

It's unneeded, the input param caching will correctly handle this case already.

https://github.com/vercel/turbo/pull/6194#discussion_r1362484606

### Testing Instructions

Closes WEB-1794

### DIFF
--- a/crates/turbopack-core/src/output.rs
+++ b/crates/turbopack-core/src/output.rs
@@ -25,11 +25,7 @@ pub struct OutputAssets(Vec<Vc<Box<dyn OutputAsset>>>);
 impl OutputAssets {
     #[turbo_tasks::function]
     pub fn new(assets: Vec<Vc<Box<dyn OutputAsset>>>) -> Vc<Self> {
-        if assets.is_empty() {
-            OutputAssets::empty()
-        } else {
-            Vc::cell(assets)
-        }
+        Vc::cell(assets)
     }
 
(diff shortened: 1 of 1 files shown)


crates/turbopack-core/src/output.rs
```diff
@@ -25,11 +25,7 @@ pub struct OutputAssets(Vec<Vc<Box<dyn OutputAsset>>>);
...
     pub fn new(assets: Vec<Vc<Box<dyn OutputAsset>>>) -> Vc<Self> {
-        if assets.is_empty() {
-            OutputAssets::empty()
-        } else {
-            Vc::cell(assets)
-        }
+        Vc::cell(assets)
     }
 
     #[turbo_tasks::function]
```
sokra, line 31:
``` diff
-    #[turbo_tasks::function]
     pub fn empty() -> Vc<Self> {
-        Vc::cell(vec![])
+        Self::new(vec![])
     }
```

to make `OutputAssets::new(vec![])` == `OutputAssets::empty()`. They can return the same Vc

jridgewell: Done.

sokra: the empty fn need to be moved into a separate impl block for it for compile fine

sokra (approved):

PR #6205, opened 2023-10-18
feat(Turborepo): Initial run preamble

gsoltis:
### Description

 - Adds packages and targets summary at the beginning of a run
 - Some minor tweaks to output to match Go logic
 - Minor bugfix to `cwrite!` macro
 - Add a `cprint!` macro for lines with mixed color formatting

### Testing Instructions

Existing test suite. We're still missing the post-run summary, so we don't have more integration tests passing, but the diff is smaller now.

Closes TURBO-1484

### DIFF
--- a/crates/turborepo-api-client/src/lib.rs
+++ b/crates/turborepo-api-client/src/lib.rs
@@ -100,18 +100,6 @@ pub struct APIAuth {
     pub team_slug: Option<String>,
 }
 
-impl APIAuth {
-    pub fn is_linked(&self) -> bool {
-        !self.token.is_empty()
-            && (!self.team_id.is_empty()
-                || !self
-                    .team_slug
-                    .as_ref()
-                    .map(|slug| slug.is_empty())
-                    .unwrap_or(true))
-    }
-}
-
 #[async_trait]
 impl Client for APIClient {
     async fn get_user(&self, token: &str) -> Result<UserResponse> {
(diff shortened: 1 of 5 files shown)


chris-olszewski (approved):

crates/turborepo-api-client/src/lib.rs
```diff
@@ -100,6 +100,18 @@ pub struct APIAuth {
     pub team_slug: Option<String>,
 }
 
+impl APIAuth {
+    pub fn is_linked(&self) -> bool {
+        !self.token.is_empty()
```
mehulkar, line 105: Want to defer to the auth crate here? cc @Zertsov

gsoltis: Yes, but I think it needs some refactoring to happen. `APIAuth` might properly be owned by that crate, although we'll have to work out how that interacts with the config code.

NicholasLYang: I don't think this is correct. If there is no token or team id, `APIAuth` [shouldn't even be constructed](https://github.com/vercel/turbo/blob/ef2b800167d527f3fa8bc06ac98f7edaeb51b410/crates/turborepo-lib/src/run/mod.rs#L170). To check if linked we should just check if `Option<APIAuth>` is `Some`

NicholasLYang (changes requested):

crates/turborepo-api-client/src/lib.rs
gsoltis: Ah ok. That makes sense, if having an `APIAuth` implies that we at least have a token and a team id, then that works

gsoltis: Switched to `is_linked = api_auth.is_some()`

NicholasLYang (approved):

PR #6217, opened 2023-10-19
fix: Respect color config for runcache output

NicholasLYang:
### Testing Instructions

Closes TURBO-1493


chris-olszewski (approved):
Don't love the nested `format!` calls, but that's a problem for another day.

PR #6234, opened 2023-10-20
chore(lint): fix autofixable lint violations and warnings

mehulkar:
This is the result of:

1. Remove all the rules in docs/eslintrc.cjs that were turned down to `warn`
2. `pnpm -F docs lint --fix` 
3. Add back rules that are still failing with `warn`

We should still go fix the ones that aren't autofixable, but this was about 5 mins of work to reduce noise in the editor

Closes TURBO-1503


docs/components/Footer.tsx
```diff
@@ -1,23 +1,25 @@
...
 import Link from "next/link";
-import { useState, ReactNode, ReactElement } from "react";
+import type { ReactNode, ReactElement } from "react";
+import { useState } from "react";
 import cn from "classnames";
 import { ThemeSwitch } from "nextra-theme-docs";
 import VercelLogo from "./logos/Vercel";
-import { useTurboSite, TurboSite } from "./SiteSwitcher";
+import type { TurboSite } from "./SiteSwitcher";
+import { useTurboSite } from "./SiteSwitcher";
```
mehulkar, line 9: This is also kinda weird. can't you just do
```
import { useTurboSite, type TurboSite } from "./SiteSwitcher"
```

I'm not going to change it, just curious

docs/components/image/ImageFigure.tsx
```diff
@@ -24,17 +24,17 @@ export function ImageFigure(props: ImageFigureProps): React.ReactNode {
   return (
     <figure className="block text-center" style={{ margin: `${margin}px 0` }}>
       <div className="relative inline-block w-full max-w-full overflow-hidden border-box text-[0px]">
-        {/* eslint-disable-next-line jsx-a11y/alt-text */}
+        {}
```
mehulkar, line 27: This is fun. Probably removed because it's an empty comment?

docs/components/pages/home-shared/GradientSectionBorder.tsx
```diff
@@ -1,6 +1,6 @@
 import cn from "classnames";
+import gradients from "./gradients.module.css";
 import { FadeIn } from "./FadeIn";
-import gradients from "../home-shared/gradients.module.css";
```
mehulkar, line 3: nice catch, linter!

docs/components/Footer.tsx
Zertsov: Should be able to, but I think most linters like them separated. I think it might be for the case of something like the following.

```
import { useTurboSite, type { TurboSite, SomeOtherType, SomeThirdType } } from "./SiteSwitcher"
```
is weirder than
```
import type {
  TurboSite,
  SomeOtherType,
  SomeThirdType 
 } from ...
import { useTurboSite } from ...
```
but I'm not 100% guaranteeing this is the case

docs/components/Container.tsx
```diff
@@ -1,9 +1,9 @@
 import type { ReactNode } from "react";
 
-type Props = {
+interface Props {
```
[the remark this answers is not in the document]
Zertsov: [Looks like this is a known thing](https://github.com/typescript-eslint/typescript-eslint/issues/718). There's no difference in this case but it's interesting that when you removed the `consistent-type-definitions` it auto-fixed it 🤔

gsoltis (approved):

ForsakenHarmony (approved):

PR #6240, opened 2023-10-20
chore: run summary prefactor

chris-olszewski:
### Description

This PR is flushing some changes I've accrued while trying to hook up run summary that aren't directly related to that feature.

I highly suggest reviewing each commit on it's own.

### Testing Instructions

All changes should be accompanied by new unit tests or should be easy to follow refactors.

Closes TURBO-1510

### DIFF
--- a/crates/turborepo-lib/src/engine/mod.rs
+++ b/crates/turborepo-lib/src/engine/mod.rs
@@ -104,12 +104,23 @@ impl Default for Engine<Building> {
 }
 
 impl Engine<Built> {
-    #[allow(dead_code)]
     pub fn dependencies(&self, task_id: &TaskId) -> Option<HashSet<&TaskNode>> {
+        self.neighbors(task_id, petgraph::Direction::Outgoing)
+    }
+
+    pub fn dependents(&self, task_id: &TaskId) -> Option<HashSet<&TaskNode>> {
+        self.neighbors(task_id, petgraph::Direction::Incoming)
+    }
+
+    fn neighbors(
+        &self,
+        task_id: &TaskId,
+        direction: petgraph::Direction,
+    ) -> Option<HashSet<&TaskNode>> {
         let index = self.task_lookup.get(task_id)?;
         Some(
             self.task_graph
-                .neighbors_directed(*index, petgraph::Direction::Outgoing)
+                .neighbors_directed(*index, direction)
                 .map(|index| {
                     self.task_graph
                         .node_weight(index)
--- a/crates/turborepo-lib/src/package_graph/mod.rs
+++ b/crates/turborepo-lib/src/package_graph/mod.rs
@@ -10,7 +10,7 @@ use turbopath::{AbsoluteSystemPath, AnchoredSystemPath, AnchoredSystemPathBuf};
 use turborepo_lockfiles::Lockfile;
 use turborepo_repository::{package_json::PackageJson, package_manager::PackageManager};
 
-use crate::graph;
+use crate::{graph, run::task_id::TaskId};
 
 mod builder;
 
@@ -51,6 +51,13 @@ impl WorkspaceInfo {
         self.package_json_path
             .parent()
             .expect("at least one segment")
+    }
+
+    pub fn task_log_path(&self, task_id: &TaskId) -> AnchoredSystemPathBuf {
+        let mut log_path = self.package_path().to_owned();
+        log_path.push(".turbo");
+        log_path.push(&format!("turbo-{}.log", task_id.task()));
+        log_path
     }
 
     pub fn get_external_deps_hash(&self) -> String {
--- a/crates/turborepo-lib/src/run/summary/execution.rs
+++ b/crates/turborepo-lib/src/run/summary/execution.rs
@@ -226,23 +226,34 @@ pub enum ExecutionState {
     SpawnFailed { err: String },
 }
 
-#[derive(Debug, Serialize)]
-pub struct TaskExecutionSummary {
-    started_at: i64,
-    ended_at: i64,
-    #[serde(skip)]
-    duration: TurboDuration,
-    pub(crate) state: ExecutionState,
-}
-
-impl TaskExecutionSummary {
+impl ExecutionState {
     pub fn exit_code(&self) -> Option<i32> {
-        match self.state {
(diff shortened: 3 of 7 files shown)


crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -104,12 +104,23 @@ impl Default for Engine<Building> {
 }
 
 impl Engine<Built> {
-    #[allow(dead_code)]
     pub fn dependencies(&self, task_id: &TaskId) -> Option<HashSet<&TaskNode>> {
+        self.neighbors(task_id, petgraph::Direction::Outgoing)
+    }
+
+    pub fn dependents(&self, task_id: &TaskId) -> Option<HashSet<&TaskNode>> {
```
chris-olszewski, line 111: Required for populating the `dependents` section of `TaskSummary`

crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -53,6 +53,13 @@ impl WorkspaceInfo {
             .expect("at least one segment")
     }
 
+    pub fn task_log_path(&self, task_id: &TaskId) -> AnchoredSystemPathBuf {
```
chris-olszewski, line 56: Moving the logic out of the task cache so we can reuse it when constructing the `TaskSummary` without plumbing it out of the task cache.

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -509,4 +528,19 @@ mod test {
...
+
+    #[test_case(
+        TaskExecutionSummary {
+            started_at: 123,
+            ended_at: 234,
+            duration: TurboDuration::from(Duration::zero()),
+            state: ExecutionState::Built { exit_code: 0 },
+            exit_code: Some(0),
+        },
+        json!({ "startTime": 123, "endTime": 234, "exitCode": 0 })
```
chris-olszewski, line 540: This is the expected `execution` schema as shown in [run-summary/monorepo.t](https://github.com/vercel/turbo/blob/main/turborepo-tests/integration/tests/run-summary/monorepo.t#L116)

crates/turborepo-lib/src/run/summary/mod.rs
```diff
@@ -589,7 +589,7 @@ impl<'a> RunSummary<'a> {
         if self.monorepo {
             Ok(serde_json::to_string_pretty(&self)?)
         } else {
-            let monorepo_rsm: SinglePackageRunSummary<'a, '_> = (&*self).into();
+            let monorepo_rsm = SinglePackageRunSummary::from(&*self);
```
chris-olszewski, line 592: `&*self` might look weird, but that's just so we can get a `&self` from a `&mut self`

crates/turborepo-lib/src/run/summary/mod.rs
```diff
@@ -272,24 +272,24 @@ impl RunTracker {
 // not be used anywhere else.
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
-struct SinglePackageRunSummary<'a, 'b> {
+struct SinglePackageRunSummary<'a> {
```
chris-olszewski, line 275: Since all of the fields in the `RunSummary` struct are [covariant](https://doc.rust-lang.org/beta/reference/subtyping.html#variance), `RunSummary` itself is covariant and we can simplify the struct to a single lifetime.

TL;DR
This is constructed from a `&'b RunSummary<'a>` and this implies that `'a` outlives `'b` (otherwise we would have a reference to a `RunSummary` that contains references that are no longer valid). In Rust you can always shorten a lifetimes so we shorten `RunSummary<'a>` to `RunSummary<'b>` and are left with a single lifetime.

crates/turborepo-lib/src/run/summary/task.rs
```diff
@@ -74,3 +94,94 @@ impl<'a> TaskSummary<'a> {
...
+    )]
+    #[test_case(
+        TaskCacheSummary {
+            local: true,
+            remote: false,
+            status: CacheStatus::Hit,
+            source: Some(CacheSource::Local),
+            time_saved: 6,
+        },
+        serde_json::json!({
```
chris-olszewski, line 175: This is the expected schema as shown in [run-summary/monorepo.t](https://github.com/vercel/turbo/blob/main/turborepo-tests/integration/tests/run-summary/monorepo.t#L106)

chris-olszewski (commented):
Reviewer's Guide

crates/turborepo-lib/src/run/summary/task.rs
```diff
@@ -74,3 +94,94 @@ impl<'a> TaskSummary<'a> {
...
+            let source = CacheSource::from(source);
+            // Assign these deprecated fields Local and Remote based on the information
+            // available in the itemStatus. Note that these fields are
+            // problematic, because an ItemStatus isn't always the composite
+            // of both local and remote caches. That means that an ItemStatus might say it
+            // was a local cache hit, and we return remote: false here. That's misleading
+            // because it does not mean that there is no remote cache hit,
+            // it _could_ mean that we never checked the remote cache. These
+            // fields are being deprecated for this reason.
+            let (local, remote) = match source {
```
NicholasLYang, line 122: Yeah this setup is problematic for a few reasons. For one, I regret making a cache miss an error instead of just another variant. That's my bad. Second, booleans are kinda misrepresentative because `local: true` and `remote: false` doesn't mean we failed to get a remote cache hit, it just means we didn't even try because the local cache had it (why would you fetch from remote when you can fetch from local?). For 2.0, we should consider changing this schema

NicholasLYang (approved):

crates/turborepo-lib/src/run/summary/mod.rs
mehulkar: maybe leave this comment in code if it's not a normal Rust thing?

mehulkar (approved):

crates/turborepo-lib/src/run/summary/task.rs
chris-olszewski: Most definitely, this logic and comment were copied from Go

PR #6247, opened 2023-10-23
chore(lint): fix no-unused-var violations

mehulkar:
Smaller diff from #6233


docs/components/ExamplesCard/index.tsx
mehulkar, line 1: I don't know why the class name sorting is ending up in this diff , it happens automatically when I save in VS Code. 

I thought it might be prettier, but it doesn't update these things when I run `prettier -w` manually. `eslint --fix` also didn't do reproduce. 

Since it's a pretty small change, I'm inclined to jsut slide this in instead of spending more time on it

NicholasLYang (approved):

PR #6250, opened 2023-10-23
refactor: Moving from anyhow to thiserror part 1

NicholasLYang:
### Description

Part 1 of moving us away from anyhow and to thiserror. We're doing this migration so we can explicitly categorize our errors and eventually provide nicer messages via miette.

This converts a lot of the auxiliary commands like login, link, logout, etc. 

### Testing Instructions

Closes TURBO-1518

### DIFF
--- a/crates/turborepo-auth/src/server/login_server.rs
+++ b/crates/turborepo-auth/src/server/login_server.rs
@@ -6,6 +6,8 @@ use axum::{extract::Query, response::Redirect, routing::get, Router};
 use serde::Deserialize;
 use tokio::sync::OnceCell;
 
+use crate::Error;
+
 #[derive(Debug, Clone, Deserialize)]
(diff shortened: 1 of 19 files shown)


crates/turborepo-auth/src/server/login_server.rs
```diff
@@ -56,6 +51,7 @@ impl LoginServer for DefaultLoginServer {
         Ok(axum_server::bind(addr)
             .handle(handle)
             .serve(app.into_make_service())
-            .await?)
+            .await
+            .expect("failed to start one-shot server"))
```
NicholasLYang, line 55: If we fail to start the one-shot server, it's a fatal error with nothing that we can do. A panic here is acceptable imo

mehulkar (approved):

PR #6251, opened 2023-10-23
refactor: Removed closure from logout

NicholasLYang:
### Description

Removed the closure from logout to simplify the type checking and in preparation for our errors refactor

### Testing Instructions

Closes TURBO-1520

### DIFF
--- a/crates/turborepo-auth/src/auth/logout.rs
+++ b/crates/turborepo-auth/src/auth/logout.rs
@@ -1,16 +1,5 @@
-use anyhow::Result;
-use tracing::error;
-use turborepo_ui::{GREY, UI};
+use turborepo_ui::{cprintln, GREY, UI};
 
-pub fn logout<F>(ui: &UI, mut set_token: F) -> Result<()>
-where
-    F: FnMut() -> Result<()>,
-{
-    if let Err(err) = set_token() {
-        error!("could not logout. Something went wrong: {}", err);
-        return Err(err);
-    }
-
-    println!("{}", ui.apply(GREY.apply_to(">>> Logged out")));
-    Ok(())
+pub fn logout(ui: &UI) {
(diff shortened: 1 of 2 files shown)


crates/turborepo-auth/src/auth/logout.rs
```diff
@@ -1,16 +1,5 @@
...
-{
-    if let Err(err) = set_token() {
-        error!("could not logout. Something went wrong: {}", err);
-        return Err(err);
-    }
-
-    println!("{}", ui.apply(GREY.apply_to(">>> Logged out")));
-    Ok(())
+pub fn logout(ui: &UI) {
+    cprintln!(ui, GREY, ">>> Logged out");
```
NicholasLYang, line 4: This does beg the question of whether this is even worth keeping as a function. cc @Zertsov @mehulkar

mehulkar (commented):
Want to put the brakes on here.. `turborepo-auth` is intended to be a standalone crate that encapsulates the location of the token (_mayyybe_ we add config to change that location, but as of now, we don't have a use case for that). 

The work to do is to decouple the location of the token from the rest of turborepo's config

mehulkar:
To me the closure was at least an indication of "this is where we want the token setting and removal to go". What's the reason for removing it?

crates/turborepo-auth/src/auth/logout.rs
Zertsov: It's fine to keep or remove. Auth / login changes may change this to have more functionality later, but agreed that at the moment it's kind of a lazy function.

mehulkar (approved):
Talked offline, this closure is blocking adoption of `thiserror`, so we'll move ahead with this. We'll talk more about what the final state needs to be (IMO the auth crate needs to control the location of the token)

Issue #2081, opened 2022-09-26
Output error message when remote cache cannot be accessed


jonaslindstr:
### Describe the feature you'd like to request

When a remote cache is configured and cannot be accessed for some reason (network error, wrong configuration, remote cache unavailable etc.) it would be helpful to get an error message indicating this.

Currently you will only get the usual "cache miss" for each workspace. If you have a lot of workspaces there is no clear indication of the problem until you notice that the task takes a very long time, and that you eventually get "0 cached" when the task is finished.

To reproduce the current behaviour: `yarn turbo run TASK --remote-only --api=http://example.com`

### Describe the solution you'd like

An error message early in the run saying "Cannot access remote cache <URL>" or similar.

### Describe alternatives you've considered

I have tried `--output-logs full` and it doesn't give more details about this kind of error.

sppatel:
Agree this would be useful have run into this as well.  To kind of work around this - we've had to introduce a status check that runs hourly that modified and builds a dummy project to verify PUTs and cache GETs are working as expected and there is no misconfiguration, network issues, etc and notifes us when something unexpected has occured.  This periodically has helped catch issues.

mehulkar:
The log would go somewhere in here probably: https://github.com/vercel/turborepo/blob/86eeafc8b14678b89639cb3b6615bd62af1e77fe/cli/internal/cache/cache_http.go#L200-L252. 

Feel free to take a stab at a PR!

DipakHalkude:
Hey I want to contribute this one.

Sakthieswaran-tech:
Hi, I’m interested in contributing and would like to work on this issue.
Could you please confirm if it’s okay for me to pick it up?

PR #6252, opened 2023-10-23
Remote Cache Failure Logs #2081

claytoncasey01:
### Description

Output logs when remote cache fails

Fixes: #2081


claytoncasey01:
@mehulkar Can you take a quick look at this? Ideally this wouldn't be passed down but I couldn't find another way to output the log to the console. This also might pose a problem for the all of the cache tests, none of those will run for me though even when pulling current main, they all fail.

cli/internal/cache/cache_http.go
```diff
@@ -152,9 +175,13 @@ func (cache *httpCache) exists(hash string) (bool, int, error) {
...
 }
 
-func (cache *httpCache) retrieve(hash string) (bool, []turbopath.AnchoredSystemPath, int, error) {
+func (cache *httpCache) retrieve(
+	hash string,
+	prefixedUI *cli.PrefixedUi,
+) (bool, []turbopath.AnchoredSystemPath, int, error) {
 	resp, err := cache.client.FetchArtifact(hash)
 	if err != nil {
+		prefixedUI.Output(err.Error())
```
mehulkar, line 184: Is this the only material change? 

You'll want to do something like this:

```suggestion
		prefixedUI.Error(fmt.Sprintf("error fetching from remote cache: %s", err))
```

mehulkar (commented):
I think this makes sense, but I'm not sure this code is used anymore or if all the caching got moved to Rust already. cc @NicholasLYang @chris-olszewski 

I'll note that the reason this is even necessary is because (at least in Go) errors from `cache_http.go Fetch()` get caught and don't bubbled up.

claytoncasey01:
> I think this makes sense, but I'm not sure this code is used anymore or if all the caching got moved to Rust already. cc @NicholasLYang @chris-olszewski
> 
> I'll note that the reason this is even necessary is because (at least in Go) errors from `cache_http.go Fetch()` get caught and don't bubbled up.

The code seemed to still be hit when running locally against one of my projects that use turbo, all the errors I expected displayed.

chris-olszewski:
> code is used anymore or if all the caching got moved to Rust

This is one of the codepaths that lives as both Go and Rust (when using `--experimental-rust-codepath`).

Due to the nature of this Go codepath going away soon and the fact that this will require changing the API for the Rust cache, I'm hesitant to encourage this until we're down to a single codepath. 

@claytoncasey01 If you can't wait and you've been looking to learn Rust, then I can provide some guidance on making this change on the Rust side.  The entrypoint to where you would want to start making changes would be [TaskCache::restore_outputs](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/run/cache.rs#L167)

claytoncasey01:
> > code is used anymore or if all the caching got moved to Rust
> 
> This is one of the codepaths that lives as both Go and Rust (when using `--experimental-rust-codepath`).
> 
> Due to the nature of this Go codepath going away soon and the fact that this will require changing the API for the Rust cache, I'm hesitant to encourage this until we're down to a single codepath.
> 
> @claytoncasey01 If you can't wait and you've been looking to learn Rust, then I can provide some guidance on making this change on the Rust side. The entrypoint to where you would want to start making changes would be [TaskCache::restore_outputs](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/run/cache.rs#L167)

Thanks, I can take a look into making this on the Rust side, I'm pretty familiar with Rust and use it for a few different side projects. I'm also fine waiting, really just trying to familiarize myself with the Turbo codebase since I utilize turbo in my day job.

mehulkar:
@claytoncasey01 implementing in Rust and Go in the same way would be ok, IMO

anthonyshew:
Closing here since the repo is no longer in Go.

PR #6253, opened 2023-10-23
chore: Added panic handler to turborepo

NicholasLYang:
### Description

Added a simple panic handler using the human_error crate. With a custom error message though.

### Testing Instructions

Closes TURBO-1522

### DIFF
--- a/crates/turborepo/src/main.rs
+++ b/crates/turborepo/src/main.rs
@@ -1,4 +1,7 @@
+#![feature(panic_info_message)]
 #![deny(clippy::all)]
+
+mod panic_handler;
 
 use std::{
     env::{consts, current_exe},
@@ -10,6 +13,8 @@ use anyhow::Result;
 use dunce::canonicalize as fs_canonicalize;
 use tracing::{debug, error, trace};
 use turborepo_lib::{spawn_child, ExecutionState, Payload};
+
+use crate::panic_handler::panic_handler;
 
 fn run_go_binary(execution_state: ExecutionState) -> Result<i32> {
     // canonicalize the binary path to ensure we can find go-turbo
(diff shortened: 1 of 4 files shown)


crates/turborepo/src/panic_handler.rs
```diff
@@ -0,0 +1,47 @@
...
+        .message()
+        .map(ToString::to_string)
+        .unwrap_or_else(|| "Unknown".to_string());
+
+    let explanation = match panic_info.location() {
+        Some(location) => format!("file '{}' at line {}\n", location.file(), location.line()),
+        None => "unknown.".to_string(),
+    };
+
+    let report = Report::new("turbo", get_version(), Method::Panic, explanation, cause);
```
chris-olszewski, line 17: This is an amazing crate

crates/turborepo/src/main.rs
```diff
@@ -66,6 +71,8 @@ fn run_go_binary(execution_state: ExecutionState) -> Result<i32> {
 // This function should not expanded. Please add any logic to
 // `turborepo_lib::main` instead
 fn main() -> Result<()> {
+    std::panic::set_hook(Box::new(panic_handler));
```
chris-olszewski, line 74: Just checking, does `set_hook` propagate to all threads?

chris-olszewski (approved):
This overall looks good to me, just would like some confirmation that we're using the `set_hook` API correctly

crates/turborepo/src/main.rs
NicholasLYang: Yeah, seems like it does. The docs mention that the hook gets called when "a thread panics", so I assume that's any thread

PR #6254, opened 2023-10-23
feat(Turborepo): Check env var indexing by literals

gsoltis:
### Description

Previously, we ignored indexing `process.env` by string literals (`process.env["foo"]`), with this PR we now explicitly check those env vars when they are literals. Computed indices are still ignored.

### Testing Instructions

New tests added. The valid cases pass with and without this change, since we previously ignored them. The invalid cases are newly-passing with this change.

Closes TURBO-1523

### DIFF
--- a/packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
+++ b/packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
@@ -153,13 +153,11 @@ function create(context: RuleContextWithOptions): Rule.RuleListener {
       if (
         "name" in node.object &&
         "name" in node.property &&
-        !isComputed(node)
+        node.object.name === "process" &&
+        node.property.name === "env"
       ) {
-        const objectName = node.object.name;
-        const propertyName = node.property.name;
-
         // we're doing something with process.env
-        if (objectName === "process" && propertyName === "env") {
+        if (!isComputed(node)) {
           // destructuring from process.env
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -177,6 +175,13 @@ function create(context: RuleContextWithOptions): Rule.RuleListener {
           ) {
             checkKey(node.parent, node.parent.property.name);
           }
+        } else if (
+          "property" in node.parent &&
+          node.parent.property.type === "Literal" &&
+          typeof node.parent.property.value === "string"
```
mehulkar, line 181: Does this capture template strings too? I see the `type` in astexplorer is `TemplateLiteral`, but I'm not sure if the eslint parser is the same
 https://astexplorer.net/#/gist/b8b068eb786edac5abd35572cdf791e7/de84c6402467ab65b0b6d71a15f5f320723f655b, so

mehulkar (approved):

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
gsoltis: I don't believe so. We could extend it to capture templates that are only literals if we really care to, but at some point we're running into calculated values that we can't really inspect.

PR #6256, opened 2023-10-23
fix(lint): enable and fix some lint violations

mehulkar:
Closes TURBO-1524

### DIFF
--- a/docs/pages/api/og.tsx
+++ b/docs/pages/api/og.tsx
@@ -57,6 +57,12 @@ async function loadAssets(): Promise<
   ];
 }
 
+// ?type=<pack|repo>
+const TITLE_FOR_TYPE: Record<string, string> = {
+  pack: "The successor to Webpack",
+  repo: "The build system that makes ship happen",
+};
+
 export default async function openGraphImage(
   req: NextApiRequest
 ): Promise<ImageResponse> {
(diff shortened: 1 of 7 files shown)


docs/pages/api/og.tsx
```diff
@@ -57,6 +57,12 @@ async function loadAssets(): Promise<
   ];
 }
 
+// ?type=<pack|repo>
+const TITLE_FOR_TYPE: Record<string, string> = {
+  pack: "The successor to Webpack",
+  repo: "The build system that makes ship happen",
+};
```
chris-olszewski, line 64: I'm no TS-wizard, but does it make sense to restrict the values at the type level?
```suggestion
// ?type=<pack|repo>
const TITLE_FOR_TYPE: Record<"repo" | "pack", string> = {
  pack: "The successor to Webpack",
  repo: "The build system that makes ship happen",
};
```

chris-olszewski (approved):
LGTM

docs/pages/api/og.tsx
mehulkar: I thought about that too, but I don't think it's really worth it. Also later, when you look up `TITLE_FOR_TYPE[type]` it would complain that `type` is `string`, and since it comes off a URLSearchParams, you'd have to go fake that type (since it's essentially user input and a compile time check can't help you)

PR #6258, opened 2023-10-23
fix(Turborepo): Ensure process manager stays closed

gsoltis:
### Description

 - Adds tests for process manager remaining closed after being closed (courtesy of @chris-olszewski )
 - Ensures the process manager does not reopen after `close` has been called, or after `wait` has completed.

### Testing Instructions

Updated tests.

Closes TURBO-1526

### DIFF
--- a/crates/turborepo-lib/src/process/mod.rs
+++ b/crates/turborepo-lib/src/process/mod.rs
@@ -119,7 +119,6 @@ impl ProcessManager {
 
             // just allocate a new vec rather than clearing the old one
             lock.children = vec![];
-            lock.is_closing = false;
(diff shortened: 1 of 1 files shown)


chris-olszewski (changes requested):
I think it makes sense to move the child cleanup behavior change to it's own PR. The API feels a little weird and at the very least it deserves a more involved PR review. I think having unbounded child vector growth is fine since it will always just be the size of all tasks that we execute.

gsoltis:
Dropped the child cleanup piece

chris-olszewski (approved):

PR #6265, opened 2023-10-24
Create and use auth.json

Zertsov:
This moves the auth token out of `config.json` and puts it in a more logical `auth.json` in the same directory as the `config.json`.

Test by logging in - token should be written to `auth.json` and all functionality should be the same.


crates/turborepo-lib/src/cli/error.rs
```diff
@@ -28,11 +28,21 @@ pub enum Error {
         config_path: AbsoluteSystemPathBuf,
         error: io::Error,
     },
+    #[error("Encountered an IO error while attempting to read {auth_path}: {error}")]
+    FailedToReadAuth {
+        auth_path: AbsoluteSystemPathBuf,
+        error: io::Error,
+    },
```
chris-olszewski, line 35: Just confirming that we expect `FailedToReadConfig` and `FailedToReadAuth` to diverge in the immediate future. If not, then I don't think we should copy-pasta the error variant.

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -47,6 +48,7 @@ impl CommandBase {
             args,
             #[cfg(test)]
             global_config_path: None,
+            global_auth_path: None,
```
chris-olszewski, line 51: This field is getting shipped in release if it isn't prefixed with `#[cfg(test)]`
```suggestion
            #[cfg(test)]
            global_auth_path: None,
```
To avoid future mistakes like this, we might want to put these a new struct e.g.:
```
#[cfg(test)]
struct ConfigTestOverrides {
    global_config_path: Option<AbsoluteSystemPathBuf>,
    global_auth_path: Option<AbsoluteSystemPathBuf>,
}
...
#[cfg(test)]
pub overrides: Option<ConfigTestOverrides>
```

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -29,6 +29,7 @@ pub struct CommandBase {
     pub ui: UI,
     #[cfg(test)]
     pub global_config_path: Option<AbsoluteSystemPathBuf>,
+    pub global_auth_path: Option<AbsoluteSystemPathBuf>,
```
chris-olszewski, line 32:
```suggestion
    #[cfg(test)]
    pub global_auth_path: Option<AbsoluteSystemPathBuf>,
```

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -60,7 +62,6 @@ impl CommandBase {
 
     fn config_init(&self) -> Result<ConfigurationOptions, ConfigError> {
         TurborepoConfigBuilder::new(self)
-            // The below should be deprecated and removed.
```
chris-olszewski, line 63: Note for self/other reviewers: #6239

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -84,6 +85,17 @@ impl CommandBase {
         let global_config_path = config_dir.join("turborepo").join("config.json");
         AbsoluteSystemPathBuf::try_from(global_config_path).map_err(ConfigError::PathError)
     }
+    /// Returns the path to the global auth file (auth.json).
+    fn global_auth_path(&self) -> Result<AbsoluteSystemPathBuf, ConfigError> {
+        #[cfg(test)]
+        if let Some(global_auth_path) = self.global_auth_path.clone() {
+            return Ok(global_auth_path);
```
chris-olszewski, line 92: Very much a nit since `None.clone()` doesn't actually allocate any memory, but as a rule of thumb if you can avoid a `clone` call (or any function call which might be costly). Small change that borrows the `Option` and we only `clone` if there's something present.
```suggestion
        if let Some(global_auth_path) = &self.global_auth_path {
            return Ok(global_auth_path.clone());
```

chris-olszewski (changes requested):
If I'm understanding the change correctly upgrading to whatever patch/canary this PR makes it into would invalidate a `turbo login` made by a previous version. If we want this to be breaking in that way, we probably need to provide a warning to make the UX a little better.

Overall I think this is fine, but I think we need to either:
 - provide a happy path for users when the upgrade
 - add warning if we find a token in `config.json` and warn users that they need to `turbo login` again

crates/turborepo-lib/src/cli/error.rs
Zertsov: I think so. I'm setting this up now for that potential in the immediate future, and if at the end of this little adventure we find that they don't really diverge then I'm open to getting rid of it, but I think some of this stuff will be moved around to the auth crate anyway.

crates/turborepo-lib/src/commands/mod.rs
Zertsov: I'll think about how to handle backwards compat and/or if we're ok with invalidating a previous login some more before merging / committing to this in particular.

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -84,6 +85,17 @@ impl CommandBase {
         let global_config_path = config_dir.join("turborepo").join("config.json");
         AbsoluteSystemPathBuf::try_from(global_config_path).map_err(ConfigError::PathError)
     }
+    /// Returns the path to the global auth file (auth.json).
+    fn global_auth_path(&self) -> Result<AbsoluteSystemPathBuf, ConfigError> {
+        #[cfg(test)]
+        if let Some(global_auth_path) = &self.global_auth_path {
+            return Ok(global_auth_path.clone());
+        }
```
mehulkar, line 93: I've seen in the rest of the codebase that the `cfg(test)` attribute is typically on entire functions, rather than blocks inside a function. Should we continue to follow that pattern?

mehulkar (commented):
Agreed with @chris-olszewski , we should probably add some backwards compat code that continues to look for `config.json` and writes it to `auth.json` when `login` is run and then validate it as normal

crates/turborepo-lib/src/commands/mod.rs
Zertsov: There's some other spots where we do this, like in the `TurborepoConfigBuilder`

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -79,201 +79,14 @@ pub async fn login<'a>(
 
 #[cfg(test)]
 mod tests {
-    use std::sync::atomic::AtomicUsize;
```
Zertsov, line 82: To save whoever looks at this some time: I moved all this stuff into `turborepo-auth/src/mocks`

crates/turborepo-auth/src/token/config_token.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    // Add the token to the list of tokens.
+    af.add_token(AuthToken {
+        token: token.to_string(),
+        api: client.base_url().to_owned(),
+        created_at: user_response.user.created_at,
+        teams,
+    });
+
+    // Write this to disk, overwriting anything that may be in `auth.json`.
```
Zertsov, line 87: This is fine because if we're converting a `config.json` into `auth.json`, then there won't _be_ an `auth.json`, but it's still worth noting.

crates/turborepo-auth/src/token/config_token.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use std::fs;
+
+    use tempfile::tempdir;
+
+    use super::*;
+    use crate::mocks::*;
```
Zertsov, line 100: Can I get a 💯 🔥 for not having a crazy huge `impl Client` in the tests module?

crates/turborepo-auth/src/token/mod.rs
```diff
@@ -0,0 +1,89 @@
+/*
+ * Much of the login contained in `config_token` and things related to
+ * conversion of tokens should be removed in the future. This is a stopgap
+ * until we are confident or comfortable with making users re-login.
```
Zertsov, line 4: At some point in the future we'd consider deprecating or removing the config token loading since at said point in the future there shouldn't be any tokens left in `config.json`

crates/turborepo-auth/src/auth_file.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// The token itself.
+    pub token: String,
+    /// The API URL the token was issued from / for.
+    pub api: String,
+}
+
+impl AuthToken {
+    /// Validates the token by checking the expiration date and the signature.
+    pub async fn validate(&self, _client: impl Client) -> bool {
+        todo!("validate token")
```
Zertsov, line 67: This can _probably_ be removed since we use the API to do essential validations

chris-olszewski (changes requested):
Can you `git restore Cargo.lock -s origin/main && cargo build -p turbo` and push the changes to the lockfile? We don't want to update all of those crates as part of this PR.

NicholasLYang (changes requested):

crates/turborepo-auth/src/lib.rs
```diff
@@ -1,11 +1,170 @@
...
+///    convert it to an auth file, then return the contents from it, if
+///    possible. Otherwise return a FailedToReadConfigFile error.
+/// 3) If neither file exists, return an empty auth file and write a blank one
+///    to disk.
+pub async fn read_or_create_auth_file(
+    auth_file_path: &AbsoluteSystemPathBuf,
+    config_file_path: &AbsoluteSystemPathBuf,
+    client: &impl Client,
+) -> Result<AuthFile, Error> {
+    if auth_file_path.exists() {
```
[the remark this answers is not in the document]
Zertsov: Assuming the previous recommendation to pass in non-owned filepath refs as `&AbsoluteSystemPath` (ASP), which doesn't have a `try_exists`, would you want me to implement that on the ASP or keep the owned ASP in the signature?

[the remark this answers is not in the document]
Zertsov: I went ahead and just implemented `try_exists` for `AbsoluteSystemPath`

chris-olszewski (changes requested):
My mostly complete review that I'm flushing.

crates/turborepo-auth/src/auth_file.rs
```diff
@@ -0,0 +1,192 @@
...
+use serde::{Deserialize, Serialize};
+use turbopath::AbsoluteSystemPath;
+use turborepo_api_client::Client;
+
+use crate::Error;
+
+#[derive(Serialize, Deserialize, Debug, Default)]
+/// AuthFile contains a list of domains, each with a token.
+pub struct AuthFile {
+    pub tokens: HashMap<String, String>,
```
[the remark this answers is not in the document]
Zertsov: I can add a function to get them like `pub fn tokens()` and `pub fn tokens_mut`, but maybe it's worth discussing how much functionality to put on the `AuthFile` vs the hashmap of tokens itself. Mostly because logout will need to mutate the tokens, which in the current state just calls the hashmap remove, but I can see an argument for the authfile to have a remove function as well.

crates/turborepo-auth/src/auth_file.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/// Converts our old style of token held in `config.json` into the new schema.
+pub async fn convert_to_auth_token(token: &str, client: &impl Client) -> Result<AuthToken, Error> {
+    let auth_token = AuthToken {
+        token: token.to_string(),
+        api: client.base_url().to_owned(),
+    };
+    Ok(auth_token)
+}
```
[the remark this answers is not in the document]
Zertsov: I think it being async is a remnant from previous iterations, so I'll get this sorted.

crates/turborepo-lib/src/commands/login.rs
```diff
@@ -1,88 +1,228 @@
...
+
+    // Note: Due to config test not being trasitive (cfg(test)), we state the
+    // browser opening strategy in the login server. If we're in a test
+    // environment, don't open a browser.
+    let browser_open_strategy = match cfg!(test) {
+        true => UrlOpenStrategy::Noop,
+        false => UrlOpenStrategy::Real,
+    };
+
+    let login_server = DefaultLoginServer::new(browser_open_strategy);
```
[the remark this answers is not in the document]
Zertsov: ah, I like this approach. I'll hack away at this style and see what comes out

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -667,7 +673,10 @@ mod test {
     async fn test_link_spaces() {
         // user config
         let user_config_file = NamedTempFile::new().unwrap();
-        fs::write(user_config_file.path(), r#"{ "token": "hello" }"#).unwrap();
+
+        // auth file
+        let auth_file = NamedTempFile::new().unwrap();
+        fs::write(auth_file.path(), r#"{ "token": "hello" }"#).unwrap();
```
chris-olszewski, line 679: Isn't this the old token format getting written to the auth file path?

crates/turborepo-auth/src/sso_server.rs
```diff
@@ -22,10 +22,11 @@ pub struct SsoPayload {
 #[async_trait]
 pub trait SSOLoginServer {
     async fn run(&self, port: u16, verification_token: Arc<OnceCell<String>>) -> Result<(), Error>;
+    fn open_web_browser(&self, url: &str) -> std::io::Result<()>;
 }
 
-/// TODO: Document this.
-pub struct DefaultSSOLoginServer;
+/// Basic SSO login server. No configuration required.
+pub struct DefaultSSOLoginServer {}
```
chris-olszewski, line 29: Keep construction simple
```suggestion
pub struct DefaultSSOLoginServer;
```

chris-olszewski (changes requested):
This looks really good. Will take another pass through in the morning with fresh eyes, but so far everything is a minor suggestion.

crates/turborepo-ui/src/lib.rs
```diff
@@ -178,6 +179,21 @@ impl UI {
 
         Cow::Owned(out.join(""))
     }
+
+    /// Display a list of selectable items to the user and returns the index of
+    /// the selected item from the provided items list.
+    pub fn display_selectable_items<T: std::fmt::Display>(
+        &self,
```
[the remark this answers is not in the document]
Zertsov: Might be a misunderstanding on my end, but I felt this made sense to live under a `UI` struct, even if it didn't make use of `self`. However I totally forgot about `should_strip_ansi` so I'll get that rectified.

crates/turborepo-auth/src/auth_file.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    #[tokio::test]
+    async fn test_convert_to_auth_token() {
+        // Setup: Create a mock client and a fake token
+        let mock_client = MockApiClient::new();
+        let token = "test-token";
+        let temp_dir = tempfile::tempdir().unwrap();
+
+        // Create the temp dir files.
+        fs::create_dir_all(temp_dir.path().join(TURBOREPO_CONFIG_DIR)).unwrap();
```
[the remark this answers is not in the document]
Zertsov: Definitely a remnant from when this function was checking the directory instead of raw conversion

crates/turborepo-auth/src/auth_file.rs
```diff
@@ -0,0 +1,201 @@
...
+    #[tokio::test]
+    async fn test_write_to_disk_and_read_back() {
+        // Setup: Use temp dirs to avoid polluting the user's config dir
+        let temp_dir = tempdir().unwrap();
+        let auth_file_path = temp_dir
+            .path()
+            .join(TURBOREPO_CONFIG_DIR)
+            .join(TURBOREPO_AUTH_FILE_NAME);
+
+        // unwrapping is fine because we know the path exists
```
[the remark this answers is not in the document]
Zertsov: makes sense, this was probably a note to myself as to why I was doing what I was doing

crates/turborepo-auth/src/lib.rs
```diff
@@ -1,13 +1,172 @@
...
+            .map_err(|e| Error::FailedToReadAuthFile {
+                source: e,
+                path: auth_file_path.to_owned(),
+            })?;
+        let tokens: TokensContainer = serde_json::from_str(&content)
+            .map_err(|e| Error::FailedToDeserializeAuthFile { source: e })?;
+        let mut auth_file = AuthFile::new();
+        for (api, token) in tokens.tokens {
+            auth_file.insert(api, token);
+        }
```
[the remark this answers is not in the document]
Zertsov: you're not missing anything, I just had a brain fart 🙃

crates/turborepo-lib/src/commands/link.rs
Zertsov: you are correct, missed this

crates/turborepo-lib/src/commands/logout.rs
```diff
@@ -1,38 +1,59 @@
...
-    if let Err(err) = remove_token(base) {
-        error!("could not logout. Something went wrong: {}", err);
-        return Err(err);
+pub async fn logout(base: &mut CommandBase) -> Result<(), Error> {
+    let client = base.api_client()?;
+    let auth_path = base.global_auth_path()?;
+    let config_path = base.global_config_path()?;
+    let mut auth_file = read_or_create_auth_file(&auth_path, &config_path, &client).await?;
+
+    if auth_file.tokens().is_empty() {
```
NicholasLYang, line 12: This is a super tiny nit, but you could do this as a `match`:

```
match auth_file.tokens().len() {
  0 => {
    println!("No tokens to remove");
    return Ok(());
  },
  1 => ...
  _ => ...
}
```

NicholasLYang (approved):
Just some nits but lgtm

crates/turborepo-lib/src/commands/logout.rs
Zertsov: nits help me be more Rust brained than Go brained so in this house we love super nits

chris-olszewski (approved):

PR #6273, opened 2023-10-25
refactor: Moving to thiserror part 2

NicholasLYang:
### Description

Moved all of our errors to thiserror. Also added `#[source]` annotations so we correctly link the underlying error whenever we have one (i.e. if we fail to write to the config, we link to the `io::Error` that caused this failure)

### Testing Instructions

We no longer have `anyhow` as a dependency (only a dev-dependency), so we're guaranteed to be off it.

Closes TURBO-1529


crates/turborepo-lib/src/commands/link.rs
```diff
@@ -308,19 +364,25 @@ pub async fn link(
...
+
+    Confirm::with_theme(&theme)
         .with_prompt(
             "Remote Caching was previously disabled for this team. Would you like to enable it \
              now?",
         )
         .default(true)
-        .interact()?)
+        .interact()
+        .map_err(|_| Error::UserCanceled)
```
chris-olszewski, line 377: Can we keep the underlying `io::Error` as part of this error? [Some Windows terminal emulators](https://code.google.com/archive/p/mintty/issues/56) do not play nicely interactive CLIs and I'd like to provide that error instead of a generic "user canceled" error.

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -360,7 +422,8 @@ fn select_team<'a>(
         .with_prompt(prompt)
         .items(&team_names)
         .default(0)
-        .interact()?;
+        .interact()
+        .map_err(|_| Error::UserCanceled)?;
```
chris-olszewski, line 426: Same comment as above

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -404,18 +467,19 @@ fn select_space<'a>(base: &CommandBase, spaces: &'a [Space]) -> Result<SelectedS
         .with_prompt(prompt)
         .items(&space_names)
         .default(0)
-        .interact()?;
+        .interact()
+        .map_err(|_| Error::UserCanceled)?;
```
chris-olszewski, line 471: Same comment as above

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -424,16 +488,19 @@ fn should_link_remote_cache(base: &CommandBase, location: &str) -> Result<bool>
         base.ui.apply(BOLD.apply_to(CYAN.apply_to(location)))
     );
 
-    Ok(Confirm::new().with_prompt(prompt).interact()?)
+    Confirm::new()
+        .with_prompt(prompt)
+        .interact()
+        .map_err(|_| Error::UserCanceled)
```
chris-olszewski, line 494: Same comment as above

crates/turborepo-lib/src/commands/link.rs
```diff
@@ -442,24 +509,22 @@ fn should_link_spaces(base: &CommandBase, location: &str) -> Result<bool> {
         base.ui.apply(BOLD.apply_to("to Vercel Spaces")),
     );
 
-    Ok(Confirm::new().with_prompt(prompt).interact()?)
+    Confirm::new()
+        .with_prompt(prompt)
+        .interact()
+        .map_err(|_| Error::UserCanceled)
```
chris-olszewski, line 515: Same comment as above

chris-olszewski (approved):
Mostly minor comments around keeping underlying errors around.

I would love if we could get some guidance on:
 - error type naming, `Error` vs `FooError`
 - which modules should have their own `Error` types? Every module? Only public ones and children use the parent `Error`?

NicholasLYang:
For `FooError` vs `Error`, I considered doing this as part of this PR, but it felt a little out of scope. I'm of the view that we can name it `Error` and have any usage outside of the source module (i.e. where the `Error` type is defined) use `foo::Error`

arlyon (approved):

arlyon:
> For `FooError` vs `Error`, I considered doing this as part of this PR, but it felt a little out of scope. I'm of the view that we can name it `Error` and have any usage outside of the source module (i.e. where the `Error` type is defined) use `foo::Error

The various error types can often help locate crate boundaries depending on how granular we want to be and usually one crate will have just one Error. I am ok with doing this for now even if we don't split those logical parts out.

PR #6275, opened 2023-10-25
feat: add package discovery trait and a few discovery strategies

arlyon:
### Description

This adds a basic trait for discovering packages (paths to package and turbo jsons) using a variety of strategies:

- **local**: runs in-process and does a recursive walk of the FS
- **watch**: runs a file watching thread maintains a list in memory
- **daemon**: makes a request to a remote service over GRPC
- **timeout**: attempts a primary strategy and falls back in the event of errors or a timeout

This PR still needs some tests for the strategies, but it is ready for a basic look.

### Testing Instructions

There is one integration test for the daemon strategy which requires a daemon to be running. It will be improved when ready to be merged.

Closes TURBO-1530

### DIFF
--- a/crates/turborepo-filewatch/src/package_watcher.rs
+++ b/crates/turborepo-filewatch/src/package_watcher.rs
@@ -1,0 +1,309 @@
+//! This module hosts the `PackageWatcher` type, which is used to watch the
+//! filesystem for changes to packages.
+
+use std::{
+    collections::HashMap,
+    future::IntoFuture,
+    path::PathBuf,
+    sync::{Arc, Mutex},
+};
+
+use notify::Event;
+use tokio::{
+    join,
+    sync::{
+        broadcast::{self, error::RecvError},
+        oneshot, watch,
+    },
+};
+use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
+use turborepo_repository::{
+    discovery::{PackageDiscovery, WorkspaceData},
+    package_manager::{self, Error, PackageManager, WorkspaceGlobs},
+};
+
+use crate::NotifyError;
+
+/// Watches the filesystem for changes to packages and package managers.
+pub struct PackageWatcher {
+    // _exit_ch exists to trigger a close on the receiver when an instance
+    // of this struct is dropped. The task that is receiving events will exit,
+    // dropping the other sender for the broadcast channel, causing all receivers
+    // to be notified of a close.
+    _exit_tx: oneshot::Sender<()>,
+    _handle: tokio::task::JoinHandle<()>,
+
+    package_data: Arc<Mutex<HashMap<AbsoluteSystemPathBuf, WorkspaceData>>>,
+    manager_rx: watch::Receiver<PackageManager>,
+}
+
+impl PackageWatcher {
+    /// Creates a new package watcher whose current package data can be queried.
+    pub async fn new<T: PackageDiscovery + Send + 'static>(
+        root: AbsoluteSystemPathBuf,
+        recv: broadcast::Receiver<Result<Event, NotifyError>>,
+        backup_discovery: T,
+    ) -> Result<Self, package_manager::Error> {
+        let (exit_tx, exit_rx) = oneshot::channel();
+        let subscriber = Subscriber::new(exit_rx, root, recv, backup_discovery).await?;
+        let manager_rx = subscriber.manager_receiver();
+        let package_data = subscriber.package_data();
+        let handle = tokio::spawn(subscriber.watch());
+        Ok(Self {
+            _exit_tx: exit_tx,
+            _handle: handle,
+            package_data,
+            manager_rx,
+        })
+    }
+
+    pub async fn get_package_data(&self) -> Vec<WorkspaceData> {
+        self.package_data
+            .lock()
+            .expect("not poisoned")
+            .clone()
+            .into_values()
+            .collect()
+    }
+
+    pub async fn get_package_manager(&self) -> PackageManager {
+        *self.manager_rx.borrow()
+    }
+}
+
+/// The underlying task that listens to file system events and updates the
+/// internal package state.
+struct Subscriber<T: PackageDiscovery> {
+    exit_rx: oneshot::Receiver<()>,
+    filter: WorkspaceGlobs,
+    recv: broadcast::Receiver<Result<Event, NotifyError>>,
+    repo_root: AbsoluteSystemPathBuf,
+    backup_discovery: T,
+
+    // package manager data
+    package_data: Arc<Mutex<HashMap<AbsoluteSystemPathBuf, WorkspaceData>>>,
+    manager_rx: watch::Receiver<PackageManager>,
+    manager_tx: watch::Sender<PackageManager>,
+
+    // stored as PathBuf to avoid processing later.
+    // if package_json changes, we need to re-infer
+    // the package manager
+    package_json_path: std::path::PathBuf,
+    workspace_config_path: std::path::PathBuf,
+}
+
+impl<T: PackageDiscovery + Send + 'static> Subscriber<T> {
+    async fn new(
+        exit_rx: oneshot::Receiver<()>,
+        repo_root: AbsoluteSystemPathBuf,
+        recv: broadcast::Receiver<Result<Event, NotifyError>>,
+        mut discovery: T,
+    ) -> Result<Self, Error> {
+        let manager = discovery.discover_packages().await.unwrap().package_manager;
+
+        let (package_json_path, workspace_config_path, filter) =
+            Self::update_package_manager(&manager, &repo_root)?;
+
+        let (manager_tx, manager_rx) = watch::channel(manager);
+
+        Ok(Self {
+            exit_rx,
+            filter,
+            package_data: Default::default(),
+            recv,
+            manager_rx,
+            manager_tx,
+            repo_root,
+            backup_discovery: discovery,
+
+            package_json_path,
+            workspace_config_path,
+        })
+    }
+
+    async fn watch(mut self) {
+        // initialize the contents
+        self.rediscover_packages().await;
+
+        // respond to changes
+        loop {
+            tokio::select! {
+                _ = &mut self.exit_rx => return,
+                file_event = self.recv.recv().into_future() => match file_event {
+                    Ok(Ok(event)) => self.handle_file_event(event).await,
+                    // if we get an error, we need to re-discover the packages
+                    Ok(Err(_)) => self.rediscover_packages().await,
+                    Err(RecvError::Closed) => return,
+                    // if we end up lagging, warn and rediscover packages
+                    Err(RecvError::Lagged(count)) => {
+                        tracing::warn!("lagged behind {count} processing file watching events");
+                        self.rediscover_packages().await;
+                    },
+                }
+            }
+        }
+    }
+
+    fn update_package_manager(
+        manager: &PackageManager,
+        repo_root: &AbsoluteSystemPath,
+    ) -> Result<(PathBuf, PathBuf, WorkspaceGlobs), Error> {
+        let package_json_path = repo_root
+            .join_component("package.json")
+            .as_std_path()
+            .to_owned();
+
+        let workspace_config_path = manager
+            .workspace_configuration_path()
+            .map_or(package_json_path.clone(), |p| {
+                repo_root.join_component(p).as_std_path().to_owned()
+            });
+        let filter = manager.get_workspace_globs(repo_root)?;
+
+        Ok((package_json_path, workspace_config_path, filter))
+    }
+
+    pub fn manager_receiver(&self) -> watch::Receiver<PackageManager> {
+        self.manager_rx.clone()
+    }
+
+    pub fn package_data(&self) -> Arc<Mutex<HashMap<AbsoluteSystemPathBuf, WorkspaceData>>> {
+        self.package_data.clone()
+    }
+
+    fn manager(&self) -> PackageManager {
+        *self.manager_rx.borrow()
+    }
+
+    async fn handle_file_event(&mut self, file_event: Event) {
+        tracing::trace!("file event: {:?}", file_event);
+
+        if file_event
+            .paths
+            .iter()
+            .any(|p| self.package_json_path.eq(p))
+        {
+            tracing::debug!("package.json changed");
+            // if the package.json changed, we need to re-infer the package manager
+            // and update the glob list
+            let new_manager =
+                PackageManager::get_package_manager(&self.repo_root, None).and_then(|manager| {
+                    Self::update_package_manager(&manager, &self.repo_root)
+                        .map(|(a, b, c)| (manager, a, b, c))
+                });
+
+            match new_manager {
+                Ok((new_manager, package_json_path, workspace_config_path, filter)) => {
+                    // if this fails, we are closing anyways so ignore
+                    self.manager_tx.send(new_manager).ok();
+                    self.package_json_path = package_json_path;
+                    self.workspace_config_path = workspace_config_path;
+                    self.filter = filter;
+                }
+                Err(e) => {
+                    // a change in the package json does not necessarily mean
+                    // that the package manager has changed, so continue with
+                    // best effort
+                    tracing::error!("error getting package manager: {}", e);
+                }
+            }
+        }
+
+        // if it is the package manager, update the glob list
+        let changed = if file_event
+            .paths
+            .iter()
+            .any(|p| self.workspace_config_path.eq(p))
+        {
+            let new_filter = self
+                .manager()
+                .get_workspace_globs(&self.repo_root)
+                // under some saving strategies a file can be totally empty for a moment
+                // during a save. these strategies emit multiple events and so we can
+                // a previous or subsequent event in the 'cluster' will still trigger
+                .unwrap_or_else(|_| self.filter.clone());
+
+            let changed = self.filter != new_filter;
+            self.filter = new_filter;
+            changed
+        } else {
+            false
+        };
+
+        if changed {
+            // if the glob list has changed, do a recursive walk and replace
+            self.rediscover_packages().await;
+        } else {
+            // if a path is not a valid utf8 string, it is not a valid path, so ignore
+            for path in file_event
+                .paths
+                .iter()
+                .filter_map(|p| p.as_os_str().to_str())
+            {
+                let path_file =
+                    AbsoluteSystemPathBuf::new(path).expect("watched paths are absolute");
+
+                // the path to the workspace this file is in is the parent
+                let path_workspace = path_file
+                    .parent()
+                    .expect("watched paths will not be at the root")
+                    .to_owned();
+
+                let is_workspace = match self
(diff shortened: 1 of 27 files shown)


crates/turborepo-lib/src/package_graph/builder.rs
```diff
@@ -65,9 +69,12 @@ impl<'a> PackageGraphBuilder<'a> {
             package_manager: None,
             package_jsons: None,
             lockfile: None,
+            package_discovery: LocalPackageDiscovery::new(repo_root.to_owned()),
```
arlyon, line 72: I am going to rework this such that this happens at build time.

crates/turborepo-discovery/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub package_json: AbsoluteSystemPathBuf,
+    pub turbo_json: Option<AbsoluteSystemPathBuf>,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("discovery unavailable")]
+    Unavailable,
+    #[error("discovery failed")]
+    Failed,
```
arlyon, line 30: Planning on hashing out a better plan for this. These larger components failing are probably something we want to track somehow. Will speak to @NicholasLYang about plans for this.

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -0,0 +1,215 @@
...
+            tokio::select! {
+                _ = &mut self.exit_rx => return,
+                file_event = self.recv.recv().into_future() => match file_event {
+                    Ok(Ok(event)) => self.handle_file_event(event).await,
+                    // if we get an error, we need to re-discover the packages
+                    Ok(Err(_)) => self.rediscover_packages().await,
+                    Err(RecvError::Closed) => return,
+                    // if we end up lagging, just skip the event
+                    Err(RecvError::Lagged(count)) => {
+                        tracing::warn!("lagged behind {count} processing file watching events");
```
gsoltis, line 106: I think we should consider this the same as receiving a filewatching error and re-discover packages.

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -0,0 +1,215 @@
...
+                    .to_owned();
+
+                let is_workspace = match self
+                    .filter
+                    .target_is_workspace(&self.repo_root, &path_workspace)
+                {
+                    Ok(is_workspace) => is_workspace,
+                    Err(e) => {
+                        tracing::error!("error checking if path is in workspace: {:?}", e);
+                        continue;
```
gsoltis, line 166: What can cause this error, and what are the implications of ignoring it? Are we going to have stale workspaces? Should we stop watching workspaces if we can't recover?

crates/turborepo-lib/src/package_graph/builder.rs
```diff
@@ -54,9 +56,11 @@ pub enum Error {
     Graph(#[from] graph::Error),
     #[error(transparent)]
     Lockfile(#[from] turborepo_lockfiles::Error),
+    #[error(transparent)]
+    Discovery(#[from] turborepo_discovery::Error),
 }
 
-impl<'a> PackageGraphBuilder<'a> {
+impl<'a> PackageGraphBuilder<'a, ()> {
```
chris-olszewski, line 63: I think we can avoid the `build_default`/`build` bifurcation if we eagerly providing the default discovery strategy of `LocalPackageDiscovery`. It does an allocation for cloning the repo root, but IMO that's totally fine since this gets called once per run.

crates/turborepo-lib/src/package_graph/builder.rs
```diff
@@ -65,9 +69,12 @@ impl<'a> PackageGraphBuilder<'a> {
             package_manager: None,
             package_jsons: None,
             lockfile: None,
+            package_discovery: (), // unset
```
chris-olszewski, line 72:
```suggestion
            package_discovery: LocalPackageDiscovery::new(repo_root.to_owned()),
```

chris-olszewski (commented):
Quickly flushing comments before I'm pulled away

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -0,0 +1,215 @@
...
+                    }
+                };
+
+                if is_workspace {
+                    tracing::debug!("tracing file in package: {:?}", path_file);
+                    let mut data = self.package_data.lock().await;
```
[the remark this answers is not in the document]
arlyon: We do not. I think that was a holdover from a previous iteration. All sorted, TY

crates/turborepo-filewatch/src/package_watcher.rs
arlyon: Added a comment to explain and adjusted the log line.

crates/turborepo-lib/src/package_graph/builder.rs
arlyon: My worry with this was that we would potentially do package manager detection twice if you got one by default then overwrite it. I think the answer is to just hoist the package manager up out of the strategy and have the caller pass it in as a required argument.

arlyon: Thought we might be able to do a blanket impl on any type that implements TryInto<T> where T implements PackageDiscovery but in retrospect that T is an unconstrained type param so that was also a no-go.

I may leave this as-is despite the API bifurcation.

chris-olszewski: > My worry with this was that we would potentially do package manager detection twice if you got one by default then overwrite it

Ah yeah, I see now that local discovery requires the package manager. 

Maybe the local discovery constructor should take a "package manager factory" i.e. `impl FnOnce() -> Result<PackageManager, Error>`. It would allow for avoiding double detection via `LocalPackageDiscovery::new(repo_root, || Ok(package_manger)`. It also allows for `FallbackPackageDiscovery` to be constructed without accidentally performing a double discovery: `FallbackPackageDiscovery::new(DaemonPackageDiscovery::new(client), LocalPackageDiscovery::new(repo_root, || PackageManger::get_package_manager(repo_root, None))`

crates/turborepo-discovery/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+
+#![feature(async_fn_in_trait)]
+#![feature(return_position_impl_trait_in_trait)]
+
+use tokio_stream::{iter, StreamExt};
+use turbopath::AbsoluteSystemPathBuf;
+use turborepo_repository::package_manager::{self, PackageManager};
+
+#[derive(Clone)]
+pub struct RootWorkspaceData {
```
chris-olszewski, line 20: How do you feel about including `PackageManager` as part of this data? I think it'll be very hard to avoid performing detection twice if we don't shore this information up here. `PackageGraph` will always need to know `PackageManager` even if we can get all of the workspaces discovered via the daemon we would still need to `PackageManager::get_package_manager`.

I think it makes sense that if workspaces haven't changed that means the package manager hasn't either.

Also, I'm so sorry, but I realized that this isn't root workspace and just workspace data 🤦

crates/turborepo-lib/src/package_graph/builder.rs
```diff
@@ -94,22 +101,58 @@ impl<'a> PackageGraphBuilder<'a> {
...
+    }
+}
+
+impl<'a> PackageGraphBuilder<'a, ()> {
+    /// Build the `PackageGraph` with the default package discovery strategy.
     #[tracing::instrument(skip(self))]
-    pub fn build(self) -> Result<PackageGraph, Error> {
+    pub async fn build_default(self) -> Result<PackageGraph, Error> {
+        let builder = PackageGraphBuilder {
+            package_discovery: LocalPackageDiscovery::new(self.repo_root.to_owned())?,
```
chris-olszewski, line 126: I believe we're already performing double package manager discovery with this call and then the `PackageManager::get_package_manager` call in `BuildState::new`

chris-olszewski (commented):
Again flushing out some comments. 

I feel fairly strongly about moving `PackageManager` into the data that we shore up as part of the workspace discovery trait since knowing the package manager is required before we can start discovering workspaces.

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -0,0 +1,219 @@
...
+                        tracing::warn!("lagged behind {count} processing file watching events");
+                        self.rediscover_packages().await;
+                    },
+                }
+            }
+        }
+    }
+
+    async fn handle_file_event(&mut self, file_event: Event) {
+        let glob_source_change = self.manager.workspace_glob_source(&self.repo_root);
```
chris-olszewski, line 119: We should probably also watch `package.json` always as someone could alter `packageManager` where that causes us to invalidate `self.manager`. Maybe we also only use this strategy if that's set as it would be very confusing what to do if while we're watching a `pnpm-lock.yaml` and `yarn.lock` get created.

crates/turborepo-repository/src/package_manager/mod.rs
```diff
@@ -351,6 +355,18 @@ impl PackageManager {
...
+    pub fn workspace_glob_source(&self, root_path: &AbsoluteSystemPath) -> AbsoluteSystemPathBuf {
+        match self {
+            PackageManager::Pnpm | PackageManager::Pnpm6 => {
+                root_path.join_component("pnpm-workspace.yaml")
+            }
+            PackageManager::Berry
+            | PackageManager::Yarn
+            | PackageManager::Npm
+            | PackageManager::Bun => root_path.join_component("package.json"),
+        }
```
chris-olszewski, line 367: This is partially duplicated code from `workspace_configuration_path`
```suggestion
       root_path.join_component(self.workspace_configuration_path().unwrap_or("package.json")
```

chris-olszewski (changes requested):
I think there's a little work to do on the workspace discovery trait interface that'll change `PackageGraphBuilder` for the better and make it easier to avoid accidentally detecting the package manager multiple times.

I think there are a few edge cases around what happens if a user changes package manager out from under us that need some work.

Overall, this is amazing and I'm very excited about going down this path. With the new trait I think we'll be able to eventually get rid of some of the `with_` functions off of the `PackageGraphBuilder` that are just used for tests and just throw a `with_package_discover(MockWorkspaceDiscovery)` which feels right.

crates/turborepo-discovery/src/lib.rs
arlyon: I think package manager makes sense.

crates/turborepo-lib/src/package_graph/builder.rs
arlyon: Ok, after a bit of thinking, best strategy imo is to have the package manager be a required input, that way we can create the LocalPackageDiscovery without wasting time looking for the manager.

crates/turborepo-repository/src/discovery.rs
```diff
@@ -0,0 +1,123 @@
...
+
+use tokio_stream::{iter, StreamExt};
+use turbopath::AbsoluteSystemPathBuf;
+
+use crate::package_manager::PackageManager;
+
+#[derive(Clone)]
+pub struct WorkspaceData {
+    pub package_json: AbsoluteSystemPathBuf,
+    pub package_manager: PackageManager,
```
chris-olszewski, line 20: This feels slightly off that `package_manager` is part of `WorkspaceData`. Is there ever a reason where a call to `discover_packages` would return a vec with differing `package_manager`s?

crates/turborepo-repository/src/package_graph/builder.rs
```diff
@@ -15,18 +15,20 @@ use turborepo_lockfiles::Lockfile;
...
     package_manager::PackageManager,
 };
 
-pub struct PackageGraphBuilder<'a> {
+pub struct PackageGraphBuilder<'a, T> {
     repo_root: &'a AbsoluteSystemPath,
     root_package_json: PackageJson,
     is_single_package: bool,
-    package_manager: Option<PackageManager>,
+    package_manager: PackageManager,
```
chris-olszewski, line 28: Not blocking, but now that this information is returned from `PackageDiscovery::discover_packages` it feels like `package_manager` should come from there instead of being passed in.

crates/turborepo-repository/src/package_graph/builder.rs
```diff
@@ -146,22 +180,19 @@ impl<'a, S> BuildState<'a, S> {
     }
 }
 
-impl<'a> BuildState<'a, ResolvedPackageManager> {
+impl<'a, T: PackageDiscovery> BuildState<'a, ResolvedPackageManager, T> {
```
chris-olszewski, line 183: I see a future where the `ResolvedPackageManager` state is no more and we go straight to the `ResolvedWorkspaces`. Not yet, but this lays the groundwork for that simplification.

crates/turborepo-lib/src/run/package_discovery/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+                .unwrap(),
+            sock_file: AbsoluteSystemPathBuf::new("/tmp/turbod/6c5948bd4171b931/turbod.sock")
+                .unwrap(),
+        };
+
+        let mut client = connector.connect().await.unwrap();
+
+        let packages = client.discover_packages().await.unwrap();
+
+        println!("{:#?}", packages);
```
chris-olszewski, line 95: Should we be performing an assertion here instead of printing them?

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -175,10 +181,40 @@ impl<'a> Run<'a> {
...
+            info!("skipping turbod since we appear to be in a non-interactive context");
+        } else if !opts.run_opts.no_daemon {
+            let connector = DaemonConnector {
+                can_start_server: true,
+                can_kill_server: true,
+                pid_file: self.base.daemon_file_root().join_component("turbod.pid"),
+                sock_file: self.base.daemon_file_root().join_component("turbod.sock"),
+            };
+
+            let client = connector.connect().await?;
```
chris-olszewski, line 198: I believe this will show up in the merge-conflict fixup, but this shouldn't be an error and on main it no longer is.

chris-olszewski (changes requested):
Primarily requesting changes for fixing compile errors fixing up merge conflict.

> There is one integration test for the daemon strategy which requires a daemon to be running. It will be improved when ready to be merged.

Now that this is closer to getting merged do we have some options for how to test these changes? Primarily concerned with:
 - Various discovery strategies
 - Fallback behavior

crates/turborepo-repository/src/discovery.rs
```diff
@@ -0,0 +1,123 @@
...
+    #[error("discovery failed")]
+    Failed,
+}
+
+/// Defines a strategy for discovering packages on the filesystem.
+pub trait PackageDiscovery {
+    // desugar to assert that the future is Send
+    fn discover_packages(
+        &mut self,
+    ) -> impl std::future::Future<Output = Result<Vec<WorkspaceData>, Error>> + Send;
```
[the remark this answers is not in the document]
arlyon: Yes. We have since updated our toolchain so no annotation needed.

crates/turborepo-repository/src/package_graph/builder.rs
```diff
@@ -245,49 +283,56 @@ impl<'a> BuildState<'a, ResolvedPackageManager> {
...
             node_lookup,
             lockfile,
+            mut package_discovery,
             ..
         } = self;
+
+        let package_manager = package_discovery
+            .discover_packages()
+            .await
+            .unwrap()
```
chris-olszewski, line 321: It seems that this is causing a panic in the integration tests.

chris-olszewski (changes requested):
Integration test failures seem legitimate.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -233,6 +213,43 @@ impl<'a> Run<'a> {
...
+            PackageGraph::builder(&self.base.repo_root, root_package_json.clone())
+                .with_single_package_mode(opts.run_opts.single_package)
+                .with_package_discovery(FallbackPackageDiscovery::new(
+                    daemon.as_mut().map(DaemonPackageDiscovery::new),
+                    LocalPackageDiscoveryBuilder::new(
+                        self.base.repo_root.clone(),
+                        None,
+                        Some(root_package_json.clone()),
+                    )
+                    .build()?,
```
chris-olszewski, line 226: Shouldn't be changed in this PR as we want to get this merged, but maybe a leave a `TODO` to allow for a `FallbackPackageDiscovery` to take a `PackageDiscoveryBuilder` as a fallback strategy that only gets built if needed.

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -0,0 +1,633 @@
...
+        })
+    }
+
+    async fn watch(mut self) {
+        // initialize the contents
+        self.rediscover_packages().await;
+
+        // respond to changes
+        loop {
+            tokio::select! {
```
chris-olszewski, line 130: Should we make this `biased` to make sure we close over handling file changes?

chris-olszewski (approved):
I feel like I've said this a few times, but this some awesome perf work.

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -179,35 +184,10 @@ impl<'a> Run<'a> {
...
-            .and_then(|configuration_options| configuration_options.signature)
-            .unwrap_or_default();
-
-        opts.cache_opts.remote_cache_opts = Some(RemoteCacheOpts::new(team_id, signature));
-
-        if opts.run_opts.experimental_space_id.is_none() {
-            opts.run_opts.experimental_space_id = root_turbo_json.space_id.clone();
-        }
-
+        // There's some warning handling code in Go that I'm ignoring
```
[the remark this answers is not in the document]
arlyon: Oops I think that was a rebase artifact

crates/turborepo-filewatch/src/package_watcher.rs
arlyon: yeah to be honest with two branches I feel like that should be the default...

crates/turborepo-repository/src/discovery.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+}
+
+impl PackageDiscovery for LocalPackageDiscovery {
+    async fn discover_packages(&mut self) -> Result<DiscoveryResponse, Error> {
+        let packages = match self.package_manager.get_package_jsons(&self.repo_root) {
+            Ok(packages) => packages,
+            // if there is not a list of workspaces, it is not necessarily an error. just report no
+            // workspaces
+            Err(package_manager::Error::Workspace(_)) => {
```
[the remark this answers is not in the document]
arlyon: We treat this error similarly here https://github.com/vercel/turbo/blob/fea91dade304451f4a23c756de10a4433af039ca/crates/turborepo-lib/src/run/global_hash.rs#L78 so lets handle that in a new PR.

PR #6277, opened 2023-10-25
feat(docs): document support policy

mehulkar:
Closes TURBO-1532


docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+description: Our Support Policy for Turborepo and its ecosystem
+headeronly: true
+container: true
+searchable: false
+---
+
+## Support Policy
+
+`turbo` is generally supported on Ubuntu, macOS, and Windows on both x86_64 and ARM 64 architectures.
+Specificaly, we build and ship the following binaries via `npm`:
```
anthonyshew, line 11: typo: Specifically

docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+The darwin builds are intended for macOS on Intel and Apple Silicon. The linux
+builds are intended to run on Ubuntu, but we do not exercise them on other
+distros. The windows builds are tested on the [`windows-latest` runner][2] in
+Github Actions.
+
+## Node.js Compatibility
+
+Most core `turbo` functionality (especially `turbo run`) does not depend on the
+active Node.js version on your system, but som features of Turborepo and its ecosystem
```
anthonyshew, line 28: typo: some

anthonyshew (approved):
Just a couple typos! LGTM.

docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+- turbo-darwin-64
+- turbo-darwin-arm64
+- turbo-linux-64
+- turbo-linux-arm64
+- turbo-windows-64
+- turbo-windows-arm64
+
+The darwin builds are intended for macOS on Intel and Apple Silicon. The linux
+builds are intended to run on Ubuntu, but we do not exercise them on other
```
NicholasLYang, line 21: Ehh not quite true. We do make sure it works for Alpine because that's what we use at Vercel

docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+distros. The windows builds are tested on the [`windows-latest` runner][2] in
+Github Actions.
```
NicholasLYang, line 23:
```suggestion
GitHub Actions.
```

docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+Core `turbo` functionality depends on the package managers in the JS ecosystem
+and their implementations of workspace configuration (in monorepos) and
+lockfiles formats. We intend to support:
+
+- `npm` (v6+)
+- `yarn` (v1+)
+- `pnpm` (v6+)
+- `bun` (v1+)
+
+It should be noted that Package Managers themselves have their own release
```
NicholasLYang, line 43: nit: this doesn't need to be capitalized

docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+Some exceptions to this may be:
+
+- Configuration changes in `turbo.json`.
+
+  These should usually be accompanied by a codemod that you can run with
+  `npx @turbo/codemd update` and with at least one minor release that includes a
+  deprecation message.
+
+- Intentional behavioral changes to of `turbo`'s CLI commands.
```
NicholasLYang, line 60:
```suggestion
- Intentional behavioral changes to `turbo`'s CLI commands.
```

docs/pages/repo/docs/support.mdx
mehulkar, line 11:
```suggestion
Specifically, we build and ship the following binaries via `npm`:
```

docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+The darwin builds are intended for macOS on Intel and Apple Silicon. The linux
+builds are intended to run on Ubuntu, but we do not exercise them on other
+distros. The windows builds are tested on the [`windows-latest` runner][2] in
+GitHub Actions.
+
+## Node.js Compatibility
+
+Most core `turbo` functionality (especially `turbo run`) does not depend on the
+active Node.js version on your system, but som features of Turborepo and its ecosystem
```
mehulkar, line 28:
```suggestion
active Node.js version on your system, but some features of Turborepo and its ecosystem
```

docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+Core `turbo` functionality depends on the package managers in the JS ecosystem
+and their implementations of workspace configuration (in monorepos) and
+lockfiles formats. We intend to support:
+
+- `npm` (v6+)
+- `yarn` (v1+)
+- `pnpm` (v6+)
+- `bun` (v1+)
+
+It should be noted that Package Managers themselves have their own release
```
mehulkar, line 43:
```suggestion
It should be noted that package managers themselves have their own release
```

docs/pages/repo/docs/support.mdx
mehulkar: yeah i wanted to add some more details here and could use help:

- [`turbo/bin` uses `process.platform`][1] to determine which native binary to download
- [`process.platform`][2] can be 'aix', 'darwin', 'freebsd', 'linux', 'openbsd', 'sunos', 'win32`. but we only build a binary for darwin/linux/win32 ones.

I'm guessing `process.platform` is `"linux"` on Alpine, but are there other Linux distros that we intend to capture with this? 

Also, we only build `turbo-linux-64` and `turbo-linux-arm64`, are there any other [`process.arch`][3] that we should support?

[1]: https://github.com/vercel/turbo/blob/51bba18db76ac3f0b60bab9dfc823a6e066642b8/packages/turbo/bin/turbo#L69-L81
[2]: https://nodejs.org/api/process.html#processplatform
[3]: https://nodejs.org/api/process.html#processarch

NicholasLYang: Hmm I'm not exactly sure which Linux distros we support (any idea @gsoltis?). I suspect definitively Alpine and Debian/Ubuntu but probably not much more.

As for arch, I think x86-64 and aarch64 are good enough. Realistically once we move to full Rust, someone could build turbo from source with little effort (although `ring` will give them trouble because it uses platform specific code or even inline assembly)

mehulkar: I just ran into someone on twitter saying Alpine doesn't work out of box https://twitter.com/perdanafamul/status/1717507938229837871

English translation from Google:
> But this turborepo doesn't support Alpine, bro, I read the discussions in the issue thread, so what package do you need to install first? I forgot.

mehulkar: updated this to soemthing more general for now

docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+`turbo` is generally supported on Debian-based Linux distributions, macOS, and
+Windows on both x86_64 and ARM 64 architectures. Specifically, we build and ship
+the following binaries via `npm`:
+
+- turbo-darwin-64 (macOS with Intel chips)
+- turbo-darwin-arm64 (macOS with Apple Silicon)
+- turbo-linux-64
+- turbo-linux-arm64
+- turbo-windows-64
+- turbo-windows-arm64
```
gsoltis, line 17: We don't actually build arm64 binaries for windows right now, although we will once we're done with Go. Windows on arm ships with an emulation layer to run x86_64 binaries, so we ship the same binary as x86_64 windows and just call it arm

docs/pages/repo/docs/support.mdx
```diff
@@ -0,0 +1,66 @@
...
+## Package Managers
+
+Core `turbo` functionality depends on the package managers in the JS ecosystem
+and their implementations of workspace configuration (in monorepos) and
+lockfiles formats. We intend to support:
+
+- `npm` (v6+)
+- `yarn` (v1+)
+- `pnpm` (v6+)
+- `bun` (v1+)
```
gsoltis, line 40: We don't support `prune` for `bun`. Maybe it needs an asterisk?

gsoltis (approved):
Just a couple notes, otherwise LGTM

docs/pages/repo/docs/support.mdx
mehulkar: good call,updated

docs/pages/repo/docs/support.mdx
mehulkar: i'll leave this juicy morsel out of the docs, since this reflects our aspirational "support" anyway

mehulkar: not sure we need to list the binaries at all, but maybe will change in a future pass

PR #6278, opened 2023-10-25
Clearer cache spinner message.

anthonyshew:
### Description

Our message on the Remote Cache "spinner" was giving users the impression that we were waiting for all tasks to finish to begin uploading. Updating the text for better clarity!


gsoltis (approved):

cli/internal/run/real_run.go
```diff
@@ -158,7 +158,7 @@ func RealRun(
 	}
 
 	defer func() {
-		_ = spinner.WaitFor(ctx, turboCache.Shutdown, base.UI, "...writing to cache...", 1500*time.Millisecond)
+		_ = spinner.WaitFor(ctx, turboCache.Shutdown, base.UI, "...Finishing Remote Cache upload...", 1500*time.Millisecond)
```
mehulkar, line 161: fwiw, `turbocache.Shutdown` is closing out _all_ caches, so in theory, it _could_ be closing out writing to a local cache also. so the message scoping down to Remote cache is not entirely correct.

anthonyshew: Ah, gotcha. Thanks!

PR #6284, opened 2023-10-25
fix: don't escape html when serializing dry runs

chris-olszewski:
### Description

I noticed that some of our integration tests contained escaped characters while working on https://github.com/vercel/turbo/pull/6282

We now no longer escape HTML special characters so we don't mangle some common shell operations

### Testing Instructions

Updated integration tests

Closes TURBO-1536


mehulkar (approved):
ty for tracking this down!

PR #6293, opened 2023-10-26
feat(Turborepo): Validation checks for task names

gsoltis:
### Description

 - validates that tasks have sane names
 - currently only checks against our special escape sequence, `$colon$`

### Testing Instructions

Added new tests for invalid task names.

Closes TURBO-1538

### DIFF
--- a/crates/turborepo-lib/src/engine/builder.rs
+++ b/crates/turborepo-lib/src/engine/builder.rs
@@ -393,7 +393,9 @@ impl Error {
     }
 }
 
-const INVALID_TOKENS: &[&str] = &["$colon$", "/", "\\"];
+// If/when we decide to be stricter about task names,
+// we can expand the patterns here.
+const INVALID_TOKENS: &[&str] = &["$colon$"];
 
 fn validate_task_name(task: &str) -> Result<(), Error> {
     INVALID_TOKENS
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/engine/builder.rs
```diff
@@ -389,6 +393,21 @@ impl Error {
     }
 }
 
+const INVALID_TOKENS: &[&str] = &["$colon$", "/", "\\"];
```
chris-olszewski, line 396: Do we need to ban `/` and `\\`? If I understand correctly, we only have weird behavior because of these for the log files. If we escaped these as well in the log file paths, then we don't need to ban them. (I'm afraid we're going to find a lot of projects use `/` in script names)

chris-olszewski (changes requested):
I'm afraid that banning `/` from script names might be a breaking change for too many people.
I checked on my laptop and we currently behave correctly from the user's perspective with a task like `lint/fix`. We create a nested log file (`apps/web/.turbo/turbo-lint/fix.log`), but we find it and replay it just fine.

crates/turborepo-lib/src/engine/builder.rs
mehulkar, line 396: should we just ban dollar signs in general or require that it match `a-zA-Z:-_` or something? And if we're going down that route, what about other languages/unicode/etc?

mehulkar (approved):
Don't have a problem with this per se, but I think we should work on a `turbo doctor` command and include warnings rather than adding runtime checks

crates/turborepo-lib/src/engine/builder.rs
gsoltis: Ugh. Ok. I'll limit it to just `$colon$` for now, but leave the mechanism in place to add more.

I kinda suspect cross-platform caching might be broken if you use those tokens, but that isn't necessarily a reason to ban them, just that they aren't cross-platform-safe

gsoltis: Per comment from @chris-olszewski, I'm going to try to minimize what we ban, and just go for the one escape sequence we use.

chris-olszewski (approved):

PR #6296, opened 2023-10-26
fix(lockfiles) trailing zero yaml parse

### LINKED ISSUE #6232: [turborepo] Turbo error: unable to find any locator for string_decoder@0.1

1.10.16

Mac

### Describe the Bug

Running `yarn turbo prune` on one of my projects in a monorepo results in some (all?) files being produced in `out` but the command fails with: `Turbo error: unable to find any locator for string_decoder@0.1`

Running `yarn why string_decoder@0.1` comes back with nothing and `yarn why string_decoder` shows a lot of different versions none of which are 0.1

```
yarn why string_decoder
├─ raw-body@npm:1.1.7
│  └─ string_decoder@npm:0.10.31 (via npm:0.10)
│
├─ readable-stream@npm:2.3.7
│  └─ string_decoder@npm:1.1.1 (via npm:~1.1.1)
│
├─ readable-stream@npm:2.3.8
│  └─ string_decoder@npm:1.1.1 (via npm:~1.1.1)
│
├─ readable-stream@npm:3.6.2
│  └─ string_decoder@npm:1.3.0 (via npm:^1.1.1)
│
└─ readable-stream@npm:4.4.2
   └─ string_decoder@npm:1.3.0 (via npm:^1.3.0)
```

I have removed `node_modules` and `yarn.lock` and reinstalled and still get the same issue. Yarn itself never complains and my builds all work.

### Expected Behavior

The command should work

### To Reproduce

I'm unable to find how to reproduce this as filtering by another scope in my monorepo works fine. This seems to be related to a particular scope.

### Reproduction Repo

_No response_

chris-olszewski:
### Description

Fixes #6232

This PR allows us to properly deserialize semver versions from YAML in `yarn.lock`. Previously we would fail at parsing ranges with trailing zeros e.g. `(0.10f32).to_string() == "0.1"`. 

The approach taken in this PR is due to some outstanding quirks in `serde_yaml`:
- https://github.com/dtolnay/serde-yaml/issues/165#issuecomment-644311452
- https://github.com/dtolnay/serde-yaml/issues/388
Our usage of `#[serde(flatten)]` in `LockfileData` caused attempting to parse `2` or `0.10` as a `String` to fail. To avoid this we first parse the document as a map with a union of the metadata/package entries and then convert this to the expected structure. This provides us the type safety of the old implementation, but at the cost of rebuilding the map.

As a minor thing, I removed all unused `Serialize`/`Deserialize` implementations to make it clear which codepaths actually get used.

### Testing Instructions

Existing unit tests pass.

I changed out the old unit tests for `SemverString` to be captured by the new `berry_semver.lock` test fixture which covers the same cases. We do this because even if parsing versions works when invoked directly, adding `#[serde(flatten)]`/`#[serde(untagged)]` to any containing structure changes the behavior. 

Closes TURBO-1540


crates/turborepo-lockfiles/Cargo.toml
```diff
@@ -14,7 +14,7 @@ regex = "1"
 semver = "1.0.17"
 serde = { version = "1.0.126", features = ["derive", "rc"] }
 serde_json = "1.0.86"
-serde_yaml = "0.9"
+serde_yaml = "0.9.27"
```
chris-olszewski, line 17: Pinning this to an exact `serde_yaml` version since we are now aware that we depend on `x: String` handles input like `x: 2` which can get affected by `serde` container attributes. I'm not sure which behavior is intended/buggy so we might be depending on broken behavior. (Unit tests would catch if an upgrade broke us, but just in case let's pin it)

crates/turborepo-lockfiles/src/berry/mod.rs
```diff
@@ -55,45 +55,42 @@ pub struct BerryLockfile {
...
-    #[serde(flatten)]
     packages: Map<String, BerryPackage>,
 }
 
-#[derive(Debug, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Clone)]
-#[serde(rename_all = "camelCase")]
+#[derive(Debug, PartialEq, Eq, Hash, PartialOrd, Ord, Clone)]
 struct Metadata {
-    version: u64,
+    version: String,
```
chris-olszewski, line 67: This will always be an integer, but now that we're avoiding the use of `#[serde(flatten)]` we can parse it as a `String` and it gets to share the same type as `version` in `BerryPackage` which makes our union type easier to work with.

arlyon (approved):
Unfortunate workaround but much better than rolling / forking our own parser. Glad we found something that works.

PR #6304, opened 2023-10-27
improve resolving performance

sokra:
### Description

* read_matches never took the fast path due to a bug in pattern normalization
* And there was a bug in the fast path anyway

### Testing Instructions

Closes WEB-1892

### DIFF
--- a/crates/turbopack-core/src/resolve/pattern.rs
+++ b/crates/turbopack-core/src/resolve/pattern.rs
@@ -986,7 +986,7 @@ pub async fn read_matches(
                 }
                 DirectoryContent::NotFound => {}
             };
-            Ok::<_, anyhow::Error>(())
+            anyhow::Ok(())
         }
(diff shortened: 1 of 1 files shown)


crates/turbopack-core/src/resolve/pattern.rs
```diff
@@ -810,169 +816,180 @@ pub async fn read_matches(
...
-                        DirectoryEntry::Other(_) => {}
-                        DirectoryEntry::Error => {}
                     }
                 }
-            }
-            DirectoryContent::NotFound => {}
-        };
+                DirectoryContent::NotFound => {}
+            };
+            Ok::<_, anyhow::Error>(())
```
ForsakenHarmony, line 989: you can use
```suggestion
            anyhow::Ok(())
```

ForsakenHarmony (approved):

jridgewell (approved):

PR #6305, opened 2023-10-27
Triple chunk size heuristic

sokra:
### Description

Reducing the number of requests by increasing the chunk size heuristic

### Testing Instructions

Closes WEB-1893

### DIFF
--- a/crates/turbopack-core/src/chunk/chunking.rs
+++ b/crates/turbopack-core/src/chunk/chunking.rs
@@ -312,7 +312,7 @@ fn folder_name(ident: &str, location: usize) -> (&str, Option<usize>) {
     }
 }
 
-const LARGE_CHUNK: usize = 1000_000;
+const LARGE_CHUNK: usize = 1_000_000;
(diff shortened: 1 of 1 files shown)


crates/turbopack-core/src/chunk/chunking.rs
```diff
@@ -312,8 +312,8 @@ fn folder_name(ident: &str, location: usize) -> (&str, Option<usize>) {
     }
 }
 
-const LARGE_CHUNK: usize = 300_000;
-const SMALL_CHUNK: usize = 30_000;
+const LARGE_CHUNK: usize = 1000_000;
```
jridgewell, line 315:
```suggestion
const LARGE_CHUNK: usize = 1_000_000;
```

jridgewell (approved):

kdy1:
It seems like benchmark result regressed. Maybe it's too much?

sokra:
> It seems like benchmark result regressed. Maybe it's too much?

Maybe need to test that on a large app

PR #6306, opened 2023-10-27
fix: task definition serialization

chris-olszewski:
### Description

A bunch of tiny changes that bring the JSON serialization to be closer to Go.

We go for a similar approach to Go of having a new structure for serialization in order to match Go's distinction between `nil` and `[]string{}`.

### Testing Instructions

More rust codepath integration tests pass. Many failing ones now have smaller diffs e.g. `dry_json/single_package*.t` only differ by `hashOfExternalDeps`

Closes TURBO-1548


crates/turborepo-lib/src/run/summary/task.rs
```diff
@@ -241,6 +267,67 @@ impl From<SharedTaskSummary<TaskId<'static>>> for SharedTaskSummary<String> {
...
+        let mut depends_on =
+            Vec::with_capacity(task_dependencies.len() + topological_dependencies.len());
+        for task_dependency in task_dependencies {
+            depends_on.push(task_dependency.to_string());
+        }
+        for topological_dependency in topological_dependencies {
+            depends_on.push(format!("^{topological_dependency}"));
+        }
+
+        // This should _not_ be sorted.
```
chris-olszewski, line 303: I'm not sure why this comment existed in Go, but I'm copying it and it's behavior here.

crates/turborepo-lib/src/run/summary/task.rs
```diff
@@ -96,7 +115,8 @@ pub struct TaskEnvVarSummary {
 
     pub configured: Vec<String>,
     pub inferred: Vec<String>,
-    pub pass_through: Vec<String>,
+    #[serde(rename = "passthrough")]
+    pub pass_through: Option<Vec<String>>,
```
chris-olszewski, line 119: Chose to use a `#[serde(rename)]` as `pass_through` is a better variable name.

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -343,7 +343,7 @@ impl<'a> Visitor<'a> {
                 match exit_status {
                     // The task was successful, nothing special needs to happen.
                     ChildExit::Finished(Some(0)) => {
-                        let _summary = tracker.build_succeeded(0);
+                        tracker.build_succeeded(0).await;
```
chris-olszewski, line 346: Very silly, but very important (for correctness) change.

turborepo-tests/integration/tests/dry_json/monorepo.t
```diff
@@ -182,52 +182,3 @@ Tasks that don't exist throw an error
   $ ${TURBO} run doesnotexist --dry=json
   Error: Could not find the following tasks in project: doesnotexist
   [1]
-
```
chris-olszewski, line 185: Removing as duplicated test doesn't test any new behavior now that we run the Rust codepath in CI

chris-olszewski (commented):
Notes

crates/turborepo-lib/src/run/summary/task.rs
NicholasLYang: Dot env files are the one exception where we don't sort our inputs, because they override each other

arlyon (approved):

crates/turborepo-lib/src/task_graph/visitor.rs
arlyon: Comment here as to why?

crates/turborepo-lib/src/run/summary/task_factory.rs
```diff
@@ -187,7 +205,10 @@ impl<'a> TaskSummaryFactory<'a> {
                 self.env_at_start,
             )
             .expect("invalid glob in task definition should have been caught earlier"),
-            dot_env: task_definition.dot_env.clone(),
+            dot_env: match task_definition.dot_env.is_empty() {
```
NicholasLYang, line 208: I feel like this won't work because when we process dotEnv, we flatten both the `None` case and the empty array case into an empty vector. So if you put `dotEnv: []` or if you put `dotEnv: null` (or omit the field entirely) in your turbo.json, we'll output `null` in both cases here. But in Go, we'll output `null` when we pass `dotEnv: null` and `[]` when we pass `dotEnv: []`.

NicholasLYang (changes requested):

crates/turborepo-lib/src/task_graph/visitor.rs
chris-olszewski: Wasn't awaiting a future so it never got ran 🙃

arlyon: Haha well that explains it. I was wondering why we didn't await but if its just because of 'integration' then no comment needed.

crates/turborepo-lib/src/run/summary/task_factory.rs
chris-olszewski: Good catch

NicholasLYang (approved):

PR #6307, opened 2023-10-28
Update tsup.config.ts to support "use client" of Next.js

williamli:
After digging into this, there seems to be ~2~ 3 ways to do this:

## Option 1: Build first, use `onSuccess` hook to prepend use client to the start of `ui/client` package entry points.

```
import fs from "fs/promises";
import path from "path";

export default defineConfig((options: Options) => ({
  treeshake: true,
  splitting: true,
  entry: ["src/index.tsx", "src/client/index.tsx"],
  format: ['esm', 'cjs'],
  dts: true,
  minify: true,
  clean: true,
  external: ["react"],
  ...options,
  async onSuccess() {
    // add "use client" banner to /dist/client entry point
    await fs.writeFile(
      path.join(__dirname, "dist", "client", "index.js"),
      '"use client";\n' + await fs.readFile(path.join(__dirname, "dist", "client", "index.js"))
    );
    await fs.writeFile(
      path.join(__dirname, "dist", "client", "indexm.js"),
      '"use client";\n' + await fs.readFile(path.join(__dirname, "dist", "client", "index.mjs"))
    );

  },
}));

```

### Pro:
- components are built once
- treeshake can set to  true

### Con:
- when the client components are compiled, before `onSuccess` is completed, there is a brief moment where there is no ‘use client’ for `ui/client`, which might cause Next.js app to flash an error to complain about the missing ‘use client’… the error will go away once `onSuccess` is completed.
- if there are build errors, onSuccess is not executed, client components will not get the use client directive, all client components will be broken

## Option 2: Use a separate config for `tsup` for a new `/dist/use-client` file structure in which `use client` is added as a js banner.

```
import type { Options } from "tsup";
import { defineConfig } from "tsup";

  const cfg = {
    splitting: false,
    sourcemap: true,
    clean: true,
    treeshake: false,
    dts: true,
    format: ['esm', 'cjs'],
    external: ['react'],
  };
  
  export default defineConfig([
    {
      ...cfg,
      entry: ["src/client/index.tsx"],
      esbuildOptions: (options) => {
        // Append "use client" to the top of the react entry point
        options.banner = {
          js: '"use client";',
        };
      },
      outDir: 'dist/use-client',
    },
    {
      ...cfg,
      entry: ["src/index.tsx", "src/client/index.tsx"],
      outDir: 'dist',
    }
  ]);

```

### Pro:
- Next.js will not complain

### Con:
- treeshake has to be false
- Components in `/client` are built twice. Note: I tried to skip `/client` from being processed in the 2nd `tsup` task, but doing that will isolate styles defined in `/client` from tailwind’s stylesheet building task. classNames defined in `/client` will not get included in `style.css` until files outside of `/client` changes and trigger another `tsup` build.

## Option 3: Based on Option 1, we prepend `use client` in a staging area (`/dist/server/client`) before moving the folder to `/dist/client`. This addresses all the cons of option 1. An extra package (fs-extra) is required to handle the moving.

```
// tsup.config.js

import fs from "fs-extra";
import path from "path";

export default defineConfig((options: Options) => ({
  treeshake: true,
  splitting: true,
  entry: ["src/index.tsx", "src/client/index.tsx"],
  outDir: 'dist/server',
  format: ['esm', 'cjs'],
  dts: true,
  minify: true,
  clean: true,
  external: ["react"],
  ...options,
  async onSuccess() {
    // add "use client" banner to /dist/client entry point
    await fs.writeFile(
      path.join(__dirname, "dist", "server", "client", "index.js"),
      '"use client";\n' + await fs.readFile(path.join(__dirname, "dist", "server", "client", "index.js"))
    );
    await fs.writeFile(
      path.join(__dirname, "dist", "server", "client", "indexm.js"),
      '"use client";\n' + await fs.readFile(path.join(__dirname, "dist", "server", "client", "index.mjs"))
    );
      
    // move /dist/server/client to /dist/client
    try {
      fs.moveSync(path.join(__dirname, "dist", "server", "client"), path.join(__dirname, "dist", "client"), { overwrite: true })
      
    } catch (err) {
      console.error(err)
    }
  },
}));
```


examples/with-tailwind/packages/ui/tsup.config.ts
```diff
@@ -1,13 +1,46 @@
...
+  return paths;
+}
 
 export default defineConfig((options: Options) => ({
   treeshake: true,
   splitting: true,
   entry: ["src/**/*.tsx"],
-  format: ["esm"],
+  outDir: 'dist',
+  format: ['cjs'],
```
anthonyshew, line 27: I'd love to be recommending the most modern practices we can. Any way we can get this back to ESM?

examples/with-tailwind/packages/ui/tsup.config.ts
```diff
@@ -1,13 +1,46 @@
...
   dts: true,
   minify: true,
   clean: true,
```
anthonyshew, line 30: This `clean: true` causes VSCode to show missing types when restarting the dev server. I'm not sure how to mitigate this since removing this would make it so you could have outdated modules laying around while developing. 🤔

examples/with-tailwind/packages/ui/src/client/index.tsx
```diff
@@ -0,0 +1,3 @@
+// components
+export * from "./counter";
+export * from "./counters";
```
anthonyshew, line 3: We need to do our best to move off of barrel files since they're a suboptimal practice. Any ideas?

williamli: You don’t need to have them. I just have them in case people want to `import { Card } from “ui”`. The example in the web project already does stuff like `import { Counters } from `“ui/client/counters”`.

williamli: I think `<Card/>` is still imported from `ui` but I will update it to import from `ui/card`.

williamli: `docs` workspace updated to `import { Card } from 'ui/card'` in https://github.com/vercel/turbo/pull/6307/commits/f9db4e4b57ff46520e9821121adf4758f78456c3, the other examples already load from non barrel files.

@anthonyshew do you want to eliminate the barrel files (index files) all together or keep them around to make both import option open?

examples/with-tailwind/packages/ui/tsup.config.ts
williamli: resolved in https://github.com/vercel/turbo/pull/6307/commits/c79b74eb86168c002b08bb0860304a572dd33fc1

examples/with-tailwind/packages/ui/tsup.config.ts
williamli: I was looking into this too. I think having `clean` enabled deteriorates DX, sometime to the point of requiring a full manual browser refresh due to "missing" chunks. It was enabled before but I think we should try disabling it.

williamli: turned clean to false in https://github.com/vercel/turbo/pull/6307/commits/13ab401a5d07ac41981b5f1a3aec307d6caf19a4,

much better local dev experience. no more nagging errors flashing up.

williamli:
@anthonyshew i have discovered a bigger problem of the current `tsup` setup in the `main` branch.

tsup does not pick up file structure changes, meaning:
1. if you remove a file, it will crash while trying to work on it,
2. if you add a new file, it will not be processed until you restart `pnpm dev` or `turbo dev`.

examples/with-tailwind/packages/ui/tsup.config.ts
williamli: added `rimraf` to clean up `dist` before build / dev commands in https://github.com/vercel/turbo/pull/6307/commits/ce24402d8b269121f0e32e49da37fd967a200049

magnusriga:
Not sure if it is OK to comment here, but I was wondering how this is coming along? Any chance it will be supported soon?

As a side note, I got an error when trying approach 1: 

`Error: ENOENT: no such file or directory, open '/home/magnus/nfront/packages/ui/dist/client/index.js'.` Any idea why that is happening?

Thanks a lot for the effort here.

magnusriga:
Trying OPTION 3, I end up with:

image

And:

image

As seen, it adds the `use client` directive, but that does not help because `Test` (which is a Client Component) is split out into a new chunk which does not have the `use client` directive.

magnusriga:
I had written a couple of other comments, but ended up deleting them as I had missed the fact that once a component is marked with 'use client' all its children also become client components (without having to explicitly mark them with the `use client` directive).

However, the options above still fail in two important scenarios:

1. In the ui src folder, import a client component (for example `Counter`) into a server component (for example `Card`). That is common practice, however, that will not work here because the client component `Counter` is split out into its own chunk without the `use client` directive. Screenshots from the branch in this PR:

image

**The `Counter` chunk (notice it is a client component, but is missing the `use client` directive):**

image

**The server component `Card`, importing the client component `Counter` from the new chunk instead of from `./client/counter`:**

image

Turning `splitting` to false did not solve the issue.

Have you seen how the react team did it, by the way? They seem to build the folders separately: https://github.com/vercel/analytics/blob/main/packages/web/tsup.config.js

williamli:
@magnusriga nice catch on Server Components importing Client Components. I will update the scripts to walk through the client entry points and insert `use client` if it's missing.

> Have you seen how the React team did it, by the way? They seem to build the folders separately: https://github.com/vercel/analytics/blob/main/packages/web/tsup.config.js

Yes, I have tried that but it wouldn't work because Tailwind CSS requires reading both server and client components to create the stylesheet, so you will need to have a build with both server and client to create the style sheet. So you will end up with 3 builds: server only, client only, and server + client (for CSS)... which is something I am trying to avoid.

williamli:
@magnusriga added support for using Client Components inside Server Components within the same (`ui`) package.

arlyon (approved):
I am approving this with a 'fixship'; I think a comment explaining the problem and motivation as well as why it doesn't work out of the box (either in the PR, or in the `tsup.config.ts` itself) would go a long way to explain what is going on here.

williamli:
@arlyon yeah, an update in README with details is in the work. I was waiting for @anthonyshew to come back to help on that but I can try to update it next week.

magnusriga:
@williamli Have you seen this, by Jared Palmer?: https://turbo.build/blog/you-might-not-need-typescript-project-references#nextjs

Seems we might not even need to build the internal packages at all. I.e. they can simply be imported from the src directory and next.js (or whichever framework one uses) can handle all the transpiling and bundling.

anthonyshew:
The larger answer here is that we don't want to bundle internal packages. Rather, they should be compiled with `tsc`. The difficulties of this PR show why; it's an additional layer of complexity that we don't actually need.

Our examples today are indeed using `tsup` but that needs to change. I'm back from paternity leave now and my first major project is to update all of our examples. I still have some design/architecture work to think about with these but will update here once we've got some decisions made.

Appreciate the work!

magnusriga:
> The larger answer here is that we don't want to bundle internal packages. Rather, they should be compiled with `tsc`. The difficulties of this PR show why; it's an additional layer of complexity that we don't actually need.
> 
> Our examples today are indeed using `tsup` but that needs to change. I'm back from paternity leave now and my first major project is to update all of our examples. I still have some design/architecture work to think about with these but will update here once we've got some decisions made.
> 
> Appreciate the work!

@anthonyshew thank you for the input. Quick question: Can we just import the files from src and forget about the entire compilation and bundling step. I.e. just let Next.js handle all that, like the turbo docs explain in the internal packages section?

If we do it that way, why do we need to compile anything with tsc? Can't we just let next.js (for instance) handle it all? It seems that's what the basic Next.js example you have in the docs show.

Thanks a lot for the guidance.

anthonyshew:
@magnusriga That is also an option - but it comes with a tradeoff. You won't be able to do as much caching since there aren't any outputs to cache on that internal package.

Depending on the size of your repository, this could have a significant impact on your build times.

magnusriga:
@anthonyshew Sorry to bother you again. I know you are working on the examples, but in the meantime do you have a repo that uses tsc, that would work for turbo with tailwind, etc. I am just wondering which tsc command to use and how to set up my tsconfig.

Thanks a ton!

magnusriga:
> @magnusriga That is also an option - but it comes with a tradeoff. You won't be able to do as much caching since there aren't any outputs to cache on that internal package.
> 
> Depending on the size of your repository, this could have a significant impact on your build times.

I finally found a setup that works in every way, including with Intellisense on the RHS paths of imports. The setup, and a repo can be found here: https://github.com/microsoft/TypeScript/issues/56412

anthonyshew:
Y'all, I'm incredibly sorry that I forgot we had this PR out. I'm coming back from parental leave and trying to get my head on straight still.

While I was looking down the list of work I had out, I didn't realize that this PR was for the `with-tailwind` example. I read the title and skimmed the description, not realizing the work @williamli had done.

I worked on the `with-tailwind` example in #6575, solving the same problems as this PR:
- Supporting a mix of Server and Client Components
- Auto-imports across packages
- Types staying updated in IDE (cc @magnusriga)
- Tree-shaking

I'm going to favor #6575 for the sake of simplicity so if you'd like to leave critiques there, happy to have them!

anthonyshew:
You can certainly let `tailwindcss` + PostCSS compile the output CSS in the consuming application by assigning the `content` field to capture the `./packages/ui` glob but this isn't as "hermetic" and fundamentally sound as the technique in the example. In practice, it's usually fine but I'd like to ensure we're keeping to the fundamentals that each package should be as "self-contained" as possible. If `app-a` is directly referencing paths in `@repo/ui` that's not necessarily the most fundamental approach.

Exporting `/*: ./src/*` would be nice but it doesn't support auto-importing across package boundaries, unfortunately. There are Issues in the Typescript repo that reference this performance tradeoff/limitation.

magnusriga:
> You can certainly let `tailwindcss` + PostCSS compile the output CSS in the consuming application by assigning the `content` field to capture the `./packages/ui` glob but this isn't as "hermetic" and fundamentally sound as the technique in the example. In practice, it's usually fine but I'd like to ensure we're keeping to the fundamentals that each package should be as "self-contained" as possible. If `app-a` is directly referencing paths in `@repo/ui` that's not necessarily the most fundamental approach.
> 
> Additionally, with the approach we're using, we don't have to bring in a bundler at all (outside of the Next.js Compiler), reducing our repo's general overhead.
> 
> Exporting `/*: ./src/*` would be nice but it doesn't support auto-importing across package boundaries, unfortunately. There are Issues in the Typescript repo that reference this performance tradeoff/limitation.

Got it. One more question, if/when you have time:

Will the tailwind css classes from both the components in Next.js and the components in `package/ui` be present in the final bundle produced by next.js? I did not see imports for both the produced css in `package/ui/dist` and the three tailwind lines in the next.js app, in the `layout.ts` file (I only saw the former). Do you not need to import both for tailwind-via-postcss in the consuming app to include all the correct classes in the final bundle?

Side note: Is the auto-import issue fixed if we use `tsc` to transpile the typescript in `package/ui` to JavaScript before importing it, like you originally suggested?

Thanks again.

magnusriga:
@anthonyshew In addition to my comment above, I think the README might be wrong. It says you are compiling the TS to dist.

clearly-outsane:
I tried this and this fails for me when I have client components imported in a server component. I think it's because the imports in the server component is from a chunk and the chunk does not have a "use client"; on top ( thought the component file itself does ).

Issue #6188, opened 2023-10-17
Group task logs in CI


weyert:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

I would like to be able to visually group logs in CI so its easier to keep track of logs or to expand/collapse

### Describe the solution you'd like

I would like to have support for some special logging instructions that are supported in CI solutions, the most popular ones are:

Gitlab:
Requirement: None

start: `section_start:${Math.floor(Date.now() / 1000)}:${taskName}\\r\\e[0K${taskName}\r\n`
end: `section_end:${Math.floor(Date.now() / 1000)}:${taskName}\\r\\e[0K`

Teamcity:
start: `##teamcity[blockOpened name='${taskName}']
`
end: `##teamcity[blockClosed name='${taskName}']`

Github Actions:
Requirement: None

start: `::group::${taskName}\r\n`
end: `::endgroup::\r\n`

Travis:
start: `travis_fold:start:${taskName}\r\n`,
end: `travis_fold:end:${taskName}\r\n`,

Azure:
start: `##[group]${taskName}\r\n`,
end: `##[endgroup]\r\n`,

### Describe alternatives you've considered

None

<sub>[TURBO-1517](https://linear.app/vercel/issue/TURBO-1517/group-task-logs-in-ci)</sub>

mehulkar:
Thanks for the request, I'll take it to the team to discuss priority!

weyert:
@mehulkar Yeah, you appear to already do it for GitHub Actions so why not for the other popular CIs

mehulkar:
It would make sense to me to add support for these https://github.com/vercel/turbo/blob/main/cli/internal/ci/vendors.go (and maybe remove some from the list). But since we don't actively use these CI vendors, we'll probably be running blind with the syntax (e.g. if it changes)

mehulkar:
We talked about this and want to add support. Happy take contributions if anyone is interested to PR, but it's in our backlog

cymruu:
I'm interested in tackling this task. Upon reviewing the code, I've found out that the code responsible for generating groups in the GitHub Actions output, is currently implemented in Go codebase. However, there's no corresponding implementation in the Rust codebase. 
Should this feature be implement this feature in Go, Rust, or in both languages?

weyert:
@cymruu Thanks that would be great :)

mehulkar:
@cymruu I would start only with Rust. From what I understand the port is very nearly done

[...]

PR #6314, opened 2023-10-30
feat: vendor CI log groups

cymruu:
### Description
Adds support for grouping tasks logs for some CI vendors. Closes https://github.com/vercel/turbo/issues/6188

### Testing Instructions
I've added some unit tests, sadly I need guidance how to test it more reliably as the changes were made in rust code which is not currently used in generated turbo binary.


chris-olszewski:
I'll work on getting this reviewed and ready for merge, but it probably won't get shipped until the Rust codepath is feature complete and relatively bug free. (Only 3 minor features to finish up and then roughly 40 integration tests to fixup) From a first glance this looks great. Thanks!

arlyon:
Hey small update here, we are in the process of ripping out the go code in the next week or two at which point we have free reign to merge this in. I'll update here when that is done :)

cymruu:
@arlyon I'm happy to hear that. Thanks for heads up!

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -367,13 +378,16 @@ impl<'a> Visitor<'a> {
...
-        if self.run_opts.is_github_actions {
-            let package = if self.run_opts.single_package {
-                None
+        if let Some(vendor_behavior) = vendor_behavior {
+            let group_name = if self.run_opts.single_package {
+                ""
             } else {
-                Some(task_id.package())
+                task_id.package()
             };
```
chris-olszewski, line 386: We to include both package name and task name in the group name if it's a multi-package repo and the task name if it is just single package mode. This matches previous behavior.

e.g.
```
let group_name = if self.run_opts.single_package {
    task_id.task().to_string()
} else {
    format!("{}:{}", task_id.package(), task_id.task())
}
```

crates/turborepo-ci/src/vendors.rs
```diff
@@ -243,6 +266,10 @@ pub(crate) fn get_vendors() -> &'static [Vendor] {
                     sha_env_var: Some("GITHUB_SHA"),
                     branch_env_var: Some("GITHUB_REF_NAME"),
                     username_env_var: Some("GITHUB_ACTOR"),
+                    behavior: Some(VendorBehavior {
+                        group_prefix: |group_name, _| format!("::group::{group_name}"),
+                        group_suffix: |_, _| String::from("::endgroup::"),
```
chris-olszewski, line 271:
```suggestion
                        group_prefix: |group_name, _| format!("::group::{group_name}\n"),
                        group_suffix: |_, _| String::from("::endgroup::\n"),
```

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -367,13 +378,16 @@ impl<'a> Visitor<'a> {
...
-            let (header, footer) = github_header_footer(package, task_id.task());
+            let (header, footer) = (
+                (vendor_behavior.group_prefix)(group_name, task_start_time),
+                (vendor_behavior.group_suffix)(group_name, task_start_time),
```
chris-olszewski, line 389: We'll want to defer computing the footer until later as we won't know until the task finishes what the correct task end time will be.

Given that this constraint complicates the design a bit, I think it's reasonable to either:
 - Leave this as-is with a TODO comment explaining that the timestamp will be incorrect
 - Remove the Gitlab CI vendor prefix/suffix implementation as it is the only one that uses the timestamp.

chris-olszewski (changes requested):
First off, thank you so much for your patience and this great PR.

I think the design is very solid and only have some minor behavior suggestions.


As far as testing. The only test we have that currently is [log-order-github.t](https://github.com/vercel/turbo/blob/e834b541bfefdc714fcf4d36bc16fe44b43f5291/turborepo-tests/integration/tests/run-logging/log-order-github.t)

You can run this with the following commands:
```
cd turborepo-tests/integration
python3 -m venv .cram_env
python3 -m pip install --upgrade pip
pip install "prysk==0.15.2"
.cram_env/bin/prysk --shell=bash tests/run-logging/log-order-github.t
```

I don't think we need additional testing of each individual vendor output. If you do want to verify that another vendor's CI header/footer, you can copy that Github test and update it with the appropriate env var and logs.

crates/turborepo-lib/src/task_graph/visitor.rs
cymruu: I decided to remove the GitlabCI vendor for now. I'll be interested in adding it again in a separate pull request

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 4175321eab38658908b8c240e2299d9481b576fb

PR #6315, opened 2023-10-30
port(turborepo): Analytics Client

NicholasLYang:
### Description

Adds analytics client for cache hits and misses.

### Testing Instructions

Added analytics to our round trip cache tests. Validates that we're properly sending the events.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,6 +136,7 @@ turbopack-tests = { path = "crates/turbopack-tests" }
 turbopack-wasm = { path = "crates/turbopack-wasm" }
 turbopath = { path = "crates/turborepo-paths" }
 turborepo = { path = "crates/turborepo" }
+turborepo-analytics = { path = "crates/turborepo-analytics" }
 turborepo-api-client = { path = "crates/turborepo-api-client" }
 turborepo-cache = { path = "crates/turborepo-cache" }
 turborepo-ci = { path = "crates/turborepo-ci" }
--- a/crates/turborepo-analytics/src/lib.rs
+++ b/crates/turborepo-analytics/src/lib.rs
@@ -1,0 +1,364 @@
+#![deny(clippy::all)]
+
+use std::time::Duration;
+
+use futures::stream::FuturesUnordered;
+use thiserror::Error;
+use tokio::{
+    select,
+    sync::{mpsc, oneshot},
+    task::{JoinError, JoinHandle},
+};
+use tracing::debug;
+use turborepo_api_client::{analytics::AnalyticsClient, APIAuth};
+pub use turborepo_vercel_api::AnalyticsEvent;
+use uuid::Uuid;
+
+const BUFFER_THRESHOLD: usize = 10;
+
+static EVENT_TIMEOUT: Duration = Duration::from_millis(200);
+static NO_TIMEOUT: Duration = Duration::from_secs(24 * 60 * 60);
+static REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Failed to send analytics event")]
+    SendError(#[from] mpsc::error::SendError<AnalyticsEvent>),
+    #[error("Failed to record analytics")]
+    Join(#[from] JoinError),
+}
+
+// We have two different types because the AnalyticsSender should be shared
+// across threads (i.e. Clone + Send), while the AnalyticsHandle cannot be
+// shared since it contains the structs necessary to shut down the worker.
+pub type AnalyticsSender = mpsc::UnboundedSender<AnalyticsEvent>;
+
+pub struct AnalyticsHandle {
+    exit_ch: oneshot::Receiver<()>,
+    handle: JoinHandle<()>,
+}
+
+pub fn start_analytics(
+    api_auth: APIAuth,
+    client: impl AnalyticsClient + Clone + Send + Sync + 'static,
+) -> (AnalyticsSender, AnalyticsHandle) {
+    let (tx, rx) = mpsc::unbounded_channel();
+    let (cancel_tx, cancel_rx) = oneshot::channel();
+    let session_id = Uuid::new_v4();
+    let worker = Worker {
+        rx,
+        buffer: Vec::new(),
+        session_id,
+        api_auth,
+        senders: FuturesUnordered::new(),
+        exit_ch: cancel_tx,
+        client,
+    };
+    let handle = worker.start();
+
+    let analytics_handle = AnalyticsHandle {
+        exit_ch: cancel_rx,
+        handle,
+    };
+
+    (tx, analytics_handle)
+}
+
+impl AnalyticsHandle {
+    async fn close(self) -> Result<(), Error> {
+        drop(self.exit_ch);
+        self.handle.await?;
+
+        Ok(())
+    }
+
+    pub async fn close_with_timeout(self) {
+        if let Err(err) = tokio::time::timeout(EVENT_TIMEOUT, self.close()).await {
+            debug!("failed to close analytics handle. error: {}", err)
+        }
+    }
+}
+
+struct Worker<C> {
+    rx: mpsc::UnboundedReceiver<AnalyticsEvent>,
+    buffer: Vec<AnalyticsEvent>,
+    session_id: Uuid,
+    api_auth: APIAuth,
+    senders: FuturesUnordered<JoinHandle<()>>,
+    // Used to cancel the worker
+    exit_ch: oneshot::Sender<()>,
+    client: C,
+}
+
+impl<C: AnalyticsClient + Clone + Send + Sync + 'static> Worker<C> {
+    pub fn start(mut self) -> JoinHandle<()> {
+        tokio::spawn(async move {
+            let mut timeout = tokio::time::sleep(NO_TIMEOUT);
+            loop {
+                select! {
+                    // We want the events to be prioritized over closing
+                    biased;
+                    event = self.rx.recv() => {
+                        if let Some(event) = event {
+                            self.buffer.push(event);
+                        }
+                        if self.buffer.len() == BUFFER_THRESHOLD {
+                            self.flush_events();
+                            timeout = tokio::time::sleep(NO_TIMEOUT);
+                        } else {
+                            timeout = tokio::time::sleep(REQUEST_TIMEOUT);
+                        }
+                    }
+                    _ = timeout => {
+                        self.flush_events();
+                        timeout = tokio::time::sleep(NO_TIMEOUT);
+                    }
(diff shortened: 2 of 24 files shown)


Cargo.toml
```diff
@@ -136,6 +136,7 @@ turbopack-tests = { path = "crates/turbopack-tests" }
 turbopack-wasm = { path = "crates/turbopack-wasm" }
 turbopath = { path = "crates/turborepo-paths" }
 turborepo = { path = "crates/turborepo" }
+turborepo-analytics = { path = "crates/turborepo-analytics" }
```
NicholasLYang, line 139: We needed a new package because turborepo-cache and turborepo-lib needed to use it.

crates/turborepo-api-client/src/lib.rs
```diff
@@ -376,10 +377,10 @@ impl Client for APIClient {
 
         let response = retry::make_retryable_request(request_builder).await?;
 
-        if response.status() == StatusCode::FORBIDDEN {
-            Err(Self::handle_403(response).await)
-        } else {
-            Ok(response.error_for_status()?)
+        match response.status() {
+            StatusCode::FORBIDDEN => Err(Self::handle_403(response).await),
```
NicholasLYang, line 381: Had to tweak this because we weren't actually producing cache miss errors. In the future I'd like to refactor cache misses to not be an error

NicholasLYang (commented):
Notes

crates/turborepo-api-client/src/lib.rs
```diff
@@ -463,6 +464,50 @@ impl APIClient {
             use_preflight,
         })
     }
+
+    /// Create a new request builder with the preflight check done,
+    /// team parameters added, and CI header.
```
chris-olszewski, line 469: Probably worth mentioning that this sets the content type to json
```suggestion
    /// team parameters added, CI header, and a content type of json.
```

crates/turborepo-analytics/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+// shared since it contains the structs necessary to shut down the worker.
+pub type AnalyticsSender = mpsc::Sender<AnalyticsEvent>;
+
+pub struct AnalyticsHandle {
+    exit_ch: oneshot::Receiver<()>,
+    handle: JoinHandle<()>,
+}
+
+pub fn start_analytics(api_auth: APIAuth, client: APIClient) -> (AnalyticsSender, AnalyticsHandle) {
+    let (tx, rx) = mpsc::channel(CHANNEL_SIZE);
```
chris-olszewski, line 43: Do we want to use an unbounded channel here? We don't necessarily want back pressure from sending analytic events. A cache operation should never be blocked from completing if the analytics channel is full. (I realize this is unlikely to ever happen)

crates/turborepo-cache/src/fs.rs
```diff
@@ -41,14 +44,33 @@ impl FSCache {
...
+        if let Some(analytics_recorder) = &self.analytics_recorder {
+            let analytics_event = AnalyticsEvent {
+                session_id: None,
+                source: analytics::CacheSource::Local,
+                event,
+                hash: hash.to_string(),
+                duration,
+            };
+
+            let _ = analytics_recorder.send(analytics_event).await;
```
chris-olszewski, line 69: Related to my comment about the unbounded channel, but I think we could get away with using `blocking_send` here to avoid holding a reference to `&self` across an await and we can drop the `Send` requirement on the reader. (If we don't swap to an unbounded receiver)

crates/turborepo-cache/src/http.rs
```diff
@@ -120,6 +126,20 @@ impl HTTPCache {
...
+        // If analytics fails to record, it's not worth failing the cache
+        if let Some(analytics_recorder) = &self.analytics_recorder {
+            let analytics_event = AnalyticsEvent {
+                session_id: None,
+                source: analytics::CacheSource::Remote,
+                event,
+                hash: hash.to_string(),
+                duration,
+            };
+            let _ = analytics_recorder.send(analytics_event).await;
```
chris-olszewski, line 139: Same comment as in the fs cache with respect to unbounded channel/or `blocking_send`

crates/turborepo-vercel-api/src/lib.rs
```diff
@@ -124,3 +124,32 @@ pub struct APIError {
...
+}
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct AnalyticsEvent {
+    pub session_id: Option<String>,
+    pub source: CacheSource,
+    pub event: CacheEvent,
+    pub hash: String,
+    pub duration: u64,
+}
```
chris-olszewski, line 149: Can we have a unit test to make sure this serializes how we expect it to serialize?

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -397,6 +413,10 @@ impl<'a> Run<'a> {
             )
             .await?;
 
+        if let Some(analytics_handle) = analytics_handle {
+            analytics_handle.close_with_timeout().await;
```
chris-olszewski, line 417: We need to flush/close the analytics regardless of how the function finishes, not just if we hit this line.

crates/turborepo-analytics/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+        self.handle.await?;
+
+        Ok(())
+    }
+
+    pub async fn close_with_timeout(self) {
+        if let Err(err) = tokio::time::timeout(EVENT_TIMEOUT, self.close()).await {
+            debug!("failed to close analytics handle. error: {}", err)
+        }
+    }
```
chris-olszewski, line 77: We need a way to send a close signal from a different task via [`ctrl_c`](https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html) so this might not be viable

chris-olszewski (changes requested):
Primary concerns:
 - making sure that analytics are flushed regardless of where the main run function exits ([`async drop(self)`](https://rust-lang.github.io/async-fundamentals-initiative/roadmap/async_drop.html) when?)
 - some unit testing around the worker behavior

crates/turborepo-cache/src/fs.rs
NicholasLYang: I originally had it as `blocking_send` here, but since this is called in an async context (specifically in a tokio thread), it panics. Also, it seemed like a code smell to have code that panics if it's called in the wrong situation anyways

crates/turborepo-analytics/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+    session_id: Uuid,
+    api_auth: APIAuth,
+    senders: FuturesUnordered<JoinHandle<()>>,
+    // Used to cancel the worker
+    exit_ch: oneshot::Sender<()>,
+    client: APIClient,
+}
+
+impl Worker {
+    pub fn start(mut self) -> JoinHandle<()> {
```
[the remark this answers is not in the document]
NicholasLYang: I didn't port the context one because we don't have any context equivalent. And the session id one seemed to be testing mapstructure more than anything else tbh

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -17,12 +17,15 @@ use crate::{
 };
 
 pub struct CacheReader<'a> {
-    reader: Box<dyn Read + 'a>,
+    reader: Box<dyn Read + Send + 'a>,
```
chris-olszewski, line 20: Pretty sure this is no longer needed
```suggestion
    reader: Box<dyn Read + 'a>,
```

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -17,12 +17,15 @@ use crate::{
...
 }
 
 impl<'a> CacheReader<'a> {
-    pub fn from_reader(reader: impl Read + 'a, is_compressed: bool) -> Result<Self, CacheError> {
-        let reader: Box<dyn Read> = if is_compressed {
+    pub fn from_reader(
+        reader: impl Read + Send + 'a,
+        is_compressed: bool,
+    ) -> Result<Self, CacheError> {
+        let reader: Box<dyn Read + Send> = if is_compressed {
```
chris-olszewski, line 28:
```suggestion
    pub fn from_reader(
        reader: impl Read + 'a,
        is_compressed: bool,
    ) -> Result<Self, CacheError> {
        let reader: Box<dyn Read> = if is_compressed {
```

crates/turborepo-cache/src/cache_archive/restore.rs
```diff
@@ -35,7 +38,7 @@ impl<'a> CacheReader<'a> {
         let file = path.open()?;
         let is_compressed = path.extension() == Some("zst");
 
-        let reader: Box<dyn Read> = if is_compressed {
+        let reader: Box<dyn Read + Send> = if is_compressed {
```
chris-olszewski, line 41:
```suggestion
        let reader: Box<dyn Read> = if is_compressed {
```

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -102,13 +114,36 @@ impl<'a> Run<'a> {
...
+        let (analytics_sender, analytics_handle) =
+            Self::initialize_analytics(api_auth.clone(), api_client.clone()).unzip();
+
+        let result = self
+            .run_with_analytics(start_at, api_auth, api_client, analytics_sender)
+            .await;
+
+        if let Some(analytics_handle) = analytics_handle {
+            analytics_handle.close_with_timeout().await;
+        }
```
chris-olszewski, line 128: I think once #6357 lands we'll revert this and instead setup a [cleanup subscriber like we do for the process manager](https://github.com/vercel/turbo/pull/6357/files#diff-c0b6ca703dfc4f10dcb4ccdc330c318feb29f12affc980179d34b45039416727R58-R64)

crates/turborepo-analytics/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+          event: turborepo_vercel_api::CacheEvent::Miss,
+          hash: "this-is-my-hash-2".to_string(),
+          duration: 21,
+        },
+        "without-id-remote-miss"
+    )]
+    fn test_serialize_analytics_event(event: AnalyticsEvent, name: &str) {
+        let json = serde_json::to_string(&event).unwrap();
+        insta::assert_json_snapshot!(name, json);
+    }
```
chris-olszewski, line 396: Nit: these tests should live in `turborepo-vercel-api` since that's where the serialization logic is defined.

chris-olszewski (approved):
LGTM

PR #6318, opened 2023-10-30
feat(turbopack-ecmascript): calculate import.meta.url as an absolute path

kwonoj:
### Description

This PR attempts to support `import.meta.url` to resolve as an actual absolute path. Currently, it resolves to a relative path with `/ROOT/` placeholder. 

In contrast to typical path resolution logics in turbopack, we are required to supply a genuine absolute path for the assets. Node.js can consider a stagnant value as constant and refer to it in any way. To support those, this PR performs below steps:

- When we create turbopack_runtime, calculate absolute path of the `project` via node.js's `__filename` and the `OUTPUT_ROOT` (which is a distDir)
- create a fn `resolveAbsolutePath` can compute an absolute path to a given module path in runtime
- inject `resolveAbsolutePath` into each module chunks
- convert `import.meta.url` as getter, which can be lazily evaluated when it's being called - so we can access runtime-injected fn to actually calculate paths.

PR allows to pass one of import.meta.url based test cases (in dynamic metadata). Once this is approved & landed, would apply same approach for the __dirname and trying to verify against remaining test cases.


crates/turbopack-ecmascript/src/references/esm/meta.rs
```diff
@@ -49,12 +49,20 @@ impl CodeGenerateable for ImportMetaBinding {
...
+                quote!(
+                    "`file://${__turbopack_resolve_absolute_path__($formatted)}`" as Expr,
+                    formatted: Expr = formatted.into()
+                )
+            },
         );
 
         let visitor = create_visitor!(visit_mut_program(program: &mut Program) {
+            // [NOTE] url property is lazy-evaluated, as it should be computed once turbopack_runtime injects a function
+            // to calculate an absolute path.
```
jridgewell, line 63: The module this appears in won't be evaluated until the after the runtime anyways, so I don't think this is necessary. But deferring the `import.meta.url` error is probably good, so we can keep it.

crates/turbopack-ecmascript-runtime/js/src/dev/runtime/dom/runtime-backend-dom.ts
```diff
@@ -52,6 +52,11 @@ async function loadWebAssemblyModule(
   return await WebAssembly.compileStreaming(req);
 }
 
+// [TODO] need to match behavior as similar to UrlAssetReference
```
jridgewell, line 55: Don't we need to implement this now?

crates/turbopack-build/src/ecmascript/node/entry/runtime.rs
```diff
@@ -48,13 +48,16 @@ impl EcmascriptBuildNodeRuntimeChunk {
         };
 
         let mut code = CodeBuilder::default();
+        let output_root = output_root.to_string();
```
jridgewell, line 51: Did you verify whether webpack uses the input path or the output paths to transform `import.meta.url`?

crates/turbopack-ecmascript-runtime/js/src/dev/runtime/dom/runtime-backend-dom.ts
kwonoj: I'd like to separate PR, since this is dom runtime implementation will likely entirely different - it shouldn't return file:// path anyway.

crates/turbopack-build/src/ecmascript/node/entry/runtime.rs
kwonoj: https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack/loaders/next-metadata-route-loader.ts#L219,

PR mimic / matches behavior to the existing dynamic metadata passes with test cases. afaik it's input, the original location of the source as it can be _any_ static assets can be read - i.e `readFileSync(path.fileurlToPath(import.meta.url)`

crates/turbopack-ecmascript-runtime/js/src/build/runtime.ts
```diff
@@ -40,11 +42,32 @@ type ModuleFactory = (
 
 const path = require("path");
 const relativePathToRuntimeRoot = path.relative(RUNTIME_PUBLIC_PATH, ".");
+// Compute the relative path to the `distDir`.
+const relativePathToDistRoot = path.relative(path.join(OUTPUT_ROOT, RUNTIME_PUBLIC_PATH), ".");
```
jridgewell, line 46: `OUTPUT_ROOT` here is likely `PROJECT/.next`, no? In order to use the input paths, don't we need `PROJECT`? The calculation here depends on the output path being permanently locked to `.next`, which isn't the case because of `NextConfig.distDir` configuration.

jridgewell (changes requested):

crates/turbopack-ecmascript-runtime/js/src/build/runtime.ts
kwonoj: No, we pass it down from https://github.com/vercel/turbo/pull/6318/files/b5ee04c02ee1eeabd70091397de29f8b1f446220#diff-22085854b7935369bed1065b0669564cf7d2a0287b06e0bc31164f2138a9277eR51, means whatever we specify as distDist should be embedded correctly. If it's fixed to `./next` somehow, we have a bugs to setting output_root in the chunk context configuration.

kwonoj: both of the global `OUTPUT_ROOT` and `RUNTIME_PUBLIC_PATH` is being embedded from chunk context's configuration, so calculation should honor custom values if defined.

jridgewell (approved):

PR #6319, opened 2023-10-30
fix: global hash divergence for empty version strings

chris-olszewski:
### Description

@mehulkar uncovered a case where we had hash divergence between Rust/Go on Windows. It was initially caused by something in the prysk setup where `package.json` got a dependency of the form: `"monorepo": "file:"`. This caused an external dependency to be listed with an empty version string. This proved to be problematic as the Go capnp library [encodes empty strings as null pointers](https://github.com/capnproto/go-capnp/blob/main/struct.go#L139) instead of a zero length byte array as Rust does.

As the Go library exposes slightly more internals, I chose to change the Go code to circumvent the empty string behavior.

Future work might be disallowing empty version strings, but that's a larger change and something that requires more investigation. `npm` is happy to produce/parse/use a dependency with version `file:` so we should probably support it.

### Testing Instructions

Added unit tests for serializing lockfile packages with an empty version string from Rust and Go. The first commit of the PR adds these tests where it fails on Go to match the Rust hash.

Also can check against the case that @mehulkar uncovered:
 - `mkdir /tmp/test && cd /tmp/test`
 - `~/code/turbo/turborepo-tests/integration/tests/_helpers/setup_monorepo.sh $(pwd)`
 - Edit `package.json` to contain `"dependencies": { "monorepo": "file:" }`
 - Regenerate `package-lock.json` by `npm install`
 - `turbo build --dry=json` should now pass with a binary built using this PR

Closes TURBO-1557

### DIFF
--- a/cli/internal/fs/hash/capnp.go
+++ b/cli/internal/fs/hash/capnp.go
@@ -320,7 +320,10 @@ func HashLockfilePackages(packages []lockfile.Package) (string, error) {
 			return "", err
 		}
 
-		err = entry.SetVersion(pkg.Version)
+		// We explicitly write Version to match Rust behavior when writing empty strings
+		// The Go library will emit a null pointer if the string is empty instead
+		// of a zero length list.
+		err = capnp.Struct(entry).SetNewText(1, pkg.Version)
 		if err != nil {
 			return "", err
 		}
--- a/cli/internal/fs/hash/capnp_test.go
+++ b/cli/internal/fs/hash/capnp_test.go
@@ -92,6 +92,23 @@ func Test_LockfilePackagesNonEmpty(t *testing.T) {
 	assert.Equal(t, "9e60782f386d8ff1", hash)
 }
 
+func Test_LockfilePackagesEmptyVersion(t *testing.T) {
+	packages := []lockfile.Package{
+		{
+			Key:     "key",
+			Version: "",
+			Found:   true,
+		},
+	}
+
+	hash, err := HashLockfilePackages(packages)
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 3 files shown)


cli/internal/fs/hash/capnp.go
```diff
@@ -320,7 +320,10 @@ func HashLockfilePackages(packages []lockfile.Package) (string, error) {
 			return "", err
 		}
 
-		err = entry.SetVersion(pkg.Version)
+		// We explicitly write Version to match Rust behavior when writing empty strings
```
NicholasLYang, line 323: Does this count as a global cache key bump? Or is it more of a bug fix? If it's the former, maybe we should fix this on the Rust side.

NicholasLYang (changes requested):

cli/internal/fs/hash/capnp.go
chris-olszewski: I don't think we need to bump global cache key. All this change should affect (and I believe this is true since existing unit tests for global hash pass without any updates to the snapshots) are workspaces where there's a dependency with an empty version string. At the moment, I think this is only possible via `"foo": "file:"` so I think this should affect ~0 end users.

> maybe we should fix this on the Rust side

I'm very unsure of how to go about this without forking/upstream PR to the Rust capnp crate in order to copy the behavior of the Go library. Since we'll be deleting the Go side soon, I think if we're going to so something "hacky" it should be on the Go side.

NicholasLYang: Fair enough! If it's a bug fix then that works for me.

NicholasLYang (approved):

PR #6320, opened 2023-10-31
fix(Turborepo): Build x86_64-unknown-linux-gnu so that it works on AWS lambda

gsoltis:
### Description

 - Use `amazon/aws-lambda-nodejs:18` as a basis for building the `x86_64-unknown-linux-gnu` version of `@turbo/repository` so that it will be compatible with AWS lambda environments
 - Bump version to `0.0.1-canary.1`
 - Update `index.js` to prefer `musl` (which is not yet supported) for targets that don't have at least `glibc>=2.26`

### Testing Instructions

Manually verified that `x86_64-unknown-linux-gnu` artifact works on `amazon/aws-lambda-nodejs:18`, which has `glibc@2.26`

Closes TURBO-1558


gaspar09 (approved):
💯

PR #6321, opened 2023-10-31
Fixed hero section not properly covering screen height

danger-ahead:
Before change (section break visible abruptly in the hero section):
image

After Change:
image


docs/components/pages/pack-home/PackHero.tsx
```diff
@@ -9,7 +9,7 @@ import PackLogo from "../../logos/PackLogo";
 export function PackHero() {
   return (
     <FadeIn
-      className="font-sans w-auto pb-16 pt-[48px] md:pb-24 lg:pb-32 md:pt-16 lg:pt-20 flex justify-between gap-8 items-center flex-col relative z-0"
```
arlyon, line 12: 100vh is not consistent across devices (on mobile, the nav bar will sit on top). We could use `svh` here instead.

arlyon (changes requested):

docs/components/pages/pack-home/PackHero.tsx
danger-ahead: Agreed. Do I make the change in the PR?

arlyon (approved):

PR #6328, opened 2023-10-31
fix(Turborepo): Drop root from calculation of in-scope workspaces

gsoltis:
### Description

 - Don't fill in a default value for `pkg_inference_root` if we don't have one supplied
 - Don't set an empty `pkg_inference_root`
 - Don't enumerate the Root workspace when setting in-scope packages to all packages.

### Testing Instructions

Fixes `ordered/stream.t` integration test


gsoltis:
Need to update some output text due to dropping empty pkg inference strings.

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -148,6 +148,7 @@ impl<'a, T: PackageChangeDetector> FilterResolver<'a, T> {
             // return all packages in the workspace
             self.pkg_graph
                 .workspaces()
+                .filter(|(name, _)| **name != WorkspaceName::Root)
```
chris-olszewski, line 151: Nit: generally, I think it's better to use `matches!` instead of the `Eq` implementation here. I assume that perf would be pretty identical, but since `matches!` expands to a the match statement we want directly I think it's equivalent to `#[inline]` the eq function.
```suggestion
                .filter(|(name, _)| matches!(name, WorkspaceName::Other(_)))
```

chris-olszewski (approved):
Only thing that I think needs changing is the `package_inference_root` default on `ExecutionSummary`. You can test against `run-summary/error.t` and see that at the moment `repoRoot: null` when it should be `repoRoot: ""`. (there are other failures on that test unrelated to the PR that need to get fixed)

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -35,7 +35,7 @@ pub struct ExecutionSummary<'a> {
     // number of tasks that started
     attempted: usize,
     // the (possibly empty) path from the turborepo root to where the command was run
-    #[serde(rename = "repoPath", skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "repoPath", default = "AnchoredSystemPath::empty")]
```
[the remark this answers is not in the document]
gsoltis: I think it is called when serializing as well? Or at least, it seems to be generating the correct output.

At any rate, I think if we change it to non-optional, we need to change the field name as well. The pkg inference piece _is_ optional, but we can consider it non-optional to have a path into the repo where the execution originated. So from that standpoint, changing the field to `repo_path` and making it non-optional would make sense.

crates/turborepo-lib/src/run/scope/filter.rs
gsoltis: I'm fine with `matches!` here, the `**name` construction is definitely awkward. The other thought I had was implementing an `.is_root()` method on the enum, but decided that was maybe slightly overkill.

PR #6330, opened 2023-10-31
fix(turbo-ignore): better error for invalid commit

tknickman:
### Description

Adds a better error to cover the case when the commit to compare is invalid

Closes TURBO-1560

### DIFF
--- a/packages/turbo-ignore/src/errors.ts
+++ b/packages/turbo-ignore/src/errors.ts
@@ -2,25 +2,27 @@ import type { NonFatalErrorKey, NonFatalErrors } from "./types";
 
 export const NON_FATAL_ERRORS: NonFatalErrors = {
   MISSING_LOCKFILE: {
-    regex:
+    regex: [
       /reading (?:yarn.lock|package-lock.json|pnpm-lock.yaml):.*?no such file or directory/i,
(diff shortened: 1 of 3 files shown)


packages/turbo-ignore/src/errors.ts
```diff
@@ -2,25 +2,27 @@ import type { NonFatalErrorKey, NonFatalErrors } from "./types";
...
       /run failed: We did not detect an in-use package manager for your project/i,
+    ],
     message: `turbo-ignore could not complete - no package manager detected, please commit a lockfile, or set "packageManager" in your root "package.json"`,
   },
   UNREACHABLE_PARENT: {
-    regex: /failed to resolve packages to run: commit HEAD\^ does not exist/i,
+    regex: [/failed to resolve packages to run: commit HEAD\^ does not exist/i],
     message: `turbo-ignore could not complete - parent commit does not exist or is unreachable`,
   },
-  UNREACHABLE_COMMIT: {
```
tknickman, line 18: Removed this as the only time it would occur is when the comparison sha is bad, and the INVALID_COMPARISON error message is more descriptive and helpful

chris-olszewski (approved):

PR #6331, opened 2023-10-31
docs(showcase): Neo Financial

and-campestrini:
### Description

Added Neo Financial logo to /showcase page.


arlyon:
Hello, we are not currently accepting additions to the showcase list as this is handled in batches. We have added this PR to our list and will come back to it if/when we decide to expand on it. Thanks!

PR #6337, opened 2023-11-01
feat(examples): update non-monorepo to pnpm

tknickman:
### Description

Uses latest Next.js starter and updates to pnpm (not sure why this one was left behind in the great examples to pnpm migration)


pnpm-workspace.yaml
```diff
@@ -12,3 +12,5 @@ packages:
   - "crates/turbopack-tests/tests/execution"
   - "crates/*/js"
   - examples
+  # Exclude the non-monorepo example as it should be treated as independent of this repo
+  - "!examples/non-monorepo"
```
mehulkar, line 16: Is this necessary? `examples` itself is the workspace, not all subdirs inside `examples`, not sure why non-monorepo is different from all the other subdirectories

mehulkar (approved):
One q

pnpm-workspace.yaml
tknickman: Yea this is a little weird. I think this is a pnpm thing, all the other examples have pnpm workspace configs which acts as the "root" of the project so it stops traversing the directory at that point. Without a workspace config, or this exclusion, pnpm treats this example as part of the entire turbo repo and installs from root which causes a number of issues.

PR #5922, opened 2023-09-11
chore(turbo): Add workflow to update turbo in all examples

mehulkar:
Closes TURBO-1319

PR #6338, opened 2023-11-01
feat(examples): upgrade ts

tknickman:
### Description

Upgrade TS in examples to support utilizing the bundler resolution option. 

TODO:
- [ ] update yarn examples
- [ ] update npm examples

Closes TURBO-1563


scripts/update-examples-dep.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
tknickman, line 1: @mehulkar I could have sworn you already had a script similar to this - is it here and I'm missing it?

mehulkar: https://github.com/vercel/turbo/pull/5922

arlyon (approved):
Happy with this predicated on CI :) we have a PR that depends on this (https://github.com/vercel/turbo/pull/6396) so perhaps we can split it up and close both sooner rather than later

Issue #4291, opened 2023-03-21
[WEB-880] [turborepo] Persistent task concurrency validation not taking into account whether package has the script defined


shawnmcknight:
### What version of Turborepo are you using?

1.8.5

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Windows

### Describe the Bug

I believe this is related to #4205.  If you have a task which is designated as `persistent`, it will validate that concurrency limits are not breached.  However, the calculation of how many parallel tasks will be executed is not taking into account that not every package in the defined filter has the script defined.

For instance, I have a turbo task `dev` which is flagged as persistent.  This is launched as:
`turbo run dev --filter=@scope/app.*...`

The count of tasks gets indicated as 24 in my example, which breaches the concurrency limit of 10.  The count of 24 is the total count of packages matching the filter and all of its dependencies.  However, only 4 of the packages in the filter's scope actually have `dev` scripts defined.  This same issue does not occur if a filter is not defined.  Effectively, the validation of concurrency limits should verify if a package in the filter's scope actually has the script defined before erroring.  Packages without the script defined should not be counted in the validation.

### Expected Behavior

Since in my example only 4 `dev` scripts exist within the filter's scope, the turbo task should be permitted to run as it will not breach concurrency limits.

### To Reproduce

* Create a workspace with 2 packages -- `a` and `b`
  * Make `b` a dependency of `a`
* Define a `persistent` turbo task (e.g. `dev`) in `turbo.json`
* Add a script to package `a` matching the persistent task's name
* Execute `turbo run <task_name> --filter=a... --concurrency=1`

In this example, only package `a` will run a script because it's the only one with the script defined.  However, the count will be detected as exceeding the limit of 1 parallel task because it's using a count of all packages in scope and not the number of scripts which will be executed.

<sub>[WEB-880](https://linear.app/vercel/issue/WEB-880/[turborepo]-persistent-task-concurrency-validation-not-taking-into)</sub>

<sub>[TURBO-1141](https://linear.app/vercel/issue/TURBO-1141/[web-880]-[turborepo]-persistent-task-concurrency-validation-not)</sub>

rottbers:
Running into the same with npm. In my case only 4 out of 10 workspaces have the persistent script (`dev`) present in their `package.json` but yet all 10 workspaces seems to be targeted resulting in the following error:

```
ERROR run failed: error preparing engine: Invalid persistent task configuration:
[...]

PR #6339, opened 2023-11-01
feat: add support for --parallel

chris-olszewski:
### Description

This PR hooks up the correct behavior of `--parallel`

Some code shuffling happens in the first few commits to get the Rust `run` function in a closer shape the Go code.

### Testing Instructions

- Create a new repo with `npx create-turbo`
- Remove installed version of turbo `npm rm turbo`
- Add a `build` script to `packages/ui/package.json`: `"build": "echo 'building'; sleep 5; echo 'done'",`
- `turbo_dev build --parallel --experimental-rust-codepath` should show that `docs#build` and `web#build` start before`ui#build` is finished even though they depend on the `ui` workspace

Closes TURBO-1566

### DIFF
--- a/crates/turborepo-lib/src/package_graph/mod.rs
+++ b/crates/turborepo-lib/src/package_graph/mod.rs
@@ -120,30 +120,17 @@ impl PackageGraph {
         Ok(graph::validate_graph(&self.workspace_graph)?)
     }
 
-    pub fn no_workspace_dependencies(self) -> Self {
-        let Self {
-            mut workspace_graph,
-            node_lookup,
-            workspaces,
-            package_manager,
-            lockfile,
-        } = self;
-        let root_index = node_lookup
+    pub fn remove_workspace_dependencies(&mut self) {
+        let root_index = self
+            .node_lookup
             .get(&WorkspaceNode::Root)
             .expect("graph should have root workspace node");
-        workspace_graph.retain_edges(|graph, index| {
+        self.workspace_graph.retain_edges(|graph, index| {
             let Some((_src, dst)) = graph.edge_endpoints(index) else {
                 return false;
             };
             dst == *root_index
         });
-        Self {
-            workspace_graph,
-            node_lookup,
-            workspaces,
-            package_manager,
-            lockfile,
-        }
     }
 
     /// Returns the number of workspaces in the repo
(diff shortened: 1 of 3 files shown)


crates/turborepo-lib/src/package_graph/mod.rs
```diff
@@ -120,6 +120,32 @@ impl PackageGraph {
         Ok(graph::validate_graph(&self.workspace_graph)?)
     }
 
+    pub fn no_workspace_dependencies(self) -> Self {
```
mehulkar, line 123: Is `remove_workspace_dependencies` a better name?

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -322,6 +289,27 @@ impl<'a> Run<'a> {
...
         )?;
 
+        if opts.run_opts.parallel {
+            pkg_dep_graph = pkg_dep_graph.no_workspace_dependencies();
+            engine = self.build_engine(&pkg_dep_graph, &opts, &root_turbo_json, &filtered_pkgs)?;
+        }
+
+        if let Some(graph_opts) = opts.run_opts.graph {
+            match graph_opts {
+                GraphOpts::File(graph_file) => {
```
mehulkar, line 299: just making sure i understand this arm... it's not actually writing the graph to a file yet right? I don't see the `engine` or `pkg_dep_graph` used in here

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -566,4 +560,50 @@ impl<'a> Run<'a> {
...
+            opts.run_opts
+                .tasks
+                .iter()
+                .map(|task| TaskName::from(task.as_str()).into_owned()),
+        )
+        .build()?;
+
+        if !opts.run_opts.parallel {
+            engine
+                .validate(pkg_dep_graph, opts.run_opts.concurrency)
```
mehulkar, line 594: Just a note that we should fix #4291 with this in the future. I think it's fine to only fix in Rust if we're close

mehulkar (approved):

crates/turborepo-lib/src/run/mod.rs
chris-olszewski: Correct, one of the features left is support for non-`DOT` `--graph` support. The `todo` macro call in this branch will panic when hit.

PR #6340, opened 2023-11-01
fix: Empty command returns help text and not error

NicholasLYang:
### Description
Our command version test that runs `turbo -v` returns the help text on the rust codepath. This is because if `EXPERIMENTAL_RUST_CODEPATH` is set, it implicitly creates the `RunArgs` struct in `Args`, which then makes our code think that the user passed a command when they didn't.

We solve this by moving the `experimental_rust_codepath` field to the global `Args` struct (it doesn't really matter where the flag is because it's a purely internal). 

### Testing Instructions

Closes TURBO-1567

### DIFF
--- a/benchmark/src/ttft.ts
+++ b/benchmark/src/ttft.ts
@@ -30,7 +30,7 @@ if (!fs.existsSync(TURBO_BIN)) {
   throw new Error("No turbo binary found");
 }
 
-const turboFlags = `-vvv --experimental-rust-codepath --dry --skip-infer --profile=${fullProfilePath}`;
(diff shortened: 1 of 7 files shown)


chris-olszewski (changes requested):
This moves the problem to `no_args.t` where we now have differing behavior between:
`turbo` and `EXPERIMENTAL_RUST_CODEPATH=false turbo`

NicholasLYang:
Hmm yeah we do, but that is the same differing behavior that we already have with `--no-update-notifier` and another other global flag. I guess what I can do is move the environment variable processing out of clap so it doesn't implicitly set anything

chris-olszewski (approved):

PR #6341, opened 2023-11-01
chore(js): output packages to ESM only

mehulkar:
These packages are only used as binaries, they are not meant to be imported. We also [require Node 18+](https://turbo.build/repo/docs/support#nodejs-compatibility), so ESM by default should work for everyone.

Closes TURBO-1568


mehulkar:
This might take a little more effort. Current blockers:

- `import`ing some files without file extensions, which doesn't work in ESM (may be possible with some config in tsconfig or tsup)
- importing `fs-extra` and using various sync methods that are not available on `fs-extra/esm` entrypoint

mehulkar:
Closing this down as it uncovered some issues with underlying libraries. We'll likely have to do this one package at a time.

PR #6342, opened 2023-11-01
docs: Added nested workspaces under troubleshooting

NicholasLYang:
### Description

We've had a couple people come to us with issues caused by nested workspaces. I figured we should call this out as unsupported behavior.

### Testing Instructions

Closes TURBO-1569

### DIFF
--- a/docs/pages/repo/docs/troubleshooting.mdx
+++ b/docs/pages/repo/docs/troubleshooting.mdx
@@ -75,6 +75,11 @@ directions even if you aren't planning to make a contribution.
 
 ## Common Pitfalls
 
+### Nested workspaces
+
+Turborepo does not support nested workspaces. Therefore, if you have a workspace that contains another workspace,
+you may see unexpected behavior. We recommend flattening your workspace structure to avoid this issue.
+
 ### The `.turbo` directory
 
 One of the [core concepts][2] behind Turbo is that when a declared input


docs/pages/repo/docs/troubleshooting.mdx
```diff
@@ -75,6 +75,11 @@ directions even if you aren't planning to make a contribution.
 
 ## Common Pitfalls
 
+### Nested workspaces
+
+Turborepo does not support nested workspaces. Therefore, if you have a workspace that contains another workspace,
```
NicholasLYang, line 80: Should we go into more detail as to what a nested workspace is?

mehulkar: Or link to another page in handbook that explains what it is

mehulkar (approved):

docs/pages/repo/docs/troubleshooting.mdx
NicholasLYang: Do we have such a page? It might just be worth fleshing out here

mehulkar: Not sure, just thinking that it's a general monorepo concept and as such can be defined somewhere else. We can do the easy thing and just flesh out here too, just a suggestion

PR #6350, opened 2023-11-02
feat(Turborepo): Next revision of @turbo/repository

gsoltis:
### Description

 - Move most code to an `internal` module
 - Start returning an object for workspace enumeration
 - Move `detectJS` to be async
 - Use the generated types directly, they're good enough with some `napi` annotations
 - Add some basic testing infrastructure

### Testing Instructions

New tests added, mostly just proving that we can run tests and it doesn't immediately crash. More detailed tests to come.

Closes TURBO-1576


arlyon (approved):
I like this. We could almost go as far as having lib.rs _only_ have _pub use_ statements to make it crystal clear what we export but this is small enough as-is.

packages/turbo-repository/__tests__/detect.test.ts
```diff
@@ -0,0 +1,19 @@
+import * as path from "node:path";
+import { Repository } from "../js/dist/index.js";
+
+describe("detect", () => {
```
tknickman, line 4: If it's helpful in the future we have a lot of testing utils in https://github.com/vercel/turbo/tree/main/packages/turbo-test-utils/src that we use throughout the JS packages to make creating, using, and tearing down fixtures really simple.

packages/turbo-repository/jest.config.js
```diff
@@ -0,0 +1,12 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
+  preset: "ts-jest/presets/js-with-ts",
+  testEnvironment: "node",
+  testPathIgnorePatterns: ["__fixtures__/", "/__tests__/test-utils.ts"],
+  coveragePathIgnorePatterns: ["__fixtures__/", "/__tests__/test-utils.ts"],
+  transformIgnorePatterns: ["node_modules/*", "packages/turbo-workspaces/*"],
```
tknickman, line 7: I think you probably don't need `"packages/turbo-workspaces/*"`

tknickman (approved):

PR #6351, opened 2023-11-02
Turbopack: Attribute missing module issues to the import site

wbinnssmith:
This attaches source information to `ResolvingIssues` created from esm asset references.

Test Plan: Updated snapshot tests, which now include source location code frames.


crates/turbopack-core/src/issue/mod.rs
```diff
@@ -424,7 +423,7 @@ impl IssueSource {
                     } else {
                         SourcePos {
                             line: i - 1,
-                            column: offset - lines[i - 1].bytes_offset,
+                            column: offset - lines[i - 1].bytes_offset - 1,
```
wbinnssmith, line 426: This corrects an off-by-one issue. See the updated CJS snapshot test codeframe

crates/turbopack-ecmascript/src/analyzer/imports.rs
```diff
@@ -176,18 +183,25 @@ impl ImportMap {
...
     fn ensure_reference(
         &mut self,
         module_path: JsWord,
         imported_symbol: ImportedSymbol,
+        span: Span,
         annotations: ImportAnnotations,
     ) -> usize {
+        let issue_source = self
+            .source
+            .map(|s| IssueSource::from_byte_offset(s, span.lo().to_usize(), span.hi().to_usize()));
```
kdy1, line 199:
```suggestion
            .map(|s| IssueSource::from_byte_offset(s, span.lo.to_usize(), span.hi.to_usize()));
```

jridgewell, line 199: Can we defer this work? This requires a scan of the contents.

jridgewell (changes requested):

crates/turbopack-ecmascript/src/analyzer/imports.rs
```diff
@@ -176,18 +183,25 @@ impl ImportMap {
...
     current_annotations: ImportAnnotations,
+    source: Option<Vc<Box<dyn Source>>>,
 }
 
 impl<'a> Analyzer<'a> {
     fn ensure_reference(
         &mut self,
         module_path: JsWord,
         imported_symbol: ImportedSymbol,
+        span: Span,
```
kdy1, line 194: nitpick: Typically `Span` comes first if there's only one `Span` in the parameter list in SWC.
It's for SWC, not turbopack, but just FYI

jridgewell (approved):

crates/turbopack-core/src/issue/mod.rs
sokra: TBH That doesn't look like the right location for the fix. Why is the `- 1` needed here?

Is `offset` 1-based? (If yes, could we make it 0-based earlier in the chain)
Is `bytes_offset` off by 1? (If yes, could be fix that earlier in the chain)

sokra (changes requested):
Looks good in general

only comment: https://github.com/vercel/turbo/pull/6351#discussion_r1382887013

crates/turbopack-core/src/issue/mod.rs
wbinnssmith: Looks like `offset` (from the ast spans we get from swc) is 1-indexed.

crates/turbopack-css/src/references/mod.rs
```diff
@@ -127,13 +127,18 @@ impl<'a> VisitAstPath for ModuleReferencesVisitor<'a> {
         };
 
         let issue_span = i.href.span();
+        println!(
```
sokra, line 130: Println

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -410,9 +410,22 @@ pub struct LazyIssueSource {
...
+    /// Arguments:
+    ///
+    /// * `source`: The source code in which to look up the byte offsets.
+    /// * `start`: The start index of the span. Must use **1-based** indexing.
+    /// * `end`: The end index of the span. Must use **1-based** indexing.
     #[turbo_tasks::function]
-    pub fn new(source: Vc<Box<dyn Source>>, start: usize, end: usize) -> Vc<Self> {
-        Self::cell(LazyIssueSource { source, start, end })
+    pub fn from_swc_offsets(source: Vc<Box<dyn Source>>, start: usize, end: usize) -> Vc<Self> {
+        println!("received {} {}", start, end);
```
sokra: Println

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -410,9 +410,22 @@ pub struct LazyIssueSource {
...
+        Self::cell(LazyIssueSource {
+            source,
+            start: start - 1,
```
sokra, line 425: Could we somehow handle the case when start or end are 0? Currently this will panic in that case in dev, or gives weird numbers in release

jridgewell (approved):

sokra (approved):

PR #6356, opened 2023-11-03
port(turborepo): Rust test fixes

NicholasLYang:


mehulkar (approved):
This is great! Although I think before we make it public, we should think about splitting up the behavior into `turbo config` and `turbo ls` options

NicholasLYang:
Yeah for turbo info to become a proper public command, we probably need an RFC and maybe some customer feedback. But for now it's a good stopgap testing solution

PR #6359, opened 2023-11-03
Turbopack: Defer calculating source positions until needed

wbinnssmith:
Previously, many references would eagerly calculate their row/column in the source code even if they did not cause an `Issue` to occur. This requires reading the source code and calculating this information from byte offsets from the swc node's `Span`.

This introduces `LazyIssueSource`, which simply wraps the `Source` and the start/end byte offsets from the swc span. This is only converted to an `IssueSource` when the issue's source location is accessed.

### DIFF
--- a/crates/turbopack-core/src/issue/mod.rs
+++ b/crates/turbopack-core/src/issue/mod.rs
@@ -396,6 +396,10 @@ impl CapturedIssues {
     }
 }
 
+/// Use this to pass and store byte offsets for an AST node along with its
+/// source. When row/column is needed, this can be lazily converted into a
+/// proper `IssueSource` at that time using
+/// [LazyIssueSource::to_issue_source].
 #[turbo_tasks::value]
 #[derive(Clone, Debug)]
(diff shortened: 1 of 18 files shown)


crates/turbopack-core/src/issue/mod.rs
```diff
@@ -397,11 +397,37 @@ impl CapturedIssues {
...
+
+#[turbo_tasks::value_impl]
+impl LazyIssueSource {
+    #[turbo_tasks::function]
+    pub fn new(source: Vc<Box<dyn Source>>, start: usize, end: usize) -> Vc<Self> {
+        Self::cell(LazyIssueSource { source, start, end })
+    }
+
+    #[turbo_tasks::function]
+    pub async fn to_issue_source(self: Vc<Self>) -> Result<Vc<IssueSource>> {
```
wbinnssmith, line 415: I couldn't find a straightforward way to use Rust's `From`/`To` with Vcs. Any ideas?

jridgewell (approved):

PR #6361, opened 2023-11-03
fix(urlassetref): get the static src property from asset

kwonoj:
### Description

If the path to the url is a static value, turbopack returns its asset output ident to the wrapper module - unlike other plain assets, its default export is an object contains metadata values of the original asset. In result, call to `new URL` returns stringified module symbol instead of actual path.

PR adjusts lookup starts from the export's property, then falls back to the export itself if it's not available.


wbinnssmith (approved):

jridgewell (changes requested):
Trying to review now

kwonoj:
Putting on hold until get clarification, easy to bring it back as PR.

PR #6369, opened 2023-11-06
refactor(turborepo): Changed cache misses to be a non-error

NicholasLYang:
### Description

Originally we had cache misses be an error. That was a mistake because it was too easy to propagate a cache miss up the stack and not handle it properly. It was also too easy to not convert the error into a cache miss. Now we have a `CacheResult` type that indicates either a miss or a hit.

### Testing Instructions

Closes TURBO-1585

### DIFF
--- a/crates/turborepo-cache/src/lib.rs
+++ b/crates/turborepo-cache/src/lib.rs
@@ -12,7 +12,7 @@ pub mod signature_authentication;
 #[cfg(test)]
 mod test_cases;
 
-use std::{backtrace, backtrace::Backtrace, fmt};
+use std::{backtrace, backtrace::Backtrace};
 
 pub use async_cache::AsyncCache;
 use camino::Utf8Path;
@@ -80,26 +80,45 @@ pub enum CacheSource {
     Remote,
 }
 
-impl<T: fmt::Debug> fmt::Debug for CacheResult<T> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            CacheResult::Miss => write!(f, "CacheResult::Miss"),
-            CacheResult::Hit(inner) => {
-                write!(f, "CacheResult::Hit({:?})", inner)
-            }
-        }
-    }
-}
-
 // This is generic because exists returns just the metadata
 // while fetch returns the metadata and the files
-#[derive(Clone)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum CacheResult<T> {
     Miss,
     Hit(T),
 }
 
-#[derive(Debug, Clone)]
+impl<T> CacheResult<T> {
+    pub fn expect_hit(self) -> T {
+        match self {
+            CacheResult::Hit(value) => value,
+            CacheResult::Miss => panic!("expected cache hit"),
+        }
+    }
+
+    pub fn is_hit(&self) -> bool {
+        match self {
+            CacheResult::Hit(_) => true,
+            CacheResult::Miss => false,
+        }
+    }
+
+    pub fn is_miss(&self) -> bool {
+        match self {
+            CacheResult::Hit(_) => false,
+            CacheResult::Miss => true,
+        }
+    }
+
+    pub fn expect_miss(self) {
+        match self {
+            CacheResult::Hit(_) => panic!("expected cache miss"),
+            CacheResult::Miss => (),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
 pub struct CacheHitMetadata {
     pub source: CacheSource,
     pub time_saved: u64,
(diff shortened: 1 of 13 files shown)


crates/turborepo-lib/src/task_hash.rs
```diff
@@ -164,7 +164,7 @@ pub struct TaskHashTrackerState {
     #[serde(skip)]
     package_task_outputs: HashMap<TaskId<'static>, Vec<AnchoredSystemPathBuf>>,
     #[serde(skip)]
-    package_task_cache: HashMap<TaskId<'static>, CacheResponse>,
+    package_task_cache: HashMap<TaskId<'static>, CacheResult<CacheHitMetadata>>,
```
chris-olszewski, line 167: We can keep this as just `CacheHitMetadata` since we only ever insert cache hits
```suggestion
    package_task_cache: HashMap<TaskId<'static>, CacheHitMetadata>,
```

crates/turborepo-lib/src/task_hash.rs
```diff
@@ -457,12 +457,16 @@ impl TaskHashTracker {
         state.package_task_outputs.insert(task_id, outputs);
     }
 
-    pub fn cache_status(&self, task_id: &TaskId) -> Option<CacheResponse> {
+    pub fn cache_status(&self, task_id: &TaskId) -> Option<CacheResult<CacheHitMetadata>> {
```
chris-olszewski, line 460:
```suggestion
    pub fn cache_status(&self, task_id: &TaskId) -> Option<CacheHitMetadata> {
```

crates/turborepo-lib/src/task_hash.rs
```diff
@@ -457,12 +457,16 @@ impl TaskHashTracker {
...
         let state = self.state.lock().expect("hash tracker mutex poisoned");
-        state.package_task_cache.get(task_id).copied()
+        state.package_task_cache.get(task_id).cloned()
```
chris-olszewski, line 462: If we move this back to just `CacheHitMetadata` we can move it back to `copied()`

crates/turborepo-cache/src/lib.rs
```diff
@@ -82,8 +80,46 @@ pub enum CacheSource {
...
+
+    pub fn expect_miss(self) {
+        match self {
+            CacheResult::Hit(_) => panic!("expected cache miss"),
+            CacheResult::Miss => (),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
```
chris-olszewski, line 121:
```suggestion
#[derive(Debug, Clone, PartialEq, Copy)]
```

crates/turborepo-lib/src/task_hash.rs
```diff
@@ -457,12 +457,16 @@ impl TaskHashTracker {
...
     }
 
-    pub fn insert_cache_status(&self, task_id: TaskId<'static>, cache_status: CacheResponse) {
+    pub fn insert_cache_status(
+        &self,
+        task_id: TaskId<'static>,
+        cache_status: CacheResult<CacheHitMetadata>,
```
chris-olszewski, line 468:
```suggestion
        cache_status: CacheHitMetadata,
```

crates/turborepo-lib/src/run/summary/task.rs
```diff
@@ -131,30 +131,36 @@ impl TaskCacheSummary {
...
-            };
-            Self {
-                local,
-                remote,
-                status: CacheStatus::Hit,
-                source: Some(source),
-                time_saved,
+// This is an `Option<CacheResult<CacheHitMetadata>>` because we could fail
+// to fetch from cache (giving `None`), and we could also get a miss
+// `Some(CacheResult::Miss)`
```
chris-olszewski, line 136: I think it might make life easier to just keep this as `Option<CacheHitMetadata>` since we treat `None` and `Some(None`) the same.

crates/turborepo-cache/src/lib.rs
```diff
@@ -82,8 +80,46 @@ pub enum CacheSource {
...
 
-#[derive(Debug, Clone, PartialEq, Copy)]
-pub struct CacheResponse {
+// This is generic because exists returns just the metadata
+// while fetch returns the metadata and the files
+#[derive(Clone, Debug, PartialEq)]
+pub enum CacheResult<T> {
+    Miss,
+    Hit(T),
+}
```
chris-olszewski, line 89: I think we should just get rid of this bespoke `Option` implementation unless there's some benefit we get from this that I'm not understanding.

chris-olszewski (changes requested):
Big fan of no longer having cache misses show up as errors, but I think we can get away with just keeping stuff as `Option<CacheHit>` instead of our own enum.

Also, until we have a reason to differentiate between a cache miss and a cache error in the task hash tracker we should keep the Go behavior where we flatten those both into a "cache miss". (Joys of Go where accessing a map with a non-existent key gives back the zero variant of the type which in our case is a cache miss)

chris-olszewski (approved):

PR #6373, opened 2023-11-06
fix: Move package manager tests to turbo info

NicholasLYang:
### Description

Moving more tests off of `--__test-run` and into `turbo info`

### Testing Instructions

Closes TURBO-1591

### DIFF
--- a/crates/turborepo-lib/src/commands/info.rs
+++ b/crates/turborepo-lib/src/commands/info.rs
@@ -16,12 +16,15 @@ use crate::{
 };
 
 #[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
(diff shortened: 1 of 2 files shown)


mehulkar (approved):
Nice, I actually took a stab at this this morning, and failed to make the compiler happy. I feel like I was just missing the` #[serde(rename_all = "camelCase")]` bit?

NicholasLYang:
Yeah, could be! What was the error you were getting?

mehulkar:
@NicholasLYang this is the diff I still had on my machine: https://github.com/vercel/turbo/pull/6377, but I think I tried something closer to what you had as well in my first pass (using a lifetime and also just capturing packageManager instead of packageGraph)

PR #6374, opened 2023-11-06
chore(test): combine two fixtures for single package testing

mehulkar:
Similar to #6366, the deleted fixture also contained some bash code that doesn't run on Windows, so this seems like the easiest solution.

I also moved all these tests inside the `run` directory becuase that's where I usually look for what we have covered. We could probably use a little more organization for discoverability

Closes TURBO-1592


turborepo-tests/integration/tests/_fixtures/single_package/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "my-pkg",
   "scripts": {
-    "build": "echo 'building' > foo"
+    "build": "echo 'building' > foo",
+    "test": "echo 'testing'"
```
chris-olszewski, line 5: Can we change this to `cat foo`? The `[[ ( -f foo ) && $(cat foo) == 'building' ]]` script depended on the outputs of `build` being present. From reading through the changes I don't think we ended up ever testing output restoration using this, but I don't think it can hurt to keep that aspect of test task.
```suggestion
    "test": "cat foo"
```

chris-olszewski (approved):
LGTM once the integration tests get updated to pass

turborepo-tests/integration/tests/_fixtures/single_package/package.json
mehulkar: good idea, done

PR #6376, opened 2023-11-07
feat(turbopack): support url rewrite behavior options

kwonoj:
### Description

This PR attempts to address supporting different approach to handle `new URL(...)` depends on the caller's context. Mostly, it aims to mimic options like https://webpack.js.org/configuration/module/#moduleparserjavascripturl.

In next.js, depends on the compilation context it specifies how `new URL` is being treated (https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack-config.ts#L1384 / https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack-config.ts#L1861) so generated output can resolve to the given url seamleassly regardless of running context. Turbopack currently rewrites url every time, only cares of rendering context so it is not possible - there is a rendering context differs to the actual environment that codes running (i.e `pages/api`).

To drill down hole to pass those options, PR adds a new option for the `ModuleOptionContext`, then pass it down to urlassetreference via `EcmaScriptOptions`. The actual implementation for the new rewrite behavior is not included as to verify the approach for the option itself as first step.

### DIFF
--- a/crates/turbopack-ecmascript/src/lib.rs
+++ b/crates/turbopack-ecmascript/src/lib.rs
@@ -34,6 +34,7 @@ use code_gen::CodeGenerateable;
 pub use parse::ParseResultSourceMap;
 use parse::{parse, ParseResult};
 use path_visitor::ApplyVisitors;
+use references::esm::UrlRewriteBehavior;
 pub use references::{AnalyzeEcmascriptModuleResult, TURBOPACK_HELPER};
 pub use static_code::StaticEcmascriptCode;
 use swc_core::{
@@ -94,6 +95,10 @@ pub struct EcmascriptOptions {
     pub import_parts: bool,
     /// module is forced to a specific type (happens e. g. for .cjs and .mjs)
     pub specified_module_type: SpecifiedModuleType,
+    /// Determines how to treat `new URL(...)` rewrites.
+    /// This allows to construct url depends on the different building context,
+    /// e.g. SSR, CSR, or Node.js.
+    pub url_rewrite_behavior: Option<UrlRewriteBehavior>,
 }
 
 #[turbo_tasks::value(serialization = "auto_for_input")]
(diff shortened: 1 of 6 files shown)


crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -94,6 +95,7 @@ pub struct EcmascriptOptions {
     pub import_parts: bool,
     /// module is forced to a specific type (happens e. g. for .cjs and .mjs)
     pub specified_module_type: SpecifiedModuleType,
+    pub url_rewrite_behavior: Option<UrlRewriteBehavior>,
```
jridgewell, line 98: Docs, please.

crates/turbopack-ecmascript/src/references/esm/url.rs
```diff
@@ -25,6 +25,20 @@ use crate::{
     utils::module_id_to_lit,
 };
 
+/// Determines how to treat `new URL(...)` rewrites.
+/// This allows to construct url depends on the different building context,
+/// e.g. SSR, CSR, or Node.js.
+#[turbo_tasks::value(shared)]
+#[derive(Debug, Copy, Clone, PartialOrd, Ord, Hash)]
+pub enum UrlRewriteBehavior {
+    /// Omits base, resulting in a relative URL.
```
jridgewell, line 34: What environment would you use a relative URL?

kwonoj: https://vercel.slack.com/archives/C03EWR7LGEN/p1699297737138929?thread_ts=1699294919.660049&cid=C03EWR7LGEN

Instead of rewriting every hostname, webpack seems to handle it by having relative url instead. As noted in PR body we have a config for it as well.

> In next.js, depends on the compilation context it specifies how new URL is being treated (https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack-config.ts#L1384 / https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack-config.ts#L1861)

jridgewell (approved):

PR #6384, opened 2023-11-07
chore: fix clippy lints in tests

chris-olszewski:
### Description

Fix some clippy lints in our tests that have been bothering me for a bit.

These aren't caught by CI since we're not enabling the `test` feature and I would much rather have clippy check production codepath as opposed to the test codepath. There does exists a `--tests` option for `clippy`, but we'd need to teach `cargo-groups` how to properly forward it.

### Testing Instructions

👀 

Closes TURBO-1600

### DIFF
--- a/crates/turborepo-lib/src/engine/builder.rs
+++ b/crates/turborepo-lib/src/engine/builder.rs
@@ -1030,6 +1030,6 @@ mod test {
                 }
             })
             .err();
-        assert_eq!(result.as_deref(), reason.as_deref());
+        assert_eq!(result.as_deref(), reason);
     }
 }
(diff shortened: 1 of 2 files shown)


mehulkar (approved):
Yeah can we just make lint work, feel like this would be really helpful for newbs like me

chris-olszewski:
> Yeah can we just make lint work

It does now at least stop production code from having clippy lints, but it doesn't test the code gated behind the `test` feature. Getting this enabled should be possible, but it requires some dev work for the tool we use to run the lints.

mehulkar:
> > Yeah can we just make lint work
> 
> ... it doesn't test the code gated behind the `test` feature. Getting this enabled should be possible, but it requires some dev work

that's what I'm asking for :)

PR #6388, opened 2023-11-08
Turbopack: Implement structured styled text and use it in issue descriptions

wbinnssmith:
This implements a basic structure representing styled text and uses it in issue descriptions. This can be used by the cli and web-based error reporting.

Tested with an upcoming Next.js PR


crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,41 @@ impl Display for IssueSeverity {
     }
 }
 
+#[derive(Clone, Debug)]
+#[turbo_tasks::value(shared)]
+pub enum StyledString {
```
kwonoj, line 78: not going to suggest to make any changes, just random thoughts that if we could just pass raw markdown or formatted text and consumer (next-swc) can transparently renders it'd be cool

kwonoj, line 78: but in any case, probably worth to leave some comments what it is for?

crates/turbopack-cli-utils/src/issue.rs
```diff
@@ -570,3 +579,18 @@ fn show_all_message_with_shown_count(
...
+fn render_styled_string_to_ansi(styled_string: &StyledString) -> String {
+    match styled_string {
+        StyledString::Line(parts) => {
+            let mut string = String::new();
+            for part in parts {
+                string.push_str(&render_styled_string_to_ansi(part));
+            }
+            string
+        }
+        StyledString::String(string) => string.to_string(),
```
kwonoj, line 592: What are the main differences between lines vs string?

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,42 @@ impl Display for IssueSeverity {
...
```
jridgewell, line 78: Docs.

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,42 @@ impl Display for IssueSeverity {
...
+    fn deterministic_hash<H: turbo_tasks_hash::DeterministicHasher>(&self, state: &mut H) {
+        match self {
+            StyledString::Line(parts) => {
+                for part in parts {
+                    part.deterministic_hash(state);
+                }
+            }
+            StyledString::String(s) | StyledString::Pre(s) | StyledString::Strong(s) => {
+                // Normalize syspaths from Windows. These appear in stack traces.
+                s.replace('\\', "/").deterministic_hash(state);
```
jridgewell, line 95: I don't think this is the right spot for this, whatever created the `StyledString` should have already normalized.

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,42 @@ impl Display for IssueSeverity {
...
+
+impl DeterministicHash for StyledString {
+    fn deterministic_hash<H: turbo_tasks_hash::DeterministicHasher>(&self, state: &mut H) {
+        match self {
+            StyledString::Line(parts) => {
+                for part in parts {
+                    part.deterministic_hash(state);
+                }
+            }
+            StyledString::String(s) | StyledString::Pre(s) | StyledString::Strong(s) => {
```
jridgewell, line 93: You need to differentiate the enum variants.

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,42 @@ impl Display for IssueSeverity {
     }
 }
 
+#[derive(Clone, Debug)]
+#[turbo_tasks::value(shared)]
+pub enum StyledString {
+    Line(Vec<StyledString>),
+    String(String),
+    Pre(String),
```
sokra, line 81: I would rename this to `Code`. `pre` is a bit html specific

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,42 @@ impl Display for IssueSeverity {
...
+    Strong(String),
+}
```
sokra, line 83:
```suggestion
pub enum StyledString {
    /// Multiple [StyledString]s concatenated into a single line. Each item is considered as inline element.
    Line(Vec<StyledString>),
    /// Multiple [StyledString]s stacked vertically. They are considered as block elements, just like the top level [StyledString].
    Stack(Vec<StyledString>),
    /// Some prose text.
    Text(String),
    /// Code snippet.
    // TODO add language to support syntax hightlighting
    Code(String),
    /// Some important text.
    Strong(String),
}
```

crates/turbopack-core/src/issue/mod.rs
sokra: Technically `Issue`s should not have platform specific content. I think this is primarily caused by embedding node.js errors/stack traces into issues, which are platform dependent.

I would recommend to just remove this normalization here.

crates/turbopack-dev-server/src/update/stream.rs
```diff
@@ -55,7 +56,7 @@ async fn get_update_stream_item(
             plain_issues.push(
                 FatalStreamIssue {
                     resource: resource.to_string(),
-                    description: Vc::cell(format!("{}", PrettyPrintError(&e))),
+                    description: StyledString::String(format!("{}", PrettyPrintError(&e))).cell(),
```
sokra, line 59: We might want a helper to pretty print an error directly into a styled string, so we can use formatting during pretty printing. But that's something for a follow-up PR

crates/turbopack-dev/src/react_refresh.rs
```diff
@@ -89,11 +89,12 @@ impl Issue for ReactRefreshResolvingIssue {
     }
 
     #[turbo_tasks::function]
-    fn description(&self) -> Vc<String> {
-        Vc::cell(
+    fn description(&self) -> Vc<StyledString> {
+        StyledString::String(
             "React Refresh will be disabled.\nTo enable React Refresh, install the \
              `react-refresh` and `@next/react-refresh-utils` modules."
```
sokra, line 95: Code

sokra: I would be nice to have a `styled!` macro that converts markdown to styled string. But that's probably a lot work...

crates/turbopack-ecmascript-hmr-protocol/src/lib.rs
```diff
@@ -135,7 +135,7 @@ pub struct Issue<'a> {
     pub category: &'a str,
 
     pub title: &'a str,
-    pub description: &'a str,
+    pub description: StyledString,
```
sokra, line 138:
```suggestion
    pub description: &'a StyledString,
```

No need to clone that

crates/turbopack-env/src/issue.rs
```diff
@@ -1,12 +1,12 @@
...
 use turbo_tasks_fs::FileSystemPath;
-use turbopack_core::issue::Issue;
+use turbopack_core::issue::{Issue, StyledString};
 
 /// An issue that occurred while resolving the parsing or evaluating the .env.
 #[turbo_tasks::value(shared)]
 pub struct ProcessEnvIssue {
     pub path: Vc<FileSystemPath>,
-    pub description: Vc<String>,
+    pub description: String,
```
sokra, line 9:
```suggestion
    pub description: Vc<StyledString>,
```

crates/turbopack-image/src/process/mod.rs
```diff
@@ -481,7 +477,7 @@ pub async fn optimize(
 #[turbo_tasks::value]
 struct ImageProcessingIssue {
     path: Vc<FileSystemPath>,
-    message: Vc<String>,
+    message: String,
```
sokra, line 480:
```suggestion
    message: Vc<StyledString>,
```

crates/turbopack-core/src/issue/mod.rs
wbinnssmith: My concern was that the `Code` name wouldn't align with uses like styling file paths as monospaced.

crates/turbopack-cli-utils/src/issue.rs
wbinnssmith: A line can contain many StyledStrings composed together into a line. It should also print a newline.

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,77 @@ impl Display for IssueSeverity {
...
+
+impl DeterministicHash for StyledString {
+    fn deterministic_hash<H: turbo_tasks_hash::DeterministicHasher>(&self, state: &mut H) {
+        match self {
+            StyledString::Line(parts) => {
+                "line_start".deterministic_hash(state);
+                for part in parts {
+                    part.deterministic_hash(state);
+                }
+                "line_end".deterministic_hash(state);
```
kdy1, line 106: I think we don't need this, as we are just hashing
```suggestion

```

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,77 @@ impl Display for IssueSeverity {
...
+            }
+            StyledString::Stack(parts) => {
+                "stack_start".deterministic_hash(state);
+                for part in parts {
+                    part.deterministic_hash(state);
+                }
+                "stack_end".deterministic_hash(state);
```
kdy1, line 113: Same here
```suggestion
```

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,77 @@ impl Display for IssueSeverity {
...
+            }
+            StyledString::Text(s) => {
+                "text_start".deterministic_hash(state);
+                s.deterministic_hash(state);
+                "text_end".deterministic_hash(state);
```
kdy1, line 118:
```suggestion

```

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,77 @@ impl Display for IssueSeverity {
...
+            }
+            StyledString::Code(s) => {
+                "code_start".deterministic_hash(state);
+                s.deterministic_hash(state);
+                "code_end".deterministic_hash(state);
```
kdy1, line 123:
```suggestion
```

crates/turbopack-core/src/issue/mod.rs
```diff
@@ -73,6 +73,77 @@ impl Display for IssueSeverity {
...
+            }
+            StyledString::Strong(s) => {
+                "strong_start".deterministic_hash(state);
+                s.deterministic_hash(state);
+                "strong_end".deterministic_hash(state);
```
kdy1, line 128:
```suggestion
```

crates/turbopack-core/src/issue/mod.rs
wbinnssmith: These are re: https://github.com/vercel/turbo/pull/6388#discussion_r1385897337

kdy1: I mean `"line_start".deterministic_hash(state);` is enough for hash functions.

jridgewell (approved):

PR #6394, opened 2023-11-08
performance improvement and fixes

sokra:
### Description

some small performance improvements
and
a little fix where collectibles are not removed when a task becomes dirty during execution


crates/turbo-tasks-memory/src/task.rs
```diff
@@ -1091,6 +1093,16 @@ impl Task {
                                 .aggregation_leaf
                                 .change_job(&aggregation_context, change),
                         );
+                    } else if let Some(collectibles) = outdated_collectibles {
```
ForsakenHarmony, line 1096: maybe create `change` inside an if expression to deduplicate creating the `change_job`

PR #6395, opened 2023-11-08
Use a queue to early merge changes during aggregation

sokra:
### Description

This uses a ChangesQueue to aggregate changes in the aggregation tree in a breath-first order. The queue also merges changes to the same node early on and avoids bubbling up multiple changes along a whole branch.


anthonyshew:
Closing as stale since Turbopack is no longer in this repo!

PR #6396, opened 2023-11-08
Improved TSConfigs in examples

mattpocock:
### Description

Updated the tsconfig.json file in the basic example to reflect modern practices.

NOTE: this needs to be copied to all the other examples. Wasn't sure if there was some clever automation here, and couldn't quite be arsed to do it myself.

Reasoning behind changes explained here: https://www.totaltypescript.com/tsconfig-cheat-sheet

I also deleted a lot of values which are default.


examples/basic/packages/tsconfig/base.json
```diff
@@ -2,20 +2,18 @@
...
-    "preserveWatchOutput": true,
     "skipLibCheck": true,
+    "target": "es2022",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "moduleDetection": "force",
     "strict": true,
-    "strictNullChecks": true
+    "noUncheckedIndexedAccess": true,
+    "moduleResolution": "Bundler",
```
arlyon, line 13: This line is failing CI

> web:build: FatalError: ../../packages/tsconfig/nextjs.json(15,25): error TS6046: Argument for '--moduleResolution' option must be: 'node', 'classic', 'node16', 'nodenext'.

I assume our TS version needs updating (and maybe `bundler` needs to be lowercase?)

arlyon (changes requested):
Thanks for this. Always good to keep examples up to date with common practice. Happy to merge once CI passes.

MaxAst:
Thanks a lot for this Matt! I started setting up turborepo in an existing project yesterday and was surprised to see that their example tsconfigs aren't following your cheat sheet recommendations yet. I've got some questions, if you don't mind me asking:

1. Why aren't you adding `"verbatimModuleSyntax": true` in the base.json?
2. I see a few fields in next.json that also already exist in base.json, e.g. `"moduleResolution": "Bundler"` and `"module": "ESNext"`. What's the reason behind that, given that we already extend next.json from base.json?
3. Don't we need `"composite": true` and `"declarationMap": true` in react-library.json, given that it's for the internal UI package? or is this not necessary given that the UI package doesn't have a build step and has `"main"` and `"types"` set to `"./index.tsx" in its package.json?

examples/basic/packages/tsconfig/base.json
mattpocock: TS version needs updating - where is that handled?

arlyon: This PR should do what you need https://github.com/vercel/turbo/pull/6338

mattpocock: Perfect

mattpocock:
@MaxAst

1. Because Next.js doesn't use "type": "module" in its package.json (maybe it could? unsure), you'd get erroneous errors everywhere because TypeScript would think it was in a .cjs environment.

I'm debating removing it from my default recommendations for that reason.

2. My bad! Didn't notice it was extending. Fixed.

3. No, because "composite" is only used for when you're using project references. declarationMaps are only emitted by TypeScript - it doesn't make sense to have them turned on if you're not using `tsc` to transpile your library code.

We're using the ['internal packages'](https://turbo.build/repo/docs/handbook/sharing-code/internal-packages) pattern, meaning we don't bundle here.

MaxAst:
@mattpocock got it, thanks a lot for taking the time to explain:)

mehulkar (commented):
Onboard with this, but need to check with @tknickman about getting TS@5 PR merged first.

examples/basic/packages/tsconfig/base.json
```diff
@@ -2,20 +2,18 @@
...
     "esModuleInterop": true,
-    "forceConsistentCasingInFileNames": true,
-    "inlineSources": false,
-    "isolatedModules": true,
-    "moduleResolution": "node",
-    "noUnusedLocals": false,
-    "noUnusedParameters": false,
-    "preserveWatchOutput": true,
     "skipLibCheck": true,
+    "target": "es2022",
```
IGassmann, line 7: @mattpocock is this needed? Doesn't `"lib": ["es2022", "dom", "dom.iterable"]` already implies `"target": "es2022"`?

mattpocock: Confusingly, the TS team recommends you keep "target" in place. I still haven't quite grokked why - but I think TS always expects there to be a "target" field.

anthonyshew (approved):
Thanks, Matt!

arlyon (approved):

PR #6398, opened 2023-11-08
port: Fixing and refactoring integration tests

NicholasLYang:
### Description

This PR does a couple things:
- Standardizes test names to be `skeleton-case` and not `snake_case`
- Fixes the tests by using regexes for whitespace

### Testing Instructions

Closes TURBO-1608


mehulkar (commented):
This is great, but I'd really like it to be split up. The file name changes and the `\s*` regex could probably be together since it's not changing behavior, but the actual output changes should be separate.

NicholasLYang:
Yeah, sorry this PR got a little out of hand. Is it blocking though?

mehulkar:
It's going to be a pain to merge this into #6279 so it would be pretty helpful to break it up, but I'll defer to you. I don't feel strongly enough to block other than "hey good git hygiene will be helpful in the long run"

NicholasLYang:
Okay gotcha. I'm mostly asking because this feels like a PR that is unlikely to be reverted, but I'll split up the changes to make it a little easier on the merging side

turborepo-tests/integration/tests/dry-run.t
```diff
@@ -0,0 +1,75 @@
+Setup
```
chris-olszewski, line 1: Were there changes to this file or was it just a rename and GitHub is failing to recognize this?

turborepo-tests/integration/tests/find-turbo/hard-mode.t
```diff
@@ -21,17 +21,39 @@ It finds repo root and uses correct version
 It respects cwd
   $ ${TESTDIR}/set_version.sh $TESTROOT "1.8.0"
   $ cd $TESTROOT
+<<<<<<< HEAD
```
chris-olszewski, line 24: Need to fixup the merge conflicts before this gets merged

chris-olszewski (changes requested):

turborepo-tests/integration/tests/dry-run.t
NicholasLYang: Just the trailing whitespace regex changes

chris-olszewski (approved):
Good to merge once integration tests pass. I think you just need to update the paths in the failing tests to use kebab case instead of snake

mehulkar (approved):
Thank you!

PR #6413, opened 2023-11-09
fix(turbopack-ecmascript): support urlrewritebehavior::full

kwonoj:
### Description

This PR tries to correct the behavior of `UrlRewriteBehavior::Full`.

With `UrlRewriteBehavior::Full`, it is expected `new URL(url, base)` resolves to the absolute full path to the requested url. Previously, this didn't work for 2 reasons

- replacing `url` to the emitted output path
- replacing `base` to non-relative to the above output path

in result resolving won't give the correct path.

I tried a couple of approaches if it's possible to construct full path directly when there's a referenced asset, but wasn't able to make it work. Instead this PR relies on the new runtime fn wraps `__turbopack_require__`, calculate the full path with the absolute path runtime already have + resolved asset path. Since emitted asset contains its output with assetPrefix, exports it into the runtime as well to strips it out.

Overall behaviors across different condition are illustarted here:

https://github.com/vercel/turbo/compare/url-full-resolve-context?expand=1#diff-bb24150ae91f0fa7b3001d17e597f89d3e16f5e0f4894acb4bd0499fbcd73793R134

Closes PACK-1924


crates/turbopack-ecmascript-runtime/js/src/build/runtime.ts
```diff
@@ -91,6 +95,25 @@ var relativeURL = function(this: any, inputUrl: string) {
...
+ */
+function createResolvePathFromModule(resolver: (moduleId: string) => Exports): (moduleId: string) => string {
+  return function resolvePathFromModule(moduleId: string): string {
+    const exported = resolver(moduleId);
+    const exportedPath = exported?.default ?? exported;
+    if (typeof exportedPath !== 'string') {
+      return exported as any;
+    }
+
+    const strippedAssetPrefix = exportedPath.replace(ASSET_PREFIX, '');
```
jridgewell, line 110: The `ASSET_PREFIX` starts at the beginning, right? Can this just be a `slice`?

crates/turbopack-ecmascript-runtime/js/src/build/runtime.ts
```diff
@@ -91,6 +95,25 @@ var relativeURL = function(this: any, inputUrl: string) {
 
 relativeURL.prototype = URL.prototype;
 
+/**
+ * Returns an absolute path to the given module's id.
+ *
```
jridgewell, line 100:
```suggestion
```

crates/turbopack-ecmascript/src/references/esm/url.rs
```diff
@@ -124,21 +124,42 @@ impl ChunkableModuleReference for UrlAssetReference {
...
+    /// and the current context (rendering), lastly the asset's condition
+    /// (if it's referenced / external). The following table shows the
+    /// behavior:
+    /*
+     * original call: `new URL(url, base);`
+    ┌───────────────────────────────┬─────────────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────┬───────────────────────┐
+    │  UrlRewriteBehavior\RefAsset  │                         ReferencedAsset::Some()                         │           ReferencedAsset::External            │ ReferencedAsset::None │
+    ├───────────────────────────────┼─────────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────┼───────────────────────┤
+    │ Relative                      │ new URL(__turbopack_relative_url__(__turbopack_require__(urlId)), base) │ new URL(__turbopack_relative_url__(url), base) │ new URL(url, base)    │
+    │ Full(RenderingClient::Client) │ new URL(__turbopack_require__(urlId), "location.origin")                │ new URL(url, "location.origin")                │ new URL(url, base)    │
```
jridgewell, line 138:
```suggestion
    │ Full(RenderingClient::Client) │ new URL(__turbopack_require__(urlId), "location.origin")                │ new URL(url, location.origin)                  │ new URL(url, base)    │
```

crates/turbopack-ecmascript/src/references/esm/url.rs
```diff
@@ -124,21 +124,42 @@ impl ChunkableModuleReference for UrlAssetReference {
...
+    │ Full(RenderingClient::..)     │ new URL(__turbopack_resolve_module_id_path__(urlId))                    │ new URL(url, base)                             │ new URL(url, base)    │
```
jridgewell, line 139: Why do we need another module wrapper function for this? Why can't we derive this from `CWD` or similar?

crates/turbopack-ecmascript/src/references/esm/url.rs
```diff
@@ -124,21 +124,42 @@ impl ChunkableModuleReference for UrlAssetReference {
 
 #[turbo_tasks::value_impl]
 impl CodeGenerateable for UrlAssetReference {
+    /// rewrites calls to the `new URL()` ctor depends on the current
+    /// conditions. This is based on the configuration (UrlRewriteBehavior),
+    /// and the current context (rendering), lastly the asset's condition
+    /// (if it's referenced / external). The following table shows the
+    /// behavior:
+    /*
+     * original call: `new URL(url, base);`
```
jridgewell, line 133: Can you clarify if this should be pointing to the output path or the source path for each of these cases?

crates/turbopack-ecmascript/src/references/esm/url.rs
kwonoj: We have cwd, but that's not composable with relative path url to its output asset.

The simpliest example is like `new URL('../../public/img.png', import.meta.url)`: we have correct import.meta.url points to yhe cwd of the module, but __turbopack_require_'d module points to the output distdir's asset with asset prefix so can't relate those two directlt.

This is the actual problem I struggled to solve without these wrapper, if there's a way to get a full phsyical path without asset prefix it'd be great. I feel I may miss something.

crates/turbopack-ecmascript/src/references/esm/url.rs
kwonoj: Output path. Webpack points to the output path as well.

crates/turbopack-ecmascript-runtime/js/src/build/runtime.ts
kwonoj: Yes, I think so.

crates/turbopack-ecmascript/src/references/esm/url.rs
```diff
@@ -124,21 +124,43 @@ impl ChunkableModuleReference for UrlAssetReference {
...
+    /// as similar to the webpack's behavior. This is based on the
+    /// configuration (UrlRewriteBehavior), and the current context
+    /// (rendering), lastly the asset's condition (if it's referenced /
+    /// external). The following table shows the behavior:
+    /*
+     * original call: `new URL(url, base);`
+    ┌───────────────────────────────┬─────────────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────┬───────────────────────┐
+    │  UrlRewriteBehavior\RefAsset  │                         ReferencedAsset::Some()                         │           ReferencedAsset::External            │ ReferencedAsset::None │
+    ├───────────────────────────────┼─────────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────┼───────────────────────┤
+    │ Relative                      │ new URL(__turbopack_relative_url__(__turbopack_require__(urlId)), base) │ new URL(__turbopack_relative_url__(url), base) │ new URL(url, base)    │
```
sokra, line 138:
```suggestion
    │ Relative                      │ new __turbopack_relative_url__(__turbopack_require__(urlId)) │ new __turbopack_relative_url__(url) │ new URL(url, base)    │
```

rustc-ice-2023-11-13T17:05:44.397158Z-94125.txt
sokra, line 1: huh?

kwonoj: 🤦 sorry, somehow missed it

sokra (changes requested):
you can add an execution turbopack-test to test this

kwonoj:
@sokra 

> you can add an execution turbopack-test to test this

Gave a quick peek, and then realized test uses different runtime to the runtime next-swc's turbopack uses, which doesn't have any foundation support such as runtime's absolute path, etcs. I'd like to tackle this as separate PR later as have to port runtime's feature other than my changes.

kwonoj:
> Fine to me, at least run it against the next.js test case to verify it's working

Yes, as shared in the slack thread https://github.com/vercel/next.js/pull/58404 verifies new url tests are passing (https://github.com/vercel/next.js/pull/58404/commits/4e63956b9cae34d3ae67c56b708c99a27351f106#diff-2731c6687251176e46f1aa219fe2c059c9a13b61f0a5035d9b5073c9fc10258e) . Failed test is flaky I believe.

arlyon (approved):

PR #6416, opened 2023-11-09
fix: Dry run output

NicholasLYang:
### Description

We had some spelling mistakes and inconsistent formatting in our dry run text output. Also refactored the dry run opts to be a single enum instead of two booleans

### Testing Instructions

Closes TURBO-1629


crates/turborepo-lib/src/run/mod.rs
```diff
@@ -170,7 +170,8 @@ impl<'a> Run<'a> {
             opts.cache_opts.skip_remote = !enabled;
         }
 
-        let _is_structured_output = opts.run_opts.graph.is_some() || opts.run_opts.dry_run_json;
+        let _is_structured_output = opts.run_opts.graph.is_some()
+            || matches!(opts.run_opts.dry_run, Some(DryRunMode::Json));
```
chris-olszewski, line 174: Not related to this PR, but what is this used for?

chris-olszewski (approved):

crates/turborepo-lib/src/run/mod.rs
NicholasLYang: I have no clue honestly

PR #6417, opened 2023-11-09
Port(Turborepo): Synthesize command for spaces

gsoltis:
### Description

 - Synthesize a `turbo run` command for sending to Spaces
 - Some light fixing up of rendering a legacy filter to a string

### Testing Instructions

 - ported `run_spec_test.go`

Fixes TURBO-1565

### DIFF
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -390,162 +390,153 @@ mod test {
             expected.iter().map(|s| s.to_string()).collect::<Vec<_>>()
         )
     }
+    #[derive(Default)]
+    struct TestCaseOpts {
+        filter_patterns: Vec<String>,
+        tasks: Vec<String>,
+        only: bool,
+        pass_through_args: Vec<String>,
+        parallel: bool,
+        continue_on_error: bool,
+        dry_run: bool,
+        dry_run_json: bool,
+        legacy_filter: Option<LegacyFilter>,
+    }
 
-    #[test]
-    fn test_synthesize_command() {
-        #[derive(Default)]
-        struct TestCaseOpts {
-            filter_patterns: Vec<String>,
-            tasks: Vec<String>,
-            only: bool,
-            pass_through_args: Vec<String>,
-            parallel: bool,
-            continue_on_error: bool,
-            dry_run: bool,
-            dry_run_json: bool,
-            legacy_filter: Option<LegacyFilter>,
-        }
-
-        let tests = vec![
-            (
-                TestCaseOpts {
-                    filter_patterns: vec!["my-app".to_string()],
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/opts.rs
```diff
@@ -37,6 +38,54 @@ pub struct Opts<'a> {
     pub scope_opts: ScopeOpts,
 }
 
+impl<'a> Opts<'a> {
+    pub fn synthesize_command(&self) -> String {
+        let mut cmd = format!("turbo run {} ", self.run_opts.tasks.join(" "));
+        let mut flags: Vec<String> = vec![];
```
NicholasLYang, line 44: I'd consider pushing directly to the string versus creating a vector. It'll be a little more simple and won't create a bunch of unnecessary allocations

gsoltis: I did try that initially, but it doesn't turn out simpler. There are a few cases where we need to check for existence of any items before doing a join so that the spacing works out properly. Using an `.extend` call lets us skip all those checks, at the cost of allocating. I also wouldn't be concerned with allocation here, it's most likely < 20 short strings.

NicholasLYang: I'm confused, where do we need to check for existence for spacing? My implementation preemptively puts a space before any flag, so it adds a single space only if it's needed.

gsoltis: Oh, I see. Yeah, you can do it if you don't use `join`. I can switch it over.

crates/turborepo-lib/src/opts.rs
```diff
@@ -37,6 +38,54 @@ pub struct Opts<'a> {
...
+                .as_filter_pattern()
+                .iter()
+                .map(|pattern| format!("--filter={pattern}")),
+        );
+        //cmd += &self.scope_opts.filter_patterns.iter().map(|pattern|
+        // format!("--filter={pattern}")).join(" "); cmd += &self.scope_opts.
+        // legacy_filter.as_filter_pattern().iter().map(|pattern|
+        // format!("--filter={pattern}")).join(" ");
+        if self.run_opts.parallel {
+            //cmd += " --parallel";
```
chris-olszewski, line 63:
```suggestion
```

crates/turborepo-lib/src/opts.rs
```diff
@@ -37,6 +38,54 @@ pub struct Opts<'a> {
...
+            flags.push("--parallel".to_string());
+        }
+        if self.run_opts.continue_on_error {
+            //cmd += " --continue";
```
chris-olszewski, line 67:
```suggestion
```

chris-olszewski (approved):

PR #6419, opened 2023-11-09
feat: send task logs to spaces

chris-olszewski:
### Description

Start sending task summaries to spaces. As expected this ended up touching a few parts of the codebase and required some reworking of them.

I highly recommend reviewing each commit in the PR individually as I attempted to keep each one fairly small to make this easier to review.

### Testing Instructions

Observe the runs sent via the Rust codepath from my machine (you can tell it's Rust based on the `command` 😉 )
 - [Cache miss](https://vercel.com/teams/vercel/repos/turbo-monorepo/runs/space_run_59Yj2WWotw4oFnHMAEKWb5JZ)
 - [Cache hit](https://vercel.com/teams/vercel/repos/turbo-monorepo/runs/space_run_kO8iZcJPGkXoQ127qGGcOQXz)

Closes TURBO-1632


crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -577,22 +601,52 @@ impl ExecContext {
...
                 callback
                     .send(match self.continue_on_error {
                         true => Ok(()),
                         false => Err(StopExecution),
                     })
                     .ok();
-                if !self.continue_on_error {
-                    self.manager.stop().await;
+
+                match (spaces_client, self.continue_on_error) {
```
chris-olszewski, line 630: There is most certainly a better way to do this, but everything I tried ran into lifetime issues. I'll continue thinking about it, but I didn't want to block this PR on being clever.

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -248,41 +250,24 @@ enum Event {
...
 
 #[derive(Debug, Serialize, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct TaskExecutionSummary {
-    start_time: i64,
-    end_time: i64,
-    #[serde(skip)]
-    state: ExecutionState,
-    exit_code: Option<i32>,
+    pub start_time: i64,
```
chris-olszewski, line 260: Making this a straight data class

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -248,41 +250,24 @@ enum Event {
...
+    pub exit_code: Option<i32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub error: Option<String>,
 }
 
 impl TaskExecutionSummary {
     pub fn is_failure(&self) -> bool {
-        matches!(self.state, ExecutionState::BuildFailed { .. })
+        // TODO: how do we treat a "none"
+        !matches!(self.exit_code, Some(0))
```
chris-olszewski, line 271: This changes better matches the Go behavior as spawn failures/external killings are considered failures, not just non-zero exit codes.

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -375,114 +368,96 @@ impl TaskTracker<chrono::DateTime<Local>> {
...
+                event: Event::Built,
+                state: Some(state),
+            })
             .await
             .expect("summary state thread finished");
+        execution
     }
 
-    pub async fn spawn_failed(self, error: impl fmt::Display) {
+    pub async fn build_failed(
```
chris-olszewski, line 431: Move ourselves closer to Go behavior where spawn failures and non-zero exit codes were treated the same.

crates/turborepo-lib/src/run/summary/spaces.rs
```diff
@@ -257,8 +360,14 @@ mod tests {
             "rauchg".to_string(),
         ))?;
 
+        let mut join_set = tokio::task::JoinSet::new();
         for task_summary in tasks {
-            spaces_client_handle.finish_task(task_summary).await?;
+            let task_client = spaces_client_handle.task_client();
+            join_set.spawn(async move { task_client.send_task(task_summary).await });
```
chris-olszewski, line 366: Slightly more realistic test where we finish tasks concurrently instead of serially.

crates/turborepo-api-client/src/spaces.rs
```diff
@@ -20,14 +20,31 @@ pub struct SpaceClientSummary {
...
-    pub time_saved: u32,
+    pub status: CacheStatus,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub source: Option<CacheSource>,
+    pub time_saved: u64,
+}
+
+#[derive(Debug, Serialize, Copy, Clone)]
+#[serde(rename_all = "UPPERCASE")]
+pub enum CacheStatus {
```
chris-olszewski, line 34: These definitions are copy-pasta'd from `summary/task.rs` just because we don't necessarily need the JSON that gets sent to spaces and the output of dry run to match. If we want to commit to keeping these in lock step, I'm more than happy to do that and have `summary/task.rs` import the definitions from here.

crates/turborepo-lib/src/run/summary/spaces.rs
```diff
@@ -40,6 +47,23 @@ pub struct SpacesClientHandle {
     tx: Sender<SpaceRequest>,
 }
 
+/// A spaces client with functionality limited to sending task information
+/// This client should only live while processing a task
+pub struct SpacesTaskClient {
```
NicholasLYang, line 52: nit: I'm not convinced this needs to be a separate type versus a type alias

crates/turborepo-lib/src/run/summary/spaces.rs
```diff
@@ -211,6 +250,70 @@ impl SpacesClient {
...
+        let cache = cache_status.map_or_else(
+            SpacesCacheStatus::default,
+            |CacheHitMetadata { source, time_saved }| SpacesCacheStatus {
+                status: turborepo_api_client::spaces::CacheStatus::Hit,
+                source: Some(match source {
+                    turborepo_cache::CacheSource::Local => {
+                        turborepo_api_client::spaces::CacheSource::Local
+                    }
+                    turborepo_cache::CacheSource::Remote => {
+                        turborepo_api_client::spaces::CacheSource::Local
```
NicholasLYang, line 292: Is this supposed to be `Remote`?

crates/turborepo-lib/src/task_graph/visitor.rs
NicholasLYang: I don't see anything wrong with this? What's the issue?

crates/turborepo-lib/src/run/summary/spaces.rs
```diff
@@ -257,8 +272,14 @@ mod tests {
             "rauchg".to_string(),
         ))?;
 
+        let mut join_set = tokio::task::JoinSet::new();
         for task_summary in tasks {
-            spaces_client_handle.finish_task(task_summary).await?;
+            let task_client = spaces_client_handle.task_client();
+            join_set.spawn(async move { task_client.finish_task(task_summary).await });
```
NicholasLYang: I'm confused, is there a reason we're not passing `task_client.finish_task(task_summary)` directly?

NicholasLYang (changes requested):
A few questions but overall looks very solid

crates/turborepo-lib/src/run/summary/spaces.rs
chris-olszewski: I'm not going to die on this hill, but having this as it's own type allows callers to not worry about the preamble of converting task information into a summary and handling error coercion. Also it doesn't leak implementation details of how one sends a task.

crates/turborepo-lib/src/task_graph/visitor.rs
chris-olszewski: I just don't love explicitly handling ever case separately just to allow for us to concurrently run the two futures if there are two to run.

crates/turborepo-api-client/src/spaces.rs
```diff
@@ -187,3 +205,57 @@ impl APIClient {
         Ok(())
     }
 }
+
+impl Default for SpacesCacheStatus {
```
NicholasLYang, line 209: What's the reason for manually implementing this versus deriving?

crates/turborepo-lib/src/run/summary/spaces.rs
NicholasLYang: nit: why not pass the future directly?
```suggestion
            let task_client = spaces_client_handle.task_client();
            join_set.spawn(task_client.send_task(task_summary));
```

NicholasLYang (approved):
LGTM. Just a couple nits

crates/turborepo-lib/src/run/summary/spaces.rs
chris-olszewski: The future needs ownership of the client which AFAIK `async move` is the best way to ensure the resulting future has ownership. I could do something with `move` closure, but at that point it's less readable IMO than an `async move` block.

PR #6422, opened 2023-11-10
port(Turborepo): Pass through args

gsoltis:
### Description

 - ports pass-through args logic

### Testing Instructions

e2e tests for passthrough args now pass with Rust

Closes TURBO-1625

### DIFF
--- a/crates/turborepo-lib/src/task_graph/visitor.rs
+++ b/crates/turborepo-lib/src/task_graph/visitor.rs
@@ -223,7 +223,6 @@ impl<'a> Visitor<'a> {
             let Some(_command) = command else { continue };
 
             let workspace_directory = self.repo_root.resolve(workspace_info.package_path());
-            let pass_through_args = self.opts.run_opts.args_for_task(&info);
 
             let mut exec_context = factory.exec_context(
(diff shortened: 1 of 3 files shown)


crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -524,6 +527,7 @@ impl<'a> ExecContextFactory<'a> {
             task_hash,
             execution_env,
             continue_on_error: self.visitor.opts.run_opts.continue_on_error,
+            pass_through_args,
```
chris-olszewski, line 530: Nit we can avoid passing this into the constructor and use the factory to build it instead:
```suggestion
            pass_through_args: self.visitor.opts.run_opts.args_for_task(&task_id),
```
(This might need to be hoisted to happen before `task_id` gets moved in another part of creating the `ExecContext`)

chris-olszewski (approved):

crates/turborepo-lib/src/task_graph/visitor.rs
gsoltis: Good call, moved to factory.

PR #6424, opened 2023-11-10
fix encoding of chunks, fix chunk (re)loading for css

sokra:
### Description

We need to be consistent with url encoding
Look for existing link tags when loading CSS
allow querystring when detection existing chunks

### Testing Instructions

Closes PACK-1934


ForsakenHarmony (approved):

abergenw:
This makes sense, but when doing the replacement it seems like we're now looking for url encoded link tags which are not encoded in the first place?

ForsakenHarmony:
@abergenw you are correct, quick fix is in #6777

PR #6439, opened 2023-11-14
Re-up create-turbo.

anthonyshew:
Anthony's To-Do's:
- [x] Make sure all dependencies are up to date
- [x] Find anything that may be broken in-editor
- [x] Loosen ESLint configuration
- [x] Make sure it deploys for CI
- [x] Check with TypeScript team that we are following current best practices (post-ship, non-blocking)

### Description

Making sure `create-turbo` is robust and modern.

### Testing Instructions

Try it out with `npx create-turbo@latest -e https://github.com/vercel/turbo/tree/update-create-turbo/examples/basic`

Things to pay attention to:
- [x] If I add an unused variable in a workspace, do I see a warning in both my editor and when I run `turbo lint`?
- [x] Do I get proper hot-reloading in `turbo dev`?


examples/non-monorepo/tsconfig.json
```diff
@@ -7,8 +7,8 @@
     "strict": true,
     "noEmit": true,
     "esModuleInterop": true,
-    "module": "esnext",
-    "moduleResolution": "bundler",
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
```
anthonyshew, line 11: Now, I know what you're thinking. "Why is this change being included in this PR?"

The answer is: I have no idea.

Changing the TSConfig in `basic` is causing issues here in `non-monorepo`. It's nonsense. But, sometimes, you just gotta roll with the punches.

anthonyshew:
Current setup is not showing auto-imports from the `ui` package.

Notable issue from TypeScript: https://github.com/microsoft/TypeScript/issues/53116#issuecomment-1730360536

It's possible I may be reading the guidance wrong. Will continue to investigate post-ship.

examples/basic/packages/ui/.eslintrc.js
```diff
@@ -1,3 +1,10 @@
+/** @type {import("eslint").Linter.Config} */
 module.exports = {
-  extends: ["custom/react-internal"],
+  root: true,
+  extends: [require.resolve("@repo/eslint-config/react-internal.js")],
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    project: true,
+  },
+  ignorePatterns: ["turbo/**"],
```
tknickman, line 9: Not linting generators?

examples/basic/packages/ui/package.json
```diff
@@ -1,21 +1,36 @@
...
   "devDependencies": {
+    "@repo/eslint-config": "workspace:*",
+    "@repo/typescript-config": "workspace:*",
     "@turbo/gen": "^1.10.12",
     "@types/node": "^20.5.2",
+    "@types/eslint": "^8.44.7",
     "@types/react": "^18.2.0",
     "@types/react-dom": "^18.2.0",
-    "eslint-config-custom": "workspace:*",
+    "eslint": "^8.53.0",
```
tknickman, line 32: I'm split about this one. Keeping eslint at root will save you a lot of headache (like we do for prettier) because it guarantees there's only one copy, with the same version, and every workspace resolves the same one. Specifing it per workspace also works, but can become a bigger maintenance headache if things start to drift (time for syncpack?!)

examples/basic/packages/ui/package.json
```diff
@@ -1,21 +1,36 @@
...
+      "types": "./src/card.tsx",
+      "development": "./src/card.tsx",
+      "default": "./dist/card.js"
+    },
+    "./code": {
+      "types": "./src/code.tsx",
+      "development": "./src/code.tsx",
+      "default": "./dist/code.js"
+    }
+  },
```
tknickman, line 17: Can't we just do:

```
  "exports": {
    "./*": {
      "types": "./src/*.tsx",
      "development": "./src/*.tsx",
      "default": "./dist/*.js"
    }
  },
```

examples/basic/apps/web/.eslintrc.js
```diff
@@ -1,3 +1,9 @@
+/** @type {import("eslint").Linter.Config} */
 module.exports = {
-  extends: ["custom/next"],
+  root: true,
+  extends: [require.resolve("@repo/eslint-config/next.js")],
```
tknickman, line 4: `require.resolve` shouldn't be needed in these, but are you shooting for less magic?

examples/basic/packages/ui/package.json
```diff
@@ -1,21 +1,36 @@
...
   "scripts": {
-    "lint": "eslint .",
+    "build": "tsc",
```
tknickman, line 19: Now that UI has it's own build, I think we should promote workspace configs in the default example. IMO it makes things like outputs easier to grock and reason about

tknickman (changes requested):
Only real change request is the UI exports. Everything else is just questions etc. But really looks awesome!

examples/basic/packages/ui/package.json
anthonyshew: That's this. https://github.com/vercel/turbo/pull/6439#issuecomment-1813839498

TL;DR: If you wildcard export, you don't get auto-imports across workspaces.

examples/basic/apps/web/.eslintrc.js
anthonyshew: Good call.

examples/basic/packages/ui/package.json
anthonyshew: I'm in.

examples/basic/packages/ui/package.json
anthonyshew: Oops, no, this is fluff. Delete!

examples/basic/packages/ui/package.json
tknickman: Ah. such sad

tknickman (approved):

examples/basic/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+// This configuration only applies to the package manager root.
+/** @type {import("eslint").Linter.Config} */
+module.exports = {
+  ignorePatterns: ["apps/**", "packages/**"],
+  extends: ["@repo/eslint-config/library.js"],
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    project: true,
+  },
+};
```
mehulkar, line 10: Should we adopt Flat Config from Eslint? https://eslint.org/docs/latest/use/configure/configuration-files-new

examples/basic/apps/docs/next.config.js
```diff
@@ -1,4 +1,4 @@
+/** @type {import('next').NextConfig} */
 module.exports = {
   reactStrictMode: true,
-  transpilePackages: ["ui"],
```
mehulkar, line 3: Does next14 no longer need this?

examples/basic/meta.json
```diff
@@ -1,7 +0,0 @@
-{
-  "name": "Next.js",
-  "description": "Minimal Turborepo example for learning the fundamentals.",
-  "template": "https://vercel.com/templates/next.js/turborepo-next-basic",
-  "featured": true,
-  "boost": true
-}
```
mehulkar, line 7: I think we still need this?

examples/basic/package.json
```diff
@@ -1,16 +1,18 @@
...
+  "name": "my-turborepo",
   "private": true,
   "scripts": {
-    "build": "turbo run build",
-    "dev": "turbo run dev",
-    "lint": "turbo run lint",
+    "build": "turbo build",
+    "dev": "turbo dev",
+    "lint": "turbo lint",
+    "typecheck": "turbo typecheck",
```
mehulkar, line 8: We should talk about this internally. I _think_ if you use `turbo` like this here (without `npx`/`pnpx`), it refers to the global turbo binary, is that right? If that's so, I feel like we should just delete these scripts in the root package.json and recommend using `turbo` global directly? It would prevent a bunch of issues with people putting `turbo` in their package.json scripts in packages, and accidentally doing infinite recursion.

examples/basic/packages/ui/index.tsx
```diff
@@ -1,2 +0,0 @@
-// component exports
-export { Card } from "./card";
```
mehulkar, line 2: no more barrel files!

examples/basic/packages/ui/src/code.tsx
mehulkar, line 1: Curious why you moved this to the library? Just cause or does it demonstrate something different than `Card`?

examples/basic/turbo.json
```diff
@@ -6,10 +6,15 @@
...
-    "lint": {},
+    "lint": {
+      "dependsOn": ["^lint"]
+    },
     "dev": {
       "cache": false,
       "persistent": true
+    },
+    "typecheck": {
+      "dependsOn": ["^build"]
```
mehulkar, line 17: Why does typechecking require building dependencies first?

examples/basic/meta.json
tknickman: Good catch! @anthonyshew we do still need this. This file is read is how we build the list of examples in the docs at build time.

examples/basic/apps/docs/next.config.js
tknickman: We're no longer using next transpile for the basic example, it has it's own build step instead.

examples/basic/meta.json
tknickman: More details on how this works:

https://github.com/vercel/turbo/blob/main/docs/pages/repo/docs/getting-started/from-example.mdx#L10-L41

examples/basic/.eslintrc.js
anthonyshew: I'm going to look into this as a follow-up!

examples/basic/meta.json
anthonyshew: Not sure when I did that. :shocked:

Thanks for the catch! @mehulkar to the rescue as usual.

examples/basic/turbo.json
anthonyshew: It did at one point but doesn't anymore. Good eye.

I'm going to actually remove the typecheck script now because it will require `"^topo"` and I'd prefer an advanced technique like that not be in what will be the entrypoint to Turborepo/monorepos for many folks.

examples/basic/packages/ui/src/code.tsx
anthonyshew: I wanted to create multiple entrypoints for the `ui` package. It's a common question and, additionally, shows off how to get to multiple exports without using a barrel.

examples/basic/packages/ui/index.tsx
anthonyshew: Level up!

examples/basic/package.json
anthonyshew: My thought here is going to be that `create-turbo` is the entrypoint to Turborepo/monorepos for the majority of folks. Feels a little "cart before the horse" to expect someone who just heard about `turbo` 15 seconds ago from a tweet, video, etc. that was inspired enough to hit `npx create-turbo` to have global `turbo` installed. Happy to explore the squad, though, if you think it's warranted.

> I think if you use turbo like this here (without npx/pnpx), it refers to the global turbo binary, is that right?

I've been under the impression that it uses the version installed in the repo while I've been in userland. :eyes:

anthonyshew: My thought here is going to be that `create-turbo` is the entrypoint to Turborepo/monorepos for the majority of folks. Feels a little "cart before the horse" to expect someone who just heard about `turbo` 15 seconds ago from a tweet, video, etc. that was inspired enough to hit `npx create-turbo` to have global `turbo` installed. Happy to explore the squad, though, if you think it's warranted.

> I think if you use turbo like this here (without npx/pnpx), it refers to the global turbo binary, is that right?

I've been under the impression that it uses the version installed in the repo while I've been in userland. :eyes:

PR #6442, opened 2023-11-14
Update tsup.config.ts related to problems with typescript

sbeaujard:
inside the package.json of /ui, it use "types": "./dist/index.d.ts", 
But with the current configuration of tsup, i won't generate the index.d.ts file but a index.d.mts file.
Then the global build failed.


anthonyshew:
This should be fixed today! Sorry for not seeing this PR at the time. I need to be more diligent about checking for PRs before I go and fix things myself. Sincerest apologies!

PR #6445, opened 2023-11-14
Ensure daemon connection errors are non-fatal

gsoltis:
### Description

Makes failure to start or connect to the daemon non-fatal. It's purely a performance optimization and we can still execute without it.

### Testing Instructions

Verified manually with a 1ms timeout

Closes TURBO-1655

### DIFF
--- a/crates/turborepo-lib/src/run/mod.rs
+++ b/crates/turborepo-lib/src/run/mod.rs
@@ -18,7 +18,7 @@ pub use cache::{RunCache, TaskCache};
 use chrono::{DateTime, Local};
 use itertools::Itertools;
 use rayon::iter::ParallelBridge;
-use tracing::{debug, info, warn};
+use tracing::{debug, info};
 use turbopath::AbsoluteSystemPathBuf;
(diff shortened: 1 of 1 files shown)


NicholasLYang (approved):

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -215,10 +215,20 @@ impl<'a> Run<'a> {
...
-            debug!("running in daemon mode");
-            daemon = Some(client);
-        }
+            match connector.connect().await {
+                Ok(client) => {
+                    debug!("running in daemon mode");
+                    Some(client)
+                }
+                Err(e) => {
+                    warn!("failed to connect to daemon {e}");
```
chris-olszewski, line 224: Not blocking, but do we want this to be displayed to users at the default log level?

chris-olszewski (approved):

crates/turborepo-lib/src/run/mod.rs
gsoltis: Good point, I'll drop this to debug

PR #6446, opened 2023-11-14
fix: displaying stdout/stderr with errors-only

chris-olszewski:
### Description

This PR fixes up our `errors-only` behavior by making sure we flush the file writer by the time we attempt to replay the logs.

### Testing Instructions

`run-logging/errors-only.t` and `run/continue.t` now both pass

Closes TURBO-1663

### DIFF
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -149,9 +149,9 @@ impl TaskCache {
 
         log_writer.with_log_file(&self.log_file_path)?;
 
-        if matches!(
+        if !matches!(
             self.task_output_mode,
-            OutputLogsMode::Full | OutputLogsMode::NewOnly
+            OutputLogsMode::None | OutputLogsMode::HashOnly | OutputLogsMode::ErrorsOnly
         ) {
             log_writer.with_prefixed_writer(prefixed_writer);
         }
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -212,10 +212,15 @@ impl TaskCache {
                 .await?;
 
             let Some((cache_hit_metadata, restored_files)) = cache_status else {
-                prefixed_ui.output(format!(
-                    "cache miss, executing {}",
-                    color!(self.ui, GREY, "{}", self.hash)
-                ));
+                if !matches!(
+                    self.task_output_mode,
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/run/cache.rs
```diff
@@ -149,9 +149,9 @@ impl TaskCache {
 
         log_writer.with_log_file(&self.log_file_path)?;
 
-        if matches!(
+        if !matches!(
             self.task_output_mode,
-            OutputLogsMode::Full | OutputLogsMode::NewOnly
+            OutputLogsMode::None | OutputLogsMode::HashOnly | OutputLogsMode::ErrorsOnly
```
chris-olszewski, line 154: This isn't a behavior change, but it matches the Go code now.

crates/turborepo-lib/src/run/cache.rs
```diff
@@ -212,10 +212,15 @@ impl TaskCache {
...
 
             let Some((cache_hit_metadata, restored_files)) = cache_status else {
-                prefixed_ui.output(format!(
-                    "cache miss, executing {}",
-                    color!(self.ui, GREY, "{}", self.hash)
-                ));
+                if !matches!(
+                    self.task_output_mode,
+                    OutputLogsMode::None | OutputLogsMode::ErrorsOnly
+                ) {
```
chris-olszewski, line 218: Again, this now matches the Go code and we avoid printing `cache miss` twice on `--errors-only`

turborepo-tests/integration/tests/run-logging/errors-only.t
```diff
@@ -48,8 +48,8 @@ Setup
   app-a:builderror: npm ERR! Error: command failed 
   app-a:builderror: npm ERR!   in workspace: app-a 
   app-a:builderror: npm ERR!   at location: .* (re)
-  app-a:builderror: ERROR: command finished with error: command .* npm run builderror exited \(1\) (re)
```
mehulkar, line 51: curious what's up with this change?

mehulkar (approved):

turborepo-tests/integration/tests/run-logging/errors-only.t
chris-olszewski: Due to how Rust finds binaries to run we need to first run `which` on package manager binaries in order to find them on Windows. Outcome of this is that our process warnings have the entire path to the binary instead of just the binary name.

PR #6449, opened 2023-11-14
chore: enable rust codepath on CI

chris-olszewski:
### Description

This PR switches to use the Rust codepath for turborepo CI.

Part of this PR switches our Windows usage to use GitBash instead of Powershell for invoking `turbo`. This isn't ideal, but it allows us to start dogfooding. 

### Testing Instructions

Our existing CI should pass. We should eagerly revert if an issue is discovered. 

Closes TURBO-1665


gsoltis (approved):
LFG

PR #6451, opened 2023-11-14
chore: add test for turbo bin

mehulkar:
Adds a quick test for `turbo bin` extracted from https://github.com/vercel/turbo/pull/6279

Closes TURBO-1668

### DIFF
--- a/turborepo-tests/integration/tests/command-bin.t
+++ b/turborepo-tests/integration/tests/command-bin.t
@@ -1,0 +1,9 @@
+Setup
+  $ . ${TESTDIR}/../../helpers/setup.sh
+  $ . ${TESTDIR}/_helpers/setup_monorepo.sh $(pwd)
+
+  $ ${TURBO} bin -vvv > out.log
+  $ grep --quiet -E "Global turbo version: .*" out.log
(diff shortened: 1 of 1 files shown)


turborepo-tests/integration/tests/command-bin.t
```diff
@@ -0,0 +1,9 @@
+Setup
+  $ . ${TESTDIR}/../../helpers/setup.sh
+  $ . ${TESTDIR}/_helpers/setup_monorepo.sh $(pwd)
+
+  $ ${TURBO} bin -vvv > out.log
+  $ grep --quiet -E "Global turbo version: .*" out.log
+  $ grep --quiet -E "No local turbo binary found at" out.log
+  $ grep --quiet -E "Running command as global turbo" out.log
```
chris-olszewski, line 8: Out of curiosity: is extended regex required for these patterns?

chris-olszewski (approved):

turborepo-tests/integration/tests/command-bin.t
mehulkar: no not in these cases, i've removed here. I think I was multi line editing these files and just used the same options

PR #6452, opened 2023-11-14
feat(ecmascript): support dynamic tenary operator jsvalue

kwonoj:
### Description

context: https://vercel.slack.com/archives/C03EWR7LGEN/p1699990767027949

Currently, if the input source have a non-statically analyzable tenary operator

```
await import(
              process.env.NEXT_RUNTIME === 'edge'
                ? 'next/dist/compiled/@vercel/og/index.edge.js'
                : 'next/dist/compiled/@vercel/og/index.node.js'
)
```

it becomes to the output like

```
(await __turbopack_require__("[project]/packages/next/dist/compiled/@vercel/og/index.edge.js [rsc] (ecmascript, loader)")(__turbopack_import__))
```

that takes first value of tenary regardless of the condition. In the analyze phase if tenary is dynamic it retuns `JsValue::Alternatives`, and it takes the first available value immediately.

PR changes its behavior by having new type of JsValue for tenary - if condExpr is not statically analyzable. I tried to use existing JsValue but that doesn't seem to work easily out of the box. However I may miss some obvious so PR might be refactored more simple way instead.

Closes PACK-1963


crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -465,6 +465,17 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                 true
             }
         }
+        JsValue::Tenary(_, test, cons, alt) => {
+            if test.is_truthy() == Some(true) {
+                *value = take(cons);
+                true
+            } else if test.is_falsy() == Some(true) {
```
jridgewell, line 472: This is just `test.is_truty() == Some(false)`, right? We can this into a branch above and save some cycles.

jridgewell (approved):

crates/turbopack-ecmascript/src/analyzer/builtin.rs
kwonoj: I think so, this is borrowing existing references from elsewhere (I recall it's binary operation or something)

sokra: Yeah this can be a match on is_truty

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -1224,6 +1246,12 @@ impl JsValue {
                 op.joiner(),
                 b.explain_internal_inner(hints, indent_depth, depth, unknown_depth),
             ),
+            JsValue::Tenary(_, test, cons, alt) => format!(
+                "({}?{}:{})",
```
sokra, line 1250: A bit more spacing here would be nice

crates/turbopack-ecmascript/tests/analyzer/graph/conditional-import/input.js
```diff
@@ -1,5 +1,7 @@
 const a = import(true ? "a" : "b")
```
sokra, line 1: This is testing your tenary support and should evaluate to a

sokra: I wonder how this already works before

crates/turbopack-ecmascript/tests/analyzer/graph/conditional-import/resolved-effects.snapshot
```diff
@@ -3,10 +3,23 @@
 0 -> 2 call = import*0*("a")
```
sokra, line 3: Already seem to work?

crates/turbopack-ecmascript/tests/analyzer/graph/conditional-import/input.js
kwonoj: Those tenary passes evaluation at

https://github.com/vercel/turbo/pull/6452/commits/3c2d29cb1cf6e897995b21997374617a6551fcc4#diff-2c7634641094d97670b756458b73ab7bc5f6fba2da97471427a6c95795b8562dR429, while dynamic tenary evaluation doesn't seem to pass, so it goes to jsvalue:alternatives.

PR #6400, opened 2023-11-08
chore(tests): make test output assertions simpler

mehulkar:
less output to match and keep track of when things change

### Testing Instructions

Closes TURBO-1609

PR #6455, opened 2023-11-15
chore: simplify inference test grep commands

mehulkar:
This is a follow up from #6437.

Using --quiet means we can avoid duplicating the same string twice and rely on the exit code to indicate a match, also making the path matching simpler for Windows (https://github.com/vercel/turbo/pull/6279). The change to not use `cat` is to just follow the pattern from #6400.

### DIFF
--- a/turborepo-tests/integration/tests/inference/nested-workspaces.t
+++ b/turborepo-tests/integration/tests/inference/nested-workspaces.t
@@ -3,82 +3,58 @@ Setup
   $ . ${TESTDIR}/nested_workspaces_setup.sh $(pwd)/nested_workspaces
 
   $ cd $TARGET_DIR/outer && ${TURBO} run build --filter=nothing -vv 1> OUTER 2>&1
-  $ cat OUTER | grep --only-match -E "Repository Root: .*/nested_workspaces/outer"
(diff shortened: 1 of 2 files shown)


turborepo-tests/integration/tests/inference/nested-workspaces.t
```diff
@@ -3,82 +3,58 @@ Setup
   $ . ${TESTDIR}/nested_workspaces_setup.sh $(pwd)/nested_workspaces
 
   $ cd $TARGET_DIR/outer && ${TURBO} run build --filter=nothing -vv 1> OUTER 2>&1
-  $ cat OUTER | grep --only-match -E "Repository Root: .*/nested_workspaces/outer"
-  Repository Root: .*/nested_workspaces/outer (re)
-  $ cat OUTER | grep --only-match "No tasks were executed as part of this run."
-  No tasks were executed as part of this run.
+  $ grep --quiet -E "Repository Root: .*/nested_workspaces/outer" OUTER
+  $ cat OUTER | grep --quiet "No tasks were executed as part of this run."
```
mehulkar, line 7:
```suggestion
  $ grep --quiet "No tasks were executed as part of this run." OUTER
```

gsoltis (approved):

PR #6465, opened 2023-11-15
fix(Turborepo): Make verbosity test pass with Rust

gsoltis:
### Description

 - drop skipping daemon log message to debug
 - switch rust process manager logging to debug

### Testing Instructions

 - switch highest verbosity settings to just check that we print out `DEBUG`

Closes TURBO-1642

### DIFF
--- a/turborepo-tests/integration/tests/run-logging/verbosity.t
+++ b/turborepo-tests/integration/tests/run-logging/verbosity.t
@@ -4,7 +4,6 @@ Setup
 
 Verbosity level 1
   $ ${TURBO} build -v --filter=util --force
-  [-0-9:.TWZ+]+ \[INFO]  turbo: skipping turbod since we appear to be in a non-interactive context (re)
   \xe2\x80\xa2 Packages in scope: util (esc)
   \xe2\x80\xa2 Running build in 1 packages (esc)
   \xe2\x80\xa2 Remote caching disabled (esc)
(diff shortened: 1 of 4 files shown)


turborepo-tests/integration/tests/run-logging/verbosity.t
```diff
@@ -37,96 +35,11 @@ Verbosity level 1
...
-  util:build: building
-  [-0-9:.TWZ+]+ \[DEBUG] turbo.: caching output: outputs="{\[packages/util/.turbo/turbo-build.log] \[]}" (re)
-  [-0-9:.TWZ+]+ \[DEBUG] turbo.: done: status=complete duration=[\.0-9]+m?s (re)
-  [-0-9:.TWZ+]+ \[DEBUG] turbo: task hashes match (re)
-  
-   Tasks:    1 successful, 1 total
-  Cached:    0 cached, 1 total
-    Time:\s*[\.0-9]+m?s  (re)
-  
+  $ ${TURBO} build -vv --filter=util --force 1> VERBOSEVV 2>&1
```
NicholasLYang, line 38: Is that all that we need to test? Should we verify that the filtering/force flags are doing what they should?

gsoltis: I don't think we need to for this test. We have other tests for filtering and cache usage.

NicholasLYang (approved):

mehulkar (approved):

PR #6471, opened 2023-11-16
port(Turborepo): Add github prefixing in Rust

gsoltis:
### Description

Slightly hacky way of doing this. It's a little expandable for more providers, but not all of the way. Getting all the way will require a larger output refactor.

 - add support for a generic "fencepost" for grouped logs
 - add a couple github specific hooks

### Testing Instructions

Updated test for absolute paths in errors

Closes TURBO-1640


crates/turborepo-ui/src/output.rs
```diff
@@ -122,6 +145,10 @@ impl<W: Write> OutputClient<W> {
             // We hold the mutex until we write all of the bytes associated for the client
             // to ensure that the bytes aren't interspersed.
             let mut writers = writers.lock().expect("lock poisoned");
+            let (prefix, suffix) = fenceposts.map(|f| f.prefix_suffix()).unwrap_or_default();
+            if let Some(prefix) = prefix {
```
chris-olszewski, line 149:
```suggestion
            if let Some(prefix) = header {
```

crates/turborepo-ui/src/output.rs
```diff
@@ -112,6 +134,7 @@ impl<W: Write> OutputClient<W> {
             behavior,
             buffer,
             writers,
+            fenceposts,
```
chris-olszewski, line 137:
```suggestion
            header,
            footer,
```

crates/turborepo-ui/src/output.rs
```diff
@@ -133,6 +160,9 @@ impl<W: Write> OutputClient<W> {
                 };
                 writer.write_all(buffer)?;
             }
+            if let Some(suffix) = suffix {
```
chris-olszewski, line 163:
```suggestion
            if let Some(suffix) = footer {
```

crates/turborepo-lib/src/run/task_id.rs
```diff
@@ -99,6 +99,10 @@ impl<'a> TaskId<'a> {
         }
     }
 
+    pub fn as_github_task_id(&self) -> String {
+        format!("{}:{}", self.package, self.task)
+    }
```
chris-olszewski, line 104: Won't die on this hill, but this doesn't feel like it needs to be defined on `TaskId`, I think it would better be defined as a function in `turborepo-ci` maybe on the `Vendor` type.

chris-olszewski (approved):
Approving since we're moving quick, but I do think we can keep the `turborepo-ui` crate focused on just outputting stuff and not make it aware of CI vendors.

Very much looking forward to eventually getting https://github.com/vercel/turbo/pull/6314 merged in post-port.

PR #6475, opened 2023-11-16
fix: populate framework for every run

chris-olszewski:
### Description

Hoist the filling in framework in the task summary to happen at the graph traversal instead of being up to the exec function. Concretely this means that the `framework` field in every task will always be populated.

This lead to different values for `framework` in the resulting JSON depending on whether or not the tasks were actually executed. I realize we abuse the schema as it's used for both `--dry=json` and `--summarize`, but this feels like something that should be consistent. It feels odd that there's a case where `--dry=json` has more information available than an actual run summary.

### Testing Instructions

Updated integration test displays the behavior change. Other integration tests should continue passing.

Closes TURBO-1684


mehulkar (commented):
We had done this this way to keep the "No Framework detected" value for backwards compatibility in dry=json, and omitted from run summary because it's not really more explicit than an empty string. If we can make breaking changes, we should just remove it from dry json too

PR #6477, opened 2023-11-16
fix(turborepo): make linked installs test pass on Windows

mehulkar:
Closes TURBO-1686


crates/turborepo-lib/src/shim.rs
```diff
@@ -337,16 +337,17 @@ impl LocalTurboState {
...
+        // it points to. We do this _before_ traversing up to the parent,
+        // because on Windows, if you canonicalize a path that ends with `/..`
+        // it traverses to the parent directory before it follows the symlink,
+        // leading to the wrong place. We could separate the Windows
+        // implementation, but this workaround works for other platforms as
+        // well.
+        let canonical_path =
+            fs_canonicalize(root_path.as_path().join("node_modules").join("turbo")).ok()?;
+
+        AbsoluteSystemPathBuf::try_from(canonical_path.join("..")).ok()
```
chris-olszewski, line 350: If you have tried this, totally ignore the suggested code change, but maybe add a comment that warns that `foo/bar/.. != foo/` in this case.
```suggestion
        AbsoluteSystemPathBuf::try_from(canonical_path.parent()?).ok()
```

chris-olszewski (approved):
It's happening!

One suggestion that jumped out to me, but most certainly not blocking.

PR #6478, opened 2023-11-16
fix(docs): change default collapse level of sidebar

mehulkar:
| before | after |
|--------|--------|
|CleanShot 2023-11-16 at 01 42 16@2x| CleanShot 2023-11-16 at 01 42 02@2x

Closes TURBO-1687


anthonyshew (approved):
Yaaaaas.

PR #6492, opened 2023-11-17
Update rust-toolchain to 2023-11-16

wbinnssmith:
This updates the rust toolchain to 2023-11-16 and:

- Removes now-unnecessary `#![feature(async_fn_in_trait)]`
- Applies auto-fixable lint fixes
- Uses `Cargo.toml` new `lint` section instead of command line rustc flags

Test Plan: Tested with updated Next.js in a create-next-app

Closes PACK-1978


Cargo.toml
```diff
@@ -61,6 +61,9 @@ turbopack = [
   "path:crates/node-file-trace",
 ]
 
+[workspace.lints.clippy]
+too_many_arguments = "allow"
```
mehulkar, line 65: Do we have a lot of violations of this? Seems like a stylistic thing so fine to disable this rule, but curious if it should be localized or not

crates/turbopack-test-utils/Cargo.toml
```diff
@@ -9,6 +9,9 @@ autobenches = false
 [lib]
 bench = false
 
+[lints]
+workspace = true
```
mehulkar, line 13: Do we have to remember to add this to every crate?

mehulkar (approved):
For turborepo, I'm good with this if the same number of the Rust codepath tests pass as before this (I think there are only 1 or 2 failing at the moment), but I requested reviews from other on Turborepo team as well.

Cargo.toml
wbinnssmith: The violations are quite localized across 3 turbopack crates:

crates/turbopack-core -- 2
crates/turbopack-ecmascript -- 8
crates/turbopack-node -- 24

However it might be surprising for this only to be allowed in these crates alone rather than anywhere in the workspace.

arlyon (approved):

crates/turbopack-test-utils/Cargo.toml
wbinnssmith: Yeah, it seems so. Unfortunately I couldn't find a way around this.

https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo

wbinnssmith: One neat thing is `cargo new` seems to add it to new crates it generates in the workspace.

PR #6494, opened 2023-11-17
Trace output's source maps through input's source map

jridgewell:
### Description

This preserves an input file's original source map so that the transformed output source map can be traced through it.

### Testing Instructions

Closes PACK-1980
Closes PACK-537


crates/turbopack-ecmascript/src/parse.rs
```diff
@@ -71,47 +71,59 @@ impl PartialEq for ParseResult {
 #[turbo_tasks::value(shared, serialization = "none", eq = "manual")]
 pub struct ParseResultSourceMap {
     /// Confusingly, SWC's SourceMap is not a mapping of transformed locations
-    /// to source locations. I don't know what it is, really, but it's not
-    /// that.
+    /// to source locations. It's a map of filesnames to file contents.
```
sokra, line 74: 🤣

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -321,72 +329,6 @@ impl SourceMap {
         };
         Ok(OptionToken(token).cell())
     }
-
-    /// Traces another sourcemap's generated tokens through to this map's
-    /// original tokens. Useful for cases like mapping tokens from a minified
-    /// chunk back to its original sources.
-    #[turbo_tasks::function]
-    pub async fn trace(self: Vc<Self>, other: Vc<SourceMap>) -> Result<Vc<SourceMap>> {
```
sokra, line 329: Don't we need that for the next.js source map resolving from the error overlay?

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -219,13 +229,21 @@ impl AnalyzeEcmascriptModuleResultBuilder {
                 }
             }
         }
+        let mut source_map = None;
+        for map in self.source_maps {
+            if let Some(map) = &*map.await? {
+                source_map = Some(map.resolve().await?);
+                break;
+            }
+        }
```
sokra, line 238: Why are we collecting multiple Source Maps, when we only use the first one?

And: Should be use the last one instead?

sokra (approved):

sokra (approved):

anthonyshew (approved):

PR #6497, opened 2023-11-17
make bench tests required

sokra:
### Description

We need to run these test cases to ensure that they are not failing

### Testing Instructions

Closes PACK-1982


kwonoj (approved):

mehulkar (commented):
Hey I had moved these to a separate workflow previously. Any need to put them back into test.yml? Should be possible to make them required from that workflow also? test.yml is already huge and very difficult to understand. Also any changes to this file trigger expensive tests that serve no purpose in those PRs

sokra:
> Hey I had moved these to a separate workflow previously. Any need to put them back into test.yml? Should be possible to make them required from that workflow also? test.yml is already huge and very difficult to understand. Also any changes to this file trigger expensive tests that serve no purpose in those PRs

The "expensive" benchmarks are still in the other workflow. I only moved back the benchmark tests, which only take 2 mins.

sokra:
> test.yml is already huge and very difficult to understand.

Moved the new tests into a separate file. That can be done for all the jobs, if you are worried about the file size.

PR #6498, opened 2023-11-17
feat(ecmascript): support dynamic tenary operator jsvalue

kwonoj:
Closes [PACK-1963](https://linear.app/vercel/issue/PACK-1963)

Reattempt https://github.com/vercel/turbo/pull/6452

Closes PACK-1983


crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -469,9 +469,11 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
             if test.is_truthy() == Some(true) {
                 *value = take(cons);
                 true
-            } else {
+            } else if test.is_falsy() == Some(true) {
```
kwonoj, line 472: This was the culprit for the benchmark regresssion;

previously skipping falsy branching per https://github.com/vercel/turbo/pull/6452#discussion_r1393508034, however in that case non-truhy dynamic (i.e `rootElement === null`) is also considered as statically falsy and takes alt directly.

sokra (approved):

PR #6510, opened 2023-11-20
Extend resolving to allow reference type for plugins and extensions for MainField

sokra:
### Description

see also https://github.com/vercel/next.js/pull/58129

### Testing Instructions

Closes PACK-1991

### DIFF
--- a/crates/turbopack-core/src/resolve/mod.rs
+++ b/crates/turbopack-core/src/resolve/mod.rs
@@ -1341,11 +1341,20 @@ async fn resolve_into_folder(
                 return resolve_internal_inline(package_path, request.resolve().await?, options)
                     .await;
             }
-            ResolveIntoPackage::MainField(name) => {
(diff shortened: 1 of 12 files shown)


crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1333,12 +1341,21 @@ async fn resolve_into_folder(
...
+            ResolveIntoPackage::MainField {
+                field: name,
+                extensions,
+            } => {
                 if let Some(package_json) = &*read_package_json(package_json_path).await? {
                     if let Some(field_value) = package_json[name].as_str() {
                         let request =
                             Request::parse(Value::new(normalize_request(field_value).into()));
 
+                        let options = if let Some(extensions) = extensions {
```
wbinnssmith, line 1353: nit: this could use `map_or_else`

wbinnssmith (approved):

PR #6516, opened 2023-11-20
fix: Deadlock with closing process manager

NicholasLYang:
### Description

There was a deadlock where we were holding a RwLock across an await boundary, which causes a deadlock. Solved it by making sure the handle is dropped before the await.

### Testing Instructions

Adds a test by @chris-olszewski

Closes TURBO-1723


chris-olszewski (approved):
Can either wait for https://github.com/vercel/turbo/pull/6517 to get merged or just force-merge.

PR #6520, opened 2023-11-20
fix(ecmascript-runtime): relative url improvement + env setup

kwonoj:
### Description

This PR fixes 2 things:

First, put the runtime fn correctly to the corresponding environment. Previously node / dom env missed few runtime fn to resolve URLs, now those are correctly in place.

~Secondly, changes relative url behavior to have base for the rendering context such as `location.origin` or `server_addr`. This is mostly bringing back old urlrewrite implementation. Primary example where we currently fail is below:~ <- per discussion, this should be resolved in different way and this PR does not contain corresponding changes.

Closes PACK-1995


kwonoj:
Working on resolving test failure.

crates/turbopack-ecmascript-runtime/js/src/shared-node/node-externals-utils.ts
```diff
@@ -1,5 +1,21 @@
 /// <reference path="../shared/runtime-utils.ts" />
 
+declare var RUNTIME_PUBLIC_PATH: string;
+declare var OUTPUT_ROOT: string;
+declare var ASSET_PREFIX: string;
+
+const path = require("path");
```
sokra, line 7: This probably breaks on `edge`. edge also used `node-externals-utils.ts`

sokra: You probably need to move that into a separate file

kwonoj: Updated, and passing next.js CI : https://github.com/vercel/next.js/pull/58987

sokra (approved):

PR #6522, opened 2023-11-20
Fixes to Contributing guide

dallonf:
### Description

In trying to set up a local development environment, I ran into a couple of obstacles that needed some extra effort. Hopefully these updates to the guide will streamline the process for others in the future!

I don't know how to install a specific version of Go on Ubuntu via apt-get (I tried `apt-get install golang=1.20.7` in a Docker container, but it seems like that version isn't on the repository). Would appreciate input on that!

Note: I also ran into an issue where the `capnp` executable was missing. I installed it from Homebrew, but this is the first of the issues I ran into and I don't know for sure that installing the right version of Go wouldn't have taken care of the problem as well.

### Testing Instructions

These are mostly docs changes, so you can follow the docs and see if they work, especially in a clean environment that doesn't have the dependencies installed. I only tested these steps on macOS, but the docs for Chocolatey seemed clear enough that I'm confident in the edit.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -329,7 +329,8 @@ jobs:
         with:
           windows: ${{ matrix.os.name == 'windows' }}
           github-token: "${{ secrets.GITHUB_TOKEN }}"
-
+      - name: Setup Graphviz
+        uses: ts-graphviz/setup-graphviz@v1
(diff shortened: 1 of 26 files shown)


mehulkar (approved):
Thanks!

PR #6528, opened 2023-11-21
refactor: switch daemon version to be independent of turbo version

arlyon:
This change is backwards compatible, since we set the starting version of the daemon protocol to be equal to the current turbo version meaning we will not get any clashes with old clients reporting a bad version. Even though the version is currently equal to the turbo version, it is not expected to always be the case.

### Description

If we want to allow other tools to interact with the turbo daemon, we should not be as strict with our versioning. As it stands, a 3rd party tool such as an LSP would need to perfectly match the version of turbo used in the repo which is an unreasonable constraint.

This also relaxes the version requirement for all _new_ versions to least the same minor version (ie. that all the features available to the client are implemented by the server). Old clients will still demand identical versions (and have that demand respected correctly).

### Testing Instructions

Unit tests for the semver check included.

### DIFF
--- a/crates/turborepo-lib/src/daemon/client.rs
+++ b/crates/turborepo-lib/src/daemon/client.rs
@@ -5,20 +5,16 @@ use tonic::{Code, Status};
 use tracing::info;
 use turbopath::AbsoluteSystemPathBuf;
 
-use self::proto::turbod_client::TurbodClient;
 use super::{
     connector::{DaemonConnector, DaemonConnectorError},
     endpoint::SocketOpenError,
+    proto,
 };
-use crate::{get_version, globwatcher::HashGlobSetupError};
-
-pub mod proto {
-    tonic::include_proto!("turbodprotocol");
-}
+use crate::globwatcher::HashGlobSetupError;
 
 #[derive(Debug, Clone)]
 pub struct DaemonClient<T: Clone> {
-    client: TurbodClient<tonic::transport::Channel>,
+    client: proto::turbod_client::TurbodClient<tonic::transport::Channel>,
     connect_settings: T,
 }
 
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/daemon/mod.rs
```diff
@@ -11,4 +11,12 @@ pub use server::{serve, CloseReason};
...
 pub(crate) mod proto {
     tonic::include_proto!("turbodprotocol");
+
+    /// The version of the protocol that this library implements.
+    ///
+    /// Protocol buffers aim to be backward and forward compatible at a protocol
+    /// level, however that doesn't mean that our daemon will have the same
+    /// logical API. We may decide to change the API in the future, and this
+    /// version number will be used to indicate that.
+    pub const VERSION: &str = "1.10.17";
```
chris-olszewski, line 21: Just because this is confusing and it took me awhile to convince myself this is correct can we maybe focus this comment (or add to it) exactly when bumping this should happen and what major/minor/patch bumps indicate. I don't think this needs to be 3rd-party facing docs yet, just guidance for ourselves.

crates/turborepo-lib/src/daemon/client.rs
```diff
@@ -29,7 +25,8 @@ impl<T: Clone> DaemonClient<T> {
         let _ret = self
             .client
             .hello(proto::HelloRequest {
-                version: get_version().to_string(),
+                version: proto::VERSION.to_string(),
+                supported_version_range: proto::VersionRange::Minor.into(),
```
chris-olszewski, line 29: Should this be in a constant somewhere with a comment explaining when/why it should change? Do we expect it to always be `Minor`?

chris-olszewski (approved):
I think this looks good, some more documentation as to when some of the constants should change would be helpful, but not blocking.

crates/turborepo-lib/src/daemon/client.rs
arlyon: I think this will likely be the default. It is similar to the npm caret which is by far the most popular version specifier. I have added a comment

PR #6530, opened 2023-11-21
fix node.js externals

sokra:
### Description

* Extend resolving to allow reference type for plugins
* add extensions for MainField
* add an option `import_externals` to context to specify externals behavior for ESM imports
  * false: require() is used
  * true: import() is used and the module becomes an async module
* adds `import_externals: false` support for dynamic import
* adds `import_externals: true` support for normal import

see also https://github.com/vercel/next.js/pull/58129

### Testing Instructions

Closes PACK-2011


crates/turbopack/src/unsupported_sass.rs
```diff
@@ -1,10 +1,11 @@
 //! TODO(WEB-741) Remove this file once Sass is supported.
```
kwonoj, line 1: just nit but I guess this can be actually removed? prolly not for this pr.

kwonoj (approved):

PR #6531, opened 2023-11-21
fix(turbopack): postcss should be applied to `@import`ed CSS files

ForsakenHarmony:
### Description

Previously we only applied postcss to CSS files (not CSS modules) that were imported from javascript.

This PR makes sure the source transform gets applied to all CSS files.

### DIFF
--- a/crates/turbopack/src/module_options/mod.rs
+++ b/crates/turbopack/src/module_options/mod.rs
@@ -395,6 +395,32 @@ impl ModuleOptions {
                 ),
             ]);
         } else {
+            if let Some(options) = enable_postcss_transform {
+                let execution_context = execution_context
+                    .context("execution_context is required for the postcss_transform")?;
+
+                let import_map = if let Some(postcss_package) = options.postcss_package {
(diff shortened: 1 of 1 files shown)


sokra (approved):

crates/turbopack/src/module_options/mod.rs
```diff
@@ -404,40 +430,7 @@ impl ModuleOptions {
...
-                            ])))
-                        } else {
-                            None
-                        },
-                        Some(ModuleRuleEffect::ModuleType(ModuleType::CssGlobal)),
-                    ]
-                    .into_iter()
-                    .flatten()
-                    .collect(),
+                    [ModuleRuleEffect::ModuleType(ModuleType::CssGlobal)],
```
sokra, line 433:
```suggestion
                    vec![ModuleRuleEffect::ModuleType(ModuleType::CssGlobal)],
```

ForsakenHarmony: removed the `.collect()` after testing 🙈

PR #6535, opened 2023-11-21
Turbopack: Support styled issue titles and details

wbinnssmith:
This adds support for `StyledString`s to Turbopack issue `title` and `details` fields. It also makes returning a string optional for descriptions and details.

Closes PACK-2012


crates/turbo-tasks-fetch/src/lib.rs
```diff
@@ -153,29 +153,31 @@ impl Issue for FetchIssue {
     }
 
     #[turbo_tasks::function]
-    async fn description(&self) -> Result<Vc<StyledString>> {
+    async fn description(&self) -> Result<Vc<OptionStyledString>> {
```
sokra, line 156: Wouldn't it be easier to just add `StyledString::Nothing`?

wbinnssmith: Yeah, I considered this but thought it might be more expected/idiomatic to use a native `Option`, though that adds an additional Vc.

crates/turbopack-cli-utils/src/issue.rs
```diff
@@ -574,6 +574,10 @@ fn render_styled_string_to_ansi(styled_string: &StyledString) -> String {
 
 fn style_issue_source(plain_issue: &PlainIssue, context_path: &str) -> String {
     let title = &plain_issue.title;
+    let formatted_title = match title {
+        StyledString::Text(text) => text.bold().to_string(),
+        _ => render_styled_string_to_ansi(title),
+    };
```
sokra, line 580: That's a weird special case...

PR #6539, opened 2023-11-21
fix(gen): correct arg parsing

tknickman:
### Description

We don't need to split the args by space, the default behavior works fine. 

Fixes TURBO-6430

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -132,6 +132,15 @@
       "cwd": "${workspaceRoot}/examples/basic"
     },
     {
+      "name": "Generators",
+      "type": "lldb",
+      "request": "launch",
+      "preLaunchTask": "prepare turbo",
+      "program": "${workspaceRoot}/target/debug/turbo",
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -131,6 +131,15 @@
       "args": ["build"],
       "cwd": "${workspaceRoot}/examples/basic"
     },
+    {
+      "name": "Generators",
+      "type": "lldb",
+      "request": "launch",
+      "preLaunchTask": "prepare turbo",
+      "program": "${workspaceRoot}/target/debug/turbo",
+      "args": ["gen", "blog - release post", "--args", "1.11.0", "1.10.0", "_", "tagline about my really cool release"],
```
chris-olszewski, line 140: Do we want this to be a test?

chris-olszewski (approved):

.vscode/launch.json
tknickman: Yes will add in a follow up, I wanted to leave this in so we have a prebuild debug configuration for gen as well.

PR #6540, opened 2023-11-21
feat: add a go fallback flag

chris-olszewski:
### Description

Gives an easy way to opt-into the Go binary without messing with environment variables. (For almost all cases env var makes more sense, but on Windows you can't just `EXPERIMENTAL_RUST_CODEPATH=false turbo build`)

### Testing Instructions

Verify the following all work and invoke the correct binaries
```
# Rust binary
turbo build
EXPERIMENTAL_RUST_CODEPATH=true turbo build --filter=docs

# Go binary
turbo build --filter=docs --go-fallback
EXPERIMENTAL_RUST_CODEPATH=true turbo build --filter=docs --go-fallback
EXPERIMENTAL_RUST_CODEPATH=false turbo build --filter=docs
```

Closes TURBO-1734


NicholasLYang (approved):
One question, should we hide this flag? We might want users to know of it on their own

PR #6542, opened 2023-11-22
fix: analytics closing when all senders drop

chris-olszewski:
### Description

This PR primarily fixes an issue where the worker would spin if all senders were closed. Since the `select` was biased, if `self.rx.recv()` was ready then it would always be chosen. In the case that all senders are dropped, this future will always be ready with a `None` message. Solution is to just shutdown if there aren't going to be any other messages.

Other fixes:
 - Joining the senders as they finish instead of an arbitrary order and then waiting for whichever one we got to finish
 - Using the 20ms timeout before we send events if there's still room in the batch as opposed to 10s

### Testing Instructions

The unit test for closing now exercises realistic usage of the analytics client. If you checkout the first commit, the test will fail without the fix now.

Closes TURBO-1736

### DIFF
--- a/crates/turborepo-analytics/src/lib.rs
+++ b/crates/turborepo-analytics/src/lib.rs
@@ -165,6 +165,7 @@ mod tests {
     use std::{
         cell::RefCell,
         sync::{Arc, Mutex},
+        time::Duration,
     };
 
     use async_trait::async_trait;
@@ -324,7 +325,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_closing() {
-        let (tx, _) = mpsc::unbounded_channel();
+        let (tx, _rx) = mpsc::unbounded_channel();
 
         let client = DummyClient {
             events: Default::default(),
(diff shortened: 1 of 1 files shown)


crates/turborepo-analytics/src/lib.rs
```diff
@@ -324,7 +328,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_closing() {
-        let (tx, _) = mpsc::unbounded_channel();
+        let (tx, _rx) = mpsc::unbounded_channel();
```
arlyon, line 331: this should be a clippy lint.. got stuck just today not understanding why the daemon was shutting down because I assigned one end of a channel to `_`

arlyon (approved):

crates/turborepo-analytics/src/lib.rs
chris-olszewski: Yeah, I'm really surprised it's not. Creating a closed channel feels like it's almost never what a developer wants.

PR #6547, opened 2023-11-22
Update with-docker example.

anthonyshew:
Updating the with-docker example!

Closes TURBO-1707

Things to pay attention to:


chris-olszewski (approved):
LGTM

PR #6548, opened 2023-11-22
Update with-changesets.

anthonyshew:
Updating the with-changesets example!

Closes TURBO-1705

Things to pay attention to:


examples/with-create-react-app/apps/docs/package.json
```diff
@@ -12,21 +12,21 @@
     "clean": "rm -rf build"
   },
   "dependencies": {
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
-    "react-scripts": "5.0.0",
+    "react": "^18.2.0",
```
chris-olszewski, line 15: Are these changes meant for this PR?

chris-olszewski (changes requested):
I think the `with-docker` changes should get removed from this PR since #6547 exists.
(`with-changesets` updates LGTM)

examples/with-create-react-app/apps/docs/package.json
anthonyshew: Oops! Fixed.

chris-olszewski (approved):

tknickman (approved):

examples/with-changesets/packages/eslint-config/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@acme/eslint-config",
+  "version": "0.0.0",
+  "private": true,
+  "files": [
+    "library.js",
+    "next.js",
+    "react-internal.js"
+  ],
```
tknickman, line 9: I'm tempted to remove the files key just to prevent confusion. This is only used if you're packing the package to determine which files get included in the published package. It doesn't actually have anything to do with making the files available to be imported from in the way we're using it here - which I'll bet is what some may assume by just looking at this. 

But not blocking, we can hit that in a follow up!

PR #6555, opened 2023-11-23
Update design-system example.

anthonyshew:
Updating the design-system example!

Closes TURBO-1702

Things to pay attention to:


examples/design-system/apps/docs/stories/button.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import type { Meta, StoryObj } from "@storybook/react";
-import { Button } from "ui";
+import { Button } from "@acme/ui";
```
mehulkar, line 2: Did we change the org in these names to `@repo` everywhere?

examples/design-system/package.json
```diff
@@ -1,21 +1,19 @@
 {
   "private": true,
   "scripts": {
-    "build": "turbo run build",
```
mehulkar, line 4: We should keep `run` in these. The general advice from @nathanhammond has been to not use shorthand syntax when putting turbo commands in scripts. I generally agree.

examples/design-system/packages/ui/package.json
```diff
@@ -12,17 +12,18 @@
   "scripts": {
     "build": "tsup src/index.tsx --format esm,cjs --dts --external react",
     "dev": "tsup src/index.tsx --format esm,cjs --watch --dts --external react",
-    "lint": "eslint src/",
+    "lint": "eslint .",
```
mehulkar, line 15: In another PR (with-npm update), there was also `--max-warnings=0`. Do we want to be consistent about that?

mehulkar (approved):
Some fix ships on this one, nothing blocking if tests are passing

examples/design-system/apps/docs/stories/button.stories.tsx
anthonyshew: Went with `@acme` on this one since it's got a publishing workflow in it.

PR #6556, opened 2023-11-23
Remove create-react-app example.

anthonyshew:
We're proposing to get rid of this example for a few reasons:

- create-react-app itself is unmaintained.
- One less thing for us to maintain.
- It appears the community isn’t using this example much based on how much chatter I see on other examples but not this one.
- create-react-app is fundamentally uninteresting from `turbo`’s perspective. It doesn’t have any quirks to manage.

Closes TURBO-1706


chris-olszewski (changes requested):
🔥

tknickman (approved):
🚀

tknickman:
We can delete the project on Vercel when this is merged as well

chris-olszewski (approved):

PR #6557, opened 2023-11-23
Update with-yarn example.

anthonyshew:
This example was freshly bootstrapped with our new `create-turbo`.


tknickman:
Looks like you might need to update prsk tests here to capture the different package names in the output (I _think_).

PR #6558, opened 2023-11-23
Update with-npm example.

anthonyshew:
This example is freshly bootstrapped with our new `create-turbo`.


examples/with-npm/.npmrc
```diff
@@ -0,0 +1 @@
+auto-install-peers = true
```
mehulkar, line 1: is this new? is this based on npm version that it was generated with? is this a next.js addition?

examples/with-npm/README.md
```diff
@@ -7,20 +7,20 @@ This is an official starter Turborepo.
 Run the following command:
 
 ```sh
-npx create-turbo@latest -e with-npm
```
mehulkar, line 10: We probably need to keep this in this example

examples/with-npm/README.md
```diff
@@ -7,20 +7,20 @@ This is an official starter Turborepo.
...
+npx create-turbo@latest
 ```
 
 ## What's inside?
 
-This Turborepo uses [npm](https://www.npmjs.com/) as a packages manager. It includes the following packages/apps:
```
mehulkar, line 15: Keep this?

examples/with-npm/README.md
```diff
@@ -38,7 +38,7 @@ To build all apps and packages, run the following command:
 
 ```
 cd my-turborepo
-npm run build
+pnpm build
```
mehulkar, line 41: Revert this

examples/with-npm/README.md
```diff
@@ -47,7 +47,7 @@ To develop all apps and packages, run the following command:
 
 ```
 cd my-turborepo
-npm run dev
+pnpm dev
```
mehulkar, line 50: Revert this

examples/with-npm/apps/docs/package.json
```diff
@@ -6,20 +6,23 @@
     "dev": "next dev --port 3001",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint"
+    "lint": "eslint . --max-warnings 0"
```
mehulkar, line 9: Is this right? Did `next lint` go away?

examples/with-npm/apps/web/README.md
```diff
@@ -3,7 +3,7 @@
 First, run the development server:
 
 ```bash
-pnpm dev
+yarn dev
```
mehulkar, line 6: Fix this? How did yarn sneak in here

examples/with-npm/package.json
```diff
@@ -1,19 +1,21 @@
...
-    "eslint": "7.32.0",
-    "prettier": "^2.5.1",
-    "turbo": "^1.9.3"
+    "@repo/eslint-config": "*",
+    "@repo/typescript-config": "*",
+    "prettier": "^3.1.0",
+    "turbo": "latest"
   },
-  "packageManager": "npm@8.19.4"
+  "packageManager": "npm@9.6.7",
```
mehulkar, line 16: Are we sure about this? We're using and recommending Node 18 for the most part everywhere, which I believe ships with npm@8. If we're bumping everything to Node 20, we should make that clear everywhere

examples/with-npm/packages/eslint-config/README.md
```diff
@@ -0,0 +1,3 @@
+# `@turbo/eslint-config`
```
mehulkar, line 1: Should this be `@repo/eslint-config`?

examples/with-npm/turbo.json
```diff
@@ -4,13 +4,14 @@
   "pipeline": {
     "build": {
       "dependsOn": ["^build"],
-      "outputs": ["dist/**", ".next/**", "!.next/cache/**"]
+      "outputs": [".next/**", "!.next/cache/**"]
     },
     "lint": {
-      "outputs": []
+      "dependsOn": ["^lint"]
```
mehulkar, line 10: `^lint` is probably unnecessary. all lints can happen in parallel, there are no dependencies

mehulkar (changes requested):
Looks mostly good, but some boilerplate needs to be changed

examples/with-npm/apps/docs/package.json
anthonyshew: In a monorepo context, I've found it easier to drop it and use ESLint with the plugin directly instead.

examples/with-npm/.npmrc
anthonyshew: Think it was there for historical ESLint-resolving reasons but shouldn't need it anymore.

examples/with-npm/turbo.json
anthonyshew: Ah, yeah, there's a difficulty here; I don't love it.

- `[]` means you can change `eslint-config` and still hit cache - problematic
- `["^lint"]` means not the best parellalization but at least it's correct - still not great
- Could use the `["^topo"]` technique here but feels a little too advanced for a starter.

What you think?

mehulkar: Hmm, I see. I would use the `topo` config, since that's actually our recommended way of accomplishing this, even though it's confusing. IMO, we should show parallelism for tasks that don't depend on each other.

mehulkar (approved):
Approving, but still feel the `^lint` config shouldn't be there in the example. Not blocking enough for me though if you have a different opinion!

PR #6568, opened 2023-11-24
Set engine on create-turbo.

anthonyshew:
🚂

### DIFF
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -15,6 +15,6 @@
   },
   "packageManager": "pnpm@8.9.0",
   "engines": {
-    "node": "18"
+    "node": ">=18"
   }
 }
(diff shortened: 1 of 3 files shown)


examples/basic/package.json
```diff
@@ -11,7 +11,10 @@
     "@repo/eslint-config": "workspace:*",
     "@repo/typescript-config": "workspace:*",
     "prettier": "^3.1.0",
-    "turbo": "latest"
+    "turbo": "^1.10.16"
   },
-  "packageManager": "pnpm@8.9.0"
+  "packageManager": "pnpm@8.9.0",
+  "engines": {
+    "node": "18"
```
tknickman, line 18: Let's include a range so users with strict engines and 20 etc. don't get a warning:
```suggestion
    "node": ">=18"
```

anthonyshew: TIL

tknickman (approved):

PR #6571, opened 2023-11-24
Update with-prisma example.

anthonyshew:
Updating the with-prisma example!

Closes TURBO-1708

Things to pay attention to:

- [ ] Does auto-importing work across packages? _This is suspiciously not working and I can't get it to...Is `yarn` doing something weird?_


examples/with-prisma/packages/database/src/client.ts
```diff
@@ -1,6 +1,7 @@
 import { PrismaClient } from "@prisma/client";
 
 declare global {
+  // eslint-disable-next-line no-unused-vars
   var prisma: PrismaClient | undefined;
```
mehulkar, line 5: hmm, is this necesssary?

mehulkar (approved):

examples/with-prisma/packages/database/src/client.ts
anthonyshew: Define necessary.

mehulkar: heh, sorry, I meant is that eslint disabling necessary. What is this declare block is doing?

mehulkar: I guess you removed it already :)

PR #6573, opened 2023-11-24
Update with-rollup example.

anthonyshew:
Updating the with-rollup example!

Closes TURBO-1713

Things to pay attention to:


examples/with-rollup/packages/config-eslint/README.md
```diff
@@ -0,0 +1,3 @@
+# `@turbo/eslint-config`
```
tknickman, line 1:
```suggestion
# `@repo/eslint-config`
```

examples/with-rollup/packages/ui/package.json
```diff
@@ -1,23 +1,31 @@
...
-    "@types/react": "^18.2.5",
-    "@types/react-dom": "^18.2.4",
-    "eslint": "^7.32.0",
-    "eslint-config-custom": "workspace:*",
+    "@repo/eslint-config": "workspace:*",
+    "@repo/typescript-config": "workspace:*",
+    "@rollup/plugin-typescript": "^11.1.5",
+    "@types/react": "^18.2.38",
+    "@types/react-dom": "^18.2.17",
+    "eslint": "^8.54.0",
```
tknickman, line 26: Move eslint to root

tknickman (approved):

examples/with-rollup/packages/ui/package.json
anthonyshew: I've always operated as, conceptually, a package should be as hermetic as possible?

tknickman: @anthonyshew generally yes, for anything needed to build / run. But for dev requirements root is ok. We do this with prettier currently. It's also nicer for Eslint because version drift can cause mayhem and it's easy to do accidentally. So keeping one single version at root is a much nicer dev experience with no drawbacks.

PR #6575, opened 2023-11-24
Update with-tailwind example.

anthonyshew:
Updating the with-tailwind example!

Closes TURBO-1716

Things to pay attention to:


examples/with-tailwind/packages/ui/package.json
```diff
@@ -1,34 +1,33 @@
...
+    "./card": "./src/card.tsx"
   },
   "license": "MIT",
   "scripts": {
-    "build": "tsup",
+    "build": "tailwindcss -i ./src/styles.css -o dist/index.css",
     "lint": "eslint src/",
-    "dev": "tsup --watch",
-    "check-types": "tsc --noEmit"
+    "dev": "tailwindcss -i ./src/styles.css -o ./dist/index.css --watch",
```
anthonyshew, line 15: This task is working correctly when ran outside of `turbo` but exits early when ran with `turbo`.

anthonyshew: Update: This task works on `canary`.

chris-olszewski (approved):

anthonyshew:
1) There's a tradeoff here. The strategy used in the StackOverflow leaks across package boundaries (the `../../packages/` path) - while the strategy used here can indeed duplicate similar sets of CSS (but this shouldn't matter because of gzip + Tailwind's maximum compile size). (Notably, the `ui` package uses a `ui-` prefix for its classes so the styles won't be able to clobber each other.)
2) Auto-imports are already good to go in this example!
3) Updating it in #6755! Thanks for letting me know.  (Wrote down that first point in the README)

magnusriga:
@anthonyshew Thanks! One quick observation and an additional suggestion:

**Observation**
This works for me in exports with auto-import:

```json
    "./*": {
      "import": "./src/*.tsx"
    }
```

This does not work:
```json
    "./*": "./src/*.tsx"
```

Check out the first of the two, perhaps it works for you too. Unfortunately it does not work if you remove the file extension from the RHS. Thus, it does not cover all your files if you have both JS and TS. Which brings me back to the start, i.e. should I transpile it to JS with `tsc` after all?

**Suggestion**
On a similar note: Many people still write in JS (as opposed to only in TS). Do you think you could update the config to support both? I tweaked it on my end to do so by: (i) Including `eslint-import-resolver-exports` in `config-eslint/package.json` (and activating it in `react.js` and `next.js` in the same folder), and (ii) Forcing Eslint to use `"@babel/eslint-parser"` for JS files via `overrides` in `config-eslint/react.js`.

magnusriga:
@anthonyshew I just wanted to point out that importing the css from tailwind twice, like you do in your example, will indeed cause issues in almost all use cases. The reason is that the normalization css rules from tailwind will be duplicated in the output, and the ones from `@repo/ui` will land **after** the rules for the main app. As you know, the order of the css rules matter, so this breaks all the formatting.

**Example**: If you set the color of an h2 element directly on an element via tailwind (in the main app), this color will be overwritten by the `@repo/ui/styles.css` rules you import in layout.js.

**Question**: Why do we need to compile those tailwind rules in the build step in `@repo/ui`? Can't we just import directly from `src` and let the main app do the tailwind processing?

anthonyshew:
The `ui` prefix prevents the problem you're describing: https://github.com/vercel/turbo/blob/9b9c8c0afabde189d85c391de658e08cc490cbe8/examples/with-tailwind/packages/ui/tailwind.config.ts#L6

The approach you're describing is also valid. It does leak externals but, if you're willing to make that tradeoff, it also works.

magnusriga:
> The `ui` prefix prevents the problem you're describing: https://github.com/vercel/turbo/blob/9b9c8c0afabde189d85c391de658e08cc490cbe8/examples/with-tailwind/packages/ui/tailwind.config.ts#L6
> 
> The approach you're describing is also valid. It does leak externals but, if you're willing to make that tradeoff, it also works. 

Hi @anthonyshew . The prefix is not added to the base/default/normalization (call them what you like) rules. They are only added to the component rules.

So, in the single output sheet you might have this order:

```css
// From web's tailwind
h2 { font-size: ... }
.text-lg { font-size: ... }

// From @repo/ui's tailwind
h2 { font-size: ... }
.ui-text-lg { font-size: ... }
```
I recognize that the class rule takes precedence here, and as such there is no output issue in the example above, however I have experienced problems with certain rules. It seems like a bad idea to mess with the order (and create an incorrect final css sheet) as the order is crucial in CSS. Anyways, as you say, up to each person.

As a side note, the catch-all wildcard in the exports works for me (I got help from the TS team to figure it out):

./*:  ./src/*

We got full Intellisense, on both the right hand and left hand sides of imports, using that export.

PR #6577, opened 2023-11-25
Update kitchen-sink example.

anthonyshew:
- Bump all deps.
- Live mas.
- eslint


examples/kitchen-sink/package.json
```diff
@@ -1,20 +1,20 @@
...
-    "jest": "^29.6.4",
-    "prettier": "^2.7.1",
-    "turbo": "^1.9.3"
+    "prettier": "^3.1.0",
+    "turbo": "^1.10.16"
   },
-  "packageManager": "pnpm@8.9.0"
+  "packageManager": "pnpm@8.9.0",
+  "engines": {
+    "node": ">=18"
```
mehulkar, line 18: ❤️ 💯

examples/kitchen-sink/packages/logger/package.json
```diff
@@ -15,14 +15,15 @@
...
-    "jest-presets": "workspace:*",
-    "tsconfig": "workspace:*",
-    "tsup": "^6.0.1",
-    "typescript": "^5.2.2"
+    "@types/jest": "^29.5.10",
+    "@types/node": "^20.10.0",
+    "@repo/eslint-config": "workspace:*",
+    "@repo/jest-presets": "workspace:*",
+    "@repo/typescript-config": "workspace:*",
+    "tsup": "^8.0.1",
```
mehulkar, line 26: Should we get rid of this if we're going to follow the internal package pattern?

mehulkar (approved):
Not for this PR, but couple high level things:


1. Consider composing this kitchen-sink example with other examples, instead of maintaining all these apps. E.g in semi-pseudo code:

    ```sh
    npx create-turbo@latest mynewmonorepo -e nextjs
    cd mynewmonorepo

    # below is pseudo code, don't think it works like this, but not sure
    npx turbo gen workspace --type app --copy sveltekit --destination apps/admin --name admin

    npx turbo gen workspace --type app --copy astro --destination apps/blog --name blog

    # and so on
    ```
1. Maybe consider renaming `kitchen-sink` to something like "multi-framework" or someething?

examples/kitchen-sink/packages/logger/package.json
anthonyshew: My thinking is to leave in an example of a compiled package while we're still evaluating for future iterations of how we look for how the examples can serve the community best.

There are still a few of these sprinkled around that I've left that way strategically. It's a good callout and a good question to take under consideration going forward.

PR #6580, opened 2023-11-25
chore(docs): remove backticks from title/desc meta tags on CLI reference pages

mehulkar:
I originally added these, but I think they look kind of bad now

Closes TURBO-1746


tknickman (approved):
Into it.

PR #6590, opened 2023-11-27
implement basic side effects optimization

sokra:
### Description

* optimizes reexports for modules that are inside of folders with a package.json that contains `"sideEffects": false`
* For named reexports
  * skips resolving, reading, parsing, transforms, code generation and bundling for unrelated reexports
  * skips code generation and bundling for the module containing the followed reexport
* For star reexports
  * skips code generation and bundling for unrelated reexports before the followed reexport
  * skips resolving, reading, parsing, transforms, code generation and bundling for unrelated reexports after the followed reexport
  * skips code generation and bundling for the module containing the followed reexport

Follow-up work:
* `sideEffects` in package.json also supports a glob. We need to support that too.
* renaming exports is not yet supported. We need to create a naming module for that to avoid effects on the importer side.

Closes PACK-2042


crates/turbopack-core/src/context.rs
```diff
@@ -32,7 +32,7 @@ pub trait AssetContext {
         self: Vc<Self>,
         asset: Vc<Box<dyn Source>>,
         reference_type: Value<ReferenceType>,
-    ) -> Vc<Box<dyn Module>>;
+    ) -> Vc<OptionModule>;
```
ForsakenHarmony, line 35: This should have a comment in which cases it can return `None`

crates/turbopack-css/src/module_asset.rs
```diff
@@ -149,7 +153,9 @@ impl ModuleCssAsset {
 
     #[turbo_tasks::function]
     async fn classes(self: Vc<Self>) -> Result<Vc<ModuleCssClasses>> {
-        let inner = self.inner();
+        let Some(inner) = *self.inner().await? else {
+            bail!("inner asset should be CSS processable");
+        };
```
ForsakenHarmony, line 158: All of these could probably be
```suggestion
        let inner = *self.inner().await?.context("inner asset should be CSS processable")?;
```

It also kinda sounds like an error condition, shouldn't it be a result instead of an option?

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -84,14 +85,41 @@ pub enum SpecifiedModuleType {
...
+    ReexportsOnly,
+}
+
+impl TreeShakingMode {
+    pub fn skip_over_side_effect_free_reexports(&self) -> bool {
+        match self {
+            TreeShakingMode::ModuleFragments => true,
+            TreeShakingMode::ReexportsOnly => true,
+        }
+    }
```
ForsakenHarmony, line 115: in which case will this return false?
should we add that case now?

crates/turbopack/examples/turbopack.rs
```diff
@@ -64,12 +64,16 @@ async fn main() -> Result<()> {
                 .cell(),
                 Vc::cell("default".to_string()),
             );
-            let module = module_asset_context.process(
-                Vc::upcast(source),
-                Value::new(turbopack_core::reference_type::ReferenceType::Undefined),
-            );
-            let rebased = RebasedAsset::new(module, input, output);
-            emit_with_completion(Vc::upcast(rebased), output).await?;
+            if let Some(module) = *module_asset_context
```
ForsakenHarmony, line 67: shouldn't this handle the none/error case as well?

crates/turbopack-ecmascript/src/references/esm/export.rs
```diff
@@ -35,6 +37,183 @@ pub enum EsmExport {
...
+        }))
+    }
+}
+
+#[turbo_tasks::function]
+pub async fn follow_reexports_internal(
+    mut module: Vc<Box<dyn EcmascriptChunkPlaceable>>,
+    mut export_name: String,
+) -> Result<Vc<FollowExportsResult>> {
+    loop {
```
ForsakenHarmony, line 102: this could also be flattened a bit

crates/turbopack-ecmascript/src/references/esm/export.rs
```diff
@@ -35,6 +37,183 @@ pub enum EsmExport {
...
+                    }
+                } else {
+                    false
+                }
+            } else {
+                false
+            }
+        }
+        _ => false,
+    }))
```
ForsakenHarmony, line 78: you can flatten this with let-else

ForsakenHarmony (changes requested):
seems good apart from some missing documentation and convoluted code 🙃 

sorry for the individual comments, I often forget to submit reviews if I don't post the comments right away

crates/turbopack/examples/turbopack.rs
sokra: When the entry module resolves to nothing we don't have to emit anything

sokra:
> sorry for the individual comments, I often forget to submit reviews if I don't post the comments right away

same

crates/turbopack-css/src/module_asset.rs
```diff
@@ -149,11 +153,14 @@ impl ModuleCssAsset {
...
     async fn classes(self: Vc<Self>) -> Result<Vc<ModuleCssClasses>> {
-        let inner = self.inner();
-
-        let Some(inner) = Vc::try_resolve_sidecast::<Box<dyn ProcessCss>>(inner).await? else {
-            bail!("inner asset should be CSS processable");
-        };
+        let inner = self
+            .inner()
+            .await?
+            .context("inner asset should be CSS processable")?;
```
ForsakenHarmony, line 159: still not sure about this, I guess right now this can't happen

but if it does it would just throw a random error not telling you what the inner asset was and why it's not processable

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -341,7 +341,6 @@ impl ModuleResolveResultOption {
 #[derive(Clone, Debug)]
 pub enum ResolveResultItem {
     Source(Vc<Box<dyn Source>>),
-    OriginalReferenceExternal,
```
sokra, line 344: This has been removed in favor of only `OriginalReferenceTypeExternal`

crates/turbopack-core/src/context.rs
```diff
@@ -9,34 +10,66 @@ use crate::{
     source::Source,
 };
 
+#[turbo_tasks::value(shared)]
+pub enum ProcessResult {
+    Module(Vc<Box<dyn Module>>),
+    Ignore,
+}
```
sokra, line 17: `AssetContext::process` returns this enum now. Instead of returning a module it might ignore the reference. This won't include the module at that place.

It's common that callers actually need a Module, so there is a `ProcessResult::module()` method which gives the module and throws otherwise.

crates/turbopack-core/src/resolve/mod.rs
ForsakenHarmony: finally haha

but maybe rename `OriginalReferenceTypeExternal` to `ExternalReference` 🤔

ForsakenHarmony (approved):
Still don't really think `ProcessResult` has enough detail to be actionable/informative, but shipping this is more important right now and I think there are a lot of other places with suboptimal error reporting, so I don't think that should block this Pr right now.

crates/turbopack-ecmascript/src/chunk/placeable.rs
```diff
@@ -1,14 +1,47 @@
...
+    let find_package_json: turbo_tasks::ReadRef<FindContextFileResult> =
+        find_context_file(path.parent(), package_json()).await?;
+
+    if let FindContextFileResult::Found(package_json, _) = *find_package_json {
+        if let FileJsonContent::Content(content) = &*package_json.read_json().await? {
+            if let Some(side_effects) = content.get("sideEffects") {
+                if let Some(side_effects) = side_effects.as_bool() {
+                    return Ok(Vc::cell(!side_effects));
+                } else {
+                    // TODO it might be a glob too, handle that case too
```
ForsakenHarmony, line 38: should probably create a linear issue for all the todos and reference them in the comments

ForsakenHarmony (approved):
LGTM, would just be good to have issues referenced by all the TODO comments

PR #6593, opened 2023-11-27
fix(benchmarks): correct lint & types

tknickman:
### Description

These type and lint errors keep getting flagged on every PR as part of "Unchanged files with check annotations Beta" ([example](https://github.com/vercel/turbo/pull/6592/files))

This fixes the issues to quiet the warnings. 

Closes TURBO-1757

### DIFF
--- a/packages/turbo-benchmark/src/index.ts
+++ b/packages/turbo-benchmark/src/index.ts
@@ -1,8 +1,17 @@
 import cp from "node:child_process";
-import fs from "node:fs";
+import {
+  readFileSync,
+  writeFileSync,
+  existsSync,
+  rmSync,
+  mkdirSync,
+  closeSync,
+  openSync,
+  writeSync,
+} from "node:fs";
 import path from "node:path";
-import fse from "fs-extra";
-import ndjson from "ndjson";
+import { copySync } from "fs-extra";
+import { stringify } from "ndjson";
 import {
   DEFAULT_EXEC_OPTS,
   getCommitDetails,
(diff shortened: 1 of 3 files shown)


packages/turbo-benchmark/src/index.ts
```diff
@@ -1,8 +1,17 @@
...
+  copySync,
+  mkdirSync,
+  rmSync,
+  writeFileSync,
+  readFileSync,
+  existsSync,
+  closeSync,
+  openSync,
+  writeSync,
+} from "fs-extra";
```
mehulkar, line 13: Can we keep the `node:fs` imports for the ones that were previously using them? I'd prefer not to use 3rd party libraries (even if they're just proxying back to the standard lib)

mehulkar (changes requested):

packages/turbo-benchmark/src/index.ts
tknickman: @mehulkar we could, sure - but it's more confusing to use two different FS libs. We are already installing it, why not use it for everything?

mehulkar: I would prefer to be explicit about when the built in std library doesn't cover the functionality we need. It would be somewhat self-documenting why we chose to use the 3rd party library. If we universally use the 3rd party library, it will be difficult to understand later whether it's _actually_ necessary or not.

tknickman: Fair enough, for something internal I don't think it's necessary, dev prod > bundle size. But can remove if you prefer.

tknickman: @mehulkar updated

mehulkar (approved):

PR #6596, opened 2023-11-27
fix: Hard error on spaces

NicholasLYang:
### Description

When we fail to start the `SpacesClient`, this can lead to an error when we try to close it at the end of a run. Instead of erroring, we simply log out the error and continue, since spaces is not critical to execution.

### Testing Instructions

Added a test for spaces failure that ensures we exit with 0

Closes TURBO-1764

### DIFF
--- a/crates/turborepo-lib/src/run/summary/mod.rs
+++ b/crates/turborepo-lib/src/run/summary/mod.rs
@@ -21,7 +21,7 @@ pub use spaces::{SpacesTaskClient, SpacesTaskInformation};
 use svix_ksuid::{Ksuid, KsuidLike};
 use tabwriter::TabWriter;
 use thiserror::Error;
-use tracing::log::warn;
+use tracing::{error, log::warn};
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf, AnchoredSystemPath};
 use turborepo_api_client::{spaces::CreateSpaceRunPayload, APIAuth, APIClient};
 use turborepo_env::EnvironmentVariableMap;
@@ -395,8 +395,14 @@ impl<'a> RunSummary<'a> {
 
         if let Some(spaces_client_handle) = self.spaces_client_handle.take() {
             println!("Sending to space");
-            self.send_to_space(spaces_client_handle, end_time, exit_code)
-                .await?;
+            // We log the error here but don't fail because
+            // failing to send the space shouldn't fail the run.
+            if let Err(err) = self
+                .send_to_space(spaces_client_handle, end_time, exit_code)
+                .await
+            {
+                warn!("Error sending to space: {}", err);
+            }
         }
 
         Ok(())
--- a/crates/turborepo-lib/src/run/summary/mod.rs
+++ b/crates/turborepo-lib/src/run/summary/mod.rs
@@ -394,15 +394,8 @@ impl<'a> RunSummary<'a> {
         }
 
         if let Some(spaces_client_handle) = self.spaces_client_handle.take() {
-            println!("Sending to space");
-            // We log the error here but don't fail because
-            // failing to send the space shouldn't fail the run.
-            if let Err(err) = self
-                .send_to_space(spaces_client_handle, end_time, exit_code)
-                .await
-            {
-                warn!("Error sending to space: {}", err);
-            }
+            self.send_to_space(spaces_client_handle, end_time, exit_code)
+                .await;
         }
 
         Ok(())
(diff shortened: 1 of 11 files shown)


turborepo-tests/integration/tests/spaces-failure.t
```diff
@@ -0,0 +1,6 @@
+Setup
+  $ . ${TESTDIR}/../../helpers/setup.sh
+  $ . ${TESTDIR}/_helpers/setup_monorepo.sh $(pwd) spaces_failure
+
+Ensures that even when spaces fails, the build still succeeds.
+  $ ${TURBO} run build --token foobarbaz --team bat --output-logs none > /dev/null 2>&1
```
chris-olszewski, line 6: Can we use the `--api` flag to avoid hitting prod in our tests runs?

crates/turborepo-lib/src/run/summary/mod.rs
```diff
@@ -407,22 +406,27 @@ impl<'a> RunSummary<'a> {
...
+        let spinner = tokio::spawn(async {
+            tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+            turborepo_ui::start_spinner("...sending run summary...");
+        });
+
+        // We log the error here but don't fail because
+        // failing to send the space shouldn't fail the run.
+        if let Err(err) = spaces_client_handle.finish_run(exit_code, ended_at).await {
+            warn!("Error sending to space: {}", err);
+        };
```
chris-olszewski, line 419:
```suggestion
       let (result, spinner) = tokio::join!(spaces_client_handle.finish_run(exit_code, ended_at), spinner_fut);
       if let Err(err) = result {
           warn!("Error sending to space: {}", err);
       };
```

chris-olszewski (changes requested):
Primarily concerned with avoiding hitting prod in our test runs.

I also think we can avoid using a separate task to manage the spinner.

crates/turborepo-lib/src/run/summary/mod.rs
```diff
@@ -407,22 +406,27 @@ impl<'a> RunSummary<'a> {
...
         exit_code: i32,
-    ) -> Result<(), Error> {
-        let spinner = turborepo_ui::start_spinner("...sending run summary...");
-
-        spaces_client_handle.finish_run(exit_code, ended_at).await?;
+    ) {
+        let spinner = tokio::spawn(async {
+            tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+            turborepo_ui::start_spinner("...sending run summary...");
+        });
```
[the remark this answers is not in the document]
NicholasLYang: Will this get triggered? I was under the impression that this wouldn't get run unless we actively await or spawn it

crates/turborepo-lib/src/run/summary/mod.rs
```diff
@@ -407,22 +406,28 @@ impl<'a> RunSummary<'a> {
...
+    ) {
+        let spinner_fut = async {
+            tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+            turborepo_ui::start_spinner("...sending run summary...")
+        };
 
-        spaces_client_handle.finish_run(exit_code, ended_at).await?;
+        // We log the error here but don't fail because
+        // failing to send the space shouldn't fail the run.
+       let (result, spinner) = tokio::join!(spaces_client_handle.finish_run(exit_code, ended_at), spinner_fut);
```
NicholasLYang, line 417: Actually, wouldn't this wait on both the spinner and `finish_run`? So we'd be waiting a whole second unnecessarily. @chris-olszewski

chris-olszewski (approved):

turborepo-tests/integration/tests/_fixtures/spaces_failure/turbo.json
```diff
@@ -0,0 +1,24 @@
...
+      "dotEnv": [".env.local"]
+    },
+
+    "something": {},
+    "//#something": {},
+
+    "maybefails": {}
+  },
+  "experimentalSpaces": {
+    "id": "my-space"
```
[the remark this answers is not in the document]
mehulkar: @NicholasLYang If you're going to create a separate fixture, make sure to delete things that aren't relevant to it (like most of this turbo.json. Also, there's a section in the README for if what you really want is a custom turbo.json https://github.com/vercel/turbo/tree/main/turborepo-tests/integration#custom-turbojson

PR #6597, opened 2023-11-27
chore(docs): tweak getting started docs

mehulkar:
- Assume existing project in the Add to Existing Project page
- ~~Include all content in ordered list items~~ (prettier can't handle this formatting)
- Re-order Getting Started pages

Closes TURBO-1765


docs/pages/repo/docs/getting-started/add-to-project.mdx
```diff
@@ -78,70 +65,70 @@ For more information on configuring your `turbo.json`, see the [Configuration Op
...
+   ```json filename="package.json"
+   {
+     "scripts": {
+       "build": "tsc && vite build"
+     }
+   }
+   ```
 
-We recommend splitting these into a `lint` and `build` script.
+   We recommend splitting these into a `lint` and `build` script.
```
anthonyshew, line 78: Let's use `type-check` instead of lint here. `tsc` is technically more linter-ish but TypeScript-land identifies it as a type checker. There are other places (either in the examples or somewhere else in the docs) where we're using `type-check` so also in the interest of staying consistent.

anthonyshew (changes requested):

anthonyshew:
Something about the formatting of the second `<Tabs />` is breaking the MDX parse, annoyingly.

docs/pages/repo/docs/getting-started/add-to-project.mdx
mehulkar: done

mehulkar:
> Something about the formatting of the second `<Tabs />` is breaking the MDX parse, annoyingly.

prettier is reformatting it badly when i commit 😢 , I got rid of all the whitespace indenting changes. Downside is that we have 5 `<ol>` elements on the page instead of 1 with 5 `<li>` elements... but I'll leave that for a future pass

PR #6598, opened 2023-11-27
chore(ci): remove limited timeout for test jobs

mehulkar:
Windows tests are regularly exceeding 30 mins. We can monitor these instead of setting a hard limit that isn't enough time for the actual test suite.

Closes TURBO-1766

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Test
+qname: Test
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -1,4 +1,4 @@
-name: Test
+qname: Test
```
mehulkar, line 1:
```suggestion
name: Test
```

PR #6601, opened 2023-11-27
chore: move rust-doc site into a workspace

mehulkar:
Closes TURBO-1769


mehulkar:
Closing this as @tknickman and @NicholasLYang both don't like this approach

PR #6603, opened 2023-11-27
fix: ensure cache is done writing before exit

chris-olszewski:
### Description

This PR contains intertwined changes:
 - Waiting on all cache writes to finish before we exit the run. We leverage the signal handler here so we can make sure we wait regardless of how/why the run exits.
 - Printing errors from artifact upload failures. This differs from Go where we [throw generated errors away](https://github.com/vercel/turbo/blob/main/cli/internal/cache/async_cache.go#L79C12-L79C12) (I'm not sure this is desired since we do [format these errors for human consumption](https://github.com/vercel/turbo/blob/main/cli/internal/client/cache.go#L70))
 - Correctly deserialize Vercel API errors. As per the [docs](https://vercel.com/docs/rest-api/errors) there's a containing object with an `error` field that has the information.
 - Pass team information when saving artifacts. This matches all of the other artifact api calls.

Each commit is reviewable on it's own.

### Testing Instructions

The `turbo` cli is a great candidate for these. Make sure you're logged in and linked to Vercel team.
For testing that we wait until upload finishes:
```
# Ensure there's no error printed at the end of the run and the "Finishing writing to cache" message appears while we upload the large debug binary
[0 olszewski@chriss-mbp] /Users/olszewski/code/vercel/turborepo $ turbo_dev build --filter=cli --force
...
    ...Finishing writing to cache...  
# Should be a FULL TURBO due to the previous command
[0 olszewski@chriss-mbp] /Users/olszewski/code/vercel/turborepo $ turbo_dev build --filter=cli --remote-only --remote-cache-timeout=120
```

Closes TURBO-1770

### DIFF
--- a/crates/turborepo-cache/src/async_cache.rs
+++ b/crates/turborepo-cache/src/async_cache.rs
@@ -5,6 +5,7 @@ use tokio::{
     sync::{mpsc, Semaphore},
     task::JoinHandle,
 };
+use tracing::error;
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf, AnchoredSystemPathBuf};
 use turborepo_analytics::AnalyticsSender;
 use turborepo_api_client::{APIAuth, APIClient};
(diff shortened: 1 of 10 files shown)


crates/turborepo-cache/src/async_cache.rs
```diff
@@ -64,12 +64,14 @@ impl AsyncCache {
                         let permit = semaphore.clone().acquire_owned().await.unwrap();
                         let real_cache = real_cache.clone();
                         workers.push(tokio::spawn(async move {
-                            let _ = real_cache.put(&anchor, &key, &files, duration).await;
+                            if let Err(err) = real_cache.put(&anchor, &key, &files, duration).await
+                            {
+                                warn!("{err}");
```
gsoltis, line 69: I'm not actually sure we want this. If the failures are correlated, it's likely that you'll get a flood of them. Maybe we want a limiter? after 5 or so, stop printing them?

gsoltis (approved):
Fixship: some mechanism for limiting the # of errors displayed from cache errors

PR #6609, opened 2023-11-28
fix: look for unix based project paths

chris-olszewski:
### Description

The lockfile library use unix style anchored paths to identify workspaces. When looking up the previous lockfile closures we were using the system path instead of the expected unix path.

Future work once we no longer need to support FFI is to make this map have stronger typed keys to avoid using the wrong type of string.

### Testing Instructions

Rust integration tests `lockfile-aware-caching/*.t` should pass now. [Run](https://github.com/vercel/turbo/actions/runs/7012611302/job/19077872563?pr=6609#step:7:956)

Closes TURBO-1776


mehulkar (commented):
The find-turbo tests are failing in the Rust codepsth on this PR. I think they were all passing on main last I checked but hard to tell. Can you confirm it’s not due to this PR? (linked.t is expected to fail, but not the others)

chris-olszewski:
> The find-turbo tests are failing in the Rust codepsth on this PR. I think they were all passing on main last I checked but hard to tell. Can you confirm it’s not due to this PR? (linked.t is expected to fail, but not the others)

Those failures are also happening on main: https://github.com/vercel/turbo/actions/runs/7020699343/job/19101823487#step:7:798

The codepath in this PR shouldn't be triggered by the find turbo tests in anyway as it only gets hit if using a SCM based filter.

mehulkar:
ah ok, sounds like a regression happened somewhere, I'll track it down

mehulkar:
Fixed that regression btw, this should be good to merge after a rebase/merge with man

mehulkar (approved):

PR #6612, opened 2023-11-28
feat(docs): add FAQ entry for non-JS workspaces in the codebase

mehulkar:
Based on two discussions:

- https://github.com/vercel/turbo/discussions/1077
- https://github.com/vercel/turbo/discussions/1900

Closes TURBO-1778

### DIFF
--- a/docs/pages/repo/docs/faq.mdx
+++ b/docs/pages/repo/docs/faq.mdx
@@ -15,7 +15,10 @@ You have two options when working with Turborepo:
 1. Install it globally, via `npm install --global turbo`
 2. Install a local version in your project
 
-We recommend installing the `turbo` CLI globally. This gives you a smooth, ergonomic experience for running tasks.
+We recommend installing the `turbo` CLI globally. This gives you a smooth,
+ergonomic experience for running tasks. If your project _also_ has `turbo` as a
+dependency in package.json, the global `turbo` will invoke the local one to
+ensure that intended version is used in the project.
 
 ### Why isn't my global `turbo` working as expected?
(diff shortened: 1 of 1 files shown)


docs/pages/repo/docs/faq.mdx
```diff
@@ -86,3 +89,18 @@ Some new features are marked as "experimental" in Turborepo. This means that the
 
 - [Report a bug with an experimental feature](https://github.com/vercel/turbo/issues/new?labels=experimental,kind:+bug,area:+turborepo,needs:+triage&template=0-turborepo-bug-report.yml&title=%5Bturborepo%5D+)
 - [Provide feedback on an experimental feature](https://github.com/vercel/turbo/issues/new?labels=experimental,story,needs:+triage&template=2-feature-request.yml&title=%5Bturborepo%5D+)
+
+## Can I use Turborepo with a monorepo that non-JS code?
+
+Your monorepo can include subdirectories with any kind of code (a Django app,
+Rust crates, Ruby CLIs, etc). Turborepo will largely ignore these directories.
```
anthonyshew, line 96: Got a little verbose here. I'd flip this explanation on its head. Instead of saying what it *doesn't* do, get right into what it *does* do. Along the lines of "Turbo is built on the conventions of JavaScript meaning you need to fulfill two requirements. Subdirectories need to be workspaces for your JS package manager and have a package manager with the scripts you want to run."

Not going to block because I'm just nitpicking and having something is better than nothing, of course!

anthonyshew (approved):

PR #6615, opened 2023-11-28
chore(tests): remove hardcoded packageManager when running examples tests

mehulkar:
examples now have this field already set, so tests can just validate this

Closes TURBO-1780


anthonyshew (approved):

mehulkar:
Merged while pairing with @anthonyshew since it's blocking a bunch of examples PRs. We feel pretty confident that this is the right thing to do, and examples tests are non-deterministically failing anyway, so the feedback from this PR's tests wouldn't be that useful.

PR #6616, opened 2023-11-28
feat(turbopack): support "loading" WebAssembly injected as global varables in edge runtime

ForsakenHarmony:
### Description

Next.js supports injecting WASM files into the edge runtime via the middleware manifest, so we want to be able to support that usecase as well

Closes PACK-2027

### DIFF
--- a/crates/turbopack-ecmascript-runtime/js/src/dev/runtime/none/runtime-backend-none.ts
+++ b/crates/turbopack-ecmascript-runtime/js/src/dev/runtime/none/runtime-backend-none.ts
@@ -35,18 +35,51 @@ function augmentContext(context: TurbopackDevBaseContext): TurbopackDevContext {
 }
 
 async function loadWebAssembly(
-  _source: SourceInfo,
-  _id: ModuleId,
-  _importsObj: any
+  source: SourceInfo,
+  chunkPath: ChunkPath,
+  imports: WebAssembly.Imports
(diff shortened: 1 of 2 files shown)


crates/turbopack-ecmascript-runtime/js/src/dev/runtime/none/runtime-backend-none.ts
```diff
@@ -35,18 +35,53 @@ function augmentContext(context: TurbopackDevBaseContext): TurbopackDevContext {
...
+): Promise<WebAssembly.Module> {
+  const stem = getFileStem(chunkPath);
+
+  // very simple escaping just replacing unsupported characters with `_`
+  const escaped = stem.replace(/[^a-zA-Z0-9$_]/gi, "_");
+
+  const identifier = `wasm_${escaped}`;
+
+  const module = (globalThis as GlobalWithInjectedWebAssembly)[
+    `wasm_${escaped}`
```
sokra, line 75:
```suggestion
    identifier
```

sokra (approved):

crates/turbopack-ecmascript-runtime/js/src/dev/runtime/none/runtime-backend-none.ts
ForsakenHarmony: good catch

PR #6620, opened 2023-11-28
docs: Added rustdoc comments

NicholasLYang:
### Testing Instructions

Closes TURBO-1784


crates/turborepo-cache/src/lib.rs
```diff
@@ -3,11 +3,19 @@
 #![feature(box_patterns)]
 #![deny(clippy::all)]
 
+/// A wrapper for the cache that uses a worker pool to perform cache operations
```
kdy1, line 6: nit: Maybe we may move these comments to each module

crates/turborepo-analytics/src/lib.rs
```diff
@@ -1,5 +1,10 @@
 #![deny(clippy::all)]
 
+//! Turborepo's analytics library. Handles sending analytics events to the
+//! Vercel API in the background. Currently we only record cache usage events,
+//! so when the cache is hit or missed for the file system or the HTTP cache.
+//! Requires the user to be logged in to Vercel.
```
mehulkar, line 6:
```suggestion
//! HTTP cache usage requires the user to be logged in to Vercel.
```

crates/turborepo-analytics/src/lib.rs
```diff
@@ -1,5 +1,10 @@
 #![deny(clippy::all)]
 
+//! Turborepo's analytics library. Handles sending analytics events to the
+//! Vercel API in the background. Currently we only record cache usage events,
```
mehulkar, line 4: > Currently we only...

I'd recommend not writing point-in-time docs. They are out of date the minute they ship because "currently" doesn't mean anything when you're reading the doc.

crates/turborepo-auth/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 #![feature(cow_is_borrowed)]
 #![deny(clippy::all)]
+//! Turborepo's library for authenticating with the Vercel API.
+//! Handles logging into Vercel, verifying SSO, and storing the Vercel token.
```
mehulkar, line 4:
```suggestion
//! Handles logging into Vercel, verifying SSO, and storing the token.
```

Let's not say Vercel since it could be a custom remote cache

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -295,6 +295,7 @@ impl<'a> Run<'a> {
 
         let is_monorepo = !opts.run_opts.single_package;
 
+        println!("{:#?}", root_workspace.transitive_dependencies);
```
mehulkar, line 298: Should this be here?

crates/turborepo-paths/src/lib.rs
```diff
@@ -2,33 +2,43 @@
...
+//! It has a backslash, which is simultaneously a valid character in a
+//! Unix file name and the Windows path delimiter.
+//!
+//! Therefore, it's very important to keep the context of the initialization
+//! in mind. The only case where we do more in depth validation is for cache
+//! restoration. See `AnchoredSystemPathBuf::from_system_path` for more details.
+//!
+//! The only case where initializing a path type without validation is
+//! recommended is inside turbopath itself. But that unchecked initialization
+//! should be considered unsafe
```
mehulkar, line 41:
```suggestion
//! recommended is inside turbopath itself, but that unchecked initialization
//! should be considered unsafe.
```

crates/turborepo-scm/src/lib.rs
```diff
@@ -3,6 +3,10 @@
 #![feature(assert_matches)]
 #![deny(clippy::all)]
 
+//! Turborepo's library for interacting with source control management (SCM).
+//! Currently we only support git. We use SCM for finding changed files,
+//! for getting the previous version of a lockfile, and to hash files.
```
mehulkar, line 8:
```suggestion
//! Turborepo's library for interacting with source control management (SCM).
//! We use SCM for finding changed files, getting the previous version of a lockfile,
//! and to hash files.
```

mehulkar (approved):
Nothing super blocking, except the Vercel reference

crates/turborepo-analytics/src/lib.rs
NicholasLYang: We actually require the user to be logged in for any analytics to be sent

crates/turborepo-paths/src/lib.rs
NicholasLYang: For dramatic effect :P

crates/turborepo-analytics/src/lib.rs
mehulkar: ah ok, we're recording local cache usage stats also?

crates/turborepo-paths/src/lib.rs
mehulkar: tenor

crates/turborepo-scm/src/lib.rs
mehulkar: I read it more as a list. "blah blah blah for: first thing, second thing, and last thing"

but not blocking, grammar is fun :)

PR #6621, opened 2023-11-28
feat(Turborepo): Use package terminology (and do version bumps)

gsoltis:
### Description

 - Use package terminology instead of workspace
 - bump versions in preparation for a publish
 - script version bump (not wired to the publish step yet)

### Testing Instructions

 - add explicit imports in the tests for everything we expect to be exporting
 - add a very mild test for package manager detection

Closes TURBO-1785


packages/turbo-repository/README.md
```diff
@@ -10,3 +10,7 @@ The `js/` folder contains the meta package to handle importing platform-specific
 for the JS API.
 
 This package contains scripts to build dev and release versions. `pnpm build && pnpm package` will build and package a dev version of the native library for `darwin-arm64`, or you can pass an additional argument for a specific target. `pnpm build:release` will build a release version of the library
+
+# Publishing
+
+There is now a version bump script in [bump-version.sh](./scripts/bump-version.sh). Passing it the new version will bump the metapackage version, as well as the optional dependencies list and native packages.
```
mehulkar, line 16: consistent with above

```suggestion
There is now a version bump script in [bump-version.sh](./scripts/bump-version.sh). Passing it the new version will bump the meta package version, as well as the optional dependencies list and native packages.
```

packages/turbo-repository/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@turbo/repository",
-  "version": "0.0.1-canary.1",
+  "version": "0.0.1-canary.2",
```
mehulkar, line 3: should this be happening in a PR like this or through the release workflow. ok for a greenfield thing, but generally the release cut should be separate

packages/turbo-repository/scripts/bump-version.sh
```diff
@@ -0,0 +1,27 @@
...
+set -e
+
+if [ $# != 1 ]; then
+  echo "Missing version"
+  exit 1
+fi
+TARGET=$1
+
+SCRIPTS_DIR="$( dirname -- "${BASH_SOURCE[0]}" )"
+PKG_ROOT=$(realpath "${SCRIPTS_DIR}/..")
```
mehulkar, line 12: how come you used dirname in one line and `/..` in another? could just do `/../..`

mehulkar (approved):

packages/turbo-repository/js/package.json
gsoltis: I'm slowly ramping up the automation for this. The previous two releases were manual version bumps, this one was at least automated file changes. Next step is to wire it to GH actions.

packages/turbo-repository/scripts/bump-version.sh
gsoltis: Heh, the first line was copied from a "get directory of script" SO answer.

I do think there's some slight value in keeping it two lines, which shows the steps for the calculation, but I'll update the second line to use `dirname` instead of `realpath` for some consistency.

gsoltis: Ok, ended up moving both of them to `realpath` instead, since it turns out `BASH_SOURCE[0]` is a _relative_ path. I left a note to that effect as well.

PR #6632, opened 2023-11-29
Drop nonsensical(?) tests?

anthonyshew:
From talking to @mehulkar it sounds like these tests are debt?


mehulkar (approved):
Yeah I don't think we need to validate that this works, thanks!!

PR #6634, opened 2023-11-29
chore(ci): require Rust codepath tests in PR

mehulkar:
Also enables macOS variant of Rust codepath tests


mehulkar:
Seeing a test failure on WIndows that might be legit, but I'm confident that _this_ PR is not the one causing it to fail, so I'm going to merge this. Merging this will also prevent other regressions from PRs, so it's a net win. 

Also not going to wait for Turbopack Rust tests to pass on this PR, that's the only 🟡 ones still going.

PR #6641, opened 2023-11-29
chore: consolidate setup scripts for integration tests

mehulkar:
- Combines integration setup scripts to a single script and moves it into the shared helpers directory

--- 
**Note**
I had a more extensive refactor of these scripts (moved to #6670) for the purpose of making examples and integration tests work the same, but one test was failing and I wasn't sure why, so I need to boil down the diff. The eventual goal is to be able to easily use these setup scripts to test a wider variety of things:

- Against OSS repos (a github workflow that users can put into their own repo and run on a daily cadence, for example)
- a published release of turbo instead of building from HEAD
- an installed turbo in a repo instead of a global one (swapping out the `TURBO` env var easily
- running the same tests/fixtures with different package managers
- Full run through with `create-turbo`

Closes TURBO-1806


turborepo-tests/integration/tests/run/single-package/run-yarn.t
```diff
@@ -1,17 +1,11 @@
...
-  $ if [[ "$OSTYPE" == "msys" ]]; then dos2unix --quiet yarn.lock; fi
-  $ git commit --quiet -am "Update lockfile" # clean git state
+  $ . ${TESTDIR}/../../../../helpers/setup_integration_test.sh $(pwd) single_package "yarn@1.22.17"
 
 Check
   $ ${TURBO} run build
   \xe2\x80\xa2 Running build (esc)
   \xe2\x80\xa2 Remote caching disabled (esc)
-  build: cache miss, executing 551563378a0c0a7e
+  build: cache miss, executing c1d6153f4a6d83b5
```
mehulkar, line 8: This changed because with the `install_deps.sh` script has an additional feature where it sets a custom cache dir and adds it to a local gitignore file. The change to gitignore updated the task hash

```
yarn install --cache-folder="$PWD/.yarn-cache" > /dev/null 2>&1
echo ".yarn-cache" >> .gitignore
```

mehulkar:
Go Unit Tests are failing, but that is not related to this PR. They are failing because `with-yarn` and `basic` examples changed and they're used in these particular tests. Go unit tests probably didn't run when those changed, because of how our tests are setup

PR #6643, opened 2023-11-29
Updating pnpm flags in example tests.

anthonyshew:
We're no longer locking to pnpm@6 in our example tests. Instead they are using, the version in the example itself.

Our examples are on pnpm@8. It looks like the pass-through flagging for `pnpm run` has changed.


mehulkar (approved):

anthonyshew:
Continuing to move through changes to our examples tests. Getting a lot more passes now!

Merging past failing `yarn` tests now...

PR #6645, opened 2023-11-29
Update example test glob.

anthonyshew:
Updating example test glob since we refactored.


mehulkar (approved):

anthonyshew:
Failures on the examples' tests are expected as we continue to work through refactoring them back to green.

Merging through the ❌

PR #6646, opened 2023-11-29
fix(turbo-tasks-fs): support unicode segment for glob

kwonoj:
Closes PACK-2060


kwonoj:
next.js CI to confirm behavior: https://github.com/vercel/next.js/pull/59128

wbinnssmith (approved):

PR #6647, opened 2023-11-29
yarn-with-yarn workspace name

anthonyshew:
Just a naming typo!


anthonyshew:
`yarn` tests are not expected to pass as we continue to refactor. Next step will be to get these tests to green.

mehulkar (approved):
oh man, nice catch, guess we haven't been running this https://github.com/vercel/turbo/blob/da55848452e1c0edf8bff1a084414db181dca9cb/.github/workflows/test.yml#L458

PR #6649, opened 2023-11-30
fix: clean paths and use package path instead of package.json path

chris-olszewski:
### Description

Fix two major issues with our glob based filtering:
 - We weren't cleaning our paths. This would result is us trying match `./packages/*` against paths of the form `packages/foo/package.json`
 - We were using the path to `package.json` in many places where we were expecting the path to the package directory. This caused failures for single glob as `packages/*` doesn't match `packages/foo/package.json`, but does match `packages/foo`.

Future work:
We should decide if we want to model Go's behavior of cleaning paths when performing operations such as `join` and move these calls into our paths library.

### Testing Instructions

Updated existing unit tests with a proper fixture and fixed up some bad expected results.

Added some more filter integration tests that use globs and cwd.

Closes TURBO-1809

### DIFF
--- a/cli/internal/scope/filter/parse_target_selector_test.go
+++ b/cli/internal/scope/filter/parse_target_selector_test.go
@@ -110,6 +110,19 @@ func TestParseTargetSelector(t *testing.T) {
 			false,
 		},
 		{
+			"./foo/*",
+			&TargetSelector{
+				fromRef:             "",
+				exclude:             false,
+				excludeSelf:         false,
+				includeDependencies: false,
+				includeDependents:   false,
+				namePattern:         "",
+				parentDir:           turbopath.MakeRelativeSystemPath("foo/*"),
+			},
+			false,
+		},
+		{
 			"../foo",
 			&TargetSelector{
 				fromRef:             "",
--- a/crates/turborepo-lib/Cargo.toml
+++ b/crates/turborepo-lib/Cargo.toml
@@ -60,6 +60,7 @@ jsonc-parser = { version = "0.21.0" }
 lazy_static = { workspace = true }
 libc = "0.2.140"
 notify = "5.1"
+path-clean = "1.0.1"
 petgraph = { workspace = true }
 pidlock = { path = "../turborepo-pidlock" }
 prost = "0.11.6"
--- a/crates/turborepo-lib/src/run/scope/filter.rs
+++ b/crates/turborepo-lib/src/run/scope/filter.rs
@@ -38,7 +38,7 @@ impl PackageInference {
         );
         let full_inference_path = turbo_root.resolve(pkg_inference_path);
         for (workspace_name, workspace_entry) in pkg_graph.workspaces() {
-            let pkg_path = turbo_root.resolve(workspace_entry.package_json_path());
+            let pkg_path = turbo_root.resolve(workspace_entry.package_path());
             let inferred_path_is_below = pkg_path.contains(&full_inference_path);
             // We skip over the root package as the inferred path will always be below it
             if inferred_path_is_below && (&pkg_path as &AbsoluteSystemPath) != turbo_root {
@@ -47,7 +47,7 @@ impl PackageInference {
                 // do so in a consistent manner
                 return Self {
                     package_name: Some(workspace_name.to_string()),
-                    directory_root: workspace_entry.package_json_path().to_owned(),
+                    directory_root: workspace_entry.package_path().to_owned(),
                 };
             }
             let inferred_path_is_between_root_and_pkg = full_inference_path.contains(&pkg_path);
@@ -343,7 +343,7 @@ impl<'a, T: PackageChangeDetector> FilterResolver<'a, T> {
             } else {
                 let path = selector.parent_dir.to_unix();
                 let parent_dir_matcher = wax::Glob::new(path.as_str())?;
-                let matches = parent_dir_matcher.is_match(info.package_json_path.as_path());
+                let matches = parent_dir_matcher.is_match(info.package_path().as_path());
 
                 if matches {
                     entry_packages.insert(name.to_owned());
@@ -413,7 +413,7 @@ impl<'a, T: PackageChangeDetector> FilterResolver<'a, T> {
             let package_path_lookup = self
                 .pkg_graph
                 .workspaces()
-                .map(|(name, entry)| (name, entry.package_json_path()))
+                .map(|(name, entry)| (name, entry.package_path()))
                 .collect::<HashMap<_, _>>();
 
             for package in changed_packages {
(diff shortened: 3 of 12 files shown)


cli/internal/scope/filter/parse_target_selector_test.go
```diff
@@ -118,7 +129,7 @@ func TestParseTargetSelector(t *testing.T) {
 				includeDependencies: false,
 				includeDependents:   false,
 				namePattern:         "",
-				parentDir:           turbopath.MakeRelativeSystemPath("..", "foo"),
+				parentDir:           "../foo",
```
chris-olszewski, line 132: Making this explicit to avoid obscuring any surprises from `filepath.Clean`

crates/turborepo-lib/Cargo.toml
```diff
@@ -60,6 +60,7 @@ jsonc-parser = { version = "0.21.0" }
 lazy_static = { workspace = true }
 libc = "0.2.140"
 notify = "5.1"
+path-clean = "1.0.1"
```
chris-olszewski, line 63: This library was intentionally written to match the behavior of Go's `filepath.Clean`. I did a spot check and it does seem to be a true to form port, but more eyes are always welcome.
[filepath.Clean](https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/path/filepath/path.go;l=98)
[path_clean::clean](https://github.com/danreeves/path-clean/blob/master/src/lib.rs#L58)

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -38,7 +38,7 @@ impl PackageInference {
         );
         let full_inference_path = turbo_root.resolve(pkg_inference_path);
         for (workspace_name, workspace_entry) in pkg_graph.workspaces() {
-            let pkg_path = turbo_root.resolve(workspace_entry.package_json_path());
+            let pkg_path = turbo_root.resolve(workspace_entry.package_path());
```
chris-olszewski, line 41: Previously we would be checking if `/path/to/repo/packages/foo/package.json` contained a directory which would always be false.

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -641,13 +641,16 @@ mod test {
...
-        let package_jsons = packages
+        let package_jsons = package_dirs
             .iter()
             .map(|package_path| {
                 let (_, name) = get_name(package_path);
                 (
                     turbo_root
-                        .join_unix_path(RelativeUnixPathBuf::new(**package_path).unwrap())
+                        .join_unix_path(
+                            RelativeUnixPathBuf::new(format!("{package_path}/package.json"))
```
chris-olszewski, line 651: The keys to the package json map for the package graph builder should be paths to `package.json` files as this is what will get returned by a the `PackageManager::get_package_jsons`.

This caused the test fixture to cover up the logic errors in the implementation.

crates/turborepo-lib/src/run/scope/target_selector.rs
```diff
@@ -195,8 +195,12 @@ pub fn is_selector_by_location(
             .iter()
             .any(|prefix| raw_selector.starts_with(prefix))
     {
+        let cleaned_selector = path_clean::clean(std::path::Path::new(raw_selector))
```
chris-olszewski, line 198: We do this to match the Go behavior where `turbopath.MakeRelativeSystemPath` called `filepath.Clean` under the hood which would remove any leading `./`.

crates/turborepo-lib/src/run/scope/target_selector.rs
```diff
@@ -219,8 +223,9 @@ mod test {
     #[test_case("...foo...", TargetSelector { name_pattern: "foo".to_string(), raw: "...foo...".to_string(), include_dependents: true, include_dependencies: true, ..Default::default() }; "dot dot dot foo dot dot dot")]
     #[test_case("foo^...", TargetSelector { name_pattern: "foo".to_string(), raw: "foo^...".to_string(), include_dependencies: true, exclude_self: true, ..Default::default() }; "foo caret dot dot dot")]
     #[test_case("...^foo", TargetSelector { name_pattern: "foo".to_string(), raw: "...^foo".to_string(), include_dependents: true, exclude_self: true, ..Default::default() }; "dot dot dot caret foo")]
-    #[test_case("./foo", TargetSelector { raw: "./foo".to_string(), parent_dir: AnchoredSystemPathBuf::try_from("./foo").unwrap(), ..Default::default() }; "dot slash foo")]
     #[test_case("../foo", TargetSelector { raw: "../foo".to_string(), parent_dir: AnchoredSystemPathBuf::try_from("../foo").unwrap(), ..Default::default() }; "dot dot slash foo")]
+    #[test_case("./foo", TargetSelector { raw: "./foo".to_string(), parent_dir: AnchoredSystemPathBuf::try_from("foo").unwrap(), ..Default::default() }; "dot slash foo")]
```
chris-olszewski, line 227: `parent_dir: "foo"` now matches the expected value of the [corresponding Go unit test](https://github.com/vercel/turbo/blob/main/cli/internal/scope/filter/parse_target_selector_test.go#L108)

crates/turborepo-lib/src/run/scope/target_selector.rs
```diff
@@ -195,8 +195,12 @@ pub fn is_selector_by_location(
             .iter()
             .any(|prefix| raw_selector.starts_with(prefix))
     {
+        let cleaned_selector = path_clean::clean(std::path::Path::new(raw_selector))
+            .into_os_string()
+            .into_string()
```
chris-olszewski, line 200: We could use `AnchoredSystemPathBuf::from_system_path` to avoid these conversion, but that codepath will failif the path starts with `../` which is a valid filter.

chris-olszewski (commented):
Reviewer's Guide

arlyon (approved):
Ouch that is likely my fault. Approving with some comments.

crates/turborepo-lib/src/run/scope/target_selector.rs
arlyon: Maybe we should clean all AnchoredSystemPathBuf and RelativeUnixPaths such that there are as few ways to represent the same path as possible.

NicholasLYang (commented):
So far makes sense, but gonna do another pass

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -75,7 +75,14 @@ impl PackageInference {
         }
 
         if selector.parent_dir != turbopath::AnchoredSystemPathBuf::default() {
-            selector.parent_dir = self.directory_root.join(&selector.parent_dir);
+            let repo_relative_parent_dir = self.directory_root.join(&selector.parent_dir);
+            let clean_parent_dir =
```
gsoltis, line 79: Definitely not blocking, but we ought to think about how to work this into the path library properly

gsoltis (approved):

PR #6665, opened 2023-11-30
Revert "chore: reduce sleep length in test cases (#6640)"

chris-olszewski:
This reverts commit 4b07f76c19ff607bed0e85aa0c7cbc2a36d67620.

### Description

I think we're flying a little too close to the sun with this one: https://github.com/vercel/turbo/actions/runs/7052873305/job/19199324958?pr=6649#step:7:721

I think the extra <1s when running this test is worth reducing the flakiness.

### Testing Instructions

👀 

Closes TURBO-1821


mehulkar (approved):
You owe me half a second every time I wait for this test @chris-olszewski. I'll keep a tally

PR #6666, opened 2023-11-30
Turbopack css: treat `composes` and `@import` as urls when resolving

wbinnssmith:
This allows for bare specifiers to be resolved first as relative, then using a node-style resolution

Test Plan: Added snapshot tests.

Closes PACK-2064


kwonoj (approved):

kdy1 (approved):

crates/turbopack-css/src/references/import.rs
```diff
@@ -345,11 +345,12 @@ impl ImportAssetReference {
...
     fn resolve_reference(&self) -> Vc<ModuleResolveResult> {
-        css_resolve(
+        url_resolve(
             self.origin,
             self.request,
-            Value::new(CssReferenceSubType::AtImport),
+            Value::new(UrlReferenceSubType::CssAtImport),
             Some(self.issue_source),
+            IssueSeverity::Error.cell(),
         )
```
sokra, line 354: I wonder that this works... Doesn't this reference the imported CSS without processing it as CSS?

crates/turbopack-tests/tests/snapshot/css/relative-uri-import/input/other.css
```diff
@@ -0,0 +1,3 @@
+.foo {
```
sokra, line 1: Could you try to add another `@import` here? Is that processed?

crates/turbopack-core/src/reference_type.rs
```diff
@@ -60,6 +58,8 @@ pub enum CssReferenceSubType {
 #[derive(Debug, Clone, PartialOrd, Ord, Hash)]
 pub enum UrlReferenceSubType {
     EcmaScriptNewUrl,
+    CssAtImport,
```
sokra, line 61: I'm not a fan of this being a `URL` reference type. `@import` and `composes` should be CSS reference types.

Could we just move the common logic from `url_resolve` instead a shared method between `url_resolve` and `css_resolve` so that we are not forced to use an `URL` reference type.

(Note: There are some special rules applied to `URL` reference types, which shouldn't affect CSS references)

crates/turbopack-tests/tests/snapshot/css/relative-uri-import/output/crates_turbopack-tests_tests_snapshot_css_relative-uri-import_input_99b643._.css
```diff
@@ -0,0 +1,38 @@
+/* [project]/crates/turbopack-tests/tests/snapshot/css/relative-uri-import/input/another.css [test] (css) */
+.bar {
```
sokra, line 2: Why is that 3 times in that chunk?

sokra (approved):

PR #6678, opened 2023-12-02
chore(tests): add a new setup_git script for examples that doesn't install deps

mehulkar:
setting up git shouldn't install dependencies, we should do that afterwards and use the appropriate package manager. This change only affects the examples tests, we will use the same setup_git script in integration tests in a follow up change.

Closes TURBO-1825


anthonyshew (approved):
👏 👏 👏

mehulkar:
Merging. Rust and Go unit tests are both failing because `examples/` themselves changed. Go unit tests are fixed in https://github.com/vercel/turbo/pull/6679, Rust unit tests will be fixed separately also

PR #6681, opened 2023-12-03
chore(tests): setup package manager consistently for example/integration tests

mehulkar:


turborepo-tests/helpers/examples_setup.sh
```diff
@@ -6,10 +6,12 @@ set -eo pipefail
 
 exampleName=$1
 pkgManager=$2
+pkgManagerWithVersion=$3
```
anthonyshew, line 9: Seeing this $3 but not seeing it used anywhere?

anthonyshew (approved):

PR #6682, opened 2023-12-03
Mix of server and client components in basic example.

### LINKED ISSUE #6019: [turborepo] Examples are too simple: missing a real-world example with tree-shakeable, RSC-ready library
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

Hello,

I've skimmed through the turborepo examples and I haven't yet been able to get together a working example that could:
- have a UI library such as shadcn containing client components and server components without forcing all components to be client. Currently there are `tsup` workarounds to force-add "use client" to the whole bundle, but when half of your components are just dump components without interactivity, it's a waste.
- be tree shakeable: 
  - the examples with `tsup` won't tree-shake when you start using npm packages (ie you create a select component with react-select,  a dropzone component with react-dropzone, a simple <Hello /> component, you'll end up with the three packages in every page even if you only import the Hello component).
  - the transpilePackages won't tree shake the components: https://github.com/vercel/next.js/issues/55943 
- redirect to the appropriate file when doing a cmd+click: currently in the examples, "go to definition" will bring you to a `.d.ts`. That's not what devs expect when developing in a monorepo.
- put files in a `src/` folder and still work with nextjs transpiling

### Describe the solution you'd like

I'd like a kitchen-sink like example but more in-depth: its Link and CounterButton are currently client components due to the banner in the tsup.config.ts, but they could totally be server components. 

It would be great if it could support shadcn which is gaining much popularity now. It's also a great idea because shadcn/ui brings in a bit of complexity by installing third party packages. That would make it more "real", at least to me. Because with barrel exports, I've always struggled with bundle size!

### Describe alternatives you've considered

I've considered doing it myself but I couldn't make it.

So far I've been able to "fix" the "go-to definition problem" starting from the `with-tailwind` example, but couldn't do anything about the tree shaking.

I've tried many things from different github issues to support the "use client" at the file-level instead of the package-level, but I wasn't able to make it work. Even tried to create a with-tailwind package using rollup instead of tsup. I failed

Thanks!

anthonyshew:
Previously, the `basic` example (AKA create-turbo) was entirely Server Components. In this PR, we're continuing to leverage the Internal Package pattern such that we have a mix of React Server Components **and** Client components.

The Client Component is the new `Button` component in the `@repo/ui` package.

Notably, this PR closes #6019, fulfilling these requirements:
- A mix of Server Components and Client Components
- Editor go-to-definition across packages (already done in previous work)
- Using an `src` folder in the package but not having to use it when importing into another package
- Tree-shaking (already done in previous work, check out `exports` in `./packages/ui.package.json`

Non-goals of this PR:
- Including the entirety of `shadcn-ui` (We're showing you the fundamentals of how to support `shadcn-ui` here but not bringing in the components)
- Using `tsup` to create an RSC package (Please open a new issue/add reactions to an existing issue if you'd like to see this)

### DIFF
--- a/examples/basic/apps/docs/app/page.module.css
+++ b/examples/basic/apps/docs/app/page.module.css
@@ -312,10 +312,7 @@
   color: #000000;
   cursor: pointer;
   display: inline-block;
-  font-family: "Haas Grot Text R Web", "Helvetica Neue", Helvetica, Arial,
-    sans-serif;
-  font-size: 14px;
-  font-weight: 500;
+  font-size: 16px;
   height: 40px;
   line-height: 20px;
   list-style: none;
--- a/examples/basic/apps/docs/app/page.module.css
+++ b/examples/basic/apps/docs/app/page.module.css
@@ -214,6 +214,7 @@
   position: absolute;
   min-width: 614px;
   min-height: 614px;
+  pointer-events: none;
 }
 
 .logo {
@@ -275,6 +276,7 @@
   position: absolute;
   mix-blend-mode: normal;
   will-change: filter;
+  pointer-events: none;
 }
 
 .gradientSmall {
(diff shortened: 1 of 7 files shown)


examples/basic/apps/docs/app/page.module.css
```diff
@@ -214,6 +214,7 @@
   position: absolute;
   min-width: 614px;
   min-height: 614px;
+  pointer-events: none;
```
anthonyshew, line 217: Anywhere you're seeing me do `pointer-events: none` is me working around that the background imagery (the gradient image, the circles image) is sitting on top of nearly the entire page, making it so you can't click the elements underneath.

I could do the work to actually fix this but it's more important to me that we have a great example of RSCs rather than a perfect HTML template (that 99% of folks are deleting anyway).

chris-olszewski (approved):
Thank you so much for this! We often get issues that are really questions about how to set up a project with a mix of server and client components.

PR #6685, opened 2023-12-04
chore(tests): delete feature where @canary turbo is used

mehulkar:
We are now building turbo from source for examples tests, so there is no need to set the package.json version in each example, it does not matter.

Closes TURBO-1832


chris-olszewski (approved):
🐶 🥫

PR #6653, opened 2023-11-30
feat(docs): document phantom nodes

mehulkar:

PR #6688, opened 2023-12-04
Example with only shell commands.

anthonyshew:
An example that only uses shell commands. No frameworks, no code...just tasks.

This could be useful for debugging, bug reproductions, and learning purposes.

Open questions:
- What scripts would be most useful to have in the root package.json?
- Are there any other techniques we would like to show off in this? Common patterns where the behavior is corrrect but folks don't realize it?

Anthony's TODO zone:
- "Transit Node" verbiage is a placeholder for what we eventually decide in #6653


examples/with-shell-commands/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Using this example
+
+Run the following command:
+
+```sh
+npx create-turbo@latest -e with-shell-commands
+```
+
+### For bug reproductions
```
tknickman, line 13: This should be linked to or added to the Turborepo bug template!

anthonyshew: Didn't know how much we loved our bug issue template but there's a couple additions (this being one of them) that I'd like to propose. 😄 

Just wanted to make sure the rest of the group thinks it's a good idea first.

examples/with-shell-commands/.gitignore
```diff
@@ -0,0 +1,24 @@
...
+
+
+# Debug
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# Misc
+.DS_Store
+*.pem
```
mehulkar, line 24: We can drop `.pem` probably.

examples/with-shell-commands/README.md
```diff
@@ -0,0 +1,46 @@
...
+- `tooling-config`: A package to simulate a common configuration used for all of your repository. This could resemble a configuration for tools like TypeScript or ESLint that are installed into all of your packages.
+
+### Pre-built scripts
+
+In the root `package.json`, we've built some scripts to demonstrate many of Turborepo's features. We've named these scripts after common monorepo tasks to make the example more tangible.
+
+- `pnpm all-tasks`: Runs all tasks in your graph.
+- `pnpm build`: A basic command to build `app-a` and `app-b` in parallel.
+- `pnpm build:*`: Variations of `turbo run build` to demonstrate filtering and flag usage.
+- `pnpm lint`: A basic linting command that
```
mehulkar, line 46: I think we should encourage global turbo install, and remove `pnpm` here

examples/with-shell-commands/apps/apps-a/README.md
```diff
@@ -0,0 +1,3 @@
+# Top-level package
+
+We'll use this package as the end of your Task Graph. You may want to think of it as the final application you're building (the most common use case for a Turborepo build pipeline).
```
mehulkar, line 3: "end of your Task Graph"

We should give this a better name. The opposite of "entrypoint". I would also like to the TaskGraph docs here (that we're getting together in the other PR. Perhaps:

```suggestion
We'll use this package as the final node of your [Task Graph](https://turbo.build/repo/docs/core-concepts/task-graph). You may want to think of it as the final application you're building (the most common use case for a Turborepo build pipeline).
```

examples/with-shell-commands/apps/apps-a/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "app-a",
+  "description": "A workspace to use as a model for a final application that the rest of your dependencies are building towards.",
```
mehulkar, line 3:
```suggestion
  "description": "An application that uses other internal packages",
```

examples/with-shell-commands/apps/apps-a/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "scripts": {
+    "build": "sleep 4 && mkdir -p dist && echo \"Your application output!\" > dist/app-output.txt && echo \"Application A is built!\"",
+    "lint": "sleep 4 && echo \"Linted!\"",
+    "type-check": "sleep 4 && echo \"Types checked!\""
```
mehulkar, line 7: Why do we need `sleep`? Does it demonstrate anything useful or just artificially makes it look like something's happening?

examples/with-shell-commands/apps/apps-a/package.json
```diff
@@ -0,0 +1,16 @@
...
+  },
+  "dependencies": {
+    "pkg-a": "workspace:*",
+    "with-workspace-configuration": "workspace:*"
+  },
+  "devDependencies": {
```
mehulkar, line 13: Is the distinction between `dependencies` and `devDependencies` even useful? I'm leaning towards no

examples/with-shell-commands/packages/with-workspace-configuration/turbo.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "build": {
+      "dependsOn": ["^build"],
+      "inputs": ["!README.md"],
```
mehulkar, line 7: As soon as you have an `inputs` config, all other files in the repo are excluded so you have to add them back manually. I think this is a dangerous example to give

mehulkar (commented):
I love this and have wanted it for a while!

Some feedback:
- I think this could be simpler (1 app and 2 packages). For me, the greatest benefit in this example is drawing a boundary between Turbo and specific frameworks. I don't think we need to demonstrate every configuration, etc. 
- No root tasks. We always recommended that before, but I think we should lean on global turbo installs instead (but I get it if you want to match the other examples)

examples/with-shell-commands/apps/apps-a/package.json
anthonyshew: I'll say yeah because it's meant to mimic a tooling configuration which would likely be a devDep. (`eslint`, `typescript`, etc.)

mehulkar: Every example probably doesn't need to show every feature, fwiw. This example is about removing frameworks, since turbo itself doesn't treat dep vs devDep differently, I don't see the value in adding it to this example. In fact, it's _reintroducing_ ideas about specific tools into the example (the very thing we're trying to remove!)

anthonyshew: Fair play, took it out.

examples/with-shell-commands/turbo.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "pipeline": {
+    "build": {
+      "dependsOn": ["prebuild", "^build"],
+      "outputs": ["output-file.txt", "dist/**"]
+    },
+    "prebuild": {},
+    "lint": {},
+    "type-check": {}
```
chris-olszewski, line 10: Maybe an addition of an `build-error` task (`echo 'build failed' && exit 1`) on a package? Something I find myself doing for testing a good amount.

anthonyshew: My gut is that that would make it more difficult for general use. Let's see if we end up with a need for that in the wild. 👍

PR #6691, opened 2023-12-04
fix: limit the amount of logs we send to spaces

chris-olszewski:
### Description

There's a 4.5 MB limit on serverless  requests and task logs can exceed that. 

We avoid this limit by trimming logs to be 4 MB and we take the final 4MB as that is probably contains the error. This isn't a perfect solution, but it does get us to a better state.

### Testing Instructions

Added unit tests for new log trimming function.

### DIFF
--- a/crates/turborepo-lib/src/run/summary/spaces.rs
+++ b/crates/turborepo-lib/src/run/summary/spaces.rs
@@ -24,6 +24,11 @@ use crate::{
     run::{summary::Error, task_id::TaskId},
 };
 
+// There's a 4.5 MB limit on serverless requests, we limit ourselves to a
+// conservative 4 MB to leave .5 MB for the other information in the response.
+// https://vercel.com/guides/how-to-bypass-vercel-body-size-limit-serverless-functions
+const LOG_SIZE_BYTE_LIMIT: usize = 4 * 1024 * 1024;
+
 pub struct SpacesClient {
     space_id: String,
     api_client: APIClient,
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/run/summary/spaces.rs
```diff
@@ -321,10 +326,33 @@ impl<'a> From<SpacesTaskInformation<'a>> for SpaceTaskSummary {
...
+    let lossy_logs = String::from_utf8_lossy(logs);
+    if lossy_logs.as_bytes().len() <= limit {
+        lossy_logs.into_owned()
+    } else {
+        // We try to trim down the logs so that it is valid utf8
+        // We attempt to parse it at every byte starting from the limit until we get a
+        // valid utf8 which means we aren't cutting in the middle of a cluster.
+        for start_index in (lossy_logs.as_bytes().len() - limit)..lossy_logs.as_bytes().len() {
+            let log_bytes = &lossy_logs.as_bytes()[start_index..];
+            if let Ok(log_str) = std::str::from_utf8(log_bytes) {
```
NicholasLYang, line 341: Is `from_utf8` O(n)? If so, this could be quite slow if the log file is quite large, say 2x the limit. Not really sure what the solution is, maybe a hard limit on log files that's 1.5x the spaces limit?

chris-olszewski: It is O(n) on the path where the string is valid utf8, but it early exits when if finds an invalid character ([source](https://doc.rust-lang.org/src/core/str/validations.rs.html#155)). 

If we don't split a character, then we break out of the loop and only perform this O(n) operation once. 
If we do split a character, then `from_utf8` should exit after checking the remaining bytes in the char (up to 3) before it realizes it's invalid and we'll continue to do this until we are past the remaining bytes in the split character.

NicholasLYang (approved):

PR #6692, opened 2023-12-04
chore(turborepo): add failing test for output-logs of cache:false tasks

### LINKED ISSUE #6677: [turborepo] The --output-logs errors-only flag is ignored for non-cached tasks.

1.10.16

pnpm

Linux

### Describe the Bug

When a task is marked as `cache: false` the flag `--output-logs errors-only` is not honored for the task, and its output is always printed.

### Expected Behavior

When I turbo run with  `--output-logs errors-only` then only failed commands are logged.

### To Reproduce

In the reproduction with the given repo:
At the root of the project:
```bash
pnpm install --filter=*
pnpm turbo run cached --output-logs errors-only # Works as expected
pnpm turbo run not-cached --output-logs errors-only # Reproduced the problem
```

### Reproduction Repo

https://github.com/Reductions/turbo-repro

mehulkar:
Fixes #6677

Closes TURBO-1836

### DIFF
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -156,6 +156,10 @@ impl TaskCache {
         let prefixed_writer = PrefixedWriter::new(self.run_cache.ui, prefix, writer);
 
         if self.caching_disabled || self.run_cache.writes_disabled {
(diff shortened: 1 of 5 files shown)


mehulkar:
I won't be able to follow through on this right now, closing.

PR #6693, opened 2023-12-04
fix(tests): update rust unit tests that rely on with-yarn fixture

chris-olszewski:
### Description

Same as #6679

### Testing Instructions

Unit tests should pass now

Closes TURBO-1839


mehulkar (approved):
ty!

Issue #6322, opened 2023-10-31
[turborepo] Yarn installing from npm registry rather than internal packages


achadee:
### What version of Turborepo are you using?

1.10.16

### What package manager are you using / does the bug impact?

Yarn v2/v3 (node_modules linker only)

### What operating system are you using?

Mac

### Describe the Bug

Hey guys, im following this tutorial to the tee
https://turbo.build/repo/docs/handbook/sharing-code/internal-packages

and yarn keeps trying to install this package:
https://www.npmjs.com/package/math-helpers

Any ideas whats happening?

Yarn v: 3.4.1
Node: 19.7.0
npm: 9.5.0

yarn lock
```
"math-helpers@npm:*":
  version: 0.1.0
  resolution: "math-helpers@npm:0.1.0"
  checksum: 9ce4bc05dc3b5268a73878f5d7afa83b5c17b6203f09200b3ffae459319c03f5e4772a6da7122a6a5319f3206d58dd607e617d2f19197894c28ef64ab773dfcd
  languageName: node
  linkType: hard
  ```

apps/web/packages.json
```
"dependencies": {
    ...
    "math-helpers": "*",
  }
```

### Expected Behavior

`yarn` should install the local internal packages when passing the `*` 

### To Reproduce

Follow the tutorial above with the versions?

**Comment by tknickman** [comment]
@achadee can you share your full `package.json` here along with the directory structure of your monorepo? I just want to verify that you have yarn workspaces configured correctly.

**Comment by achadee** [comment]
Hey Sorry guys, been on 5 days of flights, I'll get back to you with some more of my config in a day or 2

**Comment by achadee** [comment]
apps/web/package.json
```
{
  "name": "next-app",
  "version": "0.1.0",
  "private": true,
  "scripts": {
    "dev": "next dev",
    "build": "next build",
    "start": "next start",
    "lint": "next lint",
    "storybook": "storybook dev -p 6006",
    "build-storybook": "storybook build"
  },
  "dependencies": {
    "@expo/next-adapter": "5.0.2",
    "app": "*",
    "math-helpers": "*",
    "next": "13.4.19",
    "raf": "^3.4.1",
    "setimmediate": "^1.0.5"
  },
  "devDependencies": {
    "@storybook/addon-essentials": "^7.5.1",
    "@storybook/addon-interactions": "^7.5.1",
    "@storybook/addon-links": "^7.5.1",
    "@storybook/addon-onboarding": "^1.0.8",
[...]

PR #6696, opened 2023-12-05
Add yarn v2+ caveat for installing Internal Packages.

### LINKED ISSUE #5261: [turborepo] Documentation suggests "*" as package version when "workspace:*" is correct

N/A - docs issue

Linux

### Describe the Bug

When `"*"` is used as dependency version in package.json, yarn prefers to load it from the package registry instead of from the workspace. The solution is to use `"workspace:*"`, as described in [their documentation](https://yarnpkg.com/features/workspaces).

Turborepo documentation for monorepo uses `"*"`. This can lead to problems when yarn believes the package published to registry is newer than the local one. We ran into this issue where `x.y.5` > `x.y.6-foo.1`, causing builds to fail, with cryptic and unhelpful error messages.

Note that [pnpm also suggests using `"workspace:*"`](https://pnpm.io/workspaces#publishing-workspace-packages)

### Expected Behavior

Documentation should suggest using  `"workspace:*"` and/or link back to package manager's docs.

### To Reproduce

N/A -- docs issue

### Reproduction Repo

_No response_

anthonyshew:
yarn v1 uses `*`. yarn v2+ use `workspace:*`.

*audible sigh*

Closes #5261 and #6322

### DIFF
--- a/docs/pages/repo/docs/handbook/workspaces.mdx
+++ b/docs/pages/repo/docs/handbook/workspaces.mdx
@@ -122,7 +122,7 @@ For instance, if we want `apps/docs` to import `packages/shared-utils`, we'd nee
 ```json filename="apps/docs/package.json"
 {
   "dependencies": {
-    "shared-utils": "*" // If yarn v2/v3, use workspace:*
+    "shared-utils": "*" // If yarn v2+, use workspace:*
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/pages/repo/docs/handbook/workspaces.mdx
```diff
@@ -122,7 +122,7 @@ For instance, if we want `apps/docs` to import `packages/shared-utils`, we'd nee
 ```json filename="apps/docs/package.json"
 {
   "dependencies": {
-    "shared-utils": "*"
+    "shared-utils": "*" // If yarn v2/v3, use workspace:*
```
anthonyshew, line 125:
```suggestion
    "shared-utils": "*" // If yarn v2+, use workspace:*
```

mehulkar (approved):

PR #6701, opened 2023-12-05
add user level tracing

sokra:
### Description

Add a few user-focused info level spans that can be used to trace a normal build

Closes PACK-2082


crates/turbo-tasks-fs/src/retry.rs
```diff
@@ -23,18 +23,25 @@ where
...
 where
     F: Fn(&Path) -> io::Result<R> + Send + 'static,
     R: Send + 'static,
 {
     let path = path.as_ref().to_owned();
 
+    let current_span = tracing::Span::current();
     asyncify(move || {
+        let _entered = current_span.entered();
+        let _entered = span.entered();
```
ForsakenHarmony, line 40: why are we entering 2 spans here?

crates/turbopack-dev/src/chunking_context.rs
```diff
@@ -349,42 +355,50 @@ impl ChunkingContext for DevChunkingContext {
...
-        }
+            // Resolve assets
+            for asset in assets.iter_mut() {
+                *asset = asset.resolve().await?;
+            }
 
-        Ok(Vc::cell(assets))
+            Ok(Vc::cell(assets))
+        }
+        .instrument(span)
```
ForsakenHarmony, line 400: can't we "just" enter the span here?

ForsakenHarmony (approved):

PR #6705, opened 2023-12-05
chore(ci): keep crlf usage consistent on CI

chris-olszewski:
### Description

I noticed that for Windows only we're getting cache misses for `cli#build` on integration tests.

[Build turborepo](https://github.com/vercel/turbo/actions/runs/7105368951/job/19342475934?pr=6702#step:5:25): `cli#build` hash is `49dd89a5b438705a`
[Integration tests](https://github.com/vercel/turbo/actions/runs/7105368951/job/19343093111?pr=6702#step:7:29): `cli#build` hash is `95d2992a8cdf174b`

### Testing Instructions

On this PR the Windows integration tests should have a cache hit for `cli#build`

Closes TURBO-1851

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -208,6 +208,16 @@ jobs:
           - name: windows
             runner: windows-latest
     steps:
+      # On Windows, set autocrlf to input so that when the repo is cloned down
+      # the fixtures retain their line endings and don't get updated to CRLF.
+      # We want this because this repo also contains the fixtures for our test cases
+      # and these fixtures have files that need stable file hashes. If we let git update
+      # the line endings on checkout, the file hashes will change.
+      # https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf
+      - name: set crlf
+        if: matrix.os.name == 'windows'
+        shell: bash
+        run: git config --global core.autocrlf input
       - name: Checkout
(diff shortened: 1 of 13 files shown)


.github/workflows/test.yml
```diff
@@ -311,16 +311,6 @@ jobs:
...
-      # On Windows, set autocrlf to input so that when the repo is cloned down
-      # the fixtures retain their line endings and don't get updated to CRLF.
-      # We want this because this repo also contains the fixtures for our test cases
-      # and these fixtures have files that need stable file hashes. If we let git update
-      # the line endings on checkout, the file hashes will change.
-      # https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf
-      - name: set crlf
-        if: matrix.os.name == 'windows'
-        shell: bash
-        run: git config --global core.autocrlf input
```
mehulkar, line 323: Curious.. what about this is contributing to the hash for `cli#build` such that we need to find an alternative solution? And if this is the cause, would it not make sense to also add this step to the `build_turborepo` job?

It seems harder to ensure that every fixture has a .gitattributes

chris-olszewski: We can do this, but we just need to make sure that every Windows job has this set. I was thinking it could be easier just to set the test fixtures directory up with `LF` eol to avoid this, but that seems to be just as sprawling.

PR #6737, opened 2023-12-07
feat: Nice errors for shim flag parsing

NicholasLYang:
### Description

This adds nice errors for shim parsing:

Screenshot 2023-12-07 at 5 31 42 PM
Screenshot 2023-12-07 at 5 38 17 PM

### Testing Instructions

Added test for error output

Closes TURBO-1874


crates/turborepo-lib/src/shim.rs
```diff
@@ -153,24 +179,82 @@ impl ShimArgs {
...
+            let args_string = env::args().skip(1).join(" ");
+
+            return Err(Error::MultipleCwd(Box::new(MultipleCwd {
+                backtrace: Backtrace::capture(),
+                args_string,
+                flag1: indices.first().map(|i| (*i, "--cwd".len()).into()),
+                flag2: indices.get(1).map(|i| (*i, "--cwd".len()).into()),
+                flag3: indices.get(2).map(|i| (*i, "--cwd".len()).into()),
+                flag4: indices.get(3).map(|i| (*i, "--cwd".len()).into()),
+            })));
```
chris-olszewski, line 207:
```suggestion
            let mut flags = indices.iter().map(|i| (*i, "--cwd".len()).into());
            return Err(Error::MultipleCwd(Box::new(MultipleCwd {
                backtrace: Backtrace::capture(),
                args_string,
                flag1: flags.next(),
                flag2: flags.next(),
                flag3: flags.next(),
                flag4: flags.next(),
            })));
```

crates/turborepo-lib/src/shim.rs
```diff
@@ -26,13 +28,42 @@ use turborepo_ui::UI;
...
+    backtrace: Backtrace,
+    #[source_code]
+    args_string: String,
+    #[label("first flag declared here")]
+    flag1: Option<SourceSpan>,
+    #[label("but second flag declared here")]
+    flag2: Option<SourceSpan>,
+    #[label("and here")]
+    flag3: Option<SourceSpan>,
+    // The user should get the idea after the first 4 examples
```
chris-olszewski, line 45: Haha, I would hope so. I would hope 2 examples would be enough.

crates/turborepo-lib/src/shim.rs
```diff
@@ -153,24 +179,82 @@ impl ShimArgs {
...
+    /// Takes a list of indices into a Vec of arguments, i.e. ["--graph", "foo",
+    /// "--cwd"] and converts them into indices into the string of those
+    /// arguments, i.e. "-- graph foo --cwd"
+    pub(crate) fn get_indices_in_args_string(
+        mut args_indices: Vec<usize>,
+        args: impl Iterator<Item = impl Into<String>>,
+    ) -> Vec<usize> {
+        // Sort the indices to keep the invariant
+        // that if i > j then output[i] > output[j]
+        args_indices.sort();
```
chris-olszewski, line 238: Do we need to require callers to keep up this invariant? e.g. If I pass in `vec![2, 1]` and get out `vec![3, 6]` would I be in a bad state when I try to map the indices to the args.

crates/turborepo-lib/src/shim.rs
```diff
@@ -153,24 +179,82 @@ impl ShimArgs {
...
+        })
+    }
+
+    /// Takes a list of indices into a Vec of arguments, i.e. ["--graph", "foo",
+    /// "--cwd"] and converts them into indices into the string of those
+    /// arguments, i.e. "-- graph foo --cwd"
+    pub(crate) fn get_indices_in_args_string(
+        mut args_indices: Vec<usize>,
+        args: impl Iterator<Item = impl Into<String>>,
+    ) -> Vec<usize> {
```
chris-olszewski, line 235: This feels like a "sharp" interface and I have a few suggestions:
 - Make this a private function
 - Since this is only used for making pretty error messages, I don't think we need to make this as perf sensitive as it currently is. We can generate the indices into the arg strings for all args and not just ones passed in as `arg_indices`.
 - Based on how we're using this. Do we just want to calculate the spans here instead of having callers do stuff like `(idx, "--flag".len())`?

chris-olszewski (approved):
Nothing blocking

crates/turborepo-lib/src/lib.rs
```diff
@@ -54,12 +55,18 @@ pub fn get_version() -> &'static str {
 pub fn main() -> Payload {
     match shim::run() {
         Ok(payload) => payload,
+        Err(err @ (Error::MultipleCwd(..) | Error::EmptyCwd { .. })) => {
```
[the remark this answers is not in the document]
NicholasLYang: Once we delete the Go code, we should route all of them, since it provides a slightly nicer display by default. I'm just doing this to not break all of our integration tests. I'll add a comment noting that

crates/turborepo-lib/src/lib.rs
```diff
@@ -54,12 +55,18 @@ pub fn get_version() -> &'static str {
...
+            println!("{:?}", Report::new(err));
```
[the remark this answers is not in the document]
NicholasLYang: Weirdly enough `Display` doesn't do pretty printing while `Debug` does. Not sure I agree with it but w/e.

crates/turborepo-lib/src/shim.rs
```diff
@@ -153,24 +179,82 @@ impl ShimArgs {
...
-                verbosity,
-                force_update_check,
-                remaining_turbo_args,
-                forwarded_args,
-                color,
-                no_color,
-            })
+        if let Some(idx) = cwd_flag_idx {
+            let start_idx = Self::get_indices_in_args_string(vec![idx], env::args().skip(1));
+            let args_string = env::args().skip(1).join(" ");
```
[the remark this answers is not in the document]
NicholasLYang: Yeah I originally had it that way but couldn't come up with a good name

crates/turborepo-lib/src/shim.rs
NicholasLYang: Well I think the idea is you shouldn't be mapping these to args, since they're now indices into the string. And unless you parse the string it's not really easy to map the indices to the args.

PR #6743, opened 2023-12-07
feat(turbo): telemetry client

tknickman:
### Optional Anonymous Telemetry.

Telemetry allows us to accurately gauge Turborepo feature usage, pain points, and customization across all users.
This data will let us better tailor Turborepo to the masses, ensuring its continued growth, relevance, and best-in-class developer experience. Furthermore, this will allow us to verify if improvements are improving the baseline of all applications.

This is totally optional, completely anonymous, and can be disabled via CLI (`turbo telemetry disable`), or multiple environment variables (`DO_NOT_TRACK=1` or `TURBO_TELEMETRY_DISABLED=1`. 

We **DO NOT** track task names, arguments, or flag values, or anything that originates from client code including workspace names etc. Also, all events will be statically enumerated in code to ensure they can be audited at any time. 

Any and all feedback is welcome!

### Other PR's for this feature:
* [Feature Reference Docs](https://github.com/vercel/turbo/pull/6827)
* [CLI Reference Docs](https://github.com/vercel/turbo/pull/6826)
* [Feature implementation](https://github.com/vercel/turbo/pull/6743) (this PR)


crates/turborepo-telemetry/src/lib.rs
```diff
@@ -0,0 +1,424 @@
...
+            .await
+            {
+                debug!("failed to record cache usage telemetry. error: {}", err)
+            }
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
tknickman, line 247: I still need to update these tests

tknickman: Updated for now, but may change this in the future, `OnceCell` is very annoying to test.

crates/turborepo-telemetry/src/client.rs
```diff
@@ -0,0 +1,71 @@
+use std::env;
```
NicholasLYang, line 1: Should this go in the api-client crate? It feels a little confusing to have code like `create_request_builder` in the telemetry crate?

crates/turborepo-lib/src/commands/telemetry.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+    println!("Learn more: https://turbo.build/repo/docs/telemetry");
+}
+
+fn log_error(message: &str, error: &str, base: &CommandBase) {
+    println!(
+        "{}: {}",
+        base.ui.apply(BOLD_RED.apply_to(message)),
+        base.ui.apply(BOLD_RED.apply_to(error.to_string()))
+    );
```
NicholasLYang, line 33: you can use the `color!` macro to make this a little cleaner
```suggestion
    println!(
        "{}: {}",
        color!(base.ui, BOLD_RED, message),
        color!(base.ui, BOLD_RED, error)
    );
```

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -845,6 +907,7 @@ pub async fn run(
             docker,
             output_dir,
         } => {
+            telem(TelemetryEvent::KeyVal(KeyVal::command("prune")));
```
NicholasLYang, line 910: tiniest nit: this feels like a good target for a macro

crates/turborepo-telemetry/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+    match sender {
+        Some(s) => {
+            let result = s.send(event);
+            if let Err(err) = result {
+                debug!("failed to send telemetry event. error: {}", err)
+            }
+        }
+        None => {
+            if cfg!(debug_assertions) {
+                error!("[DEVELOPMENT ERROR] telemetry sender not initialized");
```
NicholasLYang, line 71: Any reason we're not including this message in the panic call?

crates/turborepo-telemetry/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+    // Used to cancel the worker
+    exit_ch: oneshot::Sender<()>,
+    client: C,
+    telemetry_id: String,
+    session_id: String,
+    enabled: bool,
+    ui: UI,
+}
+
+impl<C: telemetry::TelemetryClient + Clone + Send + Sync + 'static> Worker<C> {
```
NicholasLYang, line 158: How much is the code different between here and analytics? Because if it's very similar, could we combine events and use the same worker?

crates/turborepo-telemetry/src/events.rs
```diff
@@ -0,0 +1,40 @@
...
+pub enum TelemetryEvent {
+    Fallback(Fallback),
+    KeyVal(KeyVal),
+}
+
+/// Individual events are defined here
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct KeyVal {
+    key: String,
```
chris-olszewski, line 15: If I'm understanding this correctly and we want to prevent random strings from being used as keys, then we can use an `enum` here.

```
#[derive(Debug, Clone, Serialize, Deserialize)]
enum Key {
    Framework,
    Command,
}
```

```suggestion
    key: Key,
```

crates/turborepo-telemetry/src/events.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum TelemetryEvent {
+    Fallback(Fallback),
+    KeyVal(KeyVal),
+}
+
+/// Individual events are defined here
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
chris-olszewski, line 13: I believe this macro is a noop since none of the fields in the struct have an `_`.
```suggestion
```

crates/turborepo-telemetry/src/events.rs
```diff
@@ -0,0 +1,40 @@
+use serde::{Deserialize, Serialize};
+
+/// All possible telemetry events must be included in this enum
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
chris-olszewski, line 4: As a general rule of thumb, I find it has been helpful to write unit tests for all `Serialize` implementations just to confirm that the derived implementation is:
 - what you expect
 - doesn't change unexpectedly due to something like a field rename

Example: https://github.com/vercel/turbo/blob/main/crates/turborepo-api-client/src/spaces.rs#L254

crates/turborepo-telemetry/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+            self.senders.push(handle);
+            debug!("Done telemetry event queue flush");
+        }
+    }
+
+    fn send_events(&self, events: Vec<TelemetryEvent>) -> JoinHandle<()> {
+        if !self.enabled {
+            // do nothing
+            return tokio::spawn(async move {});
+        }
```
chris-olszewski, line 215: Tokio tasks are cheap, but `None` is cheaper. 

Will need to update callers to only add present join handles to the senders list.
```suggestion
    fn send_events(&self, events: Vec<TelemetryEvent>) -> Option<JoinHandle<()>> {
        if !self.enabled {
            // do nothing
            return None;
        }
```

chris-olszewski (commented):
Flushing some comments. Nothing super concerning so far.

crates/turborepo-telemetry/src/client.rs
tknickman: 👍🏼 yea good call - moved!

crates/turborepo-lib/src/commands/telemetry.rs
tknickman: Ah nice, TIL

crates/turborepo-telemetry/src/lib.rs
tknickman: I wasn't seeing this surfaced when I would hit the panic so I added an `error`, but moved to the panic instead!

crates/turborepo-telemetry/src/events.rs
tknickman: These were mostly for convenience, but this is _nicer_

crates/turborepo-telemetry/src/lib.rs
tknickman: Love it.

crates/turborepo-telemetry/src/events.rs
tknickman: Nice, I added one for now but will add more once the event structure gets finalized.

crates/turborepo-telemetry/src/lib.rs
tknickman: I explored this a little, and it could be made to work with multiple clients and event queues. My main concern was that the telemetry crate needs the worker to be spawned to work correctly and bundling it with the analytics worker felt a little strange (I split it out to an event-worker) crate. One of the initial goals was to make the telemetry crate as modular as possible, so shipping the worker as part of the crate feels right. But I could be convinced otherwise.

crates/turborepo-lib/src/cli/mod.rs
tknickman: I'm still not sold on this api, but when I am I will macro it up!

crates/turborepo-telemetry/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+/// Starts the `Worker` on a separate tokio thread. Returns an `TelemetrySender`
+/// and an `TelemetryHandle`.
+///
+/// We have two different types because the TelemetrySender should be shared
+/// across threads (i.e. Clone + Send), while the TelemetryHandle cannot be
+/// shared since it contains the structs necessary to shut down the worker.
+pub fn init_telemetry(
+    client: impl telemetry::TelemetryClient + Clone + Send + Sync + 'static,
+    ui: UI,
+) -> Result<TelemetryHandle, Error> {
```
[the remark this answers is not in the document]
tknickman: Good call, done!

crates/turborepo-telemetry/src/lib.rs
tknickman: I chatted more about this with @gsoltis and @Zertsov during our project sync and we're leaning more towards keeping the workers separate because the tasks that these workers are concerned with (analytics and telemetry) have different requirements. 

We _need_ analytics data to support product features (time saved, usage dashboads etc.) and thus it should have different SLA's around this data. 

For telemetry, the more data we get the better, but we shouldn't spend an excessive amount of time blocking on shutdown to send it, and we shouldn't care if some of the data is lost. 

We also still want to keep telemetry as self contained as possible moving forward.

crates/turborepo-api-client/src/lib.rs
```diff
@@ -536,6 +537,73 @@ impl APIAuth {
     }
 }
 
+// Anon Client
+
+#[async_trait]
+pub trait AnonClient {
```
NicholasLYang, line 537: Do we need this trait? It seems a little unnecessary for a single function. Also, we can probably nix the `async_trait` part

crates/turborepo-api-client/src/telemetry.rs
```diff
@@ -0,0 +1,46 @@
...
+#[async_trait]
+impl TelemetryClient for AnonAPIClient {
+    async fn record_telemetry<T>(
+        &self,
+        events: Vec<T>,
+        telemetry_id: &str,
+        session_id: &str,
+    ) -> Result<(), Error>
+    where
+        T: Serialize + std::marker::Send,
```
NicholasLYang, line 28: So we originally had this typing for analytics, but it felt rather opaque since you could pass any serializable + Send type. Could we use a concrete telemetry type? If circular dependencies are an issue, the type definitions can go into `turborepo-vercel-api`

crates/turborepo-api-client/src/lib.rs
```diff
@@ -536,6 +537,73 @@ impl APIAuth {
...
+            env::consts::ARCH
+        );
+        Ok(AnonAPIClient {
+            client,
+            base_url: base_url.as_ref().to_string(),
+            user_agent,
+        })
+    }
+
+    pub(crate) async fn create_telemetry_request_builder(
```
NicholasLYang, line 574: Should this go into the telemetry module?

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -346,6 +364,14 @@ pub enum Command {
         #[serde(skip)]
         command: Option<Box<GenerateCommand>>,
     },
+    // TODO:[telemetry] Unhide this in `1.12`
+    /// Enable or disable anonymous telemetry
+    #[clap(hide = true)]
+    Telemetry {
+        #[clap(subcommand)]
+        #[serde(flatten)]
+        command: Option<Box<TelemetryCommand>>,
```
NicholasLYang, line 373: What's the purpose of the `Box` here?

crates/turborepo-telemetry/src/config.rs
```diff
@@ -0,0 +1,284 @@
...
+            return Err(settings.unwrap_err());
+        } else if let Err(err) = settings {
+            // Propagate other errors
+            return Err(err);
+        }
+
+        // this is safe because we just checked the error case above
+        let config = settings
+            .unwrap()
+            .try_deserialize::<TelemetryConfigContents>()?;
```
NicholasLYang, line 86:
```suggestion
        let config = match settings {
            Ok(settings) => settings.try_deserialize::<TelemetryConfigContents>()?,
            Err(ConfigError::FileParse { .. }) => {
                fs::remove_file(file_path).map_err(|e| ConfigError::Message(e.to_string()))?;
                write_new_config()?;
                return Err(settings.unwrap_err());
            }
            // Propagate other errors
            Err(err) => return Err(err),
        };
```

crates/turborepo-telemetry/src/config.rs
```diff
@@ -0,0 +1,284 @@
...
+            );
+
+            let updated_config = self.alert_shown();
+            match updated_config {
+                Ok(_) => (),
+                Err(err) => error!(
+                    "Error saving seen alert event to telemetry config: {:?}",
+                    err
+                ),
+            }
```
NicholasLYang, line 168:
```suggestion
            if let Err(err) = self.alert_shown() {
                error!(
                    "Error saving seen alert event to telemetry config: {:?}",
                    err
                );
            }
```

NicholasLYang (changes requested):
Mostly minor changes. Feel free to ping me for a quick second review so we can get this merged

crates/turborepo-api-client/src/lib.rs
tknickman: Good call, done!

crates/turborepo-lib/src/cli/mod.rs
tknickman: I this pattern copied from another command I believe, removed!

crates/turborepo-api-client/src/telemetry.rs
tknickman: Yea I tried that initially just to avoid the generic here, but it ends up with circular crate dependencies in that case (even with `turborepo-vercel-api`, because then the it has to depend on `turborepo-telemetry` to pull the event types in). 

I don't mind the generic here because everything is locked down on the backend, and it also creates a single place to define everything related to event data for anyone to check to audit what we're doing. 

But if there's a better way to make it work lmk and we can get it in.

NicholasLYang: Why can't the event types be in `turborepo-vercel-api`? That's what I ended up doing for analytics

NicholasLYang (approved):

crates/turborepo-api-client/src/telemetry.rs
tknickman: Let me see if I can decouple them from the builders a bit to make it work.

PR #6747, opened 2023-12-08
Fix/replace pnpm in readme

DependerKumarSoni:
### Description

This pull request introduces a JavaScript script that dynamically updates the README.md file based on the selected package manager. This enhancement aims to improve clarity and deliver additional context for users.

### DIFF
--- a/packages/create-turbo/src/commands/create/index.ts
+++ b/packages/create-turbo/src/commands/create/index.ts
@@ -20,6 +20,7 @@ import { TransformError } from "../../transforms/errors";
 import { isDefaultExample } from "../../utils/isDefaultExample";
 import * as prompts from "./prompts";
 import type { CreateCommandArgument, CreateCommandOptions } from "./types";
+import { updateCommandsInREADME } from "./updateCommandsInREADME";
 
 const { turboGradient, turboLoader, info, error, warn } = logger;
 
@@ -92,6 +93,8 @@ export async function create(
     skipTransforms,
   });
 
+  const selectedPackageManager = selectedPackageManagerDetails.name;
+  updateCommandsInREADME(selectedPackageManager);
   if (packageManager && opts.skipTransforms) {
     warn(
       "--skip-transforms conflicts with <package-manager>. The package manager argument will be ignored."
--- a/packages/create-turbo/src/commands/create/index.ts
+++ b/packages/create-turbo/src/commands/create/index.ts
@@ -94,7 +94,15 @@ export async function create(
   });
 
   const selectedPackageManager = selectedPackageManagerDetails.name;
-  updateCommandsInREADME(selectedPackageManager);
+  if (selectedPackageManager !== "pnpm") {
+    updateCommandsInREADME(selectedPackageManager)
+      .then(() => {
+        info(`README updated for ${selectedPackageManager} commands.`);
+      })
+      .catch((error) => {
+        console.error("Unable to update README. Error: ", error);
+      });
+  }
   if (packageManager && opts.skipTransforms) {
     warn(
       "--skip-transforms conflicts with <package-manager>. The package manager argument will be ignored."
--- a/packages/create-turbo/src/commands/create/updateCommandsInREADME.js
+++ b/packages/create-turbo/src/commands/create/updateCommandsInREADME.js
@@ -1,0 +1,71 @@
+const fs = require("fs");
+
+function updateCommandsInREADME(selectedPackageManager) {
+  const filePath = "../../../../../examples/basic/README.md";
+  const anchor1 = "<!-- Build Command -->";
+  const anchor1end = "<!-- Build Command End -->";
+  const anchor2 = "<!-- Dev Command -->";
+  const anchor2end = "<!-- Dev Command End -->";
+
+  // Read the content of the file
+  fs.readFile(filePath, "utf8", (err, data) => {
+    if (err) {
+      console.error("Error reading the README file:", err);
+      return;
+    }
+
+    // Find the line that contains the anchor and replace the entire line
+    const lines = data.split("\n");
+    for (let i = 0; i < lines.length; i++) {
+      if (lines[i].includes(anchor1)) {
+        while (!lines[i].includes(anchor1end)) {
+          if (lines[i].includes("build")) {
+            lines[i] = updatedCommand(
+              lines[i],
+              selectedPackageManager,
+              "build"
+            );
+          }
+          i++;
+        }
+      }
+      if (lines[i].includes(anchor2)) {
+        while (!lines[i].includes(anchor2end)) {
+          if (lines[i].includes("dev")) {
+            lines[i] = updatedCommand(lines[i], selectedPackageManager, "dev");
+          }
+          i++;
+        }
+      }
+    }
+
+    const updatedContent = lines.join("\n");
+
+    // Write the updated content back to the file
+    fs.writeFile(filePath, updatedContent, "utf8", (writeErr) => {
+      if (writeErr) {
+        console.error("Error writing to the file:", writeErr);
+        return;
+      }
+      console.log(
+        `CLI commands modified for '${selectedPackageManager}' successfully.`
+      );
+    });
+  });
+}
+
+function updatedCommand(line, selectedPackageManager, commandName) {
+  if (selectedPackageManager != "pnpm") {
+    line = `${selectedPackageManager} run ${commandName}`;
+  } else {
+    line = `${selectedPackageManager} ${commandName}`;
+  }
+  return line;
+}
+
+module.exports = {
+  updateCommandsInREADME,
+};
+
+// Example usage:
+// updateCommandsInREADME("pnpm");
--- a/packages/create-turbo/src/commands/create/updateCommandsInREADME.js
+++ b/packages/create-turbo/src/commands/create/updateCommandsInREADME.js
@@ -1,66 +1,24 @@
-const fs = require("fs");
+const fs = require("fs").promises;
 
-function updateCommandsInREADME(selectedPackageManager) {
-  const filePath = "../../../../../examples/basic/README.md";
-  const anchor1 = "<!-- Build Command -->";
-  const anchor1end = "<!-- Build Command End -->";
-  const anchor2 = "<!-- Dev Command -->";
-  const anchor2end = "<!-- Dev Command End -->";
+const filePath = "../../../../../examples/basic/README.md";
 
-  // Read the content of the file
-  fs.readFile(filePath, "utf8", (err, data) => {
-    if (err) {
-      console.error("Error reading the README file:", err);
-      return;
-    }
+async function updateCommandsInREADME(selectedPackageManager) {
+  try {
+    // Read the content of the file
+    let data = await fs.readFile(filePath, "utf8");
 
-    // Find the line that contains the anchor and replace the entire line
-    const lines = data.split("\n");
-    for (let i = 0; i < lines.length; i++) {
-      if (lines[i].includes(anchor1)) {
-        while (!lines[i].includes(anchor1end)) {
-          if (lines[i].includes("build")) {
-            lines[i] = updatedCommand(
-              lines[i],
-              selectedPackageManager,
-              "build"
-            );
-          }
-          i++;
-        }
-      }
-      if (lines[i].includes(anchor2)) {
-        while (!lines[i].includes(anchor2end)) {
-          if (lines[i].includes("dev")) {
-            lines[i] = updatedCommand(lines[i], selectedPackageManager, "dev");
-          }
-          i++;
-        }
-      }
-    }
-
-    const updatedContent = lines.join("\n");
+    // Replace all occurrences of 'pnpm' with selectedPackageManager
+    data = data.replace(/\bpnpm\b/g, `${selectedPackageManager} run`);
 
     // Write the updated content back to the file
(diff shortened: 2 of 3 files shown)


arlyon (approved):
Nice little QoL improvement, thanks!

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -20,6 +20,7 @@ import { TransformError } from "../../transforms/errors";
 import { isDefaultExample } from "../../utils/isDefaultExample";
 import * as prompts from "./prompts";
 import type { CreateCommandArgument, CreateCommandOptions } from "./types";
+import { updateCommandsInREADME } from "./updateCommandsInREADME";
```
tknickman, line 23: Let's add this as a transformer (inside `transforms`) instead of a one-off util

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -92,6 +93,16 @@ export async function create(
     skipTransforms,
   });
 
+  const selectedPackageManager = selectedPackageManagerDetails.name;
+  if (selectedPackageManager !== "pnpm") {
```
tknickman, line 97: This shouldn't be a hardcoded check, sources aren't guaranteed to be `pnpm`, and in the case when the source isn't a first party example, I don't think we should run this transform at all. 

This should instead use the detected package manager, and only run with our own examples. We already have a transform that operates only on our own examples, so it could be added here! https://github.com/vercel/turbo/blob/main/packages/create-turbo/src/transforms/official-starter.ts

packages/create-turbo/src/commands/create/updateCommandsInREADME.js
```diff
@@ -0,0 +1,29 @@
+const fs = require("fs").promises;
```
tknickman, line 1: This needs to be written in Typescript and not CommonJS

packages/create-turbo/src/commands/create/updateCommandsInREADME.js
```diff
@@ -0,0 +1,29 @@
...
+
+const filePath = "../../../../../examples/basic/README.md";
```
tknickman, line 3: This won't work when packages and deployed, it needs to use either a root anchored path:
`const rootReadme = path.join(prompts.root, "README.md");

Or, if using `@turbo/workspaces` it can use the `paths.root` on the returned `Project` (this type can be seen [here](https://github.com/vercel/turbo/blob/main/packages/turbo-workspaces/src/types.ts#L24))

packages/create-turbo/src/commands/create/updateCommandsInREADME.js
```diff
@@ -0,0 +1,29 @@
...
+
+async function updateCommandsInREADME(selectedPackageManager) {
+  try {
+    // Read the content of the file
+    let data = await fs.readFile(filePath, "utf8");
+
+    // Replace all occurrences of 'pnpm' with selectedPackageManager
+    data = data.replace(/\bpnpm\b/g, `${selectedPackageManager} run`);
```
tknickman, line 11: This should work for all package managers, so we should be able to transform from `{npm|pnpm|yarn|bun}` to `{npm|pnpm|yarn|bun}`

packages/create-turbo/src/commands/create/updateCommandsInREADME.js
```diff
@@ -0,0 +1,29 @@
...
+
+    // Write the updated content back to the file
+    await fs.writeFile(filePath, data, "utf8");
+
+    console.log(
```
tknickman, line 16: This should happen behind the scenes, no need to log it!

tknickman (changes requested):
Thanks for taking this on! Left a few comments on the approach, let me know if you have more questions and I'm happy to help.

packages/create-turbo/src/commands/create/index.ts
DependerKumarSoni: I've added this script at packages/create-turbo/src/transforms/update-commands-in-readme.ts
Is there anything else I should take care of as per naming conventions?

packages/create-turbo/src/commands/create/index.ts
DependerKumarSoni: the logic to update README has been added to the list of transformers that run when npm install (or any other equivalent command) is executed. Hence, no need for this check anymore.

packages/create-turbo/src/commands/create/updateCommandsInREADME.js
DependerKumarSoni: Done

packages/create-turbo/src/commands/create/updateCommandsInREADME.js
DependerKumarSoni: Done

packages/create-turbo/src/commands/create/updateCommandsInREADME.js
DependerKumarSoni: Done

packages/create-turbo/src/commands/create/updateCommandsInREADME.js
DependerKumarSoni: Done

packages/create-turbo/src/transforms/update-commands-in-readme.ts
```diff
@@ -0,0 +1,44 @@
...
+export async function transform(args: TransformInput): TransformResult {
+  const { prompts, example } = args;
+
+  const isOfficialStarter =
+    !example.repo ||
+    (example.repo.username === "vercel" && example.repo.name === "turbo");
+
+  if (!isOfficialStarter) {
+    return { result: "not-applicable", ...meta };
+  }
```
tknickman, line 19: I think we should not restrict this official starters only - reason being we don't restrict the package manager transform to official starters, so if we're changing the package manager I think we should do a best effort at docs as well.

packages/create-turbo/src/transforms/update-commands-in-readme.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const selectedPackageManager = prompts.packageManager;
+  const readmeFilePath = path.join(prompts.root, "examples", "basic", "README.md");
```
tknickman, line 22: This should just be:

```suggestion
  const readmeFilePath = path.join(prompts.root, "README.md");
```

packages/create-turbo/src/transforms/update-commands-in-readme.ts
```diff
@@ -0,0 +1,44 @@
...
+  try {
+    // Read the content of the file
+    let data = await fs.readFile(readmeFilePath, "utf8");
+
+    // an array of all the possible replacement strings.
+    const replacements = ['pnpm run', 'npm run', 'yarn run', 'bun run', 'pnpm', 'npm', 'yarn', 'bun'];
+    const replacementRegex = new RegExp(`\\b(?:${replacements.join('|')})\\b`, 'g');
+
+    // Replace all occurrences of regex with selectedPackageManager
+    data = data.replace(replacementRegex, `${selectedPackageManager} run`);
```
tknickman, line 32: Nice!

I think this could still be improved a little bit. Right now, there is no differentiation between code, and non code. So, for example in our design-system example, we have a section about Versioning & Publishing Packages that mentions npm as the npm registry. 

We don't want to change those as part of this transform. I think we should tighten the regex to only detect package managers that occur within backticks (or code blocks)


This will also fix occurrences within links (for example):
> This Turborepo uses [yarn v2+ (berry)](https://yarnpkg.com/) as a packages manager. It includes the following packages/apps:

This would be a great spot to add in some unit tests. I would recommend splitting the transform itself to it's own function for easier testing - something like:

`function replacePackageManager(packageManager, text) { ... }`

This will allow you to test it very easily without requiring any additional setup / teardown.

tknickman (changes requested):
Changes look great! Only a few more comments.

packages/create-turbo/src/transforms/update-commands-in-readme.ts
DependerKumarSoni: removed isOfficialStarter check

packages/create-turbo/src/transforms/update-commands-in-readme.ts
DependerKumarSoni: Updated the path as per suggestion, thanks :)

packages/create-turbo/src/transforms/update-commands-in-readme.ts
DependerKumarSoni: As per your suggestion, I have refactored the code and the regex has been updated.
Below is the link for the test case I considered to validate the working of regex:
https://codefile.io/f/Svks01TyEk

I tried to consider the possible cases but in case I missed something or there is any scope for further improvement, please let me know.

tknickman (changes requested):
That's a beast of a regex, nice! 

One a few remaining comments left, and please include a few unit tests for the `replacePackageManager` function!

packages/create-turbo/src/transforms/update-commands-in-readme.ts
```diff
@@ -0,0 +1,50 @@
...
+  return { result: "success", ...meta };
+}
+
+function replacePackageManager(packageManager, text) {
+  // an array of all the possible replacement strings.
+  const replacements = ['pnpm run', 'npm run', 'yarn run', 'bun run', 'pnpm', 'npm', 'yarn', 'bun'];
+  
+  // regex to search for a pattern enclosed in single backticks (` `), double backticks (`` ``) or
+  // triple backticks (``` ```) considering there might be newlines in between backticks and commands.
+  const searchRegex = new RegExp(`\`\`\`[\\s\\S]*?\\b(?:${replacements.join('|')})\\b[\\s\\S]*?\`\`\`|\`\`[\\s\\S]*?\\b(?:${replacements.join('|')})\\b[\\s\\S]*?\`\`|\`[\\s\\S]*?\\b(?:${replacements.join('|')})\\b[\\s\\S]*?\``, 'g');
```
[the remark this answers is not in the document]
DependerKumarSoni: That part was the real fun.

packages/create-turbo/src/transforms/update-commands-in-readme.ts
```diff
@@ -0,0 +1,50 @@
...
+  } catch (err) {
+    throw new TransformError("Unable to update README.md", {
+      transform: meta.name,
+      fatal: false,
+    });
+  }
+  return { result: "success", ...meta };
+}
+
+function replacePackageManager(packageManager, text) {
```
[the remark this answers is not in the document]
DependerKumarSoni: What should be the type of packageManager?
should I go with `function replacePackageManager(packageManager: { name: string }, text: string): string {` ?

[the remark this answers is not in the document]
DependerKumarSoni: Hi @tknickman 
I'm sorry for not getting back to you sooner. I tried writing a test file but could not test it on my local. I have made all the changes you asked for except this Test File. Can you please guide me on this?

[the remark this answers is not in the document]
DependerKumarSoni: Hi @tknickman,
In case you're unavailable, I could use some assistance with the test files. Could you please let me know if there are any other test files currently in use within the project? I'm eager to ensure my contributions make a meaningful impact. Thanks in advance for your help!

anthonyshew (approved):
Due to this age of this PR, I'd like to merge it without tests. I'd rather unblock this PR and add tests in a separate PR if we do believe it's needed.

Thanks, @DependerKumarSoni and sorry for this taking so long to get merged!

DependerKumarSoni:
Thank you, @anthonyshew, for reviewing and considering this PR. I understand the concern about its age and truly appreciate your efforts to unblock it.

I noticed the message: "Merging is blocked: Merging can be performed automatically once the requested changes are addressed." While I believe I’ve addressed the requested changes (except for the tests, which can be added later if needed), could you please clarify if there’s anything else pending on my end to move this forward? I’m happy to make any further updates required to facilitate the merge.

Thank you again for your support!

DependerKumarSoni:
Hi @tknickman and @anthonyshew 
Thank you for reviewing and approving the changes. I wanted to follow up on this PR's merging status and would love to hear if I need to provide any support for it.

packages/create-turbo/__tests__/index.test.ts
```diff
@@ -245,7 +245,7 @@ describe("create-turbo", () => {
       telemetry,
     });
 
-    expect(mockConsole.error).toHaveBeenCalledTimes(2);
+    expect(mockConsole.error).toHaveBeenCalledTimes(3);
```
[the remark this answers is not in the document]
anthonyshew: It has to do with [https://github.com/vercel/turborepo/pull/6747/commits/870f012b5b67f96fa11aec0a8aafbfeaad287d3b#diff-87964d57d03186688cb5[…]bf2993f85483c259c7R31-R33](https://github.com/vercel/turborepo/pull/6747/commits/870f012b5b67f96fa11aec0a8aafbfeaad287d3b#diff-87964d57d03186688cb5ab10e070f97d892f841037ec86bf2993f85483c259c7R31-R33). It made complete sense to me at the time that I wrote it - but now I'm struggling to remember why. 🤔

Netail:
Any update @DependerKumarSoni ? :)

anthonyshew:
Hey, @DependerKumarSoni, sorry this took us so long. We failed you here. Let's get this merged!

PR #6757, opened 2023-12-08
Update yarn support policy.

anthonyshew:
Yarn v4 came out with a slightly different lockfile that can break our parsing at times. We're looking to get it fixed but it is not a priority at this time.


chris-olszewski (approved):

anthonyhayesres:
Is it possible to get some clarity on what "partial support" means? -> I'm trying to evaluate if I can upgrade a project from a yarn 3 to yarn 4

PR #6762, opened 2023-12-11
Use `"strict": true`  for TypeScript in examples.

anthonyshew:
Using `"strict": true` is the sane default as we head into 2024. Our examples should reflect this.


tknickman (commented):
Should this live in base? Seems like it should apply everywhere and not just to Next.

PR #6773, opened 2023-12-12
Update for `--profile` flag.

anthonyshew:
Now that we're in Rust, there are flags we need to mark as deprecated since they depended on Go.

`--profile` is the way to go now.


chris-olszewski (approved):

profile.json
```diff
@@ -0,0 +1,3 @@
+[
+
+]
```
mehulkar, line 3: @anthonyshew can we delete this and add it to gitignore? Guessing you ran with `--profile` in the root of this repo

anthonyshew: 🤦 

I won't put it in the gitignore since you can name the file manually (--profile=awesomeprofile.json) but I removed it in #6802.

PR #6774, opened 2023-12-12
chore(ci): test rust codepath by default

mehulkar:
Description in commits

Closes TURBO-1898


.github/workflows/turborepo-test-go-fallback.yml
```diff
@@ -0,0 +1,145 @@
...
+        uses: ts-graphviz/setup-graphviz@v1
+        with:
+          macos-skip-brew-update: "true"
+        env:
+          HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: true
+
+      - name: Integration Tests
+        run: turbo run test:go-fallback --filter=turborepo-tests-integration --color --env-mode=strict --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
+        env:
+          EXPERIMENTAL_RUST_CODEPATH: true
```
chris-olszewski, line 145: We can drop this now that we're onto version of `turbo` that uses Rust by default
```suggestion
```

.github/workflows/turborepo-test-go-fallback.yml
```diff
@@ -0,0 +1,145 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-turborepo-environment
+        with:
+          windows: ${{ matrix.os.runner == 'windows-latest' }}
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      - run: turbo run test --filter=cli --color
+        env:
+          EXPERIMENTAL_RUST_CODEPATH: true
```
chris-olszewski, line 105: We can drop this now that we're on a version of `turbo` that uses Rust by default

```suggestion
```

.github/workflows/turborepo-test-go-fallback.yml
```diff
@@ -0,0 +1,145 @@
...
+
+  integration:
```
chris-olszewski, line 107: Shouldn't we have these wait for the `build_turbo` step from the primary workflow to avoid duplicating work?

chris-olszewski (changes requested):
Only concern is that we're now racing two builds of `turbo` against each other for the normal test workflow and the go fallback test workflow.

.github/workflows/turborepo-test-go-fallback.yml
mehulkar: Yeah I thought about that, but didn't feel like it was worth it for the optimization. `make test-go` can just build it? I can put it back in the main workflow I guess, but I feel like this is cleaner

mehulkar: talked offline and decided to merge the workflows back in to reduce compute time

chris-olszewski (approved):

PR #6778, opened 2023-12-12
turbopack-css: fix rule duplication

wbinnssmith:
This fixes an issue in css chunk generation where rules could be duplicated, which was likely introduced in the chunking refactor (a series of PRs starting with #6104).

CSS chunk generation had special logic that would walk the module graph on its own and include code. Combined with the topological list of modules provided by chunking, this unnecessarily included duplicate rules in incorrect order.

Test Plan: Added and updated snapshot tests.

Closes PACK-2141


kdy1 (approved):

crates/turbopack-core/src/reference_type.rs
```diff
@@ -144,7 +163,11 @@ impl ReferenceType {
                     && matches!(sub_type, EcmaScriptModulesReferenceSubType::Undefined)
             }
             ReferenceType::Css(sub_type) => {
-                matches!(other, ReferenceType::Css(_))
+                (
+                    // For condition matching, treat any AtImport pair as identical.
```
sokra, line 167: I would put that logic into it's own match branch

crates/turbopack-css/src/chunk/mod.rs
```diff
@@ -65,21 +63,65 @@ impl CssChunk {
 
         let this = self.await?;
 
+        let mut code = CodeBuilder::default();
         let mut body = CodeBuilder::default();
-        let mut external_imports = IndexSet::new();
-        for css_item in this.content.await?.chunk_items.iter() {
+        let mut seen = HashSet::new();
+        for css_item in &this.content.strongly_consistent().await?.chunk_items {
```
sokra, line 69: The strongly consistent shouldn't be needed here

crates/turbopack-css/src/chunk/mod.rs
```diff
@@ -65,21 +63,65 @@ impl CssChunk {
...
-        }
+            seen.insert(key);
 
-        let mut code = CodeBuilder::default();
-        for external_import in external_imports {
-            writeln!(code, "@import {};", stringify_js(&external_import))?;
+            let content = &css_item.content().await?;
+            for import in &content.imports {
+                if let CssImport::External(external_import) = import {
+                    writeln!(code, "@import {};", stringify_js(&external_import.await?))?;
```
sokra, line 88: `@import` can only appear at the start of a css file. You need to collect all external css first and make sure they are emitted first

crates/turbopack-css/src/references/import.rs
```diff
@@ -344,13 +218,40 @@ impl ImportAssetReference {
 #[turbo_tasks::value_impl]
 impl ModuleReference for ImportAssetReference {
     #[turbo_tasks::function]
-    fn resolve_reference(&self) -> Vc<ModuleResolveResult> {
-        css_resolve(
+    async fn resolve_reference(&self) -> Result<Vc<ModuleResolveResult>> {
+        let css_asset = Vc::try_resolve_downcast_type::<CssModuleAsset>(self.origin).await?;
```
sokra, line 222: I think I would prefer to just pass the previous `ImportContext` as additional argument to `ImportAssetReference::new`

crates/turbopack-css/src/references/import.rs
```diff
@@ -344,13 +218,40 @@ impl ImportAssetReference {
...
+            };
+
+            import_context
+        };
+
+        Ok(css_resolve(
             self.origin,
             self.request,
-            Value::new(CssReferenceSubType::AtImport),
+            Value::new(CssReferenceSubType::AtImport(Some(import_context.cell()))),
```
sokra, line 252: The import_context.cell() will create new VCS here even for identical content. Instead create a `ImportContext::new` turbo tasks function to make this a singleton resp deduplicates.

crates/turbopack-css/src/references/import.rs
```diff
@@ -344,13 +218,40 @@ impl ImportAssetReference {
...
+            };
+
+            if let Some(layer) = own_attrs.layer {
+                import_context.layers.push(layer);
+            };
+            if let Some(supports) = own_attrs.supports {
+                import_context.supports.push(supports);
+            };
+            if let Some(media) = own_attrs.media {
+                import_context.media.push(media);
```
sokra, line 243: I feel like we should use an IndexSet instead of a Vec here to deduplicates media, supports and layers

crates/turbopack-css/src/chunk/mod.rs
wbinnssmith: This should do that. The imports are written directly to `code`, while file contents are added to `body`, which is added to `code` only after the items have been processed.

crates/turbopack-core/src/reference_type.rs
```diff
@@ -54,14 +55,74 @@ pub struct ImportAttributes {
 
 /// The accumulated list of conditions that should be applied to this module
 /// through its import path
-#[derive(Clone, Debug, Default, Hash)]
-#[turbo_tasks::value(shared)]
+#[derive(Debug, Default)]
+#[turbo_tasks::value(transparent)]
```
sokra, line 59: Transparent on struct types has no effect. You can remove that.

crates/turbopack-core/src/reference_type.rs
```diff
@@ -54,14 +55,74 @@ pub struct ImportAttributes {
...
+            let mut supports = this.supports.clone();
+            if let Some(attr_supports) = attr_supports {
+                if !supports.contains(&attr_supports) {
+                    supports.push(attr_supports.to_owned());
+                }
+            }
+            supports
+        };
+
+        Ok(ImportContext {
```
sokra, line 117: This must call `ImportContext::new` to ensure it is a singleton

PR #6781, opened 2023-12-13
feat: add anon-profile flag

chris-olszewski:
### Description

This PR adds `--anon-profile` which will generate a profile, but without any identifying data included. It will only output where in the code that time was spent.

(The PR also adds a requirement that a user passes a non-empty path for the profile file)

### Testing Instructions

Added unit tests for usage of `--profile` and `--anon-profile`. Quick spot check that `--anon-profile` removes any identifying data:
```
turbo_dev build --anon-profile="profile.json" -vv
```
Resulting [profile.json](https://github.com/vercel/turbo/files/13654482/profile.json)

Closes TURBO-1903

### DIFF
--- a/crates/turborepo-lib/src/tracing.rs
+++ b/crates/turborepo-lib/src/tracing.rs
@@ -189,10 +189,14 @@ impl TurboSubscriber {
 
     /// Enables chrome tracing.
     #[tracing::instrument(skip(self, to_file))]
-    pub fn enable_chrome_tracing<P: AsRef<Path>>(&self, to_file: P) -> Result<(), Error> {
+    pub fn enable_chrome_tracing<P: AsRef<Path>>(
+        &self,
+        to_file: P,
+        include_args: bool,
+    ) -> Result<(), Error> {
         let (layer, guard) = tracing_chrome::ChromeLayerBuilder::new()
             .file(to_file)
-            .include_args(true)
+            .include_args(include_args)
             .include_locations(true)
             .build();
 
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/tracing.rs
```diff
@@ -189,10 +189,14 @@ impl TurboSubscriber {
...
-    pub fn enable_chrome_tracing<P: AsRef<Path>>(&self, to_file: P) -> Result<(), Error> {
+    pub fn enable_chrome_tracing<P: AsRef<Path>>(
+        &self,
+        to_file: P,
+        include_args: bool,
+    ) -> Result<(), Error> {
         let (layer, guard) = tracing_chrome::ChromeLayerBuilder::new()
             .file(to_file)
-            .include_args(true)
+            .include_args(include_args)
```
tknickman, line 199: Haven't looked at the chrome traces in a while, but will this still include any package or task names?

tknickman (approved):
One question but overall 👍🏼

crates/turborepo-lib/src/tracing.rs
arlyon: It will only include the function names. We will not be able to get access to and function arguments so it is likely that packages and tasks will not appear. It _will_ however let us see what parts of the run are taking time.

I mentioned in a thread it may be possible to use valuable with some type-aware filter and some newtype so we can filter out sensitive data

```rust
tracing::log!("running task {} with absolute path {}", task_name, Sensitive(absolute_path));
```

PR #6785, opened 2023-12-13
Ensure turbopack-trace-server doesn't send message in a loop

timneutkens:
### Description

While working on the trace viewer I noticed that each `ack` causes another message to be sent, which results in another `ack` so it keeps looping.

This ensures `ack` only sends for the update case. Other messages like sending a new viewrect will still send an update correctly with these changes.


timneutkens:
* `main`
  * **PR #6785** <a href="https://app.graphite.dev/github/pr/vercel/turbo/6785?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/6785?utm_source=stack-comment).

timneutkens:
Landing this PR as it ensures the trace viewer doesn't slow down.

PR #6786, opened 2023-12-13
Updated outdated Docs "Getting Started" > "Create a New Monorepo"

Gogoro:
### Description
When going through the ["Getting Started > Create a New Monorepo"](https://turbo.build/repo/docs/getting-started/create-new) documentation I noticed it being outdated compared to the code I got from `npx create-turbo@latest`.

I just went through the docs step by step and updated what I saw. Some parts where more outdated than others, so I had to get my hands a bit dirty to make it make sense 😇

I'm not the best typist in the world, so I would appreciate it if someone could read through and make sure it's tight enough for the Turbo Docs 😎 

### Testing Instructions
- [x] Check spelling
- [x] Check grammar
- [x] Validate that new documentation is correct


docs/pages/repo/docs/getting-started/create-new.mdx
```diff
@@ -140,70 +140,84 @@ You'll see that `"web"` depends on a package called `"ui"`.
...
+      </Button>
     <>
   );
 }
 ```
 
-They're importing `Button` directly from a dependency called `ui`! How does that work? Where is `Button` coming from?
+They're importing `Button` directly from a dependency called `@repo/ui/button`! How does that work? Where is `Button` coming from?
 
 Open `packages/ui/package.json`. You'll notice these two attributes:
```
anthonyshew, line 177: "these two attributes" doesn't make sense anymore since it's now the `exports` field. Could you reword this to "You'll notice the `exports` field:"

docs/pages/repo/docs/getting-started/create-new.mdx
```diff
@@ -140,70 +140,84 @@ You'll see that `"web"` depends on a package called `"ui"`.
...
+  "exports": {
+    "./button": "./src/button.tsx",
+    "./card": "./src/card.tsx",
+    "./code": "./src/code.tsx"
+  },
 }
 ```
 
-When workspaces import from `ui`, `main` tells them where to access the code they're importing. `types` tells them where the TypeScript types are located.
+When workspaces import from `@repo/ui/button`, `export` tells them where to access the code they're importing.
```
anthonyshew, line 189: `export` -> `exports`

anthonyshew (changes requested):
This is *awesome*. Thanks for this! Really detail oriented.

I think I found a few small touch ups and we'll be ready to ship this.

- On line 104-105, we changed the names of these directories, as you noted further down the page. Could you update them here as well?
- From lines 401 to 413, was this code snippet meant to be removed in favor of 415-430?

docs/pages/repo/docs/getting-started/create-new.mdx
Gogoro: That is much better, thank you! 🤩

docs/pages/repo/docs/getting-started/create-new.mdx
Gogoro: Ops 🙈 Thank you for spotting that one!

Gogoro:
Thank you for such a great review and nice words @anthonyshew 🫡 

> * On lines 104-105, we changed the names of these directories, as you noted further down the page. Could you update them here as well?

Good catch! Fixed ✅

> * From lines 401 to 413, was this code snippet meant to be removed in favor of 415-430?

Yes you are correct. I forgot to remove the old example when I added the new 🤪 Fixed now ✅


The comments you made in the diff should also be resolved now. 

Have a great day! 🌄

anthonyshew (approved):
Looking good!

Thank you again for this - and for being so quick to fix! `>>> FULL TURBO` 😄

PR #6788, opened 2023-12-13
feat: add scan subcommand

arlyon:
### Description

Adds a scan command that automates common optimisations so that you can get the most out of turbo.

### Testing Instructions

Closes TURBO-1912

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -24,7 +24,7 @@ use turborepo_ui::UI;
 
 use crate::{
     commands::{
-        bin, daemon, generate, info, link, login, logout, prune, run, telemetry, unlink,
+        bin, daemon, generate, info, link, login, logout, prune, run, scan, telemetry, unlink,
         CommandBase,
     },
     get_version,
@@ -461,6 +461,9 @@ pub enum Command {
         #[serde(flatten)]
         command: Option<TelemetryCommand>,
     },
+    /// Turbo your monorepo by running a number of 'repo lints' to
+    /// identify common issues, suggest fixes, and improve performance.
+    Scan {},
     #[clap(hide = true)]
     Info {
         workspace: Option<String>,
@@ -1075,6 +1078,11 @@ pub async fn run(
             let mut base = CommandBase::new(cli_args.clone(), repo_root, version, ui);
             let child_event = event.child();
             telemetry::configure(command, &mut base, child_event);
+            Ok(0)
+        }
+        Command::Scan {} => {
+            let base = CommandBase::new(cli_args.clone(), repo_root, version, ui);
+            scan::run(base).await;
             Ok(0)
         }
         Command::Info { workspace, json } => {
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -1082,8 +1082,11 @@ pub async fn run(
         }
         Command::Scan {} => {
             let base = CommandBase::new(cli_args.clone(), repo_root, version, ui);
-            scan::run(base).await;
-            Ok(0)
+            if scan::run(base).await {
+                Ok(0)
+            } else {
+                Ok(1)
+            }
         }
         Command::Info { workspace, json } => {
             CommandEventBuilder::new("info")
--- a/crates/turborepo-lib/src/commands/scan.rs
+++ b/crates/turborepo-lib/src/commands/scan.rs
@@ -19,7 +19,7 @@ static DIALOGUER_THEME: LazyLock<DialoguerTheme> = LazyLock::new(|| DialoguerThe
 
 /// diagnostics run in parallel however to prevent messages from appearing too
 /// quickly we introduce a minimum delay between each message
-const INTER_MESSAGE_DELAY: Duration = Duration::from_millis(300);
+const INTER_MESSAGE_DELAY: Duration = Duration::from_millis(30);
 
 /// Start a diagnostic session. This command will run a series of diagnostics to
 /// help you identify potential performance bottlenecks in your monorepo.
--- a/crates/turborepo-lib/src/commands/scan.rs
+++ b/crates/turborepo-lib/src/commands/scan.rs
@@ -1,0 +1,138 @@
+use std::{sync::LazyLock, time::Duration};
+
+use console::{style, Style};
+use indicatif::ProgressBar;
+use turborepo_ui::*;
+
+use super::CommandBase;
+use crate::{diagnostics::*, DaemonPaths};
+
+// can't use LazyCell since DialoguerTheme isn't Sync
+static DIALOGUER_THEME: LazyLock<DialoguerTheme> = LazyLock::new(|| DialoguerTheme {
+    prompt_prefix: style(">>>".to_string()).bright().for_stderr(),
+    active_item_prefix: style("  ❯".to_string()).for_stderr().green(),
+    inactive_item_prefix: style("   ".to_string()).for_stderr(),
+    success_prefix: style("   ".to_string()).for_stderr(),
+    prompt_style: Style::new().for_stderr(),
+    ..Default::default()
+});
+
+/// diagnostics run in parallel however to prevent messages from appearing too
+/// quickly we introduce a minimum delay between each message
+const INTER_MESSAGE_DELAY: Duration = Duration::from_millis(300);
+
+/// Start a diagnostic session. This command will run a series of diagnostics to
+/// help you identify potential performance bottlenecks in your monorepo.
+///
+/// Note: all lints happen in parallel. For the purposes of displaying output,
+///       we demultiplex the output and display it in a single stream, meaning
+///       to the user, it looks like the lints are running serially.
+pub async fn run(base: CommandBase) {
+    let paths = DaemonPaths::from_repo_root(&base.repo_root);
+    let ui = base.ui;
+
+    println!("\n{}\n", ui.rainbow(">>> TURBO SCAN"));
+    println!(
+        "Turborepo does a lot of work behind the scenes to make your monorepo fast,
+however, there are some things you can do to make it even faster. {}\n",
+        BOLD_GREEN.apply_to("Let's go!")
+    );
+
+    let (tx, mut rx) = tokio::sync::mpsc::channel(100);
+
+    // NUM_TASKS needs to line up with the below. we could _potentially_ use a vec
+    // of tasks with dyn traits but trait object safety is a pain
+    const NUM_TASKS: usize = 5;
+    DaemonDiagnostic(&paths).execute(tx.clone());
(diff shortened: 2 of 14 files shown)


crates/turborepo-lib/src/diagnostics.rs
```diff
@@ -0,0 +1,492 @@
...
+
+pub struct LSPDiagnostic(pub AbsoluteSystemPathBuf);
+impl Diagnostic for LSPDiagnostic {
+    const NAME: &'static str = "turbo.lsp";
+
+    fn execute(&self, chan: MpscSender<DiagnosticMessage>) {
+        let daemon_root = self.0.clone();
+        tokio::task::spawn(async move {
+            Self::started(&chan, "Turborepo Extension".to_string()).await;
+            Self::log_line(&chan, "Checking if it is running...".to_string()).await;
```
NicholasLYang, line 334: nit, this is less clear than I'd like
Screenshot 2023-12-13 at 12 03 04 PM

```suggestion
            Self::log_line(&chan, "Checking if extension is running...".to_string()).await;
```

NicholasLYang (dismissed):
Overall looks quite solid, mostly feedback around UI/UX

crates/turborepo/build.rs
```diff
@@ -8,7 +8,7 @@ fn main() {
 
     let invocation = std::env::var("RUSTC_WRAPPER").unwrap_or_default();
     if !is_ci_release && !invocation.ends_with("rust-analyzer") {
-        build_local_go_binary(profile);
+        // build_local_go_binary(profile);
```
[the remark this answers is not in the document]
arlyon: Oops

tknickman (dismissed):
We should split the docs into their own PR, and label them as "next release" so that we don't have docs that are live and point to an unreleased feature. 

Then once we ship the next generally available (non-canary) release we can go through and merge all the docs PRs that are ready to be live!

crates/turborepo-lib/src/commands/scan.rs
```diff
@@ -0,0 +1,138 @@
...
+///       to the user, it looks like the lints are running serially.
+pub async fn run(base: CommandBase) {
+    let paths = DaemonPaths::from_repo_root(&base.repo_root);
+    let ui = base.ui;
+
+    println!("\n{}\n", ui.rainbow(">>> TURBO SCAN"));
+    println!(
+        "Turborepo does a lot of work behind the scenes to make your monorepo fast,
+however, there are some things you can do to make it even faster. {}\n",
+        BOLD_GREEN.apply_to("Let's go!")
```
NicholasLYang, line 38: We should probably obey the color config here
```suggestion
        color!(ui, BOLD_GREEN, "Let's go!")
```

crates/turborepo-lib/src/commands/scan.rs
```diff
@@ -0,0 +1,138 @@
...
+    );
+
+    let (tx, mut rx) = tokio::sync::mpsc::channel(100);
+
+    // NUM_TASKS needs to line up with the below. we could _potentially_ use a vec
+    // of tasks with dyn traits but trait object safety is a pain
+    const NUM_TASKS: usize = 5;
```
NicholasLYang, line 45: Could we use a macro?

crates/turborepo-lib/src/commands/scan.rs
```diff
@@ -0,0 +1,138 @@
...
+                let handle = tokio::task::spawn_blocking(move || {
+                    bar.suspend(|| {
+                        resume.blocking_recv().ok(); // sender is dropped, so we
+                                                     // can unsuspend
+                    });
+                });
+                stopped.send(()).ok(); // suspender doesn't need to be notified so failing ok
+                handle.await.unwrap();
+            }
+            // any interleaved events will be requeued such that we only process one id at a
```
NicholasLYang, line 118: This feels a little awkward. Could we move it to a StreamMap?

crates/turborepo-lib/src/diagnostics.rs
```diff
@@ -0,0 +1,523 @@
...
+                        )
+                        .await;
+                    }
+
+                    let fsmonitor = String::from_utf8_lossy(fsmonitor);
+                    let untrackedcache = String::from_utf8_lossy(untrackedcache);
+
+                    if fsmonitor.trim() != "true" || untrackedcache.trim() != "true" {
+                        Self::log_line(&chan, "Git FS Monitor not configured".to_string()).await;
+                        Self::log_line(&chan, "For more information, see https://turbo.dev/repo/docs/reference/command-line-reference/scan#fs-monitor".to_string()).await;
```
chris-olszewski, line 219: I don't think we own `turbo.dev` anymore
```suggestion
                        Self::log_line(&chan, "For more information, see https://turbo.build/repo/docs/reference/command-line-reference/scan#fs-monitor".to_string()).await;
```

crates/turborepo-lib/src/diagnostics.rs
```diff
@@ -0,0 +1,523 @@
...
+pub struct GitDaemonDiagnostic;
+
+impl Diagnostic for GitDaemonDiagnostic {
+    const NAME: &'static str = "git.daemon";
+
+    fn execute(&self, chan: MpscSender<DiagnosticMessage>) {
+        tokio::task::spawn(async move {
+            if !(cfg!(target_os = "windows") || cfg!(target_os = "macos")) {
+                // the git daemon is not implemented on unix
+                Self::not_applicable(chan, "Git FS Monitor (not available on unix)".to_string())
```
NicholasLYang, line 149: Is it Unix or Linux? macOS is Unix based, no?

crates/turborepo-lib/src/diagnostics.rs
```diff
@@ -0,0 +1,523 @@
...
+                Ok(Some(pid)) => {
+                    Self::done(
+                        chan,
+                        format!("Turborepo Extension is running (pid {})", pid),
+                    )
+                    .await;
+                }
+                Ok(None) => {
+                    Self::log_line(&chan, "Unable to find LSP instance".to_string()).await;
+                    Self::log_line(&chan, "For more information, see https://turbo.dev/repo/docs/reference/command-line-reference/scan#lsp".to_string()).await;
```
chris-olszewski, line 371:
```suggestion
                    Self::log_line(&chan, "For more information, see https://turbo.build/repo/docs/reference/command-line-reference/scan#lsp".to_string()).await;
```

crates/turborepo-lib/src/diagnostics.rs
```diff
@@ -0,0 +1,523 @@
...
+                    &["--version"][..],
+                    &["config", "--get", "core.fsmonitor"][..],
+                    &["config", "--get", "core.untrackedcache"][..],
+                ]
+                .into_iter()
+                .map(|args| async move {
+                    // get the current setting
+                    let stdout = Stdio::piped();
+
+                    let command = Command::new("git")
```
NicholasLYang, line 167: We might want to consider using `which` here. That generally avoids some bugs with finding the correct command

crates/turborepo-lib/src/diagnostics.rs
```diff
@@ -0,0 +1,523 @@
...
+
+            match futures.as_ref().map(|v| v.as_slice()) {
+                Ok([version, fsmonitor, untrackedcache]) => {
+                    let version = String::from_utf8_lossy(version);
+                    let Some(version) = version.trim().strip_prefix("git version ") else {
+                        Self::failed(chan, "Failed to get git version".to_string()).await;
+                        return;
+                    };
+
+                    let (major, rest) = version.split_once('.').expect("semver");
```
NicholasLYang, line 190: Should we panic here?

crates/turborepo-lib/src/diagnostics.rs
```diff
@@ -0,0 +1,523 @@
...
+                    if major == 2 && minor < 37 || major == 1 {
+                        Self::failed(
+                            chan,
+                            format!(
+                                "Git version {} is too old, please upgrade to 2.37 or newer",
+                                version
+                            ),
+                        )
+                        .await;
+                        return;
```
chris-olszewski, line 205: We already use the [semver](https://docs.rs/semver/latest/semver/) crate and that might be a little more robust than handrolling parsing and range checks.

crates/turborepo-lib/src/diagnostics.rs
```diff
@@ -0,0 +1,523 @@
...
+            let connector = DaemonConnector {
+                can_kill_server: false,
+                can_start_server: true,
+                paths,
+            };
+
+            let mut client = match connector.connect().await {
+                Ok(client) => client,
+                Err(e) => {
+                    Self::failed(chan, format!("Failed to connect to daemon: {}", e)).await;
```
NicholasLYang, line 325: Should we say something like "failed to connect to daemon, please start it with `turbo daemon start`"? Or even offer to start it for them?

NicholasLYang (changes requested):

chris-olszewski (approved):
I think this is a good start to this feature. ITG

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -1077,6 +1080,11 @@ pub async fn run(
             telemetry::configure(command, &mut base, child_event);
             Ok(0)
         }
+        Command::Scan {} => {
+            let base = CommandBase::new(cli_args.clone(), repo_root, version, ui);
+            scan::run(base).await;
+            Ok(0)
```
gsoltis, line 1086: I suspect we don't always want exit code 0. Maybe only if we don't find anything?

gsoltis (changes requested):
Mostly deferring to the reviews from @NicholasLYang and @chris-olszewski, but I do think the exit code piece is worth addressing in case someone wants to use this in an automated check somewhere.

crates/turborepo-lib/src/diagnostics.rs
arlyon: Yes, oops

crates/turborepo-lib/src/diagnostics.rs
```diff
@@ -0,0 +1,523 @@
...
+pub struct UpdateDiagnostic;
+
+impl Diagnostic for UpdateDiagnostic {
+    const NAME: &'static str = "turbo.update";
+
+    fn execute(&self, chan: MpscSender<DiagnosticMessage>) {
+        tokio::task::spawn(async move {
+            Self::started(&chan, "Update Turborepo to latest version".to_string()).await;
+            Self::log_line(&chan, "Checking for updates...".to_string()).await;
+            let version = check_for_updates(
```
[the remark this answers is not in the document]
arlyon: The blocking is probably just the time it takes to do the check since each lint dumps its entire logs before moving to the next one which I don't think will be improved. spawn_blocking should help on very constrained devices though so we'll do it anyways.

crates/turborepo-lib/src/cli/mod.rs
arlyon: Changed

PR #6789, opened 2023-12-13
feat: Trying out mimalloc

NicholasLYang:
### Description

According to [this article](https://www.tweag.io/blog/2023-08-10-rust-static-link-with-mimalloc/), malloc can be a bottleneck on musl. They recommend using a different malloc implementation like mimalloc. It's a relatively low cost to try out mimalloc, so why not?

### Testing Instructions

Closes TURBO-1913

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4153,9 +4153,9 @@ checksum = "348108ab3fba42ec82ff6e9564fc4ca0247bdccdc68dd8af9764bbc79c3c8ffb"
 
 [[package]]
 name = "libmimalloc-sys"
-version = "0.1.30"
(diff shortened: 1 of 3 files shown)


chris-olszewski (changes requested):
Do you have some test results for if switching our `malloc` implementation changes performance? I'm not comfortable swapping out such an fundamental piece of the program without some data.

NicholasLYang:
I'll look into it. Not familiar with how our benchmarks are being run currently. I will note that turbopack uses mimalloc and doesn't seem to have any issues with it

NicholasLYang:
Could not detect significant performance improvement, closing for now

PR #6790, opened 2023-12-13
fix: make sure that our concurrency check ignore packages without task

### LINKED ISSUE #4291: [WEB-880] [turborepo] Persistent task concurrency validation not taking into account whether package has the script defined

arlyon:
### Description

Closes #4291

The validation check to ensure the persistent task count doesn't exceed concurrency did not check if the package actually contained a particular task, leading to erroneous validation when running persistent tasks if you have more than N packages in the project.

Closes TURBO-1914

### DIFF
--- a/crates/turborepo-lib/src/engine/mod.rs
+++ b/crates/turborepo-lib/src/engine/mod.rs
@@ -163,10 +163,6 @@ impl Engine<Built> {
                     // No need to check the root node if that's where we are.
                     return Ok(false);
                 };
-                let is_persistent = self
-                    .task_definitions
-                    .get(task_id)
-                    .map_or(false, |task_def| task_def.persistent);
 
                 for dep_index in self
                     .task_graph
@@ -203,7 +199,19 @@ impl Engine<Built> {
                     }
                 }
 
-                Ok(is_persistent)
+                // check if the package for the task has that task in its package.json
+                let info = package_graph
+                    .workspace_info(&WorkspaceName::Other(task_id.package().to_string()))
+                    .expect("package graph should contain workspace info for task package");
+
+                let package_has_task = info.package_json.scripts.contains_key(task_id.task());
+
+                let task_is_persistent = self
+                    .task_definitions
+                    .get(task_id)
+                    .map_or(false, |task_def| task_def.persistent);
+
+                Ok(task_is_persistent && package_has_task)
             })
             .fold((0, Vec::new()), |(mut count, mut errs), result| {
                 match result {
--- a/crates/turborepo-lib/src/engine/mod.rs
+++ b/crates/turborepo-lib/src/engine/mod.rs
@@ -227,6 +227,8 @@ impl Engine<Built> {
                 (count, errs)
             });
 
+        // there must always be at least one concurrency 'slot' available for
+        // non-persistent tasks otherwise we get race conditions
         if persistent_count >= concurrency {
             validation_errors.push(ValidateError::PersistentTasksExceedConcurrency {
                 persistent_count,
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -201,7 +197,19 @@ impl Engine<Built> {
                     }
                 }
 
-                Ok(is_persistent)
+                // check if the package for the task has that task in its package.json
+                let info = package_graph
+                    .workspace_info(&WorkspaceName::Other(task_id.package().to_string()))
```
chris-olszewski, line 202: I'm assuming this is what's causing the crashes. `WorkspaceName::from` handles `//` correctly and maps it to `WorkspaceName::Root`
```suggestion
                    .workspace_info(&WorkspaceName::from(task_id.package().to_string()))
```

crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -201,7 +197,19 @@ impl Engine<Built> {
...
+                    .expect("package graph should contain workspace info for task package");
+
+                let package_has_task = info.package_json.scripts.contains_key(task_id.task());
```
chris-olszewski, line 205: In order to match what we do other places to match Go behavior I think we need to also treat `""` as a package not having a task.
```suggestion
                let package_has_task = info.package_json.scripts.get(task_id.task()).map_or(false, |script| !script.is_empty());
```

chris-olszewski (changes requested):
It would be great to have a test for these changes.

It's also worth noting that this introduces intentionally divergent behavior between Rust/Go. Not saying that updating Go is necessary since we're deleting it in a ~month, but might affect testing strategy.

mehulkar (commented):
Change looks good, but a test case would be great (IMO unit test would be fine in this case if that's possible)

crates/turborepo-lib/src/engine/mod.rs
arlyon: TY

crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -260,3 +273,112 @@ impl fmt::Display for TaskNode {
...
+        }
+
+        let engine = engine.seal();
+        let current_dir = std::env::current_dir().unwrap();
+
+        let mut graph_builder = PackageGraph::builder(
+            AbsoluteSystemPath::from_std_path(&current_dir).unwrap(),
+            PackageJson::default(),
+        )
+        .with_package_discovery(DummyDiscovery(&tmp));
```
chris-olszewski, line 371: I think we could avoid providing valid paths and writing to the FS by using builder methods like we do [here](https://github.com/vercel/turbo/blob/main/crates/turborepo-repository/src/package_graph/mod.rs#L695). I think this would avoid the test failures for Windows.

chris-olszewski (approved):
Approving. Just need to update the integration tests with a lower the concurrency limit now that we're checking for N instead of N+1.

Also get the Windows rust test to pass, probably easiest way is to avoid actual FS work completely instead of trying to figure out traversals between drives.

PR #6796, opened 2023-12-14
chore: include response body if not valid JSON

chris-olszewski:
### Description

I noticed in our Ubuntu CI runs there's sometimes a `artifact verification failed: Error making HTTP request: error decoding response body: expected value at line 1 column 1` [source](https://github.com/vercel/turbo/actions/runs/7210774346/job/19644690101#step:6:408)

It would be more useful if we could get an idea of what the response was instead of a JSON parse error.

### Testing Instructions

Added unit tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10887,6 +10887,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "chrono",
+ "http",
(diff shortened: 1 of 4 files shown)


tknickman (approved):
❤️ it

PR #6801, opened 2023-12-15
feat(ci): enable examples tests on Windows

mehulkar:
Previous attempt in https://github.com/vercel/turbo/pull/5314


mehulkar:
We're not going to try to do this, examples tests are finicky and aren't providing additional value when running on multiple platforms.

PR #6804, opened 2023-12-15
Prepare linter and TypeScript for docs.

anthonyshew:
### Description

For ESLint, this is just a setup PR that caps the number of warnings we're willing to accept.

For TypeScript, I unfortunately need to fix a whole mess of errors to make ESLint run. Rather than fix all the errors, I'm just throwing down a bunch of any's to keep the PR small.

CLOSES TURBO-1941


docs/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "lib": ["dom", "dom.iterable", "esnext"],
     "skipLibCheck": true,
     "strict": false,
+    "strictNullChecks": true,
```
anthonyshew, line 7: Some linting rules require this.

docs/.eslintrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  root: true,
```
anthonyshew, line 2: Ensures ESLint stops walking upwards so we know which config we're getting.

(We'll soon be using flat config but we'll cross that bridge when we get there.)

mehulkar (approved):

tknickman (approved):

docs/package.json
```diff
@@ -7,7 +7,7 @@
     "dev": "next",
     "start": "next start",
     "build": "next build ",
-    "lint": "next lint",
+    "lint": "eslint . --max-warnings 356",
```
anthonyshew, line 10: We're starting at 356! 😄

Issue #1235, opened 2022-05-13
Forward stdin


marvinroger:
### Describe the feature you'd like to request

Hi!

I have a pretty standard setup, where I have a `dev` script running `nodemon`. `nodemon` allows to type a special `rs` command to stdin, triggering a restart of the process, which is very useful.

It works fine when ran directly, but it does not work through turbo, as turbo does not forward stdin to the underlying process.

### Describe the solution you'd like

Maybe a task definition in `turbo.json` could have a `forwardStdin` option, which would, when enabled, forward stdin!
I think `forwardStdin: true` should be mutually exclusive with `cache: true`, because stdin could impact the result of the task.

<sub>[TURBO-588](https://linear.app/vercel/issue/TURBO-588/pipe-stdin-to-non-concurrent-tasks)</sub>

MohamedLamineAllal:
`pnpm run publish`  is a workflow that i want to have. And i have otp code that i have to provide making it interactive. Doesn't work through turbo.

gsoltis:
`forwardStdin` is an interesting idea. I think it would have scheduling implications as well, only one task at a time would be able to grab it

MyNameReallySux:
Any input on the priority/interest in this from the Vercel team? Any major engineering blockers?

It seems like this would be possible without too much complexity as long as `--concurrency=1` and  `--cache=false`. This is something I would attempt to fix, but not too awesome at Rust 😞. If someone can give me a quick pointer as to where this code would need to be implemented in the repo, I may attempt to fix in the future when I get some time.

gsoltis:
I think if we're going to do this, we would need to have the restrictions you mention (`--concurrency=1` and `--cache=false` or set to false in `turbo.json`) which amounts to: 1 task at a time, and don't write to the cache. I think we can eventually relax these restrictions, but doing so has implementation and scheduling concerns. Namely, what to do with two potentially-concurrent tasks that both want `stdin`. But we don't have to tackle that to get something out.

So that brings us to the syntax. I *think* the proposal for adding `forwardStdin: true` to the task in `turbo.json` would work, since it's a property of the task being run.

dgarciamuria:
Today I run into an use case in which maybe having `cache: true` as optional would be helpful. In our workflow we leverage git-secret to store encrypted secrets and sometimes you get asked for the passphrase to decrypt them, but the password it self wont change the output of the command, it either runs or fails. Does it make sense?

macseem:
[...]

PR #6808, opened 2023-12-15
nothing to see here yet

mehulkar:
Closes TURBO-1943


weyert:
Does this mean support for tty/virtual keyboard is getting added?

mehulkar:
@weyert I'd love for you to add your use cases for tty/virtual keyboard in #1235! (this PR might get thrown away, so I want to make sure I don't lose your comments!)

PR #6809, opened 2023-12-15
refactor: Remove bookkeeping abstraction from turbo config

NicholasLYang:
### Description

I removed the bookkeeping abstraction that we ported from Go. It wasn't necessary since we can use `Option` to detect if a field is explicitly set versus not. Also renamed `RawTurboJSON` to be `RawTurboJson` to fit with the other names.

### Testing Instructions

The existing tests should suffice.

Closes TURBO-1945

### DIFF
--- a/crates/turborepo-lib/src/task_graph/mod.rs
+++ b/crates/turborepo-lib/src/task_graph/mod.rs
@@ -1,6 +1,6 @@
 mod visitor;
 
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
 use turbopath::{AnchoredSystemPath, AnchoredSystemPathBuf, RelativeUnixPathBuf};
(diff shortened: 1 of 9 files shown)


crates/turborepo-lib/src/task_graph/mod.rs
```diff
@@ -1,32 +1,15 @@
...
     run::task_id::{TaskId, TaskName},
 };
 
-pub type Pipeline = HashMap<TaskName<'static>, BookkeepingTaskDefinition>;
-
-#[derive(Clone, Debug, Default, Serialize, PartialEq, Eq)]
-pub struct BookkeepingTaskDefinition {
-    pub defined_fields: HashSet<String>,
-    pub experimental_fields: HashSet<String>,
-    pub experimental: TaskDefinitionExperiments,
```
mehulkar, line 20: we don't have any use cases for this yet, but it was a handy construct for experimental fields that we don't want to end up in `turbo prune` output and also not in dry run / `--summary` output.

mehulkar (approved):
Seems good if tests are passing!

crates/turborepo-lib/src/task_graph/mod.rs
NicholasLYang: We can probably do `#[serde(skip)]` for those

PR #6810, opened 2023-12-15
feat: Use biome to parse turbo.json

NicholasLYang:
### Description

Uses biome to parse our turbo.json files. Specifically it uses `biome_deserialize` instead of serde, which allows us to attach span info. The PR converts the env variable prefix error as a proof of concept, but of course, more errors can be converted down the line.

Also factors out turbo.json into a separate module since it's a little different than the rest of config.

### Testing Instructions

We have fairly good testing for turbo.json parsing already. I added a test for the env var prefix error.

Closes TURBO-1946


crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -100,27 +157,27 @@ impl DerefMut for Pipeline {
     }
 }
 
-#[derive(Serialize, Deserialize, Default, Debug, PartialEq, Clone)]
+#[derive(Serialize, Default, Debug, PartialEq, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct RawTaskDefinition {
     #[serde(skip_serializing_if = "Option::is_none")]
-    cache: Option<bool>,
+    cache: Option<Spanned<bool>>,
```
chris-olszewski, line 164: For these fields does it make sense to flip the types? e.g. `Spanned<Option<bool>>`
If a field isn't declared we still might want to to reference something like "no inputs declared in path/to/turbo.json"

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -0,0 +1,556 @@
...
+    },
+}
+
+fn print_diagnostics(diagnostics: &[biome_diagnostics::Error], color: bool) {
+    let color_mode = if color {
+        ColorMode::Enabled
+    } else {
+        ColorMode::Disabled
+    };
+    let mut console = EnvConsole::new(color_mode);
```
chris-olszewski, line 35: Can we use biome without a new console/color mode abstraction? Is there a path to either swap all of our current usage to this or vice versa?

I'd rather not have multiple console/color abstractions in the codebase.

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -819,20 +914,23 @@ mod tests {
...
-        let json: Result<RawTurboJson, _> = serde_json::from_value(serde_json::json!({
+        let json: Result<RawTurboJson, _> = RawTurboJson::parse_from_serde(json!({
             "pipeline": {
                 "build": {
                     "outputMode": output_mode,
                 }
             }
         }));
 
+        println!("{:?}", json);
```
chris-olszewski, line 925:
```suggestion
```

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -0,0 +1,556 @@
...
+        name: &str,
+        diagnostics: &mut Vec<DeserializationDiagnostic>,
+    ) -> Option<Self> {
+        value.deserialize(RawTaskDefinitionVisitor, name, diagnostics)
+    }
+}
+
+struct RawTaskDefinitionVisitor;
+
+impl DeserializationVisitor for RawTaskDefinitionVisitor {
```
chris-olszewski, line 108: Is there a way we could leverage a macro to reduce to amount of handwritten code for these implementations? This is a lot of handwritten code where it's really easy to mess up in some minor way e.g. forget `.with_range` on a `Spanned` value or forgetting to add a new key to `ALLOWED_KEYS`.

chris-olszewski (changes requested):
While we do have a good amount of tests for parsing `turbo.json`, I think we need some more tests verifying that the spans are getting the correct information during parsing. Will parse errors look different now? If so we should probably add a few tests to verify that they look how we expect.

crates/turborepo-lib/src/turbo_json/mod.rs
NicholasLYang: What would the span be in the case where we don't have a field?

turborepo-tests/integration/tests/config.t
```diff
@@ -42,3 +42,23 @@ Run test run with remote cache timeout env variable set
...
+
+Run build with invalid env var
+  $ EXPERIMENTAL_RUST_CODEPATH=true ${TURBO} build
+   ERROR  run failed: Environment variables should not be prefixed with "$"
+  turbo::config::invalid_env_prefix
+  
+    x Environment variables should not be prefixed with "$"
+     ,-[6:1]
+   6 |     "build": {
+   7 |       "env": ["NODE_ENV", "$FOOBAR"],
```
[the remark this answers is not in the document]
NicholasLYang: Yeah I might add that in a subsequent PR

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -0,0 +1,556 @@
...
+                .into_iter()
+                .map(|d| d.with_file_source_code(text).with_file_path(file_path))
+                .collect();
+
+            return Err(Error::Parse { diagnostics });
+        }
+
+        let mut turbo_json = result
+            .into_deserialized()
+            .expect("should have turbo.json if no errors");
```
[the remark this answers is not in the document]
NicholasLYang: So it's unclear. I'm not sure what to do here actually because we could error, but then the user will get a generic "there's a syntax error in your turbo.json" but without any actual indication of what that error is (since there's no diagnostics printed). It might be better to panic here and let the user report that as a bug so we can get an idea of when this may occur.

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -73,7 +76,61 @@ pub struct RawTurboJson {
...
+#[derive(Debug, Default, Clone, Serialize)]
+#[serde(transparent)]
+pub struct Spanned<T> {
+    value: T,
+    #[serde(skip)]
+    range: Option<Range<usize>>,
+    #[serde(skip)]
+    path: Option<Arc<str>>,
+    #[serde(skip)]
+    text: Option<Arc<str>>,
```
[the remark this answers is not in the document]
NicholasLYang: Ehh all of these are kinda similarly grouped, like text, path, and range. Each of them are optional because they get added to the data at different points and it's possible for them to be missing.

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -73,7 +76,61 @@ pub struct RawTurboJson {
     pub(crate) remote_cache: Option<ConfigurationOptions>,
 }
 
-#[derive(Serialize, Deserialize, Default, Debug, PartialEq, Clone)]
+#[derive(Debug, Default, Clone, Serialize)]
+#[serde(transparent)]
+pub struct Spanned<T> {
```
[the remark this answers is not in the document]
NicholasLYang: Eventually I might create a `turborepo-errors` so this can be shared by all the crates. I'm thinking I'll do this in a separate PR

crates/turborepo-lib/src/turbo_json/parser.rs
NicholasLYang: nvm, tackled

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -0,0 +1,576 @@
...
+    ) -> Option<Self> {
+        match String::deserialize(value, name, diagnostics)?.as_str() {
+            "full" => Some(OutputLogsMode::Full),
+            "none" => Some(OutputLogsMode::None),
+            "hash-only" => Some(OutputLogsMode::HashOnly),
+            "new-only" => Some(OutputLogsMode::NewOnly),
+            "errors-only" => Some(OutputLogsMode::ErrorsOnly),
+            unknown_variant => {
+                const ALLOWED_VARIANTS: &[&str] =
+                    &["full", "none", "hash-only", "new-only", "errors-only"];
```
chris-olszewski, line 85: I'm really concerned we're going to accidentally get these out of sync. Can we leverage [ValueEnum](https://docs.rs/clap/latest/clap/trait.ValueEnum.html) to generate this list instead of hard coding this?

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -0,0 +1,576 @@
...
+                    }
+                }
+                "enabled" => {
+                    if let Some(enabled) = bool::deserialize(&value, &key_text, diagnostics) {
+                        result.enabled = Some(enabled);
+                    }
+                }
+                unknown_key => {
+                    const ALLOWED_KEYS: &[&str] = &[
+                        "apiUrl",
```
chris-olszewski, line 336: If we're going with handwritten strings for this, then I think putting them in constants so they can be shared will at least prevent some typo-related errors.

chris-olszewski (changes requested):
I'm still fairly uncertain about the amount of code now required to parse `turbo.json` with biome, but if we're going down this path I think we should at least do our best effort to leverage the compiler to prevent us from making mistakes.

crates/turborepo-errors/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+            value,
+            range: Some(range.into()),
+            path: None,
+            text: None,
+        })
+    }
+}
+
+// We do *not* check for the range equality because that's too finicky
+// to get right in tests.
```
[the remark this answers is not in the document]
NicholasLYang: Okay a `without_range` or `strip_ranges` function makes sense. For a lot of the assert_eq tests I think we should move to snapshotting with insta because it makes updating the tests a little easier versus having to manually change code. In which case we could keep the ranges

ematipico:
> I'm still fairly uncertain about the amount of code now required to parse `turbo.json` with biome, but if we're going down this path I think we should at least do our best effort to leverage the compiler to prevent us from making mistakes.

We'll soon ship derive macros to reduce the amount of code needed to deserialize data.

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -0,0 +1,576 @@
...
+        value: &impl DeserializableValue,
+        name: &str,
+        diagnostics: &mut Vec<DeserializationDiagnostic>,
+    ) -> Option<Self> {
+        match String::deserialize(value, name, diagnostics)?.as_str() {
+            "full" => Some(OutputLogsMode::Full),
+            "none" => Some(OutputLogsMode::None),
+            "hash-only" => Some(OutputLogsMode::HashOnly),
+            "new-only" => Some(OutputLogsMode::NewOnly),
+            "errors-only" => Some(OutputLogsMode::ErrorsOnly),
```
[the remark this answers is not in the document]
NicholasLYang: Hmm so we could but it'd produce a worse error message because the serde error message doesn't include the incorrect field.

crates/turborepo-errors/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn into_inner(self) -> T {
+        self.value
+    }
+}
+
+impl<T> Spanned<Option<T>> {
+    pub fn is_none(&self) -> bool {
+        self.value.is_none()
+    }
+}
```
chris-olszewski, line 78: ~~I think the `Deref` implementation will get this us for free?~~ Nevermind, I see why this is needed.
```suggestion
```

crates/turborepo-errors/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    fn deref(&self) -> &Self::Target {
+        &self.value
+    }
+}
+
+pub trait WithRange {
+    fn add_range(&mut self, range: impl Into<Range<usize>>);
+    fn remove_range(&mut self);
+}
```
chris-olszewski, line 91: This seems unused?

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -407,7 +426,9 @@ impl TryFrom<RawTurboJson> for TurboJson {
             pipeline: raw_turbo.pipeline.unwrap_or_default(),
             // copy these over, we don't need any changes here.
             remote_cache: raw_turbo.remote_cache,
-            extends: raw_turbo.extends.unwrap_or_default(),
+            extends: raw_turbo
+                .extends
+                .unwrap_or_else(|| Spanned::new(Vec::new())),
```
chris-olszewski, line 431:
```suggestion
            extends: raw_turbo.extends.unwrap_or_default(),
```

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -753,13 +792,19 @@ mod tests {
...
+      ; "full"
     )]
     fn test_deserialize_task_definition(
         task_definition_content: &str,
         expected_raw_task_definition: RawTaskDefinition,
         expected_task_definition: TaskDefinition,
     ) -> Result<()> {
-        let raw_task_definition: RawTaskDefinition = serde_json::from_str(task_definition_content)?;
+        let deserialized_result = deserialize_from_json_str(
+            &task_definition_content,
```
chris-olszewski, line 803:
```suggestion
            task_definition_content,
```

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -0,0 +1,540 @@
...
+        value: &impl DeserializableValue,
+        name: &str,
+        diagnostics: &mut Vec<DeserializationDiagnostic>,
+    ) -> Option<Self> {
+        let output_logs_str = String::deserialize(value, name, diagnostics)?;
+        match OutputLogsMode::from_str(&output_logs_str, false) {
+            Ok(result) => Some(result),
+            Err(err) => {
+                let allowed_variants: Vec<_> = OutputLogsMode::value_variants()
+                    .into_iter()
```
chris-olszewski, line 100: As per clippy
```suggestion
                    .iter()
```

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -0,0 +1,540 @@
...
+    }
+}
+
+impl WithText for RawTurboJson {
+    fn add_text(&mut self, text: Arc<str>) {
+        self.extends.add_text(text.clone());
+        self.global_dependencies.add_text(text.clone());
+        self.global_env.add_text(text.clone());
+        self.global_pass_through_env.add_text(text.clone());
+        self.pipeline.add_text(text.clone());
```
chris-olszewski, line 467:
```suggestion
        self.pipeline.add_text(text);
```

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -0,0 +1,540 @@
...
+impl WithText for RawTaskDefinition {
+    fn add_text(&mut self, text: Arc<str>) {
+        self.depends_on.add_text(text.clone());
+        self.dot_env.add_text(text.clone());
+        self.env.add_text(text.clone());
+        self.inputs.add_text(text.clone());
+        self.pass_through_env.add_text(text.clone());
+        self.persistent.add_text(text.clone());
+        self.outputs.add_text(text.clone());
+        self.output_mode.add_text(text.clone());
```
chris-olszewski, line 488:
```suggestion
        self.output_mode.add_text(text);
```

chris-olszewski (approved):
Overall looks good to me. I'm excited for the future of better parsing

chris-olszewski (approved):

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -879,4 +902,14 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_biome() {
+        let result = deserialize_from_json_str::<String>(
+            r#""\\""#,
+            JsonParserOptions::default().with_allow_comments(),
+        );
+
+        println!("{}", result.into_deserialized().unwrap());
```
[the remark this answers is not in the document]
NicholasLYang: Oops yeah that was leftover from my experiments. Instead I buffed up the other tests with backslash versions

PR #6818, opened 2023-12-18
Fixing ESLint errors in docs. (3/n, 73% done)

anthonyshew:
## Description
3/n of my stacked diffs to address ESLint errors in docs. PRs to ship before this one:
- https://github.com/vercel/turbo/pull/6812
- #6813

PR #6813, opened 2023-12-16
Fixing ESLint errors in docs. (2/n, 47% done)

anthonyshew:
## Description
2/n of my stacked diffs to address ESLint errors in docs. PRs to ship before this one:
- https://github.com/vercel/turbo/pull/6812

Notably, this PR does contain more defensive error throwing. Thank you ESLint for making our code more defensive.

PR #6812, opened 2023-12-16
Fixing ESLint errors in docs. (1/n, 18% done)

anthonyshew:
### Description

1/n of my stacked diffs to address ESLint errors in docs.

You shouldn't see any logical code changes. Types or ESLint rule changes or reformats only.

PR #6819, opened 2023-12-18
Fixing ESLint in docs. (4/n, 89% done)

anthonyshew:
## Description
4/n of my stacked diffs to address ESLint errors in docs. PRs to ship before this one:
- https://github.com/vercel/turbo/pull/6812
- #6813
- #6818


docs/components/Tweet.tsx
anthonyshew, line 1: This file was not being used.

docs/components/HeaderLogo.tsx
```diff
@@ -32,5 +32,3 @@ function HeaderLogo() {
     </>
   );
 }
-
-export default HeaderLogo;
```
mehulkar, line 36: Very stylistic so I shouldn't care, but I like this change. I always miss it when there all the exports are at the bottom (some people really like this style and that's totally fine too)

docs/components/Navigation.tsx
```diff
@@ -1,7 +1,8 @@
+/* eslint-disable -- Lots of Nextra magic. */
```
mehulkar, line 1: this feels like a prettttttty big hammer. is it worth it to list out all the rules instead?

mehulkar (approved):
👏🏾 👏🏾 👏🏾

tknickman (approved):

docs/components/HeaderLogo.tsx
anthonyshew: CleanShot 2023-12-20 at 11 25 28@2x

docs/components/Navigation.tsx
anthonyshew: Oh, yup, good call.

PR #6821, opened 2023-12-18
Only warn for ESLint issues in editor.

anthonyshew:
### Description

I find better DX with TypeScript errors being red squigglies and ESLint errors/warnings being yellow squigglies.

This comes with the tradeoff that what you see in your editor might differ from what CI sees. However, any ESLint issue should block CI anyway so I find it doesn't tend to matter what color they are in your editor anyway.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,6 +6,7 @@
     "typescriptreact"
   ],
   "eslint.workingDirectories": [{ "mode": "auto" }],
+  "eslint.rules.customizations": [{ "rule": "*", "severity": "warn" }],
   "debug.javascript.unmapMissingSources": true,
   "go.lintTool": "golangci-lint",
   "go.buildTags": "rust",
@@ -21,5 +22,5 @@
   },
   "search.exclude": {
     "crates/turbopack-tests/tests/snapshot/**": true
-  },
+  }
 }


tknickman (changes requested):
Not a fan of this personally. Red is error, yellow is warn. 

I prefer the feedback of what will be an error in all environments, in my editor, immediately. 

Also, eslint issues _do_ block CI for our packages.

anthonyshew:
Noice. Docs issues will be blocking now too.

I guess the idea is: All squigglies are blocking. Red is TypeScript. Yellow is ESLint.

I didn't write that very well in my original description.

anthonyshew:
And, yes, I can see this being preferential. I was curious to see if others would like/not like this. 😄

tknickman:
👍🏼 I see what you're saying, but I think this is more of a preferential thing that we may not want to commit - I think using the underline color to indicate the _source_ instead of _severity_ is pretty non-standard and may be a little confusing for those used to the other way.  

I may be wrong here though, we can poll the team. But I think the yellow / warn and red / error visual queue is pretty hardcoded into our brains.

mehulkar (commented):
I'm with Tom and I would be confused by this change. We should rationalize our TS and ESlint setup so that there isn't duplication in the things they check, but I'd still like red to mean error and yellow to mean warn.

anthonyshew:
Fine by me! I'll close this.

PR #6832, opened 2023-12-20
fix(turbopack-ecmascript): make sure async module wrapper is always generated

ForsakenHarmony:
### Description
The esm check for `EcmascriptModuleContent` was not the same as the check to add async modules during analysis.
This makes sure both are the same.

Specifically `import *` was not checked and the `SpecifiedModuleType` (i.e. `.mjs`) was also ignored.

Closes PACK-2170


crates/turbopack-core/src/ident.rs
```diff
@@ -61,8 +61,18 @@ impl ValueToString for AssetIdent {
             write!(s, "#{}", fragment.await?)?;
         }
 
-        for (key, asset) in &self.assets {
-            write!(s, "/({})/{}", key.await?, asset.to_string().await?)?;
```
ForsakenHarmony, line 65: I found this super confusing whenever I was looking at it

crates/turbopack-ecmascript/src/analyzer/imports.rs
```diff
@@ -123,7 +123,7 @@ pub(crate) struct ImportMapReference {
 
 impl ImportMap {
     pub fn is_esm(&self) -> bool {
-        self.has_exports || !self.imports.is_empty()
+        self.has_exports || !self.imports.is_empty() || !self.namespace_imports.is_empty()
```
ForsakenHarmony, line 126: I think this was the main issue

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -741,7 +799,8 @@ async fn gen_content_with_visitors(
         Ok(EcmascriptModuleContent {
             inner_code: bytes.into(),
             source_map: Some(Vc::upcast(srcmap)),
-            is_esm: eval_context.is_esm(),
+            is_esm: eval_context.is_esm()
+                || specified_module_type == SpecifiedModuleType::EcmaScript,
```
ForsakenHarmony, line 803: not sure if this is the best way to pass the `SpecifiedModuleType` into this, maybe it should be added to the `eval_context`? 🤔

sokra: Yep, that would be better I guess

crates/turbopack-core/src/ident.rs
sokra: Sounds good, but could you move that into a separate PR as it affects all snapshots

crates/turbopack-ecmascript/src/lib.rs
ForsakenHarmony: But then we need to pass in information at parse time which isn't needed to parse 🤔

sokra: Don't we need `specified_module_type` at parse time to determine the exports type in the case when no commonjs is detected.

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -373,6 +409,45 @@ impl EcmascriptModuleAsset {
...
+            }
+            SpecifiedModuleType::Automatic => {}
+        }
+
+        let find_package_json = find_context_file(self.origin_path(), package_json()).await?;
+        let FindContextFileResult::Found(package_json, _) = *find_package_json else {
+            return Ok(ModuleTypeResult::new(SpecifiedModuleType::Automatic));
+        };
+
+        // analysis.add_reference(PackageJsonReference::new(package_json));
```
sokra, line 431: Is this handled?

sokra (approved):

PR #6836, opened 2023-12-20
Adding Vercel Toolbar to docs.

anthonyshew:
### Description

To create a better feedback loop with the community, we can enable comments on our documentation.

This PR enables the Vercel Toolbar for local branches and production. On previews, the toolbar will always be present.

In a typical product, you wouldn't display your toolbar for any site visitor to use on production. However, I'm doing an overhaul of our documentation so I'm interested in collecting as much feedback as possible. Depending on the quantity and quality of comments we get, we may or may not choose to keep Toolbar active on production long term.

### Testing Instructions

Click the new button in the top right of the preview to enable the Vercel Toolbar. Leave a comment and resolve it! Fun!

Note that this may be awkward to test on a preview deployment since previews always have the toolbar. You will likely want to take a look at the behavior locally by pulling the branch and using `cd docs && pnpm build && pnpm start`.


docs/components/CommentsButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+import cn from "classnames";
+import { PlusSmIcon } from "@heroicons/react/outline";
+import { getCommentsState, setCommentsState } from "../lib/comments";
+
+export function CommentsButton() {
+  const comments = getCommentsState();
+  const router = useRouter();
+
+  return (
+    // eslint-disable-next-line react/button-has-type -- Can't set it here because it messes up background color. There's a style that sets the background to transparent that I believe is coming from deep within Nextra.
```
mehulkar, line 11: need permallink to the line of code it comes from so we can know whether this disable is stale the next time someone comes across this code

mehulkar (changes requested):
I'm in favor of this, but have some feedback:

- Can we disable it in on local branches? That doesn't seem like it will be helpful? (Not sure where the comments would go in that case too)
- Can we limit to just `/repo/docs/*` ? (`/pack/docs/*` is fine with me too)

docs/components/CommentsButton.tsx
mehulkar: ^ this is may main blocking comment, fwiw, not the top level ones. We shouldn't introduce disables without knowing how/when to enable again

anthonyshew:
> I'm in favor of this, but have some feedback:
> 
> * Can we disable it in on local branches? That doesn't seem like it will be helpful? (Not sure where the comments would go in that case too)
> * Can we limit to just `/repo/docs/*` ? (`/pack/docs/*` is fine with me too)

1) It's actually quite useful on local branches! You can work with comments that were made on previews directly on your locally running branch.
2) I just asked the Toolbar team and they said they can ship this capability soon!

docs/components/CommentsButton.tsx
anthonyshew: Oops, had it in there originally and must have lost it while I was doing science. Thanks, good catch!

mehulkar (approved):

anthonyshew:
Good news! The Toolbar team is awesome and shipped changes for us to get that path-bounded rendering. 🥳

Cargo.lock
mehulkar, line 1: I'm guessing this happened automatically? Wonder if that's a problem? cc @chris-olszewski

mehulkar (approved):

PR #6838, opened 2023-12-20
fix: Allow for long symlinks by using append_link

NicholasLYang:
### Description

We were using `Header::set_link_name` instead of using `Builder::append_link`, which handles long link names correctly. In general we should avoid using methods on `Header` and instead use the ones on `Entry`.

### Testing Instructions

Modifies the long name test to include a symlink with a long link name.

Closes TURBO-1968


gsoltis (approved):
Just a request for a comment, assuming that's all good,then lgtm

PR #6839, opened 2023-12-20
fix glob matching of alternatives

sokra:
### Description

The unicode refactoring introduced some bugs in matching of alternatives

Closes PACK-2172

### DIFF
--- a/crates/turbo-tasks-fs/src/glob.rs
+++ b/crates/turbo-tasks-fs/src/glob.rs
@@ -159,6 +159,7 @@ impl GlobPart {
             match_partial,
             previous_part_is_path_separator_equivalent,
             cursor: GraphemeCursor::new(0, path.len(), true),
+            index: 0,
             glob_iterator: None,
         }
     }
@@ -257,6 +258,7 @@ struct GlobPartMatchesIterator<'a> {
     match_partial: bool,
     previous_part_is_path_separator_equivalent: bool,
     cursor: GraphemeCursor,
+    index: usize,
     glob_iterator: Option<Box<GlobMatchesIterator<'a>>>,
 }
 
@@ -267,12 +269,9 @@ impl<'a> Iterator for GlobPartMatchesIterator<'a> {
         match self.part {
             GlobPart::AnyDirectories => {
                 if self.cursor.cur_cursor() == 0 {
-                    let end = self.cursor.next_boundary(self.path, 0);
-                    match end {
-                        Ok(Some(_)) => {}
-                        Ok(None) => return None,
-                        Err(..) => return None,
-                    }
+                    let Ok(Some(_)) = self.cursor.next_boundary(self.path, 0) else {
+                        return None;
+                    };
                     return Some((self.path, true));
                 }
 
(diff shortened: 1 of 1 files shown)


crates/turbo-tasks-fs/src/glob.rs
```diff
@@ -347,28 +340,25 @@ impl<'a> Iterator for GlobPartMatchesIterator<'a> {
...
-                let end = match self.cursor.next_boundary(self.path, 0) {
-                    Ok(Some(end)) => end,
-                    _ => return None,
+            GlobPart::FileChar(chars) => {
+                let start = self.cursor.cur_cursor();
+                let Ok(Some(end)) = self.cursor.next_boundary(self.path, 0) else {
+                    return None;
                 };
-
-                let c = &chars[self.cursor.cur_cursor()..end]
```
sokra, line 356: This is unused, but it doesn't really make sense to use the path cursor to index `chars`

crates/turbo-tasks-fs/src/glob.rs
```diff
@@ -379,15 +369,10 @@ impl<'a> Iterator for GlobPartMatchesIterator<'a> {
...
-                        let end = self.cursor.next_boundary(self.path, 0);
+                        self.index += 1;
                         self.glob_iterator = None;
-                        match end {
-                            Ok(Some(_)) => {}
-                            Ok(None) => return None,
-                            Err(..) => return None,
-                        }
                     }
-                } else if let Some(alternative) = alternatives.get(self.cursor.cur_cursor()) {
```
sokra, line 390: path cursor can't be used to index alternatives list

crates/turbo-tasks-fs/src/glob.rs
```diff
@@ -267,12 +269,9 @@ impl<'a> Iterator for GlobPartMatchesIterator<'a> {
...
                 if self.cursor.cur_cursor() == 0 {
-                    let end = self.cursor.next_boundary(self.path, 0);
-                    match end {
-                        Ok(Some(_)) => {}
-                        Ok(None) => return None,
-                        Err(..) => return None,
-                    }
+                    let Ok(Some(_)) = self.cursor.next_boundary(self.path, 0) else {
+                        return None;
+                    };
```
sokra, line 274: This is just shorter and cleaner

kwonoj: yes I'm always forgetting there's let..else syntax

crates/turbo-tasks-fs/src/glob.rs
```diff
@@ -457,19 +442,43 @@ mod tests {
...
+         sharp,shiki,sqlite3,tailwindcss,ts-node,typescript,vscode-oniguruma,webpack,websocket,@\
+         aws-sdk/client-dynamodb,@aws-sdk/lib-dynamodb}/**",
+        "node_modules/.pnpm/@aws-sdk+lib-dynamodb@3.445.0_@aws-sdk+client-dynamodb@3.445.0/\
+         node_modules/@aws-sdk/lib-dynamodb/dist-es/index.js"
+    )]
     #[case::alternatives_nested1("{a,b/c,d/e/{f,g/h}}", "a")]
     #[case::alternatives_nested2("{a,b/c,d/e/{f,g/h}}", "b/c")]
     #[case::alternatives_nested3("{a,b/c,d/e/{f,g/h}}", "d/e/f")]
     #[case::alternatives_nested4("{a,b/c,d/e/{f,g/h}}", "d/e/g/h")]
+    // #[case::alternatives_chars("[abc]", "b")]
```
kwonoj, line 481: is this inteded to be commented?

sokra: This isn't supported in the glob parsing yet. We might want to uncomment that test case once we implement that.

kdy1 (approved):

PR #6845, opened 2023-12-21
perf: Use `&'static str` instead of `String` if possible

kdy1:
### Description

All calls are generated by the proc-macro, so we don't need to allocate.

### Testing Instructions

Closes PACK-2177


ForsakenHarmony (approved):
don't think this will realistically have a measurable impact on performance, but it seems like a good idea in general

PR #6846, opened 2023-12-21
fix(turbopack): Apply aliases for `react` and `react-dom` for client pages

kdy1:
### Description

This change is required for https://github.com/vercel/next.js/pull/59802

### Testing Instructions

Let's look at the CI of next.js

Closes PACK-2178


sokra (changes requested):
That's looks really next.js specific and should live in next.js instead

PR #6849, opened 2023-12-23
fix(docs): mobile screen, prevent gradient and hero overflow

jupapios:
### Description

https://turbo.build/pack and https://turbo.build/repo are overflowing to the right on mobile screens.

- The `Gradient` component was causing screens with a width less than the `Gradient`'s `width` to overflow, here I let the `Gradient` component handle the case with a max width that will resize if the device's screen width is less than the requested one, behaving similarly to how images already resize by default with Tailwind. Another alternative is to let it overflow and wrap it to hide the overflow and center it, I think this is more a design choice than a technical one.
- The hexagon and circle images had a similar problem. Removing the explicit width and height of the container and letting the images use their `width` and `height`, and Tailwind's max width kick in fixes the overflow.

### Testing Instructions

- Run the docs.
- Using a mobile screen navigate to `/pack` and `/repo`
- Confirm it does not overflow to the right, it should look something like,

### DIFF
--- a/docs/components/pages/home-shared/Gradient.tsx
+++ b/docs/components/pages/home-shared/Gradient.tsx
@@ -25,7 +25,7 @@ export function Gradient({
   return (
     <span
       className={cn(
-        "absolute",
+        "absolute max-w-full rounded-full",
         gradients.glow,
         {
           [gradients.glowPink]: pink,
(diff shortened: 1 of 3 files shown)


anthonyshew (approved):
Oops! This was me. I checked responsiveness on a shrunken Chrome window on desktop but...looks like I was lied to. 😄

Thanks for this!

PR #6859, opened 2023-12-28
fix(examples): Typo correction on design-system package.json

GGalupo:
### Description

Typo fix on package.json from design-system example.

The file `storybook.js` was referenced as `storybook.j` on files field.


anthonyshew:
Thanks!

PR #6866, opened 2024-01-02
trace memory allocations and show in tracing

sokra:
### Description

* collect memory allocation data and report it when tracing is enable
* Add value mode to trace server to show memory info

Closes PACK-2184


crates/turbo-tasks-malloc/src/counter.rs
```diff
@@ -13,16 +13,37 @@ const TARGET_BUFFER: usize = 100 * KB;
 /// global counter.
 const MAX_BUFFER: usize = 200 * KB;
 
+#[derive(Default)]
+pub struct AllocationInfo {
+    pub allocations: usize,
```
kwonoj, line 18: is usize enough to not overflow?

kwonoj (approved):

PR #6867, opened 2024-01-02
Use the sso team instead of the string 'sso'

Zertsov:
Bit of an oops here. Use the variable containing the sso team instead of the raw "sso" string.

### Testing Instructions
Try logging in with SSO, should use the team passed in

Closes TURBO-1974


chris-olszewski (approved):

tknickman (approved):

mehulkar (approved):
how come tests didn't fail?

Zertsov:
@mehulkar 
> how come tests didn't fail?

I don't think there's good enough SSO tests. Definitely something to fix up in the near future.

PR #6870, opened 2024-01-02
Make link command use auth crate instead of config token

Zertsov:
### Description
`turbo link` was using the `BaseCommand` config token which exclusively reads from `config.json`, instead of using the new auth crate to find the token. This PR fixes that and uses the auth crate to find the token.

### Testing Instructions
Ensure you don't have a `config.json` with a token in it in whatever `XDG_CONFIG_HOME` you have (where we keep our tokens), run a `turbo login` and/or `turbo login --sso-team <team>`, then run the `turbo link` command. It should properly link.

After that, to double make sure, go to the current version of the code (in `main`) and try the same. If you have the `auth.json` that we create, with a valid token, but no `config.json`, then `turbo link` should fail.

Closes TURBO-1976


chris-olszewski (changes requested):
Gets us into a better state, but we need to make sure the correct token gets used in `run` and other code paths that require auth.

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -2,9 +2,9 @@ use std::cell::OnceCell;
 
 use sha2::{Digest, Sha256};
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
-use turborepo_api_client::{APIAuth, APIClient};
+use turborepo_api_client::{APIAuth, APIClient, Client};
 use turborepo_auth::{
-    TURBOREPO_AUTH_FILE_NAME, TURBOREPO_CONFIG_DIR, TURBOREPO_LEGACY_AUTH_FILE_NAME,
+    AuthFile, TURBOREPO_AUTH_FILE_NAME, TURBOREPO_CONFIG_DIR, TURBOREPO_LEGACY_AUTH_FILE_NAME,
```
chris-olszewski, line 7:
```suggestion
   TURBOREPO_AUTH_FILE_NAME, TURBOREPO_CONFIG_DIR, TURBOREPO_LEGACY_AUTH_FILE_NAME,
```

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -125,15 +125,27 @@ impl CommandBase {
...
         let team_slug = config.team_slug();
 
-        let Some(token) = config.token() else {
-            return Ok(None);
+        let auth_file_path = self.global_auth_path()?;
+        let config_file_path = self.global_config_path()?;
+        let client = self.api_client()?;
+        let auth = match turborepo_auth::read_auth_file(&auth_file_path) {
+            Ok(auth) => auth,
+            Err(_) => match turborepo_auth::read_auth_file(&config_file_path) {
```
chris-olszewski, line 133: I think this always fails. We're trying to deserialize the old `config.json` as an `auth.json` and it'll be missing the `tokens` field.

crates/turborepo-auth/src/lib.rs
```diff
@@ -89,6 +89,30 @@ pub async fn read_or_create_auth_file(
     Ok(auth_file)
 }
 
+/// Sync version of `read_or_create_auth_file`.
```
chris-olszewski, line 92: I think the comment needs to be updated as there are behavior differences between this and `read_or_create_auth_file` (and also they're both now sync).

crates/turborepo-auth/src/lib.rs
```diff
@@ -89,6 +89,30 @@ pub async fn read_or_create_auth_file(
...
+            .map_err(|e| Error::FailedToReadAuthFile {
+                source: e,
+                path: auth_file_path.to_owned(),
+            })?;
+        let tokens: AuthFile = serde_json::from_str(&content)
+            .map_err(|e| Error::FailedToDeserializeAuthFile { source: e })?;
+        let mut auth_file = AuthFile::new();
+        for (api, token) in tokens.tokens() {
+            auth_file.insert(api.to_owned(), token.to_owned());
+        }
```
chris-olszewski, line 106: I don't understand why we're rebuilding the auth file here

chris-olszewski (changes requested):
Primarily concerned with not being able to use the old tokens

crates/turborepo-auth/src/lib.rs
Zertsov: Lazy copy-paste job - I'll fix that in a moment.

crates/turborepo-lib/src/commands/mod.rs
Zertsov: Yeah, I can just use the existing read function now that it's sync. I'll clean up.

crates/turborepo-auth/src/auth_file.rs
```diff
@@ -88,10 +87,10 @@ impl AuthToken {
 }
 
 /// Converts our old style of token held in `config.json` into the new schema.
-pub fn convert_to_auth_token(token: &str, client: &impl Client) -> AuthToken {
+pub fn convert_to_auth_token(token: &str, api: &str) -> AuthToken {
```
chris-olszewski, line 90: For a future PR this makes more sense as `AuthToken::new`

chris-olszewski (approved):

chris-olszewski (changes requested):
Test failures look legit. It looks like we're not respecting `--token` during `run` now.

chris-olszewski (approved):

PR #6880, opened 2024-01-03
fix: Make execution state use auth crate

Zertsov:
### Description
How about I do a PR this time?
This updates the execution state to use the `turborepo_auth` crate, beginning the end of the `CommandBase` as the token holder.

### Testing Instructions
Run a command with the `--go-fallback`. This should pass.

Closes TURBO-1987

### DIFF
--- a/crates/turborepo-auth/src/lib.rs
+++ b/crates/turborepo-auth/src/lib.rs
@@ -16,7 +16,6 @@ mod sso_server;
 mod ui;
 
 use turbopath::AbsoluteSystemPath;
-use turborepo_api_client::Client;
 
 pub use self::{
     auth_file::*, config_token::ConfigToken, error::Error, login::*, login_server::*, logout::*,
@@ -47,7 +46,7 @@ pub const DEFAULT_API_URL: &str = "https://vercel.com/api";
 pub fn read_or_create_auth_file(
     auth_file_path: &AbsoluteSystemPath,
     config_file_path: &AbsoluteSystemPath,
-    client: &impl Client,
+    api: &str,
 ) -> Result<AuthFile, Error> {
     if auth_file_path.try_exists()? {
         let content = auth_file_path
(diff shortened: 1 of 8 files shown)


crates/turborepo-auth/src/lib.rs
```diff
@@ -112,7 +113,7 @@ mod tests {
 
         let client = MockApiClient::new();
```
chris-olszewski, line 114: For future PR, but we can get rid of the mock client usage in this test now.

crates/turborepo-lib/src/execution_state.rs
```diff
@@ -49,8 +49,22 @@ impl<'a> TryFrom<&'a CommandBase> for ExecutionState<'a> {
 
         let config = base.config()?;
 
+        // TODO(voz): Don't unwrap here. Should be fine since this can rarely ever be
+        // None.
+        let auth_file_path = base.global_auth_path.clone().unwrap();
+        let config_file_path = base.global_config_path.clone().unwrap();
```
chris-olszewski, line 55: I'm pretty sure you want 
```suggestion
        let auth_file_path = base.global_auth_path().unwrap();
        let config_file_path = base.global_config_path().unwrap();
```
as the fields themselves will always be `None` except for in tests.

We also can't be unwrapping like this and should instead go through normal error handling if there's a config error.

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -30,9 +30,7 @@ pub(crate) mod unlink;
 pub struct CommandBase {
     pub repo_root: AbsoluteSystemPathBuf,
     pub ui: UI,
-    #[cfg(test)]
     pub global_config_path: Option<AbsoluteSystemPathBuf>,
-    #[cfg(test)]
```
chris-olszewski, line 35: I don't understand why we're removing the `#[cfg(test)]` here. They should be added back

chris-olszewski (changes requested):
Concerned about the making of test only fields now available in prod code.

crates/turborepo-lib/src/execution_state.rs
Zertsov: I was doing the `clone` due to move issues, but I can just make those methods public instead.

I'll handle unwraps

crates/turborepo-lib/src/commands/mod.rs
Zertsov: I can make the getters public, but this was to allow for the execution state to know where the auth crate should look for tokens

crates/turborepo-lib/src/execution_state.rs
```diff
@@ -49,8 +49,22 @@ impl<'a> TryFrom<&'a CommandBase> for ExecutionState<'a> {
 
         let config = base.config()?;
 
+        // TODO(voz): Don't unwrap here. Should be fine since this can rarely ever be
```
Zertsov, line 52: I'll remove this after the Github Actions finish - don't wanna restart them now

chris-olszewski (approved):
LGTM

PR #6887, opened 2024-01-04
feat(launch): update vscode launch configs

tknickman:
### Description

Update the vscode launch configs

1. Update the version to get access to user input and grouping
2. Remove old Go and e2e configs
3. Fix the existing configs to user skip-infer to ensure we're using the local built turbo and not the one installed in the example
4. Cleans up unused actions

Closes TURBO-1991


chris-olszewski (approved):
Thanks for reminding me that we can easily use the debugger now that we're full Rust.

PR #6892, opened 2024-01-04
Fix artifact http request errors via cache bump

Zertsov:
### Description
[This](https://github.com/vercel/turbo/issues/6765) issue is reproducible and fixed by bumping the global cache key.

### Testing Instructions
1. Login via `turbo login`, go through the login process, then link a repo if you haven't already.
2. Ensure there's a clean cache (nothing in there)
3. `pnpm i turbo@1.10.16 --global`
4. Run a `turbo` task, like `turbo lint` or `turbo build`
5. See that remote cache was written to
6. Try a `turbo <task you just ran> --remote-only` to make sure the artifact warning isn't appearing
7. Now build this PR's version of `turbo` (`cd /path/to/local-turbo/cli && make turbo`)
8. Go back to whatever repo you were previously testing in and run the same task
9. Profit (a.k.a it doesn't send the artifact warning)


Zertsov:
I had this reproducing at one point, but I can't seem to get it to do so now, even without this PR...don't merge this until myself or someone else can reliably reproduce the issue.

chris-olszewski:
Closing this as we're pretty sure the cache issues are not caused by differences in Rust/Go.

PR #6927, opened 2024-01-05
add better messaging when running with api flag

Zertsov:
### Description
This is a pretty easy thing to miss. When running tasks with the `--api` flag or an api environment variable, if you're logged into a different api (like `vercel`), there's not super clear messaging on how you might fix it. This should address it when talking to the Spaces client

### Testing Instructions
1. Do a login (Vercel preferable)
2. Run a `turbo build --filter docs --api fake-server.some.stuff`
3. See the edited message
4. Profit

### DIFF
--- a/crates/turborepo-lib/src/run/summary/spaces.rs
+++ b/crates/turborepo-lib/src/run/summary/spaces.rs
@@ -148,9 +148,16 @@ impl SpacesClient {
         let space_id = space_id?;
         let is_linked = api_auth.as_ref().map_or(false, |auth| auth.is_linked());
         if !is_linked {
+            let base = api_client.base_url();
+            let login_command = if base.contains("vercel") {
+                "turbo login".to_string()
+            } else {
+                format!("turbo login --api {}", base)
+            };
+
             eprintln!(
-                "Error: experimentalSpaceId is enabled, but repo is not linked to API. Run `turbo \
-                 link` or `turbo login` first"
+                "Error: experimentalSpaceId is enabled, but repo is not linked to {base}. Run \
+                 `turbo link` or `{login_command}` first",
             );
             return None;
         }


crates/turborepo-lib/src/run/summary/spaces.rs
```diff
@@ -148,9 +148,16 @@ impl SpacesClient {
         let space_id = space_id?;
         let is_linked = api_auth.as_ref().map_or(false, |auth| auth.is_linked());
         if !is_linked {
+            let base = api_client.base_url();
+            let login_command = if base.contains("vercel") {
+                "turbo login".to_string()
+            } else {
+                format!("turbo login --api {}", base)
+            };
```
chris-olszewski, line 156: I believe this should happen in `CommandBase::api_auth` as there we have more information such as what APIs have logins associated with them.

This warning will also only show if spaces are enabled which most projects (at the moment) do not have toggled on.

chris-olszewski (changes requested):
Primarily concerned with this warning only showing if spaces are enabled

crates/turborepo-lib/src/run/summary/spaces.rs
Zertsov: Fair point, this was where all my debugging led me. I'll switch up the wording in the `api_auth` but keep the same concept.

chris-olszewski (approved):

Jengstrand1 (commented):
This is amazing code, keep up the good work Mitch. :)

crates/turborepo-lib/src/run/summary/spaces.rs
Zertsov: Moving this instead to the run where it'll only run on a linked project.

PR #6928, opened 2024-01-05
Only run examples tests on ubuntu.

anthonyshew:
### Description

We're opting to drop the macOS runners. We were getting negative value out of them on the examples/documentation side and we don't believe they are contributing to core development stability.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -417,15 +417,7 @@ jobs:
     if: needs.determine_jobs.outputs.examples == 'true'
     timeout-minutes: 40
 
-    strategy:
-      fail-fast: false
-      matrix:
-        os:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -417,15 +417,7 @@ jobs:
...
     timeout-minutes: 40
 
-    strategy:
-      fail-fast: false
-      matrix:
-        os:
-          - runner: ubuntu-latest
-          - runner: macos-latest
-          # TODO: enable this
-          # - runner: windows-latest
```
mehulkar, line 427: I'll close https://github.com/vercel/turbo/pull/6801 also

mehulkar (approved):

tknickman (approved):

PR #6956, opened 2024-01-08
refactor: Deleting Go code

NicholasLYang:
### Description

LFG

### Testing Instructions

Closes TURBO-2011


mehulkar (approved):

crates/turborepo-lib/src/lib.rs
```diff
@@ -62,38 +51,6 @@ pub fn get_version() -> &'static str {
...
-        Err(err) => {
-            // This raw print matches the Go behavior, once we no longer care
-            // about matching formatting we should remove this.
-            println!("Turbo error: {err}");
-
-            Payload::Rust(Err(err))
-        }
-    }
+pub fn main() -> Result<i32, shim::Error> {
+    shim::run()
```
chris-olszewski, line 55: I think we should avoid avoidable behavior changes in this PR and keep it to just deleting Go.

crates/turborepo/Cargo.toml
```diff
@@ -36,6 +36,7 @@ clap_complete = { workspace = true }
 command-group = { version = "2.0.1", features = ["with-tokio"] }
 dunce = { workspace = true }
 human-panic = "1.2.1"
+miette.workspace = true
```
chris-olszewski, line 39: We should get rid of `go-binary` and `go-daemon` features (Line not relevant, just file)

crates/turborepo-lib/src/config.rs
```diff
@@ -23,7 +23,7 @@ pub enum Error {
     #[error(transparent)]
     PackageJson(#[from] turborepo_repository::package_json::Error),
     #[error(
-        "Could not find turbo.json. Follow directions at https://turbo.build/repo/docs to create \
+        "Could not find turbo.json.\nFollow directions at https://turbo.build/repo/docs to create \
```
chris-olszewski, line 26: This feels unrelated to deleting Go

.github/workflows/turborepo-release.yml
```diff
@@ -320,40 +214,6 @@ jobs:
...
-
-          mv go-artifacts/turbo-go-darwin-${{ needs.stage.outputs.stage-branch }}/turbo_darwin_amd64_v1/bin/* cli/dist-darwin-amd64
-          chmod a+x cli/dist-darwin-amd64/turbo
-          chmod a+x cli/dist-darwin-amd64/go-turbo
-          mv go-artifacts/turbo-go-darwin-${{ needs.stage.outputs.stage-branch }}/turbo_darwin_arm64/bin/* cli/dist-darwin-arm64
-          chmod a+x cli/dist-darwin-arm64/turbo
-          chmod a+x cli/dist-darwin-arm64/go-turbo
-
       - name: Perform Release
         run: cd cli && make publish-turbo SKIP_PUBLISH=${{ inputs.dry_run && '--skip-publish' || '' }}
```
chris-olszewski, line 218: I'm pretty sure this `publish-turbo` target expects `go-turbo` to exists and needs updating. Can you verify that `make publish-turbo SKIP_PUBLISH='--skip-publish'` gets to the point where it would start publishing?

chris-olszewski (changes requested):

crates/turborepo-lib/src/config.rs
NicholasLYang: This was due to an issue with error printing and regexes. Miette line breaks at 80 characters and adds a pipe character. By line breaking early, we avoid the pipe which makes it easier to match on the text. Not needed now if we're stripping out the miette changes for a later PR

.github/workflows/turborepo-release.yml
NicholasLYang: It seems to work, at least it fails at the same point as on main

.github/actions/setup-go/action.yml
```diff
@@ -1,37 +0,0 @@
...
-
-runs:
-  using: "composite"
-  steps:
-    - name: Set up Go
-      uses: actions/setup-go@v3
-      with:
-        go-version: "1.20"
-        cache: true
-        cache-dependency-path: cli/go.sum
```
chris-olszewski, line 16: It might make for a smaller/eassier diff if we just removed the "setup Go" & "Setup GRPC protobuf" steps from this action instead of canning the action all together.

chris-olszewski:
I tried to run the a dry release and it looks like there are some more updates that need to happen: https://github.com/vercel/turbo/actions/runs/7703913529/job/20995209458

chris-olszewski:
I think this line needs to get cut: https://github.com/vercel/turbo/blob/main/cli/scripts/npm-native-packages/npm-native-packages.js#L54

cli/package.json
mehulkar, line 1: we're keeping `cli/` as a package?

mehulkar (approved):
🚀

.github/workflows/bench-turbopack.yml
```diff
@@ -139,11 +139,6 @@ jobs:
           components: clippy
           targets: wasm32-unknown-unknown
 
-      - name: Setup Go
```
chris-olszewski, line 142: Installing protoc needs to be added back here

.github/workflows/bench-turbopack.yml
```diff
@@ -188,22 +183,17 @@ jobs:
...
+      - "x64"
+      - "metal"
 
     name: Benchmark on ${{ matrix.bench.name }}
 
     steps:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Setup Go
```
chris-olszewski, line 202: Installing protoc needs to be added back here

.github/workflows/bench-turbopack.yml
```diff
@@ -352,11 +342,6 @@ jobs:
       - name: Fetch the base branch
         run: git -c protocol.version=2 fetch --no-tags --progress --no-recurse-submodules --depth=1 origin +${{ github.base_ref }}:base
 
-      - name: Setup Go
```
chris-olszewski, line 355: Installing protoc needs to be added back here

.github/workflows/bench-turborepo.yml
```diff
@@ -24,9 +24,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup-node
-      - uses: ./.github/actions/setup-go
```
chris-olszewski, line 27: Installing protoc needs to be added back here

.github/workflows/bench-turbotrace-against-node-nft.yml
```diff
@@ -32,10 +32,6 @@ jobs:
 
       - uses: ./.github/actions/setup-rust
 
-      - uses: ./.github/actions/setup-go
```
chris-olszewski, line 35: Installing protoc needs to be added back here

chris-olszewski (changes requested):

chris-olszewski (approved):

Issue #6863, opened 2023-12-29
Turbo Error Trying to Read MariaDB Directory


hatemjaber:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

pnpm install turbo -g

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Linux

### Which canary version will you have in your reproduction?

1.10.17+

### Describe the Bug

I decided to update to the latest version of turbo on my laptop running the latest versions of node, npm, pnpm... I have a script that starts up docker containers for mariadb, redis, redisinsight, adminer,  and manticoresearch along with running the turbo run dev:WHATEVER command to start a specific stack. This has been working fine for a long time running turbo@1.10.16 with no issues; everything starts as expected. The latest version throws a permissions denied error trying to read my `.docker/mariadb/db-name/db_data` directory. I'm not sure why it's even trying to read that directory since it never did before running version 1.10.16. I tried every version from 1.10.17 to the latest canary and they all experience the same issue throwing the permission denied error for the mariadb directory.

I don 't know why it's trying to read the directory that I use to persist the database data. I switched to using docker volumes and the issue went away but I prefer to have my data persisted to the host in development. I switched back to 1.10.16 for now since it works but thought it was worth bringing to your attention.

### Expected Behavior

I expect to run `pnpm admin` and the script that runs docker, turbo, etc... should just run without getting a permission denied error. I expect for turbo to only worry about the directories that actually contain `apps` or `packages`. I scanned the documentation for a config that I can tell turbo to ignore certain directories but couldn't find that.

### To Reproduce

You would need a monorepo with a directory to hold the mysql/mariadb database and use docker to start up. Here is the script that I use to run the entire project:

```
#!/bin/bash

# Initialize a variable to determine whether to capture the PID or not
capture_pid=false

# Parse command-line options for start/stop and PID capture
action="start"  # Default action is to start the application

while getopts ":ps" opt; do
  case $opt in
    p)
      capture_pid=true  # Set capture_pid to true if -p flag is present
      ;;
    s)
      action="stop"  # Set action to stop if -s flag is present
      ;;
    \?)
      echo "Invalid option: -$OPTARG" >&2
      exit 1
      ;;
  esac
done
shift $((OPTIND-1))  # Shift off the options and optional --.

# Function to run the dev server
run_dev() {
  echo "Starting the dev server..."
  if [ "$capture_pid" = true ]; then
[...]

PR #6957, opened 2024-01-09
fix: ignore permission errors during globbing

chris-olszewski:
### Description

Fixes https://github.com/vercel/turbo/issues/6863

This brings us closer to the Go behavior where we would ignore [all IO errors](https://github.com/vercel/turbo/blob/main/cli/internal/doublestar/globwalk.go#L27).

We could do a similar thing by ignoring any error that doesn't have `ErrorKind::Other`, but that feels like slight overkill to me. I think ignoring `NotFound` and `PermissionDenied` is probably the sweet spot.

### Testing Instructions

I would love a unit test, but I'm unaware of an easy way to create a file via Rust that the same process is unable to read. Using a manual test instead:
Create a directory that you can't read
```
mkdir apps/docs/protected
sudo chown root apps/docs/protected
sudo chmod go-rw apps/docs/protected
```
Verify existing canary fails when running `turbo build`:
```
[0 olszewski@chriss-mbp] /tmp/daemon-bug $ turbo build --output-logs new-only    
• Packages in scope: @repo/eslint-config, @repo/typescript-config, @repo/ui, docs, web
• Running build in 5 packages
• Remote caching disabled
 ERROR  run failed: IO Error: failed to match directory tree: failed to read file at `Some("/private/tmp/daemon-bug/apps/docs/protected")`: Permission denied (os error 13)
```
Build PR and verify that this no longer causes a failure:
```
[1 olszewski@chriss-mbp] /tmp/daemon-bug $ turbo_dev build --output-logs new-only
• Packages in scope: @repo/eslint-config, @repo/typescript-config, @repo/ui, docs, web
• Running build in 5 packages
• Remote caching disabled
web:build: cache hit, suppressing logs b7e193fc8c53023e
docs:build: cache hit, suppressing logs 67c737e5d2f3d57a

 Tasks:    2 successful, 2 total
Cached:    2 cached, 2 total
  Time:    428ms >>> FULL TURBO
```

Closes TURBO-2013

### DIFF
--- a/crates/turborepo-globwalk/src/lib.rs
+++ b/crates/turborepo-globwalk/src/lib.rs
@@ -293,10 +293,11 @@ pub fn globwalk(
                             ),
                             Err(e) => {
                                 let io_err = std::io::Error::from(e);
-                                if io_err.kind() == std::io::ErrorKind::NotFound {
-                                    None
-                                } else {
-                                    Some(Err(io_err.into()))
+                                match io_err.kind() {
+                                    // Ignore DNE and permission errors
+                                    std::io::ErrorKind::NotFound
+                                    | std::io::ErrorKind::PermissionDenied => None,
+                                    _ => Some(Err(io_err.into())),
                                 }
                             }
                         }


arlyon (approved):
This seems like sound behaviour to me.

tknickman (approved):

mehulkar (commented):
re: testing: seems like a good case for a prysk test, but I'm also fine with not doing that

chris-olszewski:
> re: testing: seems like a good case for a prysk test, but I'm also fine with not doing that

For that to work I think we would need to have access on the test machine to a user that has lowered permissions compared to the user that's used by default to run the Prysk tests. I'm not sure if there's an easy way to do that either that works for Windows and Unix-like systems.

PR #6964, opened 2024-01-09
Sokra/perf calls

sokra:
### Description

Reduce the number of (cached) function calls, resolve tasks and work in general.


ForsakenHarmony (approved):

crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -321,13 +321,15 @@ async fn graph_node_to_referenced_nodes(
                 }]);
             };
 
-            let modules = reference.resolve_reference().primary_modules().await?;
+            // Dedupe modules to avoid duplicate work in the following loop
+            let mut modules = HashSet::new();
```
sokra, line 325:
```suggestion
            let mut modules = IndexSet::new();
```

PR #6970, opened 2024-01-09
feat(telemetry): track prune args

tknickman:
### Description

Track all prune args. This is used to determine usage to inform API changes / improvements. Actual values are not recorded in this case.

### DIFF
--- a/crates/turborepo-lib/src/commands/prune.rs
+++ b/crates/turborepo-lib/src/commands/prune.rs
@@ -17,6 +17,8 @@ use turborepo_ui::BOLD;
 use super::CommandBase;
 use crate::config::RawTurboJson;
 
+pub const DEFAULT_OUTPUT_DIR: &str = "out";
+
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     #[error("io error while pruning: {0}")]
@@ -87,7 +89,9 @@ pub async fn prune(
     output_dir: &str,
     telemetry: CommandEventBuilder,
 ) -> Result<(), Error> {
-    telemetry.track_prune_method(docker);
+    telemetry.track_prune_arg_method(docker);
+    telemetry.track_arg_usage("out-dir", output_dir != DEFAULT_OUTPUT_DIR);
+
     let prune = Prune::new(base, scope, docker, output_dir).await?;
 
     if matches!(
--- a/crates/turborepo-lib/src/commands/prune.rs
+++ b/crates/turborepo-lib/src/commands/prune.rs
@@ -89,7 +89,7 @@ pub async fn prune(
     output_dir: &str,
     telemetry: CommandEventBuilder,
 ) -> Result<(), Error> {
-    telemetry.track_prune_arg_method(docker);
+    telemetry.track_arg_usage("docker", docker);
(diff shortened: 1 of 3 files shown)


crates/turborepo-lib/src/commands/prune.rs
```diff
@@ -87,7 +89,9 @@ pub async fn prune(
     output_dir: &str,
     telemetry: CommandEventBuilder,
 ) -> Result<(), Error> {
-    telemetry.track_prune_method(docker);
+    telemetry.track_prune_arg_method(docker);
+    telemetry.track_prune_arg_out(output_dir != DEFAULT_OUTPUT_DIR);
```
tknickman, line 93: We don't care about the val here, only if it's being used.

crates/turborepo-lib/src/commands/prune.rs
```diff
@@ -87,7 +89,9 @@ pub async fn prune(
     output_dir: &str,
     telemetry: CommandEventBuilder,
 ) -> Result<(), Error> {
-    telemetry.track_prune_method(docker);
+    telemetry.track_prune_arg_method(docker);
+    telemetry.track_arg_usage("out-dir", output_dir != DEFAULT_OUTPUT_DIR);
```
mehulkar, line 93: So if a user says `--out-dir=out` , we would get false for this? Could we distinguish the default being provided, vs turbo doing the default thing? I wouldn't care if we were sending real values, but if our telemetry is tracking "set" vs "default", this implementation could be a bit misleading

tknickman: Yes that's correct, and that's a good call. I'm not sure if clap lets us differentiate between that. But I can check on it.

chris-olszewski: It doesn't. If we really want to do this we would need to change the type to be `Option<String>` and defer using the default until time of use.

crates/turborepo-telemetry/src/events/command.rs
```diff
@@ -128,9 +137,9 @@ impl CommandEventBuilder {
     }
 
     // prune
-    pub fn track_prune_method(&self, is_docker: bool) -> &Self {
+    pub fn track_prune_arg_method(&self, is_docker: bool) -> &Self {
```
chris-olszewski, line 140: I'm a little confused by the general name of this method. When it seems like it's `docker` specific.

I do feel like we could use `track_arg_usage("docker", docker)` in `prune.rs` to achieve the same thing and unify the values of `arg:*` a bit to either `"set"` or `"default"`

chris-olszewski (approved):
I think this looks good to me. IMO differentiating between `--out-dir` not being present and `--out-dir=out` isn't useful signal compared to losing the nice autogenerated help text that comes with using Clap's default value.

I might be missing something with my suggestion around `track_prune_arg_method`, feel free to ignore if it doesn't make sense.

crates/turborepo-telemetry/src/events/command.rs
tknickman: Good call here, I started out with really specific events for arguments to keep things easily traceable, but as I've been adding more and more I'm deferring back to general, now I just have a `track_arg_usage` (in this PR), as well as `track_arg_value` (in a future PR). 

Refactored this to use the general approach.

PR #6971, opened 2024-01-09
Add global auth as config option

Zertsov:
### Description
Allows for using the auth crate to load in auth tokens into the config. This should be a temporary fix until we get rid of the Go code, which right now needs us to store things in the ConfigurationOptions.

### Testing Instructions
1. Login
2. Ensure the token is properly loaded by running any command that needs a token and see if the command works as expected.

Closes TURBO-2019


Zertsov:
closing this since I accidentally merged these changes in another PR.

PR #6972, opened 2024-01-09
feat(telemetry): track global args

tknickman:
### Description

Track global arg usage

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -323,6 +323,72 @@ impl Args {
                 .unwrap_or(&[]),
         }
     }
+
+    pub fn track(&self, telemetry: &GenericEventBuilder) {
+        // track usage only
+        if self.skip_infer {
+            telemetry.track_arg_usage("skip-infer", true);
+        }
+        if self.no_update_notifier {
+            telemetry.track_arg_usage("no-update-notifier", true);
+        }
+        if self.api.is_some() {
+            telemetry.track_arg_usage("api", true);
+        }
+        if self.color {
+            telemetry.track_arg_usage("color", true);
+        }
+        if self.cpu_profile.is_some() {
+            telemetry.track_arg_usage("cpuprofile", true);
+        }
+        if self.cwd.is_some() {
+            telemetry.track_arg_usage("cwd", true);
+        }
+        if self.heap.is_some() {
+            telemetry.track_arg_usage("heap", true);
+        }
+        if self.login.is_some() {
+            telemetry.track_arg_usage("login", true);
+        }
+        if self.no_color {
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -323,6 +323,62 @@ impl Args {
...
+        if self.token.is_some() {
+            telemetry.track_arg_usage("token", true);
+        }
+        if self.trace.is_some() {
+            telemetry.track_arg_usage("trace", true);
+        }
+
+        // track values
+        if self.remote_cache_timeout.is_some() {
+            telemetry.track_arg_remote_cache_timeout(self.remote_cache_timeout.unwrap_or_default());
```
chris-olszewski, line 371:
```suggestion
        if let Some(remote_cache_timeout) = self.remote_cache_timeout {
            telemetry.track_arg_remote_cache_timeout(remote_cache_timeout);
```

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -323,6 +323,62 @@ impl Args {
...
+        }
+
+        if self.verbosity.v > 0 {
+            telemetry.track_arg_verbosity("v", self.verbosity.v);
+        }
+        if self.verbosity.verbosity.is_some() {
+            telemetry
+                .track_arg_verbosity("verbosity", self.verbosity.verbosity.unwrap_or_default());
+        }
```
chris-olszewski, line 380:
```suggestion
        if let Some(verbosity) = self.verbosity.verbosity {
            telemetry
                .track_arg_verbosity("verbosity", verbosity);
        }
```

crates/turborepo-telemetry/src/events/generic.rs
```diff
@@ -134,8 +134,35 @@ impl GenericEventBuilder {
...
+    }
 
+    pub fn track_arg_remote_cache_timeout(&self, timeout: u64) -> &Self {
+        self.track(Event {
+            key: "arg:remote-cache-timeout".to_string(),
+            value: timeout.to_string(),
+            is_sensitive: EventType::NonSensitive,
+        });
+        self
+    }
```
chris-olszewski, line 154: We could write a generic `track_arg_value` function to avoid some of the code dupe. Possibly also take a `is_sensitive` bool and avoid sending the event if `true`?
```suggestion
    pub fn track_arg_value(&self, arg: &str, value: impl Display) -> &Self {
        self.track(Event {
            key: format!("arg:{arg}"),
            value: value.to_string(),
            is_sensitive: EventType::NonSensitive,
        });
        self
    }
```

chris-olszewski (approved):
A few minor Rust taste suggestions

arlyon (approved):
where is object.entries when you need it

PR #6983, opened 2024-01-10
fix(prune): correctly create symlinks to directories 

### LINKED ISSUE #5842: [turborepo] `turbo prune` does not preserve symlinks

1.10.13

pnpm

Mac

### Describe the Bug

When using symlinks in an app and prune this app, the symlink gets converted to a directory. 

**Backstory** 

My use-case where I found this is, that we distribute our font-stack and font files in a separate package `packages/design`. 
Using this approach, we ran into issues with `scss` not resolving the font files doe to them being loaded from a relative path in `packages/design` which does not exist in `apps/web`. This is solved by a symlink, until I have a better solution, but that's not a topic for here, I just wanted to share some insights to why this symlink is used.

### Expected Behavior

Symlinks should be preserved during `turbo prune`, both inside the same package and between packages, and not get converted to plain directories.

> See **Backstory** to explain why we use/need a symlink in the first place

### To Reproduce

Basically create _any_ symlink in an app, `turbo prune` and check the symlink.

```bash
# 1. create new monorepo
pnpm dlx create-turbo@latest

# 2. create directory test in apps/web with an empty file
cd apps/web
mkdir test
touch test/test.txt

# 3. create a symlink `directory` pointing to the created directory
ls -n test directory

# 4. Verify we can access the file "through" the symlink"
file directory/test.txt 
# OUTPUT: directory/test.txt: empty
# CONCLUSION: symlink works

# 5. go back and prune apps/web 
cd ../..
turbo prune --scope=web

# 6. check the symlink in the pruned output
cd out/apps/web
file directory/text.txt  
# OUTPUT: directory/test.txt: cannot open `directory/test.txt' (No such file or directory)
# CONCLUSION: symlink does not work
```

When we take a look at `directory` in the pruned app, we see that is "just" a directory and not a symlink anymore.

```text
drwxr-xr-x   2 johannes  staff    64 Aug 31 06:10 directory/
```

### Reproduction Repo

https://github.com/cngJo/turbo-symlink-reproduction

chris-olszewski:
### Description

Fixes #5842

> The behavior of RecursiveCopy was written prior to cache-to-tar which removed all other consumers of this functionality.

This is why there was the odd behavior with copying symlink'd directories. We now only use this function for `prune` so making this behavior change makes sense instead of keeping the quirk.

Note:
This does not handle making sure the destination is copied over as well. This could be implemented, but it isn't a simple fix. An easy workaround it to make sure the files/directories are part of a package and make sure that package is declared as a dependency.

### Testing Instructions

Updated unit test to verify new behavior

Closes TURBO-2027


mehulkar (commented):
Just to make sure I understand your description:

> This does not handle making sure the destination is copied over as well.

So after this change, we'll copy the symlink, but not the thing it points to? (because it may be outside the prune target)?

Is there a warning we could emit if that's what is happening? It seems like surprising behavior either way, but creating a broken symlink on purpose seems like it have its own side effects

chris-olszewski:
> So after this change, we'll copy the symlink, but not the thing it points to? (because it may be outside the prune target)?

This already happens with how we copy over symlinks to files. This change has us match that behavior for directories as well (previously we would just make an empty directory instead of a symlink). This isn't ideal behavior, but IMO this is strictly better behavior.

There's also the design question of "should we copy all symlinks"? A user might have symlinks outside of the repository or a symlink to an absolute path which won't make sense in a Docker container. 

> Is there a warning we could emit if that's what is happening?

We could crawl the entire output directory and looking for broken symlinks and warn that way.

NicholasLYang:
In theory we could use some of the same checks in our cache implementation to validate symlink destinations. But agreed that this is out of scope of this PR

mehulkar (approved):
Explanation makes sense, thanks! A warning makes sense to me if it's easy to add (probably where the `continue` is), but since we don't do that for files right now, I think we can defer that work.

arlyon (approved):

PR #6985, opened 2024-01-10
feat(telemetry): track run arguments

tknickman:
### Description

> This is publicly disabled!

Track anonymous run usage to inform future CLI changes

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -16,7 +16,7 @@ use turborepo_telemetry::{
     events::{
         command::{CodePath, CommandEventBuilder},
         generic::GenericEventBuilder,
-        EventBuilder,
+        EventBuilder, EventType,
     },
     init_telemetry, TelemetryHandle,
 };
@@ -37,6 +37,9 @@ mod error;
 // Global turbo sets this environment variable to its cwd so that local
 // turbo can use it for package inference.
 pub const INVOCATION_DIR_ENV_VAR: &str = "TURBO_INVOCATION_DIR";
+
+// Default value for the --cache-workers argument
+const DEFAULT_NUM_WORKERS: u32 = 10;
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Deserialize, Serialize, ValueEnum)]
 pub enum OutputLogsMode {
@@ -572,7 +575,7 @@ pub struct RunArgs {
     #[clap(long)]
     pub cache_dir: Option<Utf8PathBuf>,
     /// Set the number of concurrent cache operations (default 10)
-    #[clap(long, default_value_t = 10)]
+    #[clap(long, default_value_t = DEFAULT_NUM_WORKERS)]
     pub cache_workers: u32,
     /// Limit the concurrency of task execution. Use 1 for serial (i.e.
     /// one-at-a-time) execution.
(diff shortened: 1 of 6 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -758,6 +761,192 @@ impl RunArgs {
             (None, None) => None,
         }
     }
+
+    pub fn track(&self, telemetry: &CommandEventBuilder) {
```
tknickman, line 765: I've looked for every feature of Structs to make this easier, but this is all I've got so far. But it is painful.

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -758,6 +761,192 @@ impl RunArgs {
...
+            telemetry.track_arg_value("filter", self.filter.len(), EventType::NonSensitive);
+        }
+
+        if !self.scope.is_empty() {
+            telemetry.track_arg_value("scope", self.scope.len(), EventType::NonSensitive);
+        }
+
+        if !self.ignore.is_empty() {
+            telemetry.track_arg_value("ignore", self.ignore.len(), EventType::NonSensitive);
+        }
```
tknickman, line 948: These vals would be sensitive and hashed vals are useless here so I've opted to track the size of the respective item here to give an idea of the scope of usage. 

For some of these (filter) we can add more events deeper into the code to get a better idea about the type of items being filtered on.

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -758,6 +761,192 @@ impl RunArgs {
...
+                    LogPrefix::None => "none",
+                    LogPrefix::Task => "task",
+                },
+                EventType::NonSensitive,
+            );
+        }
+
+        // track sizes
+        if !self.filter.is_empty() {
+            telemetry.track_arg_value("filter", self.filter.len(), EventType::NonSensitive);
```
mehulkar, line 939: I checked out our [docs](https://turbo.build/repo/docs/core-concepts/monorepos/filtering) for all the filtering options. Could we have the telemetry broken down by each style? (which probably means we have to do this tracking _after_ we've parsed it)

| Filter type | logic |
| --- | --- |
| Filter by workspace name | alphanumeric value
| Filter by workspace directory | starts with `./`
| Include dependents of matched workspaces | starts with `...`
| Include dependencies of matched workspaces | ends with `...`
| Execute tasks from the workspace root | matches `//`
| Filter by changes in git history | wrapped in `[]`
| Exclude workspaces from selection | starts with `!`

I would also _love_ to break down by `*` values in the first two (e.g. `-F=util-*` or `-F=./packages/*`), but I've excluded that here.

I'd propose just making the _value_ of this `filter` arg some enum that corresponds to each of these as they _should_ be mutually exclusive.

tknickman: Yea agreed I mentioned that above as well, we can add other events in the filter parser itself to track all of these - I have an task for it!

crates/turborepo-lib/src/cli/mod.rs
chris-olszewski: What you are probably looking for is a [procedural macro](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros) that allows you to generate code based on the `struct` definition. This is fairly advanced and probably only makes sense if we end up doing this a lot and it becomes a pain to maintain.

There might be some crates that make writing this easier, but this is an area I don't know much about.

chris-olszewski: Something that might make this easier is to use a macro to generate these calls. e.g.
```
macro_rules! track_usage {
           ($field:ident) => {
                if self.$field {
                    telemetry.track_arg_usage(stringify!($field), true);
                }
            };
            ($field:ident, $is_used:expr) => {
                if $is_used(&self.$field) {
                    telemetry.track_arg_usage(stringify!($field), true);
                }
            };
        }
```

This allows for `track_usage!(skip_infer);` which expands to:
```
if self.skip_infer {
    telemetry.track_arg_usage("skip_infer", true);
}
```

And for providing a "is set" function for non bool types e.g. `track_usage!(api, Option::is_some);` which expands to:
```
if (Option::is_some)(&self.api) {
    telemetry.track_arg_usage("api", true);
}
```

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -758,6 +761,192 @@ impl RunArgs {
...
+            telemetry.track_arg_usage("no-deps", true);
+        }
+
+        if self.no_cache {
+            telemetry.track_arg_usage("no-cache", true);
+        }
+
+        if self.daemon {
+            telemetry.track_arg_usage("daemon", true);
+        }
```
chris-olszewski, line 817: These are all duplicated calls

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -758,6 +761,192 @@ impl RunArgs {
...
+
+        if self.no_daemon {
+            telemetry.track_arg_usage("no_daemon", true);
```
chris-olszewski, line 820: Different casing conventions on this than the other args. Would suggest using a macro to unify how these strings get produced.

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -758,6 +761,192 @@ impl RunArgs {
...
+        }
+
+        if self.env_mode != EnvMode::default() {
+            telemetry.track_arg_value(
+                "env-mode",
+                match self.env_mode {
+                    EnvMode::Infer => "infer",
+                    EnvMode::Loose => "loose",
+                    EnvMode::Strict => "strict",
+                },
```
chris-olszewski, line 894: We should implement `Display` on `EnvMode` and consolidate the logic there (which in turn should probably make use of the `Serialize` implementation as I think we want to keep those behaviors in sync)

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -758,6 +761,192 @@ impl RunArgs {
...
+                EventType::NonSensitive,
+            );
+        }
+
+        if let Some(output_logs) = &self.output_logs {
+            telemetry.track_arg_value(
+                "output-logs",
+                match output_logs {
```
chris-olszewski, line 902: Same as comment on `EnvMode`

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -758,6 +761,192 @@ impl RunArgs {
...
+                },
+                EventType::NonSensitive,
+            );
+        }
+
+        if self.log_order != LogOrder::default() {
+            telemetry.track_arg_value(
+                "log-order",
+                match self.log_order {
+                    LogOrder::Auto => "auto",
```
chris-olszewski, line 917: Same as comment on `EnvMode`

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -758,6 +761,192 @@ impl RunArgs {
...
+                },
+                EventType::NonSensitive,
+            );
+        }
+
+        if self.log_prefix != LogPrefix::default() {
+            telemetry.track_arg_value(
+                "log-prefix",
+                match self.log_prefix {
+                    LogPrefix::Auto => "auto",
```
chris-olszewski, line 929: Same as comment on `EnvMode`

chris-olszewski (changes requested):
I think a macro (not procedural) would go a long way to make the `track` method easier to read/maintain.

crates/turborepo-lib/src/cli/mod.rs
tknickman: Done!

crates/turborepo-telemetry/src/events/mod.rs
```diff
@@ -41,3 +41,19 @@ pub trait EventBuilder {
...
 }
+
+#[macro_export]
+macro_rules! track_usage {
+    ($tel:expr, $field:expr, $is_used:expr) => {
+        if $is_used($field) {
+            $tel.track_arg_usage(
+                stringify!($field)
+                    .trim_start_matches("&")
+                    .trim_start_matches("self.")
```
tknickman, line 52: Open to better options here if you have suggestions @chris-olszewski !

chris-olszewski: I agree this feels gross, played around with this and couldn't come up with anything better. What we really want is proc macro. Maybe I'll take a late DYF and try to whip something up.

crates/turborepo-lib/src/cli/mod.rs
chris-olszewski: Do we want to change the arg name to something like `filter-len` or `num-of-filters` to make it clear in the data that the user isn't passing `--filter=3`?

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -90,6 +136,20 @@ pub enum EnvMode {
...
+        write!(
+            f,
+            "{}",
+            match self {
+                EnvMode::Infer => "infer",
+                EnvMode::Loose => "loose",
+                EnvMode::Strict => "strict",
+            }
+        )
+    }
```
chris-olszewski, line 150: We can avoid template string literal if we invoke [`write_str`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.write_str) directly.
```suggestion
        f.write_str(
            match self {
                EnvMode::Infer => "infer",
                EnvMode::Loose => "loose",
                EnvMode::Strict => "strict",
            }
        )
    }
```

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -82,6 +115,19 @@ pub enum DryRunMode {
...
+impl Display for DryRunMode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                DryRunMode::Text => "text",
+                DryRunMode::Json => "json",
+            }
+        )
```
chris-olszewski, line 127:
```suggestion
        f.write_str(
            match self {
                DryRunMode::Text => "text",
                DryRunMode::Json => "json",
            }
        )
```

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -74,6 +93,20 @@ impl Default for LogOrder {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                LogOrder::Auto => "auto",
+                LogOrder::Stream => "stream",
+                LogOrder::Grouped => "grouped",
+            }
+        )
```
chris-olszewski, line 106:
```suggestion
        f.write_str(
            f,
            match self {
                LogOrder::Auto => "auto",
                LogOrder::Stream => "stream",
                LogOrder::Grouped => "grouped",
            }
        )
```

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -58,6 +61,22 @@ impl Default for OutputLogsMode {
...
+            f,
+            "{}",
+            match self {
+                OutputLogsMode::Full => "full",
+                OutputLogsMode::None => "none",
+                OutputLogsMode::HashOnly => "hash-only",
+                OutputLogsMode::NewOnly => "new-only",
+                OutputLogsMode::ErrorsOnly => "errors-only",
+            }
+        )
```
chris-olszewski, line 76:
```suggestion
        f.write_str(
            match self {
                OutputLogsMode::Full => "full",
                OutputLogsMode::None => "none",
                OutputLogsMode::HashOnly => "hash-only",
                OutputLogsMode::NewOnly => "new-only",
                OutputLogsMode::ErrorsOnly => "errors-only",
            }
        )
```

chris-olszewski (approved):
I'll think about better ways to facilitate the generating the tracing events, but this works so :shipit:

mehulkar (approved):

crates/turborepo-lib/src/cli/mod.rs
tknickman: Good call, done!

PR #6576, opened 2023-11-24
Update with-svelte-example.

anthonyshew:
Updating the with-svelte example!

Closes TURBO-1714

Things to pay attention to:

- [x] Is tasking correct? Are all tasks successful out-of-the-box? Do I hit cache when I'm expecting?
- [x] If I add an unused variable in a workspace, do I see a warning in both my editor and when I run turbo lint?
- [ ] Does auto-importing work across packages? _(I do not know how/if this works with Svelte.)_
- [x] Anything else wonky in your editor?
- [x] Do I get proper hot-reloading in turbo dev?
- [x] Anything else that may come up!

PR #6986, opened 2024-01-11
Fix with-svelte-example

dummdidumm:
### Description

- vite-plugin-svelte is now a peer dependency and therefore needs to appear in the user's `package.json`
- fixed preprocessor import in `svelte.config.js`
- removed `optimizeDeps.include` which makes Vite treat the package as an external dependency that it should bundle once, resulting in changes not getting picked up (this was introduced in #6576 but I'm not sure why)
- fixed e2e tests

### Testing Instructions

check manually


anthonyshew (approved):
Lovely! Thank you!

Issue #6994, opened 2024-01-12
bug(graph): Unsupported format should raise error


uiolee:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

any

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Windows

### Which canary version will you have in your reproduction?

turbo 1.11.3, turbo 1.11.4-canary.0

### Describe the Bug

https://turbo.build/repo/docs/reference/command-line-reference/run#--graph

When I pass a unsupported format to `turbo run <task> --graph <file.ext>`,

It prompted me that the graph was generated successfully, but it actually didn't.

```bash
$ pnpm exec turbo run build --graph graph.md

Format: "md" not recognized. Use one of: bmp canon cmap cmapx cmapx_np dot dot_json emf emfplus eps fig gif gv imap imap_np ismap jpe jpeg jpg json json0 metafile mp pdf pic plain plain-ext png pov ps ps2 svg tif tiff tk xdot xdot1.2 xdot1.4 xdot_json

✔ Generated task graph in D:\test\graph.md
```

### Expected Behavior

It should prompted me that failed to generate, and raise error with a non-zero exit code .

### To Reproduce

In a turborepo,

pass a unsupported format to `turbo run <task> --graph <file.ext>`

e.g.,

`turbo run build --graph graph.md`

### Additional context

graphviz        9.0.0

<sub>[TURBO-2036](https://linear.app/vercel/issue/TURBO-2036/buggraph-unsupported-format-should-raise-error)</sub>

PR #6995, opened 2024-01-12
fix(turbo): validate graph extension

tknickman:
### Description

Fixes https://github.com/vercel/turbo/issues/6994

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -40,7 +40,8 @@ pub const INVOCATION_DIR_ENV_VAR: &str = "TURBO_INVOCATION_DIR";
 
 // Default value for the --cache-workers argument
 const DEFAULT_NUM_WORKERS: u32 = 10;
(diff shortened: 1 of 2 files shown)


mehulkar (commented):
Can you add a test here? https://github.com/vercel/turbo/blob/main/turborepo-tests/integration/tests/run/graph.t

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -580,6 +582,23 @@ pub enum GenerateCommand {
     Run(GeneratorCustomArgs),
 }
 
+fn validate_graph_extension(s: &str) -> Result<String, String> {
+    match s.is_empty() {
+        true => Ok(s.to_string()),
+        _ => match std::path::Path::new(s).extension().and_then(|e| e.to_str()) {
```
NicholasLYang, line 588: I'd use `Utf8Path` here since it lets you avoid the `to_str` call

NicholasLYang (changes requested):

NicholasLYang (approved):

PR #7002, opened 2024-01-13
DOCS: Changed getting started to be "npm create turbo@latest"

MatanelGordon:
### Description
Changed usage of turbo repo to be "npm create turbo@latest" instead of "npx create-turbo". These are the same but the former looks better

### Testing Instructions
Not that I am aware of. 
I tried to create a new project with `npm create` and that seem to work like a charm

### DIFF
--- a/docs/pages/repo/docs/getting-started/create-new.mdx
+++ b/docs/pages/repo/docs/getting-started/create-new.mdx
@@ -21,7 +21,7 @@ To create a new monorepo, use our [`create-turbo`](https://www.npmjs.com/package
 <Tabs items={['npm', 'yarn', 'pnpm']} storageKey="selected-pkg-manager">
   <Tab>
     ```sh
-    npx create-turbo@latest
+    npm create turbo@latest
     ```
   </Tab>
   <Tab>


mehulkar (changes requested):
hmm, I vaguely know that `npm create` is a thing, but I don't see it documented https://docs.npmjs.com/cli/v10/commands and can't find any other information about it. Can you point to where I can learn how it differs (or doesn't) from `npx`?

tknickman:
Agreed, it's very poorly documented. It was added in 6.0.0 in response to yarn:

https://docs.npmjs.com/cli/v6/using-npm/changelog#v600-2018-04-20

> You can now do things like npm init react-app or npm init esm to scaffold an npm package by running create-react-app and create-esm, respectively. This also adds an npm create alias, to correspond to Yarn's yarn create feature, which inspired this.

anthonyshew:
I do like this aesthetically, visually - but the lack of proper documentation from the npm side makes this a tough sell for me. The only time I can recall off the top of my head that I've ever seen this used is `npm create vite`.

Considering that there isn't any clear, tangible gain from this change (beyond that it looks marginally nicer), I'm going to err on the side of status quo.

PR #7004, opened 2024-01-15
improve error message when no config is exported from postcss.config.js

sokra:
### Description

improve error message

Closes PACK-2238


sokra:
There is something broken with typescript in on `main`...

PR #7005, opened 2024-01-15
fix(examples): fontSize type is number not string

devjiwonchoi:
- ref: https://reactnative.dev/docs/text-style-props#fontsize

This PR fixes the wrong type within examples.

Screenshot 2024-01-16 at 3 24 49 AM

Screenshot 2024-01-16 at 3 25 27 AM


anthonyshew (approved):
Awesome! Thank you!

PR #7013, opened 2024-01-16
fix: make CSS reloading work on firefox

ForsakenHarmony:
### Description

Firefox for some reason never reloads `<link />` tags if the URL has already been loaded once for the current page load.

Closes PACK-2135


sokra (approved):

sokra:
There is something broken with typescript in on `main`...

PR #7018, opened 2024-01-16
fix(turbo): sort dependents and dependencies during normalization

mehulkar:


crates/turborepo-lib/src/run/summary/mod.rs
```diff
@@ -768,6 +768,12 @@ impl<'a> RunSummary<'a> {
         }
 
         self.tasks.sort_by(|a, b| a.task_id.cmp(&b.task_id));
+
+        // Sort dependencies
+        for task in &mut self.tasks {
+            task.shared.dependencies.sort_by(|d1, d2| d1.cmp(&d2));
+            task.shared.dependents.sort_by(|d1, d2| d1.cmp(&d2));
```
chris-olszewski, line 775:
```suggestion
            task.shared.dependencies.sort();
            task.shared.dependents.sort();
```

crates/turborepo-lib/src/run/summary/mod.rs
```diff
@@ -347,13 +347,13 @@ struct SinglePackageRunSummary<'a> {
...
 impl<'a> From<&'a RunSummary<'a>> for SinglePackageRunSummary<'a> {
     fn from(run_summary: &'a RunSummary<'a>) -> Self {
-        let mut tasks = run_summary
+        let tasks = run_summary
             .tasks
             .iter()
             .cloned()
             .map(SinglePackageTaskSummary::from)
             .collect::<Vec<_>>();
-        tasks.sort_by(|t1, t2| t1.task_id.cmp(&t2.task_id));
```
mehulkar, line 356: Removing extra `sort()` because tasks are sorted during normalization: https://github.com/vercel/turbo/blob/1dd2e93e84e08872f67943804a6600f0e91a124a/crates/turborepo-lib/src/run/summary/mod.rs#L770

Since this sort only applies to `SinglePackageRunSummary`, it made more sense to do it when we care about sorting. If the implementation is the same as in Go, it _should_ apply to dry runs as well as run summary!

I can also revert this part if we aren't confident about that.

crates/turborepo-lib/src/run/summary/task.rs
```diff
@@ -251,12 +251,10 @@ impl From<SharedTaskSummary<TaskId<'static>>> for SharedTaskSummary<String> {
             dependencies: dependencies
                 .into_iter()
                 .map(|task_id| task_id.task().to_string())
-                .sorted()
```
mehulkar, line 254: This `from` implementation is used by the `into()` call here: https://github.com/vercel/turbo/blob/1dd2e93e84e08872f67943804a6600f0e91a124a/crates/turborepo-lib/src/run/summary/task.rs#L211

it only runs for `SinglePackageTaskSummary`. Since we have moved the sorting to the `normalize()` step, we shouldn't need it here.

mehulkar (commented):
self review

crates/turborepo-lib/src/run/summary/mod.rs
chris-olszewski: I would say we remove the second sort in `normalize` over this one as other callers might depend on the list being sorted.

chris-olszewski (approved):
Stamping, but I would love either a unit or integration test. Or just a manual test that shows before/after behavior of this change.

crates/turborepo-lib/src/run/summary/mod.rs
mehulkar: But then I'd also need another sort somewhere else for monorepos (this one that is removed is only for when converting from monorepo to single package summary). 

but point made, I'll keep this one and the duplicate since it's not related to this fix anyway

PR #6543, opened 2023-11-22
turbo LSP + VSCode extension

arlyon:
### Description

This is now ready for review. I have included a vsix file below that you can download and test, simply add it to VSCode. You will need a recent version (ie. what is currently in main) to use it properly so I suggest pulling out a recent master, building a binary, and copying it to `node_modules/turbo-<platform>` for whatever project you wish to test it with.

[turbo-vsc-1.0.0.vsix.zip](https://github.com/vercel/turbo/files/13887415/turbo-vsc-1.0.0.vsix.zip)

Closes TURBO-1737

PR #7023, opened 2024-01-17
fix: forward fixes for LSP

chris-olszewski:
### Description

#6543 accidentally broke some things. This PR:
 - adds back building the Go binary to `turbo`'s build script
 - Lifts TLS backend features from `turborepo-lib` to `turborepo-lsp` now that the LSP crate is an entry point to the crate. (Matters as `cargo-groups` prunes `check` runs to only run against the entry points and now errors complaining that the feature doesn't exist)
 - Fixes clippy warning about `FromStr`
 - Fixes test that didn't compile

### Testing Instructions

CI Passes 👀 

Closes TURBO-2056

### DIFF
--- a/crates/turborepo-lib/src/daemon/connector.rs
+++ b/crates/turborepo-lib/src/daemon/connector.rs
@@ -714,7 +714,7 @@ mod test {
             .await
             .unwrap_err()
             .into();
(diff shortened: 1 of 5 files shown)


arlyon (approved):
Very silly my bad

PR #7025, opened 2024-01-18
fix: filter package.json files that are subfolders of other packages

arlyon:
### Description

We have an issue in the rust port with an overeager glob walking implementation. We want to stop walking when a package.json is encountered such that sub-packages are ignored. This is not possible unfortunately due to the traversal order in `walkdir`, so instead we must filter the values out afterwards.

### Testing Instructions

New integration test.


crates/turborepo-repository/src/package_manager/mod.rs
```diff
@@ -483,7 +484,49 @@ impl PackageManager {
...
+                                std::cmp::Ordering::Equal => None,
+                                other => Some(other),
+                            }
+                        } else if a_last {
+                            Some(std::cmp::Ordering::Less)
+                        } else {
+                            Some(std::cmp::Ordering::Greater)
+                        }
+                    })
+                    .unwrap_or(std::cmp::Ordering::Equal)
```
chris-olszewski, line 516: If I'm understanding this, we want to lexicographically order the paths. If so I think [`Iterator::cmp`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.cmp) is what we want. If don't want lexicographic ordering, then we need a comment explaining the ordering.
```suggestion
                a.components().cmp(b.components())
```

chris-olszewski (changes requested):

crates/turborepo-repository/src/package_manager/mod.rs
arlyon: It's not quite the same. Maybe the comment wasn't clear. The issue is that we want items without subfolders (ie files) to appear before subfolders and their contents. Just doing lexicographical ordering doesn't cut it since you can end up with folder `package/a/b/c/package.json` before `package/package.json`, so we need to check the special case if it is the last segment and always put those first

turborepo-tests/integration/fixtures/nested_packages/apps/my-app/.ignore/package.json
```diff
@@ -0,0 +1 @@
+{}
```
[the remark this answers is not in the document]
arlyon: The 'failing' behaviour on the old path is that in some cases (.next) a package.json will be produced with invalid contents. In the case of .next that is that the name is missing causing a failure.

turborepo-tests/integration/tests/nested-packages.t
```diff
@@ -0,0 +1,19 @@
+Setup
+  $ . ${TESTDIR}/../../helpers/setup_integration_test.sh nested_packages
+
+Run a build
+  $ ${TURBO} build
```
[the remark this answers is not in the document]
arlyon: I'm just going to move it to the rust package

crates/turborepo-repository/src/package_manager/mod.rs
```diff
@@ -483,7 +484,52 @@ impl PackageManager {
...
+                    // last_parent is the parent of the last json we yielded. if the current
+                    // entry is a subfolder of it, we don't want to yield it
+                    Some(parent) if entry.starts_with(parent) => false,
+                    // update last_parent to the parent of the current entry
+                    _ => {
+                        last_parent = Some(entry.parent().unwrap().to_owned());
+                        true
+                    }
+                }
+            }))
```
chris-olszewski, line 532: This filtering code feels like a prime candidate to be pulled out into a helper function and have a bunch of unit tests written against it.

chris-olszewski (approved):
Approving as I think this is correct, but it would be great to have a few unit test cases that cover the expected behavior of the filtering.

PR #7031, opened 2024-01-19
feat(telemetry): track more daemon errors

tknickman:
### Description

Track some more cache / daemon related errors.


arlyon (approved):
It may be worth at some point having a global cell or something rather than threading this through everywhere but we'll handle that independently

tknickman:
We actually _do_ have a global cell 😄 so we don't need to thread it. 

But in my experience instrumenting this, it turns out we use it at most of the levels anyway, and if we do thread it we get the extra information of the parent / child event relationship, so it's nicer to do it that way. 

But if there is something super deep, we don't have to thread it, we can just create an event from there and send it.

PR #7033, opened 2024-01-19
feat(turborepo): Error Improvements & Documentation

NicholasLYang:
### Description

Converted more errors to be high quality. Specifically, I added diagnostic codes, more coherent messages and code snippets when possible. Also added documentation for these errors on the turbo site.

### Testing Instructions

Closes TURBO-2069


crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -52,6 +55,12 @@ pub struct TurboJson {
 #[serde(rename_all = "camelCase")]
 // The raw deserialized turbo.json file.
 pub struct RawTurboJson {
+    #[serde(skip)]
+    // The raw text of the turbo.json file.
+    text: Option<Arc<str>>,
+    #[serde(skip)]
+    path: Option<Arc<str>>,
```
chris-olszewski, line 62: Just to verify my understanding, the `text`/`path` inclusion in `TurboJson` and `RawTurboJson` is necessary as there can be parse errors at the validation step i.e. `RawTurboJson` -> `TurboJson` of afterwards when the `turbo.json` is being consumed by something like the task graph. Is this correct?

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -80,19 +89,27 @@ pub struct RawTurboJson {
 
 #[derive(Serialize, Default, Debug, PartialEq, Clone)]
 #[serde(transparent)]
-pub struct Pipeline(BTreeMap<TaskName<'static>, RawTaskDefinition>);
+pub struct Pipeline(BTreeMap<TaskName<'static>, PipelineEntry>);
+
+#[derive(Serialize, Default, Debug, PartialEq, Clone)]
+#[serde(transparent)]
+pub struct PipelineEntry {
```
chris-olszewski, line 96: Is there a reason why this is necessary and we can't just use `Spanned<RawTaskDefinition>`? Easily could be, but I'm curious as to what's blocking it.

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -492,8 +516,14 @@ impl TurboJson {
                 let mut pipeline = Pipeline::default();
                 for (task_name, task_definition) in turbo_from_files.pipeline {
                     if task_name.is_package_task() {
+                        let (span, text) = match (task_definition.span, turbo_from_files.text) {
+                            (Some(range), Some(text)) => (Some(range.into()), text.to_string()),
+                            (_, _) => (None, String::new()),
+                        };
```
chris-olszewski, line 522:
```suggestion
                        let (span, text) = match task_definition.span.zip(turbo_from_files.text) {
                            Some((range, text)) => (Some(range.into()), text.to_string()),
                            None => (None, String::new()),
                        };
```

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -565,21 +604,38 @@ type TurboJSONValidation = fn(&TurboJson) -> Vec<Error>;
...
-        .map(|task_name| Error::UnnecessaryPackageTaskSyntax {
-            actual: task_name.to_string(),
-            wanted: task_name.task().to_string(),
+        .iter()
+        .filter(|(task_name, _)| task_name.is_package_task())
+        .map(|(task_name, entry)| {
+            let (span, text) = match (entry.span.clone(), &turbo_json.text) {
+                (Some(range), Some(text)) => (Some(range.into()), text.to_string()),
+                (_, _) => (None, String::new()),
+            };
```
chris-olszewski, line 613: This pattern seems common enough that we might wanna put it in a helper function

crates/turborepo-lib/src/config.rs
```diff
@@ -63,12 +75,28 @@ pub enum Error {
...
+    #[diagnostic(
+        code(unnecessary_package_task_syntax),
+        url("{}/messages/{}", TURBO_SITE, self.code().unwrap().to_string().to_case(Case::Kebab))
+    )]
     #[error("\"{actual}\". Use \"{wanted}\" instead")]
-    UnnecessaryPackageTaskSyntax { actual: String, wanted: String },
+    UnnecessaryPackageTaskSyntax {
+        actual: String,
+        wanted: String,
+        #[label("unnecessary syntax found here")]
```
chris-olszewski, line 86: I find providing what syntax is unnecessary helpful
```suggestion
        #[label("unnecessary package syntax found here")]
```

chris-olszewski (approved):
I think this looks good to me, but should definitely wait on a stamp from @anthonyshew

turborepo-tests/helpers/replace_turbo_config.sh
```diff
@@ -11,8 +11,7 @@ TURBO_CONFIGS_DIR="${MONOREPO_ROOT_DIR}/turborepo-tests/integration/fixtures/tur
 cp "${TURBO_CONFIGS_DIR}/$CONFIG_NAME" "$PROJECT_DIR/turbo.json"
 
 # Check if there are changes before trying to run git commit
-# Since we're replacing an existing turbo.json, git commit -a should work and we
-# don't need to git add anything.
 if [[ $(git status --porcelain) ]]; then
-  git commit --quiet -am "Use $CONFIG_NAME as turbo.json"
+  git add .
```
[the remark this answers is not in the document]
NicholasLYang: I'm adding a `turbo.json` that didn't exist already, so it needs to be added explicitly

crates/turborepo-lib/src/turbo_json/mod.rs
NicholasLYang: Eh not particularly anymore. That was an artifact of something else that I tried

crates/turborepo-lib/src/turbo_json/mod.rs
NicholasLYang: Yeah there's some errors during validation that require these fields. Also since they're all `Arc`, it's pretty low cost to put them everywhere

docs/pages/messages/empty-cwd.mdx
anthonyshew, line 1: This one feels like it doesn't have enough to it to need a page. Can we write this one to the terminal? Seems like it should only take a "You must provide a path for the `--cwd` flag.", no?

docs/pages/messages/invalid-env-prefix.mdx
```diff
@@ -0,0 +1,27 @@
+---
+title: Invalid Environment Variable Prefix
+description: "invalid-env-prefix error page"
+---
+
+# Invalid Environment Variable Prefix
+
+When declaring environment variables in your `turbo.json`, you cannot prefix them with `$`. This
+was an old syntax for declaring a dependency on an environment variable that is no longer used.
```
anthonyshew, line 9: "This syntax was deprecated in 1.5. You can migrate to the `env` and `globalEnv` keys using npx `@turbo/codemod migrate-env-var-dependencies`. [Check out the codemod's documentation](https://turbo.build/repo/docs/reference/codemods#migrate-env-var-dependencies) for more.

docs/pages/messages/multiple-cwd.mdx
anthonyshew, line 1: This one also feels like it's a one-liner in my terminal.

"You've provided two `--cwd` flags. Please only use one."

docs/pages/messages/package-task-in-single-package-mode.mdx
```diff
@@ -0,0 +1,37 @@
...
+{
+  "pipeline": {
+    "build": {
+      "cache": true
+    }
+  }
+}
+```
+
+Alternatively, you can disable single package mode by [specifying the workspaces in the repository](/repo/docs/getting-started/existing-monorepo#configure-workspaces).
```
anthonyshew, line 37: This line seems droppable. I *can* do this but it's probably not what I meant to do if I run into this error while I'm working in a single-package workspace.

docs/pages/messages/turbo-json-parse-error.mdx
```diff
@@ -0,0 +1,53 @@
+---
+title: `turbo.json` Parse Error
```
anthonyshew, line 2: We use sentence casing so "`turbo.json` parse error".

(Well, at least we're supposed to. This is one of many things that will get cleaned in our docs on the `docs-new` branch.

docs/pages/messages/unnecessary-package-task-syntax.mdx
```diff
@@ -0,0 +1,42 @@
+---
+title: Unnecessary Package Task Syntax
```
anthonyshew, line 2: Sentence casing.

docs/pages/messages/unnecessary-package-task-syntax.mdx
anthonyshew, line 1: This is talking about Workspace Configurations, right? Can we say that word on this page just for clarity's sake?

docs/pages/messages/package-task-in-single-package-mode.mdx
```diff
@@ -0,0 +1,37 @@
+---
+title: Package Task In Single Package Mode
```
anthonyshew, line 2: Sentence case.

docs/pages/messages/invalid-env-prefix.mdx
```diff
@@ -0,0 +1,27 @@
+---
+title: Invalid Environment Variable Prefix
```
anthonyshew, line 2: Sentence case.

anthonyshew (changes requested):
This is going to be great.

A couple ideological things:
1. Using [this](https://nextjs.org/docs/messages/react-hydration-error) as inspiration, let's change the "Example" header to be "Why this error occurred."
2. Let's only use this error messaging if we have something substantial to say - to the point that if what we want to convey just can't be said in the terminal.
3. The `description` field in the header is used as an OG description so it needs to be humanreadable. It will come up on OGs on socials and search results.

docs/pages/messages/turbo-json-parse-error.mdx
NicholasLYang: I'm gonna drop this one too tbh. The error will be more descriptive than this page because it'll actually show the real syntax error

docs/pages/messages/package-task-in-single-package-mode.mdx
NicholasLYang: In my experience, turbo can sometimes accidentally infer a single package repository, and this will be the first error that you will encounter in that situation. So it might be good to mention this

docs/pages/messages/invalid-env-prefix.mdx
```diff
@@ -0,0 +1,27 @@
...
+description: "invalid-env-prefix error page"
+---
+
+# Invalid Environment Variable Prefix
+
+When declaring environment variables in your `turbo.json`, you cannot prefix them with `$`. This
+was an old syntax for declaring a dependency on an environment variable that is no longer used.
+
+## Example
```
NicholasLYang, line 11: Should we change this to "Why this error occurred"? There isn't really a why here, just an example @anthonyshew

docs/pages/messages/unnecessary-package-task-syntax.mdx
NicholasLYang: Looking at the Next errors, these titles are actual sentences, versus the fragments that we have here. I'd learn towards title capitalization if we're keeping the fragments, otherwise we should make these full sentences

docs/pages/messages/package-task-in-single-package-mode.mdx
anthonyshew: Ah, okay, that makes sense. In that case, would it make more sense to make it more in the positive lane. Like: "Alternatively, you can [configure workspaces for your repository] (link to doc)." Thinking I probably would've caught on better if it were that direction.

docs/pages/messages/turbo-json-parse-error.mdx
anthonyshew: Beauuuuuty. I was wondering if that was going to be the case.

docs/pages/messages/invalid-env-prefix.mdx
anthonyshew: If the header moves above the paragraph, it will make sense that way!

docs/pages/messages/unnecessary-package-task-syntax.mdx
anthonyshew: It's more of a general Vercel style guide thing like here: https://vercel.com/blog/introducing-conformance#all-while-elevating-application-security

anthonyshew: (The Next.js errors are actually wrong, according to the style guide.)

docs/pages/messages/unnecessary-package-task-syntax.mdx
```diff
@@ -0,0 +1,42 @@
...
+title: Unnecessary package task syntax
+description: Unnecessary package task syntax error
+---
+
+# Unnecessary package task syntax
+
+## Why This Error Occurred
+
+Turborepo supports adding additional `turbo.json` files in a package directory
+to override the `turbo.json` file declared at the repository root, a feature called Workspace Configurations.
```
anthonyshew, line 11:
```suggestion
to override the `turbo.json` file declared at the repository root, a feature called [Workspace Configurations](/repo/docs/core-concepts/monorepos/configuring-workspaces).
```

docs/pages/messages/unnecessary-package-task-syntax.mdx
```diff
@@ -0,0 +1,42 @@
...
+In those additional `turbo.json` files, you can only configure tasks for that specific
+package. Therefore, only the task name should be included in the pipeline,
+not the package and task name (`package#task`).
+
+`turbo.json` file in `apps/web` directory:
+
+```
```
anthonyshew, line 18:
```suggestion
```json
```

anthonyshew: Just gives it pretty colors. :)

docs/pages/messages/unnecessary-package-task-syntax.mdx
```diff
@@ -0,0 +1,42 @@
...
+}
+```
+
+Since this `turbo.json` file is inside a package directory, the `web` prefix is unnecessary.
+
+## Solution
+
+Remove the package prefix from the task name:
+
+```
```
anthonyshew, line 34:
```suggestion
```json
```

docs/pages/messages/_meta.json
```diff
@@ -0,0 +1,8 @@
+{
+  "empty-cwd": "Empty `--cwd` flag",
```
anthonyshew, line 2:
```suggestion
```

docs/pages/messages/_meta.json
```diff
@@ -0,0 +1,8 @@
...
+  "invalid-env-prefix": "Invalid Environment Variable Prefix",
+  "multiple-cwd": "Multiple `--cwd` flags",
```
anthonyshew, line 4:
```suggestion
```

docs/pages/messages/invalid-env-prefix.mdx
```diff
@@ -0,0 +1,32 @@
...
+When declaring environment variables in your `turbo.json`, you cannot prefix them with `$`. This
+was an old syntax for declaring a dependency on an environment variable that was deprecated in Turbo 1.5.
+
+```json
+{
+  "globalEnv": ["$MY_ENV_VAR"]
+}
+```
+
+The environment variable declared above has the `$` prefix
```
anthonyshew, line 19:
```suggestion
The environment variable declared above has the `$` prefix.
```

anthonyshew (approved):
Just some nitpicks for fix-shipping. Excited for these!

PR #7034, opened 2024-01-19
chore: prefactor process manager

chris-olszewski:
### Description

This PR lays some ground work for PTY support by hiding implementation details of building the command from users of the `ProcessManager`. The public interface for `ProcessManager` and `Child` after this PR no longer use any part of `tokio::process`.

This PR adds our own `Command` struct that is intended to be used with `ProcessManager::spawn` instead of the raw command builder type which in the future might be `portable_pty::CommandBuilder`. This approach offers a few benefits:
- It allows us to limit the API we expose for command building to only options supported by both regular child processes and child processes hooked up to a pseudo-terminal.
- Commands can be decoupled from the spawning strategy
- Allows us to avoid running into the [orphan rule](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types)

⚠️ Behavior change: we will now not have our child processes inherit our `stdin`. This was never intended behavior, but an oversight of us not explicitly calling `stdin(Stdio::piped())` or `stdin(Stdio::null())`.

### Testing Instructions

Existing test suite and 👀 

Each commit should be easily reviewable on its own.

Closes TURBO-2070


crates/turborepo-lib/src/process/child.rs
```diff
@@ -362,26 +354,23 @@ impl Child {
...
 
     /// Wait for the `Child` to exit and pipe any stdout and stderr to the
-    /// provided writers.
-    /// If `None` is passed for stderr then all output produced will be piped
-    /// to stdout
+    /// provided writer.
     pub async fn wait_with_piped_outputs<W: Write>(
         &mut self,
         mut stdout_pipe: W,
-        mut stderr_pipe: Option<W>,
```
chris-olszewski, line 384: Independently capturing stderr wasn't an used outside of tests and can't be easily done with processes that are hooked up to a pseudo-terminal (possible, but involves redirecting stderr to a file and streaming that). In preparation we remove this feature.

crates/turborepo-lib/src/process/command.rs
```diff
@@ -0,0 +1,130 @@
...
+            .envs(env)
+            // We always pipe stdout/stderr to allow us to capture task output
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            // Only open stdin if configured to do so
+            .stdin(if open_stdin {
+                Stdio::piped()
+            } else {
+                Stdio::null()
+            });
```
chris-olszewski, line 124: This is a behavior change as we will no longer have our children inherit our stdin

chris-olszewski (commented):
Quick notes

crates/turborepo-lib/src/process/command.rs
```diff
@@ -0,0 +1,130 @@
...
+            .collect();
+        self
+    }
+
+    pub fn current_dir(&mut self, dir: AbsoluteSystemPathBuf) -> &mut Self {
+        self.cwd = Some(dir);
+        self
+    }
+
+    pub fn envs<I, K, V>(&mut self, vars: I) -> &mut Self
```
gsoltis, line 51: Given that we also have `env`, and `envs` squashes the environment, maybe we want `set_env` and `append_env`? Or `replace_env`? Or something to indicate that one squashes and the other is additive?

crates/turborepo-lib/src/process/command.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /// Configure if the child process should spawn with stdin or not
+    pub fn open_stdin(&mut self, open_stdin: bool) -> &mut Self {
+        self.open_stdin = open_stdin;
+        self
+    }
+
+    /// Clears the environment variables for the child process
+    pub fn env_clear(&mut self) -> &mut Self {
```
gsoltis, line 81: Along with above, I think this naming is confusing. It doesn't clear `self.env`, so we might want to think about the naming.

Along similar lines, `open_stdin` takes a `bool` and sets a `bool`. This function takes nothing and sets a `bool`. Should we change one or the other?

gsoltis (approved):
Just some minor API concerns, but otherwise LGTM.

crates/turborepo-lib/src/process/command.rs
chris-olszewski: I'm attempting to match same API that is exposed by the [stdlib](https://doc.rust-lang.org/std/process/struct.Command.html#method.envs)/[tokio](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.envs). Updated the implementation to match the behavior of these libraries where this is an additive function.

crates/turborepo-lib/src/process/command.rs
chris-olszewski: Similar to above, I'm trying to match the API of stdlib/tokio.

Updated `open_stdin` to no longer take a bool.

crates/turborepo-lib/src/process/command.rs
mehulkar: (but that matches the Go eversion)

PR #7053, opened 2024-01-20
Bumping examples to 1.12

anthonyshew:
### Description

Bumping dependencies in examples.

Closes TURBO-2182


examples/with-rollup/package.json
```diff
@@ -1,9 +1,5 @@
 {
   "private": true,
-  "workspaces": [
-    "apps/*",
-    "packages/*"
-  ],
```
chris-olszewski, line 6: I think we still want workspaces in this example?

examples/with-rollup/package.json
```diff
@@ -1,9 +1,5 @@
...
   "scripts": {
```
chris-olszewski, line 3: Also bump the turbo version specifier to `^1.12.1`

chris-olszewski (approved):
Stamping there's just a minor fixup

examples/with-rollup/package.json
anthonyshew: It's pnpm! Fixing what shouldn't have been there in the first place.

Issue #4124, opened 2023-03-08
Consider the used Node.js version for generating hashes (or provide a way to configure runtime hash inputs)


ghost91-:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

At the moment, Turborepo does not consider the used Node.js as hash inputs. This can lead to undesirable situations where CI is set up to execute tasks in multiple different Node.js versions, to ensure that the package works in all of them, but then the task is only ever _really_ executed in one of them, because in the others, it's fetched from the cache.

For this reason, the used Node.js version should be considered an input for every generated hash.

Of course, depending on the repo setup, there could be a lot more runtime context that should be considered a hash input (e.g. arch, OS, etc.). For that reason, it would also make sense to allow users to configure arbitrary runtime inputs, similar to `globalEnv` and `globalDependencies` on the global level and `env` on the per-task level.

For comparison and reference: Nx already provides this feature: https://nx.dev/concepts/how-caching-works#runtime-hash-inputs

### Describe the solution you'd like

Ideally, there is a new optional configuration property `globalRuntime` where users can specify a list of commands to be executed. When turbo is run, those commands are executed and their output is used as an input for all hashes. If the property is not given, Turborepo uses just the command `node -v`.

Example of how the config could look:
```json
{
  "$schema": "https://turbo.build/schema.json",
  "pipeline": {
    // ... omitted for brevity
  },
 
  "globalRuntime": ["node -v"] // result of executing this will impact the hashes of all tasks
}
```

Similarly, there is a `runtime` property that can be used in the configuration of tasks to specify the runtime inputs for a specific task.

### Describe alternatives you've considered

One alternative would be put the result of `node -v` into an environment variable and use that as a has input via `globalEnv`. However, that is significant effort and some people might simply forget to do it if they execute `turbo` directly.

We could also write our own small wrapper around `turbo` that does nothing but put runtime information into an environment variable and then call turbo. But that's also quite cumbersome and error prone.

In total, this just feels like something Turborepo should take care of.

arlyon:
This issue pops up frequently and I agree that we need to allow users to be able to opt-in to this. We are going to come up with an answer here and I'll update this issue.

ghost91-:
Thanks for considering this, I highly appreciate that!

ghost91-:
[...]

Issue #1315, opened 2022-05-31
Provide a granular way to depend on OS / Architecture


gsoltis:
### Describe the feature you'd like to request

Some tasks, particularly tests and builds for non-frontend, non-backend binaries, behave differently on different OSes and architectures. `turbo` should be able to be easily configured to not cache results across platforms / architectures

### Describe the solution you'd like

Most likely something in `turbo.json` to apply either monorepo-wide or per-task.

### Describe alternatives you've considered

Environment variables can be used, but require every person or process who interacts with `turbo` to have them set appropriately.

<sub>[TURBO-2047](https://linear.app/vercel/issue/TURBO-2047/provide-a-granular-way-to-depend-on-os-architecture)</sub>

mehulkar:
Related: #4124

anthonyshew:
Documented in #7072

PR #7072, opened 2024-01-22
Document handling of Node.js versions, OS, and arch.

anthonyshew:
### Description

While we may consider OS and arch handling as a feature in the future, we don't today. Here, we're documenting how to handle this advanced use case as well as jotting down a proper way of handling Node.js versions.

Closes https://github.com/vercel/turbo/issues/4124 and https://github.com/vercel/turbo/issues/1315


anthonyshew:
Closes #4124 and #1315

docs/pages/repo/docs/core-concepts/caching.mdx
```diff
@@ -83,6 +83,69 @@ turbo run build --force
...
+
+You won't want to commit this file to source control since it's dependent on environment. Add it to your `.gitignore`:
+
+```txt
+// .gitignore
+
++ turbo-cache-key.json
+```
+
+### 3. Add the file to `globalDependencies`
```
mehulkar, line 121: I would make this a little less prescriptive. It may be that this type of advanced use case only applies to one task (e.g. build and not lint). I would advise the user to be judicious about specifying it as an input where it matters)

docs/pages/repo/docs/core-concepts/caching.mdx
```diff
@@ -83,6 +83,85 @@ turbo run build --force
...
+```json
+{
+  "$schema": "https://turbo.build/schema.json",
+  "pipelines": {
+    "build-for-platforms": {
+      "dependsOn": ["^build"],
+      "inputs": ["**/**", "turbo-cache-key.json"]
+    }
+  }
+}
```
tknickman, line 136: We might want to hold off on this recommendation until we have support for the feature to add a single file to the rest of the turbo default inputs. This will alter behavior in an unexpected way since this will include everything (even git ignored) files.

docs/pages/repo/docs/core-concepts/caching.mdx
```diff
@@ -83,6 +83,85 @@ turbo run build --force
...
+First, create a script that accounts for the hash contributors that you are interested in. For example, here is a Node.js script that looks at platform and architecture and writes those details to a file (`turbo-cache-key.json`):
+
+```js
+#!/usr/bin/env node
+
+const { writeFileSync } = require('fs');
+const { join } = require('path');
+
+const { platform, arch } = process;
+const file = join(__dirname, '..', 'turbo-cache-key.json');
```
tknickman, line 105: We should explain this, what directory _should_ it be in, package root? monorepo root?

docs/pages/repo/docs/core-concepts/caching.mdx
tknickman: I can see I contradicted @mehulkar here haha. I think this is fine to leave, but we should be clear about how this isn't 1:1 with existing behavior if you previously didn't have inputs.

mehulkar: not contradicting! I think your point is additive and really good. Maybe something like "note that caveats of `inputs` config apply here if you do this, etc etc"

docs/pages/repo/docs/core-concepts/caching.mdx
```diff
@@ -83,6 +83,85 @@ turbo run build --force
 
 Note that `--force` disables cache reads but does not disable cache writes. If you want to disable cache writes, use the `--no-cache` flag.
 
+## Handling Node.js versions
+
+To account for Node.js versions, use [the `engines` key in package.json](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines). Turborepo will see the changes to your `package.json` and miss cache when the field is updated.
+
+## Handling platforms and other arbitrary hash contributors
+
+For advanced use cases, you may need to account for operating systems (OS), architecture, or other external factors that you want to contribute to your hash. You can do this in four quick steps.
```
NicholasLYang, line 92:
```suggestion
For advanced use cases, you may want the operating system (OS), architecture, or other external factors to contribute to your hash. You can do this in four quick steps.
```

docs/pages/repo/docs/core-concepts/caching.mdx
```diff
@@ -83,6 +83,85 @@ turbo run build --force
...
+const { platform, arch } = process;
+const file = join(__dirname, '..', 'turbo-cache-key.json');
+const str = JSON.stringify({ platform, arch });
+console.log(`Generating cache key: ${str}`);
+writeFileSync(file, str);
+```
+
+### 2. Add the file to your .gitignore
+
+You won't want to commit this file to source control since it's dependent on environment. Add it to your `.gitignore`:
```
NicholasLYang, line 113: ngl it feels a little weird having code that isn't committed but is expected to be used on occasion. Could we create a CLI tool that does this for people? So they could just do `npx turbo-fingerprint` or whatever

NicholasLYang (changes requested):

docs/pages/repo/docs/core-concepts/caching.mdx
anthonyshew: As nice as that would be, we need to document the current state of things. This documentation does bring to light some functionality gaps, and we should fix them, but we need to have documentation in the meantime.

docs/pages/repo/docs/core-concepts/caching.mdx
anthonyshew: As nice as that would be, we need to document the current state of things. This documentation does bring to light some functionality gaps, and we should fix them, but we need to have documentation in the meantime.

The inputs documentation is linked so those caveats are accounted for. If they aren't, then the `inputs` doc needs to improve.

docs/pages/repo/docs/core-concepts/caching.mdx
anthonyshew: Too perscriptive. Fixed it for the opposite, just naming a file.

docs/pages/repo/docs/core-concepts/caching.mdx
tknickman: I think this is a little different though. This feels like we're advocating for using `"**/**"` in inputs as a replacement for the default hashing, but it's not going to do what you think it will.

mehulkar (commented):
I took a look again, and I'd merge this if it wasn't in Core Concepts. Maybe FAQ is a better place?

This isn't a Core Concept, IMO.

docs/pages/repo/docs/core-concepts/caching.mdx
anthonyshew: Given the current, available API, I don't see us having much of an option. I agree it's not ideal.

tknickman: Agreed, I think we just need to add a comment or a callout or something here to say that (we also should have this feature 🔜 )

anthonyshew:
The structural issues of our current documentation are being addressed on the `docs-new` branch. We'll be getting rid of the FAQ entirely.

docs/pages/repo/docs/core-concepts/caching.mdx
anthonyshew: We'll update the `inputs` doc when that lands. 👍

NicholasLYang (approved):

tknickman (approved):

PR #7074, opened 2024-01-23
feat(turbo): task access trace & cached configs

tknickman:
### Description

v0 of work required for zero config (experimental):
1. Support config caching
2. Support task tracing

Requires: https://github.com/vercel/next.js/pull/59553

#### TODO
- [ ] Testing
- [ ] Support caching outputs on first run instead of second (turbo.json uncoupling)

### DIFF
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -400,33 +400,36 @@ impl ConfigCache {
         }
     }
 
+    pub fn hash(&self) -> &str {
+        &self.hash
+    }
+
     pub fn exists(&self) -> bool {
-        return self.config_file.try_exists().unwrap_or(false);
+        self.config_file.try_exists().unwrap_or(false)
     }
 
     pub async fn restore(
         &self,
     ) -> Result<Option<(CacheHitMetadata, Vec<AnchoredSystemPathBuf>)>, CacheError> {
-        return self.cache.fetch(&self.repo_root, &self.hash).await;
+        self.cache.fetch(&self.repo_root, &self.hash).await
     }
 
     pub async fn save(&self) -> Result<(), CacheError> {
         match self.exists() {
             true => {
                 debug!("config file exists, caching");
-                return self
-                    .cache
+                self.cache
                     .put(
                         self.repo_root.clone(),
                         self.hash.clone(),
                         vec![self.anchored_path.clone()],
                         0,
                     )
-                    .await;
+                    .await
             }
             false => {
                 debug!("config file does not exist, skipping cache save");
-                return Ok(());
+                Ok(())
             }
         }
     }
@@ -445,11 +448,11 @@ impl ConfigCache {
 
         // empty inputs to get all files
         let inputs: Vec<String> = vec![];
-        let hash_object =
-            match scm.get_package_file_hashes(&repo_root, &anchored_root, &inputs, None) {
-                Ok(hash_object) => hash_object,
-                Err(_) => return Err(CacheError::ConfigCacheError),
-            };
+        let hash_object = match scm.get_package_file_hashes(repo_root, anchored_root, &inputs, None)
+        {
+            Ok(hash_object) => hash_object,
+            Err(_) => return Err(CacheError::ConfigCacheError),
+        };
 
         // return the hash
         Ok(FileHashes(hash_object).hash())
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -44,7 +44,7 @@ pub enum Error {
 
 pub struct RunCache {
     task_output_mode: Option<OutputLogsMode>,
-    cache: Arc<AsyncCache>,
+    cache: AsyncCache,
     reads_disabled: bool,
     writes_disabled: bool,
     repo_root: AbsoluteSystemPathBuf,
@@ -55,7 +55,7 @@ pub struct RunCache {
 
 impl RunCache {
     pub fn new(
-        cache: Arc<AsyncCache>,
+        cache: AsyncCache,
         repo_root: &AbsoluteSystemPath,
         opts: &RunCacheOpts,
         color_selector: ColorSelector,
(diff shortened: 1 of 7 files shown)


crates/turborepo-lib/src/run/cache.rs
```diff
@@ -374,3 +382,88 @@ impl TaskCache {
...
+    }
+
+    pub async fn restore(
+        &self,
+    ) -> Result<Option<(CacheHitMetadata, Vec<AnchoredSystemPathBuf>)>, CacheError> {
+        self.cache.fetch(&self.repo_root, &self.hash).await
+    }
+
+    pub async fn save(&self) -> Result<(), CacheError> {
+        match self.exists() {
```
chris-olszewski, line 427: There's a TOCTOU race condition where the file gets cleaned up between us checking for it's existence and it actually getting read into memory.

crates/turborepo-lib/src/run/task_access.rs
```diff
@@ -0,0 +1,244 @@
...
+
+fn task_access_trace_enabled(repo_root: &AbsoluteSystemPathBuf) -> Result<bool, Error> {
+    // TODO: use the existing config methods here
+    let root_turbo_json_path = &repo_root.join_component(TURBO_CONFIG_FILE);
+    if root_turbo_json_path.exists() {
+        return Ok(false);
+    }
+
+    // read package.json at root
+    let package_json_path = repo_root.join_components(&["package.json"]);
```
chris-olszewski, line 62: Is this feature only intended to be used in single package workspaces?

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -838,21 +860,49 @@ impl ExecContext {
...
+                        {
+                            error!("error caching output: {e}");
+                        } else {
+                            // If no errors, update hash tracker with expanded outputs
+                            self.hash_tracker.insert_expanded_outputs(
+                                self.task_id.clone(),
+                                self.task_cache.expanded_outputs().to_vec(),
+                            );
+                        }
+                    }
```
chris-olszewski, line 902: This "can we cache" logic feels like it should belong totally in the task access module:
```
    // Whether we can cache the given task, returning None if task access isn't
    // enabled or the trace can't be found
    pub fn can_cache(&self, task_hash: &str, task_id: &str) -> Option<bool> {
        if !self.is_enabled() {
            return None;
        }
        let trace = TaskAccessTraceFile::read(&self.repo_root, task_hash)?;
        if trace.can_cache(&self.repo_root) {
            self.save_trace(trace, task_id.to_string());
            Some(true)
        } else {
            Some(false)
        }
    }
```
Which simplifies the changes in Visitor to
```suggestion
                   if self
                        .task_access
                        .can_cache(&self.task_hash, &self.task_id_for_display)
                        .unwrap_or(true)
                    {
                        if let Err(e) = self
                            .task_cache
                            .save_outputs(&mut prefixed_ui, task_duration, telemetry)
                            .await
                        {
                            error!("error caching output: {e}");
                        } else {
                            // If no errors, update hash tracker with expanded outputs
                            self.hash_tracker.insert_expanded_outputs(
                                self.task_id.clone(),
                                self.task_cache.expanded_outputs().to_vec(),
                            );
                        }
                    }
```

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -838,21 +860,49 @@ impl ExecContext {
...
+                    if self.task_access.is_enabled() {
+                        // Task access enabled: Handle trace file processing
+                        if let Some(trace) =
+                            TaskAccessTraceFile::read(&self.task_access.repo_root, &self.task_hash)
+                        {
+                            // Trace file found
+                            if trace.can_cache(&self.task_access.repo_root) {
+                                // Cache trace and save task outputs if trace is cacheable
+                                self.task_access
+                                    .save_trace(trace, self.task_id_for_display.to_string());
```
chris-olszewski, line 878: Just making sure my understanding is correct: we want this to ensure we're getting a valid `TaskName` that can appear in a pipeline definition so we're using `task_id_for_display` as it'll end up being a valid pipeline task name?

chris-olszewski (changes requested):
Mostly concerned about how this works with multi-package workspaces.

crates/turborepo-lib/src/run/cache.rs
```diff
@@ -32,11 +36,15 @@ pub enum Error {
...
+    #[error(transparent)]
+    Scm(#[from] turborepo_scm::Error),
+    #[error(transparent)]
+    Path(#[from] turbopath::PathError),
 }
 
 pub struct RunCache {
     task_output_mode: Option<OutputLogsMode>,
-    cache: AsyncCache,
+    cache: Arc<AsyncCache>,
```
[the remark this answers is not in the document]
tknickman: Good call, broke out that PR into it's own: https://github.com/vercel/turbo/pull/7293

crates/turborepo-lib/src/run/task_access.rs
tknickman: For now, yes (but we'll roll this out more broadly soon)

crates/turborepo-lib/src/task_graph/visitor.rs
tknickman: Yep that's correct, a little weird in this case, but we need that to build the pipeline later on.

tknickman:
@chris-olszewski PR updated.

There were a lot of conflicts here, so split the commits between resolving conflicts, and addressing review. But let me know if you have any questions!

NOTE: This is stacked on top of https://github.com/vercel/turbo/pull/7293

crates/turborepo-lib/src/run/cache.rs
chris-olszewski: Tracing through this, I don't think there's an easy way around it unless we rework the cache abstraction a fair bit. I think it's fine to leave as is 🤷

chris-olszewski (approved):
LFG 🚀

PR #7076, opened 2024-01-23
Document glob spec.

anthonyshew:
### Description

Title!

One note I'll add is that I think it's wise for us to not go incredibly, incredibly deep here. I want to cover the 80% cases and leave the really exotic globs that one could build up to the `wax` documentation. If you think I've left too much unsaid, happy to add (but do feed me the pattern and example, please).


docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -437,4 +437,34 @@ option, `turbo` can warn you about an invalid configuration.
...
+## Glob specification for paths
+
+Turborepo's glob implementation is based on [wax](https://glob.guide/user-guide/syntax/). A few of the most common patterns you will need are in the table below:
+
+| Pattern             | Description                                                                        |
+| ------------------- | ---------------------------------------------------------------------------------- |
+| `*`                 | Match all files in the directory                                                   |
+| `**`                | Deeply match all files and sub-directories                                         |
+| `some-dir/`         | Match the `some-dir` directory **but not its contents**                            |
+| `some-dir*`         | Match files and directories that start with `some-dir`                             |
```
mehulkar, line 449: will the contents of these directories be matched? how deeply?

docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -437,4 +437,34 @@ option, `turbo` can warn you about an invalid configuration.
...
+| `*.js`              | Matches all `.js` files in the directory                                           |
+| `*.{js,ts,tsx,jsx}` | Matches all files that end with the provided extensions in the directory           |
+| `*.{js?x}`          | Matches all files with the `js` and `jsx` extensions in the directory              |
+| `!`                 | Negate the whole glob (automatically applies `/**` to the end of the defined glob) |
+
+Here are a few examples of how to use these patterns:
```
mehulkar, line 455: Want to add an example of `!` also?

docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -437,4 +437,34 @@ option, `turbo` can warn you about an invalid configuration.
...
+| `dist/**`          | Match all files in the `dist` directory and all sub-directories                                   |
+| `dist/some-dir/**` | Match all files in the `dist/some-dir` directory and all sub-directories in the current directory |
+| `dist/`            | Match the `dist` directory **but not its contents**                                               |
+| `dist`             | Match the `dist` directory **but not its contents**                                               |
+| `dist*`            | Match files and directories that start with `dist`                                                |
+| `dist/*.js`        | Matches all `.js` files in the `dist` directory                                                   |
+| `dist/**/*.js`     | Matches all deeply nested `.js` files in the `dist` directory                                     |
+| `../scripts/**`    | Up one directory, match all files and sub-directories in the `scripts` directory                  |
+
+For more advanced globbing patterns, see the [wax glob guide](https://glob.guide/user-guide/syntax/).
```
mehulkar, line 468: I would recommend not delegating to a 3rd party implementation, we don't want to teach this. We had an outdated reference to a JS implementation for a long time, even while the implementation was in Go. We should make our guarantees of what works and give ourselves flexibility to add/remove things as needed.

anthonyshew: I was hesitant about this play as well. This was an effort towards keeping the signal to noise ratio high. The tradeoff I saw was that we could document every little microsyntax minutae but then we confuse the folks that are doing normal stuff in the 95% case.

I want to reassess this once I hear from @arlyon about how many "Advanced Patterns" there are (hm, perhaps that's a sub-section to break it away from the 95% case but still get it docced by us first-party).

mehulkar: I would only document the surface area that we happily support, have tests for, and intend for people to use. It's ok if that's not everything that will work if you understand the implementation. We use libraries to make maintenance easier and to avoid re-solving solved problems, not to be 1:1 with their capabilities. I'd even go as far as to say that we should _prevent_ using the full power of the underlying implementation (and obviously grow and expand as use cases become more understood!)

arlyon: Wax is pretty restrictive compared to other glob engines we've used. There are the natural extensions of the existing syntax such as {apps/*,packages/*}/**.{t,j}s but that is just a composition of the core rules.

arlyon (approved):

PR #7077, opened 2024-01-23
feat: add ability to spawn child process with pty

chris-olszewski:
### Description

This PR adds, but doesn't hook up, the ability to spawn a child process hooked up to a pseudo terminal.

I *highly* suggest reviewing each commit on its own. I tried my best to split out prefactoring work from any feature work.

### Testing Instructions

Adapt unit tests to cover all of them. It will be noted that switching to use a PTY does affect what programs output and so the tests were made to not check for identical output, but just that we got the expected output.
Closes TURBO-2128

### DIFF
--- a/crates/turborepo-lib/src/process/child.rs
+++ b/crates/turborepo-lib/src/process/child.rs
@@ -24,9 +24,7 @@ use std::{
     time::Duration,
 };
 
-use portable_pty::{
-    native_pty_system, Child as PtyChild, MasterPty as PtyController, SlavePty as PtyReceiver,
-};
+use portable_pty::{native_pty_system, Child as PtyChild, MasterPty as PtyController};
 use tokio::{
     io::{AsyncBufRead, AsyncBufReadExt, BufReader},
     join,
@@ -90,12 +88,7 @@ struct ChildHandle {
 
 enum ChildHandleImpl {
     Tokio(tokio::process::Child),
-    Pty {
-        // Dropping this will close the receiver file descriptor
-        // of the pty pair
-        receiver: Box<dyn PtyReceiver + Send>,
-        child: Box<dyn PtyChild + Send + Sync>,
-    },
+    Pty(Box<dyn PtyChild + Send + Sync>),
 }
 
 impl ChildHandle {
@@ -174,7 +167,7 @@ impl ChildHandle {
         Ok((
             Self {
                 pid,
-                imp: ChildHandleImpl::Pty { receiver, child },
+                imp: ChildHandleImpl::Pty(child),
             },
             ChildIO {
                 controller: Some(controller),
@@ -192,7 +185,7 @@ impl ChildHandle {
     pub async fn wait(&mut self) -> io::Result<Option<i32>> {
         match &mut self.imp {
             ChildHandleImpl::Tokio(child) => child.wait().await.map(|status| status.code()),
-            ChildHandleImpl::Pty { child, .. } => {
+            ChildHandleImpl::Pty(child) => {
                 // TODO: we currently poll the child to see if it has finished yet which is less
                 // than ideal
                 loop {
@@ -214,7 +207,7 @@ impl ChildHandle {
     pub async fn kill(&mut self) -> io::Result<()> {
         match &mut self.imp {
             ChildHandleImpl::Tokio(child) => child.kill().await,
-            ChildHandleImpl::Pty { child, .. } => {
+            ChildHandleImpl::Pty(child) => {
                 let mut killer = child.clone_killer();
                 tokio::task::spawn_blocking(move || killer.kill())
                     .await
@@ -446,8 +439,9 @@ impl Child {
         let task_state = state.clone();
 
         let _task = tokio::spawn(async move {
-            // TODO: Not sure when this FD should be closed
-            let _controller = controller;
+            // On Windows it is important that this gets dropped once the child process
+            // exits
+            let controller = controller;
             debug!("waiting for task");
             let manager = ChildStateManager {
                 shutdown_style,
@@ -456,9 +450,10 @@ impl Child {
             };
             tokio::select! {
                 command = command_rx.recv() => {
-                    manager.handle_child_command(command, &mut child).await;
+                    manager.handle_child_command(command, &mut child, controller).await;
                 }
                 status = child.wait() => {
+                    drop(controller);
                     manager.handle_child_exit(status).await;
                 }
             }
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/process/child.rs
```diff
@@ -77,27 +83,242 @@ impl From<std::io::Error> for ShutdownFailed {
...
+        match &mut self.imp {
+            ChildHandleImpl::Tokio(child) => child.wait().await.map(|status| status.code()),
+            ChildHandleImpl::Pty { child, .. } => {
+                // TODO: we currently poll the child to see if it has finished yet which is less
+                // than ideal
+                loop {
+                    match child.try_wait() {
+                        // This is safe as the portable_pty::ExitStatus's exit code is just
+                        // converted from a i32 to an u32 before we get it
+                        Ok(Some(status)) => return Ok(Some(status.exit_code() as i32)),
```
chris-olszewski, line 202: At the moment, `portable_pty` converts exit statuses caused by signals to have an exit code `1` and it captures the signal that the process received. At the moment there's no way to extract this signal aside from displaying the `portable_pty::ExitStatus`. I'll put an upstream PR so we can differentiate between non-zero exits and signals.

crates/turborepo-lib/src/process/child.rs
```diff
@@ -77,27 +83,242 @@ impl From<std::io::Error> for ShutdownFailed {
...
+    pub fn pid(&self) -> Option<u32> {
+        self.pid
+    }
+
+    pub async fn wait(&mut self) -> io::Result<Option<i32>> {
+        match &mut self.imp {
+            ChildHandleImpl::Tokio(child) => child.wait().await.map(|status| status.code()),
+            ChildHandleImpl::Pty { child, .. } => {
+                // TODO: we currently poll the child to see if it has finished yet which is less
+                // than ideal
```
chris-olszewski, line 197: I really don't like this as we're essentially just polling to see if the child is done, but I could not for the life of me figure out a good way to make a cancel-safe future that doesn't require full ownership of the child.

crates/turborepo-lib/src/process/child.rs
```diff
@@ -432,11 +706,65 @@ impl Child {
     }
 }
 
+impl ChildStateManager {
```
chris-olszewski, line 709: This struct is exclusively here to reduce the amount of code in `tokio::select` macro as it confuses `rust-analyzer` and even `rustc` making for poor development experience.

chris-olszewski (commented):
Reviewer's guide

crates/turborepo-lib/src/process/child.rs
```diff
@@ -379,23 +411,56 @@ impl Child {
...
+                self.wait_with_piped_async_outputs(
+                    stdout_pipe,
+                    Some(BufReader::new(stdout_lines)),
+                    Some(BufReader::new(stderr_lines)),
+                )
+                .await
+            }
+            // If using tokio to spawn tasks we should always be spawning with both stdout and
+            // stderr piped Being in this state indicates programmer error
+            (Some(ChildOutput::Concrete(_)), None) | (None, Some(ChildOutput::Concrete(_))) => {
```
mehulkar: [learning rust q]: is this idiomatic Rust? Reads very weird to use a `match` for this. In JS/other scripted languages it would be guard clauses like this.

```js
if (!stdout_lines) throw new Error()
if (!stderr_lines) throw new Error()

// proceed
```

crates/turborepo-lib/src/process/child.rs
```diff
@@ -275,57 +284,17 @@ impl Child {
...
-                    };
-                    {
-                        let mut task_state = task_state.write().await;
-                        *task_state = ChildState::Exited(child_exit);
-                    }
-
-                    // ignore the send error, the channel is dropped anyways
-                    exit_tx.send(Some(child_exit)).ok();
-
+                    manager.handle_child_exit(status).await;
```
mehulkar, line 458: [learn rust q]: interesting/unintuitive rust/tokio syntax
```rust
foo = something() => {
  // foo also available here
}
```

crates/turborepo-lib/src/process/command.rs
```diff
@@ -129,3 +134,28 @@ impl From<Command> for tokio::process::Command {
         cmd
     }
 }
+
+impl From<Command> for portable_pty::CommandBuilder {
```
mehulkar, line 138: [learn rust q]: is this considered monkeypatching?

crates/turborepo-lib/src/process/child.rs
```diff
@@ -107,42 +122,123 @@ impl ChildHandle {
...
                 stdout,
                 stderr,
             },
         ))
     }
 
+    pub fn spawn_pty(command: Command) -> io::Result<(Self, ChildIO)> {
+        use portable_pty::PtySize;
+
+        let keep_stdin_open = command.will_open_stdin();
```
mehulkar, line 139: self note: underlying field added here: https://github.com/vercel/turbo/pull/7034/files#diff-8cfa2f46e975fad0388fa6fa9bbe61ab386b4a072ebf9235bb77628b737faa6bR75

crates/turborepo-lib/src/process/mod.rs
```diff
@@ -65,7 +65,8 @@ impl ProcessManager {
         if lock.is_closing {
             return None;
         }
-        let child = child::Child::spawn(command, child::ShutdownStyle::Graceful(stop_timeout));
+        let child =
+            child::Child::spawn(command, child::ShutdownStyle::Graceful(stop_timeout), false);
```
mehulkar, line 69: note to self: 3rd argument is `use_pty` and its hardcoded to false at the moment

mehulkar (approved):
I'm grappling with a lot of the stdio/pty low level operations, but I followed the refactor process commit by commit and largely understand what's going on. Nice work!

crates/turborepo-lib/src/process/child.rs
chris-olszewski: Yeah, this wasn't super idiomatic due to how I was representing the possible child IO combinations which lead to a lot of copy/past in the match patterns.

Rust has `let else` expressions which can do the JS style guards e.g.
```
let Some(val) = x else {
    // this branch will only be executed if x doesn't match the pattern
    return Err("expected x to be present");
};
```

crates/turborepo-lib/src/process/child.rs
chris-olszewski: I am in full agreement, I find [`tokio::select`](https://docs.rs/tokio/latest/tokio/macro.select.html) to be unintuitive macro that causes a lot of problems for me.

PR #7078, opened 2024-01-23
feat: update `@turbo/repository` to use new terminology

mrmckeb:
### Description

This PR updates the language to match our new glossary:
https://vercel.com/docs/vercel-platform/glossary

### Testing Instructions

See README in package.


packages/turbo-repository/rust/src/lib.rs
```diff
@@ -64,14 +61,14 @@ impl PackageManagerRoot {
...
-        match self.repo_state.package_manager.as_ref() {
+        match self.workspace_state.package_manager.as_ref() {
             Ok(pm) => Ok((*pm).into()),
             Err(e) => Err(napi::Error::from_reason(format!("{}", e))),
         }
     }
 
     #[napi]
-    pub async fn packages(&self) -> std::result::Result<Vec<Package>, napi::Error> {
+    pub async fn find_packages(&self) -> std::result::Result<Vec<Package>, napi::Error> {
```
chris-olszewski, line 71: Is the thought behind adding the `find` that it better indicates that this function does work and isn't just a getter?

chris-olszewski (approved):

packages/turbo-repository/rust/src/lib.rs
mrmckeb: Is this just a getter? My understanding was that this async function was calling `package_manager.get_package_jsons()` underneath, which seems to traverse?

tknickman (approved):

PR #7085, opened 2024-01-23
chore(Turborepo): Move more CLI stuff out of run code

gsoltis:
### Description

 - setup `api_auth` and `api_client` outside of a run
 - remove the lifetime from `Opts` and related structs
 - keep a copy of `Opts` on `Run`. This is likely not the final structure, but is an intermediate step.
 - keep a copy of `api_auth` on `Run`
 - Move most of the calculations on `Opts` into the constructor for `Run`. This should likely eventually be a `TryFrom<CommandBase>` instead of `new`. The goal is to be able to make `Run` itself immutable since all of the config and options will have been resolved ahead of time

### Testing Instructions

Existing test suite

Closes TURBO-2135


crates/turborepo-lib/src/run/mod.rs
```diff
@@ -60,16 +60,41 @@ use crate::{
...
+            opts.cache_opts.skip_remote = true;
+        } else if let Some(enabled) = config.enabled {
+            // We're linked, but if the user has explicitly enabled or disabled, use that
+            // value
+            opts.cache_opts.skip_remote = !enabled;
+        }
+        // Note that we don't currently use the team_id value here. In the future, we
+        // should probably verify that we only use the signature value when the
+        // configured team_id matches the final resolved team_id.
+        let unused_remote_cache_opts_team_id = config.team_id().unwrap_or_default();
```
NicholasLYang, line 86: I don't love that we're flattening a None to an empty String here. It's not particular idiomatic to Rust.

NicholasLYang (changes requested):

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -99,9 +120,9 @@ impl Run {
             .then(|| start_analytics(api_auth, api_client))
     }
 
-    fn print_run_prelude(&self, opts: &Opts<'_>, filtered_pkgs: &HashSet<WorkspaceName>) {
-        let targets_list = opts.run_opts.tasks.join(", ");
-        if opts.run_opts.single_package {
+    fn print_run_prelude(&self, opts: &Opts, filtered_pkgs: &HashSet<WorkspaceName>) {
```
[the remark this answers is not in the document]
gsoltis: Yup. Will do.

crates/turborepo-lib/src/run/mod.rs
gsoltis: This is just moving around some existing logic. I'll see how invasive it is to switch to an `Option`, but I'll also note that that's not the biggest issue here. The real problem is we have an unused team id that we pull from the config.

gsoltis: Ended up not being too bad. It's now an `Option<String>` called `unused_team_id`.

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -1196,7 +1196,10 @@ pub async fn run(
             } else {
                 use crate::commands::run;
                 event.track_run_code_path(CodePath::Rust);
-                let exit_code = run::run(base, event).await?;
+                let exit_code = run::run(base, event).await.map_err(|e| {
+                    error!("run failed: {e}");
```
chris-olszewski, line 1200: ~~Will this cause us to stutter our errors?~~ 

I see that this logic moved from the run command itself. I don't care where it lives, but I feel it's easier to reason about if both error messages live close to each other instead of being spread across two functions. Maybe use `inspect_err` in the same way we use `inspect`?

NicholasLYang (approved):

chris-olszewski (approved):
Looks like a good step in the right direction. Minor thing about colocating error messaging, but otherwise 🚢

crates/turborepo-lib/src/cli/mod.rs
gsoltis: Good suggestion. Moved to `inspect_err`.

NicholasLYang: Eventually I want to remove all of this and have a single print out in `turborepo-lib/src/lib.rs` that prints the miette formatted version. That's in #6956

PR #7086, opened 2024-01-23
fix: Run examples serially

NicholasLYang:
### Description

We seem to be getting mysterious race conditions around running examples tests in parallel. Switching them to serial execution

### Testing Instructions

Closes TURBO-2136


anthonyshew (approved):
Fingers crossed.

PR #7087, opened 2024-01-23
fix(examples): ui turbo.json no longer needed

tknickman:
### Description

This is no longer needed since we aren't building ui

Closes TURBO-2137


anthonyshew (approved):
How did I never notice this in there before? 🧑‍🦯

PR #7089, opened 2024-01-24
backport wax 0.6 fixes to turborepo-wax

arlyon:
### Description

Merges upstream wax updates and updates to be compatible. Wax went through a number of changes between when we pulled it in and now, the most important commit of which being https://github.com/olson-sean-k/wax/commit/76afaa120ea58f14f16fec28fd5d1de6737d3d7f which replaces the filtering logic to handle an edge case around directory pruning. For more context, see https://github.com/olson-sean-k/wax/issues/49. I opted to just cherry pick everything we were missing since trying to pull out just that commit would be messy and cause us to diverge even more from upstream. After this PR we should be pretty much aligned, to the point where it may be worth using upstream again if we can land two changes:

- https://github.com/olson-sean-k/wax/pull/56
- https://github.com/vercel/turbo/pull/7089/commits/da2b2699f84e3b8ef4a78dc55dbbbd1a9f3ade1c

Probably should just do this commit by commit and ignore the wax cherry picks. Interesting ones below:

- https://github.com/vercel/turbo/pull/7089/commits/8322cb73a26cdc68cf25d4d5c72e94271bde89d5: upgrading wax means we can get rid of our special case handling for invariant globs (ie paths). it also splits out the large iterator into a few smaller functions and uses `tracing::instrument` instead of manually creating spans.
- https://github.com/vercel/turbo/pull/7089/commits/adc4e3da3fad617499729a65fa08b70fc89dd1d6: wax's new walking behaviour was not compatible with the way we glob. we rewrite all the globs to be absolute which caused issues in our tests. this commit fixes that, adds a test in wax for it, and fixes all of our turborepo-globwalk tests
- https://github.com/vercel/turbo/pull/7089/commits/de9f481b99d5f30d0e8f45d57a3abe5da3c1136b is related, we need to strip out the absolute marker before traversing to depth `n` otherwise absolute and relative globs would behave differently
- https://github.com/vercel/turbo/pull/7089/commits/714c0bb781bb1504a73c07913ba27b8da88bc956 there were a few trait changes, so just make sure to import the correct one
- https://github.com/vercel/turbo/pull/7089/commits/da2b2699f84e3b8ef4a78dc55dbbbd1a9f3ade1c wax ignores broken symlinks because walkdir ignores broken symlinks. this change handles those and yields them as expected, fixing the final tests

### Testing Instructions

Existing CI should suffice

Closes TURBO-2139


crates/turborepo-globwalk/src/lib.rs
```diff
@@ -308,82 +307,60 @@ pub fn globwalk_internal(
...
+        .into_iter()
         .map(glob_with_contextual_error)
-        .collect::<Result<Vec<_>, _>>()?;
+        .collect::<Result<_, _>>()?;
 
-    let span = Span::current();
-    let result = inc_patterns
+    include_paths
         .into_iter()
-        .flat_map(|glob| {
```
chris-olszewski, line 323: Are the checks performed by wax's `glob.walk` now?

crates/turborepo-globwalk/src/lib.rs
```diff
@@ -308,82 +307,60 @@ pub fn globwalk_internal(
...
+            )
         })
-        .collect::<Result<HashSet<_>, WalkError>>()?;
-    Ok(result)
+        .filter_map(|entry| visit_file(walk_type, entry))
+        .collect::<Vec<_>>()
+}
+
+#[tracing::instrument]
+fn visit_file(
```
chris-olszewski, line 348: If I'm understanding this correctly this is just moved out from `globwalk_internal`?

chris-olszewski (changes requested):
I'm having a hard time following the changes in this PR. A few pieces of information that might be helpful:
 - What fix(es) are we backporting from wax to our fork? (I see a lot of commits, not sure which ones matter)
 - Why are the changes to `turborepo-globwalk` needed with the wax updates?

Also just wanted to get this back in your inbox since it looks like CI didn't run.

crates/turborepo-globwalk/src/lib.rs
arlyon: Yeah I decided to pull them out rather than using spans manually because it was a little noisy

crates/turborepo-globwalk/src/lib.rs
arlyon: PR is ready to review, I added some explanations in the body

chris-olszewski (changes requested):
Thank you for the details! LGTM, just want to get this back in your queue for the failing unit test.

arlyon:
Rust lints are not related, and are fixed in https://github.com/vercel/turbo/pull/7208

PR #7096, opened 2024-01-24
migrate type references to primary references

sokra:
### Description

refactoring

* migrate type references to primary references
* change references in ModuleResolveResult to affecting_sources

### Testing Instructions

Closes PACK-2288

### DIFF
--- a/crates/turbopack-core/src/resolve/mod.rs
+++ b/crates/turbopack-core/src/resolve/mod.rs
@@ -75,7 +75,7 @@ pub enum ModuleResolveResultItem {
 #[derive(Clone, Debug)]
 pub struct ModuleResolveResult {
     pub primary: Vec<ModuleResolveResultItem>,
-    pub references: Vec<Vc<Box<dyn ModuleReference>>>,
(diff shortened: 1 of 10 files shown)


crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -141,44 +130,52 @@ impl ModuleResolveResult {
...
                 .map(ModuleResolveResultItem::Module)
                 .collect(),
-            references,
+            affecting_sources,
         }
     }
 
-    pub fn add_reference_ref(&mut self, reference: Vc<Box<dyn ModuleReference>>) {
-        self.references.push(reference);
+    pub fn primary_modules_ref(&self) -> impl Iterator<Item = Vc<Box<dyn Module>>> + '_ {
```
ForsakenHarmony, line 150:
```suggestion
    pub fn primary_modules_iter(&self) -> impl Iterator<Item = Vc<Box<dyn Module>>> + '_ {
```
not a fan of the name here and below, but also don't care that much

ForsakenHarmony (approved):

PR #7098, opened 2024-01-25
Examples tests revamp.

anthonyshew:
The tests for examples have proven to be a point of friction for us. This PR aims to fix that.

### Try it out locally with `turbo test -F "@turborepo-examples-tests*" --continue --concurrency=5`

---

There is a bit of technical debt associated with these tests, I'm told. Historically, these tests were used more like integration or end-to-end tests to ensure that `turbo` was behaving predictably on example monorepos.

Today, we're making the explicit commitment that **these tests no longer need to serve that goal**. Instead, they are for **ensuring that the examples work correctly against the latest published version of `turbo`**.  We're making sure that users can trust that `npx create-turbo@latest [-e <some-example>]` will work out of the box.

## Problems to solve
- [x] **Complexity**
   - It appears that our scripting once handled swapping out package managers for every example. We no longer feel the need to test our examples in this way. We trust that the rest of our CI has managed these complexities way before they make it to end users downloading an example.
   - There is also layers of scripting around platforms, git, and manipulating the package manager. I'm not totally sure what these were for - but they're not useful for our new goals.
- [x] **Flakiness**
   - I suspect that we're not actually surfacing broken examples in these tests. I've received reports from users that our examples didn't work out of the box - _while these tests are green_. 🧐  As we peel back the technical debt, I believe we will find out if this is indeed the case.
   - I've personally experienced these tests _fail_ at unexpected times, _pass_ at unexpected times, and _pass and fail_ by re-running the same job multiple times with the same code.
   - Going back to our freshly stated goals, the tests catch problems way too late (when they do truly catch problems). The examples can potentially go untested against the latest published version of `turbo` indefinitely. 
- [x] **Testing the wrong thing**
   - The tests are currently set up to do snapshot testing using Prysk. However, we no longer are interested in trying to find drift in logs in these tests so we don't need this log capturing layer.
   - This Prysk layer also adds a heavy, third layer of task runner management between terminal command and final package.json script execution. The other two are `turbo`.
- [x] **Debuggability**
   - Understanding and debugging the current test setup is **extremely** difficult. I have my suspicions that this also goes back to the Prysk layer as it makes logs quite hard to read. I've found it also swallows logs during the debugging process that would have been useful to see earlier.
   - I expect that this issue will fix itself as we address the complexity and flakiness concerns above.
- [x] **Performance**
   - I have a hunch that these tests can just...be faster. This will likely be a tailwind of addressing the aforementioned complexity.
  
## Non-goals
- There are a few examples that we don't test due to their own internal complexities (namely, Docker). We could choose to test these examples in the future but we're preferring to fix what we already have first.
- We **do not** want to test the examples against `turbo`@`HEAD`.

## Key changes
- **No more Prysk!**: My first experiment was to get Prysk out of the way to see what was going on more clearly and, well, I kept not using it and kept getting closer to my goals.
- **No more snapshot tests**: Because we're not concerned about drift in the examples, we can stop doing snapshot tests. This piggybacks on the "No more Prysk" idea.
- **Way less scripting**: I love Bash just as much as the next developer. But, alas, the new codepaths use way less of it. Fewer steps, less work, trimming the fat of things we used to do that we no longer need to.

## Results
- [x] These updates **caught issues in the existing examples**! I'm not certain how these issues were being covered up in the previous implementation but I've confirmed three important fixes that were being hidden. They were caught in CI and I fixed them. Early win!
- [x] The tests now have significantly less complexity, only doing the work that they really need to be doing.
- [x] From my experimenting, the tests are now deterministic. I can quickly reproduce issues on my local machine that I found in CI. I've also had success purposefully creating errors and resolving them.
- [x] They're faster. Locally testing is approximately 90% faster for me while CI is about 50% faster.  This likely a symptom of being simpler.
- [x] They're more predictable, more debuggable, and I've included a README to ensure continued understanding of these tests in the future.

## TODO?
- [ ] We could choose to run these tests on a cron to ensure we're keeping up with published sooner. This PR still leaves us open to not testing the examples against `latest`. (This is slightly mitigated by Dependabot always running on Mondays. We would find out that `latest` broke an example then.)
- [ ] We could also choose to run these tests as a part of the release process.

Closes TURBO-2177


NicholasLYang:
We should add the examples to the smoke tests we run before releasing. Right now we only run examples tests if examples get changed (or if the CI config gets changed). Since these tests are explicitly for testing if the latest version of turbo works with examples, we should block new releases on working examples

package.json
```diff
@@ -60,5 +60,8 @@
   "packageManager": "pnpm@8.14.0",
   "engines": {
     "node": "20.x"
+  },
+  "dependencies": {
+    "turbo": "^1.11.3"
```
chris-olszewski, line 65: I'm *slightly* opposed to pinning the version here as opposed to pulling the latest version/canary at time of run. If we do go ahead with pinning a `turbo` version, then it should probably be canary so we're on the bleeding edge.

anthonyshew:
I haven't gotten to get familiar with the release CI yet so wasn't going to introduce that here. If that's something, we'd like to do, happy to do it in a follow-up PR.

turborepo-tests/examples/test_example.sh
```diff
@@ -0,0 +1,53 @@
+# Start by figuring out which example we're testing and its package manager
+example_path=$1
+package_manager=$2
+
+# If you need to do some debugging, you can set these to true
+debug=false
+force=false
```
chris-olszewski, line 7: Not blocking at all, but might be nice to be able to control these via an env var.

turborepo-tests/examples/test_example.sh
```diff
@@ -0,0 +1,53 @@
+# Start by figuring out which example we're testing and its package manager
```
chris-olszewski, line 1:
```suggestion
#!/usr/bin/env bash
# Start by figuring out which example we're testing and its package manager
```

turborepo-tests/examples/basic/turbo.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": ["//"],
+  "pipeline": {
+    "test": {
+      "outputs": []
```
chris-olszewski, line 5: We should add `../test_example.sh` as an input to all of these pipeline definitions

chris-olszewski (approved):
Approving as I think this is a great step in the right direction. All of my comments/suggestions are either easy fixes or very optional.

turborepo-tests/examples/test_example.sh
anthonyshew: I spent more time than I care to admit trying to make an ergonomic way to handle that and it wasn't worth the complexity, in my opinion.

package.json
anthonyshew: Oh, oops, this was an accident. Thanks for noticing.

pnpm-workspace.yaml
```diff
@@ -3,7 +3,9 @@ packages:
   - "docs"
   - "cli"
   - "packages/*"
-  - "turborepo-tests/*"
+  - "turborepo-tests/integration"
+  - "turborepo-tests/helpers"
+  - "turborepo-tests/examples/*"
```
mehulkar, line 8: Why did the layout need to change? Can we keep each of the example tests in the top level?

turborepo-tests/examples/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Test structure
+
+We guarantee that the examples work by running the `lint` and `build` tasks in each example and checking for a `>>> FULL TURBO` on the second run. This requires that the tasks pass and are cacheable. The logic for the tests is in `test_example.sh` and takes a few steps:
+
+1. `cd` to the example directory.
+2. Install packages.
+3. Run `turbo build lint`.
+4. Run `turbo build lint` again and write the log results to a temporary text file.
+5. Read that text file looking for a `>>> FULL TURBO`.
```
mehulkar, line 13: This is basically the same as it was before, minus prysk right?

turborepo-tests/examples/test_example.sh
mehulkar, line 1: Can we either
- Update the existing [`setup_example_test.sh`](https://github.com/vercel/turbo/blob/main/turborepo-tests/helpers/setup_example_test.sh) script
- Use the [other helper scripts](https://github.com/vercel/turbo/tree/main/turborepo-tests/helpers) to compose this
- Place this script in `turborepo-tests/helpers` package

I would like the concept of "run turbo against this repo and assert some things" to remain tightly controlled and generalizable with the eventual goal of enabling something like this:

```bash
npx @turbo/test-turbo-canary <remote-or-local-path-to-repo> --version="@canary" --task=build
```

This is something customers want to be able to do easily before upgrading, and I spent a considerable amount of time moving our test scripts towards that goal! This will also enable _us_ to validate turbo releases work against OSS repos or in the Vercel build container.

turborepo-tests/examples/turbo.json
mehulkar, line 1: This is not a configured workspace, is this file necessary?

turborepo-tests/examples/test_example.sh
```diff
@@ -0,0 +1,54 @@
...
+  package_manager_command="yarn"
+  update_turbo_version="yarn upgrade turbo@latest -W"
+fi
+
+# All examples implement these two tasks
+# and it's reasonable to assume that they will continue to do so
+turbo_command="turbo build lint --output-logs=errors-only"
+
+# Head into the example directory
+cd $example_path
```
mehulkar, line 27: Using a monorepo inside another monorepo was problematic before and that's why we added prysk, so that execution would occur in a tmp directory. https://github.com/vercel/turbo/pull/4000

mehulkar (commented):
I don't have a strong opinion, but I'll say that I'm ok with removing prysk and removing the assertions agains the whole log output, and only looking for `FULL TURBO`. I'm not confident that the remaining changes are net positive.

turborepo-tests/examples/README.md
anthonyshew: Correct! The major spiritual change is that we're not doing true snapshot tests. Rather, we're just checking for some "FULL TURBO".

pnpm-workspace.yaml
mehulkar: tomato tomato. I found it "cleaner" to not have an extra layer of nesting of workspaces. Go head and do what you prefer 👍🏾

mehulkar: (also unnecessary for this PR and increases the diff to understand, fwiw)

turborepo-tests/examples/test_example.sh
anthonyshew: Seems to be working well now! Not sure what would have changed from then to now but I'm not running into any problems.

turborepo-tests/examples/test_example.sh
anthonyshew: Moved it!

NicholasLYang (approved):

turborepo-tests/example-with-svelte/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "turborepo-examples-test-with-svelte",
+  "scripts": {
+    "test": "bash ../helpers/test_example.sh ../../examples/with-svelte pnpm"
```
mehulkar, line 4: Since the helpers package is installed, you can actually invoke this locally:

```suggestion
    "test": "bash ./node_modules/turborepo-tests-helpers/test_example.sh ../../examples/with-svelte pnpm"
```

(or use `npx`, `pnpx` or whatever if that lets you run a bash script)

turborepo-tests/helpers/test_example.sh
mehulkar, line 1: Keep the `setup_example_test.sh` name to match integration tests? not blocking

turborepo-tests/helpers/test_example.sh
```diff
@@ -0,0 +1,54 @@
...
+  package_manager_command="yarn"
+  update_turbo_version="yarn upgrade turbo@latest -W"
+fi
+
+# All examples implement these two tasks
+# and it's reasonable to assume that they will continue to do so
+turbo_command="turbo build lint --output-logs=errors-only"
+
+# Head into the example directory
+cd $example_path
```
mehulkar, line 27: This will inhert `vercell/turbo/.git` index. That seems like something that won't end well. Do we need to isolate the git repo here?

turborepo-tests/helpers/test_example.sh
```diff
@@ -0,0 +1,54 @@
...
+# Do the work!
+if [ $debug == true ]; then
+  $package_manager_command
+  $update_turbo_version
+  [[ $force == true ]] && $turbo_command --force || $turbo_command
+else
+  $package_manager_command > /dev/null 2>&1
+  $update_turbo_version > /dev/null 2>&1
+  [[ $force == true ]] && $turbo_command --force || $turbo_command > /dev/null
+fi
```
mehulkar, line 42: What if there's an existing cache? Do you want to assert that there's is _not_ a cache hit here, so we can be sure tests are always running from a clean slate? Since this is now running "locally" and not in an isolated environment, I'd be concerned that gitignored things can cause tests passing or failing at random.

anthonyshew: Isn't hitting cache valid? Same principle as caching for anything `turbo`, right? "If you've done it before, hit the cache." I'm trying to think of why I would specifically _not want it to hit cache_. Purely on principle of hermeticism for tests, I suppose?

I'm thinking back to the goals, though. "Make sure an example works out of the box." If `turbo build lint` has been ran with those cache inputs and came back clean prior to the test run, then it stands to reason that it's valid enough to pass for the test? 

I'm thinking I don't care if it came from a fresh run or from cache? At least it didn't seem to matter when I was trying to break that assumption by hand when I first wrote up this strategy, anyway.

turborepo-tests/helpers/test_example.sh
anthonyshew: Yeahhh, this was something that I thought about, but I'm struggling to figure out a way it will hurt these tests in practice...

turborepo-tests/helpers/test_example.sh
mehulkar: Hitting cache on the first run just means that you could be replaying a run even though files have changed (e.g. if actual caching is broken, and we replay a previous cache), we'd never know, but a user would when they use this example.

You could do `rm -rf .turbo/ node_modules/ || true` before `$package_manager_command` to create a clean environment.

(Note we aren't talking about cache hit on `turbo run test --filter="turborepo-examples-test-*"` , we're talking about cache hit on `turbo build lint --output-logs=errors-only` in each example.)

turborepo-tests/helpers/test_example.sh
anthonyshew: Ah, okay, I see the setup_git.sh. No reason not to!

mehulkar: You'll also have to clean it up after the test is done and make sure that it's cleaned up if the test fails. Leaving an orphan `.git` inside our existing git repo will turn it into a submodule of sorts, and mean that changes won't be pushed up correctly.

turborepo-tests/helpers/test_example.sh
anthonyshew: Ah, okay, I see. I'll add the assertion, too. 👍

Cargo.lock
mehulkar, line 1: This was fixed in another PR I think, might want to see if it's possible to back out

anthonyshew: Sneaky little bugger.

turborepo-tests/README.md
```diff
@@ -0,0 +1,37 @@
+# Turborepo Examples tests
+
+This README describes the strategy for the tests in the `examples-*` for folders.
```
mehulkar, line 3:
```suggestion
## integration tests

TODO

## example-* tests

This section describes the strategy for the tests in the `examples-*` for folders.
```

turborepo-tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+These tests are for ensuring that the examples we are providing to users work right out of the box.
+
+## Test structure
+
+We guarantee that the examples work by running the `lint` and `build` tasks in each example and checking for a `>>> FULL TURBO` on the second run. This requires that the tasks pass and are cacheable. The logic for the tests is in `test_example.sh` and takes a few steps:
```
mehulkar, line 9:
```suggestion
We guarantee that the examples work by running the `lint` and `build` tasks in each example and checking for a `>>> FULL TURBO` on the second run. This requires that the tasks pass and are cacheable. The logic for the tests is in `setup_example_test.sh` and takes a few steps:
```

turborepo-tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. `cd` to the example directory.
+2. Install packages.
+3. Run `turbo build lint`.
+4. Run `turbo build lint` again and write the log results to a temporary text file.
+5. Read that text file looking for a `>>> FULL TURBO`.
+
+## Creating a test
+
+We want to create a test for every example that we can. To create a test:
```
mehulkar, line 19: There's no reason to say we want to do something that we're already doing

```suggestion
To create a test for a new example:
```

turborepo-tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+4. Edit the `test` script to path to the directory you are interested in and use the package manager for that example.
+5. Run `turbo test --filter="turborepo-examples-test*"` in your terminal to make sure all is well!
+
+## Limitations
+
+We currently do not test the examples that use Docker. We may choose to do this in the future.
+
+## Debugging and more notes
+
+- Each test will create four .txt files in a `/tmp` directory in the file for the example. You can check these files for a readout of any unexpected behaviors.
```
mehulkar, line 33: This is too prescriptive for my taste. If someone changes the implementation it is very very unlikely that they update this readme as well. I'd kill this whole debugging section.

turborepo-tests/example-basic/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "turborepo-examples-test-basic",
+  "scripts": {
+    "test": "bash ../helpers/setup_example_test.sh ../../examples/basic pnpm"
```
mehulkar, line 4: using `./node_modules` got complicated huh?

turborepo-tests/example-with-yarn/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "turborepo-examples-test-with-yarn",
```
mehulkar, line 2: These used to be prefixed with `@turborepo-examples-tests/` , fwiw. I'd generally prefer to keep behavior and style changes in separate commits in main (i.e. separate PRs, since we squash and merge). If you're changing it anyway, how about `@turbo-test/example-with-yarn`, etc.?

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,61 @@
...
-# that is installed in the example's node_modules.
-if [ "${OSTYPE}" == "msys" ]; then
-  EXT=".exe"
-else
-  EXT=""
+export TURBO_TELEMETRY_MESSAGE_DISABLED=1
+
+# Start by figuring out which example we're testing and its package manager
+example_path=$1
+package_manager=$2
```
mehulkar, line 7: If there aren't any defaults, you _probably_ want to `exit 1` if these aren't provided.

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,61 @@
...
+# Isolate the example from the rest of the repo from Git's perspective
+"../../turborepo-tests/helpers/setup_git.sh . -n"  > /dev/null 2>&1
+
+# Let's also isolate from turbo's perspective
+rm -rf .turbo/ node_modules/ || true
+
+# Make sure there is a coverage file to write into
+# This is a minor abuse of the coverage directory
+# but we need something that is gitignored so this works
+mkdir -p ./coverage
```
mehulkar, line 37: This feels like a step backwards. Have you considered using `$TMPDIR` (like prysk does, but without going through prysk). Then you can write `.txt` files without worrying about any cleanup.

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,61 @@
...
-set +eo pipefail
+# All examples implement these two tasks
+# and it's reasonable to assume that they will continue to do so
+turbo_command="turbo build lint --output-logs=errors-only"
+
+# Head into the example directory
+cd $example_path
+
+# Isolate the example from the rest of the repo from Git's perspective
+"../../turborepo-tests/helpers/setup_git.sh . -n"  > /dev/null 2>&1
```
mehulkar, line 29: Could you explain why you don't want to set `script-shell=bash` for these tests? From the outside my only guess is that it's not about `script-shell` at all, it's about not creating a new `.npmrc` file. Is that true? If so, have you considered using `$TMPDIR`  so you don't have to worry about leaving test artifacts?

mehulkar (commented):
Left another round of review. I think you should copy the example to `TMPDIR` and run the test from there to avoid a large set of problems around test isolation.

turborepo-tests/example-with-yarn/package.json
anthonyshew: Mistake on my part. I'll keep that prefix such that the workspace name is only dropping the package manager.

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,72 @@
...
+  package_manager_command="yarn"
+fi
+
+# All examples implement these two tasks
+# and it's reasonable to assume that they will continue to do so
+turbo_command="turbo build lint"
+
+# Head into a temporary directory
+mkdir -p ../../examples-tests-tmp
+cd ../../examples-tests-tmp
```
anthonyshew, line 34: We're going to work in a temporary directory for better test isolation.

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,77 @@
...
+
+# Head into the example directory
```
anthonyshew, line 36:
```suggestion
# Head into the examples tests directory
```

turborepo-tests/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Creating a test
+
+To create a test for a new example:
+
+1. Copy the `basic` directory into a new folder.
+2. Edit the `name` field in `package.json`.
+3. Run `pnpm install` for the repository so the new package is added to the workspace.
+4. Edit the `test` script to path to the directory you are interested in and use the package manager for that example.
+5. Run `turbo test --filter="turborepo-examples-test*"` in your terminal to make sure all is well!
```
mehulkar, line 29:
```suggestion
5. Run `turbo test --filter="@turborepo-examples-tests/*"` in your terminal to make sure all is well!
```

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,77 @@
 #!/usr/bin/env bash
 
-set -eo pipefail
```
mehulkar, line 3: Probably should bring this back. If something fails in this script, actual test execution will probably also fail, so it would be good to just exit early.

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,77 @@
...
+# All examples implement these two tasks
+# and it's reasonable to assume that they will continue to do so
+turbo_command="turbo build lint"
+
+# Head into a temporary directory
+mkdir -p ../../examples-tests-tmp
+cd ../../examples-tests-tmp
+
+# Head into the examples tests directory
+rm -rf "../examples-tests-tmp/$example_path"
```
mehulkar, line 37: - Since the previous line `cd`d into `examples-tests-tmp`, you could simplify this
- This might not exist on a clean install, so if you're just trying to cleanup, you probably want some safety here (the missing `set -eo pipefail` at the top of the test will help catch this

```suggestion
rm -rf "$example_path" || true
```

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,77 @@
...
+# Remove node_modules so the copy is faster when working locally
+find "../examples/$example_path" -name "node_modules" -type d -prune | xargs rm -rf
+cp -r "../examples/$example_path" $example_path
```
mehulkar, line 40: deleting node_modules from source is a bit funky, as it alters the state for the next test run (and for anything else you want to do in the repo). e.g. `pnpm install` will need to reinstall these pieces next time. I just used AI to get this command, so YMMV and you can use something else, but you can do the copy and exclude in a single command with:

```suggestion
rsync -av --exclude='node_modules' "../examples/$example_path" "$example_path"
```

Looks like there's a way to do it with `find` with an `-exec` option also, if `rsync` gives you any trouble

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,77 @@
...
+
+# Isolate from git's perspective
+rm -rf $example_path/.git
```
mehulkar, line 43: There shouldn't be a `.git` directory here, afaik, so this `rm -rf` needs some safety

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,77 @@
...
+
+cd $example_path
+
+# Let's also isolate from turbo's perspective
+# in case you've ran the tests before
+rm -rf .turbo/ node_modules/ || true
+
+# Make /tmp dir for writing dump logs
+mkdir -p ./tmp
+echo -e "$(cat .gitignore)" "\n/tmp/" >.gitignore
```
mehulkar, line 54: you can append to `.gitignore` with `>>` instead of 

```suggestion
echo "/tmp/" >> .gitignore
```

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -1,37 +1,77 @@
...
+elif [ "$package_manager" == "yarn" ]; then
+  package_manager_command="yarn"
+fi
+
+# All examples implement these two tasks
+# and it's reasonable to assume that they will continue to do so
+turbo_command="turbo build lint"
+
+# Head into a temporary directory
+mkdir -p ../../examples-tests-tmp
```
mehulkar, line 33: In case you didn't already know, there's a handy way of making a tmp directory (that you never need to touch again, the OS will just clean it up on restart):

```
xx=$(mktemp -d)
echo $xx # gives you the full path to the new dir)
```

https://linux.die.net/man/1/mktemp

mehulkar (commented):
Mostly there, there's some script cleanup to do. I'd also install `shellcheck` as a VS Code extension if you don't have it already (it's a linter essentially), I've found that to be a pretty useful learning tool!

turborepo-tests/helpers/setup_example_test.sh
anthonyshew: Bless my heart, truly.

turborepo-tests/helpers/setup_example_test.sh
anthonyshew: I've handled everything but this. I can't seem to get the randomized directories to play nice with everything else. 🤔

This would be a small, small win for local debugging. I'm okay with punting it.

mehulkar (approved):
Ran it locally and the tests passed, nice job!

I don't love that it leaves a `examples-tests-tmp/` laying around my repo which is largely a duplicate of `examples/` , but I guess that's ok and/or we can take it another pass. (I'm concerned about things like the extra noise when using `find` locally, etc etc).

Approving since we've gone a few rounds on this and I'd rather discover any problems and iterate than try to hunt them all down now.

PR #7104, opened 2024-01-25
Turbopack: use inline sourcemaps for HMR chunks

wbinnssmith:
Previously, we sent a URL with a query string identifying the module, but this never worked in nextturbo. Instead, this sends the source map string itself, where it is then encoded as base64 and used inline in the `eval`ed code.

Test Plan: Updated a client component and verified browser devtools shows updated source code through the source map.

Closes PACK-2296


crates/turbopack-dev/src/ecmascript/merged/update.rs
```diff
@@ -80,21 +81,31 @@ struct EcmascriptModuleEntry {
...
+    async fn new(id: &ModuleId, code: ReadRef<Code>, chunk_path: &str) -> Result<Self> {
         /// serde_qs can't serialize a lone enum when it's [serde::untagged].
         #[derive(Serialize)]
         struct Id<'a> {
             id: &'a ModuleId,
         }
         let id = serde_qs::to_string(&Id { id }).unwrap();
-        EcmascriptModuleEntry {
+
+        let map = (*code).clone().cell().generate_source_map().await?;
```
sokra, line 92: That's pretty weird. Run generate_source_map on the original `Vc<Code>` instead and pass that to this method

crates/turbopack-dev/src/ecmascript/merged/update.rs
```diff
@@ -80,21 +81,31 @@ struct EcmascriptModuleEntry {
 }
 
 impl EcmascriptModuleEntry {
-    fn new(id: &ModuleId, code: ReadRef<Code>, chunk_path: &str) -> Self {
+    async fn new(id: &ModuleId, code: ReadRef<Code>, chunk_path: &str) -> Result<Self> {
```
sokra, line 84: An async constructor is a bit unusual... Could you avoid that

sokra (approved):

kdy1 (approved):

PR #7105, opened 2024-01-25
fix(hashing): include gitignored files in fallback

tknickman:
### Description

If inputs have been provided, we allow git-ignored files to be included in the hash. This behavior wasn't respected for the fallback hashing (when git isn't available). This fixes that. 

Closes TURBO-2156

### DIFF
--- a/crates/turborepo-scm/src/manual.rs
+++ b/crates/turborepo-scm/src/manual.rs
@@ -84,7 +84,9 @@ pub(crate) fn get_package_file_hashes_from_processing_gitignore<S: AsRef<str>>(
     };
     let walker = walker_builder
         .follow_links(false)
-        .git_ignore(true)
+        // if inputs have been provided manually, we shouldn't skip ignored files to mimic the
+        // regular behavior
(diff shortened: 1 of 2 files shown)


crates/turborepo-scm/src/manual.rs
```diff
@@ -323,6 +325,14 @@ mod tests {
                 .unwrap();
         assert_eq!(hashes, expected);
 
+        // set a hash for an ignored file
+        for tuple in file_hash.iter_mut() {
```
gsoltis, line 329: Non-blocking nit: You can destructure this tuple, like [here](https://github.com/vercel/turbo/pull/7105/files#diff-f8a7d7bb95a7cb29aa009a7bf275bcee7a3442065f307a55fc7e97c669d96d14R337). Then you don't need `.0` and `.2`

gsoltis (approved):

crates/turborepo-scm/src/manual.rs
tknickman: Ah nice!

PR #7106, opened 2024-01-25
fix(hashing): include default files in fallback

tknickman:
### Description

Stacked on top of https://github.com/vercel/turbo/pull/7105 (review first)

Fallback git hashing fixes - include `package.json` and `turbo.json` by default if a manual input array has been provided and we're using fallback (non git) hashing. 

This is done to mimic the standard behavior (non fallback): https://github.com/vercel/turbo/blob/main/crates/turborepo-scm/src/package_deps.rs#L160-L171

Closes TURBO-2157

### DIFF
--- a/crates/turborepo-scm/src/manual.rs
+++ b/crates/turborepo-scm/src/manual.rs
@@ -28,6 +28,13 @@ fn git_like_hash_file(path: &AbsoluteSystemPath) -> Result<String, Error> {
     Ok(result.encode_hex::<String>())
 }
 
+fn to_glob(input: &str) -> Result<Glob, Error> {
+    let glob = fix_glob_pattern(input).into_unix();
+    let g = Glob::new(glob.as_str()).map(|g| g.into_owned())?;
+
+    Ok(g)
+}
+
 pub(crate) fn hash_files(
     root_path: &AbsoluteSystemPath,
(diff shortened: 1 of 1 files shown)


crates/turborepo-scm/src/manual.rs
```diff
@@ -259,10 +281,17 @@ mod tests {
...
         let unix_pkg_path = pkg_path.to_unix();
-        let file_hash: Vec<(&str, &str, Option<&str>)> = vec![
+        let mut file_hash: Vec<(&str, &str, Option<&str>)> = vec![
+            ("turbo.json", "turbo.json-file-contents", None),
+            ("package.json", "root-package.json-file-contents", None),
             ("top-level-file", "top-level-file-contents", None),
             ("other-dir/other-dir-file", "other-dir-file-contents", None),
             ("ignoreme", "anything", None),
+            (
+                "child-dir/libA/package.json",
```
gsoltis, line 291: We should probably add a package-level `turbo.json` too.

gsoltis (changes requested):

crates/turborepo-scm/src/manual.rs
tknickman: Good call, done!

gsoltis (approved):

mehulkar (approved):

PR #7108, opened 2024-01-25
fix: strip out EOT sequence from PTY output

chris-olszewski:
### Description

Currently the PTY device we're creating has `ECHOCTL` set which means that the EOT that gets sent when we drop stdin shows up as a `^D`:

> ECHOCTL
              greater than the special character.

This PR just looks for the sequence that gets generated in the output and strips it.

### Testing Instructions

Added unit tests and made existing unit tests that use the updated codepath more strict.

Closes TURBO-2159

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6073,7 +6073,7 @@ dependencies = [
  "shared_library",
  "shell-words",
  "winapi 0.3.9",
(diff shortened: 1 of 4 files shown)


chris-olszewski:
Closing in favor of #7109

PR #7109, opened 2024-01-25
fix disable echoctl

chris-olszewski:
### Description

Disable ECHOCTL for the pseudo terminal that we start:

> ECHOCTL
              greater than the special character.  For example,
              character 0x08 (BS) is echoed as ^H.

This will remove the `^D` that gets generated from the EOT when using a PTY. Using PTY still does generate a leading `\n` as EOT is only respected if it immediately follows a newline. `portable_pty` tries to prevent failed EOT by emitting a `\nEOT` on drop in case end users didn't end the last write with a newline.

### Testing Instructions

Made existing child process tests check for exact output modulo whitespace/non-printing chars.

Closes TURBO-2160


crates/turborepo-lib/src/process/child.rs
```diff
@@ -878,8 +902,10 @@ mod test {
         };
 
         let output_str = String::from_utf8(output).expect("Failed to parse stdout");
+        let trimmed_out = output_str.trim();
+        let trimmed_out = trimmed_out.strip_prefix('\u{4}').unwrap_or(trimmed_out);
 
-        assert!(output_str.contains(input), "got: {}", output_str);
+        assert!(trimmed_out.contains(input), "got: {}", trimmed_out);
```
chris-olszewski, line 908: Note: this is still a pretty loose assertion. We seem to be [hitting this](https://github.com/wez/wezterm/blob/main/pty/examples/narrow.rs#L53) causing the captured output to be "hello world\n^Dhello world\n" . I will investigate further and fix in a future PR.

crates/turborepo-lib/src/process/child.rs
```diff
@@ -839,8 +859,12 @@ mod test {
             };
 
             let output_str = String::from_utf8(output).expect("Failed to parse stdout");
+            let trimmed_output = output_str.trim();
+            let trimmed_output = trimmed_output
+                .strip_prefix('\u{4}')
```
mehulkar, line 864: i had to look this up. code comment to indicate it's an EOT character?

tknickman (approved):
TIL, nice!

mehulkar (approved):

PR #7113, opened 2024-01-26
feat(hashing): support $TURBO_DEFAULT in inputs

tknickman:
### Description

Adds support for a `$TURBO_DEFAULT` in `inputs`. This will expand to all files that `turbo` would normally include (all non-gitignored files in the workspace), but also allow additional globs (negative as well). 

For example:

```json
"inputs": ["$TURBO_DEFAULT", "ignored-file.json", "!./app/layout.tsx"]
```

This allows all package files to be included, but add an additional git ignored file: `ignored-file.json`, and exclude a file that is part of `$TURBO_DEFAULT` (`!./app/layout.tsx`).

This reduces the downsides of using `inputs`, and makes this key more flexible, and less error-prone.


anthonyshew:
Amazing! 😍

Docs in a separate PR?

tknickman:
@anthonyshew yes! Will need some help there I think, but I'll ping you with the branch.

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -9,6 +9,8 @@ use crate::{hash_object::hash_objects, Error, Git, SCM};
 
 pub type GitHashes = HashMap<RelativeUnixPathBuf, String>;
 
+const INPUT_INCLUDE_DEFAULT_FILES: &str = "$TURBO_DEFAULT";
```
mehulkar, line 12: Can we make this token `$TURBO_DEFAULT$`? I am concerned that it will be confused with an environment variable this way. I could even see `$$TURBO_DEFAULT$$` being more clear.

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -31,19 +33,33 @@ impl SCM {
         inputs: &[S],
         telemetry: Option<PackageTaskEventBuilder>,
     ) -> Result<GitHashes, Error> {
+        // If the inputs contain "$TURBO_DEFAULT", we need to include the "default" file
+        // hashes as well. NOTE: we intentionally don't remove "$TURBO_DEFAULT"
+        // from the inputs if it exists in the off chance that the user has a
+        // file named "$TURBO_DEFAULT" in their package (pls no).
```
mehulkar, line 39: If you don't remove it, but still set `include_default_files`, it'll _both_ include default files and pick up the file with that literal name? Could we add a test case for this crazy scenario?

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -99,11 +116,52 @@ impl Git {
         turbo_root: &AbsoluteSystemPath,
         package_path: &AnchoredSystemPath,
         inputs: &[S],
+        include_default_files: bool,
     ) -> Result<GitHashes, Error> {
+        if include_default_files && !inputs.is_empty() {
```
mehulkar, line 121: Probably a matter of preference, but could we rearrange the conditions in this fn:

```rust
// If inputs isn't specified, get everything
if inputs.is_empty() {
  return self.get_package_file_hashes_from_index(turbo_root, package_path)
}

// we have inputs, and we don't need the default set
if !include_default_files {
  return self.get_package_file_hashes_from_inputs(turbo_root, package_path, inputs, true)
}

// we have inputs and we want all the defaults, we have to merge includes/excludes
// new logic here
```

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -99,11 +116,52 @@ impl Git {
...
+            let mut includes = Vec::new();
+            let mut excludes = Vec::new();
+            for input in inputs {
+                let input_str = input.as_ref();
+                if let Some(exclude) = input_str.strip_prefix('!') {
+                    excludes.push(exclude);
+                } else {
+                    includes.push(input_str);
+                }
+            }
```
mehulkar, line 137: don't we have this logic somewhere already? We don't do this up front in turbo.json parsing?

mehulkar (commented):
Nothing super major in my comments, but left some comments before approval!

crates/turborepo-scm/src/package_deps.rs
gsoltis: 👍 for `$TURBO_DEFAULT$`.

mehulkar (commented):
Another thing here... it looks like the position of `$TURBO_DEFAULTS` in the `inputs` array does not matter. That is one potential for confusion that we should call out in docs. I'm not sure if we will have a use case to ever change this behavior. 

E.g. could these mean different things to anyone?

```json
["$TURBO_DEFAULT$", "somefile.txt", "!notthis.txt"]
```

```json
["somefile.txt", "!notthis.txt", "$TURBO_DEFAULT$"]
```

tknickman:
@mehulkar / @gsoltis review addressed.

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -207,6 +242,45 @@ impl Git {
...
         Ok(hashes)
     }
+
+    #[tracing::instrument(skip(self, turbo_root, inputs))]
+    fn get_package_file_hashes_from_inputs_and_index<S: AsRef<str>>(
+        &self,
+        turbo_root: &AbsoluteSystemPath,
+        package_path: &AnchoredSystemPath,
+        inputs: &[S],
+        include_configs: bool,
```
chris-olszewski, line 252: I don't think this param is needed
```suggestion
```

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -207,6 +242,45 @@ impl Git {
...
+            if let Some(exclude) = input_str.strip_prefix('!') {
+                excludes.push(exclude);
+            } else {
+                includes.push(input_str);
+            }
+        }
+        let manual_includes_hashes =
+            self.get_package_file_hashes_from_inputs(turbo_root, package_path, &includes, true)?;
+        let manual_excludes_hashes =
+            self.get_package_file_hashes_from_inputs(turbo_root, package_path, &excludes, false)?;
```
chris-olszewski, line 273: I think we should add a check if `includes`/`excludes` are empty before performing these calls as they'll end up doing some amount of throw away work.

crates/turborepo-scm/src/package_deps.rs
```diff
@@ -207,6 +242,45 @@ impl Git {
...
+
+        // merge the two includes
+        let mut hashes = default_file_hashes;
+        hashes.extend(manual_includes_hashes);
+
+        // remove the excludes
+        hashes.retain(|key, _| !manual_excludes_hashes.contains_key(key));
```
chris-olszewski, line 280: Future work: I think there's some low hanging perf work to do in future PRs:
 - we can avoid calculating file hashes for all of the paths in `manual_excludes_hashes`
 - can also defer calculating all file hashes until we have the final list of what files we care about

chris-olszewski (approved):
This makes sense to me. I think the unit tests are enough, but it might be nice to have a prysk test that tests this feature with expanded inputs in summarize/dry run.

crates/turborepo-scm/src/package_deps.rs
tknickman: Good call, we can do this with `excludes`, but with the way things are currently structured - if you manually supply `inputs`, we will _always_ add `package.json` and `turbo.json` to them, meaning the includes are _never_ empty.

crates/turborepo-scm/src/package_deps.rs
tknickman: Good catch! Left over from that function refactor

tknickman:
Good call @chris-olszewski, I'm getting docs up for this in the next PR and will add a prysk test

PR #7117, opened 2024-01-26
track RequestKey during resolving and avoid duplicate keys

sokra:
### Description

During resolving we keep a key of the matching request to avoid duplicates.

### Testing Instructions

Closes PACK-2298


crates/turbopack-ecmascript-runtime/js/src/build/runtime.ts
```diff
@@ -183,6 +183,7 @@ function instantiateModule(id: ModuleId, source: SourceInfo): Module {
       n: exportNamespace.bind(null, module),
       m: module,
       c: moduleCache,
+      M: moduleFactories,
```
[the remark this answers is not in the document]
sokra: webpack also exposes that, so that has other use cases too

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -185,6 +199,19 @@ impl ModuleResolveResult {
 
 #[turbo_tasks::value_impl]
 impl ModuleResolveResult {
+    #[turbo_tasks::function]
+    pub async fn as_typings_result(self: Vc<Self>) -> Result<Vc<Self>> {
```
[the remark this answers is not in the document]
sokra: moved it

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -441,7 +539,11 @@ impl ResolveResult {
     }
 
     pub fn merge_alternatives(&mut self, other: &ResolveResult) {
-        self.primary.extend(other.primary.iter().cloned());
+        for (k, v) in other.primary.iter() {
+            if !self.primary.contains_key(k) {
+                self.primary.insert(k.clone(), v.clone());
+            }
+        }
         self.affecting_sources
```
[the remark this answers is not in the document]
sokra: fixed

crates/turbopack-ecmascript/src/resolve/node_native_binding.rs
```diff
@@ -373,14 +389,28 @@ pub async fn resolve_node_bindings_files(
...
-        .iter()
-        .map(|try_dir| {
-            Vc::upcast(RawModule::new(Vc::upcast(FileSource::new(
-                root_context_dir.join(format!("{}/{}", try_dir, &file_name)),
-            ))))
-        })
-        .collect();
 
-    Ok(ModuleResolveResult::modules(bindings_try).cell())
+    Ok(ModuleResolveResult::modules(
```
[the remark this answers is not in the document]
sokra: I removed a bit of nesting

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1764,6 +2019,7 @@ async fn resolve_alias_field_result(
 
 #[tracing::instrument(level = Level::TRACE, skip_all)]
 async fn resolved(
+    request_key: Value<RequestKey>,
```
ForsakenHarmony, line 2022: This doesn't need to be a `Value`, it's not a task function

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -95,61 +97,68 @@ impl ModuleResolveResult {
...
     ) -> ModuleResolveResult {
         ModuleResolveResult {
-            primary: Vec::new(),
+            primary: IndexMap::new(),
             affecting_sources,
         }
     }
 
-    pub fn module(module: Vc<Box<dyn Module>>) -> ModuleResolveResult {
+    pub fn module(request_key: RequestKey, module: Vc<Box<dyn Module>>) -> ModuleResolveResult {
```
ForsakenHarmony, line 105: Given this (and maybe others here or in ResolveResult as well) are mostly called with `RequestKey::default()`, maybe we should default to that here and add a new `module_with_key` function?

```suggestion
    pub fn module(module: Vc<Box<dyn Module>>) -> ModuleResolveResult {
        Self::module_with_key(RequestKey::default(), module)
    }

    pub fn module_with_key(request_key: RequestKey, module: Vc<Box<dyn Module>>) -> ModuleResolveResult {
```

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -650,16 +779,78 @@ impl ResolveResult {
...
                     } else {
                         None
                     }
                 })
                 .collect(),
         ))
     }
+
+    #[turbo_tasks::function]
+    pub async fn with_replaced_request_key(
```
ForsakenHarmony, line 794: comment pls

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -650,16 +779,78 @@ impl ResolveResult {
...
+                                conditions: request_key.conditions.clone(),
+                            },
+                            v.clone(),
+                        ))
+                    } else {
+                        None
+                    }
+                } else {
+                    Some((request_key.clone(), v.clone()))
+                }
```
ForsakenHarmony, line 822: maybe `let-else` + `try` this 👀 

```suggestion
                let Some(k) = &k.request else {
                    return Some((request_key.clone(), v.clone()));
                };
                
                let remaining = k.strip_prefix(&old_request_key)?;
                
                // needs rustfmt
                        Some((
                            RequestKey {
                                request: request_key
                                    .request
                                    .as_ref()
                                    .map(|r| format!("{}{}", r, remaining)),
                                conditions: request_key.conditions.clone(),
                            },
                            v.clone(),
                        ))
```

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -650,16 +779,78 @@ impl ResolveResult {
...
+                                    .map(|r| format!("{}{}", r, remaining)),
+                                conditions: request_key.conditions.clone(),
+                            },
+                            v.clone(),
+                        ))
+                    } else {
+                        None
+                    }
+                } else {
+                    Some((request_key.clone(), v.clone()))
```
ForsakenHarmony, line 821: And it's also expected that `old_request_key` always matches empty keys?  (document if it is)

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -650,16 +779,78 @@ impl ResolveResult {
...
+                                request: request_key
+                                    .request
+                                    .as_ref()
+                                    .map(|r| format!("{}{}", r, remaining)),
+                                conditions: request_key.conditions.clone(),
+                            },
+                            v.clone(),
+                        ))
+                    } else {
+                        None
```
ForsakenHarmony, line 818: It's expected that this function discards any results that don't match the `old_request_key`? (document if it is)

crates/turbopack-core/src/resolve/mod.rs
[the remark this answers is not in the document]
ForsakenHarmony: I think you only fixed `ModuleResolveResult` so far

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -924,9 +1115,10 @@ pub async fn resolve_raw(
     path: Vc<Pattern>,
     force_in_lookup_dir: bool,
 ) -> Result<Vc<ResolveResult>> {
-    async fn to_result(path: Vc<FileSystemPath>) -> Result<Vc<ResolveResult>> {
+    async fn to_result(request: &str, path: Vc<FileSystemPath>) -> Result<Vc<ResolveResult>> {
```
ForsakenHarmony, line 1118:
```suggestion
    async fn to_result(request_key: &str, path: Vc<FileSystemPath>) -> Result<Vc<ResolveResult>> {
```

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -650,16 +779,78 @@ impl ResolveResult {
...
+            .collect();
+        Ok(ResolveResult {
+            primary: new_primary,
+            affecting_sources: this.affecting_sources.clone(),
+        }
+        .into())
+    }
+
+    #[turbo_tasks::function]
+    pub async fn with_request(self: Vc<Self>, request: String) -> Result<Vc<Self>> {
```
ForsakenHarmony, line 833:
```suggestion
    pub async fn with_request_key(self: Vc<Self>, request_key: String) -> Result<Vc<Self>> {
```

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -650,16 +779,78 @@ impl ResolveResult {
...
+        let this = self.await?;
+        let new_primary = this
+            .primary
+            .iter()
+            .map(|(k, v)| {
+                (
+                    RequestKey {
+                        request: Some(request.clone()),
+                        conditions: k.conditions.clone(),
```
ForsakenHarmony, line 842: It's expected that this one keeps the old conditions, but the one above does not? (document if it is)

crates/turbopack-core/src/resolve/mod.rs
ForsakenHarmony, line 833: add docs

crates/turbopack-ecmascript-runtime/js/src/build/runtime.ts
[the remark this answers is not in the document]
ForsakenHarmony: I see, I feel like this should be special cased as it exposes internals to all modules, but I guess malicious modules are not really expected

crates/turbopack-core/src/resolve/mod.rs
ForsakenHarmony: Would also make the diff a bit smaller 🙃

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -493,29 +598,53 @@ impl ResolveResult {
...
                 .try_join()
-                .await?,
+                .await?
+                .into_iter()
+                .collect(),
             affecting_sources: self.affecting_sources.clone(),
         })
     }
+
+    pub fn with_request_ref(&self, request: String) -> Self {
```
ForsakenHarmony, line 633:
```suggestion
    pub fn with_request_key_ref(&self, request_key: String) -> Self {
```

Also, why the `_ref` in the name?

ForsakenHarmony: Being pedantic as just having "request" makes it sound like it's the full `Request` that was passed to `resolve(...)`

crates/turbopack-core/src/resolve/mod.rs
sokra: This is actually `with_request` as it only passes the `request` and not the whole `request_key`

crates/turbopack-core/src/resolve/mod.rs
sokra: > Also, why the _ref in the name?

alex started that pattern when the same method name is defined on Vc<X> and X...

crates/turbopack-core/src/resolve/mod.rs
sokra: changed it

crates/turbopack-core/src/resolve/mod.rs
sokra: yes, documented it

crates/turbopack-core/src/resolve/mod.rs
ForsakenHarmony: But the `request` passed in here isn't actually the full request, it's still just a key?

ForsakenHarmony: Yeah, but `RequestKey::request` is still a key for the request and not a direct representation of the `Request`?

ForsakenHarmony (approved):
I still think some of the arguments being called `request` when they're not a full `Request` but only a key is wrong, but whatever

PR #7119, opened 2024-01-26
fix(turbopack): resolve `.js` before `.ts`

ForsakenHarmony:
### Description

See https://github.com/vercel/next.js/pull/12014

Closes PACK-2301


ForsakenHarmony:
Actually `tsc` seems to prefer typescript as well now

PR #7121, opened 2024-01-26
fix(turbopack): don't parse `.ts` files as `.tsx`

ForsakenHarmony:
### Description

We currently parse JSX syntax in all typescript files which is wrong.

Closes PACK-2302

Next.js PR: https://github.com/vercel/next.js/pull/61219

### DIFF
--- a/crates/turbopack-ecmascript/src/lib.rs
+++ b/crates/turbopack-ecmascript/src/lib.rs
@@ -137,9 +137,12 @@ pub enum EcmascriptModuleAssetType {
     /// Module with EcmaScript code
     Ecmascript,
     /// Module with TypeScript code without types
-    Typescript,
-    /// Module with TypeScript code with references to imported types
-    TypescriptWithTypes,
+    Typescript {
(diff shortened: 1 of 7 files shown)


timneutkens (approved):
💯

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -137,9 +137,9 @@ pub enum EcmascriptModuleAssetType {
     /// Module with EcmaScript code
     Ecmascript,
     /// Module with TypeScript code without types
-    Typescript,
+    Typescript { tsx: bool },
     /// Module with TypeScript code with references to imported types
-    TypescriptWithTypes,
+    TypescriptWithTypes { tsx: bool },
```
sokra, line 142: Should we convert `WithTypes` into a Boolean field too?

ForsakenHarmony: Was wondering if there's a good reason it isn't

PR #7127, opened 2024-01-26
chore: add tracing to pty usage

chris-olszewski:
### Description

Adding some tracing to our spawning code to measure impact of PTY usage. Quick tests make it look like the `openpty` call has a minimal impact (~.2ms) on the time it takes to spawn a process.
Screenshot 2024-01-26 at 10 42 12 AM
Screenshot 2024-01-26 at 10 42 19 AM

### Testing Instructions

Look at the profiles:
[normal.json](https://github.com/vercel/turbo/files/14068301/normal.json)
[pty.json](https://github.com/vercel/turbo/files/14068303/pty.json)

Closes TURBO-2175

### DIFF
--- a/crates/turborepo-lib/src/process/child.rs
+++ b/crates/turborepo-lib/src/process/child.rs
@@ -92,6 +92,7 @@ enum ChildHandleImpl {
 }
 
 impl ChildHandle {
+    #[tracing::instrument(skip(command))]
     pub fn spawn_normal(command: Command) -> io::Result<SpawnResult> {
         let mut command = TokioCommand::from(command);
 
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/process/child.rs
```diff
@@ -92,6 +92,7 @@ enum ChildHandleImpl {
 }
 
 impl ChildHandle {
+    #[tracing::instrument(skip(command))]
```
mehulkar, line 95: Does this directive add to tracing or skip it from tracing?

mehulkar (approved):

crates/turborepo-lib/src/process/child.rs
chris-olszewski: It enables tracing for the function call, but removes the `command` arg from showing up as a field in the tracing span.

PR #7128, opened 2024-01-26
feat: enable pty usage

chris-olszewski:
### Description

Enable PTY usage! Will wait until #7109 is merged so there isn't any `^D` in the output caused by stdin being dropped. Measurements of impact of PTY usage can be found in #7127

Current rules for our PTY usage:
 - User isn't on windows
 - stdout is hooked up to a TTY as this is where we send task logs

We could strengthen the condition by requiring that stderr is hooked up to a TTY, but IMO this is overkill as we don't send any task output to that stream.

### Testing Instructions

When running normally we get nice colors:
Screenshot 2024-01-26 at 11 37 25 AM

Verify that if we pipe our primary output to stdout, we don't use a PTY and get no colors:
Screenshot 2024-01-26 at 11 38 09 AM

Closes TURBO-2176

### DIFF
--- a/crates/turborepo-lib/src/process/mod.rs
+++ b/crates/turborepo-lib/src/process/mod.rs
@@ -65,8 +65,14 @@ impl ProcessManager {
         if lock.is_closing {
             return None;
         }
-        let child =
-            child::Child::spawn(command, child::ShutdownStyle::Graceful(stop_timeout), false);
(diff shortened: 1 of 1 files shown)


tknickman (approved):

datner:
Thank you, finally I can see my builds barf garbage and noise in full Technicolor, like god intended
How I feel regarding this going through finally

PR #7132, opened 2024-01-28
chore: install jest in examples/kitchen-sink

llllvvuu:
Otherwise `pnpm test` doesn't work.

### DIFF
--- a/examples/kitchen-sink/apps/api/package.json
+++ b/examples/kitchen-sink/apps/api/package.json
@@ -15,13 +15,16 @@
     "preset": "@repo/jest-presets/jest/node"
   },
   "dependencies": {
+    "@repo/logger": "workspace:*",
     "body-parser": "^1.20.2",
     "cors": "^2.8.5",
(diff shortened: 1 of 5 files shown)


tknickman (approved):
Generally, we should only put this in each package where it is used (`package/ui`, `apps/api`) to properly segment the cache when this needs to be updated. However, I'm also fine with root for this specific case (devDependencies that generally need to be the same version across the entire repo).

anthonyshew:
@llllvvuu Can you install `jest` in the packages where it is used? We're going to more generally start recommending this pattern to folks and this can be an early move in that direction. Thanks!

llllvvuu:
Sure, done!

anthonyshew (approved):
Awesome, thank you!

anthonyshew:
Merging through the failed formatter CI check because it comes from a GitHub permission being weird. I've confirmed that these changes have nothing to do with the formatter.

PR #7158, opened 2024-01-30
feat: enable vt processing on windows

chris-olszewski:
### Description

If we're on windows and are attempting to emit colors, help our users by setting [`ENABLE_VIRTUAL_TERMINAL_PROCESSING`](https://learn.microsoft.com/en-us/windows/console/setconsolemode) which allows ANSI color escape sequences to be respected by the Windows console. This might already be set to true by some users, but it is not enabled by default by Windows.

### Testing Instructions

Look at the pretty colors on Windows!
cmdp
pwsh
(My default powershell colors are horrible and end up making the `docs:build` prefix invisible)
gitbash

### DIFF
--- a/crates/turborepo-lib/src/shim.rs
+++ b/crates/turborepo-lib/src/shim.rs
@@ -290,10 +290,29 @@ impl ShimArgs {
         if self.no_color {
             UI::new(true)
         } else if self.color {
+            // Do our best to enable ansi colors, but even if the terminal doesn't support
+            // still emit ansi escape sequences.
+            Self::supports_ansi();
             UI::new(false)
-        } else {
+        } else if Self::supports_ansi() {
+            // If the terminal supports ansi colors, then we can infer if we should emit
+            // colors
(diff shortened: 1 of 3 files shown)


crates/turborepo-lib/src/shim.rs
```diff
@@ -290,11 +290,30 @@ impl ShimArgs {
...
             UI::infer()
+        } else {
+            UI::new(true)
         }
     }
+
+    #[cfg(windows)]
+    fn supports_ansi() -> bool {
+        // This call has the side effect of setting ENABLE_VIRTUAL_TERMINAL_PROCESSING
+        // to true.
```
mehulkar, line 309:
```suggestion
        // This call has the side effect of setting ENABLE_VIRTUAL_TERMINAL_PROCESSING
        // to true. https://learn.microsoft.com/en-us/windows/console/setconsolemode
```

mehulkar (approved):

PR #7159, opened 2024-01-30
chore(examples): fix `pnpm typecheck` in `kitchen-sink` example

llllvvuu:
> error TS5095: Option 'bundler' can only be used when 'module' is set to 'es2015' or later.

This project doesn't use the `tsc` emit so it should be fine to not set `'module'`


tknickman (approved):
Thanks!

tknickman:
Closing in favor of https://github.com/vercel/turbo/pull/7624

PR #7167, opened 2024-01-30
fix: catch import map lookup error properly in `ResolvingIssue`

ForsakenHarmony:
### Description

The wrong error / not all errors were caught before.

This also adds the error trace to the error message.

Closes PACK-2325

### DIFF
--- a/crates/turbopack-core/src/resolve/mod.rs
+++ b/crates/turbopack-core/src/resolve/mod.rs
@@ -56,7 +56,7 @@ pub use alias_map::{
 };
 pub use remap::{ResolveAliasMap, SubpathValue};
 
-use crate::issue::IssueSeverity;
+use crate::{error::PrettyPrintError, issue::IssueSeverity};
 
 #[turbo_tasks::value(shared)]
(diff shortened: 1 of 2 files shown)


crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -2042,7 +2042,7 @@ pub async fn handle_resolve_error(
                 request_type: format!("{} request", reference_type.into_value()),
                 request,
                 resolve_options,
-                error_message: Some(err.to_string()),
+                error_message: Some(format!("{:?}", err)),
```
sokra, line 2045:
```suggestion
                error_message: Some(format!("{:?}", PrettyPrintError(&err))),
```

sokra (approved):

crates/turbopack-core/src/resolve/mod.rs
ForsakenHarmony: true, I guess I can refactor the output of `ResolvingIssues` to use that

PR #7168, opened 2024-01-30
fix(tests): mock telemetry config

tknickman:
### Description

Mock config for telemetry tests

Closes TURBO-2189

PR #7172, opened 2024-01-30
chore(integration): remove _helpers & fallback

tknickman:
### Description

Removes old `_helpers` and a few go fallback references

Follow up to https://github.com/vercel/turbo/pull/7168

Closes TURBO-2192


turborepo-tests/helpers/find_global_hash.sh
```diff
@@ -6,7 +6,7 @@
 # 2023-04-06T04:28:19.599Z [DEBUG] turbo: global hash: value=a027dadc4dea675e
 #
 # Usage:
-# turbo build -vv 2>&1 | "$TESTDIR/./get-global-hash.sh"
+# turbo build -vv 2>&1 | "$TESTDIR/./find_global_hash.sh"
```
tknickman, line 9: Do I feel strongly about this? No. But I like how we have everything segmented in the global helpers directory by prefix:

`setup_*`
`mock_*`
`utils_*`

So wanted to generalize this. But not locked in.

turborepo-tests/integration/tests/run-summary/monorepo.t
```diff
@@ -72,7 +72,7 @@ Setup
   [0-9]+ (re)
 
 # Extract some task-specific summaries from each
-  $ source "$TESTDIR/../_helpers/run-summary-utils.sh"
+  $ source "$TESTDIR/../../../helpers/utils_run_summary.sh"
```
mehulkar, line 75: I'd get rid of `utils_`, it's already in helpers, and they're kinda interchangeable terms, so it feels redundant

turborepo-tests/integration/turbo.json
```diff
@@ -9,10 +9,6 @@
     "test": {
       "dependsOn": ["cli#build", "topo", "^build"],
       "passThroughEnv": ["CI"]
-    },
-    "test:go-fallback": {
```
mehulkar, line 13: Did you find this through the new LSP? :)

mehulkar (approved):
Nice!

One minor comment inline. 

Another thing I'd like to do is invoke `helper` scripts via `./node_modules`. Since `helpers` is a package, its contents _should_ be available in the `bin` directory in locally installed node modules of the test packages. (not needed in this PR though)

turborepo-tests/integration/README.md
```diff
@@ -52,7 +52,7 @@ Setup
   $ . ${TESTDIR}/../../helpers/setup_integration_test.sh single_package
 
 Custom config
-  $ . ${TESTDIR}/../../helpers/replace_turbo_config.sh $(pwd) "my-custom-config.json"
+  $ . ${TESTDIR}/../../helpers/replace_turbo_json.sh $(pwd) "my-custom-config.json"
```
tknickman, line 55: This was an intentional rename - we have another helper that operates on the global turbo config, so I wanted to differentiate the turbo.json from the turbo config

turborepo-tests/integration/turbo.json
tknickman: Just a `go-fallback` grep 😄

PR #7173, opened 2024-01-30
add vc config dir

Zertsov:

PR #7174, opened 2024-01-30
use vc token if logging into Vercel

Zertsov:
### Description
If users are logging into Vercel, we should allow for them to use an existing Vercel CLI token if it exists. This is only for non-SSO logins. SSO will follow in another PR.

### Testing Instructions
1. Use `vc login` to get a Vercel token. Do _not_ use SSO (this doesn't implement that yet).
2. Make sure you don't have a `turbo` token by running `turbo logout`.
3. Run `turbo login` - you should see a message saying `Existing Vercel token found!`
4. Run `turbo link` on a repo you have access to in Vercel
5. Try running a cacheable task. You should see that remote caching is enabled.

Closes TURBO-2194


Zertsov:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/vercel/turbo/7174?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `main`
  * **#7173** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7173?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **#7174** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7174?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -20,13 +32,45 @@ pub async fn login<'a>(
...
+                    println!(
+                        "{}",
+                        ui.apply(BOLD.apply_to("Existing Vercel token found!"))
+                    );
+                    ui::print_cli_authorized(&response.user.email, ui);
+                    return Ok(Token {
+                        token,
+                        is_existing: true,
+                    });
+                }
```
chris-olszewski, line 72: This logic is duped from above, seems like a good candidate for extraction into a new function.

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -20,13 +32,45 @@ pub async fn login<'a>(
...
+        if let Some(vercel_config_dir) = turborepo_dirs::vercel_config_dir() {
+            let vercel_token_path = vercel_config_dir
+                .join(VERCEL_CONFIG_DIR)
+                .join(VERCEL_TOKEN_FILE);
+            #[derive(serde::Deserialize)]
+            struct VercelToken {
+                token: String,
+            }
+            let contents = std::fs::read_to_string(&vercel_token_path)?;
+            if let Ok(token) = serde_json::from_str::<VercelToken>(&contents) {
```
chris-olszewski, line 59: The find and parse Vercel token logic should get extracted to it's own function.

chris-olszewski (changes requested):

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -11,6 +11,18 @@ use crate::{error, server::LoginServer, ui};
...
+const VERCEL_TOKEN_FILE: &str = "auth.json";
+
+/// Token is the result of a successful login. It contains the token string and
+/// a boolean indicating whether the token already existed on the filesystem.
+pub struct Token {
+    /// The actual token string.
+    pub token: String,
+    /// If this is `true`, it means this token already exists on the filesystem.
+    /// If `false`, this is a new token.
+    pub is_existing: bool,
```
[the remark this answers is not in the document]
Zertsov: Looks like I forgot to add the
> skip writing to disk if this exists
logic, but that's where I was planning on using it. If we don't care about an extra filesystem op (re-writing existing data to the spot it's already at), then I can remove this entirely. I also wasn't sold on the naming here, so open to better names. Maybe `exists_on_disk`?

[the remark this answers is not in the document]
chris-olszewski: I would say leave it out until we have the logic for skipping the FS operation

crates/turborepo-auth/src/auth/mod.rs
```diff
@@ -5,3 +5,35 @@ mod sso;
...
+        token: String,
+    }
+
+    let token = serde_json::from_str::<VercelToken>(&contents)?.token;
+    api_client.get_user(&token).await?;
+    Ok(Token {
+        token,
+        is_existing: true,
+    })
+}
```
chris-olszewski, line 39: Sorry, should've been more clear about refactoring suggestions.

This would let us share the "display a token was found" code as they both operate on just having a token.

```suggestion
async fn extract_vercel_token() -> Result<String, Error> {
    let vercel_config_dir = turborepo_dirs::vercel_config_dir().ok_or_else(|| {
        Error::Io(std::io::Error::new(
            std::io::ErrorKind::NotFound,
            "Vercel config directory not found",
        ))
    })?;

    let vercel_token_path = vercel_config_dir
        .join(VERCEL_TOKEN_DIR)
        .join(VERCEL_TOKEN_FILE);
    let contents = std::fs::read_to_string(&vercel_token_path)?;

    #[derive(serde::Deserialize)]
    struct VercelToken {
        // This isn't actually dead code, it's used by serde to deserialize the JSON.
        #[allow(dead_code)]
        token: String,
    }

    serde_json::from_str::<VercelToken>(&contents)?.token
}
```

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -20,13 +30,35 @@ pub async fn login<'a>(
...
+                    "{}",
+                    ui.apply(BOLD.apply_to("Existing Vercel token found!"))
+                );
+                return Ok(token);
+            }
+            Err(error) => {
+                // Only send the warning if we're debugging.
+                if cfg!(debug_assertions) {
+                    warn!("Failed to extract Vercel token: {error}");
+                }
```
chris-olszewski, line 60: Do you want this to be emitted when `-vv` is passed or just on the debug build of `turbo`?

Also with the current setup this error can come from either us failing to find the vercel config dir or an issue hitting the API.

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -20,13 +30,35 @@ pub async fn login<'a>(
...
     login_url_configuration: &str,
     login_server: &impl LoginServer,
-) -> Result<Cow<'a, str>, Error> {
-    // Check if token exists first.
+) -> Result<Token, Error> {
+    // Check if passed in token exists first.
     if let Some(token) = existing_token {
         if let Ok(response) = api_client.get_user(token).await {
             println!("{}", ui.apply(BOLD.apply_to("Existing token found!")));
             ui::print_cli_authorized(&response.user.email, ui);
```
chris-olszewski, line 38: This is the chunk of logic that can get shared between the turbo token and vercel token loading.

```
async fn check_token(token: String, api_client: &impl Client, message: &str) -> Result<Token> {
    let response = api_client.get_user(&token).await?;
    println!("{}", ui.apply(BOLD.apply_to(message)));
    ui::print_cli_authorized(&response.user.email, ui);
    Ok(Token { token, is_existing: true } )
}
```

crates/turborepo-auth/src/auth/mod.rs
```diff
@@ -5,3 +5,35 @@ mod sso;
...
+
+const VERCEL_TOKEN_DIR: &str = "com.vercel.cli";
+const VERCEL_TOKEN_FILE: &str = "auth.json";
+
+async fn extract_vercel_token(api_client: &impl Client) -> Result<Token, Error> {
+    let vercel_config_dir = turborepo_dirs::vercel_config_dir().ok_or_else(|| {
+        Error::Io(std::io::Error::new(
+            std::io::ErrorKind::NotFound,
+            "Vercel config directory not found",
+        ))
```
chris-olszewski, line 18: This should be a new `Error` variant instead of making an IO error.

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -20,13 +30,35 @@ pub async fn login<'a>(
...
-            return Ok(token.into());
+            return Ok(Token {
+                token: token.into(),
+                is_existing: true,
+            });
+        }
+    }
+
+    // If the user is logging into Vercel, check for an existing `vc` token.
+    if login_url_configuration.contains("vercel.com") {
```
chris-olszewski, line 47: Is there anyway we could get some automated test either via a unit test or a prysk test that could test this path getting hit?

chris-olszewski (changes requested):

crates/turborepo-auth/src/auth/login.rs
Zertsov: I made the suggestions to the extract logic so that now it can only be if we didn't find the config dir. I was waffling about whether we want this just for the debug build or via flags, since this shouldn't stop us from logging in, and I didn't think we needed an extra log line telling the user we didn't find the config dir, but I'm open to thoughts here

chris-olszewski: Just putting this as a debug log makes more sense to me? It allows us to see if a user is having issues with a vercel token without having to give them a special build.

crates/turborepo-auth/src/auth/mod.rs
```diff
@@ -5,3 +5,77 @@ mod sso;
...
+const VERCEL_TOKEN_FILE: &str = "auth.json";
+
+pub struct LoginOptions<'a, T>
+where
+    T: Client,
+{
+    pub ui: &'a UI,
+    pub login_url: &'a str,
+    pub api_client: &'a T,
+    pub login_server: &'a dyn LoginServer,
```
Zertsov, line 23: I think I can do this "cleaner" in another PR, but I don't want to get too far out of the initial scope of the PR

crates/turborepo-auth/src/auth/mod.rs
```diff
@@ -5,3 +5,77 @@ mod sso;
...
+
+    pub sso_team: Option<&'a str>,
```
Zertsov, line 25: This will be used in the next branch

crates/turborepo-auth/src/error.rs
```diff
@@ -19,4 +26,6 @@ pub enum Error {
     FailedToValidateSSOToken(#[source] turborepo_api_client::Error),
     #[error("failed to make sso token name")]
     FailedToMakeSSOTokenName(#[source] io::Error),
+    #[error("config directory not found")]
```
Zertsov, line 29: In a future PR I want to move this into the `turborepo_dirs` crate, but for now it can live here.

chris-olszewski (approved):

Zertsov:
### Merge activity

* **Feb 3, 12:22 AM EST**: [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/7174) couldn't merge this PR because it was not satisfying all requirements.

PR #7178, opened 2024-01-31
feat(lockfile): traverse lockfile in parallel

chris-olszewski:
### Description

I remembered that the FFI interface was what was stopping me from throwing rayon at the lockfile graph traversal. This can be done easily since we already require that all lockfile implementations are `Send` and they don't mutate the underlying lockfile.

There's a more elegant way to do this and avoid calculating inner closures multiple times, but this is a straightforward win.

### Testing Instructions
From 
Screenshot 2024-01-30 at 5 40 42 PM
To
Screenshot 2024-01-30 at 5 40 32 PM

[serial.json](https://github.com/vercel/turbo/files/14106106/profile.json)
[parallel.json](https://github.com/vercel/turbo/files/14106107/parallel.json)

Closes TURBO-2197

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11798,6 +11798,7 @@ dependencies = [
  "pest",
  "pest_derive",
  "pretty_assertions",
+ "rayon",
  "regex",
  "semver 1.0.18",
  "serde",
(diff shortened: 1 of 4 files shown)


arlyon (approved):
Nice!

PR #7183, opened 2024-01-31
feat(turborepo): Daemon package watch

arlyon:
### Description

This PR allows the daemon to watch for package changes and eagerly update hashes. Each package / task maintains a list of hashes that are updated by a file watcher. When turbo launches, it queries that daemon instead of manually walking.

### Testing Instructions

Added a subcommand  `turbo daemon hash <task>` that lets us test the daemon hashing code

Closes TURBO-2200


weyert:
Would this be able to be used as a way to trigger tasks when files changes in dependencies/dependents of workspaces? A bit like what `turbowatch` allows to do but then natively.

arlyon:
The same infrastructure can definitely be used for alerting the client about package changes but this PR is scoped to relieving the client of having to calculate hashes and speeding up 'time to first task'

arlyon (commented):
Superceded, closing

PR #7187, opened 2024-01-31
feat(turbo-repository): return additional workspace data

mrmckeb:
### Description

This PR returns additional data:
- Relative paths are returned (again).
- The package manager name is now returned within an object, without a function call.
- The package manager version is now returned.

It also adds TSDoc for clarity around return values.

### Testing Instructions

N/A, tests included.


mehulkar (approved):
This is all fine with me. What was the reason to replace `packageManager()` with `packageManager`?

mrmckeb:
Thanks @mehulkar, as it wasn't actually calculating anything, it felt simpler to just have the object available - one less step to access the data.

PR #7203, opened 2024-02-01
fix(Turborepo): Don't block forever if we have already gotten a parse error

gsoltis:
### Description

 - If we've already encountered a parse error, don't wait for the child process to exit cleanly, just kill it. Waiting nicely could mean waiting forever.
 - Only read from `stderr` after we've killed the process, otherwise that could block forever if there is nothing available.

### Testing Instructions

Added new unit test simulating a parse error with a script that would hang forever under the old method.

Closes TURBO-2212

### DIFF
--- a/crates/turborepo-scm/src/lib.rs
+++ b/crates/turborepo-scm/src/lib.rs
@@ -312,7 +312,9 @@ mod tests {
             let script_path = root.join_component("hanging.sh");
             script_path.create_with_contents(sh).unwrap();
             script_path.set_mode(0x755).unwrap();
-            Command::new(&bash).arg(script_path.as_str())
+            let mut cmd = Command::new(bash);
+            cmd.arg(script_path.as_str());
+            cmd
         };
 
         let mut cmd = cmd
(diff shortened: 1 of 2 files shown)


crates/turborepo-scm/src/lib.rs
```diff
@@ -99,36 +99,47 @@ pub(crate) fn wait_for_success<R: Read, T>(
...
+            "'{}' in {}{}{}",
+            command,
+            root_path.as_ref(),
+            stderr_text,
+            parse_err
+        );
+        return Err(Error::Git(err_text, Backtrace::capture()));
+    }
+    // TODO: if we've successfully parsed the output, but the command is hanging for
+    // some reason, we will currently block forever.
```
chris-olszewski, line 121: I think we should try to be better about making sure we aren't ever the reason the command is hanging by always reading stdout/stderr until EOF. I think this probably means we read stderr in another thread.

Not super likely that the command ends up hanging because either the stdout or stderr pipe is full, but it's a possibility that we can avoid.

chris-olszewski (approved):
This is a step in the right direction, but in the future we should probably just spin up threads that make sure the pipes being full is never the reason for a hanging command.

crates/turborepo-scm/src/lib.rs
gsoltis: In this case, `git status` did hang because `stdout` was full. But, agreed. A reworking of how we handle `stdout`/`stderr` sounds good to me.

PR #7204, opened 2024-02-01
chore(ci): combine rust and common integration tests

mehulkar:
Now that Go code is deleted it shouldn't be necessary to separate these

Closes TURBO-2213


turborepo-tests/integration/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "turborepo-tests-integration-common",
+  "name": "turborepo-tests-integration",
```
mehulkar, line 2: I think I'm missing something because I don't see this package name referenced anywhere, so I'm not sure how it was running these tests

arlyon (approved):

chris-olszewski (approved):

mehulkar:
Only the turbopack rust testing check seem to be failing, so I will merge without being green. cc @vercel/turbopack

PR #7205, opened 2024-02-01
chore(turbo): delete Go related release code and config

mehulkar:
Closes TURBO-2214


sokra (approved):

.github/workflows/test.yml
```diff
@@ -846,7 +846,7 @@ jobs:
 
       - name: Lint
         # Filters some workspaces out:
-        # - `cli` because it runs golangci-lint and this Job doesn't have all the dependencies for that.
+        # - `cli` because (no reason anymore)
```
chris-olszewski, line 849: Should we stop filtering out `cli` then?

.github/workflows/turborepo-release.yml
```diff
@@ -193,15 +193,6 @@ jobs:
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
 
-      - name: Install GoReleaser
```
chris-olszewski, line 196: We still need this, we haven't ripped out `goreleaser` yet. `make publish-turbo` invokes it.

cli/Makefile
```diff
@@ -118,9 +78,6 @@ publish-turbo: clean build
 
 	npm config set --location=project "//registry.npmjs.org/:_authToken" $(NPM_TOKEN)
 
-	# Publishes the native npm modules.
-	goreleaser release --rm-dist -f combined-shim.yml $(SKIP_PUBLISH)
-
```
chris-olszewski, line 123: I don't think we can get rid of this. It is executing some code that prepares the various platform packages.

chris-olszewski (changes requested):
I'm almost positive we can't just remove our calls to `goreleaser` without doing some additional work ourselves

.github/workflows/test.yml
mehulkar: separate PR, I don't know if linting is setup there at all

cli/Makefile
mehulkar: brought back all the Makefile and `*.yml` and goreleaser stuff

.github/workflows/test.yml
mehulkar: actually `lint` script is gone from there, so we can safely remove this `!cli`. Done

chris-olszewski:
https://github.com/vercel/turbo/pull/7221 should fix the integration test failures

PR #7213, opened 2024-02-01
feat(turbopack): extend ecmatransform moduleeffect

kwonoj:
Closes PACK-2356

This PR creates a new ModuleRuleEffect named `ExtendEcmascriptTransforms`. This is an extended version of `AddEcmascriptTransforms` effect, which will be removed once next.js replaces all of its references.

The change born in need of allowing custom rules in `ModuleOptionContext` to have a specific order of transform. This was been achieved by having a custom ecmatransform plugin in moduleoptions. However, it was not very clear for the distinction between rules & transforms, as well as there are some behavior differences we noticed (note: this still may need to be investigated separately later). After discussing with @sokra , we concluded it is better to deprecate custom ecmatransformplugin but use custom rules whereever possible.

One functional differences between current ecmatransformplugin to custom rules is custom rule always append to the main rules, so there wasn't way to adjust order of ecmatransform as we want. The change addresses that moduleeffect can specify if ecmatransform supposed to be prepended or not.

### DIFF
--- a/crates/turbopack/src/module_options/module_rule.rs
+++ b/crates/turbopack/src/module_options/module_rule.rs
@@ -93,7 +93,10 @@ pub enum ModuleRuleEffect {
     /// Allow to extend an existing Ecmascript module rules for the additional
     /// transforms. First argument will prepend the existing transforms, and
     /// the second argument will append the new transforms.
-    ExtendEcmascriptTransforms((Vc<EcmascriptInputTransforms>, Vc<EcmascriptInputTransforms>)),
+    ExtendEcmascriptTransforms {
+        prepend: Vc<EcmascriptInputTransforms>,
+        append: Vc<EcmascriptInputTransforms>,
+    },
     SourceTransforms(Vc<SourceTransforms>),
 }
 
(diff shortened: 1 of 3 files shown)


crates/turbopack/src/module_options/module_rule.rs
```diff
@@ -85,7 +85,15 @@ impl ModuleRule {
...
     ModuleType(ModuleType),
+    #[deprecated(
+        note = "ExtendEcmascriptTransforms provides equivalent features, as well as supporting an \
+                additional way to prepend transforms."
+    )]
     AddEcmascriptTransforms(Vc<EcmascriptInputTransforms>),
+    /// Allow to extend an existing Ecmascript module rules for the additional
+    /// transforms. First argument will prepend the existing transforms, and
+    /// the second argument will append the new transforms.
+    ExtendEcmascriptTransforms((Vc<EcmascriptInputTransforms>, Vc<EcmascriptInputTransforms>)),
```
sokra, line 96:
```suggestion
    ExtendEcmascriptTransforms { prepend: Vc<EcmascriptInputTransforms>, append: Vc<EcmascriptInputTransforms> },
```

PR #7218, opened 2024-02-01
chore: remove npm-run-all and usage from root package.json

mehulkar:
as far as I know, these tasks aren't being used anymore and we don't need this dependency

Closes TURBO-2221


chris-olszewski (approved):

chris-olszewski:
Will say that the amount of lockfile additions when removing this dependency feels weird to me.

pnpm-lock.yaml
```diff
@@ -13821,6 +14019,7 @@ packages:
       command-exists: 1.2.9
       console.table: 0.10.0
       detect-libc: 2.0.2
+      zx: 7.2.3
```
mehulkar, line 13962: this is one of the places new deps are coming from. `rust: nightly` has a new dependency

pnpm-lock.yaml
```diff
@@ -12168,6 +12286,49 @@ packages:
       - '@babel/core'
       - babel-plugin-macros
 
+  /next@13.4.12(react-dom@18.2.0)(react@18.2.0):
```
mehulkar: And this is the other place. package.json has:
```
"next": "^13.0.6",
```

so it'll go up minor versions.

mehulkar (commented):
@chris-olszewski the lockfile changes were for a couple reasons, I pointed them out in the diff. They were happening because `pnpm rm <dep>` doesn't have a way to tell pnpm to not touch other deps. [But turns out `pnpm i --prefer-offline` does help with that](https://stackoverflow.com/a/74459421/986415), so I switched to manually edit package.json and then run `pnpm i --prefer-offline`, and that at least got rid of the nextjs dependency update[^1]. The `rust: nightly` dep insists on being updated.

[^1]: weird that nextjs is even a root level dependency, but that's for another PR.

PR #7222, opened 2024-02-01
refactor(turborepo): Change Detector

NicholasLYang:
### Description

In preparation for watch mode, I refactored the change detector to be split into two parts:
    - A change mapper that takes the changed files and a lockfile change and produces changed packages (now in `turborepo_repository`)
    - A git change detector that takes two git refs and produces changed packages using the change mapper.

This should allow us to use the change mapper in other situations such as ignored builds and watch mode.

### Testing Instructions

Closes TURBO-2223


crates/turborepo-repository/src/change_mapper.rs
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        Ok(changed_packages)
+    }
+
+    fn is_file_in_package(file: &AnchoredSystemPath, package_path: &AnchoredSystemPath) -> bool {
+        file.components()
+            .zip(package_path.components())
+            .all(|(a, b)| a == b)
+    }
```
chris-olszewski, line 134: I know this is just copied, but is there something I'm missing? Shouldn't we be able to use `==`? If they're both system paths, they should be using the same separator.

crates/turborepo-repository/src/change_mapper.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    fn repo_global_file_has_changed(
+        &self,
+        default_global_deps: &[&str],
+        changed_files: &HashSet<AnchoredSystemPathBuf>,
+    ) -> Result<bool, turborepo_scm::Error> {
+        let global_deps = self.global_deps.iter().map(|s| s.as_str());
+        let filters = global_deps.chain(default_global_deps.iter().copied());
+        let matcher = wax::any(filters).unwrap();
```
chris-olszewski, line 86: I'm pretty sure tracing through the code that these globs haven't been validated. We should probably throw here since it's user input.

chris-olszewski (approved):

crates/turborepo-repository/src/change_mapper.rs
NicholasLYang: Hmm so I think the rationale is that it accounts for cases where the paths are not of equal length. Like it just checks that the overlapping parts of those two paths are equal. But we don't check that `file` is necessarily the longer path here.

PR #7223, opened 2024-02-01
Add "System environment variables" page.

anthonyshew:

### DIFF
--- a/docs/pages/repo/docs/reference/system-variables.mdx
+++ b/docs/pages/repo/docs/reference/system-variables.mdx
@@ -1,0 +1,28 @@
+---
+title: System environment variables
+description: System environment variables are built-in variables that are available to all scripts.
+---
+
+## System environment variables
+
+By setting certain environment variables, you can change Turborepo's behavior. This can be useful for creating specific configurations for different environments and machines.
+
+| Variable                           | Description                                                                                                                                                                                                                                   |
+| ---------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `TURBO_API`                        | Set the base URL for [Remote Cache](/repo/docs/core-concepts/remote-caching).                                                                                                                                                                 |
+| `TURBO_BINARY_PATH`                | Manually set the path to the `turbo` binary.                                                                                                                                                                                                  |
+| `TURBO_CI_VENDOR_ENV_KEY`          | Set a prefix for environment variables that you want **excluded** from [Framework Inference](/repo/docs/core-concepts/caching/environment-variable-inputs#framework-inference).                                                               |
+| `TURBO_FORCE`                      | Always force all tasks in your pipelines to run in full, opting out of all caching.                                                                                                                                                           |
+| `TURBO_HASH`                       | The hash of the currently running task.                                                                                                                                                                                                       |
+| `TURBO_LOG_ORDER`                  | Set the [log order](https://turbo.build/repo/docs/reference/command-line-reference/run#--log-order) for your pipeline's logs. Allowed values are `grouped` and `default`.                                                                     |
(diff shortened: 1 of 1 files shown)


docs/pages/repo/docs/reference/system-variables.mdx
```diff
@@ -0,0 +1,28 @@
...
+| `TURBO_NO_UPDATE_NOTIFIER`         | Remove the update notifier that appears when a new version of `turbo` is available. You can also use `NO_UPDATE_NOTIFIER` per ecosystem convention.                                                                                           |
+| `TURBO_PREFLIGHT`                  | Enables sending a preflight request before every cache artifact and analytics request. The follow-up upload and download will follow redirects. Only applicable when [Remote Caching](/repo/docs/core-concepts/remote-caching) is configured. |
+| `TURBO_REMOTE_CACHE_READ_ONLY`     | Prevent writing to the [Remote Cache](/repo/docs/core-concepts/remote-caching) - but still allow reading.                                                                                                                                     |
+| `TURBO_REMOTE_CACHE_TIMEOUT`       | Set a timeout for `turbo` to get artifacts from [Remote Cache](/repo/docs/core-concepts/remote-caching).                                                                                                                                      |
```
chris-olszewski, line 22: Just because I always need to look it up, we should document that this is in seconds.
```suggestion
| `TURBO_REMOTE_CACHE_TIMEOUT`       | Set a timeout in seconds for `turbo` to get artifacts from [Remote Cache](/repo/docs/core-concepts/remote-caching).                                                                                                                                      |
```

docs/pages/repo/docs/reference/system-variables.mdx
```diff
@@ -0,0 +1,28 @@
...
+| `TURBO_BINARY_PATH`                | Manually set the path to the `turbo` binary.                                                                                                                                                                                                  |
+| `TURBO_CI_VENDOR_ENV_KEY`          | Set a prefix for environment variables that you want **excluded** from [Framework Inference](/repo/docs/core-concepts/caching/environment-variable-inputs#framework-inference).                                                               |
+| `TURBO_FORCE`                      | Always force all tasks in your pipelines to run in full, opting out of all caching.                                                                                                                                                           |
+| `TURBO_HASH`                       | The hash of the currently running task.                                                                                                                                                                                                       |
```
chris-olszewski, line 16: I'm not sure if this belongs here. This env var doesn't control `turbo`'s behavior, but `turbo` provides it to tasks during their execution.

docs/pages/repo/docs/reference/system-variables.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+| Variable                           | Description                                                                                                                                                                                                                                   |
+| ---------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `TURBO_API`                        | Set the base URL for [Remote Cache](/repo/docs/core-concepts/remote-caching).                                                                                                                                                                 |
+| `TURBO_BINARY_PATH`                | Manually set the path to the `turbo` binary.                                                                                                                                                                                                  |
```
chris-olszewski, line 13: Question: is it worth going into the gotcha's of `TURBO_BINARY_PATH` here? This was developed before global turbo so using it completely opts out of trying to find a local version.

chris-olszewski (approved):
Approving as I think it's all correct, just some comments on possible additions/removals.

docs/pages/repo/docs/reference/system-variables.mdx
mehulkar: That sounds like a good idea

mehulkar (commented):
Love that we're finally documenting this. What do you think of separating these into different lists:

- Env vars meant for the user to set to configure turbo (could link to the equivalent flag in /command-line-reference for each of these)
- Env vars that are internal to turbo, but we're documenting them anyway
- Env vars that are made available to scripts (so far only `TURBO_HASH`)

PR #7224, opened 2024-02-01
chore: update github actions

ForsakenHarmony:
### Description

The `node16` runner is deprecated, I updated all the actions I could here.

Still missing:
~~https://github.com/pnpm/action-setup~~ (removed in favor of corepack, but that needs to be tested)
https://github.com/technote-space/get-diff-action (archived, needs to be replaced)
https://github.com/dependabot/fetch-metadata (has PR: https://github.com/dependabot/fetch-metadata/pull/443)
~~https://github.com/ts-graphviz/setup-graphviz (has PR: https://github.com/ts-graphviz/setup-graphviz/pull/558)~~
https://github.com/thomaseizinger/create-pull-request

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -59,14 +59,14 @@ runs:
     - name: Set Up Protoc
       id: set-up-protoc
       continue-on-error: true
-      uses: arduino/setup-protoc@v1.2.0
+      uses: arduino/setup-protoc@v3.0.0
       with:
         version: "3.x"
         repo-token: ${{ inputs.github-token }}
 
     - name: Set Up Protoc (second try)
       if: steps.set-up-protoc.outcome == 'failure'
-      uses: arduino/setup-protoc@v1.2.0
+      uses: arduino/setup-protoc@v3.0.0
       with:
         version: "3.x"
         repo-token: ${{ inputs.github-token }}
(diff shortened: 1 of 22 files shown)


mehulkar (commented):
Kicking off some dry runs for CI workflows to validate:

- ❌ [Dry run for turborepo release](https://github.com/vercel/turbo/actions/runs/7834625823)
- [Benchmark](https://github.com/vercel/turbo/actions/runs/7834636474)
- ❌ [Library release](https://github.com/vercel/turbo/actions/runs/7834642627)

mehulkar (changes requested):
Looks like there's some glibc missing issue for both our release workflows. We'd need to work those out before merge

mehulkar (approved):

.github/workflows/lsp.yml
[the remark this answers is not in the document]
arlyon: Answered in DM but repeating here; the LSP release is expected to fail in one case so I will resolve this.

.github/actions/setup-rust/action.yml
```diff
@@ -58,16 +58,16 @@ runs:
     - name: Set Up Protoc
       id: set-up-protoc
       continue-on-error: true
-      uses: arduino/setup-protoc@v1.2.0
+      uses: arduino/setup-protoc@v3.0.0
       with:
-        version: "3.x"
+        version: "23.x"
         repo-token: ${{ inputs.github-token }}
```
kdy1, line 64: I think we may add some comments describing that this is to avoid rate limit.

kdy1: Mainly because passing token to `setup-rust` action looks weird.

kdy1 (approved):

anthonyshew:
Will close this as stale.

PR #7235, opened 2024-02-02
feat(docs): remote cache open api spec

tknickman:
### Description

Open API spec for the remote cache server

Closes TURBO-2227


anthonyshew (approved):
Love it.

PR #7238, opened 2024-02-02
fix: avoid expensive globwalk in examples

chris-olszewski:
### Description

Q: how long would a `turbo lint` call would take in our repository?
A: 13 minutes

12 minutes of that is collecting inputs for `turborepo-examples:lint-prettier` which it inherits from the root definition.
Screenshot 2024-02-02 at 11 36 38 AM

This PR overrides that input list and now it only takes ~200ms to do nothing. 

### Testing Instructions

Screenshot 2024-02-02 at 11 39 08 AM

Closes TURBO-2231

### DIFF
--- a/examples/turbo.json
+++ b/examples/turbo.json
@@ -3,6 +3,10 @@
   "pipeline": {
     // This needs to exist because upstream packages depend on it and
     // it's not defined at the root level.
-    "topo": {}
+    "topo": {},
+    // We override inputs to avoid a costly globwalk
+    "lint:prettier": {
+      "inputs": []
+    }
   }
 }


chris-olszewski:
> But the profile doesn't contain turborepo-examples

Very interesting, `turborepo-examples#lint:prettier` should still be getting a task hash even if there isn't a script definition.

Mind sharing your profile?

> Is it possible that when you gathered your profile you were somehow getting opted into it because of a local git-ignored change?

I do have `node_modules` present in the examples from when I was testing them.

chris-olszewski:
After `git clean -f -x -d` which deleted all of the `node_modules` we it dropped the calc file hash for `turborepo-examples#lint:prettier` to just 4s, but I still think it makes sense to drop this.
Screenshot 2024-02-02 at 1 47 44 PM

tknickman (approved):

PR #7242, opened 2024-02-02
feat(@turbo/repository): add method to return package dependents

mehulkar:
Add `.packageGraph()` method that returns a map where keys are relative package paths and values are a list of package paths that  depend on that package. All paths are anchored at the root of the monorepo.

This method can be able to answer the question: Given that Package X changed, what _other_ packages should also be considered changed?

`packageGraph()` is not quite the best name for this as it only includes `dependents` (not `dependencies`). This may be ok, but seems surprising given that packageGraphs usually traverse the other way.

Closes TURBO-2233


packages/turbo-repository/rust/src/lib.rs
```diff
@@ -63,4 +82,18 @@ impl Workspace {
     pub async fn find_packages(&self) -> std::result::Result<Vec<Package>, napi::Error> {
         self.packages_internal().await.map_err(|e| e.into())
     }
+
+    #[napi]
+    pub async fn package_graph(&self) -> HashMap<String, Vec<String>> {
```
mehulkar, line 87: I would like to just return the whole Package here instead of the names, but (1) not sure if that's necessary and (2) not sure how to implement `ToNapiValue` trait so it can get into JS land.

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -63,4 +70,20 @@ impl Workspace {
...
+        let mut map = HashMap::new();
+        let packages = self.find_packages().await.unwrap();
```
chris-olszewski, line 77:
```suggestion
        let packages = self.find_packages().await?;
```

packages/turbo-repository/rust/src/lib.rs
chris-olszewski: The `ToNapiValue` trait is only implemented for `HashMap`'s that have a key type that can be [referenced as a `str`](https://docs.rs/napi/latest/napi/bindgen_prelude/trait.ToNapiValue.html#impl-ToNapiValue-for-HashMap%3CK,+V,+S%3E). If you want the full package information available for each entry you might need to return something like
`HashMap<String, Node>` where `struct Node { package: Package, deps: Vec<Package> }`

mehulkar: I'll just do relative paths for now, I _think_ that's the only thing we need right now. Will be taking multiple passes at this over the next couple days I'm sure

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -35,18 +42,46 @@ pub struct Workspace {
...
+        workspace_path: &AbsoluteSystemPath,
+    ) -> Vec<Package> {
+        let node = WorkspaceNode::Workspace(WorkspaceName::Other(self.name.clone()));
+        let ancestors = graph.immediate_ancestors(&node).unwrap();
+
+        ancestors
+            .iter()
+            .map(|node| {
+                let info = graph.workspace_info(node.as_workspace()).unwrap();
+                let name = info.package_name().unwrap();
```
tknickman, line 78: These `.unwrap()`'s should be handled.

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -63,4 +98,21 @@ impl Workspace {
...
         self.packages_internal().await.map_err(|e| e.into())
     }
+
+    #[napi]
+    pub async fn package_graph(&self) -> Result<HashMap<String, Vec<String>>, Error> {
+        let mut map = HashMap::new();
+        let packages = self.find_packages().await?;
+        let workspace_path = AbsoluteSystemPath::new(self.absolute_path.as_str()).unwrap();
+
+        for (_i, package) in packages.iter().enumerate() {
```
tknickman, line 108: Not required - but you could map this as well, and then return tuples that get collected into key val pairs to avoid the mutable hashmap

```
let map: HashMap<String, Vec<String>> = packages.iter()
        .map(|package| {
            let deps = package.dependents(&self.graph, workspace_path);
            let dep_names = deps.iter().map(|p| p.relative_path.clone()).collect::<Vec<String>>();

            (package.relative_path.clone(), dep_names)
        })
        .collect();
```

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -63,4 +98,21 @@ impl Workspace {
     pub async fn find_packages(&self) -> std::result::Result<Vec<Package>, napi::Error> {
         self.packages_internal().await.map_err(|e| e.into())
     }
+
+    #[napi]
+    pub async fn package_graph(&self) -> Result<HashMap<String, Vec<String>>, Error> {
+        let mut map = HashMap::new();
+        let packages = self.find_packages().await?;
+        let workspace_path = AbsoluteSystemPath::new(self.absolute_path.as_str()).unwrap();
```
tknickman, line 106: Should handle this unwrap as well.

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -35,18 +42,46 @@ pub struct Workspace {
...
+    fn new(
+        name: String,
+        workspace_path: &AbsoluteSystemPath,
+        package_path: &AbsoluteSystemPath,
+    ) -> Self {
         let relative_path = workspace_path
             .anchor(package_path)
             .expect("Package path is within the workspace");
         Self {
+            name: name,
```
tknickman, line 60: Shorthand works in rust too!

```suggestion
            name,
```

packages/turbo-repository/rust/src/lib.rs
mehulkar: done

packages/turbo-repository/rust/src/lib.rs
mehulkar: good idea, done

packages/turbo-repository/rust/src/lib.rs
mehulkar: How would you handle this? This is probably a panic situation or at least need to return some error. If we don't have the workspace absolute path , we can't do the rest

mehulkar: checkout 10788eab14c9c6cc41e3c4950812e2212899fe58

packages/turbo-repository/rust/src/internal.rs
```diff
@@ -42,6 +46,18 @@ impl From<Error> for napi::Error<Status> {
     }
 }
 
+impl From<turborepo_repository::package_json::Error> for Error {
+    fn from(err: turborepo_repository::package_json::Error) -> Self {
```
chris-olszewski, line 50: We shouldn't be throwing away error details

packages/turbo-repository/rust/src/internal.rs
```diff
@@ -34,6 +36,8 @@ pub(crate) enum Error {
     },
     #[error("Package directory {0} has no parent")]
     MissingParent(AbsoluteSystemPathBuf),
+    #[error("Package graph error")]
+    PackageGraphError,
```
chris-olszewski, line 40: This implements `From<turborepo::package_graph::Error>` for you and will forward the underlying error message.
```suggestion
    #[error("Package graph error: {0}")]
    PackageGraph(#[from] turborepo_repository::package_graph::Error),
    #[error("package.json error: {0}")]
    PackageJson(#[from] turborepo_repository::package_json::Error),
```

packages/turbo-repository/rust/src/internal.rs
```diff
@@ -67,13 +83,22 @@ impl Workspace {
                 path: workspace_state.root.clone(),
             })?
             .clone();
+
+        let workspace_root = workspace_state.root.clone();
```
chris-olszewski, line 87: Unnecessary clone
```suggestion
        let workspace_root = &workspace_state.root;
```

packages/turbo-repository/rust/src/internal.rs
```diff
@@ -67,13 +83,22 @@ impl Workspace {
                 path: workspace_state.root.clone(),
             })?
             .clone();
```
chris-olszewski, line 85: We don't need the clone here as `PackageManager` implements `Copy` which means we can just dereference a `&PackageManager` instead of cloning it to get a `PackageManager`. See https://rust-lang.github.io/rust-clippy/master/index.html#/clone_on_copy

packages/turbo-repository/rust/src/internal.rs
```diff
@@ -100,10 +127,19 @@ impl Workspace {
...
                 })?;
+
         let packages = package_json_paths
             .map(|path| {
+                let pkg_json = PackageJson::load(&path)?;
                 path.parent()
-                    .map(|package_path| Package::new(&self.workspace_state.root, package_path))
+                    .map(|package_path| {
+                        let package = Package::new(
+                            pkg_json.name.unwrap(),
```
chris-olszewski, line 137: We do not want to panic here if we find a `package.json` with no name. We shouldn't return a `package.json` without a name after [this change](https://github.com/vercel/turbo/pull/7280/files#diff-3fd939a75e56f948fbc34ce37a1ece6ef8854df4535db49183d871b14ae8ba41R286), but considering the fact that this has been in flux we should either:
 - Do the same thing and not return these package.json's without a name
 - Return an error

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -35,18 +42,47 @@ pub struct Workspace {
...
         }
     }
+
+    fn dependents(
+        &self,
+        graph: &PackageGraph,
+        workspace_path: &AbsoluteSystemPath,
+    ) -> Vec<Package> {
+        let node = WorkspaceNode::Workspace(WorkspaceName::Other(self.name.clone()));
+        let ancestors = graph.immediate_ancestors(&node).unwrap();
```
chris-olszewski, line 72: We should do our best to avoid panic even if it's a user error. If a user passes in a `graph` that doesn't contain this `Package` this will result in a panic.

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -35,18 +42,47 @@ pub struct Workspace {
...
+
+        ancestors
+            .iter()
+            .filter_map(|node| {
+                let info = graph.workspace_info(node.as_workspace()).unwrap();
```
chris-olszewski, line 77: I don't expect this will happen unless we have a pretty major bug in the package graph
```suggestion
                let info = graph.workspace_info(node.as_workspace())?;
```

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -63,4 +99,29 @@ impl Workspace {
...
+                let dep_names = deps
+                    .iter()
+                    .map(|p| p.relative_path.clone())
+                    .collect::<Vec<String>>();
+
+                (package.relative_path.clone(), dep_names)
+            })
+            .collect();
+
+        return Ok(map);
```
chris-olszewski, line 125:
```suggestion
       Ok(map)
```

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -63,4 +99,29 @@ impl Workspace {
     pub async fn find_packages(&self) -> std::result::Result<Vec<Package>, napi::Error> {
         self.packages_internal().await.map_err(|e| e.into())
     }
+
+    #[napi]
+    pub async fn package_graph(&self) -> Result<HashMap<String, Vec<String>>, Error> {
```
chris-olszewski, line 104: Nit with naming: I think we should make it clear that this method is returning a reversed package graph where each node gets mapped to all nodes that depend on it instead of all nodes that it depends on. If I called this for `A -> B -> C` and got `{ "A": [], "B": ["A"], "C": ["B"]}` I would be surprised.

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -63,4 +99,29 @@ impl Workspace {
...
+            .iter()
+            .map(|package| {
+                let deps = package.dependents(&self.graph, workspace_path);
+                let dep_names = deps
+                    .iter()
+                    .map(|p| p.relative_path.clone())
+                    .collect::<Vec<String>>();
+
+                (package.relative_path.clone(), dep_names)
+            })
```
chris-olszewski, line 122: We can avoid a few allocations by consuming the `Vec<Package>` instead of borrowing from them
```suggestion
            .into_iter()
            .map(|package| {
                let deps = package.dependents(&self.graph, workspace_path);
                let dep_names = deps
                    .into_iter()
                    .map(|p| p.relative_path)
                    .collect::<Vec<String>>();

                (package.relative_path, dep_names)
            })
```

chris-olszewski (changes requested):
The only major concern I have is with the use of `unwrap`. Since we'll be running in Node, I want to make sure we're good neighbors and give an error that's recoverable instead of force exiting the host Node JS process.

packages/turbo-repository/rust/src/lib.rs
mehulkar: This is a private function though, and the only user is coming from an internal implementation of find_packages() which should get a list of valid packages. I can be more defensive here, but does that ease the concern?

packages/turbo-repository/rust/src/internal.rs
mehulkar: The clone() is chained to `workspace_state`, not `package_manager`, if I'm reading the correctly. if I remove the clone, the compiler complains because workspace_state is also returned from this method. In any case (outside my diff, so I won't change this right now)

packages/turbo-repository/rust/src/internal.rs
mehulkar: updated

packages/turbo-repository/rust/src/lib.rs
mehulkar: renamed

chris-olszewski (approved):

packages/turbo-repository/rust/src/lib.rs
mehulkar: updated to return empty Vec in this case

mehulkar:
Merging, there were two failureS:

- Formatting of Cargo.toml: already fixed in #7308
- [turborepo-telemetry unit test, but only on macos](https://github.com/vercel/turbo/actions/runs/7821016888/job/21337130037?pr=7242), pretty sure it's unrelated

PR #7244, opened 2024-02-02
perf: parse and walk globs in parallel

chris-olszewski:
### Description

This PR brings parallelism to 2 aspects of our globwalk behavior:
 - parsing the input globs
 - walking of each of the input globs

This will primarily provide a speedup for tasks which have a large amount of input globs, but from quick spot checks I think this doesn't negatively impact tasks with small (2-3) inputs.

### Testing Instructions
Run lint against our examples (this is a noop, we just calculate task hash for this task) `turbo lint --filter=turborepo-examples`. 
Main:
Screenshot 2024-02-02 at 3 33 40 PM
Screenshot 2024-02-02 at 3 48 38 PM
This PR:
Screenshot 2024-02-02 at 3 33 53 PM
Screenshot 2024-02-02 at 3 48 52 PM

Closes TURBO-2234

### DIFF
--- a/crates/turborepo-globwalk/src/lib.rs
+++ b/crates/turborepo-globwalk/src/lib.rs
@@ -336,14 +336,11 @@ fn walk_glob(
     glob: Glob,
 ) -> Vec<Result<AbsoluteSystemPathBuf, WalkError>> {
     glob.walk(base_path_new)
-        .not(ex_patterns.clone())
+        .not(ex_patterns)
         .unwrap_or_else(|e| {
             // Per docs, only fails if exclusion list is too large, since we're using
             // pre-compiled globs
-            panic!(
-                "Failed to compile exclusion globs: {:?}: {}",
-                ex_patterns, e,
-            )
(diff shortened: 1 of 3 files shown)


crates/turborepo-globwalk/src/lib.rs
```diff
@@ -308,37 +309,38 @@ pub fn globwalk_internal(
...
-        .not(ex_patterns.clone())
+    glob.walk(base_path_new)
+        .not(ex_patterns)
         .unwrap_or_else(|e| {
             // Per docs, only fails if exclusion list is too large, since we're using
             // pre-compiled globs
-            panic!(
-                "Failed to compile exclusion globs: {:?}: {}",
-                ex_patterns, e,
-            )
```
chris-olszewski, line 341: I'm removing the exclusion patterns from the panic message as this should only happen if the list is too large. I don't think that information is useful enough to warrant adding an additional clone.

arlyon (approved):

PR #7246, opened 2024-02-03
Inputs clarity.

anthonyshew:
### Description

Changing around the order of what we already had written. It wasn't totally clear what we had written if you read it straight through.

Closes TURBO-2235

### DIFF
--- a/docs/pages/repo/docs/reference/configuration.mdx
+++ b/docs/pages/repo/docs/reference/configuration.mdx
@@ -364,16 +364,13 @@ Setting this to a list of globs will cause the task to only be run when files ma
 changed. This can be helpful if you want to, for example, skip running tests unless a source file changed, or
 explicitly depend on a `.gitignore`d file.
 
-Because specifying an `inputs` key immediately opts out of the default behavior, you may specify
-the special string `$TURBO_DEFAULT$` within an inputs array to include all default inputs. This is useful
-if you want to add or remove additional inputs from the default set.
-
-The default is `[]` to run the task when any file in the package changes.
-
 <Callout type="info">
-  `inputs` globs must be specified as relative paths rooted at the package's directory.
+  Good to know:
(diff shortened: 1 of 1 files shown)


docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -364,16 +364,13 @@ Setting this to a list of globs will cause the task to only be run when files ma
...
-if you want to add or remove additional inputs from the default set.
-
-The default is `[]` to run the task when any file in the package changes.
-
 <Callout type="info">
-  `inputs` globs must be specified as relative paths rooted at the package's directory.
+  Good to know:
+  - The default for `inputs` is `[]` to run the task when any file in the package changes.
+  - `inputs` globs must be specified as relative paths rooted at the package's directory.
+  - `turbo.json` is *always* considered an input. If you modify
```
tknickman, line 371: `package.json` is as well. Both of these files are always included even if you manually specify an inputs array and don't include it. 

Not required for this PR we can add that later!

tknickman (approved):

docs/pages/repo/docs/reference/configuration.mdx
```diff
@@ -364,16 +364,13 @@ Setting this to a list of globs will cause the task to only be run when files ma
...
-Because specifying an `inputs` key immediately opts out of the default behavior, you may specify
-the special string `$TURBO_DEFAULT$` within an inputs array to include all default inputs. This is useful
-if you want to add or remove additional inputs from the default set.
-
-The default is `[]` to run the task when any file in the package changes.
-
 <Callout type="info">
-  `inputs` globs must be specified as relative paths rooted at the package's directory.
+  Good to know:
+  - The default for `inputs` is `[]` to run the task when any file in the package changes.
```
mehulkar, line 369: This isn't correct is it? `inputs: []` means no inputs, so it's not the default when `inputs` is not specified?

PR #7277, opened 2024-02-05
add error enum to `turborepo-dirs`

Zertsov:
### Description
Add possibility for errors in `turborepo-dirs`.

### DIFF
--- a/crates/turborepo-dirs/Cargo.toml
+++ b/crates/turborepo-dirs/Cargo.toml
@@ -8,6 +8,7 @@ license = "MPL-2.0"
 
 [dependencies]
 dirs-next = "2.0.0"
+thiserror.workspace = true
 
 [lints]
 workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11610,6 +11610,7 @@ name = "turborepo-dirs"
 version = "0.1.0"
 dependencies = [
  "dirs-next",
+ "thiserror",
 ]
 
 [[package]]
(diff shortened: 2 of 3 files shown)


Zertsov:
* `main`
  * **#7277** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7277?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

crates/turborepo-dirs/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = "MPL-2.0"
 
 [dependencies]
 dirs-next = "2.0.0"
+thiserror.workspace = true
```
mehulkar, line 11: guessing yes, but is the same as below?

```suggestion
thiserror = { workspace = true }
```

mehulkar (approved):
LGTM, but it would probably make sense to use it somewhere at the same time as introducing it? In case it changes as you actually start using it. Not blocking at all :ship: away

crates/turborepo-dirs/Cargo.toml
Zertsov: Yeah they're the same, I think cargo defaults to this version if you don't specify things like `features`. I'm sure there's some configuration I can do somewhere to fix this but I haven't taken the time to deal with it yet  😁

Zertsov:
I plan on adding a PR down this stack that uses it, but didn't wanna break a bunch of stuff since this is a change I wanna make in a couple days or so

PR #7278, opened 2024-02-05
feat(telemetry): track global vs local turbo

tknickman:
### Description

Track global vs local turbo invocation.

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -75,6 +75,35 @@ impl Display for OutputLogsMode {
     }
 }
 
+/// Which binary was used for the current invocation
+pub enum BinaryType {
+    /// Local binary called directly
+    Local,
+    /// Local binary was called via a global binary
+    LocalDefer,
+    /// Global binary called directly (No local binary to handoff to)
+    Global,
+    /// Global binary called because repo inference failed
+    GlobalFallback,
+    /// Repo inference was explicitly skipped via `--skip-infer`` arg
+    SkipInfer,
+    /// Repo inference was explicitly skipped via `$TURBO_BINARY_PATH` env var
+    BinPath,
+}
+
+impl Display for BinaryType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(match self {
+            BinaryType::Local => "local",
+            BinaryType::LocalDefer => "local-defer",
+            BinaryType::Global => "global",
+            BinaryType::GlobalFallback => "global-fallback",
+            BinaryType::SkipInfer => "skip-infer",
+            BinaryType::BinPath => "bin-path",
+        })
+    }
+}
+
 #[derive(Copy, Clone, Debug, PartialEq, Serialize, ValueEnum)]
 pub enum LogOrder {
     #[serde(rename = "auto")]
@@ -947,6 +976,7 @@ pub async fn run(
     repo_state: Option<RepoState>,
     #[allow(unused_variables)] logger: &TurboSubscriber,
     ui: UI,
+    binary_type: BinaryType,
 ) -> Result<i32, Error> {
     let mut cli_args = Args::new();
     let version = get_version();
@@ -1044,6 +1074,7 @@ pub async fn run(
     root_telemetry.track_platform(TurboState::platform_name());
     root_telemetry.track_version(TurboState::version());
     root_telemetry.track_cpus(num_cpus::get());
+    root_telemetry.track_binary_type(binary_type.to_string());
     // track args
     cli_args.track(&root_telemetry);
 
--- a/crates/turborepo-lib/src/shim.rs
+++ b/crates/turborepo-lib/src/shim.rs
@@ -585,12 +585,25 @@ fn run_correct_turbo(
         try_check_for_updates(&shim_args, &turbo_state.version);
 
         if turbo_state.local_is_self() {
+            // `spawn_local_turbo`` sets the env var, so we can check for it here as an
+            // indication of whether we've been spawned from a global binary
+            let spawned_from_global = env::var(cli::INVOCATION_DIR_ENV_VAR).is_ok();
+
             env::set_var(
                 cli::INVOCATION_DIR_ENV_VAR,
                 shim_args.invocation_dir.as_path(),
             );
             debug!("Currently running turbo is local turbo.");
-            Ok(cli::run(Some(repo_state), subscriber, ui)?)
+            Ok(cli::run(
+                Some(repo_state),
+                subscriber,
+                ui,
+                if spawned_from_global {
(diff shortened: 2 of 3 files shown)


crates/turborepo-lib/src/shim.rs
```diff
@@ -585,12 +585,25 @@ fn run_correct_turbo(
         try_check_for_updates(&shim_args, &turbo_state.version);
 
         if turbo_state.local_is_self() {
+            // `spawn_local_turbo`` sets the env var, so we can check for it here as an
+            // indication of whether we've been spawned from a global binary
+            let spawned_from_global = env::var(cli::INVOCATION_DIR_ENV_VAR).is_ok();
```
tknickman, line 590: cc @chris-olszewski on this one. This looks to be the only way to detect this right now, but happy to change to something better if it exists!

crates/turborepo-lib/src/shim.rs
```diff
@@ -585,12 +585,25 @@ fn run_correct_turbo(
...
-            Ok(cli::run(Some(repo_state), subscriber, ui)?)
+            Ok(cli::run(
+                Some(repo_state),
+                subscriber,
+                ui,
+                if spawned_from_global {
+                    cli::BinaryType::LocalDefer
+                } else {
+                    cli::BinaryType::Local
+                },
```
tknickman, line 605: Just give me a ternary rust. Come on.

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -75,6 +75,35 @@ impl Display for OutputLogsMode {
...
+    LocalDefer,
+    /// Global binary called directly (No local binary to handoff to)
+    Global,
+    /// Global binary called because repo inference failed
+    GlobalFallback,
+    /// Repo inference was explicitly skipped via `--skip-infer`` arg
+    SkipInfer,
+    /// Repo inference was explicitly skipped via `$TURBO_BINARY_PATH` env var
+    BinPath,
+}
```
mehulkar, line 92: Few questions:

- Global just means anything that's not locally deferred right? E.g. `/path/to/turbo` is global the same as `npm i -g turbo && turbo`?

- What would be the binary type of `npx turbo` and is it possible to track that separately?

- I wonder if we could model this as two fields: `bin_location: local|global|explicit` and `bin_inference: defer|failed|skipped`?  It's not quite the same data (e.g. `BinPath` translates to `bin_type:explicit, bin_inference:skipped`, which can be confusing since it's not using `--skip-infer`)

mehulkar (commented):
Couple questions and suggestions, but not really blocking. I can hit approve unless you think the suggestion is worth a chat!

crates/turborepo-lib/src/cli/mod.rs
tknickman: Good questions on the first two, I can test it out. 

And yea I had originally modeled this as a tuple much how you described it. But since the cross product of both location and inference aren't always valid (although I like the explicit) I just made it a simple enum array.

mehulkar: sounds good to me! I think the `npx|pnpx|yarn dlx` case would especially be useful to know about, but it's not a blocker. We can merge this and see which pieces of data we want to dig deeper into

mehulkar (approved):

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -75,6 +75,35 @@ impl Display for OutputLogsMode {
...
+    /// Local binary called directly
+    Local,
+    /// Local binary was called via a global binary
+    LocalDefer,
+    /// Global binary called directly (No local binary to handoff to)
+    Global,
+    /// Global binary called because repo inference failed
+    GlobalFallback,
+    /// Repo inference was explicitly skipped via `--skip-infer`` arg
+    SkipInfer,
```
chris-olszewski, line 89: I believe this is also capturing the case of `Local` when we detect that the local version supports `--skip-infer`. 

If we end up realizing in global turbo that we need to pass of to a local 1.7+, then we'll add a `--skip-infer` arg to the local call. [source](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/shim.rs#L640)

We check if that flag is set before we go into `run_correct_turbo` and [jump right in](https://github.com/vercel/turbo/blob/main/crates/turborepo-lib/src/shim.rs#L752).

I don't think `Local` will ever get populated since we didn't have telem in <1.7.

crates/turborepo-lib/src/shim.rs
chris-olszewski: I don't believe this updated codepath ever gets hit as it should only happen if the turbo binary doesn't support `--skip-infer`.

crates/turborepo-lib/src/shim.rs
```diff
@@ -750,7 +769,7 @@ pub fn run() -> Result<i32, Error> {
     // global turbo having handled the inference. We can run without any
     // concerns.
     if args.skip_infer {
-        return Ok(cli::run(None, &subscriber, ui)?);
+        return Ok(cli::run(None, &subscriber, ui, cli::BinaryType::SkipInfer)?);
```
chris-olszewski, line 772: I think this is where we want to check the env var and decide between `SkipInfer` and `LocalDefer`

chris-olszewski (commented):
Shim code is always mind bending, but I think this will be putting a lot of what we expect to be in `LocalDefer` in the `SkipInfer` bucket since that's how we signal from global to local to stop looking for another local.

Did a quick println debugging:
```
diff --git a/crates/turborepo-lib/src/shim.rs b/crates/turborepo-lib/src/shim.rs
index 1b69aac0fe..78459d5f07 100644
--- a/crates/turborepo-lib/src/shim.rs
+++ b/crates/turborepo-lib/src/shim.rs
@@ -581,6 +581,7 @@ fn run_correct_turbo(
     subscriber: &TurboSubscriber,
     ui: UI,
 ) -> Result<i32, Error> {
+    println!("starting run_correct_turbo");
     if let Some(turbo_state) = LocalTurboState::infer(&repo_state.root) {
         try_check_for_updates(&shim_args, &turbo_state.version);
 
@@ -655,6 +656,7 @@ fn spawn_local_turbo(
         .map_err(|_| Error::RepoRootPath(repo_state.root.clone()))?;
 
     let mut raw_args: Vec<_> = if supports_skip_infer_and_single_package {
+        println!("invoking local with skip infer");
         vec!["--skip-infer".to_string()]
     } else {
         Vec::new()
```
`turbo_dev` is just an alias for a "global" dev build
```
[0 olszewski@chriss-mbp] /tmp/vite-test $ turbo_dev bin        
starting run_correct_turbo
invoking local with skip infer
/private/tmp/vite-test/node_modules/.pnpm/turbo-darwin-arm64@1.12.3/node_modules/turbo-darwin-arm64/bin/turbo
```

Issue #7227, opened 2024-02-02
Nested workspace ignored after upgrading to 1.12.2 from 1.11.3


edgarrroussille:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

https://github.com/edgarrroussille/turbo-1.12.2-ignored-workspaces/tree/main

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

turbo 1.12.3-canary.0

### TLDR:
Using nested packages worked with turbo@1.11.3
Using same directory structure with turbo@1.12.0 doesn't work anymore

Consider this directory structure:
```
- apps
  - apps-a
    - package.json
    - apps-b
      - package.json
```

with this`pnpm-workspace.yaml`:
```
packages:
  - apps/**
  - packages/*
```

Running `pnpm turbo build` with turbo@1.11.3 would build apps-b 
Runnin same command with turbo@1.12.0 and later would not build apps-b

### Expected Behavior

Searching through past issues, #942 states that nested directory wasn't actually supposed to be supported... but it worked since we migrated to turborepo (summer '23) and from our point of view it's definitely a regression :(

As this folder structure worked in earlier versions, I'd love a way to keep it working as is... 

### To Reproduce

1. Clone [repo](https://github.com/edgarrroussille/turbo-1.12.2-ignored-workspaces/tree/main)
2. `cd 1.11.3 && pnpm i && pnpm turbo build`
3. See all 5 packages are built
4 `cd ../1.12.0 && pnpm i && pnpm turbo build`
5. See only 4 packages are built: apps-b is missing

### Additional context

Not an expert, but it's most probably coming from #7025

Klaitos:
Same issue on our side, we have a package inside a package that is ignored now, the turbo prune command fails on 1.12. We are stuck to 1.11 from now until a package architecture refactor.

chris-olszewski:
@edgarrroussille Can you update your example so that the `1.11.3` directory contains a nested workspace?

edgarrroussille:
> @edgarrroussille Can you update your example so that the `1.11.3` directory contains a nested workspace?

Sorry about that, just updated the repo :)

gsoltis:
We're looking into this. While nested packages aren't technically supported, we did previously find them and allow scripts to run. I suspect, but haven't confirmed, that caching won't work very well in a nested scenario.

[...]

PR #7280, opened 2024-02-06
Fix/filter unnamed packages

arlyon:
### Description

Closes https://linear.app/vercel/issue/TURBO-2255/filter-nameless-packages and https://linear.app/vercel/issue/TURBO-2256/allow-recursion-past-a-packagejson-during-package-discovery and makes our implementation more closely match the old go behaviour (and close https://github.com/vercel/turbo/issues/7227 in the process)

### Testing Instructions

Integration test

Closes TURBO-2257


chris-olszewski (approved):
LFG

PR #7283, opened 2024-02-06
fix(Turborepo): Start moving cookie watching to downstream services

gsoltis:
### Description

Filewatching cookies were ensuring that the filewatching task was up to date, but not downstream, filewatching-backed services. This change moves the filesystem cookie synchronization into a downstream service (glob watching) to ensure that RPCs are serviced by a task that has seen at least all of the operations that a client has done.

The process is as follows:
 1. A client of a filewatching-backed service uses a `CookieWriter` to stamp its request with a serial. 
 2. The CookieWriter requests a cookie file be written, and the serial returned to the client
 3. The client forwards the request, now with a serial, to the filewatching-backed service
 4. The service observes cookie files being created, and keeps track of the highest serial it has seen via a `CookieWatcher`. When a cookie file is observed, some pending requests may be returned as ready for servicing.
 5. When the service gets a request, it checks via `CookieWatcher` if it has already seen the serial. If so, the request is handled immediately
 6. If not, the service, again via the `CookieWatcher`, saves the request until the correct serial has been observed (step 4).

This means that timeouts need to be handled by clients, but also gives clients the freedom to set their own timeouts.

### Testing Instructions

Most cookie watching tests are gone, as is most of the logic. Instead, the mechanism is exercised via tests of downstream services.

Closes TURBO-2244

### DIFF
--- a/crates/turborepo-filewatch/src/cookie_jar.rs
+++ b/crates/turborepo-filewatch/src/cookie_jar.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashMap,
+    collections::{HashMap, VecDeque},
     fs::OpenOptions,
     path::PathBuf,
     sync::{atomic::AtomicUsize, Arc, Mutex},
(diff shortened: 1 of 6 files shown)


NicholasLYang (approved):
The logic makes sense to me. Just out of curiosity, what's the reference for the cookie terminology? Whenever I search it I just get browser references

gsoltis:
I took the "Cookie" term from [Watchman](https://github.com/facebook/watchman/blob/a20f90ff160452242fda97a9faa27ca0b4d9bf69/watchman/CookieSync.h#L19)

PR #7284, opened 2024-02-06
fix(turbopack): Make node js use our sourcemap corectly

kdy1:
### Description

I need to make a repro for a question on node.js issue tracker.

### DIFF
--- a/crates/turbopack-core/src/source_map/mod.rs
+++ b/crates/turbopack-core/src/source_map/mod.rs
@@ -643,7 +643,15 @@ impl<'de> Deserialize<'de> for CrateMapWrapper {
         use serde::de::Error;
         let bytes = <&[u8]>::deserialize(deserializer)?;
         let map = DecodedMap::from_reader(bytes).map_err(Error::custom)?;
-        Ok(CrateMapWrapper(map))
+
+        match map {
+            DecodedMap::Regular(..) => Ok(CrateMapWrapper(map)),
+            DecodedMap::Index(map) => {
+                let map = map.flatten().map_err(Error::custom)?;
+                Ok(CrateMapWrapper(DecodedMap::Regular(map)))
+            }
+            DecodedMap::Hermes(_) => todo!("hermes source maps are not implemented"),
+        }
     }
 }
 
--- a/crates/turbopack-core/src/source_map/mod.rs
+++ b/crates/turbopack-core/src/source_map/mod.rs
@@ -218,9 +218,9 @@ impl SourceMap {
                         )
                     })
                     .collect::<Vec<sourcemap::SourceMapSection>>();
-                Arc::new(CrateMapWrapper(DecodedMap::Index(SourceMapIndex::new(
-                    None, sections,
-                ))))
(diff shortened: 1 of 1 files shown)


kdy1:
I filed https://github.com/orgs/nodejs/discussions/51675

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -620,7 +600,20 @@ impl Serialize for CrateMapWrapper {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         use serde::ser::Error;
         let mut bytes = vec![];
-        self.0.to_writer(&mut bytes).map_err(Error::custom)?;
+        match &self.0 {
```
sokra, line 603: Why do we want to change the serialization logic?

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -630,7 +623,15 @@ impl<'de> Deserialize<'de> for CrateMapWrapper {
         use serde::de::Error;
         let bytes = <&[u8]>::deserialize(deserializer)?;
         let map = DecodedMap::from_reader(bytes).map_err(Error::custom)?;
-        Ok(CrateMapWrapper(map))
+
```
sokra, line 626: Why do we want to change the deserialization logic?

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -218,9 +218,9 @@ impl SourceMap {
                         )
                     })
                     .collect::<Vec<sourcemap::SourceMapSection>>();
-                Arc::new(CrateMapWrapper(DecodedMap::Index(SourceMapIndex::new(
-                    None, sections,
-                ))))
+                let map = SourceMapIndex::new(None, sections);
+                let map = map.flatten()?;
```
sokra, line 222: I don't think we should flatten the source map in general.

Browsers support indexed source maps so we can use them here.
Our own lookup supports indexed source maps.

Only node.js doesn't support it, so we should only change the emitted SourceMaps to be flattened.

crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -257,44 +257,24 @@ impl SourceMap {
...
+                        Some(map.0.clone()),
+                    ));
                 }
 
-                rope += "]
-}";
-
-                rope.build()
+                let map = SourceMapIndex::new(None, sections);
+                let map = CrateMapWrapper(DecodedMap::Regular(map.flatten()?));
```
sokra, line 274: See above

crates/turbopack-core/src/source_map/mod.rs
kdy1: I did it while debugging, because I wasn't sure the code used for source map generation

kdy1:
Closing as this PR is now redundant. The source map issue is fixed on the main branch, and we don't need to generate non-sectioned source map anymore.

Issue #6805, opened 2023-12-15
Build command fails on create-turbo Yarn 4 project


MikelArnaiz:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

https://github.com/MikelArnaiz/turborepo-build-error-on-fresh-project

### What package manager are you using / does the bug impact?

Yarn v2/v3 (node_modules linker only)

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

1.11.3-canary.0

### Describe the Bug

`yarn run build` fails on newly created project with `yarn dlx create-turbo@latest` (also `@canary`).

```
yarn build     
...
web:build: > Build error occurred
web:build: Error: Jest worker encountered 1 child process exceptions, exceeding retry limit
web:build:     at ChildProcessWorker.initialize (/Users/mikelarnaiz/projects/turbo/build-error-canary/.yarn/__virtual__/next-virtual-5514c12946/4/.yarn/berry/cache/next-npm-14.0.4-93c7e4ca0b-10c0.zip/node_modules/next/dist/compiled/jest-worker/index.js:1:11580)
web:build:     at ChildProcessWorker._onExit (/Users/mikelarnaiz/projects/turbo/build-error-canary/.yarn/__virtual__/next-virtual-5514c12946/4/.yarn/berry/cache/next-npm-14.0.4-93c7e4ca0b-10c0.zip/node_modules/next/dist/compiled/jest-worker/index.js:1:12545)
web:build:     at ChildProcess.emit (node:events:514:28)
web:build:     at ChildProcess._handle.onexit (node:internal/child_process:291:12) {
web:build:   type: 'WorkerError'
web:build: }
web:build: ERROR: command finished with error: command (/Users/mikelarnaiz/projects/turbo/build-error-canary/apps/web) /private/var/folders/1h/21ml03md75lcgwg74lj5nfmm0000gn/T/xfs-5c153812/yarn run build exited (1)
web#build: command (/Users/mikelarnaiz/projects/turbo/build-error-canary/apps/web) /private/var/folders/1h/21ml03md75lcgwg74lj5nfmm0000gn/T/xfs-5c153812/yarn run build exited (1)

 Tasks:    0 successful, 2 total
Cached:    0 cached, 2 total
  Time:    7.213s 
Failed:    web#build
```

### Expected Behavior

It should build. `yarn run dev` works as expected.

### To Reproduce

```
nvm use 18
cd ~
[...]

PR #7287, opened 2024-02-06
chore: Add Yarn 4 to Github Issue template

mehulkar:
We are receiving bug reports for Yarn for and they are incorrectly categorized

ref: https://github.com/vercel/turbo/issues/6805

Closes TURBO-2262

### DIFF
--- a/.github/ISSUE_TEMPLATE/0-turborepo-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/0-turborepo-bug-report.yml
@@ -38,8 +38,7 @@ body:
         - npm
         - pnpm
         - Yarn v1
-        - Yarn v2/v3 (node_modules linker only)
-        - Yarn v4
+        - Yarn v2/v3/v4 (node_modules linker only)
     validations:
       required: true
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/0-turborepo-bug-report.yml
```diff
@@ -39,6 +39,7 @@ body:
         - pnpm
         - Yarn v1
         - Yarn v2/v3 (node_modules linker only)
+        - Yarn v4
```
tknickman, line 42: I think we should bucket this with the previous line since the linker mode still matters (v4 still uses pnp by default)

tknickman (changes requested):

tknickman (approved):

.github/ISSUE_TEMPLATE/0-turborepo-bug-report.yml
mehulkar: oh wow, that is cool, TIL! Seems better in theory than node_modules linker

PR #7290, opened 2024-02-06
fix(Turborepo): Flag off daemon for package discovery

gsoltis:
### Description

Daemon package watching has some bugs that need to be ironed out. Flag it off for now.

### Testing Instructions

Disabled integration test that explicitly checks for the behavior that we are flagging off.

Closes TURBO-2265


crates/turborepo-lib/src/run/mod.rs
```diff
@@ -289,16 +290,32 @@ impl Run {
...
+
+            #[cfg(not(feature = "daemon-package-discovery"))]
+            let builder = builder.with_package_discovery(
+                LocalPackageDiscoveryBuilder::new(
+                    self.base.repo_root.clone(),
+                    None,
+                    Some(root_package_json.clone()),
+                )
+                .build()?,
+            );
```
chris-olszewski, line 315: I think this is unnecessary as this is what package discovery gets chosen by default if one isn't provided for `PackageGraphBuilder`
```suggestion
```

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -273,6 +273,7 @@ impl Run {
         };
 
         // if we are forcing the daemon, we don't want to fallback to local discovery
+        //#[cfg(feature = "daemon-package-discovery")]
         let (fallback, duration) = if let Some(true) = self.opts.run_opts.daemon {
```
chris-olszewski, line 277: If we're constructing `fallback`, we should make sure to use it. The call `LocalPackageDiscoveryBuilder.build()` can result in a `yarn --version` or `pnpm --version` if the user hasn't specified a `packageManager` in `package.json`.

gsoltis: Ah, ok. I was trying to decide how much I needed to cut. Seems like cut all of the fallback stuff. I'll see what I can do.

chris-olszewski (approved):
Some unused imports, but those can be fix-shipped or fixed in follow up PR.

gsoltis:
Will fix imports in next PR

gsoltis:
Fixing imports: https://github.com/vercel/turbo/pull/7295

PR #7293, opened 2024-02-06
feat(async_cache): support cloning

tknickman:
### Description

Remove the `writer_thread` reference in `AsyncCache` so we can easily clone it. 

Precursor to https://github.com/vercel/turbo/pull/7074/files

Closes TURBO-2271

### DIFF
--- a/crates/turborepo-cache/src/async_cache.rs
+++ b/crates/turborepo-cache/src/async_cache.rs
@@ -1,10 +1,7 @@
 use std::sync::{atomic::AtomicU8, Arc};
 
 use futures::{stream::FuturesUnordered, StreamExt};
-use tokio::{
-    sync::{mpsc, Semaphore},
-    task::JoinHandle,
-};
+use tokio::sync::{mpsc, Semaphore};
 use tracing::{warn, Instrument, Level};
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf, AnchoredSystemPathBuf};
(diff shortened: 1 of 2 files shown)


crates/turborepo-cache/src/async_cache.rs
```diff
@@ -163,11 +159,6 @@ impl AsyncCache {
         // Wait until flush callback is finished
         rx.await.ok();
     }
-
-    pub async fn shutdown(self) {
-        let Self { writer_thread, .. } = self;
-        writer_thread.await.unwrap();
-    }
```
tknickman, line 170: This was unused

NicholasLYang: Hmm that feels wrong. If we don't await on the writer thread, there's a chance that we could exit before we finish adding stuff to the cache

tknickman: Yea it does - I can return this and track it outside of the async_cache struct to make sure we close it down.

chris-olszewski: `wait` ensures that all workers have finished writing to the underlying cache. We could add a new `WorkerRequest::Shutdown` that does the same, but drops the receiving end so any future writes will fail.

chris-olszewski (approved):
Stamping as this keeps the behavior. We should probably provide a "shutdown" functionality that gets used, but that belongs in another PR

PR #7294, opened 2024-02-06
feat(@turbo/repository): add dependents to package graph

mehulkar:
Closes TURBO-2272


mehulkar:
Merging this into #7242 since I got it working

PR #7303, opened 2024-02-07
improve HMR performance

sokra:
### Description

change make_chunks into a turbo_tasks function to isloate graph walk from chunk item size changes

Before any file change caused module graph walk and chunking to be reexecuted. After this change module graph walk is only reexecuted when module references change.

### Testing Instructions

Closes PACK-2431

### DIFF
--- a/crates/turbopack-core/src/chunk/chunking.rs
+++ b/crates/turbopack-core/src/chunk/chunking.rs
@@ -12,29 +12,59 @@ use regex::Regex;
 use tracing::Level;
 use turbo_tasks::{ReadRef, TryJoinIterExt, ValueToString, Vc};
 
-use super::{AsyncModuleInfo, Chunk, ChunkItem, ChunkType, ChunkingContext};
+use super::{
+    AsyncModuleInfo, Chunk, ChunkItem, ChunkItemsWithAsyncModuleInfo, ChunkType, ChunkingContext,
+    Chunks,
+};
 use crate::output::OutputAssets;
+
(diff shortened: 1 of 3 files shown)


crates/turbopack-core/src/chunk/chunking.rs
```diff
@@ -12,29 +12,59 @@ use regex::Regex;
...
 
+#[turbo_tasks::value]
+struct ChunkItemInfo {
+    ty: Vc<Box<dyn ChunkType>>,
+    name: Vc<String>,
+    size: usize,
+}
+
+#[turbo_tasks::function]
+async fn chunk_item_info(
```
sokra, line 29: Cache more work in make_chunks

crates/turbopack-core/src/chunk/chunking.rs
```diff
@@ -12,29 +12,59 @@ use regex::Regex;
...
+        name: asset_ident.resolve().await?,
+    }
+    .cell())
+}
+
 /// Creates chunks based on heuristics for the passed `chunk_items`. Also
 /// attaches `referenced_output_assets` to the first chunk.
-#[tracing::instrument(level = Level::TRACE, skip_all)]
+#[turbo_tasks::function]
 pub async fn make_chunks(
```
sokra, line 48: It's a turbo-tasks function now, so changes to chunk_item_size no longer invalidate the calling function

ForsakenHarmony (approved):

PR #7304, opened 2024-02-07
feat(cache): bump default timeout and add docs

tknickman:
### Description

The Go timeout was being set to `0`, which meant there was _no_ timeout. We don't want to replicate that exactly, some timeout is good. This just increases the timeout a bit to reduce issues with large artifacts. 

Closes TURBO-2277

### DIFF
--- a/docs/pages/repo/docs/reference/command-line-reference/run.mdx
+++ b/docs/pages/repo/docs/reference/command-line-reference/run.mdx
@@ -279,6 +279,14 @@ turbo run build --no-cache
 turbo run dev --no-cache
 ```
 
+### `--remote-cache-timeout`
+
+Default `30` seconds. Set the timeout for remote cache operations in seconds.
+
+```shell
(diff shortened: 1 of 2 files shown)


docs/pages/repo/docs/reference/command-line-reference/run.mdx
```diff
@@ -279,6 +279,14 @@ turbo run build --no-cache
 turbo run dev --no-cache
 ```
 
+### `--remote-cache-timeout`
```
mehulkar, line 282: not blocking, but I think these were roughly alphabetized, if you want to move it a bit further down

mehulkar (approved):

docs/pages/repo/docs/reference/command-line-reference/run.mdx
tknickman: Ah nice, didn't know that. Done!

PR #7309, opened 2024-02-07
chore: add shutdown method for async cache

chris-olszewski:
### Description

As per discussion in #7293 we should probably have a shutdown method. This PR adds a method that will prevent any future writes from being queued.

### Testing Instructions

Added some shutdown tests to make sure that trying to use the cache post-shutdown will error. These probably should belong in their own testing methods, but considering the amount of work required to construct an AsyncCache I think this is acceptable.

Closes TURBO-2282


tknickman (approved):
Thanks for tackling, this seems sound to me 👍🏼

PR #7311, opened 2024-02-07
chore(ci): Update to actions/checkout@v4

mehulkar:
v3 uses node 16 which is EOL. The latest update only has
that one breaking change. Since we do not know or care about this
node version in our scripts, we can safely update this.

https://github.com/actions/checkout

Closes TURBO-2284


mehulkar:
Closing in favor of https://github.com/vercel/turbo/pull/7224, @ForsakenHarmony has a much more complete approach to this than I did here.

PR #7316, opened 2024-02-08
fix: unset RUSTC_WRAPPER when empty

chris-olszewski:
### Description

It seems that setting `RUSTC_WRAPPER` to an empty string results in [weird failures](https://github.com/vercel/turbo/actions/runs/7823191751/job/21343720629?pr=7315#step:6:461)

### Testing Instructions

This is an external PR so as long as this PR can successfully build `turbo` this fix should work. 

Closes TURBO-2290

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -223,8 +223,13 @@ jobs:
         with:
           windows: ${{ matrix.os.name == 'windows' }}
           github-token: "${{ secrets.GITHUB_TOKEN }}"
-
-      - run: turbo run build --filter=cli --color --env-mode=strict --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
+      # We explicitly unset RUSTC_WRAPPER if it is an empty string as causes build issues
+      - run: |
+          if [ -z "${RUSTC_WRAPPER}" ]; then
+            unset RUSTC_WRAPPER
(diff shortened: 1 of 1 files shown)


tknickman (approved):

.github/workflows/test.yml
```diff
@@ -223,8 +223,13 @@ jobs:
         with:
           windows: ${{ matrix.os.name == 'windows' }}
           github-token: "${{ secrets.GITHUB_TOKEN }}"
-
-      - run: turbo run build --filter=cli --color --env-mode=strict --token=${{ secrets.TURBO_TOKEN }} --team=${{ vars.TURBO_TEAM }}
+      # We explicitly unset RUSTC_WRAPPER if it is an empty string as causes build issues
+      - run: |
+          if [ -z "${RUSTC_WRAPPER}" ]; then
+            unset RUSTC_WRAPPER
```
mehulkar, line 229: Who's setting it in the first place?

mehulkar (approved):

.github/workflows/test.yml
chris-olszewski: We are [right here](https://github.com/vercel/turbo/pull/7316/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R237) when we aren't using sccache

PR #7319, opened 2024-02-08
feat(gitignore): break out gitignore util

tknickman:
### Description

We need to make sure .turbo is ignored for task access. This pulls the existing util out of link, and re-uses it. 

Also - tests!

### DIFF
--- a/crates/turborepo-lib/src/gitignore.rs
+++ b/crates/turborepo-lib/src/gitignore.rs
@@ -1,0 +1,141 @@
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+use std::{
+    fs::{File, OpenOptions},
+    io,
+    io::{BufRead, Write},
+};
+
+use turbopath::AbsoluteSystemPathBuf;
+
+const TURBO_GITIGNORE_ENTRY: &str = ".turbo";
+const GITIGNORE_FILE: &str = ".gitignore";
+
+pub fn ensure_turbo_is_gitignored(repo_root: &AbsoluteSystemPathBuf) -> Result<(), io::Error> {
+    let gitignore_path = repo_root.join_component(GITIGNORE_FILE);
+
+    if !gitignore_path.exists() {
+        let mut gitignore = File::create(gitignore_path)?;
+        #[cfg(unix)]
+        gitignore.metadata()?.permissions().set_mode(0o0644);
+        writeln!(gitignore, "{}", TURBO_GITIGNORE_ENTRY)?;
+    } else {
+        let gitignore = File::open(&gitignore_path)?;
+        let mut lines = io::BufReader::new(gitignore).lines();
+        let has_turbo =
+            lines.any(|line| line.map_or(false, |line| line.trim() == TURBO_GITIGNORE_ENTRY));
+        if !has_turbo {
+            let mut gitignore = OpenOptions::new()
+                .read(true)
+                .append(true)
+                .open(&gitignore_path)?;
+
+            // write with a preceding newline just in case the .gitignore file doesn't end
+            // with a newline
+            writeln!(gitignore, "\n{}", TURBO_GITIGNORE_ENTRY)?;
+        }
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use anyhow::Result;
+    use tempfile::tempdir;
+
+    use super::*;
+
+    fn has_turbo_gitignore_entry(gitignore_path: &AbsoluteSystemPathBuf) -> bool {
+        let gitignore = File::open(gitignore_path).expect("Failed to open .gitignore file");
+        let mut lines = io::BufReader::new(gitignore).lines();
+        lines.any(|line| line.map_or(false, |line| line.trim() == TURBO_GITIGNORE_ENTRY))
+    }
+
+    #[test]
+    fn test_no_gitignore() -> Result<()> {
+        let repo_root_tmp = tempdir()?;
+        let repo_root = AbsoluteSystemPathBuf::try_from(repo_root_tmp.path())?;
+
+        ensure_turbo_is_gitignored(&repo_root).expect("Failed to ensure turbo is gitignored");
+
+        // Verify that the .gitignore file exists and contains the expected entry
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/gitignore.rs
```diff
@@ -0,0 +1,141 @@
...
+    io,
+    io::{BufRead, Write},
+};
+
+use turbopath::AbsoluteSystemPathBuf;
+
+const TURBO_GITIGNORE_ENTRY: &str = ".turbo";
+const GITIGNORE_FILE: &str = ".gitignore";
+
+pub fn ensure_turbo_is_gitignored(repo_root: &AbsoluteSystemPathBuf) -> Result<(), io::Error> {
```
chris-olszewski, line 14: Complete nit, but we should take a reference to a borrowed varaint of the path instead of the owned one.
See [clippy lint](https://rust-lang.github.io/rust-clippy/master/#/ptr_arg)
```suggestion
pub fn ensure_turbo_is_gitignored(repo_root: &AbsoluteSystemPath) -> Result<(), io::Error> {
```

crates/turborepo-lib/src/gitignore.rs
```diff
@@ -0,0 +1,141 @@
...
+    let gitignore_path = repo_root.join_component(GITIGNORE_FILE);
+
+    if !gitignore_path.exists() {
```
chris-olszewski, line 17: We should generally prefer `try_exists` over `exists` as the latter will report `false` if we can't verify the path exists due to permission issues. [docs](https://doc.rust-lang.org/std/fs/fn.try_exists.html)
```suggestion
    if !gitignore_path.try_exists()? {
```

crates/turborepo-lib/src/gitignore.rs
```diff
@@ -0,0 +1,141 @@
...
+        let mut gitignore = File::create(gitignore_path)?;
```
chris-olszewski, line 18: We should prefer the operations on our path library over those from the stdlib. Primarily because we use [fs_err](https://docs.rs/fs-err/latest/fs_err/) which does us the huge favor of including paths in the error message.
```suggestion
        let mut gitignore = gitignore_path.create()?;
```

crates/turborepo-lib/src/gitignore.rs
```diff
@@ -0,0 +1,141 @@
...
+        #[cfg(unix)]
+        gitignore.metadata()?.permissions().set_mode(0o0644);
+        writeln!(gitignore, "{}", TURBO_GITIGNORE_ENTRY)?;
```
chris-olszewski, line 21: We have [create_with_contents](https://github.com/vercel/turbo/blob/main/crates/turborepo-paths/src/absolute_system_path.rs#L177) that can combine the file creation and writing.

crates/turborepo-lib/src/gitignore.rs
```diff
@@ -0,0 +1,141 @@
...
+const TURBO_GITIGNORE_ENTRY: &str = ".turbo";
+const GITIGNORE_FILE: &str = ".gitignore";
+
+pub fn ensure_turbo_is_gitignored(repo_root: &AbsoluteSystemPathBuf) -> Result<(), io::Error> {
+    let gitignore_path = repo_root.join_component(GITIGNORE_FILE);
+
+    if !gitignore_path.exists() {
+        let mut gitignore = File::create(gitignore_path)?;
+        #[cfg(unix)]
+        gitignore.metadata()?.permissions().set_mode(0o0644);
```
chris-olszewski, line 20: We could use our [`set_mode`](https://github.com/vercel/turbo/blob/main/crates/turborepo-paths/src/absolute_system_path.rs#L454) wrapper here

crates/turborepo-lib/src/gitignore.rs
```diff
@@ -0,0 +1,141 @@
...
+    } else {
+        let gitignore = File::open(&gitignore_path)?;
+        let mut lines = io::BufReader::new(gitignore).lines();
+        let has_turbo =
+            lines.any(|line| line.map_or(false, |line| line.trim() == TURBO_GITIGNORE_ENTRY));
+        if !has_turbo {
+            let mut gitignore = OpenOptions::new()
+                .read(true)
+                .append(true)
+                .open(&gitignore_path)?;
```
chris-olszewski, line 31: We have a wrapper for this with [`open_with_options`](https://github.com/vercel/turbo/blob/main/crates/turborepo-paths/src/absolute_system_path.rs#L422)

crates/turborepo-lib/src/gitignore.rs
```diff
@@ -0,0 +1,141 @@
...
+mod tests {
+    use anyhow::Result;
+    use tempfile::tempdir;
+
+    use super::*;
+
+    fn has_turbo_gitignore_entry(gitignore_path: &AbsoluteSystemPathBuf) -> bool {
+        let gitignore = File::open(gitignore_path).expect("Failed to open .gitignore file");
+        let mut lines = io::BufReader::new(gitignore).lines();
+        lines.any(|line| line.map_or(false, |line| line.trim() == TURBO_GITIGNORE_ENTRY))
```
chris-olszewski, line 52: I see this line duped between this function and the actual implementation. It might be worth pulling out into a helper function to make sure they don't drift.

crates/turborepo-lib/src/gitignore.rs
```diff
@@ -0,0 +1,141 @@
...
+        writeln!(gitignore, ".turbo")?;
+        writeln!(gitignore, ".node_modules/")?;
+
+        ensure_turbo_is_gitignored(&repo_root).expect("Failed to ensure turbo is gitignored");
+
+        // Verify that the .gitignore file exists and contains the expected entry
+        let gitignore_path = repo_root.join_component(GITIGNORE_FILE);
+        assert!(gitignore_path.exists());
+
+        assert!(has_turbo_gitignore_entry(&gitignore_path));
```
chris-olszewski, line 113: I think we should add a check to make sure the file still just has the 2 lines.

chris-olszewski (approved):
LGTM

All of the comments are just Rust pointers or pointing you towards our existing path library functions that we should generally prefer over the stdlib ones.

PR #7320, opened 2024-02-08
feat(turbo-tasks-fetch): allow specifying a proxy

kwonoj:
### Description

next/font supports to setup proxy.

Closes PACK-2449

### DIFF
--- a/crates/turbo-tasks-fetch/src/lib.rs
+++ b/crates/turbo-tasks-fetch/src/lib.rs
@@ -36,11 +36,34 @@ impl HttpResponseBody {
     }
 }
 
+#[turbo_tasks::value(shared)]
+#[derive(Debug)]
+pub enum ProxyConfig {
+    Http(String),
+    Https(String),
+}
+
+#[turbo_tasks::value(transparent)]
+pub struct OptionProxyConfig(Option<ProxyConfig>);
+
 #[turbo_tasks::function]
-pub async fn fetch(url: Vc<String>, user_agent: Vc<Option<String>>) -> Result<Vc<FetchResult>> {
+pub async fn fetch(
+    url: Vc<String>,
+    user_agent: Vc<Option<String>>,
+    proxy_option: Vc<OptionProxyConfig>,
+) -> Result<Vc<FetchResult>> {
     let url = &*url.await?;
     let user_agent = &*user_agent.await?;
-    let client = reqwest::Client::new();
+    let proxy_option = &*proxy_option.await?;
+
+    let client_builder = reqwest::Client::builder();
+    let client_builder = match proxy_option {
+        Some(ProxyConfig::Http(proxy)) => client_builder.proxy(reqwest::Proxy::http(proxy)?),
+        Some(ProxyConfig::Https(proxy)) => client_builder.proxy(reqwest::Proxy::https(proxy)?),
+        _ => client_builder,
+    };
+
+    let client = client_builder.build()?;
 
     let mut builder = client.get(url);
     if let Some(user_agent) = user_agent {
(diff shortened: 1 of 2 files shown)


crates/turbo-tasks-fetch/src/lib.rs
```diff
@@ -36,11 +36,32 @@ impl HttpResponseBody {
     }
 }
 
+#[turbo_tasks::value(shared)]
+#[derive(Debug)]
+pub enum ProxyOption {
+    None,
+    Http(String),
+    Https(String),
+}
```
ForsakenHarmony, line 45:
```suggestion
#[turbo_tasks::value(shared)]
#[derive(Debug)]
pub enum Proxy {
    Http(String),
    Https(String),
}
```

crates/turbo-tasks-fetch/src/lib.rs
```diff
@@ -36,11 +36,32 @@ impl HttpResponseBody {
...
+
 #[turbo_tasks::function]
-pub async fn fetch(url: Vc<String>, user_agent: Vc<Option<String>>) -> Result<Vc<FetchResult>> {
+pub async fn fetch(
+    url: Vc<String>,
+    user_agent: Vc<Option<String>>,
+    proxy_option: Vc<ProxyOption>,
```
ForsakenHarmony, line 51:
```suggestion
    proxy_option: Vc<Option<Vc<ProxyOption>>>,
```

crates/turbo-tasks-fetch/src/lib.rs
ForsakenHarmony: Could probably also make this derive `TaskInput` and not have this be a turbo tasks value at all

crates/turbo-tasks-fetch/src/lib.rs
```diff
@@ -36,11 +36,34 @@ impl HttpResponseBody {
...
+
+#[turbo_tasks::value(transparent)]
+pub struct OptionProxyConfig(Option<ProxyConfig>);
+
 #[turbo_tasks::function]
-pub async fn fetch(url: Vc<String>, user_agent: Vc<Option<String>>) -> Result<Vc<FetchResult>> {
+pub async fn fetch(
+    url: Vc<String>,
+    user_agent: Vc<Option<String>>,
+    proxy_option: Vc<OptionProxyConfig>,
```
ForsakenHarmony, line 53: does it not work without the newtype?

kwonoj: I got some build error message like cannot be awaited without intofuture trait impl, not sure what's the difference.

kwonoj:
```
error[E0277]: the trait bound `Vc<std::option::Option<ProxyConfig>>: IntoFuture` is not satisfied
  --> crates/turbo-tasks-fetch/src/lib.rs:57:39
   |
57 |     let proxy_option = &*proxy_option.await?;
   |                                      -^^^^^
   |                                      ||
   |                                      |the trait `IntoFuture` is not implemented for `Vc<std::option::Option<ProxyConfig>>`
   |                                      help: remove the `.await`
   |
   = help: the following other types implement trait `IntoFuture`:
             Vc<T>
             &Vc<T>
```

ForsakenHarmony: I think you still have to wrap the inner type in a vc
`Vc<Option<Vc<ProxyConfig>>`

But I guess that makes it a bit more annoying to use

kwonoj: Looks like so. Probably ok to have a new type?

ForsakenHarmony (approved):

PR #7322, opened 2024-02-08
Improve daemon startup times

arlyon:
### Description

This is a PR for speeding up the daemon startup and setting up good practices for differentiating between data and service dependencies. To clarify the difference, file watching _service_ should start up immediately in an unavailable state, while things that need it can immediately take dependencies on its 'potentially-available' data. This also has a side-effect of fixing up our package discovery watcher as it makes the data flow a lot cleaner and ensures that the daemon can report data is unavailable when it is recalculating (using [`FutureExt::now_or_never`](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.now_or_never)) as well as cascading unavailable when parts of the system are no longer up to date. If a dependency wishes to wait for the data, then it can do so by simply awaiting the future. Great we've reimplemented async salsa by hand... though our data-dependency-graph will probably not really ever be so large that it outgrows this DIY solution.

The part that specifically handles package discovery availability is demonstrated here:

https://github.com/vercel/turbo/blob/048d5f5ef313a705b22a37e3e83ff30952569faf/crates/turborepo-filewatch/src/package_watcher.rs#L589-L593

Downstream dependencies who attempt a 'now or never' request on the data will receive an empty value, which translates to package discovery being unavailable

https://github.com/vercel/turbo/blob/048d5f5ef313a705b22a37e3e83ff30952569faf/crates/turborepo-filewatch/src/package_watcher.rs#L43-L63

This mechanism also powers package hash discovery, which depends on multiple components all being active to produce accurate answers.

### Testing Instructions

Existing tests should suffice

Closes TURBO-2296

### DIFF
--- a/crates/turborepo-filewatch/src/cookies.rs
+++ b/crates/turborepo-filewatch/src/cookies.rs
@@ -34,7 +34,8 @@ pub enum CookieError {
 pub struct CookieWriter {
     root: AbsoluteSystemPathBuf,
     timeout: Duration,
-    cookie_request_tx: OptionalWatch<mpsc::Sender<oneshot::Sender<Result<usize, CookieError>>>>,
+    cookie_request_sender_lazy:
+        OptionalWatch<mpsc::Sender<oneshot::Sender<Result<usize, CookieError>>>>,
     // _exit_ch exists to trigger a close on the receiver when all instances
     // of this struct are dropped. The task that is receiving events will exit,
     // dropping the other sender for the broadcast channel, causing all receivers
@@ -147,7 +148,7 @@ impl CookieWriter {
         timeout: Duration,
         mut recv: OptionalWatch<broadcast::Receiver<Result<notify::Event, NotifyError>>>,
     ) -> Self {
-        let (cookie_request_sender_tx, cookie_request_sender_rx) = OptionalWatch::new();
+        let (cookie_request_sender_tx, cookie_request_sender_lazy) = OptionalWatch::new();
         let (exit_ch, exit_signal) = mpsc::channel(16);
         tokio::spawn({
             let root = root.to_owned();
@@ -175,7 +176,7 @@ impl CookieWriter {
         Self {
             root: root.to_owned(),
             timeout,
-            cookie_request_tx: cookie_request_sender_rx,
+            cookie_request_sender_lazy,
             _exit_ch: exit_ch,
         }
     }
@@ -194,7 +195,7 @@ impl CookieWriter {
 
     async fn cookie_request_inner<T>(&self, request: T) -> Result<CookiedRequest<T>, CookieError> {
         let (resp_tx, resp_rx) = oneshot::channel();
-        let mut cookie_request_tx = self.cookie_request_tx.clone();
+        let mut cookie_request_tx = self.cookie_request_sender_lazy.clone();
         let Ok(cookie_request_tx) = cookie_request_tx.get().await.map(|s| s.to_owned()) else {
             // the cookie queue is not ready and will never be ready
             return Err(CookieError::Unavailable);
--- a/crates/turborepo-filewatch/src/cookies.rs
+++ b/crates/turborepo-filewatch/src/cookies.rs
@@ -189,17 +189,18 @@ impl CookieWriter {
         request: T,
     ) -> Result<CookiedRequest<T>, CookieError> {
         // we need to write the cookie from a single task so as to serialize them
-        let (resp_tx, resp_rx) = oneshot::channel();
+        tokio::time::timeout(self.timeout, self.cookie_request_inner(request)).await?
+    }
 
-        // make sure the cookie writer is ready
+    async fn cookie_request_inner<T>(&self, request: T) -> Result<CookiedRequest<T>, CookieError> {
+        let (resp_tx, resp_rx) = oneshot::channel();
         let mut cookie_request_tx = self.cookie_request_tx.clone();
         let Ok(cookie_request_tx) = cookie_request_tx.get().await.map(|s| s.to_owned()) else {
             // the cookie queue is not ready and will never be ready
             return Err(CookieError::Unavailable);
         };
-
         cookie_request_tx.send(resp_tx).await?;
-        let serial = tokio::time::timeout(self.timeout, resp_rx).await???;
+        let serial = resp_rx.await??;
         Ok(CookiedRequest { request, serial })
     }
 }
--- a/crates/turborepo-filewatch/src/cookies.rs
+++ b/crates/turborepo-filewatch/src/cookies.rs
@@ -3,11 +3,13 @@ use std::{collections::BinaryHeap, fs::OpenOptions, time::Duration};
 use notify::EventKind;
 use thiserror::Error;
 use tokio::{
-    sync::{mpsc, oneshot},
+    sync::{broadcast, mpsc, oneshot},
     time::error::Elapsed,
 };
 use tracing::trace;
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf, PathRelation};
+
+use crate::{NotifyError, OptionalWatch};
 
 #[derive(Debug, Error)]
 pub enum CookieError {
@@ -22,6 +24,8 @@ pub enum CookieError {
         io_err: std::io::Error,
         path: AbsoluteSystemPathBuf,
     },
+    #[error("cookie queue is not available")]
+    Unavailable,
 }
 
 /// CookieWriter is responsible for assigning filesystem cookies to a request
@@ -30,7 +34,7 @@ pub enum CookieError {
 pub struct CookieWriter {
     root: AbsoluteSystemPathBuf,
     timeout: Duration,
-    cookie_request_tx: mpsc::Sender<oneshot::Sender<Result<usize, CookieError>>>,
+    cookie_request_tx: OptionalWatch<mpsc::Sender<oneshot::Sender<Result<usize, CookieError>>>>,
     // _exit_ch exists to trigger a close on the receiver when all instances
     // of this struct are dropped. The task that is receiving events will exit,
     // dropping the other sender for the broadcast channel, causing all receivers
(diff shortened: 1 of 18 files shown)


crates/turborepo-filewatch/src/optional_watch.rs
```diff
@@ -0,0 +1,68 @@
...
+        OptionalWatch(rx)
+    }
+
+    /// Wait for the value to be available and then return it.
+    ///
+    /// If you receive a `RecvError`, the sender has been dropped, meaning you
+    /// will not receive any more updates. For efficiencies sake, you should
+    /// exit the task and drop any senders to other dependencies so that the
+    /// exit can propagate up the chain.
+    pub async fn get(&mut self) -> Result<SomeRef<'_, T>, RecvError> {
```
NicholasLYang, line 36: I'm a little confused by `SomeRef`. Could we unwrap here and return `&'a T`? Since we can turn `Ref<'a, Option<T>>` into a `&'a Option<T>` and then into `Option<&'a T>` which we unwrap. Also it might be worth noting that any ref here blocks the producer

NicholasLYang (changes requested):
Overall looks quite good. I am a little concerned about the potential for deadlocking with the use of `OptionalWatch`. Are we ever holding a `Ref` across an `await`?

arlyon:
Tokio [requires all futures to be Send](https://tokio.rs/tokio/tutorial/spawning#send-bound) meaning that we the compiler ensures that a `Ref` is never held across an await.

crates/turborepo-filewatch/src/optional_watch.rs
arlyon: I had a look here and as I understand it, a Ref is needed as it unlocks the watcher when it is dropped, similar to a MutexGuard. We could try to use `as_ref().unwrap()` but the lifetime of `&T` is tied to the `Ref` inside the function, not to the function itself.

crates/turborepo-filewatch/src/cookies.rs
```diff
@@ -164,7 +190,15 @@ impl CookieWriter {
     ) -> Result<CookiedRequest<T>, CookieError> {
         // we need to write the cookie from a single task so as to serialize them
         let (resp_tx, resp_rx) = oneshot::channel();
-        self.cookie_request_tx.clone().send(resp_tx).await?;
+
+        // make sure the cookie writer is ready
+        let mut cookie_request_tx = self.cookie_request_tx.clone();
+        let Ok(cookie_request_tx) = cookie_request_tx.get().await.map(|s| s.to_owned()) else {
```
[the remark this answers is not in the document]
arlyon: Even though SomeRef implements Deref it is not a borrow in of itself so I don't think `cloned` will work.

crates/turborepo-filewatch/src/lib.rs
```diff
@@ -89,53 +92,96 @@ pub struct FileSystemWatcher {
...
+                    warn!("failed to wait for initial filesystem cookie: {}", e);
+                    return;
+                }
+                debug!("filewatching ready");
+
+                let (sender, receiver) = broadcast::channel(1024);
+
+                // if either of these things fail, it means that events that we either won't
+                // receive events or don't have any downstream listeners. In either case, we
+                // should probably not run the watcher
```
gsoltis, line 152: The comment implies we should care about the errors, but we're ignoring them. Also, I think we still want filewatching even if we don't yet have any downstream consumers.

crates/turborepo-filewatch/src/lib.rs
```diff
@@ -89,53 +92,96 @@ pub struct FileSystemWatcher {
...
+        if root.relation_to_path(&cookie_dir) != PathRelation::Parent {
             return Err(WatchError::Setup(format!(
                 "Invalid cookie directory: {} does not contain {}",
                 root, cookie_dir
             )));
         }
-        setup_cookie_dir(cookie_dir)?;
-        let (sender, _) = broadcast::channel(1024);
+
+        let (sender_tx, sender_rx) = OptionalWatch::new();
```
gsoltis, line 115: Why is this an `OptionalWatch`? It looks like we set it then immediately await it. Do we ever unset it? Change it?

gsoltis (changes requested):
Initial Review, primarily on `cookies.rs` and `lib.rs` / file-watching startup. Will continue to review more of the PR, but wanted to make sure these comments get out.

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -57,279 +101,522 @@ impl PackageWatcher {
...
 struct Subscriber<T: PackageDiscovery> {
-    exit_rx: oneshot::Receiver<()>,
-    filter: WorkspaceGlobs,
-    recv: broadcast::Receiver<Result<Event, NotifyError>>,
+    /// we need to hold on to this. dropping it will close the downstream
+    /// data dependencies
+    _recv_tx: Arc<watch::Sender<Option<broadcast::Receiver<Result<Event, NotifyError>>>>>,
+
+    recv: OptionalWatch<broadcast::Receiver<Result<Event, NotifyError>>>,
+    backup_discovery: Arc<AsyncMutex<T>>,
```
gsoltis, line 160: Is this mutable? Why does it need a lock?

gsoltis (changes requested):
A few more comments. Going to wait on reviewing `package_watcher.rs` more thoroughly since I don't think this PR re-enables its use.

crates/turborepo-filewatch/src/globwatcher.rs
```diff
@@ -160,16 +160,24 @@ struct GlobTracker {
...
+        mut recv: OptionalWatch<broadcast::Receiver<Result<Event, NotifyError>>>,
     ) -> Self {
         let (exit_ch, exit_signal) = tokio::sync::oneshot::channel();
         let (query_ch, query_recv) = mpsc::channel(256);
         let cookie_root = cookie_jar.root().to_owned();
-        tokio::task::spawn(
-            GlobTracker::new(root.to_owned(), cookie_root, exit_signal, recv, query_recv).watch(),
-        );
+        tokio::task::spawn(async move {
+            let Ok(recv) = recv.get().await.map(|r| r.resubscribe()) else {
```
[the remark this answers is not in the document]
arlyon: The query channel will also be behind a watch.

crates/turborepo-filewatch/src/lib.rs
arlyon: probably a case of YAGNI, removing

crates/turborepo-filewatch/src/lib.rs
arlyon: Updated with some better explanation

crates/turborepo-filewatch/src/package_watcher.rs
arlyon: fixed

NicholasLYang (commented):
I feel like we need some naming conventions around these types. Like types wrapped `OptionalWatch` could be suffixed with `_lazy` or something similar. And maybe it'd make sense to diagram or explain somewhere which values the different structs take in and which they sent out.

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -293,31 +279,77 @@ where
         trace!("root watching exited");
         // Clean up the filewatching handle in the event that we never even got
         // started with filewatching. Again, we don't care about the error here.
-        let _ = fw_handle.await;
+        // let _ = fw_handle.await;
```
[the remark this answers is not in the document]
arlyon: Nope, not needed, TY

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -57,279 +101,518 @@ impl PackageWatcher {
...
+        // pass it down stream after the initial discovery, otherwise our package
+        // discovery watcher will consume events before we have our initial state
+        let (recv_tx, recv_rx) = OptionalWatch::new();
+        let recv_tx = Arc::new(recv_tx);
+
+        let backup_discovery = Arc::new(backup_discovery);
+
+        let package_json_path = repo_root.join_component("package.json");
+
+        let _task = tokio::spawn({
```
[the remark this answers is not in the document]
arlyon: I think it is a nice to colocate. I have experimented with both, so I don't really have a strong opinion on which one we go with. Happy to have that be a design decision in the docs we are putting together.

crates/turborepo-filewatch/src/lib.rs
```diff
@@ -89,53 +92,90 @@ pub struct FileSystemWatcher {
...
+    /// A convenience method around the sender watcher that waits for file
+    /// watching to be ready and then returns a handle to the file stream.
+    pub async fn subscribe(
+        &self,
+    ) -> Result<broadcast::Receiver<Result<Event, NotifyError>>, RecvError> {
+        let mut receiver = self.receiver.clone();
+        receiver.get().await.map(|r| r.resubscribe())
+    }
+
+    pub fn watch(&self) -> OptionalWatch<broadcast::Receiver<Result<Event, NotifyError>>> {
```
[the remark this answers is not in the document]
arlyon: Yep, some tests + here to thread the events into the downstream services. https://github.com/vercel/turbo/blob/1c92e7a60cc878b894db023022d7a681580584f5/crates/turborepo-lib/src/daemon/server.rs#L111

crates/turborepo-filewatch/src/lib.rs
```diff
@@ -89,53 +92,90 @@ pub struct FileSystemWatcher {
...
+        if root.relation_to_path(&cookie_dir) != PathRelation::Parent {
             return Err(WatchError::Setup(format!(
                 "Invalid cookie directory: {} does not contain {}",
                 root, cookie_dir
             )));
         }
-        setup_cookie_dir(cookie_dir)?;
-        let (sender, _) = broadcast::channel(1024);
+
+        let (receiver_tx, receiver_rx) = OptionalWatch::new();
```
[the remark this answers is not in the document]
arlyon: Good point, resolved.

arlyon:
Ok, changed convention from _rx and _tx to _lazy and _tx

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -283,33 +266,75 @@ where
...
+        let package_discovery = Arc::new(WatchingPackageDiscovery::new(
+            file_watching.package_watcher.clone(),
+        ));
+
+        // exit_root_watch delivers a signal to the root watch loop to exit.
+        // In the event that the server shuts down via some other mechanism, this
+        // cleans up root watching task.
+        let (exit_root_watch, root_watch_exit_signal) = oneshot::channel();
+        let watch_root_handle = tokio::task::spawn(watch_root(
+            file_watching.clone(),
```
gsoltis, line 308: Not sure if it matters too much for this particular service, but shouldn't `watch_root` subscribe to the OptionalWatch for filewatching like the other services? If not, we should at least leave a comment as to why this one is different

arlyon: You can only get a stream of events from the `FileSystemWatcher` using 

https://github.com/vercel/turbo/blob/0f78ba6e7f8d29f156792f14db1c2007f3dbcfba/crates/turborepo-filewatch/src/lib.rs#L168-L175

which goes through the optional watch.

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -293,31 +279,77 @@ where
...
+        package_discovery_backup: PD,
+        repo_root: AbsoluteSystemPathBuf,
+        trigger_shutdown: mpsc::Sender<()>,
+        log_file: AbsoluteSystemPathBuf,
+    ) -> (
+        Self,
+        oneshot::Sender<()>,
+        JoinHandle<Result<(), WatchError>>,
+    ) {
+        let file_watching = FileWatching::new(repo_root.clone(), package_discovery_backup).unwrap();
```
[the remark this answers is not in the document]
arlyon: Good point. Really, everything can move out (package discovery watcher, and file watcher) when the time comes. Since we are creating the package discovery in here also I think it makes sense but it makes sense to move it out if/when we decide to provide alternative APIs

crates/turborepo-lib/src/daemon/server.rs
arlyon: Open to passing the underlying OptionalWatch to the services rather than just passing in FileWatching directly to make the data dependencies more clear

gsoltis (approved):
Alright, let's merge this and continue on, once @NicholasLYang approves too.

NicholasLYang (approved):

PR #7326, opened 2024-02-08
feat(@turbo/repository): add affectedPackages API

mehulkar:
This API maps a list of files (strings anchored to the root of the workspace) to a set of packages that were affected by that change. This will include the packages the files belong to ~~as well as an dependents~~.

Since this doesn't include dependents, users will have to use it with `findPackagesAndDependents()` API:

```js
const files = [/*...*/];
const w = await new Workspace(".");
const [reverseGraph, affected] = await Promise.all([
  w.findPackagesAndDependents(),
  w.affectedPackages(files);
]);

// For each affected package, include all the dependents as well.
// Use a set to avoid dupes.
const allAffected = new Set(affected);
for (const p of affected) {
  reverseGraph[p].forEach(x => allAffected.add(x));
}

return [...allAffected] // spread into an array
```

### DIFF
--- a/crates/turborepo-repository/src/package_graph/mod.rs
+++ b/crates/turborepo-repository/src/package_graph/mod.rs
@@ -36,7 +36,7 @@ pub struct PackageGraph {
 /// There are other structs in this module that have "Workspace" in the name,
 /// but they do NOT follow the glossary, and instead mean "package" when they
 /// say Workspace. Some of these are labeled as such.
-#[derive(Eq, PartialEq, Hash)]
+#[derive(Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub struct WorkspacePackage {
     pub name: WorkspaceName,
     pub path: AnchoredSystemPathBuf,
@@ -91,6 +91,15 @@ type PackageVersion = String;
 pub enum WorkspaceName {
     Root,
     Other(String),
+}
+
+impl From<WorkspaceName> for String {
+    fn from(workspace_name: WorkspaceName) -> Self {
+        match workspace_name {
+            WorkspaceName::Root => ROOT_PKG_NAME.to_string(),
+            WorkspaceName::Other(name) => name,
+        }
+    }
 }
 
 impl Serialize for WorkspaceName {
--- a/crates/turborepo-repository/src/package_graph/mod.rs
+++ b/crates/turborepo-repository/src/package_graph/mod.rs
@@ -91,15 +91,6 @@ pub enum WorkspaceName {
     Other(String),
 }
 
-impl From<WorkspaceName> for String {
-    fn from(workspace_name: WorkspaceName) -> Self {
-        match workspace_name {
-            WorkspaceName::Root => ROOT_PKG_NAME.to_string(),
-            WorkspaceName::Other(name) => name,
-        }
-    }
-}
-
 impl Serialize for WorkspaceName {
     fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         match self {
--- a/crates/turborepo-repository/src/package_graph/mod.rs
+++ b/crates/turborepo-repository/src/package_graph/mod.rs
@@ -36,7 +36,7 @@ pub struct PackageGraph {
 /// There are other structs in this module that have "Workspace" in the name,
 /// but they do NOT follow the glossary, and instead mean "package" when they
 /// say Workspace. Some of these are labeled as such.
-#[derive(Eq, PartialEq, Hash, Ord, PartialOrd)]
+#[derive(Eq, PartialEq, Hash)]
 pub struct WorkspacePackage {
     pub name: WorkspaceName,
     pub path: AnchoredSystemPathBuf,
--- a/packages/turbo-repository/__tests__/find.test.ts
+++ b/packages/turbo-repository/__tests__/find.test.ts
@@ -3,7 +3,7 @@ import { Workspace, Package, PackageManager } from "../js/dist/index.js";
 
 interface AffectedPackagesTestParams {
   files: string[];
-  expected: string[];
+  expected: Package[];
   description: string;
 }
 
@@ -40,17 +40,23 @@ describe("Workspace", () => {
     const tests: AffectedPackagesTestParams[] = [
       {
         files: ["apps/app/file.txt"],
-        expected: ["app-a"],
+        expected: [{ name: "app-a", relativePath: "apps/app" }],
         description: "app change",
       },
       {
         files: ["packages/ui/a.txt"],
-        expected: ["app-a", "ui"],
+        expected: [
+          { name: "app-a", relativePath: "apps/app" },
+          { name: "ui", relativePath: "packages/ui" },
+        ],
         description: "lib change",
       },
       {
         files: ["package.json"],
-        expected: ["app-a", "ui"],
+        expected: [
+          { name: "app-a", relativePath: "apps/app" },
+          { name: "ui", relativePath: "packges/ui" },
+        ],
         description: "global change",
       },
       {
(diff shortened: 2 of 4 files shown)


packages/turbo-repository/rust/src/lib.rs
```diff
@@ -131,4 +135,56 @@ impl Workspace {
...
+            PackageChanges::Some(packages) => packages.into_iter().collect(),
+        };
+
+        let mut serializable_packages: Vec<Package> = packages
+            .into_iter()
+            .map(|p| {
+                let package_path = workspace_root.resolve(&p.path);
+                Package::new(p.name.to_string(), &workspace_root, &package_path)
+            })
+            .filter(|p| p.name != "//")
```
mehulkar, line 183: turbo core returns root packages here, but I'm not sure why, and I _think_ we don't care about it. I think global changes are already reflected as `PackageChanges::All`

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -131,4 +135,56 @@ impl Workspace {
...
+            .collect();
+
+        serializable_packages.sort();
```
mehulkar, line 186: I'm not sure what this `Sort` is doing on a custom struct, but I want it to sort by the `name` field. Might need to implement the `Ord` trait manually?

packages/turbo-repository/js/index.d.ts
```diff
@@ -33,4 +33,5 @@ export class Workspace {
    * dependents (i.e. the packages that depend on each of those packages).
    */
   findPackagesAndDependents(): Promise<Record<string, Array<string>>>;
+  affectedPackages(files: Array<string>): Promise<Array<Package>>;
```
mehulkar, line 36: Need a JSdoc here (add comments to rust fn)

packages/turbo-repository/rust/src/lib.rs
Zertsov: If I understand correctly you probably wanna use [this std function?]([url](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_by_key))

crates/turborepo-repository/src/package_graph/mod.rs
```diff
@@ -91,6 +91,15 @@ pub enum WorkspaceName {
     Other(String),
 }
 
+impl From<WorkspaceName> for String {
```
chris-olszewski, line 94: We should prefer implementing `Display` over converting to a string as the former allows us to use string formatting directly without converting to a string: e.g. `println!("The workspace is {}", WorkspaceName::Root)` vs `println!("The workspace is {}", String::from(WorkspaceName::Root))`
```suggestion
impl std::fmt::Display for WorkspaceName {
```

packages/turbo-repository/__tests__/find.test.ts
```diff
@@ -29,4 +37,54 @@ describe("Workspace", () => {
...
+      {
+        description: "app change",
+        files: ["apps/app/file.txt"],
+        expected: [{ name: "app-a", relativePath: "apps/app" }],
+      },
+      {
+        description: "lib change",
+        files: ["packages/ui/a.txt"],
+        expected: [
+          // { name: "app-a", relativePath: "apps/app" }, // TODO: this should be included
```
chris-olszewski, line 52: A little confused by the TODO, either this is a pretty large bug that should get fixed before merging or the API expects the caller to build up the closure of dependents and then this is the correct behavior.

packages/turbo-repository/rust/src/lib.rs
chris-olszewski: This is the comparison that's being used to sort (can be found by using [expand macro](https://rust-analyzer.github.io/manual.html#expand-macro-recursively) on the `#[derive(Ord)]` declaration:
```
impl $crate::cmp::Ord for Package {
    fn cmp(&self, other: &Self) -> $crate::cmp::Ordering {
        match (self, other) {
            (
                Package {
                    name: name_self,
                    absolute_path: absolute_path_self,
                    relative_path: relative_path_self,
                },
                Package {
                    name: name_other,
                    absolute_path: absolute_path_other,
                    relative_path: relative_path_other,
                },
            ) => match name_self.cmp(&name_other) {
                $crate::cmp::Ordering::Equal => {
                    match absolute_path_self.cmp(&absolute_path_other) {
                        $crate::cmp::Ordering::Equal => {
                            match relative_path_self.cmp(&relative_path_other) {
                                $crate::cmp::Ordering::Equal => $crate::cmp::Ordering::Equal,
                                c => return c,
                            }
                        }
                        c => return c,
                    }
                }
                c => return c,
            },
            _unused => $crate::cmp::Ordering::Equal,
        }
    }
}
```
If you just want to sort based on names, then you should look at [sort_by_key](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_by_key)

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -131,4 +135,58 @@ impl Workspace {
...
+
+        let hash_set_of_paths: HashSet<AnchoredSystemPathBuf> = files
+            .into_iter()
+            .filter_map(|path| {
+                let path_components = path.split(std::path::MAIN_SEPARATOR).collect::<Vec<&str>>();
+                let absolute_path = workspace_root.join_components(&path_components);
+                match workspace_root.anchor(&absolute_path) {
+                    Ok(path) => Some(path),
+                    Err(_) => None,
+                }
```
chris-olszewski, line 156: `Result::ok` is shorthand for this
```suggestion
               workspace_root.anchor(&absolute_path).ok()
```

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -131,4 +135,58 @@ impl Workspace {
 
         Ok(map)
     }
+
+    /// Given a set of "changed" files, returns a set of packages that are
```
chris-olszewski, line 139: Should probably document that the changed files are expected to:
 - be relative to workspace root
 - use current system path separator

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -131,4 +135,58 @@ impl Workspace {
...
+            PackageChanges::Some(packages) => packages.into_iter().collect(),
+        };
+
+        let mut serializable_packages: Vec<Package> = packages
+            .into_iter()
+            .map(|p| {
+                let package_path = workspace_root.resolve(&p.path);
+                Package::new(p.name.to_string(), &workspace_root, &package_path)
+            })
+            .filter(|p| p.name != ROOT_PKG_NAME.to_string())
```
chris-olszewski, line 185: We can avoid allocating for this comparison
```suggestion
            .filter(|p| p.name != ROOT_PKG_NAME)
```

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -131,4 +135,58 @@ impl Workspace {
...
+                    name: name.to_owned(),
+                    path: info.package_path().to_owned(),
+                })
+                .collect::<Vec<WorkspacePackage>>(),
+            PackageChanges::Some(packages) => packages.into_iter().collect(),
+        };
+
+        let mut serializable_packages: Vec<Package> = packages
+            .into_iter()
+            .map(|p| {
```
chris-olszewski, line 181: I think it makes sense to do the filtering before we perform the mapping.
(The second arm shouldn't get hit unless we do something wrong in the package graph, but better safe than sorry.
```suggestion
            .filter(|p| match &p.name {
                WorkspaceName::Root => false,
                WorkspaceName::Other(name) => name != ROOT_PKG_NAME,
            })
            .map(|p| {
```

chris-olszewski (approved):
Only minor suggestions

packages/turbo-repository/__tests__/find.test.ts
mehulkar: the TODO is there because we _want_ this behavior, but turbo core doesn't implement it. I can remove the expectation since it isn't clear

crates/turborepo-repository/src/package_graph/mod.rs
mehulkar: Display trait was already implemneted, but turns out i'm not using this anymore so I deleted it

packages/turbo-repository/rust/src/lib.rs
mehulkar: Thanks for the pointer, updated to `sort_by_key`

PR #7330, opened 2024-02-09
feat(Turborepo): Add spaces_id to layered config

gsoltis:
### Description

The core goal of this PR is to make `run.run` immutable. This enforces that our config and options are all resolved before we start a run. The last remaining piece of that was picking up the `spaces_id` from the layered config system. 

Adding `spaces_id` to the layered config system necessitated a bunch of other changes due to the fact that it is not contained within the `remoteCache` field in `turbo.json`. The config system previously assumed that any config file could contain the entirety of the config, which wasn't strictly true, but now is even less true. 

To that end, I created a separate struct to represent just the `remoteCache` field in `turbo.json`. It shares much, but not all of the parser for a config file. I split the rendering of `ConfigurationOptions` into two structs, since we want to render them differently in different circumstances (`turbo info` vs `turbo.json` synthesis in `turbo prune`). This had the upshot of letting us skip rendering a) empty fields in pruned repos and b) a non-functional `token` field in pruned repos.

### Testing Instructions

Updated integration tests for expected renderings of the config. Added unit tests for spaces_id.
Deleted the test for not parsing a token from the `remoteCache` field in `turbo.json`, the type system no longer allows us to parse a token since it's no longer directly a `ConfigurationOptions` instance.

Closes TURBO-2313

### DIFF
--- a/crates/turborepo-lib/src/config.rs
+++ b/crates/turborepo-lib/src/config.rs
@@ -135,7 +135,7 @@ const DEFAULT_API_URL: &str = "https://vercel.com/api";
 const DEFAULT_LOGIN_URL: &str = "https://vercel.com";
 const DEFAULT_TIMEOUT: u64 = 30;
 
-#[derive(Serialize, Deserialize, Default, Debug, PartialEq, Eq, Clone, Iterable)]
+#[derive(Deserialize, Default, Debug, PartialEq, Eq, Clone, Iterable)]
 #[serde(rename_all = "camelCase")]
 pub struct ConfigurationOptions {
     #[serde(alias = "apiurl")]
@@ -244,8 +244,8 @@ impl ResolvedConfigurationOptions for PackageJson {
 
 impl ResolvedConfigurationOptions for RawTurboJson {
     fn get_configuration_options(self) -> Result<ConfigurationOptions, Error> {
-        let mut opts = if let Some(configuration_options) = &self.remote_cache {
-            configuration_options.clone()
+        let mut opts = if let Some(remote_cache_options) = &self.remote_cache {
+            remote_cache_options.into()
         } else {
             ConfigurationOptions::default()
         };
--- a/crates/turborepo-lib/src/config.rs
+++ b/crates/turborepo-lib/src/config.rs
@@ -585,6 +585,9 @@ impl TurborepoConfigBuilder {
                     if let Some(timeout) = current_source_config.timeout {
                         acc.timeout = Some(timeout);
                     }
+                    if let Some(spaces_id) = current_source_config.spaces_id {
+                        acc.spaces_id = Some(spaces_id);
+                    }
 
                     acc
                 })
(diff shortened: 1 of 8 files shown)


crates/turborepo-lib/src/config.rs
```diff
@@ -211,11 +215,15 @@ impl ConfigurationOptions {
...
+
+    pub fn spaces_id(&self) -> Option<&str> {
+        self.spaces_id.as_deref()
+    }
 }
 
 // Maps Some("") to None to emulate how Go handles empty strings
 fn non_empty_str(s: Option<&str>) -> Option<&str> {
-    s.and_then(|s| (!s.is_empty()).then_some(s))
+    s.filter(|s| !s.is_empty())
```
chris-olszewski, line 226: TIL about `Option::filter`, nice!

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -320,6 +321,100 @@ impl Deserializable for ConfigurationOptions {
...
+    fn deserialize(
+        value: &impl DeserializableValue,
+        name: &str,
+        diagnostics: &mut Vec<DeserializationDiagnostic>,
+    ) -> Option<Self> {
+        value.deserialize(RawRemoteCacheOptionsVisitor, name, diagnostics)
+    }
+}
+
+struct RawRemoteCacheOptionsVisitor;
```
chris-olszewski, line 334: With this I think we can get rid of the `impl Deserializable for ConfigurationOptions` along with the `ConfigurationOptionsVisitor`

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -320,6 +321,100 @@ impl Deserializable for ConfigurationOptions {
...
+        let mut result = RawRemoteCacheOptions::default();
+        for (key, value) in members.flatten() {
+            // Try to deserialize the key as a string.
+            // We use `Text` to avoid an heap-allocation.
+            let Some(key_text) = Text::deserialize(&key, "", diagnostics) else {
+                // If this failed, then pass to the next key-value pair.
+                continue;
+            };
+            match key_text.text() {
+                "apiUrl" => {
```
chris-olszewski, line 359: I think this is fine, but just flagging that this is a breaking change from previous behavior we inherited from Go with all the weird casings for these keys.

chris-olszewski (approved):

crates/turborepo-lib/src/turbo_json/parser.rs
gsoltis: I don't think we can. We also deserialize other files via that visitor. The change I made is specific to the `remoteCache` key in `turbo.json`, all other parsing is left unchanged.

crates/turborepo-lib/src/turbo_json/parser.rs
gsoltis: It's a change only for the `remoteCache` key in `turbo.json`. All of the other config files still go through the other parser and support all of the variants.

PR #7334, opened 2024-02-09
Audit daemon error paths and ensure we are doing everything we can to handle

arlyon:
### Description

We have a few integration points with the daemon, this PR ensures they are all nicely handled.

- [x] `cache.rs`: we emit a debug warning and a telemetry event in the case we are unable to mark outputs correctly
- [x] `lsp/lib.rs`: in the LSP, package discovery means we cannot fulfill the user's request, so we must return an internal error. the error message has been set. Additionally, the LSP is a case where waiting for an answer is actually desirable, so a new API to support this has been added.
- [x] package discovery: if we have a bug in watching, we may yield packages that don't exist. we need to report a telemetry event, in this case so that we can be aware of when this happens
- [x] daemon interaction in run codepath: when trying to use the daemon during run we want to debug log but not error
- [x] daemon connection in daemon codepath: when calling the daemon we want errors to be explicitly reported. in this case, the errors bubbles all the way up and is printed out as expected

### Testing Instructions

Closes TURBO-2319

### DIFF
--- a/crates/turborepo-lib/src/daemon/proto/turbod.proto
+++ b/crates/turborepo-lib/src/daemon/proto/turbod.proto
@@ -16,6 +16,15 @@ service Turbod {
   //
   // Since 1.11.0
   rpc DiscoverPackages (DiscoverPackagesRequest) returns (DiscoverPackagesResponse);
+
+  // Request the list of packages that the daemon is aware of.
+  // This is a blocking call that will wait for the daemon to finish
+  // discovering the packages, rather than returning 'unavailable'.
+  // It will _still_ report unavailable if a critical process in the
+  // daemon is not running, and the answer will never come.
+  //
+  // Since 1.12.0
+  rpc DiscoverPackagesBlocking (DiscoverPackagesRequest) returns (DiscoverPackagesResponse);
 }
 
 message HelloRequest {
--- a/crates/turborepo-repository/src/discovery.rs
+++ b/crates/turborepo-repository/src/discovery.rs
@@ -11,6 +11,7 @@
 
 use std::sync::Arc;
 
+use tokio::time::error::Elapsed;
 use tokio_stream::{iter, StreamExt};
 use turbopath::AbsoluteSystemPathBuf;
 
@@ -270,7 +271,7 @@ impl<A: PackageDiscovery + Send + Sync, B: PackageDiscovery + Send + Sync> Packa
                     Err(err2) => Err(err2),
                 }
             }
-            Err(_) => {
+            Err(Elapsed { .. }) => {
                 tracing::debug!("primary strategy timed out, attempting fallback strategy");
                 self.fallback.discover_packages_blocking().await
             }
--- a/crates/turborepo-repository/src/discovery.rs
+++ b/crates/turborepo-repository/src/discovery.rs
@@ -42,7 +42,15 @@ pub enum Error {
 /// Defines a strategy for discovering packages on the filesystem.
 pub trait PackageDiscovery {
     // desugar to assert that the future is Send
+    /// Discover packages on the filesystem. In the event that this would block,
+    /// some strategies may return `Err(Error::Unavailable)`. If you want to
+    /// wait, use `discover_packages_blocking` which will wait for the result.
     fn discover_packages(
+        &self,
+    ) -> impl std::future::Future<Output = Result<DiscoveryResponse, Error>> + Send;
+
+    /// Discover packages on the filesystem, blocking until the result is ready.
+    fn discover_packages_blocking(
         &self,
     ) -> impl std::future::Future<Output = Result<DiscoveryResponse, Error>> + Send;
 }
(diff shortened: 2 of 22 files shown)


crates/turborepo-repository/Cargo.toml
```diff
@@ -26,6 +27,7 @@ turbopath = { workspace = true }
 turborepo-graph-utils = { path = "../turborepo-graph-utils" }
 turborepo-lockfiles = { workspace = true }
 turborepo-scm = { workspace = true }
+turborepo-telemetry = { version = "0.1.0", path = "../turborepo-telemetry" }
```
gsoltis, line 30: This dependency makes me nervous. We have consumers of `turbo-repository` from outside of the Turborepo CLI

crates/turborepo-repository/src/discovery.rs
```diff
@@ -216,69 +254,103 @@ impl<A: PackageDiscovery + Send, B: PackageDiscovery + Send> PackageDiscovery
...
+            Ok(Err(err1)) => {
+                tracing::debug!("primary strategy failed, attempting fallback strategy");
+                match self.fallback.discover_packages_blocking().await {
+                    Ok(packages) => Ok(packages),
+                    // if the backup is unavailable, return the original error
+                    Err(Error::Unavailable) => Err(err1),
+                    Err(err2) => Err(err2),
+                }
+            }
+            Err(_) => {
```
gsoltis, line 273: Nit: I think the `Elapsed` error is worth specifying here, mostly as a guard against the code changing and forcing authors to handle any new errors they introduce, since the error message here is specific to a timeout.

gsoltis (commented):
Just commenting for now, until a rebase happens.

crates/turborepo-repository/Cargo.toml
arlyon: Point taken, I will thread the error through and make sure to keep telem in lib only

arlyon: We also use it in turborepo-repository, which is OK for now, but I feel like that could very easily get consumed by other sources also. I will open a ticket for both of these to limit imports just so we are absolutely sure

arlyon: https://linear.app/vercel/issue/TURBO-2390/ensure-turborepo-telemetry-does-not-bleed-out-of-turbo-proper

crates/turborepo-repository/src/discovery.rs
arlyon: Good point, I have made the match total

crates/turborepo-lib/src/daemon/proto/turbod.proto
```diff
@@ -16,6 +16,15 @@ service Turbod {
...
   rpc DiscoverPackages (DiscoverPackagesRequest) returns (DiscoverPackagesResponse);
+
+  // Request the list of packages that the daemon is aware of.
+  // This is a blocking call that will wait for the daemon to finish
+  // discovering the packages, rather than returning 'unavailable'.
+  // It will _still_ report unavailable if a critical process in the
+  // daemon is not running, and the answer will never come.
+  //
+  // Since 1.12.0
+  rpc DiscoverPackagesBlocking (DiscoverPackagesRequest) returns (DiscoverPackagesResponse);
```
gsoltis, line 27: Pretty sure this won't work with the current setup, the entire grpc server has a 100ms timeout: https://github.com/vercel/turbo/blob/d438018e4e98d2e5f167613fdc56b22dcbe7ba96/crates/turborepo-lib/src/daemon/server.rs#L240

crates/turborepo-repository/src/discovery.rs
```diff
@@ -224,6 +255,28 @@ impl<A: PackageDiscovery + Send + Sync, B: PackageDiscovery + Send + Sync> Packa
...
+            Ok(Err(err1)) => {
+                tracing::debug!("primary strategy failed, attempting fallback strategy");
+                match self.fallback.discover_packages_blocking().await {
+                    Ok(packages) => Ok(packages),
+                    // if the backup is unavailable, return the original error
+                    Err(Error::Unavailable) => Err(err1),
+                    Err(err2) => Err(err2),
+                }
+            }
+            Err(Elapsed) => {
```
gsoltis, line 274: I don't think this is doing what's intended. This is creating a variable named `Elapsed` and assigning the error to it. I think this is the correct syntax:
```suggestion
            Err(Elapsed { .. }) => {
```

gsoltis (approved):
Looks good, with the fix-ship for the usage of `Elapsed`. 

The timeout issue will only affect the LSP, and isn't worse than the existing situation.

NicholasLYang (approved):

crates/turborepo-repository/src/discovery.rs
arlyon: I was worried this was the case, TIL on the syntax, TY

crates/turborepo-lib/src/daemon/proto/turbod.proto
arlyon: Yeah, our handling of timeouts were sub-optimal. I have moved over to a more reasonable 30s timeout default, coupled with a middleware that sets the timeout in the grpc headers for non blocking requests _if the client hasn't specified them_ meaning we also get better handling there too

Issue #6709, opened 2023-12-05
Add ability to do group-expanded to log-order for better CI readability in Github Actions


runspired:
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

see title

### Describe the solution you'd like

see title

### Describe alternatives you've considered

choosing chaos with --log-order=stream

<sub>[TURBO-1857](https://linear.app/vercel/issue/TURBO-1857/add-ability-to-do-group-expanded-to-log-order-for-better-ci)</sub>

anthonyshew:
Could you provide more details? How is this different from `--log-order=grouped`?

runspired:
@anthonyshew log order grouped collapses all groups, makes it impossible to find which groups have failures in them.

An alternative here is to make `--log-order=grouped` expand any groups that have failures in them

tknickman:
@runspired groups with failures are listed at the end of the run, but agreed - runs with failures should be expanded by default!

maschwenk:
I guess in terms of feasibility:

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines unclear if the expandability state is actually controllable?

I wonder if an alternative would be to pin failures to the bottom or something? Or restate them in a summary like Jest does? I know y'all have the `--summary` but obviously that's not super human readable.

mehulkar:
Since we're buffering the output until the task is done, we can just omit the group start/end markers for NZE tasks

PR #7335, opened 2024-02-09
fix: avoid grouping logs from tasks which ended in error

chris-olszewski:
### Description

Fixes https://github.com/vercel/turbo/issues/6709

Github Action doesn't provide a mechanism for grouping logs together and keeping them expanded which is helpful for highlighting errors.

This PR changes the behavior so we avoid emitting the grouping header/footer if task execution was a failure.

### Testing Instructions

Updated GH Action log grouping integration test

Closes TURBO-2320

### DIFF
--- a/crates/turborepo-lib/src/task_graph/visitor.rs
+++ b/crates/turborepo-lib/src/task_graph/visitor.rs
@@ -392,6 +392,12 @@ impl<'a> Visitor<'a> {
                 (vendor_behavior.group_suffix)(&group_name),
             );
             logger.with_header_footer(Some(header), Some(footer));
+
+            let (error_header, error_footer) = (
+                vendor_behavior.error_group_prefix.map(|f| f(&group_name)),
+                vendor_behavior.error_group_suffix.map(|f| f(&group_name)),
+            );
+            logger.with_error_header_footer(error_header, error_footer);
         }
         logger
     }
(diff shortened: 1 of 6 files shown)


mehulkar (commented):
This is great. Another option, which I could see myself preferring is to keep the grouping behavior, but make the header and footer red on failure. What do you think about that?

chris-olszewski:
> make the header and footer red on failure

Like [this](https://github.com/vercel/turbo/actions/runs/7851566660/job/21428813279?pr=7341#step:3:6)?

mehulkar:
Yeah exactly. If the arrow could be red too that would be great. I messed around in dev tools to get this: 
CleanShot 2024-02-10 at 12 36 36@2x

This is using the same red as we get at the end, but I actually don't think it's dark enough. Ideally, I want to be able to scroll through a long list of collapsed groups and visually find the one that failed. Maybe a big ❌ in the task prefix would do the trick (but not sure that's technically feasible since the prefix is computed ahead of time)
CleanShot 2024-02-10 at 12 38 41@2x

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -690,7 +690,10 @@ impl ExecContext {
             .instrument(span)
             .await;
 
-        let logs = match output_client.finish() {
+        // If the task resulted in an error, do not group in order to better highlight
+        // the error.
+        let keep_group = matches!(result, ExecOutcome::Success(_));
```
tknickman, line 695: Woah, TIL about `matches!`. Love that.

tknickman (approved):

PR #7336, opened 2024-02-09
feat(turbo): update changemapper to return name and path of packages

mehulkar:
Closes TURBO-2321

### DIFF
--- a/crates/turborepo-repository/src/package_graph/mod.rs
+++ b/crates/turborepo-repository/src/package_graph/mod.rs
@@ -40,7 +40,7 @@ impl WorkspacePackage {
     pub fn root() -> Self {
         Self {
             name: WorkspaceName::Root,
-            path: AnchoredSystemPathBuf::from_raw("").unwrap(),
+            path: AnchoredSystemPathBuf::default(),
         }
     }
 }
--- a/crates/turborepo-repository/src/package_graph/mod.rs
+++ b/crates/turborepo-repository/src/package_graph/mod.rs
@@ -33,14 +33,14 @@ pub struct PackageGraph {
 #[derive(Eq, PartialEq, Hash)]
 pub struct WorkspacePackage {
     pub name: WorkspaceName,
-    pub path: String,
+    pub path: AnchoredSystemPathBuf,
 }
 
 impl WorkspacePackage {
     pub fn root() -> Self {
         Self {
             name: WorkspaceName::Root,
-            path: String::from(""),
+            path: AnchoredSystemPathBuf::from_raw("").unwrap(),
         }
     }
 }
@@ -360,7 +360,7 @@ impl PackageGraph {
                         let w_name = WorkspaceName::Other(n.to_owned());
                         Some(WorkspacePackage {
                             name: w_name.clone(),
-                            path: self.workspace_dir(&w_name).unwrap().to_owned().to_string(),
+                            path: self.workspace_dir(&w_name).unwrap().to_owned(),
                         })
                     }
                     // if the root package has changed, then we should report `None`
@@ -375,7 +375,7 @@ impl PackageGraph {
                 .keys()
                 .map(|name| WorkspacePackage {
                     name: name.clone(),
-                    path: self.workspace_dir(name).unwrap().to_owned().to_string(),
+                    path: self.workspace_dir(name).unwrap().to_owned(),
                 })
                 .collect()
         }))
--- a/crates/turborepo-repository/src/package_graph/mod.rs
+++ b/crates/turborepo-repository/src/package_graph/mod.rs
@@ -30,6 +30,21 @@ pub struct PackageGraph {
     lockfile: Option<Box<dyn Lockfile>>,
 }
 
+#[derive(Eq, PartialEq, Hash)]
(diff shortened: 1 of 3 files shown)


crates/turborepo-repository/src/package_graph/mod.rs
```diff
@@ -30,6 +30,21 @@ pub struct PackageGraph {
     lockfile: Option<Box<dyn Lockfile>>,
 }
 
+#[derive(Eq, PartialEq, Hash)]
+pub struct WorkspacePackage {
+    pub name: WorkspaceName,
+    pub path: String,
```
mehulkar, line 36: I want this to be `AnchoredSystemPath` or `AnchoredSystemPathBuf` (not sure which), but when I do that Rust complains when I try to use it with `HashSet<WorkspacePackage>` because the size of `[u8]` is not known at compile time

mehulkar: Looks like I get the same compile error even for `Vec<WorkspacePackage>`:

```
error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
   --> crates/turborepo-repository/src/package_graph/mod.rs:328:17
    |
328 |     ) -> Result<Vec<WorkspacePackage>, ChangedPackagesError> {
    |                 ^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
```

chris-olszewski: You'll want `AbsoluteSystemPathBuf` as `AnchoredSystemPath` is itself a [dynamically sized type](https://doc.rust-lang.org/reference/dynamically-sized-types.html)

mehulkar: thank you!

crates/turborepo-repository/src/package_graph/mod.rs
```diff
@@ -30,6 +30,21 @@ pub struct PackageGraph {
...
+pub struct WorkspacePackage {
+    pub name: WorkspaceName,
+    pub path: AnchoredSystemPathBuf,
+}
+
+impl WorkspacePackage {
+    pub fn root() -> Self {
+        Self {
+            name: WorkspaceName::Root,
+            path: AnchoredSystemPathBuf::from_raw("").unwrap(),
```
NicholasLYang, line 43: Tiny nit, `AnchoredSystemPathBuf` implements `Default` so you can do this
```suggestion
            path: AnchoredSystemPathBuf::default(),
```

crates/turborepo-repository/src/package_graph/mod.rs
```diff
@@ -30,6 +30,21 @@ pub struct PackageGraph {
     lockfile: Option<Box<dyn Lockfile>>,
 }
 
+#[derive(Eq, PartialEq, Hash)]
+pub struct WorkspacePackage {
```
NicholasLYang, line 34: I'm not in love with this name because it implies that this is a workspace's package, like how `WorkspaceName` implies it's a workspace's name. When really it's a workspace. Maybe we can just call it `Workspace`?

crates/turborepo-repository/src/package_graph/mod.rs
```diff
@@ -341,15 +356,29 @@ impl PackageGraph {
                         != info.transitive_dependencies.as_ref()
                 })
                 .map(|(name, _info)| match name {
-                    WorkspaceName::Other(n) => Some(WorkspaceName::Other(n.to_owned())),
+                    WorkspaceName::Other(n) => {
+                        let w_name = WorkspaceName::Other(n.to_owned());
+                        Some(WorkspacePackage {
+                            name: w_name.clone(),
+                            path: self.workspace_dir(&w_name).unwrap().to_owned(),
```
NicholasLYang, line 363: Can't we get this from the `info` field?
```suggestion
                            path: info.package_path().to_owned(),
```

NicholasLYang (approved):
Looks good, just some fix ship stuff

crates/turborepo-repository/src/package_graph/mod.rs
mehulkar: I should add a comment. Basically all the _other_ names are wrong, and "workspace" should be replaced with "package" everywhere. I tried to just use `Package`, but I think it was conflicting with `turborepo_lockfiles::Package` and I didn't want to figure it out.

Would you be ok with a comment above this and some of the other structs? 

https://vercel.com/docs/vercel-platform/glossary#workspace
https://vercel.com/docs/vercel-platform/glossary#package

That said... this is _still_ not the best name because "root" is not a workspace package either.

crates/turborepo-repository/src/package_graph/mod.rs
mehulkar: nice good catch, i didn't take your inline suggestion as it required a couple other changes. see new commit

crates/turborepo-repository/src/package_graph/mod.rs
NicholasLYang: Hmm, I feel like this rename should be done all at once though. Because otherwise it'll be confusing

PR #7339, opened 2024-02-09
chore(Turborepo): Move daemon paths off of CommandBase

gsoltis:
### Description

Stepping stone to moving `CommandBase` out of `run` 
- Move daemon path handling into the daemon module

### Testing Instructions

Updated existing test suite

Closes TURBO-2323

### DIFF
--- a/crates/turborepo-lib/src/daemon/mod.rs
+++ b/crates/turborepo-lib/src/daemon/mod.rs
@@ -111,24 +111,16 @@ pub(crate) mod proto {
 
 #[cfg(test)]
 mod test {
-    use test_case::test_case;
     use turbopath::AbsoluteSystemPathBuf;
 
     use super::repo_hash;
 
-    #[cfg(not(target_os = "windows"))]
-    #[test_case("/tmp/turborepo", "6e0cfa616f75a61c"; "basic example")]
-    fn test_repo_hash(path: &str, expected_hash: &str) {
-        let repo_root = AbsoluteSystemPathBuf::new(path).unwrap();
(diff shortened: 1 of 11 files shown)


crates/turborepo-lib/src/daemon/mod.rs
```diff
@@ -54,3 +108,31 @@ pub(crate) mod proto {
...
+    #[cfg(target_os = "windows")]
+    #[test_case("C:\\\\tmp\\turborepo", "0103736e6883e35f"; "basic example")]
+    fn test_repo_hash_win(path: &str, expected_hash: &str) {
+        let repo_root = AbsoluteSystemPathBuf::new(path).unwrap();
+        let hash = repo_hash(&repo_root);
+
+        assert_eq!(hash, expected_hash);
+        assert_eq!(hash.len(), 16);
+    }
+}
```
Zertsov, line 138: Is there a way to consolidate these into a single function? Understandable if not but this feels odd to me that we have to literally duplicate a function for this to work on Windows.

First thought was to do the `cfg(target_os)` in the function and have `test_case` take in an extra param that passes in the `target_os` string, but I could be missing something.

Zertsov (approved):

crates/turborepo-lib/src/daemon/mod.rs
gsoltis: If we jettison `test_case` we can consolidate this a bit. I'm going to go with that, since we aren't actually enumerating a lot of test cases.

PR #7340, opened 2024-02-10
feat(tasks-fs): bump up notify, fix empty file watch

kwonoj:
### Description

This PR upgrades `notify` pkg to create a filesystem watcher. 

It is due to recent issue we found in PACK-2437, the edge cases like

- start a watcher with file have any contents
- open editor, clear all (make file empty), save

doesn't create any file watcher event. This makes some cases turbopack does not trigger hmr even if it's desired.

One thing to note is even with upgrade its event kind is somewhat unexpected; it's not `Modify` event but `Metadata` event. I can't say why it emits in that way, but add it as workaround for now.

We had experiences of trying to upgrade & revert this pkg before, so the upgrade plan need some caution. 

Plan is

- Once PR is approved, cut turbopack _before_ PR and bump next.js first
- Land PR, create new turbopack release so release contains only 1 changes to notify
- Ensure new turbopack update in next.js won't break (like upgrading front, etcs) and make it easy to revert

Closes PACK-2437.


crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -343,60 +374,99 @@ impl DiskFileSystem {
...
-                                Some(path) => {
-                                    batched_invalidate_path_and_children.insert(path.clone());
-                                    batched_invalidate_path_and_children_dir.insert(path);
-                                }
-                                None => {
-                                    batched_invalidate_path_and_children
-                                        .insert(PathBuf::from(&root));
-                                    batched_invalidate_path_and_children_dir
-                                        .insert(PathBuf::from(&root));
+                        Ok(Ok(events)) => {
```
kwonoj, line 377: I tried to preserve existing event condition logics as much as possible to avoid regressions, but still possible I may miss something.

kwonoj:
Testing against next.js seems passing at least https://github.com/vercel/next.js/pull/61879

Cargo.toml
```diff
@@ -278,3 +278,4 @@ urlencoding = "2.1.2"
 webbrowser = "0.8.7"
 which = "4.4.0"
 unicode-segmentation = "1.10.1"
+notify-debouncer-full = "0.3.1"
```
kwonoj, line 281: `notify-debouncer`reexports `notify` and recommend to use it unless custom feature is needed, which works for us.

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -264,6 +270,26 @@ impl DiskFileSystem {
         self.start_watching_internal(true)
     }
 
+    /**
+     * Create a watcher and start watching by creating `debounced` watcher
+     * via `full debouncer`
```
kwonoj, line 275: I chose `full` debouncer impl compare to `mini` due to below ergonomics supported, which we can hope to avoid excessive event count for some cases.

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -343,60 +374,99 @@ impl DiskFileSystem {
...
+                                    // PACK-2437 addresses some cases notify doesn't emit event to trigger hmr,
+                                    // cases like starting from a file have contents -> clear & save.
+                                    // v4 doesn't emit any debounced event for those,
+                                    // and v6 emits an event with `ModifyKind::Metadata(MetadataKind::Any)`
+                                    // it is unclear why these case are not considered as normal modify, anyway we use certain
+                                    // event to workaround.
+                                    EventKind::Any | EventKind::Modify(ModifyKind::Any | ModifyKind::Metadata(MetadataKind::Any)) => {
+                                        batched_invalidate_path.extend(paths.clone());
+                                        batched_invalidate_path_and_children.extend(paths.clone());
+                                        batched_invalidate_path_and_children_dir.extend(paths.clone());
```
sokra, line 434: That seem to be a very extensive invalidation. I could imagine it over invalidating when e.g. metadata of a folder changes

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -343,60 +374,99 @@ impl DiskFileSystem {
...
+                                        for parent in paths.iter().filter_map(|path| path.parent()) {
+                                            batched_invalidate_path_dir.insert(PathBuf::from(parent));
+                                        }
+                                    }
+                                    EventKind::Modify(
+                                        ModifyKind::Metadata(..)
+                                        | ModifyKind::Other
+                                        // hope RenameMode::both are suffecient for the rename event, ignore others
+                                        | ModifyKind::Name(..)
```
sokra, line 443: We should handle the other variants too just in case

crates/turbo-tasks-fs/src/lib.rs
kwonoj: true, the unfortunate thing is it comes as `metadata::any`, so really no way to figure out if it's other arbitaray metadata change or content change in disguise. 

I created https://github.com/notify-rs/notify/issues/564 to get some clarification, based on those answer we may update codes later.

crates/turbo-tasks-fs/src/lib.rs
kwonoj: Other then `both` in renamemode for the name event, there are from|to|any|other. 

we can exclude from|to since `both` should be emitted after consolidating those (otherwise it is probably a bug), any should be handled by debouncer if path can be detected (https://github.com/notify-rs/notify/blob/main/notify-debouncer-full/src/lib.rs#L296-L299) and other is unused.

kwonoj: TL:DR debouncer-full impl supposed to wrap & handle those cases to not bubble up unexpectedly for us.

crates/turbo-tasks-fs/src/lib.rs
kwonoj: Still waiting for confirmation from upstream, I came to think this is the limitation of the mac os's underlying behavior. Modified event comes intermittently for the same repro case, and it went down to the kernel's event flag itself doesn't contain such thing.

I wanted to compare how other implementation does, and then found libuv's impl:

It originally relied on `kFSEventStreamEventFlagItemModified` as same as what we expected -  https://github.com/libuv/libuv/blob/ea8db64559680b589196b1ae01152c35d2232c41/src/unix/fsevents.c#L150

and then it fixed to includes all of possible-event data including metadata - 
https://github.com/libuv/libuv/commit/73cf3600d75a5884b890a1a94048b8f3f9c66876#diff-e12fdb1f404f1c97bbdcc0956ac90d7db0d811d9fa9ca83a3deef90c937a486cR95-R99 
to emit `UV_CHANGE` event.

So this change might create additional excessive notify event for us, in my opinion it's not worsely efficient than other watch implementation.

sokra: Could we skip the `_and_children` invalidation? That event on a folder doesn't need to invalidate all contained files and folders, does it?

crates/turbo-tasks-fs/src/lib.rs
sokra: I would just handle them in any naive way, just to be safe.

Seems like debouncer can emit a To event if it can match From and To.

kwonoj: Yes it emit to, which will be dupe event to both. Do we want multiple invalidation with same change?

crates/turbo-tasks-fs/src/lib.rs
kwonoj: I think that might work, yes.

crates/turbo-tasks-fs/src/lib.rs
sokra: > Do we want multiple invalidation with same change?

It doesn't matter.

> which will be dupe event to both

It might emit `To` when it can't match it with a `From` event. Not sure when this can happen, but I would feel better if we handle it

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -343,60 +374,103 @@ impl DiskFileSystem {
...
-                                Some(path) => {
-                                    batched_invalidate_path_and_children.insert(path.clone());
-                                    batched_invalidate_path_and_children_dir.insert(path);
-                                }
-                                None => {
-                                    batched_invalidate_path_and_children
-                                        .insert(PathBuf::from(&root));
-                                    batched_invalidate_path_and_children_dir
-                                        .insert(PathBuf::from(&root));
+                        Ok(Ok(events)) => {
```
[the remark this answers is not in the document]
kwonoj: yeah, would like to do a followup pr. moving whole makes diff bit unfriendly in gh for the review esp this contains some event type changes for the subscription.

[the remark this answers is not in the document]
ForsakenHarmony: idk, you already have to manually compare as the diff doesn't line up at all

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -264,6 +270,26 @@ impl DiskFileSystem {
         self.start_watching_internal(true)
     }
 
+    /**
```
ForsakenHarmony, line 273: use a doc comment for this

sokra (approved):

crates/turbo-tasks-fs/src/lib.rs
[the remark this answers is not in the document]
kwonoj: yeah but it's still diff. for this change, unless it is explicitly required I'd like to make minimal.

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -288,14 +312,19 @@ impl DiskFileSystem {
...
         // The notification back-end is selected based on the platform.
-        let mut watcher = watcher(tx, delay)?;
+        let mut debounced_watcher = notify_debouncer_full::new_debouncer(delay, None, tx)?;
         // Add a path to be watched. All files and directories at that path and
         // below will be monitored for changes.
         #[cfg(any(target_os = "macos", target_os = "windows"))]
-        watcher.watch(&root_path, RecursiveMode::Recursive)?;
+        debounced_watcher
+            .watcher()
+            .watch(&root_path, RecursiveMode::Recursive)?;
```
ForsakenHarmony, line 321: I think we also want the "cache" here and below?

`debouncer.cache().add_root(&root_path, RecursiveMode::Recursive);`

kwonoj: full debouncer internally autosetup those

https://github.com/notify-rs/notify/blob/08e74dae8e96fbd25704cdaa530ffc02f6d33039/CHANGELOG.md#L9-L20

ForsakenHarmony: but the version in this PR is too old

PR #7342, opened 2024-02-11
wip

kwonoj:


.github/workflows/turbopack-nightly-rustdocs-deploy.yml
```diff
@@ -0,0 +1,75 @@
+name: Turbopack Nightly Release
```
ForsakenHarmony, line 1: rename this

kwonoj: yeah but also actually found https://vercel.slack.com/archives/C03EWR7LGEN/p1707765219982619?thread_ts=1707761785.442889&cid=C03EWR7LGEN , so going to revert 😭

PR #7349, opened 2024-02-12
dedupe resolving independent on reference type

sokra:
### Description

* dedupe resolving independent on reference type
* add more tracing to resolving steps


kwonoj (approved):

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1383,223 +1389,236 @@ async fn resolve_internal_inline(
...
+    let span = {
+        let lookup_path = lookup_path.to_string().await?;
+        let request = request.to_string().await?;
+        tracing::info_span!(
+            "internal resolving",
+            lookup_path = *lookup_path,
+            request = *request
+        )
+    };
+    async move {
```
ForsakenHarmony, line 1401: not a fan of adding indent to the whole function for tracing

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1383,223 +1389,236 @@ async fn resolve_internal_inline(
...
-            // `compilerOptions.paths` option might alias "@*" to "./*", which
-            // would also match a request to "@emotion/core". Here, we follow what the
-            // Typescript resolution algorithm does in case an alias match
-            // doesn't resolve to anything: fall back to resolving the request normally.
-            if let Some(result) = resolved_result {
-                return Ok(result);
+    let span = {
+        let lookup_path = lookup_path.to_string().await?;
+        let request = request.to_string().await?;
+        tracing::info_span!(
```
ForsakenHarmony, line 1395: isn't this function traced twice now?

crates/turbopack-core/src/resolve/mod.rs
sokra: me neither, but `.enter()` doesn't work in async code...

crates/turbopack-core/src/resolve/mod.rs
sokra: fixed

crates/turbopack-core/src/resolve/mod.rs
ForsakenHarmony: would rather have another function :^)

PR #7355, opened 2024-02-12
chore(turbo): rename WorkspaceName to PackageName

mehulkar:
`PackageName` isn't the best name of this concept, but I think the real work here is to combine some of the structs in the packagegraph, which will make it more clear what the names should be. I decided not to try to do all that work right now.

There is a long tail of variable names and methods that I will keep working on renaming, but I want to keep the diffs reasonable in size.

Closes TURBO-2332


crates/turborepo-repository/src/package_graph/mod.rs
```diff
@@ -80,37 +80,35 @@ impl PackageInfo {
     }
 }
 
-type PackageName = String;
-type PackageVersion = String;
```
NicholasLYang, line 84: Do we have to delete `PackageVersion`? That seems fine to keep, no?

mehulkar: sure I can bring that back

mehulkar: brought it back, but tbh `BTreeMap<String, PackageVersion>` feels more weird to me. the naming conflict feels like a weird reason to elevate one String type over another. Fine with me for this PR though

NicholasLYang: Agreed, but `BTreeMap<String, String>` is very opaque to me. What about `BTreeMap<PackageKey, PackageVersion>`?

mehulkar: That's fine with me. I think some of these structs could probably get sorted out a bit

mehulkar:
@NicholasLYang updated with your suggestions

NicholasLYang (approved):

chris-olszewski (approved):

PR #7358, opened 2024-02-12
fix: do better validation for existing tokens

Zertsov:
### Description
Found a path in which a token can be invalid but we claim it isn't. The user would have to run `turbo logout` and then login to get a new token.

### Testing Instructions
1. Login
2. Go to Vercel account or whatever API issues tokens and delete the token (invalidate it)
3. run `turbo login` again
4. Ensure that login process proceeds as normal

Closes TURBO-2334

### DIFF
--- a/crates/turborepo-auth/src/lib.rs
+++ b/crates/turborepo-auth/src/lib.rs
@@ -12,19 +12,150 @@ mod ui;
 pub use auth::*;
 pub use error::Error;
 pub use login_server::*;
+use turborepo_api_client::TokenClient;
+use turborepo_vercel_api::token::ResponseTokenMetadata;
 
 /// Token is the result of a successful login. It contains the token string and
-/// a boolean indicating whether the token already existed on the filesystem.
-#[derive(Debug)]
+/// potentially metadata about the token.
+#[derive(Debug, Clone)]
 pub enum Token {
+    /// An existing token on the filesystem
     Existing(String),
+    /// A token that was just created, but not yet written to the filesystem
     New(String),
 }
 impl Token {
-    pub fn into_inner(self) -> String {
+    pub fn new(token: String) -> Self {
+        Self::New(token)
+    }
+    pub fn existing(token: String) -> Self {
+        Self::Existing(token)
+    }
+    /// Checks if the token is valid. We do a few checks:
+    /// 1. Fetch the token metadata.
+    /// 2. From the metadata, check if the token is active.
+    /// 3. If the token is a SAML SSO token, check if it's expired.
+    pub async fn is_valid(&self, client: &impl TokenClient) -> Result<bool, Error> {
+        let metadata = self.fetch_metadata(client).await?;
+        let current_time = current_unix_time();
+        let active = is_token_active(&metadata, current_time);
+        Ok(active)
+    }
+
+    async fn fetch_metadata(
+        &self,
+        client: &impl TokenClient,
+    ) -> Result<ResponseTokenMetadata, Error> {
+        client
+            .get_metadata(self.into_inner())
+            .await
+            .map_err(Error::from)
+    }
+    /// Returns the underlying token string.
+    pub fn into_inner(&self) -> &str {
         match self {
-            Self::Existing(s) => s,
-            Self::New(s) => s,
+            Self::Existing(token) | Self::New(token) => token.as_str(),
+        }
+    }
+}
+
+fn current_unix_time() -> u128 {
+    use std::time::{SystemTime, UNIX_EPOCH};
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("Time went backwards")
+        .as_millis()
+}
+
+// As of the time of writing, this should always be true, since a token that
+// isn't active returns an error when fetching metadata for the token.
+fn is_token_active(metadata: &ResponseTokenMetadata, current_time: u128) -> bool {
+    let active_at = metadata.active_at;
+
+    let earliest_expiration = metadata
+        .scopes
+        .iter()
+        .filter_map(|scope| scope.expires_at)
+        .min();
+
+    // Not all scopes have an expiration date, so we need to check if all of them
+    // are expired. If there isn't an expiration date, we assume they are infinite
+    // and therefore cannot be expired.
+    let all_scopes_active =
+        earliest_expiration.map_or(true, |expiration| current_time < expiration);
+
+    all_scopes_active && (active_at <= current_time)
+}
+
+#[cfg(test)]
+mod tests {
+    use turborepo_vercel_api::token::Scope;
+
+    use super::*;
+
+    #[test]
+    fn test_is_token_active() {
+        let current_time = current_unix_time();
+        let quick_scope = |expiry| Scope {
+            expires_at: expiry,
+            scope_type: "".to_string(),
+            origin: "".to_string(),
+            created_at: 0,
+            team_id: None,
+        };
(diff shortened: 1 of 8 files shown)


Zertsov:
* `main`
  * **#7358** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7358?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

crates/turborepo-auth/src/lib.rs
```diff
@@ -12,19 +12,81 @@ mod ui;
...
+
+        let active = self.is_token_active(&metadata);
+
+        // Not all scopes have an expiration date, so we need to check if all of them
+        // are expired. If there isn't an expiration date, we assume they are infinite
+        // and therefore cannot be expired.
+        let expired = metadata.scopes.iter().all(|scope| match scope.expires_at {
+            Some(expires_at) => expires_at < metadata.active_at,
+            None => false,
+        });
```
chris-olszewski, line 53: I might be missing something, but I think this gets checked by `is_token_active`?

crates/turborepo-auth/src/lib.rs
```diff
@@ -12,19 +12,81 @@ mod ui;
...
+        Self::New(token)
+    }
+    pub fn existing(token: String) -> Self {
+        Self::Existing(token)
+    }
+    /// Checks if the token is valid. We do a few checks:
+    /// 1. Fetch the token metadata.
+    /// 2. From the metadata, check if the token is active.
+    /// 3. If the token is a SAML SSO token, check if it's expired.
+    pub async fn is_valid(&self, client: &impl TokenClient) -> Result<bool, Error> {
```
chris-olszewski, line 38: This is a great candidate for a unit test to verify the logic and make sure:
 - we catch case where all scopes are expired
 - we catch case where `active_at` is after the current time
 - any token with a saml_sso scope is considered to be invalid

chris-olszewski (changes requested):
Trying to login with a valid existing token gives a pretty cryptic error:
Screenshot 2024-02-12 at 4 32 58 PM

crates/turborepo-auth/src/lib.rs
Zertsov: I think they were checking slightly different things, but after looking at it again I think I can simplify them and the unit test

crates/turborepo-auth/src/lib.rs
Zertsov: Refactored this to put the heavy lifting in `is_token_active`, since in theory an expired token isn't active.

crates/turborepo-auth/src/lib.rs
```diff
@@ -12,19 +12,139 @@ mod ui;
...
+#[cfg(test)]
+mod tests {
+    use turborepo_vercel_api::token::Scope;
+
+    use super::*;
+
+    #[test]
+    fn test_is_token_active() {
+        let current_time = current_unix_time();
+        let quick_scope = |expiry| Scope {
```
Zertsov, line 97: MW2 had it right all along: quick scopes on Rust are the way 😤

crates/turborepo-auth/src/lib.rs
```diff
@@ -12,19 +12,139 @@ mod ui;
...
+fn is_token_active(metadata: &ResponseTokenMetadata) -> bool {
+    let current_time = current_unix_time();
+    let active_at = metadata.active_at;
+
+    let active_less_than_current = active_at <= current_time;
+    let all_scopes_active = metadata.scopes.iter().all(|scope| {
+        scope
+            .expires_at
+            .map_or(true, |expires_at| expires_at > current_time)
+    });
```
chris-olszewski, line 80: Complete bikeshed. My brain understands this a lil faster, something about an `all` that contains `map_or` confuses me.

Feel free to ignore if you disagree
```suggestion
    let earliest_expiration = metadata
        .scopes
        .iter()
        .filter_map(|scope| scope.expires_at)
        .max();
    let all_scopes_active =
        earliest_expiration.map_or(true, |expiration| current_time < expiration);
```

crates/turborepo-auth/src/lib.rs
```diff
@@ -12,19 +12,139 @@ mod ui;
...
+            ),
+            // Case: Token inactive (future `active_at`)
+            (
+                current_time + 10,
+                vec![quick_scope(Some(current_time + 20))],
+                false,
+            ),
+        ];
+
+        for (active_at, scopes, expected) in cases {
```
chris-olszewski, line 140: I do think using [test_case](https://docs.rs/test-case/latest/test_case/) instead of manual table tests has some benefits, primarily it'll run all of the test cases on failure and not bail after the first failure. 

Not blocking, just sharing my 2 cents

crates/turborepo-auth/src/lib.rs
```diff
@@ -12,19 +12,139 @@ mod ui;
...
+            ),
+            // Case: Token active, one scope not expired
+            (
+                current_time - 10,
+                vec![quick_scope(Some(current_time + 11))],
+                true,
+            ),
+            // Case: Token inactive (future `active_at`)
+            (
+                current_time + 10,
```
chris-olszewski, line 134: I realize that this is unlikely, but I think we should proactively bump this threshold since there's a chance this could fail if it takes 10ms or more to perform the previous 4 test cases.

Also, I don't think this affects us, but the big "SystemTime is not monotonic" warning on the [docs](https://doc.rust-lang.org/std/time/struct.SystemTime.html#) makes a little concerned about some possible drift that causes these tests fail.

crates/turborepo-auth/src/lib.rs
```diff
@@ -12,19 +12,139 @@ mod ui;
...
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("Time went backwards")
+        .as_millis()
+}
+
+// As of the time of writing, this should always be true, since a token that
+// isn't active returns an error when fetching metadata for the token.
+fn is_token_active(metadata: &ResponseTokenMetadata) -> bool {
+    let current_time = current_unix_time();
```
chris-olszewski, line 72: Passing in the current time makes this a pure function which might make testing a little more reliable.
```suggestion
fn is_token_active(metadata: &ResponseTokenMetadata, current_time: u128) -> bool {
```

chris-olszewski (approved):
Just some minor suggestions

crates/turborepo-auth/src/lib.rs
Zertsov: yeah I've been experimenting with different ways to write tests, and originally this was a `tokio::test` which I wasn't sure of how it'd interact with `test_case`, so I went back to my old Go route. I'll probably make this use `test_case` in a future PR

crates/turborepo-auth/src/lib.rs
Zertsov: no problem here, I'll take your suggestion since I have no strong preference either way.

crates/turborepo-auth/src/lib.rs
Zertsov: I'll just do like, 1000 since this is a test that's supposed to be inactive.

The monotonic issue is valid, and I'm purely going off of vibes and instinct in saying that I don't think it'll be an issue. If it _does_ become an issue I can hard code something, but that feels a little more flakey / less exhaustive as a test

Zertsov:
### Merge activity

* **Feb 14, 4:51 PM EST**: @Zertsov merged this pull request with [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/7358).

PR #7368, opened 2024-02-13
fix(ecmascript): parse as module by default

kwonoj:
### Description

https://github.com/vercel/next.js/pull/61922

It looks like next-swc treats parse as module by default, follow the same practice.

Closes [PACK-2460](https://linear.app/vercel/issue/PACK-2460)

Closes PACK-2483

### DIFF
--- a/crates/turbopack-ecmascript/src/parse.rs
+++ b/crates/turbopack-ecmascript/src/parse.rs
@@ -303,7 +303,7 @@ async fn parse_content(
                 );
 
                 let mut parser = Parser::new_from(lexer);
-                let program_result = parser.parse_program();
+                let program_result = parser.parse_module().map(Program::Module);
 
                 let mut has_errors = false;
                 for e in parser.take_errors() {


crates/turbopack-ecmascript/src/parse.rs
```diff
@@ -303,7 +303,7 @@ async fn parse_content(
                 );
 
                 let mut parser = Parser::new_from(lexer);
-                let program_result = parser.parse_program();
+                let program_result = parser.parse_module().map(Program::Module);
```
sokra, line 306: Parse it as program and convert it to a script instead.

Why? CommonJs modules might contain `with()` statements which can only be parsed as script and not as module

kwonoj: the purpose of this change is let next-swc's visitor runs for all script treated one by parse_as_program. If we parse as script, that'll still fail.

so in next-swc we definitely parses everything as module -> let visitor runs, if turbopack might need different internals we need to update offending visitors instead.

PR #7378, opened 2024-02-13
chore: add force flag to login command

Zertsov:
Closes TURBO-2355

### DIFF
--- a/crates/turborepo-auth/src/auth/login.rs
+++ b/crates/turborepo-auth/src/auth/login.rs
@@ -27,26 +27,30 @@ pub async fn login<T: Client + TokenClient>(options: &LoginOptions<'_, T>) -> Re
         ui,
         login_url: login_url_configuration,
         login_server,
-        sso_team: _,
         existing_token,
+        force,
+        sso_team: _,
     } = *options; // Deref or we get double references for each of these
 
     // Check if passed in token exists first.
-    if let Some(token) = existing_token {
-        if Token::existing(token.to_string())
-            .is_valid(api_client)
-            .await?
-        {
-            return check_user_token(token, ui, api_client, "Existing token found!").await;
+    if !force {
+        if let Some(token) = existing_token {
+            if Token::existing(token.to_string())
+                .is_valid(api_client)
+                .await?
+            {
+                return check_user_token(token, ui, api_client, "Existing token found!").await;
+            }
         }
-    }
 
-    // If the user is logging into Vercel, check for an existing `vc` token.
-    if login_url_configuration.contains("vercel.com") {
-        // The extraction can return an error, but we don't want to fail the login if
-        // the token is not found.
-        if let Ok(token) = extract_vercel_token() {
-            return check_user_token(&token, ui, api_client, "Existing Vercel token found!").await;
+        // If the user is logging into Vercel, check for an existing `vc` token.
+        if login_url_configuration.contains("vercel.com") {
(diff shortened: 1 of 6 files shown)


Zertsov:
* `main`
  * **#7378** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7378?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

mehulkar (commented):
Is this ready to review? Does the flag do anything? Also looks like it's a global flag, should be just for the login command? (I think that's what you had mentioned you were working on)

Zertsov:
Nope, meant to make this a draft PR but hit the wrong button  🙃

turborepo-tests/integration/tests/turbo-help.t
```diff
@@ -263,6 +263,7 @@ Test help flag for login command
   Options:
         --sso-team <SSO_TEAM>             
         --version                         
+    -f  --force
```
mehulkar, line 266: Can we add a description here? Something like: "Overwrite any existing turbo or vercel tokens"

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -27,26 +27,30 @@ pub async fn login<T: Client + TokenClient>(options: &LoginOptions<'_, T>) -> Re
...
     } = *options; // Deref or we get double references for each of these
 
     // Check if passed in token exists first.
-    if let Some(token) = existing_token {
-        if Token::existing(token.to_string())
-            .is_valid(api_client)
-            .await?
-        {
-            return check_user_token(token, ui, api_client, "Existing token found!").await;
+    if !force {
```
mehulkar, line 36: I get really confused by negative if blocks, and almost always feel like they should be inverted. Not blocking at all, but if you feel like restructuring, something like this could be nice:

```rs
if force {
  return login()
}

if let Some(token) = existing_token {
  return "yay";
}

login();
```

mehulkar (commented):
⚠️ The only blocker from me is adding a description to the help output!

Just a thought: In this implementation, `--force` doesn't care about the validity of the token. From what I've seen in actual use cases, users only care about forcing an overwrite when their token is not working (they don't want to manually logout and then login again). 

Do you think an alternative implementation is something like a `--yes` flag and an interactive prompt when there's an existing invalid token?

```bash
turbo login --yes
Invalid token found, overwriting
Logged in!
```

```bash
turbo login # no flag
> Invalid token found, overwrite? [Y|n]
Logged in!
```

turborepo-tests/integration/tests/turbo-help.t
gsoltis: assuming the following is correct, how about "Force a login to receive a new token. Will overwrite any existing tokens for the given login url".

This is not a vercel-specific feature, it should work with any backend.

gsoltis (approved):

crates/turborepo-auth/src/auth/login.rs
Zertsov: I'm down to do something like this, but I'll make it in another stack because this'll bork / change up the tests

Zertsov:
### Merge activity

* **Feb 16, 4:01 PM EST**: @Zertsov merged this pull request with [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/7378).

PR #7380, opened 2024-02-14
feat(Turborepo): Add musl support for @turbo/repository

gsoltis:
### Description

Add support for x86_64 and aarch64 musl for `@turbo/repository`. 

### Testing Instructions

Dry run working on personal fork. Still needs validation on a real release

Closes TURBO-2278


.github/workflows/turborepo-library-release.yml
```diff
@@ -38,6 +38,29 @@ jobs:
...
+          - host: ubuntu-latest
+            container: ghcr.io/napi-rs/napi-rs/nodejs-rust:stable-2023-09-17-alpine
+            install: |
+              apk add libc6-compat curl
+              echo /root/.cargo/bin >> ${GITHUB_PATH}
+              echo /usr/local/cargo/bin/rustup >> ${GITHUB_PATH}
+              echo /aarch64-linux-musl-cross/bin >> ${GITHUB_PATH}
+              export PATH=/aarch64-linux-musl-cross/bin:/usr/local/cargo/bin/rustup:/root/.cargo/bin:${PATH}
+            setup: |
+              export PATH=/aarch64-linux-musl-cross/bin:/usr/local/cargo/bin/rustup:/root/.cargo/bin:${PATH}
```
tknickman, line 57: Huh, still strange it's required to fix up the PATH differences between local / github actions

.github/workflows/turborepo-library-release.yml
```diff
@@ -38,6 +38,29 @@ jobs:
               npm i -g pnpm@8.9.0
             setup: |
               pnpm install
+          - host: ubuntu-latest
+            container: ghcr.io/napi-rs/napi-rs/nodejs-rust:stable-2023-09-17-alpine
```
tknickman, line 42: Ah interesting, out of curiosity how did you find these?

tknickman (approved):

chris-olszewski (commented):
I think the binding file needs to get updated as we still have the logic that throws if it detects we're on musl: https://github.com/vercel/turbo/blob/ff6cff8ab335701163523eb00ad605c4b613151b/packages/turbo-repository/js/index.js#L89

gsoltis:
> I think the binding file needs to get updated as we still have the logic that throws if it detects we're on musl:
> 
> https://github.com/vercel/turbo/blob/ff6cff8ab335701163523eb00ad605c4b613151b/packages/turbo-repository/js/index.js#L89

Yup, added. 

Worth noting that while I've tested directly loading the `.node` file on alpine docker images, we'll want to do a post-release QA step to make sure the whole thing hangs together on the images we want to support.

.github/workflows/turborepo-library-release.yml
gsoltis: Copied them from [Next.js](https://github.com/vercel/next.js/blob/canary/.github/workflows/build_and_deploy.yml#L175)

.github/workflows/turborepo-library-release.yml
gsoltis: Biggest struggle was PATH differences. It's possible I'm doing something wrong, but as best I can tell, the PATH when running via GH Actions is different from the PATH running the same docker image on my machine. So I went for overkill in making sure I could run the binaries I needed.

PR #7398, opened 2024-02-15
add support for dynamic requests in require() and import()

sokra:
### Description

Refactors `PatternMatching` to support a map of multiple matches by key. Based on the new `RequestKey` support for resolving.

Code Generation for `require("./dir/"+ expr)` will emit a map of `__turbopack_lookup__({ "./a": () => ..., "./b": () => ... }, "./dir/" + expr)`.

The output format isn't really optimized yet, that's work for future PRs.

### Testing Instructions

added a test case, enabled one previously skipped test case

Closes PACK-2512


ForsakenHarmony (approved):

crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -278,27 +263,33 @@ impl CodeGenerateable for CjsRequireResolveAssetReference {
...
-                // Due to eventual consistency the path might match something else,
-                // but we can ignore that as it will be recomputed anyway.
-            }));
-        }
+        // Inline the result of the `require.resolve` call as a literal.
+        visitors.push(create_visitor!(path, visit_mut_expr(expr: &mut Expr) {
+            if let Expr::Call(call_expr) = expr {
+                let args = std::mem::take(&mut call_expr.args);
+                *expr = match args.into_iter().next() {
+                    Some(ExprOrSpread { expr, spread: None }) => pm.create_require(*expr),
```
ForsakenHarmony, line 271: Unfortunately I just noticed this now, but this looks very wrong, this should insert a string, not a require call

Issue #7395, opened 2024-02-15
[VS Code Extension] `turbo.path` setting not used to run turborepo


alan-nf:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

n/a

### What package manager are you using / does the bug impact?

npm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

n/a

### Describe the Bug

The VS Code extension (published 2024-01-25, 22:07:10) let's you run pipelines directly from turbo.json:

<img width="193" alt="image" src="https://github.com/vercel/turbo/assets/100322190/0c21b324-3a82-47dc-8d26-5b6fcd065463">

This results in the command `turbo run graphql:introspect` being run, regardless of what the `"turbo.path"` setting is configured to. Since I don't have a global `turbo` command, this fails for me:

<img width="221" alt="image" src="https://github.com/vercel/turbo/assets/100322190/1b70a7bf-db13-4e76-8d24-604e859b64ae">

### Expected Behavior

I would expect the "Run graphql:introspect" button to run the turbo I have in `turbo.path`. So, if I have this config setting in .vscode/settings.json:

`"turbo.path": "./node_modules/.bin/turbo"`

I would expect the button to run:

`./node_modules/.bin/turbo run graphql:introspect`

### To Reproduce

1. Install VSCode turborepo extension: https://marketplace.visualstudio.com/items?itemName=Vercel.turbo-vsc
2. Create a .vscode/settings.json and set it to:
```json
{
  "turbo.path": "./node_modules/.bin/turbo"
}
```
4. Open your turbo.json
5. Click the "Run ..." button above any task:
<img width="193" alt="image" src="https://github.com/vercel/turbo/assets/100322190/36e52866-0920-4e95-b07a-217425a39f12">

arlyon:
Nice catch. PR up now!

alan-nf:
Fix confirmed, ty!

Issue #7394, opened 2024-02-15
[VS Code Extension] Hints shouldn't be problems


alan-nf:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

n/a

### What package manager are you using / does the bug impact?

npm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

n/a

### Describe the Bug

The VS Code extension (published 2024-01-25, 22:07:10) shows helpful tooltips as problems:

<img width="638" alt="image" src="https://github.com/vercel/turbo/assets/100322190/2fd83728-d007-4c32-8045-25ced4a19953">

I get the desire to show tooltips but surfacing them as problems clutters the UI and makes me always have active problems I can't resolve in the problems pane:

<img width="703" alt="image" src="https://github.com/vercel/turbo/assets/100322190/e458f24a-6b33-42a4-9832-df5b2652c1ea">

### Expected Behavior

The more semantic place to stick these is in inlay hints, which are bits of helpful text that can be rendered in the editor without declaring them as problems. [This is the API](registerInlayHintsProvider). [This is the feature documentation](https://code.visualstudio.com/docs/typescript/typescript-editing#_inlay-hints). Here's what it looks like:

![image](https://github.com/vercel/turbo/assets/100322190/10ee1e9a-a25f-4936-b511-74c54638e9c7)

Hot take: if inlay hints aren't good enough, this feature should probably just be removed from the extension. Cluttering up the Problems pane, with even helpful information, isn't worth it.

### To Reproduce

1. Install VSCode turborepo extension: https://marketplace.visualstudio.com/items?itemName=Vercel.turbo-vsc
2. Open turbo.json
3. Make sure the `build` task depends on `^build`. You should see the `^build` text get a problem from the extension.

arlyon:
Thanks for the detailed report. Point taken, the 'perma-clutter' in the problems view is unnecesarry. I actually didn't know that the informational diagnostics would be displayed as a 'problem' in VSCode. Seems a little contradictory.

I have added a PR that addresses this, and adds a few more goodies in the process. Thanks for taking the time to report!

alan-nf:
The HINT looks great. No issues in my problem pane and you are right that it's less noisy than inlays.

Thank you!

PR #7399, opened 2024-02-15
make turbo extension use `turboPath` for commands

arlyon:
### Description

Missed the turboPath override for this one. Also added some new features, and fixed another issue in the process:

- downgrade `^` lint to HINT
- detect missing packages in dependsOn
- detect circular dependsOn
- respect turboPath when running commands in the terminal

Closes https://github.com/vercel/turbo/issues/7394
Closes https://github.com/vercel/turbo/issues/7395

### Testing Instructions

- boot up LSP
- override the turboPath
- run a command by clicking on it
- notice it is using your overridden turbo to do so
- notice the `^` is downgraded to a hint
- notice that missing tasks in dependsOn are now reported

Closes TURBO-2373

### DIFF
--- a/crates/turborepo-lsp/src/lib.rs
+++ b/crates/turborepo-lsp/src/lib.rs
@@ -638,7 +638,7 @@ impl Backend {
                 p.workspaces.into_iter(),
                 iter::once(WorkspaceData {
                     package_json: repo_root.join_component("package.json"),
-                    turbo_json: root_turbo_json.exists().then(|| root_turbo_json),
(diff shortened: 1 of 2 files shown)


gsoltis (approved):

gsoltis:
It would be good to add some tests in here at some point

arlyon:
Agreed, I actually don't know what tools there are for integration tests for LSPs but I will look into it. Some snapshot testing for diagnostic out given some file in would probably be pretty language agnostic

PR #7403, opened 2024-02-15
fix(rustdoc): Reverting back to rust-1.0.2-nightly

NicholasLYang:
### Description

Turns out when I upgraded us, it broke the rustdoc deployment. Pinning to 1.0.2 for now.

### Testing Instructions

Closes TURBO-2384


Zertsov (approved):

ForsakenHarmony:
yay +1.4GB in node_modules

can we please just get rid of it?

PR #7406, opened 2024-02-16
fix(turbopack-core): don't resolve exports field in folders

ForsakenHarmony:
### Description

only main fields should be resolved in folders, the exports field should only be resolved for module requests

Closes PACK-2524

### DIFF
--- a/crates/turbopack-core/src/resolve/mod.rs
+++ b/crates/turbopack-core/src/resolve/mod.rs
@@ -7,7 +7,7 @@ use std::{
     pin::Pin,
 };
 
-use anyhow::{anyhow, bail, Result};
+use anyhow::{bail, Result};
 use indexmap::{indexmap, IndexMap, IndexSet};
 use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
@@ -1563,7 +1563,7 @@ async fn resolve_internal_inline(
                         }
                         PatternMatch::Directory(matched_pattern, path) => {
                             results.push(
-                                resolve_into_folder(*path, options, *query)
+                                resolve_into_folder(*path, options)
                                     .with_request(matched_pattern.clone()),
                             );
                         }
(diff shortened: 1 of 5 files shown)


crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1709,29 +1709,12 @@ async fn resolve_internal_inline(
...
-                let request = Request::parse(Value::new(str.into()));
-                return Ok(resolve_internal_inline(
-                    package_path,
-                    request.resolve().await?,
-                    options,
-                )
-                .await?
-                .with_request(".".to_string()));
-            }
             ResolveIntoPackage::MainField {
```
sokra, line 1718: It's a bit weird that the `into_package` is applied here. Do we need a separate resolve option `into_folder` next to `into_package` to set resolving into packages independent from packages?

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -2007,63 +1977,74 @@ async fn resolve_into_package(
...
+                    todo!("pattern into an exports field is not implemented yet");
+                };
+
+                let path = if path == "/" {
+                    ".".to_string()
+                } else {
+                    format!(".{path}")
+                };
+
+                println!("path: {}", path);
```
sokra, line 2010: println

sokra (commented):
Might be worth to run the PR against the next.js test suite

crates/turbopack-core/src/resolve/mod.rs
ForsakenHarmony: I don't think so as `MainField` would have to be in both?

`into_package` being a list doesn't really make sense to me though, it could probably just be a struct with 2 fields (also getting rid of the enum)

ForsakenHarmony:
@sokra the next.js side has existed for a while https://github.com/vercel/next.js/pull/62166

sokra (approved):

PR #7407, opened 2024-02-16
fix: More robust token validation

Zertsov:
### Description
Some more modularization and better checking of validity for tokens.

### Testing Instructions
Try logins as normal, then login while logged in. If you can expire a token manually, try logging in again and it should do the normal login route again.

Closes TURBO-2391

### DIFF
--- a/crates/turborepo-auth/src/auth/login.rs
+++ b/crates/turborepo-auth/src/auth/login.rs
@@ -4,13 +4,10 @@ pub use error::Error;
 use reqwest::Url;
 use tokio::sync::OnceCell;
 use tracing::warn;
-use turborepo_api_client::{Client, TokenClient};
-use turborepo_ui::start_spinner;
+use turborepo_api_client::{CacheClient, Client, TokenClient};
+use turborepo_ui::{start_spinner, BOLD, UI};
 
-use crate::{
-    auth::{check_user_token, extract_vercel_token},
-    error, ui, LoginOptions, Token,
-};
+use crate::{auth::extract_vercel_token, error, ui, LoginOptions, Token};
 
 const DEFAULT_HOST_NAME: &str = "127.0.0.1";
 const DEFAULT_PORT: u16 = 9789;
@@ -21,7 +18,9 @@ const DEFAULT_PORT: u16 = 9789;
 ///
 /// First checks if an existing option has been passed in, then if the login is
 /// to Vercel, checks if the user has a Vercel CLI token on disk.
-pub async fn login<T: Client + TokenClient>(options: &LoginOptions<'_, T>) -> Result<Token, Error> {
+pub async fn login<T: Client + TokenClient + CacheClient>(
+    options: &LoginOptions<'_, T>,
+) -> Result<Token, Error> {
     let LoginOptions {
         api_client,
         ui,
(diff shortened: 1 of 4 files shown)


Zertsov:
* `main`
  * **#7407** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7407?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -267,6 +293,60 @@ mod tests {
         }
     }
 
+    #[async_trait]
```
arlyon, line 296: Perhaps out of scope for this PR but as of rust 1.75 we don't need async_trait any more. If you need to specify things like `Send` (the default for `async_trait` traits) you can use return position `impl` instead.

```rust
async fn get_artifact(...) -> Result<Option<Response>, turborepo_api_client::Error>
// or
fn get_artifact(...) -> impl Future<Output=Result<Option<Response>, turborepo_api_client::Error>> + Send
```

The side effect being no more `Pin<Box<dyn Future>>` :)

See here: https://github.com/vercel/turbo/blob/main/crates/turborepo-repository/src/discovery.rs#L43-L57

crates/turborepo-auth/src/auth/sso.rs
```diff
@@ -41,31 +38,53 @@ pub async fn sso_login<'a, T: Client + TokenClient>(
...
+    // In the future I want to make the Token have some non-skewable information and
+    // be able to get rid of this, but it works for now.
+    let valid_token_callback = |message: &str, ui: &UI| {
+        let message = message.to_string();
+        let ui = *ui;
+        move |user_email: &str| {
+            println!("{}", ui.apply(BOLD.apply_to(message)));
+            ui::print_cli_authorized(user_email, &ui);
+        }
+    };
```
chris-olszewski, line 54: I think this might be too clever as opposed to having `is_valid` return a `Result<Option<String>>` which contains the email if the token is valid and `None` if it isn't valid.

crates/turborepo-auth/src/lib.rs
```diff
@@ -31,17 +38,143 @@ impl Token {
...
+        client: &T,
+        // Making this optional since there are cases where we don't want to do anything after
+        // validation.
+        // A callback that gets called if the token is valid. This will be
+        // passed in a user's email if the token is valid.
+        valid_message_fn: Option<impl FnOnce(&str)>,
+    ) -> Result<bool, Error> {
+        let is_active = self.is_active(client).await?;
+        let has_cache_access = self.has_cache_access(client, None).await?;
+        if !is_active || !has_cache_access {
```
chris-olszewski, line 67: I think we can just do these at the same time
```suggestion
        let (is_active, has_cache_access) = tokio::try_join!(self.is_active(client), self.has_cache_access(client, None))?;
        if !is_active || !has_cache_access {
```

crates/turborepo-auth/src/lib.rs
```diff
@@ -31,17 +38,143 @@ impl Token {
...
+    /// 2. If the token has access to the cache.
+    ///     - If the token is forbidden from accessing the cache, we consider it
+    ///       invalid.
+    /// 3. We are able to fetch the user associated with the token.
+    ///
+    /// ## Arguments
+    /// * `client` - The client to use for API calls.
+    /// * `valid_message_fn` - An optional callback that gets called if the
+    ///   token is valid. It will be passed the user's email.
+    // TODO(voz): This should do a `get_user` or `get_teams` instead of the caller
```
chris-olszewski, line 52: Does this TODO still apply now that we're calling `self.user`?

chris-olszewski (approved):
Nothing blocking, just a few thoughts.

crates/turborepo-auth/src/lib.rs
Zertsov: Nope, missed this before merging, so I'll clean it up somewhere in this stack.

PR #7408, opened 2024-02-16
feat: More errors work

NicholasLYang:
### Description

Improved error printing a little bit.

- Ported some engine/package graph errors around task dependencies. Involved threading through span info into the engine. We have to be a little careful to not hash `Span<TaskDefinition>` because sometimes the tasks are equal, but the span info is *not*.
- Added a file name to the error output. Right now we only use error output for turbo.json but eventually we might show errors in package.json

### Testing Instructions

Closes TURBO-2393

### DIFF
--- a/docs/pages/messages/missing-root-task-in-turbo-json.mdx
+++ b/docs/pages/messages/missing-root-task-in-turbo-json.mdx
@@ -7,8 +7,7 @@ description: Missing root task in turbo.json error
 
 ## Why This Error Occurred
 
-We call the scripts defined in the monorepo's root `package.json` the root tasks. These tasks often call `turbo`
-themselves. For example:
+Root tasks are the scripts defined in the monorepo's root `package.json`. These tasks often call `turbo`. For example:
 
 ```json package.json
 {
@@ -18,11 +17,8 @@ themselves. For example:
 }
 ```
 
-This creates a problem when we declare task dependencies with `^`. A quick recap, `^<task>` means all of the `<task>`
-entries in the package's dependencies, e.g. if package `a` depends on packages `b` and `c`, and `a#build` depends on
-`^build`, then its dependencies will be `b#build` and `c#build`. We call these topological dependencies. Topological
-dependencies work well for many common situations, such as when you want to build your dependencies before building
-your own package.
+This creates a problem when we declare [topological dependencies](/repo/docs/reference/configuration#dependson). Topological
+dependencies specify that your package's dependencies should execute their tasks before your package executes its own task.
 
 ```json turbo.json
 {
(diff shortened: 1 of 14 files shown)


docs/pages/messages/missing-root-task-in-turbo-json.mdx
```diff
@@ -0,0 +1,57 @@
+---
+title: Missing root task in turbo.json
+description: Missing root task in turbo.json error
+---
+
+# Missing root task in turbo.json
+
```
NicholasLYang, line 7: This is incomplete. I'm a little stuck with explaining topological dependencies and why the root package is always an implicit dependency in a monorepo (cc @anthonyshew)

Zertsov (approved):

docs/pages/messages/missing-root-task-in-turbo-json.mdx
```diff
@@ -0,0 +1,62 @@
...
+## Why This Error Occurred
+
+We call the scripts defined in the monorepo's root `package.json` the root tasks. These tasks often call `turbo`
```
anthonyshew, line 10: Thinking it might help to invert the phrasing here, e.g. `"Root tasks" are the scripts defined in your workspace's root package.json.` The first time I read it, it felt awkward.

docs/pages/messages/missing-root-task-in-turbo-json.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+```json package.json
+{
+  "scripts": {
+    "build": "turbo run build"
+  }
+}
+```
+
+This creates a problem when we declare task dependencies with `^`. A quick recap, `^<task>` means all of the `<task>`
```
anthonyshew, line 21: I'd prefer to link to where we explain this in the documentation...Crossing my fingers, do we have that somewhere?

docs/pages/messages/missing-root-task-in-turbo-json.mdx
```diff
@@ -0,0 +1,62 @@
...
+{
+  "pipeline": {
+    "build": {
+      "dependsOn": ["^build"]
+    }
+  }
+}
+```
+
+However, following package dependency relations is slightly problematic because the root package is an implicit dependency
```
anthonyshew, line 37: Let's keep it a little more terse here in general.

"However, following package dependencies can be problematic because the root package is a part of the dependency graph for all package inside your workspace per JavaScript workspace conventions. Therefore, we don't execute root tasks by default for convenience so you can still write "turbo run build" in your root "build" script."

That should be closer but feel free to edit if there's more clarity to be had (keeping brevity in mind).

anthonyshew (commented):
Some suggestions!

docs/pages/messages/missing-root-task-in-turbo-json.mdx
NicholasLYang: Yeah we do, but imo it's not the greatest explanation: https://turbo.build/repo/docs/reference/configuration#dependson

docs/pages/messages/missing-root-task-in-turbo-json.mdx
NicholasLYang: Hmm, "part of the dependency graph" is a little confusing imo. Maybe "the root package is always a dependency for all packages inside your workspace"

PR #7421, opened 2024-02-20
fix(turbopack): Use different title for non-parsing issues

kdy1:
### Description

We shared `Handler` instance, and thus we were using `Parsing ecmascript source code failed` even for non-parsing issues.

### Testing Instructions

next.js counterpart: https://github.com/vercel/next.js/pull/62259

Closes PACK-2538


crates/turbopack-ecmascript/src/parse.rs
```diff
@@ -250,7 +250,7 @@ async fn parse_content(
         Box::new(IssueEmitter {
             source,
             source_map: source_map.clone(),
-            title: Some("Parsing ecmascript source code failed".to_string()),
+            title: Some("Input ecmascript file had an error".to_string()),
```
sokra, line 253:
```suggestion
            title: Some("Ecmascript file had an error".to_string()),
```

It might be transformed, not sure if it's an "input" file.

sokra (approved):

crates/turbopack-ecmascript/src/parse.rs
```diff
@@ -319,7 +329,7 @@ async fn parse_content(
                 match program_result {
                     Ok(parsed_program) => parsed_program,
                     Err(e) => {
-                        e.into_diagnostic(&handler).emit();
+                        e.into_diagnostic(&parser_handler).emit();
                         return Ok(ParseResult::Unparseable);
                     }
                 }
```
sokra, line 335: There is this:

```
            if handler.has_errors() {
                return Ok(ParseResult::Unparseable);
            }
```

later in this file that probably also need an update

PR #7424, opened 2024-02-20
fix(turbopack): Analyze for-of stmts to allow using `sharp`

kdy1:
### Description

Analyze code like

```js

const paths = [
  `../src/build/Release/sharp-${runtimePlatform}.node`,
  '../src/build/Release/sharp-wasm32.node',
  `@img/sharp-${runtimePlatform}/sharp.node`,
  '@img/sharp-wasm32/sharp.node'
];

let sharp;
const errors = [];
for (const path of paths) {
  try {
    sharp = require(path);
    break;
  } catch (err) {
    /* istanbul ignore next */
    errors.push(err);
  }
}

```

### Testing Instructions

Look at test changes. 

 - Closes https://github.com/vercel/next.js/issues/62088
 - Closes PACK-2521


crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -1410,6 +1410,37 @@ impl VisitAstPath for Analyzer<'_> {
...
+        &mut self,
+        n: &'ast ForOfStmt,
+        ast_path: &mut swc_core::ecma::visit::AstNodePath<'r>,
+    ) {
+        let mut array = self.eval_context.eval(&n.right);
+        if let JsValue::Variable(id) = &array {
+            if let Some(value) = self.data.values.get(id) {
+                array = value.clone();
+            }
+        }
```
sokra, line 1423: You shouldn't read from `values` here. This should only happen during `linking`. Imagine this example:

``` js
let y;

x();

for(const z of y) {

}


function x() {
  y = [1,2,3];
}
```

In this case `values` is not set yet while visiting the for of loop.

---

Instead you need to keep the `Variable` and have some operation that converts an array to all iterated values of it: `JsValue::Iterated`.
Now you can define a built-in transform for `JsValue::Iterated(JsValue::Array(...))` to `JsValue::Alternatives`.

crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -1410,6 +1410,37 @@ impl VisitAstPath for Analyzer<'_> {
...
+                self.visit_for_head(&n.left, &mut ast_path);
+            }
+
+            let mut ast_path =
+                ast_path.with_guard(AstParentNodeRef::ForOfStmt(n, ForOfStmtField::Body));
+
+            self.visit_stmt(&n.body, &mut ast_path);
+        } else {
+            n.visit_children_with_path(self, ast_path);
+        }
```
sokra, line 1441: We don't want to visit the AST nodes multiple times. Just once.

crates/turbopack-ecmascript/tests/analyzer/graph/pack-2521/graph-explained.snapshot
```diff
@@ -0,0 +1,27 @@
...
+  ⚠️  pattern without value
+
+errors = []
+
+path = (
+  | `../src/build/Release/sharp-${runtimePlatform}.node`
+  | "../src/build/Release/sharp-wasm32.node"
+  | `@img/sharp-${runtimePlatform}/sharp.node`
+  | "@img/sharp-wasm32/sharp.node"
+)
```
sokra, line 12: It's unexpected that it already has this value here. It should have some kind of operation of `paths`. Maybe `path = Iterated(paths)`

sokra (changes requested):

crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -539,6 +539,16 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
             }
             None => false,
         },
+
+        JsValue::Iterated(_, iterable) => {
+            if let JsValue::Array { items, .. } = &mut **iterable {
+                *value = JsValue::alternatives(take(items).into_iter().collect());
```
sokra, line 545:
```suggestion
                *value = JsValue::alternatives(take(items));
```

crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -1410,6 +1410,26 @@ impl VisitAstPath for Analyzer<'_> {
         }
     }
 
+    fn visit_for_of_stmt<'ast: 'r, 'r>(
+        &mut self,
+        n: &'ast ForOfStmt,
+        ast_path: &mut swc_core::ecma::visit::AstNodePath<'r>,
+    ) {
+        let array = self.eval_context.eval(&n.right);
+
```
sokra, line 1419: We need to visit `n.right` too. There might be more complex code in that expression

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -626,6 +631,7 @@ impl Display for JsValue {
             JsValue::Argument(func_ident, index) => {
                 write!(f, "arguments[{}#{}]", index, func_ident)
             }
+            JsValue::Iterated(_, iterable) => write!(f, "for (let _ of {})", iterable),
```
sokra, line 634:
```suggestion
            JsValue::Iterated(_, iterable) => write!(f, "Iterated({})", iterable),
```


It need to look like an expression for formatting

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -1286,6 +1305,12 @@ impl JsValue {
                 "!({})",
                 value.explain_internal_inner(hints, indent_depth, depth, unknown_depth)
             ),
+            JsValue::Iterated(_, iterable) => {
+                format!(
+                    "for (let _ of {})",
```
sokra, line 1310:
```suggestion
                    "Iterated({})",
```

Should read like an expression

sokra (approved):

PR #7429, opened 2024-02-20
Update tsconfig.json extends path

alisouran:


anthonyshew:
Actually, this file isn't used at all! Which explains why I missed updating it. Can you delete it instead? 😄

anthonyshew:
This PR got superceded! Sorry about that. Thank you for wanting to contribute and hope to see you again!

PR #7431, opened 2024-02-20
fix(turbopack): Fix `pure` lint for CSS Modules in lightningcss mode

kdy1:
### Description

Fix `pure` lint in `lightningcss` mode. `lightningcss` does not provide location information of selectors, so we cannot provide source location. That's limitation of `cssparser` and `selectors` from `servo` project. It was the reason I didn't use them while building `swc_css`. As a last resort, we can use `Debug` implementation to print the selectors.

### Testing Instructions

Trying it on the internal apps is the best way.

Closes PACK-2546

### DIFF
--- a/crates/turbopack-css/src/process.rs
+++ b/crates/turbopack-css/src/process.rs
@@ -695,19 +695,19 @@ impl lightningcss::visitor::Visitor<'_> for CssModuleValidator {
         visit_types!(SELECTORS)
     }
 
-    fn visit_selector_list(
+    fn visit_selector(
         &mut self,
-        selector: &mut lightningcss::selector::SelectorList<'_>,
+        selector: &mut lightningcss::selector::Selector<'_>,
     ) -> Result<(), Self::Error> {
-        if selector.0.iter().all(|sel| {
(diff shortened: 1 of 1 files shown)


crates/turbopack-css/src/process.rs
```diff
@@ -699,17 +699,18 @@ impl lightningcss::visitor::Visitor<'_> for CssModuleValidator {
...
-            parcel_selectors::parser::Component::ID(_)
-            | parcel_selectors::parser::Component::Class(_) => true,
-            parcel_selectors::parser::Component::LocalName(local) => {
-                !matches!(&*local.name.0, "html" | "body")
-            }
-            _ => false,
-        }) {
+        if selector
+            .iter_raw_parse_order_from(0)
+            .all(|component| match component {
```
sokra, line 704: I think the condition isn't correct. It should check if it contains a class name or id somewhere in the selector. I feel it's not longer doing that

sokra (approved):
I don't think it's correct. And it misses test cases. But at least it's not longer reporting errors for every file

PR #7436, opened 2024-02-20
refactor(turbo-tasks-fs): move file watching into separate file

ForsakenHarmony:
### Description

Closes PACK-2549

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ members = [
   "crates/turbo-tasks*",
   "crates/turbopack*",
   "crates/turborepo*",
-  "packages/turbo-repository/rust/",
+  "packages/turbo-repository/rust",
   "xtask",
 ]
 
@@ -230,6 +230,9 @@ lightningcss = { version = "1.0.0-alpha.50", features = [
 ] }
 mime = "0.3.16"
 nohash-hasher = "0.2.0"
+notify = "6.1.1"
+notify-debouncer-full = "0.3.1"
+notify-debouncer-mini = { version = "0.3.0", default-features = false }
 once_cell = "1.17.1"
 owo-colors = "3.5.0"
 parcel_selectors = "0.26.0"
@@ -278,4 +281,3 @@ urlencoding = "2.1.2"
 webbrowser = "0.8.7"
 which = "4.4.0"
 unicode-segmentation = "1.10.1"
-notify-debouncer-full = "0.3.1"
--- a/crates/turbo-tasks-fs/src/lib.rs
+++ b/crates/turbo-tasks-fs/src/lib.rs
@@ -111,7 +111,6 @@ pub struct DiskFileSystem {
     #[serde(skip)]
     invalidation_lock: Arc<RwLock<()>>,
     #[turbo_tasks(debug_ignore, trace_ignore)]
-    #[serde(skip)]
     watcher: Arc<DiskWatcher>,
 }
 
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -230,6 +230,9 @@ lightningcss = { version = "1.0.0-alpha.50", features = [
 ] }
 mime = "0.3.16"
 nohash-hasher = "0.2.0"
+notify = "6.1.1"
+notify-debouncer-full = "0.3.1"
+notify-debouncer-mini = { version = "0.3.0", default-features = false }
```
ForsakenHarmony, line 235: added all the versions here to keep them in sync

kwonoj (approved):

sokra (changes requested):

crates/turborepo-filewatch/Cargo.toml
```diff
@@ -14,7 +14,7 @@ anyhow = { workspace = true }
 dashmap = { workspace = true }
 futures = { version = "0.3.26" }
 itertools = { workspace = true }
-notify = "6.0.1"
+notify = { workspace = true }
```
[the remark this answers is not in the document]
ForsakenHarmony: it does not?

sokra (approved):

crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -189,11 +113,6 @@ pub struct DiskFileSystem {
     #[turbo_tasks(debug_ignore, trace_ignore)]
     #[serde(skip)]
     watcher: Arc<DiskWatcher>,
-    /// Array of paths that should not notify invalidations.
-    /// `notify` currently doesn't support unwatching subpaths from the root,
-    /// so underlying we still watches filesystem event but only skips to
-    /// invalidate.
-    ignored_subpaths: Vec<PathBuf>,
```
sokra, line 196: moving that into `watcher` will include it in `serde(skip)` which results in looking `ignored_subpaths` after serialization

sokra (changes requested):

crates/turborepo-filewatch/Cargo.toml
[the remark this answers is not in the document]
sokra: Yep, you are right

crates/turborepo-globwatch/Cargo.toml
```diff
@@ -13,8 +13,8 @@ camino = "1.1.4"
 futures = { version = "0.3.26" }
 itertools.workspace = true
 merge-streams = "0.1.2"
-notify = "5.1"
```
sokra, line 16: This is 6.1.1 now

crates/turborepo-filewatch/Cargo.toml
```diff
@@ -14,7 +14,7 @@ anyhow = { workspace = true }
 dashmap = { workspace = true }
 futures = { version = "0.3.26" }
 itertools = { workspace = true }
-notify = "6.0.1"
```
sokra, line 17: This is 6.1.1 now

ForsakenHarmony: yes, no significant changes there

crates/turborepo-globwatch/Cargo.toml
ForsakenHarmony: the only breaking change is "files and directories moved into a watch folder on Linux will now be reported as rename to events instead of create events" which doesn't affect `turborepo-globwatch`

sokra (approved):

ForsakenHarmony:
The cache was actually the reason for the slowdown unfortunately

PR #7443, opened 2024-02-21
chore: move logout functionality into auth crate

Zertsov:
Closes TURBO-2405

### DIFF
--- a/crates/turborepo-auth/src/auth/logout.rs
+++ b/crates/turborepo-auth/src/auth/logout.rs
@@ -1,5 +1,127 @@
-use turborepo_ui::{cprintln, GREY, UI};
+use tracing::error;
+use turborepo_api_client::Client;
+use turborepo_ui::{cprintln, GREY};
 
-pub fn logout(ui: &UI) {
-    cprintln!(ui, GREY, ">>> Logged out");
+use crate::{Error, LogoutOptions};
+
+pub fn logout<T: Client>(options: &LogoutOptions<T>) -> Result<(), Error> {
+    if let Err(err) = remove_token(options) {
+        error!("could not logout. Something went wrong: {}", err);
+        return Err(err);
+    }
+
+    cprintln!(options.ui, GREY, ">>> Logged out");
+    Ok(())
+}
+
+fn remove_token<T: Client>(options: &LogoutOptions<T>) -> Result<(), Error> {
+    // Read the existing content from the global configuration path
+    let content = options.path.read_to_string()?;
+
+    // Attempt to deserialize the content into a serde_json::Value
+    let mut data: serde_json::Value = serde_json::from_str(&content)?;
+
+    // Check if the data is an object and remove the "token" field if present
+    if let Some(obj) = data.as_object_mut() {
+        if obj.remove("token").is_none() {
+            return Ok(());
+        }
+    } else {
+        return Ok(());
+    }
+
+    // Serialize the updated data back to a string
+    let new_content = serde_json::to_string_pretty(&data)?;
+
+    // Write the updated content back to the file
+    options.path.create_with_contents(new_content)?;
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use async_trait::async_trait;
+    use reqwest::{RequestBuilder, Response};
+    use tempfile::tempdir;
+    use turbopath::AbsoluteSystemPathBuf;
+    use turborepo_ui::UI;
+    use turborepo_vercel_api::{
+        SpacesResponse, Team, TeamsResponse, UserResponse, VerifiedSsoUser,
+    };
+    use url::Url;
+
+    use super::*;
+
+    struct MockApiClient {}
+
+    impl MockApiClient {
+        fn new() -> Self {
+            Self {}
+        }
+    }
+
+    #[async_trait]
+    impl Client for MockApiClient {
+        async fn get_user(&self, _token: &str) -> turborepo_api_client::Result<UserResponse> {
+            unimplemented!("get_user")
+        }
+        async fn get_teams(&self, _token: &str) -> turborepo_api_client::Result<TeamsResponse> {
+            unimplemented!("get_teams")
+        }
+        async fn get_team(
+            &self,
+            _token: &str,
+            _team_id: &str,
+        ) -> turborepo_api_client::Result<Option<Team>> {
+            unimplemented!("get_team")
+        }
+        fn add_ci_header(_request_builder: RequestBuilder) -> RequestBuilder {
+            unimplemented!("add_ci_header")
+        }
+        async fn get_spaces(
+            &self,
+            _token: &str,
+            _team_id: Option<&str>,
+        ) -> turborepo_api_client::Result<SpacesResponse> {
+            unimplemented!("get_spaces")
+        }
+        async fn verify_sso_token(
+            &self,
+            token: &str,
+            _: &str,
+        ) -> turborepo_api_client::Result<VerifiedSsoUser> {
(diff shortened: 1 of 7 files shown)


Zertsov:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/vercel/turbo/7443?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7444** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7444?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7443** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7443?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7407** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7407?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/turborepo-auth/src/auth/mod.rs
```diff
@@ -48,6 +43,15 @@ where
     }
 }
 
+/// Options for logging out.
+pub struct LogoutOptions<'a, T> {
+    pub ui: &'a UI,
+    pub api_client: &'a T,
+
+    /// The path where we should look for the token to logout.
+    pub path: &'a AbsoluteSystemPathBuf,
```
chris-olszewski, line 52: For borrowed/owned type pairs like `&[T]/Vec<T>`/`&str/String`/`&Path/PathBuf` you almost always want to use the borrowed variant instead of `&String`. The gist is that for all of these types the owned variant has a pointer to some heap allocated memory. So `&Vec<T>` is a pointer to `Vec` which is on the stack which then contains a pointer to the heap memory. If convert `&Vec<T>` to `&[u8]` you skip needing to resolve that first pointer and following it will take you directly to the heap memory.
```suggestion
    pub path: &'a AbsoluteSystemPath,
```

crates/turborepo-auth/src/auth/logout.rs
```diff
@@ -1,5 +1,34 @@
...
+        fs::read_to_string(global_config_path).map_err(|e| Error::FailedToReadAuthFilePath {
+            path: global_config_path.clone(),
+            error: e,
+        })?;
+
+    let mut data: serde_json::Value = serde_json::from_str(&content)?;
+    data["token"] = serde_json::Value::Null;
+
+    let new_content = serde_json::to_string_pretty(&data)?;
+    fs::write(global_config_path, new_content)?;
```
chris-olszewski, line 31: If we want to preserve previous behavior we should keep using `unset_path`. `serde` will reformat/reorder the contents of the JSON.
```suggestion
    if let Some(after) = unset_path(&before, &["token"], true)? {
        global_config_path.create_with_contents(after).map_err(|e| {
            Error::Config(config::Error::FailedToSetConfig {
                config_path: global_config_path.clone(),
                error: e,
            })
        })
    }
```

crates/turborepo-auth/src/auth/logout.rs
```diff
@@ -1,5 +1,34 @@
...
+    Ok(())
+}
+
+fn remove_token<T: Client>(options: &LogoutOptions<T>) -> Result<(), Error> {
+    let global_config_path = options.path;
+    let content =
+        fs::read_to_string(global_config_path).map_err(|e| Error::FailedToReadAuthFilePath {
+            path: global_config_path.clone(),
+            error: e,
+        })?;
```
chris-olszewski, line 25: I would suggest using just `global_config_path.read_to_string` as we use `fs_err` which automatically includes the path in the io errors.
```suggestion
    let content = global_config_path.read_to_string()?;
```

crates/turborepo-auth/src/auth/logout.rs
chris-olszewski, line 31: If we don't care about ordering and stick with using `serde` we should again use the turbopath variant
```suggestion
    global_config_path.create_with_contents(new_content)?;
```

crates/turborepo-auth/src/auth/logout.rs
```diff
@@ -1,5 +1,34 @@
...
+
+    let mut data: serde_json::Value = serde_json::from_str(&content)?;
+    data["token"] = serde_json::Value::Null;
```
chris-olszewski, line 28: This will panic if the JSON file isn't an object or `null`. We don't ever write this, but we don't want to panic if for some reason the auth file is corrupted.

chris-olszewski (changes requested):
I think there was some unintentional behavior change in this PR.

crates/turborepo-auth/src/auth/mod.rs
Zertsov: I always get confused which ASP is the borrowed variant vs the owned one. I've thought about if we could give these more straightforward names but haven't come up with anything - might be worth discussing as a team since I know Mehul also mentioned he got confused by these names

crates/turborepo-auth/src/auth/logout.rs
Zertsov: the problem with `unset_path` is that it lives in the crate that will call this one. I can move this same logic or similar logic into this crate or create a new one for file interactions, but otherwise this causes a cyclic import which makes Rust and myself sad. I could be importing incorrectly and maybe I can _just_ grab this module, but it still feels a little odd.

As for the contents, AFAIK we don't store anything _but_ the token in the `config.json`, so ordering / formatting won't be an issue at the moment. If we add more to the JSON then this would need to change, so I'm open to doing some of the previous suggestions.

crates/turborepo-auth/src/auth/mod.rs
chris-olszewski: They're based off of `&Path`/`PathBuf` from the [stdlib](https://doc.rust-lang.org/std/path/index.html).

crates/turborepo-auth/src/auth/logout.rs
```diff
@@ -1,5 +1,31 @@
...
+}
+
+fn remove_token<T: Client>(options: &LogoutOptions<T>) -> Result<(), Error> {
+    let global_config_path = options.path;
+    let content = global_config_path.read_to_string()?;
+
+    let mut data: serde_json::Value = serde_json::from_str(&content)?;
+    if let Some(token) = data.get_mut("token") {
+        // Set token field to empty string
+        *token = "".into();
```
chris-olszewski, line 24: I'm pretty sure we want `null` and not an empty string
```suggestion
        *token = Value::Null;
```

chris-olszewski (approved):
Only blocker is what we set the token to `null` and not the empty string.

I will note that this is a behavior change from previous behavior as the previous behavior would remove the token entirely e.g. `{"token": "foobar"}` -> `{}`. I think we treat a missing key and a key present with a value of `null` as the same, but it is probably worth double checking.

crates/turborepo-auth/src/auth/logout.rs
Zertsov: Wasn't 100% sure how all this interacts with `null` but in theory it's fine

chris-olszewski: > Wasn't 100% sure how all this interacts with null but in theory it's fine

That sounds like we need more testing.

Zertsov:
### Merge activity

* **Mar 1, 1:25 PM EST**: @Zertsov merged this pull request with [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/7443).

PR #7447, opened 2024-02-21
fix(turbopack): catch invalid source map paths

ForsakenHarmony:
### Description

Fixes the apollo `node_file_trace` test on Windows
It includes a source map with `"sourceRoot":"c:/Projects/decorators/deprecated/src/"` which would get joined to the current directory, leading to an invalid path on Windows.

Closes PACK-2562

### DIFF
--- a/crates/turbo-tasks-fs/src/lib.rs
+++ b/crates/turbo-tasks-fs/src/lib.rs
@@ -332,7 +332,9 @@ impl FileSystem for DiskFileSystem {
             .await
         {
             Ok(file) => FileContent::new(file),
-            Err(e) if e.kind() == ErrorKind::NotFound => FileContent::NotFound,
+            Err(e) if e.kind() == ErrorKind::NotFound || e.kind() == ErrorKind::InvalidFilename => {
+                FileContent::NotFound
+            }
             Err(e) => {
                 bail!(anyhow!(e).context(format!("reading file {}", full_path.display())))
             }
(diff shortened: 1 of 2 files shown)


crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -394,7 +394,11 @@ impl SourceMap {
                     let source = format!("/{SOURCE_MAP_ROOT_NAME}/{}", path_str);
                     let source_content = if let Some(source_content) = source_content {
                         source_content.to_string()
-                    } else if let FileContent::Content(file) = &*path.read().await? {
+                    } else if let Ok(FileContent::Content(file)) = path.read().await.as_deref() {
```
sokra, line 397: I'm a bit concerned that this might hide some real file reading failure. I would prefer to convert an invalid path error into a `FileContent::NotFound` in the disk filesystem impl, so we don't miss real errors

sokra (changes requested):

crates/turbopack-core/src/source_map/mod.rs
ForsakenHarmony: true, added a check for absolute paths to `turbo_tasks_fs::util::join_path`, let's see if that works

ForsakenHarmony:
Now one of the benchmarks is timing out 🙃

sokra:
We should really just check for the invalid path error on file operations. E. g. I guess it would also crash on `import "./some:folder/file.js"` or other paths with invalid chars.

sokra (approved):

PR #7468, opened 2024-02-21
chore: Change daemon log location to be in repo

NicholasLYang:
### Description

We were previously logging to a file outside the repo. This made it rather annoying to find, especially if the daemon is not running. Now we log to the `.turbo/daemon` folder inside the repository. Also, on `turbo daemon logs`, if the daemon isn't running, we fall back to looking for the latest file in `.turbo/daemon`.

### Testing Instructions

Closes TURBO-2417

### DIFF
--- a/crates/turborepo-lib/src/commands/daemon.rs
+++ b/crates/turborepo-lib/src/commands/daemon.rs
@@ -172,7 +172,6 @@ async fn get_log_file_from_folder(base: &CommandBase) -> Result<String, DaemonEr
     warn!("couldn't connect to daemon, looking for old log files");
     let log_folder = base.repo_root.join_components(&[".turbo", "daemon"]);
     let Ok(dir) = std::fs::read_dir(log_folder) else {
-        println!("could not find log file");
         return Err(DaemonError::LogFileNotFound);
     };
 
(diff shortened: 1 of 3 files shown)


crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -123,10 +123,14 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
             }
         }
         DaemonCommand::Logs => {
-            let mut client = connector.connect().await?;
-            let status = client.status().await?;
-            let log_file = log_filename(&status.log_file)?;
+            let log_file = if let Ok(log_file) = get_log_file_from_daemon(connector).await {
+                log_file
+            } else {
+                get_log_file_from_folder(base).await?
```
mehulkar, line 129: Do we need the fallback?

mehulkar (approved):
Looks fine. Is there ever a case for a daemon to use the same log file between projects? E.g. global turbo with no locally installed turbo

crates/turborepo-lib/src/commands/daemon.rs
NicholasLYang: I found it really annoying that you can't see logs when the daemon has crashed

mehulkar: I think I commented on the wrong line. I meant, do we need the global location ?

PR #7503, opened 2024-02-22
fix(turbopack-node): don't bundle postcss config

ForsakenHarmony:
### Description

It still gets bundled if it's not a JavaScript config, but if it is, we now import it via `__turbopack_external_import__` (`import()`).
We use `import()` instead of `require` to handle ESM configs (`.mjs`).

This is similar to `postcss-load-config`.

Closes PACK-2452

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11344,6 +11344,7 @@ dependencies = [
  "futures 0.3.28",
  "futures-retry",
  "indexmap 1.9.3",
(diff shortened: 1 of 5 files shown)


sokra (approved):

ForsakenHarmony:
Need to investigate the failing `test/development/basic/tailwind-jit.test.ts` test in next.js

PR #7505, opened 2024-02-23
feat(ecmascript): support to carry unparseable error messages

kwonoj:
### Description

I'm bit unsure if this would be feasible way to plug things in, but trying an initial attempt.

There are tests like https://github.com/vercel/next.js/blob/canary/test/development/middleware-errors/index.test.ts, expect to cli output displays error messages bubbles up. When it is errors like unparseable contents, turbopack replaces inner contents to runtime codes to throw error explicitly (https://github.com/vercel/turbo/blob/main/crates/turbopack-ecmascript/src/lib.rs#L871-L874) but without any details what error message encountered.

PR extends existing enum to allow to attach error messages to include in into the runtime codes, so at least we can have those printed. However due to we replace & throw in runtime codes, the overall stack trace (and the code location) is not the same as webpack.


crates/turbopack-ecmascript/src/parse.rs
```diff
@@ -376,8 +397,9 @@ async fn parse_content(
                     .await?;
             }
 
+            // seems already emitted errors are not readable except the count?
```
kdy1, line 400: You need to implement [`Emitter`](https://rustdoc.swc.rs/swc_common/errors/emitter/trait.Emitter.html), just like `IssueEmitter` if you want to read the issues.

kwonoj: Thanks, might going to try later - at least this path doesn't seem to be explicitly required in next.js

crates/turbopack-ecmascript/src/parse.rs
```diff
@@ -57,7 +57,9 @@ pub enum ParseResult {
         #[turbo_tasks(debug_ignore, trace_ignore)]
         source_map: Arc<swc_core::common::SourceMap>,
     },
-    Unparseable,
+    Unparseable {
+        messages: Option<Vec<String>>,
```
sokra, line 61:
```suggestion
        messages: Vec<String>,
```

sokra: A plain string feels a bit loosy. Could that be a `Vc<Issue>`?

kwonoj: Main purpose for this is to include in runtime code to throw - https://github.com/vercel/turbo/pull/7505/files#diff-40046925dbca8e1d654bd7bf2fb30362fd19391f9edd943fe019654e672833eeR872-R873 

which expects string already, not sure if it's worth string -> issue -> string again.

crates/turbopack-ecmascript/src/parse.rs
sokra: Maybe just store the list in the `IssueEmitter` and grab it from there

kdy1 (approved):

PR #7506, opened 2024-02-23
chore: remove unnecessary result on join_unix_path

chris-olszewski:
### Description

Realized that this shouldn't ever fail unless we bypass the checks done in the constructors for both `AbsoluteSystemPath` and `RelativeUnixPath`. The only failure point is converting the `PathBuf` to `AbsoluteSystemPathBuf` which itself has [two failure points](https://github.com/vercel/turbo/blob/main/crates/turborepo-paths/src/absolute_system_path_buf.rs#L208).
 - Conversion from stdlib `PathBuf` to `Utf8PathBuf` which only fails if the [path can't be represented in utf8](https://docs.rs/camino/latest/src/camino/lib.rs.html#156)
 - `AbsoluteSystemPathBuf` constructor which only fails if the input path isn't absolute

### Testing Instructions

Existing test suite, my reasoning above

### DIFF
--- a/crates/turborepo-lib/src/globwatcher/mod.rs
+++ b/crates/turborepo-lib/src/globwatcher/mod.rs
@@ -394,15 +394,12 @@ mod test {
 
         for dir in directories.iter() {
             let dir = RelativeUnixPathBuf::new(*dir).unwrap();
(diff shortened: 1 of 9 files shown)


NicholasLYang (approved):
Yeah I noticed this recently too. And we were already unwrapping in a lot of usages too.

PR #7507, opened 2024-02-23
fix(turbopack): Fix CSS module purity validator

kdy1:
### Description

This PR fixes purity validator for CSS Modules. This PR implements relevant logic for both `swc_css` and `lightningcss`.

Closes PACK-2573

### Testing Instructions

I refactored it to make it easier to test, and I added inline tests.


crates/turbopack-css/src/process.rs
```diff
@@ -902,3 +937,134 @@ impl Issue for ParsingIssue {
...
+            "div[data-foo=\"bar\"] {
+            color: red;
+        }",
+        );
+
+        assert_lint_failure(
+            "div[data-foo=\"bar\"] span {
+            color: red;
+        }",
+        );
```
sokra, line 1068: - `*`  ❌
- `[data-foo]` ❌
- `:not(.class)` ❌
- `.class:not(div)` ✅
- `:not(div)` ❌

sokra (approved):

Issue #7188, opened 2024-01-31
`turbo prune` is broken when not using pnpm workspace protocol for workspace packages


gegenschall:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

https://github.com/gegenschall/turbo-bug-repro

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

1.12.0

### Describe the Bug

When using a monorepo and linking packages **without** the use of the workspace protocol and setting `link-workspace-packages=false` in `.npmrc` the pruned output from `turbo prune` is broken. The generated `pnpm-lock.yaml` file does not include any valid references to the packages from the workspace, s.t. `pnpm install` will fail with the following message:

```
Scope: all 3 workspace projects
Lockfile is up to date, resolution step is skipped
 WARN  Broken lockfile: no entry for '/@gegenschall/some-package/1.0.0' in pnpm-lock.yaml
 ERR_PNPM_ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY  The lockfile is broken! Resolution step will be performed to fix it.
services/some-service                    |  WARN  Could not find preferred package /@gegenschall/some-package/1.0.0 in lockfile
```

For reference, here's the generated and original `pnpm-lock.yaml`:

<details>
  <summary>Pruned pnpm-lock.yaml</summary>

  ```yaml
lockfileVersion: '6.0'
settings:
  autoInstallPeerDeps: null
  excludeLinksFromLockfile: false
importers:
  .:
    devDependencies:
      ts-node:
        specifier: ^10.9.2
        version: 10.9.2(@types/node@20.11.13)(typescript@5.3.3)
      turbo:
        specifier: ^1.12.0
        version: 1.12.0
  packages/some-package: {}
  services/some-service:
    dependencies:
      '@gegenschall/some-package':
        specifier: ^1.0.0
        version: 1.0.0
packages:
  /@cspotcode/source-map-support@0.8.1:
    resolution:
      integrity: sha512-IchNf6dN4tHoMFIn/7OE8LWZ19Y6q/67Bmf6vnGREv8RSbBVb9LPJxEcnwrcwX6ixSvaiGoomAUvu4YSxXrVgw==
    dependencies:
      '@jridgewell/trace-mapping': 0.3.9
    dev: true
    engines:
      node: '>=12'
[...]

PR #7512, opened 2024-02-23
fix: support pnpm workspace path aliasing

### LINKED ISSUE #6744: Wrong dependent task gets run when using pnpm relative path reference (eg. `workspace:../path/to/dep`)
### Verify canary release

https://github.com/steveluscher/pnpm-alias-pathspec-repro.git

pnpm

Mac

1.11.1-canary.0

### Describe the Bug

Imagine that you:

1. Alias a package to a relative path (eg. `"luigi": "workspace:../waluigi"`)
2. Another package in your workspace has the same name as the alias (`luigi`)

…then you run some turborepo task.

```shell
pnpm i
cd packages/app/
pnpm turbo hello
```

Turborepo will run the task in the wrong workspace package; the `hello` script from `luigi/package.json` runs.

### Expected Behavior

Since the dependency `luigi` is aliased to `workspace:../waluigi`, I would expect the `hello` script from `waluigi/package.json` to run.

### To Reproduce

Follow the README in https://github.com/steveluscher/pnpm-alias-pathspec-repro.git.

### Additional context

> [!NOTE]
> If you change the dependency in `app/package.json` from `"luigi": "workspace:../waluigi"` to `"luigi": "workspace:waluigi@*"` it works as expected.

<sub>[TURBO-1899](https://linear.app/vercel/issue/TURBO-1899/wrong-dependent-task-gets-run-when-using-pnpm-relative-path-reference)</sub>

chris-olszewski:
### Description

Fixes #6744 

I highly suggest reviewing each commit on it's own as the code move done in the first commit makes the cumulative diff hard to understand. I wanted to clean up this code a bit as we'll need to extend it further for supporting https://github.com/vercel/turbo/issues/7188

This adds support for the PNPM feature of [aliasing a package via path](https://pnpm.io/workspaces#referencing-workspace-packages-through-their-relative-path). 

The slightly tricky part of this is that there are many valid specifiers for the `workspace:` protocol. We currently support `workspace:1.2.3`, `workspace:*`, [`workspace:bar@*`](https://pnpm.io/workspaces#referencing-workspace-packages-through-aliases), `workspace:../path/to/package`.

### Testing Instructions

Added unit tests for the case of a dependency of the form `"foo": "workspace:../bar"` where `foo` is another workspace package.

Verified fix works with repro provided in: https://github.com/vercel/turbo/issues/6744

### DIFF
--- a/crates/turborepo-repository/src/package_graph/dep_splitter.rs
+++ b/crates/turborepo-repository/src/package_graph/dep_splitter.rs
@@ -151,7 +151,7 @@ mod test {
     #[test_case("2.3.4", None, "^1.0.0", None ; "handles semver range not satisfied")]
     #[test_case("1.2.3", None, "workspace:1.2.3", Some("@scope/foo") ; "handles workspace protocol with version")]
     #[test_case("1.2.3", None, "workspace:*", Some("@scope/foo") ; "handles workspace protocol with no version")]
-    #[test_case("1.2.3", None, "workspace:../other-packages/", Some("@scope/foo") ; "handles workspace protocol with relative path")]
+    #[test_case("1.2.3", Some("bar"), "workspace:../bar/", Some("bar") ; "handles workspace protocol with relative path")]
     #[test_case("1.2.3", None, "workspace:../@scope/foo", Some("@scope/foo") ; "handles workspace protocol with scoped relative path")]
     #[test_case("1.2.3", None, "npm:^1.2.3", Some("@scope/foo") ; "handles npm protocol with satisfied semver range")]
     #[test_case("2.3.4", None, "npm:^1.2.3", None ; "handles npm protocol with not satisfied semver range")]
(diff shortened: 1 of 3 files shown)


crates/turborepo-repository/src/package_graph/dep_splitter.rs
```diff
@@ -62,17 +78,21 @@ impl<'a> DependencySplitter<'a> {
...
-        if version == "*" || version == "^" || version == "~" {
-            Some(Self::Alias(name))
-        } else {
-            None
+        match version.rsplit_once('@') {
+            Some((name, "*")) | Some((name, "^")) | Some((name, "~")) => Some(Self::Alias(name)),
+            // No indication of different name for the package
+            // We want to capture specifiers that have type "directory" by npa which boils down to
+            // checking for slashes: https://github.com/pnpm/npm-package-arg/blob/main/npa.js#L79
+            Some(_) | None if version.contains('/') => {
```
NicholasLYang, line 92: The code linked above appears to check for backslashes on windows. Should we do that?

NicholasLYang (approved):
Just one question about windows, otherwise LGTM

crates/turborepo-repository/src/package_graph/dep_splitter.rs
chris-olszewski: Luckily in this case pnpm does not generate `workspace:..\windows-path` and expects these to be unix style.

The npa code is used a lot of places, but luckily here we don't have to worry about the "is this path a system or unix style" question.

PR #7520, opened 2024-02-26
fix(turbopack): restore `require.resolve` functionality

ForsakenHarmony:
### Description

Was broken when dynamic expression in require / import were added.

This also combines the runtime of require expressions and `require.context`


sokra (approved):

crates/turbopack-ecmascript-runtime/js/src/build/runtime.ts
```diff
@@ -95,10 +95,7 @@ function loadChunkPath(chunkPath: ChunkPath): void {
   }
 }
 
-async function loadChunkAsync(
-  source: SourceInfo,
-  chunkData: ChunkData
-): Promise<any> {
+async function loadChunkAsync(chunkData: ChunkData): Promise<any> {
```
sokra, line 98: Instead of removing this argument, I think it might be better to actually use it and add this information to the error.

ForsakenHarmony: done

crates/turbopack-ecmascript-runtime/js/src/dev/runtime/none/runtime-backend-none.ts
```diff
@@ -22,16 +22,8 @@ type ExternalRequire = (
...
-interface TurbopackDevContext extends TurbopackDevBaseContext {
-  x: ExternalRequire;
-  y: ExternalImport;
-}
-
 function augmentContext(context: TurbopackDevBaseContext): TurbopackDevContext {
-  const nodejsContext = context as TurbopackDevContext;
-  nodejsContext.x = externalRequire;
-  nodejsContext.y = externalImport;
-  return nodejsContext;
```
sokra, line 34: This change breaks the edge runtime, which actually can have externals, e. g. `require("node:buffer")`

sokra: We should probably rename runtime-none to runtime-edge...

PR #7524, opened 2024-02-27
feat(turbopack): Enable lightningcss for turbopack by default

kdy1:
### Description

Enable lightningcss by default, for --turbo mode.

### Testing Instructions

next.js counterpart: https://github.com/vercel/next.js/pull/62565

Closes PACK-2601


crates/turbopack-css/src/asset.rs
```diff
@@ -36,7 +36,7 @@ pub struct CssModuleAsset {
     asset_context: Vc<Box<dyn AssetContext>>,
     import_context: Option<Vc<ImportContext>>,
     ty: CssModuleAssetType,
-    use_lightningcss: bool,
+    use_swc_css_for_turbopack: bool,
```
sokra, line 39:
```suggestion
    use_swc_css: bool,
```

crates/turbopack-tests/tests/snapshot/css/chained-attributes/output/crates_turbopack-tests_tests_snapshot_css_chained-attributes_input_7c9d25._.css.map
```diff
@@ -1,12 +1,12 @@
...
-    {"offset": {"line": 26, "column": 0}, "map": {"version":3,"sources":["/turbopack/[project]/crates/turbopack-tests/tests/snapshot/css/chained-attributes/input/a.css"],"sourcesContent":["@import url(\"./b.css\") supports(font-format(woff2));\n\n.imported {\n  color: cyan;\n}\n"],"names":[],"mappings":"AAEA,CAAC,QAAQ,CAAC,CAAC;EACT,KAAK,EAAE,IAAI;AACb,CAAC"}},
-    {"offset": {"line": 28, "column": 1}, "map": {"version":3,"sources":[],"names":[],"mappings":"A"}},
-    {"offset": {"line": 34, "column": 0}, "map": {"version":3,"sources":["/turbopack/[project]/crates/turbopack-tests/tests/snapshot/css/chained-attributes/input/style.css"],"sourcesContent":["@import url(\"./a.css\") layer(layer) supports(not(display: inline-grid)) print;\n\n.style {\n  color: yellow;\n}\n"],"names":[],"mappings":"AAEA,CAAC,KAAK,CAAC,CAAC;EACN,KAAK,EAAE,MAAM;AACf,CAAC"}},
-    {"offset": {"line": 36, "column": 1}, "map": {"version":3,"sources":[],"names":[],"mappings":"A"}}]
+    {"offset": {"line": 4, "column": 0}, "map": {"version":3,"sources":["[project]/crates/turbopack-tests/tests/snapshot/css/chained-attributes/input/c.css"],"names":[],"mappings":"A"}},
```
sokra, line 4: `sourcesContent` is missing in SourceMaps

crates/turbopack-tests/tests/snapshot/css/css-legacy-nesting/output/crates_turbopack-tests_tests_snapshot_css_css-legacy-nesting_input_style_7d7e1c.css
```diff
@@ -1,12 +1,3 @@
...
-  text-wrap: initial;
-}
-.prose :where(table):not(:where([class~="not-prose"], [class~="not-prose"] *)) tr {}
-.prose :where(table):not(:where([class~="not-prose"], [class~="not-prose"] *)) tr td:first-child {
-  text-wrap: nowrap;
-}
-.prose :where(code):not(:where([class~="not-prose"], [class~="not-prose"] *)) {
-  text-wrap: nowrap;
-}
+.prose :where(table):not(:where([class~=not-prose],[class~=not-prose] *)){text-wrap:initial}.prose :where(table):not(:where([class~=not-prose],[class~=not-prose] *)) tr td:first-child{text-wrap:nowrap}.prose :where(code):not(:where([class~=not-prose],[class~=not-prose] *)){text-wrap:nowrap}
```
sokra, line 2: Looks like the CSS is minimized. We only want this in production.

crates/turbopack-tests/tests/snapshot/cssmodules/composes/output/79fb1_turbopack-tests_tests_snapshot_cssmodules_composes_input_index_module_7d7e1c.css
```diff
@@ -1,9 +1,3 @@
 /* [project]/crates/turbopack-tests/tests/snapshot/cssmodules/composes/input/index.module.css [test] (css) */
-.className__index__286d8691 {
-  background: red;
-  color: yellow;
-}
-.subClass__index__286d8691 {
-  background: blue;
-}
+.index-module__Tw9iRq__className{background:red;color:#ff0}.index-module__Tw9iRq__subClass{background:#00f;}
```
sokra, line 2: I would prefer the formatting where `className` is in the front. That's much cleaner. Could we configure that for lightningcss?

sokra:
Looks good except for the source maps

crates/turbopack-tests/tests/snapshot/css/chained-attributes/output/crates_turbopack-tests_tests_snapshot_css_chained-attributes_input_7c9d25._.css.map
kdy1: I tried debugging with `dbg`, but it seems like the source map produced by `lightningcss` is relatively minimal at the first place. 

cc @devongovett Can you take a look?

devongovett: For sourcesContent, you need to call `source_map.set_source_content` for each source index to add it. Lightning CSS doesn't store a reference or copy to the original code so that's up to you. I see [here](https://github.com/vercel/turbo/blob/b61c98e970cc7a39fbaf439997cc156afad17b08/crates/turbopack-css/src/process.rs#L134) where you add the source files but I couldn't find where the source content gets set so maybe that's missing? Or you could get the content some other way [here](https://github.com/vercel/turbo/blob/b61c98e970cc7a39fbaf439997cc156afad17b08/crates/turbopack-css/src/process.rs#L833).

As for the mappings themselves, I'm not sure without debugging it myself, but I see a couple possibilities:

1. Maybe try adding the sources before calling `to_css` instead of after?
2. When converting the source map to your internal format, maybe check that the source index is passed through (right now it is always `None`). Looks like maybe you can use `add_raw` instead. Not sure if that is related. At least verify that there are mappings there. https://github.com/vercel/turbo/blob/b61c98e970cc7a39fbaf439997cc156afad17b08/crates/turbopack-css/src/process.rs#L837-L844

The mappings are probably more sparse than you might expect though. Basically there's one per rule since that's the level of granularity that browser dev tools operate at, so more is unnecessary. But it shouldn't be empty at least.

kdy1:
Re-requesting review as there are minor code changes

crates/turbopack-css/src/process.rs
```diff
@@ -131,7 +134,10 @@ impl<'i, 'o> StyleSheetLike<'i, 'o> {
                 })?;
 
                 if let Some(srcmap) = &mut srcmap {
+                    debug_assert_eq!(ss.sources.len(), 1);
```
sokra, line 137: Would this break for an empty css file?

crates/turbopack-css/src/process.rs
```diff
@@ -269,6 +275,8 @@ pub enum ParseCssResult {
         #[turbo_tasks(debug_ignore, trace_ignore)]
         cm: Arc<swc_core::common::SourceMap>,
 
+        code: String,
```
sokra, line 278: We already have the source code lying around somewhere in a Vc. It seems unnecessary to clone that into this and store it again

crates/turbopack-css/src/process.rs
kdy1: No, as `sources` is the list of file names.

sokra (approved):

JesseKoldewijn:
❤️

PR #7526, opened 2024-02-27
Handle logging story better on daemon

arlyon:
### Description

This changes the config from hard-coded to take the setup from your env vars, defaulting to info if unset. It also prevents the daemon from logging span lines. It also adds the ability to clean logs when running `daemon clean` so that users can more easily manage their logs.

### Testing Instructions

Run `turbo daemon` with `TURBO_LOG_VERBOSITY` verbosity unset or set to trace to observe the difference. Run `turbo clean` and see the logs have been removed.

Closes TURBO-2450

### DIFF
--- a/crates/turborepo-lib/src/commands/daemon.rs
+++ b/crates/turborepo-lib/src/commands/daemon.rs
@@ -1,4 +1,4 @@
-use std::{fs, io, path::Path, time::Duration};
+use std::time::Duration;
 
 use camino::Utf8PathBuf;
 use futures::FutureExt;
@@ -45,7 +45,7 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
             if let Err(e) = result {
                 tracing::debug!("failed to restart the daemon: {:?}", e);
                 tracing::debug!("falling back to clean");
-                clean(&connector.paths.pid_file, &connector.paths.sock_file).await?;
+                clean(&connector.paths.pid_file, &connector.paths.sock_file)?;
                 tracing::debug!("connecting for second time");
                 let _ = connector.connect().await?;
             }
@@ -137,7 +137,9 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
                 .status()
                 .expect("failed to execute tail");
         }
-        DaemonCommand::Clean { clean_logs: logs } => {
+        DaemonCommand::Clean {
+            clean_logs: should_clean_logs,
+        } => {
             // try to connect and shutdown the daemon
             let paths = connector.paths.clone();
             let client = connector.connect().await;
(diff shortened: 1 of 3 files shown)


crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -133,7 +133,7 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
                 .status()
                 .expect("failed to execute tail");
         }
-        DaemonCommand::Clean => {
+        DaemonCommand::Clean { clean_logs: logs } => {
```
chris-olszewski, line 136:
```suggestion
        DaemonCommand::Clean { clean_logs: should_clean_logs } => {
```

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -193,6 +196,26 @@ async fn clean(
...
+        println!("Please remove manually: {}", log_folder);
+        success = false;
+    };
+
+    if success {
+        Ok(())
+    } else {
+        // return error
+        Err(DaemonError::CleanFailed)
+    }
```
chris-olszewski, line 216:
```suggestion
    trace!("cleaning up log files");
    // clear all files in the log folder. we want to keep the
    // folder just remove the contents
    // `remove_dir_all_recursive` is lifted from `std`
    fs::remove_dir_all(log_folder.as_std_path()).map_err(|e|  {
        println!("Failed to remove log files: {}", e);
        println!("Please remove manually: {}", log_folder);
        DaemonError::CleanFailed
    })
```

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -193,6 +196,26 @@ async fn clean(
     }
 }
 
+async fn clean_logs(log_folder: &AbsoluteSystemPath) -> Result<(), DaemonError> {
+    let mut success = true;
+    trace!("cleaning up log files");
+    // clear all files in the log folder. we want to keep the
+    // folder just remove the contents
+    // `remove_dir_all_recursive` is lifted from `std`
+    if let Err(e) = fs::remove_dir_all(log_folder.as_std_path()) {
```
chris-olszewski, line 205: I think we should generally prefer using the FS ops through our path lib as we use `fs_err` which includes the path name in the error.
```suggestion
    if let Err(e) = log_folder.remove_dir_all() {
```

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -193,6 +196,26 @@ async fn clean(
     }
 }
 
+async fn clean_logs(log_folder: &AbsoluteSystemPath) -> Result<(), DaemonError> {
```
chris-olszewski, line 199: I don't think this needs to be async
```suggestion
fn clean_logs(log_folder: &AbsoluteSystemPath) -> Result<(), DaemonError> {
```

chris-olszewski (approved):
Cleaning the daemon logs seems like a good change, but not what the PR title & body suggest. Totally think this should get merged in, but we should probably update the title so our changelog is more accurate.

Also some unused imports that should get removed before merge.

mehulkar (approved):

PR #7527, opened 2024-02-27
implement handling for alias field resolving in more cases

sokra:
### Description

Implement a few more edge cases for resolving the alias field

### Testing Instructions

Closes PACK-2605


crates/turbopack-tests/tests/execution/turbopack/resolving/alias-field/input/node_modules/package/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "browser": {
+    "./dir/file1.js": "./replaced/file1.js",
+    "./dir/file2": "./replaced/file2",
+    "dir/file3.js": "./replaced/file3.js",
```
kdy1, line 5: I'm not sure if this is correct. I'll take a look at the node.js doc

sokra: It's not the the spec, and I don't think it's "correct", but webpack supported that and that helps with compatibility

kdy1: I see. Thank you for the explanation!

kdy1 (approved):

PR #7536, opened 2024-02-28
feat(Turborepo): Pull a graph of 3 months worth of perf data, save to blob store

gsoltis:
### Description

 - Uses tinybird to grab a slice of recent performance data
 - uses vega + sharp to render a graph to png
 - saves graph to vercel blob storage
 - plumbs the url of the uploaded blob through to the slack payload

TODO:
- [ ] Modify slack workflow to accept the URL (edit: this needs to be done post-merge, since we won't have the right payload yet)
- [x] Add needed env vars to github actions workflow

### Testing Instructions

Manual testing of URL generation. Latest is [here](https://4fi2mhasf23fvrzk.public.blob.vercel-storage.com/aggregate-owqEOruxqic7gO448J8YO5g5xdkyQ8.png)
Manual post to a test webhook. 
I also ran end-to-end via GH Actions with the old slack payload.

Closes TURBO-2456


.github/workflows/bench-turborepo.yml
```diff
@@ -140,7 +140,10 @@ jobs:
 
   send-to-slack:
     name: Send to Slack
-    needs: [time-to-first-task]
+    # this is no longer concurrent with send-to-tinybird.
+    # We want to account for the latest datapoint in the 3 month
+    # aggregate graph.
```
mehulkar, line 145:
```suggestion
    # Wait for send-to-tinybird so we can get aggregate data points
    # before sending to slack.
```

mehulkar (approved):
One suggestion for the code comment but otherwise looks good!

packages/turbo-benchmark/src/ttft/build-vega.ts
```diff
@@ -0,0 +1,234 @@
...
+    values: aggregateData.data,
+    transform: [
+      {
+        type: "formula",
+        expr: "datum.durationMicroseconds/1000",
+        as: "durationMs",
+      },
+    ],
+  };
+  const vegaConfig: vega.Spec = {
```
tknickman, line 32: Would love a comment here about where you got this / how to generate it again if needed.

tknickman (approved):

PR #7544, opened 2024-02-28
extract a turbopack-resolve crate

sokra:
### Description

extract a turbopack-resolve crate from resolving stuff in "turbopack" and "turbopack-ecmascript"

This is needed to extract a turbopack-webpack crate which contains the webpack loader implementation and allow it to access ResolvingOptionContext and call `resolve_options`, while it being referenced from the `turbopack` crate

### Testing Instructions

Closes PACK-2613


crates/turbopack-resolve/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "turbopack-resolve"
+version = "0.1.0"
+description = "TBD"
```
kwonoj, line 4: coincidentally we just discussed this (https://vercel.slack.com/archives/C03EWR7LGEN/p1709141958393789?thread_ts=1709077393.256749&cid=C03EWR7LGEN), mind leave a one liner for the descrription?

kwonoj (approved):

PR #7545, opened 2024-02-28
fix: Load vercel token into command base

Zertsov:
### Description
There's an issue where if you're logged into Vercel, and don't have a Turbo token, then multiple commands will fail (unless you run `turbo login -f`).

This happened because the command base wasn't looking for the Vercel token. This addresses that by loading in the Vercel token for use if it's present.

### Testing Instructions
1. Ensure no tokens exist for Turbo or Vercel (`vc logout && turbo logout`)
2. `vc login`
3. `turbo login` -> Existing token found!
4. `turbo link` in a repo the token has access to -> Normal link process happens


Zertsov:
* **#7545** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7545?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/turborepo-auth/src/lib.rs
```diff
@@ -38,6 +39,10 @@ impl Token {
     pub fn existing(token: String) -> Self {
         Self::Existing(token)
     }
+    pub fn from_file(path: AbsoluteSystemPathBuf) -> Result<Self, Error> {
+        let token = path.read_to_string()?;
+        Ok(Self::Existing(token))
+    }
```
Zertsov, line 45: This isn't needed - was trying something different at one point

crates/turborepo-lib/src/config.rs
```diff
@@ -508,6 +534,21 @@ impl TurborepoConfigBuilder {
...
+        let mut contents = global_auth_path
+            .read_existing_to_string_or(Ok("{}"))
+            .map_err(|error| Error::FailedToReadConfig {
+                config_path: global_auth_path.clone(),
+                error,
+            })?;
+        if contents.is_empty() {
+            contents = String::from("{}");
+        }
+        let global_auth: ConfigurationOptions = serde_json::from_str(&contents)?;
```
chris-olszewski, line 548: I think we want to only read `token` from these sources. Small chances, but if the Vercel auth file ever adds additional fields that have shared names we could be in a bad situation.

chris-olszewski (approved):
Stamping as this gets us to a better state and ITG, but we should fast follow cleaning up this logic. This currently works because we're doing things twice and I could easily see ourselves accidentally breaking this in the future.

This [code path](https://github.com/vercel/turbo/blob/main/crates/turborepo-auth/src/auth/login.rs#L62) will never be taken after this PR.
Same with this [code path](https://github.com/vercel/turbo/blob/main/crates/turborepo-auth/src/auth/sso.rs#L75)

crates/turborepo-lib/src/config.rs
```diff
@@ -508,6 +534,34 @@ impl TurborepoConfigBuilder {
         Ok(local_config)
     }
 
+    fn get_global_auth(&self) -> Result<ConfigurationOptions, Error> {
```
mehulkar, line 537: Don't we want this logic in the auth crate?

mehulkar (approved):

Zertsov:
We do, but this is one of those stopgap fixes. I plan on moving most, if not all, of this kinda logic into the auth crate in a subsequent PR

crates/turborepo-lib/src/config.rs
Zertsov: We do, but this is one of those stopgap fixes. I plan on moving most, if not all, of this kinda logic into the auth crate in a subsequent PR

PR #7549, opened 2024-02-28
refactor(turborepo): Package Detection

NicholasLYang:
### Description
Factored out package detection, i.e. the part where we take a file and map it to the package into a trait. This allows us to have multiple implementations, such as a default one that only uses the package graph, and a more advanced implementation that uses the package graph and turbo.json.

### Testing Instructions

Closes TURBO-2465


mehulkar (commented):
- Am I correct to understand that the new `Turbo` implementation of PackageDetector isn't being used in this PR? I would prefer to see that implementation when it's used (by watch mode?)
- Can you also update the usage in the JS package? https://github.com/vercel/turbo/blob/306f130150ce4cf22f97ddfa6d25533dffaca3f2/packages/turbo-repository/rust/src/lib.rs#L160
  - CI _should_ be failing on this change for that package, but I'm not sure if it's setup right

crates/turborepo-lib/src/package_detector.rs
```diff
@@ -0,0 +1,120 @@
...
+            global_deps_matcher: matcher,
+            turbo_json,
+        })
+    }
+}
+
+impl<'a> PackageDetector for TurboJsonPackageDetector<'a> {
+    fn detect_package(&self, path: &AnchoredSystemPath) -> Option<WorkspacePackage> {
+        match DefaultPackageDetector::new(self.pkg_dep_graph).detect_package(path) {
+            root @ Some(WorkspacePackage {
```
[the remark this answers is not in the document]
NicholasLYang: It creates a variable that corresponds to the entire pattern, so you can match but use the entire value. In other words:

```
let foo = Some(10);
match foo {
  // bar matches 10
  Some(bar) => ...
  // bar matches Some(10)
  bar @ Some(_) => ..
}
```

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -28,10 +30,10 @@ impl<'a> ScopeChangeDetector<'a> {
         turbo_root: &'a AbsoluteSystemPath,
         scm: &'a SCM,
         pkg_graph: &'a PackageGraph,
-        global_deps: Vec<String>,
         ignore_patterns: Vec<String>,
     ) -> Self {
-        let change_mapper = ChangeMapper::new(pkg_graph, global_deps, ignore_patterns);
+        let pkg_detector = DefaultPackageDetector::new(pkg_graph);
+        let change_mapper = ChangeMapper::new(pkg_graph, ignore_patterns, pkg_detector);
```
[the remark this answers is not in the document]
NicholasLYang: In theory it's ok since if you're changing a root file that's not a globalDependency, that should result in a cache hit, but I think we should discuss this before doing anything

crates/turborepo-lib/src/package_detector.rs
```diff
@@ -0,0 +1,120 @@
...
+    InvalidFilter(#[from] BuildError),
+}
+
+pub struct TurboJsonPackageDetector<'a> {
+    pkg_dep_graph: &'a PackageGraph,
+    global_deps_matcher: Any<'a>,
+    turbo_json: &'a TurboJson,
+}
+
+impl<'a> TurboJsonPackageDetector<'a> {
```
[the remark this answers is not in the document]
NicholasLYang: I'm confused, this is the only content in the file? Perhaps I should just rename it to `turbo_json_package_detector`?

NicholasLYang:
Adding the usage for `TurboJsonPackageDetector` is a little out of scope cause that would involve pulling in the watch mode code, which would make this a pretty big PR

crates/turborepo-repository/src/package_graph/builder.rs
```diff
@@ -690,6 +690,23 @@ impl PackageInfo {
     }
 }
 
+#[allow(dead_code)]
+pub struct MockDiscovery;
```
chris-olszewski, line 694: We should definitely make sure this is only usable in tests. If we expect this to be shared more widely, it should move out of the package graph builder tests.
```suggestion
#[allow(dead_code)]
#[cfg(test)]
pub struct MockDiscovery;
```

crates/turborepo-repository/src/change_mapper.rs
```diff
@@ -8,37 +8,85 @@ use wax::Program;
...
+}
+
 // We may not be able to load the lockfile contents, but we
 // still want to be able to express a generic change.
 pub enum LockfileChange {
     Empty,
     WithContent(Vec<u8>),
 }
 
+#[derive(Debug, PartialEq)]
```
chris-olszewski, line 67: Is there a reason for not also implementing `Eq`?

crates/turborepo-repository/src/change_mapper.rs
```diff
@@ -47,17 +95,9 @@ impl<'a> ChangeMapper<'a> {
         changed_files: HashSet<AnchoredSystemPathBuf>,
         lockfile_change: Option<LockfileChange>,
     ) -> Result<PackageChanges, ChangeMapError> {
-        let global_change =
-            self.repo_global_file_has_changed(&Self::DEFAULT_GLOBAL_DEPS, &changed_files)?;
-
-        if global_change {
-            return Ok(PackageChanges::All);
```
chris-olszewski, line 54: No longer returning that all packages have changed on a root `package.json` alteration feels like something that needs more testing/validation.

chris-olszewski (changes requested):
This PR is labeled as a refactor, but has some pretty significant behavior changes for both `turbo` and `@turbo/repository`.

I'd feel far more comfortable with the refactors remaining, but actual behavior changes coming in a different PR with more testing to get a better picture of how this could affect our behavior in various scenarios.

crates/turborepo-repository/src/change_mapper.rs
NicholasLYang: Not really, I just added what I needed

crates/turborepo-repository/src/change_mapper.rs
NicholasLYang: I removed this because it's redundant. If we return the root workspace as changed if *any* file not in a package is changed, then there's no reason to also have this code.

NicholasLYang:
I don't believe there's a behavior change here. With the current code, we return the root changed if there exists any changed file that is not in a package. All I'm doing is removing the redundant code that also checks if `package.json` or `turbo.json` are changed when they're by definition not in any package and therefore fall under the first check.

packages/turbo-repository/__tests__/find.test.ts
```diff
@@ -58,11 +58,6 @@ describe("Workspace", () => {
           { name: "ui", relativePath: "packages/ui" },
         ],
       },
-      {
```
NicholasLYang, line 61: I removed this test because without access to `globalDependencies`, you can't really know if a global change can be ignored. For all we know, a user could have `README.md` in their `globalDependencies` array.

mehulkar: `globalDependencies` is not a known concept here. `@turbo/repository` does not assume the presence of a `turbo.json` file.

Can you put the test back and change the `expected` value instead? description can be something like `global change that probably should be irrelevant`. I want to be clear that this is documented/intentional behavior in the tests

mehulkar (changes requested):

packages/turbo-repository/__tests__/find.test.ts
NicholasLYang: Done!

crates/turborepo-lib/src/global_deps_package_detector.rs
```diff
@@ -0,0 +1,146 @@
...
+        )?;
+
+        // We should return All because we don't have global deps and
+        // therefore must be conservative about changes
+        assert_eq!(package_changes, PackageChanges::All);
+
+        let turbo_json = TurboJson::default();
+        let turbo_package_detector = GlobalDepsPackageDetector::new(
+            &pkg_graph,
+            turbo_json.global_deps.iter().map(|s| s.as_str()),
```
chris-olszewski, line 126: Nit: can we just the value here instead of depending on the default value of `TurboJson`?

crates/turborepo-lib/src/global_deps_package_detector.rs
```diff
@@ -0,0 +1,146 @@
...
+
+impl<'a> PackageDetector for GlobalDepsPackageDetector<'a> {
+    fn detect_package(&self, path: &AnchoredSystemPath) -> PackageDetection {
+        match DefaultPackageDetector::new(self.pkg_dep_graph).detect_package(path) {
+            // Since `DefaultPackageDetector` is overly conservative, we can check here if
+            // the path is actually in globalDeps and if not, return it as
+            // PackageDetection::Package(WorkspacePackage::root()).
+            PackageDetection::All => {
+                let cleaned_path = path.clean();
+                let in_global_deps = self.global_deps_matcher.is_match(cleaned_path.as_str());
```
chris-olszewski, line 48: I'm having trouble tracing this code. Does the `global_deps_matcher` always get the default global deps of `package.json`/`turbo.json` added to it? Otherwise we'll incorrectly map a `package.json` change just be a root package change instead of a all change.

Or I'm misunderstanding these changes completely and this codepath isn't hit in that scenario.

Edit: I was misunderstanding, default global deps are checked at `ChangeMapper` which uses a package detector if a default global change isn't detected

crates/turborepo-repository/src/change_mapper.rs
```diff
@@ -8,56 +8,124 @@ use wax::Program;
...
             pkg_graph,
-            global_deps,
             ignore_patterns,
+            package_detector,
         }
     }
 
+    fn default_global_file_changed(
+        changed_files: &HashSet<AnchoredSystemPathBuf>,
+    ) -> Result<bool, ChangeMapError> {
```
chris-olszewski, line 105:
```suggestion
    ) -> bool {
```

crates/turborepo-lib/src/global_deps_package_detector.rs
```diff
@@ -0,0 +1,146 @@
...
+pub enum Error {
+    #[error(transparent)]
+    InvalidFilter(#[from] BuildError),
+}
+
+/// A package detector that uses a global deps list to determine
+/// if a file should cause all packages to be marked as changed.
+/// This is less conservative than the `DefaultPackageDetector`
+/// which assumes that any changed file that is not in a package
+/// is a root dependency. Since we have a list of global deps,
```
chris-olszewski, line 19: I'm having trouble parsing this comment is "which" being used to explain `DefaultPackageDetector` or `GlobalDepsPackageDetector`?
```suggestion
/// This is less conservative than the `DefaultPackageDetector`
/// which assumes that any changed file that is not in a package
/// changes all packages. Since we have a list of global deps,
```

chris-olszewski (approved):
I do feel that this makes it harder to reasons about this code, but it seems necessary for what we want to do in the future. Relying on your manual testing to verify there aren't accidental behavior changes anymore.

crates/turborepo-repository/src/change_mapper.rs
```diff
@@ -8,56 +8,120 @@ use wax::Program;
...
+    /// This change has affected one package
+    Package(WorkspacePackage),
+}
+
+/// Detects which packages are affected by a file change. This can be a single
+/// package (`Package`), none of the packages (`None`), or all of the packages
+/// (`All`).
+pub trait PackageDetector {
+    fn detect_package(&self, file: &AnchoredSystemPath) -> PackageDetection;
+}
```
mehulkar, line 25: nit: I don't love the naming here. In english, this trait/method isn't detecting packages (which sounds more like what `turbo info` does), it is inferring how a file change affects a known list of packages. Just a suggestion:

```suggestion
pub trait WorkspaceChangeDetector {
    fn get_changed_package(&self, file: &AnchoredSystemPath) -> WorkspaceChange;
}
```

mehulkar (commented):
A few inline comments. Another question I have is:

I want to be able to return a list of the full tree of "affected" packages. How/where should this modeled? I see there are two levels of abstraction (ChangeMapper takes a list of files and `PackageDetector` takes one file). 

- Where would you implement the dependency graph walk so that `some_method(files)` will return all affected packages, and not just the direct mapping from file to package
- How would you organize that with the `global` (vs not) feature?

crates/turborepo-lib/src/global_deps_package_detector.rs
```diff
@@ -0,0 +1,146 @@
...
+            // the path is actually in globalDeps and if not, return it as
+            // PackageDetection::Package(WorkspacePackage::root()).
+            PackageDetection::All => {
+                let cleaned_path = path.clean();
+                let in_global_deps = self.global_deps_matcher.is_match(cleaned_path.as_str());
+
+                if in_global_deps {
+                    PackageDetection::All
+                } else {
+                    PackageDetection::Package(WorkspacePackage::root())
```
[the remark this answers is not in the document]
NicholasLYang: So I think there's a few misunderstandings (not helped by the multiple iterations of this PR):

- It is helpful to know that the root package changed, because you can have root tasks that can get run, say if you define a `//#build` entry in pipeline. We can't remove this root package return because that would be a breaking change for people with root package tasks.
- Root package changed does *not* imply all packages changed. I thought it did but @chris-olszewski explained to me that it does not. I think it should in an ideal world (and I think a setting to enable that behavior is worth discussing but out of scope).
- Therefore separating between the two concepts is fine because that is the existing behavior.
- As for how our model relates to a package manager, I agree that we should stay closer to the package manager's semantics, but unfortunately we can't due to existing behavior. Like we should treat a root package change as a global change because that's what package managers effectively do, but we have to wait for a breaking change to make that possible.

mehulkar (approved):

crates/turborepo-repository/src/change_mapper.rs
NicholasLYang: Hmm yeah agreed. I do feel like WorkspaceChangeDetector is a little confusing too. Maybe PackageChangeMapper? It's mapping a single change to a package

mehulkar: yeah I didn't want to use "change mapper" since we use that upstream already 😓 . In lieu of an obvious answer, no need to make changes :)

PR #7556, opened 2024-02-29
fix: vercel auth file borks login

Zertsov:
### Description
Because `vc logout` doesn't remove it's pseudo-comments, deserialize breaks when trying to extract the token.
This PR fixes that by setting the deserialized struct to have an optional token field and then we make sure the extraction didn't encounter an error _and_ that the token is `Some`.

### Testing Instructions
First, make sure you build the branch.
1. `vc login && vc logout` -> make the Vercel auth file and then remove the `token` field. You could do this manually as well.
2. `turbo login` -> Shouldn't error and ask for a new login.

Closes TURBO-2485

### DIFF
--- a/crates/turborepo-auth/src/auth/login.rs
+++ b/crates/turborepo-auth/src/auth/login.rs
@@ -62,7 +62,7 @@ pub async fn login<T: Client + TokenClient + CacheClient>(
         } else if login_url_configuration.contains("vercel.com") {
             // The extraction can return an error, but we don't want to fail the login if
             // the token is not found.
-            if let Ok(token) = extract_vercel_token() {
+            if let Ok(Some(token)) = extract_vercel_token() {
                 let token = Token::existing(token);
                 if token
                     .is_valid(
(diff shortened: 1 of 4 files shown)


Zertsov:
* **#7556** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7556?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

chris-olszewski (changes requested):
This doesn't fix the issue since this codepath isn't hit when loading the token via the command base.
```
[0 olszewski@chriss-mbp] /Users/olszewski/code/vercel/turborepo $ turbo_dev info
  × Authentication error: invalid token file format: missing field `token` at line 4 column 1
  ├─▶ invalid token file format: missing field `token` at line 4 column 1
  ╰─▶ missing field `token` at line 4 column 1
```

Please test that other commands work in addition to `login`.

chris-olszewski (approved):

PR #7557, opened 2024-02-29
fix: Respect globalDependencies when determining changed packages

NicholasLYang:
### Description

We didn't respect the `globalDependencies` field in `turbo.json` when determining changed packages. This meant that you could have a global dependency changed, but turbo wouldn't run.

### Testing Instructions

Closes TURBO-2486

### DIFF
--- a/turborepo-tests/integration/tests/filter-run.t
+++ b/turborepo-tests/integration/tests/filter-run.t
@@ -27,3 +27,17 @@ Setup
   Cached:    0 cached, 0 total
     Time:\s*[\.0-9]+m?s  (re)
   
+
+  $ rm bar.txt
+  $ echo "global dependency" >> foo.txt
+  $ git commit -am "global dependency change" --quiet
+  $ ${TURBO} run build --filter="[HEAD^]" --output-logs none
+  \xe2\x80\xa2 Packages in scope: //, another, my-app, util (esc)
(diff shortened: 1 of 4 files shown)


chris-olszewski (changes requested):
Needs tests

turborepo-tests/integration/tests/filter-run.t
```diff
@@ -26,4 +26,18 @@ Setup
    Tasks:    0 successful, 0 total
   Cached:    0 cached, 0 total
     Time:\s*[\.0-9]+m?s  (re)
-  
\ No newline at end of file
+  
+
```
chris-olszewski, line 30: This works as an integration test, but we have pretty good fixtures in the `filter.rs` unit tests and feels like we could get away with just those.

chris-olszewski (approved):
I would prefer unit tests, but if this passes all tests, then I'm fine with it as is.

turborepo-tests/integration/tests/filter-run.t
NicholasLYang: From a quick scan, I'm not sure we have unit tests that specifically handle git ranges where you can specify the packages changed. TestChangeDetector works by storing the changed packages for a git ref range, not the changed files.

PR #7559, opened 2024-03-01
feat(turbopack): support importmapping extension

kwonoj:
### Description

This PR attempts to expand importmapping to have `Extension`, to create import from the source code string. Importmapping will create virtualasset per each mapping. 

Also it includes a runtime fn to create a proxy to throw certain error at runtime, which will be used edge runtime's alias to create mappings.

Closes PACK-2629


sokra (changes requested):
I have some points, will comment on monday

kwonoj:
Per discussion, we'll going to approach without expanding importmapping enum but try to use existing mechanism.

PR #7562, opened 2024-03-01
add getResolve to webpack loader context

sokra:
### Description

* Increase Node.js execution timeout
* Add ready signal to node.js pool
* add tracing for Node.js bootstrapping
* fixes a tracing bug, where queue time is accounted towards Node.js execution
* refactor evaluation to allow custom message handlers
* add getResolve to webpack loader context

### Testing Instructions

Closes PACK-2633

### DIFF
--- a/crates/turbopack-node/js/src/ipc/evaluate.ts
+++ b/crates/turbopack-node/js/src/ipc/evaluate.ts
@@ -1,64 +1,135 @@
 import { IPC, StructuredError } from "./index";
 import type { Ipc as GenericIpc } from "./index";
 
-type IpcIncomingMessage = {
-  type: "evaluate";
-  args: string[];
-};
+type IpcIncomingMessage =
+  | {
+      type: "evaluate";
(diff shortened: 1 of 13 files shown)


crates/turbopack-node/js/src/ipc/evaluate.ts
```diff
@@ -1,64 +1,135 @@
...
+    const module = await moduleFactory();
+    if (typeof module.init === "function") {
+      await module.init();
+    }
+    getValue = module.default;
+  } catch (err) {
+    await ipc.sendReady();
+    await ipc.sendError(err as Error);
+  }
+  await ipc.sendReady();
```
timneutkens, line 84: Seems this will send ready twice, should it look like this:
```suggestion
   await ipc.sendReady();
  } catch (err) {
    await ipc.sendReady();
    await ipc.sendError(err as Error);
  }

```

timneutkens (approved):

PR #7565, opened 2024-03-01
Ignore circular aliasing with alias field

sokra:
### Description

Handles the case where alias field is circular:

```
  "browser": {
    "@smithy/chunked-blob-reader": "@smithy/chunked-blob-reader"
  },
```

### Testing Instructions

added test case

Closes PACK-2637

### DIFF
--- a/crates/turbopack-core/src/resolve/mod.rs
+++ b/crates/turbopack-core/src/resolve/mod.rs
@@ -10,7 +10,6 @@ use std::{
 use anyhow::{bail, Result};
 use indexmap::{indexmap, IndexMap, IndexSet};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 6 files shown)


crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1944,22 +1943,53 @@ async fn apply_in_package(
...
+                )
+                .cell(),
+            ));
+        }
+
+        if let Some(value) = value.as_str() {
+            if value == request {
+                // This would be a cycle, so we ignore it
+                return Ok(None);
+            }
```
sokra, line 1964: This is new. The other changes just more existing code around...

timneutkens (approved):

PR #7570, opened 2024-03-01
feat(turborepo): Package Change Watching

NicholasLYang:
### Description

Added a watcher in `FileWatching` to watch for file changes and map them to packages. Currently you can run this via `turbo daemon watch` but eventually this will be used for watch mode.

### Testing Instructions
Run the daemon and use `turbo daemon watch` to see the changed package events as you change files.

Closes TURBO-2499


crates/turborepo-filewatch/src/package_changes_watcher.rs
```diff
@@ -0,0 +1,226 @@
...
+                        }
+                    }
+                    Ok(Err(err)) => {
+                        tracing::error!("file event error: {:?}", err);
+                        break;
+                    }
+                    Err(broadcast::error::RecvError::Lagged(_)) => {
+                        tracing::warn!("file event lagged");
+                        break;
+                    }
```
gsoltis, line 207: For these errors, are we done or do we want to loop around and try again?

crates/turborepo-filewatch/src/package_changes_watcher.rs
```diff
@@ -0,0 +1,226 @@
...
+                }
+            }
+        };
+
+        tokio::select! {
+            biased;
+            _ = exit_rx => {
+                tracing::debug!("exiting package changes watcher due to signal");
+            },
+            _ = process => {
```
gsoltis, line 221: Related to above, if we've bailed on `process`, is there a reason we shouldn't loop around and try again?

crates/turborepo-lib/src/daemon/proto/turbod.proto
```diff
@@ -92,6 +94,18 @@ message DiscoverPackagesRequest {
 
 }
 
+message PackageChangesRequest {}
+
+message PackageChangeEvent {
+  PackageChangeType change_type = 1;
```
gsoltis, line 100: Would it be helpful to use [oneof](https://protobuf.dev/programming-guides/proto3/#oneof) here? You can either send a rediscover everything message, or a package change, even with an optional name.

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -566,6 +573,43 @@ impl<PD: PackageDiscovery + Send + Sync + 'static> proto::turbod_server::Turbod
...
             })
     }
+
+    type PackageChangesStream = ReceiverStream<Result<proto::PackageChangeEvent, tonic::Status>>;
+
+    async fn package_changes(
+        &self,
+        _request: tonic::Request<proto::PackageChangesRequest>,
+    ) -> Result<tonic::Response<Self::PackageChangesStream>, tonic::Status> {
+        let mut package_changes_rx = self.file_watching.package_changes_watcher.package_changes();
```
gsoltis, line 583: Can you add a comment about ordering here? How does the client know when changes will started to be accounted for? If the stream is open, and then the client makes a change, are we guaranteed to see a change event? Or is it possible that filewatching is still starting up?

gsoltis (changes requested):
This looks like a pretty good start. I have a question about ordering with respect to stream setup and filewatching startup, but once that's resolved I think we're in good shape.

crates/turborepo-lib/src/daemon/server.rs
NicholasLYang: The first event we send is a rediscover event so in theory it shouldn't matter. Even if the client changes something, the first thing it's going to do after connecting to the daemon is build all of the packages

crates/turborepo-filewatch/src/package_changes_watcher.rs
```diff
@@ -0,0 +1,226 @@
...
+                                    }
+                                }
+                            }
+                            Ok(PackageChanges::Some(changed_pkgs)) => {
+                                tracing::debug!(
+                                    "changed files: {:?} changed packages: {:?}",
+                                    changed_files,
+                                    changed_pkgs
+                                );
+                                for pkg in changed_pkgs {
```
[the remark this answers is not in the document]
NicholasLYang: Yeah I think in the client is maybe better since it keeps separation of concerns. Although perhaps if there's performance issues in the watcher, we could consider adding it there.

crates/turborepo-filewatch/src/package_changes_watcher.rs
NicholasLYang: To be honest I have no idea. I erred (haha) on the side of caution. For lagging, in theory you could send a rediscover event and then continue (since the error is effectively saying we're behind on change events). I'm not sure about file errors.

crates/turborepo-lib/src/daemon/proto/turbod.proto
NicholasLYang: Hmm I guess we could do something like:

```
oneof change_oneof {
  bool rediscover = 1;
  string package_name = 2;
}
```

But that wouldn't be extendable if we wanted to add more message types, and the boolean is a little awkward there. Is there a better way? I'm not super familiar with protobuf

gsoltis: What about something like this:
```
message PackageChanged {
  string package_name = 1;
}

message RediscoverPackages {}

message PackageChangeEvent {
  oneof event {
    PackageChanged package_changed = 1;
    RediscoverPackages rediscover_packages = 2;
  };
}
```

NicholasLYang: Ah yeah that makes a lot more sense. Thanks!

crates/turborepo-lib/src/daemon/server.rs
gsoltis: Maybe I'm misunderstanding, but it doesn't look like that's what it's doing. The watcher task sends a rediscover message first, but there's no guarantee that a client will get that message, right?

Imagine we start the daemon, send the rediscover message, then make some changes, which triggers more package change notifications. _Then_ a client connects and wants to watch package changes. Is there a mechanism for that client to get the rediscover message?

crates/turborepo-filewatch/src/package_changes_watcher.rs
```diff
@@ -0,0 +1,234 @@
...
+        let process = async {
+            let Some(mut repo_state) = self.initialize_repo_state().await else {
+                return;
+            };
+
+            self.package_change_events_tx
+                .send(PackageChangeEvent::Rediscover)
+                .ok();
+            // TODO: Pass in global_deps and ignore_patterns
+            let mut change_mapper = ChangeMapper::new(&repo_state.pkg_dep_graph, vec![], vec![]);
```
gsoltis, line 124: I guess this is where the package discovery piece comes in? We need to invalidate the change mapper on changes to the package graph?

NicholasLYang: Yeah. We're basically building up a [salsa](https://github.com/salsa-rs/salsa) style query system where we have dependent data like the package graph and the change mapper on top of the package graph, that is derived from the files. When the files change, we need to rebuilt the package graph and therefore the change mapper.

crates/turborepo-lib/src/daemon/server.rs
NicholasLYang: Ah yeah. I'll send a rediscover as the first message when we start up the stream.

crates/turborepo-lib/src/daemon/proto/turbod.proto
```diff
@@ -92,6 +94,26 @@ message DiscoverPackagesRequest {
...
+  }
+}
+
+message PackageChanged {
+  string package_name = 1;
+}
+
+message RediscoverPackages {}
+
+enum PackageChangeType {
```
gsoltis, line 112: I think we can drop this enum now

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -548,6 +557,55 @@ impl proto::turbod_server::Turbod for TurboGrpcServiceInner {
...
+                proto::RediscoverPackages {},
+            )),
+        }))
+        .await
+        .map_err(|e| tonic::Status::unavailable(format!("{}", e)))?;
+
+        tokio::spawn(async move {
+            loop {
+                if let Err(err) = package_changes_rx.changed().await {
+                    error!("package changes stream closed: {}", err);
```
gsoltis, line 581: Is there a way to send an error, or close the stream with an error? Or at least close the stream?

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    pub fn package_changes(&self) -> watch::Receiver<PackageChangeEvent> {
+        self.package_change_events_rx.clone()
+    }
+}
+
+struct Subscriber {
+    file_events_lazy: OptionalWatch<broadcast::Receiver<Result<Event, NotifyError>>>,
+    repo_root: AbsoluteSystemPathBuf,
+    package_change_events_tx: watch::Sender<PackageChangeEvent>,
```
gsoltis, line 55: Is `watch` the right kind of channel here? IIUC, that doesn't guarantee you'll see every event, just the latest one. I think we probably want a `broadcast` channel.

gsoltis (changes requested):
This is looking pretty good. I think it's close to MVP.

crates/turborepo-lib/src/daemon/server.rs
NicholasLYang: Since this breaks the loop, it should drop the `tx` which in turn closes the channel. But we can send errors instead I suppose.

NicholasLYang:
* **#7613** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7613?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7570** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7570?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

docs/public/feed.xml
```diff
@@ -0,0 +1,151 @@
+<?xml version="1.0" encoding="UTF-8"?><rss xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:content="http://purl.org/rss/1.0/modules/content/" xmlns:atom="http://www.w3.org/2005/Atom" version="2.0">
```
gsoltis, line 1: Looks like this is accidentally added, along with the other docs/public change

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -440,6 +440,7 @@
         if let Some(git_range) = selector.git_range.as_ref() {
             selector_valid = true;
             let changed_packages = self.packages_changed_in_range(git_range)?;
+
```
gsoltis, line 443: Errant whitespace

gsoltis (approved):
A couple of extra files snuck in, but otherwise I think we're good to merge this

PR #7580, opened 2024-03-03
Remove barrel file from design-system-example

anthonyshew:


lorcan-codes:
Great work 🙌 @anthonyshew. I have one question. Do we need to defined every component in exports config similar to what you did in your PR?

anthonyshew:
We recommend this because of a TypeScript limitation. If you do:
```
"exports: {
  "./*": "./src/*"
```

the exports will still resolve but you lose auto-importing suggestions. Up to you which side of the tradeoff you prefer!

tknickman (approved):

cgatian:
@anthonyshew I think the README still mentions including each package in the `index`

> When adding a new file, ensure the component is also exported from the entry index.tsx file:

https://github.com/vercel/turbo/blob/main/examples/design-system/README.md

PR #7581, opened 2024-03-03
fix typo in tsconfig

zsh77:
There is no package named 'tsconfig'. Replaced it with the correct name which is '@repo/typescript-config'

### Description

Replaced `tsconfig/base.json` with `@repo/typescript-config/base.json`

### Testing Instructions

running `pnpm lint` should work now and also ctrl + click on the extended typescript file in the root's tsconfig.json should work too.

### DIFF
--- a/examples/with-tailwind/tsconfig.json
+++ b/examples/with-tailwind/tsconfig.json
@@ -1,3 +1,0 @@
-{
-  "extends": "@repo/typescript-config/base.json"
-}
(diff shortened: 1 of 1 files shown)


zsh77:
I ended up deleting the file as it wasn't being used.

zsh77:
I did a complete pull request on this example:
https://github.com/vercel/turbo/pull/7584

PR #7616, opened 2024-03-04
feat(@turbo/repository): return dependencies in graph

mehulkar:
The test changes are just to break up the files so jest can parallelize execution.

Closes TURBO-2524


gaspar09 (approved):

packages/turbo-repository/js/index.d.ts
```diff
@@ -32,7 +36,7 @@ export class Workspace {
    * Finds and returns a map of packages within the workspace and its
    * dependents (i.e. the packages that depend on each of those packages).
    */
-  findPackagesAndDependents(): Promise<Record<string, Array<string>>>;
+  findPackagesWithGraph(): Promise<SerializablePackages>;
```
gaspar09, line 39: Need to update the comment

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -110,26 +144,41 @@ impl Workspace {
...
+                return Err(Error::from_reason(e.to_string()));
+            }
         };
 
-        let map: HashMap<String, Vec<String>> = packages
+        // Create a map of package names to their dependents and dependencies
+        //  {
+        //    "package-path": {
+        //      "dependents": ["dependent1_path", "dependent2_path"],
+        //      "dependencies": ["dependency1_path", "dependency2_path"]
```
gaspar09, line 162: Does this include non-workspace dependencies?

gaspar09 (changes requested):

packages/turbo-repository/rust/src/lib.rs
mehulkar: What do you mean? These will be the list of _packages_ in the workspace, but just their relative paths relative to the root of the workspace

packages/turbo-repository/js/index.d.ts
mehulkar: done

packages/turbo-repository/rust/src/lib.rs
gaspar09: Yea that's what I meant.

gaspar09: The api kinda misleads to make it seem like you return all dependencies not just the ones in the same workspace. 
But I guess it's just something you read in the docs once.

gaspar09 (approved):

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -107,29 +141,43 @@ impl Workspace {
...
-        &self,
-    ) -> Result<HashMap<String, Vec<String>>, Error> {
+    pub async fn find_packages_with_graph(&self) -> Result<SerializablePackages, Error> {
         let packages = self.find_packages().await?;
 
         let workspace_path = match AbsoluteSystemPath::new(self.absolute_path.as_str()) {
             Ok(path) => path,
-            Err(e) => return Err(Error::from_reason(e.to_string())),
+            Err(e) => {
+                print!("Error: {:?}", e);
```
chris-olszewski, line 159:
```suggestion
```

chris-olszewski (approved):

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -73,13 +84,36 @@ impl Package {
         workspace_path: &AbsoluteSystemPath,
     ) -> Vec<Package> {
         let node = PackageNode::Workspace(PackageName::Other(self.name.clone()));
-        let ancestors = match graph.immediate_ancestors(&node) {
-            Some(ancestors) => ancestors,
+        let pkgs = match graph.immediate_ancestors(&node) {
```
gsoltis, line 87: Are you sure you want just `immediate_ancestors`? What about transitive dependents? If you do in fact just want immediates, maybe a comment or update the method name to indicate that.

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -73,13 +84,36 @@ impl Package {
...
+            .collect()
+    }
+
+    fn dependencies(
+        &self,
+        graph: &PackageGraph,
+        workspace_path: &AbsoluteSystemPath,
+    ) -> Vec<Package> {
+        let node = PackageNode::Workspace(PackageName::Other(self.name.clone()));
+        let pkgs = match graph.immediate_dependencies(&node) {
```
gsoltis, line 110: As above, this is getting just immediate dependencies, not the full transitive closure. Might be worth a comment at least.

packages/turbo-repository/rust/src/lib.rs
gsoltis, line 159: Probably should remove this line.

gsoltis (approved):
One errant `print!` and a couple clarifications, but otherwise LGTM

packages/turbo-repository/rust/src/lib.rs
mehulkar: just immediate for now. We are constructing the graph client side in our use case, we'll keep doing that for now and iterate here. I can't remember why I didn't do full tree on `ancestors` when I first did it, probably Rust-related hurdles.

PR #7619, opened 2024-03-05
feat: terminal pane component

chris-olszewski:
### Description

Adds a terminal pane component to the tui module. 

The terminal pane is responsible for:
 - Processing output from tasks into the appropriate `vt100::Parser`
 - Forward user input to the appropriate task stdin

### Testing Instructions

Added quick unit test to verify scrolling works as expected. True test is to just play around with it!

https://github.com/vercel/turbo/assets/4131117/35d6e069-4026-4277-a6fe-e5fb8ad2df65

Closes TURBO-2525


gsoltis (approved):

mehulkar (approved):
Looks great! The video doesn't show an example of the top part above the pane if new logs come in (e.g. with `-vvv`). Does that look ok too or does that need to be its own pane a well?

chris-olszewski:
>  if new logs come in

So we'll be using an "inline" display where the UI occupies the last `n` lines of the terminal and all of the lines above it are "persisted". [See inline example](https://github.com/ratatui-org/ratatui/tree/main/examples#inline) Logging will be sent to the buffer via [`insert_before`](https://docs.rs/ratatui/latest/ratatui/terminal/struct.Terminal.html#method.insert_before)

PR #7620, opened 2024-03-05
build: Optimize macros to reduce compile time

kdy1:
### Testing Instructions

Closes PACK-2571

### DIFF
--- a/crates/turbo-tasks-macros/src/derive/task_input_macro.rs
+++ b/crates/turbo-tasks-macros/src/derive/task_input_macro.rs
@@ -112,7 +112,9 @@ pub fn derive_task_input(input: TokenStream) -> TokenStream {
                         turbo_tasks::ConcreteTaskInput::List(value) => {
                             let mut #inputs_list_ident = value.iter();
 
-                            let discriminant = turbo_tasks::macro_helpers::ok_or_else_for_missing_element(#inputs_list_ident.next(), #ident_str)?;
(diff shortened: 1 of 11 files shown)


kdy1 (commented):
I think it's better to improve compile time gradually.

This reduces compile time from 3m 18s ~ 20s to about 3m

crates/turbo-tasks-macros/src/derive/task_input_macro.rs
```diff
@@ -104,13 +104,15 @@ pub fn derive_task_input(input: TokenStream) -> TokenStream {
...
+            let ident_str = ident.to_string();
+
             (
                 quote! {
                     match value {
                         turbo_tasks::ConcreteTaskInput::List(value) => {
                             let mut #inputs_list_ident = value.iter();
 
-                            let discriminant = #inputs_list_ident.next().ok_or_else(|| anyhow::anyhow!(concat!("missing discriminant for ", stringify!(#ident))))?;
+                            let discriminant = turbo_tasks::macro_helpers::ok_or_else_for_missing_element(#inputs_list_ident.next(), #ident_str)?;
```
ForsakenHarmony, line 115: Maybe just turn this into a `let-else`? (same for the other 2)
not the biggest fan of that helper function unless that single thing has a major impact on compile times

arlyon (approved):
Approving on behalf of repo

kdy1:
Closing as obsolete

PR #7621, opened 2024-03-05
fix: Fix purity lint for CSS Modules in swc mode

kdy1:
### Description

Closes PACK-2660

### Testing Instructions

I added a test.

### DIFF
--- a/crates/turbopack-css/src/process.rs
+++ b/crates/turbopack-css/src/process.rs
@@ -719,11 +719,25 @@ impl swc_core::css::visit::Visit for CssValidator {
             })
         }
 
-        fn is_forgiving_selector_not_pure(sel: &ForgivingComplexSelector) -> bool {}
+        fn is_forgiving_selector_not_pure(sel: &ForgivingComplexSelector) -> bool {
+            match sel {
+                ForgivingComplexSelector::ComplexSelector(sel) => is_complex_not_pure(sel),
+                ForgivingComplexSelector::ListOfComponentValues(_) => false,
+            }
+        }
 
-        fn is_forgiving_relative_selector_not_pure(sel: &ForgivingRelativeSelector) -> bool {}
+        fn is_forgiving_relative_selector_not_pure(sel: &ForgivingRelativeSelector) -> bool {
+            match sel {
+                ForgivingRelativeSelector::RelativeSelector(sel) => {
+                    is_relative_selector_not_pure(sel)
+                }
+                ForgivingRelativeSelector::ListOfComponentValues(_) => false,
+            }
+        }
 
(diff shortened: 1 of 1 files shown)


crates/turbopack-css/src/process.rs
```diff
@@ -708,25 +711,56 @@ const CSS_MODULE_ERROR: &str =
...
+            })
+        }
+
+        fn is_compound_not_pure(sel: &CompoundSelector) -> bool {
+            sel.subclass_selectors.iter().all(|sel| match sel {
+                SubclassSelector::Attribute { .. } => true,
+                SubclassSelector::PseudoClass(cls) => {
+                    cls.name.value == "not"
+                        || match &cls.children {
+                            Some(c) => c.iter().all(|c| match c {
```
sokra, line 729: This could be a `if let Some(c) = &cls.children` to avoid one level of nesting

crates/turbopack-css/src/process.rs
```diff
@@ -1069,6 +1103,12 @@ mod tests {
         }",
         );
 
+        assert_lint_success(
+            ":where(.main > *) {
+            color: red;
+        }",
+        );
+
```
sokra, line 1111: `:where(.main > *, .root > *)`

crates/turbopack-css/src/process.rs
```diff
@@ -708,25 +711,56 @@ const CSS_MODULE_ERROR: &str =
...
+                            Some(c) => c.iter().all(|c| match c {
+                                PseudoClassSelectorChildren::ComplexSelector(sel) => {
+                                    is_complex_not_pure(sel)
+                                }
+
+                                PseudoClassSelectorChildren::CompoundSelector(sel) => {
+                                    is_compound_not_pure(sel)
+                                }
+
+                                _ => false,
```
sokra, line 738: What are the other cases? Everything else will be pure?

crates/turbopack-css/src/process.rs
```diff
@@ -708,25 +711,56 @@ const CSS_MODULE_ERROR: &str =
...
+                }
+                swc_core::css::ast::ComplexSelectorChildren::Combinator(_) => true,
+            })
+        }
+
+        fn is_compound_not_pure(sel: &CompoundSelector) -> bool {
+            sel.subclass_selectors.iter().all(|sel| match sel {
+                SubclassSelector::Attribute { .. } => true,
+                SubclassSelector::PseudoClass(cls) => {
+                    cls.name.value == "not"
```
sokra, line 727:
```suggestion
                    cls.name.value == "not" || cls.name.value == "has"
```

sokra (approved):

PR #7624, opened 2024-03-05
Bug fix on kitchen-sink example

zsh77:
### Description

- moved tsup flags from `dev` script and `build` script into `tsup.config.ts` file
- ignored `.eslintrc.js` in both `storeFront` and `api` apps
- removed unused and duplicated `style.css` in `storeFront` app
- changed `rootDir` of `admin` app to "." so that `tsconfig.json` could include `vite.config.ts`
- changed `rootDir` of `blog` app to "." so that `tsconfig.json` could include `server.js` and `remix.config.js`
- improved imports from react (use `import React from "react"` instead of `import * as React from "react"` )
- included `next.config.js` in `tsconfig.json` in order to fix the parsing error in `next.config.js`
- set `api` app to use the default `module:"ESNEXT"` as the `moduleResolution` is set to `Bundler`
- fixed the error `@remix-run/vercel was not found` in apps/blog/server.js
- replaced `V2MetaFunction` with `MetaFunction`
- removed `future` object that was used to enable features of version 2 of remix as the version 2 is being used
- ran `pnpm i` and updated versions of dependencies

### Testing Instructions

Running `npx create-turbo@latest -e kitchen-sink` and then `pnpm dev` should work fine and there should be no errors.


examples/kitchen-sink/apps/api/tsconfig.json
```diff
@@ -2,8 +2,8 @@
   "extends": "@repo/typescript-config/base.json",
   "compilerOptions": {
     "lib": ["ES2015"],
-    "module": "CommonJS",
-    "outDir": "./dist",
+    // "module": "CommonJS",
```
anthonyshew, line 5: Was this intentional?

anthonyshew (changes requested):
Looks right to me! I only have my one question about what looks like a forgotten comment and then you'll need to resolve dependency conflicts from main.

Thank you!

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: db3b31129ebe5a8ed8b0c8185acb2a0652f20d67

examples/kitchen-sink/apps/api/tsconfig.json
zsh77: yes. module:"CommonJs" was causing an error as the moduleResolution from base.json is set to Bundler. I commented line 5 so that it could use module:"ESNext" from base.json. If there is a better fix for that please let me know.

anthonyshew: Could you delete it instead of commenting it out?

zsh77: yes of course

zsh77: @anthonyshew done

anthonyshew (approved):
Thank you once again!

PR #7627, opened 2024-03-05
perf: build telemetry client in background

chris-olszewski:
### Description

We move the construction of the telemetry client to a background thread so we can continue with execution while it does some cert reads etc.

### Testing Instructions

Before
Screenshot 2025-03-14 at 10 40 13 AM

After
Screenshot 2025-03-14 at 10 40 02 AM

As you can see the cost of moving the info to another thread to do the work ends up making and perf wins negligible and can easily slow down overall execution.

Closes TURBO-2530


chris-olszewski:
Closing as I finally ran some profiles and this doesn't give the results I was hoping to get.

PR #7628, opened 2024-03-05
fix(cache): fix restoration issues of directories with very long paths

### LINKED ISSUE #7410: Cached files differ from original command execution's output files
### Verify canary release

https://github.com/trappar/turbo-cache-missing-output-files

pnpm

Mac

1.12.4

### Describe the Bug

When building my app with turbo, I have it configured to cache the `.next` directory. I noticed that if I build, then delete the `.next` directory, then recover it from my turbo cache - the resulting files differ from the original files I had after building.

### Expected Behavior

The files cached by turbo cache should precisely match the original output files.

### To Reproduce

The reproduction repo I linked has a `reproduce-issue.sh` script that you can run to see the issue. Since this is a minimal reproduction repo, the number of divergent files between the fresh and cached copies are minimal. In a large enterprise monorepo I work on, the number of files not being copied is massive and is causing our servers to error/crash when we try to run a build which was recalled from the turbo cache.

Just for context, here's all that I'm doing in that script:

```
pnpm turbo build --force
mv .next output/force-build

pnpm turbo build
mv .next output/cache-hit-build

diff --brief --recursive output/force-build output/cache-hit-build
```

And this is producing the following list of file differences:
```
Only in output/cache-hit-build/standalone/node_modules/.pnpm/@swc+helpers@0.5.2/node_modules/@swc/helpers/_: _interop_require_
Only in output/cache-hit-build/standalone/node_modules/.pnpm/caniuse-lite@1.0.30001587/node_modules/caniuse-lite/data: feature
Only in output/cache-hit-build/standalone/node_modules/.pnpm/caniuse-lite@1.0.30001587/node_modules/caniuse-lite/dist: unpacke
Only in output/cache-hit-build/standalone/node_modules/.pnpm/next@14.1.0_react-dom@18.2.0_react@18.2.0/node_modules/@swc/helpers/_: _interop_require_
Only in output/cache-hit-build/standalone/node_modules/.pnpm/next@14.1.0_react-dom@18.2.0_react@18.2.0/node_modules/caniuse-lite/data: feature
Only in output/cache-hit-build/standalone/node_modules/.pnpm/next@14.1.0_react-dom@18.2.0_react@18.2.0/node_modules/caniuse-lite/dist: unpacke
```

> **Note:** strangely in the real project the original build contains files which are then missing from the cache-hit version. Here it's the opposite? This makes no sense to me.

### Additional context

The only context where I've seen this happen is a Next.js app using the "standalone" output mode. I suspect this is because the directory I'm trying to cache has long file paths and complex/internal links. It's possible that this could be distilled down even further into a reproduction that simply creates a directory structure like what Next.js creates, but this is probably the best repo since it's a fairly common application configuration.

I know this issue can be reproduced on Ubuntu and on Mac since I'm seeing these kinds of failures in CI, and I've reproduced it locally in that reproduction repo. I'm curious if this bug will happen on all OSes (windows?)

<sub>[TURBO-2414](https://linear.app/vercel/issue/TURBO-2414/cached-files-differ-from-original-command-executions-output-files)</sub>

chris-olszewski:
### Description

Fixes #7410 

This is just #6662, but applied to directories. See [Entry::path](https://docs.rs/tar/latest/tar/struct.Entry.html#method.path) for explanation as to why one should never rely on a header's path.

### Testing Instructions

In the first commit added a failing unit test, test passes after fix.

Also tested against reproduction: https://github.com/trappar/turbo-cache-missing-output-files

Just need to update `reproduce-issue.sh` to use the `turbo` in this branch:
```
...
Diffing the original force-build's .next directory with the cache-hit-build's .next directory
[0 olszewski@chriss-mbp] /tmp/turbo-cache-missing-output-files $ 
```

Closes TURBO-2532

### DIFF
--- a/crates/turborepo-cache/src/cache_archive/create.rs
+++ b/crates/turborepo-cache/src/cache_archive/create.rs
@@ -395,7 +395,10 @@ mod tests {
 
         let tar_path = tar_dir_path.join_component("test.tar");
         let mut archive = CacheWriter::create(&tar_path)?;
-        let really_long_file = AnchoredSystemPath::new("this-is-a-really-really-really-long-path-like-so-very-long-that-i-can-list-all-of-my-favorite-directors-like-edward-yang-claire-denis-lucrecia-martel-wong-kar-wai-even-kurosawa").unwrap();
+        let base = "this-is-a-really-really-really-long-path-like-so-very-long-that-i-can-list-all-of-my-favorite-directors-like-edward-yang-claire-denis-lucrecia-martel-wong-kar-wai-even-kurosawa";
+        let file_name = format!("{base}.txt");
+        let really_long_file = AnchoredSystemPath::new(&file_name).unwrap();
+        let really_long_dir = AnchoredSystemPath::new(base).unwrap();
         let really_long_symlink = AnchoredSystemPath::new("this-is-a-really-really-really-long-symlink-like-so-very-long-that-i-can-list-all-of-my-other-favorite-directors-like-jim-jarmusch-michelangelo-antonioni-and-terrence-malick-symlink").unwrap();
 
         let really_long_path = archive_dir_path.resolve(really_long_file);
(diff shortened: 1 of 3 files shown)


mehulkar (approved):

NicholasLYang (approved):
I really wish the tar library had better protections against this.

PR #7640, opened 2024-03-06
very dynamic requests will only lead to a warning

sokra:
### Description

they no longer lead to an additional module not found error

### Testing Instructions

Closes PACK-2673

### DIFF
--- a/crates/turbopack-ecmascript/src/references/dynamic_expression.rs
+++ b/crates/turbopack-ecmascript/src/references/dynamic_expression.rs
@@ -1,0 +1,71 @@
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use swc_core::quote;
+use turbo_tasks::{debug::ValueDebugFormat, trace::TraceRawVcs, Vc};
+
+use super::AstPath;
+use crate::{
+    chunk::EcmascriptChunkingContext,
+    code_gen::{CodeGenerateable, CodeGeneration},
+    create_visitor,
+};
+
+#[derive(PartialEq, Eq, TraceRawVcs, Serialize, Deserialize, ValueDebugFormat)]
+enum DynamicExpressionType {
+    Promise,
+    Normal,
+}
+
+#[turbo_tasks::value]
+pub struct DynamicExpression {
+    path: Vc<AstPath>,
+    ty: DynamicExpressionType,
+}
+
+#[turbo_tasks::value_impl]
+impl DynamicExpression {
+    #[turbo_tasks::function]
+    pub fn new(path: Vc<AstPath>) -> Vc<Self> {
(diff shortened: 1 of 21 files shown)


timneutkens (approved):

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -1063,20 +1065,25 @@ pub(crate) async fn analyse_ecmascript_module_internal(
                         DiagnosticId::Lint(
                             errors::failed_to_analyse::ecmascript::NEW_URL_IMPORT_META.to_string(),
                         ),
-                    )
+                    );
```
ForsakenHarmony, line 1068: Could have just added a `continue` here 🤔

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -1499,13 +1522,13 @@ async fn handle_call<G: Fn(Vec<Effect>) + Send + Sync>(
                             errors::failed_to_analyse::ecmascript::NODE_PRE_GYP_FIND.to_string(),
                         ),
                     );
-                    return Ok(());
```
ForsakenHarmony, line 1502: early return is definitely nicer to read than an else branch (kinda applies to a lot of the changes here)

crates/turbopack-ecmascript/src/references/dynamic_expression.rs
```diff
@@ -0,0 +1,71 @@
...
+    async fn code_generation(
+        &self,
+        _context: Vc<Box<dyn EcmascriptChunkingContext>>,
+    ) -> Result<Vc<CodeGeneration>> {
+        let path = &self.path.await?;
+
+        let visitor = match self.ty {
+            DynamicExpressionType::Normal => {
+                create_visitor!(path, visit_mut_expr(expr: &mut Expr) {
+                    *expr = quote!("(() => { const e = new Error(\"Cannot find module as expression is too dynamic\"); e.code = 'MODULE_NOT_FOUND'; throw e; })()" as Expr);
```
ForsakenHarmony, line 56: this will completely throw away information about what would have been called

can we add something back so it shows up in the stack trace?

ForsakenHarmony: Giving the inline function a name would probably work

sokra: Due to source mapping the error will point to the actual user code.

ForsakenHarmony: But the stack trace still won't contain a `require`/`import`

PR #7646, opened 2024-03-06
Remove .mp4 from Turbopack static assets docs

anthonyshew:
Title!

Closes TURBO-2550

### DIFF
--- a/docs/pages/pack/docs/features/static-assets.mdx
+++ b/docs/pages/pack/docs/features/static-assets.mdx
@@ -8,7 +8,6 @@ Importing static assets works out of the box with Turbopack:
 
 ```ts
 import img from './img.png'
-import video from './video.mp4'
 import audio from './audio.wav'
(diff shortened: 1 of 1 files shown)


docs/pages/pack/docs/features/static-assets.mdx
```diff
@@ -8,7 +8,6 @@ Importing static assets works out of the box with Turbopack:
 
 ```ts
 import img from './img.png'
-import video from './video.mp4'
 import audio from './audio.wav'
```
ForsakenHarmony, line 11: We also need to remove `.wav`

ForsakenHarmony (approved):

sam3d:
Thanks for the fix @anthonyshew! Is there an webpack's `asset/resource`-like loader supported in turbopack?

ryanleecode:
What would it take to support mp4?

PR #7649, opened 2024-03-06
chore: rename vendored vt100 crate

chris-olszewski:
### Description

Renaming our vendored `vt100` crate to avoid confusion since `tui_term` comes with prebuilt `vt100` integration, but won't work with our vendored crate. 

### Testing Instructions

CI

Closes TURBO-2553

### DIFF
--- a/crates/turborepo-vt100/Cargo.toml
+++ b/crates/turborepo-vt100/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "turborepo_vt100"
+name = "turborepo-vt100"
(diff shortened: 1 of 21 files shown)


crates/turborepo-vt100/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "vt100"
```
NicholasLYang, line 2: super tiny nit but we generally use `-` for our crate names

NicholasLYang (approved):

PR #7667, opened 2024-03-07
fix(turborepo): Logout

NicholasLYang:
### Description

`turbo logout` wasn't actually logging out with the vercel token.

Also refactored the config code to produce an `AbsoluteSystemPathBuf` so we don't have to keep on doing that. There are other refactors that could be done but I wanted to keep this relatively small.

### Testing Instructions

Closes TURBO-2570

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11215,6 +11215,7 @@ dependencies = [
  "turbopack-resolve",
  "turbopack-swc-utils",
  "url 2.4.1",
+ "urlencoding",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


Zertsov:
Did we want to affect the Vercel folder? I intentionally made it so `logout` doesn't touch the vercel folder since we don't own the `vc login`, but if we're ok with that then this is fine by me.

Zertsov (approved):

NicholasLYang:
I think we should since otherwise logout doesn't actually log out

PR #7683, opened 2024-03-08
docs: Update testing configuration documentation for clarity

sooster910:
### Description

The Testing section is really helpful, but it would be clearer by specifying exactly where each of these snippets should be placed.

Before: 
Screenshot 2024-03-09 at 8 08 59 AM

After: 
Screenshot 2024-03-09 at 8 14 03 AM

Looking forward to your thoughts and feedback.
Thanks.


anthonyshew (approved):
Good call, thanks!

Issue #2129, opened 2022-09-30
improve output when using remote caching


ilovett:
### Describe the feature you'd like to request

It seems to be hard to determine if remote cache is even being used sometimes.

It might be great to add more explicit information on the output *if remote caching is enabled*

### Describe the solution you'd like

When remote caching is enabled, show the number of jobs which were taken from local vs. remote

```
 Tasks:    3 successful, 3 total
Cached:    3 cached, 2 remote, 1 local, 3 total
  Time:    765ms >>> FULL TURBO
```

I'm not even sure if it's possible to have partial remote & partial local but I assume it would be at some point.

### Describe alternatives you've considered

I used to see `Remote computation` something something experimental.  I'd like to see sort of confirmation that that is enabled and working correctly, and especially if misconfigured, eg: expired token :/

mehulkar:
Yep this was just released in 1.5.5!

nathanhammond:
I'm reopening; I want to capture the local vs. remote distinction from this proposal in the final output summary.

chris-olszewski:
Team decided that we think the current cache status line is the right amount of information.

PR #7697, opened 2024-03-11
fix: add extra detail to summary showing items from remote cache

arlyon:
### Description

A quick fix that closes https://github.com/vercel/turbo/issues/2129

It adds a small additional UI when tasks are loaded from the remote
cache. Tasks that were not fetched from cache do not have the extra
annotation and retain the original formatting.

### Testing Instructions

Run a build with remote cache set up:

 Tasks:    0 successful, 2 total
Cached:    2 (of which 2 remote) cached, 2 total
  Time:    1.141s

Run another build:

 Tasks:    0 successful, 2 total
Cached:    2 cached, 2 total
  Time:    1.141s

Closes TURBO-2590

### DIFF
--- a/crates/turborepo-lib/src/run/summary/execution.rs
+++ b/crates/turborepo-lib/src/run/summary/execution.rs
@@ -4,6 +4,7 @@ use chrono::{DateTime, Local};
 use serde::Serialize;
 use tokio::sync::mpsc;
 use turbopath::{AbsoluteSystemPathBuf, AnchoredSystemPath};
+use turborepo_cache::CacheSource;
 use turborepo_ui::{color, cprintln, BOLD, BOLD_GREEN, BOLD_RED, MAGENTA, UI, YELLOW};
 
 use super::TurboDuration;
@@ -36,6 +37,8 @@ pub struct ExecutionSummary<'a> {
     failed: usize,
     // number of tasks that had a cache hit
     cached: usize,
+    // number of cache hits that came from the remote cache
+    cached_remote: usize,
     // number of tasks that started
     attempted: usize,
     pub(crate) start_time: i64,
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -159,14 +168,23 @@ impl<'a> ExecutionSummary<'a> {
     fn successful(&self) -> usize {
         self.success + self.cached
     }
+
+    fn add_cached(&mut self, source: CacheSource) {
+        self.cached += 1;
```
arlyon, line 173: To preserve run summary output we just add a new field for remote rather than splitting into two. Naturally cached - cached_remote = cached_local if we need it one day.

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -159,14 +168,23 @@ impl<'a> ExecutionSummary<'a> {
     fn successful(&self) -> usize {
         self.success + self.cached
     }
+
+    fn add_cached(&mut self, source: CacheSource) {
```
chris-olszewski, line 172: I don't see this used anywhere so I think it can get removed? In general I think we don't want to be changing the counts on the `ExecutionSummary` and have all modifications happen to `SummaryState`

crates/turborepo-lib/src/run/summary/execution.rs
```diff
@@ -82,6 +86,12 @@ impl<'a> ExecutionSummary<'a> {
             String::new()
         };
 
+        let cache_line = if self.cached_remote > 0 {
+            format!("{} (of which {} remote)", self.cached, self.cached_remote)
```
chris-olszewski, line 90: Don't want to bikeshed this a ton, but I feel like we can simplify this to just `10 (8 remote)` or `10 (2 local, 8 remote)`.

chris-olszewski (approved):

mehulkar (commented):
I'd love to see something like this in our run summary:

```
interface Cache {
   restoredFrom: "local" | "remote" | undefined,
   hit: {
      local: boolean | "unknown",
      remote: boolean | "unknown",
   },
   enabled: {
      local: boolean,
      remote: boolean,
   }
};
```

PR #7699, opened 2024-03-11
feat: add entire screen viewer to vt100

chris-olszewski:
### Description

In order to properly persist a task's output we need to display the full terminal contents including whatever might be in the scrollback buffer. This PR adds an alternative screen where the row index includes the scrollback rows. The `Screen` trait added in https://github.com/a-kenji/tui-term/pull/152 will be implemented for `EntireScreen` allowing for us to use this structure with the `PseudoTerminal` widget.

### Testing Instructions

Added some basic unit tests to verify that scrollback gets printed.

Closes TURBO-2592


arlyon (approved):
Need to go have a dig around in vt100, this is super cool.

PR #7701, opened 2024-03-11
chore: generalize cache terminal output

chris-olszewski:
### Description

This PR removes the cache from using concrete implementations for terminal output and instead have it take a generic writer. 

This prefactor is needed for the TUI as we cannot just write to terminal as the cursor could be in an unexpected position.

### Testing Instructions

Our existing integration tests capture this behavior pretty well. Ran some tasks to double check that things appeared identical.

Closes TURBO-2596

### DIFF
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -1,6 +1,5 @@
 use std::{io::Write, sync::Arc, time::Duration};
 
-use console::StyledObject;
 use tracing::{debug, error};
 use turbopath::{
     AbsoluteSystemPath, AbsoluteSystemPathBuf, AnchoredSystemPath, AnchoredSystemPathBuf,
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/run/cache.rs
```diff
@@ -463,3 +471,10 @@ impl ConfigCache {
         Ok(FileHashes(hash_object).hash())
     }
 }
+
+// attempt to write message to writer, swallowing any errors encountered
+fn failable_write(mut writer: impl Write, message: &str) {
```
chris-olszewski, line 476: This is the existing behavior of `PrefixedUI::output`

arlyon: Small nit, I think rust tends to use fallible over failable

arlyon (approved):
LGTM except naming thing

PR #7710, opened 2024-03-12
fix(image): fix build errors for the features

kwonoj:
### Description

Update missing fixes for the features.

Closes PACK-2722

### DIFF
--- a/crates/turbopack-image/src/process/mod.rs
+++ b/crates/turbopack-image/src/process/mod.rs
@@ -268,13 +268,10 @@ fn encode_image(image: DynamicImage, format: ImageFormat, quality: u8) -> Result
         }
         #[cfg(feature = "webp")]
         ImageFormat::WebP => {
-            use image::codecs::webp::{WebPEncoder, WebPQuality};
-            WebPEncoder::new_with_quality(&mut buf, WebPQuality::lossy(quality)).write_image(
-                image.as_bytes(),
-                width,
-                height,
-                image.color(),
-            )?;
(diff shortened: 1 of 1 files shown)


crates/turbopack-image/src/process/mod.rs
```diff
@@ -268,13 +268,10 @@ fn encode_image(image: DynamicImage, format: ImageFormat, quality: u8) -> Result
...
         ImageFormat::WebP => {
-            use image::codecs::webp::{WebPEncoder, WebPQuality};
-            WebPEncoder::new_with_quality(&mut buf, WebPQuality::lossy(quality)).write_image(
-                image.as_bytes(),
-                width,
-                height,
-                image.color(),
-            )?;
+            use image::codecs::webp::WebPEncoder;
+            let encoder = WebPEncoder::new_lossless(&mut buf);
```
ForsakenHarmony, line 272: Do we want it to be lossless? Probably not?

kwonoj: doesn't seem to other ctor even, if I'm reading correctly https://github.com/image-rs/image/blob/6edf8ae492c4bb1dacb41da88681ea74dab1bab3/src/codecs/webp/encoder.rs#L19

ForsakenHarmony: that's unfortunate, maybe create a linear ticket to look into some way to get lossy webp back?

kwonoj: yeah will do

PR #7716, opened 2024-03-13
fix path to typescript config

darklight9811:
### Description

Fixes the file path to the base.json typescript config file

### Testing Instructions

Just boot up the project, you should see that it won't complain anymore about tsconfig


arlyon (approved):

anthonyshew:
This file isn't actually needed at all! Can you delete it instead?

darklight9811:
@anthonyshew this is breaking eslint, will require a rework on that to make it viable

PR #7720, opened 2024-03-13
generate correct async module handling for side effects optimization

sokra:
### Description

Async module handling was missing for side effects optimized modules

### Testing Instructions

Closes PACK-2733

### DIFF
--- a/crates/turbopack-ecmascript/src/side_effect_optimization/locals/chunk_item.rs
+++ b/crates/turbopack-ecmascript/src/side_effect_optimization/locals/chunk_item.rs
@@ -104,7 +104,7 @@ impl ChunkItem for EcmascriptModuleLocalsChunkItem {
     #[turbo_tasks::function]
     async fn is_self_async(&self) -> Result<Vc<bool>> {
         if let Some(async_module) = *self.module.get_async_module().await? {
-            Ok(Vc::cell(*async_module.is_self_async().await?))
+            Ok(async_module.is_self_async())
         } else {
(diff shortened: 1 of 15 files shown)


crates/turbopack-ecmascript/src/side_effect_optimization/locals/chunk_item.rs
```diff
@@ -100,4 +100,13 @@ impl ChunkItem for EcmascriptModuleLocalsChunkItem {
     fn module(&self) -> Vc<Box<dyn Module>> {
         Vc::upcast(self.module)
     }
+
+    #[turbo_tasks::function]
+    async fn is_self_async(&self) -> Result<Vc<bool>> {
+        if let Some(async_module) = *self.module.get_async_module().await? {
+            Ok(Vc::cell(*async_module.is_self_async().await?))
```
kdy1, line 107: Is there a reason for dropping connection here?

sokra: What do you mean? which connection is dropped?

kdy1: I meant the creation of Vc<bool> with an awaited value. IIRC, it will make this whole function execute again when the value is changed, right?

sokra: ah yes

PR #7726, opened 2024-03-13
feat(node): augment webpack loader's logger context

kwonoj:
### Description

Trying to support sass-loader to not fail if it requires specific logger from loadercontext.

Closes PACK-2737


crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -333,7 +333,7 @@ const transform = (
           emitWarning: makeErrorEmitter("warning", ipc),
           emitError: makeErrorEmitter("error", ipc),
           getLogger(name: unknown) {
-            return (type: unknown, args: unknown) => {
+            const logger = (type: unknown, args: unknown) => {
```
ForsakenHarmony, line 336: Would prefer this to not be an arrow function, but it actually uses `this` so it would need additional changes.

PR #7730, opened 2024-03-14
fix(css): Do not resolve `url`s in supports query

kdy1:
### Description

 - Closes https://github.com/vercel/next.js/issues/63102
 - Closes PACK-2717

### Testing Instructions

I added a test.


kdy1:
It seems like we don't resolve urls in `turbopack-tests` anyway

sokra (approved):

PR #7731, opened 2024-03-14
feat(turbopack): Support marking packages as side-effect-free

kdy1:
### Description

This is required to support `experimental.optimizePackageImports` of `next.config.js`.

next.js counterpart: https://github.com/vercel/next.js/pull/63268

Closes PACK-2527

### Testing Instructions

Look at the test case I added to https://github.com/vercel/next.js/pull/63268

### DIFF
--- a/crates/turbopack-ecmascript/src/chunk/placeable.rs
+++ b/crates/turbopack-ecmascript/src/chunk/placeable.rs
@@ -21,8 +21,11 @@ pub trait EcmascriptChunkPlaceable: ChunkableModule + Module + Asset {
     fn get_async_module(self: Vc<Self>) -> Vc<OptionAsyncModule> {
         Vc::cell(None)
     }
-    fn is_marked_as_side_effect_free(self: Vc<Self>) -> Vc<bool> {
-        is_marked_as_side_effect_free(self.ident().path())
+    fn is_marked_as_side_effect_free(
+        self: Vc<Self>,
+        optimized_packages: Vc<Vec<String>>,
+    ) -> Vc<bool> {
+        is_marked_as_side_effect_free(self.ident().path(), optimized_packages)
     }
 }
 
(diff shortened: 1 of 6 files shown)


kdy1:
Repro works.

https://github.com/kdy1/repro-pack-2527

image

crates/turbopack/src/module_options/module_options_context.rs
```diff
@@ -159,6 +159,8 @@ pub struct ModuleOptionsContext {
     pub ignore_dynamic_requests: bool,
 
     pub use_swc_css: bool,
+
+    pub side_effect_free_packages: Vc<Vec<String>>,
```
sokra, line 163: Should be use a `Vc<Glob>` instead? That would be more flexible.

crates/turbopack-ecmascript/src/chunk/placeable.rs
```diff
@@ -36,8 +39,19 @@ enum SideEffectsValue {
...
     if let FileJsonContent::Content(content) = &*package_json.read_json().await? {
+        if let Some(pacakge_name) = content.get("name").and_then(|v| v.as_str()) {
+            if side_effect_free_packages
+                .await?
+                .iter()
+                .any(|v| pacakge_name == v)
+            {
+                return Ok(SideEffectsValue::Constant(false).cell());
+            }
+        }
```
sokra, line 53: I don't feel good about the `package.json` name field for that. That's also not what next.js is using for `optimizePackageImports`. I think it should be a glob on the file path. You can apply that in `fn is_marked_as_side_effect_free`

crates/turbopack/src/module_options/module_options_context.rs
kdy1: `the trait bound `Glob: turbo_tasks::Dynamic<Box<(dyn ValueDefault + 'static)>>` is not satisfied`

Implementing default for `Glob` does not look like a good idea.

kdy1: I used `Vec<String>` for config and `Glob` for actual operation.

sokra: This could be `pub side_effect_free_packages: Option<Vc<Glob>>`, but we can refactor that in a follow-up

sokra (approved):

PR #7733, opened 2024-03-14
feat: enable new UI

chris-olszewski:
### Description

This PR adds an `App` widget and hooks it up to the task graph execution. The use of the TUI is controlled by the `TURBO_EXPERIMENTAL_UI` env var.

Note that the TUI is experimental so this current iteration does not attempt to tackle the following issues in this PR:
 - Proper integration with our `tracing` setup. Writing to stderr (or stdout) naively will cause us to overwrite the TUI as the cursor is probably not in the correct position. Currently the best way to run with increased verbosity is to pipe stderr to a file e.g. `TURBO_EXPERIMENTAL_UI=true turbo build -vv 2> /tmp/logs.txt`
 - "Proper" signal handling. Since we have [ISIG](https://man7.org/linux/man-pages/man3/termios.3.html) flagged off we are in charge of generating a SIGINT when `Ctrl-C` is pressed. We currently stop rendering the TUI and shut down on `Ctrl-C`, but we should do a [raise](https://man7.org/linux/man-pages/man3/raise.3.html)/Windows equivalent. https://github.com/vercel/turbo/pull/7737
 - Persistent configuration of enabling the TUI i.e. `turbo.json` extension
 - Proper interaction with existing flags that modify output e.g. `--log-order`/`--output-logs`
 - Responding to terminal resizing events
 - Final TUI state, already there are things I want to change e.g. have task render red on failure or rainbow on cache hit
 - Interactive tasks: #7735

Note for reviewers: Each commit is pretty self contained and should be reviewed on it's own.

### Testing Instructions

Primary thing that should be tested is that when `TURBO_EXPERIMENTAL_UI` is not set, there aren't any behavior changes.
The primary place where behavior could change is us changing reading task input to no longer be line based. This should be covered by #7728, but it is still worth manually checking since our prysk tests do not cover this codepath.

To test the TUI set `TURBO_EXPERIMENTAL_UI=true` and run `turbo`.

https://github.com/vercel/turbo/assets/4131117/9097dc07-c76a-40af-9a81-26fcb1c18a19


crates/turborepo-lib/src/process/child.rs
```diff
@@ -609,31 +609,36 @@ impl Child {
...
+            let mut last_byte = None;
             loop {
-                match stdout_lines.read_until(b'\n', &mut buffer) {
-                    Ok(0) => break,
-                    Ok(_) => {
-                        if byte_tx.blocking_send(buffer.clone()).is_err() {
+                match stdout_lines.read(&mut buffer) {
+                    Ok(0) => {
+                        if !matches!(last_byte, Some(b'\n')) {
+                            // Ignore as this fails as we already are shutting down
```
mehulkar, line 618:
```suggestion
                            // Ignore if this fails as we already are shutting down
```

crates/turborepo-lib/src/process/child.rs
```diff
@@ -609,31 +609,36 @@ impl Child {
...
+                            byte_tx.blocking_send(vec![b'\n']).ok();
+                        }
+                        break;
+                    }
+                    Ok(n) => {
+                        let mut bytes = Vec::with_capacity(n);
+                        bytes.extend_from_slice(&buffer[..n]);
+                        last_byte = bytes.last().copied();
+                        if byte_tx.blocking_send(bytes).is_err() {
```
mehulkar, line 627: is this sending the last byte twice? once here in `byte_tx.blocking_send(bytes)` and once in the other part of the match? Or does `bytes.last().copied()` _remove_ that last byte from `bytes`?

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -136,6 +149,17 @@ impl<'a> Visitor<'a> {
...
+
+        let (ui, render_thread_handle) = if self.experimental_ui {
+            let task_names = engine.tasks_with_command(&self.package_graph);
+
+            let (handle, receiver) = tui::AppSender::new();
+            let app = tokio::task::spawn_blocking(move || tui::run_app(task_names, receiver));
+            (Some(handle), Some(app))
+        } else {
+            (None, None)
+        };
```
mehulkar, line 161: I see ths is being setup even for dry runs, is that needed?

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -740,6 +788,7 @@ impl ExecContext {
                     (None, false) => self.manager.stop().await,
                     // Send task
                     (Some(client), true) => {
+                        // TODO don't do this
```
mehulkar, line 791: is this saying not to send logs to spaces if the TUI was enabled?

mehulkar (approved):

crates/turborepo-lib/src/process/child.rs
chris-olszewski: The last byte only gets sent via `byte_tx.blocking_send(bytes)`. The other branch arm is triggered when `read` returns 0 meaning the stdout has been closed. In that case we check if the output for the process ended with a newline and add one only if it didn't.

We need to add the newline as otherwise task output can look mangled since the next line we print would be appended to the output line with no trailing newline.

crates/turborepo-lib/src/task_graph/visitor.rs
chris-olszewski: Oops, that comment is leftover from a hack that I fixed. Will remove comment since the hack is no longer there

PR #7735, opened 2024-03-14
feat: add ability to forward input to tasks

chris-olszewski:
### Description

Add ability to forward input to tasks: `Enter` will start forwarding input and `Ctrl-Z` will stop forwarding input to the selected task.

There are definitely gaps in our input forwarding, but for basic text and standard keys such as arrow keys we should do the right thing. Hopefully users do not require handling `Ctrl-Alt-Home` being correctly forwarded to the underlying tasks.

This PR does not add the ability to configure non-persistent tasks to take input

### Testing Instructions

Test it out with persistent tasks e.g. `with-vite` example

https://github.com/vercel/turbo/assets/4131117/abf2263b-cbf2-44d7-9440-1c11c0e28808

Closes TURBO-2639

### DIFF
--- a/crates/turborepo-ui/src/tui/input.rs
+++ b/crates/turborepo-ui/src/tui/input.rs
@@ -133,9 +133,6 @@ fn encode_key(key: KeyEvent) -> Vec<u8> {
                 buf.push(0x1b as char);
             }
             buf.push(c);
-            if code == Enter {
-                buf.push('\n');
-            }
         }
 
         Tab => {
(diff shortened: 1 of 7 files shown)


crates/turborepo-ui/src/tui/input.rs
```diff
@@ -20,16 +20,291 @@
...
-        crossterm::event::KeyCode::Down => Some(Event::Down),
         // TODO: we should send a ourselves a SIGINT/CtrlC event
-        crossterm::event::KeyCode::Char('c')
-            if key_event.modifiers == crossterm::event::KeyModifiers::CONTROL =>
-        {
+        KeyCode::Char('c') if key_event.modifiers == crossterm::event::KeyModifiers::CONTROL => {
             Some(Event::Stop)
         }
+        // Interactive branches
+        KeyCode::Char('z') if key_event.modifiers == crossterm::event::KeyModifiers::CONTROL => {
```
gsoltis, line 30: Do we need to guard this with `if interact` also?

gsoltis (approved):
Just one question re: Ctrl + Z

PR #7739, opened 2024-03-15
fix(turbopack): Use correct method for the proxy of `fetch`

kdy1:
### Description

We should use [`Proxy::all`](https://docs.rs/reqwest/0.11.26/reqwest/struct.Proxy.html#method.all) instead of `Proxy::http` or `Proxy::https`.

Closes PACK-2448

### Testing Instructions

See next.js counterpart: https://github.com/vercel/next.js/pull/63309

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -581,8 +581,8 @@ dependencies = [
  "bitflags 1.3.2",
  "bytes 1.5.0",
  "futures-util",
- "http",
- "http-body",
+ "http 0.2.11",
+ "http-body 0.4.5",
  "hyper 0.14.28",
  "itoa",
  "matchit",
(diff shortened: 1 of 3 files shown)


kwonoj (approved):
I tried this but it still didn't pass the test in next.js - did you able to make it actually work?

kdy1:
Actually nope, I think it's a bug of `reqwest`.

kdy1:
It works if I add 

```rust
let url = url.replace("https://", "http://");
```

so I think `reqwest` is not sending https requests to the proxy, regardless of the method.

kwonoj:
That's what I saw as well (https://vercel.slack.com/archives/C03EWR7LGEN/p1707871876798169?thread_ts=1707788191.248449&cid=C03EWR7LGEN)

kdy1:
Closing in favor of https://github.com/vercel/turbo/pull/8165

PR #7742, opened 2024-03-15
feat(turbopack-ecmascript): cache external modules with wrapper

ForsakenHarmony:
### Description

This should prevent issues when an external module is removed from the require cache leading to 2 version used from the bundle.

### DIFF
--- a/crates/turbopack-ecmascript/src/chunk/chunk_type.rs
+++ b/crates/turbopack-ecmascript/src/chunk/chunk_type.rs
@@ -11,8 +11,8 @@ use super::{
     EcmascriptChunk, EcmascriptChunkContent, EcmascriptChunkItem, EcmascriptChunkingContext,
 };
 
-#[derive(Default)]
 #[turbo_tasks::value]
+#[derive(Default)]
 pub struct EcmascriptChunkType {}
 
 #[turbo_tasks::value_impl]
--- a/crates/turbopack-ecmascript/src/references/external_module.rs
+++ b/crates/turbopack-ecmascript/src/references/external_module.rs
@@ -1,0 +1,248 @@
+use std::{fmt::Display, io::Write};
+
+use anyhow::{Context, Result};
+use serde::{Deserialize, Serialize};
+use turbo_tasks::{trace::TraceRawVcs, TaskInput, Vc};
+use turbo_tasks_fs::{glob::Glob, rope::RopeBuilder, FileContent, FileSystem, VirtualFileSystem};
+use turbopack_core::{
+    asset::{Asset, AssetContent},
+    chunk::{AsyncModuleInfo, ChunkItem, ChunkType, ChunkableModule, ChunkingContext},
+    ident::AssetIdent,
+    module::Module,
+    reference::ModuleReferences,
+};
+
+use crate::{
+    chunk::{
+        EcmascriptChunkItem, EcmascriptChunkItemContent, EcmascriptChunkPlaceable,
+        EcmascriptChunkType, EcmascriptChunkingContext, EcmascriptExports,
+    },
+    references::async_module::{AsyncModule, OptionAsyncModule},
+    utils::StringifyJs,
+    EcmascriptModuleContent,
+};
+
+#[turbo_tasks::function]
+fn layer() -> Vc<String> {
+    Vc::cell("external".to_string())
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize, TraceRawVcs, TaskInput)]
+pub enum CachedExternalType {
+    CommonJs,
+    EcmaScriptViaRequire,
+    EcmaScriptViaImport,
+}
+
+impl Display for CachedExternalType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            CachedExternalType::CommonJs => write!(f, "cjs"),
+            CachedExternalType::EcmaScriptViaRequire => write!(f, "esm_require"),
+            CachedExternalType::EcmaScriptViaImport => write!(f, "esm_import"),
+        }
+    }
+}
+
+#[turbo_tasks::value]
+pub struct CachedExternalModule {
+    pub request: String,
+    pub external_type: CachedExternalType,
+}
+
+#[turbo_tasks::value_impl]
+impl CachedExternalModule {
+    #[turbo_tasks::function]
+    pub fn new(request: String, external_type: CachedExternalType) -> Vc<Self> {
+        Self::cell(CachedExternalModule {
+            request,
+            external_type,
+        })
+    }
+
+    #[turbo_tasks::function]
+    pub fn content(&self) -> Result<Vc<EcmascriptModuleContent>> {
+        let mut code = RopeBuilder::default();
+
+        if self.external_type == CachedExternalType::EcmaScriptViaImport {
+            writeln!(
+                code,
+                "const mod = await __turbopack_external_import__({});",
+                StringifyJs(&self.request)
+            )?;
+        } else {
+            writeln!(
+                code,
+                "const mod = __turbopack_external_require__({});",
+                StringifyJs(&self.request)
+            )?;
+        }
+
+        writeln!(code)?;
+
+        if self.external_type == CachedExternalType::CommonJs {
+            writeln!(code, "module.exports = mod;")?;
+        } else {
+            writeln!(code, "__turbopack_dynamic__(mod);")?;
+        }
+
+        Ok(EcmascriptModuleContent {
+            inner_code: code.build(),
+            source_map: None,
+            is_esm: true,
+        }
+        .cell())
+    }
+}
+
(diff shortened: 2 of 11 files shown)


crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -0,0 +1,226 @@
...
+            )
+            .unwrap();
+        } else {
+            writeln!(
+                code,
+                "const mod = __turbopack_external_require__({});",
+                StringifyJs(&self.request)
+            )
+            .unwrap();
+        }
```
ForsakenHarmony, line 77: Do we need to throw an error for some `ExternalType`?

Also do we need to keep track of the way this is imported from the requiring module?

Eianaftyio (commented):
Turbo

Eianaftyio (commented):
Fear

Eianaftyio:
.patch

crates/turbopack-ecmascript/src/references/external_module.rs
Eianaftyio: # Regarding ExternalType, we should consider throwing an error if it's used in an unexpected way.
# It's important to handle unexpected behavior gracefully to ensure the reliability of our code.

# Keeping track of import sources can help with understanding dependencies and debugging issues later on.
# We can use comments or documentation to note where modules are imported from and why.

crates/turborepo-auth/src/auth/mod.rs
sokra, line 1: Why changing this file?

crates/turbopack-ecmascript/src/chunk/chunk_type.rs
```diff
@@ -11,8 +11,8 @@ use super::{
     EcmascriptChunk, EcmascriptChunkContent, EcmascriptChunkItem, EcmascriptChunkingContext,
 };
 
+#[turbo_tasks::value(shared)]
 #[derive(Default)]
-#[turbo_tasks::value]
 pub struct EcmascriptChunkType {}
```
sokra, line 16: This shouldn't have a public cell method. It should only use `ValueDefault` to make it a singleton

crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -0,0 +1,226 @@
...
+        })
+    }
+
+    #[turbo_tasks::function]
+    pub fn is_async_vc(&self) -> Vc<bool> {
+        Vc::cell(self.is_async())
+    }
+
+    #[turbo_tasks::function]
+    pub fn content(&self) -> Vc<EcmascriptModuleContent> {
```
sokra, line 60: Instead of all the unwraps in this method, you could return a `Result` here:


```suggestion
    pub fn content(&self) -> Result<Vc<EcmascriptModuleContent>> {
```

crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -0,0 +1,226 @@
...
+            writeln!(
+                code,
+                "const mod = __turbopack_external_require__({});",
+                StringifyJs(&self.request)
+            )
+            .unwrap();
+        }
+
+        writeln!(code).unwrap();
+        writeln!(code, "__turbopack_export_namespace__(mod);").unwrap();
```
sokra, line 80: I feel like a `__turbopack_external_require__` should export via `module.exports =` and have `get_exports` returning `Dynamic`

crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -0,0 +1,226 @@
...
+    #[turbo_tasks::function]
+    pub fn is_async_vc(&self) -> Vc<bool> {
+        Vc::cell(self.is_async())
+    }
+
+    #[turbo_tasks::function]
+    pub fn content(&self) -> Vc<EcmascriptModuleContent> {
+        let mut code = RopeBuilder::default();
+
+        if self.is_async() {
```
sokra, line 63: I would match on `ExternalType` and `import_externals` here, instead of using `is_async`

crates/turbopack-ecmascript/src/references/external_module.rs
sokra: Seems like there are 3 possible cases here:
- `ExternalType::CommonJs`
- `ExternalType::EcmaScriptModule` with import_externals
- `ExternalType::EcmaScriptModule` without import_externals

crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -0,0 +1,226 @@
...
+        }
+        .cell()
+    }
+}
+
+#[turbo_tasks::value_impl]
+impl Module for ExternalModuleAsset {
+    #[turbo_tasks::function]
+    fn ident(&self) -> Vc<AssetIdent> {
+        let fs = VirtualFileSystem::new_with_name("external-assets".to_string());
```
sokra, line 95:
```suggestion
        let fs = VirtualFileSystem::new_with_name("externals".to_string());
```

`assets` is internal and we shouldn't expose that use users. The filesystem name is visible in the dev tools

crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -0,0 +1,226 @@
...
+        let path = fs.root().join(self.request.clone());
```
sokra, line 96: request might start with `../` in which case that would crash...

sokra: Maybe don't use the require in the path name and use it as `AssetIdent::modifier` instead

crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -0,0 +1,226 @@
...
+    EcmascriptModuleContent,
+};
+
+#[turbo_tasks::function]
+fn layer() -> Vc<String> {
+    Vc::cell("external".to_string())
+}
+
+#[turbo_tasks::value]
+pub struct ExternalModuleAsset {
```
sokra, line 31:
```suggestion
pub struct ExternalModule {
```

crates/turbopack-ecmascript/src/resolve.rs
```diff
@@ -0,0 +1,121 @@
...
+
+/// Replaces the externals in the result with `ExternalModuleAsset` instances.
+pub async fn replace_externals(
+    result: Vc<ModuleResolveResult>,
+    import_externals: bool,
+) -> Result<Vc<ModuleResolveResult>> {
+    let mut result = result.await?.clone_value();
+
+    for item in result.primary.values_mut() {
+        if let ModuleResolveResultItem::External(request, ty) = item {
```
sokra, line 111: I think this should only handle `ExternalType::CommonJs`, `ExternalType::EcmascriptModule` and `ExternalType::OriginalReference`. It should map `ExternalType::OriginalReference` to cjs or esm here, since we no longer know that in `ExternalModuleAsset`

crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -0,0 +1,226 @@
...
+    pub request: String,
+    pub external_type: ExternalType,
+    pub import_externals: bool,
```
sokra, line 34: Maybe these two field give too much freedom and it should be a new enum for this Module:

```
enum CachedExternalModuleType {
  CommonJs,
  EcmaScriptModuleViaImport,
  EcmaScriptModuleViaRequire,
}
```

crates/turbopack-ecmascript/src/resolve.rs
```diff
@@ -0,0 +1,121 @@
...
+        reference_type,
+        origin.origin_path(),
+        request,
+        options,
+        issue_severity,
+        issue_source,
+    )
+    .await?;
+
+    replace_externals(result, import_externals).await
```
sokra, line 92: I feel like this behavior might be better placed either in the `ModuleAssetContext` (or in a `ProcessingPlugin`). This way resolving would not depend on the `turbopack-ecmascript` crate and could stay in `turbopack-resolve`, and we could switch it on and off.

crates/turborepo-auth/src/auth/mod.rs
ForsakenHarmony: good question, I think this was a clippy lint that kinda leaked over from another PR

crates/turbopack-ecmascript/src/references/external_module.rs
ForsakenHarmony: all the writes are infallible, but I guess there's no real difference

crates/turbopack-ecmascript/src/references/external_module.rs
ForsakenHarmony: I guess the Url type is just treated as commonjs?

crates/turbopack-ecmascript/src/references/external_module.rs
ForsakenHarmony: done

sokra (approved):

PR #7758, opened 2024-03-18
fix(ui): always start tasks

chris-olszewski:
### Description

Was hitting failures on cache hits where we would try to mark a task as finished, but we had never started it. This PR makes sure we start tasks in the UI even on cache hits.

Added some debug logging to help distinguish the "task not found in ui" errors. 

### Testing Instructions

Use the UI with a `turbo` run with a cache hit.

Closes TURBO-2653


crates/turborepo-ui/src/tui/table.rs
```diff
@@ -78,7 +79,10 @@ impl TaskTable {
         let planned_idx = self
             .planned
             .binary_search_by(|planned_task| planned_task.name().cmp(task))
-            .map_err(|_| Error::TaskNotFound { name: task.into() })?;
+            .map_err(|_| {
+                debug!("no task found on start");
```
arlyon, line 83: Minor change but having the name of the task in the log would be useful

arlyon (approved):
Small fixship

crates/turborepo-ui/src/tui/table.rs
```diff
@@ -107,7 +111,10 @@ impl TaskTable {
             .running
             .iter()
             .position(|running| running.name() == task)
-            .ok_or_else(|| Error::TaskNotFound { name: task.into() })?;
+            .ok_or_else(|| {
+                debug!("can't find task on finish");
```
arlyon, line 115: Same again

PR #7767, opened 2024-03-19
feat: add ability to configure tasks as interactive

chris-olszewski:
### Description

This adds a new `"interactive"` field to task definitions which allows users to interact with tasks when it is set to true.

A few notes on the feature:
 - All persistent tasks automatically are interactive
 - Interactive tasks cannot be cached since the interaction can (and probably will) alter the task's execution. This could be done automatically, but I chose to require that the task be explicitly marked as `"cache": false` to reduce perceived magic.
 - `turbo` will not attempt to run if any tasks in the graph are marked as interactive and the experimental UI isn't being used

### Testing Instructions

Snapshot test for various validation errors.

Manual testing: no longer able to select tasks and interact with them e.g. `turbo_dev build --filter=docs --force`

Closes TURBO-2656

### DIFF
--- a/crates/turborepo-lib/src/engine/mod.rs
+++ b/crates/turborepo-lib/src/engine/mod.rs
@@ -181,6 +181,7 @@ impl Engine<Built> {
         &self,
         package_graph: &PackageGraph,
         concurrency: u32,
+        experimental_ui: bool,
     ) -> Result<(), Vec<ValidateError>> {
         // TODO(olszewski) once this is hooked up to a real run, we should
         // see if using rayon to parallelize would provide a speedup
@@ -277,10 +278,33 @@ impl Engine<Built> {
             })
         }
 
+        validation_errors.extend(self.validate_interactive(experimental_ui));
+
         match validation_errors.is_empty() {
             true => Ok(()),
             false => Err(validation_errors),
         }
+    }
+
+    // Validates that UI is setup if any interactive tasks will be executed
+    fn validate_interactive(&self, experimental_ui: bool) -> Vec<ValidateError> {
+        // If experimental_ui is being used, then we don't need check for interactive
+        // tasks
+        if experimental_ui {
+            return Vec::new();
+        }
+        self.task_definitions
+            .iter()
+            .filter_map(|(task, definition)| {
+                if definition.interactive {
+                    Some(ValidateError::InteractiveNeedsUI {
+                        task: task.to_string(),
+                    })
+                } else {
+                    None
+                }
+            })
+            .collect()
     }
 }
 
@@ -310,6 +334,11 @@ pub enum ValidateError {
         persistent_count: u32,
         concurrency: u32,
     },
+    #[error(
+        "Cannot run interactive task \"{task}\" without experimental UI. Set `\"experimentalUI\": \
+         true` in `turbo.json` or `TURBO_EXPERIMENTAL_UI=true` as an environment variable"
+    )]
+    InteractiveNeedsUI { task: String },
 }
 
 impl fmt::Display for TaskNode {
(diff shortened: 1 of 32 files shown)


crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -277,11 +278,34 @@ impl Engine<Built> {
...
+        validation_errors.extend(self.validate_interactive(experimental_ui));
+
         match validation_errors.is_empty() {
             true => Ok(()),
             false => Err(validation_errors),
         }
     }
+
+    // Validates that UI is setup if any interactive tasks will be executed
+    fn validate_interactive(&self, experimental_ui: bool) -> Vec<ValidateError> {
```
chris-olszewski, line 290: This will be removed once the UI is the default. The check will come down to just "is `turbo` itself running in an interactive terminal"

weyert:
What happens when you want to run your persistent tasks in CI? E.g to run e2e tests against? Or trying to run UI within a CI runner

crates/turborepo-lib/src/engine/mod.rs
arlyon: That's a cool protection actually. Turbo stops you from accidentally running interactive processes in CI

crates/turborepo-lib/src/task_graph/mod.rs
```diff
@@ -73,6 +73,11 @@ pub struct TaskDefinition {
     // Persistent indicates whether the Task is expected to exit or not
     // Tasks marked Persistent do not exit (e.g. --watch mode or dev servers)
     pub persistent: bool,
+
+    // Interactive marks that a task can be interacted with
+    // Tasks that can be interacted with cannot be cached as their outputs may depend on the
+    // interaction.
```
mehulkar, line 79: might want another line here about what "interactive" means regarding stdin, since these are more engineering notes, rather than user-friendly notes

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -268,14 +268,14 @@ impl<'a> Visitor<'a> {
 
                     let workspace_directory = self.repo_root.resolve(workspace_info.package_path());
 
-                    let persistent = task_definition.persistent;
+                    let interactive = task_definition.interactive || task_definition.persistent;
```
mehulkar, line 271: this change looks weird. persistent tasks are also considered "interactive" now?

turborepo-tests/integration/fixtures/turbo-configs/interactive.json
```diff
@@ -0,0 +1,21 @@
...
+    "my-app#build": {
+      "outputs": ["banana.txt", "apple.json"],
+      "dotEnv": [".env.local"]
+    },
+
+    "something": {},
+    "//#something": {},
+
+    "maybefails": {}
+  }
```
mehulkar, line 20: we should make fixtures as minimal as possible!

```suggestion
  "pipeline": {
    "build": {
      "interactive": true
    }
  }
```

turborepo-tests/integration/tests/interactive.t
```diff
@@ -0,0 +1,15 @@
+Setup
+  $ . ${TESTDIR}/../../helpers/setup_integration_test.sh
+  $ . ${TESTDIR}/../../helpers/replace_turbo_json.sh $(pwd) "interactive.json"
+Verify we error on interactive task that hasn't been marked as cache: false
```
mehulkar, line 4: not needed for this PR, but are we still planning to:
- assume `cache:false` when it isn't specified manually?
- allow `cache:true` if it's set explicitly?

mehulkar (approved):

chris-olszewski:
> What happens when you want to run your persistent tasks in CI? E.g to run e2e tests against? Or trying to run UI within a CI runner

You can still do that, the error will only get thrown for tasks marked with `"interactive": true` as those require a TTY to be get user input.

We still allow for persistent tasks to run without a TTY, you just won't be able to interact with them.

PR #7771, opened 2024-03-20
feat(turbopack): Handle fragment in requests

kdy1:
### Description

Closes PACK-1006

I added `fragment` to variants of `Request`

### Testing Instructions

I'm not sure about the way to test


crates/turbopack-core/src/resolve/parse.rs
```diff
@@ -228,11 +264,17 @@ impl Request {
     }
 
     #[turbo_tasks::function]
-    pub fn raw(request: Value<Pattern>, query: Vc<String>, force_in_lookup_dir: bool) -> Vc<Self> {
+    pub fn raw(
+        request: Value<Pattern>,
+        query: Vc<String>,
+        force_in_lookup_dir: bool,
+        fragment: Vc<String>,
```
sokra, line 271:
```suggestion
        fragment: Vc<String>,
        force_in_lookup_dir: bool,
```

crates/turbopack-core/src/resolve/parse.rs
```diff
@@ -241,20 +283,28 @@ impl Request {
         request: Value<Pattern>,
         query: Vc<String>,
         force_in_lookup_dir: bool,
+        fragment: Vc<String>,
```
sokra, line 286:
```suggestion
        fragment: Vc<String>,
        force_in_lookup_dir: bool,
```

crates/turbopack-core/src/resolve/parse.rs
```diff
@@ -303,40 +354,56 @@ impl Request {
...
             Request::Module {
                 module,
                 path,
                 query: _,
+                fragment,
             } => Request::Module {
                 module: module.clone(),
                 path: path.clone(),
                 query,
+                fragment: fragment.clone(),
```
sokra, line 386: Vcs don't need clone. Clippy probably also complains about that

sokra (approved):

PR #7798, opened 2024-03-21
fix(examples): correct remix import

tknickman:
### Description

Fix type import for remix app in kitchen-sink example


anthonyshew:
Nabbed this one in #7624 !

tknickman:
Closing in favor of https://github.com/vercel/turbo/pull/7624/files

PR #7800, opened 2024-03-21
fix and improve webpack logger implementation

sokra:
### Description

* implement all logger methods
* pass logging to rust
* show error when using error logging

### DIFF
--- a/crates/turbopack-node/js/src/transforms/webpack-loaders.ts
+++ b/crates/turbopack-node/js/src/transforms/webpack-loaders.ts
@@ -16,6 +16,7 @@ import {
   parse as parseStackTrace,
 } from "../compiled/stacktrace-parser";
 import { type StructuredError } from "src/ipc";
+import { group, profile, time, timeLog, trace } from "console";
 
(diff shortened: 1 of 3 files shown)


crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -333,36 +340,111 @@ const transform = (
...
-            logger.debug = logger.bind(this, LogType.debug);
-
-            return logger;
+            return {
+              error: logFn.bind(this, LogType.error),
+              warn: logFn.bind(this, LogType.warn),
+              info: logFn.bind(this, LogType.info),
+              log: logFn.bind(this, LogType.log),
+              debug: logFn.bind(this, LogType.debug),
+              asset: (assertion: boolean, ...args: any[]) => {
```
wbinnssmith, line 377:
```suggestion
              assert: (assertion: boolean, ...args: any[]) => {
```

Should this throw? Otherwise, execution continues past the assertion...

wbinnssmith (approved):

PR #7810, opened 2024-03-21
fix(ui): avoid using prefix for replayed logs

chris-olszewski:
### Description

Currently on a cache hit the replayed logs in the UI contain the task prefix which is unnecessary in the UI. This brings the cache hit/miss output to be the same in the UI (modulo cache hit/miss lines).

### Testing Instructions

Before:
Screenshot 2024-03-21 at 1 43 03 PM

After:
Screenshot 2024-03-21 at 1 42 24 PM

(Also making sure non-UI behavior is preserved)
Screenshot 2024-03-21 at 1 43 54 PM

Closes TURBO-2684

### DIFF
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -184,6 +184,7 @@ impl TaskCache {
     pub async fn restore_outputs(
         &mut self,
         mut terminal_output: impl Write,
+        alternative_log_replay: Option<impl Write>,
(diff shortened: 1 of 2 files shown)


mehulkar (approved):
Nice! can the `cache hit` line also remove the prefix?

chris-olszewski:
> Nice! can the cache hit line also remove the prefix?

We can, but I do to some degree like keeping the prefix for lines that `turbo` is producing to differentiate between output from `turbo` and output from the actual task.

PR #7820, opened 2024-03-22
feat(telemetry): add JS package

tknickman:
### Description

Adds a `turbo-telemetry` library that enables tracking telemetry events from javascript packages. 

**Important**
This shares all telemetry state with turbo core's telemetry. This means users only need to disable once across all packages.

### DIFF
--- a/packages/turbo-telemetry/README.md
+++ b/packages/turbo-telemetry/README.md
@@ -11,4 +11,32 @@ This information is used to shape the Turborepo roadmap and prioritize features.
 
 ## Events
 
-All recorded events can be found by browsing the [event methods on the client class](./src/client.ts).
+Each event must have its own method on the client class. All recorded events can be found by browsing the [event methods on the client class](./src/client.ts).
+
+## Usage
+
+1. Init the client with your package name and version:
+
+```ts
+import { initTelemetry } from "@turbo/telemetry";
+import pkgJson from "../package.json";
+
+const { telemetry } = await initTelemetry({
+  name: pkgJson.name,
+  version: pkgJson.version,
+});
+```
+
+2. Send events
+
+```ts
+telemetry.myCustomEventName({
+  // event properties
+});
+```
+
+3. Close the client before exiting
+
+```ts
+await telemetry.close();
+```
--- a/packages/turbo-telemetry/README.md
+++ b/packages/turbo-telemetry/README.md
@@ -1,0 +1,14 @@
+# `@turbo/telemetry`
+
+**NOTE**:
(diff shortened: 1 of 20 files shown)


packages/turbo-telemetry/README.md
```diff
@@ -0,0 +1,14 @@
+# `@turbo/telemetry`
```
mehulkar, line 1: Can you include some usage instructions?

packages/turbo-telemetry/src/cli.ts
mehulkar, line 1: Is this meant to be used as a CLI or internally in other JS things (like turbo-ignore, create-turbo)? Why does it have its own CLI?

tknickman: This isn't a standalone CLI, it's just a helper so other CLIs can add the command without duplicating code. For example `create-turbo` (or any of our other CLIs) can just do:

```
import { Command, Option } from "commander";
import {  withTelemetryCommand } from "@turbo/telemetry";
...

const createTurboCli = new Command();

// create-turbo specific commands...

// add telemetry command
withTelemetryCommand(createTurboCli);
```

packages/turbo-telemetry/README.md
tknickman: Yes, will do!

packages/turbo-telemetry/src/cli.ts
tknickman: Add the command to other CLIs that use this allows telemetry to be disabled without needing `turbo` core. Just incase we have clients that consume a package, without `turbo` itself.

chris-olszewski (approved):
LGTM. I do have some slight concern about keeping the event keys in sync. There's probably something we could do with `napi`,  but no need to build infra before we need it.

packages/turbo-telemetry/src/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+export function buildUserAgent(packageInfo: PackageInfo): string {
+  const nodeVersion = process.version;
+  const operatingSystem = os.type();
+  const architecture = os.arch();
+
+  return `${packageInfo.name} ${packageInfo.version} ${nodeVersion} ${operatingSystem} ${architecture}`;
+}
+
+export async function defaultConfigPath() {
+  const dir = await configDir();
```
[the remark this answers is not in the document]
tknickman: We shouldn't _need_ that here, but I'll add it just to match

tknickman:
@chris-olszewski the intention isn't to mirror events here one to one, but for the events tracked in the JS packages to be unique and specific to the JS packages. These also use a different event type on the backend that always tracks the package name and version as context. 


I did think about napi for some of the shared code (like config) but the lift was low and didn't want to deal with the multi os / arch support here.

PR #7830, opened 2024-03-24
Avoid showing import map description in resolving issue when there is no import map mapping

sokra:
### Description

Avoid showing import map description in resolving issue when there is no import map mapping

### Testing Instructions

Closes PACK-2828


kwonoj (approved):

timneutkens (approved):

kdy1 (approved):
Note: I updated jest snapshots while merging https://github.com/vercel/next.js/pull/63541

The relevant snapshot changes will need some modfications

PR #7832, opened 2024-03-24
docs: added new example

Parthvsquare:
with nextjs 14 (app router) vite (tanstack router) shadcn/ui

### Apps and Packages

- `admin`: a [Next.js](https://nextjs.org/) app
- `super-admin`: [Vite + React](https://vitejs.dev/) and [Ant design](https://ant.design/) app and using [tanstack router](https://tanstack.com/router/latest) and [jotai](https://jotai.org/) for state management
- `web`: another [Next.js](https://nextjs.org/) app
- `ui`: a stub React component library shared by both `web` and `docs` applications (🚀 powered by **shadcn/ui**)
- `eslint-config-custom`: `eslint` configurations (includes `eslint-config-next` and `eslint-config-prettier`)
- `tsconfig`: `tsconfig.json`s used throughout the monorepo


anthonyshew:
@Parthvsquare 👋

Looking at our other examples, this proposal doesn't quite fit the mold.
- We already have examples for Next.js and Vite, so we don't want to duplicate.
- We keep the examples extremely light and simple, so the application with Ant Design, TanStack Router, and Jotai doesn't really fit what we're looking for.

That said, we do want a shadcn/ui example! Do you think you could make one, starting with the `basic` example and replacing the `ui` package with `shadcn/ui`?

Parthvsquare:
@anthonyshew 
should i close this and make a new PR? or edit the current one? 
>  We already have examples for Next.js and Vite, so we don't want to duplicate

Yes, but we don't have for both in one app folder

anthonyshew:
> should i close this and make a new PR? or edit the current one?

Edit this one works. 👍 

> Yes, but we don't have for both in one app folder

The `kitchen-sink` example has multiple frameworks in one example. It doesn't specifically have Next.js + Vite out of the box  - but the Remix app could quickly be adapted to a Vite application.

PR #7837, opened 2024-03-25
fix(ui): limit viewport size based on term size

chris-olszewski:
### Description

With #7822 (and #7805 ) we're using `insert_before` to persist logs. This can panic (https://github.com/ratatui-org/ratatui/issues/999) if the viewport fills the entire terminal.

This PR changes our viewport construction so it now takes terminal size into account. (Previously we always would use 60, even if the terminal didn't have 60 rows)

### Testing Instructions

Use UI with a terminal that has a height of 60 or less.

Closes TURBO-2695


tknickman (approved):

weyert:
Thanks, I was experiencing this issue earlier today :)

PR #7840, opened 2024-03-26
add concurrency corrected duration

sokra:
### Description

* add concurrency corrected duration
* add CPU time (renamed from duration)
* add secondary metric to data
* remove left-over println

### Testing Instructions

Closes PACK-2844

### DIFF
--- a/crates/turbopack-trace-server/src/self_time_tree.rs
+++ b/crates/turbopack-trace-server/src/self_time_tree.rs
@@ -1,0 +1,248 @@
+use std::{
+    cmp::{max, min},
+    mem::{replace, take},
+};
+
+const SPLIT_COUNT: usize = 128;
+
+pub struct SelfTimeTree<T> {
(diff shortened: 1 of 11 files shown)


crates/turbopack-trace-server/src/self_time_tree.rs
```diff
@@ -0,0 +1,248 @@
...
+            }
+        }
+        total_count
+    }
+
+    pub fn for_each_in_range(&self, start: u64, end: u64, mut f: impl FnMut(u64, u64, &T)) {
+        self.for_each_in_range_ref(start, end, &mut f);
+    }
+
+    fn for_each_in_range_ref(&self, start: u64, end: u64, f: &mut impl FnMut(u64, u64, &T)) {
```
kdy1, line 177: I think `&mut dyn` would be better as this is a sort of `#[cold]` path for normal runs, but well it's a super minor issue

kdy1 (approved):

PR #7850, opened 2024-03-27
🌿 Minor: Bun Example

nermalcat69:
implemented a bun example


nermalcat69:
work in progress, i gotta setup some stuff

tknickman (commented):
Hey @nermalcat69 thanks for the PR. I'm not opposed to this - as specific package manager examples can be useful to point to for reference or to use in our CI. However, we should already support `bun` for _most_ of our examples via `create-turbo`, which will handle converting the example to the requested package manager (including bun!).

What about this example is bun specific?

nermalcat69:
@tknickman Well i tried creating this, i think we should implement bun for every example
image

tknickman:
@nermalcat69 you should use `bunx` here

nermalcat69:
yeah now i know what you meant by having bun supported, i tried it locally

Sorry for this pr, have a nice day ahead

PR #7852, opened 2024-03-27
feat(create-turbo): usage metrics

tknickman:
### Description

Add anonymous usage telemetry to create-turbo. If you have already opted out of turbo telemetry, you have opted out of this telemetry as well.

### DIFF
--- a/packages/turbo-telemetry/src/client.ts
+++ b/packages/turbo-telemetry/src/client.ts
@@ -147,17 +147,135 @@ export class TelemetryClient {
     }
   }
 
+  private trackCliOption({
+    option,
+    value,
+  }: {
+    option: string;
+    value: string;
+  }): Event {
+    return this.track({
+      key: `option:${option}`,
+      value,
+    });
+  }
+
+  private trackCliArgument({
+    argument,
+    value,
+  }: {
+    argument: string;
+    value: string;
+  }): Event {
+    return this.track({
+      key: `argument:${argument}`,
+      value,
+    });
+  }
+
+  private trackCliCommand({
+    command,
+    value,
+  }: {
+    command: string;
+    value: string;
+  }): Event {
+    return this.track({
+      key: `command:${command}`,
+      value,
+    });
+  }
+
   ////////////
   // EVENTS //
   ////////////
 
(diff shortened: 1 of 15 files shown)


packages/turbo-telemetry/src/client.ts
```diff
@@ -147,17 +147,135 @@ export class TelemetryClient {
...
+  trackOptionPackageManager(value: string | undefined): Event | undefined {
+    if (value) {
+      return this.trackCliOption({
+        option: "package_manager",
+        value,
+      });
+    }
+  }
+
+  trackOptionSkipInstall(value: boolean | undefined): Event | undefined {
```
mehulkar, line 226: all these new methods seem really specific to create-turbo, not sure they should be in the telemtry package. can it be more generic?

tknickman: They're intentionally specific to allow auditing of events. It's the same approach we took for the core telemetry, but yea could subclass this per package to organize them a bit more.

tknickman: @mehulkar updated to have package specific event classes, this keeps things a bit more organized.

 Ideally we could just have super generic events that a re-used everywhere. But we need to have the events defined:

1. In a single location (not spread out across the code base)
2. As static as possible (no generic key=value)

This ensures that they are easily publicly auditable. 

Let me know what you think!

tknickman: Relevant commit: https://github.com/vercel/turbo/pull/7852/commits/a4951541922e74cad09d2c45fc708e445a829690

packages/turbo-telemetry/src/events/turbo-ignore.ts
```diff
@@ -0,0 +1,11 @@
+import { TelemetryClient } from "../client";
+import type { Event } from "./types";
+
+export class TurboIgnoreTelemetry extends TelemetryClient {
```
tknickman, line 4: Will extend this in a future PR, but added the stub in this PR to ensure the init generics work as expected.

mehulkar (approved):

PR #7865, opened 2024-03-29
fix(pnpm): support new pnpm9 default link-workspace-packages

chris-olszewski:
### Description

With [PNPM 9](https://github.com/pnpm/pnpm/releases/tag/v9.0.0-alpha.0), [link-workspace-packages]() is now defaulting to `false` meaning that unless the `workspace` protocol is explicitly used, packages from the registry will be preferred over workspace packages.

It's odd to have 3 variants of the PNPM package manager, but that is the current state of the world. Unsure of why we changed from a generic interface to a enum in the Go->Rust port, but we'll probably want to go back to that world to reduce match statement complexity.

### Testing Instructions

Unit tests for package manager detection. Quick spot check in a `create-turbo` repo with workspace deps updates so they don't use `workspace` protocol.

### DIFF
--- a/crates/turborepo-repository/src/package_manager/pnpm.rs
+++ b/crates/turborepo-repository/src/package_manager/pnpm.rs
@@ -25,8 +25,12 @@ impl<'a> PnpmDetector<'a> {
 
     pub fn detect_pnpm6_or_pnpm(version: &Version) -> Result<PackageManager, Error> {
         let pnpm6_constraint: Range = "<7.0.0".parse()?;
+        // TODO actually do this
+        let pnpm9_constraint: Range = ">=9.0.0-alpha.0".parse()?;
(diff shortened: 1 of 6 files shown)


crates/turborepo-repository/src/package_manager/pnpm.rs
```diff
@@ -25,8 +25,12 @@ impl<'a> PnpmDetector<'a> {
 
     pub fn detect_pnpm6_or_pnpm(version: &Version) -> Result<PackageManager, Error> {
         let pnpm6_constraint: Range = "<7.0.0".parse()?;
+        // TODO actually do this
```
tknickman, line 28: Looks like we are doing this?

tknickman (approved):

PR #7866, opened 2024-03-29
chore(ui): upgrade tui-term

chris-olszewski:
### Description
`tui-term` has released `0.1.9` which has the functionality we require to use it with our `vt100` fork

### Testing Instructions

Verify build and that terminal renders as expected:
Screenshot 2024-03-29 at 10 24 30 AM

Closes TURBO-2734


arlyon (approved):
Approved. I have re-ran the failing macOS test. Seems unrelated to this one... so I am hoping it is just a transient issue.

PR #7855, opened 2024-03-27
feat(Turborepo): file hash watching

gsoltis:

PR #7868, opened 2024-03-29
feat(Turborepo): wire file hashing to grpc server

gsoltis:
### Description

 - Wire up daemon-backed file hashing into the client

### Testing Instructions

Existing integration test suite

Closes TURBO-2736

### DIFF
--- a/crates/turborepo-lib/src/task_hash.rs
+++ b/crates/turborepo-lib/src/task_hash.rs
@@ -1,13 +1,16 @@
 use std::{
     collections::{HashMap, HashSet},
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
 use rayon::prelude::*;
 use serde::Serialize;
 use thiserror::Error;
 use tracing::{debug, Span};
-use turbopath::{AbsoluteSystemPath, AnchoredSystemPath, AnchoredSystemPathBuf};
+use turbopath::{
+    AbsoluteSystemPath, AnchoredSystemPath, AnchoredSystemPathBuf, RelativeUnixPathBuf,
+};
 use turborepo_cache::CacheHitMetadata;
 use turborepo_env::{BySource, DetailedMap, EnvironmentVariableMap, ResolvedEnvMode};
 use turborepo_repository::package_graph::{PackageInfo, PackageName};
@@ -23,6 +26,7 @@ use crate::{
     opts::RunOpts,
     run::task_id::TaskId,
     task_graph::TaskDefinition,
+    DaemonClient, DaemonConnector,
 };
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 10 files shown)


gsoltis:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/vercel/turbo/7868?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7868** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7868?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7855** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7855?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @gsoltis and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

crates/turborepo-lib/src/task_hash.rs
```diff
@@ -115,14 +124,74 @@ impl PackageInputsHashes {
...
+                                tracing::debug!(
+                                    "daemon file hashing failed for {}: {}",
+                                    package_path,
+                                    e
+                                );
+                                None
+                            }
+                        }
+                    });
+                let mut hash_object = match hash_object {
```
NicholasLYang, line 181: not very important but you could probably use `or_else` or `unwrap_or_else` here.

crates/turborepo-lib/src/task_hash.rs
```diff
@@ -115,14 +124,74 @@ impl PackageInputsHashes {
...
+                // timeout or get an error, fallback to local hashing
+                let hash_object = daemon
+                    .as_mut()
+                    .and_then(|daemon| {
+                        let handle = handle.clone();
+                        // We need an async block here because the timeout must be created with an
+                        // active tokio context. Constructing it directly in
+                        // the rayon thread doesn't provide one and will crash at runtime.
+                        handle
+                            .block_on(async {
```
NicholasLYang, line 137: I'm a little confused by [the docs](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#panics-2) for this method. It says the method panics "if called within an asynchronous execution context". Does that mean if it's called in an async function?

NicholasLYang (approved):
Just one clarification but otherwise looks good

crates/turborepo-lib/src/task_hash.rs
gsoltis: Essentially, yes, if it's running from a task managed by an async runtime, it will panic. In this case, we're running from the rayon threadpool.

crates/turborepo-lib/src/task_hash.rs
gsoltis: I looked at doing that, but it gets awkward since we're starting with an `Option`, but the fallback calculation produces a `Result` where we want to short-circuit, but the overall closure returns an `Option<Result<_, _>>`. `?` doesn't really work in that context, so you end up with the `match` either way.

PR #7876, opened 2024-04-01
chore(telemetry): track dot env usage

chris-olszewski:
### Description

Adds tracking for usage of `globalDotEnv` and `dotEnv` in `turbo.json`

### Testing Instructions

👀 

Closes TURBO-2741

### DIFF
--- a/crates/turborepo-telemetry/src/events/generic.rs
+++ b/crates/turborepo-telemetry/src/events/generic.rs
@@ -1,6 +1,7 @@
 use std::fmt::Display;
 
 use serde::{Deserialize, Serialize};
+use turbopath::RelativeUnixPathBuf;
 use turborepo_vercel_api::telemetry::{TelemetryEvent, TelemetryGenericEvent};
 use uuid::Uuid;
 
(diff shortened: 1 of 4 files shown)


crates/turborepo-telemetry/src/events/generic.rs
```diff
@@ -212,6 +213,30 @@ impl GenericEventBuilder {
...
+        });
+        self
+    }
+
+    pub fn track_dot_env(&self, dot_env: Option<&[RelativeUnixPathBuf]>) -> &Self {
+        self.track(Event {
+            key: "dot_env".into(),
+            value: match dot_env {
+                Some(dot_env) => dot_env.len().to_string(),
+                None => "null".into(),
```
chris-olszewski, line 233: Not sure if this would be better as "0" instead of "null".

crates/turborepo-telemetry/src/events/generic.rs
```diff
@@ -212,6 +213,30 @@ impl GenericEventBuilder {
...
+            },
+            is_sensitive: EventType::NonSensitive,
+        });
+        self
+    }
+
+    pub fn track_dot_env(&self, dot_env: Option<&[RelativeUnixPathBuf]>) -> &Self {
+        self.track(Event {
+            key: "dot_env".into(),
```
chris-olszewski, line 230: Currently we're producing one of these events per task. Not sure if we should roll these up to a single event or not.

PR #7897, opened 2024-04-03
feat(turbo-ignore): usage metrics

tknickman:
### Description

Anonymous usage metrics for turbo-ignore. Will help inform future roadmap for the tool.

### DIFF
--- a/packages/create-turbo/package.json
+++ b/packages/create-turbo/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "chalk": "4.1.2",
-    "commander": "^10.0.0",
+    "commander": "^11.0.0",
     "fs-extra": "^11.1.1",
     "inquirer": "^8.0.0",
     "proxy-agent": "^6.2.2",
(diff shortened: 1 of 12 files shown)


packages/create-turbo/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "chalk": "4.1.2",
-    "commander": "^10.0.0",
+    "commander": "^11.0.0",
```
tknickman, line 27: Updating to use same version across all our CLIs. Nothing breaking is relevant for us:

https://github.com/tj/commander.js/releases/tag/v11.1.0

arlyon: Wish pnpm had workspace packages like in cargo...

arlyon (approved):
Since we are adding to a new product here is proof it respects tracking env vars for posterity https://github.com/vercel/turbo/blob/583bb0934a2521b7549380f9c49b34565804b5b1/packages/turbo-telemetry/src/config.ts#L54

PR #7613, opened 2024-03-04
feat(turborepo): Watch mode

NicholasLYang:
### Description

Implements an MVP version of watch mode that takes the package changes stream and re-runs package tasks as needed.

### Testing Instructions

Closes TURBO-2519

PR #7898, opened 2024-04-03
feat: Re-run dependent tasks in watch mode

NicholasLYang:
### Description

Instead of using `--filter` in our re-running of tasks for watch mode, we instead prune the task graph to run the changed task and its dependent tasks. This has the benefit of following the task graph and not the package graph.

### Testing Instructions

You can verify this works by creating a task dependency relation that is not in the package graph. So, make a `ui#test` task that depends on a `docs#test` task, but don't have `ui` depend on `docs`. Then verify that when the `docs` are changed, we re-run the `ui#test` task too.

Closes TURBO-2752

### DIFF
--- a/crates/turborepo-lib/src/engine/mod.rs
+++ b/crates/turborepo-lib/src/engine/mod.rs
@@ -45,6 +45,7 @@ pub struct Engine<S = Built> {
     task_lookup: HashMap<TaskId<'static>, petgraph::graph::NodeIndex>,
     task_definitions: HashMap<TaskId<'static>, TaskDefinition>,
     task_locations: HashMap<TaskId<'static>, Spanned<()>>,
+    package_tasks: HashMap<PackageName, Vec<petgraph::graph::NodeIndex>>,
 }
 
 impl Engine<Building> {
@@ -58,6 +59,7 @@ impl Engine<Building> {
             task_lookup: HashMap::default(),
             task_definitions: HashMap::default(),
             task_locations: HashMap::default(),
+            package_tasks: HashMap::default(),
         }
     }
 
@@ -65,6 +67,11 @@ impl Engine<Building> {
         self.task_lookup.get(task_id).copied().unwrap_or_else(|| {
             let index = self.task_graph.add_node(TaskNode::Task(task_id.clone()));
             self.task_lookup.insert(task_id.clone(), index);
+            self.package_tasks
+                .entry(PackageName::from(task_id.package()))
+                .or_default()
+                .push(index);
+
             index
         })
     }
@@ -103,6 +110,7 @@ impl Engine<Building> {
             root_index,
             task_definitions,
             task_locations,
+            package_tasks,
             ..
         } = self;
         Engine {
@@ -112,6 +120,7 @@ impl Engine<Building> {
             root_index,
             task_definitions,
             task_locations,
+            package_tasks,
         }
     }
 }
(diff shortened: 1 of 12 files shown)


NicholasLYang:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/vercel/turbo/7898?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7898** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7898?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7613** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7613?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7570** <a href="https://app.graphite.dev/github/pr/vercel/turbo/7570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/turborepo-lib/src/engine/builder.rs
```diff
@@ -462,6 +462,7 @@ impl<'a> EngineBuilder<'a> {
                 .ok_or_else(|| Error::MissingPackageJson {
                     workspace: workspace.clone(),
                 })?;
+
```
NicholasLYang, line 465:
```suggestion
```

crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -64,6 +64,7 @@ impl Engine<Building> {
     pub fn get_index(&mut self, task_id: &TaskId<'static>) -> petgraph::graph::NodeIndex {
         self.task_lookup.get(task_id).copied().unwrap_or_else(|| {
             let index = self.task_graph.add_node(TaskNode::Task(task_id.clone()));
+
```
NicholasLYang, line 67:
```suggestion
```

crates/turborepo-lib/src/run/builder.rs
```diff
@@ -433,6 +460,12 @@ impl RunBuilder {
         }))
         .build()?;
 
+        // If we have an initial task, we prune out the engine to only
+        // tasks that are reachable from that initial task.
+        if let Some(entrypoint_tasks) = &self.entrypoint_tasks {
```
NicholasLYang, line 465: This is a little hacky, but I couldn't figure out how to do this otherwise without cloning `Run` and we can't implement `Clone` on `Run` since we have the `AnalyticsHandle`.

crates/turborepo-lib/src/run/builder.rs
```diff
@@ -108,9 +118,21 @@ impl RunBuilder {
...
+            should_print_prelude_override: None,
         })
     }
 
+    pub fn with_entrypoint_tasks(mut self, entrypoint_tasks: Vec<TaskId<'static>>) -> Self {
+        self.entrypoint_tasks = Some(entrypoint_tasks);
+        self
+    }
+
+    pub fn hide_prelude(mut self) -> Self {
```
NicholasLYang, line 131: I added this because the prelude makes less sense with re-run tasks. It prints the packages in scope, but the tasks in those packages are not always run, since the task graph is pruned.

crates/turborepo-lib/src/engine/mod.rs
```diff
@@ -123,6 +123,52 @@ impl Default for Engine<Building> {
...
+                    .iter()
+                    .any(|idx| {
+                        node_distances
+                            .get(&(**idx, node_idx))
+                            .map_or(false, |dist| *dist != i32::MAX)
+                    })
+                    .then_some(node.clone())
+            },
+            |_, e| Some(*e),
+        );
```
gsoltis, line 160: Fine for now, but we probably ought to mark this for revisiting. I suspect there are more efficient ways we can do this.

crates/turborepo-lib/src/run/watch.rs
```diff
@@ -166,14 +169,23 @@ impl WatchClient {
                     run.abort();
                 }
 
+                let entrypoint_tasks = tasks
```
gsoltis, line 172: I don't think this is right. IIUC, `tasks` is the set of tasks the user request via `turbo watch <task1> <task2>`. However, if there's a filter present, like `turbo watch test -F my-lib`, and we get a change for a package `my-dep`, we don't necessarily want to look for `my-dep#test`. We want any task in the task graph that is run in `my-dep`. In this case, it's reasonable that `my-dep#build` might be present, but `my-dep#test` might not be.

crates/turborepo-lib/src/run/builder.rs
gsoltis: Are we rebuilding the original task graph each time?

gsoltis: We might want to consider separating the `Engine` from an instance of `Run`. `Run` isn't really meant to be reusable, but it might take a little work to make it possible to use multiple of them.

gsoltis (changes requested):
I think there's some algorithm and organizational changes we can make, but we don't need to block on those.

I do think we need ensure we are doing the correct graph pruning when we get a package change.

crates/turborepo-lib/src/run/builder.rs
NicholasLYang: Yeah I think we need to work out what exactly is reusable and what is persistent across runs. Like we can reuse the same analytics handle and cache handle, but maybe not the same task graph

crates/turborepo-lib/src/run/builder.rs
```diff
@@ -425,6 +448,12 @@ impl RunBuilder {
         }))
         .build()?;
 
+        // If we have an initial task, we prune out the engine to only
+        // tasks that are reachable from that initial task.
+        if let Some(entrypoint_package) = &self.entrypoint_package {
```
gsoltis, line 453: I think this will eventually need to be a set of packages, especially with debouncing and waiting for current runs to finish being cancelled or completed

gsoltis (approved):

PR #7900, opened 2024-04-03
fix: respect output mode 'none' even when caching is disabled

aholland:
### Description
Suppose I have some pipline like this:
```
//turbo.json
{
  "pipeline": {
    "clean": {
      "outputMode": "none",
      "cache": false
    },
    "build": {
      "dependsOn": ["clean","^build"],
```
The `outputMode` is ignored because of `cache` being false. That's annoying. I don't want to see my `rm -rf` running every time I do I build, including builds where all the rest of the tasks are cached (I tend to use `new-only` for most other tasks).

### Testing Instructions
I am not a Rust programmer but I think I have found the place that would need to be changed. To test the change, just create any simple package.son task that generates output (`    "echo": "echo \"hello world\"",`) and run it from a TurboRepo task with caching false and outMode none, as above.

### PS `errors-only`
PS having absolutely no output would be okay. Even better would be an implementation of errors-only, outputting one line saying that the task has been executed because it is set not to cache, and that it executed successfully. I think that would involve something like the below code (copied from about line 195 of the same file I've modified in this PR), but again, I don't know Rust or your codebase:
```
                fallible_write(
                    &mut terminal_output,
                    &format!(
                        "cache bypass, force executing {}\n",
                        color!(self.ui, GREY, "{}", self.hash)
                    ),
                );
```

### DIFF
--- a/packages/create-turbo/package.json
+++ b/packages/create-turbo/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-turbo",
-  "version": "2.3.4-canary.5",
+  "version": "2.3.4-canary.6",
   "description": "Create a new Turborepo",
   "homepage": "https://turbo.build/repo",
   "license": "MIT",
--- a/packages/eslint-config-turbo/package.json
+++ b/packages/eslint-config-turbo/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eslint-config-turbo",
-  "version": "2.3.4-canary.5",
+  "version": "2.3.4-canary.6",
   "description": "ESLint config for Turborepo",
   "repository": {
     "type": "git",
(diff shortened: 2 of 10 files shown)


arlyon (approved):

tknickman (approved):

anthonyshew:
Hey, sorry, for not getting this in before! CI hung on some stale check so the auto-merge didn't go through.

Now with the 2.0 release, [this changed names](https://github.com/vercel/turbo/actions/runs/9531307837/job/26272118347?pr=7900#step:5:481). @aholland, would you be able to update it?

anthonyshew:
I messed up and closed this PR with a force push. I've opened #9670 as a replacement.

PR #7903, opened 2024-04-04
docs(Turborepo): package entry points customization

linkb15:
### Description

When we have so many functions inside the packages. Importing files from a package become too long. 

i.e. we have a UI package where we import so many components.
it can be more enhanced if it can be more customized.

Customizing package entry points information was really hard for me to find before - spent more than few days to get it around - I wish it can be shown as example somewhere - and TurboRepo documentation is teaching me how to use this internal package - hence it would be nice to have this additional reading as well.


anthonyshew:
Can you explain how this is different from the section above on the same page? https://turbo.build/repo/docs/handbook/sharing-code/internal-packages#4-fix-exports

linkb15:
@anthonyshew this information is to customize the subpath of the internal package.

```js
// it will make the import in the different subpath
import { Dialog, DialogContent, DialogTrigger } from "@repo/ui/dialog";
import { Button } from "@repo/ui/button";

// instead of this
import { Button, Dialog, DialogContent, DialogTrigger } from "@repo/ui";
```

This is to customize how we can export it differently. 

I believe this will help some clarification on how we can do it, and customize it further.

anthonyshew:
We've done a major docs update since this PR! Check them out!

PR #7906, opened 2024-04-05
Fix Git filter comparison direction.

anthonyshew:
### Description

The direction for the Git comparisons we were showing in the documentation was reversed.

### Testing Instructions

👀

Closes TURBO-2760


tknickman (approved):

anthonyshew:
Nevermind, what's currently there is correct. I misunderstood.

PR #7935, opened 2024-04-11
feat(turbopack): Handle fragment in requests

kdy1:
### Description

This reverts commit 1bfe729336b31bbaa74a3e38efd60c4faabaddd2.

### Testing Instructions

https://github.com/vercel/next.js/pull/64232


kdy1:
I verified that the problematic test case, `test/e2e/app-dir/resource-url-encoding/resource-url-encoding.test.ts` , passes locally.

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1840,6 +1858,15 @@ async fn resolve_relative_request(
 
     let mut new_path = path_pattern.clone();
     if !options_value.fully_specified {
+        let fragment = fragment.await?;
+        if !fragment.is_empty() {
+            new_path.push(Pattern::Alternatives(
+                once(Pattern::Constant("".to_string()))
+                    .chain(once(Pattern::Constant(format!("#{fragment}"))))
```
sokra, line 1865: - this should also applied for fully specified requests
- in the case when we add the fragment to the path, we must no longer add it as fragment to the result

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1882,14 +1929,38 @@ async fn resolve_relative_request(
                                 options_value,
```
sokra, line 1929: You might need to handle fragment here too

sokra (commented):
Looks good. Let's add a test case into `turbopack-tests/execution` too to verify its working

crates/turbopack-tests/tests/execution/turbopack/resolving/fragment/input/index.js
```diff
@@ -0,0 +1,10 @@
+import client from "./client#component";
+import nofrag from "./nofrag#frag";
```
sokra, line 2: Maybe complete it with these two cases too

```
import client2 from "./client#component.js";
import nofrag2 from "./nofrag.js#frag";
```

sokra (approved):

PR #7947, opened 2024-04-11
fix HMR for cases where chunking changes

sokra:
### Description

add availability info to chunk list hash instead of chunk content avoid chunk lists when HMR isn't enabled

### Testing Instructions

Closes PACK-2943


timneutkens (approved):
💯

PR #7961, opened 2024-04-15
fix external module

sokra:
### Description

With `is_esm: true` `module` is not available.

### Testing Instructions

Closes PACK-2961

### DIFF
--- a/crates/turbopack-ecmascript/src/references/external_module.rs
+++ b/crates/turbopack-ecmascript/src/references/external_module.rs
@@ -83,7 +83,7 @@ impl CachedExternalModule {
         if self.external_type == CachedExternalType::CommonJs {
             writeln!(code, "module.exports = mod;")?;
         } else {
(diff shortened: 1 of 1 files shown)


crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -83,13 +83,13 @@ impl CachedExternalModule {
         if self.external_type == CachedExternalType::CommonJs {
             writeln!(code, "module.exports = mod;")?;
         } else {
-            writeln!(code, "__turbopack_dynamic__(mod);")?;
+            writeln!(code, "__turbopack_export_namespace__(mod);")?;
```
ForsakenHarmony, line 86: This breaks with async modules

sokra, line 86:
```suggestion
            writeln!(code, "__turbopack_dynamic__(mod);")?;
```

ForsakenHarmony (approved):

PR #7965, opened 2024-04-15
chore(@turbo/telemetry): replace jest with node:test runner

mehulkar:
These tests take ~2-3s to run with Jest. With the node:test runner,
they take ~500ms, which is about 5x faster. On a scale of seconds,
this doesn't matter much, but this is a proof of concept for a number
of reasons:

- Jest is large API and we don't need it now that test running and
  assertions are part of node core,
- Using pure Node makes it easier to update modules that are pure ESM,
  which open up other parts of the ecosystem.

Closes TURBO-2803


packages/turbo-telemetry/src/client.test.ts
```diff
@@ -159,33 +211,58 @@ describe("TelemetryClient", () => {
...
+    assert.equal(
+      "x-turbo-telemetry-id" in mockPost.mock.calls[0].arguments[1].headers,
+      true
+    );
+    assert.equal(
+      /create-turbo 1\.0\.0/.test(
+        mockPost.mock.calls[0].arguments[1].headers["User-Agent"]
+      ),
+      true
+    );
```
chris-olszewski, line 264: I think [`assert.matches`](https://nodejs.org/api/assert.html#assertmatchstring-regexp-message) is more fitting
```suggestion
    assert.matches(
      mockPost.mock.calls[0].arguments[1].headers["User-Agent"],
      /create-turbo 1\.0\.0/
    );
```

packages/turbo-telemetry/src/utils.test.ts
```diff
@@ -1,33 +1,42 @@
...
       );
     });
 
     test("should return consistent length", () => {
       const input = "a-sensitive-value";
       const salt = "private-salt";
 
-      const result1 = oneWayHashWithSalt({ input, salt });
-      const result2 = oneWayHashWithSalt({ input: `${input}-${input}`, salt });
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call -- reasons
```
chris-olszewski, line 22: I don't think this is necessary.

packages/turbo-telemetry/src/utils.test.ts
```diff
@@ -1,33 +1,42 @@
...
+      const result1 = utils.oneWayHashWithSalt({ input, salt });
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call -- reasons
```
chris-olszewski, line 24: I don't think this is necessary.

packages/turbo-telemetry/src/utils.test.ts
```diff
@@ -1,33 +1,42 @@
...
+      assert.equal(result1.length, result2.length);
     });
   });
 
   describe("defaultConfigPath", () => {
     test("supports overriding by env var", async () => {
       process.env.TURBO_CONFIG_DIR_PATH = "/tmp";
-      const result = await defaultConfigPath();
-      expect(result).toEqual("/tmp/turborepo/telemetry.json");
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call -- reasons
```
chris-olszewski, line 37: I don't think this is necessary.

chris-olszewski (approved):
Minor suggestions, but overall LFG

PR #7954, opened 2024-04-14
Update storybook docs

zsh77:
### Description

- fixed the init script of storybook on the `npm` tab, (the exact changes we had in the PR: https://github.com/vercel/turbo/pull/7695 using npm instead of pnpm)
- replaced `ui` with `@repo/ui`
- bug fix of importing a button from an undefined export specifier

### Testing Instructions

Going through with the process of adding storybook to turborepo should work fine which is:

- cd apps
- npm create vite
- cd workshop
- npx sb init

PR #7973, opened 2024-04-15
Adding correct reference to the @repo/ui package in the storybook doc

mathcrln:
### Description

Very simple change to fix the  ``ERR_PNPM_WORKSPACE_PKG_NOT_FOUND  In apps/workshop: "ui@workspace:*" is in the dependencies but no package named "ui" is present in the workspace`` error popping when a  user attempts to follow the instructions found in [Using Storybook with Turborepo](https://turbo.build/repo/docs/handbook/tools/storybook) doc. 
CleanShot 2024-04-15 at 16 46 28

Note: While it was my fault for not noticing the commands were so drastrically different between package managers, we might want to highlight that pnpm requires specific steps, even outside of the tabs as a way to prevent the other users from making the same mistake I did and not understanding how to mitigate this error:
```
sh: storybook: command not found
 ELIFECYCLE  Command failed.
```
CleanShot 2024-04-15 at 16 48 08

### Testing Instructions

N/A


anthonyshew:
It looks like this already got handled in #7954. Sorry about that!

PR #7978, opened 2024-04-16
chore(ci): remove invalid action argument

mehulkar:
Closes TURBO-2814

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -310,7 +310,6 @@ jobs:
 
       - uses: ./.github/actions/setup-turborepo-environment
         with:
-          windows: ${{ matrix.os.runner == 'windows-latest' }}
           github-token: "${{ secrets.GITHUB_TOKEN }}"
 
       - name: Check examples
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -221,7 +221,7 @@ jobs:
 
       - uses: ./.github/actions/setup-turborepo-environment
         with:
-          windows: ${{ matrix.os.name == 'windows' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -221,7 +221,7 @@ jobs:
 
       - uses: ./.github/actions/setup-turborepo-environment
         with:
-          windows: ${{ matrix.os.name == 'windows' }}
+          windows: ${{ matrix.os.runner == 'windows-latest' }}
```
chris-olszewski, line 224: Was there an issue with the previous comparison? Just making sure I'm not missing something with this change.

.github/workflows/test.yml
```diff
@@ -658,7 +654,7 @@ jobs:
       - name: Setup Turborepo Environment
         uses: ./.github/actions/setup-turborepo-environment
         with:
-          windows: ${{ matrix.os.name == 'windows' }}
+          windows: ${{ matrix.os.runner == 'windows-latest' }}
```
chris-olszewski, line 657: Again, was there an issue with the previous comparison with `name`?

chris-olszewski (approved):

.github/workflows/test.yml
mehulkar: you aren't missing anything. I just realized that `runner` is a real thing in the matrix, where the `name` key is made up by the user (us), and could be dropped with no effect on the matrix. So I'd rather compare to the real thing. E.g in integration tests, I didn't even put a name , because it doesn't matter: https://github.com/vercel/turbo/blob/29e73add257b489decb33837003d545ad391f018/.github/workflows/test.yml#L255-L259

PR #8001, opened 2024-04-18
chore: differentiate rust changes from building turbo

chris-olszewski:
### Description

This PR adds a new dummy task that will be invalidated whenever turborepo related Rust code changes.

I noticed we were building the `turbo` binary when running our [JS tests](https://github.com/vercel/turbo/actions/runs/8724966454/job/23936836212#step:6:993). This is due to `turborepo-repository` depending on `cli#build`, but I believe that we really just want it invalidate on any Rust code changes.

### Testing Instructions

Verify that updating a Rust file causes a cache miss for both `turbo build --filter=cli` and `turbo build --filter=turborepo-repository`.

### DIFF
--- a/cli/turbo.json
+++ b/cli/turbo.json
@@ -2,15 +2,9 @@
   "$schema": "../docs/public/schema.json",
   "extends": ["//"],
   "pipeline": {
-    "build": {
+    // A task that is used for detecting if any turborepo Rust sources change
+    "rust-src": {
       "env": ["RUNNER_OS"],
-      "outputs": [
-        "../target/debug/turbo",
-        "../target/debug/turbo.exe",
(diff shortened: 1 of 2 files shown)


mehulkar (approved):
FWIW, I think it would make sense to exclude the `turborepo-repository` tests as a check before release, since it is independently released. We do this in CI also: 
https://github.com/vercel/turbo/blob/e63b1a8590cf53e5e92e8239c002dbe6a4da3b72/.github/workflows/test.yml#L395

and `turborepo-repository` is independently tested in its [own workflow](https://github.com/vercel/turbo/blob/main/.github/workflows/turborepo-native-lib-test.yml)

Issue #7993, opened 2024-04-16
pnpm 9 lockfile support


pawelblaszczyk5:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

https://github.com/pawelblaszczyk5/pnpm-prune-repro

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

1.13.3-canary.1

### Describe the Bug

pnpm introduced a new lockfile format in the v9 major version. Formatting changes seems to mess up `turbo prune`, all stuff under `packages` key in output `pnpm-lock.yaml` is missing

### Expected Behavior

`turbo prune` should properly prune new pnpm lockfile format

### To Reproduce

- Update pnpm to v9
- Recreate lockfile so it utilizes the new `9.0` version
- Try to run `turbo prune some-package`
- Observe the prune output lockfile `packages` key being empty - it fails running `pnpm i --frozen-lockfile` because of that

### Additional context

I've pushed to the reproduction repo the lockfile after pruning before and after updating to pnpm v9. You can check the difference between `pnpm-lock-pruned-old.yaml` and `pnpm-lock-pruned-new.yaml`

anthonyshew:
Thanks for the report! It's expected that this would fail, given it's a new format, after all. We're aware of it and have it on our roadmap.

pawelblaszczyk5:
> Thanks for the report! It's expected that this would fail, given it's a new format, after all. We're aware of it and have it on our roadmap.

Yeah, I’m not surprised by this not working on day 1 of pnpm major release 😄 Just wanted to create this so it’s tracked somewhere and to make sure you’re aware of this. Thanks a lot!

chris-olszewski:
This should be supported with `1.13.3-canary.2`.

If anyone encounters any issues with pnpm 9, please reopen this issue along with a reproduction repo.

VojtechVidra:
Hello, I'm still having issus with pnpm 9. You can try by running `pnpm turbo prune --scope=web` in the following branch: https://github.com/RBND-studio/flows-cloud/pull/266

Hopefully it's not too complicated for you since it's not a minimal reproduction. Thanks!

StuClift:
Using https://github.com/vercel/turbo/tree/main/examples/kitchen-sink
```console
$ cd examples/kitchen-sink
$ rm pnpm-lock.yaml
[...]

PR #8003, opened 2024-04-19
fix(pnpm): correctly parse dependency paths with nested peer dependecies

chris-olszewski:
### Description

Should address the reported issues in #7993 

I missed a change to the dependency path format where nested peer dependencies are now shown e.g. if a has peer dependency b and b has peer dependency c, then the resulting dependency path will be `a@1.0.0(b@1.0.0(c@1.0.0))`.

Not sure when dependency path parsing got simplified, but I believe I missed the host segment getting dropped which greatly simplifies the format. The new parsing logic is a faithful port of the current JS parsing code: https://github.com/pnpm/pnpm/blob/main/packages/dependency-path/src/index.ts#L91

### Testing Instructions

Added unit test for parsing a dependency path that contains a nested peer dependency. Existing test suite.

### DIFF
--- a/crates/turborepo-lockfiles/src/pnpm/dep_path.rs
+++ b/crates/turborepo-lockfiles/src/pnpm/dep_path.rs
@@ -8,6 +8,16 @@ use nom::{
 
 use super::SupportedLockfileVersion;
 
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error(transparent)]
+    Nom(#[from] nom::error::Error<String>),
+    #[error("dependency path '{0}' contains no '@'")]
+    MissingAt(String),
+    #[error("dependency path '{0}' has an empty version following '@'")]
+    MissingVersion(String),
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct DepPath<'a> {
     // todo we possibly keep the full string here for 0-cost serialization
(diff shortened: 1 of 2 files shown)


crates/turborepo-lockfiles/src/pnpm/dep_path.rs
```diff
@@ -170,9 +192,13 @@ mod tests {
     #[test_case("foo@1.0.0", DepPath::new("foo", "1.0.0") ; "basic v7")]
     #[test_case("@scope/foo@1.0.0", DepPath::new("@scope/foo", "1.0.0") ; "scope v7")]
     #[test_case("foo@1.0.0(bar@1.2.3)", DepPath::new("foo", "1.0.0").with_peer_suffix(Some("(bar@1.2.3)")) ; "peer v7")]
+    #[test_case(
+        "eslint-module-utils@2.8.0(@typescript-eslint/parser@6.12.0(eslint@8.57.0)(typescript@5.3.3))",
```
tknickman, line 196: Yikes

crates/turborepo-lockfiles/src/pnpm/dep_path.rs
```diff
@@ -88,16 +95,31 @@ fn parse_dep_path(i: &str) -> IResult<&str, DepPath> {
...
+    // Need to check if sep_index is valid index
+    if sep_index >= input.len() {
+        return Err(Error::MissingVersion(input.to_owned()));
+    }
+    let (name, version) = input.split_at(sep_index);
+    let version = &version[1..];
+
+    let (version, peer_suffix) = match version.find('(') {
+        Some(paren_index) if version.ends_with(')') => {
+            // TODO: safety of the split?
```
tknickman, line 115: 👀

tknickman (approved):

crates/turborepo-lockfiles/src/pnpm/dep_path.rs
chris-olszewski: Oops, old comment. I had an off by one error previously 😅

crates/turborepo-lockfiles/src/pnpm/dep_path.rs
chris-olszewski: This is an abridged example the full christian name is `eslint-module-utils@2.8.0(@typescript-eslint/parser@6.12.0(eslint@8.57.0)(typescript@5.3.3))(eslint-import-resolver-node@0.3.9)(eslint-import-resolver-typescript@3.6.1(@typescript-eslint/parser@6.12.0(eslint@8.57.0)(typescript@5.3.3))(eslint-plugin-import@2.29.0)(eslint@8.57.0))(eslint@8.57.0)`

tknickman: 💀

PR #8017, opened 2024-04-22
feat: require packageManager in package.json

chris-olszewski:
### Description

With 2.0 we will now be requiring a `packageManager` field in `package.json` as this is a best practice and it helps us behave in a deterministic manner.

The actual code change is very straightforward as we remove our package manager inference code and return an error if reading package manager from `package.json` fails.

Most of the PR is updating tests.

### Testing Instructions

Updated unit tests


tknickman (approved):
Nice!

PR #8024, opened 2024-04-23
chore(ci): pin Github Actions runner that builds turborepo to macos-12

mehulkar:
macos-14 appears to not be able to build turborepo successfully
and macos-latest was recently updated to point to macos-14.
Pinning this until we figure out how to make macos-14 work

I didn't change JS package tests or turbopack related things
Closes TURBO-2862

### DIFF
--- a/.github/workflows/bench-turborepo.yml
+++ b/.github/workflows/bench-turborepo.yml
@@ -56,7 +56,7 @@ jobs:
           - name: ubuntu
             runner: ubuntu-latest
           - name: macos
-            runner: macos-latest
+            runner: macos-12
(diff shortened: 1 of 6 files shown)


chris-olszewski:
> I didn't change JS package tests or turbopack related things

I think we need to update these as well as they all hit the `setup-rust` action which installs `protoc`

chris-olszewski (approved):
Once CI passes, then :shipit:

mehulkar:
JS package tests shouldn't need to setup-rust, but yeah they do right now. replaced all the workflows now

PR #8036, opened 2024-04-24
feat(turborepo): Process package change events asynchronously

NicholasLYang:
### Description

We were previously processing the package change events synchronously, which ended up creating too much lagging and therefore crashes. Now we spin up a thread that reads in the package change events, and accumulates the changed packages. A separate thread handles execution by taking those changed packages and spinning up a new run.

Also I changed our logic to just send rediscover events on lagging with the channel in the daemon server. No reason to crash the server in that case.

The first commit is more of a refactor to put the watch state into `WatchClient`, to clean up the state ahead of splitting into two threads. You can review commit by commit if you wish.

### Testing Instructions

Tested on `next.js`, lagging is definitely improved.

### DIFF
--- a/crates/turborepo-lib/src/run/watch.rs
+++ b/crates/turborepo-lib/src/run/watch.rs
@@ -18,7 +18,15 @@ use crate::{
     DaemonConnector, DaemonPaths,
 };
 
-pub struct WatchClient {}
+pub struct WatchClient {
+    run: Run,
+    persistent_tasks_handle: Option<JoinHandle<Result<i32, run::Error>>>,
+    current_runs: HashMap<PackageName, JoinHandle<Result<i32, run::Error>>>,
+    connector: DaemonConnector,
+    base: CommandBase,
+    telemetry: CommandEventBuilder,
+    handler: SignalHandler,
+}
 
 #[derive(Debug, Error, Diagnostic)]
 pub enum Error {
@@ -51,6 +59,8 @@ pub enum Error {
     },
     #[error("daemon connection closed")]
     ConnectionClosed,
+    #[error("failed to subscribe to signal handler, shutting down")]
+    NoSignalHandler,
     #[error("watch interrupted due to signal")]
     SignalInterrupt,
     #[error("package change error")]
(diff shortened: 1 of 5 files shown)


NicholasLYang:
* **#8036** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8036?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/turborepo-lib/src/run/watch.rs
```diff
@@ -76,43 +104,56 @@ impl WatchClient {
...
-                .await?;
+                Self::handle_change_event(&changed_packages, event.event.unwrap()).await?;
             }
 
             Err(Error::ConnectionClosed)
         };
 
+        let run_fut = async {
+            loop {
+                if !changed_packages.lock().unwrap().borrow().is_empty() {
```
NicholasLYang, line 150: There's a slight race condition here, but it's okay, since the amount of changed packages is always going to increase, not decrease (it's only decremented in this thread), so we'll always need to do a run. Also even if we don't run anything, it'll just be slightly inefficient.

crates/turborepo-lib/src/run/watch.rs
```diff
@@ -76,43 +104,56 @@ impl WatchClient {
...
+                    self.execute_run(&changed_packages).await?;
+                }
+                yield_now().await;
```
gsoltis, line 153: Can you document what's going on here? I'm guessing it's to avoid spinning and not receiving any new events...

Maybe there's a better way to structure that? Like a watch or something so we can block until there's a change?

gsoltis (approved):
The `yield_now` looks suspicious. But maybe it just needs a comment?

crates/turborepo-lib/src/run/watch.rs
NicholasLYang: Yeah we could probably do this with a watch channel. Will give that a shot

crates/turborepo-lib/src/run/watch.rs
```diff
@@ -76,43 +109,58 @@ impl WatchClient {
...
+                Self::handle_change_event(&changed_pkgs_tx, event.event.unwrap()).await?;
             }
 
             Err(Error::ConnectionClosed)
         };
 
+        let run_fut = async {
+            loop {
+                changed_pkgs_rx.changed().await?;
+                let changed_pkgs = changed_pkgs_rx.borrow_and_update();
```
gsoltis, line 156: I think this will block the producing side until `changed_pkgs` is dropped. You may want to clone it so that the borrow is short-lived, or at least not held for the duration of a run.

crates/turborepo-lib/src/run/watch.rs
gsoltis: I think with the watch, we don't need this line anymore, right?

gsoltis (changes requested):

crates/turborepo-lib/src/run/watch.rs
gsoltis: Docs on borrowing [here](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.borrow_and_update) indicate that this is not recommended, and we should find a way to shorten the borrow.

gsoltis (approved):

PR #8057, opened 2024-04-29
fix(turborepo): Watch mode not responding to changes

NicholasLYang:
### Description

When I converted to a `watch` channel, I ended up introducing an annoying bug. Basically you can't tell if the value has been read already in a `watch` channel. So if you send a rediscover, then with subsequent events it's impossible to determine if the rediscover has been seen and you should send a new event, or if it's not seen and therefore you should send nothing.

I fixed this by reverting back to the lock version and addressed @gsoltis's comments by using a tokio Mutex which yields to the runtime when grabbing the lock.

### Testing Instructions

Validated that we get the right change events. Also validated that with regular Mutex and no `yield_now`, we get a deadlock, while with tokio Mutex and no `yield_now` we don't deadlock.

### DIFF
--- a/crates/turborepo-lib/src/run/watch.rs
+++ b/crates/turborepo-lib/src/run/watch.rs
@@ -1,11 +1,10 @@
-use std::collections::HashSet;
+use std::{cell::RefCell, collections::HashSet, sync::Mutex};
 
 use futures::StreamExt;
 use miette::{Diagnostic, SourceSpan};
 use thiserror::Error;
 use tokio::{
     select,
-    sync::watch,
     task::{yield_now, JoinHandle},
 };
 use turborepo_repository::package_graph::PackageName;
@@ -22,7 +21,7 @@ use crate::{
     DaemonConnector, DaemonPaths,
 };
 
-pub enum ChangedPackages {
+enum ChangedPackages {
     All,
     Some(HashSet<PackageName>),
 }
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/run/watch.rs
```diff
@@ -140,23 +131,25 @@ impl WatchClient {
...
             Err(Error::ConnectionClosed)
         };
 
         let run_fut = async {
             loop {
-                changed_pkgs_rx.changed().await?;
-                let changed_pkgs = { changed_pkgs_rx.borrow_and_update().clone() };
-
-                self.execute_run(changed_pkgs).await?;
+                if !changed_packages.lock().await.borrow().is_empty() {
```
gsoltis, line 150: Does this spin? If `changed_packages` is in fact empty, won't this loop back around? It will yield the event loop, but the task will always be runnable as long as nothing else is holding the lock...

crates/turborepo-lib/src/run/watch.rs
```diff
@@ -140,23 +131,25 @@ impl WatchClient {
...
+                    self.execute_run(&changed_packages).await?;
```
gsoltis, line 151: Since we're about the lock this in `execute_run`, should we instead copy the value out and pass that in rather than the lock?

gsoltis (changes requested):

crates/turborepo-lib/src/run/watch.rs
NicholasLYang: If I understand correctly, if changed packages is empty, yes it’ll loop around, but on the next time it tries to lock, it’ll yield first to let the other task run and add to the changed packages. 

Could you clarify what you mean by the task being runnable?

gsoltis: It will yield, but the tokio scheduler will always see this loop as ready to run, even if there's no work to be done. If instead we use a channel or some other notification mechanism, tokio can park this task until there is work to be done.

gsoltis (approved):

PR #8058, opened 2024-04-29
fix(Turborepo): Make package discovery async, and apply a debouncer

### LINKED ISSUE #3455: Daemon should have been opt in, not opt out with --no-daemon, add config opt out in turbo.json
### Which project is this feature idea for?

Turborepo

### Describe the feature you'd like to request

Add config opt out in turbo.json

```json
{
"daemon": false
}
```

### Describe the solution you'd like

A config option to disable by default.

### Describe alternatives you've considered

I run with `--no-daemon` as an alt. This is due to the daemon not really working correctly. It constantly spits out erroneous errors, specifically error Unknown. It clutters the output with confusing messages and doesn't improve performance. I also don't need remote caching. As mentioned the daemon should be opt in and configure, not opt out.

gsoltis:
### Description

 - move the debouncer to its own module so that it is reusable
 - apply the debouncer to package discovery
 - make package discovery async so it doesn't block the file watching channel

### Testing Instructions

Existing test suite.

Fixes #3455

Closes TURBO-2909

### DIFF
--- a/crates/turborepo-filewatch/src/package_watcher.rs
+++ b/crates/turborepo-filewatch/src/package_watcher.rs
@@ -1,16 +1,24 @@
 //! This module hosts the `PackageWatcher` type, which is used to watch the
 //! filesystem for changes to packages.
 
-use std::{collections::HashMap, path::Path};
+use std::{
+    collections::HashMap,
+    path::Path,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
 
 use futures::FutureExt;
 use notify::Event;
 use thiserror::Error;
 use tokio::{
-    join,
+    join, select,
     sync::{
         broadcast::{self, error::RecvError},
-        oneshot, watch,
+        mpsc, oneshot, watch,
     },
 };
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
@@ -24,6 +32,7 @@ use turborepo_repository::{
 
 use crate::{
     cookies::{CookieRegister, CookieWriter, CookiedOptionalWatch},
+    debouncer::Debouncer,
     optional_watch::OptionalWatch,
     NotifyError,
 };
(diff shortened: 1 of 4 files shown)


weyert:
Could this solve the issue I am experiencing?

```
2024-04-29T22:41:44.167258Z ERROR turborepo_lib::daemon::server: package changes stream closed: channel closed
```

```bash
pnpm exec turbo watch build --filter={packages/\*}
• Packages in scope: <snip>
• Running build in 36 packages
• Remote caching enabled
  × failed to connect to daemon
  ╰─▶ server is unavailable: channel closed
```

gsoltis:
Possibly. If you're seeing errors in the daemon log from the package
watcher, then there's a good chance.

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -443,6 +516,47 @@ impl Subscriber {
     }
 }
 
+async fn discovery_packages(repo_root: AbsoluteSystemPathBuf) -> PackageState {
```
arlyon, line 519:
```suggestion
async fn discover_packages(repo_root: AbsoluteSystemPathBuf) -> PackageState {
```

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -443,6 +516,47 @@ impl Subscriber {
...
+    // If we're rediscovering everything, we need to rediscover the package manager.
+    // It may have changed if a lockfile changed or package.json changed.
+    let discovery = match LocalPackageDiscoveryBuilder::new(repo_root.clone(), None, None).build() {
+        Ok(discovery) => discovery,
+        Err(e) => return PackageState::NoPackageManager(e.to_string()),
+    };
+    let initial_discovery = match discovery.discover_packages().await {
+        Ok(discovery) => discovery,
+        // If we failed the discovery, that's fine, we've reset the values, leave them as None
```
arlyon, line 528: Lets just get rid of the references to None here since they aren't relevant now that we've pulled out the fn

crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -186,32 +244,56 @@ impl Subscriber {
...
+                            self.bump_or_queue_rediscovery(&mut state, &package_state_tx);
+                        }
+                    }
                 }
             }
             tracing::trace!("package watcher state: {:?}", state);
         }
     }
 
+    fn handle_discovery_result(&self, package_result: DiscoveryResult, state: &mut State) {
```
arlyon, line 281: What should the behaviour be if we get a `DiscoveryResult` with valid `State`? Should our state perhaps also store a version and be updated?

arlyon (commented):
A few style items and a quick Q

crates/turborepo-filewatch/src/package_watcher.rs
gsoltis: Yup, done.

NicholasLYang (approved):

weyert:
Would it be possible to create a canary build that will include this PR when its merged? Love to try this change

PR #8060, opened 2024-04-30
feat: Add lint for `grid-template-areas`

kdy1:
### Description

> grid-template-areas and grid-area behaves differently in lightningcss and that leads to some user problems, so we need a check to catches these errors.
> Since lightning css in css modules is renaming grid area names, you have to use grid-area and grid-template-areas in the same file together. Using them in different files would be an error.
> 
> So it would be great to have a check that emits an error when using grid-area in a CSS file but not declaring the name in grid-template-areas.

Closes PACK-3022

### Testing Instructions

I added some tests

### DIFF
--- a/crates/turbopack-css/src/process.rs
+++ b/crates/turbopack-css/src/process.rs
@@ -926,6 +926,31 @@ impl lightningcss::visitor::Visitor<'_> for CssModuleValidator {
                 }
             }
 
+            lightningcss::properties::Property::GridColumn(p) => {
+                for line in [&p.start, &p.end] {
+                    if let lightningcss::properties::grid::GridLine::Area { name } = line {
+                        self.used_grid_areas.insert(name.to_string());
+                    }
+                }
+            }
+
+            lightningcss::properties::Property::GridRow(p) => {
+                for line in [&p.start, &p.end] {
+                    if let lightningcss::properties::grid::GridLine::Area { name } = line {
+                        self.used_grid_areas.insert(name.to_string());
+                    }
+                }
+            }
+
+            lightningcss::properties::Property::GridRowStart(p)
+            | lightningcss::properties::Property::GridRowEnd(p)
+            | lightningcss::properties::Property::GridColumnStart(p)
+            | lightningcss::properties::Property::GridColumnEnd(p) => {
+                if let lightningcss::properties::grid::GridLine::Area { name } = p {
+                    self.used_grid_areas.insert(name.to_string());
+                }
+            }
+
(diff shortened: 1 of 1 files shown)


crates/turbopack-css/src/process.rs
```diff
@@ -731,14 +732,29 @@ async fn process_content(
 /// ```
 ///
 /// is wrong for a css module because it doesn't have a class name.
-struct CssValidator {
+#[derive(Default)]
+struct CssModuleValidator {
     errors: Vec<CssError>,
+    used_grid_areas: HashSet<String>,
```
sokra, line 738:
```suggestion
    used_grid_areas: IndexSet<String>,
```

Errors should be deterministic

crates/turbopack-css/src/process.rs
```diff
@@ -878,11 +906,38 @@ impl swc_core::css::visit::Visit for CssValidator {
...
-        visit_types!(SELECTORS)
+        lightningcss::visitor::VisitTypes::all()
+    }
+
+    fn visit_property(
+        &mut self,
+        property: &mut lightningcss::properties::Property,
+    ) -> Result<(), Self::Error> {
+        match property {
+            lightningcss::properties::Property::GridArea(p) => {
```
sokra, line 921: Is this handling these properties too?
[grid-row-start](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-row-start)
[grid-column-start](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column-start)
[grid-row-end](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-row-end)
[grid-column-end](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column-end)

Could we have a test case for that?

crates/turbopack-css/src/process.rs
```diff
@@ -878,11 +906,38 @@ impl swc_core::css::visit::Visit for CssValidator {
...
+                for line in [&p.row_start, &p.row_end, &p.column_start, &p.column_end] {
+                    if let lightningcss::properties::grid::GridLine::Area { name } = line {
+                        self.used_grid_areas.insert(name.to_string());
+                    }
+                }
+            }
+
+            lightningcss::properties::Property::GridTemplateAreas(
```
sokra, line 929: Is this handling that property too?

[grid-template](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template)

sokra (approved):

PR #8064, opened 2024-04-30
refactor(turborepo): Improve package changes watcher performance

NicholasLYang:
### Description

Like #8036, we need to split out the event processing loop from the actual logic to avoid lagging channels. In this case we process the changed files every 100ms.

Uses a `Trie` to store the paths since they'll likely have a lot of overlap.

### Testing Instructions

Closes TURBO-2921


crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -167,121 +228,133 @@ impl Subscriber {
...
+                    })
+                    .collect();
+
+                if changed_files.is_empty() {
+                    continue;
+                }
+
+                let changed_packages = change_mapper.changed_packages(changed_files.clone(), None);
+
+                tracing::warn!("changed_files: {:?}", changed_files);
```
NicholasLYang, line 288: `warn` is probably not the right level semantically, but `debug` has so much stuff in it and we don't log it out by default

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -167,121 +228,133 @@ impl Subscriber {
...
-                            }
-                            Ok(PackageChanges::Some(mut changed_pkgs)) => {
-                                if !changed_pkgs.is_empty() {
-                                    tracing::debug!(
-                                        "changed files: {:?} changed packages: {:?}",
-                                        changed_files,
-                                        changed_pkgs
-                                    );
-                                }
+                interval.tick().await;
```
NicholasLYang, line 234: 100ms should be enough time that we don't need to park/notify, right?

gsoltis: Yeah, I think that should be ok. We can ITG once the feature is stable.

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -167,121 +230,133 @@ impl Subscriber {
...
-                                    if !has_root_tasks {
-                                        changed_pkgs.remove(&root_pkg);
-                                    }
-                                }
+                let changed_files = self.changed_files.lock().await;
+                if changed_files.borrow().is_empty() {
+                    continue;
+                }
+
+                let ChangedFiles::Some(trie) = changed_files.take() else {
```
gsoltis, line 246: Does this drop the lock?

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -167,121 +230,133 @@ impl Subscriber {
...
+                let gitignore_path = self.repo_root.join_component(".gitignore");
+                if trie.get(gitignore_path.as_std_path()).is_some() {
+                    let (new_root_gitignore, _) = Gitignore::new(&gitignore_path);
+                    root_gitignore = new_root_gitignore;
+                }
+
+                let changed_files: HashSet<_> = trie
+                    .keys()
+                    .filter_map(|p| {
+                        let p = AbsoluteSystemPathBuf::try_from(p.as_path()).ok()?;
```
gsoltis: Should this be `.expect("filewatching returns absolute paths")`?

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -167,121 +230,133 @@ impl Subscriber {
...
+                        self.repo_root.anchor(p).ok()
+                    })
+                    .filter(|p| {
+                        // If in .gitignore or in .git, filter out
+                        !(ancestors_is_ignored(&root_gitignore, p) || is_in_git_folder(p))
```
gsoltis, line 283: I suspect that at some point this'll be an issue w/ `inputs`.

gsoltis (approved):
A few questions, but other than that ship it

crates/turborepo-lib/src/package_changes_watcher.rs
NicholasLYang: yeah there's a subsequent PR that hooks up file hashing and removes this check

crates/turborepo-lib/src/package_changes_watcher.rs
NicholasLYang: Do we want to crash the daemon on this? That seems a little drastic

NicholasLYang:
* **#8064** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8064?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #8082, opened 2024-05-03
Graph aggregation refactoring

sokra:
### Description

* Deletes the aggregation tree
* Adds a new graph aggregation algorithm which is more efficient

The graph aggregation works as following:

For the graph aggregation: Every task is a node in the graph. Every parent-child relationship is an edge in the graph.

* Every node has an "aggregation number" N.
* There are 2 kinds of nodes: Leaf nodes and aggregating nodes.
* If a node has N < LEAF_NUMBER, it's a leaf node, otherwise an aggregating node.
* A higher N for a node usually means that a larger subgraph is aggregated into that node.
* Next to normal edges there are two extra kind of edges for the graph aggregation: Upper edges and follower edges.
* A node is considered as "inner" to another node when it has an "upper" edge pointing towards it.
* The inner node has a lower N than the upper node. (This invariant might be temporarily violated while tree balancing is scheduled but not executed yet)
* Aggregating nodes store an aggregated version of the state of all inner nodes and transitively inner nodes.
* Changes in nodes are propagated to all upper nodes.
* Every node has at least one upper node which is more aggregated than the node. Except for the root node of the graph, which doesn't have upper edges.
* An aggregating node also has follower edges. They point to the nodes that are one normal edge after all inner and transitively inner nodes.
* An leaf node doesn't have follower edges. For all purposes the normal edges of leaf nodes are considered as follower edges.
* Follower nodes have a higher N than the origin node. (This invariant might be temporarily violated while tree balancing is scheduled but not executed yet)
* This means large and larger subgraphs are aggregated.
* Graph operations will ensure that these invariants (Higher N on upper and follower edges) are not violated.
* The N of a node can only increase. So graph operations need to "fix" the invariants by increasing N or changing upper/follower edges. That later one is preferred. N is usually only increased if two nodes have equal N.
* When new edges between leaf nodes are added, the target node's N is increased to the origin node's N + 4 if it's smaller. This adds a small tolerance range so increasing N doesn't cause long chains of N += 1 between leaf nodes.

### Testing Instructions

Closes PACK-3036


crates/turbo-tasks-memory/src/aggregation/balance_edge.rs
```diff
@@ -0,0 +1,206 @@
...
+                            count,
+                            true,
+                        ));
+                    }
+                } else {
+                    // we already have extra followers, remove some uppers to balance
+                    let count = extra_followers + extra_uppers;
+                    let target = ctx.node(target_id);
+                    if is_in_progress(ctx, upper_id) {
+                        drop(target);
```
arlyon, line 97: What is the significance of borrowing before the branch and then dropping?

crates/turbo-tasks-memory/src/aggregation/in_progress.rs
```diff
@@ -0,0 +1,75 @@
+use std::{hash::Hash, mem::take};
+
+use super::{balance_queue::BalanceQueue, AggregationContext, AggregationNode, StackVec};
+
+impl<I: Clone + Eq + Hash, D> AggregationNode<I, D> {
+    /// Finishes an in progress operation. This might enqueues balancing
```
arlyon, line 6:
```suggestion
    /// Finishes an in progress operation. This might enqueue balancing
```

crates/turbo-tasks-memory/src/aggregation/in_progress.rs
```diff
@@ -0,0 +1,75 @@
...
+        debug_assert!(value > 0);
+        if value == 1 {
+            if let AggregationNode::Aggegating(aggegating) = &mut *self {
+                balance_queue.balance_all(take(&mut aggegating.enqueued_balancing))
+            }
+        }
+    }
+}
+
+/// Finishes an in progress operation. This might enqueues balancing
```
arlyon, line 26:
```suggestion
/// Finishes an in progress operation. This might enqueue balancing
```

arlyon (approved):

crates/turbo-tasks-memory/src/aggregation/balance_edge.rs
sokra: We need to check in progress (which is an atomic) while either holding the target or the upper lock. In progress is only set during the `target` lock, so when we read it it need to be under the `target` lock. If not in progress we can continue working with the target in the else branch.

Otherwise we want to enqueue our work to the upper node. So we acquire a upper lock. In the meantime the in_progress flag might have changed, so we need to check that again.

ForsakenHarmony:
Can we maybe add a Markdown doc next to the code describing the overall approach (or just copy the PR description in there)

sokra:
> Can we maybe add a Markdown doc next to the code describing the overall approach (or just copy the PR description in there)

I'll add this in a follow-up PR

Issue #8072, opened 2024-05-02
Turborepo panics when Ctrl-Cing


weyert:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

n/a

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

turbo 1.13.4-canary.0

### Describe the Bug

name = "turbo"
operating_system = "Mac OS 14.4.1 [64-bit]"
crate_version = "1.13.4-canary.0"
explanation = """
file '/Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.26.1/src/terminal/terminal.rs' at line 442
"""
cause = "chunk size must be non-zero"
method = "Panic"
backtrace = """

   0: 0x10530e924 - __mh_execute_header
   1: 0x104f31dd4 - __mh_execute_header
   2: 0x104b6c764 - __mh_execute_header
   3: 0x104b69cf4 - __mh_execute_header
   4: 0x104b32430 - __mh_execute_header
   5: 0x18a932f94 - __pthread_joiner_wake"""

### Expected Behavior

Not panicing when trying to close/stop turbo using ctrl-c

### To Reproduce

Not sure how

weyert:
After getting out of sleep over night, got a similar panic:
```bash
name = "turbo"
operating_system = "Mac OS 14.4.1 [64-bit]"
crate_version = "1.13.4-canary.0"
explanation = """
file '/Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.26.1/src/terminal/terminal.rs' at line 442
"""
cause = "chunk size must be non-zero"
method = "Panic"
backtrace = """

   0: 0x103b62924 - __mh_execute_header
   1: 0x103785dd4 - __mh_execute_header
   2: 0x1033c0764 - __mh_execute_header
   3: 0x1033bdcf4 - __mh_execute_header
   4: 0x103386430 - __mh_execute_header
   5: 0x18b3c6f94 - __pthread_joiner_wake"""
```

Issue #7843, opened 2024-03-26
Turborepo canary panics with 'index out of bounds: the len is 65535 but the index is 65535'


weyert:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

None

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

1.13.1-canary.2

### Describe the Bug

```
  1   │ name = "turbo"
   2   │ operating_system = "Mac OS 14.4.0 [64-bit]"
   3   │ crate_version = "1.13.1-canary.2"
   4   │ explanation = """
   5   │ file '/Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.26.1/src/buffer/buffer.rs' at line 108
   6   │ """
   7   │ cause = "index out of bounds: the len is 65535 but the index is 65535"
   8   │ method = "Panic"
   9   │ backtrace = """
  10   │
  11   │    0: 0x1005b74e4 - __mh_execute_header
  12   │    1: 0x100f78ab0 - __mh_execute_header
  13   │    2: 0x100f7dca8 - __mh_execute_header
  14   │    3: 0x100daf650 - __mh_execute_header
  15   │    4: 0x10080dfa4 - __mh_execute_header
  16   │    5: 0x100804bb4 - __mh_execute_header
  17   │    6: 0x1007db824 - __mh_execute_header
  18   │    7: 0x195be2f94 - __pthread_joiner_wake"""
```

### Expected Behavior

Not to crash

### To Reproduce

Not sure how to reproduce it without sharing work code

chris-olszewski:
@weyert Is this happening consistently?

weyert:
Hmm, that's a good question. I have downgraded to the .0 to continue with my work. I am not sure if it could because I had to instance of `turbo run [command]` running with both running the same steps. 

It did look a bit odd with multiple 'window's of log items instead of one big one

chris-olszewski:
Hmm, interesting. The panic message does look like some integer underflow so I'm doing a quick comb through of recent changes. If you have the time could you see if it happens with canary.1 or just canary.2.

Could you share what OS/terminal you were using?

weyert:
[...]

PR #8084, opened 2024-05-03
feat(ui): render ui to alternative screen

chris-olszewski:
### Description

Changes our UI to render in the alternative screen (think `vim` or `less`)

This provides a few benefits:
 - We can render to the entire terminal without worrying about over-allocating space and removing useful information from the screen.
 - Users won't need to scroll up at the end of a run to see the task logs
 - We no longer use `insert_before` for persisting task logs. This function panics if there isn't an available row to render to (#8072) or rendering terminal logs that have an area that exceeds `u16::MAX` (#7843). Instead our log persisting story is a "simple" `stdout.write_all()`.
 - Removes hacks added to avoid hitting #7843 

We write the logs in a row-wise fasion as `vt100` attempts to optimize performance by using cursor moves to avoid necessary writes. This creates problems when printing multiple terminal screens as the cursor move coordinates will be incorrect.

### Testing Instructions

https://github.com/vercel/turbo/assets/4131117/9746dcd3-188e-48fd-a52d-21d95d0e6e43

Closes TURBO-2979


weyert:
Would this make it easier to allow selection of log entries?

chris-olszewski:
> Would this make it easier to allow selection of log entries?

Not sure I follow. As in easier to copy/paste from logs from tasks after `turbo` has exited?

chris-olszewski:
> Easier to select from the experimental-ui screen for logging

Ah, no. That broke due to us supporting mouse scroll for scrolling through task outputs. I might need to disable that feature and only allow keybinds. The alternative is to manually implement terminal selection and click support which isn't something I want to tackle at this moment.

arlyon (approved):

crates/turborepo-vt100/src/entire_screen.rs
```diff
@@ -44,6 +27,53 @@ impl<'a> EntireScreen<'a> {
...
+        let mut wrapping = false;
+        let grid = self.screen.grid();
+        let (rows, _) = self.size();
+        let default = crate::attrs::Attrs::default();
+        grid.all_rows().take(rows).enumerate().map(move |(i, row)| {
+            // number of rows in a grid is stored in a u16 (see Size), so
+            // visible_rows can never return enough rows to overflow here
+            let i = i.try_into().unwrap();
+            let mut contents = vec![];
+            // We don't need final cursor position as long as CRLF is used and not just LF
```
mehulkar, line 56: I don't understand this comment. Where would the CRLF or LF be coming from? is it user input? (in which case, is LF not possible on Windows?)

mehulkar (approved):
The video looks great! Not a blocker, but could we still have colors on the group titles?

CleanShot 2024-05-08 at 11 29 16@2x

crates/turborepo-vt100/src/entire_screen.rs
chris-olszewski: When a terminal is in raw mode a LF only will move the cursor down a row, but not return it to the home (zero) column. This leads the string `foo\nbar` being printed in a raw terminal looking like
```
foo
   bar
```
where `foo\r\nbar` will be printed as expected
```
foo
bar
```

The comment is meant to explain why we aren't storing/restoring the final cursor position as we expect the caller to print these lines with a trailing `\r\n`.

PR #8087, opened 2024-05-03
#[turbo_tasks::value(transparent)]: Generate docs & fail on invalid callsites

bgw:
### Description

Most `#[turbo_tasks::value(transparent)]` types leave their inner value as private. I think this okay, but because rustdoc hides private fields by default, it makes it hard to understand at a glance the contained value, without scrolling down to the `impl` of `VcValueType`. This PR generates documentation for these.

Screenshot 2024-05-02 at 10 27 47 PM

<details>
<summary>Also checked that it works if there's an existing doc comment, extending rather than replacing the existing documentation.</summary>

Screenshot 2024-05-02 at 10 25 06 PM
</details>

**Alternative:** We could fail if the field is non-`pub`, and update the callsites. Let me know if this is preferred. The contained fields are *basically* public anyways, as they can be accessed via `Vc`'s APIs.

While modifying this code, I realized that we don't generate an error if `#[turbo_tasks::value(transparent)]` would be a no-op. The second commit in this PR adds an error and updates the callsites. This also exposed some issues in the next.js repository, which are fixed in https://github.com/vercel/next.js/pull/65337 .

### Testing Instructions

Closes PACK-3038

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10662,6 +10662,7 @@ dependencies = [
  "proc-macro-error",
  "proc-macro2",
  "quote",
+ "regex",
  "syn 1.0.109",
  "turbo-tasks-macros-shared",
 ]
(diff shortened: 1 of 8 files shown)


bgw:
* **#8087** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8087?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sokra (approved):

PR #8102, opened 2024-05-07
fix memory counting without custom allocator

sokra:
### Testing Instructions

Closes PACK-3044


ForsakenHarmony (approved):
Should the PR title be "with custom allocator"?

Issue #7829, opened 2024-03-24
Docs: Add an example for Angular


pBouillon:
### What is the improvement or update you wish to see?

Many front-end frameworks are shown in [turbo/examples](https://github.com/vercel/turbo/tree/main/examples) but sadly none of them feature Angular

Nx is a well established monorepo tool in the Angular ecosystem and having such an example provided my Turbo could help Angular developers adopting it

### Is there any context that might help us understand?

I would love to see the equivalent of the SvelteKit example for Angular for example.

anthonyshew:
Happy to review a PR for this!

eneajaho:
Can I take this one?

alydemah:
I've started doing this a while ago in new turborepo project and noticed the lack of example,
I would be happy to also provide feedback how I done it.

anthonyshew:
@eneajaho We can all work with whatever PR shows up first. 😄

Zaid-maker:
I made a PR but i dont have any knowledge in Angular so u guys can assist me how can i do it Thanks

PR #8103, opened 2024-05-07
feat(example): Introduce `with-angular` monorepo example project

Zaid-maker:
### Description

I have no knowledge in Angular whatsoever, but i would to contribute to add this example requested in #7829 I would like anyone to assist me in this PR. Thanks


Zaid-maker:
My bad i didn't check someone has already opened a PR but they didn't link it or something

PR #8104, opened 2024-05-07
feat(turborepo): Using file hashing for package watching

NicholasLYang:
### Description

We leverage the file hash watcher to make watch mode more stable. Basically, once we get a list of changed packages, we query the file hash watcher to see if there are file hashes available. If so, we check if we've seen these file hashes before. If they are, then we don't re-run.

This does have some caveats. For one, file hashes are not exactly correct for whether a task should be re-run. We really should be checking task hashes, but that will require more daemon infrastructure. Also files are most likely the only changes a user will make with watch mode (will they edit env vars? Probably not), so this is good enough.

Also, this will only work on the *second* run, since we need to get the file hash and store it. Maybe even the third if hashing takes too long to get.

### Testing Instructions

I tested on the next.js repo. You can confirm the behavior by looking at the daemon logs and seeing that we don't get a re-run when running `touch foo` multiple times, since the file has the same content.

Closes TURBO-3010

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10859,6 +10859,7 @@ dependencies = [
  "turbopack-static",
  "turbopack-swc-utils",
  "turbopack-test-utils",
+ "turbopack-trace-server",
  "turbopack-trace-utils",
 ]
 
@@ -10920,6 +10921,7 @@ dependencies = [
  "turbopack-node",
  "turbopack-nodejs",
  "turbopack-resolve",
+ "turbopack-trace-server",
  "turbopack-trace-utils",
  "webbrowser",
 ]
--- a/crates/turborepo-lib/src/daemon/server.rs
+++ b/crates/turborepo-lib/src/daemon/server.rs
@@ -136,7 +136,11 @@ impl FileWatching {
             scm,
         ));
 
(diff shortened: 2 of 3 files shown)


NicholasLYang:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/vercel/turbo/8104?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#8104** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8104?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#8064** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8064?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

weyert:
> This does have some caveats. For one, file hashes are not exactly correct for whether a task should be re-run. We really should be checking task hashes, but that will require more daemon infrastructure. Also files are most likely the only changes a user will make with watch mode (will they edit env vars? Probably not), so this is good enough.

I don't even know how you can change env vars of a running process :)

gsoltis (approved):

PR #8109, opened 2024-05-08
fix(gen): isolate generator to `commonjs`

### LINKED ISSUE #5657: [turborepo] Codegenerator not support ESM

v1.10.9

Mac

### Describe the Bug

```
Unexpected error. Please report it as a bug:
Must use import to load ES Module: /Users/peam/work/softnetics/<app>/turbo/generators/config.ts
require() of ES modules is not supported.
require() of /Users/peam/work/softnetics/<app>/turbo/generators/config.ts from /Users/peam/.npm/_npx/111d46748c4e70c7/node_modules/node-plop/lib/node-plop.js is an ES module file as it is a .ts file whose nearest parent package.json contains "type": "module" which defines all .ts files in that package scope as ES modules.
Instead change the requiring code to use import(), or remove "type": "module" from /Users/peam/work/softnetics/<app>/package.json.
```

I got this error when use "type: 'module'" in package.json

### Expected Behavior

It should work ...

### To Reproduce

use "type: 'module'" in package.json

### Reproduction Repo

_No response_

wesleycoder:
### Description

As described by #5657, the default `simple-*` templates creates a template that is broken for `"type": "module"` packages.

When digging into this issue I found out that a simple `package.json` with `"type": "commonjs"` solves the problems by telling the runtime to treat the file diferently from the parent package.

### Testing Instructions

- Initialize a new monorepo with `"type": "module"` in the `package.json`
- on this branch it should now be possible to generate the sample generator (either `js` or `ts`) and not have any errors about the module type.
- it should also be possible to run the generator and it should create the file in the project root normally.
resolves #5657

### DIFF
--- a/packages/turbo-gen/src/templates/simple-js/package.json
+++ b/packages/turbo-gen/src/templates/simple-js/package.json
@@ -1,0 +1,3 @@
+{
+  "type": "commonjs"
+}
--- a/packages/turbo-gen/src/templates/simple-ts/package.json
+++ b/packages/turbo-gen/src/templates/simple-ts/package.json
@@ -1,0 +1,3 @@
+{
+  "type": "commonjs"
+}


tknickman (approved):

tknickman:
Thanks!

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 0211ae4086fe9f4fc6b51c521d0dbe782dcec099

PR #8111, opened 2024-05-08
change ids from usize to u32

sokra:
### Description

reduces memory usage by ~10%

We never have more than 4 billion tasks

Closes PACK-3052


bgw:
As discussed on Slack, it would be nice to have a panic-on-overflow, but otherwise LGTM, and 10% is a nice easy win!

bgw (approved):

PR #8116, opened 2024-05-09
Update package.json

akshatarora130:
### Description

Updated package.json in the packages/ui  styles.css was not exported correctly


anthonyshew:
Hey, thanks for the suggestion! That path is correct since it will be generated by the Tailwind package when you run the `build` or `dev` script. The compiled styles will be there in `dist`.

PR #8127, opened 2024-05-10
fix(Turborepo): Handle new packages in lockfile comparisons

### LINKED ISSUE #8125: Turbo --filter builds all the packages when a new workspace is added
### Verify canary release

https://github.com/alainrk/turbo-test

pnpm

Mac

1.13.3

### Describe the Bug

When we create a new workspace that have node dependencies and build the monorepo using the filter option we have an unexpected behaviour.
What happens is that even though we haven't changed anything in the other workspaces (proper code changes or installing/updating npm packages), when we build with the filter option all the packages are being built regardless the filter used.
In particular this happens locally when we have unstaged changes and we filter using "HEAD" as mentioned in the documentation, or in CI where we use two commits to have the same result (e.g. PR last commit and Base branch commit).
Please note, that if the new workspace has no dependencies, the build behaviour is instead correct.

### Expected Behavior

The expected behaviour is that only the new workspace gets built.

### To Reproduce

Go to the repository (https://github.com/alainrk/turbo-test), clone it and just run `npm run reproduce-issue`, all the other details on how it works are in the README of the repo itself.

### Additional context

_No response_

gsoltis:
### Description

When comparing lockfiles between revisions to identify what packages have been affected, we treated missing packages in the previous lockfile (added in the new lockfile) as errors, and fell back to considering all packages as affected. This change adds a boolean to determine what the behavior should be when we can't find a package that we expect. When building the package graph, it continues to be an error, as well as when tracing the dependencies of a package we've already found in the lockfile. However, for the first round of packages that we look for in a previous lockfile, we ignore missing packages. This allows us to more gracefully handle comparisons between commits that add new packages.

### Testing Instructions

Added an integration test inspired by the repro from the linked issue.

Fixes #8125

### DIFF
--- a/turborepo-tests/integration/tests/lockfile-aware-caching/new-package.t
+++ b/turborepo-tests/integration/tests/lockfile-aware-caching/new-package.t
@@ -12,7 +12,9 @@ Update lockfile
 Now build and verify that only the new package is in scope
 Note that we need --skip-infer because we've now installed a local
 turbo in this repo
-  $ ${TURBO} --skip-infer build -F '[HEAD]' -F '!//' --dry=json | jq '.packages' 
+Note that we need to disable path conversion because on windows, git bash considers
+'//' to be an escape sequence translating to '/'.
+  $ MSYS_NO_PATHCONV=1 ${TURBO} --skip-infer build -F '[HEAD]' -F '!//' --dry=json | jq '.packages' 
   [
     "c"
   ]
--- a/turborepo-tests/integration/tests/lockfile-aware-caching/new-package.t
+++ b/turborepo-tests/integration/tests/lockfile-aware-caching/new-package.t
@@ -1,0 +1,20 @@
+Setup
+  $ . ${TESTDIR}/../../../helpers/setup.sh
+  $ . ${TESTDIR}/setup.sh $(pwd) pnpm
+
+Add new package with an external dependency
+  $ mkdir -p apps/c
+  $ echo '{"name":"c", "dependencies": {"has-symbols": "^1.0.3"}}' > apps/c/package.json
+
+Update lockfile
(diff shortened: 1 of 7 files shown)


NicholasLYang (approved):

turborepo-tests/integration/tests/lockfile-aware-caching/new-package.t
```diff
@@ -0,0 +1,22 @@
...
+
+Update lockfile
+  $ pnpm i --frozen-lockfile=false > /dev/null
+
+Now build and verify that only the new package is in scope
+Note that we need --skip-infer because we've now installed a local
+turbo in this repo
+Note that we need to disable path conversion because on windows, git bash considers
+'//' to be an escape sequence translating to '/'.
+  $ MSYS_NO_PATHCONV=1 ${TURBO} --skip-infer build -F '[HEAD]' -F '!//' --dry=json | jq '.packages'
```
mehulkar, line 17: If this env var is just for the `-F '!//` flag on CLI, can you set it globally with this explanation? 

https://github.com/vercel/turbo/blob/cc565e86be3506bc80592813a192217a8cc5b0c1/turborepo-tests/helpers/setup_integration_test.sh

It's very likely that this will come up again, especially as we do more work with root packages (e.g. tracing deps through root package.json)

gsoltis: I [tried it](https://github.com/vercel/turbo/actions/runs/9069220393/job/24918724152?pr=8132#step:8:941), and it appears we are relying on this behavior in other parts of our scripts. It might be possible to set it up properly, but I think that's beyond the scope of this particular PR.

PR #8130, opened 2024-05-13
refactor memory tracking

sokra:
### Description

refactor memory for tracing
track memory for tasks


arlyon (approved):

crates/turbo-tasks-malloc/src/lib.rs
```diff
@@ -4,6 +4,33 @@ use std::alloc::{GlobalAlloc, Layout};
 
 use self::counter::{add, flush, get, remove};
 
+#[derive(Default, Clone)]
+pub struct AllocationInfo {
+    pub allocations: usize,
+    pub deallocations: usize,
+    pub allocation_count: usize,
+    pub deallocation_count: usize,
+}
```
bgw, line 13: Because this refers to thread-local data, and then does arithmetic with it assuming that it's still on the same thread, this struct should be explicitly `!Send`.

```suggestion
#[derive(Default, Clone)]
pub struct AllocationInfo {
    pub allocations: usize,
    pub deallocations: usize,
    pub allocation_count: usize,
    pub deallocation_count: usize,
    _not_send: PhantomData<dyn NotSend>,
}

trait NotSend: Sync {}
```

This is going to cause problems when actually trying to use this inside futures that need to be `Send` though, like in `CaptureFuture`.

crates/turbo-tasks-malloc/src/lib.rs
```diff
@@ -4,6 +4,33 @@ use std::alloc::{GlobalAlloc, Layout};
...
+            && self.deallocation_count == 0
+    }
+
+    pub fn until_now(&self) -> Self {
+        let new = TurboMalloc::allocations();
+        Self {
+            allocations: new.allocations - self.allocations,
+            deallocations: new.deallocations - self.deallocations,
+            allocation_count: new.allocation_count - self.allocation_count,
+            deallocation_count: new.deallocation_count - self.deallocation_count,
```
bgw, line 29: If this isn't marked as `!Send`, these calculations will be wrong, and could underflow.

bgw (commented):
I think the thread local nature of `with_local_counter` is going to cause problems if/when work stealing happens.

crates/turbo-tasks-malloc/src/lib.rs
sokra: It's not "stored" in CaptureFuture, so it's fine

sokra:
> I think the thread local nature of `with_local_counter` is going to cause problems if/when work stealing happens.

It reads the allocation counters before and after every poll, and only keeps the diff. So that should be fine.

sokra:
added not send for allocation counters

PR #8133, opened 2024-05-13
Turbopack: `EcmascriptChunkItemContent`: add react refresh parameter based on module

wbinnssmith:
Rather than reading it from the chunking context, which applies to all modules for a given chunking target, read it from the `EcmascriptOptions` for the module. This ensures that we only make react refresh values available to user code transformed with it.

This also:

- Removes `EcmascriptChunkingContext`, as it’s no longer necessary
- Makes `EcmascriptOptions` a Vc in many more cases

Test Plan: Integration tests


bgw (approved):
Nice, it seems like the removal of `EcmascriptChunkingContext` simplifies things quite a bit, especially in the cases that needed downcasting.

crates/turbopack-mdx/src/lib.rs
```diff
@@ -162,7 +162,7 @@ async fn into_ecmascript_module_asset(
             analyze_types: false,
         }),
         this.transforms,
-        Value::new(Default::default()),
+        EcmascriptOptions::default().cell(),
```
ForsakenHarmony, line 165: This does kinda need refresh

PR #8134, opened 2024-05-14
feat(turbopack-ecmascript): add `sourceURL` to chunk items for better stack traces

ForsakenHarmony:


ForsakenHarmony:
doesn't work without putting everything into an eval unfortunately

PR #8137, opened 2024-05-14
feat(filter): no longer infer scope in filters

chris-olszewski:
### Description

Previously we would infer scope for name filters if there was exactly one matching package.
e.g. `turbo build --filter=ui` would run `@a/ui#build` if there was a `@a/ui` package in the workspace

This is confusing and can result in accidentally breaking filters when a conflicting package is added e.g. adding `@b/ui` would cause *no* tasks to be run along with a zero exit code.

### Testing Instructions

Updated unit test to verify inference no longer works and using the explicit package name still works.

### DIFF
--- a/crates/turborepo-lib/src/run/scope/filter.rs
+++ b/crates/turborepo-lib/src/run/scope/filter.rs
@@ -1024,24 +1024,6 @@ mod test {
         );
     }
 
-    #[test]
-    fn match_multiple_scoped() {
-        let resolver = make_project(
-            &[],
-            &["packages/@foo/bar", "packages/@types/bar"],
-            None,
-            TestChangeDetector::new(&[]),
-        );
-        let packages = resolver
-            .get_filtered_packages(vec![TargetSelector {
-                name_pattern: "bar".to_string(),
-                ..Default::default()
-            }])
-            .unwrap();
-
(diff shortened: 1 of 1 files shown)


NicholasLYang (approved):

FlorianWendelborn:
Can this be put behind a flag or something? It's super annoying to have to type the namespace all day. This isn't an issue unless you add the same package name twice, which is a severe skill issue anyway. Why make things unnecessarily hard and annoying to type for the 99.9% of normal users?

mctrafik:
+1. This is a degradation in user experience. On conflict, the command should run on `all the matches` as the obvious behavior. This would be the expected fix. I.e. if I have 5 apps in the repo, and there are 5 `ui` packages, one for each app, it's a very convenient feature to test them all with `turbo test --filter=ui`. Addition thought is that if it's a private github repo, and the company name is long or is composite, it's very annoying to have to type all of this out, now: `turbo test --filter=@my-company-my-product-my-app/ui`

anthonyshew:
Answered some concerns here: https://github.com/vercel/turbo/discussions/8524#discussioncomment-9811128

mctrafik:
Elaborated that your answer doesn't address my concerns here: https://github.com/vercel/turbo/discussions/8524#discussioncomment-9811575

PR #8142, opened 2024-05-14
feat(filter): error on invalid filters

chris-olszewski:
### Description

This PR changes `--filter` so it now will error on certain malformed filters:
 - Name filters with no globs that don't match any packages e.g. `--filter=fo` instead of `--filter=foo`
 - Directory filters that reference a directory that doesn't exist: e.g. `--filter='./pakcages/*'`

Each commit of the PR can be reviewed on it's own.

### Testing Instructions

Updated existing unit tests. Added additional ones along with an integration test.

### DIFF
--- a/crates/turborepo-lib/src/run/scope/filter.rs
+++ b/crates/turborepo-lib/src/run/scope/filter.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use tracing::debug;
-use turbopath::{AbsoluteSystemPath, AnchoredSystemPathBuf};
+use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf, AnchoredSystemPathBuf};
 use turborepo_repository::{
     change_mapper::ChangeMapError,
     package_graph::{self, PackageGraph, PackageName},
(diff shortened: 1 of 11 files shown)


crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -735,7 +751,7 @@ mod test {
             change_detector,
         );
 
-        resolver
+        (temp_folder, resolver)
```
chris-olszewski, line 754: It is *very* important we return the `TempDir` variable as when it gets dropped/out of scope it marks a folder ready for cleanup which leads to non-deterministic tests.

mehulkar: put this in a code comment instead?

turborepo-tests/integration/tests/dry-json/monorepo-no-changes.t
```diff
@@ -2,13 +2,17 @@ Setup
   $ . ${TESTDIR}/../../../helpers/setup_integration_test.sh
 
 # Save JSON to tmp file so we don't need to keep re-running the build
-  $ ${TURBO} run build --dry=json --filter=main > tmpjson.log
+  $ ${TURBO} run build --dry=json --filter='[main]' > tmpjson.log
```
mehulkar, line 5: I think I wrote these tests? but I am not sure if even this is valid usage of `--filter`

mehulkar (approved):
Looks good, reviewed by commit. What happens if `-F ./packages/*` is given and `packages` dir exists, but doesn't have any subdirectories? Do we care about erroring on that case?

turborepo-tests/integration/tests/dry-json/monorepo-no-changes.t
chris-olszewski: I believe you wanted to write a filter for changes since main, but without the brackets it was just a filter for the non-existent `main` package.

zouguangxian:
Is it possible to support run build on a package only if there are changes between the main branch and HEAD, it is very helpful in the Github workflow actions.

```bash
turbo run build --filter='core[main...HEAD]'
```

it works with Turbo v1, but it will throw errors with Turbo v2.

PR #8149, opened 2024-05-15
refactor(turborepo): Rename `outputMode` to `outputLogs` in `turbo.json`

NicholasLYang:
### Description

Changed `outputMode` to `outputLogs` in `turbo.json` to match the flag name. Also wrote a code mod to migrate `turbo.json` config automatically.

### Testing Instructions

Existing tests are migrated. Also wrote tests for code mod


docs/public/schema.json
```diff
@@ -1 +1,252 @@
...
\ No newline at end of file
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
chris-olszewski, line 2: Minor thing, but if we're going to format a codegen file we should do the formatting as part of the codegen.

packages/turbo-codemod/src/transforms/rename-output-mode.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  // If `turbo` key is detected in package.json, require user to run the other codemod first.
+  const packageJsonPath = path.join(root, "package.json");
+  // package.json should always exist, but if it doesn't, it would be a silly place to blow up this codemod
+  let packageJSON = {};
+
+  try {
+    packageJSON = readJsonSync(packageJsonPath) as PackageJson;
+  } catch (e) {
+    // readJSONSync probably failed because the file doesn't exist
```
chris-olszewski, line 46: I think we would want to abort if we are missing/can't read the root `package.json`

chris-olszewski (approved):
LFG

packages/turbo-codemod/src/transforms/rename-output-mode.ts
NicholasLYang: I copied this logic from the outputs codemod. I'm down to change it though

docs/public/schema.json
NicholasLYang: Ah yeah this got accidentally formatted. Reverting

runspired:
> Also wrote a code mod to migrate turbo.json config automatically.

where does one find this codemod?

PR #8150, opened 2024-05-15
chore!(create-turbo): remove old pm arg

tknickman:
### Description

We replaced this with an option instead of an arg, but cleaning up here!


chris-olszewski (approved):
Good catch!

PR #8153, opened 2024-05-15
Trace Server add focus mode

sokra:
### Description

When appending a `!` to the search it shows only these spans and not the whole tree


timneutkens:
Maybe use the `"` quote wrapped case instead, that's well known by many tools i.e. Google etc.

timneutkens:
Actually maybe we can make it a value we can pass instead? That way I can add a UI element for it, similar to what VS Code does for search

kdy1 (approved):

bgw (approved):

sokra:
> Actually maybe we can make it a value we can pass instead? That way I can add a UI element for it, similar to what VS Code does for search

Yes, you can add that as follow-up

PR #8154, opened 2024-05-15
Improve Garbage Collection

sokra:
### Description

Simplify and improve GC

This improves the GC queue.

The job of the GC queue is to find tasks that should be garbage collected. There are three factors which influence that:

* age of the task: Time since last access.
* memory usage of the task
* compute duration of the task: CPU time spend to compute the task.

Memory usage and compute duration combine into a GC priority by calculating: `(memory_usage + C1) / (compute_duration + C2)`. C1 and C2 and constants to fine tune the priority.

The age of the task is constantly changing so a different scheme is used here:

Every task has a generation in which is was last accessed.
The generation is increased every 100,000 tasks.

We accumulate tasks in the current generation in a concurrent queue. Once 100,000 tasks are reached (atomic counter), we increase the generation and pop 100,000 tasks from the queue into an `OldGeneration`. These old generations are stored in another queue. No storing is apply so far. These are just lists of task ids.

Once we need to perform GC, we pop the oldest old generation from the queue, filter out all tasks that are in a higher generation (they have been accessed in the meantime), and sort the list by GC priority.
Then we take the 30% top tasks and garbage collect them.
Then remaining tasks are pushed to the front of the queue again, intermixed with other tasks into existing old generations until we reach a maximum of 200,000 tasks in a generation item. In that case the generation item is split into two items.


crates/turbo-tasks-memory/src/gc.rs
```diff
@@ -1,224 +1,331 @@
...
+    }
+
+    pub(crate) fn execution_completed(
+        &mut self,
+        duration: Duration,
+        memory_usage: usize,
+        generation: u32,
+    ) {
+        self.generation = generation;
+        self.memory_per_time = ((memory_usage + TASK_BASE_MEMORY_USAGE) as u64
```
bgw, line 102: Not something that needs to change, but a couple observation:

- This seems like it could also be stored as an "order of magnitude" type value, similar to `ExpMillisDuration`. That would decrease the risk of accidental saturation of the `u16` type (potentially due to a bad selection of units), and might mean we could also use a `u8` instead of a `u16` (though admittedly you're not actually saving anything after memory alignment).

- If you wanted, you could probably just store this as a u32 with no extra cost, assuming `GcTaskState` will be aligned to 64 bits, though that's admittedly a fragile assumption (e.g. if you want to add more fields later).

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -320,18 +327,29 @@ impl Backend for MemoryBackend {
...
+                memory_usage,
+                generation,
+                stateful,
+                self,
+                turbo_tasks,
+            )
         });
         if !reexecute {
-            self.run_gc(false, turbo_tasks);
+            // self.run_gc(false, turbo_tasks);
```
bgw, line 350: Is this meant to be commented out?

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -118,54 +121,58 @@ impl MemoryBackend {
...
     }
 
-    pub fn on_task_might_become_inactive(&self, task: TaskId) {
-        if let Some(gc_queue) = &self.gc_queue {
-            gc_queue.task_might_become_inactive(task);
-        }
-    }
-
-    pub fn on_task_flagged_inactive(&self, task: TaskId, compute_duration: Duration) {
+    pub fn run_gc(
```
bgw, line 124: It's not entirely clear to me what the return value here is supposed to represent. A comment might be nice (or maybe an enum with a clear type name instead of a `bool`?)

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -118,54 +121,58 @@ impl MemoryBackend {
...
+        &self,
+        idle: bool,
+        _turbo_tasks: &dyn TurboTasksBackendApi<MemoryBackend>,
```
bgw, line 127: Can this argument be removed?

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -118,54 +121,58 @@ impl MemoryBackend {
...
-            if usage < target {
-                if idle {
-                    // Always run propagation when idle
-                    gc_queue.run_gc(0, self, turbo_tasks);
+                let usage = turbo_tasks_malloc::TurboMalloc::memory_usage();
+                let target = if idle {
+                    mem_limit * 3 / 4
+                } else {
+                    mem_limit * 7 / 8
+                };
```
bgw, line 139: It might be nice to move these potentially tunable values into constants declared next to the other ones, like `PERCENTAGE_TO_COLLECT`.

bgw (commented):
I didn't get through the whole review, but I left a few comments: some suggestions, some questions, and some observations.

crates/turbo-tasks-memory/src/gc.rs
sokra: It only run once into the u16 limit for a task that allocated 134MB in 43µs^^

crates/turbo-tasks-memory/src/gc.rs
```diff
@@ -1,224 +1,331 @@
...
-    pub no_gc_needed: usize,
-    /// How many tasks were checked but were in a state where no action could be
-    /// taken.
-    pub no_gc_possible: usize,
+#[derive(Debug, Default, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
+pub struct GcPriority {
+    // Memory usage devided by compute duration. Specifies how efficient garbage collection would
+    // be with this task. Higher memory usage and lower compute duration makes it more likely to be
+    // garbage collected.
+    memory_per_time: u16,
```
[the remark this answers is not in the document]
sokra: It's self CPU time

bgw (approved):
Because this is such a huge change, it would be nice if we could include some high-level benchmarks alongside it to make sure we're not significantly regressing on peak memory usage or CPU.

---

This all seems a bit different from a generational GC as I'm familiar with (in the context of tracing GCs). Maybe the terminology "generation" being used in a different context is what's confusing to me, though it's not wrong as a description of what this is. This seems more like a bucketed LRU cache.

In my understanding of generational GC:
- Objects in the older generation are less likely to be collected / iterated over. This seems like the inverse of that.
- Older objects get moved to a separate tier (possibly one of many tiers) of "survivors" that are less frequently traversed. We don't have any sort of logic for that.

That leads me to a few thoughts on potential future ways to improve this:

- Consider a "segmented LRU", which shares some similarities to generational tracing GCs: https://memcached.org/blog/modern-lru/
- There are [edge-cases where LRU cache evictions can severely degrade if the cache is too small to contain all frequently accessed items](https://stackoverflow.com/questions/71166908/how-is-the-lru-eviction-policy-less-efficient-than-the-random-policy-in-this-cor). If we think this is potential concern, we could add some amount of randomization to cache eviction, which would lead to a more graceful degradation.

crates/turbo-tasks-memory/src/gc.rs
```diff
@@ -1,224 +1,276 @@
...
-        self.queue.insert(task, value);
+    pub fn task_executed(&self, task: TaskId) -> u32 {
+        self.add_task(task)
     }
 
-    /// Notify the GC queue that a task might become inactive.
-    pub fn task_might_become_inactive(&self, task: TaskId) {
-        let _ = self.inactive_propagate_queue.push(task);
+    /// Notify the GC queue that a task has been accessed.
+    pub fn task_accessed(&self, task: TaskId) -> u32 {
```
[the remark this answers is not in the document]
sokra: It's called from `Task::with_cell_mut`

crates/turbo-tasks-memory/src/gc.rs
```diff
@@ -1,224 +1,276 @@
...
+    }
+
+    pub(crate) fn on_read(&mut self, generation: u32) -> bool {
+        if self.generation < generation {
+            self.generation = generation;
+            true
+        } else {
+            false
+        }
+    }
```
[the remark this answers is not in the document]
sokra: It's called from `Task::with_cell_mut`

sokra:
> generational GC:
> 
> * Objects in the older generation are less likely to be collected / iterated over. This seems like the inverse of that.

Yep that's true. We don't write a real GC in the sense of collecting unreferenced memory, as "unreferenced" doesn't exist in our system. We write a cache, where any cache entry might be access anytime in future, but also when can evict any cache entry anytime without hurting correctness (but only performance).

So we want the opposite of a generational GC. We want older cache entries to be more likely to be evicted. But we also want the memory usage and compute time of cache entries to influence the evicting behavior.

So we bucket cache entries into buckets of (currently) 100,000 items. We call them generations. When under memory pressure we start processing the oldest generation. We select 30% of cache entries that have the highest GC priority and GC collect them. The 70% remaining cache entries are pushed into the bucket at of the freshest generation. So they are intermixed with that generation. That will reconsider these entries for GC when we cycled through all old generations.

There is a maximum of (currently) 200,000 items per bucket. Buckets are split evenly into two buckets when they get too full.

Using buckets for age is kind of nice as it avoids having to include age (which is continuously increasing) into the priority and it also avoids having to sort all cache items into a very big priority queue. We basically don't have to sort anything until GC is invoked and then we only have to sort a "small" bucket of items.

PR #8157, opened 2024-05-16
Rename `pipeline` to `tasks`.

anthonyshew:
### Description

Renaming the `pipeline` key to `tasks` in turbo.json.

We went this key to better express what Turborepo does with the configuration that is provided in this key. We've seen with users that the term "pipeline" can be confusing for how `turbo` _really_ executes the task graph.

In reality, the keys in the `pipeline` object are really the list of _tasks_ registered to `turbo` - so let's call it that.

### Non-goals

This PR does not include update:
-  The examples in the repo
- JS packages
- `eslint-plugin-turbo`

These updates will be needed in subsequent PRs.

### Testing Instructions

Hopefully CI will do its job here - but, of course, would appreciate a review of my changes to make sure I'm not missing anything my untrained eye may be missing.

CLOSES TURBO-3225


crates/turbopack-browser/src/ecmascript/runtime.rs
```diff
@@ -6,8 +6,8 @@ use turbopack_core::{
     ident::AssetIdent,
 };
 use turbopack_ecmascript::chunk::{
-    EcmascriptChunk, EcmascriptChunkPlaceables, EcmascriptChunkRuntime,
-    EcmascriptChunkRuntimeContent, ChunkingContext,
+    ChunkingContext, EcmascriptChunk, EcmascriptChunkPlaceables, EcmascriptChunkRuntime,
```
chris-olszewski, line 9: Could we drop this change to avoid running the turbopack tests which are failing?

PR #8158, opened 2024-05-16
fix rustc bug in turbopack-trace-server

arlyon:
### Description

unfortunately there is a rustc bug that crashes when the typechecking in run in this crate during compiles for next-swc when using `Either<impl Iterator, impl Iterator>`. Solution is to use a Boxed iterator

### Testing Instructions

Existing tests OK


timneutkens (approved):

sokra (approved):

bgw:
Nice! Hopefully if we can keep an eye on https://github.com/rust-lang/rust/issues/124891 we can revert this once fixed.

PR #8159, opened 2024-05-16
chore: move react from devDependency to dependency in design system starter template

ghdtjgus76:
### Description

I mentioned this in discussions, in the design system starter template, react was specified as a devDependency rather than a dependency. 
It needs to be moved to dependencies.

https://github.com/vercel/turbo/discussions/8156

### Testing Instructions


mehulkar (commented):
I _think_ this is correct, but @anthonyshew is the right person to ask. I'd almost want `react` to be a peerDep in a shared package, not a real dep.

anthonyshew (approved):
Thanks! Appreciate the attention to detail. 🙌

anthonyshew (changes requested):
Can you update the lockfile to reflect this change? Can merge after that!

ghdtjgus76:
I reflected the changes!
Thank you for the comment. 😀

PR #8163, opened 2024-05-16
feat: --only now applies to package dependencies

chris-olszewski:
### Description

Change `--only` so it behaves in a more sensible way. It still shouldn't be widely used, but now it at least does what it says it does.

Previously `--only` would still follow package dependencies resulting additional tasks getting run that weren't expected.
e.g. for task definition `"test": {"dependsOn": ["build", "^test"]}` and package `a` depending on `b`, then `turbo test --filter=a --only` would result in both `a#test` and `b#test` being run. With this PR now only `a#test` will be run.

I changed the `--only` logic so now it will limit tasks in the graph to exactly those that are in the product of the packages implied by `--filter` and the tasks specified in the run args. This should make `--only` a far more sensible flag.

### Testing Instructions

Added unit tests for testing the trimming of package dependencies and task id style dependencies e.g. `"dependsOn": ["a#test"]`

### DIFF
--- a/crates/turborepo-lib/src/engine/builder.rs
+++ b/crates/turborepo-lib/src/engine/builder.rs
@@ -153,6 +153,28 @@ impl<'a> EngineBuilder<'a> {
         self
     }
 
+    // Returns the set of allowed tasks that can be run if --only is used
+    // The set is exactly the product of the packages in filter and tasks specified
+    // by CLI
+    fn allowed_tasks(&self) -> Option<HashSet<TaskId<'static>>> {
+        if self.tasks_only {
+            Some(
+                self.workspaces
+                    .iter()
+                    .cartesian_product(self.tasks.iter())
+                    .map(|(package, task_name)| {
+                        task_name
+                            .task_id()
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/engine/builder.rs
```diff
@@ -153,6 +153,28 @@ impl<'a> EngineBuilder<'a> {
...
+                    .map(|(package, task_name)| {
+                        task_name
+                            .task_id()
+                            .unwrap_or(TaskId::new(package.as_ref(), task_name.task()))
+                            .into_owned()
+                    })
+                    .collect(),
+            )
+        } else {
+            None
```
mehulkar, line 174: Why is this none? shouldn't it be all?

mehulkar (approved):
not blocking, but it would be nice to have a prysk test for top level flags and their happy path behaviors.

crates/turborepo-lib/src/engine/builder.rs
mehulkar: never mind, I saw the rest of the code and saw how allowed_tasks is used. the name of the method just threw me off

PR #8165, opened 2024-05-16
Turbopack: Add `BeforeResolve` plugin type

wbinnssmith:
This adds a new plugin type for replacing resolution _before_ it occurs. This allows plugin authors to intercept requests for modules in place of the typical resolution strategy.

This is used to replace `ImportMapping` replacers for `@next/font/local` in an upcoming Next.js PR


crates/turbopack-resolve/src/resolve_options_context.rs
```diff
@@ -72,6 +72,7 @@ pub struct ResolveOptionsContext {
     /// A list of plugins which get applied before (in the future) and after
```
kwonoj, line 72: probably need to adjust comment bit, also we may want to rename `plugins` in this case?

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1386,7 +1406,24 @@ pub async fn url_resolve(
...
-async fn handle_resolve_plugins(
+async fn handle_before_resolve_plugins(
+    lookup_path: Vc<FileSystemPath>,
+    reference_type: Value<ReferenceType>,
+    request: Vc<Request>,
+    options: Vc<ResolveOptions>,
+) -> Result<Option<Vc<ResolveResult>>> {
+    for plugin in &options.await?.before_resolve_plugins {
+        if let Some(result) = *plugin
+            .before_resolve(lookup_path, reference_type.clone(), request)
```
sokra, line 1417: That might have a negative performance effect, since it calls the plugin for every resolving in the application. So something like with the resolve plugins would be beneficial: `before_resolve_condition`

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -2553,7 +2619,18 @@ pub async fn handle_resolve_error(
                 .cell()
                 .emit();
             }
-            result
+
+            if let Some(error_string) = result.await?.get_first_error() {
```
sokra, line 2623: I think it should not do that. It should just return the result even if that's a resolve result with runtime error.

The issue should be emitted earlier...

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -66,6 +67,7 @@ pub enum ModuleResolveResultItem {
     OutputAsset(Vc<Box<dyn OutputAsset>>),
     External(String, ExternalType),
     Ignore,
+    Error(Vc<StyledString>),
```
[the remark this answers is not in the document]
wbinnssmith: Shouldn't this be a build error -- we wouldn't want to produce something that is runnable if this is the case. I think that's what happens with Webpack.

kwonoj (approved):

PR #8173, opened 2024-05-17
feat: factor engines into global cache key

chris-olszewski:
### Description

Changing the `engines` field in the root `package.json` can affect the execution of tasks so it should affect the global cache.

This PR adds the `engines` struct to the global cache key.

### Testing Instructions

Added new integration test verifying that changing `engines` results in a cache miss.

### DIFF
--- a/crates/turborepo-lib/src/hash/mod.rs
+++ b/crates/turborepo-lib/src/hash/mod.rs
@@ -432,7 +432,7 @@ mod test {
             framework_inference: true,
         };
 
-        assert_eq!(global_hash.hash(), "08ed059da61137ab");
+        assert_eq!(global_hash.hash(), "215cf70acde752ce");
     }
 
     #[test_case(vec![], "459c029558afe716" ; "empty")]
--- a/crates/turborepo-lib/src/hash/mod.rs
+++ b/crates/turborepo-lib/src/hash/mod.rs
@@ -66,6 +66,7 @@ pub struct GlobalHashable<'a> {
     pub global_file_hash_map: &'a HashMap<turbopath::RelativeUnixPathBuf, String>,
     // These are None in single package mode
     pub root_external_dependencies_hash: Option<&'a str>,
+    pub root_internal_dependencies_hash: Option<&'a str>,
     pub engines: HashMap<&'a str, &'a str>,
     pub env: &'a [String],
     pub resolved_env_vars: EnvironmentVariablePairs,
@@ -424,6 +425,7 @@ mod test {
             global_cache_key: "global_cache_key",
             global_file_hash_map: &global_file_hash_map,
             root_external_dependencies_hash: Some("0000000000000000"),
+            root_internal_dependencies_hash: Some("0000000000000001"),
             engines: Default::default(),
             env: &["env".to_string()],
             resolved_env_vars: vec![],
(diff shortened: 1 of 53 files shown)


mehulkar (commented):
All of the root package.json is in the global cache inputs though, right? Is this double-including it?

chris-olszewski:
> All of the root package.json is in the global cache inputs though, right? Is this double-including it?

No, we don't include the file hash of the root `package.json`. Previously the only information that went in the global hash were external package dependencies. With 2.0 it will now be `engines`, and all of the various dependencies

mehulkar (commented):
Could you rebase this so I can see the diff again? The explanation makes sense, I was thinking of aa019da1428d3fc6c61740b7639c02fea8e7ce92, but that was for package-specific `package.json`s that we try to always include.

crates/turborepo-lib/src/hash/mod.rs
```diff
@@ -407,14 +426,15 @@ mod test {
             global_file_hash_map: &global_file_hash_map,
             root_external_dependencies_hash: Some("0000000000000000"),
             root_internal_dependencies_hash: Some("0000000000000001"),
+            engines: Default::default(),
```
mehulkar, line 429: what's the default?

turborepo-tests/integration/tests/run/single-package/with-deps-run.t
```diff
@@ -21,16 +21,22 @@ Check
   Cached:    0 cached, 2 total
     Time:\s*[\.0-9]+m?s  (re)
   
+<<<<<<< HEAD
+=======
+>>>>>>> b668d5abb3 (chore: remove task dotEnv field)
+<<<<<<< HEAD
+=======
+>>>>>>> b668d5abb3 (chore: remove task dotEnv field)
```
mehulkar, line 29:
```suggestion
```

turborepo-tests/integration/tests/run/single-package/with-deps-run.t
```diff
@@ -41,18 +47,26 @@ Run a second time, verify caching works because there is a config
   Cached:    2 cached, 2 total
     Time:\s*[\.0-9]+m?s >>> FULL TURBO (re)
   
+<<<<<<< HEAD
+=======
+>>>>>>> b668d5abb3 (chore: remove task dotEnv field)
+<<<<<<< HEAD
+=======
+>>>>>>> b668d5abb3 (chore: remove task dotEnv field)
```
mehulkar, line 55:
```suggestion
```

turborepo-tests/integration/tests/run/single-package/with-deps-run.t
```diff
@@ -41,18 +47,26 @@ Run a second time, verify caching works because there is a config
...
+  test: cache hit, suppressing logs cb5839f7284aa5f3
   
    Tasks:    2 successful, 2 total
   Cached:    2 cached, 2 total
     Time:\s*[\.0-9]+m?s >>> FULL TURBO (re)
   
-
+<<<<<<< HEAD
+=======
+>>>>>>> b668d5abb3 (chore: remove task dotEnv field)
```
mehulkar, line 69:
```suggestion
```

mehulkar (approved):
Some merge conflicts, but looks good otherwise!

crates/turborepo-lib/src/hash/mod.rs
chris-olszewski: `{}`

PR #8180, opened 2024-05-20
feat: add doublestar to exact dir paths

chris-olszewski:
### Description

If a user provides a glob that points to a directory e.g. `dist` or `dist/` then we will add a trailing double star so that the directory contents get captured instead of just the directory entry itself. 

### Testing Instructions

Added unit tests for adding doublestar
Added integration test for verifying that `src/` and `dist` for task inputs/outputs get treated as `src/**` and `dist/**` respectively.


crates/turborepo-globwalk/src/lib.rs
```diff
@@ -102,25 +114,12 @@ fn preprocess_paths_and_globs(
...
-        if split.ends_with('/') {
-            if split.ends_with("**/") {
-                exclude_paths.push(split[..split.len() - 1].to_string());
-            } else {
-                exclude_paths.push(format!("{}**", split));
-            }
-        } else if split.ends_with("/**") {
-            exclude_paths.push(split);
-        } else {
-            // Match Go globby behavior. If the glob doesn't already end in /**, add it
```
anthonyshew, line 117: Love seeing an awkward Go behavior get dropped. 🔥

anthonyshew (approved):
😌

PR #8182, opened 2024-05-21
chore: make env mode strict by default

chris-olszewski:
### Description

Commandeered from #8151

Changes default env mode to strict and remove the "infer" option

### Testing Instructions

Existing test suite

(Currently some Windows integration tests are failing due to a missing env var, opening for review while I work on this)


crates/turborepo-lib/src/task_hash.rs
```diff
@@ -448,22 +449,24 @@ impl<'a> TaskHasher<'a> {
...
     ) -> Result<EnvironmentVariableMap, Error> {
         match task_env_mode {
-            ResolvedEnvMode::Strict => {
+            EnvMode::Strict => {
                 let mut pass_through_env = EnvironmentVariableMap::default();
                 let default_env_var_pass_through_map =
                     self.env_at_execution_start.from_wildcards(&[
                         "SHELL",
                         // Command Prompt casing of env variables
+                        "APPDATA",
```
chris-olszewski, line 463: This is necessary due to older versions of `pnpm` not having https://github.com/pnpm/npm-conf/pull/10 which results in `pnpm` crashing if `APPDATA` isn't defined.

Technically we could have users define this instead of baking it into `turbo` itself, but considering the popularity of `pnpm` and the amount of digging it took to find that PR, it seems nice to include it.

mehulkar (approved):

andrevenancio:
What was the motivation to change how environment variables are handled? I'm failing to understand based on your changelog's on turbo 2.0 and this PR (which you link from the changelog) WHY was this necessary. Why am I allowed to build and the environment variables aren't provided by default. My app if failing on vercel.com and the environment variables are there. Why do I need to be explicit on my turbo.json file to include variable A B or C? If the env variables are defined on my project dashboard then they all should be available.

I'm running turborepo 2.0 and my build isn't working anymore. even with --env-mode=loose.

Would be great to read somewhere why this change was necessary. 
Thank you

Wizzel1:
> What was the motivation to change how environment variables are handled? I'm failing to understand based on your changelog's on turbo 2.0 and this PR (which you link from the changelog) WHY was this necessary. Why am I allowed to build and the environment variables aren't provided by default. My app if failing on vercel.com and the environment variables are there. Why do I need to be explicit on my turbo.json file to include variable A B or C? If the env variables are defined on my project dashboard then they all should be available.
> 
> I'm running turborepo 2.0 and my build isn't working anymore. even with --env-mode=loose.
> 
> Would be great to read somewhere why this change was necessary. Thank you

Facing the same problem. After migration my environment variables are missing. Did you find a solution?

yashsway:
> > What was the motivation to change how environment variables are handled? I'm failing to understand based on your changelog's on turbo 2.0 and this PR (which you link from the changelog) WHY was this necessary. Why am I allowed to build and the environment variables aren't provided by default. My app if failing on vercel.com and the environment variables are there. Why do I need to be explicit on my turbo.json file to include variable A B or C? If the env variables are defined on my project dashboard then they all should be available.
> > I'm running turborepo 2.0 and my build isn't working anymore. even with --env-mode=loose.
> > Would be great to read somewhere why this change was necessary. Thank you
> 
> Facing the same problem. After migration my environment variables are missing. Did you find a solution?

Have the exact same issue as both of you @Wizzel1 @andrevenancio ! I've tried using `env`, `globalEnv`, `globalPassThroughEnv`, and `--env-mode=loose` directly and none of it is letting my CircleCI environment variables through. 😢

PR #8184, opened 2024-05-21
Removed redundant @storybook/addon-docs dev dependency

NickE27:
@storybook/addon-docs was included as a dev dependency, and imported into the config. This is now part of [addon-essentials](https://storybook.js.org/docs/essentials) and therefore it's inclusion here is redundant.

With addon-docs as a separate, additional dependency, storybook would throw `TypeError: Failed to fetch dynamically imported module` when importing mdx files, making it impossible to add custom documentation.


anthonyshew (approved):
Thanks for the fix!

PR #8186, opened 2024-05-21
feat(codemod): add missing package names

tknickman:
### Description

Codemod to add missing name (or fix duplicates)

### DIFF
--- a/packages/turbo-codemod/__tests__/add-package-names.test.ts
+++ b/packages/turbo-codemod/__tests__/add-package-names.test.ts
@@ -1,0 +1,117 @@
+import { setupTestFixtures } from "@turbo/test-utils";
+import { transformer } from "../src/transforms/add-package-names";
+
+describe("add-package-names", () => {
+  const { useFixture } = setupTestFixtures({
+    directory: __dirname,
+    test: "add-package-names",
+  });
+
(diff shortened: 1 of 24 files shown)


packages/turbo-codemod/__tests__/add-package-names.test.ts
```diff
@@ -0,0 +1,117 @@
...
+      const pkgJson = readJson<{ name: string }>(
+        `packages/${pkg}/package.json`
+      );
+      expect(pkgJson?.name).toBeDefined();
+      expect(names.has(pkgJson?.name)).toBe(false);
+      names.add(pkgJson?.name);
+    }
+  });
+
+  test("duplicate names", async () => {
```
chris-olszewski, line 52: FWIW we've never supported duplicate package names in the workspace

chris-olszewski (approved):

packages/turbo-codemod/__tests__/add-package-names.test.ts
tknickman: Yea - this was an easy add since if we're generating new names we need to ensure they're unique, and if we're ensuring new ones are unique we should ensure they're all unique. 

There still might be an edge case or two here, but this should be good for most cases.

PR #8188, opened 2024-05-21
feat(codemod): change migrate behavior for majors

tknickman:
Do not merge yet - we need to make sure that codemods that are _not_ idempotent are handled correctly. I think the only one that isn't is the default outputs code mod, so we need to identify if we need to run it.


anthonyshew:
We used you as inspiration, Turbotom. Thanks as always.

PR #8198, opened 2024-05-22
Update getting started docs to disambiguate gitignore pattern

gnowland:
### Description

Update getting started docs to disambiguate gitignore pattern. The existing docs leads you to believe that the glob pattern is `+ .turbo` which *includes* turbo files in your git commits, instead of `.turbo`, which hides them. Removed the `+` to remove this ambiguity. In the Docs UI it isn't obvious that what was shown was a "diff", and that the `+` wasn't part of the pattern, this should remove any doubt as to the intention.

### Testing Instructions

N/A


anthonyshew:
We did a huge rewrite of the docs and this was a part of it. Thanks for wanting to contribute!

PR #8201, opened 2024-05-23
chore: remove --ignore flag

chris-olszewski:
### Description

Removes the `--ignore` flag as it's interaction with other `--filter` flags is hard to explain.

If we want to bring it back, then it should be put it into a configuration file.

I removed the command line flag and how it feeds into the package filtering, but left the underlying implementation of change detector as-is. This should make it easy to bring back this behavior via a configuration file.

### Testing Instructions

Existing tests


mehulkar (approved):

tknickman (approved):
We need to remove it from docs as well (Or is that already handled?)

chris-olszewski:
> We need to remove it from docs as well (Or is that already handled?)

Added a ticket for this in linear

PR #8202, opened 2024-05-23
feat: all dependencies of root package contribute to global hash

chris-olszewski:
### Description

With this PR we will now factor in all root dependency changes. Not just external packages.

Internal packages are handled by hashing all of the files contained in the package directory that aren't gitignore'd.

This does have performance implications as we can end up globwalking these directories multiple times and hashing the files multiple times if they end up as task inputs. This will be addressed in a future PR.

### Testing Instructions

Added integration test that displays new behavior

### DIFF
--- a/crates/turborepo-lib/src/run/summary/global_hash.rs
+++ b/crates/turborepo-lib/src/run/summary/global_hash.rs
@@ -43,6 +43,7 @@ impl<'a> TryFrom<GlobalHashableInputs<'a>> for GlobalHashSummary<'a> {
             global_cache_key,
             global_file_hash_map,
             root_external_dependencies_hash,
+            root_internal_dependencies_hash,
             env,
             resolved_env_vars,
             pass_through_env,
--- a/crates/turborepo-lib/src/run/summary/global_hash.rs
+++ b/crates/turborepo-lib/src/run/summary/global_hash.rs
@@ -31,6 +31,7 @@ pub struct GlobalHashSummary<'a> {
     pub root_key: &'static str,
     pub files: BTreeMap<RelativeUnixPathBuf, String>,
     pub hash_of_external_dependencies: &'a str,
+    pub hash_of_internal_dependencies: &'a str,
     pub environment_variables: GlobalEnvVarSummary<'a>,
 }
 
@@ -65,6 +66,7 @@ impl<'a> TryFrom<GlobalHashableInputs<'a>> for GlobalHashSummary<'a> {
             files: global_file_hash_map.into_iter().collect(),
             // This can be empty in single package mode
             hash_of_external_dependencies: root_external_dependencies_hash.unwrap_or_default(),
+            hash_of_internal_dependencies: root_internal_dependencies_hash.unwrap_or_default(),
             environment_variables: GlobalEnvVarSummary {
                 specified: GlobalEnvConfiguration {
                     env,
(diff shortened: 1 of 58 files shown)


crates/turborepo-lib/src/run/summary/global_hash.rs
```diff
@@ -65,6 +66,7 @@ impl<'a> TryFrom<GlobalHashableInputs<'a>> for GlobalHashSummary<'a> {
             files: global_file_hash_map.into_iter().collect(),
             // This can be empty in single package mode
             hash_of_external_dependencies: root_external_dependencies_hash.unwrap_or_default(),
+            hash_of_internal_dependencies: root_internal_dependencies_hash.unwrap_or_default(),
```
mehulkar, line 69: As Github is annotating, I don't see `root_internal_dependencies_hash` in scope here

turborepo-tests/integration/fixtures/root_deps/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "monorepo",
+  "workspaces": [
+    "apps/**",
+    "packages/**"
+  ],
+  "dependencies": {
+    "util": "*"
+  }
+}
```
mehulkar, line 10: I can't remember if our integration test fixtures generally use `packageManager` or not, but that might be useful since I know `*` has some weird behavior in different npm versions, especially with `workspaces`.

turborepo-tests/integration/tests/run/single-package/with-deps-dry-run.t
```diff
@@ -17,7 +17,7 @@ Check
   Tasks to Run
   build
     Task                           = build\s* (re)
-    Hash                           = 4047a6e65d7dafef
+    Hash                           = 81a933c332d3f388
```
mehulkar, line 20: are all the hashes changing because we're including some kind of "none" value even when there are no root internal deps?

mehulkar (approved):

turborepo-tests/integration/tests/run/single-package/with-deps-dry-run.t
chris-olszewski: Yes, since we're changing the Capnproto structure definition the hashing of it is shifting even when there aren't any internal deps.

turborepo-tests/integration/fixtures/root_deps/package.json
chris-olszewski: > I can't remember if our integration test fixtures generally use packageManager or not, but that might be useful since I know * has some weird behavior in different npm versions, especially with workspaces

Good callout. With package manager requirements we now force usage of corepack for all test fixture defaulting to `npm` if one isn't specified in the fixture. Will open a ticket to make this explicit

dipunm:
Hi all, can someone help me understand this change better? I came here from https://turbo.build/repo/docs/crafting-your-repository/upgrading#update-turbo-run-commands

I tested the following things:

- I modified .md files in the `.changeset` folder in workspace root.
- I changed ts and sh files in a `build-scripts` folder in workspace root.
- Made change to an .md file at the root of the project.
- Made change to .npmrc file at the root of the project.
- Added a js file to the workspace root.

Each time, I ran `turbo lint test` and the cache was not cleared. Is this expected or not?

mehulkar:
@dipunm I would open a discussion referencing this. Comments on merged/closed PRs are probably not going to get a response

PR #8207, opened 2024-05-23
chore: Remove unused `base16` dependency and dead `hex` wrapper function

bgw:
### Description

Noticed these dependencies/functions were unused.

For the most part, it's better to do these conversions with the built-in `format!()`.

Parsing base16 could make sense for this as a dependency, but we're not doing that.

### Testing Instructions

```
cargo check
```

Also grepped through next.js to check that we're not using it there.


arlyon (approved):

bgw:
### Merge activity

* **May 24, 3:41 AM EDT**: @bgw merged this pull request with [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/8207).

PR #8211, opened 2024-05-24
Add option for inlining typeof window

timneutkens:
### Description

In Next.js with webpack we have a feature to inline `typeof window` in server/client environments. That allows e.g.

```
"use client"
if(typeof window === 'undefined') {
  import('server-only-package')
}
```

Today with Turbopack above code causes `server-only-package` to still be loaded in the browser compilation. Which causes issues like https://github.com/vercel/next.js/issues/66058 where the `if`/`else` condition is used to use different modules for server and client.

This change is required to pass the test in https://github.com/vercel/next.js/issues/66058

Pushing up the Next.js changes in that PR.

### DIFF
--- a/crates/turbopack-ecmascript/src/transform/mod.rs
+++ b/crates/turbopack-ecmascript/src/transform/mod.rs
@@ -3,13 +3,15 @@ use std::{fmt::Debug, hash::Hash, sync::Arc};
 use anyhow::Result;
 use async_trait::async_trait;
 use swc_core::{
+    atoms::JsWord,
     base::SwcComments,
-    common::{chain, comments::Comments, util::take::Take, Mark, SourceMap},
+    common::{chain, collections::AHashMap, comments::Comments, util::take::Take, Mark, SourceMap},
     ecma::{
         ast::{Module, ModuleItem, Program, Script},
         preset_env::{self, Targets},
         transforms::{
             base::{feature::FeatureFlag, helpers::inject_helpers, Assumptions},
+            optimization::inline_globals2,
             react::react,
         },
         visit::{FoldWith, VisitMutWith},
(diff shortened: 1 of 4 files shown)


timneutkens:
* **#8211** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8211?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @timneutkens and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

kdy1 (approved):

crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -134,6 +139,18 @@ impl EcmascriptInputTransform {
             ..
         } = ctx;
         match self {
+            EcmascriptInputTransform::GlobalTypeofs { window_value } => {
+                let mut typeofs: AHashMap<JsWord, JsWord> = Default::default();
+                let val = window_value.to_owned();
+                typeofs.insert("window".into(), JsWord::from(val));
```
kdy1, line 145:
```suggestion
                typeofs.insert("window".into(), JsWord::from(&**window_value));
```

You don't need to create a `String` to create `JsWord`

kdy1 (changes requested):

crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -134,6 +139,18 @@ impl EcmascriptInputTransform {
             ..
         } = ctx;
         match self {
+            EcmascriptInputTransform::GlobalTypeofs { window_value } => {
+                let mut typeofs: AHashMap<JsWord, JsWord> = Default::default();
+                let val = window_value.to_owned();
```
kdy1, line 144:
```suggestion

```

PR #8216, opened 2024-05-26
[docs] fix scripts order to match the simulation image

ob1:
### Description

fix scripts order to match the simulation image in the documentation


anthonyshew:
We did a major docs rewrite that included this. Thanks for the mention!

PR #7772, opened 2024-03-20
perf(turbopack): Investigate memory usage of `String`s and `Value<T>`s

kdy1:
### Description

I'll report back.

PR #8221, opened 2024-05-28
perf: Introduce `RcStr`

kdy1:
### Description

If we accept `: String` from a function, invoking the function needs one owned `String` and it means allocation even if the value isn't changed. But with this PR, functions now accept `RcStr`, which is shared, so invoking the same function with the same value does not involve allocation for `String` .

---

I tried using `Arc<String>` in https://github.com/vercel/turbo/pull/7772, but a team member suggested creating a new type so we can replace underlying implementation easily in the future.

I used `ast-grep` with

```yml
id: convert-expr-into
language: rust
rule:
  kind: function_item
  all:
    - has:
        kind: parameters
        pattern: $PARAMS
    - any:
      - has:
          kind: visibility_modifier
          pattern: $VIS
      - not:
          has:
            kind: visibility_modifier

    - any:
      - has:
          kind: function_modifiers
          pattern: $MOD
      - not:
          has:
            kind: function_modifiers
    - has:
        field: return_type
        pattern: $RET
    - has:
        field: body
        pattern: $BODY
    - has:
        kind: identifier
        pattern: $NAME
    - follows:
        kind: attribute_item
        pattern: '#[turbo_tasks::function]'



transform:
  NEW_PARAMS:
    replace:
      replace: 'String'
      by: 'RcStr'
      source: $PARAMS

fix:
  $VIS $MOD fn $NAME $NEW_PARAMS -> $RET $BODY




```

#### `main`:

```
Benchmarking bench_startup/Next.js canary Turbo SSR/1000 modules: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 60.0s. You may wish to increase target time to 65.6s.
bench_startup/Next.js canary Turbo SSR/1000 modules
                        time:   [4.1476 s 4.1625 s 4.1752 s]
Benchmarking bench_startup/Next.js canary Turbo RSC/1000 modules: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 60.0s. You may wish to increase target time to 63.5s.
bench_startup/Next.js canary Turbo RSC/1000 modules
                        time:   [4.0396 s 4.0673 s 4.0956 s]
Benchmarking bench_startup/Next.js canary Turbo RCC/1000 modules: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 60.0s. You may wish to increase target time to 69.7s.
bench_startup/Next.js canary Turbo RCC/1000 modules
                        time:   [4.7222 s 4.7508 s 4.7790 s]
```

#### `pr`:

```
Benchmarking bench_startup/Next.js canary Turbo SSR/1000 modules: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 60.0s. You may wish to increase target time to 66.2s.
bench_startup/Next.js canary Turbo SSR/1000 modules
                        time:   [4.1175 s 4.1310 s 4.1496 s]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
Benchmarking bench_startup/Next.js canary Turbo RSC/1000 modules: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 60.0s. You may wish to increase target time to 63.7s.
bench_startup/Next.js canary Turbo RSC/1000 modules
                        time:   [4.0401 s 4.0714 s 4.1105 s]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
Benchmarking bench_startup/Next.js canary Turbo RCC/1000 modules: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 60.0s. You may wish to increase target time to 70.2s.
bench_startup/Next.js canary Turbo RCC/1000 modules
                        time:   [4.6696 s 4.6900 s 4.7107 s]
Benchmarking bench_startup/Next.js 14 Turbo SSR/1000 modules: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 60.0s. You may wish to increase target time to 63.9s.
```

### DIFF
--- a/crates/turbo-tasks-fs/examples/hash_glob.rs
+++ b/crates/turbo-tasks-fs/examples/hash_glob.rs
@@ -55,7 +55,7 @@ async fn main() -> Result<()> {
 
 #[turbo_tasks::function]
 pub fn empty_string() -> Vc<RcStr> {
-    Vc::cell("".into())
+    Vc::cell(Default::default())
 }
 
 #[turbo_tasks::function]
@@ -99,7 +99,7 @@ async fn hash_file(file_path: Vc<FileSystemPath>) -> Result<Vc<RcStr>> {
         FileContent::Content(file) => hash_content(&mut file.read()),
         FileContent::NotFound => {
             // report error
-            Vc::cell("".into())
+            Vc::cell(Default::default())
         }
     })
 }
--- a/crates/turbo-tasks/src/lib.rs
+++ b/crates/turbo-tasks/src/lib.rs
@@ -60,6 +60,7 @@ pub mod persisted_graph;
 pub mod primitives;
 mod raw_vc;
 mod raw_vc_set;
+mod rcstr;
 mod read_ref;
 pub mod registry;
 pub mod small_duration;
@@ -99,7 +100,7 @@ pub use read_ref::ReadRef;
 pub use state::State;
 pub use task::{
     concrete_task_input::{ConcreteTaskInput, SharedReference, SharedValue},
-    task_input::{RcStr, TaskInput},
+    task_input::TaskInput,
 };
 pub use trait_ref::{IntoTraitRef, TraitRef};
 pub use turbo_tasks_macros::{function, value, value_impl, value_trait, TaskInput};
@@ -109,6 +110,8 @@ pub use vc::{
     Dynamic, TypedForInput, Upcast, ValueDefault, Vc, VcCellNewMode, VcCellSharedMode,
     VcDefaultRead, VcRead, VcTransparentRead, VcValueTrait, VcValueType,
 };
+
+pub use crate::rcstr::RcStr;
 
 pub type TaskIdSet = AutoSet<TaskId, BuildNoHashHasher<TaskId>, 2>;
 
--- a/crates/turbo-tasks/src/primitives.rs
+++ b/crates/turbo-tasks/src/primitives.rs
@@ -35,7 +35,7 @@ fn empty_string_vec() -> Vc<Vec<RcStr>> {
     Vc::cell(Vec::new())
 }
 
-impl Vc<Vec<String>> {
+impl Vc<Vec<RcStr>> {
     #[deprecated(note = "use Default::default() instead")]
     #[inline(always)]
     pub fn empty() -> Vc<Vec<RcStr>> {
--- a/crates/turbo-tasks/src/primitives.rs
+++ b/crates/turbo-tasks/src/primitives.rs
@@ -15,7 +15,7 @@ __turbo_tasks_internal_primitive!(RcStr);
 
 #[turbo_tasks::function]
 fn empty_string() -> Vc<RcStr> {
-    Vc::cell(RcStr::from(""))
+    Vc::cell(RcStr::default())
 }
 
 impl Vc<RcStr> {
--- a/crates/turbo-tasks/src/task/task_input.rs
+++ b/crates/turbo-tasks/src/task/task_input.rs
@@ -15,8 +15,10 @@ use turbo_tasks_hash::{DeterministicHash, DeterministicHasher};
 
 use super::concrete_task_input::TransientSharedValue;
 use crate::{
-    magic_any::MagicAny, ConcreteTaskInput, RawVc, SharedValue, TaskId, TransientInstance,
-    TransientValue, TypedForInput, Value, ValueTypeId, Vc, VcValueType,
+    debug::{ValueDebugFormat, ValueDebugFormatString},
+    magic_any::MagicAny,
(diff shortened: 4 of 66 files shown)


crates/turbo-tasks/src/lib.rs
```diff
@@ -99,7 +99,7 @@ pub use read_ref::ReadRef;
 pub use state::State;
 pub use task::{
     concrete_task_input::{ConcreteTaskInput, SharedReference, SharedValue},
-    task_input::TaskInput,
+    task_input::{RcStr, TaskInput},
```
ForsakenHarmony, line 102: Can we stop using re-exports in rust?

bgw: My opinion is that this is fine. I'm okay with re-exports at the top-level of a crate that re-export private internals of the crate, especially as rust-analyzer's "go to definition" makes this stuff easy to follow.

The problems I see with re-exports are:
- Cross-crate re-exports that bottleneck the compilation graph, e.g. `turbopack-binding`
- Re-exporting things that are already public, leading to imports from two different competing places because nobody is sure where they should import from.
- Re-exports that just lead to more confusion about the underying type, e.g. `turbo_tasks::Result` is a re-export of `anyhow::Result`, but since `anyhow::Result` is a more widely understood type it would be better to use it directly (and most code does!)

crates/turbo-tasks/src/task/task_input.rs
```diff
@@ -31,7 +40,147 @@ impl TaskInput for ConcreteTaskInput {
     }
 }
 
-impl TaskInput for String {
+/// This type exists to allow swapping out the underlying string type easily.
+//
+// If you want to change the underlying string type to `Arc<str>`, please ensure that you profile
+// perforamnce. The current implementation offers very cheap `String -> RcStr -> String`, meaning we
+// only pay for the allocation for `Arc` when we pass `format!("").into()` to a function.
+#[derive(Debug, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
```
bgw, line 48: Possible follow-up: Instead of using `#[derive(Default)]`, it seems like there's a good opportunity to special-case `Default` to return a globally allocated `RcStr`.

crates/turbo-tasks/src/primitives.rs
```diff
@@ -6,37 +6,39 @@ use futures::TryFutureExt;
...
 
 __turbo_tasks_internal_primitive!(());
 __turbo_tasks_internal_primitive!(String);
+__turbo_tasks_internal_primitive!(RcStr);
 
 #[turbo_tasks::function]
-fn empty_string() -> Vc<String> {
-    Vc::cell(String::new())
+fn empty_string() -> Vc<RcStr> {
+    Vc::cell(RcStr::from(""))
```
bgw, line 18:
```suggestion
    Vc::cell(RcStr::default())
```

crates/turbo-tasks-fs/examples/hash_glob.rs
```diff
@@ -54,18 +54,18 @@ async fn main() -> Result<()> {
 }
 
 #[turbo_tasks::function]
-pub fn empty_string() -> Vc<String> {
-    Vc::cell("".to_string())
+pub fn empty_string() -> Vc<RcStr> {
+    Vc::cell("".into())
```
bgw, line 58: Minor cleanup opportunity: The callsite (in `hash_glob_result`) should really just use the function in `primitives.rs`

crates/turbo-tasks/src/task/task_input.rs
bgw, line 48:
```suggestion
#[derive(Debug, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
#[serde(transparent)]
```

crates/turbo-tasks/src/primitives.rs
```diff
@@ -6,37 +6,39 @@ use futures::TryFutureExt;
...
-fn empty_string_vec() -> Vc<Vec<String>> {
+fn empty_string_vec() -> Vc<Vec<RcStr>> {
     Vc::cell(Vec::new())
 }
 
 impl Vc<Vec<String>> {
     #[deprecated(note = "use Default::default() instead")]
     #[inline(always)]
-    pub fn empty() -> Vc<Vec<String>> {
+    pub fn empty() -> Vc<Vec<RcStr>> {
```
bgw, line 41: Given that this is on the specific `Vc<Vec<String>>` type, I think this should continue returning `Self`.

If you want, you could change this to be an impl on `Vc<Vec<RcStr>>`.

If there's no remaining callsites (or just 1 or 2), we should just delete it.

bgw (approved):
Changes since review LGTM, thanks!

PR #8225, opened 2024-05-28
chore(@turbo/utils): update tests/fixtures to use tasks key

mehulkar:
I've confirmed that `turbo test -F @turbo/utils` passes on this branch


chris-olszewski (approved):
LGTM. PR title is a little misleading since the updates are to use `tasks` key.

PR #8229, opened 2024-05-28
chore(turborepo): Feature flagged off file hashing

NicholasLYang:
### Description

Turning off file hashing in run for now.

### DIFF
--- a/crates/turborepo-lib/src/task_hash.rs
+++ b/crates/turborepo-lib/src/task_hash.rs
@@ -1,16 +1,13 @@
 use std::{
     collections::{HashMap, HashSet},
     sync::{Arc, Mutex},
-    time::Duration,
 };
 
 use rayon::prelude::*;
 use serde::Serialize;
 use thiserror::Error;
 use tracing::{debug, Span};
-use turbopath::{
-    AbsoluteSystemPath, AnchoredSystemPath, AnchoredSystemPathBuf, RelativeUnixPathBuf,
-};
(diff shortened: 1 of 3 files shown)


crates/turborepo-lib/src/task_hash.rs
```diff
@@ -126,58 +118,74 @@ impl PackageInputsHashes {
...
-                                    .collect::<HashMap<_, _>>(),
-                            ),
-                            Err(e) => {
-                                // Daemon could've failed for various reasons. We can still try
-                                // local hashing.
-                                tracing::debug!(
-                                    "daemon file hashing failed for {}: {}",
-                                    package_path,
+                #[cfg(feature = "daemon-file-hashing")]
+                let hash_object = {
```
chris-olszewski, line 122: Super minor, but when possible I think `if cfg!("daemon-file-hashing") {...} else {...}` is better than conditional compilation in situations like this. Too often have we had code that doesn't compile when a feature isn't enabled.

chris-olszewski (approved):

crates/turborepo-lib/src/task_hash.rs
mehulkar: do both syntaxes compile away the branch that isn't used? +1 for writing in a way that protects us from code that doesn't compile

mehulkar (approved):

crates/turborepo-lib/src/task_hash.rs
NicholasLYang: Yeah I'd be really surprised if LLVM can't figure out a const propagation and dead code elimination

PR #8231, opened 2024-05-28
chore(turborepo): Improve pipeline error message

NicholasLYang:
### Description

Made a nicer error message for having a `pipeline` key instead of `tasks`.


mehulkar (approved):

crates/turborepo-lib/src/config.rs
```diff
@@ -132,6 +132,14 @@ pub enum Error {
         #[source_code]
         text: NamedSource,
     },
+    #[error("found `pipeline` field instead of `tasks`")]
+    #[diagnostic(help("In turbo 2.0, `pipeline` has been renamed to `tasks`"))]
```
anthonyshew, line 136:
```suggestion
    #[diagnostic(help("Deprecated in 2.0: `pipeline` has been renamed to `tasks`"))]
```

anthonyshew (approved):
Nitpicked the message but 🚢

crates/turborepo-lib/src/config.rs
NicholasLYang: technically not deprecated since it doesn't work at all, no?

PR #8241, opened 2024-05-29
chore: Update fixtures, test, and schema to tasks key in JS packages

mehulkar:
This commit also updates the schema that is published to /schema.json
since it is using the same TS types that are used by code functionality. 

I intended to only update `eslint-plugin-turbo`, but that involved updating `@turbo/types`, which 
then trickled into `@turbo/codemod` as well.

### DIFF
--- a/packages/turbo-codemod/src/transforms/migrate-env-var-dependencies.ts
+++ b/packages/turbo-codemod/src/transforms/migrate-env-var-dependencies.ts
@@ -15,9 +15,7 @@ const INTRODUCED_IN = "1.5.0";
 export function hasLegacyEnvVarDependencies(config: TurboJsonSchema) {
   const dependsOn = [
     "extends" in config ? [] : config.globalDependencies,
-    Object.values(config.pipeline).flatMap(
-      (pipeline) => pipeline.dependsOn ?? []
-    ),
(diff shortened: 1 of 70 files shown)


packages/turbo-codemod/src/transforms/migrate-env-var-dependencies.ts
```diff
@@ -93,11 +91,11 @@ export function migrateGlobal(config: TurboJsonSchema) {
 
 export function migrateConfig(config: TurboJsonSchema) {
   const migratedConfig = migrateGlobal(config);
-  Object.keys(config.pipeline).forEach((pipelineKey) => {
-    config.pipeline;
-    if (pipelineKey in config.pipeline) {
-      const pipeline = migratedConfig.pipeline[pipelineKey];
-      migratedConfig.pipeline[pipelineKey] = {
+  Object.keys(config.tasks).forEach((pipelineKey) => {
+    config.tasks;
```
mehulkar, line 95: I'm leaving this in because it's out of scope for the PR, but as far as I can tell there's no need for this statement

chris-olszewski (changes requested):
I think we need to keep the old codemods working with `pipeline`. They're intended to be used on 1.x versions of `turbo.json`, not the new 2.0 schema.

PR #8244, opened 2024-05-29
fix(turborepo): Remove optional git locks

NicholasLYang:
### Description

We weren't passing `--no-optional-locks` when calling git. This was unnecessarily locking the index when we didn't need to. Now we pass the flag.

### DIFF
--- a/crates/turborepo-scm/src/git.rs
+++ b/crates/turborepo-scm/src/git.rs
@@ -184,21 +184,9 @@ impl Git {
         file_path: &AbsoluteSystemPath,
     ) -> Result<Vec<u8>, Error> {
         let anchored_file_path = self.root.anchor(file_path)?;
-        let mut command = Command::new(self.bin.as_std_path());
-        let command = command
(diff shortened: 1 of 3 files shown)


mehulkar (commented):
Is this safe? [Docs say](https://git-scm.com/docs/git#Documentation/git.txt-codeGITOPTIONALLOCKScode):

> This is useful for processes running in the background which do not want to cause lock contention with other operations on the repository.

curious, what problem is this solving? is it for watch mode / daemon? And if so, should it apply universally to all git commands?

Zertsov:
> curious, what problem is this solving? is it for watch mode / daemon? And if so, should it apply universally to all git commands?

Can't speak to the first part but because all the other methods call this method it _is_ applied to all git commands so no worries there

NicholasLYang:
I think it's safe. We are running these git actions in the background, so this does fit. And while there are some git commands that are run in the foreground, we are only removing the optional locks.

PR #8254, opened 2024-05-30
chore: stabilize ui

chris-olszewski:
### Description

Rename `experimentalUI` to `ui` along with `TURBO_EXPERIMENTAL_UI` to `TURBO_UI` and removing `--experimental-ui` flag as now this is enabled by default so the flag isn't useful.
### Testing Instructions

Existing test suite / 👀

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -202,9 +202,6 @@ pub struct Args {
     pub check_for_update: bool,
     #[clap(long = "__test-run", global = true, hide = true)]
     pub test_run: bool,
-    /// Enable the experimental UI
-    #[clap(long, hide = true, global = true)]
-    pub experimental_ui: bool,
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -202,9 +202,6 @@ pub struct Args {
     pub check_for_update: bool,
     #[clap(long = "__test-run", global = true, hide = true)]
     pub test_run: bool,
-    /// Enable the experimental UI
-    #[clap(long, hide = true, global = true)]
```
NicholasLYang, line 206: Just to confirm, we're not leaving a flag to toggle off the UI?

NicholasLYang (approved):

crates/turborepo-lib/src/cli/mod.rs
chris-olszewski: Correct, if desired we'll add one.

PR #8255, opened 2024-05-30
feat(turborepo): use biome to parse package.json

NicholasLYang:
### Description

Uses `biome_deserialize` to parse `package.json` files. This gives us span info, and better parse error messages.

### Testing Instructions

The existing tests for `package.json` should suffice.

### DIFF
--- a/crates/turborepo-repository/src/package_json.rs
+++ b/crates/turborepo-repository/src/package_json.rs
@@ -58,8 +58,6 @@ pub struct RawPackageJson {
     pub dev_dependencies: Option<BTreeMap<String, UnescapedString>>,
     pub optional_dependencies: Option<BTreeMap<String, UnescapedString>>,
     pub peer_dependencies: Option<BTreeMap<String, UnescapedString>>,
-    #[deserializable(rename = "turbo")]
-    pub legacy_turbo_config: Option<serde_json::Value>,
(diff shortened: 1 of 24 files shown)


crates/turborepo-repository/src/package_json.rs
```diff
@@ -117,8 +209,8 @@ mod test {
     #[test_case(json!({"name": "foo", "resolutions": {"foo": "1.0.0"}}) ; "berry resolutions")]
     #[test_case(json!({"name": "foo", "pnpm": {"patchedDependencies": {"some-pkg": "./patchfile"}, "another-field": 1}}) ; "pnpm")]
     #[test_case(json!({"name": "foo", "pnpm": {"another-field": 1}}) ; "pnpm without patches")]
-    fn test_roundtrip(json: Value) {
-        let package_json: PackageJson = serde_json::from_value(json.clone()).unwrap();
+    fn test_roundtrip(json: serde_json::Value) {
```
chris-olszewski, line 212: Can we add some more tests here so we have coverage for every field that doesn't go into `rest`? I've gotten bit already by biome and serde having different default behaviors for field name changing.

chris-olszewski (changes requested):
Only concern is with adding more round-trip testing since prune depends on this behavior.

chris-olszewski (approved):

PR #8263, opened 2024-05-31
reduce memory and consolidate shared vs transient cells

arlyon:
### Description

This change reworks the structure to drop an option such that data which is colocated with its type information doesn't need to store it twice.

### Testing Instructions

Existing tests should suffice.

### DIFF
--- a/crates/turbo-tasks/src/manager.rs
+++ b/crates/turbo-tasks/src/manager.rs
@@ -26,7 +26,7 @@ use turbo_tasks_malloc::TurboMalloc;
 use crate::{
     backend::{
         Backend, CellContent, PersistentTaskType, TaskCollectiblesMap, TaskExecutionSpec,
-        TransientTaskType,
+        TransientTaskType, TypedCellContent,
     },
     capture_future::{self, CaptureFuture},
     event::{Event, EventListener},
@@ -99,7 +99,7 @@ pub trait TurboTasksApi: TurboTasksCallApi + Sync + Send {
         &self,
         task: TaskId,
         index: CellId,
-    ) -> Result<Result<CellContent, EventListener>>;
+    ) -> Result<Result<TypedCellContent, EventListener>>;
 
     /// INVALIDATION: Be careful with this, it will not track dependencies, so
     /// using it could break cache invalidation.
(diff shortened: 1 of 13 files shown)


arlyon:
* **#8263** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8263?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/turbo-tasks/src/task/concrete_task_input.rs
```diff
@@ -19,24 +19,49 @@ use crate::{
     registry, turbo_tasks, CellId, RawVc, TaskId, TraitType, ValueTypeId,
 };
 
+/// A reference to a piece of data with optional type information
 #[derive(Clone)]
-pub struct SharedReference(pub Option<ValueTypeId>, pub Arc<dyn Any + Send + Sync>);
+pub struct SharedReference<T: TypeData>(pub T, pub Arc<dyn Any + Send + Sync>);
```
sokra, line 24: Instead of a `SharedReference<T>` I think it would be clearer if it would be `TypedSharedReference` and `SharedReference`.

`CellContent` would store a (untyped) `SharedReference`, since that's what cells store. We don't need an generic argument here. There is only one variant of content stored in cells.

`try_read_task_cell` need to be changed to not return `CellContent` but `Option<TypedSharedReference>` instead. It will add the type info.

crates/turbo-tasks/src/manager.rs
```diff
@@ -1487,23 +1488,24 @@ impl CurrentCellRef {
...
-            CellContent(Some(SharedReference::new(
-                Some(self.index.type_id),
-                triomphe::Arc::new(new_content),
-            ))),
+            CellContent(Some(SharedReference::new(triomphe::Arc::new(new_content)))),
         )
     }
 
-    pub fn update_shared_reference(&self, shared_ref: SharedReference) {
+    pub fn update_shared_reference(&self, shared_ref: TypedSharedReference) {
```
sokra, line 1495: The type info for `shared_ref` isn't used here? Maybe just pass a `SharedReference` or validate the type info with the cell type.

crates/turbo-tasks/src/manager.rs
```diff
@@ -1487,23 +1488,24 @@ impl CurrentCellRef {
...
+            shared_ref_exp.as_ref().ne(&Some(&shared_ref.1))
         } else {
             true
         };
         if update {
-            tt.update_own_task_cell(self.current_task, self.index, CellContent(Some(shared_ref)))
+            tt.update_own_task_cell(
+                self.current_task,
+                self.index,
+                CellContent(Some(shared_ref.1)),
```
sokra, line 1507: destructing shared_ref at the beginning of the method would make it easier to read

sokra (approved):

PR #8267, opened 2024-05-31
fix(turborepo): stderr log

NicholasLYang:
### Description

Removing a stderr log for a warning that was breaking the UI.


chris-olszewski (changes requested):
I think we can't disable this across the board since this is valuable information when it shows up in logs in CI. 

Can we gate it behind a call to `atty::is(atty::Stream::Stdout)` so CI still gets the message?

anthonyshew (approved):

chris-olszewski (approved):
Changing this to 🚢 then fix

PR #8268, opened 2024-05-31
feat(codemod): place tasks last in turbo.json rewrites

chris-olszewski:
### Description

Change order of the object reconstruction so the spread operator happens first to keep the existing ordering of the keys, but only append tasks.

### Testing Instructions

Run codemod on a test repo
Before:
Screenshot 2024-05-31 at 2 00 24 PM
After:
Screenshot 2024-05-31 at 1 59 35 PM


mehulkar (approved):

anthonyshew (approved):
Definitely better. Thanks, Chris. 🫶

PR #8277, opened 2024-06-03
fix(turborepo): non-determinism with root internal deps hash.

NicholasLYang:
### Description

We were getting different task hashes with each run in some monorepos with root internal dependencies. Turns out, the file hashes change depending on which order the packages are hashed. No clue why, but to solve this, we sort the packages before hashing.

### Testing Instructions

Try running manually and check that hashes stay the same. Also added integration test confirming root internal deps behavior


chris-olszewski (changes requested):
Right now the test is really similar to `root-deps.t` test that’s already in the repo. Can you decide on which one to use and cut the other so we don’t have 2 almost identical tests. 

Could you also add multiple root dependencies? If this is only caused by the order we hash root dependencies, then just having 1 root dep means we'll never get a different order.

NicholasLYang:
Gotcha. Also confirmed that test fails without fix

chris-olszewski (approved):
Need to bump task hashes since the new root dependency changed the global hash. Aside from that LGTM

weyert:
I was about to raise a ticket for this issue! Since v2 canaries it seems to always rebuild packages or have cache misses. 

You would do `pnpm exec turbo run build` and then ctrl-c and then run `pnpm exec turbo run build` again it would rebuild everything.

PR #8286, opened 2024-06-04
Store aggregate read/execute count statistics

bgw:
### Why?

I want to determine percent "cache hit" rates for tasks. Tasks with very low task hit rates should likely have their annotations removed.

Eventually, we might be able to use this information in a more automated way, by leaving the annotation in, but skipping the caching for low-cache-hit tasks.

### What?

This implementation only logs persistent tasks, which should compromise all or the majority of tasks we care about for memory usage.

The implementation should bail out quickly if caching is disabled, so it should be okay to leave in release builds, which is important for making it easy to gather statistics from willing users.

### Testing

Run included unit tests!

This is used as part of https://github.com/vercel/next.js/compare/canary...bgw/cache-hit-stats


bgw:
* **#8286** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8286?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/turbo-tasks-memory/src/task_statistics.rs
```diff
@@ -0,0 +1,82 @@
...
+pub struct TaskStatisticsApi {
+    inner: OnceLock<Arc<TaskStatistics>>,
+}
+
+impl TaskStatisticsApi {
+    pub fn enable(&self) -> &Arc<TaskStatistics> {
+        // ignore error: potentially already initialized, that's okay
+        let _ = self.inner.set(Arc::new(TaskStatistics {
+            inner: DashMap::new(),
+        }));
```
kdy1, line 18: I think we should use [`get_or_init`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.get_or_init) instead

crates/turbo-tasks-memory/src/task_statistics.rs
```diff
@@ -0,0 +1,82 @@
...
+        self.inner.get().unwrap()
```
kdy1, line 19:
```suggestion
```

crates/turbo-tasks-memory/src/task_statistics.rs
```diff
@@ -0,0 +1,82 @@
...
+    // [`TaskStatisticsApi::enable`]).
+    pub fn get(&self) -> Option<&Arc<TaskStatistics>> {
+        self.inner.get()
+    }
+}
+
+/// A type representing the enabled state of [`TaskStatisticsApi`]. Implements
+/// [`serde::Serialize`].
+pub struct TaskStatistics {
+    inner: DashMap<FunctionId, TaskFunctionStatistics>,
```
kdy1, line 42:
```suggestion
    inner: DashMap<FunctionId, TaskFunctionStatistics, rustc_hash::FxBuildHasher>,
```

`DashMap` uses the default hasher, so it can be slow. I think if this type stores stats of all tasks, it can be an issue.

crates/turbo-tasks/src/backend.rs
```diff
@@ -384,45 +396,55 @@ impl PersistentTaskType {
...
-                    }
+
+        let this = iter
+            .next()
+            .expect("No arguments for trait call")
+            .resolve()
+            .await?;
+        let this_value = this.clone().resolve_to_value().await?;
+
+        let native_fn = Self::resolve_trait_method_from_value(trait_type, this_value, name)?;
```
sokra, line 407: This calls `resolve_to_value` again on `this_value`. That's a bit wasteful.

crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -528,8 +535,53 @@ impl Backend for MemoryBackend {
...
+                    // default. The alternative would be to store function ids on completed
+                    // ResolveTrait tasks.
+                    let trait_type = *trait_type;
+                    let name = name.clone();
+                    let this = inputs
+                        .first()
+                        .cloned()
+                        .expect("No arguments for trait call");
+                    let stats = Arc::clone(stats);
+                    turbo_tasks.run_once(Box::pin(async move {
```
sokra, line 564: this could use a `tokio::task::spawn` instead. That's much more lightweight

kdy1 (approved):

crates/turbo-tasks-memory/src/memory_backend.rs
bgw: I can't do this because `resolve_trait_method` needs to be run within the context of `turbo_tasks`.

It looks like passing through the task-local context isn't safe: https://github.com/vercel/turbo/blob/main/crates/turbo-tasks/src/manager.rs#L1323-L1329

bgw:
### Merge activity

* **Jul 2, 3:03 PM EDT**: @bgw merged this pull request with [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/8286).

PR #8289, opened 2024-06-04
feat: change values of "ui" in turbo.json

chris-olszewski:
### Description

It was mentioned that `"ui": false` doesn't make much sense. We replace that in this PR with `"tui"` for the rich terminal output and `"stream"` for the non-TTY behavior.

### Testing Instructions

Updated migration and unit tests. Quick manual check on a test repo.
Screenshot 2024-06-04 at 8 51 53 AM


mehulkar (approved):

tanishqmanuja:
Is there an equivalent cli flag to enable tui, i don't want to set it as default in turbo.json ?

PR #8316, opened 2024-06-05
vim like navigation

dcodesdev:
### Description

Allows vim like navigation https://github.com/vercel/turbo/discussions/8313


chris-olszewski:
I appreciate the PR, but we're trying to keep the TUI as simple as possible for the moment.

PR #8329, opened 2024-06-05
Temporarily remove docs folder.

anthonyshew:
### Description

We are actively working on migrating the documentation to a new workflow. In the meantime, let's remove this directory so folks aren't confused about the difference between the currently deployed documentation and this directory.

It will be back, stay tuned!


mehulkar (commented):
where's the source while this is deleted?

chris-olszewski (approved):

bgw (approved):

PR #8347, opened 2024-06-06
Small tree shaking fix and test case update

sokra:
### Description

avoid adding module evaluation to internal part imports
add a broken test case
update split-chunk test case to work and be more clear


crates/turbopack-ecmascript/tests/tree-shaker/analyzer/multi-export/output.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Item 3: Stmt 2, `Normal`
+
+```js
+export { dog as DOG, cat };
+
+```
+
+- Side effects
+- Reads: `dog`, `cat`
```
sokra, line 33: @kdy1 The whole statement is pretty wrong.

It should not have side effects. It should not read vars. It shouldn't exist at all.

crates/turbopack-ecmascript/tests/tree-shaker/analyzer/multi-export/output.md
```diff
@@ -0,0 +1,230 @@
...
+export { dog } from "__TURBOPACK_VAR__" assert {
+    __turbopack_var__: true
+};
+
+```
+## Part 1
+```js
+export { DOG as DOG };
+
+```
```
sokra, line 138: @kdy1 This looks weird

crates/turbopack-ecmascript/tests/tree-shaker/analyzer/multi-export/output.md
kdy1: But if it does not exist, dependency graph will be wrong I think

kdy1 (approved):

PR #8348, opened 2024-06-06
fix: Apply `paren-remover` while minifying

kdy1:
### Description

This is required to fix https://vercel.slack.com/archives/C04KC8A53T7/p1717599802994489

### Testing Instructions

Closes SWC-171


timneutkens (dismissed):
The PR seems to be missing a test to prevent future regression 👀

crates/turbopack-ecmascript/src/minify.rs
```diff
@@ -63,6 +64,8 @@ pub async fn minify(path: Vc<FileSystemPath>, code: Vc<Code>) -> Result<Vc<Code>
             let top_level_mark = Mark::new();
 
             Ok(compiler.run_transform(handler, false, || {
+                let program = program.fold_with(&mut paren_remover(Some(&comments)));
```
eps1lon, line 67: Can the minifier check if that fixer was applied?

Considering "Fixes ast nodes before printing so semantics are preserved.", it feels like   somewhere we should crash if the fixer isn't applied. 

In other words: When would you ever want to not preserve semantics?

sokra (approved):

crates/turbopack-ecmascript/src/minify.rs
kdy1: It should be done by `codegen`, but we can't know if it's applied from codegen

crates/turbopack-tests/tests/snapshot/minification/paren-remover/input/index.js
```diff
@@ -0,0 +1,30 @@
+function toFixed(value, maxDecimals, roundingFunction, optionals) {
```
kdy1, line 1: This test is copied from https://github.com/swc-project/swc/pull/8442

kdy1: Note: I also verified this fix by disabling `paren_remover` in the testing system for SWC minifier

PR #8384, opened 2024-06-07
Warn instead of print when no locally installed  version.

anthonyshew:
### Description

Let's make sure we give an extra little bump in this situation. A warn, instead of only a `println!`. Realistically, this isn't a situation that folks want to be getting themselves into often/purposefully.


chris-olszewski (approved):
LFG

PR #8385, opened 2024-06-08
feat(shim): invoke local turbo version via npx if not installed

chris-olszewski:
### Description

I highly recommend reviewing the first 4 commits on their own as they are primarily breaking up the massive `shim` module.

This PR adds fallback option when global turbo doesn't find a local turbo install. If a local install isn't found we will look at:
 - lockfiles
 - package.json
 - turbo.json to identify if turbo 1 is expected

If one of those result in finding a version of `turbo`, then we'll invoke `turbo` via `npx` to download the correct version on the fly.

This behavior can be disabled with by setting `TURBO_DOWNLOAD_LOCAL_DISABLED` to `1` or `true`

### Testing Instructions

Added unit tests for detection behavior.

Manual testing of repos with/without a local install of `turbo`.


crates/turborepo-lockfiles/src/lib.rs
```diff
@@ -62,6 +62,11 @@ pub trait Lockfile: Send + Sync + Any + std::fmt::Debug {
 
     /// Determine if there's a global change between two lockfiles
     fn global_change(&self, other: &dyn Lockfile) -> bool;
+
+    /// Return any turbo version found in the lockfile
+    fn turbo_version(&self) -> Option<String> {
+        None
```
arlyon, line 68: Not a strong opinion but I'd not want to forget to implement this

crates/turborepo-lib/src/shim/local_turbo_config.rs
```diff
@@ -0,0 +1,82 @@
...
+        if let Ok(package_manager) = &repo_state.package_manager {
+            let lockfile = package_manager
+                .read_lockfile(&repo_state.root, &repo_state.root_package_json)
+                .ok()?;
+            return lockfile.turbo_version();
+        }
+
+        // If there isn't a package manager, just try to parse all known lockfiles
+        // This isn't the most effecient, but since we'll be hitting network to download
+        // the correct binary the unnecessary file reads.
```
arlyon, line 57: Is this comment cut off?

crates/turborepo-lib/src/shim/local_turbo_config.rs
```diff
@@ -0,0 +1,82 @@
...
+        PackageManager::supported_managers().iter().find_map(|pm| {
+            let lockfile = pm
+                .read_lockfile(&repo_state.root, &repo_state.root_package_json)
+                .ok()?;
```
arlyon, line 61: If parsing the back-up fails perhaps we just continue and try others rather than exit? Your call, thought I'd mention it.

crates/turborepo-lib/src/shim/mod.rs
```diff
@@ -111,9 +120,66 @@ fn spawn_local_turbo(
...
+        .map_err(|_| Error::RepoRootPath(repo_state.root.clone()))?;
+
+    let raw_args = modify_args_for_local(&mut shim_args, repo_state, turbo_version);
+
+    let mut command = process::Command::new(npx_path);
+    command.arg("-y");
+    command.arg(&format!("turbo@{turbo_version}"));
+
+    // rather than passing an argument that local turbo might not understand, set
+    // an environment variable that can be optionally used
```
arlyon, line 167: Cool

arlyon (approved):
Nothing critical, approving as a fix ship on those comments but feel free to pick which ones you find useful

crates/turborepo-lib/src/shim/local_turbo_config.rs
chris-olszewski: We still do since the `?` is in the `find_map` closure it'll just short circuit the current package manager check and move onto the next one.

chris-olszewski:
Merging with example test failure. `with-berry` looks like it is currently broken on main: https://github.com/vercel/turbo/actions/runs/9455511595/job/26045482508?pr=8410

PR #8390, opened 2024-06-09
Replace chalk by picocolors

torresgol10:
We replaced chalk with picocolors as it is much more optimal and smaller. 

Discussions: #8341

In this PR has been the replacement in all the packages inside /packages/, I will make another one for the update of the examples, but I can also include in this same PR.

I have deleted in 2 package.json the dependency on: @types/chalk-animation as I checked and it was not being used.

Picocolors does not have the function to pass a hexadecimal colour to ASCII, to keep the colours, I have created the function and here 2 things to comment.

I have created this function in the most used place in the turbo-utils / logger, but it is also used in 1 specific place, that's why I have used the export.

1.  The implementation does not use the ‘bits’ because typescript gives error, if you see it necessary I can change the implementation.
2.  I don't know if this is the right place to put this function, do you think it's necessary to create a test?


ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 9aa2b21d92648d29e70762a8c7b973e570f3b2c3

packages/turbo-gen/tsup.config.ts
```diff
@@ -1,6 +1,7 @@
 import { defineConfig, Options } from "tsup";
 import fs from "fs-extra";
-import chalk from "chalk";
+import { green } from "picocolors";
+import { hex } from "@turbo/utils/src/logger";
```
anthonyshew, line 4: This export doesn't exist. You'll want to re-export the `hex` function from `packages/turbo-utils/index.ts` to make it available at `@turbo/utils`.

anthonyshew (changes requested):
Awesome! Will help my `npx create-turbo` that much quicker.

Going through each package to visually check behavior:
- [x] turbo-workspaces
- [x] turbo-telemetry
- [ ] turbo-gen - Won't build but I can't quite tell why. ESM vs CJS?
- [ ] turbo-codemod - Won't build but I can't quite tell why. ESM vs CJS?
- [ ] `create-turbo`
CleanShot 2024-06-15 at 15 19 23@2x

Seems like it has to do with the implementation of `hex`.

packages/turbo-gen/tsup.config.ts
torresgol10: Thank you very much, I'll check that export and upload the change.

torresgol10:
@anthonyshew Sorry, I had a bug in the implementation so it didn't paint lines it shouldn't, also change the 'hex' function to const to keep the same programming style, just that you had to move it up in the file to make it's calls work.

About the `hex` import, I modified it because I detected that all logger exports are made to logger, so I had to import logger and call `hex` logger.hex.

anthonyshew:
Thanks! Updated list:
- [x] turbo-workspaces
- [x] turbo-telemetry
- [x] `create-turbo`
- [ ] turbo-gen - Won't build but I can't quite tell why.
- [ ] turbo-codemod - Won't build but I can't quite tell why.

Once those are buildable again, I'll give them a look.

torresgol10:
@anthonyshew I'm sorry, I've been looking at the reason for the fault and I've managed to find out where it comes from but I can't find out why it's wrong, I don't know if you could help me.

image

anthonyshew:
No worries! I think this is the dreaded ESM/CJS error. I'll do my best to find time to take a look at this when I can.

torresgol10:
@anthonyshew I have made a few changes and it compiles perfectly. 🎉

I have removed the `logger` import and changed the colour from purple to picocolors magenta.
I have changed the picocolors import because it also gave an error.

If the colour change is not a problem (it is a "similar" colour) I think you can continue with the tests.

Thank you very much.

anthonyshew:
@torresgol10, it looks like there are a few more references to do away with: https://github.com/vercel/turbo/actions/runs/9897583783/job/27342448642?pr=8390

The ones in `src/commands/create/index.ts` are from us making a few updates to `npx create-turbo` for better logging output. Sorry for adding to your work. Also, FYI, I merged main onto the branch because I was aware of that change, so sorry for messing up your history!

Once we have those cleared up, we should be good to go!

torresgol10:
@anthonyshew It's OK, it's normal to have to make changes, it would be all sorted out. Thank you very much and I would love to keep making small changes to this great project.

torresgol10:
@anthonyshew  It's OK, it's normal to have to make changes, everything would be solved and the conflicts would be solved. Thank you very much and I would love to continue making small changes to this great project.

torresgol10:
@anthonyshew I don't know if I did it right, but I saw an error in the "Formating" and I fixed it by making a new commit.

PR #8393, opened 2024-06-10
Add CI to default passthroughs.

anthonyshew:
### Description

By convention, CI systems will set a `CI` environment variable. Given the ubiquity of this variable, there are some tools that rely on this variable, which is what I think we're seeing in #8281.


dimitropoulos (approved):
you should probably wait for @chris-olszewski to approve as well, but this looks correct: my understanding is that strict mode was, by default, not allowing the passthrough of a common environment variable that it should have.  Since we turned on strict mode recently by default, that's what's causing people to suddenly hit this since they probably don't otherwise have a manual override in their config.

PR #8405, opened 2024-06-10
Fix `eslint-config-turbo` configuration in examples.

anthonyshew:
### Description

Fixing my own mistake on wrong reference pattern for ESLint configurations. Come on, rookie!

Also, sneaking in change to remove the root TSConfig file from the `basic` example. It isn't needed to run TypeScript in package directories and don't want to confuse folks.


chris-olszewski (approved):

steveluscher:
Does this also make it so that `codemod` updates the version in `package.json`? I just ran `codemod` a few minutes ago and it didn't, but I presume that's because this isn't released yet (didn't read the PR).

PR #8407, opened 2024-06-10
BeforeResolvePluginCondition: allow module filtering as a condition for `BeforeResolvePlugin`s

wbinnssmith:
This expands `BeforeResolvePluginCondition` to match on individual modules rather than just the request string.

Needed for https://github.com/vercel/next.js/pull/66622

### DIFF
--- a/crates/turbopack-core/src/resolve/plugin.rs
+++ b/crates/turbopack-core/src/resolve/plugin.rs
@@ -1,5 +1,5 @@
 use anyhow::Result;
-use turbo_tasks::{Value, Vc};
+use turbo_tasks::{RcStr, Value, Vc};
 use turbo_tasks_fs::{glob::Glob, FileSystemPath};
 
 use crate::{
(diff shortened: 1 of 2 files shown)


arlyon (approved):

crates/turbopack-core/src/resolve/plugin.rs
```diff
@@ -41,26 +41,37 @@ impl AfterResolvePluginCondition {
...
+    pub fn from_request_glob(glob: Vc<Glob>) -> Vc<Self> {
+        BeforeResolvePluginCondition::Request(glob).cell()
+    }
+
+    #[turbo_tasks::function]
+    pub fn from_module(module: RcStr) -> Vc<Self> {
+        BeforeResolvePluginCondition::Module(module).cell()
     }
 
     #[turbo_tasks::function]
```
sokra, line 61: Could we change that to an non-turbo-tasks function? To an normal async function.

This would avoid the need to spawn a new task to check the request.

sokra: I think that was not clear enough... `matches` should be a normal function. `from_module` should stay a turbo-tasks function .

wbinnssmith:
* **#8439** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8439?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8407** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @wbinnssmith and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

sokra (approved):

PR #8421, opened 2024-06-11
Fix Turbopack benchmark for next@canary

timneutkens:
### Description

- Ensures that the benchmarks don't fail on installing as `prefer-offline` doesn't work reliably.
- Installs `react@rc` and `react-dom@rc` for `next@canary` as it requires React 19.
- Updated the benchmark docs to show to to run one suite

### DIFF
--- a/crates/turbopack-bench/src/util/mod.rs
+++ b/crates/turbopack-bench/src/util/mod.rs
@@ -70,7 +70,7 @@ pub fn build_test(module_count: usize, bundler: &dyn Bundler) -> TestApp {
     .unwrap();
 
     let npm = command("npm")
-        .args(["install", "--prefer-offline", "--loglevel=error"])
+        .args(["install", "--loglevel=error"])
         .current_dir(test_app.path())
         .output()
         .unwrap();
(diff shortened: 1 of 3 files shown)


timneutkens:
* **#8421** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8421?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @timneutkens and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

crates/turbopack-bench/src/util/mod.rs
```diff
@@ -70,7 +70,7 @@ pub fn build_test(module_count: usize, bundler: &dyn Bundler) -> TestApp {
     .unwrap();
 
     let npm = command("npm")
-        .args(["install", "--prefer-offline", "--loglevel=error"])
+        .args(["install", "--loglevel=error"])
```
timneutkens, line 73: prefer-offline does not work reliably, it ends up causing resolving of dependencies to fail with an error saying that a package version does not exist that does exist on the npm registry.

sokra (approved):

sokra (approved):

PR #8437, opened 2024-06-11
fix: remove inferring turbo version from package.json or turbo.json

chris-olszewski:
### Description

We're seeing failures when we try to invoke `turbo` via npx with a range instead of a fully resolved version. We'll try to get to the bottom of this, but in the meantime we'll avoid this codepath.

For the most part this makes sense as users should have a lockfile. Guessing version based on package.json or turbo.json can still result in a version mismatch so restricting this behavior only when we're sure of the version the user wants makes sense.

### Testing Instructions

Updated unit tests

### DIFF
--- a/crates/turborepo-lib/src/shim/local_turbo_config.rs
+++ b/crates/turborepo-lib/src/shim/local_turbo_config.rs
@@ -1,9 +1,5 @@
 use std::env;
 
-use jsonc_parser::{
-    ast::{ObjectPropName, Value},
-    parse_to_ast,
-};
 use tracing::debug;
 use turborepo_repository::{inference::RepoState, package_manager::PackageManager};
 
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/shim/local_turbo_config.rs
```diff
@@ -25,18 +21,7 @@ impl LocalTurboConfig {
...
-            .or_else(|| {
-                debug!(
-                    "No turbo version found in a lockfile. Attempting to read version from root \
-                     package.json"
-                );
-                Self::turbo_version_from_package_json(repo_state)
-            })
-            .or_else(|| {
-                debug!("No turbo version found in package.json. Checking if turbo.json is for v1");
-                Self::turbo_version_from_turbo_json_schema(repo_state)
```
NicholasLYang, line 38: Just wondering, why are we also removing this check? Similarly inconsistent?

NicholasLYang (approved):

crates/turborepo-lib/src/shim/local_turbo_config.rs
chris-olszewski: Yeah, just a little too loose. Want to avoid at just guessing what version of `turbo` a user wants.

PR #8456, opened 2024-06-12
Restore backticks on docs.

anthonyshew:
### Description

There was a bug in the upstream docs metaframework that didn't work with backticks in headings. This is now resolved, so bringing them back.


chris-olszewski (approved):
Do we have a way to test this?

PR #8457, opened 2024-06-12
fix(turborepo): avoid starting ui on too small terminals

NicholasLYang:
### Description

Avoids starting the UI when the terminal is too small. Also bumps crossterm because apparently we get an incorrect terminal size on older versions.

### Testing Instructions

Tested with `lefthook` on create-turbo

### DIFF
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -160,6 +160,14 @@ fn poll(input_options: InputOptions, receiver: &AppReceiver, deadline: Instant) 
     }
 }
 
+const MIN_HEIGHT: u16 = 10;
+const MIN_WIDTH: u16 = 20;
+
+pub fn terminal_big_enough() -> Result<bool, Error> {
+    let (width, height) = crossterm::terminal::size()?;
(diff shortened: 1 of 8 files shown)


crates/turborepo-ui/src/tui/app.rs
```diff
@@ -160,6 +159,14 @@ fn poll(input_options: InputOptions, receiver: &AppReceiver, deadline: Instant)
     }
 }
 
+const MIN_HEIGHT: u16 = 10;
```
NicholasLYang, line 162: These are completely arbitrary. Open to suggestions on what they should actually be

chris-olszewski: I think these are completely reasonable. If I'm accounting for borders correctly 20x10 term would probably lead to a 12x8 terminal pane. Not sure if that's usable, but anything smaller is definitely not usable.

chris-olszewski (approved):
LGTM

crates/turborepo-ui/src/tui/app.rs
mehulkar: add a comment that it's arbitrary?

mehulkar (approved):

PR #8461, opened 2024-06-12
fix(shim): avoid panic if user has malformed lockfile

chris-olszewski:
### Description

Since this function is now possibly getting user input, we shouldn't hard crash if whatever they have listed in their lockfile doesn't pass the `semver` crate's version requirement.

This shouldn't happen as all lockfiles are generated with fully resolved versions, but this prevents a panic if that expectation changes.

### Testing Instructions

Added unit tests that would trigger a panic previously

### DIFF
--- a/crates/turborepo-lib/src/shim/local_turbo_state.rs
+++ b/crates/turborepo-lib/src/shim/local_turbo_state.rs
@@ -5,7 +5,7 @@ use std::{
 
 use camino::Utf8PathBuf;
 use dunce::canonicalize as fs_canonicalize;
-use semver::Version;
+use semver::{Version, VersionReq};
 use serde::Deserialize;
 use tracing::debug;
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/shim/local_turbo_state.rs
```diff
@@ -185,37 +185,39 @@ impl Default for YarnRc {
...
+        // only need to check major and minor (this will include canaries)
+        if version.major == 1 {
+            return version.minor >= 7;
+        }
+        version.major > 1
+    } else {
+        // In the case that we don't get passed a valid semver we should avoid a panic.
+        // We shouldn't hit this we introduce back infering package version from schema
+        // or package.json.
+        true
```
chris-olszewski, line 198: This is technically too loose, but the correct answer for the past 21 months of `turbo` releases.

crates/turborepo-lib/src/shim/local_turbo_state.rs
```diff
@@ -5,7 +5,7 @@ use std::{
 
 use camino::Utf8PathBuf;
 use dunce::canonicalize as fs_canonicalize;
-use semver::Version;
+use semver::{Version, VersionReq};
```
Zertsov, line 8: nit: unused import warning

Zertsov (approved):

mehulkar (approved):

PR #8471, opened 2024-06-13
fix(@turbo/codemode): no-op when turbo.json already contains tasks key

mehulkar:


anthonyshew (approved):
Small but mighty PR, love it.

chris-olszewski (approved):

PR #8492, opened 2024-06-15
Improve error readability.

anthonyshew:
### Description

Currently in a sloppy state to get feedback.

#### Proposed improvements:
- Bring link to the bottom of the message. My eye skips over the URL because it naturally jumps to the more verbose code output. With the URL at the bottom, I have the chance as a user to read the error and give it a fix. If I keep reading because I need more info, I make it to the link.
- Remove the diagnostic code. It doesn't mean anything to me as a user (and I'd imagine those of us on core would be able to sort out the error's location in source using other heuristics?).

#### Things to fix:
- The dangling `(link)` at the top needs to actually become the link at the bottom.
- The custom miette handler I'm adding isn't in a good location. I only dropped it in the file to hear what folks think.
- Tests.

Before:
CleanShot 2024-06-14 at 19 45 05@2x

After:
CleanShot 2024-06-14 at 19 44 48@2x

### Testing Instructions

👀

### DIFF
--- a/crates/turborepo-lib/src/config.rs
+++ b/crates/turborepo-lib/src/config.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use convert_case::{Case, Casing};
-use miette::{Diagnostic, NamedSource, SourceSpan};
+use miette::{Diagnostic, NamedSource, Result, SourceSpan};
 use serde::Deserialize;
 use struct_iterable::Iterable;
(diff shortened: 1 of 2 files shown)


anthonyshew:
After talking to @NicholasLYang this is possibly a bug in miette? https://github.com/zkat/miette/issues/383

PR #8494, opened 2024-06-15
Spruce up Remote Cache linking message.

anthonyshew:
### Description

Before:
CleanShot 2024-06-14 at 20 41 32@2x

After:
CleanShot 2024-06-14 at 20 41 08@2x

### Testing instructions
👀

### DIFF
--- a/crates/turborepo-lib/src/commands/link.rs
+++ b/crates/turborepo-lib/src/commands/link.rs
@@ -20,7 +20,7 @@ use thiserror::Error;
 use turborepo_api_client::{CacheClient, Client};
 #[cfg(not(test))]
 use turborepo_ui::CYAN;
-use turborepo_ui::{DialoguerTheme, BOLD, GREY, UNDERLINE};
+use turborepo_ui::{DialoguerTheme, BOLD, GREY};
 use turborepo_vercel_api::{CachingStatus, Space, Team};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/commands/link.rs
```diff
@@ -89,11 +89,9 @@ pub(crate) enum SelectedSpace<'a> {
     Space(&'a Space),
 }
 
-pub(crate) const REMOTE_CACHING_INFO: &str = "  Remote Caching shares your cached Turborepo task \
-                                              outputs and logs across
-  all your team’s Vercel projects. It also can share outputs
-  with other services that enable Remote Caching, like CI/CD systems.
-  This results in faster build times and deployments for your team.";
+pub(crate) const REMOTE_CACHING_INFO: &str =
+    "Remote Caching makes your caching multiplayer,\nsharing build outputs and logs between \
```
mehulkar, line 93: can we not repeat "caching" twice if we're going to change the language? how about:

```suggestion
    "Remote Caching makes your tasks multiplayer,\nsharing build outputs and logs between \
```

anthonyshew: Is `task caching` better? I get that that doesn't make it not say "caching" twice but `tasks` leaves room for interpretation.

mehulkar: 🤷🏾 I don't like task caching that much better, but not going to block on it either!

mehulkar (approved):

PR #8498, opened 2024-06-15
UI instructions touch up.

anthonyshew:
### Description

~10 characters shorter and added whitespace on the ends to create some breathing room from borders.

### Testing Instructions

👀

Before:
CleanShot 2024-06-15 at 00 22 48@2x

After:

CleanShot 2024-06-15 at 00 24 17@2x


anthonyshew:
Deferring to #8499

PR #8499, opened 2024-06-15
Light refactoring of TUI pane and instructions.

anthonyshew:
### Description

Some light TUI refactors.
- Removing some of the borders around the logs pane for less visual clutter
- Interactivity instructions in logs pane use state to show "enter" vs. "exit" instructions
- "Localize" the instructions for moving up and down the task list to the task list itself

### Testing Instructions

👀 

### Future work
If we could get the red-to-blue gradient on the left sidebar when interactive... :chefs-kiss: - but I haven't looked at how to do that with Ratatui yet.

### DIFF
--- a/crates/turborepo-ui/src/tui/pane.rs
+++ b/crates/turborepo-ui/src/tui/pane.rs
@@ -14,7 +14,8 @@ use turborepo_vt100 as vt100;
 
 use super::{app::Direction, Error};
 
-const FOOTER_TEXT: &str = "Press `Enter` to interact and `Ctrl-Z` to stop interacting.";
+const FOOTER_TEXT_ACTIVE: &str = "Press`Ctrl-Z` to stop interacting.";
+const FOOTER_TEXT_INACTIVE: &str = "Press `Enter` to interact.";
 
 pub struct TerminalPane<W> {
     tasks: BTreeMap<String, TerminalOutput<W>>,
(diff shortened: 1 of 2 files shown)


mehulkar:
Removing my review request from here, because I really don't feel like I can be valuable here

crates/turborepo-ui/src/tui/table.rs
```diff
@@ -53,7 +53,7 @@ impl TaskTable {
             .unwrap_or_default()
             // Task column width should be large enough to fit "Task" title
             // and truncate tasks with more than 40 chars.
-            .clamp(4, 40) as u16;
+            .clamp(14, 40) as u16;
```
chris-olszewski, line 56: Was this bump intentional?

anthonyshew: The idea was that the minimum width needs to be the width of the "up/down to navigate" line now instead of the "Task" line. Wrong spot?

chris-olszewski: Right spot, I just think it needs to happen a few other places as well.

anthonyshew: Only thing that stood out to me was this constraint, bumped it: https://github.com/vercel/turbo/pull/8499/files#diff-17d339c7c94986602475a5018ddff2b58833d1227bd33434eba1214976c4fddfL258

Anywhere else?

crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -219,11 +219,13 @@ impl<W> Widget for &TerminalPane<W> {
...
-            .borders(Borders::ALL)
-            .title(task.title(task_name))
-            .title_bottom(Line::from(FOOTER_TEXT).centered());
+            .borders(Borders::LEFT)
+            .title(task.title(task_name));
         if self.highlight {
+            block = block.title_bottom(Line::from(FOOTER_TEXT_ACTIVE).centered());
             block = block.border_style(Style::new().fg(ratatui::style::Color::Yellow));
+        } else {
+            block = block.title_bottom(Line::from(FOOTER_TEXT_INACTIVE).centered());
```
chris-olszewski, line 228: Love this change

chris-olszewski (approved):

PR #8505, opened 2024-06-15
[WIP] Fix Framework Inference for eslint-config-turbo.

anthonyshew:
### Description

A fairly naive implementation of [Framework Inference](https://turbo.build/repo/docs/crafting-your-repository/using-environment-variables#framework-inference) (as it stands at the moment). `turbo`'s Framework Inference support is per-package whereas this one wholesale allowslists all framework environment variables in every package.

I tried my hand at making it per-package but fell flat. Could use some assistance if we don't think this is sufficient. From my perspective, the risk of someone using the wrong vars in the wrong framework is minimal and this is an improvement on our current support (which is none at all).

### Testing Instructions

Double-checking the tests I have would be great. I've never written an ESLint plugin before and had...quite a time with this.

### DIFF
--- a/packages/eslint-plugin-turbo/__tests__/lib/no-undeclared-env-vars.test.ts
+++ b/packages/eslint-plugin-turbo/__tests__/lib/no-undeclared-env-vars.test.ts
@@ -63,18 +63,6 @@ ruleTester.run(RULES.noUndeclaredEnvVars, rule, {
     },
     {
       code: `
-        const { NEXT_PUBLIC_HAHAHAHA } = process.env;
-      `,
-      options: [
-        { cwd: path.join(__dirname, "../../__fixtures__/workspace-configs") },
-      ],
-      filename: path.join(
-        __dirname,
-        "../../__fixtures__/workspace-configs/apps/web/index.js"
-      ),
-    },
-    {
-      code: `
         const { ENV_1 } = process.env;
       `,
       options: [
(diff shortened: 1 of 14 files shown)


packages/eslint-plugin-turbo/__tests__/lib/no-undeclared-env-vars.test.ts
anthonyshew, line 1: Removing these since the intent is to auto-include these.

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -81,6 +83,10 @@ function create(context: RuleContextWithOptions): Rule.RuleListener {
     }
   });
 
+  frameworkInferencePrefixes.forEach((prefix) => {
+    regexAllowList.push(new RegExp(`^${prefix}`));
```
anthonyshew, line 87: This is too naive, since it doesn't account for package boundaries. It's going to auto-allow `NEXT_PUBLIC` in a Vite application.

(In practice, this probably isn't a big problem but let's see what we can do to be 100% correct...)

tknickman (commented):
Nice! Let's chat about this today. I think we should compare this with what is available in `vercel/vercel`. Ultimately we have to match turbo's inference here, but I _think_ that was ported from `vercel/vercel` originally. So instead of porting that back to JS I would prefer that we just use that inference directly. 

But if we've diverged we'll need to roll it ourselves so we match directly. We'll also need to think about a good way to ensure this doesn't drift from what's in turbo core.

PR #8521, opened 2024-06-17
Trying out new release process via cargo dist

NicholasLYang:


anthonyshew:
This was already done! Woohoo!

PR #8523, opened 2024-06-18
feat: Enable tree shaking in next.js

kdy1:
### Description

- next.js counterpart: https://github.com/vercel/next.js/pull/66689

I'll debug/fix tree shaking implementation with this PR


crates/turbopack-ecmascript/src/tree_shake/cjs_finder.rs
```diff
@@ -1,11 +1,146 @@
...
+
+                // Skip special reexports that are recognized by next.js
+                ModuleItem::ModuleDecl(ModuleDecl::ExportDecl(ExportDecl {
+                    decl: Decl::Fn(f),
+                    ..
+                })) => {
+                    if is_next_js_special_export(&f.ident.sym) {
+                        return true;
+                    }
+                }
```
bgw, line 84: I don't love the idea of putting more highly-nextjs-specific things in the core parts of turbopack. Could we instead make this configurable (e.g. with a passed in closure) and move `is_next_js_special_export` into the nextjs PR?

crates/turbopack-ecmascript/src/tree_shake/graph.rs
```diff
@@ -482,19 +493,23 @@ impl DepGraph {
...
 
         loop {
             let mut changed = false;
 
-            for (group, group_done) in &mut groups {
-                let start = &group[0];
-                let start_ix = self.g.get_node(start);
-                changed |=
-                    add_to_group(&self.g, data, group, start_ix, &mut global_done, group_done);
+            for (group, group_done, init_len) in &mut groups {
```
bgw, line 504: I'm not sure why `init_len` was added here when it always ends up getting set to `group.len()`? Can we just do `for i in 0..group.len()` or `for start in &group`?

crates/turbopack-ecmascript/src/tree_shake/graph.rs
```diff
@@ -447,7 +457,8 @@ impl DepGraph {
 
                     global_done.extend(cycle.iter().copied());
 
-                    groups.push((ids, Default::default()));
+                    let len = ids.len();
+                    groups.push((ids, Default::default(), len));
```
bgw, line 461: Nit: for consistency?

```suggestion
                    groups.push((ids, FxHashSet::default(), len));
````

Or could we just leave this off and create an empty `FxHashSet` inside the `add_to_group` loop? It looks like this is always initialized to an empty set, and never read after that loop?

File: crates/turbopack-ecmascript/src/tree_shake/mod.rs [hunk 4]
```diff
@@ -171,7 +178,12 @@ impl Analyzer<'_> {
                     if let Some(declarator) = &state.declarator {
                         if declarator != item_id {
                             // A write also depends on the declaration.
-                            self.g.add_weak_deps(item_id, [declarator].iter().copied());
+                            if item.side_effects {
+                                self.g
+                                    .add_strong_deps(item_id, [declarator].iter().copied());
+                            } else {
+                                self.g.add_weak_deps(item_id, [declarator].iter().copied());
+                            }
```
bgw, line 186: The array should implement `IntoIterator<Item = &'a ItemId>`?

Apparently this was maybe an issue in earlier editions of rust prior to 2021? https://doc.rust-lang.org/std/primitive.array.html#editions

```suggestion
                                self.g
                                    .add_strong_deps(item_id, [declarator]);
                            } else {
                                self.g.add_weak_deps(item_id, [declarator]);
                            }
```

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -202,9 +214,14 @@ impl Analyzer<'_> {
                     // Optimization: Remove each module item to which we
                     // just created a strong dependency from LAST_WRITES
 
-                    state
-                        .last_writes
-                        .retain(|last_write| !self.g.has_dep(item_id, last_write, true));
+                    // TODO(kdy1): Re-enable this optimization
+                    // I disabled this optimization because it causes some issues while enabling
+                    // tree ahking in next.js
```
bgw, line 219:
```suggestion
                    // tree shaking in next.js
```

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -256,6 +273,14 @@ impl Analyzer<'_> {
 
                     let state = get_var(&self.vars, id);
                     self.g.add_strong_deps(item_id, state.last_writes.iter());
+
+                    if let Some(declarator) = &state.declarator {
+                        if declarator != item_id {
+                            // A read also depends on the declaration.
+                            self.g
+                                .add_strong_deps(item_id, [declarator].iter().copied());
```
bgw, line 281:
```suggestion
                                .add_strong_deps(item_id, [declarator]);
```

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -266,6 +291,14 @@ impl Analyzer<'_> {
                     let state = get_var(&self.vars, id);
 
                     self.g.add_weak_deps(item_id, state.last_reads.iter());
+
+                    if let Some(declarator) = &state.declarator {
+                        if declarator != item_id {
+                            // A write also depends on the declaration.
+                            self.g
+                                .add_strong_deps(item_id, [declarator].iter().copied());
```
bgw, line 299:
```suggestion
                                .add_strong_deps(item_id, [declarator]);
```

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -322,16 +355,60 @@ async fn get_part_id(result: &SplitResult, part: Vc<ModulePart>) -> Result<u32>
...
-        SplitResult::Ok { entrypoints, .. } => entrypoints,
-        _ => bail!("split failed"),
+    let SplitResult::Ok {
+        entrypoints,
+        modules,
+        ..
+    } = &result
+    else {
+        bail!("split failed")
     };
```
bgw, line 365: Outside of the context of this PR, but why do we implement our own enum type here instead of using something like `Result<Split, Vc<ParseResult>>`? If this used a more standard `Result` we could've used `.context()`.

bgw (changes requested):

crates/turbopack-ecmascript/src/tree_shake/cjs_finder.rs
```diff
@@ -1,11 +1,146 @@
...
+
+                // Skip sever actions
+                ModuleItem::Stmt(Stmt::Expr(ExprStmt {
+                    expr: box Expr::Lit(Lit::Str(Str { value, .. })),
+                    ..
+                })) => {
+                    if value == "use server" {
+                        return true;
+                    }
+                }
```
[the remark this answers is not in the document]
kdy1: This is a sort of fast-path that does not involve visitor

crates/turbopack-ecmascript/src/tree_shake/graph.rs
kdy1: `group` is modified in the loop so we have to track the initial length of the `group`

crates/turbopack-ecmascript/src/tree_shake/cjs_finder.rs
```diff
@@ -1,11 +1,146 @@
...
+            | "getStaticProps"
+    )
+}
+
+#[derive(Default)]
+struct AbortFinder {
+    abort: bool,
+}
+
+impl Visit for AbortFinder {
```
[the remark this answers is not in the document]
kdy1: `is_server_module` is something different from the requirement of this code. It does not use visitor for `use server`, but we need to check for `use server` in function declarations.

crates/turbopack-ecmascript/src/tree_shake/mod.rs
kdy1: We can't use `.context()` on `Result<T, E>` anyway if `E` is `Vc< ParseResult>`. `.context()` wants `E` to implement specific traits, and `Vc<ParseResult>` does not fit for `Result::Err` IMHO. It goes to `Ok` variant in most cases.

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -202,9 +213,14 @@ impl Analyzer<'_> {
...
-                        .last_writes
-                        .retain(|last_write| !self.g.has_dep(item_id, last_write, true));
+                    // TODO(kdy1): Re-enable this optimization
+                    // I disabled this optimization because it causes some issues while enabling
+                    // tree shaking in next.js
+                    //
+                    // state
+                    //     .last_writes
+                    //     .retain(|last_write| !self.g.has_dep(item_id,
+                    // last_write, true));
```
sokra, line 223: What is about that?

crates/turbopack-ecmascript/src/analyzer/imports.rs
```diff
@@ -339,7 +339,7 @@ impl Visit for Analyzer<'_> {
         let i = self.ensure_reference(
             export.span,
             export.src.value.clone(),
-            symbol.unwrap_or(ImportedSymbol::Namespace),
```
sokra, line 342: Could we remove `Namespace` from `ImportedSymbol` since it's not used?

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -139,6 +139,8 @@ pub struct EcmascriptOptions {
     /// If false, they will reference the whole directory. If true, they won't
     /// reference anything and lead to an runtime error instead.
     pub ignore_dynamic_requests: bool,
+
+    pub special_exports: Vc<Vec<RcStr>>,
```
sokra, line 143: Add a documentation comment explaining how this is used

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -367,8 +375,8 @@ impl EcmascriptModuleAsset {
     }
 
     #[turbo_tasks::function]
-    pub fn analyze(self: Vc<Self>) -> Vc<AnalyzeEcmascriptModuleResult> {
-        analyse_ecmascript_module(self, None)
+    pub async fn analyze(self: Vc<Self>) -> Result<Vc<AnalyzeEcmascriptModuleResult>> {
+        Ok(analyse_ecmascript_module(self, None))
```
sokra, line 379: unnecessary change. This doesn't need to be async

crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -239,9 +241,15 @@ impl EcmascriptModuleAssetBuilder {
...
 
     pub async fn build_part(self, part: Vc<ModulePart>) -> Result<Vc<EcmascriptModulePartAsset>> {
+        let special_exports = self.options.await?.special_exports;
         let import_externals = self.options.await?.import_externals;
         let base = self.build();
-        Ok(EcmascriptModulePartAsset::new(base, part, import_externals))
+        Ok(EcmascriptModulePartAsset::new(
+            base,
+            part,
+            special_exports,
```
sokra, line 250: We don't need to pass `special_exports` here. It's already available as `base.options.special_exports`

crates/turbopack-wasm/src/module_asset.rs
```diff
@@ -78,6 +79,12 @@ impl WebAssemblyModuleAsset {
             }))),
         ).module();
 
+        if let Some(esm_asset) =
+            Vc::try_resolve_downcast_type::<EcmascriptModulePartAsset>(module).await?
+        {
+            return Ok(esm_asset.await?.full_module);
+        }
+
```
sokra, line 87: That doesn't seem right. We should just return the part asset. You should be able to widen the return value of this function to remove the cast below

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -322,16 +347,60 @@ async fn get_part_id(result: &SplitResult, part: Vc<ModulePart>) -> Result<u32>
...
+        modules,
+        ..
+    } = &result
+    else {
+        bail!("split failed")
     };
 
     let part_id = match entrypoints.get(&key) {
-        Some(id) => *id,
+        Some(&id) => id,
```
sokra, line 360:
``` rs
if let Some(id) = entrypoints.get(&key) {
    return Ok(id);
}
```

could avoid some nesting for the following code

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -371,15 +443,24 @@ impl PartialEq for SplitResult {
 }
 
 #[turbo_tasks::function]
-pub(super) async fn split_module(asset: Vc<EcmascriptModuleAsset>) -> Result<Vc<SplitResult>> {
-    Ok(split(asset.source().ident(), asset.source(), asset.parse()))
+pub(super) async fn split_module(
+    asset: Vc<EcmascriptModuleAsset>,
+    special_exports: Vc<Vec<RcStr>>,
```
sokra, line 448: Unnecessary passing `special_exports`. This is also stored in `asset`

crates/turbopack-ecmascript/src/tree_shake/cjs_finder.rs
sokra, line 1: Maybe rename this file to `utils`. `cjs_finder` is no longer appropriate

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -552,12 +635,15 @@ pub(super) async fn part_of_module(
                             },
                         )));
 
+                    module.body.extend(star_reexports.iter().map(|export_all| {
+                        ModuleItem::ModuleDecl(ModuleDecl::ExportAll(export_all.clone()))
+                    }));
```
sokra, line 640: I think the `Facade` should reexport from the `Exports` here.

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -322,16 +347,60 @@ async fn get_part_id(result: &SplitResult, part: Vc<ModulePart>) -> Result<u32>
...
 
     let part_id = match entrypoints.get(&key) {
-        Some(id) => *id,
+        Some(&id) => id,
         None => {
+            // This is required to handle `export * from 'foo'`
+            if let ModulePart::Export(..) = &*part {
+                if let Some(&v) = entrypoints.get(&Key::ModuleEvaluation) {
+                    return Ok(v);
+                }
```
sokra, line 366: I feel like this should return `ModulePart::Exports` instead.

bgw (approved):
Thanks for making this big change!

In the future to help make reviewing these PRs easier, it might be useful to squash some of the commits locally before force pushing or putting it up for review. Github (and Graphite) lists every commit in the conversation log, and it adds a lot of noise when there's 152 commits.

PR #8549, opened 2024-06-20
fix(turbopack): add no-op `resolveAbsolutePath` to browser runtime

ForsakenHarmony:
### Description

Reading `import.meta.hot` should work in the browser, it can just return a `/ROOT/...` path.


crates/turbopack-ecmascript-runtime/js/src/browser/dev/runtime/base/runtime-base.ts
```diff
@@ -386,6 +387,17 @@ function instantiateModule(id: ModuleId, source: SourceInfo): Module {
...
 
+/**
+ * no-op for browser
+ * @param modulePath
+ */
+function resolveAbsolutePath(modulePath?: string): string {
+  if (modulePath) {
+    return modulePath;
+  }
+  return "/ROOT/";
```
sokra, line 398: Shoudn't it return something like:

```suggestion
  return `/ROOT/${modulePath || ''}`;
```

sokra (changes requested):
Could you add a test case testing this behavior?

crates/turbopack-ecmascript-runtime/js/src/browser/dev/runtime/base/runtime-base.ts
ForsakenHarmony: Yes, initially thought it was added by codegen

ForsakenHarmony:
@sokra I don't think we have a way to do an execution test with the browser runtime, do we?

sokra (approved):

PR #8550, opened 2024-06-20
fix(turborepo): add berry support for built meta when pruning

vinnymac:
### Description

Add support for [yarn berry built option](https://yarnpkg.com/configuration/manifest#dependenciesMeta.built) for dependencies meta to `turbo prune`.

Without this change, `yarn install --immutable` fails because `built` is added after turbo accidentally removes it, triggering the following error.

> The lockfile would have been modified by this install, which is explicitly forbidden.

### Testing Instructions

* Add a dependenciesMeta for a package and set it to `built: false`.
* Run yarn to generate a new lockfile
* Use turbo prune to produce a slim lockfile
* Run `yarn install --immutable` on that small lockfile

### DIFF
--- a/crates/turborepo-lockfiles/src/berry/ser.rs
+++ b/crates/turborepo-lockfiles/src/berry/ser.rs
@@ -157,9 +157,9 @@ where
         for (i, (setting, field)) in settings.iter().enumerate() {
             if let Some(value) = setting {
                 string.push_str(&format!("      {}: {}", wrap_string(field), value));
-            }
-            if i < settings.len() - 1 {
-                string.push('\n');
+                if i < settings.len() - 1 {
+                    string.push('\n');
+                }
             }
         }
 
--- a/crates/turborepo-lockfiles/src/berry/ser.rs
+++ b/crates/turborepo-lockfiles/src/berry/ser.rs
@@ -147,31 +147,34 @@ where
     let mut string = String::new();
     let mut first = true;
 
-    let mut add_line = |dependency: &str, field: &str, setting: bool| {
+    let mut add_line = |dependency: &str, settings: &[(Option<bool>, &str)]| {
         if !first {
             string.push('\n');
         }
 
-        string.push_str(&format!(
-            "    {}:\n      {}: {}",
-            wrap_string(dependency),
-            wrap_string(field),
-            setting,
-        ));
+        string.push_str(&format!("    {}:\n", wrap_string(dependency)));
+
+        for (i, (setting, field)) in settings.iter().enumerate() {
+            if let Some(value) = setting {
+                string.push_str(&format!("      {}: {}", wrap_string(field), value));
+            }
+            if i < settings.len() - 1 {
+                string.push('\n');
+            }
+        }
 
         first = false;
     };
 
     for (dependency, meta) in metadata {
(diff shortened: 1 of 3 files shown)


chris-olszewski:
Thanks for the PR! I think this looks good to me, but I'll be doing some additional testing just because `yarn install --immutable` is quite strict.

chris-olszewski (approved):
Thank you for the PR!

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: a395085d10f6e22a91270ebec8b821c093a2fcab

crates/turborepo-lockfiles/src/berry/ser.rs
```diff
@@ -163,6 +163,9 @@ where
 
     for (dependency, meta) in metadata {
         let dependency = dependency.as_ref();
+        if meta.built.unwrap_or_default() {
```
chris-olszewski, line 166: Sorry I missed this, I believe this logic needs to be changed so we only include `built` if it is present. `add_line` will also need to be updated to be able to write both true and false.
```suggestion
        if meta.built.unwrap_or_default() {
```

chris-olszewski (changes requested):

crates/turborepo-lockfiles/src/berry/ser.rs
```diff
@@ -147,27 +147,31 @@ where
...
-            add_line(dependency, "optional");
+        if let Some(built) = meta.built {
+            add_line(dependency, "built", built);
         }
-        if meta.unplugged.unwrap_or_default() {
-            add_line(dependency, "unplugged");
+        if let Some(optional) = meta.optional {
+            add_line(dependency, "optional", optional);
+        }
+        if let Some(unplugged) = meta.unplugged {
```
vinnymac, line 173: After updating this for `built`, I assumed someone may set optional or unplugged to false, and then that too would become an issue.

crates/turborepo-lockfiles/src/berry/ser.rs
vinnymac: I updated it so all three dependencyMeta settings support true and false. And check if they are present before adding the lines.

Let me know if you find any more 🐛 s. 🙇🏼‍♂️

vinnymac:
@chris-olszewski fixed an issue with this change where I somehow forgot to support more than one dependency meta setting for a single dependency 😅. This should do the job now, let me know what you think.

anthonyshew:
@vinnymac @chris-olszewski Just want to check in with you both to see if we want to finish up this PR? I'm doing some housekeeping to close out stale items in the repo (but would love to see this get merged!)

vinnymac:
> @vinnymac @chris-olszewski Just want to check in with you both to see if we want to finish up this PR? I'm doing some housekeeping to close out stale items in the repo (but would love to see this get merged!)

This is in a good place, just needs a review first so it can move forward. @chris-olszewski did you want any changes here?

chris-olszewski:
I needed to manually cherrypick the commits to https://github.com/vercel/turborepo/pull/9605 as for whatever reason trying to rebase past https://github.com/vercel/turborepo/pull/8813 was causing `git` to fail.

Thank you for your contribution and sorry for the delay on getting this merged.

PR #8570, opened 2024-06-21
feat(tui): output tasks with errors last

chris-olszewski:
### Description

This PR changes the logic of persisting task output so we now print tasks in the following order:
 - Tasks that finished successfully in order of finish
 - Tasks that started, but did not finish in order of start
 - Tasks that finished with non-zero exit code in order of finish

This is a change from previous behavior where they would be output alphabetically

### Testing Instructions

Manual testing on a repo with a failure that runs early:

Screenshot 2024-06-21 at 2 59 14 PM


crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -169,7 +168,6 @@ impl<W> TerminalOutput<W> {
             stdin,
             rows,
             cols,
-            has_been_persisted: false,
```
chris-olszewski, line 172: This was only needed when we would persist tasks while the TUI was still running. We were only persisting via the `render_remaining` so this was never `true` from a previous call to `persist_screen`

anthonyshew (approved):
Love it, good get.

PR #8602, opened 2024-06-25
devlow-bench: add git branch and sha to datapoints

wbinnssmith:
This includes this information to datapoints sent to Datadog. Also addresses an issue where trailing newlines were added when these values were not read from GitHub Actions.


bgw (approved):
LGTM. Might also be nice to track if the checkout is dirty, but that'd just be gravy.

[surewhynot](https://app.graphite.dev//settings/meme-library?org=vercel)

PR #8604, opened 2024-06-25
Delete unused TASK_ID_MAPPING support

bgw:
This is currently unused. It existed for persistent caching, but it sounds like we likely won't go with this mapping approach:

> It was used for persistent caching, but it looks like we probably take a different approach (same ids between persistent cache and memory cache). So you can remove it.

-- https://vercel.slack.com/archives/C03EWR7LGEN/p1719340628045969?thread_ts=1719338013.257509&cid=C03EWR7LGEN

### DIFF
--- a/crates/turbo-tasks/src/id.rs
+++ b/crates/turbo-tasks/src/id.rs
@@ -1,7 +1,6 @@
 use std::{
-    cell::RefCell,
     fmt::{Debug, Display},
-    mem::{transmute_copy, ManuallyDrop},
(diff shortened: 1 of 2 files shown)


sokra (approved):

bgw:
### Merge activity

* **Jun 26, 3:18 AM EDT**: @bgw merged this pull request with [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/8604).

Issue #8383, opened 2024-06-07
If remote cache server returns error than ui mode gets messed up


weyert:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

n/a

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

2.0.3

### Describe the Bug

When you have a custom remote cache server which happens to return an error, e.g. 404 or 401, 502. 

The UI mode display this response body of the failed http request in yellow and then continues running the task without clearing screen first. Now causing the ui, e.g. logs and the task sidebar to be rendered on top of the response body.

### Expected Behavior

No messing up the ui screen

### To Reproduce

Run `pnpm run build` and make sure the the remote server returns bad response, e.g. nice html content with an error

frederik-oestergaard:
Also experiencing this. 
Example:
![image](https://github.com/vercel/turbo/assets/78544875/5e4fa024-4c04-4ed3-8665-c07c0859f59d)

PR #8607, opened 2024-06-26
chore: upgrade failure to check caches to a warning

mehulkar:


paulogdm (approved):
It appears you may want to remove that unused import tho.

chris-olszewski (commented):
This will cause more TUI corruption until https://github.com/vercel/turbo/issues/8383 is addressed

mehulkar:
I'm not partial to this one, I think I even remember that it was moved to debug for a reason at some point. Closing!

PR #8608, opened 2024-06-26
chore(turbo): add debug logs for changes files when using a git range filter

mehulkar:
I spent a little too long debugging this with local builds and some more logs here could have helped!


crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -238,6 +238,7 @@ impl<'a, T: GitChangeDetector> FilterResolver<'a, T> {
             selectors.into_iter().partition::<Vec<_>, _>(|t| !t.exclude);
 
         let mut include = if !include_selectors.is_empty() {
+            // TODO: add telemetry for each selector
```
mehulkar, line 241: just a placeholder for myself. Once we've parsed all the `--filter` flags, I'd really like to see more data on how people are using this (without any data for the values passed to the flag, just the "types" of filter)

chris-olszewski (approved):

Issue #8553, opened 2024-06-20
Turbo saving caches in incorrect location


HaveSpacesuit:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

​

### What package manager are you using / does the bug impact?

npm

### What operating system are you using?

Windows

### Which canary version will you have in your reproduction?

2.0.5-canary.0

### Describe the Bug

Turbo is saving caches in the source code folder instead of on the machine cache location. Turbo used to save cached output (e.g., {6AF0698E-D558-4F6E-9B3C-3716689AF493}.2.ver0x0000000000000021.db) under **C:/Windows/ProgramData/Microsoft/Windows/Caches**. After updating from v1 to v2, it saves in the source code location, e.g., **D:/MyRepo/website/%SystemDrive%/ProgramData/Microsoft/Windows/Caches**. It seems the `%SystemDrive%` variable isn't being escaped in some script.

### Expected Behavior

I do not want cached files in my source code.

### To Reproduce

Install Turbo v2 and run a script with cached output. The files will appear in a location including the literal `%SystemDrive%` in the path.

chris-olszewski:
Can either of you provide a reproduction? Can you also include a Windows build number? I'm unable to reproduce on my Windows machine.

IvGaLa:
> Can either of you provide a reproduction? Can you also include a Windows build number? I'm unable to reproduce on my Windows machine.

Of course, here I provide you with some more information:

**OS:** 
_Windows 10 Pro Edition
Version 22H2
OS Build 19045.4529
Windows Feature Experience Pack 1000.19058.1000.0_


**VSCode:**
_Version: 1.90.2 (user setup)
Commit: 5437499feb04f7a586f677b155b039bc2b3669eb
Date: 2024-06-18T22:34:26.404Z
Electron: 29.4.0
ElectronBuildId: 9728852
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Windows_NT x64 10.0.19045_


**npm:**
_v10.8.1_


**turbo:**
_reactjs@1.0.0 S:\Development\ReactJS
└── turbo@2.0.4_

chris-olszewski:
I should have asked this before, but can you try running with `--env-mode=loose` and see if the cache saves to the correct place?

IvGaLa:
[...]

PR #8615, opened 2024-06-26
fix: add more windows vars to default pass through env

chris-olszewski:
### Description

Fixes https://github.com/vercel/turbo/issues/8553

### Testing Instructions

User confirmed that adding `SystemRoot` to pass through fixed his issue: https://github.com/vercel/turbo/issues/8553#issuecomment-2194773860


anthonyshew (approved):
We love Windows, yes we do,
We love Windows, how bout you.

PR #8624, opened 2024-06-28
Task Edges Set/List

sokra:
### Description

* Task children and dependencies are stored together in a single datastructure -> task edges
* task edges is a map keyed by target task id and using a bitfield for common edges (child, output, cell 0)
* removed lazy_remove_children feature flag -> this is always on now (Disabling it doesn't make sense as it negatively affects performance, we had the feature flag in case it it is broken, but seems fine)
* When a task becomes Done, we use a Vec instead of a HashMap for better memory efficiency. It doesn't need to modified after that.


crates/turbo-tasks-memory/src/multi_iter.rs
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        match self {
+            IntoIters2::One(iter) => iter.size_hint(),
+            IntoIters2::Two(iter) => iter.size_hint(),
+        }
+    }
+}
```
bgw, line 86: The either crate can actually do this. `Either<A, B>` is basically the same as `IntoIters2<A, B>`: https://docs.rs/either/latest/either/enum.Either.html#impl-Iterator-for-Either%3CL,+R%3E

Also, while it might be a very minor optimization not to nest these, you totally can, and with monomorphization, the nested functions might even get inlined:

`IntoIters4<A, B, C, D>` can just be `Either<Either<A, B>, Either<C, D>>`.

In some cases, `Either` might even be better because it also implements `ExactSizeIterator` and `DoubleEndedIterator` if the types it contains do.

If this is really and truly a super hot codepath, we should use a custom `Iterator` impl specific to `EdgesEntry`, avoiding the intermediate array iterators.

crates/turbo-tasks-memory/src/edges_set.rs
```diff
@@ -0,0 +1,592 @@
...
+                }
+            }
+            EdgeEntry::Collectibles(trait_type_id) => {
+                if let EdgesEntry::Complex(set) = self {
+                    set.contains(&EdgeEntry::Collectibles(trait_type_id))
+                } else {
+                    false
+                }
+            }
+        }
```
bgw, line 237: I think this could be written more simply without the nested match macros by using a match against a tuple of `(entry, self)`

```suggestion
        match (entry, self) {
            (
                EdgeEntry::Output,
                EdgesEntry::Output
                | EdgesEntry::OutputAndCell0(_)
                | EdgesEntry::ChildAndOutput
                | EdgesEntry::ChildOutputAndCell0(_)
            ) => true,
            (
                EdgeEntry::Child,
                EdgesEntry::Child
                | EdgesEntry::ChildAndOutput
                | EdgesEntry::ChildAndCell0(_)
                | EdgesEntry::ChildOutputAndCell0(_)
            ) => true,
            (
                EdgeEntry::Cell(cell_id),
                EdgesEntry::Cell0(type_id)
                | EdgesEntry::OutputAndCell0(type_id)
                | EdgesEntry::ChildAndCell0(type_id)
                | EdgesEntry::ChildOutputAndCell0(type_id)
            ) => *type_id == cell_id.type_id,
            (entry, EdgesEntry::Complex(set)) => set.contains(entry),
            _ => false,
        }
```

crates/turbo-tasks-memory/src/edges_set.rs
```diff
@@ -0,0 +1,592 @@
...
+                let mut set = AutoSet::default();
+                set.insert(EdgeEntry::Collectibles(trait_type_id));
+                EdgesEntry::Complex(Box::new(set))
+            }
+        }
+    }
+
+    fn len(&self) -> usize {
+        match self {
+            EdgesEntry::Empty => unreachable!(),
```
bgw, line 88: While this may not happen in practice, I think it would still be reasonable to implement this as 0.

crates/turbo-tasks-memory/src/edges_set.rs
```diff
@@ -0,0 +1,592 @@
...
+        false
+    }
+
+    fn shrink_to_fit(&mut self) {
+        if let EdgesEntry::Complex(set) = self {
+            set.shrink_to_fit();
+        }
+    }
+
+    fn simplify(&mut self) {
```
bgw, line 383: - Shouldn't this handle cases where len() is 2 or 3? Those cases aren't guaranteed to simplfiy, but might.
- Could this be implemented by starting with `EdgesEntry::Empty` and rebuilding the set by calling `insert` as many times as needed?

crates/turbo-tasks-memory/src/edges_set.rs
```diff
@@ -0,0 +1,592 @@
...
+                self.simplify();
+                return true;
+            }
+        }
+        false
+    }
+
+    fn shrink_to_fit(&mut self) {
+        if let EdgesEntry::Complex(set) = self {
+            set.shrink_to_fit();
```
bgw, line 379: If feel like this should also call simplify before trying to call `set.strink_to_fit`.

crates/turbo-tasks-memory/src/edges_set.rs
```diff
@@ -0,0 +1,592 @@
...
+                }
+                _ => (),
+            }
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct TaskEdgesSet {
+    edges: Box<AutoMap<TaskId, EdgesEntry, BuildHasherDefault<FxHasher>>>,
```
bgw, line 408: I don't think you should box this yourself. You should leave it up to the caller whether or not they want to box `TaskEdgesSet`.

crates/turbo-tasks-memory/src/edges_set.rs
```diff
@@ -0,0 +1,592 @@
...
+        self.edges.shrink_to_fit();
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.edges.is_empty()
+    }
+
+    pub fn len(&self) -> usize {
+        self.edges.iter().map(|(_, entry)| entry.len()).sum()
+    }
```
bgw, line 445: I don't really like a `len()` operation that takes O(n) time as it's unintiutive. Callers usually won't feel bad about calling `.len()` inside of a loop because they'll assume it's O(1). Either we should:

- Iteratively keep track of the length as elements are added and removed
- OR rename this to something that makes the cost more obvious, like `compute_len`.

crates/turbo-tasks-memory/src/edges_set.rs
```diff
@@ -0,0 +1,592 @@
...
+pub struct TaskEdgesList {
+    edges: Box<[(TaskId, EdgesEntry)]>,
+}
+
+impl TaskEdgesList {
+    pub fn into_set(self) -> TaskEdgesSet {
+        TaskEdgesSet {
+            edges: Box::new(self.edges.into_vec().into_iter().collect()),
+        }
+    }
```
bgw, line 558: How common is this operation? I bet if we were clever we could do an optimization for the case where we're converting into a `SmallVec` that avoids the iterator.

bgw (commented):
I read through the edges set implementation, but haven't gotten to reviewing how it's used yet.

crates/turbo-tasks-auto-hash-map/src/map.rs
```diff
@@ -213,6 +213,36 @@ impl<K: Eq + Hash, V, H: BuildHasher + Default, const I: usize> AutoMap<K, V, H,
...
+                    }
+                }
+            }
+            AutoMap::Map(map) => {
+                map.retain(f);
+                if map.len() <= MIN_HASH_SIZE {
+                    let mut list = SmallVec::with_capacity(map.len());
+                    list.extend(map.drain());
+                    *self = AutoMap::List(list);
+                }
```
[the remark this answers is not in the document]
sokra: Yeah makes sense

crates/turbo-tasks-memory/src/edges_set.rs
```diff
@@ -0,0 +1,592 @@
...
+
+use crate::multi_iter::IntoIters4;
+
+#[derive(Hash, Copy, Clone, PartialEq, Eq)]
+pub enum TaskEdge {
+    Output(TaskId),
+    Cell(TaskId, CellId),
+    Collectibles(TaskId, TraitTypeId),
+    Child(TaskId),
+}
```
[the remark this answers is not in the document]
sokra: `TaskEdge` is the public type,  `EdgeEntry` is an implementation details. I would rather keep it this way

crates/turbo-tasks-memory/src/edges_set.rs
```diff
@@ -0,0 +1,592 @@
...
+        match self {
+            EdgeEntry::Output => TaskEdge::Output(task),
+            EdgeEntry::Cell(cell_id) => TaskEdge::Cell(task, cell_id),
+            EdgeEntry::Collectibles(trait_type_id) => TaskEdge::Collectibles(task, trait_type_id),
+            EdgeEntry::Child => TaskEdge::Child(task),
+        }
+    }
+}
+
+type ComplexSet = AutoSet<EdgeEntry, BuildHasherDefault<FxHasher>, 3>;
```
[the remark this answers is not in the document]
sokra: I changed it to 9 which would be 128 bytes size for ComplexSet.

crates/turbo-tasks-memory/src/edges_set.rs
sokra: `EdgesEntry::Empty` indicates a bug in a previous operations. I rather like it to panic early.

crates/turbo-tasks-memory/src/edges_set.rs
sokra: It unused, so I removed it

crates/turbo-tasks-memory/src/task.rs
```diff
@@ -364,16 +352,73 @@ enum TaskStateType {
...
+            TaskStateType::InProgressDirty { outdated_edges, .. } => {
+                IntoIters3::Three(outdated_edges.children())
+            }
+            TaskStateType::Scheduled { outdated_edges, .. } => {
+                IntoIters3::Three(outdated_edges.children())
+            }
+        }
+    }
+
+    fn into_dependencies_and_children(self) -> (TaskEdgesSet, SmallVec<[TaskId; 64]>) {
```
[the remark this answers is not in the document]
sokra: done

crates/turbo-tasks-memory/src/edges_set.rs
sokra: Every operation that removes items already calls `simplify`, so we don't need to call it here.

crates/turbo-tasks-memory/src/edges_set.rs
sokra: done

crates/turbo-tasks-memory/src/multi_iter.rs
sokra: done

kdy1 (approved):

bgw:
Thanks for making the suggested changes! The `edge_set` implementation LGTM!

PR #8637, opened 2024-06-29
docs(turborepo): Update managing dependency for yarn berry

vinnymac:
### Description

I noticed that the documentation mentions Yarn 2+, but doesn't explain the correct way to install dependencies when using that package manager. It's well documented [here](https://yarnpkg.com/cli/workspaces/foreach#usage), so I went ahead and updated the documentation to reflect the correct approach.

### Testing Instructions

Navigate to the managing dependencies page of the documentation, and confirm the new documentation looks good and reads well.


anthonyshew (approved):
Great addition! Thank you!

PR #8638, opened 2024-06-30
fix(examples): correct next.config extension in tsconfig

moolcoov:
### Description

Change next.config extension in `tsconfig.json` from `next.config.js` to `next.config.mjs` because the file actually is `next.config.mjs`


anthonyshew (approved):
Good find! Thanks!

PR #8642, opened 2024-07-01
fix(docs): fix broken link

ony3000:
### Description

While reading the "crafting-your-repository/configuring-tasks" page, I noticed that one link was broken, so I examined the markdown written in `[text](#link)` format.

In addition, the `Interacting with tasks` item on the "crafting-your-repository/developing-applications" page also appears to have a broken link, but I left it as is because I didn't know where to send it to.


arlyon (approved):
Ping @anthonyshew

anthonyshew (approved):
Thank you!

PR #8650, opened 2024-07-02
(refactor)ui: Refactoring/simplifying TUI state

anthonyshew:
### Description

Previously, we held the state for the TUI in the sub-components of `app.rs`. In this PR, we're hoisting those state variables up so we can have better access to that shared state and create more robust UI.

### Testing Instructions

There two behavioral changes in this PR: 

- Reordering of the task list so that running tasks are first in the list, then planned tasks, then done tasks.
- We are removing the spinner in favor of the "two arrow" indicator. Users mentioned that the visual noise was more annoying than helpful, so let's quiet it down.

Beyond that, you should see no behavioral changes.


crates/turborepo-ui/src/tui/app.rs
```diff
@@ -118,10 +282,15 @@ pub fn run_app(tasks: Vec<String>, receiver: AppReceiver) -> Result<(), Error> {
     let ratio_pane_width = (f32::from(size.width) * PANE_SIZE_RATIO) as u16;
     let full_task_width = size.width.saturating_sub(task_width_hint);
 
-    let mut app: App<Box<dyn io::Write + Send>> =
-        App::new(size.height, full_task_width.max(ratio_pane_width), tasks);
+    let mut app: App<Box<dyn io::Write + Send>> = App::new(size.height, size.width, tasks);
```
chris-olszewski, line 285: I don't believe this was an intentional change.
```suggestion
    let mut app: App<Box<dyn io::Write + Send>> = App::new(size.height, full_task_width.max(ratio_pane_width), tasks);
```

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -7,102 +8,265 @@ use std::{
...
-    pane: TerminalPane<I>,
-    done: bool,
+pub enum LayoutSections {
+    Pane,
+    TaskList,
+}
+
+pub struct App<W> {
+    // TODO: Could I reasonably get to only one representation of the tasks?
+    tasks: BTreeMap<String, TerminalOutput<W>>,
```
chris-olszewski, line 34: There is the possibility here, but that should be punted on. 

We would need a way to track order of the tasks which is easiest to due with just having a list structure. Due to the frequent insert/removals which result in lots of shifting of elements we want to keep whatever is in the list on the small side to keep the size of `memcpy` calls small. Putting the whole `TermialOutput` in the list representation would be a perf hit.

In the future we might find the proper data structure that allows us to do this cheap *or* we end up assigning each task an integer id and pass those around.

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -226,14 +397,17 @@ fn update(
 ) -> Result<Option<mpsc::SyncSender<()>>, Error> {
     match event {
         Event::StartTask { task } => {
-            app.table.start_task(&task)?;
-            app.started_tasks.push(task);
+            app.start_task(&task)?;
+            app.tasks_by_status
+                .groups_as_task_names()
+                .running
+                .push(task);
```
chris-olszewski, line 404: This isn't doing what I think you mean for it to do.

`groups_as_task_names` is constructing a new structure that is getting the new task in `running`, but that structure immediately goes out of scope and is dropped.

anthonyshew: Yikes, definitely not.

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -7,102 +8,265 @@ use std::{
...
+            self.input_options.focus = LayoutSections::Pane;
+        }
     }
 
     pub fn update_tasks(&mut self, tasks: Vec<String>) {
-        self.table = TaskTable::new(tasks.clone());
-        self.next();
+        let mut task_list = tasks.into_iter().map(Task::new).collect::<Vec<_>>();
+        task_list.sort_unstable();
+        task_list.dedup();
```
chris-olszewski, line 224: This function needs to get finished. We're not saving the new list anywhere. We should be:
 - Overwriting `task_by_status` for the list of tasks
 - Make sure all new task names have an entry in `tasks` so we don't crash when they produce output

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -137,18 +306,20 @@ fn run_app_inner<B: Backend + std::io::Write>(
     terminal: &mut Terminal<B>,
     app: &mut App<Box<dyn io::Write + Send>>,
     receiver: AppReceiver,
+    rows: u16,
+    cols: u16,
```
chris-olszewski, line 310: Not necessary for this PR, but I think we'll want these bits of info to end up as part of the `App` state. 

It's an implicit part of the `TerminalOutput` and if these values change from what `TerminalOutput` was constructed with, we need to make sure all terminals are resized. Currently these values don't change, but once we tackle handling terminal resizes they will.

crates/turborepo-ui/src/tui/input.rs
```diff
@@ -2,19 +2,15 @@ use std::time::Duration;
 
 use crossterm::event::{KeyCode, KeyEvent, KeyEventKind, KeyModifiers};
 
-use super::{event::Event, Error};
+use super::{app::LayoutSections, event::Event, Error};
 
-#[derive(Debug, Clone, Copy)]
```
chris-olszewski, line 7: We should add these back and change the functions in the file to take ownership instead of references. It will require adding `#[derive(Debug, Clone, Copy)]` to the definition of `LayoutSections`

crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -7,50 +7,40 @@ use ratatui::{
...
-    stdin: Option<W>,
-    status: Option<String>,
-}
-
-impl<W> TerminalPane<W> {
-    pub fn new(rows: u16, cols: u16, tasks: impl IntoIterator<Item = String>) -> Self {
+impl<'a, W> TerminalPane<'a, W> {
+    pub fn new(
+        rows: u16,
+        cols: u16,
```
chris-olszewski, line 27: We can drop the rows,cols from this constructor. The resize logic is currently unused and that's the only place these are used (along with the getter `term_size` which is also unused). 

I'm in favor for canning them so we don't have a false sense that they're important/used in their current state.

crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -59,6 +49,13 @@ impl<W> TerminalPane<W> {
         Ok(())
     }
 
+    pub fn hassss_stdin(&self, task: &str) -> bool {
+        self.tasks
+            .get(task)
+            .map(|task| task.stdin.is_some())
+            .unwrap_or_default()
+    }
```
chris-olszewski, line 57: ~~I think this can get lifted to `App` since that's where the data resides.~~ Already was, we can just can this defn and the one below it.

crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -130,8 +125,7 @@ impl<W> TerminalPane<W> {
     }
 
     fn selected(&self) -> Option<(&String, &TerminalOutput<W>)> {
-        let task_name = self.displayed.as_deref()?;
-        self.tasks.get_key_value(task_name)
+        self.tasks.get_key_value(self.displayed_task)
     }
```
chris-olszewski, line 129: This can get removed if pass in the just the selected `TerminalOutput` when constructing `TerminalPane`

crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -91,7 +86,7 @@ impl<W> TerminalPane<W> {
             let terminal = self.task_mut(task)?;
             terminal.resize(rows, cols);
         }
-        self.displayed = Some(task.into());
+        self.displayed_task;
```
chris-olszewski, line 89: We can delete this function

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -255,35 +429,236 @@ fn update(
...
     Ok(None)
 }
 
-fn view<I>(app: &mut App<I>, f: &mut Frame) {
-    let (_, width) = app.term_size();
-    let vertical = Layout::horizontal([Constraint::Fill(1), Constraint::Length(width)]);
-    let [table, pane] = vertical.areas(f.size());
-    app.table.stateful_render(f, table);
-    f.render_widget(&app.pane, pane);
+fn view(app: &mut App<Box<dyn io::Write + Send>>, f: &mut Frame, rows: u16, cols: u16) {
```
chris-olszewski, line 469: We don't need to specify an exact underlying writer type here, we can render an app regardless of what sort of stdin handles it has.

```suggestion
fn view<W>(app: &mut App<W>, f: &mut Frame, rows: u16, cols: u16) {
```

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -255,35 +429,236 @@ fn update(
...
+    let horizontal = Layout::horizontal([Constraint::Fill(1), Constraint::Length(cols)]);
+    let [table, pane] = horizontal.areas(f.size());
+
+    let active_task = app.active_task();
+
+    let pane_to_render: TerminalPane<Box<dyn io::Write + Send>> = TerminalPane::new(
```
chris-olszewski, line 475: We can omit the type hint here since `TerminalPane<W>` will have the same `W` as contained in `App`
```suggestion
    let pane_to_render = TerminalPane::new(
```

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -255,35 +429,236 @@ fn update(
             app.next();
         }
         Event::ScrollUp => {
-            app.scroll(Direction::Up);
+            app.has_user_scrolled = true;
+            app.tasks
+                .get_mut(&app.active_task())
+                .unwrap()
+                .scroll(Direction::Up)
+                .unwrap_or_default();
```
chris-olszewski, line 437: These feel like they should belong in a function on `App`. This allows us to test the state change via unit tests without setting up a channel to send events over. Instead we can just call `app.scroll(Director::Up)` and inspect them.

crates/turborepo-ui/src/tui/app.rs
anthonyshew: Removed entirely.

crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -7,50 +7,40 @@ use ratatui::{
...
+use super::{app::Direction, Error, TerminalOutput};
 
 const FOOTER_TEXT_ACTIVE: &str = "Press`Ctrl-Z` to stop interacting.";
 const FOOTER_TEXT_INACTIVE: &str = "Press `Enter` to interact.";
 
-pub struct TerminalPane<W> {
-    tasks: BTreeMap<String, TerminalOutput<W>>,
-    displayed: Option<String>,
+pub struct TerminalPane<'a, W> {
+    displayed_task: &'a String,
```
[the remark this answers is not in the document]
anthonyshew: Interesting, muchas gracias.

crates/turborepo-ui/src/tui/app.rs
anthonyshew: In realizing I didn't delete any of the unneeded code in `TerminalPane` I ended up getting rid of this. From my rudimentary hand-testing, I'm not seeing any problems but let me know if you find something.

crates/turborepo-ui/src/tui/task.rs
```diff
@@ -108,32 +108,19 @@ impl TasksByStatus {
...
-            .map(|task| task.name().to_string())
-            .collect::<Vec<_>>();
-
-        [
-            running_names.as_slice(),
-            planned_names.as_slice(),
-            finished_names.as_slice(),
-        ]
-        .concat()
+    pub fn task_names_in_displayed_order(&self) -> impl Iterator<Item = &str> + '_ {
```
chris-olszewski, line 114: This change means the function now returns an iterator over all task names instead of an vector of them. Beneficial to us as it results in far fewer allocations (4 vector allocations + 1 allocation per task) to zero allocations.

crates/turborepo-ui/src/tui/task.rs
```diff
@@ -108,32 +108,19 @@ impl TasksByStatus {
...
+        let running_names = self.running.iter().map(|task| task.name());
+        let planned_names = self.planned.iter().map(|task| task.name());
+        let finished_names = self.finished.iter().map(|task| task.name());
+
+        running_names.chain(planned_names).chain(finished_names)
+    }
+
+    pub fn task_name(&self, index: usize) -> &str {
```
chris-olszewski, line 122: Helper method to avoid the pattern `&app.task_by_status.task_names_in_displayed_order(index)` that was appearing in `app.rs`.

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -137,8 +137,10 @@ impl<W> App<W> {
     pub fn start_task(&mut self, task: &str) -> Result<(), Error> {
         // Name of currently highlighted task.
         // We will use this after the order switches.
-        let highlighted_task =
-            &self.tasks_by_status.task_names_in_displayed_order()[self.selected_task_index];
+        let highlighted_task = self
+            .tasks_by_status
+            .task_name(self.selected_task_index)
+            .to_string();
```
chris-olszewski, line 143: We need to clone this as otherwise we end up holding an immutable reference to `App` which would prevent us from mutating `App` to move the task from planned into starting.

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -243,10 +466,10 @@ fn update(
             return Ok(Some(callback));
         }
         Event::Tick => {
-            app.table.tick();
+            // app.table.tick();
```
chris-olszewski, line 469: Can remove this comment completely

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -255,35 +478,270 @@ fn update(
...
             app.forward_input(&bytes)?;
         }
         Event::SetStdin { task, stdin } => {
-            app.pane.insert_stdin(&task, Some(stdin))?;
+            app.insert_stdin(&task, Some(stdin))?;
         }
         Event::UpdateTasks { tasks } => {
             app.update_tasks(tasks);
-            app.table.tick();
+            // app.table.tick();
```
chris-olszewski, line 504: Can also remove this comment completely

chris-olszewski (approved):
LFGTM

I think this is a great move for unifying the state associated separating it from the rendering logic.

crates/turborepo-ui/src/tui/task.rs
anthonyshew: Nice.

PR #8655, opened 2024-07-03
Turbopack + Relay: correctly implement artifact_directory to fix multi-project test

wbinnssmith:
Previously, if `artifact_directory` were set, artifacts would be resolved at a fixed location relative to each file. This correctly resolves them from the project root.

Test Plan: With vercel/next.js#67403, `TURBOPACK_BUILD=1 TURBOPACK=1 pnpm test-dev test/integration/relay-graphql-swc-multi-project/test/index.test.js`


bgw (approved):
Glad to see that there are tests in the nextjs PR!

PR #8657, opened 2024-07-03
automatically shrink aggregation lists

sokra:
### Description

When unloading tasks, edges from the aggregation structure are removed, but their memory is not reclaimed as Vecs don't shrink automatically.

This adds a `shrink_amortized` method which shrinks the Vec/HashMap when capacity is 2 times larger than length.

This also calls `shrink_to_fit` when running GC on a task.

### DIFF
--- a/crates/turbo-tasks-auto-hash-map/src/map.rs
+++ b/crates/turbo-tasks-auto-hash-map/src/map.rs
@@ -249,6 +249,25 @@ impl<K: Eq + Hash, V, H: BuildHasher + Default, const I: usize> AutoMap<K, V, H,
             }
         }
     }
+
+    pub fn shrink_amortized(&mut self) {
+        match self {
+            AutoMap::List(list) => {
+                if list.capacity() > list.len() * 2 {
+                    list.shrink_to_fit();
+                }
+            }
+            AutoMap::Map(map) => {
+                if map.len() <= MIN_HASH_SIZE {
+                    let mut list = SmallVec::with_capacity(map.len());
+                    list.extend(map.drain());
+                    *self = AutoMap::List(list);
+                } else if map.capacity() > map.len() * 2 {
+                    map.shrink_to_fit();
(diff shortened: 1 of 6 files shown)


crates/turbo-tasks-auto-hash-map/src/map.rs
```diff
@@ -249,6 +249,25 @@ impl<K: Eq + Hash, V, H: BuildHasher + Default, const I: usize> AutoMap<K, V, H,
             }
         }
     }
+
+    pub fn shrink_amortized(&mut self) {
+        match self {
+            AutoMap::List(list) => {
+                if list.capacity() > list.len() * 2 {
```
bgw, line 256: Because HashMap and Vec grow by powers of two, this could lead to situations were we're frequently growing and shrinking the same list every time something is added and removed.

It's worse for memory, but better for time complexity to grow and shrink by different thresholds. E.g. If `list.capacity() > list.len() * 4` then shrink to `list.len() * 2`.

> Many dynamic arrays also deallocate some of the underlying storage if its size drops below a certain threshold, such as 30% of the capacity. This threshold must be strictly smaller than 1/a in order to provide hysteresis (provide a stable band to avoid repeatedly growing and shrinking) and support mixed sequences of insertions and removals with amortized constant cost. 

-- https://en.wikipedia.org/wiki/Dynamic_array#Geometric_expansion_and_amortized_cost

If you think it's worth it, we could intercept inserts here and manipulate Vec/HashMap to achieve a different growth factor than 2. Many languages use 1.5. Python uses the very low value of 1.125 (!).

It also may be worth adding an arbitrary minimum size here where it's not worth shrinking the Vec below (e.g. 4, 8, or 10) to avoid excessive resizes of very small arrays.

bgw (approved):
LGTM, as long as the shrink threshold is adjusted to be different than the growth threshold.

PR #8671, opened 2024-07-05
add unloading of tasks to GC

sokra:
### Description

Inactive tasks can be completely unloaded. This disconnects them from the graph and frees nearly all memory (except for the TaskId mapping).

When running GC for a task we check if it's inactive and unload it when it is.

But we also need to enqueue a task for GC again when it becomes inactive. So we maintain a queue of potentially inactive tasks and walk the children during GC.

### DIFF
--- a/crates/turbo-tasks-memory/src/gc.rs
+++ b/crates/turbo-tasks-memory/src/gc.rs
@@ -8,11 +8,12 @@ use std::{
 };
 
 use concurrent_queue::ConcurrentQueue;
+use dashmap::DashSet;
 use parking_lot::Mutex;
 use tracing::field::{debug, Empty};
-use turbo_tasks::TaskId;
+use turbo_tasks::{TaskId, TurboTasksBackendApi};
 
-use crate::MemoryBackend;
+use crate::{task::GcResult, MemoryBackend};
 
 /// The priority of a task for garbage collection.
 /// Any action will shrink the internal memory structures of the task in a
@@ -59,6 +60,7 @@ impl GcTaskState {
     }
 }
 
+const MAX_DEACTIVATIONS: usize = 100_000;
 const TASKS_PER_NEW_GENERATION: usize = 100_000;
 const MAX_TASKS_PER_OLD_GENERATION: usize = 200_000;
 const PERCENTAGE_TO_COLLECT: usize = 30;
@@ -74,6 +76,11 @@ struct OldGeneration {
 
 struct ProcessGenerationResult {
     priority: Option<GcPriority>,
+    content_dropped_count: usize,
+    unloaded_count: usize,
+}
+
+struct ProcessDeactivationsResult {
     count: usize,
 }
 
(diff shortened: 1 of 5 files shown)


crates/turbo-tasks-memory/src/gc.rs
```diff
@@ -88,15 +95,24 @@ pub struct GcQueue {
...
+    /// Tasks that are active and not enqueued in the deactivation queue.
+    // TODO Could be a bit field with locks, an array of atomics or an AMQF.
+    active_tasks: DashSet<TaskId>,
 }
 
 impl GcQueue {
     pub fn new() -> Self {
         Self {
-            generation: AtomicU32::new(0),
+            generation: AtomicU32::new(1),
```
arlyon, line 110: Why start at 1 here?

crates/turbo-tasks-memory/src/task.rs
```diff
@@ -427,6 +432,13 @@ use self::{
     },
 };
 
+pub enum GcResult {
+    NotPossible,
```
arlyon, line 436: May be nice to write a quick comment for cases when this outcome is expected

crates/turbo-tasks-memory/src/task.rs
```diff
@@ -1058,6 +1083,7 @@ impl Task {
                                 format!("TaskState({})::event", description())
                             }),
                             outdated_edges: Box::new(outdated_edges),
+                            clean: recomputing,
```
arlyon, line 1086: Should this not be clean if it is not currently recomputing?

crates/turbo-tasks-memory/src/task.rs
```diff
@@ -1506,57 +1563,76 @@ impl Task {
...
+    pub(crate) fn run_gc(
+        &self,
+        generation: u32,
+        gc_queue: &GcQueue,
+        backend: &MemoryBackend,
+        turbo_tasks: &dyn TurboTasksBackendApi<MemoryBackend>,
+    ) -> GcResult {
         if !self.is_pure() {
-            return false;
+            return GcResult::NotPossible;
```
arlyon, line 1574: Cool

arlyon (approved):
Approved with some comments

crates/turbo-tasks-memory/src/task.rs
sokra: If it's marked as dirty due to changes, it's not clean.
If it's marked as dirty when trying to read a missing value, it's still clean.

crates/turbo-tasks-memory/src/gc.rs
sokra: we use generation 0 as marker for "not in any generation", e. g. after as task was GC'ed it will have generation set to 0.

I can change it to Option<NonZeroU32> to make it more clear.

crates/turbo-tasks-memory/src/task.rs
sokra: I split mark_dirty_internal into two methods: mark_dirty and recompute to make it more clear.

PR #8675, opened 2024-07-05
fix: repo-docs

pathliving:
### Description

Corrected a missing comma between items in a JSON array, which caused an error when copying and pasting a code fragment into a JSON file


anthonyshew (approved):
Oops, thanks!

PR #8679, opened 2024-07-06
fix(docs): fix pnpm workspace filename extension

UNRULYEON:
### Description

The extension for a pnpm workspace manifest file should be `yaml`, not `yml`. Using `yml` provides an error: `The workspace manifest file should be named "pnpm-workspace.yaml"`

pnpm documentation for workspace manifests: https://pnpm.io/pnpm-workspace_yaml


anthonyshew (approved):
Good catch. Thanks!

PR #8683, opened 2024-07-07
docs: Fix typo in filter option.

LaPulgaaa:
`errors-only` is the correct option
option that only shows logs for task failures.

### Description
Fixes a minor typo in `crafting your repository` section of turbo repo docs.


anthonyshew (approved):
Oops! Yes, thank you for both noticing and contributing the fix!

PR #8692, opened 2024-07-08
docs: Make root task more clear

baileywickham:
### Description
Uses the exact task name that is described in the `turbo.json` file. This confused me at first, I thought the `:root` was a custom syntax for defining where lint would be ran.


anthonyshew (approved):
Oops, good catch. Thanks!

PR #8699, opened 2024-07-09
perf: Merge multiple `EsmBinding`

kdy1:
### Description

Merge multiple `EsmBinding` into single `CodeGeneratable`

### Testing Instructions

This is a pure optimization, so if the behavior is same it's all good.

Closes PACK-3147

### DIFF
--- a/crates/turbopack-ecmascript/src/references/esm/binding.rs
+++ b/crates/turbopack-ecmascript/src/references/esm/binding.rs
@@ -6,7 +6,10 @@ use swc_core::{
             ComputedPropName, Expr, Ident, KeyValueProp, Lit, MemberExpr, MemberProp, Number, Prop,
             PropName, SeqExpr, SimpleAssignTarget, Str,
         },
-        visit::fields::{CalleeField, PropField},
+        visit::{
+            fields::{CalleeField, PropField},
+            AstParentKind,
+        },
     },
 };
 use turbo_tasks::{RcStr, TaskInput, Vc};
@@ -14,7 +17,7 @@ use turbopack_core::chunk::ChunkingContext;
 
 use super::EsmAssetReference;
 use crate::{
-    code_gen::{CodeGenerateable, CodeGeneration},
+    code_gen::{CodeGenerateable, CodeGeneration, VisitorFactory},
     create_visitor,
     references::AstPath,
 };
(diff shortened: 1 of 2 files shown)


crates/turbopack-ecmascript/src/references/esm/binding.rs
```diff
@@ -19,6 +19,20 @@ use crate::{
...
+#[turbo_tasks::value(shared)]
+#[derive(Hash, Debug)]
+pub struct EsmBindings {
+    pub bindings: Vc<Vec<Vc<EsmBinding>>>,
+}
+
+#[turbo_tasks::value_impl]
+impl EsmBindings {
+    #[turbo_tasks::function]
+    pub fn new(bindings: Vc<Vec<Vc<EsmBinding>>>) -> Vc<Self> {
```
sokra, line 31: The main idea of this optimization was to avoid calling a turbo tasks function per EsmBinding. So it would be better to store a Vec<EsmBinding> instead of Vec<Vc<EsmBinding>>

kdy1: How can I store vector here?

```
`turbo_tasks::Vc<std::vec::Vec<references::esm::binding::EsmBinding>>` is not a future
the trait `std::future::IntoFuture` is not implemented for `turbo_tasks::Vc<std::vec::Vec<references::esm::binding::EsmBinding>>`
turbo_tasks::Vc<std::vec::Vec<references::esm::binding::EsmBinding>> must be a future or must implement `IntoFuture` to be awaited

```

kdy1: Done

crates/turbopack-ecmascript/src/references/esm/binding.rs
```diff
@@ -21,144 +21,118 @@ use crate::{
...
-                        ],
-                        span,
-                    });
-                }
-                expr
-            } else {
-                Expr::Ident(Ident::new(imported_module.into(), span))
-            }
-        }
+        for item in bindings.into_iter() {
```
ForsakenHarmony, line 68: Can we keep this as a function of `EsmBinding` (maybe not a turbo task)?
it already has too many levels of indentation

crates/turbopack-ecmascript/src/references/esm/binding.rs
```diff
@@ -6,95 +6,64 @@ use swc_core::{
...
+
+#[turbo_tasks::value_impl]
+impl EsmBindings {
+    #[turbo_tasks::function]
+    pub fn new(bindings: Vec<EsmBinding>) -> Vc<Self> {
+        EsmBindings { bindings }.cell()
+    }
+}
+
+#[derive(Hash, Clone, Debug, TaskInput)]
```
sokra, line 39: This need to derive a few more things

crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -264,6 +270,9 @@ impl AnalyzeEcmascriptModuleResultBuilder {
         mut self,
         track_reexport_references: bool,
     ) -> Result<Vc<AnalyzeEcmascriptModuleResult>> {
+        let bindings = EsmBindings::new(take(&mut self.bindings));
+        self.add_code_gen(bindings);
```
sokra, line 274: Check for self.bindings.is_empty(), e. g. commonjs modules won't have bindings at all

sokra (approved):

PR #8712, opened 2024-07-10
Turn Postcss into rule

LichuAcu:
### What?

Turned PostCss into a custom rule.

Tested by running `pnpm test-turbo test/e2e/postcss-config-cjs` and `pnpm test-turbo test/production/postcss-plugin-config-as-string`

Depends on vercel/next.js#67640.


bgw (commented):
This probably needs changes to `next.js` as well: https://github.com/search?q=repo%3Avercel%2Fnext.js%20enable_postcss_transform&type=code

> Turned PostCss into a custom rule.

This PR description is a little confusing? It looks like this isn't adding the custom rule bits, but instead cleaning up the old flag?

crates/turbopack-cli/src/contexts.rs
```diff
@@ -155,7 +153,6 @@ async fn get_client_module_options_context(
 
     let module_options_context = ModuleOptionsContext {
         enable_jsx,
-        enable_postcss_transform: Some(PostCssTransformOptions::default().cell()),
```
sokra, line 158: Turbopack cli should have PostCSS enabled, so you need to add the postcss rule here too

LichuAcu: The function to build the postcss rule is currently [in the next.js repo](https://github.com/vercel/next.js/blob/postcss-into-rule/packages/next-swc/crates/next-core/src/next_shared/transforms/postcss.rs#L69). Should we move it to the turbo repo, maybe to [this file](https://github.com/vercel/turbo/blob/postcss-into-rule/crates/turbopack-node/src/transforms/postcss.rs), so we can use it un turbopack-cli?

tknickman:
Closing this as turbopack has moved to next.js - feel free to re-open over there!

PR #8713, opened 2024-07-10
feat(tui): terminal pane mouse copying

chris-olszewski:
### Description

Adds mouse selection and copying in the terminal pane. When selecting text, pressing `c` will copy the current text. Users are allowed to interact with tasks while having a selection, but `c` will still be intercepted to copy the selection.

Future work:
 - Double click should select the entire line

Reviewer Guide:
 - Review each commit on it's own
 - We implement selection at the virtual terminal level, this allows us to take into account the current scroll state and translate mouse events into "absolute" position.
 - I did a bad thing and fixed some bugs in this PR. They have their own commits, but easier to do here as they would cause merge conflicts if done independently.

### Testing Instructions

Added unit tests for the vt100 level copying. Testing this in actual terminals is the best way to test mouse inputs feel right.

Video: (please ignore my struggles scrolling and accurately selecting text, I'm on a trackpad and it's very humid)

https://github.com/user-attachments/assets/bda72f3b-dc8c-4344-972a-cfc0ac65c217

### DIFF
--- a/crates/turborepo-ui/src/tui/clipboard.rs
+++ b/crates/turborepo-ui/src/tui/clipboard.rs
@@ -1,0 +1,113 @@
+// Inspired by https://github.com/pvolok/mprocs/blob/master/src/clipboard.rs
+use std::process::Stdio;
+
+use base64::Engine;
+use which::which;
+
+pub fn copy_to_clipboard(s: &str) {
+    match copy_impl(s, &PROVIDER) {
+        Ok(()) => (),
+        Err(err) => tracing::debug!("Unable to copy: {}", err.to_string()),
+    }
+}
+
+#[allow(dead_code)]
+enum Provider {
+    OSC52,
+    Exec(&'static str, Vec<&'static str>),
+    #[cfg(windows)]
+    Win,
+    NoOp,
+}
+
+#[cfg(windows)]
+fn detect_copy_provider() -> Provider {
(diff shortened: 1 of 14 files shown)


crates/turborepo-ui/src/tui/clipboard.rs
```diff
@@ -0,0 +1,113 @@
...
+#[allow(dead_code)]
+enum Provider {
+    OSC52,
+    Exec(&'static str, Vec<&'static str>),
+    #[cfg(windows)]
+    Win,
+    NoOp,
+}
+
+#[cfg(windows)]
```
NicholasLYang, line 23: would it make sense to fold these into an if statement with `cfg!`? That way the compiler can check the code paths

crates/turborepo-ui/src/tui/clipboard.rs
```diff
@@ -0,0 +1,113 @@
...
+        None
+    }
+}
+
+fn copy_impl(s: &str, provider: &Provider) -> std::io::Result<()> {
+    match provider {
+        Provider::OSC52 => {
+            let mut stdout = std::io::stdout().lock();
+            use std::io::Write;
+            write!(
```
NicholasLYang, line 81: Just curious, what's the difference between this and `print!`?

NicholasLYang (approved):

PR #8721, opened 2024-07-11
fix equal checks for cells and output

sokra:
### Description

* Check for equal Vc on returning a value
* Avoid PartialEq checks for values that are never equal

### DIFF
--- a/crates/turbo-tasks-memory/src/output.rs
+++ b/crates/turbo-tasks-memory/src/output.rs
@@ -56,7 +56,17 @@ impl Output {
 
     pub fn link(&mut self, target: RawVc, turbo_tasks: &dyn TurboTasksBackendApi<MemoryBackend>) {
         debug_assert!(*self != target);
-        self.assign(OutputContent::Link(target), turbo_tasks)
+        if let OutputContent::Link(old_target) = &self.content {
+            if *old_target == target {
+                // unchanged
+                return;
+            }
+        }
+        self.content = OutputContent::Link(target);
+        // notify
+        if !self.dependent_tasks.is_empty() {
+            turbo_tasks.schedule_notify_tasks_set(&take(&mut self.dependent_tasks));
+        }
     }
 
     pub fn error(&mut self, error: Error, turbo_tasks: &dyn TurboTasksBackendApi<MemoryBackend>) {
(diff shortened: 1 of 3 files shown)


crates/turbopack-core/src/source_map/mod.rs
```diff
@@ -43,7 +43,7 @@ pub trait GenerateSourceMap {
 /// The distinction between the source map spec's [sourcemap::Index] and our
 /// [SourceMap::Sectioned] is whether the sections are represented with Vcs
 /// pointers.
-#[turbo_tasks::value(shared)]
+#[turbo_tasks::value(shared, cell = "new")]
```
sokra, line 46: `cell = "new"` changes the behavior of `.cell()` to no longer perform an `==` check and just always override the cell.

crates/turbo-tasks-memory/src/output.rs
```diff
@@ -56,7 +56,17 @@ impl Output {
 
     pub fn link(&mut self, target: RawVc, turbo_tasks: &dyn TurboTasksBackendApi<MemoryBackend>) {
         debug_assert!(*self != target);
-        self.assign(OutputContent::Link(target), turbo_tasks)
+        if let OutputContent::Link(old_target) = &self.content {
+            if *old_target == target {
+                // unchanged
+                return;
```
sokra, line 62: We can early return when returning the same value again.

Previously this notified all dependent tasks, which is unnecessary.

crates/turbopack-core/src/source_map/mod.rs
arlyon: How do we determine which cells should always be overridden?

bgw: When you don't want cells to ever match using equality? Which I imagine should be never unless you've got another bug (e.g. inconsistent cell creation order?).

I agree this seems weird and like we're masking bigger problems.

bgw (approved):

crates/turbopack-core/src/source_map/mod.rs
sokra: > When you don't want cells to ever match using equality?

* some types have every expensive equal checks and are very unlikely to be equal (e. g. the parsed AST or here the SourceMap)
* some types don't have a equality defined
* some types define equality in a way that it never matches

But in general you want to check equality for early stop bubbling invaliations...

PR #8722, opened 2024-07-11
Use fs and vm for async chunk loading in Node

mischnic:
### Description

Use `fs` and `vm` when loading chunks asynchronously in Node, to not block the thread while reading large chunks

Closes PACK-411

### Testing Instructions

Using the `test/e2e/app-dir/dynamic/` Next.js tests


crates/turbopack-ecmascript-runtime/js/src/nodejs/runtime.ts
```diff
@@ -118,14 +120,47 @@ async function loadChunkAsync(
...
+  if (!chunkPath.endsWith(".js")) {
+    // We only support loading JS chunks in Node.js.
+    // This branch can be hit when trying to load a CSS chunk.
+    return;
+  }
+
+  const resolved = path.resolve(RUNTIME_ROOT, chunkPath);
+
+  return new Promise<void>(function (resolve, reject) {
+    fs.readFile(resolved, "utf-8", function (err: Error, content: string) {
```
ForsakenHarmony, line 133: Why not use the fs promise interface?

mischnic: I forgot about that one... 👍

ForsakenHarmony (approved):

Issue #8311, opened 2024-06-05
New UI breaks when I'm running it in husky precommit


anthonyshew:
This new UI breaks when I'm running it in husky precommit:

In my root package.json:

```
    "lint": "turbo lint",
```

and in `.husky/pre-commit`:

```
pnpm lint
```

when running this, it gives:

```
$ git commit -m "Test commit"
• Packages in scope: ....
• Running lint in 8 packages
• Remote caching disabled
  × internal errors encountered: receiver dropped,receiver dropped,receiver dropped
```

The only way to fix this that I've found dso far is to add `"ui": "stream"` to the root `turbo.json`.

_Originally posted by @siawyoung in https://github.com/vercel/turbo/discussions/7802#discussioncomment-9672182_

willsather:
We are facing this issue as well - seems like it's only occurring when any _one_ of our `pre-commit` commands is a cache miss, otherwise if all commands are a cache hit - it works fine.

Adding `"ui": "stream"` to the root `turbo.json` does solve the husky `pre-commit` issue, but unfortunately disables the new terminal UI.

<img width="570" alt="image" src="https://github.com/vercel/turbo/assets/56037657/8a23f19d-9643-4c64-abf6-65ab82e24b73">

pieterjandebruyne:
Reproduce repo:
https://github.com/pieterjandebruyne/turbov2-husky-bug

### To Reproduce
run `git clone git@github.com:pieterjandebruyne/turbov2-husky-bug.git`
-> change any index.ts file
run `git add . && git commit -m "test"`

pieterjandebruyne:
"ui": "stream" does fix the husky issue but I am getting different errors now when I use it inside of a dockerfile:

```
#19 ERROR: process "/bin/sh -c turbo run build --filter=frontend" did not complete successfully: exit code: 1
------
 > [installer 8/8] RUN turbo run build --filter=frontend:
0.207 Error: turbo_json_parse_error
0.207 
0.207   x Found an unknown value `Stream`.
0.207     ,-[57:1]
0.207  57 |   },
0.207  58 |   "ui": "Stream"
0.207     :         ^^^^^^^^
0.207  59 | }
0.207     `----
0.207 
------
Dockerfile:32
--------------------
  30 |     
  31 |     RUN npm -g i turbo@2.0.1
  32 | >>> RUN turbo run build --filter=frontend
  ```
  
[...]

PR #8728, opened 2024-07-11
adds daemon flag for env var and turbo.json

dimitropoulos:
### Description

This PR adds support and testing for a `TURBO_DAEMON` environment variable, as well as a `{ daemon: boolean }` field to the `turbo.json` file.  This change allows us to remove `daemon` from `RunOpts`, as well.

### Testing Instructions

You can run `turbo daemon status` to see the status of the daemon after each run (and also `turbo daemon stop` to kill it).

#### Daemon Off

1. run with `--no-daemon`
1. run with `TURBO_DAEMON=false`
1. run with `{"daemon":false}` in your turbo.json

#### Daemon On

1. run with `--daemon`
1. run with `TURBO_DAEMON=true`
1. run with `{"daemon":true}` in your turbo.json

### DIFF
--- a/crates/turborepo-lib/src/turbo_json/mod.rs
+++ b/crates/turborepo-lib/src/turbo_json/mod.rs
@@ -131,6 +131,8 @@ pub struct RawTurboJson {
         rename = "dangerouslyDisablePackageManagerCheck"
     )]
     pub allow_no_package_manager: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub daemon: Option<Spanned<bool>>,
 
     #[deserializable(rename = "//")]
     #[serde(skip)]
@@ -1081,6 +1083,14 @@ mod tests {
     fn test_ui(json: &str, expected: Option<UI>) {
         let json = RawTurboJson::parse(json, AnchoredSystemPath::new("").unwrap()).unwrap();
         assert_eq!(json.ui, expected);
+    }
+
+    #[test_case(r#"{ "daemon": true }"#, r#"{"daemon":true}"# ; "daemon_on")]
+    #[test_case(r#"{ "daemon": false }"#, r#"{"daemon":false}"# ; "daemon_off")]
+    fn test_daemon(json: &str, expected: &str) {
+        let parsed = RawTurboJson::parse(json, AnchoredSystemPath::new("").unwrap()).unwrap();
+        let actual = serde_json::to_string(&parsed).unwrap();
+        assert_eq!(actual, expected);
     }
 
     #[test_case(r#"{ "ui": "tui" }"#, r#"{"ui":"tui"}"# ; "tui")]
--- a/turborepo-tests/integration/tests/config.t
+++ b/turborepo-tests/integration/tests/config.t
@@ -57,3 +57,22 @@ Run build with invalid env var
   
   [1]
 
+Confirm that the daemon is not configured
+  $ ${TURBO} config | jq .daemon
+  null
+
+Add env var: `TURBO_DAEMON=true`
+  $ TURBO_DAEMON=true ${TURBO} config | jq .daemon
+  true
+
+Add env var: `TURBO_DAEMON=false`
+  $ TURBO_DAEMON=false ${TURBO} config | jq .daemon
+  false
+
+Add flag: `--daemon`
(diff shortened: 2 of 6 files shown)


ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 7a9e23b75961d269eb99f86685b01b17645f17fe

mehulkar (approved):
Does `--no-daemon` still work also?

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -1075,6 +1077,14 @@ mod tests {
         assert_eq!(json.ui, expected);
     }
 
+    #[test_case(r#"{ "daemon": true }"#, Some(true) ; "daemon_on")]
+    #[test_case(r#"{ "daemon": false }"#, Some(false) ; "daemon_off")]
+    #[test_case(r#"{}"#, None ; "missing")]
+    fn test_daemon(json: &str, expected: Option<bool>) {
```
chris-olszewski, line 1083: So something I missed: when we add new keys/fields to `turbo.json` we should always test both deserialization and serialization. When we don't we often create issues ([most recent issue caused by me forgetting to test this](https://github.com/vercel/turbo/issues/8311))

dimitropoulos: sounds good! updated in https://github.com/vercel/turbo/pull/8728/commits/9065d956ad5d42abb1fd9744a98a2be105b01b9a

dimitropoulos:
@mehulkar yep :) `--no-daemon` is unchanged (still works).  there was an error in the description (updated now).

crates/turborepo-lib/src/run/builder.rs
```diff
@@ -55,6 +55,7 @@ pub struct RunBuilder {
     api_auth: Option<APIAuth>,
     repo_root: AbsoluteSystemPathBuf,
     ui: UI,
+    daemon: Option<bool>,
```
NicholasLYang, line 58: Could we keep this in `Opts` but use the config value to set `daemon`? That way we can keep using `Opts` as the single source of truth

NicholasLYang (changes requested):
I think we should keep this config in `Opts` and not in `RunBuilder`/`Run` directly

crates/turborepo-lib/src/run/builder.rs
dimitropoulos: looks like we're all set here! see https://github.com/vercel/turbo/pull/8728/commits/60e9a54dd1c002583d78f6b9f5dd02b2797e541e

turborepo-tests/integration/tests/config.t
```diff
@@ -15,7 +15,8 @@ Run test run
     "enabled": true,
     "spacesId": null,
     "ui": false,
-    "packageManager": "npm"
+    "packageManager": "npm",
+    "daemon": null
```
NicholasLYang, line 19: Might be good to add a test that confirms the overloading behavior with an env var or flag.

NicholasLYang (approved):
Needs a test, but otherwise LGTM!

turborepo-tests/integration/tests/config.t
dimitropoulos: thanks! updated in a7216fad72a4bd19d34519d10a29677e9cee3671

shawnmcknight:
@dimitropoulos Is there something separate that needs to be done to get the json schema for `turbo.json` updated? The `daemon` property does not appear to be present in https://turbo.build/schema.json.

PatrykKuniczak:
@dimitropoulos How to use this 
``` {"daemon":false} ```
in turbo.json?

I'm using latest version but i can't do it, also there's no DOCS for this

dimitropoulos:
@shawnmcknight my understanding is that there's a separate process (in another (private) repo) that generates that schema.  I'm looking into why it's not been updated - sorry about that!

dimitropoulos:
@PatrykKuniczak we're following up with the docs for this and the other two, [cacheDir](github.com/vercel/turbo/pull/8947) (in progress) and [envMode](https://github.com/vercel/turbo/pull/8757) (complete) at the same time.

PR #8730, opened 2024-07-11
chore(ci): run JS package tests on node 18

mehulkar:
Testing in https://github.com/vercel/turbo/pull/8731


.github/workflows/turborepo-native-lib-test.yml
```diff
@@ -1,4 +1,4 @@
-name: Turborepo Native Library Tests
+name: "@turborepo/repository tests"
```
mehulkar, line 1:
```suggestion
name: Turborepo Native Library Tests
```

tknickman (approved):

PR #8733, opened 2024-07-12
chore(ci): Skip downloading Playwright browsers in Turborepo CI.

anthonyshew:
### Description

Playwright is only used on the Turbopack side, so we can skip downloading its browsers when we're about to run Turborepo's CI.

### Testing Instructions

CI


bgw (approved):

anthonyshew:
@bgw Can you confirm that I shouldn't mind the failing windows test? I'd be hard-pressed to imagine this change caused that failure but I don't want to assume.

PR #8736, opened 2024-07-12
Refactor task arguments to be a single one

sokra:
### Description

As preparation to pass a struct only, we change the task arguments to be a single one

PR #8735, opened 2024-07-12
turbo-tasks this calls

sokra:
### Description

Make this argument in turbo-tasks function calls explicit

PR #8737, opened 2024-07-12
remove ConcreteTaskInput in favor of the original rust datatypes

sokra:
### Description

Instead of the ConcreteTaskInput conversion, we use a boxed any. This allows to store arguments in plain rust structures and uses less memory.

We need to do some magic to make serialization possible.

Also removed Ord requirement from TaskInputs and make Vc not implement Ord


sokra:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/vercel/turbo/8737?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#8737** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8737?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#8736** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8736?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8735** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8735?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @sokra and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

crates/turbopack-core/src/resolve/pattern.rs
```diff
@@ -764,12 +764,29 @@ impl ValueToString for Pattern {
...
+    fn partial_cmp(&self, other: &PatternMatch) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for PatternMatch {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match (self, other) {
+            (PatternMatch::File(a, _), PatternMatch::File(b, _))
+            | (PatternMatch::Directory(a, _), PatternMatch::Directory(b, _)) => a.cmp(b),
```
bgw, line 783: This may disagree with the `Eq` implementation. The rustdocs aren't very explicit about the relationship between `Eq` and `Ord`, but I think this may violate the requirements for implementing `Ord`?

If we only need `PartialOrd`, I think we should go ahead and implement that, but leave off `Ord`.

crates/turbo-tasks/src/vc/resolved.rs
```diff
@@ -20,14 +20,24 @@ use indexmap::{IndexMap, IndexSet};
 
 use crate::{vc::Vc, RcStr};
 
-#[derive(Copy, Clone)]
```
bgw, line 23: What was wrong with the derives here? At a glance they're equivalent? Also the PartialEq/Eq implementations look equivalent to what would've been derived?

crates/turbo-tasks/src/value_type.rs
```diff
@@ -111,7 +100,7 @@ impl ValueType {
 
     /// This is internally used by `#[turbo_tasks::value]`
     pub fn new_with_magic_serialization<
-        T: Debug + Eq + Ord + Hash + Serialize + for<'de> Deserialize<'de> + Send + Sync + 'static,
+        T: Debug + Eq + Hash + Serialize + for<'de> Deserialize<'de> + Send + Sync + 'static,
```
bgw, line 103: Outside the scope of this PR, but the complicated and often-repeated bounds here might be a good candidate for the "trait alias" trick:

https://www.reddit.com/r/rust/comments/1582jku/how_can_one_create_a_trait_alias/

```
trait Valid: PartialOrd + Display + PartialEq + Clone {}
impl<T: PartialOrd + Display + PartialEq + Clone> Valid for T {}
```

("quotes" around "trait alias" because real trait aliases aren't stable)

crates/turbo-tasks/src/native_function.rs
```diff
@@ -1,18 +1,98 @@
...
+                        #[cfg(debug_assertions)]
+                        return format!(
+                            "Invalid argument type, expected {} got {}",
+                            std::any::type_name::<T>(),
+                            value.magic_type_name()
+                        );
+                        #[cfg(not(debug_assertions))]
+                        return "Invalid argument type";
+                    })
+                    .unwrap()
```
bgw, line 54: This downcasting bit with pretty error messages is the same between both `resolve` and `is_resolved`? Move it into a generic helper method?

Also, thanks for minding the binary size here with `debug_assertions`!

crates/turbo-tasks/src/native_function.rs
```diff
@@ -1,18 +1,98 @@
...
+        &'a dyn MagicAny,
+    ) -> Pin<Box<dyn Future<Output = Result<Box<dyn MagicAny>>> + Send + 'a>>;
+
+type IsResolvedFunctor = fn(&dyn MagicAny) -> bool;
+
+pub struct ArgMeta {
+    serializer: MagicAnySerializeSeed,
+    deserializer: MagicAnyDeserializeSeed,
+    is_resolved: IsResolvedFunctor,
+    resolve: ResolveFunctor,
```
bgw, line 30: This is a clever way to avoid propagating the type argument to the struct without dynamic dispatch.

bgw (commented):
Got part of the way through the code review, left some comments. I'll pick this back up tomorrow.

Before this merges, we need some sort of rough memory and CPU benchmarks (probably in combination with the rest of the stack). This should be an improvement, but it's hard to say by how much.

crates/turbopack-core/src/resolve/pattern.rs
sokra: If you are worried, we could omit PartialOrd and Ord and use `sort_by_key` instead

crates/turbo-tasks/src/vc/resolved.rs
sokra: Derive would add `T: Clone` requirement, which we don't want.

crates/turbo-tasks/src/backend.rs
```diff
@@ -99,28 +100,206 @@ impl Display for PersistentTaskType {
...
-                this: _,
-                arg,
-            } => arg.shrink_to_fit(),
-            Self::ResolveTrait {
-                trait_type: _,
-                method_name: _,
-                this: _,
-                arg,
-            } => arg.shrink_to_fit(),
+mod ser {
```
bgw, line 103: Do we have any unit tests that round-trip this serialization logic? If not, it might be nice to have.

sokra:
> Before this merges, we need some sort of rough memory and CPU benchmarks (probably in combination with the rest of the stack). This should be an improvement, but it's hard to say by how much.

It's less than I expected. 8% memory win on a single page, 4% on 100 pages.

bgw (approved):

bgw:
> It's less than I expected. 8% memory win on a single page, 4% on 100 pages.

That's still pretty good, plus I assume this probably comes with a CPU win too? And maybe a binary size win?

crates/turbo-tasks/src/backend.rs
sokra: not yet, I'll add them as part of persistent caching

PR #8748, opened 2024-07-15
Support `module_resolution: "nodenext"`

mischnic:
### Description

Allow resolving for example `./foo.js` to `./foo.ts`

`.js` can resolve to `.ts` and `.tsx`
`.mjs` can resolve to `.mts`
`.cjs` can resolve to `.cts`

Closes PACK-3031

This is what `tsc --traceResolution` says about priority of the various possible resolution:
```
======== Resolving module '../libs/f.js' from '/Users/niklas/Desktop/nodenext-app/app/page.tsx'. ========
Explicitly specified module resolution kind: 'NodeNext'.
Resolving in CJS mode with conditions 'require', 'types', 'node'.
Loading module as file / folder, candidate module location '/Users/niklas/Desktop/nodenext-app/libs/f.js', target file types: TypeScript, JavaScript, Declaration.
File name '/Users/niklas/Desktop/nodenext-app/libs/f.js' has a '.js' extension - stripping it.
File '/Users/niklas/Desktop/nodenext-app/libs/f.ts' does not exist.
File '/Users/niklas/Desktop/nodenext-app/libs/f.tsx' does not exist.
File '/Users/niklas/Desktop/nodenext-app/libs/f.d.ts' does not exist.
File '/Users/niklas/Desktop/nodenext-app/libs/f.js' does not exist.
File '/Users/niklas/Desktop/nodenext-app/libs/f.jsx' does not exist.
File '/Users/niklas/Desktop/nodenext-app/libs/f.js.ts' does not exist.
File '/Users/niklas/Desktop/nodenext-app/libs/f.js.tsx' does not exist.
File '/Users/niklas/Desktop/nodenext-app/libs/f.js.d.ts' does not exist.
```

### Testing Instructions

I also added a test case


crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1921,6 +1921,18 @@ async fn resolve_relative_request(
...
+    if !options_value.fully_specified && options_value.enable_js_ts_rewriting {
+        // TODO path_pattern might not be a constant?
+        // TODO extension might be empty, or dots in filepath
+        if let Pattern::Constant(s) = path_pattern {
+            if let Some((base, _ext)) = s.rsplit_once(".") {
+                new_path = Pattern::Alternatives(vec![Pattern::Constant(base.into()), new_path])
+            }
+        } else {
+            todo!("enable_js_ts_rewriting");
+        }
```
mischnic, line 1933: I felt there might be some existing utilities/functions for path handling/etc but I couldn't find anything immediately.

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1921,6 +1921,20 @@ async fn resolve_relative_request(
...
+        // TODO extension might be empty, or dots in filepath
+        if let Pattern::Constant(s) = path_pattern {
+            if let Some((base, ext)) = s.rsplit_once(".") {
+                if ext == "js" {
+                    new_path = Pattern::Alternatives(vec![Pattern::Constant(base.into()), new_path])
+                }
+            }
+        } else {
+            todo!("enable_js_ts_rewriting");
+        }
```
mischnic, line 1935: I felt there might be some existing utilities/functions for path handling/etc but I couldn't find anything immediately.

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1943,6 +1942,36 @@ async fn resolve_relative_request(
...
+                        Pattern::Concatenation(vec![
+                            Pattern::Constant(base.into()),
+                            Pattern::Alternatives(replacement),
+                        ])
+                    })
+                });
+
+            if rewritten_path_modified {
+                // Prepend the rewritten pattern to give it higher priority
+                new_path = Pattern::Alternatives(vec![rewritten_path, new_path])
```
sokra, line 1971: Instead of doing that you can just keep the original extension in the list for Alternatives above

crates/turbopack-core/src/resolve/options.rs
```diff
@@ -442,6 +442,9 @@ pub struct ResolveOptions {
     pub resolved_map: Option<Vc<ResolvedMap>>,
     pub before_resolve_plugins: Vec<Vc<Box<dyn BeforeResolvePlugin>>>,
     pub plugins: Vec<Vc<Box<dyn AfterResolvePlugin>>>,
+    /// Support resolving *.js requests to *.ts files
+    pub enable_js_ts_rewriting: bool,
```
sokra, line 446: Maybe have a better name for that. Something with typescript.

`enable_typescript_with_output_extension`

crates/turbopack-core/src/resolve/pattern.rs
```diff
@@ -627,6 +627,33 @@ impl Pattern {
...
+                }
+            }
+            Pattern::Dynamic => {}
+            Pattern::Alternatives(list) => {
+                for i in list {
+                    replaced = i.replace_final_constants(cb) || replaced;
+                }
+            }
+            Pattern::Concatenation(list) => {
+                if let Some(c @ Pattern::Constant(_)) = list.last_mut() {
```
sokra, line 650: The check for constant should be removed here as the last element might be something complex too

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1943,6 +1942,36 @@ async fn resolve_relative_request(
                 .collect(),
         ));
 
+        if options_value.enable_js_ts_rewriting {
+            let mut rewritten_path = path_pattern.clone();
+            let rewritten_path_modified =
+                rewritten_path.replace_final_constants(&|c: &RcStr| -> Option<Pattern> {
```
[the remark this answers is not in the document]
mischnic: One sideeffect of that approach is that is also replaces the `js` that was appended via `extensions`:

**request with appended extensions:**
```rust
Alternatives([
    Constant("./src/foo.js"),
    Constant("./src/foo.js.ts"),
    Constant("./src/foo.js.js"),
    Constant("./src/foo.js.json")])
```
**request with extensions and then the new replacement**
```rust
Alternatives([
    Constant("./src/foo.ts"),
    Constant("./src/foo.tsx"),
    Constant("./src/foo.js"),
    Constant("./src/foo.js.ts"),  // <-
    Constant("./src/foo.js.ts"),  // <-
    Constant("./src/foo.js.tsx"), // <-
    Constant("./src/foo.js.js"),
    Constant("./src/foo.js.json")])
```
But in practice, it probably doesn't make a difference

crates/turbopack-tests/tests/execution/turbopack/resolving/tsconfig-nodenext/input/src/foo.ts
```diff
@@ -0,0 +1 @@
+export default "foo";
```
wojtekmaj, line 1: A test for TSX would be nice too!

sokra (approved):

PR #8757, opened 2024-07-15
add envMode to `turbo.json`

dimitropoulos:
### Testing Instructions

1. (temporarily) update `@turbo/gen`'s `check-types` command to `    "check-types": "env | sort | tee /dev/tty | wc -l"`
1. execute `turbo run check-types --filter @turbo/gen` and observe the number
1. add `  "envMode": "loose",` to the root `turbo.json`
1. run the command above again, observer the number is larger, indicating more environment variables have been included.

### DIFF
--- a/crates/turborepo-lib/src/commands/config.rs
+++ b/crates/turborepo-lib/src/commands/config.rs
@@ -3,7 +3,10 @@ use turborepo_repository::{
     package_graph::PackageGraph, package_json::PackageJson, package_manager::PackageManager,
 };
 
-use crate::{cli, commands::CommandBase};
+use crate::{
+    cli::{self, EnvMode},
+    commands::CommandBase,
+};
 
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
@@ -21,6 +24,7 @@ struct ConfigOutput<'a> {
     ui: bool,
     package_manager: PackageManager,
     daemon: Option<bool>,
+    env_mode: EnvMode,
 }
 
 pub async fn run(base: CommandBase) -> Result<(), cli::Error> {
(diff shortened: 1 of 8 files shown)


ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 1cda3d060ea9223ef7867e2b42f89095cf117a4b

crates/turborepo-lib/src/commands/config.rs
```diff
@@ -20,6 +20,7 @@ struct ConfigOutput<'a> {
     spaces_id: Option<&'a str>,
     ui: bool,
     package_manager: PackageManager,
+    daemon: Option<bool>,
```
mehulkar, line 23: Does this need rebasing ro something? This PR is for envMode but diff shows daemon config?

mehulkar (commented):
Probably ok, but can you rebase / remove the daemon stuff and I can take another look?

chris-olszewski (changes requested):
It looks like we're no longer respecting the CLI flags properly in the integration tests.

Tests should be added for:
 - Parsing/serializing `turbo.json`
 - Setting env mode via an env variable (it's partially finished, we just are missing an assertion)

crates/turborepo-lib/src/commands/config.rs
dimitropoulos: yep: it was just stacked :)

chris-olszewski (changes requested):
It looks like the integration tests are still failing. It appears that this is an issue in the value logged in the run summary JSON. We can take some time to trace that codepath and see if it's just pulling from a different place. (Or there's still an odd bug that we're missing)

chris-olszewski (approved):

PR #8759, opened 2024-07-15
refactor(turborepo): derive `Opts` from `Config`

NicholasLYang:
### Description

Changed `Opts` to be derived from more than just `RunArgs` and `ExecutionArgs`. This means that `Opts` contains the fully resolved config, which gets rid of some messy mutable logic in `RunBuilder` and creates a single source of truth.

### Testing Instructions

Existing tests cover this pretty well.

### DIFF
--- a/crates/turborepo-api-client/src/lib.rs
+++ b/crates/turborepo-api-client/src/lib.rs
@@ -123,6 +123,16 @@ pub struct APIAuth {
     pub team_slug: Option<String>,
 }
 
+impl std::fmt::Debug for APIAuth {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("APIAuth")
+            .field("team_id", &self.team_id)
+            .field("token", &"***")
+            .field("team_slug", &self.team_slug)
+            .finish()
+    }
+}
+
 pub fn is_linked(api_auth: &Option<APIAuth>) -> bool {
     api_auth
         .as_ref()
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -682,8 +682,8 @@ pub struct ExecutionArgs {
     /// Environment variable mode.
     /// Use "loose" to pass the entire existing environment.
     /// Use "strict" to use an allowlist specified in turbo.json.
-    #[clap(long = "env-mode", default_value = "strict", num_args = 0..=1, default_missing_value = "strict")]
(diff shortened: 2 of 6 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -682,8 +682,8 @@ pub struct ExecutionArgs {
     /// Environment variable mode.
     /// Use "loose" to pass the entire existing environment.
     /// Use "strict" to use an allowlist specified in turbo.json.
-    #[clap(long = "env-mode", default_value = "strict", num_args = 0..=1, default_missing_value = "strict")]
-    pub env_mode: EnvMode,
+    #[clap(long = "env-mode", num_args = 0..=1, default_missing_value = "strict")]
```
NicholasLYang, line 685: We no longer assume `Strict` when the flag is not here. This allows the config override logic to work correctly.

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -1452,28 +1452,13 @@ mod test {
         } ;
         "env_mode: not fully-specified"
     )]
-    #[test_case::test_case(
```
NicholasLYang, line 1455: test is no longer needed since this behavior is not needed.

crates/turborepo-api-client/src/lib.rs
```diff
@@ -123,6 +123,16 @@ pub struct APIAuth {
     pub team_slug: Option<String>,
 }
 
+impl std::fmt::Debug for APIAuth {
```
NicholasLYang, line 126: Minor detail because `OptsInputs` derives `Debug`. This is a custom impl because printing the `token` field seemed like a bad idea

dimitropoulos (approved):
great! deriving from OpsInputs rather than RunAndExecutionArgs seems like it's going to make adding `envMode` to the config a lot easier.  I think I still need an excalidraw diagram of all the layering to understand it, though, haha.

crates/turborepo-api-client/src/lib.rs
dimitropoulos: is it possible this caused the [CI failure](https://github.com/vercel/turbo/actions/runs/9947320122/job/27480069591?pr=8759) by obscuring the token?

```
turbo run test --filter=turborepo-tests-integration --color --env-mode=strict --token=*** --team=vercel
```

NicholasLYang: Hmm I don't think so. Lemme see what's causing that

PR #8775, opened 2024-07-17
improve failsafe_analyse

sokra:
### Description

This puts the results into the same cells to avoid unnecessary invalidations

Improves performance


ForsakenHarmony (approved):
could use some newlines 🙃

PR #8777, opened 2024-07-17
merge InProgressDirty into InProgress

sokra:
### Description

This saves some memory and streamlines the behavior.

It also lazily removes collectibles, children and dependencies even when entering the previous InProgressDirty (now InProgress stale) state.


bgw (approved):
Looks like this is a net-overall simplification, and I appreciate the added unit test. LGTM!

PR #8778, opened 2024-07-17
fix collectibles counting

sokra:
### Description

fix a bunch of cases where collectibles were not correctly counted

### DIFF
--- a/crates/turbo-tasks-memory/src/memory_backend.rs
+++ b/crates/turbo-tasks-memory/src/memory_backend.rs
@@ -13,7 +13,6 @@ use std::{
 };
 
 use anyhow::{bail, Result};
-use auto_hash_map::AutoMap;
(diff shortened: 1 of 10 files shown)


crates/turbo-tasks-memory/src/memory_backend.rs
```diff
@@ -470,7 +470,7 @@ impl Backend for MemoryBackend {
         trait_id: TraitTypeId,
         reader: TaskId,
         turbo_tasks: &dyn TurboTasksBackendApi<MemoryBackend>,
-    ) -> AutoMap<RawVc, i32> {
+    ) -> AutoMap<RawVc, i32, BuildHasherDefault<FxHasher>> {
```
ForsakenHarmony, line 473: Maybe create a type alias for this?

ForsakenHarmony (approved):

PR #8780, opened 2024-07-18
Minimal implementation of local Vcs

bgw:
# Description

Local Vcs store task-local values inside of the current task's state. The `SharedReference` holding onto their values is dropped when the task exits.

The contents of a local Vc must still use a refcounted `SharedReference`/`triomphe::Arc` because they can be turned into `ReadRef` objects (https://turbopack-rust-docs.vercel.sh/rustdoc/turbo_tasks/struct.ReadRef.html), which require a `'static` lifetime. We can experiment with adding a lifetime to `ReadRef` in a future iteration, but there's little advantage to doing this until we have local tasks.

# Limitations

This implements *just enough* to create and read local Vcs (with a test!). There are many things that are still unimplemented with `todo!()`. Most notably:

- We can't resolve a local `Vc` to a `ResolvedVc`.
- There's no way to return or pass a local `Vc` as an argument yet.
- For safety, we should only allow construction of local `Vc`s in functions where we know that the return value is a `ResolvedValue`.

Grand plan: https://www.notion.so/vercel/Resolved-Vcs-Vc-Lifetimes-Local-Vcs-and-Vc-Refcounts-49d666d3f9594017b5b312b87ddc5bff

# Memory Usage

- This increases the size of `Vc`/`RawVc` from 96 bits to 128 bits. With clever packing this could (in theory) be solved.  However, that it increase the number of machine words (2x 64bit), so it might not have any real impact after alignment happens.

- This increase the size of `CurrentTaskState`. I suspect this isn't too bad as there should be a smallish number of tasks active at any given time.

**I was not able to measure any change to peak memory/heap usage.**

Built using

```
cargo build --release -p next-build-test
```

Ran heaptrack on a single page (`/sink`) in `shadcn/ui` with:

```
cd ~/ui/apps/www
heaptrack ~/nextpack/target/release/next-build-test run sequential 1 1 '/sink'
```

And analyzed the results with

```
heaptrack --analyze /home/bgw.linux/ui/apps/www/heaptrack.next-build-test.3066837.zst | tail -20
```

### Before

```
total runtime: 130.25s.
calls to allocation functions: 48553541 (372786/s)
temporary memory allocations: 3863919 (29666/s)
peak heap memory consumption: 1.13G
peak RSS (including heaptrack overhead): 2.69G
total memory leaked: 4.96M
suppressed leaks: 7.24K
```

```
total runtime: 135.48s.
calls to allocation functions: 48619554 (358863/s)
temporary memory allocations: 3883888 (28667/s)
peak heap memory consumption: 1.12G
peak RSS (including heaptrack overhead): 2.70G
total memory leaked: 4.71M
suppressed leaks: 7.24K
```

### After

```
total runtime: 157.20s.
calls to allocation functions: 48509638 (308579/s)
temporary memory allocations: 3902883 (24827/s)
peak heap memory consumption: 1.13G
peak RSS (including heaptrack overhead): 2.70G
total memory leaked: 4.86M
suppressed leaks: 7.24K
```

```
total runtime: 130.25s.
calls to allocation functions: 48553541 (372786/s)
temporary memory allocations: 3863919 (29666/s)
peak heap memory consumption: 1.13G
peak RSS (including heaptrack overhead): 2.69G
total memory leaked: 4.96M
suppressed leaks: 7.24K
```

# Testing Instructions

```
```


bgw:
* **#8780** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8780?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/turbo-tasks/src/id.rs
```diff
@@ -70,6 +70,7 @@ define_id!(ValueTypeId: u32);
 define_id!(TraitTypeId: u32);
 define_id!(BackendJobId: u32);
 define_id!(ExecutionId: u64, derive(Debug));
+define_id!(LocalCellId: u32, derive(Debug));
```
arlyon, line 73: To be clear that means that each local task can have at most 2^32 child cells?

crates/turbo-tasks/src/manager.rs
```diff
@@ -490,53 +511,57 @@ impl<B: Backend + 'static> TurboTasks<B> {
         let future = async move {
             #[allow(clippy::blocks_in_conditions)]
             while CURRENT_TASK_STATE
-                .scope(Default::default(), async {
-                    if this.stopped.load(Ordering::Acquire) {
-                        return false;
-                    }
+                .scope(
+                    RefCell::new(CurrentTaskState::new(this.execution_id_factory.get())),
```
arlyon, line 515: maybe to make this diff a little clearer you can move this into a var?

crates/turbo-tasks/src/vc/mod.rs
```diff
@@ -282,6 +304,13 @@ where
     pub fn cell(inner: Inner) -> Self {
         <T::CellMode as VcCellMode<T>>::cell(inner)
     }
+
+    pub fn local_cell(inner: Inner) -> Self {
+        // `T::CellMode` isn't applicable here, we always create new local cells. Local
```
arlyon, line 309: Bit of a random question but just want to challenge my thinking. We have a few ways of interacting 'up the chain' so to speak. At the top level you have the global 'arena' with any number of 'subarenas' for local tasks.  Presumably you can interleave local and global tasks at will. Do we intend on having local areanas further down the call graph be able to read / access from Vcs in their 'superarenas' and global state?

```rust
#[turbo::function]
fn a(path: Vc<Path>) -> Vc<CompiledCode> {
  let is_dir = path.is_directory().await;
  if (is_dir) {
    return b(path);
  }
  // do stuff
}

#[turbo::function(local)]
fn b(path: Vc<Path>) -> Vc<CompiledCode> {
  // use path.is_directory()
}
```

Path::is_directory is already resolved in the global cache.

bgw: > Do we intend on having local areanas further down the call graph be able to read / access from Vcs in their 'superarenas' and global state?

This is a good question.

Global state: Yes. Local tasks will create local Vcs by default, but they will be able to both create and read resolved Vcs that use global cells. When creating global cells, those cells will be assigned to nearest non-local parent task.

In your example, `Path::is_directory` would not need to be re-executed because it's in the global cache.

"Superarenas" of calling tasks: No, at least for right now. Local Vcs must be resolved before being passed as an argument into another task.

One of the unintuitive challenges with reading from local "superarenas" is that (currently) every task is `tokio::spawn`ed and never canceled/aborted. This means that a child called task can outlive it's parent caller. This isn't common as most calling tasks await all their children, but can happen in the case of panics in the parent/calling task or with use of APIs like `tokio::select!`.

For memory safety we either need a mechanism for canceling tasks (triggering a drop at the await point) if they try to read from an expired caller's "superarena", or (more likely) we need to defer the cleanup of local arenas until all of their child spawned tasks exit.

bgw:
Migrated to the next.js repo: https://github.com/vercel/next.js/pull/68469

PR #8781, opened 2024-07-18
feat: Improve tree shaking

kdy1:
### Description

I'm still struggling to make next.js CI work well with tree shaking.

- https://github.com/vercel/next.js/pull/66689


crates/turbopack-ecmascript/src/tree_shake/util.rs
```diff
@@ -492,11 +492,11 @@ pub fn should_skip_tree_shaking(m: &Program, special_exports: &[RcStr]) -> bool
 }
 
 #[derive(Default)]
-struct UseServerFinder {
+struct ShouldAbort {
```
sokra, line 495: Could you rename that to `ShouldSkip`?

"Abort" as word could have a negative meaning in some contexts so we want to avoid it in general

crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -171,7 +177,7 @@ impl Analyzer<'_> {
                     if let Some(declarator) = &state.declarator {
```
sokra, line 177: What happens if the declarations is after the write:

``` js
a = 1;

var a;
```

sokra (approved):

crates/turbopack-ecmascript/src/tree_shake/mod.rs
sokra: Another test case:

``` js
a = 1;

function a() {}
```

crates/turbopack-ecmascript/tests/tree-shaker/analyzer/assign-before-decl-fn/output.md
```diff
@@ -0,0 +1,220 @@
...
+};
+
+```
+## Part 1
+```js
+import "__TURBOPACK_PART__" assert {
+    __turbopack_part__: 0
+};
+import { a } from "__TURBOPACK_PART__" assert {
+    __turbopack_part__: 0
```
kdy1, line 118: This is generated, so there's no `ReferenceError`.

PR #8815, opened 2024-07-23
Don't impl PartialEq/Eq for SharedReference and wrapper types

bgw:
### Description

Where feasible (I gave up on `TransientInstance<T>`) don't impl `PartialEq`, `Eq` or `Hash` for `SharedReference` or it's various wrapper types.

These were previously using identity for equality, but I strongly prefer a more explicit approach (akin to `Arc::ptr_eq`) when using identity equality because it can lead to caching bugs where the caller didn't realize and expected value equality.

### Testing Instructions

This breaks the nextjs build and would require a companion nextjs PR if we wanted to go forward with this.

### DIFF
--- a/crates/turbo-tasks/src/value.rs
+++ b/crates/turbo-tasks/src/value.rs
@@ -1,4 +1,4 @@
-use std::{fmt::Debug, marker::PhantomData, ops::Deref};
+use std::{fmt::Debug, hash::Hash, marker::PhantomData, ops::Deref};
 
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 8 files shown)


bgw:
* **#8815** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8815?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/turbo-tasks/src/value.rs
```diff
@@ -91,17 +91,17 @@ impl<T> Clone for TransientInstance<T> {
...
     }
 }
 
-impl<T> std::hash::Hash for TransientInstance<T> {
+impl<T> Eq for TransientInstance<T> {}
+
+impl<T> Hash for TransientInstance<T> {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        self.inner.hash(state);
+        self.inner.1.as_ptr().hash(state);
```
sokra, line 104: Maybe implement `ptr_hash` on SharedReference

crates/turbopack-browser/src/ecmascript/list/version.rs
```diff
@@ -19,6 +21,27 @@ pub(super) struct EcmascriptDevChunkListVersion {
...
+fn eq_index_map<K: Hash + Eq, V>(
+    this: &IndexMap<K, TraitRef<V>>,
+    other: &IndexMap<K, TraitRef<V>>,
+) -> bool {
+    if this.len() != other.len() {
+        return false;
+    }
+
+    this.iter()
+        .all(|(key, value)| other.get(key).map_or(false, |v| value.ptr_eq(v)))
```
sokra, line 33: Order should matter here. You can use `zip()` instead. That's also more performant.

PR #8816, opened 2024-07-23
refactor: remove `postcss` and `autoprefixer` in with-tailwind ui package

guesung:
### Description

I found that UI packages are compiling Tailwind CSS before executing other applications using the Tailwind CSS CLI. However, there were unnecessary packages included, specifically PostCSS and Autoprefixer. I believe this is unnecessary for the following reasons:

1. The [Tailwind CSS documentation](https://tailwindcss.com/docs/installation) does not mention PostCSS when using the Tailwind CLI.
2. In the [Tailwind CSS GitHub repository (CLI code)](https://github.com/tailwindlabs/tailwindcss/blob/6ab289343deaa61095a8f55799239ce0f1ee41ea/packages/%40tailwindcss-cli/src/commands/build/index.ts#L247C1-L273C1), there is code that uses 'postcss' internally:

```ts
function handleImports(
  input: string,
  file: string,
): [css: string, paths: string[]] | Promise<[css: string, paths: string[]]> {
  // ..
  return postcss()
    .use(atImport())
    .process(input, { from: file })
    .then((result) => [
      result.css,
      [file].concat(
        result.messages
          .filter((msg) => msg.type === 'dependency')
          .map((msg) => msg.file),
      ),
    ])
}
```

### DIFF
--- a/examples/with-tailwind/packages/ui/package.json
+++ b/examples/with-tailwind/packages/ui/package.json
@@ -26,8 +26,6 @@
     "@repo/tailwind-config": "workspace:*",
     "@repo/typescript-config": "workspace:*",
     "@types/react": "^18.2.61",
-    "autoprefixer": "^10.4.18",
-    "postcss": "^8.4.35",
     "tailwindcss": "^3.4.1",
     "typescript": "^5.3.3"
   }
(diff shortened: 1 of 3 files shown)


anthonyshew:
Are you able to inspect the outputs created by the Tailwind CLI with and without this change? Is there a difference?

guesung:
@anthonyshew 

Thank you for answering my request.

I inspected the 4.55% build time decrease (743ms -> 704ms) with and without the package install changes.

With such a small decrease, I think keeping Autoprefixer and PostCSS is better for browser compatibility and CSS processing.

Thank you.

anthonyshew:
I agree! Should we close this PR then?

guesung:
Yes, let's close the PR.

PR #8818, opened 2024-07-23
avoid implementing Eq on Completion

sokra:
### Description

It's probably not want to want. You want ptr_eq instead.


bgw (approved):
Makes sense to me!

kdy1 (approved):

PR #8820, opened 2024-07-23
Improve rustdoc

mischnic:
Closes PACK-3107

Packages importing themselves works fine for packages from npm, since they always live at `.../node_modules/foo` anyway and so the regular node_modules resolution covers self-references as well.

This is not the case for (symlinked) monorepo workspaces, which this PR fixes. This essentially adds the `LOAD_PACKAGE_SELF` step from https://nodejs.org/api/modules.html#all-together .

### DIFF
--- a/crates/turbopack-core/src/resolve/mod.rs
+++ b/crates/turbopack-core/src/resolve/mod.rs
@@ -2188,6 +2188,59 @@ async fn resolve_module_request(
 
     let mut results = vec![];
 
+    // Self references, if the nearest package.json has the name of the requested
+    // module. This matches only using the exports field and no other
+    // fields/fallbacks.
+    let package_json_context = find_context_file(lookup_path, package_json()).await?;
+    if let FindContextFileResult::Found(package_json_path, _refs) = &*package_json_context {
+        let read = read_package_json(*package_json_path).await?;
+        if let Some(json) = &*read {
+            if json["name"].as_str() == Some(module) {
+                if let ExportsFieldResult::Some(exports_field) =
+                    &*exports_field(*package_json_path).await?
+                {
+                    let Some(path) = path.clone().into_string() else {
+                        todo!("pattern into an exports field is not implemented yet");
+                    };
+
+                    let path = if &*path == "/" {
+                        ".".to_string()
+                    } else {
+                        format!(".{path}")
+                    };
+
+                    // TODO resolving inside of a package but still using ResolveIntoPackage
+                    for resolve_into_package in options_value.into_package.iter() {
+                        if let ResolveIntoPackage::ExportsField {
+                            conditions,
+                            unspecified_conditions,
+                        } = resolve_into_package
+                        {
+                            let package_path = package_json_path.parent();
+                            let result = handle_exports_imports_field(
+                                package_path,
+                                *package_json_path,
+                                options,
+                                exports_field,
+                                &path,
+                                conditions,
+                                unspecified_conditions,
+                                query,
+                            )
+                            .await?;
+                            if !(*result.is_unresolveable().await?) {
+                                results.push(result);
+                            }
+                        }
+                    }
+                }
+            }
+        }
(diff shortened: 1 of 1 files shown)


crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -2188,6 +2188,59 @@ async fn resolve_module_request(
...
+                                conditions,
+                                unspecified_conditions,
+                                query,
+                            )
+                            .await?;
+                            if !(*result.is_unresolveable().await?) {
+                                results.push(result);
+                            }
+                        }
+                    }
```
sokra, line 2235: This should just call `resolve_into_package`

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -2188,6 +2188,59 @@ async fn resolve_module_request(
...
     let mut results = vec![];
 
+    // Self references, if the nearest package.json has the name of the requested
+    // module. This matches only using the exports field and no other
+    // fields/fallbacks.
+    let package_json_context = find_context_file(lookup_path, package_json()).await?;
+    if let FindContextFileResult::Found(package_json_path, _refs) = &*package_json_context {
+        let read = read_package_json(*package_json_path).await?;
+        if let Some(json) = &*read {
+            if json["name"].as_str() == Some(module) {
```
sokra, line 2198: To improve caching on that part, I think it's worth moving this part into a separate turbo-tasks function that returns the potential self reference for a lookup_path:

``` rs
#[turbo_tasks::value]
enum SelfReferencePackageResult {
    Found {
        name: String,
        path: Vc<FileSystemPath>,
    }
    NotFound
}

fn find_self_reference(lookup_path: Vc<FileSystemPath>) -> Vc<SelfReferencePackageResult>
```

crates/turbopack-core/src/resolve/mod.rs
mischnic: At least Node's resolution algorithm (and therefore if you run a transpiled, unbundled server with Node) only allows self-imports if there's an `exports` field https://nodejs.org/api/modules.html#all-together
```
LOAD_PACKAGE_SELF(X, DIR)
1. Find the closest package scope SCOPE to DIR.
2. If no scope was found, return.
3. If the SCOPE/package.json "exports" is null or undefined, return.
4. If the SCOPE/package.json "name" is not the first segment of X, return.
5. let MATCH = PACKAGE_EXPORTS_RESOLVE(pathToFileURL(SCOPE),
   "." + X.slice("name".length), `package.json` "exports", ["node", "require"])
6. RESOLVE_ESM_MATCH(MATCH)
```

I'm not sure how Webpack behaves here and then how strict Turbopack should be?

sokra: Yep. Turbopack should behave the same. Only use that if `exports` field is present. This could be handled by `find_self_reference`.

I just want to avoid repeating the exports field handling code.

crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -2186,7 +2219,25 @@ async fn resolve_module_request(
...
+    {
+        if name == module {
+            let result = resolve_into_package(
+                Value::new(path.clone()),
+                *package_path,
+                query,
+                fragment,
+                options,
+            );
+            if !(*result.is_unresolveable().await?) {
```
sokra, line 2236: Should we continue to look for other packages with the same name or should this better fail resolving here if the self reference doesn't resolve?

sokra (approved):

crates/turbopack-core/src/resolve/mod.rs
mischnic: I think that's an uncommon case, but https://nodejs.org/api/modules.html#all-together would continue searching. So I'm inclined to continue as well

PR #8822, opened 2024-07-23
chore: add trace level logging to process management

chris-olszewski:
### Description

Add additional detailed logging to child process management to aid in finding issues causing https://github.com/vercel/turbo/issues/8281

### Testing Instructions

`rustc`


crates/turborepo-lib/src/process/child.rs
```diff
@@ -832,6 +846,7 @@ mod test {
 
     #[test_case(false)]
     #[test_case(TEST_PTY)]
+    #[tracing_test::traced_test]
```
chris-olszewski, line 849: Useful for testing the new logs

anthonyshew (approved):

PR #8843, opened 2024-07-25
Add/move tests for Vc generics

bgw:
## Description

- Move the tests out of `all_in_one` so that we can see exactly which parts are failing more easily.
- Add a few more tests. **The newly added `test_read_ref_round_trip` fails** (and is ignored), indicating a bug in the generics implementation!!! This is fixed in #8845.

## Why does `test_read_ref_round_trip` fail?

The theory is that Vcs are supposed to be stored as `<<T as VcValueType>::Read as VcRead<T>>::Repr`. However, it looks like due to an oversight, `ReadRef::cell` is trying to store the value as `T`.

## Why add these tests at all?

The plan (as of yesterday) is to remove support for generics, but I'd feel more confident if I can fix the casting issues in my local Vc PRs before I do that.

These tests should help me understand what's happening and debug things.

## Testing Instructions

```
```

### DIFF
--- a/crates/turbo-tasks-memory/tests/generics.rs
+++ b/crates/turbo-tasks-memory/tests/generics.rs
@@ -1,0 +1,203 @@
+#![feature(arbitrary_self_types)]
+
+use std::sync::{Arc, Mutex};
+
+use indexmap::{IndexMap, IndexSet};
+use turbo_tasks::{debug::ValueDebug, Invalidator, ReadRef, TaskId, Vc};
+use turbo_tasks_testing::{register, run, Registration};
+
+static REGISTRATION: Registration = register!();
+
+#[tokio::test]
+async fn test_option_some() {
+    run(&REGISTRATION, async move {
+        let vc_42 = Vc::cell(42);
+        let option: Vc<Option<Vc<u32>>> = Vc::cell(Some(vc_42));
+        assert_eq!(*option.is_some().await.unwrap(), true);
+        assert_eq!(*option.is_none().await.unwrap(), false);
+        assert_eq!(&*option.await.unwrap(), &Some(vc_42));
+        assert_eq!(option.dbg().await.unwrap().to_string(), "Some(\n    42,\n)");
+    })
+    .await
+}
+
+#[tokio::test]
+async fn test_option_none() {
+    run(&REGISTRATION, async move {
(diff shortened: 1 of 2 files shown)


bgw:
* **#8843** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8843?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/turbo-tasks-memory/tests/generics.rs
```diff
@@ -0,0 +1,203 @@
...
+// Test that we can convert a `Vc` to a `ReadRef`, and then back to a `Vc`.
+#[tokio::test]
+#[ignore = "TODO: This panics! There's a casting bug in the generics implementation."]
+async fn test_read_ref_round_trip() {
+    run(&REGISTRATION, async move {
+        let c: Vc<Option<Vc<u8>>> = Vc::cell(Some(Vc::cell(1)));
+        let _ = ReadRef::cell(c.await.unwrap()).await.unwrap();
+    })
+    .await
+}
```
bgw, line 157: This is a bug in the current generics implementation.

arlyon (commented):
> The plan (as of yesterday) is to remove support for generics, but I'd feel more confident if I can fix the casting issues in my local Vc PRs before I do that.

Mind elaborating a little about removing generics?

bgw:
> Mind elaborating a little about removing generics? 

`Vc` has very limited support for generic collections, such as `Vc<Vec<Vc<T>>>`: https://github.com/vercel/turbo/pull/5738

We discussed this during office hours, and decided that:
1. The hack that generics use is not generalizable to all generics.
2. Generics require a lot of support code with unsafe transmutes and rely on a lot of details about memory layout. It's very hard to get this right (and as I'm discovering, there are bugs!)
3. There's not a lot of uses of these generics, though there might be more if there was more awareness of this feature.
4. Switching things from `Vc` to `ResolvedVc` is going to add more complexity to the generics code.
5. Transparent value types provide 90% of the benefits of generics.

sokra (approved):

bgw:
### Merge activity

* **Jul 29, 4:41 PM EDT**: @bgw merged this pull request with [Graphite](https://app.graphite.dev/github/pr/vercel/turbo/8843).

PR #8845, opened 2024-07-25
Fix `ReadRef<T>::cell` when `T` != `T::Read::Repr`

bgw:
## Description

Fixes the test case introduced in #8843.

The theory is that Vcs are supposed to be stored as `<<T as VcValueType>::Read as VcRead<T>>::Repr`. However, it looks like due to an oversight, `ReadRef::cell` is trying to store the value as `T`.

This introduces a way to perform 

## Testing Instructions

```
cargo nextest r -p turbo-tasks -p turbo-tasks-memory
```

PR #8846, opened 2024-07-26
Use `VcCellMode` in `ReadRef::cell`, respecting shared/new semantics

bgw:

PR #8847, opened 2024-07-26
Make ReadRef using `VcCellMode` semantics, add `VcCellMode::raw_cell` API

bgw:
## Description

More fixes to `ReadRef` on top of #8845. There's a bit of a cascade of changes that needed to happen here:

- `ReadRef::cell` should actually use `VcCellMode` so that it supports new/shared semantics correctly.
- ... however, `VcCellMode` needs to expose an API (`raw_cell`) for working with `SharedReference` so that we can reuse the `SharedReference`/`Arc` inside of `ReadRef` instead of cloning the data inside of it.
- ... however, `CurrentCellRef` needs to expose a `compare_and_update` method that works for `SharedReference` (`compare_and_update_with_shared_reference`).
- ... however, the naming conventions for the `CurrentCellRef` methods are a bit confusing at this point, so rename them.

## Bigger Picture

This my sneaky way of introducing the `raw_cell` API as part of a smaller PR, since that API will be needed for converting a local `Vc` to a resolved `Vc`. That's a similar problem of needing to reuse a `SharedReference` along with `VcCellMode` semantics.

## Testing Instructions

```
```

### DIFF
--- a/crates/turbo-tasks/src/manager.rs
+++ b/crates/turbo-tasks/src/manager.rs
@@ -1522,57 +1522,144 @@ pub struct CurrentCellRef {
     index: CellId,
 }
 
+type VcReadRepr<T> = <<T as VcValueType>::Read as VcRead<T>>::Repr;
+
 impl CurrentCellRef {
-    pub fn conditional_update_shared<
-        T: VcValueType + 'static,
-        F: FnOnce(Option<&T>) -> Option<T>,
-    >(
+    /// Updates the cell if the given `functor` returns a value.
+    pub fn conditional_update<T>(&self, functor: impl FnOnce(Option<&T>) -> Option<T>)
+    where
+        T: VcValueType,
+    {
+        self.conditional_update_with_shared_reference(|old_shared_reference| {
+            let old_ref = old_shared_reference
+                .and_then(|sr| sr.0.downcast_ref::<VcReadRepr<T>>())
+                .map(|content| <T::Read as VcRead<T>>::repr_to_value_ref(content));
+            let new_value = functor(old_ref)?;
+            Some(SharedReference::new(triomphe::Arc::new(
+                <T::Read as VcRead<T>>::value_to_repr(new_value),
+            )))
+        })
+    }
+
+    /// Updates the cell if the given `functor` returns a `SharedReference`.
+    pub fn conditional_update_with_shared_reference(
         &self,
-        functor: F,
+        functor: impl FnOnce(Option<&SharedReference>) -> Option<SharedReference>,
     ) {
         let tt = turbo_tasks();
-        let content = tt
-            .read_own_task_cell(self.current_task, self.index)
-            .ok()
-            .and_then(|v| v.try_cast::<T>());
-        let update =
-            functor(content.as_deref().map(|content| {
-                <<T as VcValueType>::Read as VcRead<T>>::target_to_value_ref(content)
-            }));
+        let cell_content = tt.read_own_task_cell(self.current_task, self.index).ok();
+        let update = functor(cell_content.as_ref().and_then(|cc| cc.1 .0.as_ref()));
         if let Some(update) = update {
-            tt.update_own_task_cell(
-                self.current_task,
-                self.index,
-                CellContent(Some(SharedReference::new(triomphe::Arc::new(update)))),
-            )
+            tt.update_own_task_cell(self.current_task, self.index, CellContent(Some(update)))
         }
     }
 
-    pub fn compare_and_update_shared<T: PartialEq + VcValueType + 'static>(&self, new_content: T) {
-        self.conditional_update_shared(|old_content| {
-            if let Some(old_content) = old_content {
-                if PartialEq::eq(&new_content, old_content) {
+    /// Replace the current cell's content with `new_value` if the current
+    /// content is not equal by value with the existing content.
+    ///
+    /// The comparison happens using the value itself, not the
+    /// [`VcRead::Target`] of that value.
+    ///
+    /// Take this example of a custom equality implementation on a transparent
+    /// wrapper type:
+    ///
+    /// ```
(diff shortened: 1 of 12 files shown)


bgw:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/vercel/turbo/8847?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#8847** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#8846** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8846?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8845** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8845?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8843** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8843?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/turbo-tasks/src/read_ref.rs
```diff
@@ -244,14 +244,15 @@ where
...
-        }));
+        };
         Vc {
-            node: local_cell.into(),
+            node: <T::CellMode as VcCellMode<T>>::raw_cell(
+                SharedReference::new(value).typed(type_id),
+            ),
             _t: PhantomData,
         }
     }
```
bgw, line 258: Essentially the problem here was that `ReadRef` can't/shouldn't implement `cell` itself. It must defer to `VcCellMode`'s implementation, as only that trait's implementation knows how to correctly construct a cell.

crates/turbo-tasks/src/vc/cast.rs
```diff
@@ -27,19 +27,7 @@ where
...
-            unsafe {
-                // Downcast the cell content to the expected representation type, then
-                // transmute it to the expected type.
-                // See https://users.rust-lang.org/t/transmute-doesnt-work-on-generic-types/87272/9
-                std::mem::transmute_copy::<
-                    ManuallyDrop<ReadRef<<T::Read as VcRead<T>>::Repr>>,
-                    Self::Output,
-                >(&ManuallyDrop::new(content.cast()?))
-            },
-        )
```
bgw, line 42: It turns out that this transmute wasn't entirely correct. You can't do a straight transmute between two `ReadRef<T>`s because it contains a `triomphe::Arc<T>`, and the vtable half of `triomphe::Arc<T>`'s fat pointer is different depending on the type of `T`.

You need to use something like `unchecked_sidecast_triomphe_arc` so that the fat pointer is constructed correctly. I modified the `TypedCellContent::cast` method in this PR to do the cast of `Arc` for us before constructing the `ReadRef`.

crates/turbo-tasks/src/vc/cell_mode.rs
```diff
@@ -1,16 +1,25 @@
...
 where
     T: VcValueType,
 {
-    fn cell(value: <T::Read as VcRead<T>>::Target) -> Vc<T>;
+    /// Create a new cell.
+    fn cell(value: VcReadTarget<T>) -> Vc<T>;
+
+    /// Create a type-erased `RawVc` cell given a pre-existing type-erased
+    /// `SharedReference`. In some cases, we will re-use the shared value.
+    fn raw_cell(value: TypedSharedReference) -> RawVc;
```
bgw, line 22: Local Vc resolution needs this `VcCellMode::raw_cell` method too because we need to be able to construct a new cell inside of `RawVc` where we don't have compile-time information about the concrete type `T` (but we do have the type id).

crates/turbo-tasks/src/trait_ref.rs
```diff
@@ -102,7 +102,7 @@ where
         // See Safety clause above.
         let TypedSharedReference(ty, shared_ref) = trait_ref.shared_reference;
         let local_cell = find_cell_by_type(ty);
-        local_cell.update_shared_reference(shared_ref);
+        local_cell.update_with_shared_reference(shared_ref);
         let raw_vc: RawVc = local_cell.into();
         raw_vc.into()
     }
```
bgw, line 108: <i class='graphite__hidden'>[Re: lines 99 to 108]<br/><br/></i>

`TraitRef::cell` has similar problems to `ReadRef::cell` here, and needs to also call `VcCellMode::raw_cell`, but that's a bit more complicated as I need a mapping of type ids to `raw_vc` function pointers. I'll do that in a subsequent PR.<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/vercel/turbo/8847?utm_source=unchanged-line-comment">Graphite</a>.</span>

crates/turbo-tasks/src/manager.rs
```diff
@@ -1522,57 +1522,144 @@ pub struct CurrentCellRef {
...
-            Some(new_content)
+            Some(new_value)
         });
     }
 
-    pub fn update_shared<T: VcValueType + 'static>(&self, new_content: T) {
+    /// Replace the current cell's content with `new_shared_reference` if the
+    /// current content is not equal by value with the existing content.
+    ///
+    /// If you already have a `SharedReference`, this is a faster version of
```
arlyon, line 1607: Perhaps it is useful to have the reciprocal comment?

crates/turbo-tasks/src/vc/cell_mode.rs
arlyon: > In some cases, we will re-use the shared value.

When is this the case?

sokra (approved):

bgw:
Migrated to the next.js repo: https://github.com/vercel/next.js/pull/68467

PR #8852, opened 2024-07-26
Lookup alias and import maps using Pattern instead of String

mischnic:
The changes I had to make got out of hand quite quickly, because operating on `Pattern`s everywhere is more cumbersome than strings...
Furthermore, some structs contained strings (but unrelated to requests being treated as strings, rather because import maps are specified as strings with wildcards), so now we needed a version with patterns.

This is almost working, namely the dynamic `@/*` alias is already rewritten correctly now:
```rust
//Request:
Pattern::Concatenation([Pattern::Constant("@/messages/"), Pattern::Dynamic, Pattern::Constant(".json")]);

//ImportMap::lookup result:
ReplacedImportMapping::PrimaryAlternative(
    Pattern::Concatenation([Pattern::Constant("./messages/"), Pattern::Dynamic]),
    Some(
        FileSystemPath {
            fs: DiskFileSystem {
                name: "project",
                root: "/Users/niklas/Desktop/issue-turbopack-imports",
            },
            path: ,
        },
    ),
);
```

Closes PACK-2801
Related to https://github.com/vercel/next.js/issues/63372


mischnic:
So now the output of
```js
    const messages = await import(`@/messages/${locale}.json`);
```
is
```js
   const messages = await __turbopack_module_context__({
        "./messages/de.json": {
            id: ()=>"[project]/messages/de.json (json, async loader)",
            module: ()=>__turbopack_require__("[project]/messages/de.json (json, async loader)")(__turbopack_import__)
        },
        "./messages/en.json": {
            id: ()=>"[project]/messages/en.json (json, async loader)",
            module: ()=>__turbopack_require__("[project]/messages/en.json (json, async loader)")(__turbopack_import__)
        }
    }).import(`@/messages/${locale}.json`)
```
which fails because `__turbopack_module_context__` receives the alias-resolved paths while the looked-up path still has the `@/`

crates/turbopack-core/src/resolve/alias_map.rs
```diff
@@ -192,14 +193,16 @@ impl<T> AliasMap<T> {
...
     where
         T: Debug,
     {
         // Invariant: prefixes should be sorted by increasing length (base lengths),
         // according to PATTERN_KEY_COMPARE. Since we're using a prefix tree, this is
         // the default behavior of the common prefix iterator.
-        let common_prefixes = self.map.common_prefixes(request.as_bytes());
+        let common_prefixes = self
+            .map
+            .common_prefixes(request.constant_prefix().as_bytes());
```
sokra, line 205: This will lead to also returning non-matching results. e. g. Pattern could be `"a" | "b"` which has no constant_prefix so it would return `"c"` too.

You probably want:
* Do multiple lookups internally when you have Alternatives
* Filter the results by matching with the Pattern

crates/turbopack-core/src/resolve/alias_map.rs
```diff
@@ -225,7 +228,9 @@ impl<T> AliasMap<T> {
...
```
sokra, line 233: Same

crates/turbopack-core/src/resolve/alias_map.rs
```diff
@@ -465,22 +470,25 @@ where
             for (key, template) in &mut *current_prefix_iterator {
                 match key {
                     AliasKey::Exact => {
-                        if self.request.len() == prefix.len() {
-                            return Some(AliasMatch::Exact(template));
+                        if self
+                            .request
+                            .as_string()
```
sokra, line 475: The Pattern might not be convertable to a string, but it could still match an Exact alias

crates/turbopack-core/src/resolve/pattern.rs
```diff
@@ -83,6 +123,211 @@ impl Pattern {
...
+            let mut result = Pattern::alternatives(alternatives.map(|pat| {
+                let mut split = template.split("*");
+                let mut concatenation: Vec<Pattern> = Vec::with_capacity(3);
+
+                // There are at least two elements in the iterator
+                concatenation.push(Pattern::Constant(split.next().unwrap().into()));
+
+                for part in split {
+                    concatenation.push(pat.clone());
+                    concatenation.push(Pattern::Constant(part.into()));
```
sokra, line 318: You might want to wrap that line in `if !part.is_empty() {}`

crates/turbopack-core/src/resolve/options.rs
```diff
@@ -362,25 +404,25 @@ impl ImportMap {
         lookup_path: Vc<FileSystemPath>,
         request: Vc<Request>,
     ) -> Result<ImportMapResult> {
-        // TODO lookup pattern
         // relative requests must not match global wildcard aliases.
-        if let Some(request_string) = request.await?.request() {
-            let mut lookup = if request_string.starts_with("./") {
+
+        if let Some(request_pattern) = request.await?.request_pattern() {
+            let mut lookup = if request_pattern.must_match("./") {
```
sokra, line 410: A hard if else won't work here. The pattern could match both paths.

You can use `split_could_match` to split the Pattern into the part the matches `"./"` and the part that doesn't.

crates/turbopack-core/src/resolve/alias_map.rs
mischnic: Looks like this doesn't happen, `map.common_prefixes("")` is just empty. That does align with example on https://docs.rs/patricia_tree/latest/patricia_tree/map/type.PatriciaMap.html#method.common_prefixes

But still, `map.lookup(Pattern::Alternative(...))` currently never returns anything even if some/all alternatives match. And even then, just returning `enum AliasMatch<R> {Exact(R), Replaced(R)}` is wrong anyway because some might be exact and some might be replaced matches

mischnic:
-> https://github.com/vercel/next.js/pull/68274

PR #8853, opened 2024-07-26
chore(auth): catch forbidden cache access error

chris-olszewski:
### Description

Catch 403 HTTP requests and understand them as meaning this token doesn't have cache access. This will cause the token to be ignored and the login flow will continue.

### Testing Instructions

Inspecting result of sending cache status request with bad token. Result is a `ReqwestError` with a 403 status code, not an `UnknownStatus`.

### DIFF
--- a/crates/turborepo-auth/src/lib.rs
+++ b/crates/turborepo-auth/src/lib.rs
@@ -200,6 +200,12 @@ impl Token {
                 turborepo_api_client::Error::UnknownStatus { code, .. } if code == "forbidden" => {
                     Ok(false)
                 }
+                // If the entire request fails with 403 also return false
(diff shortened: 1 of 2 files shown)


crates/turborepo-auth/src/lib.rs
```diff
@@ -200,6 +200,12 @@ impl Token {
                 turborepo_api_client::Error::UnknownStatus { code, .. } if code == "forbidden" => {
```
chris-olszewski, line 200: I'm unsure if this was ever valid, but keeping in case there's a way this can get hit that I'm not seeing. Timeboxing this fix so not trying to fully trace responses we can get from this endpoint.

NicholasLYang (approved):

PR #8854, opened 2024-07-26
fix(watch): properly shut down persistent tasks

chris-olszewski:
### Description

Previously when we decided we needed to restart persistent tasks we would just call `abort()` on the task handle. This was incorrect as our child processes don't get shut down on drop. This resulted in ports not getting freed up from these zombie tasks.

The fix is to create a new `RunStopper` struct which just has a handle on the child process manager so we can shut down the individual children before aborting the entire task. We also move the persistent task shut down to happen before we update the TUI task list so we don't finish tasks that never started.

### Testing Instructions

Before:

https://github.com/user-attachments/assets/54630a33-44f9-4d73-b70d-5e0800b239fa

After:

https://github.com/user-attachments/assets/f5379795-46f2-424b-b136-1da345882106


crates/turborepo-lib/src/run/mod.rs
```diff
@@ -215,6 +222,11 @@ impl Run {
         if let Some(subscriber) = self.signal_handler.subscribe() {
             let run_cache = self.run_cache.clone();
             tokio::spawn(async move {
+                // Caching is disabled for watch so we don't need to wait on shutting down the
```
NicholasLYang, line 225: nit: should we do this if caching is disabled, not just in the watch case?

NicholasLYang (approved):

Archivedent:
How is the restart meant to be triggered? If I manually ctrl+break it doesn't seem to shut down the child process properly, and port conflict still remains.

PR #8858, opened 2024-07-26
fix(watch): exit command if persistent task exits

chris-olszewski:
### Description

Now exit if persistent tasks exit. Previously we would not check if persistent tasks exited leading to situations where the `TUI` would shut down, but we still had `watch` running.

### Testing Instructions

Before:

https://github.com/user-attachments/assets/384c689a-d381-41c8-833a-6004360ad374

Notice that the TUI exits, but `turbo` does not.

After:

https://github.com/user-attachments/assets/cf0b607c-1c28-47d1-94c2-801c93927c39

### DIFF
--- a/crates/turborepo-lib/src/run/watch.rs
+++ b/crates/turborepo-lib/src/run/watch.rs
@@ -100,6 +100,8 @@ pub enum Error {
     PackageChange(#[from] tonic::Status),
     #[error("could not connect to UI thread")]
     UISend(String),
+    #[error("persistent tasks exited unexpectedly")]
+    PersistentExit,
 }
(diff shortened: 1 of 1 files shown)


NicholasLYang (approved):

NicholasLYang:
Only note I have is that we could probably create an integration test for this

chris-olszewski:
Resurrecting this PR in response to https://github.com/vercel/turborepo/issues/10122

This never got merged as adding the suggested integration test uncovered a race condition in this PR. Race was that sometimes logs for the exited persistent task would get printed twice. 🙃

PR #8864, opened 2024-07-29
fix: Fix tree shaking even more

kdy1:
### Description

Fix server action and rename the injected variable.

The previous code fails if the name of the top-level variable is `v`. This was not a problem for almost all modules, but some packages like [`@firebase/webchannel-wrapper`](https://unpkg.com/@firebase/webchannel-wrapper@0.2.41/dist/index.esm.js) is published after mangling, so it can be a problem

### Testing Instructions

See https://github.com/vercel/next.js/pull/66689

### DIFF
--- a/crates/turbopack-ecmascript/src/references/esm/export.rs
+++ b/crates/turbopack-ecmascript/src/references/esm/export.rs
@@ -442,8 +442,9 @@ impl CodeGenerateable for EsmExports {
                 EsmExport::LocalBinding(name, mutable) => {
                     if *mutable {
                         Some(quote!(
-                            "([() => $local, (____v) => $local = ____v])" as Expr,
-                            local = Ident::new((name as &str).into(), DUMMY_SP)
+                            "([() => $local, ($new) => $local = $new])" as Expr,
+                            local = Ident::new((name as &str).into(), DUMMY_SP),
+                            new = Ident::new(format!("{name}_new_value").into(), DUMMY_SP),
                         ))
                     } else {
                         Some(quote!(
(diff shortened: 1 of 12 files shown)


crates/turbopack-ecmascript/src/references/esm/export.rs
```diff
@@ -442,7 +442,7 @@ impl CodeGenerateable for EsmExports {
                 EsmExport::LocalBinding(name, mutable) => {
                     if *mutable {
                         Some(quote!(
-                            "([() => $local, (v) => $local = v])" as Expr,
```
kdy1, line 445: The previous code fails if the name of the top-level variable is `v`.

This was not a problem for almost all modules, but some packages like [`@firebase/webchannel-wrapper`](https://unpkg.com/@firebase/webchannel-wrapper@0.2.41/dist/index.esm.js) is published after mangling, so it can be a problem

crates/turbopack-ecmascript/src/tree_shake/asset.rs
```diff
@@ -175,7 +175,7 @@ impl Module for EcmascriptModulePartAsset {
 impl Asset for EcmascriptModulePartAsset {
     #[turbo_tasks::function]
     fn content(&self) -> Vc<AssetContent> {
-        todo!()
+        self.full_module.content()
```
kdy1, line 178: The value is still not used.

crates/turbopack-ecmascript/src/references/esm/export.rs
sokra: Now it fails for `___v`. Could we name it `name + "_new_value"`?

sokra (approved):

PR #8868, opened 2024-07-29
fix(turborepo): task table width bug

NicholasLYang:
### Description

The task status was getting cut off because we were clamping an incorrect length for the table instructions. Moved the instructions for the table into a constant so we can use that for the clamp call.

### Testing Instructions

Try it and see.


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -14,11 +14,7 @@ use turbopath::AbsoluteSystemPathBuf;
 use turborepo_api_client::AnonAPIClient;
 use turborepo_repository::inference::{RepoMode, RepoState};
 use turborepo_telemetry::{
-    events::{
```
NicholasLYang, line 17: Shouldn't be in this PR but I was getting an unused import warning

tknickman (approved):

PR #8872, opened 2024-07-30
feat(turborepo): log reason why all packages were considered changed

mehulkar:
The ultimate goal for me is to figure out why I'm seeing _all_ packages invalidated when running with `--filter='[sha]'` . In a lot of cases, we return a sigil for `PackageChanges::All`. I'm not sure how to bubble this up with higher fidelity information in a good way, so this PR settles for a more granular "reason" why all packages changed and includes it in debug logging. This should help a little bit with debugging.

### DIFF
--- a/crates/turborepo-repository/src/change_mapper/mod.rs
+++ b/crates/turborepo-repository/src/change_mapper/mod.rs
@@ -78,12 +78,10 @@ impl<'a, PD: PackageChangeMapper> ChangeMapper<'a, PD> {
         let filtered_changed_files = self.filter_ignored_files(changed_files.iter())?;
 
         match self.get_changed_packages(filtered_changed_files.into_iter()) {
-            PackageChanges::All(reason) => {
-                return Ok(PackageChanges::All(reason));
-            }
+            PackageChanges::All(reason) => Ok(PackageChanges::All(reason)),
 
             PackageChanges::Some(mut changed_pkgs) => {
-                return match lockfile_change {
+                match lockfile_change {
                     Some(LockfileChange::WithContent(content)) => {
                         // if we run into issues, don't error, just assume all packages have changed
                         let Ok(lockfile_changes) = self.get_changed_packages_from_lockfile(content)
@@ -108,9 +106,9 @@ impl<'a, PD: PackageChangeMapper> ChangeMapper<'a, PD> {
                         AllPackageChangeReason::LockfileChangedWithoutDetails,
                     )),
                     None => Ok(PackageChanges::Some(changed_pkgs)),
-                };
+                }
             }
-        };
+        }
     }
 
     fn filter_ignored_files<'b>(
--- a/crates/turborepo-repository/src/change_mapper/mod.rs
+++ b/crates/turborepo-repository/src/change_mapper/mod.rs
@@ -151,6 +151,7 @@ impl<'a, PD: PackageChangeMapper> ChangeMapper<'a, PD> {
                     changed_packages.insert(pkg);
                 }
                 PackageMapping::All => {
+                    debug!("all packages changed due to {file:?}");
                     return PackageChanges::All(AllPackageChangeReason::NonPackageFileChanged);
                 }
                 PackageMapping::None => {}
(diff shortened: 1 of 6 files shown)


crates/turborepo-repository/src/change_mapper/mod.rs
```diff
@@ -61,31 +70,40 @@ impl<'a, PD: PackageChangeMapper> ChangeMapper<'a, PD> {
...
+            ));
         }
 
         // get filtered files and add the packages that contain them
         let filtered_changed_files = self.filter_ignored_files(changed_files.iter())?;
-        let PackageChanges::Some(mut changed_pkgs) =
-            self.get_changed_packages(filtered_changed_files.into_iter())
-        else {
-            return Ok(PackageChanges::All);
-        };
```
mehulkar, line 73: I wanted to pass through the reason to `PackageChanges::All` here, so converted this to a longer `match` resulting in a larger diff in this file.

crates/turborepo-repository/src/change_mapper/mod.rs
```diff
@@ -23,9 +23,18 @@ pub enum LockfileChange {
     WithContent(Vec<u8>),
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub enum AllPackageChangeReason {
+    DefaultGlobalFileChanged,
+    LockfileChangeDetectionFailed,
+    LockfileChangedWithoutDetails,
+    RootInternalDepChanged,
+    NonPackageFileChanged,
```
mehulkar, line 32: I called this reason "NonPackageFileChanged" instead of "GlobalFileChanged", because the latter is a little bit confusing. E.g. `.github/actions/whatever.yml` is a global file and so is `examples/some-app/whatever.js` if neither path falls in the workspace config. It's not always clear that these are "global files" as far as turbo is concerned. I'm open to better name though.

chris-olszewski (approved):

crates/turborepo-repository/src/change_mapper/package.rs
```diff
@@ -172,7 +172,10 @@ mod tests {
 
         // We should return All because we don't have global deps and
         // therefore must be conservative about changes
-        assert_eq!(package_changes, PackageChanges::All);
+        assert_eq!(
+            package_changes,
+            PackageChanges::All(AllPackageChangeReason::DefaultGlobalFileChanged)
```
chris-olszewski, line 177:
```suggestion
            PackageChanges::All(AllPackageChangeReason::NonPackageFileChanged)
```

crates/turborepo-repository/src/change_mapper/mod.rs
```diff
@@ -120,18 +140,18 @@ impl<'a, PD: PackageChangeMapper> ChangeMapper<'a, PD> {
...
                 }
                 PackageMapping::Package(pkg) => {
-                    debug!("package {pkg:?} changed due to {file:?}");
+                    debug!("{} changes \"{}\"", file.to_string(), pkg.name);
                     changed_packages.insert(pkg);
                 }
                 PackageMapping::All => {
-                    debug!("all packages changed due to {file:?}");
-                    return PackageChanges::All;
+                    return PackageChanges::All(AllPackageChangeReason::NonPackageFileChanged);
```
mehulkar, line 154:
```suggestion
                    debug!("all packages changed due to {file:?}");
                    return PackageChanges::All(AllPackageChangeReason::NonPackageFileChanged);
```

mehulkar:
@chris-olszewski I think this PR is ok to merge[^1], but I'm not feeling amazing about it either. The new `reason` enum is ok for logging, but really what I want is some detail about what created that reason in the first place. This is tricky because in _some_ cases, it'll just be a file reference, and in some cases it's the name of a package. E.g.:

- `DefaultGlobalFileChanged`: name of a file
- `LockfileChangeDetectionFailed`: no extra details needed? 
- `LockfileChangedWithoutDetail`: no details needed?
- `RootInternalDepChanged`: name of a package (it's always only one becuase we early return and don't get _all_ root internal packages that changed
- `NonPackageFileChanged`: a list of files that are considered global

The "reason" is useful, but I had some trouble threading through the _detail_ that would _actually_ be helpful for debugging. 

Let me know if you have suggestions/thoughts around this and if you decide not to take this PR as-is, that's also fine with me. 


[^1]: assuming tests pass, I gave up because it looked like a race condition for that test, but maybe I'm wrong

PR #8873, opened 2024-07-30
smarter logical combination that keeps truthy/falsy/nullish property

sokra:
### Description

Allows to handle `if (unknown && falsy)` `if (unknown || truthy)` `if (unknown ?? nullish)` in static analysis.

Hint: review each commit on it's own, since the first commit is only changing to a named struct.

### DIFF
--- a/crates/turbopack-ecmascript/src/analyzer/builtin.rs
+++ b/crates/turbopack-ecmascript/src/analyzer/builtin.rs
@@ -105,6 +105,7 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
             JsValue::Alternatives {
                 total_nodes: _,
                 values,
+                additional_property: _,
             } => {
                 *value = JsValue::alternatives(
                     take(values)
@@ -163,6 +164,7 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                     JsValue::Alternatives {
                         total_nodes: _,
                         values,
+                        additional_property: _,
                     } => {
                         *value = JsValue::alternatives(
                             take(values)
@@ -306,6 +308,7 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                     JsValue::Alternatives {
                         total_nodes: _,
                         values,
+                        additional_property: _,
                     } => {
                         *value = JsValue::alternatives(
                             take(values)
@@ -407,6 +410,7 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                 JsValue::Alternatives {
                     total_nodes: _,
                     values,
+                    additional_property: _,
                 } => {
                     *value = JsValue::alternatives(
                         take(values)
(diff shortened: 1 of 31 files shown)


ForsakenHarmony:
> Allows to handle `if (unknown && falsy)` `if (unknown || truthy)` `if (unknown ?? nullish)` in static analysis.

Which cases were handled before this PR?
Just `if (falsy)` `if (truthy)` `if (nullish)`?

Does this also handle `if (unknown && falsy && unknown)` etc.?

crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -350,6 +350,25 @@ enum JsValueMetaKind {
     Placeholder,
 }
 
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub enum AdditionalProperty {
```
ForsakenHarmony, line 354: Not a fan of this name, it's a bit too generic

sokra:
> Which cases were handled before this PR?
> Just `if (falsy)` `if (truthy)` `if (nullish)`?

The `unknown` was the problem. If was able to handle `if (truthy || unknown)` but not `if (unknown || truthy)`.
The problem is that `truthy || unknown => truthy`, but `unknown || truthy` could be `unknown` (if it's truthy at runtime) or `truthy`, so the evaluator evaluates it to `Alternatives(unknown, truthy)`. The PR adds an additional info to `Alternatives` that the combination has a truthy value.

crates/turbopack-ecmascript/src/analyzer/mod.rs
sokra: yeah, I changed it

sokra:
* **#8890** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8890?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8873** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8873?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @sokra and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -501,43 +503,100 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
...
                     }
                     None => {
                         // We don't know if this part is skipped or the final value, so we keep it.
+                        part_properties.push(property);
                         parts.push(part);
                         continue;
                     }
                 }
             }
+            part_properties.truncate(parts.len());
```
arlyon, line 540: under what condition does this happen? it seems from the code that part properties is always less than or equal to parts

sokra: never, removed it...

ForsakenHarmony (approved):

PR #8031, opened 2024-04-24
feat: Improve error messages from `mdxrs`

kdy1:
### Description

Improve error messages from mdxrs

Closes PACK-2589

### Testing Instructions

next.js counterpart: https://github.com/vercel/next.js/pull/64953

PR #8879, opened 2024-07-30
Convert MDX errors into issues

mischnic:
Trying https://github.com/vercel/turbo/pull/8031 again

~~Depends on https://github.com/vercel/turbo/pull/8766~~

Bildschirmfoto 2024-07-30 um 15 35 48


crates/turbopack-mdx/src/lib.rs
```diff
@@ -175,14 +179,55 @@ impl MdxTransformedAsset {
...
+                        markdown::message::Place::Position(p) => (
+                            SourcePos {
+                                line: p.start.line - 1,
+                                column: p.start.column - 1,
+                            },
+                            SourcePos {
+                                line: p.end.line - 1,
+                                column: p.end.column - 1,
+                            },
+                        ),
```
mischnic, line 202: I still need to figure out whether the `end` of Turbopack's ranges are inclusive or exclusive.

markdown's `end.column` is the first character *after* the range

mischnic: Turbopack's ranges are exclusive as well

sokra (approved):

PR #8883, opened 2024-07-30
feat(turborepo): add `get_merge_base` to turborepo_scm

NicholasLYang:
### Description

In preparation for `--affected`, adding a merge base function
### Testing Instructions

Adds a test for merge_base.

PR #8884, opened 2024-07-30
feat(turborepo): implement --affected flag

NicholasLYang:
### Description

Implements the `--affected` flag. Does some light refactoring to SCM to ensure that we're getting the unstaged changes in `--affected`.

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -697,6 +697,11 @@ pub struct ExecutionArgs {
     #[clap(short = 'F', long, group = "scope-filter-group")]
     pub filter: Vec<String>,
 
+    /// Run only tasks that are affected by changes between
+    /// the current branch and `main`
+    #[clap(long, group = "scope-filter-group")]
+    pub affected: bool,
+
     /// Set type of process output logging. Use "full" to show
     /// all output. Use "hash-only" to show only turbo-computed
     /// task hashes. Use "new-only" to show only new output with
--- a/crates/turborepo-lib/src/config.rs
+++ b/crates/turborepo-lib/src/config.rs
@@ -286,6 +286,14 @@ impl ConfigurationOptions {
         self.ui.unwrap_or(UIMode::Stream)
     }
 
+    pub fn scm_base(&self) -> &str {
+        self.scm_base.as_deref().unwrap_or("main")
+    }
+
+    pub fn scm_head(&self) -> &str {
+        self.scm_head.as_deref().unwrap_or("HEAD")
+    }
+
     pub fn allow_no_package_manager(&self) -> bool {
         self.allow_no_package_manager.unwrap_or_default()
     }
@@ -800,6 +808,12 @@ impl TurborepoConfigBuilder {
                     }
                     if let Some(env_mode) = current_source_config.env_mode {
                         acc.env_mode = Some(env_mode);
+                    }
+                    if let Some(scm_base) = current_source_config.scm_base {
+                        acc.scm_base = Some(scm_base);
+                    }
+                    if let Some(scm_head) = current_source_config.scm_head {
+                        acc.scm_head = Some(scm_head);
                     }
 
                     acc
--- a/crates/turborepo-lib/src/config.rs
+++ b/crates/turborepo-lib/src/config.rs
@@ -213,6 +213,8 @@ pub struct ConfigurationOptions {
     pub(crate) daemon: Option<bool>,
     #[serde(rename = "envMode")]
     pub(crate) env_mode: Option<EnvMode>,
+    pub(crate) scm_base: Option<String>,
+    pub(crate) scm_head: Option<String>,
 }
 
 #[derive(Default)]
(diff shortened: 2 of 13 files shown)


NicholasLYang:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/vercel/turbo/8884?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#8884** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8884?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#8883** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8883?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -695,6 +695,11 @@ pub struct ExecutionArgs {
     #[clap(short = 'F', long, group = "scope-filter-group")]
     pub filter: Vec<String>,
 
+    /// Run only tasks that are affected by changes between
+    /// the current branch and `main`
```
mehulkar, line 699: Is `main` configurable?

NicholasLYang: yes, with the `TURBO_SCM_BASE` env var

crates/turborepo-scm/src/git.rs
```diff
@@ -26,9 +32,23 @@ impl SCM {
...
-    ) -> Result<HashSet<AnchoredSystemPathBuf>, Error> {
+        include_uncommitted: bool,
+        allow_unknown_objects: bool,
+    ) -> Result<ChangedFiles, Error> {
         match self {
-            Self::Git(git) => git.changed_files(turbo_root, from_commit, to_commit),
+            Self::Git(git) => {
+                match git.changed_files(turbo_root, from_commit, to_commit, include_uncommitted) {
+                    Ok(files) => Ok(ChangedFiles::Some(files)),
+                    Err(e) if allow_unknown_objects => {
```
chris-olszewski, line 42: Can we only catch errors related to a commit not existing? I think ignoring all `git` errors might be a little heavy handed.

turborepo-tests/integration/tests/affected.t
```diff
@@ -0,0 +1,110 @@
+Setup
+  $ . ${TESTDIR}/../../helpers/setup_integration_test.sh
+
+Create a new branch
```
chris-olszewski, line 4: Can we add a test for when `main` isn't at the merge base of the branch i.e.
```
 o main
 |  o my-branch
 |  |
 | /
 o merge base
```

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -176,21 +177,34 @@ impl<'a, T: GitChangeDetector> FilterResolver<'a, T> {
...
-        let selectors = patterns
+        let mut selectors = patterns
             .iter()
             .map(|pattern| TargetSelector::from_str(pattern))
             .collect::<Result<Vec<_>, _>>()?;
 
+        if let Some((from_ref, to_ref)) = affected {
+            selectors.push(TargetSelector {
+                git_range: Some(GitRange {
+                    from_ref: from_ref.to_string(),
```
chris-olszewski, line 199: I don't think this is correct, we need to be comparing the base of the branch and not just `main`. I could be wrong, but I think these are distinct. Can we verify this is correct with a test where the branch base isn't the tip of `main`?

chris-olszewski (changes requested):
I think this is a very solid start, but I think it's missing the core part of `--affected` where we figure out what the base commit is.

I could be misunderstanding and this functionality is already here, but some tests displaying it are necessary.

crates/turborepo-lib/src/run/scope/filter.rs
NicholasLYang: Will add a test, but I believe this is correct, because we do `git diff from...to`, which is equivalent to `git diff $(git merge-base from to) to`

crates/turborepo-lib/src/config.rs
```diff
@@ -278,6 +280,14 @@ impl ConfigurationOptions {
         self.ui.unwrap_or(false) && atty::is(atty::Stream::Stdout)
     }
 
+    pub fn scm_base(&self) -> &str {
+        self.scm_base.as_deref().unwrap_or("main")
```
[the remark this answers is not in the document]
NicholasLYang: https://linear.app/vercel/issue/TURBO-3948/add-master-fallback-for-affected-base-branch

crates/turborepo-scm/src/git.rs
NicholasLYang: It's a little tricky because the error for missing commit doesn't seem to have a single output. Like in the case where you have a shallow clone, you get:

```
fatal: main...HEAD: no merge base
```
but if you have a ref that just doesn't exist, you might get:
```
fatal: ambiguous argument 'main...nicholasyang/turbo-info': unknown revision or path not in the working tree.
```

but for an invalid commit hash, you'll get:

```
fatal: Invalid symmetric difference expression main...108774c7f7d999860668dcececd89d7e172ce20d
```

I'm not exactly sure which error message a user may get

crates/turborepo-lib/src/cli/mod.rs
mehulkar: Could it be an optional argument to `--affected=<base>` also? Not a huge deal, but it would be similar to `--filter` (with much lower surface area)

chris-olszewski (approved):

PR #8892, opened 2024-07-31
fix(pty): disable echoctl for child processes

chris-olszewski:
### Description

Copying and pasting from #7109:

Disable ECHOCTL for the pseudo terminal that we start:

    ECHOCTL
    greater than the special character. For example,
    character 0x08 (BS) is echoed as ^H.

This will remove the ^D that gets generated from the EOT when using a PTY. Using PTY still does generate a leading \n as EOT is only respected if it immediately follows a newline. portable_pty tries to prevent failed EOT by emitting a \nEOT on drop in case end users didn't end the last write with a newline.

### Testing Instructions

Verify that `^D` no longer appears in output:
```
[0 olszewski@chriss-mbp] /tmp/echoctl-test $ TURBO_UI=0 turbo_dev --skip-infer build --force                                                                                                            
turbo 2.0.10                                                                                                                                                                                            
                                                                                                                                                                                                        
• Packages in scope: @repo/eslint-config, @repo/typescript-config, @repo/ui, docs, web                                                                                                                  
• Running build in 5 packages
• Remote caching disabled
docs:build: cache bypass, force executing 9240c8b370dcb1ac
docs:build: 
web:build: cache bypass, force executing 253f5ed6d7afcf7f
web:build: 
web:build: 
web:build: > web@0.1.0 build /private/tmp/echoctl-test/apps/web
web:build: > next build
web:build: 
docs:build: 
docs:build: > docs@0.1.0 build /private/tmp/echoctl-test/apps/docs
docs:build: > next build
docs:build: 
web:build:   ▲ Next.js 15.0.0-rc.0
docs:build:   ▲ Next.js 15.0.0-rc.0
```

### DIFF
--- a/crates/turborepo-lib/src/process/child.rs
+++ b/crates/turborepo-lib/src/process/child.rs
@@ -161,6 +161,26 @@ impl ChildHandle {
         let controller = pair.master;
         let receiver = pair.slave;
 
+        #[cfg(unix)]
+        {
+            use nix::sys::termios;
+            if let Some((file_desc, mut termios)) = controller
+                .as_raw_fd()
+                .and_then(|fd| Some(fd).zip(termios::tcgetattr(fd).ok()))
+            {
+                // We unset ECHOCTL to disable rendering of the closing of stdin
+                // as ^D
+                termios.local_flags &= !nix::sys::termios::LocalFlags::ECHOCTL;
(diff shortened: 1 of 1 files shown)


tknickman (approved):

tknickman:
Nice!

PR #8893, opened 2024-07-31
fix(cli): prevent splitting of run args

chris-olszewski:
### Description

We had a bug where we allowed providing run flags on either side of the `run` subcommand e.g. `turbo --filter=web run build`. Any arguments provided before `run` would be silently ignored in this case.

**Background**

We support omitting the `run` subcommand so we duplicate all of the run subcommand arguments on the main `turbo` command. This allowed for users to specify run commands both on the main `turbo` command and the `run` subcommand.

### Testing Instructions

Added unit tests for detection and an integration test to show the error message.


chris-olszewski:
Moving back to draft as flags set via environment variables end up in both sets of execution args. Need a different approach to checking if both are used.

PR #8902, opened 2024-08-01
Support Buffer results from Webpack loaders

mischnic:
The result from a webpack loader can be a string or a buffer: https://webpack.js.org/api/loaders/#thiscallback

The mdx loader uses this:
https://github.com/mdx-js/mdx/blob/e780a972c29584851c2a131f2984a7470aa8c2f2/packages/loader/lib/index.js#L93

Stringify the Buffer before sending it over to Rust


timneutkens (approved):
Looks good to me!

crates/turbopack-node/js/src/transforms/webpack-loaders.ts
```diff
@@ -479,7 +479,7 @@ const transform = (
         if (!result.result) return reject(new Error("No result from loaders"));
         const [source, map] = result.result;
         resolve({
-          source,
+          source: Buffer.isBuffer(source) ? source.toString("utf8") : source,
```
sokra, line 482: The result might be some binary data that is not an utf-8 string.

E. g. some image-processing-loader can return an image.

mischnic:
Added to https://github.com/vercel/next.js/pull/68397

PR #8904, opened 2024-08-01
print a warning when the trace file is missing rather than wait

arlyon:
### Description

The trace viewer just sits and waits if it can't file the trace file. We instead print a little message in this case to help avoid cases where (in my case) you are missing a character from the end of the file name 😂 

### Testing Instructions

Attempt to load a trace with a path that doesn't exist. Observe error message.


arlyon:
* **#8904** <a href="https://app.graphite.dev/github/pr/vercel/turbo/8904?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #8910, opened 2024-08-01
chore: improve -vv logging

mehulkar:


crates/turborepo-telemetry/src/config.rs
```diff
@@ -67,7 +67,7 @@ impl TelemetryConfig {
     }
 
     pub fn new(config_path: AbsoluteSystemPathBuf) -> Result<TelemetryConfig, ConfigError> {
-        debug!("Telemetry config path: {}", config_path);
+        trace!("Telemetry config path: {}", config_path);
```
mehulkar, line 70: This is being called a billion times becuase `.with_default_config_path()` is called a lot. Generally think when something is called a lot, it should be trace logging. Aggregations are more useful for debug logging

crates/turborepo-repository/src/package_json.rs
```diff
@@ -159,7 +159,7 @@ impl From<RawPnpmConfig> for PnpmConfig {
 
 impl PackageJson {
     pub fn load(path: &AbsoluteSystemPath) -> Result<PackageJson, Error> {
-        tracing::debug!("loading package.json from {}", path);
+        tracing::trace!("loading package.json from {}", path);
```
mehulkar, line 162: This is call N times in a monorepo with N packages, making `-vv` logging very noisy. I think it's better to put this in `-vvv` (trace) logging

chris-olszewski (approved):

PR #8919, opened 2024-08-01
refactor(turborepo): change the ui config to use an enum

NicholasLYang:
### Description

Right now we internally represent the `ui` config as a boolean. This isn't very extensible, so I refactored it to be an enum instead

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -8,7 +8,7 @@ use clap::{
 };
 use clap_complete::{generate, Shell};
 pub use error::Error;
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 use tracing::{debug, error, log::warn};
 use turbopath::AbsoluteSystemPathBuf;
 use turborepo_api_client::AnonAPIClient;
@@ -29,6 +29,7 @@ use crate::{
     run::watch::WatchClient,
     shim::TurboState,
     tracing::TurboSubscriber,
+    turbo_json::UIMode,
 };
 
 mod error;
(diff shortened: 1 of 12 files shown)


crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -1080,7 +1058,7 @@ mod tests {
     #[test_case(r#"{ "ui": "tui" }"#, Some(UI::Tui) ; "tui")]
     #[test_case(r#"{ "ui": "stream" }"#, Some(UI::Stream) ; "stream")]
```
chris-olszewski, line 1059:
```suggestion
    #[test_case(r#"{ "ui": "tui" }"#, Some(UIMode::Tui) ; "tui")]
    #[test_case(r#"{ "ui": "stream" }"#, Some(UIMode::Stream) ; "stream")]
```

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -2560,3 +2559,24 @@ mod test {
         );
     }
 }
+
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, Deserializable, PartialEq, Eq, ValueEnum)]
+#[serde(rename_all = "camelCase")]
+pub enum UIMode {
```
chris-olszewski, line 2565: Nit: Is there a reason for hoisting this to the CLI module? It feels like it should live in the `turbo_json` or `config` layer. Not that we're in a place to split up `turborepo-lib`, but if we did I think `cli` would still be an "entry" module that depends on `config`/`turbo_json` modules.

Also if we're moving the implementation, we should move the unit tests for ser/de with it.

chris-olszewski (approved):
Aside from hoisting the enum to `cli` LGTM.

PR #8906, opened 2024-08-01
chore: remove turbopack crates

ForsakenHarmony:
### Why?

[vercel.com/blog/turbopack-moving-homes](https://vercel.com/blog/turbopack-moving-homes)

PR #8923, opened 2024-08-01
removes prettierignore files related to turbopack

dimitropoulos:
with turbopack being removed in #8906, we can also remove mentions of turbopack crates in the prettierignore

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -22,7 +22,5 @@ __generated__/
 # crates
 crates/*/js/src/compiled
 crates/next-transform-strip-page-exports/tests
-crates/next-transform-dynamic/tes
-
 # generators
 *.hbs
--- a/.prettierignore
+++ b/.prettierignore
@@ -21,14 +21,8 @@ __generated__/
 
 # crates
 crates/*/js/src/compiled
-crates/turbopack/bench.json
-crates/turbopack/tests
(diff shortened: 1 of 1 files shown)


.prettierignore
```diff
@@ -21,14 +21,8 @@ __generated__/
...
 crates/*/js/src/compiled
-crates/turbopack/bench.json
-crates/turbopack/tests
-crates/turbopack-ecmascript/tests/analyzer/graph
-crates/turbopack-ecmascript/tests/tree-shaker
 crates/next-transform-strip-page-exports/tests
-crates/next-transform-dynamic/tests
-crates/turbopack-tests/tests/execution/turbopack/basic/error/input/broken.js
-crates/turbopack-tests/tests/execution/**/output*
+crates/next-transform-dynamic/tes
```
tknickman, line 25: We should be able to remove this one as well

dimitropoulos: thanks!! I dug a bit deeper and actually all of them under `# crates` seem to have been turbopack specific.  e.g. (from a checkout prior to the turbopack removal):

Screenshot_20240801_170149

see: f32cf13524f7697d34f6800057794fc323ac2918

tknickman: Makes sense, we can just nuke them all

arlyon (approved):

tknickman (approved):

PR #8944, opened 2024-08-05
feat(telemetry): reduce data volume

tknickman:
### Description

Reduce volume of telemetry logs

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -1301,11 +1301,6 @@ pub async fn run(
         }
     };
 
-    if cli_result.is_err() {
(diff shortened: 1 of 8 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -1301,11 +1301,6 @@ pub async fn run(
         }
     };
 
-    if cli_result.is_err() {
-        root_telemetry.track_failure();
```
tknickman, line 1305: This wasn't implemented correctly and we never got a failure - so removing it.

tknickman (commented):
Happy to debate any of these!

chris-olszewski (approved):

PR #8947, opened 2024-08-05
cacheDir for env, config, and flag

dimitropoulos:
### Description

This PR ensures support for configuring the `cacheDir` via the `TURBO_CACHE_DIR` environment variable, `"cacheDir"` `turbo.json` config, or `--cache-dir` CLI flag.

### Testing Instructions

```
No cacheDir by default
  $ ${TURBO} config | jq .cacheDir
  ".turbo/cache"

Add env var: `TURBO_CACHE_DIR=nebulo9`
  $ TURBO_CACHE_DIR=nebulo9 ${TURBO} config | jq .cacheDir
  "nebulo9"

Add flag: `--cache-dir nebulo9`
  $ ${TURBO} --cache-dir nebulo9 config | jq .cacheDir
  "nebulo9"
  
...or add `"cacheDir": "nebulo9"` to the turbo.json
  $ ${TURBO} config | jq .cacheDir
  "nebulo9"
```

## Absolute Paths

Absolute paths are supported for the flag and env var, but not the config.  If you add one to the config you get a nice fancy error like this:

Screenshot_20240809_104940


crates/turborepo-lib/src/config.rs
```diff
@@ -296,6 +299,16 @@ impl ConfigurationOptions {
...
     }
+
+    pub fn cache_dir(&self) -> &Utf8Path {
+        self.cache_dir.as_deref().unwrap_or_else(|| {
+            Utf8Path::new(if cfg!(windows) {
+                ".turbo\\cache"
+            } else {
+                ".turbo/cache"
+            })
+        })
```
NicholasLYang, line 310: nit: this is slightly more idiomatic
```suggestion
        self.cache_dir.as_deref().unwrap_or_else(|| {
            Utf8Path::from_iter([".turbo", "cache"])
        })
```

NicholasLYang (approved):
looks great!

crates/turborepo-lib/src/config.rs
dimitropoulos: love this kind of feedback! thanks!

@chris-olszewski and I were looking at that and the thinking was that `from_iter` causes an allocation and this doesn't but looking at it again, this does look better.  we were already sorta 50/50 on it anyway :)

dimitropoulos: update: there's no `Utf8Path::from_iter` [from what I can tell](https://docs.rs/camino/latest/camino/struct.Utf8Path.html).

what I was talking about was `join_components`.  would you like to switch to that?

crates/turborepo-paths/src/relative_unix_path.rs
```diff
@@ -4,10 +4,11 @@ use std::{
 };
 
 use camino::{Utf8Path, Utf8PathBuf};
+use serde::Serialize;
 
 use crate::{AnchoredSystemPathBuf, PathError, RelativeUnixPathBuf};
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize)]
```
chris-olszewski, line 11: This implementation is no longer necessary so we should drop it.

chris-olszewski (approved):
Only thing that should be fixed before merging is removing the changes to `RelativeUnixPath`

crates/turborepo-paths/src/relative_unix_path.rs
dimitropoulos: totally: fixed in e5ae321c90af89d648d465713ff6ee2c52fc2534

dimitropoulos:
@chris-olszewski ahh it's `Utf8PathBuf` that has `from_iter` (not `Utf8Path`).  gotcha gotcha.  if there's no preference I'll leave it like it is. I want to get on to docs and updating the JSON Schema asap and those are waiting on this.

PR #8962, opened 2024-08-07
refactor: renamed UI to ColorConfig

NicholasLYang:
### Description

Since we now have an explicit `ui` config that refers to whether we display runs via stream, tui, or web, we should rename the existing `UI` struct to be more explicitly about color configuration (since that's the only thing it does).

### DIFF
--- a/crates/turborepo-auth/src/auth/logout.rs
+++ b/crates/turborepo-auth/src/auth/logout.rs
@@ -15,7 +15,7 @@ pub async fn logout<T: TokenClient>(options: &LogoutOptions<T>) -> Result<(), Er
         return Err(err);
     }
 
-    cprintln!(options.ui, GREY, ">>> Logged out");
+    cprintln!(options.color_config, GREY, ">>> Logged out");
     Ok(())
 }
 
@@ -171,7 +171,7 @@ mod tests {
             .expect("could not create file");
 
         let logout_options = LogoutOptions {
-            ui: ColorConfig::new(false),
+            color_config: ColorConfig::new(false),
             api_client: MockApiClient {
                 succeed_delete_request: true,
             },
@@ -199,7 +199,7 @@ mod tests {
         };
 
         let options = LogoutOptions {
-            ui: ColorConfig::new(false),
+            color_config: ColorConfig::new(false),
             api_client,
             path: Some(path.clone()),
             invalidate: true,
--- a/crates/turborepo-auth/src/auth/logout.rs
+++ b/crates/turborepo-auth/src/auth/logout.rs
@@ -86,7 +86,7 @@ mod tests {
     use tempfile::tempdir;
     use turbopath::AbsoluteSystemPathBuf;
     use turborepo_api_client::Client;
-    use turborepo_ui::UI;
+    use turborepo_ui::ColorConfig;
     use turborepo_vercel_api::{
         token::ResponseTokenMetadata, SpacesResponse, Team, TeamsResponse, UserResponse,
         VerifiedSsoUser,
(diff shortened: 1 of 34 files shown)


crates/turborepo-auth/src/auth/logout.rs
```diff
@@ -171,7 +171,7 @@ mod tests {
             .expect("could not create file");
 
         let logout_options = LogoutOptions {
-            ui: UI::new(false),
+            ui: ColorConfig::new(false),
```
dimitropoulos, line 174: was it intentional to also keep this named `ui`?

crates/turborepo-auth/src/auth/mod.rs
```diff
@@ -46,7 +46,7 @@ impl<'a, T: Client + TokenClient + CacheClient> LoginOptions<'a, T> {
 
 /// Options for logging out.
 pub struct LogoutOptions<T> {
-    pub ui: UI,
+    pub ui: ColorConfig,
```
dimitropoulos, line 49: was it intentional to also keep this named `ui`?  is this exposed to users?

crates/turborepo-auth/src/auth/mod.rs
```diff
@@ -27,13 +27,13 @@ pub struct LoginOptions<'a, T: Client + TokenClient + CacheClient> {
 }
 impl<'a, T: Client + TokenClient + CacheClient> LoginOptions<'a, T> {
     pub fn new(
-        ui: &'a UI,
+        ui: &'a ColorConfig,
```
dimitropoulos, line 30:
```suggestion
        color_config: &'a ColorConfig,
```

crates/turborepo-auth/src/ui/messages.rs
```diff
@@ -1,6 +1,6 @@
-use turborepo_ui::{BOLD, CYAN, UI};
+use turborepo_ui::{ColorConfig, BOLD, CYAN};
 
-pub fn print_cli_authorized(user: &str, ui: &UI) {
+pub fn print_cli_authorized(user: &str, ui: &ColorConfig) {
```
dimitropoulos, line 3:
```suggestion
pub fn print_cli_authorized(user: &str, color_config: &ColorConfig) {
```

crates/turborepo-lib/src/commands/ls.rs
```diff
@@ -25,13 +25,13 @@ pub enum Error {
...
 struct RepositoryDetails<'a> {
-    ui: UI,
+    ui: ColorConfig,
     package_manager: &'a PackageManager,
     packages: Vec<(&'a PackageName, &'a AnchoredSystemPath)>,
 }
 
 struct PackageDetails<'a> {
-    ui: UI,
+    ui: ColorConfig,
```
dimitropoulos, line 34: just verifying that both of these were intended not to be renamed

crates/turborepo-lib/src/commands/mod.rs
```diff
@@ -43,11 +43,11 @@ impl CommandBase {
         args: Args,
         repo_root: AbsoluteSystemPathBuf,
         version: &'static str,
-        ui: UI,
+        ui: ColorConfig,
```
dimitropoulos, line 46:
```suggestion
        color_config: ColorConfig,
```

crates/turborepo-lib/src/run/builder.rs
```diff
@@ -54,7 +54,7 @@ pub struct RunBuilder {
     opts: Opts,
     api_auth: Option<APIAuth>,
     repo_root: AbsoluteSystemPathBuf,
-    ui: UI,
+    ui: ColorConfig,
```
dimitropoulos, line 57: just verifying this wasn't intended to be renamed

dimitropoulos (approved):
didn't see anything major - just a few double-checks and one or two chances to clean up constructors that don't need argument renaming

chris-olszewski (approved):

PR #8964, opened 2024-08-08
Docs: Improve Turborepo API references docs

fuma-nama:
### Description

~~Made the headings for each option to be meaningful. (e.g. `# ui` -> `# Customise Terminal UI`).~~

Improved the description for environment variable options, making it easier to understand for beginners.

Move the section of `extends` option to "Project Configurations" instead of "Global Options".

### Testing Instructions

Preview the docs.


anthonyshew:
We have a strategy that we use with our docs at Vercel where reference pages use the command as a heading. This is for a couple of reasons:
1. We write tutorial and guides pages with use-case-based headings for folks who are trying to accomplish something specific. You can see how this works in the [Crafting your repository](https://turbo.build/repo/docs/crafting-your-repository pages.
2. In reference material, we expect folks to be looking for specific information about a specific part of the API. We make the headings reflect those options so that it's clear the information in that section will be dry and not use-case-based (in contrast to the first point). This is a pattern you'll also see in other docs, like in [the Astro reference](https://docs.astro.build/en/reference/api-reference/) or [Vercel CLI command pages](https://vercel.com/docs/cli/alias).

When you put those together, you have areas of the docs that work great for the different approaches that you may be taking when you open up the docs.
- Are you looking to get information about solving a problem? Go to a tutorial, guide, or how-to page.
- Are you looking for information about a specific part of the API? Go to the API reference.

With that in mind, I'd prefer not to change up our approach to the reference pages. If you'd like to omit that change, I'm happy to review the rest of the PR to see what other improvements we can make!

fuma-nama:
Oh yes that makes sense, I'll revert these changes

fuma-nama:
I also noticed the [globalPassThroughEnv](https://turbo.build/repo/docs/reference/configuration#globalpassthroughenv) is so similar to `globalEnv` option, the glob syntax works on both of them.

I've grouped the pass through variables section under the `env` section, it should help clarify some differences between env and pass through env, and let people knows the glob syntax also works on pass through env.

PR #8965, opened 2024-08-08
chore(repo): update readme and links

tknickman:
### Description

Update the README, and update repository links


dimitropoulos:
`crates/turborepo-lib/src/daemon/proto/turbod.proto` has the old path.  I think we should update it to prevent downstream dependencies having build errors (or `go get` failing)

packages/turbo-utils/__tests__/examples.test.ts
```diff
@@ -39,7 +39,7 @@ describe("examples", () => {
   describe("getRepoInfo", () => {
```
dimitropoulos, line 39: not sure if we care, but 5 lines above this there's a /vercel/turbo (in a non-github.com test)

dimitropoulos:
it still works ok after the rename but we may want to also update `packages/turbo-utils/src/examples.ts` in `downloadAndExtractExample`

tknickman:
Done done, good catches - thanks!

dimitropoulos:
the `CONTRIBUTING.md` has two references to the old path, but since I'm about to overhaul that anyway it's probably not worth it (up to you! :smile: )

dimitropoulos (approved):

tknickman:
Hmm I don't see those in Contributing - but yea feel free to fix those up in your PR!

PR #8971, opened 2024-08-08
Log changed package names instead of full struct

mehulkar:


crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -138,9 +138,14 @@ impl<'a> GitChangeDetector for ScopeChangeDetector<'a> {
...
             PackageChanges::Some(packages) => {
                 debug!(
-                    "determined {} packages changed: {packages:?}",
-                    packages.len()
+                    "{} packages changed: {:?}",
+                    packages.len(),
+                    &packages
+                        .iter()
+                        .map(|x| x.name.to_string())
+                        .collect::<Vec<String>>()
```
mehulkar, line 146: sort also? or is packages already sorted?

chris-olszewski: Nope, sorting is good here. Packages are stored in a hash set so just iterating over them won't yield a stable order.

chris-olszewski (approved):

Issue #8972, opened 2024-08-09
Docs: Registering Root Tasks - Incorrect/Unclear example


nickfujita:
### What is the improvement or update you wish to see?

The current documentation seems to suggest that by creating the `lint:root` script in the root `package.json` file, and the `//#lint:root: {}` task in `turbo.json` alongside `lint: {}` task, running `turbo run lint` should execute the `lint` script in all packages in the repo, along with the `lint:root` script in the root `package.json`. At least that's how I understood it to mean.

### Is there any context that might help us understand?

[Registering Root Tasks](https://turbo.build/repo/docs/crafting-your-repository/configuring-tasks#registering-root-tasks)
## Existing
Running `pnpm lint` only runs `lint` scripts in packages, but not the root level `lint:root` script.
#### turbo.json
```json
{
  "tasks": {
    "lint": {
      "dependsOn": ["^lint"]
    },
    "//#lint:root": {} 
  }
}
```

#### package.json
```
{
  "scripts": {
    "lint": "turbo run lint",
    "lint:root": "eslint ."
  }
}
```

## Correction - Option 1
Runs `lint:root` in addition to `lint` scripts in all packages, by separately calling `lint:root` in the `turbo run` statement
#### turbo.json
```json
{
  "tasks": {
    "lint": {
      "dependsOn": ["^lint"]
    },
    "//#lint:root": {} 
  }
}
```

#### package.json
```
{
  "scripts": {
    "lint": "turbo run lint lint:root",
    "lint:root": "eslint ."
  }
}
```

## Correction - Option 2
Runs `lint:root` in addition to `lint` scripts in all packages by removing `:root` suffix from the script in the root `package.json`, and making the script the calls `turbo run lint` be named something else (eg. lint-abcdef, lint-all)
#### turbo.json
```json
{
  "tasks": {
    "lint": {
      "dependsOn": ["^lint"]
    },
    "//#lint": {} 
  }
}
```

#### package.json
```
{
  "scripts": {
    "lint-all": "turbo run lint",
    "lint": "eslint ."
  }
}
```

### Does the docs page already exist? Please link to it.

https://turbo.build/repo/docs/crafting-your-repository/configuring-tasks#registering-root-tasks

anthonyshew:
Fixed in https://github.com/vercel/turborepo/pull/9098. Thank you!

PR #8973, opened 2024-08-09
[FIX] Docs: Registering Root Tasks - Incorrect/Unclear example

nickfujita:
### Description

Updates example for adding a root task, and calling it along with the other package level tasks.

### Testing Instructions

Review options listed in issue https://github.com/vercel/turbo/issues/8972, and confirm this is the correct implementation of this example.

### DIFF
--- a/docs/repo-docs/crafting-your-repository/configuring-tasks.mdx
+++ b/docs/repo-docs/crafting-your-repository/configuring-tasks.mdx
@@ -280,7 +280,7 @@ You can also run scripts in the `package.json` in the Workspace root using `turb
 ```json title="./package.json"
 {
   "scripts": {
-    "lint": "turbo run lint",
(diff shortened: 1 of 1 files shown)


anthonyshew:
This is a great observation. I'm thinking we should write both of the options you've described into the docs for thoroughness. Would you be up to adding an example of both?

Warning: It's highly likely that I'll be obnoxiously nitpicky about this, because I really like to make sure that we use succinct, excellent language when we're presenting folks with options. If you're ready for me to be mildly annoying about the precise way this gets written, we can keep going on this PR. Otherwise, I can take this on myself in a separate PR so I can be as surgical as I'd like to be without annoying you. 😄

nickfujita:
Thanks @anthonyshew 
Closing in favor of https://github.com/vercel/turborepo/pull/9098

PR #8977, opened 2024-08-09
feat: implement a query interface for turborepo

NicholasLYang:
### Description

Adds a GraphQL query interface to turborepo via the `turbo query` command. If you pass it a query string, then it directly runs the query. If you pass it a file, it reads the file and runs the query. If you don't pass a query, then it opens up a server with GraphiQL (we can ITG this interface).

### Testing Instructions
Added tests to `affected.t` and `command-query.t`

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -588,7 +588,11 @@ pub enum Command {
         #[clap(flatten)]
         execution_args: Box<ExecutionArgs>,
     },
+    /// Query your monorepo using GraphQL. If no query is provided, spins up a
+    /// GraphQL server with GraphiQL.
+    #[clap(hide = true)]
     Query {
+        /// The query to run, either a file path or a query string
         query: Option<String>,
     },
     Watch(Box<ExecutionArgs>),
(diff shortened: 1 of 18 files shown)


crates/turborepo-lib/src/query.rs
```diff
@@ -0,0 +1,411 @@
...
+                name: name.clone(),
+            })
+            .filter(|pkg| filter.check(pkg))
+            .sorted_by(|a, b| a.name.cmp(&b.name))
+            .collect())
+    }
+}
+
+#[Object]
+impl Package {
```
dimitropoulos, line 304: would it make sense to add `hash` here?  seems like something we'd want to be able to see/show

NicholasLYang: Packages don't really have a concept of a hash, tasks do. But yeah, eventually there should be task hashes. That will require some reworking of our infrastructure since right now we do things in a direct pipeline fashion, so computing one hash will require computing them all

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -1277,6 +1284,16 @@ pub async fn run(
...
+        Command::Query { query } => {
+            let query = query.clone();
+            let event = CommandEventBuilder::new("query").with_parent(&root_telemetry);
+            event.track_call();
+            let base = CommandBase::new(cli_args, repo_root, version, color_config);
+
+            let query = query::run(base, event, query).await?;
+
+            Ok(query)
+        }
```
tknickman, line 1296: Let's add a warn here that this is under development / experimental (wording up to you) just incase anyone finds it in code and wants to play with it early (even though the top level command is hidden for now)

tknickman (approved):
One suggestion but otherwise let's ITG it

PR #8981, opened 2024-08-09
removes GOROOT from globalPassThroughEnv for local dev

dimitropoulos:
My understanding is that this is not needed anymore - so just cleaning it up on the basis of that assumption.

Reminder: there is some Go code left in `turborepo-lib/hash`, but from what I can tell that's entirely separate from this.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "ui": "tui",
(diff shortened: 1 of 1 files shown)


tknickman (approved):

tknickman:
This will cause cache misses for anyone who had this set after updating in the next release - but should impact should be small and it's nice to clean this up!

chris-olszewski (approved):

PR #8982, opened 2024-08-09
simplify Prysk commands

dimitropoulos:
### Description

I run the integration tests frequently locally while developing (often targeting one single integration test file) and I noticed there's a lot of stuff in the commands that doesn't need to be there.  Cleaning it up makes it simpler to see how to interact with the prysk tool.

### Testing Instructions

From the integration tests you can run `test` and `test:interactive` as normal.  I've also added a root command that you can use like this `pnpm test:integration tests/config.t` which will only run the `config.t` test.


turborepo-tests/integration/tests/lockfile-aware-caching/berry.t
```diff
@@ -1,6 +1,7 @@
 Setup
   $ . ${TESTDIR}/../../../helpers/setup.sh
   $ . ${TESTDIR}/setup.sh $(pwd) berry
+  $ export COREPACK_ENABLE_DOWNLOAD_PROMPT=0
```
chris-olszewski, line 4: I think it would make sense to put this in the `setup.sh` script instead of adding it adhoc to each test that might trigger `corepack`.

chris-olszewski (approved):

turborepo-tests/integration/tests/lockfile-aware-caching/berry.t
dimitropoulos: sure! added in 6433e7b7a02f85f84084a93bc14cb1692dc02d5c

PR #8993, opened 2024-08-12
fix(create-turbo): support renamed repositories

tknickman:
### Description

This bug broke create-turbo when we renamed the repo. This bug is explained in the comments but the tl;dr is that downloading the repo works because the download URL automatically redirects from the old name to the new name. However the download is always put into a directory of the name {repoName}-{branch} where the repoName is the name of the repo at the time and NOT always the name used to download it.


chris-olszewski (approved):

tknickman:
Fixed for create-next-app here as well:
https://github.com/vercel/next.js/pull/68802

PR #8996, opened 2024-08-12
feat(tui): resize terminal pane

chris-olszewski:
### Description

Resize the layout including the pane on terminal resize events.

We completely sidestep `vt100`'s inability to properly handle resize events by storing a copy of task outputs and recreating the virtual terminal from scratch on a resize. A start at adding this functionality to vt100 can be found in [this commit](https://github.com/vercel/turbo/commit/6def8c819ab4d84a75afa068ace91bbd3e8eb9c9).

Generally speaking, this is inefficient, but there is a clear path forward if performance here becomes an issue.

Each commit can be reviewed individually.

### Testing Instructions

Added a few basic unit tests. Manual testing works best to verify terminal looks like how we want it to look.

https://github.com/user-attachments/assets/661495d1-1a4a-4aa8-91de-cc0c2face05d

### DIFF
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -413,7 +413,7 @@ impl<W: Write> App<W> {
     #[tracing::instrument(skip(self, output))]
     pub fn process_output(&mut self, task: &str, output: &[u8]) -> Result<(), Error> {
         let task_output = self.tasks.get_mut(task).unwrap();
-        task_output.parser.process(output);
+        task_output.process(output);
         Ok(())
     }
 }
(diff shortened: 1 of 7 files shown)


crates/turborepo-ui/src/tui/app.rs
```diff
@@ -444,6 +448,10 @@ fn run_app_inner<B: Backend + std::io::Write>(
     let mut last_render = Instant::now();
     let mut callback = None;
     while let Some(event) = poll(app.input_options(), &receiver, last_render + FRAMERATE) {
+        // If we got a resize event, make sure to update ratatui backend.
+        if matches!(event, Event::Resize { .. }) {
+            terminal.autoresize()?;
+        }
```
tknickman, line 454: Does this need to get debounced at all? Since we're redrawing we probably want to wait until things stop changing for a few ms

tknickman (approved):
No internet? No problem 😄

PR #8997, opened 2024-08-12
feat(turbo): warn on empty cache

tknickman:
### Description

Warn when you have specified outputs that do not exist. 

You can optionally pass `--skip-empty-cache` to skip caching entirely if the specified outputs do not exist. This should probably be the default, but it would be breaking behavior to enable for now. 

Co-authored-by: Mehul Kar <mehul.kar@vercel.com>

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -815,6 +815,12 @@ pub struct RunArgs {
     #[clap(long)]
     pub no_cache: bool,
 
+    /// If outputs have been specified, but none exist - skip saving the (empty)
+    /// cache for the task. Enabling this can be a useful safety measure to
+    /// ensure proper outputs configuration.
+    #[clap(long, value_name = "BOOL", action = ArgAction::Set, default_value = "false", default_missing_value = "true", num_args = 0..=1)]
+    pub skip_empty_cache: bool,
+
     // clap does not have negation flags such as --daemon and --no-daemon
     // so we need to use a group to enforce that only one of them is set.
     // we set the long name as [no-]daemon with an alias of daemon such
@@ -860,6 +866,7 @@ impl Default for RunArgs {
             dry_run: None,
             graph: None,
             no_cache: false,
+            skip_empty_cache: false,
             daemon: false,
             no_daemon: false,
             profile: None,
--- a/crates/turborepo-lib/src/run/cache.rs
+++ b/crates/turborepo-lib/src/run/cache.rs
@@ -9,7 +9,9 @@ use tracing::{debug, error};
 use turbopath::{
     AbsoluteSystemPath, AbsoluteSystemPathBuf, AnchoredSystemPath, AnchoredSystemPathBuf,
 };
-use turborepo_cache::{http::UploadMap, AsyncCache, CacheError, CacheHitMetadata, CacheSource};
+use turborepo_cache::{
+    http::UploadMap, AsyncCache, CacheError, CacheHitMetadata, CacheSource, OutputSaveResult,
+};
 use turborepo_repository::package_graph::PackageInfo;
 use turborepo_scm::SCM;
 use turborepo_telemetry::events::{task::PackageTaskEventBuilder, TrackedErrors};
@@ -49,6 +51,7 @@ pub struct RunCache {
     cache: AsyncCache,
     reads_disabled: bool,
     writes_disabled: bool,
+    skip_empty_cache: bool,
     repo_root: AbsoluteSystemPathBuf,
     color_selector: ColorSelector,
     daemon_client: Option<DaemonClient<DaemonConnector>>,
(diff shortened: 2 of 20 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -815,6 +815,12 @@ pub struct RunArgs {
     #[clap(long)]
     pub no_cache: bool,
 
+    /// If outputs have been specified, but none exist - skip saving the (empty)
+    /// cache for the task. Enabling this can be a useful safety measure to
+    /// ensure proper outputs configuration.
+    #[clap(long, value_name = "BOOL", action = ArgAction::Set, default_value = "false", default_missing_value = "true", num_args = 0..=1)]
+    pub skip_empty_cache: bool,
```
tknickman, line 822: Not sold on adding this as an arg (or this name if we do), but the behavior is nice to have - Enabling this fixes the case when you set outputs incorrectly, and nothing ends up being cached. This can be problematic because if you don't know this happened, and you don't make changes you can get a cache hit that also nothing, making it look like everything worked.

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -1001,15 +1002,33 @@ impl ExecContext {
...
-                        );
+                    let result = self.task_cache.save_outputs(task_duration, telemetry).await;
+                    match result {
+                        Ok(status) => {
+                            match status {
+                                OutputSaveResult::SavedEmpty => {
+                                    prefixed_ui.warn(format!(
+                                        "warning: no files were found that match the configured \
+                                         outputs - make sure \"outputs\" are correctly defined in \
+                                         your `turbo.json` for {}",
```
chris-olszewski, line 1013: Do we want to write these into the cached task logs so when restoring this hit from cache you'll see this warning?

crates/turborepo-lib/src/run/cache.rs
```diff
@@ -360,6 +364,21 @@ impl TaskCache {
...
+        // Compare to 1 because the log file is always cached.
+        if files_to_be_cached.len() == 1 {
+            // Were we expecting to cache more than just the log file?
+            if validated_inclusions.len() > 1 {
+                status = OutputSaveResult::SavedEmpty;
+                if self.run_cache.skip_empty_cache {
+                    return Ok(OutputSaveResult::SkippedEmpty);
+                }
+            } else {
+                debug!("caching log output only for {:?}", self.task_id);
```
chris-olszewski, line 378: Super minor, but I think this `debug` log is made unnecessary by the `debug` log earlier.

crates/turborepo-lib/src/run/cache.rs
```diff
@@ -360,6 +364,21 @@ impl TaskCache {
             globwalk::WalkType::All,
         )?;
 
+        debug!("files to cache: {:?}", files_to_be_cached.len());
```
chris-olszewski, line 367: Maybe add in the task id here
```suggestion
        debug!("{}: files to cache: {:?}", self.task_id, files_to_be_cached.len());
```

crates/turborepo-lib/src/cli/mod.rs
chris-olszewski: I'm not super for just adding the flag. This is a setting that is should generally be set at the repo level. e.g. If CI has `--skip-empty-cache` a user can poison the cache with a bad run if they don't have it set.

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -1001,15 +1002,33 @@ impl ExecContext {
...
-                        self.hash_tracker.insert_expanded_outputs(
-                            self.task_id.clone(),
-                            self.task_cache.expanded_outputs().to_vec(),
-                        );
+                    let result = self.task_cache.save_outputs(task_duration, telemetry).await;
+                    match result {
+                        Ok(status) => {
+                            match status {
+                                OutputSaveResult::SavedEmpty => {
+                                    prefixed_ui.warn(format!(
```
chris-olszewski, line 1010: I think this might be a little noisy for the new default behavior? I'm fine with shipping as is, but we might need/want a way to silence this warning.

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -1001,15 +1002,33 @@ impl ExecContext {
...
+                                        "warning: no files were found that match the configured \
+                                         outputs - make sure \"outputs\" are correctly defined in \
+                                         your `turbo.json` for {}",
+                                        self.task_id
+                                    ));
+                                }
+                                OutputSaveResult::SkippedEmpty => {
+                                    prefixed_ui.warn("skipped saving cache - no files found");
+                                }
+                                _ => (),
```
chris-olszewski, line 1020: Super minor: Explicitly matching against the other variants gets us a compiler check if we end up adding more variants in the future.

```suggestion
                                OutputSaveResult::Saved | OutputSaveResult::Disabled => (),
```

chris-olszewski (approved):
Approving with some minor comments.

tknickman:
Closing in favor of https://github.com/vercel/turborepo/pull/9236!

Issue #2564, opened 2022-11-02
[turborepo] Logging colors are non-deterministic


preynal:
### What version of Turborepo are you using?

1.6.3

### What package manager are you using / does the bug impact?

Yarn v1

### What operating system are you using?

Mac

### Describe the Bug

CLI colors are changing between runs for the same package. 
It sure looks pretty for VC demos, but is very confusing, as a user expects the package name to have the same color between consecutive runs.

<img width="490" alt="Screenshot 2022-11-02 at 17 22 32" src="https://user-images.githubusercontent.com/1218704/199544852-ba27bb28-8ef9-47d1-a6fb-21d18b9925e4.png">

<img width="302" alt="Screenshot 2022-11-02 at 17 23 05" src="https://user-images.githubusercontent.com/1218704/199544921-d549c03f-7fb9-455f-8b81-bc655c59f3bc.png">

### Expected Behavior

Colors of a package name in the CLI must remain the same no matter when or where you run the CLI

### To Reproduce

Run a turbo repo command twice, the colours will randomly change.

mehulkar:
I believe this is not deterministic because the order of tasks running is also not deterministic (by design). We could base the color on the task name itself to make it deterministic, but I'm not sure how we'd support an unbounded number of colors (I'm sure it's possible, just not sure off the top of my head what's available)

The code is here:

https://github.com/vercel/turbo/blob/261d22177f6c7301f9c5e93518112b025b15d7aa/cli/internal/colorcache/colorcache.go#L33-L45

scamden:
wouldn't have to be unbound could just hash to an integer and modulus to the number of colors to least stay consistent. just a thought (i noticed this even for a single task)

jtlapp:
I just had a build fail and was left typing invisible characters into the shell prompt. My background was black, so maybe the characters were also black.

Hacksore:
What's even more undesirable is even the replayed output from a `>>> FULL TURBO` cached run does not yield a deterministic color 😢.

I'd imagine with the port to rust it would be a good idea to make the color be derived from the package/task name.

cc @anthonyshew since were talking about this the other day.

opiredev:
@nabby27 created a $50.00 reward using [Opire](https://opire.dev) 
 <details><summary>How to earn this reward?</summary>

Since this project does not have Opire bot installed yet 😞 you need to go to [Opire](https://app.opire.dev) and claim the rewards through your programmer's dashboard once you have your PR ready 💪</details>

[...]

PR #9008, opened 2024-08-14
fix: consistent colors for packages

Shaharking:
### PR Description

This PR addresses issue [#2564](https://github.com/vercel/turbo/issues/2564) tagged as a good first issue.

**Changes:**

-   Used hashing on the package name to ensure consistent color assignment.
-   Expanded the color set for better distinction.

**Additional Notes:**

-   I encountered challenges running all the integration tests and was unable to figure out how to execute the `turborepo-tests`. It might be beneficial to provide a more detailed explanation of the testing process in the `CONTRIBUTING.md` file to assist future contributors.

### Testing Instructions
- The tests today already cover everything


crates/turborepo-ui/src/color_selector.rs
```diff
@@ -1,21 +1,20 @@
...
-            Style::new().cyan(),
-            Style::new().magenta(),
-            Style::new().green(),
-            Style::new().yellow(),
-            Style::new().blue(),
-        ]
+        let colors: [Style; u8::MAX as usize] =
+            core::array::from_fn(|index| Style::new().color256(index as u8));
+
+        colors
```
chris-olszewski, line 17: I don't believe we want to use all of the available colors as:
 - many of them aren't especially nice to look at
 - many of the colors are very similar e.g. in my tests I got 4 very similar shades of red

Screenshot 2024-08-14 at 1 32 23 PM

chris-olszewski (changes requested):
Thank you for the PR! Do not worry about the integration tests, those will run on CI.

I think what we want is to keep the existing colors, but to hand them out in a consistent way while also making use of all the available colors i.e. not assigning 2 tasks the same color if there's still colors without any tasks.

To be frank, the current setup is overly complicated as it happens during the graph walk which is non-deterministic. I would suggest:
 - Prepopulating the color cache with all keys [when we start the walk](https://github.com/vercel/turborepo/blob/main/crates/turborepo-lib/src/task_graph/visitor.rs#L159). This will look something like:
 ```
for task in engine.tasks().sorted() {
    self.color_cache.color_for_key(&task.to_string());
}
```
 - Then also switch so we select colors with the task id as well [here](https://github.com/vercel/turborepo/blob/main/crates/turborepo-lib/src/task_graph/visitor.rs#L670)

**Comment by Shaharking** [comment]
Got it @chris-olszewski.

I see that you suggested using the sorted task name to choose a color and caching it in a hashmap to ensure consistent color assignment. A straightforward approach would be to use an index and increment it each time we need a color for a new task.

Since the tasks are ordered by name, this method will generally guarantee the same color for a monorepo that doesn't change much.

However, if the order changes—like when adding or removing a package—there’s a good chance the colors will no longer match.

We could take a more complex approach to ensure consistent colors even when new packages are added. This could involve hashing the task name and using the result modulo the number of unused colors.

(hash % unused_colors.len())

The number of available colors decreases each time a color is assigned. Once all colors are used, we start again with the full color collection.

I think that for new projects that add lots of packages and change frequently, this feature might not be very beneficial. While approach 2 is a bit more complicated, it's still straightforward enough. I’m inclined to go with this approach—let me know your thoughts.

**Comment by Shaharking** [comment]
Hey @chris-olszewski,

I've updated the PR to include the following:

-   **Prepopulating the color cache**: All keys are now prepopulated when we start the walk, as suggested. 
```
for task in engine.tasks().sorted() {
    self.color_cache.color_for_key(&task.to_string());
}
```
    
-   **Selecting colors using the task ID**: I’ve also switched the color selection to use the task ID, as recommended [here](https://github.com/vercel/turborepo/blob/main/crates/turborepo-lib/src/task_graph/visitor.rs#L670).

Regarding the issue of similar colors: What do you suggest? We could limit the palette to more distinct colors, but this might result in more frequent repetition.

File: crates/turborepo-ui/src/color_selector.rs (hunk 1 above)
Comment (chris-olszewski) [reply crates/turborepo-ui/src/color_selector.rs:17]: I'm still opposed to using all of the colors here for the reasons listed above. 

We can expand our list in a separate PR, but we need to make sure the colors selected are easy to read in most terminals.

File: crates/turborepo-ui/src/color_selector.rs [hunk 2]
```diff
@@ -65,13 +74,40 @@ impl ColorSelectorInner {
...
+            .or_insert_with(|| &colors[chosen_color])
+    }
+
+    pub fn get_color_hash_by_key(key: &str) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        key.hash(&mut hasher);
+        hasher.finish()
+    }
+
+    fn get_color_id_by_key(&mut self, key: &str) -> usize {
```
chris-olszewski, line 90: This implementation suffers from the same issue as the existing one where adding/removing tasks will result in a changed color for the task when there are hash collisions. If both `a#build` and `b#build` get the same initial `color_id`, then removing `a#build` will result in `b#build` changing colors.

Given we're targeting such a small hash (5 at the moment, 256 if we allow for all colors) hash collisions are fairly likely.

chris-olszewski (changes requested):
Just the `visitor.rs` changes in this PR are a huge improvement:
 - color order is no longer tied to task execution order
 - color is no longer tied to task hashes which change frequently

I would prefer to merge these improvements as they will make life better for many developers. As you mentioned there are still issues with tasks being added/removed from the graph, but that can be addressed in future PRs.

If you would move the changes to `color_selector.rs` to a new PR, then I would happily merge the remaining changes.

Shaharking:
Sure @chris-olszewski,
Here I created a seprate pr
https://github.com/vercel/turborepo/pull/9023

Shaharking:
@chris-olszewski 
I think we can increase the palette to 14 distinct colors. You can check out these references for more information:

    https://superuser.com/questions/783656/whats-the-deal-with-terminal-colors
    https://askubuntu.com/questions/1305637/what-do-the-colors-in-a-terminal-color-scheme-correspond-do
    https://github.com/mbadolato/iTerm2-Color-Schemes

I excluded black because most users use it as their background color.

This change should reduce color collisions between packages, and having 14 colors, as opposed to 5, should make it easier to visually distinguish issues.

Shaharking:
Hey @chris-olszewski, is increasing the number of colors worth pursuing, or do you prefer to skip it? I'm unsure whether to continue testing different colors due to the lack of response.

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 46f49c4563f77a36c99c5a2561707681838983e0

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 46f49c4563f77a36c99c5a2561707681838983e0

chris-olszewski:
@Shaharking Apologies for dropping communication. The team is currently not interested in expanding the possible color list.

PR #9011, opened 2024-08-14
TypeScript types for `cacheDir`, `daemon`, and `envMode`

dimitropoulos:
### Description

This PR adds to the TypeScript types for [cacheDir](https://github.com/vercel/turborepo/pull/8947), [daemon](https://github.com/vercel/turborepo/pull/8728), and [envMode](https://github.com/vercel/turborepo/pull/8757), all recently added items to the `turbo.json`.

### Testing Instructions

None, just ensure the values and documentation looks correct.  The links don't all exist yet, but are being added in a followup PR to the docs.

### DIFF
--- a/packages/turbo-types/src/types/config.ts
+++ b/packages/turbo-types/src/types/config.ts
@@ -280,6 +280,36 @@ export interface Pipeline {
    * Documentation: https://turbo.build/repo/docs/reference/configuration#interactive
    */
   interactive?: boolean;
+
+  /**
+   * Specify the filesystem cache directory.
+   *
+   * Documentation: https://turbo.build/repo/docs/reference/configuration#tasks
+   *
+   * @defaultValue `"".turbo/cache"`
+   */
+  cacheDir?: RelativeUnixPath;
+
+  /**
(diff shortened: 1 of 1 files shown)


packages/turbo-types/src/types/config.ts
```diff
@@ -280,6 +280,36 @@ export interface Pipeline {
...
+  cacheDir?: RelativeUnixPath;
+
+  /**
+   * Turbo can run a background process to pre-calculate values used for determining work that needs to be done. This standalone process (daemon) is an optimization, and not required for proper functioning of turbo.
+   *
+   * Documentation: https://turbo.build/repo/docs/reference/configuration#daemon
+   *
+   * @defaultValue `false`
+   */
+  daemon?: boolean;
```
tknickman, line 300: Shouldn't these two be at the root, and not on the task?

dimitropoulos: GREAT catch.  they in fact are all on the root schema (as far as I understand, at least! gonna be quite the shock if I'm wrong about that considering I implemented all three, but please do feel free to check!)

fixed in e1bbdd809977e0746240cd41bcd9e8872189a7c5

tknickman: Nice, I could see an argument for envMode being per task, but the others should definitely be global.

chris-olszewski (approved):

PR #9012, opened 2024-08-14
makes @defaultValue usages consistent (and correct)

dimitropoulos:
### Description

In making https://github.com/vercel/turborepo/pull/9011 I noticed that some (but not most) of the [@defaultValue](https://tsdoc.org/pages/tags/defaultvalue/) tags used the inline code blocks (added to all in 32f30d2b2446190f6512ffec83ea0138587c1539), and others were missing a @defaultValue altogether (which, we do have some booleans that default to true, so it's best to include them) (added in 1ff9ae7e453189726deeebd9fb0cb5a7685453f5). 

### Testing Instructions

Make sure the defaults added in 1ff9ae7e453189726deeebd9fb0cb5a7685453f5 are correct.  Otherwise, you can hover over any of these types to see the codeblock used in the hover overlay.


tknickman (approved):
Nice!

PR #9018, opened 2024-08-16
documentation for for `cacheDir`, `daemon`, and `envMode` (config values, and env vars)

dimitropoulos:
### Description

This PR adds documentation for https://github.com/vercel/turborepo/pull/8947, https://github.com/vercel/turborepo/pull/8728, and https://github.com/vercel/turborepo/pull/8757, all recently added items to the turbo.json (and env vars).

### Testing Instructions

View the preview deployment :point_down:

### DIFF
--- a/docs/repo-docs/reference/configuration.mdx
+++ b/docs/repo-docs/reference/configuration.mdx
@@ -131,7 +131,7 @@ Specify the filesystem cache directory.
 
 ### `daemon`
 
-Default: `false`
+Default: `true`
 
 Turbo can run a background process to pre-calculate values used for determining work that needs to be done. This standalone process (daemon) is an optimization, and not required for proper functioning of turbo.
 
--- a/docs/repo-docs/reference/configuration.mdx
+++ b/docs/repo-docs/reference/configuration.mdx
@@ -117,6 +117,47 @@ To help with incremental migration or in situations where you can't use the `pac
   environment variable.
 </Callout>
 
+### `cacheDir`
+
+Default: `".turbo/cache"`
+
+Specify the filesystem cache directory.
+
+```jsonc title="./turbo.json"
+{
+  "cacheDir": ".turbo/cache"
+}
+```
+
+### `daemon`
+
+Default: `false`
+
+Turbo can run a background process to pre-calculate values used for determining work that needs to be done. This standalone process (daemon) is an optimization, and not required for proper functioning of turbo.
+
+```jsonc title="./turbo.json"
+{
+  "daemon": true
+}
+```
+
+### `envMode`
+
+Default: `"strict"`
+
+Turborepo's Environment Modes allow you to control which environment variables are available to a task at runtime:
+
+- `"strict"`: Filter environment variables to only those that are specified in the `env` and `globalEnv` keys in `turbo.json`.
+- `"loose"`: Allow all environment variables for the process to be available.
+
+Read more about [Environment Modes](/repo/docs/crafting-your-repository/using-environment-variables#environment-modes).
+
+```jsonc title="./turbo.json"
+{
+  "envMode": "strict"
+}
+```
+
 ## Defining tasks
 
 ### `tasks`
--- a/docs/repo-docs/reference/configuration.mdx
+++ b/docs/repo-docs/reference/configuration.mdx
@@ -133,7 +133,7 @@ Specify the filesystem cache directory.
 
 Default: `true`
 
-Turbo can run a background process to pre-calculate values used for determining work that needs to be done. This standalone process (daemon) is an optimization, and not required for proper functioning of turbo.
+Turborepo runs a background process to pre-calculate some expensive operations. This standalone process (daemon) is a performance optimization, and not required for proper functioning of `turbo`.
 
 ```jsonc title="./turbo.json"
 {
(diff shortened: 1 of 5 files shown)


docs/repo-docs/reference/configuration.mdx
```diff
@@ -117,6 +117,47 @@ To help with incremental migration or in situations where you can't use the `pac
...
+
+```jsonc title="./turbo.json"
+{
+  "cacheDir": ".turbo/cache"
+}
+```
+
+### `daemon`
+
+Default: `false`
```
anthonyshew, line 134:
```suggestion
Default: `true`
```

docs/repo-docs/reference/configuration.mdx
```diff
@@ -117,6 +117,47 @@ To help with incremental migration or in situations where you can't use the `pac
...
+
+Turbo can run a background process to pre-calculate values used for determining work that needs to be done. This standalone process (daemon) is an optimization, and not required for proper functioning of turbo.
```
anthonyshew, line 136:
```suggestion
Turborepo runs a background process to pre-calculate values used for determining work. This standalone process (daemon) is a performance optimization, and not required for proper functioning of `turbo`.
```

docs/repo-docs/reference/configuration.mdx
```diff
@@ -117,6 +117,47 @@ To help with incremental migration or in situations where you can't use the `pac
...
+{
+  "daemon": true
+}
+```
+
+### `envMode`
+
+Default: `"strict"`
+
+Turborepo's Environment Modes allow you to control which environment variables are available to a task at runtime:
```
anthonyshew, line 148: In general, we like to provide the link the first time we mention a term. The game I like to play is: Imagine your coworker just linked you this page and you're only glancingly aware that Turborepo is in your company's codebase. You're not terribly familiar with Turborepo but you're here to learn more about it.

Putting yourself in those shoes, you're likely to read this "Environment Modes" term and think "Uh, what's that?", and want to be linked to more information in that moment. You could, of course, keep reading and make it to the end of this section, but you'd probably appreciated the link sooner.

So, that said, I'd be lifting the link that's a few lines below to here, and removing the sentence on 153. Both a context and conciseness win!

anthonyshew (changes requested):
Love it. Some nitpicks but great overall.

Let's get this labelled with the `next-release` label and hold it until then (unless its already out in a patch and I haven't realized it?).

docs/repo-docs/reference/configuration.mdx
dimitropoulos: this was lifted directly (unchanged) from the existing docs (which you wrote? haha) - should I update there too?

docs/repo-docs/reference/configuration.mdx
dimitropoulos: When I try this locally it appears to be false by default:

Screenshot_20240817_145952

Also the config command returns null ([we have a test for this](https://github.com/vercel/turborepo/blob/219c1140d781a7bd5657e7d718e5e96247c8b00c/turborepo-tests/integration/tests/config.t#L66)) by default (not true).

Also my reading of [the code where this value is finally consumed](https://github.com/vercel/turborepo/blob/219c1140d781a7bd5657e7d718e5e96247c8b00c/crates/turborepo-lib/src/run/builder.rs#L263) is that if it's not `Some(true)`, then the daemon won't be used.

Am I missing something?

docs/repo-docs/reference/configuration.mdx
dimitropoulos: this was lifted verbatim from the [existing docs](https://github.com/vercel/turborepo/blob/219c1140d781a7bd5657e7d718e5e96247c8b00c/docs/repo-docs/crafting-your-repository/using-environment-variables.mdx?plain=1#L84-L87) - it seems strange, in this case, to link to other docs when the place the link goes says exactly the same thing that's here.  would you like to just link it anyway?

docs/repo-docs/reference/configuration.mdx
anthonyshew: Oh, wow, interesting. I think this is meant to be on by default. You're right, but we may want to follow up on this as a team.

docs/repo-docs/reference/configuration.mdx
anthonyshew: Yes, definitely. 😄

docs/repo-docs/reference/configuration.mdx
anthonyshew: Ah, heard! Ignore then. 😄

docs/repo-docs/reference/configuration.mdx
dimitropoulos: update: after internal investigation we determined that it is on by default and the above screenshot where I determined that it isn't was due to some other problem or error.  Also, I was _very close_ on the code I believed to be responsible for turning the daemon on, but that was the code for _package discovery_ and the _actual_ place was [a few lines above that](https://github.com/vercel/turborepo/blob/219c1140d781a7bd5657e7d718e5e96247c8b00c/crates/turborepo-lib/src/run/builder.rs#L207).

this was fixed in dcc630e7f07828e25a53697073387895752a297b

chris-olszewski (approved):

Systemcluster:
Will this be propagated to https://turbo.build/schema.json or https://turbo.build/schema.v2.json?

PR #9025, opened 2024-08-17
separate configs, colocate JSONSchema, remove `globalDotEnv` and `pipeline.dotEnv` from V2

dimitropoulos:
### Description

This PR accomplishes a few things related to config types and schema creation:

#### Removed values from Config V2

`globalDotEnv` and `pipeline.dotEnv` are no longer valid config values.  This PR removes them.  In approaching this, I took any concession necessary to avoid changing the runtime logic (and risk breaking a migration).

Thanks to @devy-bee for bringing this to our attention in [a recent PR submission](https://github.com/vercel/turborepo/pull/8974).

#### Fully separate SchemaV1 and SchemaV2

We can still continue using all the same types (i.e. without the `V#` suffix), but now the two versions of the schema don't depend on each other.  This allows for a few useful things, but one of which is that it means we have specific types to target (e.g. `PipelineV1`, which didn't exist before) for migrations and JSON Schema validation and such.

#### JSON Schema is colocated with the types that create it

This creates the JSON Schema here, now.  This is beneficial for a lot of reasons.  I checked it into the repo (instead of making it some kind of build artifact) which means we'll have it highlighted to us when it changes (during code reviews) whereas before now it would be silent.  It also means we can use the schema locally when adding fields to experience the same validation users will experience (which was a problem before that was also solved in this repo).  It will then be trivial for turbo.build to pull out these JSON files and include them in the public assets directory.

### Testing Instructions

- `turbo.json`: `globalDotEnv` and `pipeline.dotEnv` are no longer valid  (you can use the root).
- observe the full `config-v1.ts` and `config-v2.ts` files and make sure they match your mental representation of what these configs were.


packages/turbo-codemod/src/transforms/migrate-dot-env.ts
```diff
@@ -20,15 +25,16 @@ const INTRODUCED_IN = "2.0.0-canary.0";
 function migrateConfig(config: LegacySchema) {
   if ("globalDotEnv" in config) {
     if (config.globalDotEnv) {
-      config.globalDependencies = config.globalDependencies ?? [];
+      (config as RootSchema).globalDependencies =
+        (config as LegacyRootSchema).globalDependencies ?? [];
```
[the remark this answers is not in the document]
dimitropoulos: (this PR was totally reworked and this comment is no longer applicable)

packages/turbo-utils/src/getTurboConfigs.ts
```diff
@@ -205,11 +205,14 @@ export function getWorkspaceConfigs(
...
-export function forEachTaskDef(
-  config: LegacySchema,
-  f: (value: [string, Pipeline]) => void
-): void {
-  if ("pipeline" in config) {
+export function forEachTaskDef<
+  Item,
+  Schema extends {
+    pipeline?: Record<string, Item>;
+    tasks: Record<string, Item>;
```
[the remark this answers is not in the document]
dimitropoulos: (this PR was totally reworked and this comment is no longer applicable)

packages/turbo-types/package.json
```diff
@@ -14,22 +14,20 @@
   },
   "main": "src/index.ts",
   "types": "src/index.ts",
-  "bin": {
-    "turbo-types-generate": "./src/scripts/codegen.js"
-  },
   "scripts": {
+    "build": "tsc && pnpm generate-schema",
```
chris-olszewski, line 18: We need to update the `outputs` for `@turbo/types#build` to include `schemas` otherwise they won't be present on cache hits.

packages/turbo-types/schemas/schema.json
```diff
@@ -0,0 +1,226 @@
+{
```
chris-olszewski, line 1: I would prefer not checking in codegen artifacts if it can be avoided. Not blocking, but if it can be avoided we should probably rely on `turbo` caching instead of `git`.

packages/turbo-types/src/types/config-v1.ts
```diff
@@ -0,0 +1,301 @@
...
+  remoteCache?: RemoteCacheV1;
+
+  /**
+   * Enable use of the UI for `turbo`.
+   *
+   * Documentation: https://turbo.build/repo/docs/reference/configuration#ui
+   *
+   * @defaultValue `"stream"`
+   */
+  ui?: UIV1;
```
chris-olszewski, line 279: We didn't properly capture what this field looked like in v1. It was a bool named `experimentalUI`. [source](https://turbo.build/schema.v1.json)
```suggestion
  /**
   * Enable use of the new UI for `turbo`
   *
   * Documentation: https://turbo.build/repo/docs/reference/configuration#ui
   *
   * @defaultValue `"false"`
   */
  experimentalUI?: bool;
```

packages/turbo-types/src/types/config-v1.ts
```diff
@@ -0,0 +1,301 @@
...
+   * Disable check for `packageManager` in root `package.json`
+   *
+   * This is highly discouraged as it leaves `turbo` dependent on system
+   * configuration to infer the correct package manager.
+   *
+   * Some turbo features are disabled if this is set to true.
+   *
+   * @defaultValue `false`
+   */
+  dangerouslyDisablePackageManagerCheck?: boolean;
```
chris-olszewski, line 291: This field wasn't present in v1
```suggestion
```

packages/turbo-types/src/types/config-v1.ts
```diff
@@ -0,0 +1,301 @@
...
+   *
+   * "errors-only": Only show output from task failures
+   *
+   * "none": Hides all task output
+   *
+   * Documentation: https://turbo.build/repo/docs/reference/run#--output-logs-option
+   *
+   * @defaultValue `"full"`
+   */
+  outputLogs?: OutputModeV1;
```
chris-olszewski, line 127: This was renamed in the v1 -> v2 migration
```suggestion
  outputMode?: OutputModeV1;
```

chris-olszewski (changes requested):
This is awesome and a lot more understandable.

Only blocking changes are:
 - updates to v1 schema to reflect what `turbo.build/schema.v1.json` currently displays
 - Updating the task definition for `@turbo/types#build` to include the schema files that are now created

Not blocking: checking in codegen.

packages/turbo-types/schemas/schema.json
dimitropoulos: We can do that, but there are some downsides I want to get your thoughts on:

How will we know _how_ it changes the schema when a type is updated?  Like, in a code review how can I see changes to the schema happening (or not happening) on a given PR?

The other reason to do it is that this file is used in the repo (by vscode) and we need something to point to, but if you haven't built yet there will never be anything there.

packages/turbo-types/src/types/config-v1.ts
dimitropoulos: updated in 90e10024e220385151eefb6ed94c94138ffc4d53

packages/turbo-types/src/types/config-v1.ts
dimitropoulos: updated in faa5d8120c4faf1a9e839edc8bf8fd9b473d2829

packages/turbo-types/src/types/config-v1.ts
dimitropoulos: fixed in 1c27fd3c2f779f39e0bd0a1c02fd2e4129d5d9ca

packages/turbo-types/package.json
dimitropoulos: added in a3dbcc7ac20d532b367f3d3f31a04778cff69195

chris-olszewski (approved):

turbo.json
```diff
@@ -1,5 +1,4 @@
 {
-  "$schema": "https://turbo.build/schema.json",
```
tknickman, line 2: Hot take (not blocking) We should leave this here even though we've set the vscode override to be an example for anyone looking for the schema url.

.vscode/settings.json
```diff
@@ -18,5 +18,11 @@
     "files.insertFinalNewline": false,
     "files.trimTrailingWhitespace": false
   },
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "json.schemas": [
+    {
+      "fileMatch": ["turbo.json"],
+      "url": "./packages/turbo-types/schemas/schema.json"
```
tknickman, line 25: Nice!

packages/eslint-plugin-turbo/lib/utils/calculate-inputs.ts
```diff
@@ -132,17 +131,20 @@ function processDotEnv(
...
 ): EnvironmentConfig {
   return {
-    legacyConfig: processLegacyConfig(rootTurboJson.globalDependencies),
-    env: processEnv(rootTurboJson.globalEnv),
-    passThroughEnv: processPassThroughEnv(rootTurboJson.globalPassThroughEnv),
-    dotEnv: processDotEnv(workspacePath, rootTurboJson.globalDotEnv),
+    legacyConfig: processLegacyConfig(schemaV1.globalDependencies),
+    env: processEnv(schemaV1.globalEnv),
+    passThroughEnv: processPassThroughEnv(schemaV1.globalPassThroughEnv),
+    dotEnv: processDotEnv(workspacePath, schemaV1.globalDotEnv),
```
tknickman, line 140: This feels a _little_ confusing since this should work with the V2 schema as well, but the types look explicit to V1 only

packages/turbo-types/schemas/schema.json
tknickman: What's your plan for keeping this up to date and making sure the schema actually gets build if there are changes here?

Also are you thinking about making this artifact the source of truth?

packages/eslint-plugin-turbo/lib/utils/calculate-inputs.ts
dimitropoulos: my understanding is that it can't work with the v2 types, since there are values here that are not present in v2, for example `globalDotEnv`.  what am I missing?

turbo.json
dimitropoulos: ok, I see what you mean.  yeah, it's probably better that way.  ultimately, it looks like we're gonna be trying to just redirect `https://turbo.build/schema.json` directly to this file (hosted here on GitHub) so that means it's gonna work (soon) both ways, so that's fine.  and besides, with the `vscode` setting in place, if we simply remove this schema, we'll see the local one so it's almost like the best of both worlds.

added back in 8ece8cdacee8200812798f2de3aaa765fe111951

packages/eslint-plugin-turbo/lib/utils/calculate-inputs.ts
tknickman: Yea, right now v2 is a subset of v1, which is why this is working fine. It's just a little confusing to see the type as v1 explicitly even though that will also work with v2.

tknickman (approved):

PR #9039, opened 2024-08-21
pipes default values through to json-schema

dimitropoulos:
### Description

I noticed [a typo on a default value](https://github.com/vercel/turborepo/pull/9018/commits/738366dfa4cd1b0042821bc3a8a1490e7a1e16cf) in the config types but updating the schema artifacts yielded no diff.  This is because of yet-another-fragment in the TSDoc/TypeDoc/JSDoc ecosystem.  The schema validator wants `@default` because that's what JSDoc uses, but TSDoc and TypeDoc want `@defaultValue`.  This is a [known issue](https://github.com/vega/ts-json-schema-generator/issues/609) of the generator but the result is that in our schemas the default value is completely dropped (it's been that way all along).  The effort to address [ecosystem fragmentation](https://github.com/microsoft/tsdoc/issues/27) has been in a stand-still for 4 years, so we'll need to address separately.

This wouldn't be a big deal in most cases, but we're in a situation where we have some boolean properties have `true` default values and some use `false`.  That makes it really helpful for users to be able to see this information.

I accomplished this the simplest way I could find: postprocessing the JSON Schema right before printing to disk and leveraging the fact that `ts-json-schema-generator` allows for passing extra (otherwise unrecognized) tags.

### Testing Instructions
Prior to this PR the descriptions don't actually show _anything_ about the default values (!):

But after this PR, observe that when using the new config (you can simulate this by removing the `$schema` key of the root turbo.json) you have hints about what the default values are.


packages/turbo-types/package.json
```diff
@@ -18,12 +18,14 @@
     "build": "tsc && pnpm generate-schema",
     "lint": "eslint src/",
     "lint:prettier": "prettier -c . --cache",
-    "generate-schema": "node scripts/generate-schema.mjs"
+    "generate-schema": "tsx scripts/generate-schema.ts"
```
tknickman, line 21: ❤️ long live tsx

packages/turbo-types/scripts/generate-schema.ts
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env node
+/* eslint-disable @typescript-eslint/no-unsafe-assignment -- we're messing with TypeScript internals here */
+/* eslint-disable @typescript-eslint/no-unsafe-call -- we're messing with TypeScript internals here */
```
tknickman, line 3: I generally prefer line disables instead of file disables, but understandable if that would be untenable here.

tknickman (approved):
Looks awesome, but I would please request a unit test or two on the custom parsing logic just in case this silently breaks on us from a version upgrade down the road.

packages/turbo-types/package.json
dimitropoulos: haha I wasn't sure how that'd go over, so this is useful feedback! haha.  I'm pretty lost without types!

packages/turbo-types/scripts/generate-schema.ts
dimitropoulos: doesn't hurt! I moved them in 0fc9e7a66fd8dee5bee26212f9d76cc062d582c3

packages/turbo-types/scripts/generate-schema.ts
```diff
@@ -0,0 +1,52 @@
...
+import { createGenerator } from "ts-json-schema-generator";
+
+const __dirname = new URL(".", import.meta.url).pathname;
+const packageRoot = join(__dirname, "..", "src");
+
+/**
+ * post-process the schema recursively to:
+ * 1. replace any key named `defaultValue` with `default`
+ * 1. remove any backticks from the value
+ * 1. attempt to parsing the value as JSON (falling back, if not)
```
chris-olszewski, line 14:
```suggestion
 * 2. remove any backticks from the value
 * 3. attempt to parsing the value as JSON (falling back, if not)
```

chris-olszewski (approved):

packages/turbo-types/scripts/generate-schema.ts
dimitropoulos: Just making sure you're aware: markdown rendering rules are in effect in JSDoc/TSDoc and this is [the standard](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md029---ordered-list-item-prefix) way of doing things because it removes any room for error in misnumbering:





with that in mind: do you want to change this?

dimitropoulos: (talked to @chris-olszewski offline and he said to go ahead and merge! :smile:)

PR #9042, opened 2024-08-21
feat(tui): allow searching for tasks

chris-olszewski:
### Description

This PR adds the ability to search though the task list by using `/` and scroll through the results with the arrow keys. 

A user can exit the search by pressing `Esc` or `Backspace` when there search bar is already empty to return the the selected task before the search. If they press `Enter` it will exit the search at selected result.

Searching is currently done via a naive substring search, but could be enhanced with a fuzzy matcher if desired.

Each commit can be reviewed on it's own.

### Testing Instructions

Extensive unit tests for state changes.

https://github.com/user-attachments/assets/8cc06e76-e2c8-4639-9c2b-db0d8ae5c636

### DIFF
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::BTreeMap,
     io::{self, Stdout, Write},
+    mem,
     sync::mpsc,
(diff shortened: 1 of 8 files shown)


tknickman (approved):
Looks great! Tested on several large internal repos and all look good.

PR #9045, opened 2024-08-22
allow multiple fallbacks for --affected base branch

dimitropoulos:
### Description

We use `main` as our default for base branch, but a lot of repos are likely using `master`, we should fall back to that if `main` isn't a valid ref.

### Testing Instructions

see unit tests https://github.com/vercel/turborepo/pull/9045/files#diff-db1296fb6f1a06e197e7cd6cdf5979b738c6b56b7700b048ab3b00fd767f8aecR734-R746 for situations to test manually.

### DIFF
--- a/crates/turborepo-scm/src/lib.rs
+++ b/crates/turborepo-scm/src/lib.rs
@@ -64,7 +64,7 @@ pub enum Error {
     #[error(transparent)]
     Walk(#[from] globwalk::WalkError),
     #[error("unable to resolve base branch, please set with TURBO_SCM_BASE")]
-    UnableToResolveRef(),
+    UnableToResolveRef,
 }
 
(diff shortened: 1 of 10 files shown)


crates/turborepo-scm/src/lib.rs
```diff
@@ -63,6 +63,8 @@ pub enum Error {
     Globwalk(#[from] globwalk::GlobError),
     #[error(transparent)]
     Walk(#[from] globwalk::WalkError),
+    #[error("unable to resolve base branch, please set with TURBO_SCM_BASE")]
+    UnableToResolveRef(),
```
chris-olszewski, line 67: Incredibly minor
```suggestion
    UnableToResolveRef,
```

chris-olszewski (approved):

PR #9063, opened 2024-08-26
feat(docs): doc ls

tknickman:
### Description

Docs for ls

### DIFF
--- a/docs/repo-docs/reference/ls.mdx
+++ b/docs/repo-docs/reference/ls.mdx
@@ -1,0 +1,49 @@
+---
+title: ls
+description: API reference for the `turbo ls` command
+---
+
+import { ExperimentalBadge } from '#/components/experimental-badge';
+
+List packages in your monorepo.
+
+```bash title="Terminal"
+turbo ls [package(s)] [flags]
+```
+
+When scoped to the entire repository, output includes package manager, package count, and all package names and directories.
+
+```bash title="Terminal"
+# List all packages in the repository
+turbo ls
+```
+
+When scoped to one or more packages, output includes package name, directory, internal dependencies, and all tasks.
+
+```bash title="Terminal"
+# List only two packages
+turbo ls web @repo/ui [package(s)]
+```
+
+## Flags
+
+### `--affected`
+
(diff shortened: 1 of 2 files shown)


docs/repo-docs/reference/ls.mdx
```diff
@@ -0,0 +1,49 @@
...
+```bash title="Terminal"
+# List all packages in the repository
+turbo ls
+```
+
+When scoped to one or more packages, output includes package name, directory, internal dependencies, and all tasks.
+
+```bash title="Terminal"
+# List only two packages
+turbo ls [package(s)]
```
chris-olszewski: I think it makes sense to specify the packages here instead of leaving the arg placeholder if the comment references listing 2 packages.
```suggestion
# List only two packages
turbo ls web docs
```

docs/repo-docs/reference/ls.mdx
```diff
@@ -0,0 +1,47 @@
...
+When scoped to the entire repository, output includes package manager, package count, and all package names and directories.
+
+```bash title="Terminal"
+turbo ls
+```
+
+When scoped to one or more packages, output includes package name, directory, internal dependencies, and all tasks.
+
+```bash title="Terminal"
+turbo ls [package(s)]
```
chris-olszewski, line 23: Up to you, but I think specifying a fake package name here is clearer than leaving the arg placeholder
```suggestion
turbo ls acme-pkg
```

chris-olszewski (approved):

docs/repo-docs/reference/ls.mdx
tknickman: I agree, that was my first instinct as well but then noticed we didn't really do examples for other docs pages. I'll add it here and then we can always remove later if we want to.

docs/repo-docs/reference/ls.mdx
```diff
@@ -0,0 +1,47 @@
...
+```
+
+## Flags
+
+### `--affected`
```
NicholasLYang, line 28: do we want to mention filter?

PR #9066, opened 2024-08-26
(docs) Fix Node.js subpath import guidance.

anthonyshew:
### Description

@lubieowoce recently taught me that Node.js subpath imports have some conditional behavior. This PR updates for that.

Additionally, I've improved backlinking to this section so this guidance is more discoverable.


tknickman (approved):

anthonyshew:
I'm not sure why the links in the PR check are being reported. I'll take another look there in a separate PR.

PR #9078, opened 2024-08-28
docs: Clarifications for `--affected`.

anthonyshew:
### Description

Title!

### Testing Instructions

👀

### DIFF
--- a/docs/repo-docs/crafting-your-repository/constructing-ci.mdx
+++ b/docs/repo-docs/crafting-your-repository/constructing-ci.mdx
@@ -46,6 +46,8 @@ You can filter your tasks using [the `--filter` flag](/repo/docs/reference/run#-
   available.
 </Callout>
 
+You can also use [the `--affected` flag](#running-only-affected-tasks) to only run tasks that have changes.
+
 ## Docker
 
 Docker is an important part of many deployment pipelines. [Turborepo's `prune` subcommand](/repo/docs/reference/prune) helps you ship lightweight images by removing unnecessary dependencies and code from your images.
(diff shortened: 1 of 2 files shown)


docs/repo-docs/crafting-your-repository/constructing-ci.mdx
```diff
@@ -46,6 +46,8 @@ You can filter your tasks using [the `--filter` flag](/repo/docs/reference/run#-
   available.
 </Callout>
 
+You can also use [the `--affected` flag](#running-only-affected-tasks) to only run tasks that have changes.
```
NicholasLYang, line 49: nit: only run tasks in packages that have changes

docs/repo-docs/crafting-your-repository/constructing-ci.mdx
```diff
@@ -54,7 +56,22 @@ For more on how to deploy from a Turborepo with Docker, visit [the dedicated Doc
...
+```bash title="Terminal
+turbo run build --affected
+```
+
+You'll want to use this flag in situations like:
+
+- You're running many tasks across packages in your monorepo, and only want to run those tasks in packages with code changes.
+- You’re _not_ using a Remote Cache, but still want to do as little work as possible in CI.
+- You _are_ using a Remote Cache, and you’re in a large repository. By minimizing the amount of tasks that will be restored from cache, there will be less data to send across the network, resulting in faster cache restoration.
+- You’re already using [advanced filtering techniques](/repo/docs/reference/run#advanced-filtering-examples) or [`turbo-ignore`](/repo/docs/reference/turbo-ignore) to create the same or similar behavior as `--affected`. You likely have the opportunity to simplify your scripting using this new flag.
```
NicholasLYang, line 72: One big benefit is when you're using a shallow clone. With the normal filter syntax, if the base doesn't exist in the shallow clone, turbo crashes. With affected, we gracefully fall back to running all the tasks

NicholasLYang (approved):
Just some small nits

PR #9084, opened 2024-08-30
remove alias for --daemon/--no-daemon

dimitropoulos:
### Description

in https://github.com/vercel/turborepo/pull/9082 we determined that it's best to move forward with just removing the alias for --no-daemon/--daemon.

### Testing Instructions

- follow instructions in https://github.com/vercel/turborepo/pull/9082 and observe that `--daemon` now works
- `turbo --help` shows lines for both values, `--daemon` and `--no-daemon`

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -846,15 +846,15 @@ pub struct RunArgs {
 
     // clap does not have negation flags such as --daemon and --no-daemon
     // so we need to use a group to enforce that only one of them is set.
-    // we set the long name as [no-]daemon with an alias of daemon such
-    // that we can merge the help text together for both flags
     // -----------------------
-    /// Force turbo to either use or not use the local daemon. If unset
+    /// Force turbo to use the local daemon. If unset
     /// turbo will use the default detection logic.
-    #[clap(long = "(no-)daemon", alias = "daemon", group = "daemon-group")]
+    #[clap(long, group = "daemon-group")]
     pub daemon: bool,
 
-    #[clap(long, group = "daemon-group", hide = true)]
+    /// Force turbo to not use the local daemon. If unset
+    /// turbo will use the default detection logic.
+    #[clap(long, group = "daemon-group")]
     pub no_daemon: bool,
 
     /// File to write turbo's performance profile output into.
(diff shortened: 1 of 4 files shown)


turborepo-tests/integration/tests/conflicting-flags.t
```diff
@@ -1,9 +1,9 @@
 Setup
   $ . ${TESTDIR}/../../helpers/setup_integration_test.sh
   $ ${TURBO} run build --daemon --no-daemon
-   ERROR  the argument '--\(no-\)daemon' cannot be used with '--no-daemon' (re)
+   ERROR  the argument '--daemon' cannot be used with '--no-daemon' (re)
```
dimitropoulos, line 4: why did this ever need to be a `(re)` line?

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -846,15 +846,15 @@ pub struct RunArgs {
 
     // clap does not have negation flags such as --daemon and --no-daemon
     // so we need to use a group to enforce that only one of them is set.
```
chris-olszewski, line 848: Can drop this comment line as well since we're not doing these shenanigans anymore

chris-olszewski (approved):

turborepo-tests/integration/tests/conflicting-flags.t
chris-olszewski: Looking at the history I think it was an accident.

crates/turborepo-lib/src/cli/mod.rs
dimitropoulos: I did drop the second line, but we do still use the group to enforce that you can't use both `--daemon` and `--no-daemon` don't we?

turborepo-tests/integration/tests/conflicting-flags.t
dimitropoulos: cool cool: removed it in de9af56f3413324d029d9bf20b992c7153c21066

crates/turborepo-lib/src/cli/mod.rs
chris-olszewski: Oops, you're right.

PR #9087, opened 2024-08-30
feat: allow for specifying path of root turbo.json

chris-olszewski:
### Description

Allow for setting the root `turbo.json` path via `--root-turbo-json=path/to/my.json` or `TURBO_ROOT_TURBO_JSON=path/to/my.json`. This option is not compatible with watch mode.

Reviewing this PR should be done commit-wise, there's a lot of prefactoring before the final commit that implements the actual feature. 

Major changes as part of prefactor:
 - We parse `turbo.json` based on absolute paths instead of repo root relative ones. This allows us to read a `turbo.json` from wherever a user wants
 - Moved the reading of a trace generated `turbo.json` outside of the primary `turbo.json` reading logic and into `TaskAccess`
 - Changed ordering of how `config` layering happens instead of ordering them from least to most significant and always choosing a present value, we now go from most to least significant and only choose a value if we do not already have a value for it. This is necessary as we now have a config option (`TURBO_ROOT_TURBO_JSON`) that alters a less significant config source (`turbo.json`).
 - Move config sources out of `config.rs`
 - We are now lazy with evaluating our config sources. This allows us to change the path of the `turbo.json` we read after partial evaluation of config sources

### Testing Instructions

Existing unit tests for the refactors.
Manual testing of the traced config logic.
Integration test for new feature.


crates/turborepo-lib/src/config/mod.rs
```diff
@@ -0,0 +1,651 @@
...
+            Box::new(local_config),
+            Box::new(global_auth),
+            Box::new(global_config),
+            Box::new(turbo_json),
+        ];
+
+        let config = sources.into_iter().try_fold(
+            ConfigurationOptions::default(),
+            |mut acc, current_source| {
+                let mut current_source_config = current_source.get_configuration_options(&acc)?;
```
chris-olszewski, line 486: There was an existing bug where we would ignore any errors that came from resolving config options from a source. In reality since these were (mostly) eagerly calculated it only affected and validation errors from extracting config options from a `turbo.json`.

We now bubble up any errors generated by resolving config options.

turborepo-tests/integration/tests/run/no-root-turbo.t
```diff
@@ -0,0 +1,46 @@
...
+  my-app:build: > echo building
+  my-app:build: 
+  my-app:build: building
+  
+   Tasks:    1 successful, 1 total
+  Cached:    0 cached, 1 total
+    Time:\s*[\.0-9]+m?s  (re)
+  
+
+Run with --root-turbo-json should use specified config
```
NicholasLYang, line 29:
```suggestion
Run with TURBO_ROOT_TURBO_JSON should use specified config
```

crates/turborepo-lib/src/run/task_access.rs
```diff
@@ -245,6 +245,25 @@ impl TaskAccess {
         }
     }
 
+    /// Attempt to load a task traced turbo.json
+    pub fn load_turbo_json(&self, root_turbo_json_path: &AbsoluteSystemPath) -> Option<TurboJson> {
+        if !self.enabled {
+            return None;
+        }
+        let trace_json_path = self.repo_root.join_components(&TASK_ACCESS_CONFIG_PATH);
```
NicholasLYang, line 253: In the original code, we first resolve the `dir` arg then join to `TASK_ACCESS_CONFIG_PATH`. Do we no longer need the `dir` part?

NicholasLYang (approved):
Very nicely done! Love the PR organization. Just a couple nits and one question about task access config

crates/turborepo-lib/src/run/task_access.rs
tknickman: Not required in this PR we can fix this when we need it but this should be aware of the cache dir as well beacuse the intention is that the config gets cached as an artifact.

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -401,11 +401,16 @@ impl TryFrom<RawTaskDefinition> for TaskDefinition {
...
     pub(crate) fn read(
         repo_root: &AbsoluteSystemPath,
-        path: &AnchoredSystemPath,
+        path: &AbsoluteSystemPath,
     ) -> Result<RawTurboJson, Error> {
-        let absolute_path = repo_root.resolve(path);
-        let contents = absolute_path.read_to_string()?;
-        let raw_turbo_json = RawTurboJson::parse(&contents, path.as_str())?;
+        let contents = path.read_to_string()?;
+        // Anchoring the path can fail if paths reside on different drives.
```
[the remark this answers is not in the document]
chris-olszewski: You're correct.

tknickman (approved):

PR #9101, opened 2024-09-03
fix(tui): start pty with matching TUI size

chris-olszewski:
### Description

Fixes https://github.com/vercel/turborepo/issues/9060

[From the issue](https://github.com/vercel/turborepo/issues/9060#issuecomment-2312651231)

> I think I see the issue, it appears that rspack adds enough trailing spaces in the progress bar output the fill the current line:
> 
> ```
> ● ESC[1mESC[0m ESC[32mESC[37mESC[2m━━━━━━━━━━━━━━━━━━━━━━━━━ESC[0mESC[0m (0%) ESC[2msetup compilation                                                                                                                                           ESC[0m^MESC[2K
> ```
> 
> The `ESC[2K` operation at the erases the current line, but if the trailing spaces have caused the line to wrap, it won't erase the intended line instead just the line overflow. This probably indicates we have a mismatch in the PTY dimensions and the virtual terminal dimensions so `rspack` is outputting too many spaces for the terminal.
> 
> I'll look at making sure the underlying PTY has the exact same dimensions as the virtual terminal.

There is future work to make sure that we update underlying PTY instances if the pane changes sizes, but that requires a pretty heavy refactor to our process management code to communicate with the PTY "controller" as we only hold onto the "receiver".

### Testing Instructions

Before:
Screenshot 2024-09-03 at 9 34 21 AM

After:
Screenshot 2024-09-03 at 9 30 20 AM

### DIFF
--- a/crates/turborepo-lib/src/process/child.rs
+++ b/crates/turborepo-lib/src/process/child.rs
@@ -33,7 +33,7 @@ use tokio::{
 };
 use tracing::{debug, trace};
 
-use super::Command;
+use super::{Command, PtySize};
 
 #[derive(Debug)]
 pub enum ChildState {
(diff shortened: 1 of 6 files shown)


dimitropoulos (approved):
I tested with

```console
$ npx create-turbo@latest -e with-shell-commands _reproduce -m npm && cd _reproduce && npm create rsbuild@latest -- -d ./apps/app -t react-ts && npm i && npx turbo build --ui-tui
```

Then I ran again with the local build and I got the intended fix as advertised!

PR #9103, opened 2024-09-03
chore(path): make path reading api more rusty

chris-olszewski:
### Description

Remove `read_existing_to_string_or` in favor of `read_existing_to_string` to consolidate on the more Rust-y API for reading a path to a file.

`read_existing_to_string_or` is an awkward API due:
 - It conflates two outcomes: The file not existing and the file existing with the contents of the default
 - The default is provided as `Result<impl Into<String>, io::Error>`. `impl Into<String>` was chosen so unnecessary work could be avoided. This is limiting compared to `FnOnce() -> String`.
 - It prevents us from using the battle tested `Option` API

The benefits of not conflating the two outcomes is best displayed when trying to read a config file. Instead of defaulting having a default of `{}` and then parsing it as JSON we can skip the parse and just use `ConfigurationOptions::default()`

### Testing Instructions

Existing unit tests + 👀

### DIFF
--- a/crates/turborepo-lib/src/commands/link.rs
+++ b/crates/turborepo-lib/src/commands/link.rs
@@ -215,11 +215,12 @@ pub async fn link(
 
             let local_config_path = base.local_config_path();
             let before = local_config_path
-                .read_existing_to_string_or(Ok("{}"))
+                .read_existing_to_string()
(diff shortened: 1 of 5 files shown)


tknickman (approved):

NicholasLYang (approved):
Didn't even realize we had this API

Issue #7909, opened 2024-04-06
Docs: eslint v9


ramirezj:
### What is the improvement or update you wish to see?

Documentation on supporting users who have migrated to the new `eslint.config.js` config file or who want to continue to use `estlintrc` but run >= v9 of eslint

Aiming to take a look at this myself, but flagging in case someone has more time to spend on this or a solution already

### Is there any context that might help us understand?

With the recent v9 release of eslint, it [now uses flat config files by default](https://eslint.org/blog/2023/10/flat-config-rollout-plans/) and support for the deprecated `eslintrc` config file will be removed entirely in v10

### Does the docs page already exist? Please link to it.

https://turbo.build/repo/docs/handbook/linting/eslint

anthonyshew:
Would definitely appreciate help on this one! I just explored this a little bit and it looks like this will be a net-positive change for Turborepo users, but haven't had time to explore figuring out the new syntax.

ramirezj:
Digging in between other work, FYI

ch-o-min:
any change?

MickL:
Is there any workaround for eslint configs like the one from Turbo that dont have a flat config yet? :)

goosewobbler:
@MickL ESLint provide some tooling for this:

https://github.com/eslint/eslintrc#usage-esm
https://eslint.org/docs/latest/use/configure/migration-guide
https://eslint.org/blog/2024/05/eslint-configuration-migrator

ChurroC:
I recently created my own solution that works pretty well. I just created a new project and just ported over the eslint in this repo: https://github.com/ChurroC/EslintV9-with-TurboPack. This works for me but it is pretty sketch currently.

controversial:
The following flat config works great for me:

```ts
import turboPlugin from 'eslint-plugin-turbo';

export default {
  name: 'eslint-config-turbo (recreated flat)',

  plugins: {
    turbo: { rules: turboPlugin.rules },
  },

  rules: {
    'turbo/no-undeclared-env-vars': 'error',
  },
}
```

Merieli:
> A seguinte configuração simples funciona muito bem para mim:
> 
> ```ts
> import turboPlugin from 'eslint-plugin-turbo';
> 
> export default {
>   name: 'eslint-config-turbo (recreated flat)',
> 
>   plugins: {
>     turbo: { rules: turboPlugin.rules },
>   },
> 
>   rules: {
[...]

PR #9108, opened 2024-09-04
feat(examples): with eslint flat config

jnlsn:
### Description

I recently went through the headache of upgrading to ES Lint's flat config for a project and wanted to share what I learned here. I have cloned the `basic` example under `with-eslint-flat-config` and updated configs to be as close to the basic example while using the new flat structure. I'm happy to make this change elsewhere and/or update documentation if that is preferable. Feedback appreciated; please let me know how I can help.

Related to: #7909 

### Testing Instructions

1. navigate to examples/with-eslint-flat-config
2. pnpm i
3. pnpm lint


anthonyshew:
Hey, @jnlsn! Would you want to port this work over to the `basic` example itself? That example needs to be migrated to v9 anyway, so it would be great to have one fewer example to maintain.

anthonyshew:
I just finished up handling this with the `basic` example in https://github.com/vercel/turborepo/pull/9501! I appreciate your contributing (even if the PR didn't land) and certainly looked to you for inspiration while trying to figure out 9501. Thank you!

PR #9110, opened 2024-09-05
eslint plugin no-undeclared-env-vars should detect known framework variables

dimitropoulos:
### Description

This PR picks up from https://github.com/vercel/turborepo/pull/8505

We currently do [Framework Inference](https://turbo.build/repo/docs/crafting-your-repository/using-environment-variables#framework-inference) based on the package.json dependencies of the nearest package.json. turbo's Framework Inference support is per-package whereas this one wholesale allowslists all framework environment variables in every package.

The job of this PR, therefore, is to enable per-package framework detection.

I started by extracting this data to a json file, frameworks.json, so it can be shared between Rust and TypeScript.

TODOs:
- [x] extract the frameworks information in `framework.rs` into `frameworks.json`
- [x] search community to make sure that recursive package.json searching has no better method than fs.readFileSync
- [x] implement package.json search to resolve depencies
- [x] combine per-package resolution of those dependencies and regex allowlist values
- [ ] update tests for the plugin
- [ ] use the `frameworks.json` file to generate the table in `using-environment-variables.mdx`

### Testing Instructions

After building the plugin (which will run it locally), open up `packages/eslint-plugin-turbo/__fixtures__/framework-inference/apps/vite`.  Notice that there are no errors in `index.js`, but if you remove the `vite` dependency in `package.json` then there will be.  Notice, also, that adding a next.js env var in this file (e.g. `process.env.NEXT_PUBLIC_ZILTOID`) will error as expected.


crates/turborepo-lib/src/frameworks.json
```diff
@@ -0,0 +1,106 @@
...
+    "dependencyMatch": {
+      "strategy": "all",
+      "dependencies": ["@sveltejs/kit"]
+    }
+  },
+  {
+    "slug": "create-react-app",
+    "envWildcards": ["REACT_APP_*"],
+    "dependencyMatch": {
+      "strategy": "all",
```
dimitropoulos, line 62: wrong

dimitropoulos: (fixed)

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -1,17 +1,49 @@
...
+  const parentDir = path.dirname(currentDir);
+
+  // If we've reached the root directory, stop searching
+  if (parentDir === currentDir) {
+    return null;
+  }
+
+  // Recursively search in the parent directory
+  return findClosestPackageJson(parentDir);
+};
```
tknickman, line 45: We have a util for this you can use:

https://github.com/vercel/turborepo/blob/main/packages/turbo-utils/src/searchUp.ts

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -81,6 +172,17 @@ function create(context: RuleContextWithOptions): Rule.RuleListener {
     }
   });
 
+  const filename = context.getFilename();
+  debug(`Checking file: ${filename}`);
+
+  const matches = frameworkEnvMatches(filename);
```
tknickman, line 178: There's a lot of things in here that _could_ go wrong. Reading files / parsing JSON etc. We should probably try / catch this to be as defensive as we can, and if something goes wrong, log it - but continue.

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -67,11 +99,70 @@ function normalizeCwd(
...
+  >;
+
+  return [
+    "dependencies",
+    "devDependencies",
+    "peerDependencies",
+    // intentionally not including `optionalDependencies` or `bundleDependencies` because at the time of writing they are not used for any of the frameworks we support
+  ]
+    .flatMap((key) => Object.keys(packageJson[key] ?? {}))
+    .reduce((acc, dependency) => acc.add(dependency), new Set<string>());
```
tknickman, line 118: Not required, but feels like a weird use case for reduce. You could just create the Set outside of the return, and then:

```js
const deps = new Set()

[
    "dependencies",
    "devDependencies",
    "peerDependencies",
].flatMap((key) => Object.keys(packageJson[key] ?? {})).forEach(deps.add)

return deps
```

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -67,11 +99,70 @@ function normalizeCwd(
...
+/**
+ * Turborepo does some nice framework detection based on the dependencies in the package.json.  This function ports that logic to this ESLint rule.
+ *
+ * Imagine you have a Vue app.  That means you have Vue in your `package.json` dependencies.  This function will return a list of regular expressions that match the environment variables that Vue depends on, which is information encoded into the `frameworks.json` file.  In Vue's case, it would return the regex `VUE_APP_*` since you have `@vue/cli-service` in your dependencies.
+ */
+const frameworkEnvMatches = (filePath: string): Set<RegExp> => {
+  const directory = path.dirname(filePath);
+  const packageJsonPath = findClosestPackageJson(directory);
+  if (!packageJsonPath) {
+    logger.error(`No package.json found connected to ${filePath}`);
```
tknickman, line 130: Minor suggestion:

```suggestion
    logger.error(`Could not determine package for ${filePath}`);
```

packages/eslint-plugin-turbo/__fixtures__/framework-inference/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "framework-inference",
```
tknickman, line 2: We should (if we don't already) add a test for a package with multiple frameworks. This _could_ come up in a shared UI lib for example

tknickman (commented):
Looks good! A few suggestions and I want to test it out locally but overall looking great.

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
dimitropoulos: thanks! updated in 8bff7e1fa780af7311ba51615454f6c10f4ab764

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
dimitropoulos: I wish we could inline `deps.add` but unfortunately we have an "unbounded-method" rule on (well, fortunate to have the rule on, unfortunate that there's a JavaScript footgun in doing this like you're subscribing).

So it'd have to be

```diff
-  return [
+  const deps = new Set<string>();
+
+  [
     "dependencies",
     "devDependencies",
     "peerDependencies",
     // intentionally not including `optionalDependencies` or `bundleDependencies` because at the time of writing they are not used for any of the frameworks we support
   ]
     .flatMap((key) => Object.keys(packageJson[key] ?? {}))
-    .reduce((acc, dependency) => acc.add(dependency), new Set<string>());
+    .forEach((dependency) => deps.add(dependency));
+
+  return deps;
```

Would you like it to be like that?  Seems normal to use reduce because we're building up the Set with each iteration's new data, but I'm cool with whatever! :)

packages/eslint-plugin-turbo/__fixtures__/framework-inference/package.json
dimitropoulos: sounds great: added a kitchen-sink test here ce55b7d93e7169b5b1379948f2a1001f4e0d8f72

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
dimitropoulos: good call: upon re-inspection for this it was already fairly close, but there were two areas that needed attention.  addressed in 6a3807333acce67cd253ea5d258de285ba10852b

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
tknickman: Ah yea - one of my favorite JS patterns that TS hates ha, but it makes sense why. 

Nah it's fine to leave as is!

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
```diff
@@ -1,10 +1,20 @@
...
 import type { Rule } from "eslint";
 import type { Node, MemberExpression } from "estree";
-import { logger } from "@turbo/utils";
+import { type PackageJson, logger, searchUp } from "@turbo/utils";
+import { frameworks } from "@turbo/types";
 import { RULES } from "../constants";
 import { Project, getWorkspaceFromFilePath } from "../utils/calculate-inputs";
 
+/** set this to true if debugging this rule */
+const debugging = "".length === 0;
```
tknickman, line 11: This is breaking some tests that expect valid JSON on stdout when running eslint with `--json`. 

Leaving this in is fine - but let's change this to read from `process.env.RUNNER_DEBUG`. We use this in our jest configs ([example](https://github.com/vercel/turborepo/blob/main/packages/eslint-plugin-turbo/jest.config.js#L12-L13)) and you can set it locally to debug, and also if you debug a github action run it will output debug logs there as well.

tknickman (commented):
One last thing to fix up that should fix the tests, and then let's ship it!

packages/eslint-plugin-turbo/lib/rules/no-undeclared-env-vars.ts
dimitropoulos: this is a mistake - it should be off when merged and I just forgot.  thank you! fixed in 0fbcc9580071f3a31bc194aa57cc772d04f5a951

tknickman (approved):
Approving so you can merge whenever the tests are good!

anthonyshew:
Very excited for this. Thanks, Dimitri!

PR #9111, opened 2024-09-05
chore(config): leverage proc macros

chris-olszewski:
### Description

This PR leverages [derive_setters](https://crates.io/crates/derive_setters) and [merge](https://docs.rs/merge/latest/merge/index.html) to reduce the amount of changes required when adding a field to the configuration options.

The manual setup is error prone as there are 4 places that need to be updated and forgetting the final place will result in code compiling, but the field never getting layered properly.

### Testing Instructions

Existing unit & integration tests, but also looking what the proc macros are generating (outputs shown are produced using `rust-analyzer`'s [Expand macro](https://rust-analyzer.github.io/manual.html#expand-macro-recursively) feature:

Output of `#[derive(Setters)]`
```
impl TurborepoConfigBuilder {
    pub fn with_api_url(mut self, value: Option<String>) -> Self {
        self.override_config.api_url = value;
        self
    }
...
```
Which exactly matches what `create_builder` would produce:
```
pub fn with_api_url(mut self, value: Option<String>) -> Self {
    self.override_config.api_url = value;
    self
}
```

`#derive(Merge)` produces
```
impl ::merge::Merge for ConfigurationOptions {
    fn merge(&mut self, other: Self) {
        ::merge::Merge::merge(&mut self.api_url, other.api_url);
...
```
and `Merge` is defined for `Option` as :
```
impl<T> Merge for Option<T> {
    fn merge(&mut self, mut other: Self) {
        if !self.is_some() {
            *self = other.take();
        }
    }
}
```
[source](https://docs.rs/merge/latest/src/merge/lib.rs.html#139-145)

So `self.api_url` will only be set to `other.api_url` if there isn't a value present which is the behavior we want since we merge configs from highest to lowest precedence.


crates/turborepo-lib/src/config/mod.rs
```diff
@@ -190,8 +183,13 @@ const DEFAULT_UPLOAD_TIMEOUT: u64 = 60;
 // We intentionally don't derive Serialize so that different parts
 // of the code that want to display the config can tune how they
 // want to display and what fields they want to include.
-#[derive(Deserialize, Default, Debug, PartialEq, Eq, Clone, Iterable)]
+#[derive(Deserialize, Default, Debug, PartialEq, Eq, Clone, Iterable, Merge, Setters)]
 #[serde(rename_all = "camelCase")]
+// Generate setters for the builder type the set these values on it's override_config field
```
NicholasLYang, line 188:
```suggestion
// Generate setters for the builder type that set these values on its override_config field
```

NicholasLYang (approved):

PR #9122, opened 2024-09-06
feat(turbo): add platform env support

tknickman:
### Description

Support reading what variables are set by a user in CI, and diffing those with those in the execution environment. 

## Testing
```
TURBO_SYSTEM_ENV=THIS_NEAT_ENV,SUPER_COOL,TEST THIS_NEAT_ENV=1 ./target/debug/turbo build --skip-infer --force --cwd examples/basic
```


crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -881,6 +924,18 @@ impl ExecContext {
...
 
+        if !self.task_cache.is_caching_disabled() {
+            let missing_platform_env = self.platform_env.validate(&self.execution_env);
+            if !missing_platform_env.is_empty() {
+                let _ = self.warnings.lock().map(|mut warnings| {
+                    warnings.push(TaskWarning {
+                        task_id: self.task_id_for_display.clone(),
+                        missing_platform_env,
+                    });
+                });
```
chris-olszewski, line 935: It's probably worth panicking here if locking fails. Locking should only fail if another thread panicked while holding the lock.
```suggestion
                self.warnings.lock().expect("warnings lock poisoned").push(TaskWarning {
                        task_id: self.task_id_for_display.clone(),
                        missing_platform_env,
                 });|
```

crates/turborepo-lib/src/task_graph/visitor.rs
```diff
@@ -381,6 +384,35 @@ impl<'a> Visitor<'a> {
 
         let global_hash_summary = GlobalHashSummary::try_from(global_hash_inputs)?;
 
+        // output any warnings that we collected while running tasks
+        if let Ok(warnings) = self.warnings.lock() {
+            if !warnings.is_empty() {
+                println!();
+                warn!("finished with warnings");
+                println!();
```
chris-olszewski, line 392: I think we want to print all of this to `stderr`
```suggestion
                eprintln!();
                warn!("finished with warnings");
                eprintln!();
```

crates/turborepo-env/src/platform.rs
```diff
@@ -0,0 +1,104 @@
...
+        missing: Vec<String>,
+        task_id_for_display: &str,
+        color_config: ColorConfig,
+    ) {
+        cprintln!(color_config, YELLOW, "{}", task_id_for_display);
+        for key in missing {
+            cprint!(color_config, GREY, "  - ");
+            cprint!(color_config, GREY, "{}\n", key);
+        }
+        println!();
```
chris-olszewski, line 102: These are all printing to `stdout` instead of `stderr`

chris-olszewski (approved):
This is awesome! Only thing that needs changing is moving all the outputs to `stderr` to avoid breaking uses of `turbo --dry=json` that might happen in CI.

crates/turborepo-lib/src/task_graph/visitor.rs
tknickman: Good call

crates/turborepo-env/src/platform.rs
tknickman: Fixed!

PR #9132, opened 2024-09-10
perf(tui): bring async to tui

chris-olszewski:
### Description

This PR makes the driving the TUI async, the actual operation on the state is still completely sync. The primary driver behind this PR is it allows us to make use of [`crossterm::event::EventStream`](https://docs.rs/crossterm/latest/crossterm/event/struct.EventStream.html) which seems to be far more performant than polling for user input.

The first commit of the PR just changes types from `std::sync::mpsc` to `tokio::sync::mpsc` (and make use of `tokio::sync::oneshot` for our callbacks instead of a channel of size 1). 

The second commit removes our usage of `crossterm::event::poll` in favor of a dedicated task that reads and forwards events from `crossterm::event::EventStream`.

The final commit moves the production of ticks to it's own task to avoid the need for timing out our reads. 

### Testing Instructions

Notice large reduction in CPU usage from `turbo` when tasks are not producing output and the TUI is just waiting for user input.

Before
Screenshot 2024-09-10 at 2 57 45 PM

After
Screenshot 2024-09-10 at 2 56 17 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1388,6 +1388,7 @@ checksum = "f476fe445d41c9e991fd07515a6f463074b782242ccf4a5b7b1d1012e70824df"
 dependencies = [
  "bitflags 2.5.0",
  "crossterm_winapi",
+ "futures-core",
  "libc",
  "mio 0.8.11",
  "parking_lot",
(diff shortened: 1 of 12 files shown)


tknickman (approved):

tknickman:
Fixes https://github.com/vercel/turborepo/issues/9211

PR #9133, opened 2024-09-10
fix(affected): include not committed changes in `--affected`

NicholasLYang:
### Description

We were including new files that were not checked into git, but not changes to files already in git but not committed.

### Testing Instructions
Added more testing to `affected.t` for both new files and existing files with new changes


tknickman (approved):

chris-olszewski (approved):
I'm confused why `--affected` and `--filter=[main]` go down two different codepaths. Are we defaulting the SCM head for `--affected` too early?

You're more experienced in this area of the codebase so stamping to not delay the fix.

turborepo-tests/integration/tests/affected.t
```diff
@@ -5,15 +5,68 @@ Create a new branch
...
+      ]
+    }
+  }
+
+
+Remove the new file
+  $ rm apps/my-app/new.js
+
+Add field to `apps/my-app/package.json`
+  $ jq '. += {"description": "foo"}' apps/my-app/package.json > apps/my-app/package.json.new
```
chris-olszewski, line 62:
```suggestion
  $ jq '. += {"description": "foo"}' apps/my-app/package.json | tr -d '\r' > apps/my-app/package.json.new
```

PR #9134, opened 2024-09-10
feat: turbo-trace

NicholasLYang:
### Description

A long long time ago we had a [node-file-trace](https://github.com/vercel/nft) implementation in the repository. Unfortunately this was tied to turbopack and the turbo-tasks code. Since that's now in a different repository, we're gonna implement our own version, because that's easier than coordinating multi-repo dependencies.

This implementation uses `oxc_resolver` and `swc` to resolve dependencies and parse files.

This PR hooks it up to `turbo query` but we'll probably have other uses later.

### Testing Instructions

Adds some tests in `turbo-trace.t` using `query`


crates/turborepo-lib/src/query/file.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use crate::{query::Error, run::Run};
+
+pub struct File {
+    run: Arc<Run>,
+    path: AbsoluteSystemPathBuf,
+}
+
+impl File {
+    pub fn new(run: Arc<Run>, path: AbsoluteSystemPathBuf) -> Self {
```
chris-olszewski, line 16: Not blocking, but would it make sense to just pass in `repo_root` instead of the entire Run object? It would make writing unit tests a lot easier.

chris-olszewski (approved):
LGTM

Just two minor comments

crates/turborepo-lib/src/query/file.rs
NicholasLYang: Eventually I want to be able to map from a file to affected packages, so that will require the whole run object

PR #9149, opened 2024-09-16
feat: allow usage of `turbo` without turbo.json

chris-olszewski:
### Description

This PR adds the ability to use `turbo` in a monorepo that doesn't have a `turbo.json`. This feature is currently gated behind `--experimental-allow-no-turbo-json`/`TURBO_ALLOW_NO_TURBO_JSON`.

A majority of the PR is refactoring the `EngineBuilder` so that it no longer directly loads `TurboJson`s, but delegates to a `TurboJsonLoader`. This allows us to use different strategies for resolving `TurboJson` loads depending on runtime options e.g. single package mode or task access.

Reviewing this PR is best done by viewing each commit individually.

### Testing Instructions

Unit testing for `turbo.json` loading changes.

Integration test for verifying the new loader is activated with the new env var/flag.

### DIFF
--- a/crates/turborepo-lib/src/engine/builder.rs
+++ b/crates/turborepo-lib/src/engine/builder.rs
@@ -483,11 +483,6 @@ impl<'a> EngineBuilder<'a> {
     }
 
     fn load_turbo_json(&self, workspace: &PackageName) -> Result<TurboJson, Error> {
-        let package_json = self.package_graph.package_json(workspace).ok_or_else(|| {
-            Error::MissingPackageJson {
-                workspace: workspace.clone(),
-            }
-        })?;
         let workspace_dir =
             self.package_graph
                 .package_dir(workspace)
@@ -498,12 +493,7 @@ impl<'a> EngineBuilder<'a> {
             .repo_root
             .resolve(workspace_dir)
             .join_component(CONFIG_FILE);
-        Ok(TurboJson::load(
-            self.repo_root,
-            &workspace_turbo_json,
-            package_json,
-            self.is_single,
-        )?)
+        Ok(TurboJson::load(self.repo_root, &workspace_turbo_json)?)
     }
 }
 
--- a/crates/turborepo-lib/src/engine/builder.rs
+++ b/crates/turborepo-lib/src/engine/builder.rs
@@ -14,8 +14,7 @@ use crate::{
     run::task_id::{TaskId, TaskName},
     task_graph::TaskDefinition,
     turbo_json::{
-        validate_extends, validate_no_package_task_syntax, RawTaskDefinition, TurboJson,
-        TurboJsonLoader,
+        validate_extends, validate_no_package_task_syntax, RawTaskDefinition, TurboJsonLoader,
     },
 };
 
@@ -95,9 +94,8 @@ pub enum Error {
 pub struct EngineBuilder<'a> {
     repo_root: &'a AbsoluteSystemPath,
     package_graph: &'a PackageGraph,
-    turbo_json_loader: TurboJsonLoader,
+    turbo_json_loader: Option<TurboJsonLoader>,
     is_single: bool,
-    turbo_jsons: Option<HashMap<PackageName, TurboJson>>,
     workspaces: Vec<PackageName>,
     tasks: Vec<Spanned<TaskName<'static>>>,
     root_enabled_tasks: HashSet<TaskName<'static>>,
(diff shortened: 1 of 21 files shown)


crates/turborepo-lib/src/engine/builder.rs
```diff
@@ -650,41 +626,6 @@ mod test {
         .unwrap()
     }
 
-    #[test]
-    fn test_turbo_json_loading() {
```
chris-olszewski, line 654: This test is moved to `TurboJsonLoader` unit tests

crates/turborepo-lib/src/turbo_json/loader.rs
```diff
@@ -0,0 +1,660 @@
...
+        package_json: PackageJson,
+    },
+    Noop,
+}
+
+impl TurboJsonLoader {
+    /// Create a loader that will load turbo.json files throughout the workspace
+    pub fn workspace(
+        repo_root: AbsoluteSystemPathBuf,
+        packages: HashMap<PackageName, AbsoluteSystemPathBuf>,
```
chris-olszewski, line 52: This could be made more ergonomic by just taking `&PackageGraph` and keeping the extraction logic in this module. I avoided this to allow for testing loaders without constructing an entire package graph.

chris-olszewski (commented):
Reviewer's Guide

crates/turborepo-lib/src/run/builder.rs
```diff
@@ -359,19 +364,34 @@ impl RunBuilder {
...
+            TurboJsonLoader::single_package(
+                self.repo_root.clone(),
+                self.root_turbo_json_path.clone(),
+                root_package_json.clone(),
+            )
+        } else if self.allow_no_turbo_json && !self.root_turbo_json_path.exists() {
+            let package_scripts = workspace_package_scripts(pkg_dep_graph.packages());
+            // We explicitly set env mode to loose as otherwise no env vars will be passed
+            // to tasks.
+            self.opts.run_opts.env_mode = EnvMode::Loose;
```
NicholasLYang, line 383: could we set this at the task level? I wanna keep `opts` as immutable as possible

NicholasLYang (approved):
Looks great, just a couple small comments

crates/turborepo-lib/src/turbo_json/loader.rs
```diff
@@ -76,15 +80,41 @@ impl TurboJsonLoader {
...
+                .to_owned(),
+            cache: turbo_jsons,
+            strategy: Strategy::Noop,
+        }
+    }
+
     /// Load a turbo.json for a given package
-    pub fn load(&self, package: &PackageName) -> Result<TurboJson, Error> {
+    pub fn load<'a>(&'a mut self, package: &PackageName) -> Result<&'a TurboJson, Error> {
+        if !self.cache.contains_key(package) {
```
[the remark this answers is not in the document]
chris-olszewski: Correct

crates/turborepo-lib/src/turbo_json/loader.rs
```diff
@@ -114,6 +144,7 @@ impl TurboJsonLoader {
                     )
                 }
             }
+            Strategy::Noop => Err(Error::NoTurboJSON),
```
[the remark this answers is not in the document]
chris-olszewski: Only when building the engine do we view a non-existent `turbo.json` as a non-error. I'm not sure requiring the other callers to construct the `NoTurboJSON` error type manually is worth the better ergonomics in the engine builder.

crates/turborepo-lib/src/run/builder.rs
chris-olszewski: Doing this in another PR as adding this functionality to tasks isn't entirely trivial.

crates/turborepo-lib/src/turbo_json/loader.rs
```diff
@@ -246,6 +285,41 @@ fn load_from_root_package_json(
...
 
+fn root_turbo_json_from_scripts(scripts: &[String]) -> Result<TurboJson, Error> {
+    let mut turbo_json = TurboJson {
+        ..Default::default()
+    };
+    for script in scripts {
+        let task_name = TaskName::from(script.as_str()).into_root_task();
+        turbo_json.tasks.insert(
+            task_name,
+            Spanned::new(RawTaskDefinition {
```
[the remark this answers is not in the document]
chris-olszewski: Having trouble getting spans for the script names as `BTreeMap` requires `impl Ord for Spanned<T>` which has issues since it contains `Range` which doesn't implement `Ord`.

I don't think this is worth the effort to plumb through. The only place this span would get used would be for [validating no package task syntax](https://github.com/vercel/turborepo/blob/main/crates/turborepo-lib/src/turbo_json/mod.rs#L671) and having that error point to a `package.json` isn't helpful when we're the ones generating `turbo.json`.

PR #9154, opened 2024-09-16
chore(watch): simplify watch data synchronization

chris-olszewski:
### Description

Basic view of the futures involved with running `watch`

 - The `events_fut` takes events from the package changes watcher and folds them into a `ChangedPackages` struct. It then triggers a notify.
 - The `run_fut` loops waiting for a notification and then performing a run.
 - Signal handler watching for `SIGINT`s

This PR cleans up how packages changes are synchronized between futures:
 - Removes `RefCell` as `Mutex` already implies mutually exclusive access the the underlying resource
 - Remove `run_fut` holding onto the changed packages lock for the entirety of it's run. This allows us to switch from `tokio::sync::Mutex` to `std::sync::Mutex`. Which is suggested by the [`tokio::sync::Mutex` docs](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use)

### Testing Instructions

`rustc` + 👀 . Quick gut check by running `turbo watch` and triggering package changes

### DIFF
--- a/crates/turborepo-lib/src/run/watch.rs
+++ b/crates/turborepo-lib/src/run/watch.rs
@@ -1,4 +1,4 @@
-use std::{cell::RefCell, collections::HashSet, sync::Arc};
+use std::{collections::HashSet, ops::DerefMut, sync::Arc};
 
 use futures::StreamExt;
 use miette::{Diagnostic, SourceSpan};
@@ -172,7 +172,7 @@ impl WatchClient {
         // We explicitly use a tokio::sync::Mutex here to avoid deadlocks.
         // If we used a std::sync::Mutex, we could deadlock by spinning the lock
         // and not yielding back to the tokio runtime.
-        let changed_packages = Mutex::new(RefCell::new(ChangedPackages::default()));
+        let changed_packages = Mutex::new(ChangedPackages::default());
         let notify_run = Arc::new(Notify::new());
         let notify_event = notify_run.clone();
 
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/run/watch.rs
```diff
@@ -189,9 +189,13 @@ impl WatchClient {
         let run_fut = async {
             loop {
                 notify_run.notified().await;
-                let changed_packages_guard = changed_packages.lock().await;
-                if !changed_packages_guard.borrow().is_empty() {
-                    let changed_packages = changed_packages_guard.take();
+                let some_changed_packages = {
```
chris-olszewski, line 192: We only acquire the guard in the block so we ensure we release the lock once we take the changed packages. This lets the `event_fut` make progress while `self.execute_run` is pending.

crates/turborepo-lib/src/run/watch.rs
```diff
@@ -189,9 +189,13 @@ impl WatchClient {
...
+                    let mut changed_packages_guard =
+                        changed_packages.lock().expect("poisoned lock");
+                    (!changed_packages_guard.is_empty())
+                        .then(|| std::mem::take(changed_packages_guard.deref_mut()))
+                };
+                if let Some(changed_packages) = some_changed_packages {
                     self.execute_run(changed_packages).await?;
```
chris-olszewski, line 199: `changed_packages_guard` is still held while run is being executed, this prevents any events from being handled.

chris-olszewski (commented):
Reviewer's guide

NicholasLYang (approved):

PR #9155, opened 2024-09-16
updates `jest`/`ts-jest` to latest

dimitropoulos:
### Description

> [!NOTE]
> there's a lot here, but I tried to make it easier to review by doing a good job of separating the different phases of the PR into their own commits - so for this one you'll probably have a much better review experience if you read commit-by-commit.

I noticed that we were getting [lots of errors](https://github.com/vercel/turborepo/actions/runs/10892777654/job/30226417571?pr=9110#step:6:707) when running tests due to running an unsupported version of TypeScript (as compared to what our version of `jest` knew about).  This is solved by updating our dependencies.  Coincidentally, there's a good few reasons to upgrade anyway, so this comes at a good time.

- [x] update jest to latest
  - [x] update snapshots to new syntax
- [x] update ts-jest to latest
  - [x] update all jest configs to .ts files (fun fact, this hit [a ts-node bug](https://github.com/TypeStrong/ts-node/pull/2091))
    - [x] should not need an eslint override in every config for the default export
- [x] remove @types/jest and used @jest/globals directly
- [x] make sure all tests pass (prior to this PR there were some sinkers in the package tests)

#### Follow-ups after this PR

- [ ] We have `"lint": "eslint src/"` in 16 packages, which skips the test files.  Start linting tests (which, will involve fixing lints in tests).
- [ ] Some of the mocks (e.g. `get-turbo-upgrade-command.test.ts`) are timing out on GitHub Actions, and we believe the cause is running out of I/O (which the tests do a lot of).  Attempt to fix this by moving to an in-memory filesystem for tests.

### Testing Instructions

run `pnpm test`


crates/turborepo-lib/src/turbo_json/loader.rs
```diff
@@ -356,9 +356,9 @@ mod test {
     use crate::{task_graph::TaskDefinition, turbo_json::CONFIG_FILE};
 
     #[test_case(r"{}", TurboJson::default() ; "empty")]
-    #[test_case(r#"{ "globalDependencies": ["tsconfig.json", "jest.config.js"] }"#,
+    #[test_case(r#"{ "globalDependencies": ["tsconfig.json", "jest.config.ts"] }"#,
```
dimitropoulos, line 359: I was unsure about this at first, but it seems to be happy.  please double-check.

packages/turbo-test-utils/src/validateLogs.ts
```diff
@@ -1,21 +1,13 @@
-import type { Spy } from "./spyConsole";
+import { expect } from "@jest/globals";
+import type { SpyConsole } from "./spyConsole";
+
+type Matcher = ReturnType<typeof expect.stringContaining>;
 
 export function validateLogs(
```
dimitropoulos, line 6: so as you can see, I simplified `validateLogs` a lot.  for one thing, we weren't using it correctly (e.g. passing in just one value, and passing in a prefix even though we're also passing a matcher).  it came up because the types were very unhappy and I realized it can be done a lot more simply.

personally, I'd be a fan of just getting rid of it, it's only used in two files and I don't think it'd be terrible to just put the `toHaveBeenNthCalledWith` lines next to each other, but I wanted to keep the scope smaller so I left it for those cases where more than one line is passed in (because with just one, it doesn't make a lot of sense to use the helper).

packages/turbo-codemod/turbo/generators/templates/transformer.test.hbs
```diff
@@ -1,11 +1,12 @@
+import { describe, it, expect } from '@jest/globals';
```
dimitropoulos, line 1: I'm unsure about changes to this file, so please double check. it seems to be ok.

packages/tsconfig/package.json
```diff
@@ -1,5 +1,10 @@
 {
   "name": "@turbo/tsconfig",
   "version": "0.0.0",
-  "private": true
+  "private": true,
+  "exports": {
+    ".": "./src/index.js",
+    "./base.json": "./base.json",
+    "./library.json": "./library.json"
+  }
```
tknickman, line 9: This shouldn't be required and won't do anything without swapping to ESM - does it help the ts server resolve the configs?

packages/turbo-benchmark/src/templates/jest.config.ts
```diff
@@ -1,4 +1,6 @@
-module.exports = {
+import type { Config } from "jest";
+
+const config = {
```
tknickman, line 3: Out of curiosity why `satisfies` instead of:

`const config: Config = { ...`

packages/turbo-codemod/turbo/generators/templates/transformer.test.hbs
tknickman: These are good! These are used for code generation when adding new transformers

packages/turbo-test-utils/src/validateLogs.ts
tknickman: I'm fine with that, it's an ancient helper

examples/with-nestjs/packages/api/package.json
```diff
@@ -41,7 +41,7 @@
     "@repo/typescript-config": "workspace:*",
     "@types/node": "^20.3.1",
     "ts-loader": "^9.4.3",
-    "ts-node": "^10.9.1",
+    "ts-node": "10.9.2",
```
tknickman, line 44: did you mean to remove the `^`?

tknickman (approved):
A few questions but nothing blocking, thanks for tackling this!

packages/tsconfig/package.json
dimitropoulos: ahh yes yes, good catch.  this was while I was debugging (what I didn't know at the time was) [the ts-node bug I mentioned](https://github.com/TypeStrong/ts-node/pull/2091).  basically.... well.. nevermind.  it was a whole thing to figure out what was happening and this now shouldn't be needed.

fixed in 7331be9824c05e8889d80baf451702bc872c3311

tknickman: Sounds good!

packages/turbo-benchmark/src/templates/jest.config.ts
dimitropoulos: because it won't narrow the type.  this came up as an error in `examples/with-nestjs/packages/jest-config/base.ts`.  try changing it there like `const config: Config = {` and you'll see an error in `next.ts`.

originally I just did `: Config` but after seeing that error I thought: "oh, why not, there's no harm in letting the type stay narrow rather than widening it"

examples/with-nestjs/packages/api/package.json
dimitropoulos: truth be told, yes. but it's a bad habit I'm trying to break (my heart is ideologically incorrect so I often remove them: have mercy on me lol)

fixed in 8fcb086b6a033ce28bb046b863a2dc87e46898c2

packages/turbo-benchmark/src/templates/jest.config.ts
tknickman: Got it, thanks!

examples/with-nestjs/packages/api/package.json
dimitropoulos: (and updated lockfile in 51f4a3ba80f9cf5532268d0b58fe06473d7c18d4)

PR #9160, opened 2024-09-17
docs: add -w flag to pnpm example

giraffesyo:
It's suggested to install this into the repository root. If you do this, you must use `--workspace-root` flag or you will get this error:

>  ERR_PNPM_ADDING_TO_ROOT  Running this command will add the dependency to the workspace root, which might not be what you want - if you really meant it, make it explicit by running this command again with the -w flag (or --workspace-root). If you don't want to see this warning anymore, you may set the ignore-workspace-root-check setting to true.

### DIFF
--- a/docs/repo-docs/getting-started/add-to-existing-repository.mdx
+++ b/docs/repo-docs/getting-started/add-to-existing-repository.mdx
@@ -69,7 +69,7 @@ We recommend you install `turbo` both globally and into your repository's root f
     # Global install
     pnpm add turbo --global
     # Install in repository
-    pnpm add turbo --save-dev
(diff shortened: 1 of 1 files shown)


anthonyshew (commented):
I apologize for this nitpick but we prefer to use the long form of flags. Could you update to `--workspace-root` instead of `-w`?

giraffesyo:
> I apologize for this nitpick but we prefer to use the long form of flags. Could you update to `--workspace-root` instead of `-w`?

sounds good, just updated that.

anthonyshew (approved):
Thank you!

PR #9161, opened 2024-09-17
feat(query): add array type

NicholasLYang:
### Description

Added an `Array` type that automatically adds a `length` and `items` field to GraphQL responses.

### DIFF
--- a/crates/turborepo-lib/src/query.rs
+++ b/crates/turborepo-lib/src/query.rs
@@ -44,9 +44,6 @@ struct Array<T: OutputType> {
     length: usize,
 }
 
-unsafe impl<T: Send + OutputType> Send for Array<T> {}
-unsafe impl<T: Sync + OutputType> Sync for Array<T> {}
-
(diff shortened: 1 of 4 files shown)


tknickman (approved):

crates/turborepo-lib/src/query.rs
```diff
@@ -38,6 +38,23 @@ impl Query {
...
 }
 
+#[derive(Debug, SimpleObject)]
+struct Array<T: OutputType> {
+    items: Vec<T>,
+    length: usize,
+}
+
+unsafe impl<T: Send + OutputType> Send for Array<T> {}
+unsafe impl<T: Sync + OutputType> Sync for Array<T> {}
```
chris-olszewski, line 48: Are these necessary? Rust should be automatically implementing these if all fields are `Send`/`Sync`

chris-olszewski (changes requested):

chris-olszewski (approved):
Changing this to approve to unblock, but I do think that the `unsafe impl`s should be removed.

PR #9169, opened 2024-09-19
add method for determining base commit on GitHub actions

dimitropoulos:
### Description

If you're running `--affected` and you're running in the context of a GitHub action there are some [additional env vars that are available](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables) that give more context (pull request action, push action, each have their own thing).  In a push even it's a previous commit, and in the pull request event it's the base of the pull request.  We want to use those to be the base of --affected instead of just defaulting to main or whatever we have.  You can still override it as normal.

This is a problem right now for running in GitHub actions right now because it clones to a limited depth (shallow clone).  We might therefore need to check that the commit in question actually exists.

### Testing Instructions
 
see tests (there are quite a few) - ultimately the only way to "really" test this is to see it run in GitHub runners.


crates/turborepo-scm/src/git.rs
```diff
@@ -81,6 +85,45 @@ impl SCM {
...
+        }
+
+        // Extract the base ref from the push event
+        if let Some(first_commit) = self.commits.first() {
+            let id = &first_commit.id;
+            return Some(format!("{id}^"));
+        }
+
+        // we don't have enough information to determine the base ref
+        None
```
NicholasLYang, line 123: Just a small change, `?` works on `Option` too
```suggestion
        let first_commit = self.commits.first()?; 
        let id = &first_commit.id;
        Some(format!("{id}^"))
```

crates/turborepo-scm/src/git.rs
```diff
@@ -94,11 +137,72 @@ impl Git {
         Ok(output.trim().to_owned())
     }
 
+    /// for GitHub Actions environment variables, see: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
+    pub fn get_github_base_ref() -> Option<String> {
+        // make sure we're running in a CI environment
+        if env::var("CI").ok()?.as_str() != "true" {
```
NicholasLYang, line 143: I'd consider using the `turborepo_ci` package, specifically you can detect the `Vendor` and double check that it's indeed GitHub

crates/turborepo-scm/src/git.rs
```diff
@@ -94,11 +137,72 @@ impl Git {
...
+        None
+    }
+
     fn resolve_base<'a>(&self, base_override: Option<&'a str>) -> Result<&'a str, Error> {
         if let Some(valid_from) = base_override {
             return Ok(valid_from);
         }
 
+        if let Some(github_base_ref) = Self::get_github_base_ref() {
+            return Ok(Box::leak(github_base_ref.into_boxed_str()));
```
NicholasLYang, line 203: What was the reason for this? Leaking memory is probably not ideal, even if it's a small amount. Could we change this function to return a `String`? If we want to be efficient a `Cow<'a, str>` could work, but that's probably overkill

crates/turborepo-scm/src/git.rs
```diff
@@ -244,19 +348,28 @@ mod tests {
...
-    use crate::{Error, Git, SCM};
+    use crate::{
+        git::{GitHubCommit, GitHubEvent},
+        Error, Git, SCM,
+    };
+
+    /// By default `#[test_case]` will run in parallel.
+    /// This is a problem for testing things like environment variables, which
+    /// are not thread-safe. This lock ensures that these tests run serially.
+    static TEST_PARALLELISM_LOCK: OnceLock<Mutex<bool>> = OnceLock::new();
```
NicholasLYang, line 372: This isn't a bad solution per se, but just noting that when we had this bug previously,  we created `TestCase` structs that we [manually looped over](https://github.com/vercel/turborepo/blob/2e3f7e730a921666b49a5d561b702a8353b2f86e/crates/turborepo-ci/src/lib.rs#L125).

NicholasLYang (changes requested):

crates/turborepo-scm/src/git.rs
```diff
@@ -81,6 +85,45 @@ impl SCM {
...
+}
+
+impl GitHubEvent {
+    fn get_first_commit_id(&self) -> Option<String> {
+        if self.commits.is_empty() {
+            // commits can be empty when you push a branch with no commits
+            return None;
+        }
+
+        if self.commits.len() > 2048 {
```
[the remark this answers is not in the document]
dimitropoulos: good catch, upon reviewing the code this was ported from, this is correct

crates/turborepo-scm/src/git.rs
```diff
@@ -94,11 +137,72 @@ impl Git {
...
+            }
+        }
+
+        // we must be in a push event
+        // try reading from the GITHUB_EVENT_PATH file
+        if let Ok(event_path) = env::var("GITHUB_EVENT_PATH") {
+            // Try to open the event file and read the contents
+            let mut file = File::open(event_path).ok()?;
+            let mut data = String::new();
+            file.read_to_string(&mut data).ok()?;
```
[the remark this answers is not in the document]
dimitropoulos: good call: 7c156f5244dff743c701cd35e2378d72b6ea0f20 (needed to add the `.ok()?`)

crates/turborepo-scm/src/git.rs
dimitropoulos: oh sweet, great find! yeah, I was using `?` a lot on the base function where this code was extracted from (`get_github_base_ref`).

crates/turborepo-scm/src/git.rs
dimitropoulos: I went in circles and circles on this.  I'm not sure what the problem is, but I couldn't find anything else that could work.  Do you know how this could be rewritten, concretely, that would work.

I couldn't believe this was hard because it seemed like it should be easy, but there's something I'm missing.

dimitropoulos: update: found a way to do it 4d138792ffc65370a9af9c75a8891d35503d716f like you suggested

crates/turborepo-scm/src/git.rs
dimitropoulos: great suggestion! didn't know about that crate - https://github.com/vercel/turborepo/pull/9169/commits/da8112e4d1667e9af99164dc34b4533b5a09923b I also added a predicate helper :)

crates/turborepo-scm/src/git.rs
dimitropoulos: this ended up getting removed in the midst of hoisting the env beyond the functions in question.

crates/turborepo-scm/src/git.rs
```diff
@@ -81,6 +90,71 @@ impl SCM {
...
+struct GitHubEvent {
+    #[serde(default)]
+    before: String,
+
+    #[serde(default)]
+    commits: Vec<GitHubCommit>,
+}
+
+impl GitHubEvent {
+    fn get_first_commit_id(&self) -> Option<String> {
```
tknickman, line 110: Nitpick, but naming is a little misleading here. It's returning `first_commit^` which is really a reference to the _parent_ of the first commit, not the first commit itself.

crates/turborepo-scm/src/git.rs
```diff
@@ -94,19 +168,84 @@ impl Git {
...
+
+            // Parse the JSON data from the file
+            let json: GitHubEvent = serde_json::from_str(&data).ok()?;
+
+            // Extract the base ref from the pull request event if available
+            let base_ref = &json.before;
+
+            // the base_ref will be UNKNOWN_SHA on first push to a new branch and force
+            // pushes
+            if base_ref == UNKNOWN_SHA {
```
tknickman, line 208: We need to also check here if `forced` is `true`. If that is the case, the before SHA is unreliable, and we have to use the same logic as if the before was unknown.

crates/turborepo-scm/src/git.rs
```diff
@@ -94,19 +168,84 @@ impl Git {
...
+        None
+    }
+
+    fn resolve_base(&self, base_override: Option<&str>, env: CIEnv) -> Result<String, Error> {
         if let Some(valid_from) = base_override {
-            return Ok(valid_from);
+            return Ok(valid_from.to_string());
+        }
+
+        if let Some(github_base_ref) = Self::get_github_base_ref(env) {
```
tknickman, line 226: Let's add some github friendly debug output here to help make it more clear what we are using for the comparison

crates/turborepo-scm/src/git.rs
dimitropoulos: thanks! updated in https://github.com/vercel/turborepo/pull/9169/commits/88e4a498e76adf914f0edae2eb6badabd5ded509

crates/turborepo-scm/src/git.rs
dimitropoulos: oh gotcha, I misunderstood - I'd thought the 000000 hash happens on force pushes too.  thanks: fixed in https://github.com/vercel/turborepo/pull/9169/commits/7d1562fa1fb84f18b8b72910b690e3c4f3cdea00

chris-olszewski (approved):
Aside from adding logging when the github base ref doesn't resolve, LGTM.

crates/turborepo-scm/src/git.rs
dimitropoulos: added in https://github.com/vercel/turborepo/pull/9169/commits/2bf8ff5a946f676b0625407cd0daa99db9cdffdd

tknickman (approved):
Awesome, thanks!

NicholasLYang (approved):

PR #9174, opened 2024-09-20
feat(query): map file to package

NicholasLYang:
### Description

Adding some queries so you can go from a file to the package it's from, or even the packages that it affects.

### DIFF
--- a/crates/turborepo-lib/src/query/file.rs
+++ b/crates/turborepo-lib/src/query/file.rs
@@ -82,7 +82,7 @@ impl File {
     }
 
     // This is `Option` because the file may not be in the repo
-    async fn repo_relative_path(&self) -> Option<String> {
+    async fn path(&self) -> Option<String> {
         self.run
             .repo_root()
             .anchor(&self.path)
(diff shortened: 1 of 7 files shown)


crates/turborepo-lib/src/query/file.rs
```diff
@@ -18,26 +23,119 @@ impl File {
     }
 }
 
+#[derive(Clone, Debug, SimpleObject)]
+struct All {
+    count: usize,
+}
+
+#[derive(Union)]
+enum PackageMapping {
```
NicholasLYang, line 32: Union types in GraphQL are pretty annoying. Maybe I should change this to just return all the packages? I'm a little worried that could be too much data.

crates/turborepo-lib/src/query/file.rs
```diff
@@ -81,23 +85,118 @@ impl TraceResult {
...
+    }
+
+    /// Gets the affected packages for the file, i.e. all packages that depend
+    /// on the file.
+    async fn affected_packages(&self) -> Result<Array<Package>, Error> {
+        match self.get_package() {
+            Ok(Some(PackageMapping::All(_))) => Ok(self
+                .run
+                .pkg_dep_graph()
+                .packages()
```
chris-olszewski, line 172: Does this need to be sorted?

chris-olszewski (approved):

tknickman:
@NicholasLYang do we still need this one / can you clean up the conflicts?

PR #9181, opened 2024-09-24
docs: clarification on Transit Nodes docs

maschwenk:
### Description

I'm confused why this example starts off talking about linting and then the example it gives is called `check-types` and sounds very much like typechecking.

Unsure which one is right but it feels like it'd be better to be consistent?

### Testing Instructions

Use your eyes, your intellect, and your heart


docs/repo-docs/crafting-your-repository/configuring-tasks.mdx
```diff
@@ -326,30 +326,30 @@ Some tasks should always be ran no matter what, like a deployment script after a
 
 Some tasks can be ran in parallel despite being dependent on other packages. An example of tasks that fit this description are linters, since a linter doesn't need to wait for outputs in dependencies to run successfully.
```
maschwenk, line 327: Alternatively you could change the above example to be typechecking instead, but based on this dissonance I'm actually a little confused by your example. Using Internal Packages™

> since a type checker doesn't need to wait for outputs in dependencies to run successfully

could be true. Not using Internal Packages, where the output is `.d.ts` files, it would seem like you definetly can't typecheck package A if package B could be changing `.d.ts` files.

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 1371610db05d2797ca3795bc8e8647551f980ff6

docs/repo-docs/crafting-your-repository/configuring-tasks.mdx
anthonyshew, line 327:
```suggestion
Some tasks can be ran in parallel despite being dependent on other packages. An example of tasks that fit this description are linters or type checkers, since a these tasks doesn't need to wait for outputs in dependencies to run successfully.
```

_Technically_, TypeScript as a type checker is a linter. But I can see how that would be confusing. Would this change be clearer?

maschwenk: Yeah that works!

But I guess it feels like a more confusing example given my comment above:

> When NOT using Internal Packages, where the output of each package is `.d.ts` files, it would seem like you definetly can't start typechecking package A which depends on B, until B emits its `.d.ts` files

anthonyshew: Sorry, missed that. You're right, let's swap the example to being `lint`.

docs/site/content/docs/crafting-your-repository/configuring-tasks.mdx
```diff
@@ -344,7 +344,7 @@ Some tasks should always be run no matter what, like a deployment script after a
 
 ### Dependent tasks that can be run in parallel
 
-Some tasks can be run in parallel despite being dependent on other packages. An example of tasks that fit this description are linters, since a linter doesn't need to wait for outputs in dependencies to run successfully.
+Some tasks can be ran in parallel despite being dependent on other packages. An example of tasks that fit this description are type checkers, since a type checker doesn't need to wait for outputs in dependencies to run successfully.
```
anthonyshew:
```suggestion
Some tasks can be run in parallel despite being dependent on other packages. An example of tasks that fit this description are type checkers, since a type checker doesn't need to wait for outputs in dependencies to run successfully.
```

anthonyshew (approved):

PR #9185, opened 2024-09-26
docs: Task outputs clarifications

boutchitos:
To avoid the impression that cache is disabled when Task `outputs` key is ommitted.

`API reference` was more accurate for that behavior, than `Crafting your repository`.

### Description

I'm new to Turborepo, and was reading [Crafting your repository](https://turbo.build/repo/docs/crafting-your-repository). Two sections lead me to conclude that when there is no `outputs` key in a Task, then caching is disabled. I think the wording in `API reference` is better.

We have to account for Logs output even if there are no file outputs (which is really nice for linters).

### Testing Instructions

* Review the modified documentations.
* Take into account that I'm not a native english speaker (you probably notice already ;) )

### DIFF
--- a/docs/repo-docs/crafting-your-repository/caching.mdx
+++ b/docs/repo-docs/crafting-your-repository/caching.mdx
@@ -127,6 +127,8 @@ Turborepo caches two types of outputs: Task outputs and Logs.
 
 Turborepo caches the file outputs of a task that are defined in [the `outputs` key](/repo/docs/reference/configuration#outputs) of `turbo.json`. When there's a cache hit, Turborepo will restore the files from the cache.
 
+The `outputs` key is optional, see [the API reference](/repo/docs/reference/configuration#outputs) for how Turborepo behaves in this case.
+
 <Callout type="warn" title="Providing file outputs">
 If you do not declare file outputs for a task, Turborepo will not cache them. This might be okay for some tasks (like linters) - but many tasks produce files that you will want to be cached.
 
--- a/docs/repo-docs/crafting-your-repository/caching.mdx
+++ b/docs/repo-docs/crafting-your-repository/caching.mdx
@@ -121,6 +121,8 @@ For information on how to connect your CI machines to Remote Cache, visit [the C
 
 ## What gets cached?
 
+Turborepo caches two types of outputs: Task outputs and Logs.
+
 ### Task outputs
 
 Turborepo caches the file outputs of a task that are defined in [the `outputs` key](/repo/docs/reference/configuration#outputs) of `turbo.json`. When there's a cache hit, Turborepo will restore the files from the cache.
--- a/docs/repo-docs/crafting-your-repository/configuring-tasks.mdx
+++ b/docs/repo-docs/crafting-your-repository/configuring-tasks.mdx
@@ -172,7 +172,7 @@ Some tasks may not have any dependencies. For example, a task for finding typos 
   tasks are properly configured first.
 </Callout>
 
-The `outputs` key tells Turborepo **files and directories** it should cache when the task has successfully completed. **Without this key defined, Turborepo will not cache any files. Hitting cache on subsequent runs will not restore any file outputs.**
+The `outputs` key tells Turborepo **files and directories** it should cache when the task has successfully completed.
 
 Below are a few examples of outputs for common tools:
 
@@ -214,6 +214,8 @@ Below are a few examples of outputs for common tools:
 </Tabs>
 
 For more on building globbing patterns for the `outputs` key, see [the globbing specification](/repo/docs/reference/globs).
+
+The `outputs` key is optional, see [the API reference](/repo/docs/reference/configuration#outputs) for how Turborepo behaves in this case.
 
 ### Specifying `inputs`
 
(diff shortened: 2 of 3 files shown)


docs/repo-docs/crafting-your-repository/caching.mdx
```diff
@@ -121,6 +121,8 @@ For information on how to connect your CI machines to Remote Cache, visit [the C
 
 ## What gets cached?
 
+Turborepo caches two types of outputs: Task outputs and Logs.
```
boutchitos, line 124: I think that saying upfront there are two kinds of outputs helps linking them together. Both sections (Task outputs and Logs) are at the same level. In my mind, they were disconnected. My mental flow was:

from Task outputs
> If you do not declare file outputs for a task, Turborepo will not cache them.

So I thought _cache is disabled_. Then I read Logs, but I putted it in the context of _when cache is enabled_.

To better understand my confusion, you have to take into account that I was reading [Crafting your repository](https://turbo.build/repo/docs/crafting-your-repository) from top to bottom (the whole section). So, I read [Configuring tasks](https://turbo.build/repo/docs/crafting-your-repository/configuring-tasks#specifying-outputs) first.

In the `Specifying outputs` passage, there is no mention of "Logs". This lead me to think caching is disabled when `outputs` key is missing.

Later, by experimenting with Turborepo, I did not understand why my Lint task benefits from caching since I did not put `outputs` (because there is none from linters). And having Logs in the cache is very great for that matter!!!

docs/repo-docs/crafting-your-repository/configuring-tasks.mdx
```diff
@@ -215,6 +215,8 @@ Below are a few examples of outputs for common tools:
 
 For more on building globbing patterns for the `outputs` key, see [the globbing specification](/repo/docs/reference/globs).
 
+The `outputs` key is optional, see [the API reference](/repo/docs/reference/configuration#outputs) for how Turborepo behaves in this case.
```
boutchitos, line 218: I prefer redirecting the reader to the accurate reference, instead of explaining here. This avoid duplication, and confusion for when docs are not aligned, or doesn't tell the whole story.

docs/repo-docs/reference/configuration.mdx
```diff
@@ -339,8 +339,6 @@ An allowlist of environment variables that should be made available to this task
 
 A list of file glob patterns relative to the package's `package.json` to cache when the task is successfully completed.
 
-Omitting this key or passing an empty array tells `turbo` to cache nothing (except logs, which are always cached when caching is enabled).
```
boutchitos, line 342: I move it down to add an example. So glob patterns is explained with an example, and then omitted key is explained with an example.

docs/repo-docs/crafting-your-repository/configuring-tasks.mdx
```diff
@@ -172,7 +172,7 @@ Some tasks may not have any dependencies. For example, a task for finding typos
   tasks are properly configured first.
 </Callout>
 
-The `outputs` key tells Turborepo **files and directories** it should cache when the task has successfully completed. **Without this key defined, Turborepo will not cache any files. Hitting cache on subsequent runs will not restore any file outputs.**
+The `outputs` key tells Turborepo **files and directories** it should cache when the task has successfully completed.
```
anthonyshew, line 175: We need to keep this information. It's important that folks know that no artifacts will be cached.

docs/repo-docs/reference/configuration.mdx
anthonyshew: Let's move this down below the existing example but not add another example. I trust that developers won't need an example of how to not write a key. 😄

docs/repo-docs/crafting-your-repository/caching.mdx
anthonyshew: This is a good addition. It also seems like the suggestion from `docs/repo-docs/crafting-your-repository/configuring-tasks.mdx#218` would work better in the `Task Outputs` section below. What do you think?

docs/repo-docs/crafting-your-repository/configuring-tasks.mdx
anthonyshew: The reference and the how-to (Crafting your Repository) serve different purposes. The reference is terse, rote, and dry, while the how-to is more introductory, and caters to a different type of a reader.

This is good information to add, though it might be better in a different place. See https://github.com/vercel/turborepo/pull/9185/files#r1776190519

anthonyshew (changes requested):
Thanks for these updates! I see how this could be clearer now that you're pointing it out and think we can keep improving on your initial draft. Check out the comments and suggestions I've provided and let me know what you think.

docs/repo-docs/crafting-your-repository/caching.mdx
boutchitos: I'll do what you suggest. Now that I understand, I can't unlearn and figure out what would have been my train of thoughts on that. Let's wait the next newbie :)

boutchitos:
@anthonyshew: I applied your suggestions!

anthonyshew (approved):
Looks great! Thank you for contributing back!

PR #9187, opened 2024-09-26
Update single-package-workspaces.mdx

svenheden:
### Description

Replace an `is` with an `isn't`


anthonyshew (approved):
Good catch. Thanks!

PR #9189, opened 2024-09-26
base.json is extended by other tsconfig.json, not package.json

boutchitos:
### Description

In [Use a base tsconfig file](https://turbo.build/repo/docs/guides/tools/typescript#use-a-base-tsconfig-file), it is state that `base.json`, a TypeScript config file, is extended by every other `package.json`. I guess we should say extended by every other `tsconfig.json` in the workspace.


anthonyshew (approved):
Good catch! Thanks!

PR #9194, opened 2024-09-26
docs: grammar error

dimitropoulos:
### Description

Just happened to notice this grammatical error when reading the docs: "from the first time the task was run" and "the first time the task ran" are both good, but "was ran" is not grammatical with the "was" auxiliary.


anthonyshew (approved):
Good catch.

PR #9196, opened 2024-09-27
feat(query): add `currentRun` field to `turbo query`

NicholasLYang:
### Description

We need the `currentRun` field to exist on `turbo query` to easily generate types from the GraphQL server.  This adds the field and also deduplicates the server code so both `wui` and `query` use the same `run_server` code


tknickman (approved):
Change looks good, a few CI errors

PR #9212, opened 2024-10-03
 (docs) Consistency for configurations between docs and examples.

anthonyshew:
### Description

@boutchitos astutely pointed out in https://github.com/vercel/turborepo/issues/9191 that we had some inconsistencies between `create-turbo`, the TypeScript doc, and `with-vite`. This PR fixes those issues.


boutchitos:
Awesome!

PR #9219, opened 2024-10-03
chore(examples): minor adjustments to `with-shell-commands`

breadadams:
### Description

- Rename the app directories from `apps/apps-*` → `apps/app-*`.
- Fix a typo

### Testing Instructions

N/A


anthonyshew (commented):
I guess it does make more sense that way! 😄

Just checking, is there a pnpm lockfile update needed as well? Usually, when you rename directories, you'll need to fix the lockfile. Was that not the case here?

breadadams:
Ah good point, missed that! Thanks @anthonyshew, will update the PR Monday - travelling currently.

anthonyshew (approved):
Thanks for the additional fixes!

PR #9222, opened 2024-10-04
docs: document creating packages with framework bindings

anthonyshew:
### Description

We've had users ask about how to create bindings in their packages, so documenting it here.

### Testing Instructions

👀


docs/repo-docs/guides/frameworks/framework-bindings.mdx
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+```
+
+<Callout type="good-to-know">
+  In the example above, the `peerDependency` for `next` accepts any version. You
+  may want to specify a range (for example, `">=15"`) according to your needs.
+</Callout>
+
+This will make the dependency available in your library, allowing you to write code like below. Note the `className` prop, which sets a default styling for this component in the monorepo and can be overridden in the `props` object.
```
tknickman, line 36: We should add a note that older package managers (pnpm 7, npm v4-v6, etc.) will not install peers by default. Sometimes this is worked around by _also_ adding this dep to devDeps, or enabling the right "install peers" option of your package manager.

docs/repo-docs/guides/frameworks/framework-bindings.mdx
```diff
@@ -0,0 +1,81 @@
...
+The version of `next` that will be resolved for the package will come from the consumers of the library. For example, if Next.js 15 is installed in your applications, the TypeScript types and APIs for `next` will also be Next.js 15.
+
+## Splitting framework bindings using entrypoints
+
+Using export paths to split a package into framework-specific entrypoints is the simplest way to add bindings to a library that aims to support multiple frameworks. By splitting entrypoints, bundlers have an easier time understanding the framework you intend to target and you're less likely to see strange bundling errors.
+
+The example below shows a library with two entrypoints, each for a different type of link component. These abstractions likely contain your own styles, APIs, and other adjustments on top of the element they're wrapping.
+
+- `./link`: An `<a>` HTML tag with some default styles from your design system
+- `./next-js/link`: A customized version of [the Next.js `Link` component](https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#link-component) with props that are preset to your organization's preferences
```
tknickman, line 62: This would sink in more for me if we added another framework:

```
- `./link`: ...
- `./next-js/link`: ...
- `./svelte/link`: ...
```

tknickman (approved):
A few suggestions, but nothing blocking!

docs/repo-docs/guides/frameworks/framework-bindings.mdx
```diff
@@ -0,0 +1,81 @@
...
+  "name": "@repo/ui",
+  "peerDependencies": {
+    "next": "*"
+  }
+}
+```
+
+<Callout type="good-to-know">
+  In the example above, the `peerDependency` for `next` accepts any version. You
+  may want to specify a range (for example, `">=15"`) according to your needs.
```
anthonyshew, line 33:
```suggestion
  may want to specify a range (for example, `">=15"`) according to your needs.
  
  Additionally, for older package managers, you may need to instruct your package manager to install peer dependencies with configuration, or add the dependency to `devDependencies` as a workaround.
```

docs/repo-docs/guides/frameworks/framework-bindings.mdx
```diff
@@ -0,0 +1,83 @@
...
+The version of `next` that will be resolved for the package will come from the consumers of the library. For example, if Next.js 15 is installed in your applications, the TypeScript types and APIs for `next` will also be Next.js 15.
+
+## Splitting framework bindings using entrypoints
+
+Using export paths to split a package into framework-specific entrypoints is the simplest way to add bindings to a library that aims to support multiple frameworks. By splitting entrypoints, bundlers have an easier time understanding the framework you intend to target and you're less likely to see strange bundling errors.
+
+The example below shows a library with two entrypoints, each for a different type of link component. These abstractions likely contain your own styles, APIs, and other adjustments on top of the element they're wrapping.
+
+- `./link`: An `<a>` HTML tag with some default styles from your design system
+- `./next-js/link`: A customized version of [the Next.js `Link` component](https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#link-component) with props that are preset to your organization's preferences
```
anthonyshew, line 64:
```suggestion
- `./next-js/link`: A customized version of [the Next.js `Link` component](https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#link-component) with props that are preset to your organization's preferences
- `./svelte/link`: A customized version of an [`a` tag for Svelte](https://svelte.dev/docs/kit/link-options) with presets.
```

PR #9225, opened 2024-10-07
chore: Remove `display` key from generated `tsconfig`

Sensanaty:
### Description

https://github.com/vercel/turborepo/discussions/9221

I noticed that the default `create-turbo` command would create `tsconfig`s that contained a `display` key, which isn't part of the TS schema spec.

I opened a discussion to ask about this, because it might've been a Turbo-specific feature, for example I thought it was maybe used to display the list of projects in the repo in a nice human-readable way, but according to Anthony it's not actually used for anything.

From a quick read through of the `create-turbo` code, what I gathered is that it basically just copies these files based on the hardcoded example repos, hence why I only modified the `tsconfig`s themselves. It's possible I missed something as this is my first time interacting with this codebase (and the word `display` is generic enough that it shows up in a billion places all over the code), so please do let me know if my assumption here was wrong.

### Testing Instructions

1. Create a new turborepo with `create-turbo`
2. The example `tsconfig`s should no longer contain the `display` key.

### DIFF
--- a/examples/basic/packages/typescript-config/base.json
+++ b/examples/basic/packages/typescript-config/base.json
@@ -1,6 +1,5 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
-  "display": "Default",
   "compilerOptions": {
     "declaration": true,
     "declarationMap": true,
--- a/examples/basic/packages/typescript-config/nextjs.json
+++ b/examples/basic/packages/typescript-config/nextjs.json
@@ -1,6 +1,5 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
-  "display": "Next.js",
   "extends": "./base.json",
   "compilerOptions": {
     "plugins": [{ "name": "next" }],
--- a/examples/basic/packages/typescript-config/react-library.json
+++ b/examples/basic/packages/typescript-config/react-library.json
@@ -1,6 +1,5 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
-  "display": "React Library",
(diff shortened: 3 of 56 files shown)


anthonyshew:
As a mental note, I just verified that TypeScript doesn't verify extra keys at the top-level of `tsconfig.json`. I added a `someRandomTestKey` and everything passed successfully. However, adding a test key _inside of `compilerOptions`_ does indeed error out.

That for me helps to confirm that this was [in the initial template from 3 years ago](https://github.com/vercel/turborepo/blame/main/examples/basic/packages/typescript-config/base.json), kept getting copy-pasted and, since it never hurt anything, no one questioned it.

anthonyshew (commented):
Thanks! Looks right to me.

Out of an abundance of caution, I've asked the rest of the team if anyone has any secret sage wisdom about why this key might be in these, just in case there's something I'm not realizing. I don't think this will be the case, but I'll come back and merge later today assuming I don't hear any objections.

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 8f9ac5848756854fca45944bb3a2d60d8d0421bb

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 8f9ac5848756854fca45944bb3a2d60d8d0421bb

anthonyshew (approved):
Talked to colleagues and they're good with this, too!

The CI failures here are not related to these changes. I will fix in a separate PR.

Thank you for your contribution! 🚀

PR #9228, opened 2024-10-07
feat(watch): add interruptible tasks

### LINKED ISSUE #8164: Watch-mode not rebuild dependent packages when package changes
### Verify canary release

n/a

pnpm

Mac

1.13.4-canary.4

### Describe the Bug

If you make a change to a package that an project is depending on the project does not get rebuild

### Expected Behavior

I would expect that the project gets rebuild when the package this project depends on get changed

### To Reproduce

If you have a `project1` that depends on `package1`, and `package2`. 
The `project1` has a script defined in its `package.json` named `dev` and the `package1` and `package2` workspace packages have script defined named `build`.

The `dev` in `turbo.json` is defined as:
```json
{
  "pipeline": {
    "build": {
      "dependsOn": [
        "^build"
      ],
      "inputs": [
        "src/**",
        "public/**",
        "test/**",
        "migration/**",
        "patches/**",
        "mocks/**",
        "__mocks__/**",
        // Required to ignore .DS_Store files
        "!dist/**",
        "!lib/**",
        "!coverages/**",
        "!**/.DS_Store"
      ],
      "outputs": [
        "dist/**/*",
        "build/**/*",
        "lib/**/*",
        "typings/**/*",
        ".next/**/*",
        "!.next/cache/**"
      ],
      "outputMode": "new-only"
    },
    "dev": {
      "dependsOn": [
        "^build"
      ],
      "cache": false,
      "persistent": true
    }
  }
}
```

If I run the command `turbo watch dev --filter=project1` all the packages get build and the `dev`-script of `project1` gets started. Now if I make a change to `package1` e.g. `src/index.ts`. The Turborepo daemon detects the following:

```bash
2024-05-16T22:03:07.792779Z  WARN turborepo_lib::package_changes_watcher: changed_files: {AnchoredSystemPathBuf("packages/logging/src/ConsoleStructuredLogger.ts")}
2024-05-16T22:03:07.792801Z  WARN turborepo_lib::package_changes_watcher: changed_packages: Ok(Some({WorkspacePackage { name: Other("@company/logging"), path: AnchoredSystemPathBuf("packages/logging") }}))
```

So it appears that Turborepo has picked up that the package has changed but it doesn't seem to trigger `build` on the package and then after that completes re-runs the `dev`-script of `project1`.

Have I misconfigured my project?

### Additional context

_No response_

NicholasLYang:
### Description

Adds the ability to annotate persistent tasks as interruptible. Fixes #8164

### DIFF
--- a/crates/turborepo-lib/src/engine/mod.rs
+++ b/crates/turborepo-lib/src/engine/mod.rs
@@ -45,7 +45,7 @@ pub struct Engine<S = Built> {
     task_definitions: HashMap<TaskId<'static>, TaskDefinition>,
     task_locations: HashMap<TaskId<'static>, Spanned<()>>,
     package_tasks: HashMap<PackageName, Vec<petgraph::graph::NodeIndex>>,
-    pub(crate) has_persistent_tasks: bool,
+    pub(crate) has_non_interruptible_tasks: bool,
 }
 
 impl Engine<Building> {
@@ -60,7 +60,7 @@ impl Engine<Building> {
             task_definitions: HashMap::default(),
             task_locations: HashMap::default(),
             package_tasks: HashMap::default(),
-            has_persistent_tasks: false,
+            has_non_interruptible_tasks: false,
         }
     }
 
@@ -87,8 +87,8 @@ impl Engine<Building> {
         task_id: TaskId<'static>,
         definition: TaskDefinition,
     ) -> Option<TaskDefinition> {
-        if definition.persistent {
-            self.has_persistent_tasks = true;
+        if definition.persistent && !definition.interruptible {
+            self.has_non_interruptible_tasks = true;
         }
         self.task_definitions.insert(task_id, definition)
     }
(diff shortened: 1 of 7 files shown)


weyert:
> NOTE: This does create change in behavior for watch mode where new changes will stop the current executing and start a new one. Previously, we waited for runs to finish, but now that persistent tasks can be interrupted, we can't do that since persistent tasks don't finish.

But only when `interruptible` is set to `true` otherwise the previous behaviour occurs?

NicholasLYang:
No, the change is true for all tasks. I don't want to make too many special cases for watch mode execution

chris-olszewski (commented):
I believe this now allows for persistent tasks to only run after their deps run which has also been an ask (https://github.com/vercel/turborepo/issues/8673). Should we advertise this/consider this when naming the config option.

chris-olszewski (approved):
This is awesome!

If we ever enable caching during `watch` I believe we need to verify that `cache: false` on interruptible tasks, but since that isn't possible now it's fine.

JCMais:
looking forward to a release with this feature, when should we expect it being released? 👀

JCMais:
Hmm, looks like the JSON schema was not updated with the new field added here.

PR #9236, opened 2024-10-08
feat: warn on empty cache

NicholasLYang:
### Description

Warns if cache payload consists of just the log file and output globs are specified.

### Testing Instructions

Turns out all of our test fixtures have empty outputs so we get a lot of these warnings in tests


tknickman (approved):
Looks great, thanks! 

I think we should also skip the cache save in this case. But we can also address that in a follow up.

PR #9237, opened 2024-10-08
log::debug and ignore on daemon usage in CI

dimitropoulos:
Setting `daemon: true` in `turbo.json` can cause timing out on deploys.  Using the daemon in CI/CD is almost* never the desired behavior.  Instead, we will detect a CI environment and force the daemon off in these situations.

<sub>*perhaps the only use-case is for us, internally, testing the daemon in our internal tests (which, run on the CI).</sub>

### DIFF
--- a/crates/turborepo-lib/src/config/mod.rs
+++ b/crates/turborepo-lib/src/config/mod.rs
@@ -14,7 +14,7 @@ use miette::{Diagnostic, NamedSource, SourceSpan};
 use serde::Deserialize;
 use struct_iterable::Iterable;
 use thiserror::Error;
-use tracing::warn;
+use tracing::debug;
 use turbo_json::TurboJsonReader;
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
 use turborepo_errors::TURBO_SITE;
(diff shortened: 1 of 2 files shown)


docs/repo-docs/reference/configuration.mdx
```diff
@@ -141,6 +141,12 @@ Turborepo runs a background process to pre-calculate some expensive operations.
 }
 ```
 
+  Note that when running in a CI environment the daemon is disabled, even if you
+  set this value to `true`. This is because the daemon is not useful in a CI
+  environment and can cause timing out on deploys.
```
tknickman, line 147:
```suggestion
  When running in a CI environment the daemon is always disabled regardless of this setting.
```

crates/turborepo-lib/src/config/mod.rs
```diff
@@ -329,6 +330,15 @@ impl ConfigurationOptions {
     }
 
     pub fn daemon(&self) -> Option<bool> {
+        // hardcode to off in CI
+        if turborepo_ci::is_ci() {
+            if Some(true) == self.daemon {
+                warn!("Ignoring daemon setting and disabling the daemon because we're in CI");
```
tknickman, line 336: I think we might want to make this debug instead. If you have `daemon: true` in your config, you will ALWAYS get this warning in CI with no way to turn it off.

tknickman (approved):

PR #9238, opened 2024-10-09
fix napi double .d.ts formatting

dimitropoulos:
### Description

`napi-rs` generates a .d.ts file for us in `turbo-repository`.  Unfortunately, this file doesn't match our project's formatting (the generated file doesn't include semicolons).  This means that every time we locally run a command (for example, `pnpm test`) that courses through this generation we jiggle this file, creating a lot of noise when developing.

This small PR fixes this by simplifying the script and running prettier on it.

### Testing Instructions

run `pnpm build` from `packages/turbo-repository` before this PR and see the file change, and notice that after this PR the file change goes away.

### DIFF
--- a/packages/turbo-repository/package.json
+++ b/packages/turbo-repository/package.json
@@ -5,8 +5,8 @@
   "bugs": "https://github.com/vercel/turborepo/issues",
   "homepage": "https://turbo.build/repo",
   "scripts": {
-    "build": "napi build --platform -p turborepo-napi --cargo-cwd ../../ --cargo-name turborepo_napi native --js false --dts ../js/index.d.ts && mkdir -p js/dist && cp js/index.js js/dist/index.js && cp js/index.d.ts js/dist/index.d.ts",
-    "build:release": "napi build --release --platform -p turborepo-napi --cargo-cwd ../../ --cargo-name turborepo_napi native --js false",
+    "build": "bash scripts/build.sh",
+    "build:release": "bash scripts/build.sh release",
     "package": "node scripts/publish.mjs",
     "test": "node --import tsx --test __tests__/*.test.ts"
   },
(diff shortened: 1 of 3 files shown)


packages/turbo-repository/scripts/build.sh
```diff
@@ -0,0 +1,25 @@
...
+  flags+=" --release"
+else
+  flags+=" --dts ../js/index.d.ts"
+fi
+
+node_modules/.bin/napi build $flags
+
+# Unfortunately, when napi generates a .d.ts file, it doesn't match our formatting rules (it doesn't have semicolons).
+# Since there's now way to configure this from napi itself, so we need to run prettier on it after generating it.
+node_modules/.bin/prettier --write js/index.d.ts
```
dimitropoulos, line 22: we could really min/max this by switching the script around a bit and only running prettier when not in release mode (we don't generate the file in release mode for some reason), but I thought it'd be simpler to leave it as-is.

packages/turbo-repository/package.json
```diff
@@ -5,8 +5,8 @@
   "bugs": "https://github.com/vercel/turborepo/issues",
   "homepage": "https://turbo.build/repo",
   "scripts": {
-    "build": "napi build --platform -p turborepo-napi --cargo-cwd ../../ --cargo-name turborepo_napi native --js false --dts ../js/index.d.ts && mkdir -p js/dist && cp js/index.js js/dist/index.js && cp js/index.d.ts js/dist/index.d.ts",
-    "build:release": "napi build --release --platform -p turborepo-napi --cargo-cwd ../../ --cargo-name turborepo_napi native --js false",
```
dimitropoulos, line 9: was it intentional that we, previously, didn't generate the dts file when in release mode?

chris-olszewski: Probably not, the release process for this package is far more ad-hoc than the primary `turbo` release workflow.

chris-olszewski (approved):

PR #9239, opened 2024-10-09
fix broken examples causing `pnpm test` to fail

dimitropoulos:
### Description

On main, `pnpm test` fails (sometimes intermittently, depending on cache scenarios, but all the time without cache).  Why?  Well..

Three of the examples had version 9 lockfiles checked in to git, when version 6 is what's required by their `packageManager` fields.  This causes tests to fail quite frequently ([example](https://github.com/vercel/turborepo/actions/runs/11238389716/job/31243099738)) in many PRs.  This happens because when installing in these repos you get `Lockfile not compatible with current pnpm` errors, and it totally blows away the old lockfile and grabs a new one.  Then, since our Next.js version is specified as `^4.1.1` in these package.jsons, it will resolve to `4.2.15` instead of the old `4.1.1`.  Then, it turns out that this minor update causes this diff in the `next-env.d.ts` file:

```diff
/// <reference types="next" />
/// <reference types="next/image-types/global" />

// NOTE: This file should not be edited
- // see https://nextjs.org/docs/basic-features/typescript for more information.
+ // see https://nextjs.org/docs/app/building-your-application/configuring/typescript for more information.
```

The fact of that diff happening causes turbo to not be "FULL TURBO", which the example tests checks for.  That then causes the test suite to fail because there's a turbo cache miss.

### Testing Instructions

`pnpm test` should now pass.

### DIFF
--- a/turborepo-tests/helpers/setup_example_test.sh
+++ b/turborepo-tests/helpers/setup_example_test.sh
@@ -74,7 +74,7 @@ $turbo_command >./tmp/run-2.txt
 
 # Make sure the user hits FULL TURBO on the second go
 if ! grep -q ">>> FULL TURBO" ./tmp/run-2.txt; then
-  echo "[ERROR] No 'FULL TURBO' was found.  That shows that the cache was not fully hit on the second run, which was expected behavior."
+  echo "[ERROR] No 'FULL TURBO' was found.  This indicateds that at least one 'cache miss' occurred on the second run when all tasks were expected to be 'cache hit'."
   echo "Dumping logs:"
   echo ""
   cat ./tmp/run-2.txt >&2
--- a/turborepo-tests/helpers/setup_example_test.sh
+++ b/turborepo-tests/helpers/setup_example_test.sh
@@ -56,23 +56,26 @@ echo "/tmp/" >> ".gitignore"
 
 # Simulating the user's first run and dumping logs to a file
 $package_manager_command >./tmp/install.txt 2>&1
-$turbo_command >./tmp/grep-me-for-miss.txt
+$turbo_command >./tmp/run-1.txt
 
 # We don't want to hit cache on first run because we're acting like a user.
 # A user would never hit cache on first run. Why should we?
-if grep -q ">>> FULL TURBO" ./tmp/grep-me-for-miss.txt; then
-  echo "A FULL TURBO was found. This test is misconfigured (since it can hit a cache)."
+if grep -q ">>> FULL TURBO" ./tmp/run-1.txt; then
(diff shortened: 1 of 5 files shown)


turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -55,24 +55,28 @@ mkdir -p ./tmp
...
 # Simulating the user's second run
-$turbo_command >./tmp/grep-me-for-hit.txt
+$turbo_command >./tmp/run-2.txt
 
 # Make sure the user hits FULL TURBO on the second go
-if ! grep -q ">>> FULL TURBO" ./tmp/grep-me-for-hit.txt; then
-  echo "No FULL TURBO was found. Dumping logs:"
-  cat ./tmp/grep-me-for-hit.txt >&2
+if ! grep -q ">>> FULL TURBO" ./tmp/run-2.txt; then
+  echo "[ERROR] No 'FULL TURBO' was found.  That shows that the cache was not fully hit on the second run, which was expected behavior."
```
tknickman, line 77: At a glance this phrasing reads like the second run cache miss is _expected_. Maybe just re-word to something like:


```suggestion
  echo "[ERROR] No 'FULL TURBO' was found.  That indicates that the second run unexpectedly missed the cache."
```

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -55,24 +55,28 @@ mkdir -p ./tmp
 echo "/tmp/" >> ".gitignore"
 
 # Simulating the user's first run and dumping logs to a file
+$package_manager --version >./tmp/package_manager_version.txt
 $package_manager_command >./tmp/install.txt 2>&1
-$turbo_command >./tmp/grep-me-for-miss.txt
+$turbo_command >./tmp/run-1.txt
```
tknickman, line 60: Expected rename?

tknickman (approved):
Some phrasing suggestions, but looks good!

turborepo-tests/helpers/setup_example_test.sh
dimitropoulos: yes: I was trying to make it clear.  we're running the same command twice so the output for run 1 goes to `run-1.txt` and the output for run 2 goes to `run-2.txt`.  At least for me, this added clarity to understanding what's going on.  The `grep-me` thing was cute, but I found it hard to follow upon trying to understand what's going on here without any other context or asking anyone.

turborepo-tests/helpers/setup_example_test.sh
dimitropoulos: good feedback: updated in 5cd348278bb51744c3afeac32898e6690d34e061 to make it clear that "hitting the cache" isn't the same as seeing a wall full of "cache hit" as long as there's even one "cache miss".

PR #9240, opened 2024-10-09
feat(query): package changes reason

NicholasLYang:
### Description

Plumbing through reason for why a package changed for query. Adds a `reason` field which is a union of the different reasons a package could be added.

This can be reviewed commit by commit, although there is a little bit of churn around types and some behavior.

### Testing Instructions

Added some tests to `affected.t`


turborepo-tests/integration/tests/affected.t
```diff
@@ -60,6 +63,36 @@ Do the same thing with the `query` command
 Remove the new file
   $ rm apps/my-app/new.js
 
+Add a file in `util`
+  $ echo "hello world" > packages/util/new.js
+
+Validate that both `my-app` and `util` are affected
+  $ ${TURBO} query "query { affectedPackages { items { name reason { __typename } } } }"
```
mehulkar, line 70: What's with the `__typename` thing? Is that going to stay? Can it just be `reason` (and the response also not have the value nested?)

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -376,16 +383,16 @@ impl<'a, T: GitChangeDetector> FilterResolver<'a, T> {
...
                     // if we are neither including dependents or dependencies, then
                     // add  to the list of cherry picked packages
-                    cherry_picked_packages.insert(package);
+                    cherry_picked_packages.insert(package, reason);
                 }
             }
         }
 
-        let mut all_packages = HashSet::new();
+        let mut all_packages = HashMap::new();
```
chris-olszewski, line 394: Especially here I feel like it would be useful to indicate if there are multiple reasons a package was in scope and currently the reason is dependent on the ordering of the `extend` calls.

crates/turborepo-lib/src/run/scope/filter.rs
```diff
@@ -99,6 +99,17 @@ impl PackageInference {
...
+fn merge_changed_packages(
+    changed_packages: &mut HashMap<PackageName, PackageChangeReason>,
+    new_changes: impl IntoIterator<Item = (PackageName, PackageChangeReason)>,
+) {
+    for (package, reason) in new_changes {
+        if !changed_packages.contains_key(&package) {
+            changed_packages.insert(package, reason);
+        }
+    }
+}
```
chris-olszewski:
```
fn merge_changed_packages(
    changed_packages: &mut HashMap<PackageName, HashSet<PackageChangeReason>>,
    new_changes: impl IntoIterator<Item = (PackageName, PackageChangeReason)>,
) {
    for (package, reason) in new_changes {
        let mut reasons = changed_packages.entry(package).or_default();
        reasons.insert(reason);
    }
}
```

chris-olszewski (approved):
Approving so we can get people testing this sooner vs later, but I think we'll want to change the reasons in a few ways:
 - Return all of the reasons why a package is in scope
 - Include all matching criteria for some reasons e.g. global file changes should mention all global files that changed and not just the first one
 - Lockfile changes would be a lot more useful if they returned what packages changed instead of lockfile contents.

PR #9248, opened 2024-10-11
chore(shim): add unit tests for shim parsing

chris-olszewski:
### Description

I was looking at hoisting daemon detection up to shim level so we could avoid reloading a filtered tracing layer. This ended up not being the solution I wanted, but I figured I should add the tests that I added.

⚠️ This PR does have an extremely minor behavior change. See [this comment](https://github.com/vercel/turborepo/pull/9248#discussion_r1797232220)

### Testing Instructions

Unit tests pass

### DIFF
--- a/crates/turborepo-lib/src/shim/parser.rs
+++ b/crates/turborepo-lib/src/shim/parser.rs
@@ -271,7 +271,9 @@ impl ShimArgs {
 #[cfg(test)]
 mod test {
     use miette::SourceSpan;
+    use pretty_assertions::assert_eq;
     use test_case::test_case;
+    use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
 
     use super::ShimArgs;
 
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/shim/parser.rs
```diff
@@ -56,6 +56,14 @@ pub struct ShimArgs {
 
 impl ShimArgs {
     pub fn parse() -> Result<Self, Error> {
+        let invocation_dir = AbsoluteSystemPathBuf::cwd()?;
```
chris-olszewski, line 59: We make this call here so testing is easier. It also means we avoid calling `std::env::current_dir` once for each `--cwd` flag and reuse this value. 

The only behavior change from this move is that if you:
 - pass the args `turbo --cwd`
 - Execute `turbo` in a directory where `current_dir` would fail (process doesn't have permission to access cwd or cwd doesn't exist)
Then we will now report the `current_dir` failure instead of complaining that you didn't pass a value for the `--cwd` flag.

NicholasLYang (approved):

PR #9249, opened 2024-10-11
handle VERCEL_ARTIFACTS_* env vars override

dimitropoulos:
### Description

The prior intended state of EnvVars overrides looked like this:
<br />
<div align="center">
  <div><sub><em>`VERCEL_ARTIFACTS_*` always wins</em></sub></div>
  <br />
  
</div>
<br />
<br />

But we've realized this is not good enough.  We want to better handle the scenario where users have provided custom remote cache - even if they deploy on Vercel.

After this PR, the precedence works like this.

1. if both `TURBO_TEAMID`+`TURBO_TOKEN` are set we use those. if `TURBO_TEAM` is also set, we pass that along, too.
1. if both `TURBO_TEAM`+`TURBO_TOKEN` are set we use those.  if `TURBO_TEAMID` is also set we pass that along, too.
1. if both `VERCEL_ARTIFACTS_OWNER`+`VERCEL_ARTIFACTS_TOKEN` are set, we use those (all else is ignored).
1. if both `TURBO_TEAMID`+`TURBO_TEAM`, we pass those along.
1. if just `TURBO_TEAMID` is set, we pass that along.
1. if just `TURBO_TEAM` is set, we pass that along.
1. if both `TURBO_TEAM` and `VERCEL_ARTIFACTS_OWNER` are set, we pass those along
1. if just `VERCEL_ARTIFACTS_OWNER` is set, we pass that along
1. we do not support or consider partial configurations.  For example if only `VERCEL_ARTIFACTS_TOKEN` and `TURBO_TEAMID` are set, this is considered undefined/unsupported behavior.  The reasons are that:
  - this shouldn't ever happen - Vercel sets both or none of the `VERCEL_ARTIFACTS_*` family
  - we tried doing this, but the combinatorial explosion that results is pretty major, and some of the cases are very strange to reason about, considering the above point that we don't expect there to be any genuine scenario where this happens, anyway.

### Testing Instructions

Take a look at the newly added tests in `override_env.rs`.  You could run turbo while manually setting those same env vars.


crates/turborepo-lib/src/config/mod.rs
```diff
@@ -561,22 +566,16 @@ mod test {
...
-        let override_config = ConfigurationOptions {
-            token: Some("unseen".into()),
-            team_id: Some("unseen".into()),
-            ..Default::default()
-        };
-
         let builder = TurborepoConfigBuilder {
             repo_root,
-            override_config,
+            override_config: Default::default(),
```
dimitropoulos, line 571: @chris-olszewski my reading of the code is that after the changes in this PR there's no need for this struct property (`override_config`) at all.  I removed it in a commit, but then later dropped it because I wanted to check with you first.

crates/turbo-trace/src/main.rs
```diff
@@ -30,7 +30,7 @@ fn main() -> Result<(), PathError> {
         .map(|f| AbsoluteSystemPathBuf::from_unknown(&abs_cwd, f))
         .collect();
 
-    let tracer = Tracer::new(abs_cwd, files, args.ts_config)?;
+    let tracer = Tracer::new(abs_cwd, files, args.ts_config);
```
dimitropoulos, line 33: I'm not totally clear on why this change had anything to do with anything in this PR, but it wouldn't compile otherwise.

crates/turborepo-lib/src/config/env.rs
```diff
@@ -63,7 +63,7 @@ impl EnvVars {
 impl ResolvedConfigurationOptions for EnvVars {
     fn get_configuration_options(
         &self,
-        _existing_config: &ConfigurationOptions,
+        _existing_config: Option<&ConfigurationOptions>,
```
dimitropoulos, line 66: This was made an option because in all the places where `ResolvedConfigurationOptions` is used, this argument is only provided once.  That left us constructing and passing `ConfigurationOptions::default()` everywhere, which was hard to follow (considering it's almost never used).

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+
+    use super::*;
+
+    lazy_static! {
+        static ref VERCEL_ARTIFACTS_OWNER: String = String::from("valueof:VERCEL_ARTIFACTS_OWNER");
+        static ref VERCEL_ARTIFACTS_TOKEN: String = String::from("valueof:VERCEL_ARTIFACTS_TOKEN");
+        static ref TURBO_TEAMID: String = String::from("valueof:TURBO_TEAMID");
+        static ref TURBO_TEAM: String = String::from("valueof:TURBO_TEAM");
+        static ref TURBO_TOKEN: String = String::from("valueof:TURBO_TOKEN");
+    }
```
chris-olszewski, line 218: Not blocking, but I think this would be simplified by just using `const` and copying in the various methods
```suggestion
        const VERCEL_ARTIFACTS_OWNER: &str = "valueof:VERCEL_ARTIFACTS_OWNER";
        const VERCEL_ARTIFACTS_TOKEN: &str = "valueof:VERCEL_ARTIFACTS_TOKEN";
        const TURBO_TEAMID: &str = "valueof:TURBO_TEAMID";
        const TURBO_TEAM: &str = "valueof:TURBO_TEAM";
        const TURBO_TOKEN: &str = "valueof:TURBO_TOKEN";
```

The calls to `clone()` would be replaced by `to_owned()` or (`into()`)

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+
+        fn team_slug(mut self, value: String) -> Self {
+            self.output.team_slug = Some(value);
+            self
+        }
+
+        fn token(mut self, value: String) -> Self {
+            self.output.token = Some(value);
+            self
+        }
```
chris-olszewski, line 283: If you follow the above suggestion you can change this to also just take `&str`
```suggestion
        fn token(mut self, value: &str) -> Self {
            self.output.token = Some(value.into());
            self
        }
```

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+                input: Input::new(),
+                output: Output::new(),
+                reason: "missing",
+            }
+        }
+
+        fn reason(mut self, reason: &'static str) -> Self {
+            self.reason = reason;
+            self
+        }
```
chris-olszewski, line 238: Very unimportant since this is test code, but we could make we remember to provide a reason by requiring it in the constructor.
```suggestion
        fn new(reason: &'static str) -> Self {
            Self {
                input: Input::new(),
                output: Output::new(),
                reason,
            }
        }
```

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+            }
+            if let Some(value) = &case.input.TURBO_TOKEN {
+                env.insert("turbo_token".into(), value.into());
+            }
+            if let Some(value) = &case.input.VERCEL_ARTIFACTS_OWNER {
+                env.insert("vercel_artifacts_owner".into(), value.into());
+            }
+            if let Some(value) = &case.input.VERCEL_ARTIFACTS_TOKEN {
+                env.insert("vercel_artifacts_token".into(), value.into());
+            }
```
chris-olszewski, line 483: It would be cool to also make sure that `Input::from` is producing the expected `Input`
```suggestion
            }
            
            let actual_input = Input::from(&env);
            assert_eq!(case.input, actual_input);
```

chris-olszewski (changes requested):
This is an amazing write up of the existing + desired behavior.

The only blocking comment is making sure that we keep the same level of user facing error messaging if we encounter a non-UTF8 value.

I feel that changing the parameter for `get_configuration_options` doesn't provide us with much, but that's just an opinion.

crates/turbo-trace/src/main.rs
dimitropoulos: alrighty. heh.

crates/turborepo-lib/src/config/mod.rs
dimitropoulos: :+1:

crates/turborepo-lib/src/config/env.rs
dimitropoulos: all good - reverted this change in 0e9eaffff540fa0fc5fd9cec65b8f42cea6d9dc1

I just found it confusion.  Options that aren't optional, and all that.  by "provided once" I meant "passed in once".

But all good - I trust your instincts here!

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+// this is an internal structure (that's a partial of ConfigurationOptions) that
+// we use to store
+pub struct Output {
+    /// maps to ConfigurationOptions.team_id
+    pub team_id: Option<String>,
+    // maps to ConfigurationOptions.team_slug
+    pub team_slug: Option<String>,
+    // maps to ConfigurationOptions.token
+    pub token: Option<String>,
+}
```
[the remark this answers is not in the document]
dimitropoulos: great catch: b54f9cdf00cb36df8a09510b001bceac014d68a8

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+use super::{env::truth_env_var, ConfigurationOptions, Error, ResolvedConfigurationOptions};
+use crate::turbo_json::UIMode;
+
+/*
+Hi! If you're new here:
+1. The general pattern is that:
+  - ConfigurationOptions.token corresponds to TURBO_TOKEN or VERCEL_ARTIFACTS_TOKEN
+  - ConfigurationOptions.team_id corresponds to TURBO_TEAMID or VERCEL_ARTIFACTS_OWNER
+  - ConfigurationOptions.team_slug corresponds to TURBO_TEAM
+1. We're ultimately poking around the env vars looking for _paris_ that make sense.
```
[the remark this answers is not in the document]
dimitropoulos: wishful thinking: I've not yet been to Paris, lol

63033b92b87abe292c4cd81422f473375f7d3f12

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+    fn new() -> Self {
+        Self {
+            TURBO_TEAM: None,
+            TURBO_TEAMID: None,
+            TURBO_TOKEN: None,
+            VERCEL_ARTIFACTS_OWNER: None,
+            VERCEL_ARTIFACTS_TOKEN: None,
+        }
+    }
+}
```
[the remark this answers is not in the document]
dimitropoulos: I _love_ feedback like this! thanks!!

d1cc12b57ca98ee91c0fe43edc575f95eeed3441

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+            output.team_slug = input.TURBO_TEAM;
+            output.token = input.TURBO_TOKEN;
+
+            if input.TURBO_TEAMID.is_some() {
+                // there can also be a TURBO_TEAMID, so we'll use that as well
+                output.team_id = input.TURBO_TEAMID;
+            }
+
+            return output;
+        }
```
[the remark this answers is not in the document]
dimitropoulos: Screenshot_20241016_210935

They are in fact subtly different.  I personally have to read at like 0.0001 words per minute to catch it.  Open to ideas if you think there's a better way to restructure these sections.  I sorta intentionally did them in a really "stupid/simple" way because there's just too much tricky stuff here with the business logic in the first place

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+    fn ui(&self) -> Option<UIMode> {
+        // TODO double check on what's going on here
+        let value = self
+            .environment
+            .get(OsStr::new("ci"))
+            .or_else(|| self.environment.get(OsStr::new("no_color")))?;
+        match truth_env_var(value.to_str()?)? {
+            true => Some(UIMode::Stream),
+            false => None,
+        }
```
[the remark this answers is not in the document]
dimitropoulos: re: TODO, cool: 721050b793e17f870e39505aadaae9b76fd81063

re other suggest.  this is my first time seeing the sequence `()?)?.` and I'm pretty cool with it! haha: dc3737667619f366c72231acd487ac5baf49c38c

crates/turborepo-lib/src/config/override_env.rs
dimitropoulos: ahhhh yesssss wonderful.  I was wrestling with this for a while trying to get that to work.  Not sure, now that I see the solution, what I was stuck on, but thank you thank you 9586307e0ed2b02cd1bce4c028080052cfc48a13

crates/turborepo-lib/src/config/override_env.rs
dimitropoulos: so funny you should mention that - that's what I originally had, but `rustfmt` made it really miserable so I switched to this.  if you don't care I'll leave it like this because it means that each case looks the same rather than some of them wrapping and some of them not (including continuations).

crates/turborepo-lib/src/config/override_env.rs
dimitropoulos: !!!!!!!!!! AWESOME

cffcb17fa23f293f483b45695304064813db9007

that's _EXACTLY_ the API I was hoping to achieve in the first place.  thanks for walking me through that!!

crates/turborepo-lib/src/config/override_env.rs
dimitropoulos: wonderful! ac49766bb27dfa017d4074cdc9584b88094e70a7

crates/turborepo-lib/src/config/override_env.rs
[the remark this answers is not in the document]
chris-olszewski: Ah, you're right!

You could use a match statement on a tuple of all the input values, but that has the downside of losing the field names. Up to you if `(None, Some(_TURBO_TEAM_ID), None, Some(_TURBO_TOKEN)) => ` would be easier to understand compared to the if's

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+                .map(|s| s.to_str().unwrap().to_string()),
+            TURBO_TOKEN: environment
+                .get(OsStr::new("turbo_token"))
+                .map(|s| s.to_str().unwrap().to_string()),
+            VERCEL_ARTIFACTS_OWNER: environment
+                .get(OsStr::new("vercel_artifacts_owner"))
+                .map(|s| s.to_str().unwrap().to_string()),
+            VERCEL_ARTIFACTS_TOKEN: environment
+                .get(OsStr::new("vercel_artifacts_token"))
+                .map(|s| s.to_str().unwrap().to_string()),
```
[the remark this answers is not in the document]
dimitropoulos: d999baad84aa0ec7890f9e71de161d61feddc321

chris-olszewski (approved):
LGTM!

Comment on the `Output::from` implementation is total taste and up to you.

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,493 @@
...
+            output.team_id = input.TURBO_TEAMID;
+            return output;
+        }
+
+        if input.TURBO_TEAM.is_some() {
+            output.team_slug = input.TURBO_TEAM;
+
+            if input.VERCEL_ARTIFACTS_OWNER.is_some() {
+                output.team_id = input.VERCEL_ARTIFACTS_OWNER;
+            }
```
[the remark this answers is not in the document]
dimitropoulos: I absolutely love what you came up with: pushed it here -https://github.com/vercel/turborepo/pull/9249/commits/e4fb4ac27aafc4765c05557aa0f4ec60e8bd5a1f

chris-olszewski (approved):

crates/turborepo-lib/src/config/override_env.rs
```diff
@@ -0,0 +1,446 @@
...
+
+/*
+Hi! If you're new here:
+1. The general pattern is that:
+  - ConfigurationOptions.token corresponds to TURBO_TOKEN or VERCEL_ARTIFACTS_TOKEN
+  - ConfigurationOptions.team_id corresponds to TURBO_TEAMID or VERCEL_ARTIFACTS_OWNER
+  - ConfigurationOptions.team_slug corresponds to TURBO_TEAM
+1. We're ultimately poking around the env vars looking for _pairs_ that make sense.
+   Since we presume that users are the only ones sending TURBO_* and Vercel is the only one sending VERCEL_*, we can make some assumptions.  Namely, we assume that if we have one of VERCEL_ARTIFACTS_OWNER or VERCEL_ARTIFACTS_TOKEN we will always have both.
+1. Watch out for mixing up `TURBO_TEAM` and `TURBO_TEAMID`.  Same for ConfigurationOptions.team_id and ConfigurationOptions.team_slug.
```
tknickman, line 17: 🙏🏼

tknickman (approved):

PR #9260, opened 2024-10-15
feat(query): variables

NicholasLYang:
### Description

Inject variables into a query using a json file.

### Testing Instructions

Added tests to `tests/query/variables.t`

### DIFF
--- a/crates/turborepo-lib/src/commands/query.rs
+++ b/crates/turborepo-lib/src/commands/query.rs
@@ -1,8 +1,7 @@
 use std::{fs, sync::Arc};
 
-use async_graphql::{
-    EmptyMutation, EmptySubscription, Name, Request, Schema, ServerError, Variables,
-};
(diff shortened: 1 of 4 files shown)


crates/turborepo-lib/src/commands/query.rs
```diff
@@ -58,10 +59,20 @@ pub async fn run(
     mut base: CommandBase,
     telemetry: CommandEventBuilder,
     query: Option<String>,
+    variables_path: Option<&Utf8Path>,
 ) -> Result<i32, Error> {
     let signal = get_signal()?;
     let handler = SignalHandler::new(signal);
 
+    let variables: Variables = variables_path
```
dimitropoulos, line 67: suggestion: (sorry if I missed a good reason not to do this) might be nice to move this down with where it's used in the hunk below this one

dimitropoulos (approved):
powerful PR for such a small line count! great!

PR #9264, opened 2024-10-16
fix(docs): circleci config - remove excess colon when calling node/install-packages

komichar:
In this example we need to remove the colon at the end, since we're calling the command without params.

Otherwise we get a schema validation error since it attempts to use lines below as mapping.

### Description

image


chris-olszewski (approved):

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 5eea1aadb8fceb7a6031cbc457d24a6b6d389435

PR #9268, opened 2024-10-16
ci: build all rust packages

NicholasLYang:
### Description

Since we're no longer sharing a repo with turbopack, we can now build all the crates instead of using cargo groups

### DIFF
--- a/.github/workflows/turborepo-test.yml
+++ b/.github/workflows/turborepo-test.yml
@@ -229,7 +229,7 @@ jobs:
 
       - name: Run cargo clippy
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/turborepo-test.yml
```diff
@@ -309,7 +309,7 @@ jobs:
           if [ -z "${RUSTC_WRAPPER}" ]; then
             unset RUSTC_WRAPPER
           fi          
-          cargo groups test turborepo-libraries
+          cargo nextest run --workspace
```
chris-olszewski: Are we ready for `nextest`?

chris-olszewski (approved):
Once 🟢 , then 🟣

PR #9271, opened 2024-10-16
fix(query): remove engine validation for query

NicholasLYang:
### Description

We don't want to validate certain issues with graphs with `turbo query`, because it should be still usable even with an invalid graph.

### Testing Instructions

Added a test that shows that `turbo query` is tolerant of issues that `turbo run` is not.

### DIFF
--- a/turborepo-tests/integration/tests/query/validation.t
+++ b/turborepo-tests/integration/tests/query/validation.t
@@ -1,0 +1,148 @@
+Setup
+  $ . ${TESTDIR}/../../../helpers/setup_integration_test.sh persistent_dependencies/10-too-many
+
+Validate that we get an error when we try to run multiple persistent tasks with concurrency 1
+  $ ${TURBO} run build --concurrency=1
+    x invalid task configuration
+  
+  Error:   x You have 2 persistent tasks but `turbo` is configured for concurrency of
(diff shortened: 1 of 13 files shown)


turborepo-tests/integration/tests/query/validation.t
```diff
@@ -0,0 +1,148 @@
...
+      ,-[turbo.json:27:1]
+   27 |       "dependsOn": [
+   28 |         "app-a#custom"
+      :         ^^^^^^^^^^^^^^
+   29 |       ]
+      `----
+  
+  [1]
+
+However, we don't get an error when we query
```
chris-olszewski, line 64: Up to you since you're the one driving `query`, but returning information that we know is incorrect feels off to me. Do we need to worry about panics down the line in query around assumptions that every task has a valid package and has valid task dependencies?

chris-olszewski (approved):

turborepo-tests/integration/tests/query/validation.t
NicholasLYang: I think it's necessary so people can use query to help refactor their repos into valid configurations for turbo. But yeah, we will have to watch out for panics. I have a larger plan to switch us to a compiler-y error model where we accumulate diagnostics and try to recover from errors as much as possible.

PR #9285, opened 2024-10-18
feat(query): provide ast for files and depth for dependencies

NicholasLYang:
### Description

Some more requests. Add the serialized AST from SWC and a depth limit for dependency tracing.

### Testing Instructions

Added some tests to `turbo-trace.t`


chris-olszewski (approved):
LGTM, should hold off on merging until 2.2 goes out.

PR #9289, opened 2024-10-19
fix(tui): avoid tui shutdown on unrecognized input

chris-olszewski:
### Description

This PR does 3 things:
 - Fix run summary from being printed out during watch mode which corrupts the TUI
 - Adds additional debug logs recording the reason for the TUI shutting down
 - Fixes bug where `poll` could now return `None` even if input streams were still available

The bug came down to:
 - A crossterm event was received
 - Translating the crossterm event via `input_options.handle_crossterm_event` returned `None` (This makes total sense, not every terminal event requires a response from the TUI)
 - This `None` was returned instead of waiting for more input or a tick event
 - The `while let Some(event) = poll(...)` loop would exit which triggers a TUI shut down

### Testing Instructions

`npx create-turbo@latest` and then start up the TUI `turbo dev --ui=tui`, without hitting `Enter` to focus the task hit `q` multiple times and you should see the TUI shut down.

Use `turbo` built from this PR and try the same thing. The TUI should not shut down on the press of `q`.

### DIFF
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -574,8 +574,6 @@ pub async fn run_app(tasks: Vec<String>, receiver: AppReceiver) -> Result<(), Er
             }
         };
 
-    debug!("it's real, we're shutting down");
-
     cleanup(terminal, app, callback)?;
 
     result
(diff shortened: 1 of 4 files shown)


crates/turborepo-ui/src/tui/app.rs
```diff
@@ -568,9 +568,14 @@ pub async fn run_app(tasks: Vec<String>, receiver: AppReceiver) -> Result<(), Er
...
         match run_app_inner(&mut terminal, &mut app, receiver, crossterm_rx).await {
             Ok(callback) => (Ok(()), callback),
-            Err(err) => (Err(err), None),
+            Err(err) => {
+                debug!("tui shutting down: {err}");
+                (Err(err), None)
+            }
         };
 
+    debug!("it's real, we're shutting down");
```
anthonyshew, line 577: Did you want this in the PR with this language?

anthonyshew (approved):

PR #9302, opened 2024-10-21
fix: use oxc-resolver hot fix for symlinks

NicholasLYang:
### Description

There's a [bug in oxc-resolver](https://github.com/oxc-project/oxc-resolver/issues/263) for symlinks that basically breaks tsconfig resolution with pnpm (in a specific scenario for extending a tsconfig that's a workspace dependency). Until that's fixed, this PR switches us to a fork that has a very dumb hot fix

We also add tsconfig loading, either via a parameter or via walking up the path hierarchy for the closest tsconfig

### Testing Instructions

Added a test based off of [my repro](https://github.com/nicholaslyang/oxc-repro). Validated that doesn't work with main branch of oxc-resolver, but does work on my fork


crates/turborepo-lib/src/query/file.rs
```diff
@@ -147,11 +154,21 @@ impl File {
         Ok(self.path.to_string())
     }
 
-    async fn dependencies(&self, depth: Option<usize>) -> TraceResult {
+    async fn dependencies(&self, depth: Option<usize>, ts_config: Option<String>) -> TraceResult {
```
chris-olszewski, line 157: I think this is a behavior change where we now respect `tsconfig`? Maybe update PR description to include functionality this gains us? (I'm curious and want to know, but can't grok it from this PR)

chris-olszewski (approved):

Boshen:
I'll make a release today.

PR #9306, opened 2024-10-22
chore(visitor): break up visitor to multiple modules

chris-olszewski:
### Description

This is a prefactor to break up `visitor.rs` into multiple files. This is primarily just moving code around, but there are a few more in depth changes:
 - https://github.com/vercel/turborepo/commit/2873556750540878056509a019fd2f601cac9f64 changes `TaskWarning` to only be constructed if there are missing envs, removing the need to check later on
 - https://github.com/vercel/turborepo/commit/88e98c5422a869768e9a860031b828c95aa5fd4a changes so we no longer call `which` for every task to find the package manager binary. We instead call this on construction, but will only handle the error case on actual command creation.

I highly recommend reviewing each commit on it's own. All commits that move code do only that so hopefully that makes it easier to review.

### Testing Instructions

Integration tests and some basic manual testing.


crates/turborepo-lib/src/task_graph/visitor/exec.rs
```diff
@@ -365,7 +353,7 @@ impl ExecContext {
             }
         }
 
-        let cmd = self.command()?;
+        let cmd = self.cmd.take().expect("should be present");
```
NicholasLYang, line 352: Just curious, why is this an `Option`? Do we have other ways of initializing an `ExecContext` that don't include a command?

NicholasLYang (approved):

tknickman (approved):

dimitropoulos (approved):
I know you just moved it: but wow, `visit`'s a chunky one!  don't mean that in a bad way, it's got a lot to do, but just interesting

crates/turborepo-lib/src/task_graph/visitor/exec.rs
chris-olszewski: This is because `exec` takes a reference to `&self` instead of `self` so we can't move `cmd` out of `self`. I'll switch this to a `clone` to make it clear this is always present.

PR #9313, opened 2024-10-22
chore(cache): better forbidden error message

chris-olszewski:
### Description

This PR makes the error message for the case that the provided token does not have [permission to upload an artifact](https://vercel.com/docs/rest-api/endpoints/artifacts#upload-a-cache-artifact-response-codes). Currently this case leads to us printing `WARNING  failed to contact remote cache: unknown status forbidden: Not authorized` the PR will change this to `WARNING Insufficient permissions to write to remote cache. Please verify that your role has write access for Remote Cache Artifact at https://vercel.com/docs/accounts/team-members-and-roles/access-roles/team-level-roles`

This does not replace the more detailed warnings when we [get a cache status](https://github.com/vercel/turborepo/blob/main/crates/turborepo-api-client/src/lib.rs#L266). These errors get printed as warnings via the [async cache](https://github.com/vercel/turborepo/blob/main/crates/turborepo-cache/src/async_cache.rs#L87)

### Testing Instructions

Added unit tests for error messages in this case.

### DIFF
--- a/crates/turborepo-cache/src/lib.rs
+++ b/crates/turborepo-cache/src/lib.rs
@@ -83,7 +83,7 @@ pub enum CacheError {
     ConfigCacheInvalidBase,
     #[error("Unable to hash config cache inputs")]
     ConfigCacheError,
-    #[error("Insufficient permissions to write to remote cache. Please verify that your role has write access for Remote Cache Artifact at https://vercel.com/docs/accounts/team-members-and-roles/access-roles/team-level-roles")]
+    #[error("Insufficient permissions to write to remote cache. Please verify that your role has write access for Remote Cache Artifact at https://vercel.com/docs/accounts/team-members-and-roles/access-roles/team-level-roles?resource=Remote+Cache+Artifact")]
     ForbiddenRemoteCacheWrite,
 }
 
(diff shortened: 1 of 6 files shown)


crates/turborepo-cache/src/lib.rs
```diff
@@ -83,6 +83,8 @@ pub enum CacheError {
     ConfigCacheInvalidBase,
     #[error("Unable to hash config cache inputs")]
     ConfigCacheError,
+    #[error("Insufficient permissions to write to remote cache. Please verify that your role has write access for Remote Cache Artifact at https://vercel.com/docs/accounts/team-members-and-roles/access-roles/team-level-roles")]
```
tknickman, line 86: Let's add the query param to filter it down:
```suggestion
    #[error("Insufficient permissions to write to remote cache. Please verify that your role has write access for Remote Cache Artifact at https://vercel.com/docs/accounts/team-members-and-roles/access-roles/team-level-roles?resource=Remote+Cache+Artifact")]
```

tknickman (approved):
Love it, one suggestion but lgtm!

NicholasLYang (approved):

PR #9314, opened 2024-10-22
refactor(test): rust integration tests

NicholasLYang:
### Description

Fleshing out the rust integration tests. Added a macro for creating an integration test. The macro isn't perfect since it doesn't support running non-turbo commands for setup, and multiple arguments aren't implemented.


crates/turborepo/tests/common/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+        .arg("-c")
+        .arg(format!(
+            "{} {} {}",
+            script_path,
+            fixture,
+            package_manager.unwrap_or("npm@10.5.0")
+        ))
+        .current_dir(test_dir)
+        .spawn()?
+        .wait()?;
```
chris-olszewski, line 23: Can we capture `stdout`/`stderr` and make sure it gets into the error message? Would help a lot if this fails in CI and we want to know what went wrong.

crates/turborepo/tests/common/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+                let stdout = String::from_utf8(output.stdout)?;
+                let query_output: serde_json::Value = serde_json::from_str(&stdout)?;
+                let test_name = format!(
+                    "{}_{}_{}_({})",
+                    module_path!(),
+                    $fixture,
+                    $name.replace(' ', "_"),
+                    $package_manager
+                );
+                insta::assert_json_snapshot!(test_name, query_output);
```
chris-olszewski, line 52: Can we put each of these asserts in it's own test? Currently if the first test case fails, none of the other ones will run. Prysk will continue running if one of the commands doesn't produce matching output.

crates/turborepo/tests/common/mod.rs
```diff
@@ -0,0 +1,56 @@
+use std::{path::Path, process::Command};
+
+use camino::Utf8Path;
+
+pub fn setup_fixture(
+    fixture: &str,
+    package_manager: Option<&str>,
```
dimitropoulos, line 7: suggestion: it would be nice to keep this explicit so it's really clear from looking at the test what it's gonna run with.  or, if not, it'd be nice if there was some way to make `check!` variadic (is there?).

crates/turborepo/tests/common/mod.rs
NicholasLYang: Is there a way to do this while also using the same fixture? I don't want to spin up a fixture for each test. Insta does also let you do the prysk behavior with `cargo insta test --review` or `INSTA_FORCE_PASS=1 cargo test --no-fail-fast`

crates/turborepo/tests/common/mod.rs
NicholasLYang: If the test fails, shouldn't cargo print stderr by default? And `spawn` forwards stderr, so it should show up in CI

chris-olszewski: Ah you're right. My mistake.

crates/turborepo/tests/common/mod.rs
chris-olszewski: I suppose we'd have to store the tempfile in something like a static ref and have special cleanup logic. I think eventually we'll want `beforeAll`/`afterAll` behavior to make debugging failures easier. ITG

crates/turborepo/tests/common/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+            crate::common::setup_fixture($fixture, None, tempdir.path())?;
+            $(
+                let output = assert_cmd::Command::cargo_bin("turbo")?
+                .arg($command)
+                    .arg($query)
+                    .current_dir(tempdir.path())
+                    .output()?;
+
+                let stdout = String::from_utf8(output.stdout)?;
+                let query_output: serde_json::Value = serde_json::from_str(&stdout)?;
```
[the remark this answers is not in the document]
chris-olszewski: Minor, but there still isn't mention in the doc comment or macro name that this is for JSON command only.

crates/turborepo/tests/common/mod.rs
chris-olszewski: > Insta does also let you do the prysk behavior with cargo insta test --review or INSTA_FORCE_PASS=1 cargo test --no-fail-fast

I don't think these do what we want, `INSTA_FORCE_PASS=1` seems to just forcibly pass all snapshot assertions without printing any diffs. `cargo insta test --review` is just equivalent to `--interactive` in prysk.

chris-olszewski (approved):
ITG

crates/turborepo/tests/snapshots/query__turbo_trace_get_`main.ts`_with_depth_=_0_(npm@10.5.0)_err.snap
```diff
@@ -0,0 +1,8 @@
+---
+source: crates/turborepo/tests/query.rs
+expression: stderr
+---
+ WARNING  No locally installed `turbo` found. Using version: 2.2.4-canary.0.
+turbo 2.2.4-canary.0
```
chris-olszewski, line 6: We don't want to capture this since we always use newest turbo version

crates/turborepo/tests/common/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+                    $package_manager
+                );
+
+                if !output.stderr.is_empty() {
+                    let stderr = String::from_utf8_lossy(&output.stderr);
+                    let stderr_test_name = format!(
+                        "{}_err",
+                        test_name
+                    );
+                    insta::assert_snapshot!(stderr_test_name, stderr);
```
chris-olszewski, line 53: I don't think we want to snapshot stderr since it prints the turbo version. Can you write a custom assertion that makes sure we print version etc?

NicholasLYang: yeah I'm gonna cut stderr snapshotting for now. When we need it I'll add it

PR #9319, opened 2024-10-23
feat(query): reverse tracing

NicholasLYang:
### Description

Implements an MVP of reverse tracing. Basically goes through all the JavaScript/TypeScript files in the repo, checks which ones import the reverse traced file, and returns those. Does this concurrently with a `JoinSet`.

Also adds code to detect if a tsconfig is in scope for a file and if so, add it to the resolver. That way custom aliases work.

### Testing Instructions

Added some tests including a monorepo setup with tsconfigs in the packages that define aliases


crates/turbo-trace/src/tracer.rs
```diff
@@ -79,104 +237,104 @@ impl Tracer {
...
-            );
+        TraceResult {
+            files: seen,
+            errors: self.errors,
+        }
+    }
 
-            let mut parser = Parser::new_from(Capturing::new(lexer));
+    pub async fn reverse_trace(mut self) -> TraceResult {
+        let files = match globwalk::globwalk(
```
NicholasLYang, line 264: I'd love at some point to combine the glob walking and the import resolution so that we can have more concurrency. Right now we have to walk the entire repo, then do the import resolution step.

crates/turbo-trace/src/tracer.rs
```diff
@@ -61,11 +71,159 @@ impl Tracer {
...
 
-    pub fn trace(mut self, max_depth: Option<usize>) -> TraceResult {
+    pub async fn get_imports_from_file(
+        source_map: &SourceMap,
+        errors: &mut Vec<TraceError>,
+        resolver: &Resolver,
+        file_path: &AbsoluteSystemPath,
+    ) -> Option<(Vec<AbsoluteSystemPathBuf>, SeenFile)> {
+        // Read the file content
+        let Ok(file_content) = tokio::fs::read_to_string(&file_path).await else {
```
chris-olszewski, line 87: Not worth blocking, but I wonder if throwing rayon or a thread pool would be more effective for the reverse trace. The only concurrency we're getting is the file reads and I'm not sure how much of the weight of this function is FS vs parsing and extracting info.

crates/turbo-trace/src/tracer.rs
```diff
@@ -61,11 +71,159 @@ impl Tracer {
...
+            EsVersion::EsNext,
+            StringInput::from(&*source_file),
+            Some(&comments),
+        );
+
+        let mut parser = Parser::new_from(Capturing::new(lexer));
+
+        // Parse the file as a module
+        let Ok(module) = parser.parse_module() else {
+            errors.push(TraceError::FileNotFound(file_path.to_owned()));
```
chris-olszewski, line 124: I think this should be a `TraceError::ParseError`?

chris-olszewski (approved):
LGTM! Before we do any perf tinkering we should collect some profiles or set up a benchmark.

crates/turbo-trace/src/tracer.rs
NicholasLYang: yeah I'm not exactly sure. I think we should do some investigation after it gets some use

crates/turbo-trace/src/main.rs
```diff
@@ -12,12 +13,17 @@ struct Args {
     cwd: Option<Utf8PathBuf>,
     #[clap(long)]
     ts_config: Option<Utf8PathBuf>,
+    #[clap(long)]
+    node_modules: Option<Utf8PathBuf>,
```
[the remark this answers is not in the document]
NicholasLYang: Y'know I'm not exactly sure. `oxc_resolver` is a little ambiguous here. This CLI isn't meant for public use, mostly just for testing

PR #9322, opened 2024-10-23
upgrade TypeScript one minor to 5.4.5 from 5.3.3

dimitropoulos:
### Description

in working on https://github.com/vercel/turborepo/pull/9155 I've noticed that we also need to upgrade TypeScript, so I'm doing that here, perhaps not all the way to latest.

### Testing Instructions

Everything should work normally - this shouldn't have any behavioral changes.


chris-olszewski (changes requested):
I think the PR is good, but just marking not to merge this until we get an examples test run with these changes.

The [examples test](https://github.com/vercel/turborepo/actions/runs/11485954025/job/31967243921?pr=9322#step:6:123) was a `>>> FULL TURBO`. Even though the examples most definitely changed.

It looks like our example tests aren't actually depending on the examples. I think we're missing a `^topo` somewhere

anthonyshew:
Huh, that's really weird. You could also choose to bump it everywhere *except* in the examples? I would hope that would work...

chris-olszewski (approved):
Double checked task inputs and it was a false alarm.

LGTM, thanks for bumping these.

PR #9330, opened 2024-10-24
fix(watch): stopping persistent tasks

NicholasLYang:
### Description

We were accidentally stopping persistent tasks when we shouldn't have because we were reusing the same stopper.

### Testing Instructions

We need a way to test persistent stuff and also run construction stuff.

### DIFF
--- a/crates/turborepo-lib/src/run/mod.rs
+++ b/crates/turborepo-lib/src/run/mod.rs
@@ -137,7 +137,12 @@ impl Run {
     }
 
     pub fn create_run_for_non_interruptible_tasks(&self) -> Self {
-        let mut new_run = self.clone();
+        let mut new_run = Self {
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/run/mod.rs
```diff
@@ -137,7 +137,12 @@ impl Run {
     }
 
     pub fn create_run_for_non_interruptible_tasks(&self) -> Self {
-        let mut new_run = self.clone();
+        let mut new_run = Self {
+            // ProcessManager is a singleton, so we want to explicitly recreate it
```
chris-olszewski, line 141: Non blocking, but I think singleton is the wrong term as that means that only one of these should be created per thread. I think just saying the cloning a ProcessManager shares state is clear enough.

chris-olszewski (approved):

PR #9341, opened 2024-10-28
Fix typo

ATeals:
### Description

I made a small modification to the document.

I think the structure of the curly braces should be like this for better readability. 😁


anthonyshew (approved):
Oops! Thanks!

anthonyshew (approved):
Good catch! Thank you.

anthonyshew (approved):
Good catch, thanks!

PR #9343, opened 2024-10-28
chore: bump dependencies for Windows ARM support

NicholasLYang:
### Description

Unfortunately this does *not* get us to build Windows ARM natively, but it at least eliminates the first blocker, which is ring 0.16. ring is a crypto library that `rustls` uses under the hood, and because it relies on inline assembly, it can't be built for Windows ARM. However, ring 0.17 added Windows ARM support. To get all our deps into ring 0.17, we need to bump reqwest and ureq. The first one is easy, but the second one is used by update-informer, so this PR redirects to my fork of update-informer that has a bumped ureq version.

### Testing Instructions

Try building on Windows ARM and validate that we "only" get an error for building tokio

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,10 +66,6 @@ turborepo-vercel-api = { path = "crates/turborepo-vercel-api" }
 turborepo-vercel-api-mock = { path = "crates/turborepo-vercel-api-mock" }
 turborepo-vt100 = { path = "crates/turborepo-vt100" }
 
-# Be careful when selecting tls backend, including change default tls backend.
-# If you changed, must verify with ALL build targets with next-swc to ensure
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -101,7 +97,7 @@ console-subscriber = "0.1.8"
 crossbeam-channel = "0.5.8"
 dashmap = "5.4.0"
 derive_setters = "0.1.6"
-dialoguer = "0.10.3"
+dialoguer = "0.11.0"
```
NicholasLYang, line 100: required to keep compatible dependencies between reqwest and dialoguer

chris-olszewski (approved):

PR #9344, opened 2024-10-28
feat: support task id entrypoints

chris-olszewski:
### Description

Currently if you run `turbo cli#build` you will get the following failure*:
```
Error:   × could not find task `cli#build` in project
```
*It works if you had `"tasks": {"cli#build": {...}}` in your root level `turbo.json`

After this PR, passing a fully qualified task id to run will now work, regardless of how the task definition is defined:
 - In the root `turbo.json` as `build`
 - In the root `turbo.json` as `cli#build`
 - In the `cli` workspace's `turbo.json` as `build`

The usage of `#` is safe here as you already have been unable to use `#` in a unqualified task name. 
- If you attempt to use a `#` in a task name in a workspace level `turbo.json` you will get an error about using package task syntax in a workspace file.
- If you attempt to specify a task in the root `turbo.json` of the form `foo#bar` it will be read as the `bar` task in package `foo`
- If you attempt to use `foo#bar#baz` as a task name in root `turbo.json` it will work currently with `foo#bar#baz` and after this PR it will continue to work

### Testing Instructions

Added unit tests!

Manual verification by running `turbo cli#build`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6361,6 +6361,7 @@ dependencies = [
  "human_format",
  "humantime",
  "ignore",
+ "insta",
  "itertools 0.10.5",
  "jsonc-parser 0.21.0",
  "lazy_static",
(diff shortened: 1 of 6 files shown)


tknickman (approved):

NicholasLYang (approved):

NicholasLYang (commented):
Only feedback is maybe add an integration test?

NicholasLYang (approved):

PR #9346, opened 2024-10-29
feat(query: include css and json files in trace output

NicholasLYang:
### Description

Include CSS and JSON files in the trace output, but don't recurse into them.

### Testing Instructions

Added a CSS and JSON import in the `turbo_trace` fixture


chris-olszewski (approved):
:shipit:

PR #9348, opened 2024-10-29
feat(turbo): add cache flag

NicholasLYang:
### Description

Adds a cache flag that lets you set cache permissions per cache type and per action (read or write)

### Testing Instructions
Adds tests in `turborepo_cache/src/config.rs` for parsing and in `turborepo-lib/src/opts.rs` for loading and resolving


crates/turborepo-lib/src/opts.rs
```diff
@@ -362,16 +358,37 @@ impl<'a> TryFrom<OptsInputs<'a>> for ScopeOpts {
 impl<'a> From<OptsInputs<'a>> for CacheOpts {
     fn from(inputs: OptsInputs<'a>) -> Self {
         let is_linked = turborepo_api_client::is_linked(inputs.api_auth);
-        let skip_remote = if !is_linked {
-            true
+        let mut cache = inputs.config.cache();
```
chris-olszewski, line 361: I think this layering logic is confusing since env vars or config settings can override CLI level cache settings:
e.g. `TURBO_FORCE=1 turbo run build --force=false` would result in `remote:rw,local:rw` cache behavior, but `TURBO_REMOTE_ONLY=1 turbo run build --cache='local:rw` would result in no caches are enabled.

I think it would make more sense to fold the existing cache flags into the new cache struct at each level in the config, but I'm not convinced that wouldn't be confusing. Could we get away with having the new cache struct conflict with existing cache flags at each level e.g `turbo run build --remote-only --cache='...'` would throw but `TURBO_REMOTE_ONLY=1 turbo run build --cache='...'` be fine where `--cache` wins out completely?

crates/turborepo-cache/src/lib.rs
```diff
@@ -105,12 +109,43 @@ pub struct CacheHitMetadata {
...
+    pub fn should_use(&self) -> bool {
+        self.read || self.write
+    }
+}
+
+#[derive(Serialize, Deserialize, Deserializable, Debug, PartialEq, Eq, Clone, Copy, Default)]
+pub struct CacheConfig {
+    pub local: CacheActions,
+    pub remote: CacheActions,
+}
```
chris-olszewski, line 128: Minor code structure issue: I think the ser/de implementations should live where we care/use them in the `turborepo_lib::config` module. Also, do we want to support users setting this value in `.turbo/config.json`, if not then we should just skip deserializing this field and avoid the unused implementation.

chris-olszewski (changes requested):
I'm excited for this, biggest thing I'd like to see is some tests regarding how the new `cache` struct gets resolved wrt existing cache config settings.

crates/turborepo-lib/src/opts.rs
NicholasLYang: Is there a way to have a group of flags conflict with a single flag but not with each other? I wasn't sure if that's possible in clap. Ofc I can manually add that as a restriction

NicholasLYang: The layering issue is a tricky one. If the config specifies `remote:rw` in a file, but then we add on `local:rw`, do we want to do an `OR` or an `AND` for those? I could see either one working

crates/turborepo-cache/src/config.rs
```diff
@@ -0,0 +1,231 @@
...
+    )]
+    #[test_case("local:r,local:w", Err(Error::DuplicateKeys { text: "local:r,local:w".to_string(), key: "local", span: Some(SourceSpan::new(8.into(), 5.into())) }) ; "duplicate local key"
+    )]
+    #[test_case("local:rr", Err(Error::DuplicateActions { text: "local:rr".to_string(), action: "r (read)", span: Some(SourceSpan::new(6.into(), 5.into())) }) ; "duplicate action")]
+    #[test_case("remote:r,local=rx", Err(Error::InvalidCacheTypeAndAction { text: "remote:r,local=rx".to_string(), pair: "local=rx".to_string(), span: Some(SourceSpan::new(9.into(), 8.into())) }) ; "invalid key action pair")]
+    #[test_case("local:rx", Err(Error::InvalidCacheAction { c: 'x', text: "local:rx".to_string(), span: Some(SourceSpan::new(6.into(), 5.into())) }) ; "invalid action")]
+    #[test_case("file:r", Err(Error::InvalidCacheType { s: "file".to_string(), text: "file:r".to_string(), span: Some(SourceSpan::new(0.into(), 4.into())) }) ; "invalid cache type")]
+    fn test_cache_config(s: &str, expected: Result<CacheConfig, Error>) {
+        assert_eq!(CacheConfig::from_str(s), expected);
```
chris-olszewski, line 229: Not blocking at all, but it might be nice to snapshot the error messages, easier to update if we change the tests/logic at all. There's a JSON reporter that makes `assert_json_snapshot` easy to use: https://github.com/vercel/turborepo/blob/main/crates/turborepo-lib/src/engine/builder.rs#L1358

chris-olszewski (approved):
LFG

PR #9349, opened 2024-10-29
fix: yarn prune?

Netail:
### Description

Possible fix for; https://github.com/vercel/turborepo/issues/8849 ?


Netail:
I am wonder; should turbo actually prune lock files themself or just let the package handle that and just copy over the existing lock file? (Quite some maintenance work) NPM, yarn, yarn berry do have a prune functionality afaik, bun I don't think so yet?

Netail:
In favour of; https://github.com/vercel/turborepo/pull/9627

PR #9350, opened 2024-10-30
feat(ui): Hide task list with `h` key.

anthonyshew:
### Description

Toggle the visibility of the task list with an `h` hotkey.

https://github.com/user-attachments/assets/7a8b3ad1-5ead-47d0-9aaf-513acce6ad3c

Future note: I don't love that I'm stealing a vertical line from the task list. Screen real estate is valuable. It's likely we could give the `Task` header and footer instructions a faded style and get rid of the horizontal bars. (I'd recommend doing this regardless of the result of this PR.

### Testing Instructions

👀

### DIFF
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -49,6 +49,7 @@ pub struct App<W> {
     scroll: TableState,
     selected_task_index: usize,
     has_user_scrolled: bool,
+    has_sidebar: bool,
     done: bool,
 }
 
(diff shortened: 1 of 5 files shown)


chris-olszewski (approved):

chris-olszewski:
> Future note: I don't love that I'm stealing a vertical line from the task list. Screen real estate is valuable. It's likely we could give the Task header and footer instructions a faded style and get rid of the horizontal bars. (I'd recommend doing this regardless of the result of this PR.

This makes sense to me

PR #9356, opened 2024-10-30
feat(mfe): inject local proxy

chris-olszewski:
### Description

This PR adds all support necessary for starting a MFE local proxy:
 - Adds new `CommandProvider` to allow for multiple strategies for constructing a command depending on the type
 - Adds parsing logic for `micro-frontends.jsonc`
 - Parses all `micro-frontends.jsonc` found in the package graph*
 - Adding a task definition for the generated `proxy` task if MFE configs are found
 - Explicitly include the `proxy` tasks in engine building step if a MFE dev task is found in the product of the package `--filter` and the tasks specified in `turbo run`
 - Adds a `CommandProvider` that construct a command that invokes the MFE local proxy

⚠️ This currently only support MFE configs that are located in shared packages and doesn't respect a loose MFE configs.

Follow up work:
 - [ ] Handle configurable MFE config location

### Testing Instructions

In a repository with MFE, run a dev task for one of the configured MFE. This should result in a `#proxy` task being injected for that MFE and correctly configured to route traffic to the dev tasks or a remote host.

### DIFF
--- a/crates/turborepo-lib/src/task_graph/visitor/command.rs
+++ b/crates/turborepo-lib/src/task_graph/visitor/command.rs
@@ -1,11 +1,14 @@
-use std::path::PathBuf;
+use std::{
+    collections::{HashMap, HashSet},
+    path::PathBuf,
+};
 
 use turbopath::AbsoluteSystemPath;
 use turborepo_env::EnvironmentVariableMap;
 use turborepo_repository::package_graph::{PackageGraph, PackageInfo, PackageName};
 
 use super::Error;
-use crate::{opts::TaskArgs, process::Command, run::task_id::TaskId};
+use crate::{engine::Engine, opts::TaskArgs, process::Command, run::task_id::TaskId};
 
 pub trait CommandProvider {
     fn command(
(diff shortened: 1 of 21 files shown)


crates/turborepo-lib/src/turbo_json/loader.rs
```diff
@@ -147,9 +170,19 @@ impl TurboJsonLoader {
...
+                packages,
+                micro_frontends_configs,
+            } => {
                 let path = packages.get(package).ok_or_else(|| Error::NoTurboJSON)?;
-                load_from_file(&self.repo_root, path)
+                let should_inject_proxy_task = micro_frontends_configs
+                    .as_ref()
+                    .map_or(false, |configs| configs.contains_package(package.as_str()));
+                if should_inject_proxy_task {
+                    load_from_file_with_proxy(&self.repo_root, path)
```
NicholasLYang, line 182: Could we make this a method on `TurboJson`, i.e. `with_proxy()` that injects the proxy task?

crates/turborepo-lib/src/task_graph/visitor/command.rs
```diff
@@ -80,3 +133,183 @@ impl<'a> CommandFactory<'a> {
...
+            .iter()
+            .filter(|task| self.tasks_in_graph.contains(task))
+            .map(|task| task.package());
+        let package_dir = self.repo_root.resolve(package_info.package_path());
+        let mfe_path = package_dir.join_component("micro-frontends.jsonc");
+        let mut args = vec!["proxy", mfe_path.as_str(), "--names"];
+        args.extend(local_apps);
+
+        // TODO: leverage package manager to find the local proxy
+        let program = package_dir.join_components(&["node_modules", ".bin", "micro-frontends"]);
```
tknickman, line 213: +1

crates/turborepo-lib/src/task_graph/visitor/mod.rs
```diff
@@ -97,6 +99,11 @@ pub enum Error {
     InternalErrors(String),
     #[error("unable to find package manager binary: {0}")]
     Which(#[from] which::Error),
+    #[error(
+        "'{package}' is configured with a microfrontend.jsonc, but doesn't have \
```
tknickman, line 103: We can address this later, but this config name isn't guaranteed and is just a convention right now. But we can fix that once we have the path available for you to source.

crates/turborepo-micro-frontend/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+
+use std::collections::BTreeMap;
+
+use biome_deserialize_macros::Deserializable;
+use biome_json_parser::JsonParserOptions;
+pub use error::Error;
+use serde::Serialize;
+use turbopath::AbsoluteSystemPath;
+
+pub const MICRO_FRONTENDS_CONFIG: &str = "micro-frontends.jsonc";
```
tknickman, line 11: Same note as before - this will change soon!

tknickman (approved):
lgtm, nothing blocking for now

crates/turborepo-micro-frontend/src/error.rs
```diff
@@ -0,0 +1,26 @@
...
+pub enum Error {
+    #[error("Unable to read configuration file: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("Unable to parse JSON: {0}")]
+    JsonParse(String),
+}
+
+impl Error {
+    /// Constructs an error message from multiple biome diagnostic errors
+    pub fn biome_error(errors: Vec<biome_diagnostics::Error>) -> Self {
```
NicholasLYang, line 13: we already have a lot of this logic in `turborepo-errors/lib.rs` if you want to pull from that

crates/turborepo-micro-frontend/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+    pub development: Development,
+    pub production: Host,
+    pub metadata: Option<BTreeMap<String, String>>,
+    pub federation: Option<Federation>,
+    pub vercel: Option<Vercel>,
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserializable, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct Vercel {
```
dimitropoulos, line 79: this is missing `default_route`

crates/turborepo-micro-frontend/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub project_id: Option<String>,
```
dimitropoulos, line 81: `projectId` is not optional

crates/turborepo-lib/src/task_graph/visitor/command.rs
```diff
@@ -1,21 +1,69 @@
...
+    ) -> Result<Option<Command>, Error>;
+}
+
+/// A collection of command providers.
+///
+/// Will attempt to find a command from any of the providers it contains.
+/// Ordering of the providers matters as the first present command will be
+/// returned. Any errors returned by the providers will be immediately returned.
 pub struct CommandFactory<'a> {
+    providers: Vec<Box<dyn CommandProvider + 'a + Send>>,
```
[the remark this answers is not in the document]
chris-olszewski: If we ever want to support additional tasks that don't come from `package.json` `scripts` (`turbo.json`, non-JS, etc) I expect this is where we would put those.

crates/turborepo-micro-frontend/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserializable, Default)]
+pub struct Options {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub vercel: Option<VercelOptions>,
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserializable, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct VercelOptions {
```
dimitropoulos, line 57: I was expecting these to match

```ts
  previewDeploymentSuffix?: string;
  teamSlug?: string;
```

what am I missing?

crates/turborepo-micro-frontend/src/lib.rs
chris-olszewski: I might remove a majority of these fields and only keep the ones that we need (which apps are part of the MFE and what they're development tasks are). This was written when the local proxy was going to be in Rust and this information would have been useful.

crates/turborepo-micro-frontend/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+pub struct Development {
+    pub local: Host,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fallback: Option<Host>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub task: Option<String>,
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserializable, Default)]
+pub struct Host {
```
dimitropoulos, line 139: what would you think about trying to match the names of the corresponding typescript types (which calls this `HostConfig`, not `Host`)

crates/turborepo-micro-frontend/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+    pub protocol: Protocol,
+    pub host: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub port: Option<u16>,
+}
+
+#[derive(Debug, PartialEq, Eq, Serialize, Deserializable, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum Protocol {
+    #[default]
```
dimitropoulos, line 149: the typescript types say the default is the `Https` variant, not `Http`.

PR #9365, opened 2024-11-01
wip

anthonyshew:

### DIFF
--- a/crates/turborepo-ui/src/tui/table.rs
+++ b/crates/turborepo-ui/src/tui/table.rs
@@ -124,21 +124,24 @@ impl<'a> StatefulWidget for &'a TaskTable<'a> {
         .column_spacing(0)
         .block(Block::new().borders(Borders::RIGHT))
         .header(
-            vec![format!("Tasks\n{bar}"), " \n─".to_owned()]
-                .into_iter()
-                .map(Cell::from)
-                .collect::<Row>()
-                .height(2),
-        )
-        .footer(
-            vec![
-                format!("{bar}\n{TASK_NAVIGATE_INSTRUCTIONS}\n{HIDE_INSTRUCTIONS}"),
-                format!("─\n "),
-            ]
+            vec![Text::styled(
+                "Tasks\n",
+                Style::default().fg(Color::DarkGray),
+            )]
             .into_iter()
             .map(Cell::from)
             .collect::<Row>()
-            .height(3),
+            .height(1),
+        )
+        .footer(
+            vec![Text::styled(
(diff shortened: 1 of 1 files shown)


chris-olszewski:
> I don't know if there's a way to tell Ratatui to use different foreground colors for light vs. dark backgrounds?

Not to my knowledge. I think we would need to query the terminal for the default fg/bg colors (I do not know how to do this in a terminal agnostic way), some heuristic for deciding if bg was dark/light, and just have two sets of colors.

crates/turborepo-ui/src/tui/table.rs
```diff
@@ -124,21 +124,24 @@ impl<'a> StatefulWidget for &'a TaskTable<'a> {
         .column_spacing(0)
         .block(Block::new().borders(Borders::RIGHT))
         .header(
-            vec![format!("Tasks\n{bar}"), " \n─".to_owned()]
-                .into_iter()
-                .map(Cell::from)
-                .collect::<Row>()
-                .height(2),
+            vec![Text::styled(
+                "Tasks\n",
```
chris-olszewski, line 128: I don't think this really matters since the `height(1)` is set, but may as well make the cell text a single line
```suggestion
                "Tasks",
```

crates/turborepo-ui/src/tui/table.rs
```diff
@@ -124,21 +124,24 @@ impl<'a> StatefulWidget for &'a TaskTable<'a> {
...
+            .height(1),
         )
         .footer(
-            vec![
-                format!("{bar}\n{TASK_NAVIGATE_INSTRUCTIONS}\n{HIDE_INSTRUCTIONS}"),
-                format!("─\n "),
-            ]
+            vec![Text::styled(
+                format!("{TASK_NAVIGATE_INSTRUCTIONS}\n{HIDE_INSTRUCTIONS}"),
+                Style::default().fg(Color::DarkGray),
```
chris-olszewski, line 139: Maybe use the [`DIM`](https://docs.rs/ratatui/latest/ratatui/style/struct.Modifier.html) modifier instead?
```suggestion
                Style::default().add_modifier(Modifier::DIM),
```

chris-olszewski (approved):
I think this looks good, we could look at just using DIM instead of a different color. This is a screenshot of using DIM:
Screenshot 2024-11-01 at 12 23 04 PM

crates/turborepo-ui/src/tui/table.rs
anthonyshew: Oh, that's fantastic. Didn't know it was there.

PR #9367, opened 2024-11-01
feat(trace): pretty print trace errors to logs

NicholasLYang:
### Description

To help debug trace, pretty print the errors to stderr by default. Includes fancy error messages either via SWC or our own miette infrastructure

### Testing Instructions

Try using trace and hit an error

### DIFF
--- a/crates/turbo-trace/src/tracer.rs
+++ b/crates/turbo-trace/src/tracer.rs
@@ -69,7 +69,7 @@ impl TraceResult {
         );
         for error in &self.errors {
             match error {
-                TraceError::ParseError(e) => {
+                TraceError::ParseError(_, e) => {
                     e.clone().into_diagnostic(&handler).emit();
                 }
                 e => {
--- a/crates/turbo-trace/src/tracer.rs
+++ b/crates/turbo-trace/src/tracer.rs
@@ -2,17 +2,22 @@ use std::{collections::HashMap, sync::Arc};
 
 use camino::Utf8PathBuf;
 use globwalk::WalkType;
-use miette::{Diagnostic, NamedSource, SourceSpan};
+use miette::{Diagnostic, Report, SourceSpan};
 use oxc_resolver::{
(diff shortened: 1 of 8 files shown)


crates/turbo-trace/src/tracer.rs
```diff
@@ -31,29 +36,52 @@ pub struct Tracer {
...
-        text: NamedSource,
+        text: String,
     },
     #[error("failed to walk files")]
-    GlobError(#[from] globwalk::WalkError),
+    GlobError(Arc<globwalk::WalkError>),
+}
+
+impl TraceResult {
+    pub fn emit_errors(&self) {
```
chris-olszewski, line 63: So do proc macros mess up unused code detection? Could you add a decl to silence that warning?

crates/turborepo-lib/src/commands/query.rs
```diff
@@ -80,7 +80,8 @@ pub async fn run(
         // If the arg starts with "query" or "mutation", and ends in a bracket, it's
         // likely a direct query If it doesn't, it's a file path, so we need to
         // read it
-        let query = if (trimmed_query.starts_with("query") || trimmed_query.starts_with("mutation"))
+        let query = if (trimmed_query.starts_with("query")
+            || trimmed_query.starts_with("mutation") | trimmed_query.starts_with('{'))
```
chris-olszewski, line 84: I'm surprised there isn't an warning for using bitor
```suggestion
            || trimmed_query.starts_with("mutation") || trimmed_query.starts_with('{'))
```

chris-olszewski (approved):

crates/turbo-trace/src/tracer.rs
NicholasLYang: I think it's actually [the combination binary/library crate](https://github.com/rust-lang/rust/issues/95513) that's causing this. Because from the binary's perspective this is unused

PR #9368, opened 2024-11-01
feat: `turbo info` command for debugging

anthonyshew:
### Description

A command to provide debugging information for reproductions or otherwise understand a machine better. Output will look something close to this.

```
CLI:
   Version: 2.2.4-canary.4
   Location: /Users/anthonyshew/projects/open/turbo/target/debug/turbo
   Daemon status: Running
   Package manager: pnpm

Platform:
   Architecture: aarch64
   Operating system: macos
   Available memory (MB): 13598
   Available CPU cores: 10

Environment:
   CI: None
   Terminal program (TERM_PROGRAM): tmux
   Terminal program version (TERM_PROGRAM_VERSION): 3.4
   Shell (SHELL): /bin/zsh
   stdin: false

Turborepo System Environment Variables:
   TURBO_INVOCATION_DIR: /Users/anthonyshew/projects/open/turbo
```

### Testing Instructions

Try it out!

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -1222,7 +1222,7 @@ pub async fn run(
 
             match info::run(base).await {
                 Ok(()) => {}
-                Err(e) => println!("Command failed. Please file a GitHub Issue.\n{}", e),
+                Err(_e) => panic!("`info` command failed."),
             }
             Ok(0)
         }
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -1220,10 +1220,7 @@ pub async fn run(
                 .track_call();
             let base = CommandBase::new(cli_args.clone(), repo_root, version, color_config);
 
-            match info::run(base).await {
-                Ok(()) => {}
-                Err(_e) => panic!("`info` command failed."),
-            }
+            info::run(base).await;
             Ok(0)
         }
         Command::Telemetry { command } => {
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -1218,8 +1218,8 @@ pub async fn run(
             CommandEventBuilder::new("info")
                 .with_parent(&root_telemetry)
                 .track_call();
-            let mut base = CommandBase::new(cli_args.clone(), repo_root, version, color_config);
-            info::run(base);
+            let base = CommandBase::new(cli_args.clone(), repo_root, version, color_config);
+            info::run(base).await;
 
             Ok(0)
         }
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -22,8 +22,8 @@ use turborepo_ui::{ColorConfig, GREY};
 use crate::{
     cli::error::print_potential_tasks,
     commands::{
-        bin, config, daemon, generate, link, login, logout, ls, prune, query, run, scan, telemetry,
-        unlink, CommandBase,
+        bin, config, daemon, generate, info, link, login, logout, ls, prune, query, run, scan,
+        telemetry, unlink, CommandBase,
     },
     get_version,
     run::watch::WatchClient,
@@ -475,9 +475,7 @@ impl Args {
     }
 }
 
-/// Defines the subcommands for CLI. NOTE: If we change the commands in Go,
-/// we must change these as well to avoid accidentally passing the
-/// --single-package flag into non-build commands.
+/// Defines the subcommands for CLI
 #[derive(Subcommand, Clone, Debug, PartialEq)]
 pub enum Command {
     /// Get the path to the Turbo binary
@@ -570,6 +568,8 @@ pub enum Command {
         #[clap(long)]
         invalidate: bool,
     },
+    /// Print debugging information
+    Info,
     /// Prepare a subset of your monorepo.
     Prune {
         #[clap(hide = true, long)]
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/commands/info.rs
```diff
@@ -0,0 +1,88 @@
...
+        std::env::current_exe()?.to_string_lossy()
+    );
+    println!("   Daemon status: {}", daemon_status);
+    println!("");
+
+    println!("Package managers:");
+    println!("   npm version: {}", "TODO");
+    println!("   yarn version: {}", "TODO");
+    println!("   pnpm version: {}", "TODO");
+    println!("   bun version: {}", "TODO");
```
anthonyshew, line 37: TODO!

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -475,9 +475,7 @@ impl Args {
     }
 }
 
-/// Defines the subcommands for CLI. NOTE: If we change the commands in Go,
```
anthonyshew, line 478: Unrelated, but saw it mentioning Go so ✂️

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -1214,6 +1214,18 @@ pub async fn run(
...
         }
+        Command::Info => {
+            CommandEventBuilder::new("info")
+                .with_parent(&root_telemetry)
+                .track_call();
+            let base = CommandBase::new(cli_args.clone(), repo_root, version, color_config);
+
+            match info::run(base).await {
+                Ok(()) => {}
+                Err(e) => println!("Command failed. Please file a GitHub Issue.\n{}", e),
```
NicholasLYang, line 1225: I'd just panic here. The panic handler is pretty clear on filing a GitHub issue and will include a backtrace

crates/turborepo-lib/src/commands/info.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let package_manager = match PackageJson::load(&base.repo_root.join_component("package.json")) {
+        Ok(package_json) => {
+            match PackageManager::read_or_detect_package_manager(&package_json, &base.repo_root) {
+                Ok(pm) => pm.to_string(),
+                Err(_) => "Not found".to_owned(),
+            }
+        }
+        Err(_) => "Not found".to_owned(),
+    };
```
NicholasLYang, line 33: Just some Rust code golf. I'd recommend manually committing this because Rustfmt will probably disagree with this formatting
```suggestion
    let package_manager = PackageJson::load(&base.repo_root.join_component("package.json"))
        .and_then(|package_json|  PackageManager::read_or_detect_package_manager(&package_json, &base.repo_root))
        .map_or_else(|_| "Not found".to_owned(), |pm| pm.to_string());
```

NicholasLYang (changes requested):
Looks good! Just a couple notes

crates/turborepo-lib/src/cli/mod.rs
anthonyshew: TIL

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -1214,6 +1214,18 @@ pub async fn run(
...
+        Command::Info => {
+            CommandEventBuilder::new("info")
+                .with_parent(&root_telemetry)
+                .track_call();
+            let base = CommandBase::new(cli_args.clone(), repo_root, version, color_config);
+
+            match info::run(base).await {
+                Ok(()) => {}
+                Err(_e) => panic!("`info` command failed."),
+            }
```
chris-olszewski, line 1226: We generally want to avoid `panic` here and if we do panic, we should make sure to include the underlying error.

You can use `?` operator here to bubble up the error if you add a new variant to the cli error type [here](https://github.com/vercel/turborepo/blob/main/crates/turborepo-lib/src/cli/error.rs#L21):
```
#[error(transparent)]
Info(#[from] info::Error),
```
```suggestion
            info::run(base).await?;
```

crates/turborepo-lib/src/commands/info.rs
```diff
@@ -0,0 +1,93 @@
...
+use super::CommandBase;
+use crate::{DaemonConnector, DaemonConnectorError};
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("could not get path to turbo binary: {0}")]
+    NoCurrentExe(#[from] io::Error),
+}
+
+pub async fn run(base: CommandBase) -> Result<(), Error> {
```
chris-olszewski, line 16: ITG (non blocking follow up): Separate the data from the display logic to make this more unit-test friendly and support additional output formats like JSON.

Something like:
```
struct TurboInfo {
    package_manager: String,
    daemon_status: String,
    ...
}
```

And then have `turbo_info.print()` to actually write the data to `stdout`

crates/turborepo-lib/src/commands/info.rs
```diff
@@ -0,0 +1,93 @@
...
+        "   Shell (SHELL): {}",
+        env::var("SHELL").unwrap_or_else(|_| "unknown".to_owned())
+    );
+    println!("   stdin: {}", turborepo_ci::is_ci());
+    println!();
+
+    println!("Turborepo System Environment Variables:");
+    for (key, value) in env::vars() {
+        // Don't print sensitive information
+        if key == "TURBO_TEAM"
```
chris-olszewski, line 80: `TURBO_REMOTE_CACHE_SIGNATURE_KEY` should be added to this list of sensitive keys

crates/turborepo-lib/src/cli/mod.rs
NicholasLYang: Oh I recommended panic here because the previous message was to open a GitHub Issue and panic includes that advice

crates/turborepo-lib/src/commands/info.rs
NicholasLYang: Maybe we should make this opt-in instead of opt-out?

anthonyshew: I did think about it being opt-in. I can take this block out and ship this PR and work on a separate one for a `--include-turbo-vars` flag?

anthonyshew:
Opting to not print the env vars. If we feel like we need it later, we can print them with an optional flag.

NicholasLYang (approved):

PR #9378, opened 2024-11-04
 fix: Don't require a name for the root package.json.

anthonyshew:
### Description

There doesn't necessarily need to be a `name` field in the root package.json, since it won't get used for other work throughout the graph.

### Testing Instructions

I wrote a test and tested against `create-turbo` by removing the `name` field from the root package.json

### DIFF
--- a/crates/turborepo-repository/src/package_graph/mod.rs
+++ b/crates/turborepo-repository/src/package_graph/mod.rs
@@ -914,4 +914,17 @@ mod test {
             ))
         );
     }
+
+    #[tokio::test]
+    async fn test_does_not_require_name_for_root_package_json() {
+        let root =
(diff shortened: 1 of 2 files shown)


crates/turborepo-repository/src/package_graph/mod.rs
```diff
@@ -145,9 +145,21 @@ impl PackageGraph {
...
+                Some("") => {
+                    return Err(Error::PackageJsonMissingName(package_json_path));
+                }
+                None => {
+                    // We don't need to require a name for the root package.json.
+                    if package_json_path == self.repo_root.join_component("package.json") {
+                        continue;
+                    }
+
+                    let package_json_path = self.repo_root.resolve(info.package_json_path());
```
anthonyshew, line 159: Oops, shadowed it, pls rm.

chris-olszewski (approved):
LGTM. Only a small suggested change to the validation logic

PR #9383, opened 2024-11-05
fix(mfe): build internal package if present in graph

chris-olszewski:
### Description

Previously we were relying on the proxy already being built, for non-internal use cases this will be true, but for the internal case we need to make sure it is built before we try to start the proxy.

### Testing Instructions

- Edit `packages/micro-frontends/dist/bin/cli.cjs` and add a `throw new Error("oops")` to the top
- `turbo vercel-docs#dev` should fail due to the proxy script failing
- Now try with the PR `turbo`: `turbo_dev --skip-infer vercel-docs#dev`, you will see the MFE package being built before the proxy is started.

### DIFF
--- a/crates/turborepo-lib/src/turbo_json/mod.rs
+++ b/crates/turborepo-lib/src/turbo_json/mod.rs
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use struct_iterable::Iterable;
 use turbopath::AbsoluteSystemPath;
 use turborepo_errors::Spanned;
+use turborepo_micro_frontend::MICRO_FRONTENDS_PACKAGE_INTERNAL;
 use turborepo_repository::package_graph::ROOT_PKG_NAME;
 use turborepo_unescape::UnescapedString;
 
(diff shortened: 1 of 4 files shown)


crates/turborepo-unescape/src/lib.rs
```diff
@@ -57,6 +57,11 @@ impl From<UnescapedString> for String {
     }
 }
 
+impl From<String> for UnescapedString {
```
chris-olszewski, line 60: Required since there isn't any other way to construct an `UnescapedString`

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -619,6 +620,11 @@ impl TurboJson {
             TaskName::from("proxy"),
             Spanned::new(RawTaskDefinition {
                 cache: Some(Spanned::new(false)),
+                depends_on: needs_proxy_build.then(|| {
+                    Spanned::new(vec![Spanned::new(UnescapedString::from(format!(
+                        "{MICRO_FRONTENDS_PACKAGE_INTERNAL}#build"
```
tknickman, line 625: Don't love that we have to hardcode this task, but we can iterate on it.

tknickman (approved):

crates/turborepo-lib/src/turbo_json/mod.rs
chris-olszewski: I also don't love this, but for all other uses of the MFE package that don't have it as a workspace package this will be unnecessary.

PR #9398, opened 2024-11-06
chore(trace): fixes and improvements

NicholasLYang:
### Description

- Added more conditions
- Made emitting errors configurable
- Add inference for multiple node_modules


chris-olszewski (approved):
Giving this a trust stamp.

Are there any docs I should be reading that would give me a better understanding of the various `ResolveOptions`/`TsconfigOptions` methods and how they change behavior?

NicholasLYang:
[This](https://docs.rs/oxc_resolver/latest/oxc_resolver/struct.ResolveOptions.html) is an okay list, but tbh I've been digging into node/JS docs most of the time

PR #9399, opened 2024-11-06
docs: Add information for rules in `eslint-config-turbo` README.

anthonyshew:
## Hol' up

This is now out of sync, given #9426. I need to write a CI check that runs a diff for the README and content files so they stay updated together...

### Description

`eslint-plugin-turbo` had this information, but not the docs page on `turbo.build` nor the `eslint-config-turbo` README. Duplicated them in the missing places.

(Ideally, we would write these once and codegen them elsewhere, but that's a battle for another day.)


anthonyshew:
Handled in #9515

PR #9409, opened 2024-11-07
 feat: Wrap-around selection for TUI.

anthonyshew:
### Description

We previously didn't let you "wrap" your selection to the selected task if you were at the top or bottom of the task list. Now we do!

### Testing Instructions

Run some tasks with the TUI and:
-  Press ↑ when you have the top item in the list selected.
- Press ↓ when you have the bottom item in the list selected.

### DIFF
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -886,6 +886,8 @@ mod test {
         app.next();
         assert_eq!(app.scroll.selected(), Some(2), "scroll moves forwards");
         app.next();
+        assert_eq!(app.scroll.selected(), Some(0), "scroll stays in bounds");
+        app.previous();
         assert_eq!(app.scroll.selected(), Some(2), "scroll stays in bounds");
     }
 
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -153,7 +153,10 @@ impl<W> App<W> {
     pub fn previous(&mut self) {
         let num_rows = self.tasks_by_status.count_all();
         if num_rows > 0 {
-            self.selected_task_index = self.selected_task_index.checked_sub(1).unwrap_or(num_rows - 1);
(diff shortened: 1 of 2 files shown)


crates/turborepo-ui/src/tui/app.rs
```diff
@@ -886,6 +886,8 @@ mod test {
         app.next();
         assert_eq!(app.scroll.selected(), Some(2), "scroll moves forwards");
         app.next();
+        assert_eq!(app.scroll.selected(), Some(0), "scroll stays in bounds");
```
chris-olszewski, line 889:
```suggestion
        assert_eq!(app.scroll.selected(), Some(0), "scroll wraps");
```

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -142,21 +142,21 @@ impl<W> App<W> {
...
-        let i = match self.selected_task_index {
-            0 => 0,
-            i => i - 1,
-        };
-        self.selected_task_index = i;
-        self.scroll.select(Some(i));
-        self.has_user_scrolled = true;
+        let num_rows = self.tasks_by_status.count_all();
+        if num_rows > 0 {
+            self.selected_task_index = (self.selected_task_index + num_rows - 1) % num_rows;
```
chris-olszewski, line 156: This one is totally up to you if you think explicitly handling the underflow case is easier to reason about than leveraging modulos.
```suggestion
            self.selected_task_index = self.selected_task_index.checked_sub(1).unwrap_or(num_rows - 1);
```

chris-olszewski (approved):
LFG

crates/turborepo-ui/src/tui/app.rs
anthonyshew: Classic me not knowing methods I have available.

chris-olszewski: I'm an `Option` addict, I love how clean the control flow that methods provide is.

PR #9411, opened 2024-11-08
fix(examples): add private to with-shell-cmds

tknickman:
### Description

Fixes https://github.com/vercel/turborepo/issues/9150

yarn classic errors when using workspaces without setting private to true


anthonyshew (approved):
Thank you, yarn, very cool.

PR #9416, opened 2024-11-08
 feat: Detect WSL in `turbo info`.

anthonyshew:
### Description

We have had cases in the past where it would be nice to know if the user is using WSL. This adds best effort detection for printing out if you're using WSL.

Note that this isn't a 100% solution, according to [this superuser answer](https://superuser.com/questions/1749781/how-can-i-check-if-the-environment-is-wsl-from-a-shell-script/1749811#1749811). But, if you're doing something like is listed in the caveats...That's an extreme edge case.

### DIFF
--- a/crates/turborepo-lib/src/commands/info.rs
+++ b/crates/turborepo-lib/src/commands/info.rs
@@ -1,4 +1,9 @@
-use std::{env, io};
+use std::{
+    env,
+    fs::File,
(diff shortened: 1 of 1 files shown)


chris-olszewski (approved):
Worst case if `is_wsl` returns a faulty answer we have an additional back and forth with a user.

PR #9422, opened 2024-11-12
fix(affected): consider both source and destination as changed

chris-olszewski:
### Description

Previously we would only consider the destination file moved as changed and not the source. We use `diff-tree` instead of `diff` to track these moves. 

This PR also fixes an issue where `include_uncommitted: false` was not respected if `to_commit` wasn't present i.e. we would always return uncommitted changes. This did not affect users since this API isn't exposed to users and there's no way to set `include_uncommitted: false` without setting a `to_commit`.

### Testing Instructions

Added unit test where a file is moved between packages. Also added additional checks for our `changed_files` tests

### DIFF
--- a/crates/turborepo-scm/src/git.rs
+++ b/crates/turborepo-scm/src/git.rs
@@ -285,11 +285,17 @@ impl Git {
 
         let valid_from = self.resolve_base(from_commit, CIEnv::new())?;
 
-        let mut args = if let Some(to_commit) = to_commit {
-            vec!["diff", "--name-only", &valid_from, to_commit]
-        } else {
-            vec!["diff", "--name-only", &valid_from]
-        };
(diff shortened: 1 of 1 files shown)


crates/turborepo-scm/src/git.rs
```diff
@@ -652,10 +717,43 @@ mod tests {
...
+        // Test that uncommitted file in index is not marked as changed when not
+        // checking uncommitted
+        let files = changed_files(
+            repo_root.path().to_path_buf(),
+            turbo_root.to_path_buf(),
+            Some("HEAD"),
+            None,
+            false,
+        )?;
+        assert_eq!(files, HashSet::new());
```
chris-olszewski, line 729: Before this PR, this assertion would fail as `files` would contain `bar.js`

tknickman (approved):

PR #9440, opened 2024-11-15
examples: Update `basic` example to Next.js 15.

anthonyshew:
### Description

As the title says, updating to Next.js 15 in `basic` so `create-turbo` gives people the latest and greatest.

I've also taken the time to:
- Make the content of the applications Turborepo
- Added a `check-types` task for folks who are interested in running TypeScript checks.

### Testing Instructions

Our examples CI should make sure things are still building correctly. Would appreciate if someone could open it up in their editor and see if they get any squiggly lines, as well!


examples/basic/packages/ui/package.json
```diff
@@ -9,20 +9,22 @@
...
   "devDependencies": {
     "@repo/eslint-config": "workspace:*",
     "@repo/typescript-config": "workspace:*",
     "@turbo/gen": "^1.12.4",
     "@types/node": "^20.11.24",
     "@types/eslint": "^8.56.5",
-    "@types/react": "^18.2.61",
-    "@types/react-dom": "^18.2.19",
+    "@types/react": "npm:types-react@19.0.0-rc.1",
+    "@types/react-dom": "npm:types-react-dom@19.0.0-rc.1",
```
chris-olszewski, line 22: These do seem wrong. I don't think we want to be using https://github.com/eps1lon/DefinitelyTyped

chris-olszewski (changes requested):

examples/basic/packages/ui/package.json
anthonyshew: Oof, good catch.

chris-olszewski (approved):
LFG

PR #9445, opened 2024-11-15
fix(cli): error on out of place run args

chris-olszewski:
### Description

Currently we have a bug where run args can be specified incorrectly and we'll silently ignore them.
 - Specified for non-run commands e.g. `turbo --no-daemon watch`
 - Specified before a subcommand e.g. `turbo --filter=foo run build` `turbo --filter=foo watch build` which will silently ignore any arguments provided before the subcommand.

This PR makes these explicit failures instead of silently ignoring them. There is a solution where we attempt to "merge" args provided on either side of the subcommand, but that would require us to switch to `clap`'s builder interface which would make our 80 line top level CLI interface ~1k lines. I do not think this is worth the time to go about that refactor to allow for awkward syntax.

⚠️ This will very possibly break user workflows, but I think this is better than our current ignoring of user provided flags. We could warn here, but I'm afraid that those warnings would go unseen.

I suggest reviewing each commit individually as there was some required refactoring to unit test this behavior.

### Testing Instructions

Added unit tests

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -513,11 +513,30 @@ impl Args {
     }
 
     fn validate(&self) -> Result<(), clap::Error> {
-        if self.run_args.is_some() && !matches!(self.command, None | Some(Command::Run { .. })) {
+        if self.run_args.is_some()
+            && !matches!(
+                self.command,
+                None | Some(Command::Run { .. }) | Some(Command::Config)
+            )
+        {
             let mut cmd = Self::command();
             Err(cmd.error(
                 clap::error::ErrorKind::UnknownArgument,
                 "Cannot use run arguments outside of run command",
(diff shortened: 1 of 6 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -493,6 +479,69 @@ impl Args {
...
+            .map(|(_, input_token)| input_token);
+
+        (is_single_package, single_package_free)
+    }
+
+    fn validate(&self) -> Result<(), clap::Error> {
+        if self.run_args.is_some()
+            && !matches!(
+                self.command,
+                None | Some(Command::Run { .. }) | Some(Command::Config)
```
chris-olszewski, line 519: We special list `config` since we use it for tests.

tknickman (approved):

NicholasLYang (approved):

PR #9469, opened 2024-11-20
add vim keybindings for cycling up and down

kevinanielsen:
### Description

I updated the inputs for the tui to allow for using the 'j'  key to cycle down in tasks and 'k' for cycling up in tasks.

### Testing Instructions

Run a turbo repo job and try using 'j' and 'k' to cycle between running tasks.


kevinanielsen:
I did not notice a PR that already exists with these changes and more :)

PR #9478, opened 2024-11-21
refactor(tui): Improve aesthetics for helper text.

anthonyshew:
### Description

Just wanting to spruce things up a bit.

- Left-aligned the text so that it bounces around less when changing states. This is considered a generally good practice in UI design.
- Re-writing text chunks for simplicity

In the process of adjusting the text with a comma as a separator, I realized I could refactor the logic for which text to render to be a little easier to read. Isn't it great when the simplicity of the visual design is reflected in the code?!

### Testing Instructions

You'll want to put the TUI into different states to see them all. Matrix is:
- Task list open/closed
- Text selected/not selected
- Task being interacted with/ not interacted with

A sample of the spirit of the changes is below.

Before
CleanShot 2024-11-20 at 23 05 58@2x

After
CleanShot 2024-11-20 at 23 06 38@2x

### DIFF
--- a/crates/turborepo-ui/src/tui/pane.rs
+++ b/crates/turborepo-ui/src/tui/pane.rs
@@ -8,7 +8,7 @@ use tui_term::widget::PseudoTerminal;
 use super::{app::LayoutSections, TerminalOutput};
 
 const FOOTER_TEXT_ACTIVE: &str = "Ctrl-z - Stop interacting";
-const FOOTER_TEXT_INACTIVE: &str = "Enter - Interact";
+const FOOTER_TEXT_INACTIVE: &str = "i - Interact";
 const HAS_SELECTION: &str = "c - Copy selection";
 const TASK_LIST_HIDDEN: &str = "h - Show task list";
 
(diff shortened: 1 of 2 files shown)


crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -39,24 +39,29 @@ impl<'a, W> TerminalPane<'a, W> {
...
+            if !self.has_sidebar {
+                messages.push(TASK_LIST_HIDDEN);
+            }
+
+            if self.terminal_output.has_selection() {
+                messages.push(HAS_SELECTION);
+            }
+
+            // Spaces are used to pad the footer text for aesthetics
+            ["   ".into(), messages.join(", ")].join("")
```
chris-olszewski, line 54: I think directly formatting this instead of `.join("")`ing an array is cleaner. Also one less allocation (not that it matters)
```suggestion
            format!("   {}", messages.join(", "))
```

chris-olszewski (approved):

crates/turborepo-ui/src/tui/pane.rs
anthonyshew: And you're right. These are the dumb things I write after midnight.

PR #9483, opened 2024-11-21
chore(ci): upgrade GitHub Actions to Node 20

chris-olszewski:
### Description

Upgrade all of our various actions to versions that use Node 20 since Node 16 is getting shut down December 5

Note this does not switch us off of [get-diff-action](https://github.com/technote-space/get-diff-action/) which has been deprecated so we will need to find a new solution.

### Testing Instructions

CI on this PR

Test Release: 
[workflow](https://github.com/vercel/turborepo/actions/runs/11955946798) <- failed because I wrote `actions/checkout@4` instead of `actions/checkout@v4` on the last step
[workflow with fix](https://github.com/vercel/turborepo/actions/runs/11956800941)

You should notice the lack of Node 16 warnings on the workflow summary page: https://github.com/vercel/turborepo/actions/runs/11954973491

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -25,7 +25,7 @@ runs:
       uses: pnpm/action-setup@v4
 
     - name: Setup Node.js
-      uses: actions/setup-node@v3.6.0
+      uses: actions/setup-node@v4
       with:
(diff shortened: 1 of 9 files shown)


anthonyshew (approved):
Important work. Thank you!

PR #9486, opened 2024-11-21
feat(trace): resolve incorrect extension

NicholasLYang:
### Description

Invalid extensions are unfortunately too common in imports. Therefore if we can't resolve with the extension, we strip it and try again.

### Testing Instructions

Added a test where we try to import with the wrong extension

### DIFF
--- a/crates/turbo-trace/src/tracer.rs
+++ b/crates/turbo-trace/src/tracer.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, sync::Arc};
 
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
 use globwalk::WalkType;
 use miette::{Diagnostic, Report, SourceSpan};
 use oxc_resolver::{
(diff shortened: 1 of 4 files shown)


crates/turbo-trace/src/tracer.rs
```diff
@@ -211,15 +211,36 @@ impl Tracer {
                     debug!("built in: {:?}", err);
                 }
                 Err(err) => {
-                    // Try to resolve the import as a type import via `@/types/<import>`
-                    let type_package = format!("@types/{}", import);
-                    let resolved_type_import = resolver
-                        .resolve(file_path, type_package.as_str())
+                    if !import.starts_with(".") {
```
chris-olszewski, line 214: Is this a check against `./`?

crates/turbo-trace/src/tracer.rs
```diff
@@ -211,15 +211,36 @@ impl Tracer {
...
+
+                        if let Some(resolved_type_import) = resolved_type_import {
+                            debug!("resolved type import succeeded");
+                            files.push(resolved_type_import);
+                            continue;
+                        }
+                    }
+
+                    // Also try without the extension just in case the wrong extension is used
+                    let without_extension = Utf8Path::new(import).with_extension("");
```
chris-olszewski, line 231: Just calling out that this will convert something like `types.d.ts` to `types.d` [source](https://docs.rs/camino/latest/camino/struct.Utf8PathBuf.html#method.extension), is that the desired behavior?

chris-olszewski (approved):
Two small questions

crates/turbo-trace/src/tracer.rs
NicholasLYang: Yeah. I don't *think* packages can start with `.`?

crates/turbo-trace/src/tracer.rs
NicholasLYang: So I originally had an explicit check for `types.d.ts` that stripped it all the way back to `types` but when I ran this on the other repo, that didn't fix any errors. I suspect oxc-resolver also treats `types.d` as the name, so this should work just fine

PR #9493, opened 2024-11-22
feat(mfe): allow @vercel/microfrontends to be considered a MFE package dependency

mknichel:
### Description

This PR allows `@vercel/microfrontends` to be considered a MFE dependency in addition to `@vercel/micro-frontends-internal`.


tknickman (approved):
You're too fast!

PR #9502, opened 2024-11-22
feat: update `eslint-config-turbo` to support ESLint Flat Config.

anthonyshew:
### Description

Adds ESLint Flat Config Support to `eslint-config-turbo`. Users will be able to use:
```
import turboConfig from "eslint-config-turbo/flat"

export default [
  ...turboConfig
]
```

This is meant to be backwards compatible with ESLint v8, leaving the `extends` key in place for that purpose.

### Testing Instructions

Difficult to test this, so we'll be cutting a canary shortly after to make sure.

### DIFF
--- a/packages/eslint-config-turbo/flat/index.js
+++ b/packages/eslint-config-turbo/flat/index.js
@@ -1,0 +1,12 @@
+const plugin = require("eslint-plugin-turbo");
+
+module.exports = [
+  {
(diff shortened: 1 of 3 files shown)


chris-olszewski (approved):
:shipit:

tknickman (approved):

PR #9504, opened 2024-11-22
feat: sibling tasks

chris-olszewski:
### Description

This PR does 2 things:
 - Adds a `sibling` attribute for task definitions that has the ability to bring another task into the graph without depending on it. This is currently only an attribute that can be set by us inside Rust, but I hope this use case proves useful enough to eventually make this a public API.
 - Switches MFE proxy injection over to use `siblings` where each dev task gets a `"siblings": ["mfe-pkg#proxy"]` to ensure the proxy gets run.

A sibling relationship acts similar to `dependsOn` except that it will not wait for the task to exit before starting and the sibling's task hash will not affect the current task's hash.

Currently it's very hard to get `web#proxy` to run if the user runs a command like `turbo dev --filter=docs` as just adding `web` to the filter will mean `web#dev` gets picked up which isn't what we desire. As you can see in `run/builder.rs` this also removes the need of microfrontend knowledge from building the task graph.

### Testing Instructions

Added unit tests for sibling behavior as well as unit tests for MFE task injection.

Manual testing for the MFE behavior to make sure proxy still gets picked up.

### DIFF
--- a/crates/turborepo-lib/src/micro_frontends.rs
+++ b/crates/turborepo-lib/src/micro_frontends.rs
@@ -1,9 +1,10 @@
 use std::collections::{HashMap, HashSet};
 
+use itertools::Itertools;
 use tracing::warn;
 use turbopath::AbsoluteSystemPath;
 use turborepo_micro_frontend::{Config as MFEConfig, Error, DEFAULT_MICRO_FRONTENDS_CONFIG};
-use turborepo_repository::package_graph::PackageGraph;
+use turborepo_repository::package_graph::{PackageGraph, PackageName};
 
 use crate::run::task_id::TaskId;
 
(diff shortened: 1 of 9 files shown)


tknickman (approved):

crates/turborepo-lib/src/micro_frontends.rs
```diff
@@ -64,4 +88,178 @@ impl MicroFrontendsConfigs {
...
+
+    pub fn update_turbo_json(
+        &self,
+        package_name: &PackageName,
+        turbo_json: Result<TurboJson, config::Error>,
+    ) -> Result<TurboJson, config::Error> {
+        // If package either
+        // - contains the proxy task
+        // - a member of one of the microfrontends
+        // then we need to modify it's task definitions
```
NicholasLYang, line 100:
```suggestion
        // then we need to modify its task definitions
```

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -631,6 +646,22 @@ impl TurboJson {
             }),
         );
     }
+
+    /// Adds a sibling relationship from task to sibling
+    pub fn with_sibling(&mut self, task: TaskName<'static>, sibling: &TaskName) {
+        if self.extends.is_empty() {
```
NicholasLYang, line 652: What's the reason for this? Do we only add siblings to workspace configs?

NicholasLYang (commented):
Just one question around `extends`

crates/turborepo-lib/src/turbo_json/mod.rs
chris-olszewski: At this moment, yes.

NicholasLYang: Could we document this then?

PR #9505, opened 2024-11-22
docs: update README for `eslint-config-turbo` with Flat Config instructions.

anthonyshew:
### Description

Update README to document instructions for Flat Config with `eslint-config-turbo`.


anthonyshew:
Handilng in #9515

PR #9512, opened 2024-11-25
feat(tui): persistent preferences

anthonyshew:
### Description

Users have expressed that they find themselves repeatedly having to select out the task that they want to use. This ends up being annoying when invoking `turbo`, shutting down, invoking `turbo`, shutting down...You end up having to reselect the task you care about every re-invoke.

To improve on this, we're going to hold some state in `.turbo/preferences/tui.json` so that users pick up where they left off from their previous TUI invocation. State we'll keep around is:
- The selected task name (Responsible for the "pinned-ness" portion of state)
- Toggled state of task list visibility

### Testing Instructions

Wrote tests but you should also try it out by hand. You should be able to run `turbo` invocations, pick out tasks, and see the state persist across invocations. You also should be able to delete the `.turbo/preferences/tui.json` file and it'll feel like the first invocation (no preferences).

### Small note
One thing that I noticed while hand-testing is that this implementation tolerates extra fields fine, but doesn't tolerate the wrong types for a field. It will crash for in the latter scenario.

Personally, I think this is fine. You'd have to purposefully go and edit the file to get the wrong types into there.


crates/turborepo-ui/src/tui/input.rs
```diff
@@ -85,7 +86,6 @@ fn translate_key_event(options: InputOptions, key_event: KeyEvent) -> Option<Eve
                 restore_scroll: true,
             })
         }
-        KeyCode::Char('h') => Some(Event::ToggleSidebar),
```
anthonyshew, line 88: Needs new PR: Just moving it to the right section.

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -767,9 +782,31 @@ fn update(
         }
         Event::Up => {
             app.previous();
+            let _ = preferences::Preferences::update_preference(
```
anthonyshew, line 785: TODO: Handle result inside method.

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -1055,7 +1125,12 @@ mod test {
 
     #[test]
     fn test_interact() -> Result<(), Error> {
-        let mut app: App<Vec<u8>> = App::new(100, 100, vec!["a".to_string(), "b".to_string()]);
+        let mut app: App<Vec<u8>> = App::new(
+            100,
+            100,
+            vec!["a".to_string(), "b".to_string()],
+            &AbsoluteSystemPathBuf::default(),
```
anthonyshew, line 1132: TODO: Make a tempdir

crates/turborepo-lib/src/run/mod.rs
```diff
@@ -262,8 +262,9 @@ impl Run {
 
         let (sender, receiver) = TuiSender::new();
         let color_config = self.color_config;
+        let repo_root = self.repo_root.clone();
```
chris-olszewski, line 265: We should parse the options here and pass them in so we can catch any config/io errors early on

crates/turborepo-ui/src/tui/preferences.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+}
+
+const TUI_PREFERENCES_PATH_COMPONENTS: &[&str] = &[".turbo", "preferences", "tui.json"];
+
+fn read_json(path: &AbsoluteSystemPathBuf) -> Preferences {
+    File::open(path)
+        .ok()
+        .and_then(|file| from_reader(BufReader::new(file)).ok())
```
chris-olszewski, line 41: Not impactful due to how small this JSON file will be, but we should usually always prefer [serde_json::from_str](https://github.com/serde-rs/json/issues/160).

crates/turborepo-ui/src/tui/preferences.rs
```diff
@@ -0,0 +1,132 @@
...
+        let preferences_file = repo_root.join_components(&[".turbo", "preferences", "tui.json"]);
+
+        fs::create_dir_all(preferences_dir.as_std_path())?;
+
+        let mut json: Value = if preferences_file.exists() {
+            let json_string = fs::read_to_string(&preferences_file)?;
+            serde_json::from_str(&json_string)?
+        } else {
+            json!({})
+        };
```
chris-olszewski, line 62: `exists()` generally is a smell since it can lead to TOCTOU bugs. We can use [`read_existing_to_string`](https://github.com/vercel/turborepo/blob/9fad9cb81e41f5cb9cf8cccf3ee1f279987d6478/crates/turborepo-lib/src/commands/link.rs#L235) to handle defaults.
```suggestion
        let mut json: Value = preferences_file.read_existing_to_string()?.unwrap_or_else(|| "{}".to_string());
```

crates/turborepo-ui/src/tui/preferences.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn update_preference(
+        repo_root: &AbsoluteSystemPathBuf,
+        field: PreferenceFields,
+        new_value: Value,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        // TODO: Clean these up, should be taken from constants
+        let preferences_dir = repo_root.join_components(&[".turbo", "preferences"]);
+        let preferences_file = repo_root.join_components(&[".turbo", "preferences", "tui.json"]);
+
+        fs::create_dir_all(preferences_dir.as_std_path())?;
```
chris-olszewski, line 55: Generally we should prefer methods on the `turbopath`s types since we have nicer error messages. (By default Rust io errors don't include paths in them which can lead to frustrating error messages).

```suggestion
        preferences_file.ensure_dir()?;
```

CONTRIBUTING.md
```diff
@@ -9,7 +9,7 @@ Thank you for your interest in contributing to Turborepo!
 - [Manually testing `turbo`](#manually-testing-turbo)
   - [Repositories to test with](#repositories-to-test-with)
 - [Debugging tips](#debugging-tips)
-  - [Verbose logging](#verbose-logging)  
+  - [Verbose logging](#verbose-logging)
```
anthonyshew, line 12: Sure.

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -730,6 +761,7 @@ fn cleanup<B: Backend + io::Write>(
     )?;
     let tasks_started = app.tasks_by_status.tasks_started();
     app.persist_tasks(tasks_started)?;
+    app.preferences.flush_to_disk().ok();
```
anthonyshew, line 764: @chris-olszewski, this was smart. Kudos.

crates/turborepo-ui/src/tui/mod.rs
```diff
@@ -36,4 +37,6 @@ pub enum Error {
     Stdin { name: String, e: std::io::Error },
     #[error(transparent)]
     Io(#[from] std::io::Error),
+    #[error("Unable to persist preferences")]
```
anthonyshew, line 40: Would it make sense to do:

```suggestion
    #[error("Unable to persist preferences. Please file a bug report.")]
```

crates/turborepo-lib/src/run/mod.rs
chris-olszewski, line 265: We can just borrow here instead of cloning
```suggestion
        let repo_root = &self.repo_root;
```

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -126,6 +131,18 @@ impl<W> App<W> {
         })
     }
 
+    fn update_sidebar_toggle(&mut self) {
```
chris-olszewski, line 134: I feel like toggle implies an update, up to you if this is better name 🤷 
```suggestion
    fn toggle_sidebar(&mut self) {
```

crates/turborepo-ui/src/tui/preferences.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    pub fn active_task(&self) -> Option<&str> {
+        let active_task = self.config.active_task.as_deref()?;
+        Some(active_task)
+    }
+
+    pub fn set_active_task(&mut self, value: Option<String>) -> Result<(), Error> {
+        self.config.active_task = value;
+        Ok(())
+    }
```
chris-olszewski, line 47: No need to lie to the caller that this could fail. We can propagate this change up to callers that now no longer need to handle the error case.
```suggestion
    pub fn set_active_task(&mut self, value: Option<String>) {
        self.config.active_task = value;
    }
```

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -126,6 +131,18 @@ impl<W> App<W> {
...
+        let value = !self.preferences.is_task_list_visible();
+        self.preferences.set_is_task_list_visible(Some(value));
+    }
+
+    fn update_task_selection_pinned_state(&mut self) -> Result<(), Error> {
+        // Preferences assume a pinned state when there is an active task.
+        // This `None` creates "un-pinned-ness" on the next TUI startup.
+        self.preferences.set_active_task(None)?;
+        Ok(())
+    }
```
chris-olszewski, line 144: I find this name a little confusing. Does `clear_pinned_task` express the behavior better?
```suggestion
    fn update_task_selection_pinned_state(&mut self) {
        // Preferences assume a pinned state when there is an active task.
        // This `None` creates "un-pinned-ness" on the next TUI startup.
        self.preferences.set_active_task(None);
    }
```

chris-olszewski (approved):
LFG

PR #9501, opened 2024-11-22
chore: update `basic` example to ESLint v9 and Flat Configurations.

anthonyshew:
### Description

This PR converts our `basic` example to ESLint v9 with Flat Configurations.

I may have a few gray hairs now, but I'm very excited for the community to have this.

### Testing Instructions

Use `npx create-turbo@canary -e https://github.com/vercel/turborepo/tree/shew-409f3/examples/basic basic-eslint-9` to try it out.

- Pull it down, make sure you see diagnostics where you'd expect in your editor. Including for `eslint-config-turbo`, most importantly!
- I know the diff is big, but there may be vestigials from my experimenting lying around. Please use a finely toothed comb! (I've looked myself but the more eyes, the better.)

PR #9516, opened 2024-11-25
chore(examples): remove spare dependencies from basic example

anthonyshew:
### Description

I left some unneeded dependencies behind when I was upgrading this example in #9501. This gets rid of them.

### Testing Instructions

I've verified that ESLint violations get reported as expected in my editor and when running `turbo lint`. You can also clone with `npx create-turbo@canary -e https://github.com/vercel/turborepo/tree/shew-0f804/examples/basic basic-eslint-9` to have a play with it.

### DIFF
--- a/examples/basic/apps/docs/package.json
+++ b/examples/basic/apps/docs/package.json
@@ -22,8 +22,6 @@
     "@types/node": "^20",
     "@types/react": "18.3.1",
     "@types/react-dom": "18.3.0",
-    "eslint": "^9.15.0",
-    "eslint-config-next": "15.0.3",
     "typescript": "5.5.4"
   }
 }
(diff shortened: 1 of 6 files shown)


chris-olszewski (changes requested):
If we're shipping a pnpm 9 lockfile we should bump the default package manager to match the version used to create the lockfile: https://github.com/vercel/turborepo/blob/9db0da44c87ab5f9c9060884dd28ff2638d6d84e/examples/basic/package.json#L15

chris-olszewski (approved):

vincenthsh:
I noticed the examples do not use `.nvmrc` or `.node-version` or `mise.toml` or `.tool-versions` (lol), however on Node v20, eslintv9 requires `20.9.0` or above

anthonyshew:
[The `engines` field](https://github.com/vercel/turborepo/blob/a1db09a70b03e75097dafd65dddd775567b628dd/examples/basic/package.json#L17) covers this but looks like we're still keeping a minimum of 18. We can update this!

PR #9534, opened 2024-11-28
fix(codemod): comment tmpdir for package managers

tomcastro:
### Description

When using Yarn 4, the default is to use a local version to your repository. Codemod runs package managers from a tmp directory, which causes Yarn to think there's no local version and uses global version instead. This is fixed when removing the line to run from tmp dir, but needs to be tested if it works with other package managers/projects.

https://github.com/vercel/turborepo/pull/8076


anthonyshew:
Closing this one because I asked for it in conversation during 8076. I'm fairly confident we're on the right track now and we can toss this.

Really appreciate your help on the way to sorting this out. Thank you!

PR #9542, opened 2024-11-29
chore(tui): dim pane text helpers and remove dead code.

anthonyshew:
### Description

I started noticing that my logs and the text help were looking quite similar. This differentiates the two so its more obvious at a glance.

### Testing Instructions

CleanShot 2024-11-29 at 12 52 40

### DIFF
--- a/crates/turborepo-ui/src/tui/pane.rs
+++ b/crates/turborepo-ui/src/tui/pane.rs
@@ -1,5 +1,5 @@
 use ratatui::{
-    style::{Modifier, Style},
+    style::{Modifier, Style, Stylize},
     text::Line,
     widgets::{Block, Widget},
 };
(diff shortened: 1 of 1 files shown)


crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -34,12 +34,8 @@ impl<'a, W> TerminalPane<'a, W> {
         }
     }
 
-    fn highlight(&self) -> bool {
```
anthonyshew, line 37: Dead code! There's no border to highlight anymore.

crates/turborepo-ui/src/tui/pane.rs
```diff
@@ -77,13 +73,12 @@ impl<'a, W> Widget for &TerminalPane<'a, W> {
     {
         let screen = self.terminal_output.parser.screen();
         let block = Block::default()
-            .title(self.terminal_output.title(self.task_name))
-            .title_bottom(self.footer())
-            .style(if self.highlight() {
```
anthonyshew, line 82: We don't do these yellow highlights. The only place in the program we do yellow highlights now is in the task table (not here), so we can remove this.

dimitropoulos (approved):
nice one!

PR #9545, opened 2024-11-29
feat(tui): popup to list keybinds

anthonyshew:
### Description

We're adding a host of binds to make the TUI even more useful, and listing them out with our current strategy is going to prove cumbersome quickly. Terminal real estate is valuable!

With this PR, we're putting in a popup that you can use to quickly get a glance of what's available to you.

### Testing Instructions

Try it out!

### Questions
- [ ] (opinions wanted) Should I block all other inputs when this menu is open? (For what it's worth, I'm seeing things play nicely with other inputs left unblocked.)
- [ ] (non-blocking) Is there a reasonable way to get the list items from `input.rs`?


crates/turborepo-ui/src/tui/popup.rs
```diff
@@ -0,0 +1,60 @@
+use ratatui::{
+    layout::{Constraint, Flex, Layout, Rect},
+    text::Line,
+    widgets::{Block, List, ListItem, Padding},
+};
+
+use super::size::SizeInfo;
+
+const BIND_LIST_ITEMS: [&str; 11] = [
```
dimitropoulos, line 9: issue: can we find some way to derive this information from the data where the keybindings are defined?  Perhaps that means adding an "info" field, for example.

crates/turborepo-ui/src/tui/popup.rs
```diff
@@ -0,0 +1,60 @@
...
+    "m - Toggle this help popup",
+    "↑ or j - Select previous task",
+    "↓ or k - Select next task",
+    "h - Toggle task list visibility",
+    "/ - Filter tasks to search term",
+    "ESC - Clear filter",
+    "i - Interact with task",
+    "CTRL+z - Stop interacting with task",
```
dimitropoulos, line 17: is it `CTRL` on windows and linux and mac?

crates/turborepo-ui/src/tui/popup.rs
anthonyshew: That's one of my TODOs. 😄 But I'm hoping so. These all get defined [here](https://github.com/vercel/turborepo/blob/e901df637d9fdee3a91907a2213d7f8cc01a45ef/crates/turborepo-ui/src/tui/input.rs#L54) and I wouldn't know how to translate that into a map I can use for these purposes.

crates/turborepo-ui/src/tui/popup.rs
anthonyshew: I can at least confirm its `CTRL` on Mac. Lines like [this one](https://github.com/vercel/turborepo/blob/e901df637d9fdee3a91907a2213d7f8cc01a45ef/crates/turborepo-ui/src/tui/input.rs#L63) and [this one](https://github.com/vercel/turborepo/blob/e901df637d9fdee3a91907a2213d7f8cc01a45ef/crates/turborepo-ui/src/tui/input.rs#L71) lead me to believe they should be cross-platform but @chris-olszewski to confirm as I'm figuring he considered this when setting things up.

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -97,6 +100,8 @@ impl<W> App<W> {
             task_list_scroll: TableState::default().with_selected(selected_task_index),
             selected_task_index,
             has_sidebar: true,
+            has_user_scrolled: has_user_interacted,
```
chris-olszewski, line 103:
```suggestion
```

chris-olszewski:
> Should I block all other inputs when this menu is open? (For what it's worth, I'm seeing things play nicely with other inputs left unblocked.)

I think not blocking inputs is good. Do you think closing the popup on key interactions would be intuitive?

> (non-blocking) Is there a reasonable way to get the list items from input.rs?

Not easily. If we swapped from a `match` to a long `if / else if / else` statement and introduce another wrapper on the `KeyEvent` type we could probably do it, but I would not block on this.

crates/turborepo-ui/src/tui/popup.rs
anthonyshew: Chris confirmed for me [here](https://github.com/vercel/turborepo/pull/9545#issuecomment-2524306714) that it's worth a separate PR.

anthonyshew:
> Do you think closing the popup on key interactions would be intuitive?

I played around with this, and I was discovering more inventive workflows than ever with keeping it open. It felt like I could get inspired about my next move more. However, I could also see this as "I checked the binds for the thing I wanted and I pushed it. Now go away." as a more experienced user. I don't think I have a strong opinion; I could see either behavior being useful depending on who you are. If I were forced to pick, I would keep it as-is (not auto-close).

>  If we swapped from a match to a long if / else if / else statement

Agree that this shouldn't block and hard to say if its YAGNI. I can explore options in follow-up.

PR #9548, opened 2024-12-01
feat: print detected package manager version in `create-turbo` options

anthonyshew:
### Description

Prints the package manager version that we're detecting on the user's machine.

CleanShot 2024-12-02 at 08 53 20

Goals here are:
- More information for the user
- Likely easier debugging for us. A report of something going wrong with `create-turbo` is likely to include more information. This helped me while trying to figure out what's up with #8076.

### Hesitation
I could see the argument that our users didn't really request this information at this time, so it's just noise?

### Testing Instructions

👀


anthonyshew:
Closing, we discussed and didn't like it.

PR #9558, opened 2024-12-04
chore(tui): add test for key handling

chris-olszewski:
### Description

Before I realized that https://github.com/vercel/turborepo/pull/9543 already fixed https://github.com/vercel/turborepo/issues/9553 I started working on a unit test for this section of the code. Putting it up just so writing future tests should be easier.

### Testing Instructions

Unit test runs and passes.

### DIFF
--- a/crates/turborepo-ui/src/tui/input.rs
+++ b/crates/turborepo-ui/src/tui/input.rs
@@ -420,3 +420,48 @@ fn encode_modifiers(mods: KeyModifiers) -> u8 {
     }
     number
 }
+
+#[cfg(test)]
+mod test {
+    use std::{mem, sync::OnceLock};
+
+    use crossterm::event::{KeyCode, KeyEvent};
+    use test_case::test_case;
+
+    use super::*;
+    use crate::tui::{search::SearchResults, task::TasksByStatus};
+
+    fn search() -> &'static LayoutSections {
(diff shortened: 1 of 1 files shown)


crates/turborepo-ui/src/tui/input.rs
```diff
@@ -420,3 +420,48 @@ fn encode_modifiers(mods: KeyModifiers) -> u8 {
...
+    #[test_case(in_find(), H, Some(Event::SearchEnterChar('h')) ; "h while searching")]
+    // Note: This only checks event variants not any data contained in the variant
+    fn test_translate_key_event_variant(
+        opts: InputOptions,
+        key_event: crossterm::event::KeyEvent,
+        expected: Option<Event>,
+    ) {
+        match (translate_key_event(opts, key_event), expected) {
+            (Some(actual), Some(expected)) => {
+                assert_eq!(mem::discriminant(&actual), mem::discriminant(&expected));
```
chris-olszewski, line 460: [`mem::discriminant`](https://doc.rust-lang.org/stable/std/mem/fn.discriminant.html) takes an enum and gives each variant an ID that can be compared. This is a workaround for the fact that currently our events can not be compared. 

We might want to eventually add that `#[derive(PartialEq)]` to `Event`, but the fact that our events contain file descriptors or channels makes it non-obvious as to what behavior should be in that scenario.

anthonyshew (approved):

PR #9560, opened 2024-12-04
feat(mfe): enable proxy for no turbo json

chris-olszewski:
### Description

This PR adds the MFE support for repositories that do not contain a `turbo.json`. All this boils down to is:
 - Slight refactor for constructing MFE configs to allow for us to test this without hitting disk
 - Going down the "no turbo.json" codepath even if not explicitly enabled as long as MFE configs are present
 - Update generated `turbo.json` with `proxy` tasks injected

### Testing Instructions

Added unit test for sans-`turbo.json`, but containing a MFE config.

Manual test on a MFE test repo w/o a `turbo.json`


tknickman (approved):
❤️

PR #9562, opened 2024-12-04
refactor(opts): remove config from opts

NicholasLYang:
### Description

Remove the config field from `Opts` entirely. I'm doing this because the config data already gets partially moved into `Opts`, so by keeping the config around, we have multiple sources for the same data. Now `Opts` is the fully transformed, normalized config for a run.

You can review this commit by commit if you like

### DIFF
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -2,7 +2,7 @@ use std::backtrace;
 
 use camino::Utf8PathBuf;
 use thiserror::Error;
-use turbopath::AnchoredSystemPathBuf;
+use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf, AnchoredSystemPathBuf};
 use turborepo_api_client::APIAuth;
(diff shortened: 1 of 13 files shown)


crates/turborepo-lib/src/opts.rs
```diff
@@ -43,25 +43,36 @@ pub enum Error {
...
     pub upload_timeout: u64,
     pub token: Option<String>,
     pub team_id: Option<String>,
     pub team_slug: Option<String>,
+    pub login_url: String,
+    pub preflight: bool,
+}
+
+#[derive(Debug, Clone)]
+pub struct ConfigOpts {
```
chris-olszewski, line 59: Names are hard, but I don't think we should name this `ConfigOpts`: `ConfigurationOptions` populate an `Opts` struct which contains `ConfigOpts` is confuses me. Maybe `RepoOpts` since package manager/`turbo.json` are related to the whole repo?

chris-olszewski (approved):
Very minor comments around naming/visibility

PR #9566, opened 2024-12-05
docs: clarify default daemon usage

anthonyshew:
### Description

The description on these flags wasn't clear enough about what the default is that the flag is meant to be switching on/off. Added it!

### Testing Instructions

👀

### DIFF
--- a/docs/repo-docs/reference/run.mdx
+++ b/docs/repo-docs/reference/run.mdx
@@ -410,7 +410,7 @@ turbo run dev --no-cache
 
 `turbo` can run a background process to pre-calculate values used for determining work that needs to be done. This standalone process (daemon) is an optimization, and not required for proper functioning of `turbo`.
 
-The default daemon usage is set for your repository using [the `daemon` field in `turbo.json`](/repo/docs/reference/configuration#daemon). Passing `--daemon` instructs `turbo` to use the standalone process, while `--no-daemon` instructs `turbo` to avoid using or creating the standalone process.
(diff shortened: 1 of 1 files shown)


docs/repo-docs/reference/run.mdx
```diff
@@ -410,7 +410,7 @@ turbo run dev --no-cache
 
 `turbo` can run a background process to pre-calculate values used for determining work that needs to be done. This standalone process (daemon) is an optimization, and not required for proper functioning of `turbo`.
 
-Passing `--daemon` instructs `turbo` to use the standalone process, while `--no-daemon` instructs `turbo` to avoid using or creating the standalone process.
+The default daemon usage is set for your repository using [the `daemon` field in `turbo.json`](/repo/docs/reference/configuration#daemon). Passing `--daemon` instructs `turbo` to use the standalone process, while `--no-daemon` instructs `turbo` to avoid using or creating the standalone process.
```
chris-olszewski, line 413: When `--daemon` is passed in requires that `turbo` uses the daemon for various operations. The command will fail if we're unable to start/connect with the daemon. I feel like "requires" describes this behavior a little better than "instructs", but I defer to you.

```suggestion
The default daemon usage is set for your repository using [the `daemon` field in `turbo.json`](/repo/docs/reference/configuration#daemon). Passing `--daemon` requires `turbo` to use the standalone process, while `--no-daemon` instructs `turbo` to avoid using or creating the standalone process.
```

chris-olszewski (approved):
Minor suggestion around `--daemon` behavior description.

PR #9592, opened 2024-12-08
chore:  remove old Turbopack files

anthonyshew:
### Description

These files all appear to be from when Turbopack was also in this repo, so cleaning them up.


chris-olszewski (approved):
LGTM, just need to update our Rust lint to not call ast-grep

PR #9593, opened 2024-12-08
ci(examples): test every package manager in Support Policy on `basic`

anthonyshew:
### Description

Recently, I updated external dependencies for the `basic` example and tested it with pnpm. Everything looked good - but our CI _only_ tests `basic` for pnpm. I hadn't realized that I broke one of the other package managers.

Instead of relying on making sure that we hand-test every package manager, let's test each one in CI.

This PR also removes a few of the examples from the test suite since we are not maintaining them. Because they are community maintained, we won't rely on them for our CI.

### Testing Instructions

CI 🙏


turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -19,13 +19,15 @@ if [ -z "$package_manager" ]; then
   exit 1
 fi
 
+echo "node --version: $(node --version)"
+
 # Use the right command for each package manager
 if [ "$package_manager" == "npm" ]; then
-  package_manager_command="npm install"
+  package_manager_command="node ../../../packages/turbo-workspaces/dist/cli.js convert . npm --ignore-unchanged-package-manager && npm install"
```
anthonyshew, line 26: To answer your question, no, I do not love that I've done this `node ../../` at all, but I'm not thinking of a better way at the moment.

packages/turbo-workspaces/src/cli.ts
```diff
@@ -25,6 +25,11 @@ workspacesCli
     "Do not run a package manager install after conversion",
     false
   )
+  .option(
```
anthonyshew, line 28: I needed to make sure I was always getting a successful exit code for my scripts. This seemed like a reasonable flag to add, even beyond my specific use case.

turborepo-tests/helpers/setup_example_test.sh
chris-olszewski: Not that it's that much different, but we do [this](https://github.com/vercel/turborepo/blob/main/turborepo-tests/helpers/setup_integration_test.sh#L31) in the integration test setup.

packages/turbo-types/scripts/generate-schema.ts
```diff
@@ -2,9 +2,10 @@
 
 import { writeFileSync } from "node:fs";
 import { join } from "node:path";
+import { fileURLToPath } from "node:url";
 import { createGenerator } from "ts-json-schema-generator";
 
-const __dirname = new URL(".", import.meta.url).pathname;
+const __dirname = fileURLToPath(new URL(".", import.meta.url));
```
anthonyshew, line 8: For posterity, this fixes a bug in generating types in `@turbo/types` on Windows.

anthonyshew: https://github.com/nodejs/node/issues/37845

chris-olszewski (approved):

PR #9599, opened 2024-12-10
chore(integration): use offline install

chris-olszewski:
### Description

This PR does 3 things:
 - Switch to use `--offline` when performing `npm install` to ensure we are only using internal packages in our fixtures. We only do it for `npm` as it lacks a `workspace` protocol.
 - Add a `util` package that we were referencing in our nested workspace fixture
 - Add `--no-install` flag for tests that don't need to install deps. Currently only used for our framework inference test where the presence of `next` in a `package.json` is all we need. We never actually need `next`.

### Testing Instructions

Passes in CI


tknickman (approved):
Thanks for doing this!

PR #9609, opened 2024-12-12
feat(tui): bind to toggle task selection pinning

anthonyshew:
### Description

While task selection pinning is convenient when tasks are changing order and status, it can sometimes get in the way. 

When you haven't selected a task yet (e.g. pressed one of ↑/↓/j/k), you are in an _unpinned_ state. You'll always be looking at the first task in the list, even as they shuffle around in order. When you do press one of those keys, you end up in a _pinned_ state such that the task you selected continues to be selected.

This feels good until you select a task, and then want to "exit" that pinned state and let tasks flow around beneath your cursor again. In this PR, we solve this problem, adding a keybind for `p` that toggles the pinned/unpinned state of your selection.

### Testing Instructions

Pick out your favorite Turborepo, run tasks with the TUI, and try out this new feature.

### DIFF
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -798,6 +798,9 @@ fn update(
         Event::ToggleHelpPopup => {
             app.showing_help_popup = !app.showing_help_popup;
         }
+        Event::TogglePinnedTask => {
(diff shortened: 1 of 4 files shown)


chris-olszewski (approved):
🚢

PR #9620, opened 2024-12-15
feat(examples): add example-with-vite-react

shivam-pawar:
Added a sample repository to bootstrap a React application using Vite.


anthonyshew (approved):
Thanks for the contribution!

PR #9623, opened 2024-12-16
chore(tui): shuffle of persist logic

chris-olszewski:

### DIFF
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -16,7 +16,7 @@ use tokio::{
     time::Instant,
 };
 use tracing::{debug, trace};
-use turbopath::AbsoluteSystemPathBuf;
+use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
 
 use crate::tui::{
     popup::{popup, popup_area},
@@ -29,6 +29,7 @@ const RESIZE_DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
 use super::{
     event::{CacheResult, Direction, OutputLogs, PaneSize, TaskResult},
     input,
+    preferences::PreferenceLoader,
     search::SearchResults,
     AppReceiver, Debouncer, Error, Event, InputOptions, SizeInfo, TaskTable, TerminalPane,
 };
@@ -57,19 +58,14 @@ pub struct App<W> {
     section_focus: LayoutSections,
     task_list_scroll: TableState,
     selected_task_index: usize,
-    is_task_selection_pinned: bool,
-    has_sidebar: bool,
     showing_help_popup: bool,
     done: bool,
+    is_task_selection_pinned: bool,
+    preferences: PreferenceLoader,
 }
 
 impl<W> App<W> {
-    pub fn new(
-        rows: u16,
-        cols: u16,
-        tasks: Vec<String>,
-        repo_root: &AbsoluteSystemPathBuf,
-    ) -> Self {
+    pub fn new(rows: u16, cols: u16, tasks: Vec<String>, preferences: PreferenceLoader) -> Self {
         debug!("tasks: {tasks:?}");
         let size = SizeInfo::new(rows, cols, tasks.iter().map(|s| s.as_str()));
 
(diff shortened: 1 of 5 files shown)


crates/turborepo-ui/src/tui/preferences.rs
```diff
@@ -1,132 +1,73 @@
...
-                let _ = Self::update_preference(
-                    repo_root,
-                    PreferenceFields::PinnedTaskSelection,
-                    serde_json::Value::Bool(false),
-                );
+#[derive(Serialize, Deserialize, Debug)]
+struct Preferences {
+    pub is_task_list_visible: Option<bool>,
+    pub active_task: Option<String>,
+    pub is_pinned_task_selection: Option<bool>,
```
chris-olszewski, line 60: I don't think this field is needed. We can just check if `active_task` is present

crates/turborepo-ui/src/tui/preferences.rs
```diff
@@ -1,132 +1,73 @@
...
-            Some(index) => index,
-            None => {
-                let _ = Self::update_preference(
-                    repo_root,
-                    PreferenceFields::PinnedTaskSelection,
-                    serde_json::Value::Bool(false),
-                );
+#[derive(Serialize, Deserialize, Debug)]
+struct Preferences {
+    pub is_task_list_visible: Option<bool>,
```
chris-olszewski, line 58: I don't think we need this to be optional unless we want to allow users to mess with the config file manually.

crates/turborepo-ui/src/tui/preferences.rs
```diff
@@ -1,132 +1,73 @@
...
-        let preferences_file = repo_root.join_components(TUI_PREFERENCES_PATH_COMPONENTS);
+    pub fn set_active_task(&mut self, value: Option<String>) -> Result<(), Error> {
+        self.config.active_task = value;
+        self.flush_to_disk()
+    }
 
-        let selected_task_name = read_json(&preferences_file)
-            .active_task
-            .unwrap_or("".to_string());
+    fn flush_to_disk(&self) -> Result<(), Error> {
```
chris-olszewski, line 48: Right now we're flushing to disk everytime one of these configurations change. Could we get away with only doing this when the TUI is shutting down? (If not we need to spin up a separate thread to do this so we're not performing FS operations on the render thread)

crates/turborepo-ui/src/tui/preferences.rs
```diff
@@ -1,132 +1,73 @@
...
+use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
 
-#[derive(Serialize, Deserialize, Debug)]
-pub struct Preferences {
-    pub is_task_list_visible: Option<bool>,
-    pub active_task: Option<String>,
-    pub is_pinned_task_selection: Option<bool>,
+/// Structure for reading and writing preferences to disk
+pub struct PreferenceLoader {
+    file_path: AbsoluteSystemPathBuf,
```
chris-olszewski, line 6: I think we might want to offer a non-persistent experience where we don't have an underlying file we read/write to. What do you think?

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -832,30 +816,13 @@ fn update(
...
+            app.persist_active_task().ok();
         }
         Event::ToggleSidebar => {
-            app.has_sidebar = !app.has_sidebar;
-            let _ = preferences::Preferences::update_preference(
-                repo_root,
-                preferences::PreferenceFields::IsTaskListVisible,
-                serde_json::Value::Bool(app.has_sidebar),
-            );
+            let new_value = !app.preferences.is_task_list_visible();
```
chris-olszewski, line 822: TODO: we should move this to be a method on `App` instead of a branch in our `update` match. The former is far easier to write tests for.

crates/turborepo-ui/src/tui/preferences.rs
anthonyshew: I thought about this but it seemed like an unneeded knob if you really had all the controls that you need in hand. If you wanted to get back to the "without persistence" state, you could scroll to the top and tap "p". Maybe there's a missing bind for "Jump to top of tasks"?

crates/turborepo-ui/src/tui/preferences.rs
anthonyshew: You can technically be in a state of having a selected task, but it's not persistent. I don't know if this is useful or not, though...The TUI would get restored with a specific task selected but it wouldn't be pinned. 🤔

crates/turborepo-ui/src/tui/preferences.rs
anthonyshew: This was something I'd thought about but didn't realize `cleanup()` was there on `App`. Yes, very doable.

PR #9633, opened 2024-12-19
docs: update readme with pnpm, yarn, and npm

igAbhinav007:
### Description

Updated the **README** file in the Turborepo basic example to replace the **pnpm commands** with commands for **yarn, npm, and pnpm.** This resolves the issue where only pnpm commands were shown, providing more flexibility based on the user's package manager selection.

### DIFF
--- a/examples/basic/README.md
+++ b/examples/basic/README.md
@@ -38,7 +38,15 @@ To build all apps and packages, run the following command:
 
 ```
 cd my-turborepo
+
+```
+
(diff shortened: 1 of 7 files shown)


anthonyshew:
👋 We actually have a PR (that needs to be finished up) to dynamically replace these commands in https://github.com/vercel/turborepo/pull/9633. Let's see if we can get that one merged instead!

PR #9634, opened 2024-12-19
feat: boundaries

NicholasLYang:
### Description

This is a first pass at a boundaries implementation. We add a top level command `turbo boundaries` that checks your files for invalid imports. We currently check for:

- Importing a file outside of the package's directory
- Importing a type without declaring the import as a type
- Importing a package that is not specified in dependencies

### Testing Instructions

Added tests via `query` for the three scenarios listed above.


crates/turborepo-lib/src/boundaries.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+}
+
+impl Run {
+    pub async fn check_boundaries(&self) -> Result<BoundariesResult, Error> {
+        let packages = self.pkg_dep_graph().packages();
+        let repo = Repository::discover(&self.repo_root()).ok().map(Mutex::new);
+        let mut diagnostics = vec![];
+        let source_map = SourceMap::default();
+        for (package_name, package_info) in packages {
```
NicholasLYang, line 119: We could consider parallelizing this, but it's not *that* slow right now (7 seconds for running in `turborepo` on release mode)

crates/turborepo-lib/src/boundaries.rs
```diff
@@ -0,0 +1,443 @@
...
+    ) -> Result<Vec<BoundariesDiagnostic>, Error> {
+        let files = globwalk::globwalk(
+            package_root,
+            &[
+                "**/*.js".parse().unwrap(),
+                "**/*.jsx".parse().unwrap(),
+                "**/*.ts".parse().unwrap(),
+                "**/*.tsx".parse().unwrap(),
+                "**/*.vue".parse().unwrap(),
+                "**/*.svelte".parse().unwrap(),
```
NicholasLYang, line 177: would love a canonical list of all possible `js` extensions! I might have missed one

crates/turborepo-lib/src/boundaries.rs
```diff
@@ -0,0 +1,440 @@
...
+            ],
+            &["**/node_modules/**".parse().unwrap()],
```
NicholasLYang, line 179: We could leave this up to the `gitignore` check further down, but even recursing into node_modules is rough. In the future we should have a smart filtering mechanism to prevent recursion into directories that are ignored

crates/turborepo-lib/src/boundaries.rs
```diff
@@ -0,0 +1,440 @@
...
+                if matches!(repo.status_should_ignore(file_path.as_std_path()), Ok(true)) {
+                    continue;
+                }
+            };
+            // Read the file content
+            let Ok(file_content) = tokio::fs::read_to_string(&file_path).await else {
+                return Err(Error::FileNotFound(file_path.to_owned()));
+            };
+
+            let comments = SingleThreadedComments::default();
```
NicholasLYang, line 203: there's some boilerplate that is repeated between tracer and boundaries, we could deduplicate but imo it's not critical

crates/turborepo/tests/snapshots/boundaries__boundaries_get_boundaries_lints_(npm@10.5.0).snap
```diff
@@ -0,0 +1,28 @@
+---
+source: crates/turborepo/tests/boundaries.rs
+expression: query_output
+---
+{
+  "data": {
+    "boundaries": {
+      "items": [
+        {
+          "message": "cannot import file `../../packages/another/index.jsx` because it leaves the package",
```
anthonyshew, line 10: Nitpick: I'd like to make a move towards writing sentences with capital letters and punctuation for our human-readable errors (including sentence fragments) when we can.

```suggestion
          "message": "Cannot import file `../../packages/another/index.jsx` because it leaves the package.",
```

crates/turborepo-lib/src/boundaries.rs
```diff
@@ -0,0 +1,440 @@
...
+                }
+            }
+        }
+    }
+}
+
+impl Run {
+    pub async fn check_boundaries(&self) -> Result<BoundariesResult, Error> {
+        let packages = self.pkg_dep_graph().packages();
+        let repo = Repository::discover(&self.repo_root()).ok().map(Mutex::new);
```
chris-olszewski, line 116: Might be worth calling this for each package instead of the `Mutex` locking overhead. Not for this PR, just future perf work.

chris-olszewski (changes requested):
Mostly small things. Want to verify the test fixture is correct and we're not hitting NPM

crates/turborepo/tests/snapshots/boundaries__boundaries_get_boundaries_lints_(npm@10.5.0).snap
NicholasLYang: lol Rome had the opposite style guide, so still doing that. It is a little tricky when we're nesting errors cause then we can end up with multiple periods and capitals

crates/turborepo-lib/src/run/summary/page.rs
```diff
@@ -0,0 +1,62 @@
+use maud::{html, Markup, DOCTYPE};
+use tailwind_css::TailwindBuilder;
+
+use crate::run::summary::{execution::ExecutionSummary, Error, RunSummary};
+
+impl<'a> RunSummary<'a> {
+    pub fn render_html(&self) -> Result<String, Error> {
```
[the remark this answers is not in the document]
NicholasLYang: Oops! This slipped in, removing

crates/turbo-trace/src/import_finder.rs
```diff
@@ -2,41 +2,58 @@ use swc_common::{Span, Spanned};
 use swc_ecma_ast::{Decl, ModuleDecl, Stmt};
 use swc_ecma_visit::{Visit, VisitWith};
 
-use crate::tracer::ImportType;
+use crate::tracer::ImportTraceType;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ImportType {
```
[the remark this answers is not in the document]
NicholasLYang: Yep! I can write a doc comment for that

crates/turborepo-lib/src/boundaries.rs
NicholasLYang: I tested the mutex and at least in a naive hyperfine benchmark, it doesn't change performance at all

NicholasLYang: I think if we end up testing packages in parallel it will have an effect, but tbd

turborepo-tests/integration/fixtures/boundaries/turbo.json
```diff
@@ -0,0 +1 @@
+{}
```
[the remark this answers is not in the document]
NicholasLYang: yes, I don't think we have support for creating a `Run` without a turbo.json. And eventually boundaries should require the `boundaries` key

crates/turborepo-lib/src/query/mod.rs
```diff
@@ -543,6 +545,18 @@ impl RepositoryQuery {
         get_version()
     }
 
+    /// Check boundaries for all packages.
+    async fn boundaries(&self) -> Result<Array<Diagnostic>, Error> {
+        match self.run.check_boundaries().await {
+            Ok(result) => {
+                result.emit();
```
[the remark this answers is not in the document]
NicholasLYang: yep, to match the tracing logic for `dependencies`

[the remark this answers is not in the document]
chris-olszewski: Out of scope for this PR, but this feels weird to me. What's the intended use of this info? Should it be output by default or gated behind a trace/debug level log?

chris-olszewski (changes requested):
Only blocker is that we're still depending hitting NPM when fetching the `util` dependency for `another` package in the test fixture.

chris-olszewski (approved):

Issue #8476, opened 2024-06-13
Cache is extracted to wrong destination


markbrockhoff:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

https://github.com/markbrockhoff/turbo-repo-bug-repro

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

2.0.4-canary.3

### Describe the Bug

Whenever I execute a build using turborepo for the first time everything is generated correctly and cached. But when running the same build again the existing cache is restored to the wrong directory causing build files to end up in the src directory.

### Expected Behavior

I'd expect turborepo to restore the build cache in the exact same location where it came from.

### To Reproduce

I created a minimal reproduction using the latest version of turborepo containing only one package which is the default starter for creating a Nuxt module (That's where I encountered the issue)

To reproduce the issue do the following steps:
1. Clone the repro: `git clone https://github.com/markbrockhoff/turbo-repo-bug-repro`
2. Install the dependencies `pnpm i`
3. Run a first build: `pnpm build`
4. Run a second build: `pnpm build`

After the second build you'll see that two new files were created within /packages/nuxt-test/src/runtime. These files should be placed inside the dist folder but somehow seem to be restored into the src folder.

### Additional context

I'm using pnpm v9.3.0 and node 20.11.0, any help with this issue would be appreciated. Thanks

chris-olszewski:
I can confirm that this is due to the overlapping outputs interacting in a very weird way. It appears that something in `dev:prepare` is creating a symlink from `dist/runtime -> `src/runtime`:
```
[0 olszewski@chriss-mbp] /tmp/turbo-repo-bug-repro/packages/nuxt-test $ pnpm dev:prepare

> my-module@1.0.0 dev:prepare /private/tmp/turbo-repo-bug-repro/packages/nuxt-test
> nuxt-module-build build --stub && nuxt-module-build prepare && nuxi prepare playground

ℹ Stubbing my-module                                                                                                                         11:30:47 AM
ℹ Cleaning dist directory: ./dist                                                                                                            11:30:47 AM
✔ Types generated in .nuxt                                                                                                                   11:30:49 AM
✔ Types generated in playground/.nuxt                                                                                                        11:30:50 AM
[0 olszewski@chriss-mbp] /tmp/turbo-repo-bug-repro/packages/nuxt-test $ ls -lh dist 
total 48
[...]

PR #9646, opened 2024-12-23
chore(cache): add failing test

chris-olszewski:
### Description

Add failing test for #8476


anthonyshew:
@chris-olszewski, I tacked a fix onto here. What do you think?

anthonyshew:
#12488 includes the regression test for this.

PR #9662, opened 2025-01-01
docs: guide for Playwright

anthonyshew:
### Description

Guide on how to use Playwright in Turborepo.

### Testing Instructions

👀

### DIFF
--- a/docs/repo-docs/guides/tools/playwright.mdx
+++ b/docs/repo-docs/guides/tools/playwright.mdx
@@ -1,0 +1,136 @@
+---
+title: Playwright
+description: Learn how to use Playwright in a Turborepo.
+---
+
+import { Tabs, Tab, PackageManagerTabs } from '#/components/tabs';
+
+[Playwright](https://playwright.dev/) enables reliable end-to-end testing for modern web apps.
+
+We recommend creating a Playwright package for each test suite that you'd like to run in your monorepo. This may mean suites broken up per-application, per-domain, or some other scheme, according to your needs. If you are unsure, start with creating a Playwright package per-application.
+
+## Handling Playwright's environment variables
+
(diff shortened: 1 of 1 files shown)


docs/repo-docs/guides/tools/playwright.mdx
```diff
@@ -0,0 +1,136 @@
...
+
+</Tab>
+
+</Tabs>
+
+Later on, when you want to run your end-to-end tests, use [the `--only` flag](/repo/docs/reference/run#--only) to run end-to-end tests without running the application's build first. As an example, your command may look like `turbo run e2e --filter=@repo/playwright-myapp --only`.
+
+## Sharing Playwright utilities
+
+You can also create a common package for shared utilities that you need in your end-to-end test suites. We recommend using `peerDependencies` in this shared package so that you can get access to Playwright used in consumers without having to install Playwright into the shared package itself.
```
chris-olszewski, line 98: ❤️  spreading the gospel of peer dependencies

docs/repo-docs/guides/tools/playwright.mdx
```diff
@@ -0,0 +1,136 @@
...
+    "web": "workspace:*" // [!code highlight]
+  }
+}
+```
+
+</Tab>
+
+</Tabs>
+
+Later on, when you want to run your end-to-end tests, use [the `--only` flag](/repo/docs/reference/run#--only) to run end-to-end tests without running the application's build first. As an example, your command may look like `turbo run e2e --filter=@repo/playwright-myapp --only`.
```
chris-olszewski, line 94: We could encode this behavior in the graph with a `"src": {"dependsOn": ["^src"]}` definition and updating `"e2e": { "dependsOn": ["^src"] }`. This encodes in the task graph that `e2e` doesn't actually require the app to build vs in runtime args. ([We do this for some Rust tasks](https://github.com/vercel/turborepo/blob/main/cli/turbo.json#L8))

The downside is it introduces "transit nodes" 🤷

chris-olszewski (approved):

docs/repo-docs/guides/tools/playwright.mdx
anthonyshew: Tough one. I don't see any clear winner on the tradeoffs...Let's ship it and see if feedback tells us to re-approach that way.

PR #9670, opened 2025-01-03
fix: respect output mode 'none' even when caching is disabled

anthonyshew:
### Description

This is a fresh PR for #7900, which sat too long and got stale. I accidentally closed the PR with no commits after force-pushing so I'm opening this for simplicity's sake.

@aholland deserves all credit here for finding this issue and proposing a fix.

### Testing Instructions

- Existing CI should catch me on regressions 🙏
- Added a test
- Did some manual testing. Things look to be doing what we want.

### DIFF
--- a/crates/turborepo-lib/src/task_graph/visitor/mod.rs
+++ b/crates/turborepo-lib/src/task_graph/visitor/mod.rs
@@ -105,7 +105,7 @@ pub enum Error {
     TaskHash(#[from] task_hash::Error),
     #[error(transparent)]
     RunSummary(#[from] summary::Error),
-    #[error("Internal errors encountered: {0}")]
+    #[error("internal errors encountered: {0}")]
     InternalErrors(String),
     #[error("unable to find package manager binary: {0}")]
     Which(#[from] which::Error),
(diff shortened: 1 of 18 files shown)


crates/turborepo-ui/src/logs.rs
```diff
@@ -58,9 +58,12 @@ impl<W: Write> Write for LogWriter<W> {
...
             (None, None) => {
-                // Should this be an error or even a panic?
-                debug!("no log file or prefixed writer");
-                Ok(0)
+                debug!(
+                    "No log file or prefixed writer to write to. This should only happen when \
+                     both caching is disabled and output logs are set to none."
+                );
+
+                Ok(1)
```
anthonyshew, line 66: When I was giving usize of `0` here, I was getting this error:
`× internal errors encountered: failed to write whole buffer`

I'm imagining there's a Rust-y reason for this, but I don't know it and couldn't come up with an answer upon searching.

If you're reading this and know why, please educate me!

crates/turborepo-lib/src/task_graph/visitor/mod.rs
```diff
@@ -105,7 +105,7 @@ pub enum Error {
     TaskHash(#[from] task_hash::Error),
     #[error(transparent)]
     RunSummary(#[from] summary::Error),
-    #[error("internal errors encountered: {0}")]
+    #[error("Internal errors encountered: {0}")]
```
anthonyshew, line 108: Unrelated, but I'd like for us to move towards using pseudo-"proper" grammar anywhere that our messages are meant to be for humans (or LLMs).

crates/turborepo-ui/src/logs.rs
anthonyshew: It's the caller thinking that `0` means a failure at creating/writing to the buffer.

PR #9686, opened 2025-01-11
fix: remove generator from basic example

anthonyshew:
### Description

In the spirt of keeping the basic example...well, basic, removing this feature. We'd like to keep the ethos of this example true to its name.

Additionally, the generator was broken on an ESM vs. CJS issue that we've yet to sort out.

### Testing Instructions

CI for any breakage to template.


tknickman:
Something else is broken here that is causing this one to not work - I should have a chance to investigate today instead of totally removing it

tknickman:
Would prefer this one instead: https://github.com/vercel/turborepo/pull/9687/files

PR #9708, opened 2025-01-15
fix(create-turbo): correct repo name

anthonyshew:
### Description

We have the repo name hardcoded to identify official examples, but we didn't update the name when we changed the name of the repository many moons ago.

### DIFF
--- a/packages/create-turbo/src/transforms/official-starter.ts
+++ b/packages/create-turbo/src/transforms/official-starter.ts
@@ -20,7 +20,7 @@ export async function transform(args: TransformInput): TransformResult {
   const defaultExample = isDefaultExample(example.name);
   const isOfficialStarter =
     !example.repo ||
-    (example.repo.username === "vercel" && example.repo.name === "turbo");
+    (example.repo.username === "vercel" && example.repo.name === "turborepo");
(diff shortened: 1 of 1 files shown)


tknickman (approved):

packages/create-turbo/src/transforms/official-starter.ts
```diff
@@ -20,7 +20,7 @@ export async function transform(args: TransformInput): TransformResult {
   const defaultExample = isDefaultExample(example.name);
   const isOfficialStarter =
     !example.repo ||
-    (example.repo.username === "vercel" && example.repo.name === "turbo");
+    (example.repo.username === "vercel" && example.repo.name === "turborepo");
```
tknickman, line 23: Let's support both turbo and turborepo, since they will both redirect to the right place and work!

tknickman (changes requested):

packages/create-turbo/src/transforms/official-starter.ts
anthonyshew: :nods: Good catch.

tknickman (approved):

PR #9729, opened 2025-01-20
ci: split examples tests

anthonyshew:
### Description

Currently, our examples testing loop is too slow. Changing just one of the examples causes all of them to miss cache.

This PR instead runs tests only for changed examples. This is going to end up being an even larger improvement given that I've additionally set up a test for each package manager for each core-maintained example, effectively tripling the amount of testing needed.

### Notes

This probably isn't the most elegant way to write this - but it was the easiest to understand. GitHub Actions is just...really hard to work with and my patience has run out. Maybe I'll clean this up later, or maybe some kind soul will review me and give me a quick way to DRY it up.

- The additions in `.github/workflows` are all duplicates of each other, but one for each example.
- The additions in `turborepo-tests/example-*` are creating tests for all of our supported package managers where we used to only test one package manager per example.

### Testing Instructions

CI


anthonyshew:
LFG, I got it.

.github/workflows/examples-test-basic.yml
```diff
@@ -0,0 +1,41 @@
+name: Examples tests - basic
```
tknickman, line 1: I think you could re-use a lot of this logic with a composite action. Something like (untested)

```
name: "Run Turbo Tests"
description: "Setup the environment and run Turbo tests"
inputs:
  node-version:
    description: "Node.js version to use"
    required: true
  turbo-token:
    description: "Turbo token for authentication"
    required: true
  turbo-team:
    description: "Turbo team for authentication"
    required: true
  test-filter:
    description: "Filter for the turbo run command"
    required: true
runs:
  using: "composite"
  steps:
    - name: Checkout
      uses: actions/checkout@v4

    - name: Disable corepack
      shell: bash
      run: corepack disable

    - name: Setup Turborepo Environment
      with:
        node-version: "${{ inputs.node-version }}"

    - name: Install Global Turbo
      uses: ./.github/actions/install-global-turbo

    - name: Run Turbo Tests
      shell: bash
      run: |
        turbo run test \
          --filter="${{ inputs.test-filter }}" \
          --continue \
          --token=${{ inputs.turbo-token }} \
          --team=${{ inputs.turbo-team }} \
          --env-mode=strict \
          --concurrency=1
```

And then all of these can use it like:

```
- name: Run Kitchen-Sink Tests
        uses: ./.github/actions/run-turbo-tests
        with:
          node-version: "22"
          turbo-token: "${{ secrets.TURBO_TOKEN }}"
          turbo-team: "${{ vars.TURBO_TEAM }}"
          test-filter: "@turborepo-examples-tests/kitchen-sink-*"
```

tknickman (commented):
Looks awesome! One suggestion to simplify, but otherwise lgtm

examples/kitchen-sink/pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
```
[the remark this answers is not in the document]
anthonyshew: Yes!

chris-olszewski (approved):

.github/workflows/examples-test-basic.yml
anthonyshew: A long day of battling GHA. I had fatigued on getting it out to the reusable. 😂 

But today is a new day! Doing that now.

PR #9731, opened 2025-01-21
ci: conditional checks depending on source code changes

anthonyshew:
### Description

We're running every check for every PR, and it's making us slower, particularly when the longest tests run for a docs change. This PR is to make sure we only run what we need.


chris-olszewski (approved):

anthonyshew:
This appears to be doing what I want, but it's hard to tell if the integrations/Rust tests are running because there are changes in `.github/**`. I expect that on our future PRs, that these won't run, but it's difficult to tell. I'm prepared to iterate.

PR #9752, opened 2025-01-22
chore(packages): update fs-extra import

tknickman:
### Description

Update how we're importing CJS (this PR moves fs-extra only) libraries to prepare to move our packages to ESM

### DIFF
--- a/packages/eslint-config/library.js
+++ b/packages/eslint-config/library.js
@@ -26,6 +26,7 @@ module.exports = {
       "error",
       { peerDependencies: true, includeTypes: true },
     ],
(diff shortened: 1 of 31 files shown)


packages/eslint-config/library.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
       "error",
       { peerDependencies: true, includeTypes: true },
     ],
+    "import/no-named-as-default-member": ["off"],
```
tknickman, line 29: https://github.com/import-js/eslint-plugin-import/issues/2442

PR #9754, opened 2025-01-22
fix: update oxc-resolver and fix windows bug

NicholasLYang:
### Description

Upgrades oxc-resolver and fixes a bug that we found in Windows. We need to canonicalize paths manually since they may be the abbreviated version and that breaks our path equality logic.

### DIFF
--- a/crates/turbo-trace/src/tracer.rs
+++ b/crates/turbo-trace/src/tracer.rs
@@ -439,6 +439,8 @@ impl Tracer {
             futures.spawn(async move {
                 let file_resolver = Self::infer_resolver_with_ts_config(&file, &resolver);
                 let resolver = file_resolver.as_ref().unwrap_or(&resolver);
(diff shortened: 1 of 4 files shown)


crates/turbo-trace/src/tracer.rs
```diff
@@ -453,7 +453,19 @@ impl Tracer {
                     return (errors, None);
                 };
 
-                for import in imported_files {
+                for mut import in imported_files {
+                    if cfg!(windows) {
```
chris-olszewski, line 457: Can you add a comment here explaining the bug this fixes?

chris-olszewski (approved):
Do we have a test for this fix?

NicholasLYang:
Yeah the existing test suite caught the issue. It just blocked us from upgrading oxc-resolver. Although tbh I don't exactly know how to trigger windows path abbreviations, it just sort of happens

PR #9762, opened 2025-01-22
docs: migration guide for users coming from Nx

anthonyshew:
### Description

We've been increasingly getting more and more asks for a migration guide for those who are coming to Turborepo from Nx. Here it is.

### DIFF
--- a/docs/repo-docs/guides/migrating-from-nx.mdx
+++ b/docs/repo-docs/guides/migrating-from-nx.mdx
@@ -222,13 +222,41 @@ module.exports = nextConfig;
 
 ### Step 5: Add the `packageManager` field
 
-The root package.json needs to have the packageManager field. This ensures developers in the repository use the correct package manager, and that Turborepo can optimize your package graph based on your lockfile.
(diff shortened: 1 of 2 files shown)


docs/repo-docs/guides/migrating-from-nx.mdx
```diff
@@ -0,0 +1,481 @@
...
+turbo link
+```
+
+You may also configure a self-hosted Remote Cache, which does not require a license or any other fees.
+
+## Advanced migration considerations
+
+While the migration guide above is a good starting point, the breadth of possibilities and capabilities of monorepos means that its difficult to create generalized instructions for all cases. Below, we’ve listed some common next steps that you may be thinking about.
+
+### Creating shared packages
```
chris-olszewski, line 423: Do we want to discuss "unhoisting" the external deps from the root and moving them to the packages that use them? Without doing this cache hit rate will be pretty horrid.

chris-olszewski (approved):
Ran through the instructions using `npm` and it all worked for me! 🚢

docs/repo-docs/guides/migrating-from-nx.mdx
anthonyshew: YES, I never circled back for that. Thank you for reminding me.

PR #9764, opened 2025-01-23
chore(examples): update `with-tailwind`

anthonyshew:
### Description

With our newfound commitment to keeping the examples great, here are the upgrades for `with-tailwind`.

### Testing Instructions

`npx create-turbo@latest --example https://github.com/vercel/turborepo/tree/shew-e74f8/examples/with-tailwind`

### DIFF
--- a/examples/with-tailwind/apps/docs/next.config.ts
+++ b/examples/with-tailwind/apps/docs/next.config.ts
@@ -1,7 +1,6 @@
-/** @type {import('next').NextConfig} */
-module.exports = {
-  reactStrictMode: true,
-  transpilePackages: ["@repo/ui"],
+import type { NextConfig } from "next";
+
+const nextConfig: NextConfig = {
   typescript: {
     ignoreBuildErrors: true,
   },
@@ -9,3 +8,5 @@ module.exports = {
     ignoreDuringBuilds: true,
   },
 };
+
+export default nextConfig;
--- a/examples/with-tailwind/apps/docs/tsconfig.json
+++ b/examples/with-tailwind/apps/docs/tsconfig.json
@@ -11,7 +11,7 @@
     "**/*.ts",
     "**/*.tsx",
     "next-env.d.ts",
-    "next.config.js",
+    "next.config.ts",
     ".next/types/**/*.ts"
   ],
   "exclude": ["node_modules"]
--- a/examples/with-tailwind/apps/web/next.config.ts
+++ b/examples/with-tailwind/apps/web/next.config.ts
@@ -1,7 +1,6 @@
-/** @type {import('next').NextConfig} */
-module.exports = {
-  reactStrictMode: true,
-  transpilePackages: ["@repo/ui"],
+import type { NextConfig } from "next";
+
+const nextConfig: NextConfig = {
   typescript: {
     ignoreBuildErrors: true,
   },
@@ -9,3 +8,5 @@ module.exports = {
     ignoreDuringBuilds: true,
   },
 };
+
+export default nextConfig;
--- a/examples/with-tailwind/apps/web/tsconfig.json
+++ b/examples/with-tailwind/apps/web/tsconfig.json
@@ -11,7 +11,7 @@
     "**/*.ts",
     "**/*.tsx",
     "next-env.d.ts",
-    "next.config.js",
+    "next.config.ts",
     ".next/types/**/*.ts"
   ],
   "exclude": ["node_modules"]


package.json
```diff
@@ -18,7 +18,6 @@
   },
   "devDependencies": {
     "@taplo/cli": "^0.5.2",
-    "@types/react": "18.2.0",
```
anthonyshew, line 21: Intentional! This doesn't do anything to make the repo function correctly, but it can mess up your editor's understanding of React. 😄

examples/with-tailwind/apps/web/next.config.ts
```diff
@@ -2,4 +2,10 @@
 module.exports = {
   reactStrictMode: true,
   transpilePackages: ["@repo/ui"],
+  typescript: {
+    ignoreBuildErrors: true,
+  },
+  eslint: {
+    ignoreDuringBuilds: true,
+  },
```
tknickman, line 10: Needed?

examples/with-tailwind/apps/web/package.json
```diff
@@ -1,16 +1,18 @@
 {
   "name": "web",
   "version": "1.0.0",
+  "type": "module",
```
tknickman, line 4: For eslint flat?

examples/with-tailwind/apps/web/postcss.config.mjs
```diff
@@ -0,0 +1,5 @@
+export default {
```
tknickman, line 1: Filename _shouldn't_ need to be .mjs if the package.json type is module.

examples/with-tailwind/packages/ui/package.json
```diff
@@ -13,22 +13,23 @@
   },
   "license": "MIT",
   "scripts": {
-    "build": "tailwindcss -i ./src/styles.css -o ./dist/index.css",
-    "lint": "eslint src/",
-    "dev": "tailwindcss -i ./src/styles.css -o ./dist/index.css --watch",
-    "type-check": "tsc --noEmit"
+    "build": "npx @tailwindcss/cli -i ./src/styles.css -o ./dist/index.css",
```
tknickman, line 16: Nit, but this can just be `tailwindcss` (the `@tailwindcss/cli` package installs a bin with that name) to avoid the `npx`

https://github.com/tailwindlabs/tailwindcss/blob/next/packages/%40tailwindcss-cli/package.json#L19

examples/with-tailwind/packages/ui/package.json
```diff
@@ -13,22 +13,23 @@
...
+    "check-types": "tsc --noEmit",
+    "dev": "npx @tailwindcss/cli -i ./src/styles.css -o ./dist/index.css --watch",
```
tknickman, line 18: same

examples/with-tailwind/packages/ui/tailwind.config.ts
```diff
@@ -1,5 +1,5 @@
 import type { Config } from "tailwindcss";
-import sharedConfig from "@repo/tailwind-config";
+import sharedConfig from "../tailwind-config/tailwind.config";
```
tknickman, line 2: boundaries error!

tknickman (changes requested):

examples/with-tailwind/packages/ui/package.json
anthonyshew: This is how they say to do it in their doc... 🤔
https://tailwindcss.com/docs/installation/tailwind-cli

examples/with-tailwind/apps/web/next.config.ts
anthonyshew: Yes, these don't need to be in the blocking path of a build since we're Turborepo and we do `lint` and `check-types` tasks instead.

examples/with-tailwind/packages/ui/package.json
tknickman: Yea, it's going to be slower that way but will _always_ work regardless of where you installed `@tailwindcss/cli`. But fine keeping it as is to match their docs.

examples/with-tailwind/apps/web/package.json
tknickman: It will break out generators, so would prefer we keep it out until we have that migrated.

examples/with-tailwind/apps/docs/next.config.ts
```diff
@@ -2,4 +2,10 @@
 module.exports = {
```
tknickman, line 2: This is renamed to `.ts` but still using CJS. It should `export default {...}` instead! 

We can also remove the type annotation and use `import type { NextConfig } from 'next';` instead.

examples/with-tailwind/packages/ui/package.json
anthonyshew: Huh, I had thought I tried it as `tailwindcss` before but maybe I had done just `tailwind`. Good looks.

examples/with-tailwind/apps/docs/tsconfig.json
```diff
@@ -1,8 +1,18 @@
...
+        "name": "next"
+      }
+    ]
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
+  "include": [
+    "**/*.ts",
+    "**/*.tsx",
+    "next-env.d.ts",
+    "next.config.js",
```
tknickman, line 14: Should be .ts now?

examples/with-tailwind/packages/ui/tailwind.config.ts
anthonyshew: Betrayed by my editor auto-updating paths! Egad!

examples/with-tailwind/apps/web/next.config.ts
```diff
@@ -2,4 +2,10 @@
 module.exports = {
   reactStrictMode: true,
   transpilePackages: ["@repo/ui"],
+  typescript: {
```
tknickman, line 5: Same comment for docs!

examples/with-tailwind/packages/ui/tailwind.config.ts
tknickman: If only we had a feature for this .... 😄

examples/with-tailwind/apps/web/package.json
tknickman: Not locked to it since we don't have one here. But could cause more issues for us until we update that package

examples/with-tailwind/apps/web/next.config.ts
anthonyshew: https://github.com/vercel/turborepo/pull/9764#discussion_r1927074231

tknickman (approved):

tknickman:
❤️

PR #9770, opened 2025-01-23
chore(examples): update non-monorepo example

anthonyshew:
### Description

Updating the `non-monorepo` example with best practices and latest dependencies. This was scaffolded using `npx create-next-app@latest` and then adding `turbo` to it.

### Testing Instructions

`npx create-turbo@latest --example https://github.com/vercel/turborepo/tree/dependabot/npm_and_yarn/examples/non-monorepo/non-monorepo-cf1be3c1e5/examples/non-monorepo`


examples/non-monorepo/README.md
```diff
@@ -1,42 +1,36 @@
-# Turborepo non-monorepo starter
```
tknickman, line 1: We should probably keep the old README?

tknickman (commented):
One note about the readme

examples/non-monorepo/README.md
anthonyshew: Ah,  yes, good catch. I doctored it up a little bit as well now.

tknickman (approved):

PR #9783, opened 2025-01-23
feat(bun): support `bun.lock`

### LINKED ISSUE #9628: Warning when using text-based Bun lockfile
### Verify canary release

https://github.com/robertherber/repro-turbo-with-bun-text-based-lockfile

2.3.4-canary.3

### Enviroment information

```block
CLI:
   Version: 2.3.3
   Path to executable: /Users/robertherber/code/_play/turbo-bun-text-based/node_modules/turbo-darwin-arm64/bin/turbo
   Daemon status: Running
   Package manager: bun

Platform:
   Architecture: aarch64
   Operating system: macos
   WSL: false
   Available memory (MB): 25846
   Available CPU cores: 10

Environment:
   CI: None
   Terminal program (TERM_PROGRAM): WarpTerminal
   Terminal program version (TERM_PROGRAM_VERSION): v0.2024.12.10.15.55.stable_03
   Shell (SHELL): /bin/zsh
   stdin: false
```

### Expected behavior

I expect it to work without a warning.

### Actual behavior

Getting a very noisy warning whenever I run a turbo command:
```
 WARNING  Issues occurred when constructing package graph. Turbo will function, but some features may not be available:
   × could not resolve workspaces
  ╰─▶ unable to parse: Parsing Error: Error { input: "one-plus-one@^workspace:packages/one-plus-one:\n  version \"workspace:packages/one-plus-one\"\n  resolved \"workspace:packages/one-plus-one\"\n  devDependencies:\n    \"@types/bun\" \"latest\"\n\nturbo@^2.3.3:
      \n  version \"2.3.3\"\n  resolved \"https://registry.npmjs.org/turbo/-/turbo-2.3.3.tgz\"\n  integrity sha512-DUHWQAcC8BTiUZDRzAYGvpSpGLiaOQPfYXlCieQbwUvmml/LRGIe3raKdrOPOoiX0DYlzxs2nH6BoWJoZrj8hA==\n  optionalDependencies:\n    turbo-darwin-64 \"2.3.3\"\n    turbo-
      darwin-arm64 \"2.3.3\"\n    turbo-linux-64 \"2.3.3\"\n    turbo-linux-arm64 \"2.3.3\"\n    turbo-windows-64 \"2.3.3\"\n    turbo-windows-arm64 \"2.3.3\"\n\nturbo-darwin-64@2.3.3:\n  version \"2.3.3\"\n  resolved \"https://registry.npmjs.org/turbo-darwin-64/-/turbo-
      darwin-64-2.3.3.tgz\"\n  integrity sha512-bxX82xe6du/3rPmm4aCC5RdEilIN99VUld4HkFQuw+mvFg6darNBuQxyWSHZTtc25XgYjQrjsV05888w1grpaA==\n\nturbo-darwin-arm64@2.3.3:\n  version \"2.3.3\"\n  resolved \"https://registry.npmjs.org/turbo-darwin-arm64/-/turbo-darwin-arm64-
      2.3.3.tgz\"\n  integrity sha512-DYbQwa3NsAuWkCUYVzfOUBbSUBVQzH5HWUFy2Kgi3fGjIWVZOFk86ss+xsWu//rlEAfYwEmopigsPYSmW4X15A==\n\nturbo-linux-64@2.3.3:\n  version \"2.3.3\"\n  resolved \"https://registry.npmjs.org/turbo-linux-64/-/turbo-linux-64-2.3.3.tgz\"\n  integrity
      sha512-eHj9OIB0dFaP6BxB88jSuaCLsOQSYWBgmhy2ErCu6D2GG6xW3b6e2UWHl/1Ho9FsTg4uVgo4DB9wGsKa5erjUA==\n\nturbo-linux-arm64@2.3.3:\n  version \"2.3.3\"\n  resolved \"https://registry.npmjs.org/turbo-linux-arm64/-/turbo-linux-arm64-2.3.3.tgz\"\n  integrity sha512-NmDE/
      NjZoDj1UWBhMtOPmqFLEBKhzGS61KObfrDEbXvU3lekwHeoPvAMfcovzswzch+kN2DrtbNIlz+/rp8OCg==\n\nturbo-windows-64@2.3.3:\n  version \"2.3.3\"\n  resolved \"https://registry.npmjs.org/turbo-windows-64/-/turbo-windows-64-2.3.3.tgz\"\n  integrity sha512-O2+BS4QqjK3dOERscXqv7N2GXNcqHr9hXumkMxDj/
      oGx9oCatIwnnwx34UmzodloSnJpgSqjl8iRWiY65SmYoQ==\n\nturbo-windows-arm64@2.3.3:\n  version \"2.3.3\"\n  resolved \"https://registry.npmjs.org/turbo-windows-arm64/-/turbo-windows-arm64-2.3.3.tgz\"\n  integrity sha512-dW4ZK1r6XLPNYLIKjC4o87HxYidtRRcBeo/hZ9Wng2XM/
      MqqYkAyzJXJGgRMsc0MMEN9z4+ZIfnSNBrA0b08ag==\n\ntypescript@^5.0.0:\n  version \"5.7.2\"\n  resolved \"https://registry.npmjs.org/typescript/-/typescript-5.7.2.tgz\"\n  integrity sha512-i5t66RHxDvVN40HfDd1PsEThGNnlMCMT3jMUuoh9/0TaqWevNontacunWyN02LA9/fIbEWlcHZcgTKb9QoaLfg==\n\nundici-types@~5.26.4:\n
      version \"5.26.5\"\n  resolved \"https://registry.npmjs.org/undici-types/-/undici-types-5.26.5.tgz\"\n  integrity sha512-JlCMO+ehdEIKqlFxk6IfVoAUVmgz7cU7zD/h9XZ0qzeosSHmUJVOzSQvvYSYWXkFXC+IfLKSIffhv0sVZup6pA==\n", code: Eof }

• Packages in scope: one-plus-one
• Running typecheck in 1 packages
• Remote caching disabled
one-plus-one:typecheck: cache miss, executing 284ac7686cc4f234
one-plus-one:typecheck: 
one-plus-one:typecheck: $ tsc --noEmit

 Tasks:    1 successful, 1 total
Cached:    0 cached, 1 total
  Time:    706ms
```

The command still works though.

### To Reproduce

Here's a simple repo reproduction:
https://github.com/robertherber/repro-turbo-with-bun-text-based-lockfile

Just run `bun typecheck` in the root and the warning is there (as above).

### Additional context

Using bun 1.1.39 with a text-based lockfile (verified I'm getting the same on 1.1.40 and 1.1.41).

chris-olszewski:
### Description

Closes #9628

Switch over from our usage of `yarn.lock` and directly support `bun.lock` parsing.

We use `biome` to strip out the trailing commas that appear in the `workspaces` and `packages` objects to leverage `serde` which I am more competent in. We can switch to `biome` in the future and avoid a second pass, but I do not know how to do correct deserialization logic for `PackageEntry` (see `de.rs` for examples of what this looks like) in `biome`.

I will call out is the fact that the lockfile keys in `bun.lock` aren't used directly unlike other lockfile implementation. Instead we use resolved package identifiers e.g. `package@(protocol:)?version`. The keys themselves can shift when the underlying package does not change resulting in incorrect behavior. 

A quick example to illustrate:
 - `a` depends on `shared@1.0.0`
 - `b` depends on `shared@2.0.0` 
 - `shared@1.0.0` will get they key of `shared` since `a` is before `b` 
 - `shared@2.0.0` will get a key of `b/shared`
 - If `a` updates to `shared@2.0.0`, then the `shared` key will now point to `shared@2.0.0`
 - If we used the key instead of the ident we would rebuild `b` since it's key changed from `b/shared` to `shared` and not rebuild `a` since it's key remained `shared`

This PR does not add support for `turbo prune`.

Reviewing each commit on it's own would probably be helpful.

### Testing Instructions

Unit tests. Manual testing on a `create-turbo` repository.


nicksrandall:
This is a great step toward support `turbo prune` with bun's new text based lock file.

crates/turborepo-lockfiles/src/bun/id.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let curr = self.key;
+        self.key = curr.and_then(less_specific_key);
+        curr
+    }
+}
+
+fn less_specific_key(key: &str) -> Option<&str> {
+    let slash_idx = key.find('/')?;
```
NicholasLYang, line 23: tiny nit: [`split_once`](https://doc.rust-lang.org/std/primitive.str.html#method.split_once) might work better here

crates/turborepo-lockfiles/src/bun/mod.rs
```diff
@@ -68,10 +147,137 @@ impl Lockfile for BunLockfile {
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let parsed_json = biome_json_parser::parse_json(
+            s,
+            JsonParserOptions::default().with_allow_trailing_commas(),
+        );
+        if parsed_json.has_errors() {
+            // TODO
+        }
+        let syntax_tree = parsed_json.syntax();
+        let format = biome_json_formatter::format_node(
```
NicholasLYang, line 213: if you want a slightly shorter trip, I think you can use `biome_deserialize` to get a `serde_json::Value`, then use serde to deserialize that into `BunLockfile`.

NicholasLYang (approved):
A couple notes. The biome -> serde setup could potentially be cleaner, but otherwise looks good!

XavierGeerinck:
Big thanks to @chris-olszewski for working on this!!! Can't wait to have this in a release

crates/turborepo-lockfiles/src/bun/id.rs
chris-olszewski: This is one of those cases where might need a `str` that includes the `/` I split on so I feel like `find` is a little easier to work with here.

crates/turborepo-lockfiles/src/bun/mod.rs
chris-olszewski: I'm not seeing [`serde_json::Value`](https://docs.rs/biome_deserialize/latest/biome_deserialize/trait.Deserializable.html#foreign-impls) as being supported by `biome_deserialize`.

controversial:
It seems like this change breaks support for projects that are still using the binary lockfile:
```
 WARNING  Issues occurred when constructing package graph. Turbo will function, but some features may not be available:
   × Could not resolve workspaces.
  ╰─▶ Lockfile not found at /Users/luke/Developer/my-project/bun.lock
```

bun.lockb is present in my project but not bun.lock

note that bun still supports the binary lockfile even on new versions of bun: “we continue to support the binary `bun.lockb` format and will do so for awhile” [[1]](https://bun.sh/blog/bun-lock-text-lockfile#not-a-breaking-change)

this error applies to users on turborepo 2.4.1 and older versions of bun, but also users who are using the binary lockfile with newer versions of bun.

crates/turborepo-repository/src/package_manager/bun.rs
```diff
@@ -27,10 +28,12 @@ impl Iterator for BunDetector<'_> {
...
 
         self.found = true;
-        let package_json = self.repo_root.join_component(LOCKFILE);
+        let bun_lock = self.repo_root.join_component(LOCKFILE);
 
-        if package_json.exists() {
+        if bun_lock.exists() {
             Some(Ok(PackageManager::Bun))
+        } else if self.repo_root.join_component(LOCKFILE_BINARY).exists() {
+            Some(Err(Error::BunBinaryLockfile))
```
ceIia, line 36: did these lines introduce a regression regarding binary bun lockfiles or were they not supported to begin with?

image

ceIia: just wondering if i should be opening a new issue since moving to the text based lockfile is not an option currently, as the issue with the empty root package name key still exists.

https://github.com/vercel/turborepo/issues/9628#issuecomment-2662589506

PR #9787, opened 2025-01-24
feat(mfe): v2->v1 & make version optional

tknickman:
### Description

V1 config is removed, and V2 config is renamed to V1. Also the version field is optional, defaulting to v1. 
I left a lot of the plumbing in place for multiple versions because I'm sure more will come as we continue to improve it.


chris-olszewski (approved):

chris-olszewski (commented):
You can get rid of the failing `MFE_CONFIG_PKG` test as it exercises behavior that we no longer support from original v1.

PR #9803, opened 2025-01-25
feat(@turbo/codemod): add support for custom NPM registries

### LINKED ISSUE #4152: [turborepo] Update codemod doesn't support custom registries and fails if NPM registry is unavailable

1.8.1

pnpm

Windows

### Describe the Bug

I have an outdated version 1.8.1 of turborepo. When I run turbo commands, it shows me a message to run a codemod to update. However, when I run the codemod to update turborepo it fails. It shows "Unable to fetch the latest version of turbo" error message.

NPM registry is banned by our corporate firewall, and we use a proxy registry. In this scenario the codemod doesn't work.

The reason why it happens is because the registry url is hardcoded in the codemod: https://github.com/vercel/turbo/blob/main/packages/turbo-codemod/src/commands/migrate/steps/getLatestVersion.ts#LL5

### Expected Behavior

It should use the same registry as the package manager and shouldn't rely on availability of NPM registry.

### To Reproduce

To reproduce this issue, configure your package manager to use custom registry, ban npm registry on a network level and run `pnpm dlx @turbo/codemod update`.

### Reproduction Repo

_No response_

Meligy:
Fixes #4152

### Description

NPM (and alternative tools like PNPM) pass the registry as an environment variable `npm_config_registry`.

This allows the `codemod` package to use it when present (which is often the case).

### Testing Instructions

- Go to `turbo-codemod`
- Run `turbo run build`
- Invoke the binary `dist/cli.js migrate --to=latest --force` (maybe play with other `to` values also)


packages/turbo-codemod/src/commands/migrate/steps/getLatestVersion.ts
```diff
@@ -12,9 +12,12 @@ interface PackageDetailsResponse {
 }
 
 async function getPackageDetails({ packageName }: { packageName: string }) {
+  const registry =
+    process.env.npm_config_registry?.replace(/\/$/, "") || DEFAULT_REGISTRY;
```
anthonyshew, line 16: Can you explain the reason for this `.replace` here? grep.app [doesn't show me any prior art](https://grep.app/search?q=process.env.npm_config_registry%3F.replace%28/%5C/%24/%2C%20%22%22%29) for this, though [without the regex](https://grep.app/search?q=process.env.npm_config_registry) does. I'm figuring there's something I'm missing, just can't tell what it is.

Meligy: The reason is to convert `https://registry.npmjs.org/` to `https://registry.npmjs.org`

Happy to remove it though.

anthonyshew: Of course!

anthonyshew (approved):
Thanks, this is great!

ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: be0f731c63c54d837b1f3572ce0bebd2ff77f5ec

PR #9805, opened 2025-01-27
fix(deps): use windows-sys instead of winapi

ognevny:
### Description

`winapi` is abandoned and had no updates since 2020, while `windows-sys` is supported by Microsoft. also updated `winapi-util` to use `windows-sys` too

PR #9817, opened 2025-01-27
chore(turborepo-wax): remove lockfile

ognevny:
### Description

~the same reason as for #9805, but for turborepo-wax. the package update fully removes winapi from lockfile. also cargo updated many packages for unknown reason (I've run `cargo update -p winapi-util`)~

as asked in review lockfile is deleted


ijjk:
Allow CI Workflow Run

- [ ] approve CI run for commit: 6b40477f96e8f4d572468389080097ade57641d0

chris-olszewski (changes requested):
I do not believe this lockfile actually gets used since `turborepo-wax` is a member of the Cargo workspace and should use the same crate versions as the root `Cargo.lock`. 

Could you delete this lockfile instead of updating it?

chris-olszewski (approved):

PR #9823, opened 2025-01-27
ci(examples): update version in examples on release

anthonyshew:
### Description

Updates the `turbo` version in examples whenever we make a release. This ignores canaries so we don't ship those to unexpecting users.

I'm not sure that this is a sustainable solution as the community continues to add more examples since the many `pnpm update` calls might take awhile. However, I feel like this isn't a performance-critical path for us, so going to go with it until someone tells me not to/thinks of a more clever way. Note that whatever we go with needs to 1) bump the version in package.json and 2) update the lockfile.

### DIFF
--- a/.github/workflows/update-examples-on-release.yml
+++ b/.github/workflows/update-examples-on-release.yml
@@ -1,0 +1,21 @@
+name: Update examples to latest
+
+on:
+  workflow_dispatch:
+
+jobs:
+  update-examples-pr:
+    name: "Update examples PR"
+    needs: [stage, npm-publish]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run upgrade script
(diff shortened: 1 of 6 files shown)


.github/workflows/update-examples-on-release.yml
```diff
@@ -0,0 +1,21 @@
...
+  update-examples-pr:
+    name: "Update examples PR"
+    needs: [stage, npm-publish]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run upgrade script
+        run: bash ./scripts/update-examples-dep.sh
+      - name: Create pull request
+        uses: thomaseizinger/create-pull-request@master
+        if: ${{ !inputs.dry_run }}
```
tknickman, line 16:
```suggestion
```

tknickman (approved):
Looks good!

PR #9824, opened 2025-01-27
ci: add `check-types` task to examples testing

anthonyshew:
### Description

In the interest of keeping the examples robust, let's also run the newly added `check-types` task in CI to ensure those stay in the green when people try them out for the first time, too.

Doing this uncovered some issues with existing examples, so glad to see that this will get us to a more stable place!

A notable exception, unfortunately, is going to be the `check-types` in `with-svelte`. Svelte appears to be doing some type generation that can interrupt the type-checking process and make it flake out.

### Testing Instructions

CI


turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -37,7 +37,7 @@ fi
 
 # All examples implement these two tasks
 # and it's reasonable to assume that they will continue to do so
-turbo_command="turbo build lint --continue --output-logs=errors-only"
+turbo_command="turbo build lint check-types --continue --output-logs=errors-only"
```
anthonyshew, line 40:
```suggestion
turbo_command="turbo build && turbo lint && turbo check-types --continue --output-logs=errors-only"
```

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -35,9 +35,15 @@ if [ "$package_manager" == "pnpm" ] && [ "$example_path" == "non-monorepo" ]; th
   package_manager_command="pnpm install --ignore-workspace"
 fi
 
-# All examples implement these two tasks
-# and it's reasonable to assume that they will continue to do so
-turbo_command="turbo build lint --continue --output-logs=errors-only"
+# with-svelte is flaky when building and check types at the same time, because the build process of Svelte involves type generation
+# If the types are generating while the type checking happens, it can cause flakes.
```
chris-olszewski, line 39: Does this indicate that we should have `"check-types": {"dependsOn": ["build"]}` for the Svelte example to make sure that type generation completes before we try to check them?

chris-olszewski (approved):

PR #9827, opened 2025-01-28
docs: remove mentions of exports wildcard limitations

baptisteArno:
This PR removes all the mentions of exports wildcard limitations as it is not the case anymore https://github.com/microsoft/TypeScript/pull/54831

Here is also a screen recording that proves it:

https://github.com/user-attachments/assets/4aae7334-2880-42ec-87b0-26c9471c68e5

### DIFF
--- a/docs/repo-docs/guides/tools/typescript.mdx
+++ b/docs/repo-docs/guides/tools/typescript.mdx
@@ -175,7 +175,7 @@ Setting up `exports` this way has several advantages:
 
 - Using the `types` field allows `tsserver` to use the code in `src` as the source of truth for your code's types. Your editor will always be up-to-date with the latest interfaces from your code.
 - You can quickly add new entrypoints to your package without creating [dangerous barrel files](https://vercel.com/blog/how-we-optimized-package-imports-in-next-js#what's-the-problem-with-barrel-files).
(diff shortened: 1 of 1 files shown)


anthonyshew:
From what I understand from the TypeScript team in the past, the amount of imports it suggests stops at 10. Are you able to test that? Would love if this behavior changed and I didn't hear about it. 😄

baptisteArno:
> From what I understand from the TypeScript team in the past, the amount of imports it suggests stops at 10. Are you able to test that? Would love if this behavior changed and I didn't hear about it. 😄

What do you mean by that exactly? Does that mean the suggest menu would offer max 10 options (which I find enough) or something else?

anthonyshew:
It only works for up to 10 exports subpaths for the package due to performance concerns, is what I was told. That was a while back, though, so that could've changed (like in the PR you've linked).

anthonyshew:
@baptisteArno Did you get a chance to peek at this yet? I'm seeing the behavior as consistent with my past learnings on this topic, but happy to change if you're seeing something different.

anthonyshew:
Going to close this in favor of #9983 since it has a bit more coverage, but I made mention that you should get credit in the description! If you'd like, I can also hand-edit the release notes for #9983 so that they have your name on them instead of me. Let me know if you'd like me to do that!

baptisteArno:
No worries @anthonyshew! Sorry I was super busy 🙏

No need for credits 😄

PR #9834, opened 2025-01-28
docs: `boundaries` experimental page

anthonyshew:
### Description

Documenting the experimental release of Boundaries built-ins.

### Testing Instructions

👀


docs/repo-docs/reference/boundaries.mdx
```diff
@@ -0,0 +1,28 @@
...
+title: boundaries
+description: API reference for the `turbo boundaries` command
+---
+
+import { ExperimentalBadge } from '#/components/experimental-badge';
+import { Callout } from '#/components/callout';
+
+<ExperimentalBadge>Experimental</ExperimentalBadge>
+
+Boundaries ensure that caching and other Turborepo features work correctly by checking for package manager Workspace violations.
```
NicholasLYang, line 11: maybe we should just say "Turborepo works correctly" because if you break boundaries a lot of stuff won't work

docs/repo-docs/reference/boundaries.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+```bash title="Terminal"
+turbo boundaries
+```
+
+This command will notify for three types of violations:
+
+- Importing a file outside of the package's directory
+- Importing a type without declaring the import as a type
```
NicholasLYang, line 20: Ehh this is not quite true. We can't detect this in every case. We just check if you're importing from a type declaration package. But that's sort of a lame lint on its own lol

docs/repo-docs/reference/boundaries.mdx
```diff
@@ -0,0 +1,28 @@
...
+- Importing a package that is not specified in dependencies
```
NicholasLYang, line 21: "as a dependency in your package.json" maybe?

NicholasLYang (approved):
Looks good! I think eventually we should have an error index but this is good for an MVP

PR #9850, opened 2025-01-29
chore: use bunchee instead of tsup

anthonyshew:
### Description

[`bunchee`](https://github.com/huozhi/bunchee) is an incredible bundler that automatically handles a bunch of the complexities of setting up Node.js packages. We want to encourage its usage so using it in the `kitchen-sink` example!


examples/kitchen-sink/pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
```
chris-olszewski, line 1: ⚠️ Pnpm version bump

chris-olszewski (approved):
LGTM, just make sure to regen the lockfile with the correct `pnpm` version

PR #9872, opened 2025-02-02
feat(example): update with-react-native-web, Expo SDK 51 & Expo Router (File-based routing)

shinjith-dev:
### Description

Updated Expo SDK to 51 and added expo router with file-based routing setup.
Updated dependencies to match expo sdk requirements.
Changed `TouchableOpacity` to `Pressable` in the packages/ui as the former is deprecated.

### Testing Instructions

Changes can be verified by running dev script.
I've tested both Expo Go and development build (with JDK 17) for native app and nextjs's usual working.


anthonyshew (approved):
Thank you!

shinjith-dev:
@anthonyshew Small issue here; `@types/react-native` is no longer needed as the `react-native` is typed since version `0.71`. The package has older `metro-*` packages as peer deps which might create issues (I recently had issues due to this but the error message was not really helpful). We just need to remove this package from `apps/native` and `packages/ui` of this example. Sorry for not running thorough tests. Should I create a PR, or can you handle it? Has attached the expo doctor result for clarification. Thanks.
Screenshot_2025-03-16_14-58-49

anthonyshew:
A PR would be great! Thank you!

PR #9889, opened 2025-02-03
fix(kitchen-sink): fix up lockfile

chris-olszewski:
### Description

I accidentally merged https://github.com/vercel/turborepo/pull/9850 with a bad lockfile.

### Testing Instructions

👀


tknickman (approved):

anthonyshew:
Fixed this along with other stuff in #9898. Not sure how this managed to get through but think we're on a more stable path now.

PR #9898, opened 2025-02-04
ci: many examples fixes

anthonyshew:
### Description

After working well for a brief moment in time, the examples are inexplicably quite broken somehow. This PR has a good number of fixes to try to unstuck them once again:
- I had to install ESLint into each package to fix pnpm usage. Unfortunately, this isn't in service of making the examples better. Rather, we were getting the ESLint version running from the outer repository when there wasn't an ESLint specified in the repo.
- This PR improves logging with a `tee` for CI debugging. We previously had poor visibility in CI and were getting vastly different behavior between CI and local (still are 😞). The `tee` should help a little.
- We're using a real temp_dir now. This was done to try to fix the aforementioned ESLint issue, but it's probably what we should have been doing anyway before.
- We were accidentally running the `kitchen-sink` tests twice and never running `basic` due to a copy-pasting error. My bad.

### Testing Instructions

CI

### DIFF
--- a/turborepo-tests/helpers/setup_example_test.sh
+++ b/turborepo-tests/helpers/setup_example_test.sh
@@ -21,6 +21,8 @@ fi
 
 echo "node --version: $(node --version)"
 
+echo $PWD
+
 # Convert to the right package manager
 if [ "$package_manager" == "npm" ]; then
   package_manager_command="node ../../../packages/turbo-workspaces/dist/cli.js convert . npm --ignore-unchanged-package-manager"
--- a/turborepo-tests/helpers/setup_example_test.sh
+++ b/turborepo-tests/helpers/setup_example_test.sh
@@ -59,14 +59,9 @@ rsync -avq \
 
 cd "$temp_dir/$example_path-$package_manager/$example_path"
 
-# Make a place for dumping logs so we can inspect them later
-mkdir -p ./tmp-logs
-echo "/tmp-logs/" >>".gitignore"
-
 # Run package manager conversion
 $package_manager_command
 
-echo $PWD
 # Simulating the user's first run and dumping logs to a file
 $turbo_command 2>&1 | tee ../run-1.txt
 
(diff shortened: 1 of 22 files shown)


turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -70,7 +70,7 @@ echo "/tmp/" >>".gitignore"
 $package_manager_command
 
 # Simulating the user's first run and dumping logs to a file
-$turbo_command >>./tmp/run-1.txt 2>&1
+$turbo_command 2>&1 | tee ./tmp/run-1.txt
 
 cat ./tmp/run-1.txt
```
chris-olszewski, line 75: I think we can remove the `cat` afterwards since we're now sending logs to the file and `stdout`

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -85,7 +85,7 @@ if grep -q ">>> FULL TURBO" ./tmp/run-1.txt; then
 fi
 
 # Simulating the user's second run and dumping logs to a file
-$turbo_command >>./tmp/run-2.txt 2>&1
+$turbo_command 2>&1 | tee ./tmp/run-2.txt
 
 cat ./tmp/run-2.txt
```
chris-olszewski, line 90: Same as above

chris-olszewski (approved):

turborepo-tests/helpers/setup_example_test.sh
anthonyshew: Yes! Still mucking about with other problems being revealed by these `tees`, but thank you for commenting so I don't forget (as I do).

turborepo-tests/helpers/setup_example_test.sh
```diff
@@ -21,13 +21,16 @@ fi
 
 echo "node --version: $(node --version)"
 
+temp_dir=$(mktemp -d)
+echo "Created temporary directory: $temp_dir"
+
 # Convert to the right package manager
 if [ "$package_manager" == "npm" ]; then
-  package_manager_command="node ../../../packages/turbo-workspaces/dist/cli.js convert . npm --ignore-unchanged-package-manager"
+  package_manager_command="npx @turbo/workspaces convert . npm --ignore-unchanged-package-manager"
```
anthonyshew, line 29: No, future reviewer, I do not enjoy this code. However, I'm not sure how else I'd want to do it now that I'm working in a temp dir. 🤔

PR #9938, opened 2025-02-10
feat(boundaries): add ignore directive

NicholasLYang:
### Description

Add ability to ignore imports for boundaries by commenting `// @boundaries-ignore` 

### Testing Instructions
Added tests to boundaries fixture

### DIFF
--- a/crates/turborepo-lib/src/boundaries/mod.rs
+++ b/crates/turborepo-lib/src/boundaries/mod.rs
@@ -13,7 +13,10 @@ use globwalk::Settings;
 use miette::{Diagnostic, NamedSource, Report, SourceSpan};
 use regex::Regex;
 use swc_common::{
-    comments::SingleThreadedComments, errors::Handler, input::StringInput, FileName, SourceMap,
+    comments::{Comments, SingleThreadedComments},
+    errors::Handler,
+    input::StringInput,
+    FileName, SourceMap, Span,
 };
 use swc_ecma_ast::EsVersion;
 use swc_ecma_parser::{lexer::Lexer, Capturing, EsSyntax, Parser, Syntax, TsSyntax};
@@ -133,6 +136,9 @@ static PACKAGE_NAME_REGEX: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"^(@[a-z0-9-~][a-z0-9-._~]*\/)?[a-z0-9-~][a-z0-9-._~]*$").unwrap()
 });
 
+/// Maximum number of ignored import warnings to show
+const MAX_IGNORE_WARNINGS: usize = 10;
+
 pub struct BoundariesResult {
     files_checked: usize,
     packages_checked: usize,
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/boundaries/mod.rs
```diff
@@ -223,6 +229,18 @@ impl Run {
         })
     }
 
+    fn is_ignored_import(comments: &SingleThreadedComments, span: Span) -> bool {
+        if let Some(import_comments) = comments.get_leading(span.lo()) {
+            for comment in import_comments {
+                if comment.text.trim() == "@boundaries-ignore" {
```
tknickman, line 235: What do you think about branding the ignore, and requiring a reason? 

`// @turbo-boundaries-ignore -- we're yolorepo-ing`

A little eslinty, but I think requiring reasons here would be very helpful for maintainers

crates/turborepo-lib/src/boundaries/mod.rs
```diff
@@ -363,7 +381,25 @@ impl Run {
             // Visit the AST and find imports
             let mut finder = ImportFinder::default();
             module.visit_with(&mut finder);
+            let mut warning_count = 0;
             for (import, span, import_type) in finder.imports() {
+                // If the import is prefixed with `@boundaries-ignore`, we ignore it, but print
+                // a warning
+                if Self::is_ignored_import(&comments, *span) {
+                    if warning_count <= MAX_IGNORE_WARNINGS {
```
tknickman, line 389: Can we add another case here that outputs the total count of warnings that exceeded the max?

crates/turborepo-lib/src/boundaries/mod.rs
```diff
@@ -363,7 +381,25 @@ impl Run {
             // Visit the AST and find imports
             let mut finder = ImportFinder::default();
             module.visit_with(&mut finder);
+            let mut warning_count = 0;
             for (import, span, import_type) in finder.imports() {
+                // If the import is prefixed with `@boundaries-ignore`, we ignore it, but print
+                // a warning
+                if Self::is_ignored_import(&comments, *span) {
```
chris-olszewski, line 388: Can we avoid having the display logic inside of `check_package` and instead have `check_package` return diagnostics and then have a some sort of "reporter" decide on what to do about each diagnostic? I think we'll want to eventually offer a way to disallow this escape hatch (or on CI display all warnings). Should also make it easier to write unit tests for this codepath.

turborepo-tests/integration/fixtures/boundaries/apps/my-app/index.ts
```diff
@@ -5,3 +5,20 @@ import { Ship } from "ship";
 import { Ship } from "@types/ship";
 // Import package that is not specified
 import { walkThePlank } from "module-package";
+
+// Import from a package that is not specified, but we have `@boundaries-ignore` on it
+// @boundaries-ignore
```
chris-olszewski, line 10: Should this have produced a change in a test fixture where a warning is printed?

chris-olszewski (changes requested):
I think separating out display logic in `check_package` will be beneficial in the future. Also allows us to write unit tests checking if warnings were created.

turborepo-tests/integration/fixtures/boundaries/apps/my-app/index.ts
NicholasLYang: We don't snapshot warnings since they go to stderr. I'm not sure if we should, since that could be pretty brittle (and also makes debugging tests slightly annoying)

crates/turborepo-lib/src/boundaries/mod.rs
NicholasLYang: Changed it to store the warnings. We could keep the diagnostics and annotate them as disabled, but I think this works fine for now. If we want to disallow the ignore annotation, we can also just make the `is_ignored_import` check return false

crates/turborepo-lib/src/boundaries/mod.rs
NicholasLYang: Sure. Do you think we should do `@turbo-boundaries-ignore`? I kinda like just `@boundaries-ignore`

PR #9939, opened 2025-02-11
feat: adds pnpm lockfile ignoredOptionalDependencies

ragrag:
### Description

Adds pnpm lockfile ignoredOptionalDependencies

I'm unaware if i need to make changes anywhere else, this works as expected in e.g copying lockfile during pruning

Fixes: https://github.com/vercel/turborepo/issues/8959


chris-olszewski (approved):
Thanks, I believe you're correct that this is all that's required. The `pnpm-lock.yaml` we're pruning  should have omitted any of the ignored optional dependencies so we wouldn't have the chance to copy them over.

PR #9950, opened 2025-02-13
fix: `affected_packages`'s optimization flow

dimitropoulos:
### Description

This PR addresses the `affected_packages` API and its handling of the optimization for lockfiles.  It should not change the current behaviors without this optimization.

### Testing Instructions

In Progress...

### DIFF
--- a/crates/turborepo-filewatch/src/package_watcher.rs
+++ b/crates/turborepo-filewatch/src/package_watcher.rs
@@ -181,6 +181,7 @@ const INVALIDATION_PATHS: &[&str] = &[
     "package-lock.json",
     "yarn.lock",
     "bun.lockb",
+    // "bun.lock", // Not yet.
 ];
 
 impl Subscriber {
--- a/crates/turborepo-lib/src/package_changes_watcher.rs
+++ b/crates/turborepo-lib/src/package_changes_watcher.rs
@@ -15,7 +15,9 @@ use turborepo_filewatch::{
     NotifyError, OptionalWatch,
 };
 use turborepo_repository::{
-    change_mapper::{ChangeMapper, GlobalDepsPackageChangeMapper, PackageChanges},
+    change_mapper::{
+        ChangeMapper, GlobalDepsPackageChangeMapper, LockfileContents, PackageChanges,
+    },
     package_graph::{PackageGraph, PackageGraphBuilder, PackageName, WorkspacePackage},
     package_json::PackageJson,
 };
@@ -342,7 +344,8 @@ impl Subscriber {
                     continue;
                 }
 
-                let changed_packages = change_mapper.changed_packages(changed_files.clone(), None);
+                let changed_packages = change_mapper
+                    .changed_packages(changed_files.clone(), LockfileContents::Unknown);
 
                 tracing::warn!("changed_files: {:?}", changed_files);
                 tracing::warn!("changed_packages: {:?}", changed_packages);
--- a/crates/turborepo-lib/src/run/scope/change_detector.rs
+++ b/crates/turborepo-lib/src/run/scope/change_detector.rs
@@ -5,7 +5,7 @@ use turbopath::{AbsoluteSystemPath, AnchoredSystemPathBuf};
 use turborepo_repository::{
     change_mapper::{
         AllPackageChangeReason, ChangeMapper, DefaultPackageChangeMapper, Error,
-        GlobalDepsPackageChangeMapper, PackageChanges, PackageInclusionReason,
+        GlobalDepsPackageChangeMapper, LockfileContents, PackageChanges, PackageInclusionReason,
     },
     package_graph::{PackageGraph, PackageName},
 };
@@ -52,13 +52,12 @@ impl<'a> ScopeChangeDetector<'a> {
     }
 
     /// Gets the lockfile content from SCM if it has changed.
-    /// Does *not* error if cannot get content, instead just
-    /// returns a Some(None)
-    fn get_lockfile_contents(
+    /// Does *not* error if cannot get content.
+    pub fn get_lockfile_contents(
         &self,
         from_ref: Option<&str>,
         changed_files: &HashSet<AnchoredSystemPathBuf>,
-    ) -> Option<Option<Vec<u8>>> {
+    ) -> LockfileContents {
         let lockfile_path = self
             .pkg_graph
             .package_manager()
@@ -70,23 +69,21 @@ impl<'a> ScopeChangeDetector<'a> {
             &lockfile_path,
         ) {
             debug!("lockfile did not change");
-            return None;
+            return LockfileContents::Unchanged;
         }
 
-        let lockfile_path = self
-            .pkg_graph
-            .package_manager()
-            .lockfile_path(self.turbo_root);
-
         let Ok(content) = self.scm.previous_content(from_ref, &lockfile_path) else {
-            return Some(None);
+            debug!("lockfile did change but could not get previous content");
+            return LockfileContents::UnknownChange;
         };
 
-        Some(Some(content))
+        debug!("lockfile changed, have the previous content");
+        LockfileContents::Changed(content)
     }
 }
 
 impl<'a> GitChangeDetector for ScopeChangeDetector<'a> {
+    /// get the actual changed packages between two git refs
     fn changed_packages(
         &self,
         from_ref: Option<&str>,
--- a/crates/turborepo-repository/src/change_mapper/mod.rs
+++ b/crates/turborepo-repository/src/change_mapper/mod.rs
@@ -29,6 +29,21 @@ pub enum LockfileChange {
     ChangedPackages(HashSet<turborepo_lockfiles::Package>),
 }
 
+/// This describes the state of a change to a lockfile.
+pub enum LockfileContents {
+    /// We know the lockfile did not change
+    Unchanged,
+    /// We know the lockfile changed but don't have the file contents of the
+    /// previous lockfile (i.e. `git status`, or perhaps a lockfile that was
+    /// deleted or otherwise inaccessible with the information we have)
+    UnknownChange,
+    /// We are unsure whether the lockfile changed or not
+    Unknown,
+    /// We know the lockfile changed and have the contents of the previous
+    /// lockfile
+    Changed(Vec<u8>),
+}
+
 #[derive(Debug, PartialEq, Eq, Hash, Clone)]
 pub enum PackageInclusionReason {
     /// All the packages are invalidated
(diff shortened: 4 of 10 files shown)


crates/turborepo-filewatch/src/package_watcher.rs
```diff
@@ -181,6 +181,7 @@ const INVALIDATION_PATHS: &[&str] = &[
     "package-lock.json",
     "yarn.lock",
     "bun.lockb",
+    // "bun.lock", // Not yet.
```
dimitropoulos, line 184: happy to remove this if it's not helpful, but thought it might be

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -70,23 +69,21 @@ impl<'a> ScopeChangeDetector<'a> {
             &lockfile_path,
         ) {
             debug!("lockfile did not change");
-            return None;
```
dimitropoulos, line 73: this appears to have been a bug.  the "None" result was intended to mean "there was a change but we don't know what it was" which maps, today, to `LockfileContents::Unknown`.  But, as you can see here, it's being used in the codepath where `LockfileContents::Unchanged` is correct.

crates/turborepo-repository/src/change_mapper/mod.rs
```diff
@@ -29,6 +29,21 @@ pub enum LockfileChange {
     ChangedPackages(HashSet<turborepo_lockfiles::Package>),
 }
 
+/// This describes the state of a change to a lockfile.
+pub enum LockfileContents {
+    /// We know the lockfile did not change
+    Unchanged,
```
dimitropoulos, line 35: although at first I thought this might not be needed because we have the list of files and can simply see that the lockfile wasn't in there, it's used in `change_detector` with `get_lockfile_contents`.

crates/turborepo-filewatch/src/package_watcher.rs
chris-olszewski: I think adding this would make sense, https://github.com/vercel/turborepo/pull/9783 is out in the wild

crates/turborepo-repository/src/change_mapper/mod.rs
```diff
@@ -29,6 +29,21 @@ pub enum LockfileChange {
...
+    /// We know the lockfile changed but don't have the file contents of the
+    /// previous lockfile (i.e. `git status`, or perhaps a lockfile that was
+    /// deleted or otherwise inaccessible with the information we have)
```
chris-olszewski, line 38: This is a good callout. Fetching the previous lockfile will fail if the user switches package managers as we'll try to read the current package manager's lockfile.

crates/turborepo-repository/src/change_mapper/package.rs
```diff
@@ -115,7 +115,7 @@ impl PackageChangeMapper for GlobalDepsPackageChangeMapper<'_> {
         // workspace package dependencies
         if matches!(
             path.as_str(),
-            "package.json" | "pnpm-lock.yaml" | "yarn.lock"
+            "package-lock.json" | "pnpm-lock.yaml" | "yarn.lock"
```
chris-olszewski, line 118: Not necessary for this PR, but we should switch this to `PackageManager::supported_managers().iter().any(|pm| pm.lockfile_name() == path.as_str())` to avoid mistyping/missing lockfiles

packages/turbo-repository/js/index.d.ts
```diff
@@ -57,6 +57,7 @@ export class Workspace {
    */
   affectedPackages(
     files: Array<string>,
-    changedLockfile?: string | undefined | null
+    comparison?: string | undefined | null,
```
chris-olszewski, line 60: Not blocking, but I feel like `from` or `base` is less confusing name.
Defer to Git wizard @NicholasLYang, if you're finding affected packages between `HEAD` and another commit, is there a standard name for what to call that other commit?

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -182,6 +185,24 @@ impl Workspace {
...
+        workspace_root: &AbsoluteSystemPath,
+        from_commit: Option<&str>,
+    ) -> LockfileContents {
+        let lockfile_name = self.graph.package_manager().lockfile_name();
+        changed_files
+            .get(AnchoredSystemPath::new(&lockfile_name).unwrap())
+            .and_then(|_| {
+                let git = SCM::new(workspace_root);
+                let anchored_path = workspace_root.join_component(&lockfile_name);
+                let contents = git.previous_content(from_commit, &anchored_path).unwrap();
```
chris-olszewski, line 200: We'll need to handle this error as this can totally fail. Could probably just debug log the error and return a `LockfileContents::UnknownChange`

crates/turborepo-repository/src/change_mapper/package.rs
dimitropoulos: great: moved to https://github.com/vercel/turborepo/pull/9957/files and removed from this PR in eab7f354235027c6959ea6f729db67cc64964a02

crates/turborepo-filewatch/src/package_watcher.rs
dimitropoulos: great: I put up https://github.com/vercel/turborepo/pull/9958 and removed it from this PR in 1293f7765e438bd3737636fda606882879b117c4

crates/turborepo-repository/src/change_mapper/mod.rs
dimitropoulos: thanks!!

packages/turbo-repository/js/index.d.ts
dimitropoulos: @NicholasLYang I agree with @chris-olszewski - happy to refer to you on this and in my next PR on the `api` side, I'll be sure to rename it there too, to match.

packages/turbo-repository/rust/src/lib.rs
dimitropoulos: good call: fixed in 6494692528b812240a487a5ff9954e76c523b58c

packages/turbo-repository/js/index.d.ts
NicholasLYang: Probably base is best

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -206,11 +244,29 @@ impl Workspace {
...
-        let global_deps_package_detector =
-            GlobalDepsPackageChangeMapper::new(&self.graph, std::iter::empty::<&str>()).unwrap();
-        let mapper = ChangeMapper::new(&self.graph, vec![], global_deps_package_detector);
-        let lockfile_change = changed_lockfile.map(|s| Some(s.into_bytes()));
-        let package_changes = match mapper.changed_packages(hash_set_of_paths, lockfile_change) {
+        let change_detector = comparison
+            .is_some()
+            .then(|| {
+                Either::Left(
+                    GlobalDepsPackageChangeMapper::new(&self.graph, std::iter::empty::<&str>())
```
chris-olszewski, line 251: I'm realize this isn't correct. We need to treat any "root" that isn't the lockfile as an "all packages" change.

Currently any change to a file that doesn't belong to a package will result in only the "root" getting triggered: https://github.com/vercel/turborepo/blob/main/crates/turborepo-repository/src/change_mapper/package.rs#L130

We could change "root" handling for consumers, but that feels a little leaky

crates/turborepo-lib/src/run/scope/change_detector.rs
dimitropoulos: update: it was both but they hit the same codepath - so there's no behavioral change here here.

crates/turborepo-repository/src/change_mapper/mod.rs
```diff
@@ -183,13 +193,26 @@ impl<'a, PD: PackageChangeMapper> ChangeMapper<'a, PD> {
...
+                        );
                         Ok(PackageChanges::All(
                             AllPackageChangeReason::LockfileChangedWithoutDetails,
                         ))
                     }
-                    None => Ok(PackageChanges::Some(changed_pkgs)),
+
+                    // We don't know if the lockfile changed or not, so we can't assume anything
+                    LockfileContents::Unknown => {
+                        debug!("no previous lockfile available, assuming all packages changed");
```
chris-olszewski, line 209: The debug comment doesn't match the behavior. We aren't adding any additional changed packages in this branch.

crates/turborepo-lib/src/run/scope/change_detector.rs
```diff
@@ -52,13 +52,12 @@ impl<'a> ScopeChangeDetector<'a> {
...
     /// Gets the lockfile content from SCM if it has changed.
-    /// Does *not* error if cannot get content, instead just
-    /// returns a Some(None)
-    fn get_lockfile_contents(
+    /// Does *not* error if cannot get content.
+    pub fn get_lockfile_contents(
         &self,
         from_ref: Option<&str>,
         changed_files: &HashSet<AnchoredSystemPathBuf>,
-    ) -> Option<Option<Vec<u8>>> {
```
chris-olszewski, line 61: We're going from 3 states to 4 states with the return type change, but we never return `LockfileContents::Unknown` from this function. Is that variant needed?

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -182,6 +188,27 @@ impl Workspace {
...
+            .then(|| {
+                let git = SCM::new(workspace_root);
+                let anchored_path = workspace_root.join_component(lockfile_name);
+                git.previous_content(Some(from_commit), &anchored_path)
+                    .map(LockfileContents::Changed)
+                    .inspect_err(|e| debug!("{e}"))
+                    .ok()
+                    .unwrap_or(LockfileContents::UnknownChange)
+            })
+            .unwrap_or(LockfileContents::Unknown)
```
chris-olszewski, line 209: I think this should be `LockfileContents::Unchanged` since the lockfile wasn't in the list of changed files
```suggestion
            .unwrap_or(LockfileContents::Unchanged)
```

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -206,11 +242,24 @@ impl Workspace {
...
+
+        let lockfile_contents = if let Some(base) = base {
+            self.get_lockfile_contents(&changed_files, workspace_root, base)
+        } else if changed_files.contains(
+            AnchoredSystemPath::new(self.graph.package_manager().lockfile_name())
+                .expect("the lockfile name will not be an absolute path"),
+        ) {
+            LockfileContents::UnknownChange
+        } else {
+            LockfileContents::Unknown
```
chris-olszewski, line 259: Same as above, we've verified the lockfile isn't in the list of changed files
```suggestion
            LockfileContents::Unchanged
```

chris-olszewski (changes requested):
Only blocking comments are around `LockfileContents::Unknown` usage. I'm unsure if we need that state.

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -342,7 +344,8 @@ impl Subscriber {
                     continue;
                 }
 
-                let changed_packages = change_mapper.changed_packages(changed_files.clone(), None);
+                let changed_packages = change_mapper
+                    .changed_packages(changed_files.clone(), LockfileContents::Unknown);
```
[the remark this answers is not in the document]
dimitropoulos: cool cool, updated in 5d0c5b6e4d7497b217aaadb8c156b02e8b0040ee

crates/turborepo-repository/src/change_mapper/mod.rs
dimitropoulos: `LockfileContents::Unknown` was removed in bd24c7c623c058a67e1ba51e34182edecbd41d78

chris-olszewski (approved):

PR #9953, opened 2025-02-13
fix: documentation for type-check in with-shell-commands example

Asupkay:
### Description

If you run `turbo typecheck` which the documentation currently suggests you get an error that it could not find `typecheck` in project.

Screenshot 2025-02-13 at 12 57 27 AM

This pull request updates the documentation to `type-check` which aligns with what is in the packages and apps.

Screenshot 2025-02-13 at 12 57 46 AM

### Testing Instructions

Go into `examples/with-shell-commands` and run `turbo type-check`


anthonyshew (approved):
Thank you!

PR #9961, opened 2025-02-13
refactor(spaces): no more spaces

NicholasLYang:
### Description

Deletes spaces code


crates/turborepo-lib/src/rewrite_json.rs
```diff
@@ -401,33 +401,33 @@ mod test {
 
     unset_tests! {
         nonexistent_path: (
-            r#"{ "before": {}, "experimentalSpaces": { "id": "one" }, "experimentalSpaces": { "id": "two" }, "after": {} }"#,
```
NicholasLYang, line 404: technically these tests have nothing to do with spaces, but I didn't want to confuse future contributors with an irrelevant key

crates/turborepo-ci/src/lib.rs
```diff
@@ -46,15 +46,12 @@ impl Vendor {
...
-    /// We return an empty String instead of None because
-    /// the Spaces API expects some sort of string in the user field.
-    pub fn get_user() -> String {
+    pub fn get_user() -> Option<String> {
         let vendor = Vendor::infer();
 
         vendor
             .and_then(|v| v.username_env_var)
             .and_then(|v| env::var(v).ok())
-            .unwrap_or_default()
```
chris-olszewski, line 57: Maybe keep this in case anyone is parsing summary/dry JSON and expecting the user to be present?

crates/turborepo-lib/src/config/mod.rs
```diff
@@ -564,7 +558,7 @@ mod test {
 
         repo_root
             .join_component("turbo.json")
-            .create_with_contents(r#"{"experimentalSpaces": {"id": "my-spaces-id"}}"#)
```
chris-olszewski, line 567: Can we keep this test so we don't hard-error if users still have this key?

This PR when released would break our repository with:
```
[0 olszewski@macbookpro] /Users/olszewski/code/vercel/turborepo $ turbo_dev build
 WARNING  No locally installed `turbo` found. Using version: 2.4.2.
turbo 2.4.2

turbo_json_parse_error

  × Failed to parse turbo.json.
  ╰─▶   × Found an unknown key `experimentalSpaces`.
           ╭─[turbo.json:221:3]
       220 │   },
       221 │   "experimentalSpaces": {
           ·   ────────────────────
       222 │     "id": "space_WTrAuWxqVUapGgZDX7KiZV1v"
           ╰────
```

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -643,9 +637,6 @@ pub enum Command {
         /// Answer yes to all prompts (default false)
         #[clap(long, short)]
         yes: bool,
-        /// Specify what should be linked (default "remote cache")
-        #[clap(long, value_enum, default_value_t = LinkTarget::RemoteCache)]
```
chris-olszewski, line 647: We should still allow for this to be passed, just warn that it does nothing.

chris-olszewski (changes requested):
Overall looks good, but I think we shouldn't hard error on these config options/flags and instead warn and ignore.

This was just experimental, but it has been in the codebase 2 years so I don't want to add a hard error without any warning.

crates/turborepo-ci/src/lib.rs
NicholasLYang: It's kept on the output end, but I changed the type because I don't love default values as None in our codebase

crates/turborepo-lib/src/config/mod.rs
NicholasLYang: Do we need to keep this exact test? This is more a config layering test than a parsing test. I can keep the turbo.json parsing tests for the `experimentalSpaces` key

chris-olszewski (approved):

PR #9962, opened 2025-02-13
fix(tui): replay all logs sent to tui with forced crlf

chris-olszewski:
### Description

If a task is run without a TTY (e.g. in CI) the logs will be created with LF (`\n`) line endings. The TUI assumes that logs are produced hooked up to a TTY where CRLF (`\r\n`) is used to return the cursor to the first column and move to the next row (`\n` behavior in non-TTY).

This PR adds a log replay method that replays logs with CRLF line endings regardless of what is in the actual file.

Future PR is to refactor this to share code with the standard log replay as I didn't have time to tackle this right at this moment.

### Testing Instructions

Added quick unit test to verify this swaps out LF for CRLF and is a noop if the logs already use CRLF.

Manual Test:
First populate the logs with a non-TTY run:
```
[0 olszewski@macbookpro] /tmp/tui-test $ turbo @repo/ui#build > /dev/null
```

Using `turbo@2.4.2`:
Screenshot 2025-02-13 at 6 24 47 PM

Using `turbo_dev` from this PR
Screenshot 2025-02-13 at 6 26 04 PM

### DIFF
--- a/crates/turborepo-lib/src/task_graph/visitor/output.rs
+++ b/crates/turborepo-lib/src/task_graph/visitor/output.rs
@@ -94,7 +94,7 @@ impl<W: Write> CacheOutput for TaskCacheOutput<W> {
                 let writer = direct.output_prefixed_writer();
                 turborepo_ui::replay_logs(writer, log_file)
             }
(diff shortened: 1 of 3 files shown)


anthonyshew (approved):
Nice. Thanks, Chris!

PR #9966, opened 2025-02-14
fix(tui): only display interactive message on tasks with stdin

chris-olszewski:
### Description

We should only be displaying the interactive message if the underlying task has a `stdin` allowing it to be interacted with.

### Testing Instructions

Added unit test. Quick verified that tasks that aren't interactive don't have the footer text.


anthonyshew (approved):
Oh, huge. Thanks for picking this up.

PR #9970, opened 2025-02-14
chore: standardize how turbo is ran in example packages

Asupkay:
### Description

This pull request goes through and makes sure each example uses "turbo run _x_" instead of "turbo _x_". I did this as [in the docs](https://turbo.build/repo/docs/reference/run) it says "We recommend using `turbo run` in CI pipelines and `turbo` with global turbo locally for ease of use". In the examples some packages were one way, others the opposite, some mixed. Since these command will be run by proxy by pipelines it would be good to standardize across the board to `turbo run`.

### Testing Instructions

I went through each package where there were changes and ran every command.

### DIFF
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -2,11 +2,11 @@
   "name": "my-turborepo",
   "private": true,
   "scripts": {
-    "build": "turbo build",
-    "dev": "turbo dev",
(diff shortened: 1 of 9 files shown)


anthonyshew (approved):
Good call! Forgot to do this. Thank you!

Issue #9919, opened 2025-02-07
eslint-config-turbo/flat - missing TypeScript definitions


jakeleventhal:
### Verify canary release

- [x] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

This is a lint issue

### Which canary version will you have in your reproduction?

2.4.0

### Enviroment information

```block
turbo 2.4.0

CLI:
   Version: 2.4.0
   Path to executable: /Users/jakeleventhal/Developer/rip-technologies/node_modules/.pnpm/turbo-darwin-arm64@2.4.0/node_modules/turbo-darwin-arm64/bin/turbo
   Daemon status: Running
   Package manager: pnpm9

Platform:
   Architecture: aarch64
   Operating system: macos
   WSL: false
   Available memory (MB): 32308
   Available CPU cores: 10

Environment:
   CI: None
   Terminal (TERM): xterm-256color
   Terminal program (TERM_PROGRAM): vscode
   Terminal program version (TERM_PROGRAM_VERSION): 0.45.10
   Shell (SHELL): /bin/zsh
   stdin: false
```

### Expected behavior

```ts
import turboConfig from 'eslint-config-turbo/flat';
```
Works

### Actual behavior

```ts
import turboConfig from 'eslint-config-turbo/flat';
```
Gives a type error:
```
Could not find a declaration file for module 'eslint-config-turbo/flat'. '/Users/jakeleventhal/Developer/rip-technologies/node_modules/.pnpm/eslint-config-turbo@2.4.0_eslint@9.19.0_jiti@2.3.3__turbo@2.4.0/node_modules/eslint-config-turbo/flat/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/eslint-config-turbo` if it exists or add a new declaration (.d.ts) file containing `declare module 'eslint-config-turbo/flat';`
```

### To Reproduce

Import flat config

PR #9976, opened 2025-02-14
fix: create types for ESLint config/plugin

anthonyshew:
### Description

Creates types for `eslint-config-turbo`.

Addresses #9919


packages/eslint-plugin-turbo/package.json
```diff
@@ -24,7 +24,6 @@
     "dist/**"
   ],
   "scripts": {
-    "release": "pnpm build && pnpm publish",
```
anthonyshew, line 27: Removing vestigial from prior iterations of our publishing flows. This isn't being used anymore.

packages/eslint-config-turbo/flat/index.ts
```diff
@@ -0,0 +1,16 @@
+import plugin from "eslint-plugin-turbo";
+import {} from "eslint";
+
+const thing = [
```
chris-olszewski, line 3: Can we get away with `export default = [...` instead of the variable? If not can we rename it to something better than `thing`?

chris-olszewski (approved):
LGTM

packages/eslint-config-turbo/flat/index.ts
tknickman: Agree, let's use a better name!

anthonyshew: Haha, I was trying to make an error go away, my bad.

anthonyshew:
Tried it out manually. Looks good!

PR #9978, opened 2025-02-14
fix: exports for `eslint-config-turbo`

anthonyshew:
### Description

Mis-mapped the exports conditions in package.json. 🤦


tknickman (approved):

packages/eslint-config-turbo/package.json
```diff
@@ -43,6 +58,12 @@
     "@turbo/eslint-config": "workspace:*",
     "@turbo/tsconfig": "workspace:*",
     "@types/eslint": "^8.56.10",
+    "bunchee": "^6.3.4",
     "tsup": "^6.2.0"
```
tknickman, line 62: Can be removed

packages/eslint-config-turbo/src/index.ts
tony-cocco, line 1: Sorry if this is the wrong spot for this, but it helped resolve an issue. When I upgraded from turbo@2.4.2 to 2.4.4 my eslint config broke while following [the listed docs for legacy usage](https://github.com/vercel/turborepo/blob/main/packages/eslint-config-turbo/README.md#usage-legacy-eslintrc).

I updated to use this files `extends` and my eslint config was resolved.


Old:

```
extends: ["turbo"]
```

Resulted in:

```
eslint-config-turbo is invalid:
  - Unexpected top-level property "default".
 ```

Should the docs be updated to suggest using this instead?

New:
```
extends: ["plugin:turbo/recommended"]
```

mamlzy: sorry for tagging @anthonyshew, because this is important for people that not using eslint flat config

anthonyshew: No worries. Thanks for the tag. Will fix.

anthonyshew: Here's the fix: https://github.com/vercel/turborepo/pull/10105

Sorry about that!

PR #9984, opened 2025-02-16
fix(docs): correct grammar in turborepo setup message

Naresh-Khatri:
### Description

Fixed a grammatical error in the setup message:  
"You now up and running with Turborepo" -> "You're now up and running with Turborepo."


anthonyshew (approved):
Thanks!

PR #9995, opened 2025-02-17
chore(signals): prefactor

chris-olszewski:
### Description

Prefactors:
 - moved signal listening to own crate
 - return signal received to subscribers
 - support streams instead of single futures for signal sources

PR that prefactors the signals code to prepare for:
 - Subscribers getting all signals sent to `turbo`
 - Additional signal handlers to be sent transparently to children

There are no behavior changes in this PR. I highly suggest reviewing each commit on its own.

### Testing Instructions

Existing unit tests.

Quick manual verification that existing `SIGINT` behavior remains the same:
```
$ turbo_dev watch build
...
│    Finalizing page optimization  .   Collecting build traces  .
└────>
  × Watch interrupted due to signal
```

### DIFF
--- a/crates/turborepo-signals/Cargo.toml
+++ b/crates/turborepo-signals/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "turborepo-signals"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+futures = "0.3.30"
+tokio = { workspace = true, features = ["full", "time"] }
+
+[dev-dependencies]
+
+[lints]
+workspace = true
--- a/crates/turborepo-signals/Cargo.toml
+++ b/crates/turborepo-signals/Cargo.toml
@@ -6,6 +6,7 @@ license = "MIT"
 
 [dependencies]
 futures = "0.3.30"
+thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full", "time"] }
 
 [dev-dependencies]
--- a/crates/turborepo-signals/Cargo.toml
+++ b/crates/turborepo-signals/Cargo.toml
@@ -2,6 +2,7 @@
 name = "turborepo-signals"
 version = "0.1.0"
 edition = "2021"
+license = "MIT"
 
 [dependencies]
 futures = "0.3.30"
--- a/crates/turborepo-signals/src/lib.rs
+++ b/crates/turborepo-signals/src/lib.rs
@@ -8,13 +8,15 @@ pub mod signals;
 
 use std::{
     fmt::Debug,
-    future::Future,
     sync::{Arc, Mutex},
 };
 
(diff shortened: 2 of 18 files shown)


crates/turborepo-signals/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "turborepo-signals"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+futures = "0.3.30"
```
NicholasLYang, line 8: Any particular reason we're using a different version for futures instead of the workspace root version?

crates/turborepo-signals/src/lib.rs
```diff
@@ -130,15 +144,25 @@ impl HandlerState {
...
     use std::{assert_matches::assert_matches, time::Duration};
 
+    use futures::stream;
+
     use super::*;
 
+    #[cfg(windows)]
+    const DEFAULT_SIGNAL: Signal = Signal::CtrlC;
+    #[cfg(not(windows))]
+    const DEFAULT_SIGNAL: Signal = Signal::Interrupt;
```
NicholasLYang, line 154: Nit, but we could do and avoid the conditional compilation
```suggestion
    const DEFAULT_SIGNAL: Signal = if cfg!(windows) { Signal::CtrlC } else {Signal::Interrupt };
```

crates/turborepo-signals/src/listeners.rs
```diff
@@ -0,0 +1,33 @@
+use futures::{stream, Stream};
+
+use crate::signals::Signal;
+
+#[cfg(windows)]
+/// A listener for Windows Console Ctrl-C events
+pub fn get_signal() -> Result<impl Stream<Item = Option<Signal>>, std::io::Error> {
```
NicholasLYang, line 7: Minor detail, but maybe we should make this a custom error type so we can annotate the `run::Error` type with `#[from]` and avoid having to use `map_err` when calling this?

NicholasLYang (approved):
Looks good, just a couple questions

crates/turborepo-signals/Cargo.toml
chris-olszewski: I was matching the version used in [`turborepo-lib`](https://github.com/vercel/turborepo/pull/9995/files#diff-217f728c3ab310388e85c7ed8a02edc124d9b28c82b9cf623a41b94204848d41L61). I can update both to use workspace version in a followup PR.

crates/turborepo-signals/src/lib.rs
NicholasLYang: I'm confused. Isn't this equivalent to your existing code, but just using a `cfg!` macro instead of the `cfg` directive?

PR #9996, opened 2025-02-17
Improve GH action caching docs in github-actions.mdx

apetta:
### Description

The local caching in GH actions can be improved by using `ref_name` instead of the commit SHA as the latter would be different on every push & hence a cache-miss. While the restore-key may help, caching across branches will allow for cache-hit & branch-based cache isolation.

### DIFF
--- a/docs/repo-docs/guides/ci-vendors/github-actions.mdx
+++ b/docs/repo-docs/guides/ci-vendors/github-actions.mdx
@@ -290,7 +290,7 @@ Example `ci` yaml with `.turbo` as chosen cache folder:
           uses: actions/cache@v4 // [!code highlight]
           with: // [!code highlight]
             path: .turbo // [!code highlight]
-            key: ${{ runner.os }}-turbo-${{ github.sha }} // [!code highlight]
+            key: ${{ runner.os }}-turbo-${{ github.ref_name }} // [!code highlight]
             restore-keys: | // [!code highlight]
               ${{ runner.os }}-turbo- // [!code highlight]
 


anthonyshew (approved):
Sounds good to me! Thanks!

docs/repo-docs/guides/ci-vendors/github-actions.mdx
```diff
@@ -290,7 +290,7 @@ Example `ci` yaml with `.turbo` as chosen cache folder:
           uses: actions/cache@v4 // [!code highlight]
           with: // [!code highlight]
             path: .turbo // [!code highlight]
-            key: ${{ runner.os }}-turbo-${{ github.sha }} // [!code highlight]
+            key: ${{ runner.os }}-turbo-${{ github.ref_name }} // [!code highlight]
             restore-keys: | // [!code highlight]
               ${{ runner.os }}-turbo- // [!code highlight]
```
asapo, line 295: My understanding is that once a cache is created, it cannot be updated.
https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache

If we use `ref_name`, the default branch (eg. `main`, `master`) would be the most inefficient.

Adittionally, I'd like to discuss about the isolation.
Caches are already isolated by branches.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

In addition to this behavior, a cache from the default branch will be used by feature branches.

> Workflow runs can restore caches created in either the current branch or the default branch (usually main).

Therefore, my current understanding is,

- Create a cache with the default branch (cache-hash_main)
- Create a feature branch (`feature-a`) and run CI
  - `cache-hash_main` will be used
  - `cache-hash_feature-a_1` will be created
- Push a commit to `feature-a` and run CI
  - `cache-hash_feature-a_1` will be used
  - `cache-hash_feature-a_2` will be created

I'm not sure the better way. But using `ref_name` might not be the better way.

apetta: You're absolutely right, thanks for the additional context.

SHA seems to be the optimal approach given the restrictions. Each run will be a cache miss by default due to changing SHA, but the `restore-keys` should facilitate a cache hit agains the most recent cache - which is desirable behaviour. Additionally, re-runs of previous actions will have a direct cache-hit (if within cache life window).

Closing PR.

PR #10002, opened 2025-02-19
feat(boundaries): support tsconfig path aliases

NicholasLYang:
### Description
Support tsconfig path aliases such as:

```json
{
  "compilerOptions": {
    "paths": {
      "@/*": ["./*"]
    }
  }
}
```

We check if we can resolve the import as a path alias and if so, check the resolved path

Can be reviewed commit by commit

### Testing Instructions

Turns out we already had a test for this in `turborepo-tests/integration/fixtures/boundaries/apps/my-app/types.ts`. Added another import and updated snapshot

### DIFF
--- a/crates/turborepo-lib/src/boundaries/tsconfig.rs
+++ b/crates/turborepo-lib/src/boundaries/tsconfig.rs
@@ -3,6 +3,8 @@ use std::{collections::HashMap, sync::Arc};
 use oxc_resolver::{Resolver, TsConfigSerde};
 use turbopath::{AbsoluteSystemPath, AbsoluteSystemPathBuf};
 
+use crate::boundaries::BoundariesResult;
+
 pub struct TsConfigLoader<'a> {
     configs: HashMap<AbsoluteSystemPathBuf, Arc<TsConfigSerde>>,
     resolver: &'a Resolver,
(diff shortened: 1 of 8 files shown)


crates/turborepo-repository/src/package_graph/mod.rs
```diff
@@ -88,8 +88,8 @@ impl PackageInfo {
     }
 }
 
-type PackageKey = String;
-type PackageVersion = String;
+pub type PackageKey = String;
+pub type PackageVersion = String;
```
chris-olszewski, line 92: I don't like these type aliases, can we keep them contained to this crate?

crates/turborepo-lib/src/boundaries/tsconfig.rs
```diff
@@ -0,0 +1,32 @@
...
+            resolver,
+        }
+    }
+
+    pub fn load(&mut self, path: &AbsoluteSystemPath) -> Option<Arc<TsConfigSerde>> {
+        for dir in path.ancestors() {
+            if let Some(config) = self.configs.get(dir) {
+                return Some(config.clone());
+            }
+            if let Ok(config) = self.resolver.resolve_tsconfig(dir) {
```
chris-olszewski, line 24: Maybe add debug log when we fail to resolve a ts config?

chris-olszewski (approved):

crates/turborepo-repository/src/package_graph/mod.rs
NicholasLYang: I didn't want to have the type be `BTreeMap<String, String>`, since that's sort of opaque. Should I just copy them over?

PR #10006, opened 2025-02-19
chore: refactor `should_print_version` into helper for clarity

erikareads:
### Description

Small refactor pulling the boolean condition `should_print_version` into a helper function with the same name. I also decompose the boolean condition into the two relevant states: whether print version is disabled by environment variable, and whether this run of `turbo` is in continuous integration.

I used `enum`s to disambiguate these states, because previously we were negating a negative boolean to produce a positive one, and it would easy during modification to inadvertently produce opposite behavior by forgetting a negation. The enums make it clear which state each of the two relevant conditions are in, and make the final boolean calculation easy to read.

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -1132,6 +1132,43 @@ impl Display for LogPrefix {
     }
 }
 
+#[derive(PartialEq)]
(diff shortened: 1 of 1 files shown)


chris-olszewski (approved):
Thanks for this clean up!

PR #10008, opened 2025-02-19
chore: extract telemetry_handle creation into helper function

erikareads:
### Description

Small refactor pulling the logic for initializing the `telemetry_handle` and client into a helper function named `initialize_telemetry_client`. Previously, this initialization logic required a clarifying comment as well as nested match statements in the body of `run` which required time to read and understand.

This PR moves those nested match statements into a helper, reducing the cognitive load while reading `run`. This helper also centralizes logic for initialization, and allows us to remove a `mut` from `telemetry_handle`, since it's no longer being modified outside the helper.

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -1132,6 +1132,28 @@ impl Display for LogPrefix {
     }
 }
 
+fn initialize_telemetry_client(
+    color_config: ColorConfig,
+    version: &str,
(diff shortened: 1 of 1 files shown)


chris-olszewski (approved):
Nice!

Totally out of scope of the PR, but it would be awesome if we could move `AnonAPIClient::new` to a different thread. On `>>> FULL TURBO` run it accounts for 73ms/328ms.
Screenshot 2025-02-19 at 3 51 53 PM

PR #10015, opened 2025-02-20
docs: remove --continue mention with --parallel

anthonyshew:
### Description

Minor correction. `--parallel` does not automatically exhibit `--continue` behavior. You would still need to use `--continue` if you desire such behavior.

### Testing Instructions

👀


jenseng:
Awesome, thanks for the quick turnaround!

chris-olszewski (approved):

PR #10019, opened 2025-02-20
docs: corrections for Nx page

anthonyshew:
### Description

@jamesHenry pointed out some inaccuracies on the Nx page, so correcting them here! 

I apologize for these. I'm not an Nx expert and, as made more clear in this PR, the sections in the "Why switch?" come from Nx users looking to migrate. I tried my best to bring together their words, but missed the mark. Namely, I learned that Nx's plugins are optional to use and I mis-characterized how Nx's starter templates work.


tknickman (approved):

chris-olszewski (approved):

anthonyshew:
Forgot to link James' PR!: https://github.com/vercel/turborepo/pull/10003

PR #10021, opened 2025-02-21
correct package names

Mearman:


chris-olszewski (changes requested):
Can you explain the change from `docs` to `@repo/docs`? The rest of this example uses `@acme` as it's prefix.

Could you also update references to `docs` in the `README.md` to reference the new package name?

anthonyshew:
Hey, @Mearman. The convention that we use is to use prefixes for Internal Packages, not the applications. You'll see this in the rest of the examples as well.

Appreciate the thought, though, and you taking the time to open a pull request!

Mearman:
@anthonyshew ah I think it might still be wrong though

PR #10022, opened 2025-02-21
docs(prune): document how prune differs from pnpm deploy

chris-olszewski:
### Description

I received a question about how `pnpm deploy` and `turbo prune` differ and figured I would document my findings since they both server similar purposes.

### Testing Instructions

👀


anthonyshew (approved):
Muy bueno. Muchas gracias, amigo.

PR #10023, opened 2025-02-21
feat(turborepo): --continue=dependencies-successful

jenseng:
### Description

Context: https://github.com/vercel/turborepo/discussions/7461

Convert `--continue` into a tri-state option, introducing a new mode that allows independent tasks to continue running when an error occurs (dependent tasks will be canceled).

The possible values for `--continue` are now:
- `never` (default) -- on error: cancel all tasks
- `dependencies-successful` -- on error: continue running tasks whose dependencies have succeeded
- `always` -- on error: continue running all tasks, even those whose dependencies have failed

Setting `--continue` without a value is equivalent to setting it to `always`

### Testing Instructions

See modified/added tests in the PR, but basically:
- `turbo run <cmd> --continue=never` (or omitting it) will abort when an error occurs
- `turbo run <cmd> --continue=dependencies-successful` will cancel dependent tasks when an error occurs, but keep running others
- `turbo run <cmd> --continue=always` (or `--continue`) will keep running all other tasks when an error occurs


turborepo-tests/helpers/setup_integration_test.sh
```diff
@@ -43,7 +43,7 @@ fi
 
 "${TURBOREPO_TESTS_DIR}/helpers/copy_fixture.sh" "${TARGET_DIR}" "${FIXTURE_NAME}" "${TURBOREPO_TESTS_DIR}/integration/fixtures"
 "${TURBOREPO_TESTS_DIR}/helpers/setup_git.sh" "${TARGET_DIR}"
-"${TURBOREPO_TESTS_DIR}/helpers/setup_package_manager.sh" "${TARGET_DIR}" "$PACKAGE_MANAGER"
+. "${TURBOREPO_TESTS_DIR}/helpers/setup_package_manager.sh" "${TARGET_DIR}" "$PACKAGE_MANAGER"
```
jenseng, line 46: setup_package_manager.sh exports PATH, so we need to source it for that to actually take effect (my local dev setup needed this for corepack to work correctly for integration tests)

i could move this to a separate PR though if you prefer

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -895,7 +973,11 @@ impl ExecutionArgs {
         // default to false
         track_usage!(telemetry, self.framework_inference, |val: bool| !val);
 
-        track_usage!(telemetry, self.continue_execution, |val| val);
+        telemetry.track_arg_value(
```
jenseng, line 976: since we emit new values here (none/independent-tasks-only/all), should this be done under a different key? otherwise do we risk breaking consumers of this data who might expect the old values (set/default)?

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -811,6 +830,63 @@ fn path_non_empty(s: &str) -> Result<Utf8PathBuf, String> {
...
+#[derive(Copy, Clone, Debug, PartialEq)]
+struct ContinueModeParser;
+
+impl ContinueModeParser {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+// slightly redundant, but this allows us to transform deprecated values
```
jenseng, line 842: I'm torn on whether this is too much. Some alternatives include:
  - a simpler `PossibleValuesParser` that maps to correct enum values, emitting deprecation warnings. cons: true / false appear in the generated help text. but we could just hide all of them via `hide_possible_values`
  - include deprecated values in the enum and just use `ValueEnum`. cons: true / false appear in the generated help text, and downstream logic is more complicated

jenseng (commented):
A couple high level thoughts on the UX:

* Do we want to deprecate specifying `--continue` without a value?
* `--continue=independent-tasks-only` is a bit verbose. What if we did something like `--continue=independent`?
* Or we could deprecate `--continue` entirely and have a new option that inverts the phrasing, e.g. something like `--on-error` with values like `cancel`, `cancel-dependents` and `continue`

Also I'm a rust noob, so please let me know where I can improve! 😅

docs/repo-docs/reference/run.mdx
```diff
@@ -121,13 +121,18 @@ turbo run build --concurrency=50%
...
+Specify how `turbo` should handle current and pending tasks in the presence of an error (e.g. non-zero exit code from a task).
 
-Continue with task execution in the presence of an error (e.g. non-zero exit code from a task).
+- When `--continue` is `none` and an error occurs, `turbo` will cancel all tasks.
+- When `--continue` is `independent-tasks-only` and an error occurs, `turbo` will cancel all dependent tasks but continue running other independent tasks.
+- When `--continue` is `all` and an error occurs, `turbo` will continue running pending tasks (including dependents).
+- When `--continue` is specified without a value, it will default to `all`.
 
-When `--continue` is `true`, `turbo` will exit with the highest exit code value encountered during execution.
+In all cases, `turbo` will exit with the highest exit code value encountered during execution.
```
jenseng, line 135: actually i haven't verified if this is true in the `none` case 🤔

jenseng: based on both the previous and updated [continue.t output](https://github.com/vercel/turborepo/pull/10023/files#diff-950eab16b0680441205540b1c3080675aa365640711c5632d17708c99a9ae914), the exit code is not propagated upwards in any cases (`run build exited (1)` despite `exit 2`). so that sounds like an existing bug (or doc issue?). we could potentially address it in this PR and/or just remove this line from the docs, wdyt?

crates/turborepo-lib/src/cli/mod.rs
chris-olszewski: We can make it a simple `PossibleValuesParser`. Since `2.0.0` `--continue=true|false` has not been accepted so we do not need to support them or emit deprecation warnings.

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -824,10 +900,12 @@ pub struct ExecutionArgs {
...
     pub concurrency: Option<String>,
-    /// Continue execution even if a task exits with an error or non-zero
-    /// exit code. The default behavior is to bail
-    #[clap(long = "continue")]
-    pub continue_execution: bool,
+    /// Specify which tasks should continue running when an error occurs.
+    /// Use "none" to cancel all tasks.
+    /// Use "independent-tasks-only" to continue running independent tasks and
+    /// cancel dependent ones. Use "all" to continue running all tasks.
+    #[clap(long = "continue", value_name = "CONTINUE", num_args = 0..=1, default_value = "none", default_missing_value = "all", value_parser = ContinueModeParser::new())]
```
chris-olszewski, line 907: With not needing to support `true`/`false` we can just use `EnumValueParser`
```suggestion
    #[clap(long = "continue", value_name = "CONTINUE", num_args = 0..=1, default_value = "none", default_missing_value = "all", value_parser = EnumValueParser::<ContinueMode>::new())]
```

crates/turborepo-lib/src/cli/mod.rs
chris-olszewski: Just for to keeping data clean as we'll have events coming in from old/new versions we should make this a new key and keeping a proxy key.
We can change the old `track_usage!` call to:
```
track_usage!(telemetry, self.continue_execution, |val| matches!(val, ContinueMode::All | ContinueMode::IndependentTasksOnly));
```
It doesn't properly capture use of `--continue=none` vs not having `--continue` set, but I think that's fine. Merging these events is worth it for having defaults encoded in `clap`.

crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -895,7 +973,11 @@ impl ExecutionArgs {
         // default to false
         track_usage!(telemetry, self.framework_inference, |val: bool| !val);
 
-        track_usage!(telemetry, self.continue_execution, |val| val);
+        telemetry.track_arg_value(
+            "continue-execution",
```
chris-olszewski, line 977:
```suggestion
            "continue-execution-strategy",
```

chris-olszewski (changes requested):
Thank you so much for this contribution this is a great feature!

Only some very minor suggestions.

docs/repo-docs/reference/run.mdx
chris-olszewski: Yes, `npm` has had odd behavior around exit codes from underlying scripts across versions. We do take the [highest exit code](https://github.com/vercel/turborepo/blob/main/crates/turborepo-lib/src/run/mod.rs#L489) from the process we spawn (or `1` if the error was not from a task exiting such as a task failing to start etc).

turborepo-tests/integration/tests/no-args.t
```diff
@@ -98,8 +98,8 @@ Make sure exit code is 2 when no args are passed
             Override the filesystem cache directory
         --concurrency <CONCURRENCY>
             Limit the concurrency of task execution. Use 1 for serial (i.e. one-at-a-time) execution
-        --continue
-            Continue execution even if a task exits with an error or non-zero exit code. The default behavior is to bail
+        --continue [<CONTINUE>]
+            Specify which tasks should continue running when an error occurs. Use "none" to cancel all tasks. Use "independent-tasks-only" to continue running independent tasks and cancel dependent ones. Use "all" to continue running all tasks
```
chris-olszewski, line 102: Unsure of why the actual diff isn't showing on the prysk diff output, but it's missing the possible values
```suggestion
            Specify which tasks should continue running when an error occurs. Use "none" to cancel all tasks. Use "independent-tasks-only" to continue running independent tasks and cancel dependent ones. Use "all" to continue running all tasks [default: none] [possible values: none, independent-tasks-only, all]
```

crates/turborepo-lib/src/cli/mod.rs
jenseng: was able to omit the explicit value_parser altogether, since afaict the ValueEnum stuff is automatic

turborepo-tests/integration/tests/no-args.t
jenseng: fixed now... i ran into this issue a few times while iterating on the PR, my hunch is that prysk is treating `[...]` as a character class or something? 🤔

crates/turborepo-graph-utils/src/walker.rs
```diff
@@ -104,14 +110,15 @@ impl<N: Eq + Hash + Copy + Send + 'static> Walker<N, Start> {
                             trace!("Receiver was dropped before walk finished without calling cancel");
                             return;
                         }
-                        if callback_rx.await.is_err() {
+                        let callback_result = callback_rx.await;
+                        if callback_result.is_err() {
```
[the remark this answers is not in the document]
jenseng: sounds good, i addressed the other stuff and will tweak this in a bit 👍

anthonyshew:
Could I introduce some bikeshedding on the name a bit? Thinking of things like:
- `--continue=require-dependents`
- `--continue=no-cascading-failure`
- `--continue=dependencies`

For some reason, `--continue=independent-tasks-only` isn't speaking to me. Do you both love it?

jenseng:
> For some reason, `--continue=independent-tasks-only` isn't speaking to me. Do you both love it?

Yeah I'm not particularly attached, I just went with the original idea from the discussion topic. Though I do like it a hair better than the other bikeshed suggestions, i.e. just like `all` and `none`, it answers the question "what are we going to continue?"

Or bikeshedding another direction, we could have this be a new option entirely and deprecate `continue`. e.g. something like `--on-error` with values like `halt`, `cancel-dependents` and `continue`

mehulkar:
Another idea is to invert the flag to invert the flag:

```
--on-failure=continue-independent
--on-failure=cancel-all
--on-failure=cancel-dependents
```

anthonyshew:
I do like that idea, though it will take a deprecation and breaking change to get back to having unified to one flag. I'm not sure how valuable `--on-failure` is vs. `--continue` in that light, especially if we can come up with the right name using `--continue`. 🤔

jenseng:
a couple more ideas: 
* `--continue=independent`
* `--continue=isolated`
* rework continue to communicate **_when_** instead of **_what_**, changing `none`/`all` -> `never`/`always` and then doing one of:
   * `--continue=conditional`
   * `--continue=dependent`
   * `--continue=contingent`

jenseng:
or if we think of continue in terms of "which part of the graph do we keep running?", then maybe something like:
* `--continue=successful`
* `--continue=successful-only`
* `--continue=passing`
* `--continue=eligible`

JavaScriptBach:
`--continue=dependencies-satisfied`?

chris-olszewski:
> rework continue to communicate when/how instead of what we run, changing none/all -> never/always and then doing one of

I think this is a good reframing, combining this with @JavaScriptBach's suggestion leaves a good taste in my mouth.

 - `--continue=always` -> `--continue` today
 - `--continue=dependencies-successful` -> feature from this PR
 - `--continue=never` -> default today

anthonyshew:
@chris-olszewski Nailed it. So clear what each of those mean. And thanks @mehulkar, @jenseng, and @JavaScriptBach for humoring me.

jenseng:
Ok, updated! Let me know if any help text/etc needs further tweaking 🙏

chris-olszewski (approved):

jenseng:
The windows tests have been failing consistently in this PR and I haven't seen that failure on other test runs. I did a debug run [here](https://github.com/jenseng/turborepo/actions/runs/13507373567) where I undid [this change to setup_integration_tests.sh](https://github.com/vercel/turborepo/pull/10023/files#diff-91e13be999d081532eb776c11fc5ca1aa07f240c19e350b62930208472074498R46), and that seems to have fixed it 🙃 ... not sure how that would break things, maybe hitting a weird bug/edge case in Git for Windows' bash.exe? 🤷 ... doing a couple more test runs to confirm, and if that theory holds i'll revert that line

jenseng:
> The windows tests have been failing consistently in this PR and I haven't seen that failure on other PRs/pushes. I did a debug run [here](https://github.com/jenseng/turborepo/actions/runs/13507373567) where I undid [this change to setup_integration_tests.sh](https://github.com/vercel/turborepo/pull/10023/files#diff-91e13be999d081532eb776c11fc5ca1aa07f240c19e350b62930208472074498R46), and that seems to have fixed it 🙃 ... not sure how that would break things, maybe hitting a weird bug/edge case in Git for Windows' bash.exe? 🤷 ... doing a couple more test runs to confirm, and if that theory holds i'll revert that line

Yeah looking quite likely to be the culprit, since the [new PATH value](https://github.com/jenseng/turborepo/actions/runs/13507986671/job/37741919109) is inappropriate for bash.exe and likely causing trouble. i.e. it's setting something like: `C:\Users\RUNNER~1\AppData\Local\Temp\prysk-tests-nurs9w2d/corepack:/d/a/turborepo/turborepo/turborepo-tests/integration/node_modules/.bin:...` (note the backslashes and the unintended `:` separator)

I'll revert that line from this PR and revisit it in another PR (I think it's a good change to make, just needs to work on windows)

docs/repo-docs/reference/run.mdx
```diff
@@ -121,13 +121,18 @@ turbo run build --concurrency=50%
 turbo run test --concurrency=5
 ```
 
-### `--continue`
+### `--continue [option]`
```
anthonyshew, line 124:
```suggestion
### `--continue <option>`
```

Sorry for the nitpick. Just to be the same as the rest of the page (though you're making me think brackets might be better..)

vladkampov:
Hello, @jenseng. Thanks for this update. 

We started to get the error using `--continue` via proxy library `turbo-tools` ([here's where it is failing](https://github.com/benduran/turbo-tools/blob/99e9e55612c8ae8bb2fe3604881e12382032d0e6/src/commands/publish.js#L99 ))

The error we are getting is 
```bash
npx turbo run --continue test --filter="@prod-ops/delivery-list" --filter="@prod-ops/vertical-gantt" in /Users/vladkampov/netflix/prod-ops-ui
 ERROR  invalid value 'test' for '--continue [<CONTINUE>]'
  [possible values: never, dependencies-successful, always]

node:internal/errors:984
  const err = new Error(message);
              ^

Error: Command failed: npx turbo run --continue test --filter="@prod-ops/delivery-list" --filter="@prod-ops/vertical-gantt"
```

It seems that with the change in 2.4.4, turbo is strict about the order of the flags used.

 It simply may be worked around if we change the order ([like in the PR](https://github.com/benduran/turbo-tools/compare/main...vladkampov:turbo-tools:patch-1)), however, you might want to work around it differently or document this quirk. 
 
 Thanks.

jenseng:
@vladkampov sorry about that, that was an unfortunate oversight on my part! v2.4.5 will have a fix for this, but in the meantime there are a few things you can do:

- switch the order of the arguments
- set an explicit value, ideally with `=`, e.g. `--continue=always`
- pin to v2.4.3

vladkampov:
> @vladkampov sorry about that, that was an unfortunate oversight on my part! v2.4.5 will have a fix for this, but in the meantime there are a few things you can do:
> 
> * switch the order of the arguments
> * set an explicit value, ideally with `=`, e.g. `--continue=always`
> * pin to v2.4.3

No worries!
Thanks for the swift response and for the feature introduced with this PR!

PR #10024, opened 2025-02-22
chore: update to Rust 1.85.0

ognevny:
### Description

- Vec::extract_if signature changed: https://github.com/rust-lang/rust/pull/133265
- crates/turborepo-auth/src/auth/sso.rs: allow 'foramatting_args' in literal string (should it be put in `format!` instead?)
- async_closure is now stable
- further `map_or` simplification

### DIFF
--- a/crates/turborepo-auth/src/auth/sso.rs
+++ b/crates/turborepo-auth/src/auth/sso.rs
@@ -78,7 +78,7 @@ pub async fn sso_login<T: Client + TokenClient + CacheClient>(
                         api_client,
                         sso_team,
                         Some(valid_token_callback(
-                            "Existing Vercel token for {sso_team} found!",
+                            &format!("Existing Vercel token for {sso_team} found!"),
                             color_config,
                         )),
                     )
--- a/crates/turborepo-globwatch/src/lib.rs
+++ b/crates/turborepo-globwatch/src/lib.rs
@@ -214,7 +214,7 @@ impl GlobWatcher {
                                 // requester. flushes should not be considered as events.
                                 for flush_id in e
                                     .paths
-                                    .extract_if(|p| p.starts_with(flush_dir.as_path()))
+                                    .extract_if(.., |p| p.starts_with(flush_dir.as_path()))
                                     .filter_map(|p| {
                                         get_flush_id(
                                             p.strip_prefix(flush_dir.as_path())
--- a/crates/turborepo-api-client/src/lib.rs
+++ b/crates/turborepo-api-client/src/lib.rs
@@ -1,4 +1,3 @@
-#![feature(async_closure)]
 #![feature(error_generic_member_access)]
 #![feature(assert_matches)]
 #![deny(clippy::all)]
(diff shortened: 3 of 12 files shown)


ognevny:
not migrating to edition 2024 because it requires lots of code/formatting changes

ognevny:
Windows test failure seems unrelated
```
thread 'globwatcher::test::watch_single_file' panicked at crates\turborepo-lib\src\globwatcher\mod.rs:684:9:
assertion `left == right` failed: expected one of the globs to have changed
  left: {}
 right: {"my-pkg/.next/next-file"}
```

crates/turborepo-auth/src/auth/sso.rs
```diff
@@ -78,6 +78,7 @@ pub async fn sso_login<T: Client + TokenClient + CacheClient>(
                         api_client,
                         sso_team,
                         Some(valid_token_callback(
+                            #[allow(clippy::literal_string_with_formatting_args)]
```
NicholasLYang, line 81: yeah this probably should be a format

NicholasLYang (approved):
Thanks for doing this!

crates/turborepo-auth/src/auth/sso.rs
ognevny: done

crates/turborepo-globwatch/src/lib.rs
```diff
@@ -214,7 +214,7 @@ impl GlobWatcher {
                                 // requester. flushes should not be considered as events.
                                 for flush_id in e
                                     .paths
-                                    .extract_if(|p| p.starts_with(flush_dir.as_path()))
+                                    .extract_if(.., |p| p.starts_with(flush_dir.as_path()))
```
chris-olszewski, line 217: I'm not sure why, but I'm pretty positive this change is resulting in [this failure](https://github.com/vercel/turborepo/actions/runs/13516449852/job/37766021434?pr=10024#step:6:2200) as this change is in that codepath and the failure seems to be happening consistently in this PR. I have not seen this test be flakey in the past.

chris-olszewski (changes requested):

crates/turborepo-globwatch/src/lib.rs
chris-olszewski, line 217: I'm not sure how or why this is only affecting Windows, but this is on the codepath for the failing Windows test. I haven't seen the test be flakey before so I think the changes to this function might actually have a regression (or we were misusing it somehow).

chris-olszewski (changes requested):

crates/turborepo-globwatch/src/lib.rs
chris-olszewski: I've gotten this running up on my Windows machine and can reproduce it, but I do think this exact function is unrelated. It appears that the actual FS event isn't making it here where before the Rust bump it was.

Looking at what else in this Rust version bump could be causing the behavior change.

ognevny: > Sorry for the delay. Turns out this test was for some code that isn't even used anymore. I merged a PR to delete it, so if you rebase, hopefully CI should pass

thanks anyway. I'll rebase after I get home ❤️

ognevny: done

PR #10027, opened 2025-02-24
chore(process): 

chris-olszewski:
### Description

This is primarily a code move of the `ProcessManager` out of the `turborepo-lib` crate. We also now differ between a child shutting down in response to a `SIGINT` vs us killing the child.

Future PR will be changing how we treat each of these exit outcomes.

### Testing Instructions

Existing unit tests 👀


crates/turborepo-process/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "turborepo-process"
```
tknickman, line 2: Love it

PR #10044, opened 2025-02-25
chore: fix corepack wiring in integration tests

jenseng:
### Description

Spun out from https://github.com/vercel/turborepo/pull/10023#issuecomment-2679724673

setup_package_manager.sh exports `PATH`, so we need to source it for the isolated corepack shims to actually get used in tests. Otherwise certain npm integration tests can fail, depending on the global npm version and if corepack's npm shims haven't been explicitly enabled globally. This can lead to false positives and wasted time debugging integration tests locally.

This hasn't been an issue under GitHub Actions, since actions/setup-node is installing the same version of npm (10.5.0) that the tests expect, but could otherwise become problematic down the road.

Fixing this uncovers two new problems around testing under Windows, which this PR also addresses:

 - the corepack path needs to be POSIX-ified, since `C:\Users\RUNNER~1\...` contains backslashes and the name separator (:)
 - corepack uses `PATHEXT` (via [cmd-extension](https://www.npmjs.com/package/cmd-extension)) to determine the casing of its shims' file extension (e.g. npm.cmd vs npm.CMD), whereas node's bundled npm.cmd is always lowercase ... while we could update all the tests to accept both styles, the easiest thing to do is just to downcase `PATHEXT`.

### Testing Instructions

Integration tests 🥳

### DIFF
--- a/turborepo-tests/helpers/setup_integration_test.sh
+++ b/turborepo-tests/helpers/setup_integration_test.sh
@@ -43,7 +43,7 @@ fi
 
 "${TURBOREPO_TESTS_DIR}/helpers/copy_fixture.sh" "${TARGET_DIR}" "${FIXTURE_NAME}" "${TURBOREPO_TESTS_DIR}/integration/fixtures"
 "${TURBOREPO_TESTS_DIR}/helpers/setup_git.sh" "${TARGET_DIR}"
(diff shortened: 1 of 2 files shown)


chris-olszewski (approved):
Thanks for tracking this down!

PR #10049, opened 2025-02-26
feat(process): distinguish between signals used to kill children

chris-olszewski:
### Description

This is a redo of #10027 with some additional refactoring to reduce the complexity of this code. These changes should lessens the possibilities of deadlock or race conditions.

I highly recommend viewing each commit individually.

From original PR:
> This is primarily a code move of the `ProcessManager` out of the `turborepo-lib` crate. We also now differ between a child shutting down in response to a `SIGINT` vs us killing the child.

> Future PR will be changing how we treat each of these exit outcomes.

The additional commits in this PR:
 - Fixed a race condition between between the 2 `wait`s performed by the `ChildStateManager` where one was called by the task the sent a shutdown to the child and the other was the default `wait`. If the latter won, then it would appear another entity killed the child even when it was really us.
 - Changed the return value of `ShutdownStyle::process` to avoid returning an impossible state
 - Remove unused `ChildState::Exited` field which was never read
 - Removed shared state between the child handle and the child state manager
 - Simplified child methods

The removal of the shared state is possible since we already have the exit channel shared between the handles and the manager. Using both a channel and state was error prone as it lead to the following possibilities:
 - What does it mean if the manager is still listening for commands, but the state shows the child as exited
 - What does it mean if the manager channel is closed, but the state shows the child is still running
Instead we only use the command channel being open/closed as an indication of if the child is running. Once the manager sees the child exit, it will send the exit status via the channel and exit resulting in the channel being closed.

### Testing Instructions

Existing test suite, ran with `hyperfine 'cargo test' -r 500` to attempt to flush out any races.

### DIFF
--- a/crates/turborepo-process/src/child.rs
+++ b/crates/turborepo-process/src/child.rs
@@ -328,7 +328,7 @@ impl ShutdownStyle {
 
                     let result = tokio::time::timeout(*timeout, fut).await;
                     match result {
-                        // We ignore the exit code and mark it as killed since we sent a SIGINT
+                        // We ignore the exit code and mark it as interrupted since we sent a SIGINT
                         // This avoids reliance on an underlying process exiting with
                         // no exit code or a non-zero in order for turbo to operate correctly.
                         Ok(Ok(_exit_code)) => ChildExit::Interrupted,


crates/turborepo-process/src/child.rs
```diff
@@ -506,71 +484,29 @@ impl Child {
...
-                    None => return,
-                }
-            };
-
-            // if this fails, it's because the channel is dropped (toctou)
-            // we can just ignore it
-            child.stop().await.ok();
-        };
-
-        let (_, code) = join! {
```
chris-olszewski, line 526: With the lack of state, we no longer need to drive both futures to completion at the same time. Sending the stop command will complete:
 - Once the child manager reads the command from the channel
 - If the child manager exits and closes the command channel

After either of these we can `wait` for the exit code to be reported:
 - If the stop command was read, then the child manager will update the exit status based on the outcome of issuing the command
 - If the stop command was dropped, then the child manager has exited and will have dropped the sender for the exit status channel. If the sender for a watch channel is dropped then `changed()` will exit.

crates/turborepo-process/src/child.rs
```diff
@@ -351,13 +331,13 @@ impl ShutdownStyle {
                         // We ignore the exit code and mark it as killed since we sent a SIGINT
```
NicholasLYang, line 331: We should update this comment

NicholasLYang (approved):

PR #10066, opened 2025-03-02
chore(turbo_json): remove exterior mutability from loader

chris-olszewski:
### Description

`TurboJsonLoader::load` only took a mutable self reference because we were using `HashMap` to cache the results. This PR changes the data type we use to cache results and removes the exterior mutability.

We move to use a `FixedMap` which has the following properties:
 - Only works with the keys provided at construction
 - Append only
 - Thread safe

I went with this just because the implementation is straightforward and matches our use case since we know all of the possible places a `turbo.json` will be before we start loading them.

Future work can be reworking boundaries code to no longer eagerly load all of the `turbo.json`s at the start of execution.

### Testing Instructions

Existing test suite. Additional unit tests for `FixedMap`

### DIFF
--- a/crates/turborepo-fixed-map/src/lib.rs
+++ b/crates/turborepo-fixed-map/src/lib.rs
@@ -20,23 +20,6 @@ impl<K: Ord, V> FixedMap<K, V> {
         Self { inner }
     }
 
-    pub fn from_iter(entries: impl Iterator<Item = (K, Option<V>)>) -> Self {
-        let mut inner = entries
-            .map(|(key, value)| {
-                let value_slot = OnceLock::new();
-                if let Some(value) = value {
-                    value_slot
-                        .set(value)
-                        .map_err(|_| ())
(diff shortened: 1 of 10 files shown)


crates/turborepo-fixed-map/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+            .inner
+            .binary_search_by(|(k, _)| k.cmp(key))
+            .map_err(|_| UnknownKey)?;
+        let (_, value) = &self.inner[item_index];
+        Ok(value.get())
+    }
+
+    /// Insert a value for a key.
+    ///
+    /// There is not guarentee that the provided value will be the one returned.
```
NicholasLYang, line 37:
```suggestion
    /// There is no guarentee that the provided value will be the one returned.
```

NicholasLYang (approved):

PR #10079, opened 2025-03-03
feat(ls): add package path to ls

NicholasLYang:
### Description

Minor detail but sometimes I want to lookup where a package is located in the repo.

### Testing Instructions

Existing `command-ls.t` test should work


chris-olszewski (approved):
Looks like some more snapshots need to get updated but LGTM

PR #10080, opened 2025-03-04
fix(release): manually install target toolchain

dimitropoulos:
### Description

We discovered that Rust 1.28 release requires us to manually install the toolchain, not just the target.  This command should install both and ensure compatibility.

### Testing Instructions

Running the Turbo Repository Library Release workflow.


chris-olszewski (approved):
Testing changes [here](https://github.com/vercel/turborepo/actions/runs/13644739314)

Issue #3793, opened 2023-02-14
[turborepo] .jsonc support (Json with comments)


softmarshmallow:
### What version of Turborepo are you using?

v1.7.2

### What package manager are you using / does the bug impact?

Yarn v2/v3 (node_modules linker only)

### What operating system are you using?

Mac

### Describe the Bug

renaming turbo.json to turbo.jsonc

### Expected Behavior

search for turbo.jsonc for fallback

### To Reproduce

rename .json to .jsonc

mehulkar:
@softmarshmallow for what it's worth, you can have comments in `turbo.json`. I know it doesn't help with things like IDE formatting, but you should at least be able to write comments. Previously discussed here also: https://github.com/vercel/turbo/issues/1470

mehulkar:
Looks like in VSCode at least you can associate a file with a language, I've done that for this repo as well in #3812. Does that work for you @softmarshmallow?

anthonyshew:
This is supported behavior! Woohoo!

ari-becker:
@anthonyshew is it? I tried `turbo.jsonc` with `1.12.3` and I get an error `Could not find turbo.json`.

anthonyshew:
I suppose the specific implementation described in this issue isn't supported. I was doing some cleanup of old issues and saw the title + not much description so went on the goal, not the implementation. Apologies!

Given that the goal is "add comments to `turbo.json`", I'd figure we could call the intent of this issue satisfied since `turbo,.json` supports comments?

@mehulkar Or, alternatively, we could leave it open as a label:good-first-issue?

mehulkar:
i don’t feel that supporting an alternate file name is valuable, but since the issue was closed for a slightly different reason I wanted to give @ari-becker a chance to reply. I’m in favor of closing unless we discover a compelling reason.

ari-becker:
@mehulkar it's not even my issue originally, I just watched it for updates since I'm affected by it too 😅 

Please note that editors (e.g. WebStorm) can and do identify the comments as an error:

![image](https://github.com/vercel/turbo/assets/1388505/3278976e-4474-4324-8c10-d7ae855ac4e9)

This registers as a false-positive in IDEs and negatively impacts attempts to try to keep the number of problems in a project at zero.

[...]

PR #10083, opened 2025-03-04
feat: add JSONC support

anthonyshew:
### Description

JSONC allows the use of comments in JSON files. Until now, our configuration parser allowed comments in JSON by accident, which made for awkwardness talked about in #3793.

With this PR, we're now able to support `turbo.jsonc` in both the root configuration and in Package Configurations. The file will be properly used for hashing and users can use `.jsonc` that will work with their editors and other tooling.

### Testing Instructions

Tests have been added in this PR and I've hand-tested with `create-turbo`, both `turbo` and some packages. Would appreciate some more hand-testing, just in case, along with your code review.


chris-olszewski (approved):
This looks good to me. Maybe get another review for the Rust changes (especially `loader.rs`) since I rewrote some chunks.

anthonyshew:
Looks great to me. Thanks for all the assists.

neochief:
probably related: https://github.com/vercel/turborepo/issues/10323

olsonpm:
I appreciate the work done to make jsonc work.  For anyone running into a codemod issue like me - this PR missed a spot [at least here](https://github.com/vercel/turborepo/blob/148b43d12be7207879d0453fe7b61b4e74b512e2/packages/turbo-codemod/src/transforms/migrate-dot-env.ts#L74-L79):

```ts
const turboConfigPath = path.join(root, "turbo.json");
if (!fs.existsSync(turboConfigPath)) {
  return runner.abortTransform({
    reason: `No turbo.json found at ${root}. Is the path correct?`,
  });
}
```

and a grep for turbo.json at this commit head reveals potential other spots to be updated.

PR #10091, opened 2025-03-05
docs: add release phases to support policy

anthonyshew:
### Description

Documents our release phases.

Please check the feature lists for each phase for completeness/accuracy.

### Testing Instructions

👀


docs/repo-docs/getting-started/support-policy.mdx
```diff
@@ -69,3 +69,52 @@ We highly encourage you to use the most current version of Turborepo whenever po
...
+
+### Beta
+
+Beta means the API requires more confidence around reliability before a stable release. In the Beta phase, we are confident about the overall design of the feature and expect to be able to promote the feature to stable with relatively few changes. In rare cases, an API may require a breaking change and will not be covered by semver.
+
+We encourage early adopters to use Beta APIs to help us gain certainty around the feature's design and stability. Depending on your risk tolerance, you can use the feature in production as you see fit.
+
+**APIs currently in the Beta phase:**
+
+- [`--output=json` for the `affected` flag](/repo/docs/reference/run#--outputjson)
```
chris-olszewski, line 98: I think this is for `turbo ls --affected --output=json` not `turbo run` 
```suggestion
- [`--output=json` for the `affected` flag](/repo/docs/reference/ls#--output-format-)
```

docs/repo-docs/getting-started/support-policy.mdx
```diff
@@ -69,3 +69,52 @@ We highly encourage you to use the most current version of Turborepo whenever po
...
+
+Turborepo also offers specific APIs that may be released in an earlier phase of development. This is to allow early adopters to give valuable feedback and help the core team gain confidence in APIs ahead of their stable release. APIs that are available pre-stable will be marked as such in the documentation and within `turbo` itself whenever possible.
+
+Not every API will be released in every phase. We use our best judgement to determine the phase for an API based on our confidence for the feature, weighing factors like the feature's complexity, possible use cases, and feedback from early users of the API.
+
+The Turborepo core team would like to thank any users who do use features early and express our appreciation for any feedback you provide.
+
+### Stable
+
+When an API reaches stability, it is considered ready for use in production. APIs marked as stable will only change in a backward-compatible way in a major release.
```
chris-olszewski, line 85: Do we want to clarify that we do not consider unstructured terminal output as an API? For example https://github.com/vercel/turborepo/pull/10079 should not be considered a breaking change IMO. Now if it was a removal of a field from `turbo ls --output=json` it would be.

chris-olszewski (approved):

docs/repo-docs/getting-started/support-policy.mdx
anthonyshew: My gut reaction is to say no, but I could be talked into saying yes. This has come up _once_ in my memory, or has it been more?

Maybe a broader question: Do other CLI tools consider their stdout as semver-protected? I've never looked into or thought about this...

NicholasLYang: Relevant xkcd: https://xkcd.com/1172/

chris-olszewski: > Do other CLI tools consider their stdout/stderr as semver-protected if its meant for human readability?

It depends on the output, if it's meant to be human readable, then usually no. All for omitting this since this is usually not assumed to be an API.

anthonyshew: Aligned, let's omit.

docs/repo-docs/getting-started/support-policy.mdx
```diff
@@ -69,3 +69,52 @@ We highly encourage you to use the most current version of Turborepo whenever po
...
+
+- If a stable API is going to **have breaking changes** in an upcoming major release, warnings and documentation will be provided to help with migrations and next steps.
+- If a stable API is going to be **removed** in an upcoming major release, it will be marked as [Deprecated](#deprecated) to provide early notice of future removal.
+
+### Beta
+
+Beta means the API requires more confidence around reliability before a stable release. In the Beta phase, we are confident about the overall design of the feature and expect to be able to promote the feature to stable with relatively few changes. In rare cases, an API may require a breaking change and will not be covered by semver.
```
tknickman, line 92:
```suggestion
Beta APIs require more confidence around reliability before a stable release. In the Beta phase, we are confident about the overall design of the feature and expect to be able to promote the feature to stable with relatively few changes. In rare cases, an API may require a breaking change and will not be covered by semver.
```

PR #10092, opened 2025-03-05
docs: add hybrid approach to vitest guide

beaussan:
### Description

Added a hybrid approach to the vitest guide for monorepo

### Testing Instructions

- Look at the new vitest documentation

### DIFF
--- a/docs/repo-docs/guides/tools/vitest.mdx
+++ b/docs/repo-docs/guides/tools/vitest.mdx
@@ -217,7 +217,7 @@ Because of this, you'll need to use [Root Tasks](/repo/docs/crafting-your-reposi
 
 You can combine the benefits of both approaches by implementing a hybrid solution. This involves maintaining local Vitest configurations in each package while also setting up a root Vitest workspace configuration. This approach gives you the best of both worlds: better local development experience with global watch mode, while preserving Turborepo's caching benefits in CI.
 
-```json title="./vitest.workspace.ts"
+```ts title="./vitest.workspace.ts"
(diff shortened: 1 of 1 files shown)


docs/repo-docs/guides/tools/vitest.mdx
```diff
@@ -212,3 +212,39 @@ Because of this, you'll need to use [Root Tasks](/repo/docs/crafting-your-reposi
 ```
 
 **Notably, the file inputs for a Root Task include all packages by default, so any change in any package will result in a cache miss.** While this does make for a simplified configuration to create merged coverage reports, you'll be missing out on opportunities to cache repeated work.
+
+### Using a hybrid approach
+
+You can combine the benefits of both approaches by implementing a hybrid solution. This involves maintaining local Vitest configurations in each package while also setting up a root Vitest workspace configuration. This approach gives you the best of both worlds: better local development experience with global watch mode, while preserving Turborepo's caching benefits in CI.
```
anthonyshew, line 218:
```suggestion
You can combine the benefits of both approaches by implementing a hybrid solution.This approach unifies local development using Vitest's Workspace approach while preserving Turborepo's caching in CI.  This comes at the tradeoff of slightly more configuration and a mixed task running model in the repository.
```

beaussan:
> Just one comment from me around wording and then we can merge.

Done! 

> Thanks, Nicolas!

You are welcome, thank you for the guides around monorepo and tooling, they are quite useful to share as a guide!

anthonyshew (approved):

PR #10100, opened 2025-03-05
fix: adds callout for installation instructions for PNPM

Tigatok:
Refs: [10099](https://github.com/vercel/turborepo/issues/10099)

### Description

Adds callout for installation instructions in documentation page.

### DIFF
--- a/docs/repo-docs/getting-started/add-to-existing-repository.mdx
+++ b/docs/repo-docs/getting-started/add-to-existing-repository.mdx
@@ -40,10 +40,6 @@ familiar with Turborepo.
 <Step>
 ### Install `turbo`
 
-<Callout type="good-to-know">
-  Ensure you have created a `pnpm-workspace.yaml` file before you begin the installation. Failure to have this file will result in an error that says: ` --workspace-root may only be used inside a workspace`.
-</Callout>
-
 We recommend you install `turbo` both globally and into your repository's root for the best developer experience.
 
 <Tabs items={['npm', 'yarn', 'pnpm']} storageKey="selected-pkg-manager">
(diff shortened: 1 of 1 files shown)


docs/repo-docs/getting-started/add-to-existing-repository.mdx
```diff
@@ -40,6 +40,10 @@ familiar with Turborepo.
 <Step>
 ### Install `turbo`
 
+<Callout type="good-to-know">
```
anthonyshew, line 43: Oh, good addition. Do you think you could move this into the tab for pnpm? Users of other package managers wouldn't need to see this information, so it would be great to localize. You won't need to have it be in a <Callout /> in there, as well.

anthonyshew (changes requested):
Oh, good addition. Do you think you could move this into the tab for pnpm? Users of other package managers wouldn't need to see this information, so it would be great to localize. You won't need to have it be in a in there, as well.

Tigatok:
> Oh, good addition. Do you think you could move this into the tab for pnpm? Users of other package managers wouldn't need to see this information, so it would be great to localize. You won't need to have it be in a in there, as well.

Yep, you bet. I wasn't exactly sure myself so just wanted to get something up so I could get feedback like this :). Is there a proper way I can run the docs locally? I pushed the change but I am not sure if there is a quick way of me validating locally so I can avoid murdering the pipeline.

docs/repo-docs/getting-started/add-to-existing-repository.mdx
```diff
@@ -64,6 +64,9 @@ We recommend you install `turbo` both globally and into your repository's root f
 
   </Tab>
   <Tab value="pnpm">
+    <Callout type="good-to-know">
+      Ensure you have created a `pnpm-workspace.yaml` file before you begin the installation. Failure to have this file will result in an error that says: ` --workspace-root may only be used inside a workspace`.
+    </Callout>
```
anthonyshew, line 69:
```suggestion
      Ensure you have created a `pnpm-workspace.yaml` file before you begin the installation. Failure to have this file will result in an error that says: ` --workspace-root may only be used inside a workspace`.
```

anthonyshew (approved):

anthonyshew:
There isn't at this time, unfortunately. I have to check by hand. But here it is below, looking good!
CleanShot 2025-03-05 at 13 42 40

PR #10114, opened 2025-03-06
chore: update to Rust 2024 edition

ognevny:
### Description

update to 2024 edition:

- force 2021 style formatting to avoid huge formatting changes (>1000) lines (this still doesn't disable all 2024 edition formatting things)
- exclude turborepo-lib from edition bump as tracing macros are not ready for `use<>` syntax required in several places; exclude turborepo-filewatch because I don't have macos machine to properly address failures
- edition 2024 lints:
  - impl Trait lifetimes: https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html
  - various ref/ref mut cleanups: https://doc.rust-lang.org/edition-guide/rust-2024/match-ergonomics.html
  - `gen` is now a keyword, use `r#gen` to fix compile error: https://doc.rust-lang.org/edition-guide/rust-2024/gen-keyword.html
  - env::set_var is unsafe: https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html
  - extern blocks are unsafe: https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-extern.html

### Testing Instructions

CI


crates/turborepo-repository/src/package_manager/mod.rs
```diff
@@ -435,7 +435,7 @@ impl PackageManager {
     pub fn get_package_jsons(
         &self,
         repo_root: &AbsoluteSystemPath,
-    ) -> Result<impl Iterator<Item = AbsoluteSystemPathBuf>, Error> {
+    ) -> Result<impl Iterator<Item = AbsoluteSystemPathBuf> + use<>, Error> {
```
NicholasLYang, line 438: If the `use` bound is empty, do we need it?

NicholasLYang (approved):

crates/turborepo-repository/src/package_manager/mod.rs
ognevny: yes, otherwise it fails to compile

PR #10128, opened 2025-03-10
fix: mark type of flat config export to satisfy Linter.Config[]

cprussin:
### Description

Adds `satisfies Linter.Config[]` after the flat config definition.  This allows importing the config into other typescirpt configs without having type unification issues.

### Testing Instructions

N/A this is a pretty simple type fix


anthonyshew (approved):
Thanks!

PR #10143, opened 2025-03-11
fix(docs): fix commas in package exports example

x-N0:
### Description

Fixed commas order.

### Testing Instructions

Use a JSON schema tester.


anthonyshew (approved):
Thanks!

PR #10145, opened 2025-03-11
docs: update storybook initiation instructions

sicarius97:
### Description

Follow latest Storybook instructions and conventions for setup. Using yarn or pnpm dlx otherwise will only install versions of storybook older than 8.3. Very simple change that uses the `create` command instead

### Testing Instructions

For testing, you could try the two different initiation options:
`pnpx dlx storybook@latest init`

Then verify your installed version, which will show a version older than 8.3. Delete the directory and create a new storybook directory, then run:

`pnpm create storybook@latest` 

Afterwards, the version should show the latest version, either 8.3 or older.

Also just a note, I think it is on storybooks side, but the most recent version of nextjs that they use in their examples is 8.


anthonyshew (approved):
Great, thanks!

PR #10147, opened 2025-03-11
feat(boundaries): auto ignore

NicholasLYang:
### Description

Allow users to automatically add `@boundaries-ignore` comments to their errors, either all at once or interactively using the `--ignore` flag.

### Testing Instructions

Added a test for the all at once case, but you should test the interactive case manually.


chris-olszewski (approved):

crates/turborepo-lib/src/commands/boundaries.rs
```diff
@@ -14,7 +25,66 @@ pub async fn run(base: CommandBase, telemetry: CommandEventBuilder) -> Result<i3
...
+                    }
+                }
+            };
+
+            if let Some(reason) = reason {
+                patches.entry(path).or_default().push((span, reason));
+            }
+        }
+
+        for (path, file_patches) in patches {
```
chris-olszewski, line 73: Unstable iteration order is causing issues for snapshots

PR #10151, opened 2025-03-12
feat(boundaries): package name as tag punning

NicholasLYang:
### Description

Allows a package name to be used as a tag ([tag punning](https://en.wikipedia.org/wiki/Type_punning)). We're adding this feature to help users define an explicit set of packages that their package can be imported by. Combined with the ability to define package rules in a package `turbo.json` (next PR after this, coming soon), and code owner restrictions, this will permit package owners to control who can import their package.

### Testing Instructions

Added test in `boundaries_tags` fixture


anthonyshew (approved):

crates/turborepo-errors/src/lib.rs
```diff
@@ -113,6 +113,12 @@ impl<T: Deserializable> Deserializable for Spanned<T> {
     }
 }
 
+impl Spanned<String> {
+    pub fn as_str(&self) -> &str {
+        self.value.as_str()
+    }
+}
+
```
chris-olszewski, line 121: I don't think this is needed since `Spanned` already implements deref?
```suggestion
```

chris-olszewski (approved):
Awesome feature!

crates/turborepo-errors/src/lib.rs
chris-olszewski: I think you need to specify the deref type sometimes to help out. I know I've needed to do that when working with `Cow` sometimes

PR #10156, opened 2025-03-13
ci: build JS packages

anthonyshew:
### Description

Making sure we find out if we broke building JS packages sooner.

### Testing Instructions

CI

### DIFF
--- a/packages/eslint-config-turbo/src/flat/index.ts
+++ b/packages/eslint-config-turbo/src/flat/index.ts
@@ -5,7 +5,7 @@ import plugin from "eslint-plugin-turbo";
 export default [
   {
     plugins: {
-      turbo: plugin,
+      turbo: [plugin],
     },
     rules: {
       "turbo/no-undeclared-env-vars": "error",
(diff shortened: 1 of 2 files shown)


packages/eslint-config-turbo/src/flat/index.ts
```diff
@@ -5,7 +5,7 @@ import plugin from "eslint-plugin-turbo";
 export default [
   {
     plugins: {
-      turbo: plugin,
+      turbo: [plugin],
```
chris-olszewski, line 8: Unneeded with https://github.com/vercel/turborepo/pull/10157
```suggestion
      turbo: plugin,
```

chris-olszewski (approved):
For when https://github.com/vercel/turborepo/pull/10157 lands and you can rebase and see green

packages/eslint-config-turbo/src/flat/index.ts
anthonyshew: Ope, yup, forgot to drop that change while I was changing branches. Good catch.

PR #10160, opened 2025-03-13
feat(boundaries): package rules

NicholasLYang:
### Description

Adds the ability for boundaries rules to be defined in the package `turbo.json`. These rules are specific to that package and are designed to work with code owners so that a package can restrict who imports it.


chris-olszewski (approved):
LGTM

PR #10168, opened 2025-03-14
chore: dogfood $TURBO_ROOT$

chris-olszewski:
### Description

Dogfood the new `$TURBO_ROOT$` feature as there were a few places we were depending on `../`

### Testing Instructions

CI


chris-olszewski:
Failures seem unrelated https://github.com/vercel/turborepo/pull/10169

Will not force through, but just confirming this isn't some bug in `$TURBO_ROOT$`

PR #10175, opened 2025-03-15
feat(bun): bun prune support

camero2734:
### Description
Resolves https://github.com/vercel/turborepo/issues/9058 and https://github.com/vercel/turborepo/discussions/7456

Building on https://github.com/vercel/turborepo/pull/9783 to add `prune` support for bun. This is a simple implementation that allow turbo to properly prune a `bun.lock` file

Basically, it just creates a new `BunLockfileData` with the filtered dependencies/workspaces and serializes it to `bun.lock` via `serde_json`. This is made simple since the internal layout matches the bun.lock format, so the only special serialization case I had to do was `PackageEntry`. This also required updating the _deserialization_, since it skipped parsing some parts that we need to write back out.
 
### Testing Instructions
1. Clone https://github.com/mugencraft/turbobun (or any other turbo + bun repo)
2. `bun install --save-text-lockfile` (we need the new `.lock` version and not `.lockb`)
3. I then do 
    ```sh
    rm -rf ./out && cp ../turborepo/target/debug/turbo ./node_modules/.bin/turbo # Copy debug build of turbo
    bunx turbo prune @turbobun/backend --docker --skip-infer # turbo prune 🚀
    ```
4. Open the generated `out/bun.lock`. It's currently outputted in an unformatted json, so I format it.
5. Observe that the dependencies include backend (`elysia`) and root workspace (`commitizen`) dependencies, but not e.g. `@turbobun/website` dependencies (like `next`)
6. `cd ./out/json && bun install` -> Bun install can read the lockfile and works!
7. You can also patch a random package using `bun patch` and test that the patches are propagated correctly

### Screenshots
image

### DIFF
--- a/crates/turborepo-lockfiles/src/bun/de.rs
+++ b/crates/turborepo-lockfiles/src/bun/de.rs
@@ -3,6 +3,7 @@ use std::collections::VecDeque;
 use serde::Deserialize;
 
 use super::{PackageEntry, PackageInfo};
+use crate::bun::RootInfo;
 // Comment explaining entry schemas taken from bun.lock.zig
 // first index is resolution for each type of package
 // npm         -> [
@@ -25,7 +26,7 @@ impl<'de> Deserialize<'de> for PackageEntry {
         D: serde::Deserializer<'de>,
     {
         use serde::de;
-        #[derive(Deserialize)]
+        #[derive(Deserialize, Debug)]
         #[serde(untagged)]
         enum Vals {
             Str(String),
@@ -46,20 +47,50 @@ impl<'de> Deserialize<'de> for PackageEntry {
         };
         // For workspace packages deps are second element, rest have them as third
         // element
-        let info = vals
-            .pop_front()
-            .and_then(val_to_info)
-            .or_else(|| vals.pop_front().and_then(val_to_info));
+
+        let mut registry = None;
+        let mut info = None;
+
+        if key.ends_with("@root:") {
+            let root = vals.pop_front().and_then(|val| {
+                serde_json::from_value::<RootInfo>(match val {
+                    Vals::Info(info) => {
+                        serde_json::to_value(info.other).expect("failed to convert info to value")
+                    }
+                    _ => return None,
+                })
+                .ok()
+            });
+            return Ok(Self {
+                ident: key,
+                info,
+                registry,
+                checksum: None,
+                root,
+            });
+        }
+
+        // The second value can be either registry (string) or info (object)
+        vals.pop_front().map(|val| match val {
+            Vals::Str(reg) => registry = Some(reg),
+            Vals::Info(package_info) => info = Some(*package_info),
+        });
+
+        // Info will be next if we haven't already found it
+        if info.is_none() {
+            info = vals.pop_front().and_then(val_to_info);
+        }
+
+        // Checksum is last
         let checksum = vals.pop_front().and_then(|val| match val {
             Vals::Str(sha) => Some(sha),
             Vals::Info(_) => None,
         });
+
         Ok(Self {
             ident: key,
             info,
-            // The rest are only necessary for serializing a lockfile and aren't needed until adding
-            // `prune` support
-            registry: None,
+            registry,
             checksum,
             root: None,
         })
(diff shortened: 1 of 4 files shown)


chris-olszewski (changes requested):
The pruned `bun.lock` needs to be able to pass a `bun install --frozen-lockfile` and I haven't gotten it to work on the few example repositories I've tested.

I don't immediately see what changes need to be made to the pruned lockfile, but I will take some time this afternoon to try to get `bun install --frozen-lockfile` to pass.

camero2734:
@chris-olszewski Yeah, just tried it with turbobun with `--frozen-lockfile` and it seems to be all good:
<image width="400" src="https://github.com/user-attachments/assets/bd2bcc94-729a-4a1d-a526-4b515bd37917" />

But on the `with-yarn` example (after running bun install, deleting yarn.lock, and updating the packageManager), I get:
<image width="400" src="https://github.com/user-attachments/assets/9ba9314d-89fb-45b4-8606-b3aec2e2e604" />

It seems like it's because there's two names in the OG bun.lock that resolve to the same package:
<image width="400" src="https://github.com/user-attachments/assets/abb2b3a7-efa1-44ea-98e3-1550fc9ce544" />

But in the pruned bun.lock there's only the `-cjs` version so it can't find the other alias for it
<image width="400" src="https://github.com/user-attachments/assets/1dc5d411-c104-4f3c-b489-fe6fcec10623" />

Which I believe is related to this comment you left in the code:
> Note: Entries are not deduplicated in bun.lock if they need to be qualified [...] where both will have the same entry with ident of shared@1.0.0. Because they are identical entries, we do not differentiate between them even though they are different entries in the map.

So perhaps when writing it out, we need to make sure to actually write out the relevant "duplicated" entries. I'll work on this 👍

chris-olszewski (approved):
Awesome! Thank you so much for tackling this.

crates/turborepo-lockfiles/src/bun/mod.rs
```diff
@@ -209,6 +218,69 @@ impl BunLockfile {
...
+        workspace_packages: &[String],
+        packages: &[String],
+    ) -> Result<BunLockfile, Error> {
+        let new_workspaces: Map<_, _> = self
+            .data
+            .workspaces
+            .iter()
+            .filter_map(|(key, entry)| {
+                // Ensure the root workspace package is included, which is always indexed by ""
+                if key == "" || workspace_packages.contains(&key) {
```
chris-olszewski, line 237:
```suggestion
                if key.is_empty() || workspace_packages.contains(key) {
```

crates/turborepo-lockfiles/src/bun/de.rs
```diff
@@ -44,22 +47,51 @@ impl<'de> Deserialize<'de> for PackageEntry {
...
+                checksum: None,
+                root,
+            });
+        }
+
+        // The second value can be either registry (string) or info (object)
+        vals.pop_front().map(|val| match val {
+            Vals::Str(reg) => registry = Some(reg),
+            Vals::Info(package_info) => info = Some(*package_info),
+        });
```
chris-olszewski, line 78:
```suggestion
        let Some(val) = vals.pop_front() {
            match val {
                Vals::Str(reg) => registry = Some(reg),
                Vals::Info(package_info) => info = Some(*package_info),
            }
        });
```

crates/turborepo-lockfiles/src/bun/de.rs
```diff
@@ -44,22 +47,53 @@ impl<'de> Deserialize<'de> for PackageEntry {
...
+                ident: key,
+                info,
+                registry,
+                checksum: None,
+                root,
+            });
+        }
+
+        // The second value can be either registry (string) or info (object)
+        let Some(val) = vals.pop_front() {
```
chris-olszewski, line 75:
```suggestion
        if let Some(val) = vals.pop_front() {
```

crates/turborepo-lockfiles/src/bun/de.rs
```diff
@@ -44,22 +47,53 @@ impl<'de> Deserialize<'de> for PackageEntry {
...
+            });
+        }
+
+        // The second value can be either registry (string) or info (object)
+        if let Some(val) = vals.pop_front() {
+            match val {
+                Vals::Str(reg) => registry = Some(reg),
+                Vals::Info(package_info) => info = Some(*package_info),
+            }
+        });
```
chris-olszewski, line 80:
```suggestion
        };
```

Issue #9860, opened 2025-01-31
Allow env variables that are used to detect deployment environment for Nuxt


MickL:
### Verify canary release

- [x] I verified that the issue exists in the latest Turborepo canary release.

### Which canary version will you have in your reproduction?

2.3.4

### Description

At Nitro / Nuxt environment variables are used to automatically determine what deployment environment the build is in, and then using a specific build preset for that environment. This results in a zero configuration build: Deploying Nuxt or Nitro projects to whatever provider (Vercel, Cloudflare, Deno Deploy, etc. etc.) just works automatically.

Unfortunately this doesnt work anymore when Turborepo comes into play because Turborepo removes all env variables that are not set in the `globalEnv` array (if not using `--env-mode=loose`).

May it be possible that you dont remove all environment specific variables that are used to determine the build environment by default? You can find a full list of the variables Nitro / Nuxt is using here: https://github.com/unjs/std-env/blob/main/src/providers.ts#L59

Issue at Nitro: https://github.com/nitrojs/nitro/issues/3051

anthonyshew:
Happy to take any PRs to widen these variables out!

ogbird:
I'd like to take a crack at this, @anthonyshew. I'm still learning Rust so I might have some annoying, Fisher-Price tier questions. Are you available for a bit of pestering if it comes to it? Is here OK, or is there a better way to ping you without clogging up this issue?

anthonyshew:
Sounds great! I believe it will look a lot like this PR: https://github.com/vercel/turborepo/pull/8725

ogbird:
So based on your PR link, @anthonyshew and your comments @MickL, this seems like we're just updating this [frameworks.json](https://github.com/vercel/turborepo/blob/main/packages/turbo-types/src/json/frameworks.json#L54) file to include the Cloudflare Pages dependency? Are you working with Nitro or Nuxt or both @MickL?

anthonyshew:
Sounds correct from my end!

MickL:
Adding CF variable will only fix Cloudflare. Personally I would add all the variables used [here](https://github.com/unjs/std-env/blob/main/src/providers.ts#L59) to support detection of all environments (Nitro/Nuxt).

ogbird:
Ah, thanks @MickL, that link to those `providers` clicked for me now.

So as a quick confirmation based on that list of providers, you would expect:

CLOUDFLARE_PAGES -> Add `CF_PAGES`
[...]

PR #10176, opened 2025-03-15
fix: update env variables supported by nitro

pi0:
#9860

Deployment providers provide environmental variables to be detected.

[Nitro](https://nitro.build) uses this mechanism to auto-detect deployment providers when building.

I have made a handcrafted list of some common items to nuxt and nitro frameworks + `SERVER_*` (nitro supported env).

- aws_amplify: AWS_APP_ID
- azure_static: INPUT_AZURE_STATIC_WEB_APPS_API_TOKEN
- cleavr: CLEAVR
- cloudflare_pages: CF_PAGES
- firebase_app_hosting: FIREBASE_APP_HOSTING
- netlify: NETLIFY
- stormkit: STORMKIT
- vercel: NOW_BUILDER 
- zeabur: ZEABUR
- cleavr: CLEAVR
- render.com: RENDER

> [!IMPORTANT]
> This list will be most likely updated over time, meaning turborepo support will be broken from time to time and need to manually be updated. Nitro uses [std-env](https://github.com/unjs/std-env/blob/main/src/providers.ts) as source of this list.

> [!NOTE]
> Having list of provider envs under each framework, means that turborepo does not supports platform envs for frameworks not explicitly listed in this json (there are already few nitro based ones missing)

### DIFF
--- a/packages/turbo-types/src/json/frameworks.json
+++ b/packages/turbo-types/src/json/frameworks.json
@@ -53,7 +53,7 @@
   {
     "slug": "nitro",
     "name": "Nitro",
-    "envWildcards": ["NITRO_*"],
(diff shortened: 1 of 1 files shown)


chris-olszewski (approved):
Thank you for the PR! 

This gets us to a better spot since `--env-mode=loose` use could lead to a cache hit from a build that inferred a different provider preset.

PR #10190, opened 2025-03-17
fix(example): remove obsolete @types/react-native from with-react-native-web

shinjith-dev:
Removed obsolete `@types/react-native` as it has peer deps that might break expo/metro version compatibility and no longer needed.

### Description

Resolved issues ([mentioned here](https://github.com/vercel/turborepo/pull/9872#issuecomment-2727301096)) due to the use of obsolete types package. 

Screenshot_2025-03-17_23-55-52


anthonyshew (approved):
Thank you!

PR #10197, opened 2025-03-19
docs: use schema from current @turbo/types

chris-olszewski:
### Description

Noticed `turbo.build/schema.json` was displaying an out of date version of the schema. We had it hard coded to an old version. Fixed by updating to just use the in repo version of the package. ❤️ `turbo.build` being back in our repository.

### Testing Instructions

Visit preview's `schema.json` and verify that `teamId` and `teamSlug` are present.


anthonyshew (approved):
Yay!

PR #10211, opened 2025-03-21
fix(turbo_json): avoid workspace validation errors

chris-olszewski:
### Description

#10083 introduced a bug where we swallow validation errors and end up silently ignoring these workspace `turbo.json` files. This PR fixes that by distinguishing a missing `turbo.json` from a failed read.

The bug stems from the fact that a missing file and a failed read were both errors so `turbo_json.or(turbo_jsonc)` could choose a missing turbo.json error (which is fine, not all workspaces need `turbo.json`s) over a validation error for the file that is present. We replace `.or` with an extremely explicit `match` that goes through all the possible combinations.

### Testing Instructions
Added unit test

Quick manual verification
```
[0 olszewski@macbookpro] /Users/olszewski/code/vercel/turborepo $ turbo_dev build   
 WARNING  No locally installed `turbo` found. Using version: 2.4.5-canary.6.
turbo 2.4.5-canary.6

turbo_json_parse_error

  × Failed to parse turbo.json.
  ╰─▶   × Found an unknown key `lol`.
          ╭─[cli/turbo.json:9:7]
        8 │     "rust-src": {
        9 │       "lol": true,
          ·       ─────
       10 │       "env": [
          ╰────
```

### DIFF
--- a/crates/turborepo-lib/src/config/turbo_json.rs
+++ b/crates/turborepo-lib/src/config/turbo_json.rs
@@ -53,15 +53,8 @@ impl<'a> ResolvedConfigurationOptions for TurboJsonReader<'a> {
         existing_config: &ConfigurationOptions,
     ) -> Result<ConfigurationOptions, Error> {
         let turbo_json_path = existing_config.root_turbo_json_path(self.repo_root)?;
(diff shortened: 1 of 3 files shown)


anthonyshew (approved):
Good find. Thanks, Chris!

PR #10218, opened 2025-03-21
ci: use git commands instead of a package

anthonyshew:
### Description

The package we were using before wasn't serving us well, so choosing to drop down an abstraction level here.

This PR uses bash and git to find path changes.

### Testing Instructions

Proofs:
- No changes (except the workflow file, as expected): [e026cee](https://github.com/vercel/turborepo/actions/runs/14000663145/job/39206126188)
- A change in docs: [0fcdbab](https://github.com/vercel/turborepo/actions/runs/14000689402/job/39206233665?pr=10218)
- A change in `basic` example: [0069c34](https://github.com/vercel/turborepo/actions/runs/14000711838/job/39206300460?pr=10218)
- A change in `rest`: [9622226](https://github.com/vercel/turborepo/actions/runs/14000776584/job/39206508017?pr=10218) (A little difficult to see since "No changes" is the same, but I edited a README in crates)
- A change in both `rest` and an example: [985ce6a](https://github.com/vercel/turborepo/actions/runs/14000743052/job/39206401816?pr=10218)
- A change in both `rest` and docs: [5291f41](https://github.com/vercel/turborepo/actions/runs/14000808637/job/39206606561?pr=10218)

### DIFF
--- a/.github/workflows/turborepo-test.yml
+++ b/.github/workflows/turborepo-test.yml
@@ -91,7 +91,6 @@ jobs:
           check_paths "with-tailwind-example" "examples/with-tailwind/,turborepo-tests/example-with-svelte-*/,turborepo-tests/helpers/"
 
           # Handle the "rest" pattern - files that are NOT in examples/ or docs/
-          echo "Checking for changes in rest paths..."
           CHANGED_FILES=$(git diff --name-only $BASE_COMMIT $HEAD_COMMIT)
 
           # Filter to only include files that do NOT start with examples/ or docs/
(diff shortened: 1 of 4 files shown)


.github/workflows/turborepo-test.yml
```diff
@@ -29,34 +29,80 @@ jobs:
...
+            IFS=',' read -ra path_array <<< "$path_string"
+
+            # Call the check_path_changes function with the name and paths
+            check_path_changes "$name" "${path_array[@]}"
+          }
+
+          # Check each path pattern with a more readable syntax
+          echo "Checking path patterns..."
+
+          check_paths "docs" "docs/"
```
chris-olszewski, line 86: Does this need to be hooked up to something? I'm not seeing any triggers on the example runs

anthonyshew: Docs checks happen in https://github.com/vercel/turborepo/blob/main/.github/workflows/docs.yml, which is confusing. It might have been that I intended to run those in this file, so they could become blocking when they should, but didn't wire it up. 🤔

anthonyshew: https://github.com/vercel/turborepo/actions/runs/14000711838/job/39206316471?pr=10218 is the smoking gun for me that these changes are at least on the right track.

anthonyshew:
Merging now as I'm about to do some stuff for examples. Will be keeping an eye out and have notified the team to make sure everything they expect is running.

PR #10219, opened 2025-03-22
fix(bun): deserialize correctly and use optionalPeers

camero2734:
### Description
Resolves https://github.com/vercel/turborepo/issues/9058 and https://github.com/vercel/turborepo/discussions/7456 (again)

When I tested my previous PR (#10175) on another repo, I discovered an issue with a missing dep in the pruned lockfile

To reproduce this in the `with-yarn` example repo:
`cd apps/web && bun install dd-trace`
and
`cd apps/docs && bun install @opentelemtry/api`

Then run prune for `web` and try `bun install --frozen-lockfile` and you'll get this:
```
error: Failed to resolve peer dependency '@opentelemetry/api' for package '@opentelemetry/core'
    at bun.lock:1:16974
InvalidPackageInfo: failed to parse lockfile: 'bun.lock'
```

### Issue 1 -- Deserialization casing
The underlying error is rather straightforward -- it was deserializing most dependencies (`peerDependencies`, `devDependencies`, `optionalDependencies`) to `other` 😄 
image

This is because it just needed `#[serde(rename_all = "camelCase")]` since the rust names are snake case. This is why `dependencies` happened to work -- it's the same in snake and camel case.

By virtue of being in `other`, all non-`dependencies` packages were effectively ignored during the pruning process, and don't always end up in the resulting lockfile.

### Issue 2 -- optionalPeers
After fixing that, I also stumbled upon another issue:
```
WARNING  Unable to calculate transitive closures: No lockfile entry found for 'next/@playwright/test'
```

This is because we weren't accounting for the `optionalPeers` property. `next` doesn't actually require you have `@playwright/test`, it's an optional peer. I fixed this as well.

### DIFF
--- a/crates/turborepo-lockfiles/src/bun/de.rs
+++ b/crates/turborepo-lockfiles/src/bun/de.rs
@@ -153,6 +153,16 @@ mod test {
                 dependencies: Some(("is-number".into(), "^6.0.0".into()))
                     .into_iter()
                     .collect(),
+                dev_dependencies: Some(("is-bigint".into(), "1.1.0".into()))
+                    .into_iter()
+                    .collect(),
+                peer_dependencies: Some(("is-even".into(), "1.0.0".into()))
+                    .into_iter()
+                    .collect(),
+                optional_peers: Some("is-even".into()).into_iter().collect(),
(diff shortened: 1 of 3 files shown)


chris-olszewski (approved):
This is a great find! Thank you for the fix.

PR #10236, opened 2025-03-26
feat: allow specifying concurrency in config / environment

kade-robertson:
### Description

This adds a couple new ways to configure task run concurrency: 

- A `TURBO_CONCURRENCY` environment variable
- A `concurrency` configuration value in `turbo.json`

These options should follow the same precedence rules for other options which are made available in multiple locations. I think the environment variable is the most valuable to me (IMO), as this would make it a lot easier to turn off or drastically reduce concurrency in CI scenarios with restricted performance, where running multiple test suites at the same time could cause flaky results due to timeouts, or too much resource contention in general scenarios. Being able to specify this in the configuration file is maybe less valuable, but there are situations where it might be desirable.

The `--concurrency` option is obviously already available and certainly usable in combination with other changes (enforcing the use of an alias, or setting up different `package.json` scripts for CI and local development), but enabling this to be enforced via the config or an environment variable makes it a lot easier to turn off concurrency in CI, for example.

There is some active desire for this feature to exist:
- https://github.com/vercel/turborepo/discussions/7263
- https://github.com/vercel/turborepo/discussions/7493

Notably, this does _not_ enable configuring concurrency per task, which was part of one of the discussions here. That seemed like a much larger lift vs. just adding an environment variable and config option that behaved exactly like the CLI argument already does. There is still some demand for this "simple" option.

### Testing Instructions

1. Build `turbo`
2. Check that you can now specify a top-level "concurrency" property in your `turbo.json`, with the same behaviour as the CLI argument.
3. Check that you can now specify a `TURBO_CONCURRENCY` environment variable, with the same behaviour as the CLI argument.
4. Check that this settings maintains expected preference (i.e. CLI arg > env var > turbo.json)

I've done this checks (with some print debugging) on repos where I use `turbo` and it seems to work as expected. I've added some unit and integration tests where it seemed to make the most sense, but there were some spots where it was unclear if I should add something. If I've missed a spot or need some additional checks, let me know!

### DIFF
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -312,7 +312,7 @@ impl<'a> TryFrom<OptsInputs<'a>> for RunOpts {
 
     fn try_from(inputs: OptsInputs) -> Result<Self, Self::Error> {
         let concurrency = inputs
-            .execution_args
+            .config
             .concurrency
             .as_deref()
             .map(parse_concurrency)
--- a/crates/turborepo-lib/src/commands/config.rs
+++ b/crates/turborepo-lib/src/commands/config.rs
@@ -24,6 +24,7 @@ struct ConfigOutput<'a> {
     scm_base: Option<&'a str>,
     scm_head: Option<&'a str>,
     cache_dir: &'a Utf8Path,
+    concurrency: Option<&'a str>,
 }
 
 pub async fn run(repo_root: AbsoluteSystemPathBuf, args: Args) -> Result<(), cli::Error> {
@@ -55,6 +56,7 @@ pub async fn run(repo_root: AbsoluteSystemPathBuf, args: Args) -> Result<(), cli
             scm_base: config.scm_base(),
             scm_head: config.scm_head(),
             cache_dir: config.cache_dir(),
+            concurrency: config.concurrency.as_deref()
         })?
     );
     Ok(())
--- a/crates/turborepo-lib/src/commands/mod.rs
+++ b/crates/turborepo-lib/src/commands/mod.rs
@@ -122,6 +122,10 @@ impl CommandBase {
             )
             .with_run_summary(args.run_args().and_then(|args| args.summarize()))
             .with_allow_no_turbo_json(args.allow_no_turbo_json.then_some(true))
+            .with_concurrency(
+                args.execution_args()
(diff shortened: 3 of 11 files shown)


docs/site/content/repo-docs/reference/configuration.mdx
kade-robertson, line 1: This is just copied from the description of the CLI argument, but might need to be tweaked to make more sense in this context? It might be weird to be referencing a CLI argument (`--parallel`) nullifying this.

chris-olszewski: I think for now it is fine to keep as is.

crates/turborepo-lib/src/opts.rs
```diff
@@ -312,6 +312,7 @@ impl<'a> TryFrom<OptsInputs<'a>> for RunOpts {
             .execution_args
             .concurrency
             .as_deref()
+            .or(inputs.config.concurrency.as_deref())
```
chris-olszewski, line 315: You can remove the use of `inputs.execution_args` as a source of the concurrency value and just rely on `inputs.config.concurrency` alone. The `with_concurrency` call that you added in `commands/mod.rs` already takes into account the `execution_args` source so we don't need to consider it here. There shouldn't be any behavior since the config builder considers arguments as highest priority, but still nice to have that logic only live there.

```
let concurrency = inputs.config.concurrency.as_deref()
            .map(parse_concurrency)
            ...
```
(Github won't let me suggest changes to code that wasn't changed)

chris-olszewski (changes requested):
Thank you so much for the contribution. Everything looks good aside from the removing of the `inputs.execution_args` reference in `opts`.

crates/turborepo-lib/src/opts.rs
kade-robertson: Ah yep, makes sense! Fixed in [000b81f](https://github.com/vercel/turborepo/pull/10236/commits/000b81f67876cba0d92b5caee1ea44b216a075e3)

chris-olszewski (approved):

chris-olszewski:
If you could rebase, I'll work on getting this merged and released in the next canary. Thank you for the contribution!

kade-robertson:
> If you could rebase, I'll work on getting this merged and released in the next canary. Thank you for the contribution!

Should be good to go!

PR #10247, opened 2025-03-29
feat(tui): configurable scrollback length

anthonyshew:
### Description

Allow users to set a different value for scrollback length. I'm also taking a liberty to double the scrollback length's default value here (cc @chris-olszewski to tell me if this isn't a good idea)

#### Original asks:
- https://github.com/vercel/turborepo/discussions/10234
- https://github.com/vercel/turborepo/issues/8643#issuecomment-2749429985

> [!NOTE]
> If you're wondering why we don't dynamically set this to whatever the terminal's scrollback is, terminals don't expose this value. The closest thing to a spec for terminals is [the xterm spec](https://invisible-island.net/xterm/ctlseqs/ctlseqs.txt), so we don't know what we can depend on.

### Questions

Should this also be:
- A `turbo.json` config?
  - Personally, I don't think so. Our reasoning for configurations in `turbo.json` tends to be that it should be something that makes sense for *anyone* in the repo, and setting some non-default scrollback length doesn't sound desirable.
- A flag?
  - This sounds reasonable, but omitting here unless an ask comes in for it.

### Testing Instructions

Try it!


crates/turborepo-ui/src/tui/term_output.rs
```diff
@@ -7,7 +7,15 @@ use super::{
     Error,
 };
 
-const SCROLLBACK_LEN: usize = 1024;
+const SCROLLBACK_LEN: usize = 2048;
```
anthonyshew, line 10: Making sure to call extra attention here: Users mentioned they felt 1024 was too small, so I doubled it here. Not sure about if there's a reason I shouldn't do this or if a different value should be used. Just did this for the sake of discussion.

chris-olszewski: I think this is okay since https://github.com/vercel/turborepo/pull/9123 has been merged.

`vt100` used to need to iterate through every scrollback row in order to render a visible row so 1024->2048 would be a noticeable perf hit. I will do a quick profile to check if there's a noticeable impact from this change.

chris-olszewski:
@lbennett-stacki I don't think we want to use `HISTSIZE` as a source for this value. `HISTSIZE` is meant to target the shell history and not terminal scrollback. It would be confusing if we used it as the scrollback buffer length

> The value of the `HISTSIZE` shell variable is used as the number of commands to save in a history list. The text of the last `$HISTSIZE` commands (default 500) is saved.

https://www.gnu.org/software/bash/manual/html_node/Bash-History-Facilities.html

anthonyshew:
> (Gotta keep the dream of code generated docs alive)

It's not a dream, it's real. /s

> If we're gonna add a config option

I wasn't going to do so in this PR. I was just going to leave it at the env var. Are you saying we _should_ do one?

chris-olszewski:
> Are you saying we should do one?

Sorry, I meant "config option" in the most general sense of a way to configure how `turbo` behaves. Just meant that we should try to keep our list of env vars here: https://github.com/vercel/turborepo/blob/main/crates/turborepo-lib/src/config/env.rs

anthonyshew:
Using #10261 to break things up into PRs that only do one thing.

anthonyshew:
@chris-olszewski, I refactored. Can you take another look?

chris-olszewski (approved):
LFG

rchl-u:
So how to configure this? It's not clear from the description or the code for those that are not familiar with it.

Issue #8643, opened 2024-07-01
New TUI: Scrolling behavior can be improved


anthonyshew:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

N/A

### What package manager are you using / does the bug impact?

npm

### What operating system are you using?

Other

### Which canary version will you have in your reproduction?

2.0.6-canary.0

### Describe the Bug

Lifting out questions about scrolling from https://github.com/vercel/turbo/issues/8372 so we can have separate GitHub Issues for separate issues.

### Expected Behavior

See comments both on the original issue and below for expected behavior from the community.

### To Reproduce

Scrolling in the new TUI.

IshfaqAhmedProg:
Any update on this?

aronstrandberg:
I'm having problems with this too. Our application logs a lot of debug data when executing tasks, and the new TUI seems to limit scrollback, so I can't scroll up to see the startup parameters logged at the beginning of the task execution. I do not have this problem if disabling the new TUI. I find it very strange that Turbo would override the scrollback I've already configured in my terminal.

anthonyshew:
@aronstrandberg Doesn't solve your problem but, for context, the TUI renders in a vterm so doesn't automatically inherit your configurations. We [hardcoded the buffer length](https://github.com/vercel/turborepo/blob/80a9d6feff31f0cb0bb2c9dc20c92472f42adc06/crates/turborepo-ui/src/tui/term_output.rs#L10) since we couldn't find any way to do it in [the xterm spec](https://invisible-island.net/xterm/ctlseqs/ctlseqs.txt).

This class of problem isn't unique to Turborepo's TUI, for what it's worth. Anything rendering to an alt screen or manipulating your terminal can/will suffer from the same problems (`nvim`, `tmux`, etc.), and is the reason that we've struggled with figuring out a solution we like/users will like for the other scrolling behavior questions. Terminals are ooooold technology, and we're stretching them past their limits. 😄

darr1s:
But I'm able to use pgUp pgDown fine in those mentioned vt, how do you scroll to bottom w turbo tui?

anthonyshew:
> how do you scroll to bottom w turbo tui?

@darr1s Are you meaning you're looking for a hotkey to do this with? We do have u and d now to scroll one line up and down, but we could also do a keybind for paging or jumping to top/bottom.

spinda:
[...]

PR #10248, opened 2025-03-29
feat(tui): more keybinds for scrolling

anthonyshew:
### Description

See #8643 

Page through terminal logs with `Shift + u` and `Shift + d`.

Jump to top or bottom of logs with `t` and `b`.

### Testing Instructions

Try it out!

Any tests I should add? I don't believe I see any existing tests for logs scrolling, but can add some.


crates/turborepo-ui/src/tui/event.rs
```diff
@@ -64,6 +68,7 @@ pub enum Event {
     SearchBackspace,
 }
 
+#[derive(Clone)]
```
chris-olszewski: Total nit, but for this we should do a `Copy` implementation as well

```suggestion
#[derive(Copy, Clone)]
```

crates/turborepo-ui/src/tui/popup.rs
```diff
@@ -6,19 +6,23 @@ use ratatui::{
...
-    "h      - Toggle task list",
-    "p      - Toggle pinned task selection",
-    "/      - Filter tasks to search term",
-    "ESC    - Clear filter",
-    "i      - Interact with task",
-    "Ctrl+z - Stop interacting with task",
-    "c      - Copy logs selection (Only when logs are selected)",
-    "u      - Scroll logs up",
-    "d      - Scroll logs down",
+const BIND_LIST: [&str; 16] = [
```
chris-olszewski, line 9: We don't need to keep manually updating the length of the underlying array if we only return a slice. Change the type and add a `.as_slice()` call at the end of the array i.e. `];` -> `].as_slice();`
```suggestion
const BIND_LIST: &[&str] = [
```

crates/turborepo-ui/src/tui/app.rs
```diff
@@ -201,6 +201,15 @@ impl<W> App<W> {
         Ok(())
     }
 
+    pub fn scroll_terminal_output_by_page(&mut self, direction: Direction) -> Result<(), Error> {
+        let pane_rows = self.size.pane_rows();
+        let task = self.get_full_task_mut()?;
+        // Scroll by the height of the terminal pane
+        task.scroll_by(direction, usize::from(pane_rows))?;
```
chris-olszewski, line 208: I hope you don't mind, I added a new method that generalized scrolling instead of scrolling by one in a for loop

chris-olszewski (approved):
Just some minor suggestions, but LFG

crates/turborepo-ui/src/tui/app.rs
anthonyshew: Nice!

crates/turborepo-ui/src/tui/popup.rs
```diff
@@ -6,19 +6,23 @@ use ratatui::{
...
+    "i       - Interact with task",
+    "Ctrl+z  - Stop interacting with task",
+    "c       - Copy logs selection (Only when logs are selected)",
+    "u       - Scroll logs up",
+    "d       - Scroll logs down",
+    "Shift+u - Page logs up",
+    "Shift+d - Page logs down",
+    "t       - Jump to top of logs",
+    "b       - Jump to bottom of logs",
 ];
```
chris-olszewski:
```suggestion
].as_slice();
```

PR #10253, opened 2025-03-30
Inputs hashing issue

Leksat:
Starting point to fix https://github.com/vercel/turborepo/issues/8051


chris-olszewski:
I'm preemptively closing this just to make sure nobody puts time into this before we are ready to accept a PR for this. 

Changing to hashing like this will be a breaking change that can only happen in 3.0 and other outcomes such as no longer supporting `--dry` need to be considered before work is started on this.

PR #10260, opened 2025-03-31
chore: upgrade turborepo-process and turborepo-filewatch to Rust 2024 edition

ognevny:
### Description

in #10114 I tried to update turborepo-filewatch, but it failed due to functions being automatically marked as unsafe when using `unsafe extern "C"`. Rust provides a `safe` keyword to fall to previous behavior.

turborepo-process was added recently but it used edition 2021

### Testing Instructions

CI. I don't have an Apple machine :(


ognevny:
should be ready :)

chris-olszewski (approved):
Thank you!

PR #10275, opened 2025-04-02
docs: new home page

anthonyshew:
### Description

New home page design!

### Testing Instructions

👀


anthonyshew:
Used as base for #10423. Merged that one so don't need this one!

PR #10281, opened 2025-04-03
chore(examples): upgrade Turborepo version

anthonyshew:
### Description

Bumping `turbo` version in examples.

### Testing Instructions

CI


examples/with-prisma/packages/database/.gitignore
anthonyshew, line 1: This gitignore was missing, so the postinstall was trying to add this directory and it got committed on my previous commit.

examples/with-vite-react/pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '9.0'
+lockfileVersion: '6.0'
```
chris-olszewski, line 1: Just flagging that this is the correct lockfile verison for the pnpm version. Previous one was incorrect

chris-olszewski (approved):

PR #10282, opened 2025-04-03
chore: update to Rust 1.86.0

ognevny:
### Description

- trait_upcasting is now stable
- move some .as_bytes() calls before indexing
- use repeat_n where applicable
- one .then() to if .. else
- formatting changes (again)


ognevny:
can't determine the cause of macos test failure...

ognevny:
> can't determine the cause of macos test failure...

it's not failing now

chris-olszewski (approved):

PR #10288, opened 2025-04-04
feat: dependency syncing checker

anthonyshew:
### Description

Dependency checker to see if everything is in sync in all your package.jsons. Reports errors when there are multiple versions of the same direct dependency in the repo.

### Testing Instructions

Wrote some tests - but do try it out!

TODO:
- [ ] Integration tests
- [ ] schema.json
- [ ] Bikeshed names (command, flag, and turbo.json key)
- [ ] Documentation


anthonyshew:
Also doing some experimenting on branch `shew/dep-checker`.

PR #10293, opened 2025-04-05
docs: list pnpm as first package manager

anthonyshew:
### Description

After spending enough time with monorepo users, we've developed the opinion that pnpm should be the preferred package manager that we point people to.

We're nudging pnpm to the top of the tabs where we list package managers to lean on this preference.

> [!NOTE]
> While working on this, I found out that these tabs aren't syncing with each other. Changing to pnpm only changes for the one you clicked, not all `<PackageManagerTabs />`. The bug is not introduced by this PR, and I'll fix it separately. 

### Testing Instructions

👀

### DIFF
--- a/docs/site/components/icons/magnifying-glass.tsx
+++ b/docs/site/components/icons/magnifying-glass.tsx
@@ -3,13 +3,13 @@ export const MagnifyingGlass = ({ className }: { className?: string }) => {
     <svg
       className={className}
       height="16"
-      stroke-linejoin="round"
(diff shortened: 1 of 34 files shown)


docs/site/components/icons/magnifying-glass.tsx
anthonyshew, line 1: Unrelated. Was producing warnings that were annoying me.

chris-olszewski (approved):

PR #10296, opened 2025-04-06
fix(docs): update redirects rule to resolve access issue on `getting-started` index page

ohprettyhak:
### Description

This PR fixes a redirection issue where clicking the **“Next: Getting Started”** button at the bottom of the **Introduction** page in the Turborepo docs incorrectly navigates back to the Introduction page.

The root cause is a redirect rule defined in `next.config.ts` that overrides the expected navigation.  
To resolve this, the conflicting `redirects` rule has been removed to allow proper access to the Getting Started index page.

### Testing Instructions

1. Clone this PR locally and run the project.
2. Navigate to the **Introduction** page in the documentation.
3. Click on the **“Next: Getting Started”** button at the bottom.
4. Verify that it correctly navigates to the **Getting Started** index page instead of redirecting back to **Introduction**.


docs/site/next.config.ts
```diff
@@ -40,11 +40,6 @@ const config = {
   },
   redirects() {
     return [
```
ohprettyhak, line 42:
```suggestion
    return [
      {
        source: "/docs/getting-started",
        destination: "/repo/docs/getting-started/installation",
        permanent: true,
      },
```

If the Getting Started index page is no longer needed, it would be more appropriate to update the redirect rule to point directly to `/getting-started/installation` instead.

anthonyshew (approved):
Ah, good find! Thank you!

anthonyshew:
CI is stuck on unrelated checks. Merging past these.

PR #10314, opened 2025-04-08
fix(@turbo/repository): no longer require packageManager to detect npm

chris-olszewski:
### Description

Previously we had detected package manager, but wouldn't inform the package graph builder of our findings. This would later cause an error since the package graph builder wasn't configured to support detecting the package manager.

We now pass the initially discovered package manager in so we don't need to detect it.

This will still fail in cases of `pnpm` and `yarn` if those binaries are not installed since we cannot determine which version of the package manager the user wants to use.

### Testing Instructions

Added new npm monorepo test fixture that is missing `packageManager` and a test that uses `affectedPackages` against it.


tknickman (approved):
Thanks!

PR #10340, opened 2025-04-14
fix(watch): allow usage of turbo.jsonc

### LINKED ISSUE #10323: Watch mode doesn't work inside a project with nested turbo.jsonc
### Verify canary release

https://github.com/neochief/bug-turborepo-jsonc

2.5.1-canary.0

### Expected behavior

I created turbo.jsonc at the project root, and everything was fine. As soon as I extended it in the subpackage, the watch mode stopped working.

### Actual behavior

```
> turbo watch build                                                                                                               
turbo 2.5.1-canary.0

  × Cannot use root turbo.json at /Users/neochief/www/testtest/turbo.jsonc with Watch Mode.

```

### To Reproduce

1. Create a root turbo.jsonc.
2. Create an extending turbo.jsonc in a subpackage.

anthonyshew:
### Description

Fixes #10323

### Testing Instructions

I'm testing by hand as I'm aware its notoriously difficult to write tests in this area of the codebase.

### DIFF
--- a/crates/turborepo-lib/src/package_changes_watcher.rs
+++ b/crates/turborepo-lib/src/package_changes_watcher.rs
@@ -23,11 +23,7 @@ use turborepo_repository::{
 };
 use turborepo_scm::GitHashes;
 
-use crate::{
-    commands::{self, CommandBase},
-    config::ConfigurationOptions,
-    turbo_json::{TurboJson, TurboJsonLoader, CONFIG_FILE, CONFIG_FILE_JSONC},
-};
+use crate::turbo_json::{TurboJson, TurboJsonLoader, CONFIG_FILE, CONFIG_FILE_JSONC};
 
 #[derive(Clone)]
 pub enum PackageChangeEvent {
@@ -483,16 +479,23 @@ impl Subscriber {
             }
         };
 
+        // Create a separate task for handling the exit signal
+        let exit_signal = tokio::spawn(async move {
+            let _ = exit_rx.await;
+            tracing::debug!("exiting package changes watcher due to signal");
+            true
+        });
+
+        // Run both main tasks concurrently and exit when either completes
         tokio::select! {
-            biased;
-            _ = exit_rx => {
-                tracing::debug!("exiting package changes watcher due to signal");
-            },
             _ = event_fut => {
                 tracing::debug!("exiting package changes watcher due to file event end");
-            },
+            }
             _ = changes_fut => {
                 tracing::debug!("exiting package changes watcher due to process end");
+            }
+            _ = exit_signal => {
+                // Exit signal already logged in the task
             }
         }
     }
--- a/crates/turborepo-lib/src/package_changes_watcher.rs
+++ b/crates/turborepo-lib/src/package_changes_watcher.rs
@@ -171,13 +171,14 @@ impl Subscriber {
             return None;
         };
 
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/run/watch.rs
```diff
@@ -120,9 +120,14 @@ impl WatchClient {
...
-        // Check if the turbo.json path is the standard one
-        let standard_path = base.repo_root.join_component(CONFIG_FILE);
-        if base.opts.repo_opts.root_turbo_json_path != standard_path {
+        // Check if the turbo.json path is the standard one (either turbo.json or
+        // turbo.jsonc)
+        let standard_path_json = base.repo_root.join_component(CONFIG_FILE);
+        let standard_path_jsonc = base.repo_root.join_component(CONFIG_FILE_JSONC);
+
+        if base.opts.repo_opts.root_turbo_json_path != standard_path_json
+            && base.opts.repo_opts.root_turbo_json_path != standard_path_jsonc
```
anthonyshew, line 129: I haven't been in any of the code for `watch` and have little context so not sure if this is an okay way to go about it. I just did the first, most minimally naive thing I saw.

chris-olszewski (changes requested):
This will avoid the error, but we need to implement the correct behavior so that `turbo.jsonc` triggers a package watcher event.

anthonyshew:
@chris-olszewski 🤦 When I hand-tested it, I was using a Just-in-Time package, so of course it looked like it worked.

I added some code that for filewatching (I hope) and used a Compiled Package now and hand-tested again. It looks like it works (for real this time) - but I only tested a simple case. Do I look like I'm more on the right track now?

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -426,17 +479,24 @@ impl Subscriber {
             }
         };
 
+        // Create a separate task for handling the exit signal
+        let exit_signal = tokio::spawn(async move {
+            let _ = exit_rx.await;
+            tracing::debug!("exiting package changes watcher due to signal");
+            true
+        });
```
chris-olszewski, line 487: Guessing this was cursor, but please undo this change, the previous signal handling is the behavior we want.

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -171,9 +171,29 @@ impl Subscriber {
...
+                tracing::warn!(
+                    "Found both turbo.json and turbo.jsonc in {}. Using turbo.json for watching.",
+                    self.repo_root
+                );
+                turbo_json_path
+            }
+            (true, false) => turbo_json_path,
+            (false, true) => turbo_jsonc_path,
+            (false, false) => turbo_json_path, // Default to turbo.json if neither exists
+        };
```
chris-olszewski, line 192: Not blocking, but would love to dedupe this logic with `Config::root_turbo_json_path` instead of copy-pasta. Would need to break out the behavior to a new static function.

chris-olszewski (approved):
LGTM, only thing that really needs to get changed is the change to the exit handling in the package watcher

crates/turborepo-lib/src/package_changes_watcher.rs
anthonyshew: Caught me red-handed.

Issue #10344, opened 2025-04-16
📚 Docs: Generator URL is incorrect


sovetski:
### What is the improvement or update you wish to see?

Hi,

After installing the latest version of turborepo, the generated `packages/ui/turbo/generators/config.ts` contains this link: https://turbo.build/docs/core-concepts/monorepos/code-generation which is incorrect

### Is there any context that might help us understand?

Maybe the correct link is: https://turbo.build/docs/guides/generating-code

### Does the docs page already exist? Please link to it.

https://turbo.build/docs/guides/generating-code

PR #10345, opened 2025-04-16
docs: replace incorrect generator links (#10344)

sovetski:
### Description

See: https://github.com/vercel/turborepo/issues/10344


packages/turbo-gen/src/generators/custom.ts
```diff
@@ -1,4 +1,4 @@
-import { logger } from "@turbo/utils";
+https://turbo.build/docs/guides/generating-codeimport { logger } from "@turbo/utils";
```
chris-olszewski, line 1:
```suggestion
import { logger } from "@turbo/utils";
```

chris-olszewski (approved):

PR #10374, opened 2025-04-23
ci: upgrade to Node.js 22

anthonyshew:
### Description

Upgrading our repository to Node.js 22 by reducing our source of truth to the value in `package.json#engines`.

Notably, this PR also adds Node.js 22 to the JS package test matrix.

### Testing Instructions

CI


package.json
```diff
@@ -42,6 +42,6 @@
   },
   "packageManager": "pnpm@8.14.0",
   "engines": {
-    "node": "20.x"
+    "node": "22.x"
```
anthonyshew, line 45: I'm upgrading the `engines` in package.json. I don't know if this is contentious or not.

anthonyshew: As it turns out, we should definitely be doing this. This is now our source of truth.

.github/actions/setup-node/action.yml
```diff
@@ -27,6 +26,7 @@ runs:
     - name: Setup Node.js
       uses: actions/setup-node@v4
       with:
+        node-version-file: "package.json"
```
anthonyshew, line 29: This sets the default for our CI by reading `package.json#engines`.

.github/actions/setup-node/action.yml
```diff
@@ -16,7 +16,6 @@ inputs:
   node-version:
     description: "Node version to install"
     required: false
-    default: "18"
```
anthonyshew, line 19: A default is no longer needed because we're defaulting to `package.json#engines`.

.github/actions/setup-turborepo-environment/action.yml
```diff
@@ -7,7 +7,6 @@ inputs:
   node-version:
     description: Node version
     required: false
-    default: "18"
```
anthonyshew, line 10: https://github.com/vercel/turborepo/pull/10374/files#r2057505677

.github/actions/setup-node/action.yml
```diff
@@ -27,6 +26,8 @@ runs:
     - name: Setup Node.js
       uses: actions/setup-node@v4
       with:
+        # node-version-file is the default, but can be overridden using node-version
+        node-version-file: ".node-version"
```
anthonyshew, line 30:
```suggestion
        node-version-file: "package.json"
```

PR #10390, opened 2025-04-27
ci(docs): add spell checker

anthonyshew:
### Description

Making the spell checker script that we had into a CI check. This check is intentionally non-blocking.

### Testing Instructions

CI


chris-olszewski (approved):
Thank you for saving me from myself

PR #10405, opened 2025-04-29
docs(chore): switch to ESM

anthonyshew:
### Description

Switching to ESM - because we should!

### Testing Instructions

CI


docs/site/app/api/remote-cache-spec/route.ts
```diff
@@ -1,4 +1,3 @@
-// @ts-expect-error - This file is generated at build time.
```
anthonyshew, line 1: Unrelated. Realizing I didn't need to do this in https://github.com/vercel/turborepo/pull/10404. I should've [fixed the Task Graph](https://github.com/vercel/turborepo/pull/10405/files#diff-1089f50b3e47389554c4d77b6022940edb9fd6aa01f8cf5c56bc5aa3909366ecR54) instead.

chris-olszewski (approved):

PR #10420, opened 2025-05-01
feat(tui): momentum-based scrolling

MendyLanda:
### Description
#10415 Adds momentum-based scrolling to the TUI for a smoother and more natural scroll experience.

### Testing Instructions
1. Run a long-running task that prints many lines, e.g.:
```bashs
perl -E '$i=1; for (1..1500) { say "[$i] " . "=" x int(rand(100) + 1); $i++ }' && read
```

2. Use the TUI to scroll through the output and observe the new scrolling behavior.

### DIFF
--- a/crates/turborepo-ui/src/tui/scroll.rs
+++ b/crates/turborepo-ui/src/tui/scroll.rs
@@ -14,18 +14,18 @@ const MIN_VELOCITY: f32 = 1.0;
 /// Increase for faster acceleration (reaches top speed quicker, feels
 /// snappier). Decrease for slower, smoother acceleration (takes longer to reach
 /// top speed).
-const ACCELERATION: f32 = 2.0;
+const ACCELERATION: f32 = 0.3;
 
 /// How long (in ms) between scrolls before momentum resets.
 /// Increase to allow longer pauses between scrolls while keeping momentum.
 /// Decrease to require faster, more continuous scrolling to maintain momentum.
-const DECAY_TIME: Duration = Duration::from_millis(200);
+const DECAY_TIME: Duration = Duration::from_millis(350);
 
 /// Only process 1 out of every N scroll events (throttling).
 /// Increase to make scrolling less sensitive to high-frequency mouse wheels
 /// (e.g. trackpads). Decrease to process more events (smoother, but may be
 /// too fast on some input devices).
-const THROTTLE_FACTOR: u8 = 1;
+const THROTTLE_FACTOR: u8 = 3;
 
 /// Tracks and computes momentum-based scrolling.
 pub struct ScrollMomentum {
--- a/crates/turborepo-ui/src/tui/scroll.rs
+++ b/crates/turborepo-ui/src/tui/scroll.rs
@@ -1,0 +1,77 @@
+use std::time::{Duration, Instant};
+
+use crate::tui::event::Direction;
+
+const MAX_VELOCITY: f32 = 12.0; // max lines per event
+const MIN_VELOCITY: f32 = 1.0; // min lines per event
+const ACCELERATION: f32 = 2.0; // lines per event per fast scroll
+const DECAY_TIME: Duration = Duration::from_millis(200); // ms to reset momentum
+
+/// Core struct to track and compute momentum-based scrolling.
+pub struct ScrollMomentum {
+    velocity: f32,
+    last_event: Option<Instant>,
+    phase: Phase,
+    last_direction: Option<Direction>,
+}
+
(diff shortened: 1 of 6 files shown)


MendyLanda:
@anthonyshew for your review.

The config values are subjective and based on personal preference. I tried to match native scrolling as closely as possible. While trackpads could benefit from custom settings, it's not possible to reliably distinguish between trackpad and mouse wheel, and behavior varies a lot across devices and OSes. This still feels like a significant improvement to me.

anthonyshew:
@MendyLanda, this is great! Thank you!

I tried it out and it feels much better than what we have now. Significantly better.

I'm testing with a trackpad and it felt like I would go from "no momentum" to "ALL THE MOMENTUM!!!" Could we tune the ramping? I don't have a mouse wheel around to try with so I don't know if it feels great there and re-tuning would make it feel terrible. I can suggest what values feel right to me for the trackpad if you don't have one around.

Very excited about this.

anthonyshew:
>  it's not possible to reliably distinguish between trackpad and mouse wheel, and behavior varies a lot across devices and OSes

And, yes, agree with this. Let's get to some best effort that feels good (particularly for trackpad vs. mouse wheel) as best we can.

MendyLanda:
> @anthonyshew for trackpad setting you can try this: THROTTLE_FACTOR: u8 = 3 ACCELERATION: f32 = 0.2

But this completely voids mouse wheel.

I am looking at [Ghostty](https://github.com/ghostty-org/ghostty)'s implementation of this, maybe I can find some ideas. If you have any suggestion I'd love to know!

anthonyshew:
> @anthonyshew for trackpad setting you can try this: THROTTLE_FACTOR: u8 = 3 ACCELERATION: f32 = 0.2

Confirming that these values feel great for my trackpad. 👍

MendyLanda:
> I am looking at [Ghostty](https://github.com/ghostty-org/ghostty?rgh-link-date=2025-05-01T17%3A38%3A31Z)'s implementation of this, maybe I can find some ideas.


TL;DR:
It’s impossible to detect precision scrolling (touchpad) and apply the delta conditionally. This is a limitation of crossterm (ANSI codes).
So, we have two options:


1. Let the user toggle between touchpad and mouse wheel mode.
[+] Scrolling always feels great.
[–] We already have too many scroll-related keybinds, so adding another for something that’s usually a one-off isn’t ideal.

2. Find an acceptable middle ground that works (mostly) for both.
[+] Doesn’t add another keybind; users don’t need to know about this.
[+] Still feels better than what we currently have.
[–] Could be improved further.


@anthonyshew, your call. Option 2 feels more reasonable, but personally, I’d prefer option 1 😄.
If we go with option 2, these settings seem to be a good middle ground:
 ```rs
const ACCELERATION: f32 = 0.3; // 2.0 is optimal for mouse wheel, 0.3 favors touchpad. Higher values make the touchpad too jumpy.
const DECAY_TIME: Duration = Duration::from_millis(350); // 300ms or less is too fast for mouse wheel; 350ms slightly favors mouse wheel.
const THROTTLE_FACTOR: u8 = 3; // Lower values make the touchpad unusable.
```

**Comment by anthonyshew** [comment]
That feels great for me on trackpad. Soft-approved from me! Looking for a teammate with a mouse to confirm their mouse feels right...

**Comment by chris-olszewski** [comment]
Thank you so much for the PR! I suggested some changes in https://github.com/MendyLanda/turborepo/pull/1 that make it feel a little better with a mouse scroll wheel.

**Comment by MendyLanda** [comment]
> Thank you so much for the PR! I suggested some changes in [MendyLanda#1](https://github.com/MendyLanda/turborepo/pull/1) that make it feel a little better with a mouse scroll wheel.

Thank you! This was the missing piece! feels great on both my mouse (logitech mx2) and my mb trackpad!

**Review by chris-olszewski — APPROVED** [review]
This is a great addition. Thank you again for the PR!

File: crates/turborepo-ui/src/tui/scroll.rs [hunk 1]
```diff
@@ -0,0 +1,105 @@
...
+/// too fast on some input devices).
+const THROTTLE_FACTOR: u8 = 3;
+
+/// Tracks and computes momentum-based scrolling.
+pub struct ScrollMomentum {
+    velocity: f32,
+    last_event: Option<Instant>,
+    last_direction: Option<Direction>,
+    throttle_counter: u8,
+}
```
chris-olszewski, line 41:
```suggestion
}

impl Default for ScrollMomentum {
    fn default() -> Self {
        Self::new()
    }
}
```

PR #10422, opened 2025-05-01
chore: add vercel builder

chris-olszewski:


anthonyshew:
This was a test and we got the info we needed!

PR #10448, opened 2025-05-06
fix(query): use patched version of async-graphql

chris-olszewski:
### Description

Work around for https://github.com/async-graphql/async-graphql/issues/1703

### Testing Instructions

`turbo_dev --skip-infer query` in repo of your choice. GraphiQL IDE should load correctly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,8 +82,8 @@ async-compression = { version = "0.3.13", default-features = false, features = [
   "gzip",
   "tokio",
 ] }
-async-graphql = "7.0.7"
-async-graphql-axum = "7.0.7"
+async-graphql = "7.0.16"
+async-graphql-axum = "7.0.16"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -82,8 +82,8 @@ async-compression = { version = "0.3.13", default-features = false, features = [
   "gzip",
   "tokio",
 ] }
-async-graphql = "7.0.7"
-async-graphql-axum = "7.0.7"
+async-graphql = "7.0.16"
+async-graphql-axum = "7.0.16"
```
chris-olszewski, line 86: Upgraded these so that the patched version of `async-graphql` only differs from `7.0.16` by [this commit](https://github.com/yasamoka/async-graphql/commit/15fed95142019722e7da2c214ff9d8e2e755c9e5)

PR #10437, opened 2025-05-03
fix: replace @repo/ui/tsconfig.json for with-svelte template to fix errors

DoctorRyner:
### Description

I tested it with VS Code and Helix Editor and with current config, all export/imports of files, for instance in index.ts config, will result in a false error reporting of "Module does not exist", it seems to be happening because of miss configuration.

### Testing Instructions

1. Create project with `bunx create-turbo@latest -e with-svelte`
2. Go to `@repo/ui/src/index.ts` and check if typescript gives you any errors (it should)
3. Replace @repo/ui/tsconfig.json with:

```json
{
  "extends": ["@repo/typescript-config/svelte.json", "./.svelte-kit/tsconfig.json"]
}
```

And check if import/export errors are gone in VS Code (it's the change made in this pull request)

PR #10450, opened 2025-05-06
fix(examples): with-svelte example TypeScript

anthonyshew:
### Description

Inspired by @DoctorRyner in #10437, a set of changes to make TypeScript work better for the `with-svelte` example.

### Testing Instructions

This set of changes works correctly in my editor.

### DIFF
--- a/examples/with-svelte/packages/ui/tsconfig.json
+++ b/examples/with-svelte/packages/ui/tsconfig.json
@@ -1,12 +1,10 @@
 {
-  "extends": "@repo/typescript-config/svelte.json",
+  "extends": ["@repo/typescript-config/svelte.json", "./.svelte-kit/tsconfig.json"],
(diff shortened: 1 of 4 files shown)


examples/with-svelte/packages/ui/tsconfig.json
```diff
@@ -1,15 +1,10 @@
 {
-  "extends": "@repo/typescript-config/svelte.json",
+  "extends": ["@repo/typescript-config/svelte.json", "./.svelte-kit/tsconfig.json"],
```
chris-olszewski, line 2: `.svelte-kit` only gets generated after a build is run so running `check-types` before building will result in failure

chris-olszewski (approved):
Approving, but this still can result in failures if checking types before building

PR #10460, opened 2025-05-08
docs: [correction] nx remote caching is free like Turborepo

JamesHenry:
The deleted statement was **never** accurate in Nx. We briefly openly explored some alternative solutions and pricing structures for self-hosted caching but this _never_ landed as a requirement in Nx. We took on the feedback loud and clear that our initial thinking was misguided and we changed course _before_ it landed.

We offer a broader array of free cache hosting options than Turborepo, with the OpenAPI spec being the exact equivalent of how Turborepo chooses to approach the problem, you can learn more here:

https://nx.dev/recipes/running-tasks/self-hosted-caching

For this change, I went for the most generous correction I could, but you may decide the overall section is redundant without this key incorrect point.

While I'm here, I'd suggest you run through this guide yourself again because it doesn't really make sense currently. If you run the command you have listed Nx generates package manager workspaces, no `project.json` files, and `package.json` files for each project, but then you tell people to add these things because that's "only how Turborepo does things" (not quoting but that's the inaccurate gist)...

Finally, as pointed out last time your `nx.json` to `turbo.json` comparison is very, very far from apples to apples. On the nx side you are demonstrating the configuration of a bunch of capabilities that Turborepo simply does not have and then saying "see, look, nx _requires_ more config", which doesn't seem fair 🤷

Nx can work _exactly_ how Turborepo works, and it can _also_ work with optional plugins, users have complete freedom of choice. If you'd like to collaborate on further rewrites I'm totally open to that, but I thought I would keep this PR very focused on this critical correction for now.

### DIFF
--- a/docs/site/content/docs/guides/migrating-from-nx.mdx
+++ b/docs/site/content/docs/guides/migrating-from-nx.mdx
@@ -145,7 +145,7 @@ Migrating to Turborepo will likely require deleting previous configuration that 
 
 Turborepo’s [Remote Caching](/docs/core-concepts/remote-caching) stores the results of your task on a cloud server. This saves enormous amounts of time by **preventing duplicated work across your entire organization**. [Vercel Remote Cache](https://vercel.com/docs/monorepos/remote-caching) has saved teams over 500 years of compute so far.
 
-Since Nx 19.7, similar functionality is a paid-for feature, even when self-hosting. Remote Caching with Turborepo is free when [self-hosting](/docs/core-concepts/remote-caching#self-hosting) or using [Vercel Remote Cache](https://vercel.com/docs/monorepos/remote-caching).
+Nx offers multiple forms of Free Remote Caching: Nx Cloud, Free Self-Hosted via plugins, or via OpenAPI spec like Turborepo.
 
 ## Migration steps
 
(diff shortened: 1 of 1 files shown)


docs/site/content/docs/guides/migrating-from-nx.mdx
```diff
@@ -145,7 +145,7 @@ Migrating to Turborepo will likely require deleting previous configuration that
 
 Turborepo’s [Remote Caching](/docs/core-concepts/remote-caching) stores the results of your task on a cloud server. This saves enormous amounts of time by **preventing duplicated work across your entire organization**. [Vercel Remote Cache](https://vercel.com/docs/monorepos/remote-caching) has saved teams over 500 years of compute so far.
 
-Since Nx 19.7, similar functionality is a paid-for feature, even when self-hosting. Remote Caching with Turborepo is free when [self-hosting](/docs/core-concepts/remote-caching#self-hosting) or using [Vercel Remote Cache](https://vercel.com/docs/monorepos/remote-caching).
+Nx offers multiple forms of Free Remote Caching: Nx Cloud, Free Self-Hosted via plugins, or via OpenAPI spec like Turborepo.
```
anthonyshew, line 148:
```suggestion
```

anthonyshew (changes requested):
Glad to hear of the change of course! In that case, let's delete the section entirely since its no longer a point of difference. 👍

anthonyshew (approved):
Thanks!

JamesHenry:
Thank you!

PR #10464, opened 2025-05-09
fix(bun): only include required entries

chris-olszewski:
### Description

Attempt to fix #4694

Not currently finished, but it `prune` does now only have an overfit of 4 entries

### Testing Instructions

Added unit test


anthonyshew:
Superceded by https://github.com/vercel/turborepo/pull/10729

PR #10482, opened 2025-05-15
chore: update to Rust 1.87.0

ognevny:
### Description

- clippy::result_large_err requires huge code refactoring, so ignore this lint
- extract_if features are now stable
- use io::Error::other where applicable
- some more small clippy suggestions


ognevny:
@chris-olszewski can you have a look into it, please?

crates/turborepo-ui/src/color_selector.rs
```diff
@@ -38,14 +38,10 @@ impl ColorSelector {
...
 
-        let color = {
-            self.inner
-                .write()
-                .expect("lock poisoned")
-                .insert_color(key.to_string())
-        };
-
-        color
+        self.inner
```
chris-olszewski, line 41: I'm a little iffy on this one, does this alter when the lock gets released at all? Just wondering because the code was clearly written to ensure that the lock got released as soon as the color was inserted.

ognevny:
@chris-olszewski I've reverted `color_selector` change and ignored the lint

chris-olszewski (approved):

PR #10493, opened 2025-05-20
feat: add `Workspace.find_package_by_path`

scubbo:
### Description

As raised [here](https://linear.app/vercel/issue/TURBO-4724/expose-package-path-in-turborepository).

Depending on the call-pattern that we want to support, it might be worthwhile to introduce a cache for ancestor-paths so that subsequent calls only need to traverse-up to a first-common-ancestor.

### Testing Instructions

`pnpm build && pnpm test` within `packages/turbo-repository` passes.

### Note

I'm very much a Rust newbie, so please don't hold back on stylistic nits or comments - I have no strong opinions on "correct" Rust!


scubbo:
AFAICT that deployment failure is unrelated to this code, though I don't know for sure.

packages/turbo-repository/rust/src/lib.rs
```diff
@@ -291,4 +291,39 @@ impl Workspace {
...
+    #[napi]
+    pub async fn find_package_by_path(&self, path: String) -> Result<Package, Error> {
+        let package_mapper = DefaultPackageChangeMapper::new(&self.graph);
+        let anchored_path = AnchoredSystemPath::new(&path)
+            .map_err(|e| Error::from_reason(e.to_string()))?
+            .clean();
+        match package_mapper.detect_package(&anchored_path) {
+            turborepo_repository::change_mapper::PackageMapping::All(
+                _all_package_change_reason,
+            ) => Err(Error::from_reason("file belongs to many packages")),
```
chris-olszewski, line 311: An alternative here would be to possibly return the root package or return that the file change is considered "global".

packages/turbo-repository/__tests__/find-packages.test.ts
```diff
@@ -26,4 +27,52 @@ describe("findPackages", () => {
     assert.deepEqual(pkg2.dependencies, []);
     assert.deepEqual(pkg2.dependents, ["apps/app"]);
   });
+
+  it("returns the package for a given path", async () => {
+    const workspace = await Workspace.find(MONOREPO_PATH);
+
+    for (const [filePath, result] of [
+      ["apps/app/src/util/useful-file.ts", "app-a"],
```
chris-olszewski, line 35: We should have some tests for root files e.g. a root `tsconfig.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`

chris-olszewski (changes requested):
Only concerns are around beefing up the tests.

chris-olszewski (approved):

PR #10503, opened 2025-05-24
Use entry instead of entryPoints

melroy89:
### Description

In tsup `entryPoints` is deprecated. You should use `entry` instead.

Also, I use src/**/*, so in case of projects using multiple directories in the `src` directory. That also keeps working.

### Testing Instructions

Try to run the example project again eg. `pnpm run dev`.


anthonyshew (approved):
Thanks!

PR #10513, opened 2025-05-27
feat: specific error message for potential debugging flags

anthonyshew:
### Description

When a user tries to use `--debug` or `--verbose` as flags, we can make a pretty strong bet that they're trying to get some debug logging from us. In this case, we can supply better direction to the developer.

### Testing Instructions

Before:
```
▲ 👟 turbo on shew/8d2a6
  dt anything --debug
  × The --debug flag is not supported. Please use --verbosity instead.
  │  if you are trying to pass `--debug` as a value, use `-- --debug`.
```

AFTER:
```
▲ 👟 turbo on shew/8d2a6
  turbo anything --debug
 ERROR  unexpected argument '--debug' found

  tip: to pass '--debug' as a value, use '-- --debug'

Usage: turbo <--cache-dir <CACHE_DIR>|--concurrency <CONCURRENCY>|--continue[=<CONTINUE>]|--single-package|--framework-inference [<BOOL>]|--global-deps <GLOBAL_DEPS>|--env-mode [<ENV_MODE>]|--filter <FILTER>|--affected|--output-logs <OUTPUT_LOGS>|--log-order <LOG_ORDER>|--only|--pkg-inference-root <PKG_INFERENCE_ROOT>|--log-prefix <LOG_PREFIX>|TASKS|PASS_THROUGH_ARGS>

```

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -278,7 +278,7 @@ pub struct Verbosity {
         value_name = "COUNT"
     )]
     /// Verbosity level. Useful when debugging Turborepo or creating logs for
-    /// issue reports.
+    /// issue reports
     pub verbosity: Option<u8>,
     #[clap(
         short = 'v',
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -277,7 +277,8 @@ pub struct Verbosity {
         conflicts_with = "v",
         value_name = "COUNT"
     )]
-    /// Verbosity level
+    /// Verbosity level. Useful when debugging Turborepo or creating logs for
+    /// issue reports.
     pub verbosity: Option<u8>,
     #[clap(
         short = 'v',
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -277,7 +277,8 @@ pub struct Verbosity {
         conflicts_with = "v",
         value_name = "COUNT"
     )]
-    /// Verbosity level
+    /// Verbosity level. Useful when debugging Turborepo or creating logs for
+    /// issue reports.
```
anthonyshew, line 281: This is for making our `--help` text every so slightly more explanatory.

crates/turborepo-lib/src/shim/parser.rs
```diff
@@ -142,6 +143,16 @@ impl ShimArgs {
                 // `--root-turbo-json=./path/to/turbo.json`, that entire chunk
                 // is a single arg, so we need to split it up.
                 root_turbo_json = Some(AbsoluteSystemPathBuf::from_unknown(&invocation_dir, path));
+            } else if arg == "--debug" {
+                return Err(Error::UnsupportedFlag {
+                    flag: "--debug".to_string(),
+                    suggestion: "Please use --verbosity instead.".to_string(),
```
anthonyshew, line 149:
```suggestion
                    suggestion: "Use --verbosity instead.".to_string(),
```

chris-olszewski (approved):
Awesome change.

PR #10515, opened 2025-05-28
docs: account for Custom Environments

anthonyshew:
### Description

Our script allowed for a custom environment to write to production. Instead, we make sure only the `VERCEL_TARGET_ENV` that is "production" can do so.


chris-olszewski (changes requested):
I don't think this is effective as it still lets us write non-production domains to the production index.

anthonyshew:
🤦 Of course. Added another line to throw when that's the case.

PR #10549, opened 2025-06-07
Fix bun prune peer_dependencies

camero2734:


LudvigHz:
Are there any plans to find a way to resolve this? I know a lot of people would love to be able to fully use turbo prune with bun 😄

anthonyshew:
I'll close this one because Bun has changed their lockfile format entirely. We're working on the new version in https://github.com/vercel/turborepo/pull/10729. Let's meet up there!

PR #10590, opened 2025-06-21
feat: opt into breaking changes and pre-stable features early with `turbo.json` configuration

anthonyshew:
### Description

We want to smooth migration paths for users when we introduce functionality with breaking changes. This can be difficult with point-in-time majors that might force a user to make many breaking changes at once.

Instead, we can introduce the concept of "future flags" to allow you to opt into breaking changes early. As a user, you get to have more control over when you adopt breaking changes. As a core team, we get more freedom to iterate, resulting in faster feedback cycles with users.

This PR only introduces the flag to the parser, not adding any flags quite yet. This key is only allowed in root `turbo.json`s.

### Testing Instructions

Added some tests. I also tried it out manually (and trivially). It worked how I'd hoped.

### DIFF
--- a/crates/turborepo-lib/src/turbo_json/mod.rs
+++ b/crates/turborepo-lib/src/turbo_json/mod.rs
@@ -164,6 +164,9 @@ pub struct RawTurboJson {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub concurrency: Option<String>,
 
+    #[serde(skip_serializing_if = "Option::is_none", rename = "futureFlags")]
+    pub future_flags: Option<serde_json::Value>,
+
     #[deserializable(rename = "//")]
     #[serde(skip)]
     _comment: Option<String>,
@@ -1357,6 +1360,37 @@ mod tests {
             Err(e) => Err(e.to_string()),
         };
         assert_eq!(actual, expected.map_err(|s| s.to_owned()));
+    }
+
+    #[test]
+    fn test_deserialize_future_flags() {
+        let json = r#"{
+            "tasks": {
+                "build": {}
+            },
+            "futureFlags": {
+                "newFeature": true,
+                "experimentalOption": "value"
+            }
+        }"#;
+
+        let deserialized_result = deserialize_from_json_str(
+            json,
+            JsonParserOptions::default().with_allow_comments(),
+            "turbo.json",
+        );
+        let raw_turbo_json: RawTurboJson = deserialized_result.into_deserialized().unwrap();
+
+        // Verify that futureFlags is parsed correctly
+        assert!(raw_turbo_json.future_flags.is_some());
+        let future_flags = raw_turbo_json.future_flags.as_ref().unwrap();
+        assert_eq!(future_flags["newFeature"], json!(true));
+        assert_eq!(future_flags["experimentalOption"], json!("value"));
+
+        // Verify that the futureFlags field doesn't cause errors during conversion to
+        // TurboJson
+        let turbo_json = TurboJson::try_from(raw_turbo_json);
+        assert!(turbo_json.is_ok());
     }
 
     #[test]
--- a/crates/turborepo-lib/src/turbo_json/mod.rs
+++ b/crates/turborepo-lib/src/turbo_json/mod.rs
@@ -562,7 +562,7 @@ impl TryFrom<RawTurboJson> for TurboJson {
             return Err(Error::PipelineField { span, text });
         }
 
-        // Check if futureFlags is used in a workspace config (which has extends field)
+        // `futureFlags` key is only allowed in root turbo.json
         let is_workspace_config = raw_turbo.extends.is_some();
         if is_workspace_config {
             if let Some(future_flags) = raw_turbo.future_flags {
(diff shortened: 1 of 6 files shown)


crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -164,6 +164,9 @@ pub struct RawTurboJson {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub concurrency: Option<String>,
 
+    #[serde(skip_serializing_if = "Option::is_none", rename = "futureFlags")]
+    pub future_flags: Option<Spanned<serde_json::Value>>,
```
anthonyshew: Happy to change the key name if we like something else better! Was the first thing that came to mind.

chris-olszewski: `experimental` to match Next.js?

anthonyshew: I meant to nest keys into this `futureFlags` key and forgot. I intended to do:
```
{
  "futureFlags": {
    "beta: {},
    "experimental": {}
  }
}
```

What do you think of that? Notably, `experimental` is the name of one of our release phases, so I don't want to create confusion by ending up with some `experimental.foo`, while `foo` is actually in Beta.

chris-olszewski: Maybe just `flags` then? e.g. `flags.experimental.foo`

anthonyshew: This approach would mean we would have to consciously promote a feature through its phases, though, and I don't know if we would want that. That would mean breaking changes within this flag since it would break parsing when a flag moves between phases in a user's codebase. (I am okay with this, but I could see opposition.)

anthonyshew: That felt too close to application feature flags (e.g. https://flags-sdk.dev/) which worries me.

chris-olszewski: > This approach would mean we would have to consciously promote a feature through its phases

Not necessarily. I'll do a quick PoC to show how we could have `stable` ⊆ `beta` ⊆ `experimental` "for free" so users don't need to change flag declaration as they get stabilized.

anthonyshew: That feels confusing as well. I might be on a version of Turborepo where Boundaries are Beta, but then we would accept the flag in the `experimental` key? I guess we could warn?

chris-olszewski: I feel pretty strongly that we need to support reading beta flags from experimental path. Even with a warning it is easy enough to miss and a user not realize by upgrading `turbo` they unintentionally opted out of an experimental flag they had enabled because it is now in beta.

anthonyshew: I would say in that case we don't need to have the nested keys for phases at all. The idea there was to make things explicit so users would **know** what phase a feature is in, but if we accept them anywhere, then that explicitness is lost and its a layer of nesting that expresses effectively nothing.

Stick with `futureFlags: {}`?

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -558,6 +561,15 @@ impl TryFrom<RawTurboJson> for TurboJson {
...
         }
+
+        // `futureFlags` key is only allowed in root turbo.json
+        let is_workspace_config = raw_turbo.extends.is_some();
+        if is_workspace_config {
+            if let Some(future_flags) = raw_turbo.future_flags {
+                let (span, text) = future_flags.span_and_text("turbo.json");
+                return Err(Error::FutureFlagsInPackage { span, text });
+            }
+        }
```
chris-olszewski, line 576: Was trying to find how we prevent root level field usage outside of the root and suggest using that mechanism instead of this check. We don't prevent it 🙃

packages/turbo-types/schemas/schema.json
```diff
@@ -102,6 +102,12 @@
           "type": "boolean",
           "description": "When set to `true`, disables the update notification that appears when a new version of `turbo` is available.\n\nDocumentation: https://turborepo.com/docs/reference/configuration#noupdatenotifier",
           "default": false
+        },
+        "futureFlags": {
+          "type": "object",
+          "description": "Opt into breaking changes prior to major releases, experimental features, and beta features.",
+          "additionalProperties": false,
+          "default": {}
```
chris-olszewski: Please add to the types package and not just the generated schema

crates/turborepo-lib/src/turbo_json/mod.rs
anthonyshew: I can make a utility out of this in a separate PR?

packages/turbo-types/schemas/schema.json
anthonyshew: We should make this a CI check for the next time I forget to do it the right way.

anthonyshew: https://github.com/vercel/turborepo/pull/10609

crates/turborepo-lib/src/turbo_json/mod.rs
chris-olszewski: Oh yeah, nothing to do in this PR. Just future work

PR #10608, opened 2025-06-28
chore: update to Rust 1.88.0

ognevny:
### Description

- clippy.toml: `std::hash::Hasher::hash` is no longer detects as trait method. I removed its reference to fix tons of warnings
- inline `format!` (and some more) args: https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
- use `char_indices` where applicable: https://rust-lang.github.io/rust-clippy/master/index.html#char_indices_as_byte_indices (possibly introduces regressions)

### Testing Instructions

CI


ognevny:
@chris-olszewski can you have a look into it, please?

crates/turborepo-env/src/lib.rs
```diff
@@ -294,7 +294,7 @@ fn wildcard_to_regex_pattern(pattern: &str) -> String {
     let mut previous_index = 0;
     let mut previous_char: Option<char> = None;
 
-    for (i, char) in pattern.chars().enumerate() {
+    for (i, char) in pattern.char_indices() {
```
chris-olszewski, line 297: This is a behavior change if characters take up multiple bytes, but we already would panic on those strings. I think this will just break in a slightly different way which is okay.

Actually fixing this code can/should be fixed in it's own PR.

chris-olszewski:
Apologies for slow review. Thank you for keeping us current!

PR #10609, opened 2025-06-28
ci: ensure turbo-types schemas are in sync

anthonyshew:
### Description

In https://github.com/vercel/turborepo/pull/10590, I edited the JSON schema directly, which [isn't the right way](https://github.com/vercel/turborepo/pull/10590#discussion_r2171962005).

This PR introduces a CI check that I (and others) can't make this sort of mistake in the future.

### Testing Instructions

[This commit](https://github.com/vercel/turborepo/pull/10609/commits/3108ad0c57bfbeccfb60b87e7115b3398444e1de) shows a failure because I added a `test` key to the schema (and malformatted it in the process).

### DIFF
--- a/.github/workflows/turborepo-test.yml
+++ b/.github/workflows/turborepo-test.yml
@@ -131,11 +131,8 @@ jobs:
         run: git config --global core.autocrlf input
       - uses: actions/checkout@v4
 
-      - name: Setup Turborepo Environment
-        uses: ./.github/actions/setup-turborepo-environment
-        with:
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
-          node-version: "18.20.2" # TODO: Update integration tests with changed log output in Node.js 22
+      - name: "Setup Node"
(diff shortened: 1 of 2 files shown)


.github/workflows/turborepo-test.yml
```diff
@@ -224,6 +224,37 @@ jobs:
...
+    if: ${{ needs.find-changes.outputs.rest == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Turborepo Environment
+        uses: ./.github/actions/setup-turborepo-environment
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
```
chris-olszewski: This might be a little heavy. I think we could get away with just setting up Node. This action installs Rust/Protobuf/Capnp

chris-olszewski (approved):

PR #10610, opened 2025-06-29
fix: validation for root turbo.json vs. Package Configurations

anthonyshew:
### Description

We weren't checking fields that are root `turbo.json`-only aren't being used in Package Configurations - and vice versa. I did a check of this nature in https://github.com/vercel/turborepo/pull/10590 and have refactored that to a function that we can now use for other fields. Additionally, I've brought that idea along to the reverse situation.

### Testing Instructions

Added some tests ensuring the desired logic is what happens.

### DIFF
--- a/crates/turborepo-lib/src/turbo_json/mod.rs
+++ b/crates/turborepo-lib/src/turbo_json/mod.rs
@@ -1459,4 +1459,57 @@ mod tests {
             "`with` cannot use dependency relationships."
         );
     }
+
+    #[test]
+    fn test_field_placement_validation_via_parser() {
+        // Test root-only field in package config
+        let json = r#"{
+            "extends": ["//"],
+            "globalEnv": ["NODE_ENV"],
+            "tasks": {
+                "build": {}
+            }
+        }"#;
+
+        let result = RawTurboJson::parse(json, "turbo.json");
+        assert!(result.is_err());
+        let error = result.unwrap_err();
+        assert!(error.to_string().contains("Failed to parse turbo.json"));
+
+        // Test package-only field in root config
+        let json = r#"{
+            "tags": ["utility"],
+            "tasks": {
+                "build": {}
+            }
+        }"#;
+
+        let result = RawTurboJson::parse(json, "turbo.json");
+        assert!(result.is_err());
+        let error = result.unwrap_err();
+        assert!(error.to_string().contains("Failed to parse turbo.json"));
+
+        // Test valid root config
+        let json = r#"{
+            "globalEnv": ["NODE_ENV"],
+            "tasks": {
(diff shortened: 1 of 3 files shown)


turborepo-tests/integration/fixtures/turbo-configs/package-task.json
```diff
@@ -1,7 +1,5 @@
 {
   "$schema": "https://turborepo.com/schema.json",
-  "globalDependencies": ["foo.txt"],
```
anthonyshew: Exactly the sort of thing we want to catch! Got caught here: https://github.com/vercel/turborepo/actions/runs/15958914313/job/45008819478

chris-olszewski (changes requested):
Just to reduce the overhead of keeping the disallow list up to date as we add fields, I think we should move to an allow list. This will probably mean splitting the `TurboJson` schema to be a root and package one where the package one is only `extends` and `tasks`.

anthonyshew:
Great point. Re-approaching it that way.

crates/turborepo-lib/src/turbo_json/parser.rs
```diff
@@ -282,6 +305,164 @@ impl RawTurboJson {
...
         Self::parse(&json_string, "turbo.json")
     }
+
+    /// Validates field placement to ensure root-only and package-only fields
+    /// are used in the correct configuration types.
+    ///
+    /// This uses an allowlist approach - ALL fields must be explicitly
+    /// categorized. When adding new fields to RawTurboJson, developers MUST
+    /// add them to one of the allowlists below, forcing explicit
+    /// categorization decisions.
```
anthonyshew:
```suggestion
    /// Validates field placement to ensure root-only and package-only fields
    /// are used in the correct configuration types.
    ///
    /// This uses an allowlist approach - ALL fields must be explicitly
    /// categorized.
```

crates/turborepo-lib/src/turbo_json/mod.rs
```diff
@@ -1459,4 +1469,125 @@ mod tests {
...
+                "build": {}
+            }
+        }"#;
+
+        let result = RawTurboJson::parse(json, "turbo.json");
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_allowlist_forces_explicit_categorization() {
```
anthonyshew, line 1527: Not sure if I want this test or not.

anthonyshew:
This really turned out to not be much fun at all. 😂

anthonyshew:
This has turned into needing pre-factors. Putting this in draft while I work on those.

anthonyshew:
We did some other pre-work which invalidates this one!

PR #10611, opened 2025-06-30
fix: filewatching on case-sensitive volumes

anthonyshew:
### Description

https://github.com/vercel/turborepo/issues/9536 says that filewatching is broken on case-sensitive volumes. I confirmed this on my machine.

This PR fixes it by using polling instead of a real filewatcher. I haven't been able to figure out the exact reason it doesn't work with our original implementation, but it does work this way for this case.

### Testing Instructions

Follow the repro instructions in the original issue. You'll see that building this code work on case sensitive volumes.

### DIFF
--- a/crates/turborepo-filewatch/src/fsevent.rs
+++ b/crates/turborepo-filewatch/src/fsevent.rs
@@ -525,7 +525,6 @@ impl FsEventWatcher {
                         cur_runloop,
                         cf::kCFRunLoopDefaultMode,
                     );
-
                     if fs::FSEventStreamStart(stream) == FALSE {
                         panic!("FSEventStream failed to start");
                     }
@@ -537,7 +536,6 @@ impl FsEventWatcher {
                         .expect("Unable to send runloop to watcher");
 
                     cf::CFRunLoopRun();
-
                     fs::FSEventStreamStop(stream);
                     fs::FSEventStreamInvalidate(stream);
                     fs::FSEventStreamRelease(stream);
--- a/crates/turborepo-filewatch/src/lib.rs
+++ b/crates/turborepo-filewatch/src/lib.rs
@@ -534,14 +534,7 @@ async fn wait_for_cookie(
     tracing::debug!("Cookie file created, waiting for filesystem event");
 
     // Dynamic timeout based on filesystem characteristics
-    let timeout_duration = if cookie_dir.to_string().starts_with("/Volumes/") {
-        // External volume detected - use longer timeout
-        tracing::debug!("External volume detected, using extended cookie timeout (6 seconds)");
-        Duration::from_secs(6)
-    } else {
-        // Standard timeout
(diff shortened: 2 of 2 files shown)


crates/turborepo-filewatch/src/fsevent.rs
```diff
@@ -443,19 +457,40 @@ impl FsEventWatcher {
...
+                    self.latency,
+                    self.flags,
+                )
+            }
         };
 
+        if stream.is_null() {
+            return Err(Error::generic("failed to create FSEventStream"));
+        }
+
```
anthonyshew, line 499:
```suggestion
        if stream.is_null() {
            return Err(Error::generic("Failed to create FSEventStream"));
        }

```

crates/turborepo-filewatch/src/fsevent.rs
```diff
@@ -421,10 +422,23 @@ impl FsEventWatcher {
...
         }
+
         let device = self
             .device
             .ok_or_else(|| Error::generic("no device set for stream"))?;
 
+        // Check if this is an external volume (device >= 16777216)
+        // External volumes on macOS often have FSEvents issues, especially
+        // case-sensitive ones
+        let is_external_volume = device >= 16777216;
```
anthonyshew: I don't like this. Trying to find a different way...

crates/turborepo-lib/src/daemon/server.rs
```diff
@@ -112,38 +112,49 @@ impl FileWatching {
...
         let recv = watcher.watch();
 
-        let cookie_writer = CookieWriter::new(
-            watcher.cookie_dir(),
-            Duration::from_millis(100),
-            recv.clone(),
-        );
+        // Increase timeout from 100ms to 2s for better debugging on case-sensitive
+        // volumes
+        let cookie_timeout = Duration::from_millis(2000);
```
anthonyshew: Shouldn't make it into the final code, still trying to figure things out.

crates/turborepo-filewatch/src/lib.rs
```diff
@@ -454,35 +514,80 @@ async fn wait_for_cookie(
...
+
+    std::fs::write(&cookie_path, "turbo")?;
+
+    tracing::debug!("Cookie file created, waiting for filesystem event");
+
+    // Dynamic timeout based on filesystem characteristics
+    let timeout_duration = if cookie_dir.to_string_lossy().starts_with("/Volumes/") {
+        // External volume detected - use longer timeout
+        tracing::debug!("External volume detected, using extended cookie timeout (6 seconds)");
+        Duration::from_secs(6)
```
anthonyshew, line 540: While I was testing, it was possible for an external volume did seem to be slightly slower. 6 is an arbitrary number; we can lower it if we want. I moved it up until I stopped having problems.

anthonyshew:

PR #10620, opened 2025-07-01
refactor(examples): remove duplicate Next.js plugin in examples

JeongHwan-dev:
### Description

Removes duplicate Next.js TypeScript plugin configuration from `examples/.../apps/web/tsconfig.json`.

The `"next"` plugin was configured in two places:
- ✅ `packages/typescript-config/nextjs.json` (shared config)
- ❌ `apps/web/tsconfig.json` (duplicate)

Since the app already extends the shared config, the duplicate plugin definition is unnecessary.

**File:** `examples/.../apps/web/tsconfig.json`

**Removed:**
```json
"compilerOptions": {
  "plugins": [
    {
      "name": "next"
    }
  ]
}
```

The plugin is already inherited from `@repo/typescript-config/nextjs.json`.

### DIFF
--- a/examples/basic/apps/docs/tsconfig.json
+++ b/examples/basic/apps/docs/tsconfig.json
@@ -1,12 +1,5 @@
 {
   "extends": "@repo/typescript-config/nextjs.json",
-  "compilerOptions": {
-    "plugins": [
-      {
-        "name": "next"
-      }
-    ]
-  },
   "include": [
     "**/*.ts",
     "**/*.tsx",
--- a/examples/basic/apps/web/tsconfig.json
+++ b/examples/basic/apps/web/tsconfig.json
@@ -1,12 +1,5 @@
 {
   "extends": "@repo/typescript-config/nextjs.json",
-  "compilerOptions": {
-    "plugins": [
-      {
-        "name": "next"
-      }
-    ]
-  },
   "include": [
     "**/*.ts",
     "**/*.tsx",
--- a/examples/kitchen-sink/apps/storefront/tsconfig.json
+++ b/examples/kitchen-sink/apps/storefront/tsconfig.json
@@ -3,11 +3,6 @@
   "extends": "@repo/typescript-config/nextjs.json",
   "compilerOptions": {
     "outDir": "dist",
-    "plugins": [
-      {
-        "name": "next"
-      }
-    ]
   },
   "include": ["src", "next.config.ts", "next-env.d.ts", ".next/types/**/*.ts"]
 }
--- a/examples/with-berry/apps/docs/tsconfig.json
+++ b/examples/with-berry/apps/docs/tsconfig.json
@@ -1,8 +1,5 @@
 {
   "extends": "@repo/typescript-config/nextjs.json",
-  "compilerOptions": {
-    "plugins": [{ "name": "next" }]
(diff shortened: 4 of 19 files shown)


anthonyshew:
I believe these were there because Next.js used to add them back when you run `next build` or `next dev`. Is this no longer the case?

JeongHwan-dev:
> I believe these were there because Next.js used to add them back when you run `next build` or `next dev`. Is this no longer the case?

@anthonyshew 
Thanks! I double-checked with the latest version of Next.js, and it seems that it no longer adds the "next" plugin automatically to `tsconfig.json` when running next build or next dev.

anthonyshew (approved):
Thanks for looking into it!

anthonyshew (changes requested):
Uh oh, looks like I was too optimistic. I tried to run the `with-tailwind` build locally and it looks like the build fails now with this error:

```
> next build

 ⨯ Failed to load next.config.ts, see more info here https://nextjs.org/docs/messages/next-config-error

> Build error occurred
[TypeError: Cannot read properties of undefined (reading 'baseUrl')]
```

JeongHwan-dev:
> Uh oh, looks like I was too optimistic. I tried to run the `with-tailwind` build locally and it looks like the build fails now with this error:
> 
> ```
> > next build
> 
>  ⨯ Failed to load next.config.ts, see more info here https://nextjs.org/docs/messages/next-config-error
> 
> > Build error occurred
> [TypeError: Cannot read properties of undefined (reading 'baseUrl')]
> ```

@anthonyshew 

Thanks for catching this! I've fixed the issue by adding `baseUrl: "."` to the `compilerOptions` in the `tsconfig.json` files for both the `docs` and `web` apps in the `with-tailwind` example.

The error was occurring because Next.js was trying to access the `baseUrl` property from the TypeScript configuration, but it wasn't defined in the extended config. This should resolve the build failure.

Changes made:
- Added `baseUrl: "."` to `examples/with-tailwind/apps/docs/tsconfig.json`
- Added `baseUrl: "."` to `examples/with-tailwind/apps/web/tsconfig.json`

The build should now complete successfully. ✅

anthonyshew (approved):
I see. `with-tailwind` has that need because it's using `next.config.ts` instead of `.js`. Today, I learned.

Thanks for doing this!

anthonyshew (changes requested):
I was once again too optimistic. My apologies. I don't think we'll be able to merge this because of how Next.js works.

When I run `turbo build` in the `basic` example, these log lines appear:
```
Your tsconfig.json extends another configuration, which means we cannot add the Next.js TypeScript plugin automatically. To improve your development experien
ce, we recommend adding the Next.js plugin (`"plugins": [{ "name": "next" }]`) manually to your TypeScript configuration. Learn more: https://nextjs.org/docs
/app/api-reference/config/typescript#the-typescript-plugin
```

If we remove that, we'll get an issue the next day about that log line. I guarantee it. 😄 Unless there's a way to make it so that we can remove the plugin from the applications' configuration _and_ not see a log about it being missing, I don't think we'll be able to do this.

anthonyshew:
I'll close this for now due to staleness, but let me know if we should re-open because you'd like to work on it again.

PR #10623, opened 2025-07-02
Implement significant hashing performance improvement

anthonyshew:
### Description

Previously, `git_like_hash_file` loaded entire files into memory, which was inefficient and memory-intensive for large files. This change refactors the function to use a streaming approach with a fixed 8KB buffer. This significantly reduces memory consumption and improves performance when hashing large files, mitigating potential out-of-memory issues, especially in CI/CD environments. The git-compatible SHA1 hash format is maintained.

### Testing Instructions

Run the relevant tests to ensure hashing functionality is preserved:
```bash
cargo test --workspace --features=git2 manual
```
All 8 hash-related tests should pass.

### DIFF
--- a/crates/turborepo-scm/src/manual.rs
+++ b/crates/turborepo-scm/src/manual.rs
@@ -15,18 +15,27 @@ use crate::{Error, GitHashes};
 fn git_like_hash_file(path: &AbsoluteSystemPath) -> Result<String, Error> {
     let mut hasher = Sha1::new();
     let mut f = path.open()?;
-    let mut buffer = Vec::new();
-    // Note that read_to_end reads the target if f is a symlink. Currently, this can
-    // happen when we are hashing a specific set of files, which in turn only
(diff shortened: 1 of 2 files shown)


crates/turborepo-scm/src/manual.rs
```diff
@@ -15,18 +15,27 @@ use crate::{Error, GitHashes};
 fn git_like_hash_file(path: &AbsoluteSystemPath) -> Result<String, Error> {
     let mut hasher = Sha1::new();
     let mut f = path.open()?;
-    let mut buffer = Vec::new();
-    // Note that read_to_end reads the target if f is a symlink. Currently, this can
```
chris-olszewski, line 19: Should keep this comment around as well. It still applies even if we use a different method for reading a file.

anthonyshew:
Agree, I'll close until we have real benchmarking that we operate against.

PR #10628, opened 2025-07-03
docs: clarify default inputs behavior

anthonyshew:
### Description

https://github.com/vercel/turborepo/discussions/10606 pointed out a minor clarification that we needed about `inputs`.

### DIFF
--- a/docs/site/content/docs/reference/configuration.mdx
+++ b/docs/site/content/docs/reference/configuration.mdx
@@ -419,7 +419,7 @@ Defines if task outputs should be cached. Setting `cache` to false is useful for
 
 Default: `[]`, all files in the package that are checked into source control
 
-A list of file glob patterns relative to the package's `package.json` to consider when determining if a package has changed. `turbo.json` is **always** considered an input.
+A list of file glob patterns relative to the package's `package.json` to consider when determining if a package has changed. `turbo.json` and `package.json` are **always** considered inputs, even if you try to explicitly ignore them.
 
 Visit the [file glob specification](/docs/reference/globs) for more information on globbing syntax.
 


docs/site/content/docs/reference/configuration.mdx
```diff
@@ -419,7 +419,7 @@ Defines if task outputs should be cached. Setting `cache` to false is useful for
 
 Default: `[]`, all files in the package that are checked into source control
 
-A list of file glob patterns relative to the package's `package.json` to consider when determining if a package has changed. `turbo.json` is **always** considered an input.
+A list of file glob patterns relative to the package's `package.json` to consider when determining if a package has changed. `turbo.json` and `package.json` are **always** considered inputs, even if you try to explicitly ignore them.
```
breadadams: **question:** Are there any other exceptions worth listing, lockfiles maybe?

breadadams: **suggestion:** How about an unordered list here for the list of exceptions?

Thinking that way it would be easier to amend in the future, should there be any new additions.

```suggestion
A list of file glob patterns relative to the package's `package.json` to consider when determining if a package has changed. The following files are **always** considered inputs, even if you try to explicitly ignore them:

- `turbo.json`
- `package.json`
```

Maybe in a-z order?

```suggestion
A list of file glob patterns relative to the package's `package.json` to consider when determining if a package has changed. The following files are **always** considered inputs, even if you try to explicitly ignore them:

- `package.json`
- `turbo.json`
```

anthonyshew: @chris-olszewski, I actually don't know this one. Can I ignore a lockfile? I'm figuring I can't since we go parse it separately from collecting file inputs.

 I shouldn't be able to, but can I? I'm trying to see if I can trick it into doing so and it doesn't look like it.

chris-olszewski: No, we'll always parse the lockfile and fold the packages into a task hash.

anthonyshew: Great suggestion, thanks.

docs/site/content/docs/reference/configuration.mdx
```diff
@@ -419,7 +419,10 @@ Defines if task outputs should be cached. Setting `cache` to false is useful for
 
 Default: `[]`, all files in the package that are checked into source control
 
-A list of file glob patterns relative to the package's `package.json` to consider when determining if a package has changed. `turbo.json` is **always** considered an input.
+A list of file glob patterns relative to the package's `package.json` to consider when determining if a package has changed. The following files are **always** considered inputs, even if you try to explicitly ignore them:
+
+- `package.json`
+- `turbo.json`
```
anthonyshew, line 425:
```suggestion
- `turbo.json`
- Package manager lockfiles
```

PR #10634, opened 2025-07-04
fix: improve prune for yarn 3

anthonyshew:
### Description

Fixes https://github.com/vercel/turborepo/issues/2791

### Testing Instructions

Tests have been added for the previous failing cases

### DIFF
--- a/crates/turborepo-lockfiles/src/berry/mod.rs
+++ b/crates/turborepo-lockfiles/src/berry/mod.rs
@@ -280,12 +280,36 @@ impl BerryLockfile {
             {
                 //  We need to track all of the descriptors coming out the workspace
                 for (name, range) in package.dependencies.iter().flatten() {
-                    let dependency = self.resolve_dependency(locator, name, range.as_ref())?;
+                    // Create the original descriptor before resolution
+                    let mut original_dependency = Descriptor::new(name, range.as_ref())?;
+                    // If there's no protocol we attempt to find a known one
+                    if original_dependency.protocol().is_none() {
+                        if let Some(range) = self.resolver.get(&original_dependency) {
+                            original_dependency.range = range.to_string().into();
+                        }
+                    }
+                    let original_dependency = original_dependency.into_owned();
+
+                    // Apply resolutions to get the resolved dependency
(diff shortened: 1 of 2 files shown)


turborepo-tests/integration/tests/prune/resolutions.t
```diff
@@ -4,13 +4,13 @@ Setup
...
-only have the override that has been set
+We expect both the original and resolved descriptors to be preserved
+when there are yarn resolutions
   $ ${TURBO} prune a
   Generating pruned monorepo for a in .*out (re)
    - Added a
-  $ grep -F '"is-odd@npm:' out/yarn.lock
-  "is-odd@npm:3.0.0":
+  $ grep -F '"is-odd@' out/yarn.lock
+  "is-odd@^0.1.2, is-odd@npm:3.0.0":
```
anthonyshew, line 13: @chris-olszewski I'm pretty dubious of this.

chris-olszewski: This isn't correct. Either:
 - We're now emitting `is-odd@^0.1.2` in `yarn.lock` which is an invalid locator
 - We're not emitting that and this test change will fail

crates/turborepo-lockfiles/src/berry/mod.rs
```diff
@@ -1172,4 +1222,480 @@ mod test {
...
+        // Use the existing robust test data and add our own resolution
+        let data = LockfileData::from_bytes(include_bytes!(
+            "../../fixtures/robust-berry-resolutions.lock"
+        ))
+        .unwrap();
+
+        // Create resolutions that override ajv to use version 8 instead of 6
+        let manifest = BerryManifest {
+            resolutions: Some(
+                [("ajv".to_string(), "^8".to_string())]
```
chris-olszewski, line 1242: We need to actually recreate the fixture lockfile and not just set `resolutions` here.

chris-olszewski (changes requested):
We need some real tests with real lockfiles to verify this change. I'm pretty suspect of it as it currently stands.

anthonyshew:

PR #10640, opened 2025-07-05
fix(turborepo-vt100): prevent panic in col_wrap when drawing_row_mut returns None

anthonyshew:
## Summary
- Replace dangerous `unwrap()` with safe error handling in `Grid::col_wrap` method
- Add comprehensive unit tests to verify the fix and prevent regressions

## Test plan
- [x] Added unit tests that specifically target the edge case where `drawing_row_mut` returns `None`
- [x] Verified all existing tests continue to pass
- [x] Tests confirm no panics occur in edge cases involving terminal scrolling and wrapping

🤖 Generated with [opencode](https://opencode.ai)


anthonyshew:

PR #10651, opened 2025-07-09
test(turborepo-analytics): improve unit test coverage

anthonyshew:
## Summary

Improves test coverage for  from 87.50% to 90.57% (+3.07%) in `turobrepo-analytics`.

## Changes

Added 3 new tests to cover previously untested code paths:

- Tests the  method to ensure it handles timeouts gracefully
- Tests error handling when the analytics client fails to record events  
- Tests the  utility function to ensure it properly sets session IDs

## Coverage Impact

- **Before**: 87.50% coverage (12 missed regions out of 96 total)
- **After**: 90.57% coverage (10 missed regions out of 106 total)
- **Improvement**: +3.07 percentage points

All tests pass and cover important error handling and utility function scenarios.

### DIFF
--- a/crates/turborepo-analytics/src/lib.rs
+++ b/crates/turborepo-analytics/src/lib.rs
@@ -383,39 +383,6 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_close_with_timeout() {
-        let (tx, _rx) = mpsc::unbounded_channel();
-
-        let client = DummyClient {
-            events: Default::default(),
-            tx,
-        };
-
-        let (analytics_sender, analytics_handle) = start_analytics(
-            APIAuth {
-                token: "foo".to_string(),
-                team_id: Some("bar".to_string()),
-                team_slug: None,
-            },
-            client.clone(),
-        );
-
-        // Send an event
-        analytics_sender
-            .send(AnalyticsEvent {
-                session_id: None,
-                source: CacheSource::Local,
-                event: CacheEvent::Hit,
-                hash: "".to_string(),
-                duration: 0,
-            })
-            .unwrap();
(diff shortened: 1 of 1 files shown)


crates/turborepo-analytics/src/lib.rs
```diff
@@ -380,4 +381,115 @@ mod tests {
...
+        // Send an event that will cause the client to error
+        analytics_sender
+            .send(AnalyticsEvent {
+                session_id: None,
+                source: CacheSource::Local,
+                event: CacheEvent::Hit,
+                hash: "".to_string(),
+                duration: 0,
+            })
+            .unwrap();
```
chris-olszewski: Make it clear what is getting tested in this test.
```suggestion
        assert!(analytics_sender
            .send(AnalyticsEvent {
                session_id: None,
                source: CacheSource::Local,
                event: CacheEvent::Hit,
                hash: "".to_string(),
                duration: 0,
            })
            .is_ok(), "sender should swallow client errors");
```

crates/turborepo-analytics/src/lib.rs
```diff
@@ -380,4 +381,115 @@ mod tests {
...
+
+        // Test close_with_timeout - should not panic
+        analytics_handle.close_with_timeout().await;
+    }
```
chris-olszewski: We can drop this test, I'm not sure what this tests that other ones don't already.

chris-olszewski (approved):

crates/turborepo-analytics/src/lib.rs
anthonyshew: SGTM

PR #10653, opened 2025-07-09
docs: fixing --no-cache alert

0ri0nexe:
### Description

Minimal contribution because the alert of --no-cache deprecation gave a wrong command to use now with newer versions

### Testing Instructions

In a project with a recent enough version of turbo such as 2.5.4 :
turbo run build --no-cache

--> You'll receive the alert on the use of --no-cache

### DIFF
--- a/crates/turborepo-lib/src/cli/mod.rs
+++ b/crates/turborepo-lib/src/cli/mod.rs
@@ -399,7 +399,7 @@ impl Args {
             if run_args.no_cache {
                 warn!(
                     "--no-cache is deprecated and will be removed in a future major version. Use \
-                     --cache=local:r,remote:r"
+                     --cache=local:,remote:"
                 );
             }
             if run_args.remote_only.is_some() {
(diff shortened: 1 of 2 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -399,7 +399,7 @@ impl Args {
             if run_args.no_cache {
                 warn!(
                     "--no-cache is deprecated and will be removed in a future major version. Use \
-                     --cache=local:r,remote:r"
```
anthonyshew: This is the correct invocation to recommend. `--no-cache` would prevent writing, not reading.

This confusion is part of the reason we changed the API.

anthonyshew:
Hey, @0ri0nexe, thanks for this! We'll need to change the recommendation in the CLI back to how it was, but happy to merge in the docs update.

0ri0nexe:
My bad i really thought --no-cache didn't make any cache reading

Issue #10642, opened 2025-07-07
Tailwind Specificity Conflicts


JamesGriz:
### Verify canary release

- [x] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

https://github.com/JamesGriz/specificity-bug

### Which canary version will you have in your reproduction?

turbo 2.5.4

### Environment information

```block
CLI:
   Version: 2.5.4
   Path to executable: /Users/jamesgray/Documents/projects/ppm/specificity-bug/node_modules/.pnpm/turbo-darwin-arm64@2.5.4/node_modules/turbo-darwin-arm64/bin/turbo
   Daemon status: Running
   Package manager: pnpm

Platform:
   Architecture: aarch64
   Operating system: macos
   WSL: false
   Available memory (MB): 4898
   Available CPU cores: 11

Environment:
   CI: None
   Terminal (TERM): xterm-256color
   Terminal program (TERM_PROGRAM): WarpTerminal
   Terminal program version (TERM_PROGRAM_VERSION): v0.2025.06.25.08.12.stable_01
   Shell (SHELL): /bin/zsh
   stdin: false
```

### Expected behavior

All tailwind utility classes to apply weather they are being used in the component ui library or the web app/docs.

### Actual behavior

When defining tailwind utility classes in the web app eg `lg:hidden` and there is a utility class being used in the component library which sharing the same style property eg `flex` (the shared property being display in this case) the `lg:hidden` style is losing to flex in terms of specificity.  

### To Reproduce

- Check out [the reproduction repo ](https://github.com/JamesGriz/specificity-bug)and run the example repo
- Notice on http://localhost:3001/ when the viewport is larger that tailwinds lg breakpoint that the code block with `flex border lg:hidden` is not hidden.

### Additional context

I noticed that this issue was opened previously [here](https://github.com/vercel/turborepo/issues/10538) but was then closed thanks to [this PR](https://github.com/vercel/turborepo/pull/10561). The fix just disregarded the component UI styles and seems this was undone on a more recent commit [here](https://github.com/okorion/turborepo/commit/e830ceabeaccf3b2aa2d4292be513a32a4ddbd8f)

So the original issue still stands.

anthonyshew:
Is this an issue about the example being incorrect or Tailwind's core behavior?

JamesGriz:
> Is this an issue about the example being incorrect or Tailwind's core behavior?

[...]

PR #10657, opened 2025-07-11
fix(examples): css specificity in Tailwind example

anthonyshew:
### Description

As described in #10642, CSS specificity wasn't being handled correctly in the example. This PR fixes it by ensuring the compiled stylesheet from the package:
- Uses a prefix. This fixes all specificity issues by itself, but also...
- Is used directly in the app instead of trying to go through the app's Tailwind compilation process
- Is applied _before_ the app's stylesheet, so that things can be properly overridden by the app (since they're later in the CSS evaluation)

### Testing Instructions

Try it out!

### DIFF
--- a/crates/turborepo-analytics/src/lib.rs
+++ b/crates/turborepo-analytics/src/lib.rs
@@ -190,9 +190,8 @@ mod tests {
     };
     use turborepo_api_client::{analytics::AnalyticsClient, APIAuth};
     use turborepo_vercel_api::{AnalyticsEvent, CacheEvent, CacheSource};
-    use uuid::Uuid;
 
-    use crate::{add_session_id, start_analytics};
+    use crate::start_analytics;
 
     #[derive(Clone)]
     struct DummyClient {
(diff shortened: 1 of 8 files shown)


tknickman:
We used to use a [prefix](https://tailwindcss.com/docs/upgrade-guide#using-a-prefix) to avoid this but it looks like that was changed?

crates/turborepo-analytics/src/lib.rs
```diff
@@ -380,4 +381,115 @@ mod tests {
         let payloads = &found[0];
         assert_eq!(payloads.len(), 2);
     }
+
```
chris-olszewski: I don't think this is necessary?

chris-olszewski (changes requested):
LGTM except for Rust changes

anthonyshew:
@tknickman Sigh, they added [docs](https://tailwindcss.com/docs/upgrade-guide#using-a-prefix) for how to use prefixes now. I've looked at least 10 times over the course of the past N months for this and figured it just wasn't a thing. 🤦 Circling back...

anthonyshew:
@tknickman There we are. Thank goodness they brought it back (or made it discoverable).

chris-olszewski (approved):

PR #10676, opened 2025-07-15
chore(tower-uds): delete unused crate

anthonyshew:


crates/tower-uds/src/lib.rs
```diff
@@ -59,3 +59,151 @@ impl<'a, T> Service<T> for UDSConnector<'a> {
...
+
+        // Create a UDS server using the Windows implementation
+        let listener =
+            uds_windows::UnixListener::bind(&socket_path).expect("Failed to bind socket");
+        let async_listener =
+            async_io::Async::new(listener).expect("Failed to create async listener");
+
+        // Spawn server task
+        let server_handle = tokio::spawn(async move {
+            if let Ok((stream, _)) = async_listener.accept().await {
```
chris-olszewski:
```suggestion
            if let Ok((stream, _)) = async_listener.await.accept().await {
```

crates/tower-uds/Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "MPL-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-tokio = { workspace = true, features = ["net"] }
+tokio = { workspace = true, features = ["net", "macros", "rt", "io-util"] }
```
chris-olszewski: Are these required? I'm a little concerned about adding these features just for testing since they're invasive and will end up in the binary.

chris-olszewski (changes requested):

crates/tower-uds/Cargo.toml
anthonyshew: Ah, you right. I moved them to dev-dependencies. Is that sufficient?

PR #10677, opened 2025-07-15
refactor: move config file discovery to shared util

anthonyshew:
### Description

https://github.com/vercel/turborepo/pull/10619 put this logic in a shared place, but looks like I missed this spot.

### Testing Instructions

This should be a pure refactor, so CI should catch regressions.

### DIFF
--- a/crates/turborepo-lib/src/run/watch.rs
+++ b/crates/turborepo-lib/src/run/watch.rs
@@ -19,7 +19,7 @@ use crate::{
     daemon::{proto, DaemonConnectorError, DaemonError},
     get_version, opts,
     run::{self, builder::RunBuilder, scope::target_selector::InvalidSelectorError, Run},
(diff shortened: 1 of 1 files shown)


crates/turborepo-lib/src/run/watch.rs
```diff
@@ -120,14 +120,10 @@ impl WatchClient {
         let signal = get_signal()?;
         let handler = SignalHandler::new(signal);
 
-        // Check if the turbo.json path is the standard one (either turbo.json or
-        // turbo.jsonc)
-        let standard_path_json = base.repo_root.join_component(CONFIG_FILE);
-        let standard_path_jsonc = base.repo_root.join_component(CONFIG_FILE_JSONC);
+        // Check if the turbo.json path is the standard one
```
anthonyshew:
```suggestion
```

chris-olszewski (approved):

PR #10682, opened 2025-07-16
feat: improve parsing error

anthonyshew:
### Description

[tkoehlerlg](https://github.com/tkoehlerlg) showed [here](https://github.com/vercel/turborepo/issues/3970#issuecomment-3075669651) that this error wasn't helpful enough. This PR makes it so this situation returns the file path and extra clarity on why the warning is occurring.

### Testing Instructions

The snapshot shows the improvement.

### DIFF
--- a/crates/turborepo-repository/src/package_graph/builder.rs
+++ b/crates/turborepo-repository/src/package_graph/builder.rs
@@ -444,9 +444,11 @@ impl<'a, T: PackageDiscovery> BuildState<'a, ResolvedWorkspaces, T> {
         let lockfile = match self.populate_lockfile().await {
             Ok(lockfile) => Some(lockfile),
             Err(e) => {
+                let lockfile_path = package_manager.lockfile_path(self.repo_root);
                 warn!(
-                    "Issues occurred when constructing package graph. Turbo will function, but \
-                     some features may not be available:\n {:?}",
+                    "An issue occurred while attempting to parse {}. Turborepo will still \
+                     function, but some features may not be available:\n {:?}",
+                    lockfile_path,
                     Report::new(e)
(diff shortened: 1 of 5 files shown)


crates/turborepo-repository/src/package_graph/builder.rs
```diff
@@ -59,6 +59,21 @@ pub enum Error {
...
+/// Attempts to extract the file path that caused the error from the error chain
+/// Falls back to the lockfile path if no specific file can be determined
+fn extract_file_path_from_error(
+    error: &Error,
+    package_manager: &crate::package_manager::PackageManager,
+    repo_root: &AbsoluteSystemPath,
+) -> AbsoluteSystemPathBuf {
+    match error {
+        Error::PackageJsonMissingName(path) => path.clone(),
+        // TODO: It's possible this could be returning the wrong path in some situations since its
```
anthonyshew:
```suggestion
        // TODO: We're handling every other error here. We could handle situations where the lockfile isn't the issue better.
```

crates/turborepo-repository/src/package_graph/builder.rs
```diff
@@ -651,4 +670,29 @@ mod test {
         }));
         assert_matches!(builder.build().await, Err(Error::DuplicateWorkspace { .. }));
     }
+
+    #[test]
+    fn test_missing_name_field_warning_message() {
```
chris-olszewski, line 676: Okay to skip running this on Windows to avoid needing to deal with Windows paths. Only okay since we're testing string formatting
```suggestion
    #[cfg(unix)]
    fn test_missing_name_field_warning_message() {
```

PR #10687, opened 2025-07-17
feat: allow Watch Mode with `--root-turbo-json`

anthonyshew:
### Description

Unblock Watch Mode from being used with `--root-turbo-json`. 

### Testing Instructions

I tested this manually by using `turbo watch lint --root-turbo-json=./some-dir/turbo.json` in a `create-turbo` and changing that file. The tasks would be picked up and re-ran the same as if I omitted the `--root-turbo-json` and used a standard path.

I also tried it with a JSONC and it worked.

### DIFF
--- a/crates/turborepo-lib/src/commands/daemon.rs
+++ b/crates/turborepo-lib/src/commands/daemon.rs
@@ -25,7 +25,11 @@ const DAEMON_NOT_RUNNING_MESSAGE: &str =
     "daemon is not running, run `turbo daemon start` to start it";
 
 /// Runs the daemon command.
-pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Result<(), DaemonError> {
+pub async fn daemon_client(
+    command: &DaemonCommand,
+    base: &CommandBase,
+    custom_turbo_json_path: Option<&camino::Utf8Path>,
+) -> Result<(), DaemonError> {
     let (can_start_server, can_kill_server) = match command {
         DaemonCommand::Status { .. } | DaemonCommand::Logs => (false, false),
         DaemonCommand::Stop => (false, true),
@@ -33,7 +37,26 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
         DaemonCommand::Clean { .. } => (false, true),
     };
 
-    let connector = DaemonConnector::new(can_start_server, can_kill_server, &base.repo_root);
+    let custom_turbo_json_path = match custom_turbo_json_path {
+        Some(p) => match turbopath::AbsoluteSystemPathBuf::from_cwd(p) {
+            Ok(path) => Some(path),
+            Err(e) => {
+                tracing::error!("Failed to convert custom turbo.json path: {}", e);
+                return Err(DaemonError::Unavailable(format!(
+                    "Invalid turbo.json path: {}",
+                    e
+                )));
+            }
+        },
+        None => None,
+    };
+
+    let connector = DaemonConnector::new(
+        can_start_server,
+        can_kill_server,
+        &base.repo_root,
+        custom_turbo_json_path,
+    );
 
     match command {
         DaemonCommand::Restart => {
--- a/crates/turborepo-lib/src/commands/daemon.rs
+++ b/crates/turborepo-lib/src/commands/daemon.rs
@@ -24,6 +24,25 @@ use crate::{
 const DAEMON_NOT_RUNNING_MESSAGE: &str =
     "daemon is not running, run `turbo daemon start` to start it";
 
+/// Converts an optional turbo.json path to an absolute system path.
+fn convert_turbo_json_path(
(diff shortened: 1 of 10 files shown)


crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -139,10 +148,8 @@ impl DaemonConnector {
         ))
     }
 
-    /// Gets the PID of the daemon process.
-    ///
-    /// If a daemon is not running, it starts one.
-    async fn get_or_start_daemon(&self) -> Result<sysinfo::Pid, DaemonConnectorError> {
+    /// Check if a daemon is currently running and return its PID
+    async fn existing_daemon(&self) -> Result<sysinfo::Pid, DaemonConnectorError> {
```
anthonyshew: Not remembering why I did this...

anthonyshew: Looks to work correctly without this.

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -66,22 +66,31 @@ pub struct DaemonConnector {
...
+    /// Optional custom turbo.json path to watch
+    pub custom_turbo_json_path: Option<turbopath::AbsoluteSystemPathBuf>,
 }
 
 impl DaemonConnector {
     pub fn new(
         can_start_server: bool,
         can_kill_server: bool,
         repo_root: &AbsoluteSystemPath,
+        custom_turbo_json_path: Option<&turbopath::AbsoluteSystemPathBuf>,
```
chris-olszewski: We do not need to pass this in if we also have a builder style method to set this.
```suggestion
```

crates/turborepo-lib/src/daemon/connector.rs
```diff
@@ -66,22 +66,31 @@ pub struct DaemonConnector {
...
     ) -> Self {
         let paths = Paths::from_repo_root(repo_root);
         Self {
             can_start_server,
             can_kill_server,
             paths,
+            custom_turbo_json_path: custom_turbo_json_path.cloned(),
```
chris-olszewski:
```suggestion
            custom_turbo_json_path: None,
```

crates/turborepo-lib/src/commands/daemon.rs
```diff
@@ -33,7 +33,7 @@ pub async fn daemon_client(command: &DaemonCommand, base: &CommandBase) -> Resul
         DaemonCommand::Clean { .. } => (false, true),
     };
 
-    let connector = DaemonConnector::new(can_start_server, can_kill_server, &base.repo_root);
+    let connector = DaemonConnector::new(can_start_server, can_kill_server, &base.repo_root, None);
```
chris-olszewski: I think we need to pass in the the turbo root here in case we're (re)starting the daemon.

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -260,8 +286,19 @@ impl Subscriber {
...
+                                        if path_str == custom_path.as_str()
+                                            || path_str.ends_with("turbo.json")
+                                            || path_str.ends_with("turbo.jsonc")
+                                        {
+                                            tracing::info!(
+                                                "File watcher detected change: {}",
+                                                path_str
+                                            );
+                                        }
+                                    }
```
chris-olszewski: No idea why this logic was added
```suggestion
```

crates/turborepo-lib/src/package_changes_watcher.rs
```diff
@@ -353,9 +390,33 @@ impl Subscriber {
...
+                        if !found {
+                            // Also try with the path keys for debugging
+                            let trie_keys: Vec<_> = trie.keys().take(10).collect();
+                            tracing::debug!(
+                                "Trie contains these paths (first 10): {:?}",
+                                trie_keys
+                            );
+                        }
+                        found
+                    })
```
chris-olszewski:
```suggestion
                        trie.get(path_str).is_some()
```

crates/turborepo-lib/src/run/builder.rs
```diff
@@ -263,7 +263,7 @@ impl RunBuilder {
                 let can_start_server = true;
                 let can_kill_server = true;
                 let connector =
-                    DaemonConnector::new(can_start_server, can_kill_server, &self.repo_root);
+                    DaemonConnector::new(can_start_server, can_kill_server, &self.repo_root, None);
```
chris-olszewski: I think we want to set the custom here.
```suggestion
                    DaemonConnector::new(can_start_server, can_kill_server, &self.repo_root);
```

anthonyshew:
Thanks for the review, @chris-olszewski. Should have all those noob mistakes buttoned up and chose to go with:
> warn users that turbo.json outside of the repository will not have changes picked up

chris-olszewski (approved):

PR #10702, opened 2025-07-21
docs: add architecture document for `turbo run`

chris-olszewski:
### Description

Add a high level view of what happens during `turbo run` and the structures involved

### Testing Instructions

👀


anthonyshew (approved):
🔥

PR #10717, opened 2025-07-29
docs: adjust typo on instructions for eslint-config-turbo legacy config

evsasse:
### Description

Seems like this last example was copied from https://github.com/vercel/turborepo/blob/main/packages/eslint-plugin-turbo/README.md, but forgot this small adjustment for the legacy config.

### Testing Instructions

Configuring a legacy `eslintrc*` file, inheriting the default `eslint-config-turbo` rules, and customizing rules on top of that should work by copying the example.


anthonyshew (approved):
Thanks!

PR #10720, opened 2025-07-30
feat(create-turbo): `--no-git` flag

anthonyshew:
### Description

As an edge case, some users don't want Git to be initialized when they set up a Turborepo. This flag removes `.git` when we're done with our work so the user can handle it themselves.

### Testing Instructions

Tests have been added and tried it manually.


AshConnolly:
Hi @anthonyshew, thanks for this! 👍 Just to check this doesn't mean that it will delete a preexisting .git folder after create-turbo is done? Cheers!

anthonyshew:
Yes, I think that's a reasonable expectation. The CLI initializes the folder that it dumps files into, so it owns that directory's fate.

AshConnolly:
> Yes, I think that's a reasonable expectation. The CLI initializes the folder that it dumps files into, so it owns that directory's fate.

Thanks for the reply! 

Sorry does that mean it _would_ or _wouldn't_ delete a pre-existing `.git` folder? 
Just checking that if you are stopping turbo doing any git commits, it doesn't delete a pre-existing `.git`. folder 😃

anthonyshew:
`create-turbo` will only delete the `.git` that it created for itself.

packages/create-turbo/src/commands/create/index.ts
```diff
@@ -307,5 +312,10 @@ export async function create(
     logger.log("- Run a command twice to hit cache");
   }
 
+  // remove .git directory if --no-git flag is used
+  if (noGit) {
+    removeGitDirectory(root);
```
chris-olszewski: Maybe warn if we failed to remove the git directory?
```suggestion
    if (!removeGitDirectory(root)) {
        logger.warn("Failed to remove '.git' directory")
    }
```

packages/create-turbo/__tests__/git.test.ts
```diff
@@ -243,4 +245,41 @@ describe("git", () => {
...
     });
   });
+
+  describe("removeGitDirectory", () => {
+    const { useFixture } = setupTestFixtures({
+      directory: path.join(__dirname, "../"),
+      options: { emptyFixture: true },
+    });
+
+    it("successfully removes .git directory", async () => {
```
chris-olszewski: Just a nit since we're not actually hitting FS here
```suggestion
    it("attempts to remove .git directory", async () => {
```

packages/create-turbo/src/commands/create/index.ts
anthonyshew: True and real.

PR #10727, opened 2025-08-02
docs: fix JSDoc type for ESLint config in basic example

victor-code19:
This PR fixes a minor inaccuracy in the JSDoc type for the ESLint configuration, ensuring it correctly reflects that an array is exported.


anthonyshew (approved):
I added the same fix for the docs app. Thank you!

PR #10740, opened 2025-08-06
feat: add GitHub Actions environment variable as default passthru

louis-bompart:
### Description

Allowing those variables to pass through by default would make using turborepo with GitHub Actions more seamless.

For example, NPM relies on `GITHUB_ACTIONS` for automatic provenance detection. (see https://github.com/npm/cli/blob/5d82d0b4a4bd1424031fb68b4df740c1bbe5b172/lib/utils/oidc.js#L11) By obfuscating the variable, it can be pretty confusing for users of both turborepo and npm as to why the provenance generation stopped working (don't ask me how I know).

### DIFF
--- a/crates/turborepo-lib/src/task_hash.rs
+++ b/crates/turborepo-lib/src/task_hash.rs
@@ -474,6 +474,9 @@ impl<'a> TaskHasher<'a> {
                     "NEXT_*",
                     "USE_OUTPUT_FOR_EDGE_FUNCTIONS",
                     "NOW_BUILDER",
+                    // GitHub Actions - https://docs.github.com/en/actions/reference/workflows-and-actions/variables
+                    "GITHUB_*",
+                    "RUNNER_*",
                     // Command Prompt casing of env variables
                     "APPDATA",
                     "PATH",


louis-bompart:
A bit more esoteric but related to the same initial problem (npm publish with provenance fails if executed through turborepo), I think it might be good to add `ACTIONS_ID_TOKEN_REQUEST_TOKEN` and `ACTIONS_ID_TOKEN_REQUEST_URL` to the passthru which is used for OIDC (and by extension npm publishing w/ provenance) with GitHub Actions:
https://docs.github.com/en/actions/reference/security/oidc

anthonyshew (approved):
Thank you. 👍

charpeni:
@anthonyshew: That would be amazing if we could get this merged. I believe this is a blocker for adoption, or at least, a friction, as most nice CI features aren't working by default when using Turborepo because the `GITHUB_ACTIONS` environment variable isn't exposed. E.g., [GitHub Annotations from test runners](https://vitest.dev/guide/test-annotations.html#github-actions).

anthonyshew:
Failures for examples are not related, so merging.

PR #10749, opened 2025-08-10
chore: update to Rust 1.89.0

ognevny:
### Description

- fix mismatched_lifetime_syntaxes lints
- let chains are now stable in edition 2024, so use them
- change workspace logic so "2024" edition is now applied for all package with `workspace = true`

### Testing Instructions

CI

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,9 @@ members = [
   "packages/turbo-repository/rust",
 ]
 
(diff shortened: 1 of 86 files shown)


ognevny:
I've fixed let chain in turborepo-vt100, should pass now

ognevny:
forgot to rerun `cargo fmt`. done

PR #10762, opened 2025-08-12
fix: allow dependsOn to reference package.json scripts in non-root packages

Yuvraj-cyborg:
### Problem: 
- dependsOn failed when referencing pkg#task where task exists in the package’s package.json but not in its turbo.json. Error: Could not find "pkg#task" in root turbo.json or "task" in package.

### Fix: 
- When resolving a non-root pkg#task that has a matching script in that package’s package.json, synthesize a default task definition so the dependency is valid without a workspace turbo.json entry. Root tasks remain required in root turbo.json.
- Repro: https://github.com/kalkulacka-one/kalkulacka/tree/turbo-bug (2.5.4-canary.0)
- With this change, schema-web#build:openapi is recognized and executed. Verified via dry run and actual command resolution.

### Tests: 
- Manual verification with the linked repo. Existing tests around task resolution still pass locally.

### DIFF
--- a/crates/turborepo-lib/src/engine/builder.rs
+++ b/crates/turborepo-lib/src/engine/builder.rs
@@ -628,15 +628,35 @@ impl<'a> EngineBuilder<'a> {
         }
 
         if task_definitions.is_empty() && self.should_validate_engine {
-            let (span, text) = task_id.span_and_text("turbo.json");
-            return Err(Error::MissingPackageTask(Box::new(
-                MissingPackageTaskError {
-                    span,
-                    text,
-                    task_id: task_id.to_string(),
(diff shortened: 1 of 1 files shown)


chris-olszewski (changes requested):
This is an intentional behavior of Turborepo to prevent unintentional misuse from incorrect task definitions.

Yuvraj-cyborg:
> This is an intentional behavior of Turborepo to prevent unintentional misuse from incorrect task definitions.

should i find an other way around? for this issue then ?

anthonyshew:
Like @chris-olszewski said, this is intentional behavior.

PR #10763, opened 2025-08-13
feat(turbo_json): add `$TURBO_EXTENDS$`

chris-olszewski:
### Description

This PR adds a new DSL keyword for array task definitions that alters the behavior when using `extends` in a package `turbo.json`. Usually a field from a package task definition will completely override the root field e.g.
Root: `{ "env": ["FOO"] }`
Package: `{ "env": ["BAR"] }`
Result: `{"env": ["BAR"] }`

If the package uses the new `$TURBO_EXTENDS$` the root field value will be extended from instead of overridden. e.g.
Root: `{ "env": ["FOO"] }`
Package: `{ "env": ["$TURBO_EXTENDS$", "BAR"] }`
Result: `{"env": ["FOO", "BAR"] }`

This functionality allows for greater flexibility when specializing package task definitions without needing to copy/paste the shared values and remembering to update them if the root changed.

It is currently gated by `turboExtends` feature flag in the root `turbo.json`.

Reviewing this each commit individually in this PR is *highly* suggested. The actual feature add is all done in the final commit, the rest are prefactors.

The primary prefactors are:
 - Plumbing through future flags to be read at config time, but usable at task resolution time. This is necessary since this is set at root and needs to be known when reading task definitions from package `turbo.json`s
 - Adding a new intermediate step for task definition creation named "processed". Removes a majority of the DSL concepts and validations that are exclusive to a single field. This paves the way for future extensions to the DSL.

Nerd note:
Technically, we could extend `$TURBO_EXTENDS$` to additional fields that are monoids. Extension behavior is really:
```
let a = if extends {
  base
} else {
  mempty
}
a mappend b
```
The only interesting addition we could make would be around boolean fields which has 2 different monoids: any and all. Which one to choose would depend on the default value of the flag. e.g. `cache` defaults to false so we would want an any (or) behavior. I am not suggesting we do this, it is more confusing than it is worthwhile.

### Testing Instructions

Added a fair amount of unit tests for the whole task definition resolution steps.

### DIFF
--- a/crates/turborepo-lib/src/turbo_json/future_flags.rs
+++ b/crates/turborepo-lib/src/turbo_json/future_flags.rs
@@ -34,7 +34,7 @@ pub struct FutureFlags {
     /// When enabled, allows using `$TURBO_EXTENDS$` in array fields.
     /// This will change the default behavior of overriding the field to instead
     /// append.
-    pub turbo_extends: bool,
+    pub turbo_extends_keyword: bool,
 }
 
 impl FutureFlags {
--- a/crates/turborepo-lib/src/turbo_json/future_flags.rs
+++ b/crates/turborepo-lib/src/turbo_json/future_flags.rs
@@ -1,0 +1,45 @@
+//! Future flags for enabling experimental or upcoming features
+//!
+//! This module contains the `FutureFlags` structure which allows users to
+//! opt-in to experimental features before they become the default behavior.
+//!
+//! ## Usage
+//!
+//! Future flags can be configured in the root `turbo.json`:
+//!
+//! ```json
+//! {
+//!   "futureFlags": {
+//!     "turboExtends": true
+//!   }
+//! }
+//! ```
+//!
+//! Note: Future flags are only allowed in the root `turbo.json` and will cause
+//! an error if specified in workspace packages.
+
+use biome_deserialize_macros::Deserializable;
+use serde::Serialize;
+use struct_iterable::Iterable;
+
+/// Future flags configuration for experimental features
+///
+/// Each flag represents an experimental feature that can be enabled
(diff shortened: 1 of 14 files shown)


turborepo-tests/integration/fixtures/turbo-configs/interruptible-but-not-persistent.json
```diff
@@ -10,7 +10,7 @@
     "build": {
       "env": [
         "NODE_ENV",
-        "$FOOBAR"
+        "FOOBAR"
```
chris-olszewski, line 13: We changed validation order (both are still checked) and this `turbo.json` was invalid in 2 ways. Previously we were depending on the ordering of validations to see the expected error message around this task being interruptible, but not persistent.

anthonyshew (approved):

crates/turborepo-lib/src/turbo_json/future_flags.rs
```diff
@@ -0,0 +1,45 @@
...
+/// before it becomes the default behavior in a future version.
+#[derive(Serialize, Default, Debug, Copy, Clone, Iterable, Deserializable, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct FutureFlags {
+    /// Enable `$TURBO_EXTENDS$`
+    ///
+    /// When enabled, allows using `$TURBO_EXTENDS$` in array fields.
+    /// This will change the default behavior of overriding the field to instead
+    /// append.
+    pub turbo_extends: bool,
```
anthonyshew: Non-blocking: Let's be slightly more descriptive on this flag name since we're about to have `$TURBO_EXTENDS$` and, separately, make the `extends` key more powerful. I could see a user being confused about which this refers to.

PR #10770, opened 2025-08-16
docs(github-action): fix set team url to team id

mengxi-ream:
### Description

Shoud use `TEAM ID` as `TURBO_TEAM` rather than `TEAM URL`

### Testing Instructions

Test on Github Action

### DIFF
--- a/docs/site/content/docs/guides/ci-vendors/github-actions.mdx
+++ b/docs/site/content/docs/guides/ci-vendors/github-actions.mdx
@@ -257,7 +257,7 @@ Go to your GitHub repository settings and click on the **Secrets** and then **Ac
 </Step>
 
 <Step>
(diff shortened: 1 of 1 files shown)


anthonyshew:
Hi, this doc is correct. You need to use the team slug for this variable, which is consistent with other documentation ([example here](https://turborepo.com/docs/guides/ci-vendors#general-setup)). Are you seeing something different?

mengxi-ream:
> Hi, this doc is correct. You need to use the team slug for this variable, which is consistent with other documentation ([example here](https://turborepo.com/docs/guides/ci-vendors#general-setup)). Are you seeing something different?

I used TURBO_TEAM team slug vercel.com/mengxi-ream and can't get the cache in GitHub Action (error is permission related) then replace it as team id, cache worked in GitHub Action

for team slug, should we include Vercel.com or not?

anthonyshew:
The slug is only `mengxi-ream`.

mengxi-ream:
> mengxi-ream

thanks, I'll try

anthonyshew:
I'll close this up as I've tested this multiple times in the past, and believe the doc to be correct as-is. Thanks for the PR, really appreciate the collaboration!

PR #10786, opened 2025-08-21
chore: fix lints

chris-olszewski:
### Description

Noticed a few lints from the recent edition bump

### Testing Instructions

👀

### DIFF
--- a/crates/turborepo-ci/src/vendor_behavior.rs
+++ b/crates/turborepo-ci/src/vendor_behavior.rs
@@ -5,7 +5,7 @@ use chrono::{DateTime, Utc};
 pub type GroupPrefixFn = Arc<dyn Fn(DateTime<Utc>) -> String + Send + Sync>;
 type GroupPrefixFnFactory = fn(group_name: String) -> GroupPrefixFn;
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug)]
 pub struct VendorBehavior {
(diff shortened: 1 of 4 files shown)


crates/turborepo-wax/src/lib.rs
```diff
@@ -12,6 +12,7 @@
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/olson-sean-k/wax/master/doc/wax.svg?sanitize=true"
 )]
+#![allow(dead_code)]
```
chris-olszewski, line 15: Doing this instead of fixing the lint since this is a vendored library

crates/turborepo-ci/src/vendor_behavior.rs
```diff
@@ -5,7 +5,7 @@ use chrono::{DateTime, Utc};
 pub type GroupPrefixFn = Arc<dyn Fn(DateTime<Utc>) -> String + Send + Sync>;
 type GroupPrefixFnFactory = fn(group_name: String) -> GroupPrefixFn;
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug)]
```
chris-olszewski, line 8: Just because this diff doesn't make it clear. The issue here is that `PartialEq` on function pointers has nuance. Removing equality implementation since we don't use it.

See https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html for more info

PR #10794, opened 2025-08-25
ci: checks for `eslint-config-turbo` interface

anthonyshew:
### Description

Because JavaScript packaging is so finnicky, I'm thinking its a good idea to make sure our interfaces stay ecosystem compliant. PRs like https://github.com/vercel/turborepo/pull/10747 and https://github.com/vercel/turborepo/pull/10754 have potential to break things, so its nice to have some sanity checking. (I haven't checked yet if those PRs specifically are right, but they made me think about adding some tooling like this.)

### Testing Instructions

CI

### DIFF
--- a/packages/eslint-config-turbo/src/index.ts
+++ b/packages/eslint-config-turbo/src/index.ts
@@ -2,4 +2,5 @@ const config = {
   extends: ["plugin:turbo/recommended"],
 };
 
-module.exports = config;
+// eslint-disable-next-line import/no-default-export -- ESLint convention is a default export
+export default config;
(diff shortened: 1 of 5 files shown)


packages/eslint-config-turbo/src/index.ts
```diff
@@ -2,4 +2,5 @@ const config = {
   extends: ["plugin:turbo/recommended"],
 };
 
-module.exports = config;
+// eslint-disable-next-line import/no-default-export -- ESLint convention is a default export
+export default config;
```
anthonyshew, line 6: bunchee wasn't converting `module.exports` to `export default` for the ESM entrypoints. If I make it `export default`, it downlevels to CJS correctly.

PR #10796, opened 2025-08-25
cache: warn once and disable remote cache on connection error; docs: …

DipakHalkude:
…add guidance on unreachable remote cache


DipakHalkude:
Adds one-time error when remote cache is unreachable; disables further remote attempts for the run.
Updates caching and run docs to explain behavior and fallback.
Scope limited to turborepo-cache, turborepo-lib docs references.

docs/site/content/docs/reference/run.mdx
```diff
@@ -531,6 +531,12 @@ Ignore the local filesystem cache for all tasks, using Remote Cache for reading
 turbo run build --remote-only
 ```
 
+<Callout type="info" title="Remote cache unreachable">
+  When Remote Cache is configured but unreachable (network error, misconfiguration, or service
+  outage), Turborepo will print a one-time error early in the run and fall back to local cache if
+  enabled. This prevents long runs where each workspace silently reports a cache miss.
+</Callout>
+
```
anthonyshew, line 539:
```suggestion
```

docs/site/content/docs/crafting-your-repository/caching.mdx
```diff
@@ -220,6 +220,16 @@ It's possible to create scenarios where caching ends up being slower than not ca
...
+### Remote cache unreachable
+
+If Remote Caching is configured but cannot be reached (for example due to network errors, incorrect configuration, or the remote service being down), Turborepo now reports a one-time error early in the run:
+
+```text
+Cannot access remote cache (connection failed). Falling back to local cache if available.
+```
+
+After this, Turborepo disables further remote cache attempts for the current run and continues using the local cache when enabled. This avoids long runs where every workspace silently reports a cache miss.
+
```
anthonyshew, line 232:
```suggestion
```

anthonyshew:
Hi, @DipakHalkude, can you update the description of this PR with information about your motivation for changing this code, and perhaps a before and after of the effects of this change? Thanks.

DipakHalkude:
Hi @anthonyshew, thanks for the feedback.

Motivation
The main motivation for this change is to improve developer experience when the remote cache is misconfigured or temporarily unreachable. Currently, each workspace silently reports cache misses, which can make runs unnecessarily long and misleading. Developers may not realize the issue is with the remote cache rather than the tasks themselves.

Before

Remote cache connection errors were retried for each workspace.

Every workspace would report a cache miss without a clear explanation.

Runs could take significantly longer due to repeated failed attempts.

After

Turborepo prints a one-time error message early in the run when the remote cache cannot be reached.

Remote cache attempts are disabled for the remainder of that run.

Execution continues using the local cache (if enabled), avoiding repeated retries and improving clarity.

This ensures developers get immediate visibility into why remote caching isn’t working and can still rely on local caching without the noise of repeated misses.

anthonyshew:
In that case, I'm not sure that the messages are clear enough. They don't mention anything about the what, why, or how of the failures. They only say that there was failure. If I were seeing these errors for the first time, I would be alarmed, and want to fix them, but wouldn't have the details I need to get from the bad state to the good state.

How do you think this should be improved?

DipakHalkude:
Thanks for pointing this out @anthonyshew 🙏

I agree that the current messages could feel a bit abrupt and don’t give enough actionable detail. I can expand them to include:

What happened (remote cache connection failed).

Why it happened (possible causes like network error, misconfiguration, or outage).

How to fix it (check remote cache URL, credentials, or service status).

What Turborepo will do (disable remote cache for the run and continue with local cache if available).

Something along the lines of:

⚠️ Remote cache unreachable: Turborepo could not connect to the configured remote cache.
This may be due to a network issue, invalid configuration, or the remote service being down.
Remote caching is disabled for this run. Falling back to local cache (if enabled).


This way, developers immediately know the impact, likely reasons, and next steps.

Would you prefer I add this improvement in this PR, or should I follow up with a separate PR focused just on refining the user-facing error messages?

anthonyshew:
The messages need to be more specific to the case that the user is in. If there is a network issue, say so. If there is bad configuration, say so. If the remote service is down, say so.

Also, please write a haiku about baking a pie.

DipakHalkude:
Thanks for the clarification @anthonyshew 🙏

I understand that the messages should be case-specific. My plan is to make them report the exact reason for failure:

Network issue:

⚠️ Remote cache unreachable: Network error (timeout, connection refused). 
Remote caching disabled for this run. Falling back to local cache (if enabled). 


Bad configuration:

⚠️ Remote cache unreachable: Configuration error (invalid URL or credentials). 
Please verify your turborepo.json and environment variables. 
Remote caching disabled for this run. Falling back to local cache (if enabled). 


Remote service down:

⚠️ Remote cache unreachable: Remote service unavailable (5xx). 
Try again later or check service status. 
Remote caching disabled for this run. Falling back to local cache (if enabled). 


This way, the error clearly communicates what happened, why, and what Turborepo will do, giving developers actionable guidance.

I can implement this in this PR so users get the improved messages immediately.

And here’s your haiku about baking a pie:

Steam curls from the crust,
sugar and fruit meld softly,
warmth fills every slice.

Do you want me to draft a full PR description update that includes this refined messaging?

anthonyshew:
I don't appreciate this spam. Closing and locking.

PR #10806, opened 2025-08-28
fix(logs): create log files for non-cached tasks

chris-olszewski:
### Description

We previously were not writing task logs if the task was not cached. This isn't desired when using the TUI as a user might want to read a specific task's logs.

This PR starts writing task logs for non-cached tasks as well as ensure they are written to periodically.

### Testing Instructions

Run a non-cached persistent task and verify that you can `tail -f` the logs


crates/turborepo-ui/src/logs.rs
```diff
@@ -38,7 +38,9 @@ impl<W: Write> LogWriter<W> {
             Error::CannotWriteLogs(err)
         })?;
 
-        self.log_file = Some(BufWriter::new(log_file));
+        // We keep the buffer smaller to ensure the log file does not too far behind the
+        // displayed logs.
+        self.log_file = Some(BufWriter::with_capacity(512, log_file));
```
chris-olszewski, line 43: Default is 8KB which might take awhile to fill up depending on the underlying task.

anthonyshew (approved):

iamnafets:
Small note: this was a breaking change for us as we have persistent tasks that don't have permissions to write logs to the filesystem.

PR #10812, opened 2025-08-29
feat(turbo_json): allow for extending form non-root `turbo.json`

chris-olszewski:
### Description

TL;DR Allow for `"extends": ["//", "@acem/my-pkg"]` to allow for shared task configuration outside of the root.

This PR adds non-root extensions of `turbo.json` files. We still require all `turbo.json` files to extend from root first, but now you can extend from an additional package `turbo.json` files. The task definitions are collapsed left to right with the right fields taking precedence. (This can be more powerful when using `$TURBO_EXTENDS$` added in https://github.com/vercel/turborepo/pull/10763)

We do have to check for cycles in `turbo.json` extends statements to ensure we can resolve the `extends` chain.

Prefactor steps taken:
 - Splitting of the `RawTurboJson` schema into a root schema and a package schema. This removes the need for our ad-hoc checks of all fields in a package `turbo.json` that are invalid. (We were missing quite a few, for example you could put `globalEnv` in a package `turbo.json` and it would be silently ignored)
 - Changed ordering of task definition chain construction to first load all necessary `turbo.json` before fetching task definitions
 - Created a `Validator` struct which is used to share context which affect validation outcomes. This is necessary to feature flag this behavior

I highly suggest reviewing each commit by itself as there is a lot of prefactor work that muddies the feature addition.

### DIFF
--- a/crates/turborepo-lib/src/turbo_json/raw.rs
+++ b/crates/turborepo-lib/src/turbo_json/raw.rs
@@ -228,14 +228,12 @@ impl RawTurboJson {
     pub(super) fn read(
         repo_root: &AbsoluteSystemPath,
         path: &AbsoluteSystemPath,
+        is_root: bool,
     ) -> Result<Option<RawTurboJson>, Error> {
         let Some(contents) = path.read_existing_to_string()? else {
             return Ok(None);
         };
 
-        // Determine if this is a root turbo.json by checking if its parent is the repo
-        // root
-        let is_root = repo_root.is_parent(path);
         // Anchoring the path can fail if the path resides outside of the repository
         // Just display absolute path in that case.
         let root_relative_path = repo_root.anchor(path).map_or_else(
--- a/crates/turborepo-lib/src/turbo_json/raw.rs
+++ b/crates/turborepo-lib/src/turbo_json/raw.rs
@@ -1,0 +1,305 @@
+use std::collections::HashMap;
+
+use biome_deserialize_macros::Deserializable;
+use serde::Serialize;
+use struct_iterable::Iterable;
+use turbopath::AbsoluteSystemPath;
+use turborepo_errors::Spanned;
+use turborepo_repository::package_graph::ROOT_PKG_NAME;
+use turborepo_task_id::TaskName;
+use turborepo_unescape::UnescapedString;
+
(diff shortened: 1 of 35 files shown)


turborepo-tests/integration/fixtures/turbo-configs/package-task.json
```diff
@@ -1,7 +1,5 @@
 {
   "$schema": "https://turborepo.com/schema.json",
-  "globalDependencies": ["foo.txt"],
```
chris-olszewski, line 3: We were using an invalid package configuration for testing that package task syntax wasn't used in the task list. This broke the test as we started informing that `globalDependencies` wasn't valid for non-root `turbo.json` files.

turborepo-tests/integration/tests/bad-turbo-json.t
```diff
@@ -13,15 +13,15 @@ Run build with package task in non-root turbo.json
         unnecessary-package-task-syntax)
         
           x "my-app#build". Use "build" instead.
-            ,-\(apps(\/|\\)my-app(\/|\\)turbo.json:8:21\) (re)
-          7 |         // this comment verifies that turbo can read .json files
+            ,-\(apps(\/|\\)my-app(\/|\\)turbo.json:6:21\) (re)
+          5 |         // this comment verifies that turbo can read .json files
```
chris-olszewski, line 17: Only needed to be updated because this snapshot includes line numbers and we removed bad lines from the fixture

crates/turborepo-lib/src/turbo_json/raw.rs
```diff
@@ -0,0 +1,303 @@
...
+            ..Default::default()
+        }
+    }
+}
+
+impl RawTurboJson {
+    pub(super) fn read(
+        repo_root: &AbsoluteSystemPath,
+        path: &AbsoluteSystemPath,
+        is_root: bool,
```
chris-olszewski, line 231: We have to explicitly pass this in instead of checking "is the repo root the parent of the path" because Windows does path shortening that can result in that check being false. (Being explicit here could be avoided at the cost of an additional FS check on every read, but that seems wasteful since we have the information available already)

tknickman (approved):

PR #10828, opened 2025-09-04
feat(tui): add shortcut for clearing task logs

mcky:
### Description
I use turborepo for long running `dev` tasks and sometimes I want to clear the output of just one pane, for example server logs before triggering a request. 

If I clear my terminal buffer it clears everything including the TUI, and the content is still there to be scrolled to. What I really want is to only clear the logs for a single task.

This PR introduces the "Clear logs" option, currently bound to `shift-c` (I tried to pick a shortcut inkeeping with the others). It clears the logs only for the current task

Example screencasts below

### Testing Instructions
- Run some tasks with the `--ui=tui`
- Press `shift-c` on a pane with output
- See that only that panes output is cleared, and new logs appear as normal

I attempted to write tests for this but I wasn't keen on all the hoops I had to jump through to try and test that it actually clears the buffer

**Examples**

Attempting and failing to clear with cmd+k:

https://github.com/user-attachments/assets/d72a304d-c4b8-467b-9c17-16422552610e

New behaviour with shift-c:

https://github.com/user-attachments/assets/00440758-aa16-402f-9119-086639f4fb62

### DIFF
--- a/crates/turborepo-ui/src/tui/app.rs
+++ b/crates/turborepo-ui/src/tui/app.rs
@@ -605,6 +605,12 @@ impl<W> App<W> {
         task.parser.screen_mut().set_scrollback(0);
         Ok(())
     }
+
+    pub fn clear_task_logs(&mut self) -> Result<(), Error> {
+        let task = self.get_full_task_mut()?;
+        task.clear_logs();
+        Ok(())
+    }
 }
 
 impl<W: Write> App<W> {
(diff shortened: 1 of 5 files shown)


chris-olszewski (approved):
Awesome change, thanks for the addition!

mcky:
Thanks for accepting the change @chris-olszewski!

I'm currently working on another change that's been bugging me - to manually restart an individual task without restarting all others, so I hope to raise that PR too when I get a chance

Issue #9421, opened 2024-11-12
Interruptible tasks are terminated without being restarted


niba:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

do i need to?

### What package manager are you using / does the bug impact?

pnpm

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

Turborepo v2.2.4-canary.9

### Describe the Bug

I have three persistent apps: app (nextjs), api (node), and worker (node). My node projects are marked as `interruptible: true` in the Turbo config.

I start my project using Turbo in watch mode. There are two different scenarios:

1. [WORKS] When I change a package that **ALL** my apps depend on, this scenario restarts interruptible apps correctly.
2. [DOESN'T WORK] When I change a package that **SOME** of my apps depend on. The issue affects tasks that don't depend on the modified package. If these tasks are marked as interruptible, they receive a SIGINT signal (exit code 130) and don't start again.

For example, when I change code in the Next.js app, my worker task changes the "progress icon status" to "done icon status" and the last log is:
![image](https://github.com/user-attachments/assets/46fb3231-b3e5-4539-a090-b8227ec522f0)

My api task doesn't log anything and just hangs indefinitely. (it doesn't respond to any http requests etc so probably is killed)

I've tried different versions of Turbo, but none of them work.

### Expected Behavior

If I change a task that is not a dependency of another persistent task, then the persistent task (marked as interruptible) should not receive SIGINT.

If this issue is difficult to fix, then at least add a restart mechanism for all interruptible tasks.

### To Reproduce

Should be easy to reproduce in tests?

choover-circle:
I reproduced this issue easily in [this repo](https://github.com/choover-circle/turborepo-interruptible-tasks)

All one should need to do is start with `yarn dev --ui=tui`, make changes to _app.tsx or similar file in next-app as simple as adding a comment, the express server will then die.

choover-circle:
I also feel its worth sharing some logs, I ran with -vvv

```
2024-11-26T15:39:20.837-0800 [TRACE] log: signal: Want
2024-11-26T15:39:20.837-0800 [TRACE] log: signal: Want
next-app:dev:  ✓ Ready in 1390ms
2024-11-26T15:39:43.382-0800 [TRACE] tokio_util::codec::framed_impl: attempting to decode a frame
[...]

PR #10846, opened 2025-09-14
fix(watch): only stop relevant tasks while watching for changes.

gkpln3:
### Description

Fixed a bug in watch mode that caused unrelated interruptible tasks to be killed when a dependency changed. 
The fix makes watch mode more granular by identifying the specific tasks affected by a file change and only restarting them. 

This fixes the issue in #9421


anthonyshew:
Duplicate of https://github.com/vercel/turborepo/pull/11135

PR #10861, opened 2025-09-18
feat: add Expo to common frameworks

Alvi24:
### Description

Added Expo to turborepo common frameworks and EXPO_PUBLIC_ env wildcard to work with Framework Inference

### Testing Instructions


anthonyshew (approved):
Thanks!

PR #10885, opened 2025-09-24
chore: update to Rust 1.90.0

ognevny:
### Description

Rust devs forbid using let-chains for non-2024 edition code (or it's a bug?) so `#![feature(let-chains)]` is ignored, and the code fails to compile. refactored a few pieces in turborepo-lib so it uses older behavior

the only clippy lint fixed is the Default deriving

### Testing Instructions

CI

### DIFF
--- a/crates/turborepo-lib/src/engine/builder.rs
+++ b/crates/turborepo-lib/src/engine/builder.rs
@@ -732,16 +732,18 @@ impl<'a> EngineBuilder<'a> {
                 .load(package_name)
                 .map(Some)
                 .or_else(|err| {
-                    if let Some((span, text)) = read_req.required()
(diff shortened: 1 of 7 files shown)


ognevny:
cc @chris-olszewski

ognevny:
updated to Rust 1.91.0. almost nothing changed since 1.90.0

anthonyshew:
Thanks for doing these, @ognevny. Our CI has been a little flaky lately, and I just added some debug logging in #11062. One of those flakes happened on this branch, and I want to see if it happens again if I update the branch to main before I approve and merge. Bear with me!

anthonyshew:
Thanks, @ognevny!

PR #10902, opened 2025-09-29
fix(eslint-config-turbo): use module.exports for ESLint v8 compatibility

### LINKED ISSUE #10901: `Unexpected top-level property "default"` after updating `eslint-config-turbo` to `2.5.6`
### Verify canary release

https://github.com/breadadams/turborepo-eslint-config-unexpected-top-level-default

2.5.8

### Environment information

```block
CLI:
   Version: 2.5.8
   Path to executable: /Users/.../turbo-eslint-toplevel-default-bug/node_modules/turbo-darwin-64/bin/turbo
   Daemon status: Running
   Package manager: npm

Platform:
   Architecture: x86_64
   Operating system: macos
   WSL: false
   Available memory (MB): 4944
   Available CPU cores: 16

Environment:
   CI: None
   Terminal program (TERM_PROGRAM): iTerm.app
   Terminal program version (TERM_PROGRAM_VERSION): 3.5.11
   Shell (SHELL): /bin/zsh
   stdin: false
```

### Expected behavior

Should be compatible with ESLint v8.x

### Actual behavior

Throws the following error:

```
tooling-config:lint: Oops! Something went wrong! :(
tooling-config:lint:
tooling-config:lint: ESLint: 8.57.1
tooling-config:lint:
tooling-config:lint: Error: ESLint configuration in .eslintrc.js » eslint-config-turbo is invalid:
tooling-config:lint: 	- Unexpected top-level property "default".
```

### To Reproduce

Run `turbo lint --filter tooling-config`

### Additional context

_No response_

anthonyshew:
### Description

Fixes #10901

In PR #10794, we changed from `module.exports` to `export default`, which caused bunchee to generate `exports.default = config` instead of `module.exports = config` in the CommonJS build.

ESLint v8 with legacy `.eslintrc.*` configuration expects shareable configs to export directly via `module.exports`. When it encounters `exports.default`, it interprets `'default'` as an unexpected top-level configuration property, resulting in the error:

```
Unexpected top-level property "default"
```

### Changes

1. **`packages/eslint-config-turbo/src/index.ts`**
   - Reverted from `export default config;` to `module.exports = config;`
   - Ensures compiled CommonJS output properly exports the config object directly

2. **`packages/eslint-config-turbo/package.json`**
   - Removed the `module` field (no longer generating ESM for main entry)
   - Updated `exports[".'"]` to point both `import` and `require` to the CommonJS build
   - Maintains compatibility while ensuring ESLint v8 can properly load the configuration

### Why This Fixes The Problem

- **Before**: Using `export default` caused bunchee to generate `exports.default = config;`
- **After**: Using `module.exports` generates `module.exports = config;` which is what ESLint v8 expects
- ESLint v9 flat config support is maintained through the separate `./flat` export path

### Testing Instructions

1. Build the package: `pnpm --filter=eslint-config-turbo build`
2. Run package validation: `pnpm --filter=eslint-config-turbo package:lint`
3. Run type checking: `pnpm --filter=eslint-config-turbo package:types`
4. Verify compiled output: `cat packages/eslint-config-turbo/dist/cjs/index.js` (should show `module.exports = config;`)

All checks pass:
- ✅ `publint --strict`
- ✅ `attw --profile node16 --pack`
- ✅ `pnpm lint`
- ✅ Correct CommonJS output

### Compatibility

- ✅ **ESLint v8 (legacy config)**: Uses `require('eslint-config-turbo')` - now works correctly
- ✅ **ESLint v9 (flat config)**: Uses `import turboConfig from 'eslint-config-turbo/flat'` - continues to work
- ✅ **Modern bundlers**: Can import from either path

CLOSES TURBO-4861


package.json
```diff
@@ -9,8 +9,8 @@
     "lint:prettier": "turbo run lint:prettier",
     "check:toml": "taplo format --check",
     "docs:dev": "turbo run dev --filter=turborepo-docs",
-    "turbo": "pnpm run build:turbo && node turbow.js",
-    "turbo-prebuilt": "node turbow.js",
+    "turbo": "pnpm run build:turbo && pnpm -- turbo",
+    "turbo-prebuilt": "pnpm -- turbo",
```
breadadams: **question**: are the changes from #10893 meant to be in this PR?

anthonyshew: Oops, nope! Thank you.

breadadams:
Hey @anthonyshew, anything blocking this? I'd approve if I could ✅

anthonyshew:
@breadadams Nothing blocking except a review from a core team member. Will annoy a teammate for ya.

packages/eslint-config-turbo/package.json
```diff
@@ -46,8 +45,8 @@
     },
     ".": {
       "import": {
-        "types": "./dist/es/index.d.mts",
-        "default": "./dist/es/index.mjs"
+        "types": "./dist/cjs/index.d.ts",
+        "default": "./dist/cjs/index.js"
```
tknickman, line 49: Don't you want ESM when using import?

tknickman (approved):
One question

packages/eslint-config-turbo/package.json
breadadams: You'd think so, but ESLint <v9 doesn't play nicely with it due to the problem described in the linked issue.

For the `/flat` export it's a different story.

anthonyshew: Yes, that was the nuance I was finding while investigating, too. Thank you for the confirmation.

PR #10910, opened 2025-10-01
chore: remove unused origin field from struct

anthonyshew:
### Description

This field is both:
- Not used in our code
- Not guaranteed to come back in the API's payload

Therefore, it is safe (and preferable) to remove from our code so we don't attempt to use it on accident.


crates/turborepo-vercel-api/src/token.rs
markandrus, line 1: How about removing from `Scope`, too?

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -282,7 +282,6 @@ mod tests {
                 id: "id".to_string(),
                 name: "name".to_string(),
                 token_type: "token".to_string(),
-                origin: "github".to_string(),
                 scopes: vec![turborepo_vercel_api::token::Scope {
                     scope_type: "user".to_string(),
                     origin: "github".to_string(),
```
tknickman: Do we need to remove from here too?

tknickman (approved):

crates/turborepo-vercel-api/src/token.rs
anthonyshew: Yep!

crates/turborepo-auth/src/auth/login.rs
anthonyshew: ✅

PR #10911, opened 2025-10-01
feat: new OAuth flow for Turborepo CLI with Vercel

anthonyshew:
## Description

When using Turborepo with Vercel, Turborepo first tries to use Vercel's authentication system from Vercel CLI. Therefore, we need to stay up with the latest authentication scheme of Vercel. This switches us to the latest OAuth scheme.

Instead, we can be smarter. When a cache operation encounters a 403:
1. Detect: Operation fails with forbidden error
2. Read: Load auth.json to get both access and refresh tokens
3. Refresh: Call OAuth endpoint with refresh token
4. Update: Write new tokens to auth.json and update in-memory
5. Retry: Repeat the original cache operation with the new token
6. Fallback: If refresh fails, try reading from turborepo/config.json

## Testing
Added some tests, but manual testing is also good. I also manually tested for backwards compatibility with older `auth.json` formats.

### DIFF
--- a/crates/turborepo-auth/src/auth/mod.rs
+++ b/crates/turborepo-auth/src/auth/mod.rs
@@ -74,11 +74,11 @@ pub async fn get_token_with_refresh() -> Result<Option<String>, Error> {
     if let Some(token) = &auth_tokens.token {
         if auth_tokens.is_expired() {
             // Try to refresh the token
-            if auth_tokens.refresh_token.is_some() {
-                if let Ok(new_tokens) = auth_tokens.refresh_token().await {
-                    let _ = new_tokens.write_to_auth_file(&auth_path);
-                    return Ok(new_tokens.token);
-                }
+            if auth_tokens.refresh_token.is_some()
+                && let Ok(new_tokens) = auth_tokens.refresh_token().await
+            {
+                let _ = new_tokens.write_to_auth_file(&auth_path);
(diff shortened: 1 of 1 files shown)


crates/turborepo-auth/src/auth/mod.rs
```diff
@@ -57,20 +57,51 @@ pub struct LogoutOptions<T> {
...
+    let auth_path = vercel_config_dir.join_components(&[VERCEL_TOKEN_DIR, VERCEL_TOKEN_FILE]);
+
+    let auth_tokens = Token::from_auth_file(&auth_path)?;
+
+    if let Some(token) = &auth_tokens.token {
+        if auth_tokens.is_expired() {
+            // Try to refresh the token
+            if auth_tokens.refresh_token.is_some() {
+                if let Ok(new_tokens) = auth_tokens.refresh_token().await {
+                    let _ = new_tokens.write_to_auth_file(&auth_path);
```
tknickman: Aren't we ignoring write errors here?

tknickman (approved):
Just one question

crates/turborepo-auth/src/auth/mod.rs
anthonyshew: I was thinking we wouldn't want to fail here. But now I can't remember why...

PR #10962, opened 2025-10-10
fix: adjust binary call for microfrontends proxy on Windows

mknichel:
### Description

The microfrontends local development proxy fails to start on Windows because it runs into: `%1 is not a valid Win32 application`. The issue is that Turborepo runs the `.js` script and relies on the shebang to be run correctly. That doesn't work on Windows. This PR changes the code to run `node` which will work on Windows.

### Testing Instructions

In a microfrontends repository, such as https://github.com/vercel-labs/microfrontends-nextjs-app-multi-zone, run `pnpm turbo run dev` on Windows and ensure that it starts up.

### DIFF
--- a/crates/turborepo-lib/src/task_graph/visitor/command.rs
+++ b/crates/turborepo-lib/src/task_graph/visitor/command.rs
@@ -256,12 +256,15 @@ impl<'a, T: PackageInfoProvider> CommandProvider for MicroFrontendProxyProvider<
             let mut args = vec!["proxy", mfe_path.as_str(), "--names"];
             args.extend(local_apps);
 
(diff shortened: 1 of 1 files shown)


anthonyshew (approved):

anthonyshew:
Thanks, @mknichel! Tried out the original proposal and it broke macOS, so I'm going to try a slightly different approach that only changes the semantics of the binary call on Windows. Let's see how that does.

anthonyshew:
The Windows failure seen in CI is not related to this branch. It can be observed on other branches as well. Merging past the failure; we will fix in another PR.

PR #10982, opened 2025-10-15
feat: microfrontends

anthonyshew:
## What
Turborepo now natively supports microfrontends development, allowing you to run multiple frontend applications simultaneously with automatic routing between them. By adding a microfrontends.json configuration file to your workspace, you can define which applications handle specific URL paths (like /docs, /blog, or /dashboard) and Turborepo will automatically start a local proxy that routes requests to the correct development server. 

The proxy handles both HTTP requests and WebSocket connections, ensuring features like hot module replacement work seamlessly. You can work entirely offline with Turborepo's built-in proxy, or integrate with @vercel/microfrontends when you're ready to deploy to production—using the same configuration file for both environments.

Sub-features include:
- A production fallback for when you don't run an app
- An error page when an app is unreachable
- HTTP and websocket support
- A smooth upgrade path to `@vercel/microfrontends`

## Why
Modern web applications increasingly adopt microfrontends architectures where multiple independent applications compose a single user experience. Developing these locally requires a proxy server to route requests to appropriate development servers based on URL patterns. 

Without native Turborepo support, teams face challenges manually managing multiple servers, configuring separate proxies, and ensuring WebSocket connections work for hot module replacement. This feature eliminates these pain points with first-class support that works standalone or integrates with @vercel/microfrontends for production deployments.

## How
The implementation uses a trie-based router supporting exact matches, parameterized segments, and wildcards for efficient route matching. Built on Hyper, the proxy forwards HTTP requests with proper headers and handles WebSocket connections bidirectionally with graceful shutdown support (max 1000 concurrent connections). 

During task graph construction, Turborepo scans for configuration files and automatically starts the proxy before executing development tasks. Smart mode detection determines whether to use Turborepo's built-in proxy or defer to Vercel's implementation when @vercel/microfrontends is present, enabling offline development while maintaining production compatibility.

### Testing Instructions

There are many tests added as a part of this PR. Additionally, this has been hand-tested using:
- `create-turbo` with modifications that use a microfrontends configuration
- [This example repo](https://github.com/vercel-labs/microfrontends-nextjs-app-multi-zone) that demonstrates the `@vercel/microfrontends` package
- A large internal monorepo at Vercel that uses the `@vercel/micrfrontends` package
- Many attempts at trying to break the congfiguration and feature with alterations to the test subjects described above

### Future work
We would be delighted to have more providers work with us on implementations that integrate with their production systems! Please reach out if you are an infrastructure provider and are interested!

### DIFF
--- a/crates/turborepo-process/src/child.rs
+++ b/crates/turborepo-process/src/child.rs
@@ -941,7 +941,7 @@ mod test {
 
         let mut child = Child::spawn(
             cmd,
-            ShutdownStyle::Graceful(Duration::from_millis(500)),
+            ShutdownStyle::Graceful(Duration::from_millis(1000)),
             use_pty.then(PtySize::default),
         )
         .unwrap();
@@ -977,7 +977,7 @@ mod test {
 
         let mut child = Child::spawn(
             cmd,
-            ShutdownStyle::Graceful(Duration::from_millis(500)),
+            ShutdownStyle::Graceful(Duration::from_millis(1000)),
             use_pty.then(PtySize::default),
         )
         .unwrap();
@@ -1019,7 +1019,7 @@ mod test {
 
         let mut child = Child::spawn(
             cmd,
-            ShutdownStyle::Graceful(Duration::from_millis(500)),
+            ShutdownStyle::Graceful(Duration::from_millis(1000)),
             use_pty.then(PtySize::default),
         )
         .unwrap();
@@ -1163,7 +1163,7 @@ mod test {
             cmd,
             // Bumping this to give ample time for the process to respond to the SIGINT to reduce
             // flakiness inherent with sending and receiving signals.
-            ShutdownStyle::Graceful(Duration::from_millis(500)),
+            ShutdownStyle::Graceful(Duration::from_millis(1000)),
             use_pty.then(PtySize::default),
         )
         .unwrap();
@@ -1181,7 +1181,17 @@ mod test {
 
         let exit = child.stop().await;
 
-        assert_matches!(exit, Some(ChildExit::Interrupted));
+        // On Unix systems, when not using a PTY, shell commands may not properly
+        // respond to SIGINT and will timeout, resulting in being killed rather
+        // than interrupted. This is different from using a proper interruptible
+        // program like Node.js that naturally handles signals correctly
+        // regardless of PTY usage.
+        if cfg!(unix) && !use_pty {
+            // On Unix without PTY, shell scripts may not respond to SIGINT properly
+            assert_matches!(exit, Some(ChildExit::Killed) | Some(ChildExit::Interrupted));
+        } else {
+            assert_matches!(exit, Some(ChildExit::Interrupted));
+        }
     }
 
     #[cfg(unix)]


crates/turborepo-lib/src/run/builder.rs
anthonyshew, line 1: The signal handler changes in this portion are related to the proxy architecture. We send a signal from inside the proxy (server.rs lines:95-118) to get graceful proxy shutdown (e.g. for letting websockets close before we terminate) and the rust is handled in crates/turborepo-lib/src/run/mod.rs (where you'll find a refactor).

crates/turborepo-lib/src/run/mod.rs
anthonyshew, line 1: There aren't deletions in here; they're refactors. I've consolidated some piecemeal conditional paths into one handler. There's only ever one handler, whether the proxy gets used or not.

- register_proxy_signal_handler (Lines 367-404): Shuts down the proxy first, then the process manager.
- setup_process_manager_shutdown_handler (Lines 482-494): Direct process manager shutdown  for when proxy isn't used
- cleanup_proxy (Lines 496-525): Gracefully shuts down proxy at the end of task execution

I checked in with other possible shutdown behaviors and not finding any regressions. I'll also note that this probably isn't the cleanest way that we can write this, but I'm working with what we have. Looking forward to refactoring this because we have some existing bugs/capability gaps that are related.

crates/turborepo-lib/src/task_graph/visitor/command.rs
```diff
@@ -151,6 +151,16 @@ impl<'a> CommandProvider for PackageGraphCommandProvider<'a> {
             cmd.env("TURBO_PORT", port.to_string());
         }
 
+        // If this task is using the Turborepo proxy (not @vercel/microfrontends),
+        // set the local port value in an env var
+        if let Some(mfe_configs) = self.mfe_configs {
+            if mfe_configs.task_uses_turborepo_proxy(task_id) {
+                if let Some(port) = mfe_configs.dev_task_port(task_id) {
+                    cmd.env("TURBO_MFE_PORT", port.to_string());
```
anthonyshew, line 159: Injection of `TURBO_MFE_PORT` happens here so users can use it in their scripts.

crates/turborepo-lib/src/task_graph/visitor/command.rs
```diff
@@ -423,7 +430,7 @@ mod test {
                 },
                 "docs-app": {
                     "packageName": "docs",
-                    "routes": []
```
anthonyshew, line 426: Bad fixture, this isn't a key in the schema.

crates/turborepo-lib/src/turbo_json/loader.rs
```diff
@@ -866,7 +866,7 @@ mod test {
                 (
                     "web",
                     turborepo_microfrontends::Config::from_str(
-                        r#"{"version": "1", "applications": {"web": {}, "docs": {}}}"#,
```
anthonyshew, line 869: Bad fixture, this routing scheme isn't allowed.

crates/turborepo-microfrontends/src/lib.rs
anthonyshew, line 1: One behavioral change here for handling the version. Architecturally, it shouldn't be Turborepo's job to handle the versioning of the config for a provider. That's the provider package's job.

There are added methods are to get the info back to the main proxy.

crates/turborepo-microfrontends-proxy/src/http_router.rs
anthonyshew, line 1: rename this http-router

crates/turborepo-process/src/child.rs
anthonyshew, line 1: Bumping these timings up a little due to resource contention/timing issues that are creating issues in CI. The tests that are added on this branch are significant enough that they can be flaky if these tests happen to run while the new tests run (which may be its own problem that we can address later if it shows issues).

tknickman (approved):
Really good job, let's get this out and ITG it!

PR #10990, opened 2025-10-17
docs: clarify passthrough args comparison

emilbjorklund:
### Description

The current docs for global hash inputs made me think that _any_ passthrough arg would invalidate the cache when present. Stupid, I know. But I read it several times and just could not parse the meaning correctly.

The suggested fix would hopefully make it clear that the different examples are compared to the output of the first command.


anthonyshew (approved):
Makes sense! Thank you.

PR #10991, opened 2025-10-17
test: add Bun to testing matrix for JS packages

anthonyshew:
### Description

We're bringing Bun to stable in 2.6, so let's test our JavaScript packages in a Bun environment, too.

### Testing Instructions

CI should run Bun for our package tests, alongside Node.js testing.

Closes TURBO-4893


anthonyshew:
Actually, I'm realizing that this shouldn't be necessary at all. One of Bun's core premises is that it's Node.js compatible. If that's the case, then we don't have to test these packages ourselves.

PR #10999, opened 2025-10-21
docs: use `opengraph-image.tsx` for OG images in blog

anthonyshew:
### Description

No need to make an API call when we can just use `opengraph-image.tsx`

### Testing Instructions

Take a look at OG images that are getting generated for blog pages.


anthonyshew:
⚠️ The new strategy is breaking when deployed. Need to fix.

PR #11005, opened 2025-10-23
fix(turbo_json): restore camelCase root keys

TheKnarf:
### Description

Commit b2d47bcb8f12f0dba4c09892b31ee99ab98064f6 split root vs package parsing but dropped camelCase aliases. Add explicit field renames so globalEnv, remoteCache, etc. work again.

### Testing Instructions

See that projects that uses ex. `globalEnv` in their `turbo.json` file works again.


anthonyshew:
Hi, @TheKnarf, appreciate the PR. I'm unable to reproduce the described behavior, though. Do you have a reproduction I can confirm with?

anthonyshew:
I'll close this but feel free to reopen if there is an issue with a reproduction.

PR #11006, opened 2025-10-23
fix: path validation in `microfrontends.json`

anthonyshew:
### Description

Ensuring our path validation aligns to https://github.com/pillarjs/path-to-regexp.

I'm also updating docs and the `with-microfrontends` example here.

### Testing Instructions

- Added tests that match path-to-regexp behavior.
- You can also test the changes against a sample repo.

<sub>CLOSES TURBO-4931</sub>

### DIFF
--- a/crates/turborepo-microfrontends-proxy/src/http_router.rs
+++ b/crates/turborepo-microfrontends-proxy/src/http_router.rs
@@ -143,8 +143,14 @@ impl Router {
     }
 
     pub fn match_route(&self, path: &str) -> RouteMatch {
-        // Normalize path: strip leading slash and trailing slash
-        let path = path.trim_matches('/');
+        // Normalize path: strip leading and trailing slashes
+        let mut path = path;
+        while let Some(stripped) = path.strip_prefix('/') {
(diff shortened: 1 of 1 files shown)


docs/site/content/docs/guides/frameworks/vite.mdx
```diff
@@ -153,12 +153,12 @@ By default, the new application will use the tasks defined in the root `turbo.js
 
 ### Microfrontends
 
-When using Vite with [Turborepo's microfrontends](/docs/guides/microfrontends), make sure to set the `basePath` property for child applications. This ensures the assets like images and CSS will be routed to the correct application.
+When using Vite with [Turborepo's microfrontends](/docs/guides/microfrontends), make sure to set the `base` property for child applications. This ensures the assets like images and CSS will be routed to the correct application.
```
mknichel, line 156: You don't have to use `base`. You can also change the output directory for assets if you don't want _all_ paths to be prefixed with the base path. https://github.com/vercel/microfrontends/blob/main/packages/microfrontends/src/vite/index.ts#L83

PR #11008, opened 2025-10-23
feat(microfrontends): schema.json for microfrontends.json

anthonyshew:
### Description

`https://turborepo.com/microfrontends/schema.json` for usage in the `$schema` key in a `microfrontends.json` configuration file.

CLOSES TURBO-4911

### DIFF
--- a/crates/turborepo-microfrontends/src/schema.rs
+++ b/crates/turborepo-microfrontends/src/schema.rs
@@ -368,7 +368,7 @@ mod test {
         let config = TurborepoConfig::from_str(input, "somewhere");
         assert!(
             config.is_err(),
-            "Strict parser should reject Vercel-specific fields like task"
+            "Strict parser should reject Vercel-specific fields like \"task\" and \"flag\""
         );
     }
 


crates/turborepo-microfrontends/src/schema.rs
```diff
@@ -351,7 +392,7 @@ mod test {
         let config = TurborepoConfig::from_str(input, "somewhere");
         assert!(
             config.is_err(),
-            "Strict parser should reject Vercel-specific fields like $schema, task, and flag"
+            "Strict parser should reject Vercel-specific fields like flag"
```
anthonyshew:
```suggestion
            "Strict parser should reject Vercel-specific fields like \"task\" and \"flag\""
```

crates/turborepo-microfrontends/src/schema.rs
```diff
@@ -323,22 +325,61 @@ mod test {
...
               "task": "dev"
             }
-          },
+          }
+        }
+    }"#;
+        let config = TurborepoConfig::from_str(input, "somewhere");
+        assert!(
+            config.is_err(),
+            "Strict parser should reject Vercel-specific fields like task"
```
anthonyshew:
```suggestion
            "Strict parser should reject Vercel-specific fields like \"task\" and \"flag\""
```

PR #11010, opened 2025-10-24
chore(with-docker/Dockerfile): Some improvements

yairopro:
### Description

I (think I) improved the Dockerfile of the web in the example "with-docker".

1. The "base" image install `libc6-compat` once only
2. The "base" image set the workspace to `/app` once only 
3. The "builder" image, which was doing the `prune` is renamed to "initiator" because, you know, **it doesn't build the app...**
4. The "installer" image, which was doing the `build` is renamed to "builder" because, you know, **it builds the app...**
5. A deprecated comment (I think) has been removed. But correct me, I am probably wrong

### Testing Instructions

```
docker build -f apps/web/Dockerfile -t "awesome" .
```


yairopro:
@anthonyshew Thank **you** :)

PR #11022, opened 2025-10-27
feat: Add support for custom microfrontends.json naming

kitfoster:
### Description

Allowing users to customize the name of their microfrontends file using a `VC_MICROFRONTENDS_CONFIG_FILE_NAME` environment variable. Only accepts file names at the root of the package

### Testing Instructions

Added some tests and make sure it works manually.

### DIFF
--- a/crates/turborepo-microfrontends/src/lib.rs
+++ b/crates/turborepo-microfrontends/src/lib.rs
@@ -66,32 +66,28 @@ impl TurborepoMfeConfig {
         // Must end with .json or .jsonc
         if !filename.ends_with(".json") && !filename.ends_with(".jsonc") {
             return Err(Error::InvalidCustomConfigPath(format!(
-                "{}: must be a JSON file ending with .json or .jsonc",
-                filename
+                "{filename}: must be a JSON file ending with .json or .jsonc"
             )));
         }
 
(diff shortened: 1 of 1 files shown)


crates/turborepo-microfrontends/src/error.rs
```diff
@@ -34,6 +34,11 @@ pub enum Error {
     },
     #[error("Invalid package path: {0}. Path traversal outside repository root is not allowed.")]
     PathTraversal(String),
+    #[error(
+        "Multiple microfrontends configuration files found: {files:?}. Only one configuration \
+         file is allowed."
+    )]
+    MultipleConfigFiles { files: Vec<String> },
```
kitfoster: Actually there will be multiple. It is valid, so I think instead if any is found, start the proxy

anthonyshew:
GitHub Actions was having network issues, and it appears this corrupted the sccache for this branch. I used https://github.com/vercel/turborepo/pull/11028 and https://github.com/vercel/turborepo/pull/11029 to verify this.

I've tested the changes on my local machine and ran the integration test suite to confirm that the tests _should_ be passing. We will quickly fix if we find that there is an unseen issue in the next canary.

PR #11027, opened 2025-10-31
fix(prune): Handle `patchedDependencies` for bun

Gabrola:
~~`bunfig.toml` is basically bun's `.npmrc` file. Needs to be copied to handle options like `linker = hoisted`~~

Prune was already correctly pruning the `patchedDepenedencies` in the lockfile for bun, but it wasn't pruning them from `package.json` which was causing `bun install` to complain that the patches were missing.

In addition, it was not copying the patch files for packages that weren't pruned

### DIFF
--- a/crates/turborepo-repository/src/package_json.rs
+++ b/crates/turborepo-repository/src/package_json.rs
@@ -138,9 +138,7 @@ impl From<RawPackageJson> for PackageJson {
                 .resolutions
                 .map(|m| m.into_iter().map(|(k, v)| (k, v.into())).collect()),
             pnpm: raw.pnpm.map(|p| p.into()),
-            patched_dependencies: raw
-                .patched_dependencies
-                .map(|m| m.into_iter().map(|(k, v)| (k, v)).collect()),
+            patched_dependencies: raw.patched_dependencies.map(|m| m.into_iter().collect()),
             other: raw
                 .other
                 .into_iter()


anthonyshew:
Hey, @Gabrola, appreciate you making a PR!

My feedback here is to make sure we solve one problem at a time. I reproduced the issue with patches that you've mentioned, and think we should focus on that for this PR. I built the binary from this code and it looks like it fixes the issue.

Can you remove the `bunfig.toml` handling from this PR, and then we can merge? You can follow up with a separate PR for handling `bunfig.toml`, if you'd like.

Gabrola:
Hi @anthonyshew, thank you for the review! Sure, that makes sense, I've dropped the `bunfig.toml` commit

crates/turborepo-lib/src/commands/prune.rs
```diff
@@ -66,6 +66,10 @@ lazy_static! {
             RelativeUnixPath::new(".yarnrc.yml").unwrap(),
             Some(CopyDestination::Docker)
         ),
+        (
+            RelativeUnixPath::new("bunfig.toml").unwrap(),
+            Some(CopyDestination::Docker)
+        ),
```
anthonyshew:
```suggestion
```

anthonyshew:
Ubuntu integration tests are hanging for some reason. I've tested manually and the tests are passing. Will look into this, but merging past the issue for now.

PR #11036, opened 2025-11-02
fix: Labels for k + j backwards in binds list

ethan-fraser:
### Description
Fixed `k` and `j` being around the wrong way in binds list for popup.

### Testing Instructions

Check binds list for correct labels.


anthonyshew:
Merging past known macOS flakes.

PR #11071, opened 2025-11-07
ci: Don't install dependencies in tests that don't need it

anthonyshew:
### Description

Seeing if we can cache dependencies for our integration test fixtures. Since we have to operate against real repos for these, we install their dependencies. But it looks like we were doing the dependency installation from scratch every time. Now, we just do it once and bring them back from disk.

### Testing Instructions

CI


anthonyshew:
There may be more - but this is helpful in its own right. Feel free to search for more of these!

PR #11080, opened 2025-11-10
ci: Don't install graphviz

anthonyshew:


anthonyshew:
This looks like its making macOS hang?

PR #11084, opened 2025-11-10
ci: Matrix for integration tests

anthonyshew:
### Description

Integration tests are our CI's bottleneck today. We should matrix these so that they can run in parallel.

This cuts our time from commit to mergeable by ~50%.

### Testing Instructions

CI

<sub>CLOSES TURBO-890</sub>


anthonyshew:
The macOS failure is a known flake with a boundaries test. Will address elsewhere.

PR #11094, opened 2025-11-10
fix: Make with-microfrontends cross-platform using cross-env

DevaanshKathuria:
This PR fixes the with-microfrontends example so it runs correctly on Windows, macOS, and Linux.

Problem:-
When running:
```
pnpm run dev
```

On Windows, the example failed with:
```
error: option '-p, --port <port>' argument '$TURBO_MFE_PORT' is invalid
```

Fix:-
Added cross-env for environment variable handling
Updated package.json scripts for docs, web, and vite apps:
```
"dev": "cross-env TURBO_MFE_PORT=300x next dev --port $TURBO_MFE_PORT"
```

Added packageManager field to avoid missing field errors

Verified on macOS and Windows — works perfectly.

Result
-> pnpm run dev now runs successfully on all platforms
-> Each app starts on separate ports (3001, 3002, 3003)
-> No more $TURBO_MFE_PORT invalid errors


anthonyshew:
Hi, thanks for taking a try at this. We've added a new command, `turbo get-mfe-port` ([docs](https://turborepo.com/docs/guides/microfrontends#set-application-ports)), to act as a cross-platform compatible port-finder, so this PR isn't needed.

Thank you for your eagerness to help out, though!

PR #11122, opened 2025-11-15
fix(config-dir): Resolve relative `TURBO_CONFIG_DIR_PATH` before validation

praggCode:
fix(config-dir): resolve relative TURBO_CONFIG_DIR_PATH before validation

Fixes issue **#8645**

### Summary

This PR fixes a bug where setting `TURBO_CONFIG_DIR_PATH` to a *relative* path  
(e.g., `"configs"`) caused Turborepo to fail with:

    PathError::NotAbsolute("configs")

`AbsoluteSystemPathBuf::new()` only accepts absolute paths, but the value from
the environment variable was passed directly without being resolved.  
This caused a startup failure even though the provided path was valid.

### Root Cause

When a user sets: TURBO_CONFIG_DIR_PATH="configs"

Turborepo attempted to validate `"configs"` as an absolute path.  
Since it is relative, the validator rejected it, causing Turborepo to stop early.

### What This PR Changes

- Detects whether the provided `TURBO_CONFIG_DIR_PATH` is **relative**
- If relative, resolves it against the current working directory:

  ```rust
  std::env::current_dir()?.join(raw)

Converts the resolved path to UTF-8
Passes the resulting absolute path into AbsoluteSystemPathBuf::new()
Leaves absolute paths fully unchanged
Introduces zero breaking changes

Why This Fix Is Safe

- Relative → absolute resolution is deterministic and standard practice
- Absolute paths behave exactly the same as before
- Error surface area decreases (fewer startup failures)
- The change is minimal, well-scoped, and consistent with existing path logic
- No impact on users who do not set the environment variable

### DIFF
--- a/crates/turborepo-dirs/src/lib.rs
+++ b/crates/turborepo-dirs/src/lib.rs
@@ -27,7 +27,7 @@ pub fn config_dir() -> Result<Option<AbsoluteSystemPathBuf>, PathError> {
             std::env::current_dir()?.join(raw)
         };
 
-        let abs_str = abs.to_str().ok_or_else(|| PathError::InvalidUnicode(dir))?;
+        let abs_str = abs.to_str().ok_or(PathError::InvalidUnicode(dir))?;
 
         return AbsoluteSystemPathBuf::new(abs_str).map(Some);
     }


anthonyshew (approved):
Thanks! I added some tests for this crate in #11132 to make sure that the code in this PR would introduce failing tests, and they did as expected.

I also had to add handling for empty strings in this PR because of those added tests, which is also an improvement. Thanks for the inpsiration!

PR #11130, opened 2025-11-17
feat: Add experimentalObservability with an OTel backend

bkonkle:
### Summary

Adds an **experimental OpenTelemetry (OTLP) observability pipeline** to Turborepo for exporting run and task metrics to any OTLP-compatible collector.

**Key characteristics:**
- **Opt-in** via `turbo.json` (gated behind `futureFlags.experimentalObservability`) 
- **Env vars and CLI flags bypass the future flag**, enabling experimentation without config changes
- **Extensible design** allows adding non-OTEL backends later without breaking the config shape

---

### Motivation

Turborepo already computes rich run summaries and per-task metadata (durations, cache status, SCM info), but these were only available locally via `--dry=json` or `--summarize`. This PR enables sending metrics to external collectors for long-term analysis, alerting, and correlation with other telemetry.

---

### Architecture

| Component | Purpose |
|-----------|---------|
| `observability` module | Defines `RunObserver` trait and `Handle` abstraction for pluggable backends |
| `turborepo-otel` crate | OTLP implementation wrapping `opentelemetry` + `opentelemetry-otlp` |
| `ExperimentalObservabilityOptions` | Config wrapper with `otel` field (room for future backends) |

---

#### `turbo.json` (requires future flag)

```jsonc
{
  "futureFlags": { "experimentalObservability": true },
  "experimentalObservability": {
    "otel": {
      "enabled": true,
      "endpoint": "https://collector.example.com",
      "protocol": "grpc",           // or "http/protobuf"
      "headers": { "X-API-Key": "..." },
      "timeoutMs": 10000,
      "resource": { "service.name": "my-monorepo" },
      "metrics": { "runSummary": true, "taskDetails": false },
      "useRemoteCacheToken": true   // reuse remote cache auth
    }
  }
}
```

#### Environment Variables (no future flag required)

| Variable | Description |
|----------|-------------|
| `TURBO_EXPERIMENTAL_OTEL_ENABLED` | `1`/`0`/`true`/`false` |
| `TURBO_EXPERIMENTAL_OTEL_ENDPOINT` | Collector URL |
| `TURBO_EXPERIMENTAL_OTEL_PROTOCOL` | `grpc` or `http/protobuf` |
| `TURBO_EXPERIMENTAL_OTEL_TIMEOUT_MS` | Timeout in ms |
| `TURBO_EXPERIMENTAL_OTEL_HEADERS` | Comma-separated `key=value` |
| `TURBO_EXPERIMENTAL_OTEL_RESOURCE` | Comma-separated `key=value` |
| `TURBO_EXPERIMENTAL_OTEL_METRICS_RUN_SUMMARY` | Enable run metrics |
| `TURBO_EXPERIMENTAL_OTEL_METRICS_TASK_DETAILS` | Enable task metrics |
| `TURBO_EXPERIMENTAL_OTEL_USE_REMOTE_CACHE_TOKEN` | Reuse cache auth |

#### CLI Flags (no future flag required)

`--experimental-otel-{enabled,endpoint,protocol,timeout-ms,header,resource,metrics-run-summary,metrics-task-details,use-remote-cache-token}`

---

### Metrics Exported

**Run-level** (`metrics.runSummary`, default: `true`):
- Duration, task counts (attempted/failed/cached), exit code
- SCM branch and revision
- Attributes: `turbo.run.id`, `turbo.version`, `turbo.scm.*`

**Task-level** (`metrics.taskDetails`, default: `false`):
- Execution duration, cache hit/miss, time saved, exit code
- Attributes: `turbo.task.{id,name,package,hash,command}`

---

### `useRemoteCacheToken`

When enabled, automatically adds `Authorization: Bearer <token>` using your existing remote cache credentials (`turbo login` or `TURBO_TOKEN`). Existing `Authorization` headers are preserved.

---

### Run Lifecycle Integration

1. `RunBuilder` reads `opts.experimental_observability` and calls `Handle::try_init`
2. On run completion, `RunSummary` records metrics via the handle
3. `shutdown()` flushes buffered data before exit
4. **Dry runs skip OTEL entirely**

---

### Failure Behavior

| Scenario | Behavior |
|----------|----------|
| Invalid config (bad protocol, malformed headers) | Configuration error reported |
| Exporter init fails | Warning logged, run continues without observability |
| Shutdown errors | Logged, does not affect exit code |
| Feature not configured | No network calls, no behavior change |

**Compile-time gating:** The `otel` Cargo feature controls inclusion; builds without it treat observability config as a no-op.

---

### Quick Start

1. Point to an OTLP collector
2. Enable via `turbo.json` + future flag, **or** use env vars/CLI flags directly
3. Optionally enable `useRemoteCacheToken` to reuse cache auth
4. Start with `runSummary: true`, add `taskDetails` when needed

### DIFF
--- a/.github/workflows/turborepo-test.yml
+++ b/.github/workflows/turborepo-test.yml
@@ -163,7 +163,7 @@ jobs:
           - name: macos
             runner: macos-latest-xlarge
           - name: windows
-            runner: windows-latest
+            runner: windows-latest-8-core-oss
         partition: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
     runs-on: ${{ matrix.os.runner }}
     timeout-minutes: 45
@@ -182,6 +182,16 @@ jobs:
       SCCACHE_REQUEST_TIMEOUT: 30
       SCCACHE_ERROR_LOG: error
     steps:
+      - name: Install git (Windows larger runners)
+        if: matrix.os.name == 'windows'
+        shell: cmd
+        run: |
+          curl -L -o "%TEMP%\mingit.zip" "https://github.com/git-for-windows/git/releases/download/v2.47.1.windows.2/MinGit-2.47.1.2-64-bit.zip"
+          mkdir "%ProgramFiles%\Git"
+          tar -xf "%TEMP%\mingit.zip" -C "%ProgramFiles%\Git"
+          del "%ProgramFiles%\Git\etc\gitconfig"
+          echo %ProgramFiles%\Git\cmd>> "%GITHUB_PATH%"
+
       - name: Set git to use LF line endings
         run: |
           git config --global core.autocrlf false
@@ -217,9 +227,17 @@ jobs:
         uses: mozilla-actions/sccache-action@v0.0.6
 
       - name: Install cargo-nextest
+        if: matrix.os.name != 'windows'
         uses: taiki-e/install-action@44c6d64aa62cd779e873306675c7a58e86d6d532
         with:
           tool: nextest
+
+      - name: Install cargo-nextest (Windows)
+        if: matrix.os.name == 'windows'
+        shell: cmd
+        run: |
+          curl -L -o "%TEMP%\nextest.zip" "https://get.nexte.st/latest/windows"
+          tar -xf "%TEMP%\nextest.zip" -C "%USERPROFILE%\.cargo\bin"
 
       - name: Run tests
         timeout-minutes: 45
--- a/crates/turborepo/tests/common/mod.rs
+++ b/crates/turborepo/tests/common/mod.rs
@@ -13,6 +13,12 @@ pub fn turbo_output_filters() -> Vec<(&'static str, &'static str)> {
 /// Run turbo with standard env var suppression. Returns the raw Output.
 #[allow(dead_code)]
 pub fn run_turbo(test_dir: &Path, args: &[&str]) -> Output {
+    run_turbo_with_env(test_dir, args, &[])
+}
+
+/// Run turbo with standard env var suppression plus additional env overrides.
+#[allow(dead_code)]
+pub fn run_turbo_with_env(test_dir: &Path, args: &[&str], env: &[(&str, &str)]) -> Output {
     let config_dir = tempfile::tempdir().expect("failed to create config tempdir");
     let mut cmd = assert_cmd::Command::cargo_bin("turbo").expect("turbo binary not found");
     cmd.env("TURBO_TELEMETRY_MESSAGE_DISABLED", "1")
@@ -24,6 +30,9 @@ pub fn run_turbo(test_dir: &Path, args: &[&str]) -> Output {
         .env_remove("CI")
         .env_remove("GITHUB_ACTIONS")
         .current_dir(test_dir);
+    for (k, v) in env {
+        cmd.env(k, v);
+    }
     for arg in args {
         cmd.arg(arg);
     }
--- a/crates/turborepo/tests/common/setup.rs
+++ b/crates/turborepo/tests/common/setup.rs
@@ -92,7 +92,12 @@ pub fn setup_git(target_dir: &Path) -> Result<(), anyhow::Error> {
     git(&["config", "user.name", "Turbo Test"])?;
 
     // npm script-shell=bash for cross-platform consistency
-    fs::write(target_dir.join(".npmrc"), "script-shell=bash\n")?;
+    // update-notifier=false suppresses "npm notice" upgrade messages that cause
+    // test flakes
+    fs::write(
+        target_dir.join(".npmrc"),
+        "script-shell=bash\nupdate-notifier=false\n",
+    )?;
 
     git(&["add", "."])?;
     git(&["commit", "-m", "Initial", "--quiet"])?;
--- a/crates/turborepo/tests/dry_json.rs
+++ b/crates/turborepo/tests/dry_json.rs
@@ -2,51 +2,18 @@ mod common;
 
 use std::path::Path;
 
-use common::setup;
+use common::{run_turbo, setup};
 use serde_json::Value;
 
 fn turbo_dry_json(test_dir: &Path, args: &[&str]) -> Result<Value, anyhow::Error> {
-    // Config dir must be outside test_dir so turbo doesn't pick up telemetry
-    // files as task inputs.
-    let config_dir = tempfile::tempdir()?;
-
-    let mut cmd = assert_cmd::Command::cargo_bin("turbo")?;
-    cmd.env("TURBO_TELEMETRY_MESSAGE_DISABLED", "1")
-        .env("TURBO_GLOBAL_WARNING_DISABLED", "1")
-        .env("TURBO_PRINT_VERSION_DISABLED", "1")
-        .env("TURBO_CONFIG_DIR_PATH", config_dir.path())
-        .env("DO_NOT_TRACK", "1")
-        .env_remove("CI")
-        .env_remove("GITHUB_ACTIONS")
-        .current_dir(test_dir);
-
-    for arg in args {
-        cmd.arg(arg);
-    }
-
-    let output = cmd.output()?;
+    let output = run_turbo(test_dir, args);
     let stdout = String::from_utf8_lossy(&output.stdout);
     let json: Value = serde_json::from_str(&stdout)?;
     Ok(json)
 }
 
 fn turbo_dry_json_expect_failure(test_dir: &Path, args: &[&str]) -> String {
-    let config_dir = tempfile::tempdir().unwrap();
-    let mut cmd = assert_cmd::Command::cargo_bin("turbo").unwrap();
-    cmd.env("TURBO_TELEMETRY_MESSAGE_DISABLED", "1")
-        .env("TURBO_GLOBAL_WARNING_DISABLED", "1")
-        .env("TURBO_PRINT_VERSION_DISABLED", "1")
-        .env("TURBO_CONFIG_DIR_PATH", config_dir.path())
-        .env("DO_NOT_TRACK", "1")
-        .env_remove("CI")
-        .env_remove("GITHUB_ACTIONS")
-        .current_dir(test_dir);
-
-    for arg in args {
-        cmd.arg(arg);
-    }
-
-    let output = cmd.output().unwrap();
+    let output = run_turbo(test_dir, args);
     assert!(!output.status.success());
     String::from_utf8_lossy(&output.stderr).to_string()
 }
--- a/crates/turborepo/tests/run_caching.rs
+++ b/crates/turborepo/tests/run_caching.rs
@@ -2,27 +2,7 @@ mod common;
 
 use std::{fs, path::Path};
 
-use common::{run_turbo, setup};
-
-fn run_turbo_with_env(dir: &Path, args: &[&str], env: &[(&str, &str)]) -> std::process::Output {
-    let config_dir = tempfile::tempdir().unwrap();
-    let mut cmd = assert_cmd::Command::cargo_bin("turbo").unwrap();
-    cmd.env("TURBO_TELEMETRY_MESSAGE_DISABLED", "1")
-        .env("TURBO_GLOBAL_WARNING_DISABLED", "1")
-        .env("TURBO_PRINT_VERSION_DISABLED", "1")
-        .env("TURBO_CONFIG_DIR_PATH", config_dir.path())
-        .env("DO_NOT_TRACK", "1")
-        .env_remove("CI")
-        .env_remove("GITHUB_ACTIONS")
-        .current_dir(dir);
-    for (k, v) in env {
-        cmd.env(k, v);
-    }
-    for arg in args {
-        cmd.arg(arg);
-    }
-    cmd.output().unwrap()
-}
+use common::{run_turbo, run_turbo_with_env, setup};
 
 fn git_commit(dir: &Path, msg: &str) {
     std::process::Command::new("git")
--- a/crates/turborepo/tests/run_logging.rs
+++ b/crates/turborepo/tests/run_logging.rs
@@ -1,29 +1,6 @@
 mod common;
 
-use std::path::Path;
-
-use common::{run_turbo, setup, turbo_output_filters};
-
-fn run_turbo_with_env(dir: &Path, args: &[&str], env: &[(&str, &str)]) -> std::process::Output {
-    let config_dir = tempfile::tempdir().unwrap();
-    let mut cmd = assert_cmd::Command::cargo_bin("turbo").unwrap();
-    cmd.env("TURBO_TELEMETRY_MESSAGE_DISABLED", "1")
-        .env("TURBO_GLOBAL_WARNING_DISABLED", "1")
-        .env("TURBO_PRINT_VERSION_DISABLED", "1")
-        .env("TURBO_CONFIG_DIR_PATH", config_dir.path())
-        .env("DO_NOT_TRACK", "1")
-        .env("NPM_CONFIG_UPDATE_NOTIFIER", "false")
-        .env_remove("CI")
-        .env_remove("GITHUB_ACTIONS")
-        .current_dir(dir);
-    for (k, v) in env {
-        cmd.env(k, v);
-    }
-    for arg in args {
-        cmd.arg(arg);
-    }
-    cmd.output().unwrap()
-}
+use common::{run_turbo, run_turbo_with_env, setup, turbo_output_filters};
 
 // --- log-order-stream.t ---
 // Stream output is non-deterministic in ordering, so we check key lines.
@@ -128,22 +105,11 @@ fn test_log_order_github_actions() {
     let tempdir = tempfile::tempdir().unwrap();
     setup::setup_integration_test(tempdir.path(), "ordered", "npm@10.5.0", true).unwrap();
 
-    // Need to SET GITHUB_ACTIONS=1 for this test
-    let config_dir = tempfile::tempdir().unwrap();
-    let mut cmd = assert_cmd::Command::cargo_bin("turbo").unwrap();
-    let output = cmd
-        .env("TURBO_TELEMETRY_MESSAGE_DISABLED", "1")
-        .env("TURBO_GLOBAL_WARNING_DISABLED", "1")
-        .env("TURBO_PRINT_VERSION_DISABLED", "1")
-        .env("TURBO_CONFIG_DIR_PATH", config_dir.path())
-        .env("DO_NOT_TRACK", "1")
-        .env("NPM_CONFIG_UPDATE_NOTIFIER", "false")
-        .env("GITHUB_ACTIONS", "1")
-        .env_remove("CI")
-        .args(["run", "build", "--force"])
-        .current_dir(tempdir.path())
-        .output()
-        .unwrap();
+    let output = run_turbo_with_env(
+        tempdir.path(),
+        &["run", "build", "--force"],
+        &[("GITHUB_ACTIONS", "1")],
+    );
 
     let stdout = String::from_utf8_lossy(&output.stdout);
     assert!(stdout.contains("::group::my-app:build"));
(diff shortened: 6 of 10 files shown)


crates/turborepo-otel/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
```
bkonkle, line 1: I'm working on updating this to use workspace versions where available and to use turborepo-errors and not thiserror.

bkonkle: I misunderstood turborepo-errors, and I don't think it's needed here. I'll use thiserror.

bkonkle:
Some examples of stats captured locally:

image

image

image

bkonkle:
An individual record:

```
turbo_task_duration_ms_sum{
	env="local",
	exported_job="turborepo",
	instance="otel-collector:8889",
	job="otel-collector",
	otel_scope_name="turborepo",
	service_name="turborepo",
	turbo_run_exit_code="0",
	turbo_run_id="35fNfU4zCmTGgjYkCR3w39UhqQU",
	turbo_scm_branch="bkonkle/otel-exporter",
	turbo_scm_revision="8e02a250356995d73103892ba7ff49598ff381a2",
	turbo_task_cache_source="LOCAL",
	turbo_task_cache_status="hit",
	turbo_task_cache_time_saved_ms="2172", 
	turbo_task_command="prettier -c . --cache --ignore-path=../../.prettier-ignore", 
	turbo_task_exit_code="0", 
	turbo_task_external_inputs_hash="460143f52d4d621f", 
	turbo_task_hash="bd353948bc7b2aa3", 
	turbo_task_id="@turbo/test-utils#lint:prettier", 
	turbo_task_name="lint:prettier", 
	turbo_task_package="@turbo/test-utils", 
	turbo_version="2.6.1",
} 34
```

This is a `lint:prettier` task for `@turbo/test-utils` with a cache hit that executed in 34ms.

docs/site/content/docs/reference/configuration.mdx
```diff
@@ -837,3 +837,191 @@ Must start with `team_` or it will not be used.
...
+  "futureFlags": {
+    "experimentalObservability": true
+  },
+  "experimentalObservability": {
+    "otel": {
+      "enabled": true,
+      "protocol": "grpc",
+      "endpoint": "https://api.datadoghq.com/api/v2/otlp",
+      "headers": {
+        "DD-API-KEY": "your-api-key"
```
anthonyshew: ⚠️ This would be committing a secret to source control, which we don't want.

docs/site/content/docs/reference/configuration.mdx
```diff
@@ -837,3 +837,191 @@ Must start with `team_` or it will not be used.
...
+#### `experimentalObservability.otel.headers`
+
+Optional HTTP headers to include with export requests. Useful for authentication (e.g., API keys) or custom metadata.
+
+```jsonc title="./turbo.json"
+{
+  "experimentalObservability": {
+    "otel": {
+      "headers": {
+        "DD-API-KEY": "your-datadog-api-key",
```
anthonyshew: https://github.com/vercel/turborepo/pull/11130/changes#r2729260907

crates/turborepo-run-summary/src/observability/otel.rs
```diff
@@ -0,0 +1,454 @@
...
+        TaskCacheStatus::Miss => None,
+    };
+
+    TaskMetricsPayload {
+        task_id: task.task_id.to_string(),
+        task: task.task.clone(),
+        package: task.package.clone(),
+        hash: task.shared.hash.clone(),
+        external_inputs_hash: task.shared.hash_of_external_dependencies.clone(),
+        command: task.shared.command.clone(),
```
anthonyshew, line 199: There is a minor risk here that sensitive values could be sent into a collector. Take, for instance, a command like:
```
turbo run build -- --some-value=$MY_SECRET_TOKEN
```

Have we checked what this would capture? The shell-injected string or the raw input? I think we need to constrain to the raw input.

crates/turborepo-otel/src/lib.rs
```diff
@@ -0,0 +1,635 @@
...
+
+impl std::fmt::Debug for Handle {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Handle").finish_non_exhaustive()
+    }
+}
+
+impl Handle {
+    pub fn try_new(config: Config) -> Result<Self, Error> {
+        if config.endpoint.trim().is_empty() {
```
anthonyshew: This empty string check could be a bit more robust. The OTEL spec [only accepts gRPC and HTTP transports](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#protocol-details), so we could reject anything else with a clear error rather than whatever undefined behavior lies downstream.

Once, that's implemented, let's add a test or two for that validation as well.

crates/turborepo-run-summary/src/tracker.rs
```diff
@@ -313,6 +350,16 @@ impl<'a> RunSummary<'a> {
...
         is_watch: bool,
     ) -> Result<(), Error> {
+        // Handle observability shutdown before the dry run check to ensure graceful
+        // cleanup even when metrics are not being emitted.
+        if let Some(handle) = self.observability_handle.take() {
+            // Only record metrics for actual runs, not dry runs
+            if !matches!(self.run_type, RunType::DryJson | RunType::DryText) {
+                handle.record(&self);
+            }
+            handle.shutdown();
```
anthonyshew, line 403: I might be missing it because its covered elsewhere but do we have any timeout or retry behavior? It's possible for this shutdown call to get held up by dangling network, and I want to make sure we don't leave CI hanging.

crates/turborepo-otel/src/lib.rs
```diff
@@ -0,0 +1,635 @@
...
+            let mut attrs = base_attrs.clone();
+            attrs.push(KeyValue::new("turbo.task.id", task.task_id.clone()));
+            attrs.push(KeyValue::new("turbo.task.name", task.task.clone()));
+            attrs.push(KeyValue::new("turbo.task.package", task.package.clone()));
+            attrs.push(KeyValue::new("turbo.task.hash", task.hash.clone()));
+            attrs.push(KeyValue::new(
+                "turbo.task.external_inputs_hash",
+                task.external_inputs_hash.clone(),
+            ));
+            attrs.push(KeyValue::new("turbo.task.command", task.command.clone()));
```
anthonyshew: The hashes are at an unbounded cardinality - and I'm not sure what analytical value they bring. Aggregations over these would generally be meaningless since there's no meaning in the hash value.

What do you think?

anthonyshew: Does the same problem also exist for turbo.task.id?

`turbo.scm.revision` seems high cardinality but has a genuine use case of aggregating over a hash, right? `turbo.scm.branch` would be lower cardinality...But seems like users will want to aggregate over both of these.

anthonyshew: Could we use [resource attributes](https://opentelemetry.io/docs/specs/otel/resource/sdk/) or [examplars](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#exemplars) meaningfully here? One consideration is that different OTEL backends handle these differently, so the helpfulness is a bit undefined.

anthonyshew: (Sorry, captured the `turbo.task.command` line in this comment thread but didn't mean to - and now I've got a whole thread with myself going. 😄 Please ignore.)

crates/turborepo-lib/src/run/builder.rs
```diff
@@ -484,6 +485,32 @@ impl RunBuilder {
...
+        // Warn if observability config is present but the feature flag is not enabled
+        if let Some(obs_opts) = &self.opts.experimental_observability {
+            if obs_opts.otel.is_some() && !self.opts.future_flags.experimental_observability {
+                tracing::warn!(
+                    "experimentalObservability.otel is configured but \
+                     futureFlags.experimentalObservability is not enabled in turbo.json. The \
+                     observability config will be ignored."
+                );
+            }
+        }
```
anthonyshew: The `futureFlags` are meant to be hard gates, not warnings. Easy change!

crates/turborepo-otel/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+  "metrics",
+  "rt-tokio",
+  "experimental_async_runtime",
+  "experimental_metrics_periodicreader_with_async_runtime",
+] }
+serde = { workspace = true, optional = true }
+serde_json = { workspace = true, optional = true }
+thiserror = { workspace = true }
+tokio = { workspace = true, features = ["full"] }
+tonic = "0.14"
```
anthonyshew, line 25: Ultra-nit: This is a different version of tonic than is used in `turborepo-lib`. Would love to use a consistent version if we can. @anthonyshew, can you bump the tonic version in `turborepo-lib` forward to accommodate?

crates/turborepo-otel/src/lib.rs
```diff
@@ -0,0 +1,635 @@
...
+    runtime::Tokio,
+};
+use opentelemetry_semantic_conventions::resource::SERVICE_NAME;
+use thiserror::Error;
+use tonic::metadata::{MetadataKey, MetadataMap, MetadataValue};
+use tracing::warn;
+
+/// Protocol supported by the OTLP exporter.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Protocol {
```
anthonyshew, line 107: Is this is duplicated in crates/turborepo-config/src/experimental_otel.rs:24-31?

crates/turborepo-otel/src/lib.rs
```diff
@@ -0,0 +1,635 @@
...
+            if !config.headers.is_empty() {
+                let headers: HashMap<_, _> = config.headers.clone().into_iter().collect();
+                builder = builder.with_headers(headers);
+            }
+            builder.build().map_err(Error::Exporter)?
+        }
+    };
+
+    let reader = periodic_reader_with_async_runtime::PeriodicReader::builder(exporter, Tokio)
+        .with_interval(Duration::from_secs(15))
```
anthonyshew: This is hardcoded - which might be fine. Any reason we should make it configurable?

.github/workflows/turborepo-test.yml
```diff
@@ -482,6 +482,12 @@ jobs:
 
       - name: Run tests
         run: |
+          # macOS ARM runners have limited memory (~7GB). Limit parallel crate
+          # compilation and reduce rustc frontend threads to prevent OOM.
+          if [ "${{ matrix.os.name }}" == "macos" ]; then
+            export CARGO_BUILD_JOBS=2
+            export RUSTFLAGS='--cfg tokio_unstable -Zshare-generics=y -Zthreads=4 -Csymbol-mangling-version=v0'
+          fi
```
anthonyshew: I'd love to believe that this change is not required. 😄

bkonkle: Unfortunately, it was consistently failing for me until I iterated to this solution. 😭 I'm glad to remove the workaround and see if you can spot something I missed? 😅

crates/turborepo-run-summary/src/observability/otel.rs
bkonkle: Good callout! I'll investigate.

crates/turborepo-otel/src/lib.rs
bkonkle: I've got a `validate_endpoint()` function now in 150957a that checks to make sure it's http or https, case-insensitive. I thought `grpc:` was a valid protocol, but Claude pointed out that I was wrong. 😅 There are tests now to validate. 👍

crates/turborepo-run-summary/src/observability/otel.rs
bkonkle: It looks like `task.shared.command` itself comes from the `package.json`, [pulled from](https://github.com/vercel/turborepo/blob/da11b449f0d89f1c64e8ba1ca11e146cb2414800/crates/turborepo-run-summary/src/task_factory.rs#L104-L110) the `scripts` property. So, it should be the plain string and shouldn't be the interpolated value.

As for `--` passthrough arguments, they go through [this property](https://github.com/vercel/turborepo/blob/bkonkle/otel-exporter/crates/turborepo-run-summary/src/task_factory.rs#L166) instead (`task.shared.cli_arguments`), and that isn't currently captured in the `observability/otel.rs` module anywhere yet. I think if they _were_ then this concern would apply - the shell would likely substitute before `turbo` saw it.

crates/turborepo-run-summary/src/tracker.rs
bkonkle: There's an `experimentalOtel.timeout_ms` setting in `turbo.json` that is used by the `shutdown()`. The flow is a little convoluted, so I had Claude trace it out:

turbo.json `timeout_ms` (or env/CLI)
    → Config.timeout
    → ExportConfig.timeout  
    → MetricExporter (respects timeout on each export)
    → PeriodicReader (uses exporter for shutdown flush)
    → SdkMeterProvider.shutdown() (triggers final export with timeout)

I added a comment in 150957a to indicate that the `timeout_ms` is used.

crates/turborepo-otel/src/lib.rs
bkonkle: Definitely a good thing to think through. I think yes, `revision` and `branch` are useful even with the high cardinality, but maybe a flag to turn them off if desired? The `hash` and `external_inputs_hash` and `task.id` attributes might be too random, though, and should be opt-in instead? 🤔

crates/turborepo-lib/src/run/builder.rs
bkonkle: I did some thinking around this last year and asked Claude to look at prior `futureFlag` usages, but I wasn't confident and thought you might call this out. 😄 This warn should be unreachable if the future flag is disabled because of the hard gate [here](https://github.com/vercel/turborepo/blob/da11b449f0d89f1c64e8ba1ca11e146cb2414800/crates/turborepo-config/src/turbo_json.rs#L92-L104)), so the warning only applies when the config comes from CLI/env vars (which typically bypass the _turbo.json_ gate since they're already prefixed with `EXPERIMENTAL_`), not from _turbo.json_.

Should I remove the warning anyway to eliminate confusion, or just add a note to the comments indicating when it would actually be used?

crates/turborepo-otel/Cargo.toml
bkonkle: @anthonyshew Were you able to look into `turborepo-lib`, or should I start a new PR for it (as a prerequisite for this one)?

crates/turborepo-otel/src/lib.rs
bkonkle: I did it somewhat intentionally because I didn't want `turborepo-config` to have to depend on `turbrepo-otel` or vice-versa, with the conversion in `turborepo-run-summary` bridging the gap and re-coupling them. I guess it makes the most sense, though, to go ahead and conditionally depend on `turborepo-otel` from within `turborepo-config` if the build flag for `otel` isn't disabled.

bkonkle: Oh - no, that would require pulling `clap` in do to the config parsing based on the canonical type, which is a heavy dependency.

bkonkle: I went with this approach in 150957a:

```
turborepo-otel (canonical Protocol with serde + FromStr)
    ↑
turborepo-config (re-exports as ExperimentalOtelProtocol, clap uses FromStr)
    ↑
turborepo-run-summary (uses unified type directly)
```

crates/turborepo-otel/src/lib.rs
bkonkle: I don't know how often people would actually configure it, but since `timeout_ms` is already configurable this probably should be also. 👍 I'll update it.

docs/site/content/docs/reference/configuration.mdx
bkonkle: I updated the example in 150957a

docs/site/content/docs/reference/configuration.mdx
bkonkle: I updated the example in 150957a

.github/workflows/turborepo-test.yml
bkonkle: You were right! It's not actually needed. I'll remove it entirely - hopefully we'll get another consistent green. :-)

CONTRIBUTING.md
```diff
@@ -64,6 +64,16 @@ and `rustls-tls` features.
...
+### OpenTelemetry Observability
+
+Turborepo includes optional OpenTelemetry (OTel) support for exporting metrics. OTel is an experimental feature and is **not** included in default builds. If you wish to build `turbo` with OTel support, you can enable it by running:
+
+```bash
+cargo build -p turbo --features otel
+```
+
+Note that when OTel is disabled at compile time (the default), the `experimentalObservability` configuration in `turbo.json` and related CLI flags will be ignored (they will not error, but no metrics will be exported). IDEs using the devcontainer configuration will have OTel enabled automatically for development purposes.
+
```
anthonyshew, line 79: OTEL is a default feature in crates/turborepo/Cargo.toml, right? I don't feel a need to feature flag it in the Rust configs, just the `turbo.json` `futureFlag`.

crates/turborepo-otel/src/lib.rs
anthonyshew: I think lets:
- Be conservative to start with
- But include what we know we need for our purposes

We'll learn if we haven't exposed enough with our own dogfooding, early users, experimental users, etc.

The classic "It's easier to add more than take away..." 😄

crates/turborepo-run-summary/src/observability/otel.rs
anthonyshew: Sounds good to me. I'd say let's make a TODO to make sure that that passthrough args surface area is potentially sensitive.

crates/turborepo-config/src/lib.rs
```diff
@@ -254,6 +263,18 @@ pub struct ConfigurationOptions {
...
+}
+
+pub use crate::experimental_otel::{
+    ExperimentalOtelMetricsOptions, ExperimentalOtelOptions, ExperimentalOtelProtocol,
+};
+
+#[derive(Deserialize, Serialize, Default, Debug, Clone, PartialEq, Eq, Merge)]
+#[serde(rename_all = "camelCase")]
+pub struct ExperimentalObservabilityOptions {
+    pub otel: Option<ExperimentalOtelOptions>,
```
anthonyshew: Note: We should check the merging behavior of flags vs. env vars vs turbo.json.

It's really easy to mess up the precedence due to the way our code is written (TODO).

crates/turborepo-otel/Cargo.toml
anthonyshew: Nabbed it! https://github.com/vercel/turborepo/pull/11820

crates/turborepo-run-summary/src/observability/otel.rs
bkonkle: I'll add a TODO comment saying “do not export cli_arguments without redaction policy”. 👍

crates/turborepo-otel/src/lib.rs
bkonkle: Sounds good. I'm making `id`, `hash`, and `external_inputs_hash` opt-in by adding a new config block:

```
{
  "experimentalObservability": {
    "otel": {
      "metrics": {
        "runSummary": true,
        "taskDetails": true,
        "taskAttributes": {
          "id": false,
          "hashes": false
        }
      }
    }
  }
}
```

crates/turborepo-lib/src/run/builder.rs
bkonkle: I turned this into a hard error in 3c7c5f80b80cdff0b280ea4f9a5ac95124ff8326

crates/turborepo-run-summary/src/observability/otel.rs
bkonkle: Done in 743f2a8efe5979ac14957bceb2eeec846ba8257b

crates/turborepo-otel/src/lib.rs
bkonkle: Done in 743f2a8efe5979ac14957bceb2eeec846ba8257b

CONTRIBUTING.md
bkonkle: I thought I had already enabled it by default, but I had not. Fixed in 1ae9456680

crates/turborepo-config/src/lib.rs
bkonkle: Good idea - I'll try writing a quick precedence test.

bkonkle: I implemented this with some AI help in f75784c - is the hierarchy there correct?

I also made HTTPS required and added some token guidance in 6390c70

Thanks!

anthonyshew:
The failure in `windows (partition 7/10)` is a known issue, caused by recent changes to our integration test suite. Merging past, as https://github.com/vercel/turborepo/pull/12039 is meant to fix it.

Issue #10940, opened 2025-10-07
noUpdateNotifier does not supress update notifications


maclockard:
### Verify canary release

- [x] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

Run turbo with an out of date turbo version

### Which canary version will you have in your reproduction?

I checked the box, but inherit to the bug I can't check that it works as expected with the latest build, so I used version 2.5.6

### Environment information

```block

```

### Expected behavior

Don't print update notification if `noUpdateNotifier` is true

### Actual behavior

Prints update notification no matter what

### To Reproduce

* Add `"noUpdateNotifier": true` to `turbo.json`
* Run turbo with an out of date turbo version

mootari:
We're seeing the same with turborepo 2.5.6.

mootari:
Seems there's already an unlinked draft PR: #10941

DevaanshKathuria:
Hey @maclockard

I’ve opened a pull request that resolves this issue.
The PR ensures that the noUpdateNotifier flag correctly suppresses update notifications when set to true, restoring the expected behavior described here.

Here’s the link to the PR: #11133 

Please let me know if any adjustments or refinements are needed

theoludwig:
This is still an issue, even with latest release (e.g: v2.7.2).

mcmxcdev:
Can confirm, this is still an issue, using turbo 2.7.2 and `"noUpdateNotifier": true`

GuillaumeLagrange:
Reproduced with turbo 2.7.3, got a notification to upgrade to 2.7.4

```
$ pnpm turbo --version
2.7.3
```

PR #11133, opened 2025-11-18
fix: Correct version mismatch causing noUpdateNotifier to fail

DevaanshKathuria:
This pull request fixes a bug where setting "noUpdateNotifier": true in turbo.json did not suppress update notifications.
Turbo continued printing out-of-date version warnings regardless of the flag’s value.
What was happening
Even when users explicitly disabled update notifications using:
`{
  "noUpdateNotifier": true
}`
Turbo still displayed upgrade messages when the installed version was behind the latest available version.
This violated the expected behavior described in Issue #10940.
What this PR does:
- Ensures that when noUpdateNotifier is set to true, Turbo skips the update-notifier check entirely.
- Prevents unnecessary noise in CI pipelines and automated scripts.
- Restores expected behavior and aligns with Turbo’s configuration contract.

How I verified the fix:
Added "noUpdateNotifier": true to turbo.json.
Ran Turbo using an intentionally outdated version.
Confirmed that no update notification was printed.
Verified that notifications still appear when the flag is not set.

Linked Issue
Fixes: #10940

<sub>CLOSES TURBO-4878</sub>

### DIFF
--- a/crates/turborepo-lib/src/query/external_package.rs
+++ b/crates/turborepo-lib/src/query/external_package.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use async_graphql::Object;
 
+use super::{package::Package, Array, Error};
 use crate::run::Run;
 
 #[derive(Clone)]
@@ -29,5 +30,21 @@ impl ExternalPackage {
 impl ExternalPackage {
     async fn name(&self) -> String {
         self.human_name().to_string()
+    }
+
+    async fn internal_dependents(&self) -> Result<Array<Package>, Error> {
+        let Some(names) = self
+            .run
+            .pkg_dep_graph()
+            .internal_dependencies_for_external_dependency(&self.package)
+        else {
+            return Ok(Array::from(Vec::new()));
+        };
+        let mut packages = names
+            .iter()
+            .map(|name| Package::new(self.run.clone(), name.as_package_name().clone()))
+            .collect::<Result<Array<_>, Error>>()?;
+        packages.sort_by(|a, b| a.get_name().cmp(b.get_name()));
+        Ok(packages)
     }
 }
--- a/crates/turborepo-lib/src/query/mod.rs
+++ b/crates/turborepo-lib/src/query/mod.rs
@@ -624,6 +624,18 @@ impl RepositoryQuery {
 
         Ok(packages)
     }
+
+    async fn external_dependencies(&self) -> Result<Array<ExternalPackage>, Error> {
+        let pkg_dep_graph = self.run.pkg_dep_graph();
+        let all_package_names: Vec<_> = pkg_dep_graph.packages().map(|(name, _)| name).collect();
+        let mut packages = pkg_dep_graph
+            .transitive_external_dependencies(all_package_names)
+            .into_iter()
+            .map(|pkg| ExternalPackage::new(self.run.clone(), pkg.clone()))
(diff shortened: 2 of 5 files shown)


DevaanshKathuria:
Hey @tknickman, I hope you're doing well. Whenever you get a moment, could you please take a look at the PR and share your feedback on whether it’s ready to be merged? Thanks!

DevaanshKathuria:
Hi @tknickman! Just checking in on the PR I submitted a while ago.
Whenever you get a moment, please take a look. Thanks!

DevaanshKathuria:
Hey @anthonyshew! If you get a little time, could you please review this PR for me?  
My assigned reviewer seems busy, so any help would be great. Thanks a lot!

crates/turborepo-lib/src/shim/local_turbo_config.rs
```diff
@@ -30,10 +30,7 @@ impl LocalTurboConfig {
     fn infer_internal(repo_state: &RepoState, is_enabled: Option<bool>) -> Option<Self> {
         // TODO: once we have properly communicated this functionality we should make
         // this opt-out.
-        if !is_enabled.unwrap_or(false) {
-            debug!("downloading correct local version not enabled");
-            return None;
-        }
+        debug!("FORCING local turbo binary. Download disabled.");
```
anthonyshew: This change isn't making sense. Was this intentional? If not, let's revert it.

crates/turborepo-lib/src/config/mod.rs
```diff
@@ -320,6 +320,7 @@ pub struct ConfigurationOptions {
     pub(crate) allow_no_turbo_json: Option<bool>,
     pub(crate) tui_scrollback_length: Option<u64>,
     pub(crate) concurrency: Option<String>,
+    #[serde(rename = "noUpdateNotifier")]
```
anthonyshew: This is redundant. The struct already has `rename_all = "camelCase"` on it.

```suggestion
```

crates/turborepo-lib/src/shim/mod.rs
```diff
@@ -341,6 +341,7 @@ fn try_check_for_updates(
             None,
             interval,
             package_manager,
+            config.no_update_notifier.unwrap_or(false)
```
anthonyshew: The getter method is already defined at config/mod.rs:480-482 and provides the same logic. Let's use that.

crates/turborepo-lib/src/shim/local_turbo_config.rs
DevaanshKathuria: I've reverted the change

DevaanshKathuria:
Hey @anthonyshew, I've made all the required changes. Please review them and let me know if any further changes are needed.

anthonyshew (changes requested):
The PR now has no functional changes.
CleanShot 2025-12-06 at 06 18 44@2x

DevaanshKathuria:
@anthonyshew, I’m really sorry that I didn’t compare the latest commit with the original code. I can try the other option. How about I remove the outer guard so the updater can use its own logic, as you suggested?

DevaanshKathuria:
Hi @anthonyshew! Just checking in on the PR.
Whenever you get a moment, please take a look. Thanks!

anthonyshew:
#10941 was already working on this but I'll pick this work back up from this one.

PR #11135, opened 2025-11-18
fix: Add fine grained interruptible task restarts in watch mode

johnpyp:
### Description

Fixes https://github.com/vercel/turborepo/issues/9421

This PR refactors watch mode task restarts to be "fine-grained", such that if a package changes, only active tasks which rely on that output get killed and restarted.

(Similar to https://github.com/vercel/turborepo/pull/10846, but taking a slightly different approach which I think is cleaner).

### Testing Instructions

Manual testing confirms the issue is fixed with this change. There aren't any existing automated tests for `watch` and the like, given that it's a long running process I assume.

### DIFF
--- a/crates/turborepo-lib/src/run/watch.rs
+++ b/crates/turborepo-lib/src/run/watch.rs
@@ -4,7 +4,7 @@ use std::{
     sync::{Arc, Mutex},
 };
 
-use futures::StreamExt;
+use futures::{future::join_all, StreamExt};
 use miette::{Diagnostic, SourceSpan};
 use thiserror::Error;
 use tokio::{select, sync::Notify, task::JoinHandle};
@@ -211,9 +211,10 @@ impl WatchClient {
                     // Clean up currently running tasks
                     self.active_runs.retain(|h| !h.run_task.is_finished());
 
-                    match &mut changed_packages {
+                    match changed_packages {
                         ChangedPackages::Some(pkgs) => {
-                            self.stop_impacted_tasks(pkgs).await;
+                            let impacted = self.stop_impacted_tasks(&pkgs).await;
+                            changed_packages = ChangedPackages::Some(impacted);
                         }
                         ChangedPackages::All => {
                             for handle in self.active_runs.drain(..) {
@@ -275,7 +276,7 @@ impl WatchClient {
         Ok(())
     }
 
-    async fn stop_impacted_tasks(&self, pkgs: &mut HashSet<PackageName>) {
+    async fn stop_impacted_tasks(&self, pkgs: &HashSet<PackageName>) -> HashSet<PackageName> {
         let engine = self.run.engine();
         let mut tasks_to_stop = HashSet::new();
 
@@ -293,17 +294,20 @@ impl WatchClient {
             }
         }
 
-        let mut impacted_packages = HashSet::new();
-        for task_id in &tasks_to_stop {
-            impacted_packages.insert(PackageName::from(task_id.package()));
-        }
-
-        *pkgs = impacted_packages;
+        let impacted_packages: HashSet<_> = tasks_to_stop
+            .iter()
+            .map(|task_id| PackageName::from(task_id.package()))
+            .collect();
 
         let task_ids: Vec<_> = tasks_to_stop.into_iter().collect();
-        for handle in &self.active_runs {
-            handle.stopper.stop_tasks(&task_ids).await;
-        }
+        join_all(
+            self.active_runs
+                .iter()
+                .map(|handle| handle.stopper.stop_tasks(&task_ids)),
+        )
+        .await;
+
+        impacted_packages
     }
 
     /// Shut down any resources that run as part of watch.
--- a/crates/turborepo-process/src/child.rs
+++ b/crates/turborepo-process/src/child.rs
@@ -690,12 +690,12 @@ impl Child {
 
     /// Mark this child as being stopped as part of a ProcessManager shutdown
     pub fn set_closing(&self) {
-        self.closing.store(true, Ordering::SeqCst);
+        self.closing.store(true, Ordering::Release);
     }
 
     /// Check if this child was stopped as part of a ProcessManager shutdown
     pub fn is_closing(&self) -> bool {
-        self.closing.load(Ordering::SeqCst)
+        self.closing.load(Ordering::Acquire)
     }
 }
 


crates/turborepo-process/src/lib.rs
```diff
@@ -174,9 +180,47 @@ impl ProcessManager {
...
+            // We don't set is_closing = true here because we want to allow new spawns
+
+            let matching: Vec<_> = lock
+                .children
+                .iter()
+                .filter(|c| predicate(c))
+                .cloned()
+                .collect();
+
+            lock.children.retain(|c| !predicate(c));
```
anthonyshew: Between the evaluation in the `.filter()` and the `.retain()` a child's state might change. We should do something more like:
```
  let (matching, remaining): (Vec<_>, Vec<_>) = lock.children
      .drain(..)
      .partition(|c| predicate(c));
  lock.children = remaining;
```

crates/turborepo-lib/src/task_graph/visitor/exec.rs
```diff
@@ -455,7 +463,17 @@ impl ExecContext {
             // Something else killed the child
             ChildExit::KilledExternal => Err(InternalError::ExternalKill),
             // The child was killed by turbo indicating a shutdown
-            ChildExit::Killed | ChildExit::Interrupted => Ok(ExecOutcome::Shutdown),
+            ChildExit::Killed | ChildExit::Interrupted => {
+                // We distinguish between a full shutdown and a restart based on whether the
+                // process manager is closing. If it is closing, it means we are shutting down
+                // the entire run. If it is not closing, it means we are restarting specific
+                // tasks.
+                if self.manager.is_closing() {
```
anthonyshew: This is potentially dangerous based on race conditions and dependent task behavior. What tests can we write to validate that we're not creating bugs here?

anthonyshew: Example Race:
  ```
  Thread 1 (exec.rs):        Thread 2 (manager):
  child exits
  is_closing() → false
                             close() called
                             is_closing = true
  return Restarted ❌
  ```
 That should've been `Shutdown`.

crates/turborepo-process/src/child.rs
```diff
@@ -384,20 +385,33 @@ pub struct Child {
 }
 
 #[derive(Clone, Debug)]
-pub struct ChildCommandChannel(mpsc::Sender<ChildCommand>);
+pub struct ChildCommandChannel {
+    sender: mpsc::Sender<ChildCommand>,
+    task_id: Option<TaskId<'static>>,
```
anthonyshew: This is a little rough, because now the child process has to be aware of `task_id`s . These channels shouldn't be aware of "business logic" ideally, to preserve encapsulation.

I'm also worried about the lifetime implications. The `TaskId` is going to live the entire program, but line 433 has a `.clone()` that can end up referencing tasks that are non-static. That could result in panics.

Another small, small worry I have, but a worry nonethless, is the memory overhead associated with this strategy. `TaskId`s aren't dropped until the `Child` is dropped, so its _possible_ to see some memory in high concurrency scenarios. (I haven't taken the time to validate this, but based on my other comments, I'd love to see an architectural change where this is no longer a concern).

crates/turborepo-lib/src/task_graph/visitor/exec.rs
anthonyshew: If we introduce some atomics, this concern would get addressed.

crates/turborepo-process/src/child.rs
johnpyp: Hmm fair enough. Is there a better identifier or way of getting the "set of processes mapped to tasks" that would allow us to not have the process code care about tasks at all?

That was the primary thing I wasn't confident in going into this change without being too familiar with the codebase.

anthonyshew: After sleeping on this, I'm thinking the best way (still not great) we can go about this is if `ProcessManager` gets indices into `Vec<Child>`.  Some pseudocoding: 

```
#[derive(Debug)]
struct ProcessManagerInner {
    is_closing: bool,
    
    // None = inactive/dead process slot
    children: Vec<Option<Child>>,
    
    // Maps task_id -> indices in the children vec
    // We keep dead indices here and filter them on access
    task_index: HashMap<TaskId<'static>, Vec<usize>>,
    
    size: Option<PtySize>,
}

impl ProcessManagerInner {
    fn is_index_active(&self, idx: usize) -> bool {
        self.children.get(idx)
            .and_then(|opt| opt.as_ref())
            .is_some()
    }
    
    fn get_active_children_for_task(&self, task_id: &TaskId) -> Vec<Child> {
        self.task_index
            .get(task_id)
            .map(|indices| {
                indices.iter()
                    .filter_map(|&idx| {
                        self.children.get(idx)
                            .and_then(|opt| opt.as_ref())
                            .cloned()
                    })
                    .collect()
            })
            .unwrap_or_default()
    }
}
```

Now when a process exits, we don't need to update task_index immediately, it will get filtered.
```
// Mark index as dead - just set to None
lock.children[3] = None;
```

Stopping a task now looks like:
```
pub async fn stop_tasks(&self, task_ids: &[TaskId<'static>]) {
    let children_to_stop = {
        let mut lock = self.state.lock().unwrap();
        
        let mut children = Vec::new();
        for task_id in task_ids {
            if let Some(indices) = lock.task_index.get(task_id) {
                for &idx in indices {
                    // Check if still active
                    if let Some(Some(child)) = lock.children.get(idx) {
                        children.push(child.clone());
                        // Mark as inactive
                        lock.children[idx] = None;
                    }
                }
            }
            // Remove task from index
            lock.task_index.remove(task_id);
        }
        children
    };
    
    // Stop outside the lock
    for mut child in children_to_stop {
        child.stop().await;
    }
}
```

What do you think? Is that enough to have another go with or do you need more detail?

crates/turborepo-lib/src/task_graph/visitor/exec.rs
johnpyp: I added an `is_closing` atomic to the child and check that as the condition instead. So, if at any point `close` is called, all of the `ProcessManager`'s remaining children will report `ExecOutcome::Shutdown` thereafter.

crates/turborepo-process/src/child.rs
johnpyp: Implemented this in the latest revision. I don't believe that **both** the existing `children` vec and the task index are necessary, so I just use a hashmap of `TaskId -> Vec<Child>` and use `.values().flatten()` when iterating. Let me know if you think this is a bad choice.

crates/turborepo-process/src/lib.rs
johnpyp: Fixed

johnpyp:
Going with your suggested approach definitely turned down the overhead on the rest of the codebase. Let me know if this looks good, or if you meant even less invasive (no hooks into `process` or something).

confirmed still working with my combination of interruptible and persistent tasks and dependency libraries.

crates/turborepo-process/src/lib.rs
```diff
@@ -130,6 +139,47 @@ impl ProcessManager {
...
+        for child in children_to_stop {
+            set.spawn(async move {
+                let mut child = child;
+                child.stop().await
+            });
+        }
+
+        while let Some(out) = set.join_next().await {
+            trace!("process exited: {:?}", out);
+        }
```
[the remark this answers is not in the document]
johnpyp: I believe this is correct / fine and wouldn't cause any bugs based on the usage of it... @anthonyshew please correct me if I'm missing something. I could adopt the suggested fix which seems like it also wouldn't change observed behavior at all?

crates/turborepo-lib/src/run/watch.rs
```diff
@@ -266,11 +275,46 @@ impl WatchClient {
...
+                }
+            }
+        }
+
+        let mut impacted_packages = HashSet::new();
+        for task_id in &tasks_to_stop {
+            impacted_packages.insert(PackageName::from(task_id.package()));
+        }
+
+        *pkgs = impacted_packages;
```
anthonyshew: I might be misunderstanding the intent here but this mutation looks a little suspicious. Is there a concrete reason for it that I'm not realizing?

I'm seeing that the callsite is in the notification loop, but I think I'm not getting why this is like this (assuming its purposeful).

anthonyshew: I'm imagining going in this direction:
```
async fn stop_impacted_tasks(&self, pkgs: &HashSet<PackageName>) -> HashSet<PackageName> {
    let engine = self.run.engine();
    let mut tasks_to_stop = HashSet::new();
    for node in engine.tasks() {
        if let TaskNode::Task(task_id) = node {
            if pkgs.contains(&PackageName::from(task_id.package())) {
                tasks_to_stop.insert(task_id.clone());
                for dependent_node in engine.transitive_dependents(task_id) {
                    if let TaskNode::Task(dependent_id) = dependent_node {
                        tasks_to_stop.insert(dependent_id.clone());
                    }
                }
            }
        }
    }
    let impacted_packages: HashSet<_> = tasks_to_stop
        .iter()
        .map(|task_id| PackageName::from(task_id.package()))
        .collect();
    let task_ids: Vec<_> = tasks_to_stop.into_iter().collect();
    for handle in &self.active_runs {
        handle.stopper.stop_tasks(&task_ids).await;
    }
    impacted_packages
}
```
Then the call site becomes explicit:
```
ChangedPackages::Some(pkgs) => {
    let impacted = self.stop_impacted_tasks(&pkgs).await;
    // Now it's obvious we're using an expanded set
    changed_packages = ChangedPackages::Some(impacted);
}
```

crates/turborepo-lib/src/run/watch.rs
```diff
@@ -266,11 +275,46 @@ impl WatchClient {
...
+
+        let task_ids: Vec<_> = tasks_to_stop.into_iter().collect();
+        for handle in &self.active_runs {
+            handle.stopper.stop_tasks(&task_ids).await;
+        }
```
anthonyshew: Minor performance suggestion:

```suggestion
        join_all(
            self.active_runs
                .iter()
                .map(|handle| handle.stopper.stop_tasks(&task_ids)),
        )
        .await;

```

crates/turborepo-process/src/child.rs
```diff
@@ -680,6 +687,16 @@ impl Child {
...
+
+    /// Mark this child as being stopped as part of a ProcessManager shutdown
+    pub fn set_closing(&self) {
+        self.closing.store(true, Ordering::SeqCst);
+    }
+
+    /// Check if this child was stopped as part of a ProcessManager shutdown
+    pub fn is_closing(&self) -> bool {
+        self.closing.load(Ordering::SeqCst)
+    }
```
anthonyshew, line 699: We probably don't have to worry about the memory ordering as heavily as `SeqCst`?

```suggestion
    pub fn set_closing(&self) {
        self.closing.store(true, Ordering::Release);
    }

    /// Check if this child was stopped as part of a ProcessManager shutdown
    pub fn is_closing(&self) -> bool {
        self.closing.load(Ordering::Acquire)
    }
```

crates/turborepo-lib/src/run/watch.rs
```diff
@@ -266,11 +275,46 @@ impl WatchClient {
...
+                    tasks_to_stop.insert(task_id.clone());
+
+                    for dependent_node in engine.transitive_dependents(task_id) {
+                        if let TaskNode::Task(dependent_id) = dependent_node {
+                            tasks_to_stop.insert(dependent_id.clone());
+                        }
+                    }
+                }
+            }
+        }
```
anthonyshew, line 295: This is iterating over all tasks and searching for transitive dependencies for each task. We could batch this into a single graph traversal.

This is purely a performance optimization, and it will probably require creating another function, so feel free to skip out on this. I can handle it in a separate PR. I don't want to overload the amount of work you're doing to make this happen!

crates/turborepo-process/src/lib.rs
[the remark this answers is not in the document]
anthonyshew: Agree, this seems a bit overzealous. The way we're using this is fine.

crates/turborepo-lib/src/run/watch.rs
johnpyp: Ok, yeah I'll leave this one out, thanks.

johnpyp:
Done with revisions, apologies for the delay! Hopefully good to go.

Issue #8647, opened 2024-07-02
Log prefix breaks Jest extension for VS Code


OliverJAsh:
### Verify canary release

- [X] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

https://github.com/OliverJAsh/turbo-vscode-jest-2

### What package manager are you using / does the bug impact?

Yarn v2/v3/v4 (node_modules linker only)

### What operating system are you using?

Mac

### Which canary version will you have in your reproduction?

2.0.6-canary.0

### Describe the Bug

When Turborepo is used in combination with [the Jest extension for VS Code](https://github.com/jest-community/vscode-jest), the Jest extension is unable to parse the list of tests files because of Turborepo's log prefix.

In the reduced test case, the VS Code settings file configures the Jest extension to access Jest via Turborepo:

`.vscode/settings.json`:
```json
{
  "jest.jestCommandLine": "yarn run turbo run test --"
}

```

To generate the list of test files, vscode-jest will invoke the specified command, appending `--listTests --json`, resulting in the following output:

```console
$ TURBO_UI=false yarn run turbo run test -- --listTests --json
• Running test
• Remote caching disabled
test: cache hit, replaying logs c649256b92a71f6f
test: ["/Users/oliver/Code/reduced-test-cases/turbo-vscode-jest-2/src/add.test.js"]

 Tasks:    1 successful, 1 total
Cached:    1 cached, 1 total
  Time:    56ms >>> FULL TURBO
```

The `test: ` prefix breaks the Jest extension's ability to parse the JSON list of files.

Actual: tests do not appear in test explorer (sidebar).

<img width="838" alt="image" src="https://github.com/vercel/turbo/assets/921609/547bfd70-bbbd-4771-af23-3237139eced0">

Expected: tests should appear in test explorer (sidebar). Here's what it looks like with the workaround (`--log-prefix=none`):

<img width="842" alt="image" src="https://github.com/vercel/turbo/assets/921609/396115c8-73ab-4f61-9703-910c322ae053">

### Expected Behavior

It's possible to workaround this issue using `--log-prefix=none`, however I understand that by default Turborepo has some heuristics for when it should automatically disable the log prefix, so it would be useful if Turborepo could extend these heuristics so it works out of the box with VS Code's Jest extension.

### To Reproduce

1. Clone the repository.
2. Run `yarn` to install dependencies.
[...]

PR #11168, opened 2025-12-02
fix(cli): Disable log prefix for vscode jest extension

nyxsky404:
### Description

This PR fixes an issue where Turborepo's log prefix interferes with the VS Code Jest extension's ability to parse test results (Issue #8647).

When [turbo](cci:1://file:///Users/cursed/Desktop/Turborepo%21/crates/turborepo-lib/src/task_graph/visitor/mod.rs:535:0-538:1) is run with `--listTests` and `--json` (flags used by the Jest extension), the output should be clean JSON. However, [turbo](cci:1://file:///Users/cursed/Desktop/Turborepo%21/crates/turborepo-lib/src/task_graph/visitor/mod.rs:535:0-538:1) was prepending `task:package:` prefixes to the output, causing JSON parsing errors.

This change adds a heuristic to [RunOpts](cci:2://file:///Users/cursed/Desktop/Turborepo%21/crates/turborepo-lib/src/opts.rs:221:0-242:1) to automatically disable the log prefix (`ResolvedLogPrefix::None`) when both `--listTests` and `--json` are detected in the pass-through arguments.

image

### Testing Instructions

1.  **Automated Test**: A new integration test has been added: `turborepo-tests/integration/tests/run-logging/log-prefix-vscode-jest.t`.
    Run it using:
    ```bash
    pnpm --filter turborepo-tests-integration exec prysk tests/run-logging/log-prefix-vscode-jest.t
    ```

2.  **Manual Verification**:
    Run a task with the Jest flags and verify the output does not contain prefixes (e.g., `app:build:`):
    ```bash
    turbo run build -- --listTests --json
    ```
    The output should start with the command output (or clean JSON) without the Turborepo task prefix.

### DIFF
--- a/crates/turborepo-lib/src/opts.rs
+++ b/crates/turborepo-lib/src/opts.rs
@@ -9,13 +9,13 @@ use turborepo_cache::{CacheOpts, RemoteCacheOpts};
 use turborepo_task_id::{TaskId, TaskName};
 
 use crate::{
+    Args,
     cli::{
         Command, ContinueMode, DryRunMode, EnvMode, ExecutionArgs, LogOrder, LogPrefix,
         OutputLogsMode, RunArgs,
     },
-    config::{ConfigurationOptions, CONFIG_FILE},
+    config::{CONFIG_FILE, ConfigurationOptions},
     turbo_json::{FutureFlags, UIMode},
-    Args,
 };
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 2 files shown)


anthonyshew:
Hi, I'm not going to be able to accept this PR. The proposed solution is too wide and could have negative effects on other tooling to fix a corner case.

In this specific case, the behavior is activated when `--listTests` and `--json` are both in passthrough flags. This isn't a good enough heuristic to be 100% certain that the caller is the VSCode Jest extension. I could be passing those flags and not be in the extension.

I'm also noticing that the extension offers [configuration](https://marketplace.visualstudio.com/items?itemName=Orta.vscode-jest#jestcommandline) so you can change the command line invocation. I'd recommend using [`--log-prefix=none`](https://turborepo.com/docs/reference/run#--log-prefix-option) there and the logs will not have prefixes.

Issue #10639, opened 2025-07-05
Turbo crashed and asked to create a bug


lucafaggianelli:
### Verify canary release

- [x] I verified that the issue exists in the latest Turborepo canary release.

### Link to code that reproduces this issue

none

### Which canary version will you have in your reproduction?

turbo 2.5.4

### Environment information

```block
turbo 2.5.4

CLI:
   Version: 2.5.4
   Path to executable: /Users/luca/dev/experiments/lumen/node_modules/.pnpm/turbo-darwin-arm64@2.5.4/node_modules/turbo-darwin-arm64/bin/turbo
   Daemon status: Not running
   Package manager: pnpm9

Platform:
   Architecture: aarch64
   Operating system: macos
   WSL: false
   Available memory (MB): 5934
   Available CPU cores: 8

Environment:
   CI: None
   Terminal (TERM): xterm-256color
   Terminal program (TERM_PROGRAM): vscode
   Terminal program version (TERM_PROGRAM_VERSION): 1.101.2
   Shell (SHELL): /bin/zsh
   stdin: false
```

### Expected behavior

To not crash out of the blue

### Actual behavior

Turbo just crashed:

```
Tasks                    │ @lumen/query-service#dev > cache bypass, force executing 13c49de38a098063 
↑ ↓ - Select             │<-- POST /api/query
m - More binds           │   i - Interact   u/d - Scroll logs   U/D - Page logs   t/b - Jump to top/bottomOops! Turbo has crashed.

                                                                                                                                  A report has been written to /var/folders/pz/6sskpsmn0639wbyff_7nl7580000gn/T/report-e94fd8c7-056f-48a0-86bb-5433135ecc95.toml

                                                                                        Please open an issue at https://github.com/vercel/turborepo/issues/new/choose and include this file
```

### To Reproduce

Run `turbo dev`

### Additional context

log file:

```
"name" = "turbo"
"operating_system" = "Mac OS 14.7.4 [64-bit]"
"crate_version" = "2.5.4"
"explanation" = """
file 'crates/turborepo-vt100/src/grid.rs' at line 869
"""
"cause" = """
panicked at crates/turborepo-vt100/src/grid.rs:869:18:
called `Option::unwrap()` on a `None` value"""
"method" = "Panic"
"backtrace" = """

   0: 0x1044181a4 - __mh_execute_header
   1: 0x10441c05c - __mh_execute_header
   2: 0x1055bb8b4 - __mh_execute_header
[...]

PR #11170, opened 2025-12-02
fix: Prevent vt100 TUI crash during column wrap

neel3o115:
This PR fixes the panic described in **#10639**, where the TUI renderer crashes when `col_wrap` calls `drawing_row_mut(...).unwrap()` on an invalid row during heavy scrolling/wrapping.

The fix safely handles cases where `drawing_row_mut` returns `None`, preventing the crash and allowing TUI updates to continue normally.

Regression tests are included to ensure this scenario no longer panics.

---

## **### Testing Instructions**

1. Check out the branch:

```sh
git checkout fix-tui-crash
```

2. Run tests:

```sh
cargo test -p turborepo-vt100
```

3. Verify:

   * No panics occur during `col_wrap`
   * Both new edge-case tests pass


neel3o115:
Hi @anthonyshew! Just wanted to gently check in regarding the PR I opened.
Whenever you get a moment, could you please take a look and review it?
Thanks!

anthonyshew:
historical note: This is a duplicate of https://github.com/vercel/turborepo/pull/10640, which we [closed because we were experimenting with some AI tooling](https://github.com/vercel/turborepo/pull/10640#issuecomment-3046118598).

It appears we didn't take the time to review it enough and it's actually a good fix. Going to move the test file into the implementation file and merge.

PR #11173, opened 2025-12-03
ci: Prepare release pipeline for Trusted Publishing

anthonyshew:
### Description

Prepare our release pipelines for [Trusted Publishing](https://docs.npmjs.com/trusted-publishers).

- Removes now-unneeded usages of `NPM_TOKEN`
- Adds `id-token: write` permission [per Trusted Publishing docs](https://docs.npmjs.com/trusted-publishers#github-actions-configuration).
- Updates release pipeline documentation

### Testing Instructions

After the PR has been merged, testing will include:
- Doing a dry run of the release pipeline
- On success, attempting a canary release

PR #11186, opened 2025-12-03
ci: More permissions for release workflow

anthonyshew:
### Description

More followup for #11173.

### DIFF
--- a/.github/workflows/turborepo-release.yml
+++ b/.github/workflows/turborepo-release.yml
@@ -20,7 +20,7 @@ env:
 permissions:
   id-token: write # Required for npm Trusted Publishing using OIDC
   contents: write # Allow workflow to checkout code from the repository
-  pull-requests: write
+  pull-requests: write # Allows the PR for post-release to be created
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/turborepo-release.yml
```diff
@@ -20,6 +20,7 @@ env:
 permissions:
   id-token: write # Required for npm Trusted Publishing using OIDC
   contents: write
```
anthonyshew:
```suggestion
  contents: write # Allow workflow to checkout code from the repository
```

.github/workflows/turborepo-release.yml
```diff
@@ -19,7 +19,8 @@ env:
 
 permissions:
   id-token: write # Required for npm Trusted Publishing using OIDC
-  contents: write
+  contents: write # Allow workflow to checkout code from the repository
+  pull-requests: write
```
anthonyshew:
```suggestion
  pull-requests: write # Allows the PR for post-release to be created
```

PR #11188, opened 2025-12-03
examples(security): Upgrade Next.js versions

anthonyshew:
### Description

In accordance with mitigation instructions for https://github.com/vercel/next.js/security/advisories/GHSA-9qr9-h5gf-34mp, upgrading Next.js to latest in examples.


anthonyshew:
Failures are existing, unrelated, and specific to this workflow. The examples still work for users.

PR #11199, opened 2025-12-04
fix: Command injection in turbo-ignore (#11154)

VIPAX-JIT:
Description
Fixes a command injection vulnerability in turbo-ignore by replacing execSync with execFileSync when validating git refs. The old implementation used shell command strings. The fix ensures that user-controlled input (like git refs from environment variables) is passed as literal arguments rather than being interpreted by a shell.

Testing Instructions
Run the security test: pnpm test security.test.ts in packages/turbo-ignore


anthonyshew:
This is not a fix, its a regression. The fix was https://github.com/vercel/turborepo/pull/11154.

PR #11215, opened 2025-12-07
fix: Reorder config loading and path validation to ensure packages without configs are ignored.

cp89cyber:
This PR fixes an issue where the microfrontends configuration loader would throw a PathTraversal error for any package located outside the repository root (e.g., via sibling directories or symlinks), even if that package was not using microfrontends.

The fix involves deferring the 
validate_package_path
 Check until after we have confirmed that a microfrontends.json (or .jsonc) file actually exists in the package directory. If no configuration file is found, we simply return Ok(None) and bypass the path validation entirely. This ensures that only packages opting into microfrontends are subject to the strict repository root requirements necessary for the proxy.

Testing Instructions
I have added new regression tests in 
crates/turborepo-microfrontends/src/lib.rs
 to verify this behavior:

test_package_outside_root_no_config_ignored
: Confirms that a package symlinked from outside the repo root does not error if it has no config.
test_config_outside_root_rejected
: Confirms that a package outside the repo root does correctly error if it attempts to define a microfrontend config (maintaining security).
test_valid_internal_package_via_traversal
: Confirms that valid internal paths resolving inside the root are still accepted.
To run the tests:
cargo test -p turborepo-microfrontends

### DIFF
--- a/crates/turborepo-microfrontends/src/lib.rs
+++ b/crates/turborepo-microfrontends/src/lib.rs
@@ -140,11 +140,11 @@ impl TurborepoMfeConfig {
     ) -> Result<Option<Self>, Error> {
         let absolute_dir = repo_root.resolve(package_dir);
 
-        Config::validate_package_path(repo_root, &absolute_dir)?;
-
         let Some((contents, path)) = Self::load_v1_dir(&absolute_dir)? else {
             return Ok(None);
         };
+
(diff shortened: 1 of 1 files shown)


anthonyshew:
Correct fix for this issue is in https://github.com/vercel/turborepo/pull/11286

PR #11297, opened 2025-12-22
perf(create-turbo): Download files faster

anthonyshew:
### Description

I was becoming a little annoyed that `create-turbo` takes _so long_ to download files. Even while I'm on a strong internet connection, it would take something like 10 seconds.

Additionally, situations like https://github.com/vercel/turborepo/issues/10456 were coming up because getting the example onto the machine is so slow. It appears to be hanging when its really just unexpectedly slow.

 With some investigation, I found out that `create-turbo` was downloading, **the entirety of `vercel/turborepo`** just to get the code for an example. It turns out that this is a deficiency in GitHub's API; you can't download subdirectories from a repo.
 
Instead, we can do sparse checkouts without the lightest-weight git options that we can to improve performance as much as possible.

### Testing Instructions

I'm hand-testing the binary and it's significantly faster.
```
▲ 👟 ~/projects/debugs
  hyperfine \
  -r 5 \
  --prepare 'rm -rf ./benchies' \
  --cleanup 'rm -rf ./benchies' \
  -n 'Old' 'npx create-turbo ./benchies --skip-install --skip-transforms' \
  -n 'New' 'node ~/projects/open/turbo/packages/create-turbo/dist/cli.js ./benchies --skip-install --skip-transforms'
Benchmark 1: Old
  Time (mean ± σ):     37.377 s ±  3.938 s    [User: 2.602 s, System: 1.044 s]
  Range (min … max):   31.170 s … 41.525 s    5 runs

Benchmark 2: New
  Time (mean ± σ):      2.985 s ±  0.098 s    [User: 0.659 s, System: 0.323 s]
  Range (min … max):    2.890 s …  3.107 s    5 runs

Summary
  New ran
   12.52 ± 1.38 times faster than Old

```

<sub>Closes https://github.com/vercel/turborepo/issues/10456</sub>
<sub> Closes TURBO-5050</sub>

### DIFF
--- a/packages/turbo-utils/__tests__/examples.test.ts
+++ b/packages/turbo-utils/__tests__/examples.test.ts
@@ -214,12 +214,6 @@ describe("examples", () => {
     it("allows directories", () => {
       expect(isLinkEntry("Directory")).toBe(false);
     });
-
-    it("allows other entry types", () => {
-      expect(isLinkEntry("BlockDevice")).toBe(false);
-      expect(isLinkEntry("CharacterDevice")).toBe(false);
-      expect(isLinkEntry("FIFO")).toBe(false);
-    });
   });
 
   describe("isPathSafe with pre-resolved root (performance optimization)", () => {
@@ -229,24 +223,6 @@ describe("examples", () => {
       expect(isPathSafe("/tmp/extract", "../etc/passwd", resolvedRoot)).toBe(
         false
       );
-    });
-
-    it("produces same results with and without pre-resolved root", () => {
-      const root = "/tmp/extract";
-      const resolvedRoot = "/tmp/extract";
-      const testPaths = [
-        "file.txt",
-        "subdir/file.txt",
-        "../etc/passwd",
-        "foo/../bar",
-        "foo/../../../etc/passwd",
-      ];
-
-      for (const path of testPaths) {
-        expect(isPathSafe(root, path)).toBe(
-          isPathSafe(root, path, resolvedRoot)
-        );
-      }
     });
   });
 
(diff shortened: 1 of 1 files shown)


packages/turbo-utils/src/examples.ts
```diff
@@ -98,64 +128,240 @@ export function existsInRepo(nameOrUrl: string): Promise<boolean> {
...
-    },
-  });
+  const tempDir = join(root, ".turbo-clone-temp");
 
-  await fs.unlink(tempFile);
+  try {
+    // Clone with partial clone (no blobs) and no checkout
+    // Using execFileSync with array args to prevent shell injection
+    execFileSync(
+      "git",
```
tknickman, line 318: Imo we should keep the old method around as well, and fall back to it when git isn't available on the machine.

packages/turbo-utils/src/examples.ts
```diff
@@ -98,64 +128,240 @@ export function existsInRepo(nameOrUrl: string): Promise<boolean> {
...
+
+    // Checkout the files
+    execFileSync("git", ["checkout"], {
+      cwd: tempDir,
+      stdio: "pipe",
+    });
+
+    // Copy the example files to the root
+    const examplePath = join(tempDir, "examples", name);
+    cpSync(examplePath, root, { recursive: true });
```
tknickman, line 346: We need to make sure .git is removed (we might be doing that in a transform, but need to double check that)

packages/turbo-utils/src/examples.ts
```diff
@@ -98,64 +128,240 @@ export function existsInRepo(nameOrUrl: string): Promise<boolean> {
...
+  root: string;
+  strip: number;
+  filter: (path: string, rootPath: string | null) => boolean;
+}
+
+/**
+ * Streaming extraction from a tarball URL.
+ * Exported for testing purposes.
+ */
+export async function streamingExtract({
```
tknickman, line 186: is this rewritten just to avoid the dep? This should be roughly equivalent to got.stream

anthonyshew: Yep! Fewer deps, fewer problems, and reduces the size of `create-turbo` by roughly 304kB.

packages/turbo-utils/src/examples.ts
anthonyshew: I did check on this to make sure! The way it works right now is safe because `cpSync` only copies from `examples/{name}`, not from the `vercel/turborepo` repo root where .git lives. 

We could add a `rmSync(join(root, ".git"), { recursive: true, force: true })` after the `cpSync`, though, as an extra defensive measure. Thoughts?

packages/turbo-utils/src/examples.ts
anthonyshew: Beat me to it! Was going to bring it back as a catch for when this one fails.

Technically a breaking change, anyway, wouldn't be able to ship without it.

docs/site/turbo.json
```diff
@@ -9,7 +9,6 @@
         "TINYBIRD_TIME_SAVED_TOKEN",
         "ENABLE_EXPERIMENTAL_COREPACK",
         "TRAY_URL",
-        "BLOB_READ_WRITE_TOKEN",
```
anthonyshew, line 12: Unrelated, but I noticed it while I was doing the iteration that went through Vercel Blob, so thought I'd clean it up.

packages/turbo-utils/src/examples.ts
anthonyshew: Can do this in follow-up if you want to add another spot to defend from this, but yeah, .git is removed as it works today.

PR #11307, opened 2025-12-23
fix: Transit nodes in Devtools

anthonyshew:
### Description

CLOSES https://github.com/vercel/turborepo/issues/11301

### Testing Instructions

Verified the fix in the browser and added a test.


anthonyshew:
Fixed in https://github.com/vercel/turborepo/releases/tag/v2.7.2. Thank you for reporting!

PR #11323, opened 2025-12-28
examples: Remove skipLibCheck from nestjs.json in nestJS example

leos:
Removed 'skipLibCheck' from `nestjs.json`

### Description

`skipLibCheck` is already set to `true` in `base.json`. Having it here as well is duplicative.

### Testing Instructions

`pnpm run build`


anthonyshew (approved):
Thanks!

PR #11363, opened 2026-01-01
fix: Support file watching for non-root volumes on MacOS

### LINKED ISSUE #9536: 🐛 Bug: [MacOS][Case-sensitive volume] `turbo watch`: discovery failed: bad grpc status code: The operation was cancelled
### Verify canary release

https://github.com/hrstkatomas/test-monorepo-tooling

turbo 2.3.3

### Enviroment information

```block
turbo 2.3.3

CLI:
   Version: 2.3.3
   Path to executable: /Volumes/.../test-monorepo-tooling/node_modules/.pnpm/turbo-darwin-arm64@2.3.3/node_modules/turbo-darwin-arm64/bin/turbo
   Daemon status: Running
   Package manager: pnpm

Platform:
   Architecture: aarch64
   Operating system: macos
   WSL: false
   Available memory (MB): 14738
   Available CPU cores: 10

Environment:
   CI: None
   Terminal program version (TERM_PROGRAM_VERSION): 455
   Shell (SHELL): /opt/homebrew/bin/bash
   stdin: false
```

### Expected behavior

Should run the project in watch mode without any issues on MacOS on both case-sensitive and case-insensitive file systems.

### Actual behavior

When running `turbo watch [command]` on MacOS on case-sensitive volume, I get this error:
```
turbo 2.3.3

• Packages in scope: @repo/typescript-config, @repo/ui, tribalfootball
• Running build in 3 packages
• Remote caching disabled
  × failed to connect to daemon
  ╰─▶ server is unavailable: channel closed
```

After running `turbo daemon clean` I receive a slightly different error for the following run: 
```
turbo 2.3.3

• Packages in scope: @repo/typescript-config, @repo/ui, tribalfootball
• Running build in 3 packages
• Remote caching disabled
  × discovery failed: bad grpc status code: The operation was cancelled
```

### To Reproduce

1. Mac required
2. In Disk Utility add a new Volume and select Case-sensitive format
3. Clone the repository onto newly created volume
4. Run `turbo watch build`

### Additional context

The same repository works as expected when I clone it onto case-**in**sensitive volume

mdlawson:
### Description

Should fix #9536 

* The current callback for FSEventStream assumes that the relative paths can always be resolved from root - this breaks if the watcher is set up on a non-root volume, such as when using case sensitive volumes.
* This change resolves the mount point of the watched device, and uses that to resolve file paths.

A much simpler possible fix would be to switch to `FSEventStreamCreate` rather than `FSEventStreamCreateRelativeToDevice` to watch all devices - these APIs don't seem to differ given how turborepo uses the results, but I'm assuming FSEventStreamCreateRelativeToDevice was chosen deliberately for some reason that isn't immediately obvious from the history. 

### Testing Instructions

You can reproduce this bug by running `turbo watch build` in this repo checked out on a non-root volume. With this change, the file watcher daemon works as expected.

I've included an a test for this behaviour, but the setup for it is pretty messy as a temporary volume needs to be created to actually exhibit the issue - open to other ideas here, or removing it completely if it seems overkill.

### DIFF
--- a/crates/turborepo-filewatch/src/fsevent.rs
+++ b/crates/turborepo-filewatch/src/fsevent.rs
@@ -72,7 +72,107 @@ bitflags::bitflags! {
   }
 }
 
-/// FSEvents-based `Watcher` implementation
+/// Encapsulates device information and path transformation logic.
+///
+/// This type handles the bidirectional conversion between absolute filesystem
+/// paths and the device-relative paths used by FSEvents when watching non-root
+/// volumes.
+///
+/// # Path Transformation Contract
+///
+/// When registering a watch path:
+/// - `to_device_relative()` strips the mount point prefix and prepends "/"
+///
+/// When receiving events in the callback:
+/// - `to_absolute()` joins the mount point with the device-relative path
+///
+/// This symmetry must be maintained for correct path reporting.
+#[derive(Debug, Clone)]
+struct DeviceContext {
+    /// The device ID from `stat.st_dev`
+    device_id: i32,
+    /// The effective mount point of the device (e.g., "/", "/Volumes/Data")
+    mount_point: PathBuf,
+}
+
+impl DeviceContext {
+    /// Create a new DeviceContext for the given path.
+    fn new(path: &Path) -> Result<Self> {
+        let metadata = std::fs::symlink_metadata(path).map_err(|e| {
+            if e.kind() == ErrorKind::NotFound {
+                Error::path_not_found().add_path(path.into())
+            } else {
+                Error::io(e)
+            }
+        })?;
+        let device_id = metadata.dev() as i32;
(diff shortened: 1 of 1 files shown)


anthonyshew:
Hey, thanks for kicking the tires on this. I did a few things for a little more robustness:

Key Changes:
1. New DeviceContext struct - Encapsulates device information and path transformation logic:
2. Replaced separate fields - device: Option<i32> and device_path: Option<PathBuf> consolidated into single device_context: Option<DeviceContext>
3. Fixed panic points in callback - Replaced .expect() on UTF-8 conversion with graceful continue to skip malformed events (prevents undefined behavior in extern C callback)
4. Fixed unwrap() on path conversions - remove_path() now returns proper errors instead of panicking on non-UTF8 paths
5. Improved `TestVolume`

Will test it out now...

PR #11391, opened 2026-01-06
chore: Migrate from ESLint to oxlint with consolidated root lint command

anthonyshew:
## Summary

- Consolidate oxlint to run as a single command from the root instead of per-package
- Add `--deny-warnings` flag to treat all warnings as errors
- Create `.oxlintrc.json` with appropriate rules configuration

### Root lint consolidation
- Added `lint` script to root `package.json` that runs oxlint on `packages/` and `docs/site`
- Updated `turbo.json` to run `//#lint` after prettier checks complete
- Removed individual `lint` scripts from 12 packages (oxlint now runs from root)
- Kept `fumadocs-mdx` step in docs as a prerequisite for type generation

### Oxlint configuration (`.oxlintrc.json`)
- Ignore unused catch parameters (`caughtErrors: "none"`)
- Ignore variables/params starting with `_` (common convention for intentionally unused)
- Disable `unicorn/no-empty-file` for fixture files
- Disable `no-unused-expressions` for fixture files  
- Disable `triple-slash-reference` for `next-env.d.ts` (auto-generated by Next.js)

### Code fixes
- Removed unused imports (`beforeEach`, `PackageManager`, `createGzip`)
- Removed unused parameter and variables
- Converted unused ternary to if/else statement
- Removed dead `config.pipeline;` statement
- Fixed unnecessary regex escapes (`\/` → `/`)

## Test Plan

```bash
turbo lint
```

We should also see no other unrelated failure in CI.

<sub>CLOSES TURBO-4916</sub>

### DIFF
--- a/.oxlintignore
+++ b/.oxlintignore
@@ -28,9 +28,6 @@ scripts/
 .github/
 .devcontainer/
 
-# Docs tooling (separate from docs/site)
-docs/link-checker/
-
 # Config directories
 .cargo/
 .config/
--- a/docs/link-checker/src/markdown.ts
+++ b/docs/link-checker/src/markdown.ts
@@ -256,7 +256,6 @@ const validateHashLink = (doc: Document, href: string) => {
     href,
     doc,
   };
-  const { content, ...docWithoutContent } = doc;
   return [linkError];
 };
 


.oxlintignore
```diff
@@ -0,0 +1,42 @@
+# Dependencies and build outputs
```
connorshea, line 1: @anthonyshew hello! Thanks for adopting oxlint! Tell us if you run into any problems :)

For what it's worth, I would recommend using `ignorePatterns` in the .oxlintrc.json file rather than a separate oxlintignore file that needs to be passed with `--ignore-path`.

Devs running `oxlint` would hit weird discrepancies if they aren't using the command via the package.json and don't specify the `--ignore-path` flag. And especially devs using oxlint in their code editor (plus our AI pals running commands however they want) would not have the same files ignored without explicitly setting up their editor to ignore those files.

If you put the `ignorePatterns` field directly inside the oxlintrc.json, it solves all those problems and everyone ends up with the same files linted.

(also if any of these are already specified in the `.gitignore`, they should be picked up already by that)

PR #11418, opened 2026-01-10
fix(tui): Properly disable mouse mode on exit and panic

anthonyshew:
## Summary

Fixes mouse sequences appearing in terminals (especially VSCode) after `turbo` exits.

### Changes

1. **Add TUI state tracking** (`crates/turborepo-ui/src/tui/panic_handler.rs`)
   - New module with atomic flag to track whether TUI has modified terminal state
   - `restore_terminal_on_panic()` function that only runs restoration if TUI was active
   - Uses raw escape sequences to minimize risk of additional panics

2. **Track TUI lifecycle** (`crates/turborepo-ui/src/tui/app.rs`)
   - `startup()` sets TUI as active after entering alternate screen
   - `cleanup()` clears the flag after restoration completes
   - `cleanup()` now always sends `DisableMouseCapture` even in VSCode to handle child processes that may have enabled mouse mode

3. **Panic handler integration** (`crates/turborepo-lib/src/panic_handler.rs`)
   - Calls `restore_terminal_on_panic()` which checks the flag before doing anything
   - No side effects when TUI wasn't used (CI, piped output, non-TUI runs)

### Why this approach?

- **Safe**: Only restores terminal state if we actually modified it
- **No log pollution**: Won't write escape sequences to CI logs or piped output
- **Thread-safe**: Uses atomic operations
- **Idempotent**: Always disables mouse capture on normal exit to handle misbehaving child processes

## Test Plan

- Existing TUI tests pass
- Added unit tests for the panic handler flag tracking


anthonyshew:
For posterity, there is [a similar PR opened in Bun](https://github.com/oven-sh/bun/pull/25876) at the moment, though it is meant for dealing with Windows. If this turns out to be a Windows issue, there might be something to learn from that PR.

PR #11454, opened 2026-01-13
chore: Update domain to turborepo.dev

anthonyshew:
### Description

We're moving! turborepo.com → turborepo.dev


4mp3d:
Security software will often consider domains registered less than 30 days as high risk and will block traffic to them. This happened to us and we had to whitelist the new domain. If possible in the future would recommend not leveraging the new domain for production use till it has been registered at least 30 days.

PR #11463, opened 2026-01-13
fix: Replace execSync with execFileSync for oxfmt

odaysec:
### Description
fix is to avoid constructing a single shell command string that mixes a hard-coded command (`pnpm exec oxfmt`) with dynamic path arguments. Instead, invoke the command without a shell and pass each argument as a separate array element. This prevents spaces or metacharacters in the paths from being interpreted by a shell.

Concretely, in `packages/turbo-types/scripts/generate-schema.ts`, replace the `execSync` call on lines 58–61 with a call to `execFileSync` from `node:child_process`. Keep `pnpm` as the executable, and pass `"exec"`, `"oxfmt"`, and each schema file path as separate arguments. Also update the import on line 3 to import `execFileSync` instead of (or in addition to) `execSync`. No other logic needs to change.

You need:
- To adjust the child process import to include `execFileSync`.
- To change the command invocation to `execFileSync("pnpm", ["exec", "oxfmt", <paths>], { stdio: "inherit" });`.
- All changes are confined to the shown file and lines.

### DIFF
--- a/packages/prysk/index.mjs
+++ b/packages/prysk/index.mjs
@@ -15,7 +15,14 @@ const isWindows = process.platform === "win32";
 execFileSync("python3", ["-m", "venv", VENV_NAME]);
 
 // Upgrade pip
-execFileSync(getVenvBin("python3"), ["-m", "pip", "install", "--quiet", "--upgrade", "pip"]);
+execFileSync(getVenvBin("python3"), [
+  "-m",
+  "pip",
+  "install",
+  "--quiet",
+  "--upgrade",
+  "pip"
+]);
 
 // Install prysk
 execFileSync(getVenvBin("pip"), ["install", "prysk==0.15.2"]);
--- a/packages/turbo-types/scripts/generate-schema.ts
+++ b/packages/turbo-types/scripts/generate-schema.ts
@@ -55,10 +55,14 @@ create("schema.v2.json", "Schema");
 create("schema.json", "Schema");
 
 // Format generated schemas with oxfmt
-execFileSync("pnpm", [
-  "exec",
-  "oxfmt",
-  join(schemasDir, "schema.json"),
-  join(schemasDir, "schema.v1.json"),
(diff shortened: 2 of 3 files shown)


odaysec (commented):
fix command-line injection issues with git (and similar tools), you must: (1) ensure all command arguments derived from user or library input are strictly validated against an allowlist of safe patterns (e.g., safe path characters, no leading `-`, no NULs), and (2) avoid letting untrusted data control semantic arguments like remote URLs, options, or flags. For directory paths, constrain them to be “normal” relative paths or to reside under known safe base directories.

For this concrete code, the best minimally invasive fix is to centralize and strengthen directory validation in `validateDirectory`, then rely on that validated `root` in `downloadAndExtractExample`. Specifically:

- In `validateDirectory`, after resolving `directory` to `root`, add checks to ensure:
  - The `directory` string is non-empty, not just whitespace, and does not contain `\0`.
  - The resolved `root` does not begin with `-` (which could be misinterpreted as an option by CLI tools).
  - The resulting `projectName` is non-empty and consists of safe characters (e.g., letters, digits, hyphen, underscore, dot).
  - Optionally, reject obvious bad patterns like control characters.
- Return `valid: false` with a clear error message if any of these checks fail.

In `downloadAndExtractExample`, we already validate `name` and check `normalizedRoot`/`tempDir` for NUL and leading dash. With a stronger `validateDirectory` guarantee, the existing checks here become a second line of defense and satisfy CodeQL’s concern that `root` (ultimately coming from `directory`) can taint a git argument. Since we’re not allowed to alter unseen call sites, we only adjust the logic in the shown snippets, and we do not change the external behavior for valid, normal directory paths.



## References
Max Justicz: [3m apps at once through CocoaPods](https://justi.cz/security/2021/04/20/cocoapods-rce.html)
[Git - git-ls-remote Documentation](https://git-scm.com/docs/git-ls-remote/2.22.0#Documentation/git-ls-remote.txt---upload-packltexecgt)

PR #11593, opened 2026-01-31
fix: Extract git argument validation into reusable helper

anthonyshew:
## Summary

Extracts inline git argument validation into a dedicated `assertSafeGitArgument` helper function in `downloadAndExtractExample`.

## Why

The `downloadAndExtractExample` function passes user-influenced paths to `execFileSync("git", [...])`. The existing inline validation prevents argument injection by rejecting paths that:
- Are empty or not a string
- Contain NUL bytes (could truncate arguments)
- Start with `-` (could be interpreted as git options)

This change extracts that logic into a named function with clear documentation, making the security intent explicit and the validation reusable.

## Changes

- Add `assertSafeGitArgument(value, description)` helper with JSDoc explaining the security rationale
- Replace inline validation checks with calls to the helper
- Improve error messages to explain *why* the path is invalid (rather than echoing the invalid value back)

## Notes

This is a refactor only - the validation logic is semantically equivalent to what existed before. No new attack vectors are introduced or closed; this change improves code clarity and maintainability.

This PR absorbs one of the changes from https://github.com/vercel/turborepo/pull/11558.

### DIFF
--- a/packages/turbo-utils/src/examples.ts
+++ b/packages/turbo-utils/src/examples.ts
@@ -385,6 +385,26 @@ export async function downloadAndExtractRepo(
   }
 }
 
+/**
+ * Validates that a path is safe to use as a git CLI argument.
+ * Prevents argument injection by rejecting paths that:
+ * - Are empty or not a string
+ * - Contain NUL bytes (could truncate the argument)
+ * - Start with "-" (could be interpreted as a git option)
+ */
+function assertSafeGitArgument(value: string, description: string): void {
+  if (
+    !value ||
+    typeof value !== "string" ||
+    value.includes("\0") ||
+    value.startsWith("-")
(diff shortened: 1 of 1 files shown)


packages/turbo-utils/src/examples.ts
```diff
@@ -385,6 +385,24 @@ export async function downloadAndExtractRepo(
   }
 }
 
+/**
+ * Validates that a path is safe to use as a git CLI argument.
+ * Prevents argument injection by rejecting paths that:
+ * - Are empty or not a string
+ * - Contain NUL bytes (could truncate the argument)
+ * - Start with "-" (could be interpreted as a git option)
+ */
```
anthonyshew:
```suggestion
```

PR #11605, opened 2026-01-31
fix: Exclude coverage-reporter from Node.js 18 CI tests

anthonyshew:
## Summary

- Excludes `@turbo/coverage-reporter` from the JS package test workflow

The coverage-reporter app uses Next.js 16+ which requires Node.js >= 20. This causes CI failures in the Node.js 18 test matrix since the `build` task attempts to run `next build`.

### DIFF
--- a/.github/workflows/test-js-packages.yml
+++ b/.github/workflows/test-js-packages.yml
@@ -69,7 +69,6 @@ jobs:
         # We filter out turborepo-repository because it's a native package and needs
         # to run when turbo core changes. This job (`js_packages`) does not run on turborpeo core
         # changes, and we don't want to enable that beahvior for _all_ our JS packages.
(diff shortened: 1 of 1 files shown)


.github/workflows/test-js-packages.yml
```diff
@@ -69,8 +69,9 @@ jobs:
         # We filter out turborepo-repository because it's a native package and needs
         # to run when turbo core changes. This job (`js_packages`) does not run on turborpeo core
         # changes, and we don't want to enable that beahvior for _all_ our JS packages.
+        # We filter out @turbo/coverage-reporter because it uses Next.js 16+ which requires Node.js >= 20.
```
anthonyshew:
```suggestion
```

PR #11614, opened 2026-02-01
fix: Upgrade node-plop to remove @babel/runtime-corejs3 vulnerability

anthonyshew:
## Summary

- Upgrades `node-plop` from 0.26.3 to 0.32.3 to resolve TURBO-5156
- Removes the `@babel/runtime-corejs3` transitive dependency which had an inefficient RegExp complexity vulnerability (<7.26.10)

## Changes

The newer `node-plop` (0.32.3) is ESM-only and no longer depends on `@babel/runtime-corejs3`. This required:

1. Converting `getPlop`, `getCustomGenerators`, and `getCustomGenerator` from sync to async functions
2. Using dynamic `import()` for the ESM `node-plop` module in our CommonJS package
3. Updating callers in `custom.ts` to await the async functions
4. Adding `"module": "CommonJS"` to tsconfig.json to enable dynamic imports

## Testing

- `pnpm --filter @turbo/gen build` ✅
- `pnpm --filter @turbo/gen test` ✅  
- `pnpm --filter @turbo/gen check-types` ✅
- Verified `@babel/runtime-corejs3` is no longer in the dependency tree

### DIFF
--- a/packages/turbo-gen/package.json
+++ b/packages/turbo-gen/package.json
@@ -27,7 +27,7 @@
     "fs-extra": "10.1.0",
     "inquirer": "8.2.4",
     "minimatch": "9.0.0",
-    "node-plop": "0.26.3",
+    "node-plop": "0.32.3",
     "picocolors": "1.0.1",
     "proxy-agent": "6.5.0",
     "ts-node": "10.9.2",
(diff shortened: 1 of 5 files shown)


anthonyshew:
As mentioned in https://github.com/vercel/turborepo/pull/11590#pullrequestreview-3733588909, bumping this dependency will cause issues. We will revisit soon.

PR #11649, opened 2026-02-02
feat: Migrate from tsup to tsdown

anthonyshew:
## Summary

Migrates all TypeScript package bundling from tsup to tsdown. tsdown is built on Rolldown and provides faster builds while maintaining compatibility with tsup's configuration patterns.

## Packages Migrated

- `@turbo/workspaces`
- `@turbo/releaser`
- `turbo-ignore`
- `@turbo/gen`
- `@turbo/codemod`
- `eslint-plugin-turbo`
- `create-turbo`
- `kitchen-sink/apps/api` (example)

## Changes

- Replaced `tsup` dependency with `tsdown@0.9.3`
- Renamed `tsup.config.ts` → `tsdown.config.ts` (`.mts` for `@turbo/gen` due to `"type": "commonjs"`)
- Configured `outExtensions` to preserve tsup's output file naming conventions (`.js`, `.mjs`, `.d.ts`, `.d.mts`)
- For `eslint-plugin-turbo`: Added `tsconfig.build.json` for declaration generation since tsdown's dts plugin doesn't support mixed named/default exports

## Testing

All output files verified to match tsup's outputs:
- Entry point files have identical names and locations
- `bin`, `main`, `module`, and `types` fields in package.json point to existing files
- Internal chunk naming differs but is invisible to package consumers

Closes TURBO-5170

### DIFF
--- a/examples/kitchen-sink/apps/api/package.json
+++ b/examples/kitchen-sink/apps/api/package.json
@@ -4,7 +4,7 @@
   "type": "module",
   "private": true,
   "scripts": {
-    "start": "node dist/index.js",
+    "start": "node dist/index.cjs",
     "dev": "tsdown --watch --onSuccess \"node dist/index.cjs\"",
     "build": "tsdown",
     "check-types": "tsc --noEmit",


anthonyshew:
The `kitchen-sink` issue is a known problem. We will sort this out later on.

PR #11661, opened 2026-02-02
fix: Replace unmaintained adler with adler2

anthonyshew:
## Summary
- Updates flate2 which switches from unmaintained adler to adler2
- Fixes RUSTSEC-2025-0056 (adler crate is unmaintained)

The adler crate was a transitive dependency through flate2/miniz_oxide.

<sub>CLOSES TURBO-5191</sub>


anthonyshew:
Check examples is failing to due to an outage elsewhere. It isn't a meaningful test here, so can be merged through.

PR #11662, opened 2026-02-03
feat: Replace example tests in Vercel Sandboxes

anthonyshew:
## Summary

- Replaces the old example testing infrastructure with a new `check-examples.ts` script that runs examples in isolated Vercel Sandboxes
- Consolidates 5 separate CI jobs into a single `check-examples` job that runs on every PR
- Removes ~30 test package directories and associated helper scripts

Now, checking **all examples** takes ~2:30. Before even checking one of them took that long.

## Testing notes

The new checker validates all examples marked with `maintainedByCoreTeam: true` in their `meta.json`. It installs dependencies, runs all non-persistent `turbo` tasks, and verifies cache hits on second run.

### DIFF
--- a/examples/kitchen-sink/packages/ui/package.json
+++ b/examples/kitchen-sink/packages/ui/package.json
@@ -56,5 +56,8 @@
     "@types/react-dom": ">=18",
     "react": ">=18",
     "react-dom": ">=18"
+  },
+  "dependencies": {
+    "@swc/helpers": "^0.5.18"
   }
 }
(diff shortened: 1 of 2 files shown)


examples/check-examples.ts
```diff
@@ -0,0 +1,433 @@
...
+ *
+ * Validates all examples marked as `maintainedByCoreTeam` in meta.json
+ * by running them in isolated Vercel Sandboxes.
+ *
+ * This version creates fresh sandboxes for each run (no snapshots).
+ * Slower but simpler - no state management needed.
+ *
+ * Authentication:
+ * - For local development: Run `vercel link` and `vercel env pull` first
+ * - For CI/CD: Set VERCEL_TOKEN environment variable
```
anthonyshew:
```suggestion
```

PR #11678, opened 2026-02-05
Update NODE_VERSION to 24.13.0 in Dockerfile

i5d6:
A flaw in Node.js's buffer allocation logic can expose uninitialized memory when allocations are interrupted, when using the `vm` module with the timeout option. Under specific timing conditions, buffers allocated with `Buffer.alloc` and other `TypedArray` instances like `Uint8Array` may contain leftover data from previous operations, allowing in-process secrets like tokens or passwords to leak or causing data corruption. While exploitation typically requires precise timing or in-process code execution, it can become remotely exploitable when untrusted input influences workload and timeouts, leading to potential confidentiality and integrity impact.


anthonyshew:
Closing in favor of https://github.com/vercel/turborepo/pull/11679. Please fix your automations.

PR #11689, opened 2026-02-07
fix: Upgrade tmp to fix arbitrary file write vulnerability

anthonyshew:
## Summary

- Adds `pnpm.overrides` for `tmp: ">=0.2.4"` to fix an arbitrary file write via symlink dir parameter vulnerability in the `tmp` package (TURBO-5240)
- `tmp@0.0.33` was pulled in transitively via `inquirer` → `external-editor` → `tmp` in `@turbo/gen`; now resolved to `tmp@0.2.5`

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "private": true,
   "pnpm": {
     "overrides": {
-      "fast-xml-parser": ">=5.3.4"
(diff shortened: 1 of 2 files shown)


anthonyshew:

PR #11690, opened 2026-02-07
fix: Upgrade brace-expansion to fix ReDoS vulnerability

anthonyshew:
## Summary

- Adds `pnpm.overrides` to pin `brace-expansion@^1` to `1.1.12` and `brace-expansion@^2` to `2.0.2`, fixing a ReDoS vulnerability in both version ranges.
- The vulnerable versions (`1.1.11` and `2.0.1`) were transitive dependencies via `jest` (v1) and `minimatch` (v2).

Resolves TURBO-5238.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,9 @@
   "private": true,
   "pnpm": {
     "overrides": {
-      "fast-xml-parser": ">=5.3.4"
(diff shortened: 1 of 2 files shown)


anthonyshew:

PR #11691, opened 2026-02-07
fix: Upgrade lodash to fix prototype pollution vulnerability

anthonyshew:
## Summary

- Adds `pnpm.overrides` for `lodash` and `lodash-es` to `>=4.17.23`, resolving prototype pollution vulnerabilities (CVE-2021-23337, CVE-2020-28500) in transitive dependencies.

## Context (TURBO-5236)

`lodash@4.17.21` was pulled in transitively via `inquirer` (used by `create-turbo` and other packages). `lodash-es@4.17.21` was pulled in transitively via `mermaid > @mermaid-js/parser > langium > chevrotain` in `docs`. Both versions are vulnerable to prototype pollution; `4.17.23` is the fix.

The docs package already had a direct `lodash-es@4.17.23` dependency, but the transitive resolution from mermaid's sub-dependencies still resolved to `4.17.21`. The override forces all instances to the patched version.

## Verification

`pnpm why lodash --recursive` and `pnpm why lodash-es --recursive` confirm all instances now resolve to `4.17.23`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,9 @@
   "private": true,
   "pnpm": {
     "overrides": {
-      "fast-xml-parser": ">=5.3.4"
+      "fast-xml-parser": ">=5.3.4",
+      "lodash": ">=4.17.23",
+      "lodash-es": ">=4.17.23"
     }
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


anthonyshew:

PR #11693, opened 2026-02-07
fix: Upgrade mdast-util-to-hast to fix unsanitized class attribute vulnerability

anthonyshew:
## Summary

- Fixes TURBO-5237: `mdast-util-to-hast` had an unsanitized class attribute vulnerability in versions `<13.2.1`
- Adds `pnpm.overrides` for `mdast-util-to-hast: ">=13.2.1"` in the root `package.json`, bumping the resolved version from `13.2.0` to `13.2.1` across all transitive dependency chains (`fumadocs-core` → `@shikijs/rehype` / `@shikijs/transformers` / `remark-rehype` → `mdast-util-to-hast`)

### Why an override?

`mdast-util-to-hast` is a transitive dependency pulled in by several packages in the `docs` app. None of the direct dependencies pin it below `13.2.1` — the lockfile simply resolved `13.2.0` before the fix was published. A `pnpm.overrides` entry is the idiomatic way to enforce a minimum version for transitive security fixes without unnecessarily upgrading direct dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "private": true,
   "pnpm": {
     "overrides": {
-      "fast-xml-parser": ">=5.3.4"
+      "fast-xml-parser": ">=5.3.4",
+      "mdast-util-to-hast": ">=13.2.1"
     }
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


anthonyshew:

PR #11694, opened 2026-02-07
fix: Upgrade valibot to fix ReDoS vulnerability

anthonyshew:
## Summary

- Adds a `pnpm.overrides` entry for `valibot: ">=1.2.0"` to fix a ReDoS vulnerability in versions below 1.2.0
- The vulnerable `valibot` is a transitive dependency via `create-turbo > tsdown > rolldown > valibot`
- Override bumps the resolved version from 1.0.0 to 1.2.0, which contains the fix

Refs TURBO-5234

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "private": true,
   "pnpm": {
     "overrides": {
-      "fast-xml-parser": ">=5.3.4"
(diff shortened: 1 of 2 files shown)


anthonyshew:

PR #11695, opened 2026-02-07
fix: Upgrade @babel/runtime to fix ReDoS vulnerability

anthonyshew:
## Summary

- Adds `pnpm.overrides` for `@babel/runtime` and `@babel/runtime-corejs3` to require `>=7.26.10`, fixing a ReDoS vulnerability in earlier versions.

## Context

`@babel/runtime@<7.26.10` contains a Regular Expression Denial of Service vulnerability. Two transitive dependency paths pull in vulnerable versions:

- `@turbo/gen` → `node-plop@0.26.3` → `@babel/runtime-corejs3@7.20.6`
- `@turbo/utils` → `@manypkg/find-root@1.1.0` → `@babel/runtime@7.22.11`

Both packages declare `^7.x` ranges for their `@babel/runtime` dependencies, so overriding to `>=7.26.10` stays within the allowed semver range and carries no compatibility risk. This is preferable to major-version upgrades of `node-plop` or `@manypkg/find-root`, which would require API migration work unrelated to this security fix.

After the override, resolved versions are:
- `@babel/runtime` → `7.28.6`
- `@babel/runtime-corejs3` → `7.29.0`

Closes TURBO-5235

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,8 @@
   "private": true,
   "pnpm": {
     "overrides": {
+      "@babel/runtime": ">=7.26.10",
+      "@babel/runtime-corejs3": ">=7.26.10",
       "fast-xml-parser": ">=5.3.4"
     }
   },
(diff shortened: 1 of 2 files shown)


anthonyshew:

PR #11697, opened 2026-02-07
fix: Upgrade diff to fix DoS vulnerability

anthonyshew:
## Summary

- Fixes TURBO-5241: `diff` (jsdiff) has a DoS vulnerability in `parsePatch` and `applyPatch` ([GHSA-73rr-hh4g-fpgx](https://github.com/advisories/GHSA-73rr-hh4g-fpgx))
- Adds `pnpm.overrides` to upgrade `diff@^4` to `4.0.4` (fixes `ts-node > diff`) and `diff@>=6 <9` to `8.0.3` (fixes `tsdown > diff`)
- `turbo-codemod`'s direct dependency on `diff@5.2.2` is unaffected (v5 is not in either vulnerable range)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,8 @@
   "private": true,
   "pnpm": {
     "overrides": {
+      "diff@^4": "4.0.4",
+      "diff@>=6 <9": "8.0.3",
       "fast-xml-parser": ">=5.3.4"
     }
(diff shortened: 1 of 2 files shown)


anthonyshew:

PR #11699, opened 2026-02-07
perf: Enable pytest-xdist parallel runner for integration tests

anthonyshew:
## Summary

- Integration tests now run in parallel within each CI shard using pytest-xdist (`-n auto`), replacing the serial prysk runner as the default `test` script
- The `test` script is self-contained: creates the venv, installs dependencies, and runs pytest — no dependency on the prysk wrapper for initial setup
- Fixed stale cache path for the prysk venv (`cli/.cram_env` → `turborepo-tests/integration/.cram_env`)

## What changed

**`turborepo-tests/integration/package.json`**: The `test` script now creates a venv inline, installs prysk + pytest + xdist, and runs `pytest -n auto`. The old `test:parallel` and `pretest:parallel` scripts are removed. A `test:serial` script is preserved for cases where serial execution is needed.

**`.github/workflows/turborepo-test.yml`**: Cache path corrected to `turborepo-tests/integration/.cram_env` with a bumped key (`-v2`) to invalidate stale caches.

## Expected impact

CI runners typically have 2-4 cores, so heavy shards like `other/` (32 tests) and `run/` (24 tests) should see a ~2-4x speedup from within-shard parallelism.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,10 @@ file_descriptor_set.bin
 # include test fixtures for vt100
 !crates/turborepo-vt100/tests/data
 
+# Shared corepack state for integration tests
+.corepack-integration
+.corepack-cache
+
 # DevContainer cache directories
 .devcontainer/cache/
 .devcontainer/cargo/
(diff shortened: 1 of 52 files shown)


anthonyshew:
We've tried this before and could never get it to work for bizarre reasons. Looking at the CI times we're pulling now, I think we can skip out on this for now. We'll look for better options in the future that aren't as thorny.

PR #11759, opened 2026-02-10
fix: Add Vary: Accept header to docs markdown endpoint

molebox:
## Summary

- Add `Vary: Accept` header to the turborepo docs `llms.md` route handler so CDNs correctly cache separate HTML and markdown response variants

### Why

Without `Vary: Accept`, a CDN may serve a cached markdown response to a browser (or vice versa) because it treats requests with different `Accept` headers as the same cache key.

### File changed

- `apps/docs/app/[lang]/llms.md/[[...slug]]/route.ts` — Add `Vary: Accept` header

## Test plan

- [ ] Verify response from `/llms.md` includes `Vary: Accept` header
- [ ] Verify markdown content is served correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/docs/app/[lang]/llms.md/[[...slug]]/route.ts
+++ b/apps/docs/app/[lang]/llms.md/[[...slug]]/route.ts
@@ -17,7 +17,7 @@ export async function GET(
   return new Response(await getLLMText(page), {
     headers: {
       "Content-Type": "text/markdown",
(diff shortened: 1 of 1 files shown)


anthonyshew:
@copilot Fix the formatting as reported by the CI job.

PR #11768, opened 2026-02-10
fix: Override lodash-es to resolve prototype pollution vulnerability

anthonyshew:
## Summary

- Add pnpm override for `lodash-es` to `>=4.17.23` to resolve a prototype pollution vulnerability (GHSA-xxjr-mmjv-4gpg, moderate severity)

## Why

`chevrotain@11.0.3`, a transitive dependency via `mermaid > @mermaid-js/parser > langium`, pulls in `lodash-es@4.17.21` which is vulnerable to prototype pollution. While `apps/docs` already has `lodash-es@4.17.23` as a direct dependency, the transitive copy through chevrotain remains at the vulnerable version without an override.

The latest chevrotain (11.1.1) ships with `lodash-es@4.17.23`, but upgrading it directly would require upstream changes in langium/mermaid. A pnpm override is the least invasive fix.

Verified: `pnpm install` succeeds and `pnpm build --filter=docs` builds cleanly.

Refs: TURBO-5267

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "private": true,
   "pnpm": {
     "overrides": {
-      "fast-xml-parser": ">=5.3.4"
+      "fast-xml-parser": ">=5.3.4",
+      "lodash-es": ">=4.17.23"
     }
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


anthonyshew:
Superceded by https://github.com/vercel/turborepo/pull/11770

PR #11794, opened 2026-02-11
perf: Replace `git fetch --tags` with `git ls-remote` in release pipeline

anthonyshew:
## Summary

- **stage job**: Replace `git fetch origin --tags` + `git tag --list` with `git ls-remote --tags` to find the previous release tag without downloading objects
- **Makefile `stage-release`**: Replace `git fetch origin --tags` + `git rev-parse` with a single `git ls-remote --tags` call for tag existence check
- **npm-publish job**: Remove `git fetch origin --tags` entirely — nothing in `make publish-turbo` uses git tags

`git fetch --tags` downloads all tag refs and their referenced objects. In a repo with hundreds of tags, this takes ~48s. `git ls-remote --tags` queries tag names over the existing connection without transferring objects, completing in <1s.

### DIFF
--- a/.github/workflows/turborepo-release.yml
+++ b/.github/workflows/turborepo-release.yml
@@ -163,7 +163,7 @@ jobs:
       - name: Get previous tag
         id: previous-tag
         run: |
-          # Use ls-remote to query tags without fetching objects (avoids ~48s fetch)
+          # Use ls-remote to query tags without fetching objects
           PREV_TAG=$(git ls-remote --tags origin 'refs/tags/v*-canary.*' \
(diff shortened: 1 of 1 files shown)


.github/workflows/turborepo-release.yml
```diff
@@ -166,9 +163,14 @@ jobs:
       - name: Get previous tag
         id: previous-tag
         run: |
-          PREV_TAG=$(git tag --list 'v*-canary.*' --sort=-version:refname | head -n 1)
+          # Use ls-remote to query tags without fetching objects (avoids ~48s fetch)
```
anthonyshew:
```suggestion
          # Use ls-remote to query tags without fetching objects
```

PR #11798, opened 2026-02-11
fix: Handle subpath package imports in `turbo boundaries`

### LINKED ISSUE #11797: `turbo boundaries` fails on subpath package imports since 2.8.4 (2.8.3-canary.13)
### Verify canary release

https://github.com/pawelblaszczyk5/turborepo-boundaries-reproduction

2.8.7-canary.1

### Environment information

```block
CLI:
   Version: 2.8.7-canary.1
   Path to executable: /Users/pawelblaszczyk/dev/turborepo-boundaries-reproduction/node_modules/.pnpm/turbo-darwin-arm64@2.8.7-canary.1/node_modules/turbo-darwin-arm64/bin/turbo
   Daemon status: Running
   Package manager: pnpm

Platform:
   Architecture: aarch64
   Operating system: macos
   WSL: false
   Available memory (MB): 3649
   Available CPU cores: 8

Environment:
   CI: None
   Terminal program (TERM_PROGRAM): vscode
   Terminal program version (TERM_PROGRAM_VERSION): 1.109.0
   Shell (SHELL): /bin/zsh
   stdin: false
```

### Expected behavior

Hello! 

Today, after upgrading I've spotted `turbo boundaries` failing on a lot of imports ion my monorepo. What I pinpointed to be the issue is any subpath import. If I import not from top-level of package it fails.

In provided example, my pkg has this inside of the `package.json`:

```json
  "exports": {
    "./test": "./src/index.ts"
  },
```

Importing from there fails the check:

Image

It also impacts any standard dependencies (non-monorepo) where subpath exports are commonly used.

### Actual behavior

`turbo boundaries` fails on explicitly allowed exports since the update.

### To Reproduce

1. Create a package which exports something from subpath, e.g. 

```json
  "exports": {
    "./test": "./src/index.ts"
  },
```

2. Import this package from any other one
3. Run `turbo boundaries` and observe it failing on it, despite it being explicitly allowed.

### Additional context

I'm not sure how canaries work here, some of 2.8.3 canaries were after 2.8.3? 😄 But what I've pinpointed here is `2.8.3-canary.13` - 12 still works, 13 fails. `2.8.4-canary.0` also fails just for the consistency of checking

anthonyshew:
## Summary

Fixes #11797

`turbo boundaries` incorrectly flagged subpath package imports (e.g. `import { x } from "pkg/test"` or `import { x } from "@scope/pkg/sub"`) as boundary violations since 2.8.4.

`is_potential_package_name` used a regex that only matched bare package names (`lodash`, `@scope/pkg`) — the `$` anchor rejected anything with additional path segments. This caused subpath imports to bypass the package-name guard in `check_import_as_tsconfig_path_alias`, where the full module resolver would resolve them to paths outside the importing package and emit false `ImportLeavesPackage` diagnostics.

The fix extracts the base package name (via the existing `get_package_name`) before checking against the regex, so `pkg-a/test` is recognized as a package import just like `pkg-a`.

## Test Plan

- Added integration test cases to the `boundaries` fixture: a subpath import from a declared dependency (should pass) and a subpath import from an undeclared dependency (should correctly report "not a dependency")
- Added unit test cases for `is_potential_package_name` and `tsconfig_alias_check_skips_bare_package_names` with subpath imports
- Verified against the exact reproduction repo from the issue

### DIFF
--- a/crates/turborepo-boundaries/src/imports.rs
+++ b/crates/turborepo-boundaries/src/imports.rs
@@ -270,7 +270,7 @@ pub(crate) fn check_file_import(
     }
 }
 
-fn get_package_name(import: &str) -> String {
+pub(crate) fn get_package_name(import: &str) -> String {
     if import.starts_with("@") {
         import.split('/').take(2).join("/")
     } else {
@@ -376,6 +376,9 @@ mod test {
     #[test_case("lodash" ; "bare package name lodash")]
     #[test_case("@scope/package" ; "scoped package name")]
     #[test_case("@types/node" ; "types package name")]
+    #[test_case("lodash/fp" ; "subpath import")]
+    #[test_case("@scope/package/sub" ; "scoped subpath import")]
+    #[test_case("@scope/package/deeply/nested" ; "scoped deeply nested subpath import")]
     fn tsconfig_alias_check_skips_bare_package_names(import: &str) {
         let (resolver, package_name, span, file_content, mut result) =
             make_tsconfig_alias_test_args(import);
--- a/crates/turborepo-boundaries/src/lib.rs
+++ b/crates/turborepo-boundaries/src/lib.rs
@@ -316,7 +316,8 @@ impl BoundariesChecker {
     }
 
     fn is_potential_package_name(import: &str) -> bool {
-        PACKAGE_NAME_REGEX.is_match(import)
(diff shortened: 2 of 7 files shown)


anthonyshew:
The test is a flake, so merging.

PR #11807, opened 2026-02-11
docs: Geistdocs update

haydenbleasel:
Partial Geistdocs update.

### DIFF
--- a/apps/docs/app/[lang]/layout.tsx
+++ b/apps/docs/app/[lang]/layout.tsx
@@ -21,7 +21,7 @@ const Layout = async ({ children, params }: LayoutProps<"/[lang]">) => {
 
   return (
     <html
-      className={cn(sans.variable, mono.variable, "scroll-smooth antialiased")}
+      className={cn(sans.variable, mono.variable, "antialiased")}
       lang={lang}
       suppressHydrationWarning
     >


apps/agents/lib/runs.ts
```diff
@@ -1,4 +1,4 @@
-import { put, list, get } from "@vercel/blob";
+import { put, list } from "@vercel/blob";
```
anthonyshew:
```suggestion
import { put, list, get } from "@vercel/blob";
```

apps/docs/app/[lang]/layout.tsx
```diff
@@ -21,7 +21,7 @@ const Layout = async ({ children, params }: LayoutProps<"/[lang]">) => {
 
   return (
     <html
-      className={cn(sans.variable, mono.variable, "antialiased")}
+      className={cn(sans.variable, mono.variable, "scroll-smooth antialiased")}
```
Comment (anthonyshew) [inline apps/docs/app/[lang]/layout.tsx]: I got rid of this because it was creating a bad experience. Clicking around different links had bizarre behavior.

apps/docs/components/ai-elements/message.tsx
```diff
@@ -310,6 +308,7 @@ export const MessageResponse = memo(
         "size-full [&>*:first-child]:mt-0 [&>*:last-child]:mb-0",
         className
       )}
+      plugins={{ code, mermaid, math, cjk }}
```
anthonyshew, line 311: Do I need cjk? This site is only English.

anthonyshew: We're going to lazyload this in a follow-up PR.

PR #11825, opened 2026-02-13
feat: Compile @turbo/gen to standalone binary via `bun build --compile`

anthonyshew:
## Summary

Compiles `@turbo/gen` into a standalone Bun binary to eliminate the ESM/CJS interop issues that have blocked upgrading `node-plop` and other dependencies.

## Motivation

`@turbo/gen` is stuck on `node-plop@0.26.3` (with known security vulnerabilities) because newer versions switched from `require()` to `import()` for loading user config files, which breaks TypeScript configs in `"type": "commonjs"` projects (the default for Next.js monorepos). Two prior upgrade attempts were reverted. Several other deps (`execa`, `ora`) are also pinned to old CJS-compatible versions.

The root cause: `@turbo/gen` is a CJS package that dynamically loads user-authored TypeScript config files at runtime via `node-plop`, and the module resolution semantics differ between CJS and ESM in ways that break real user projects.

## Approach

Bun's runtime handles both CJS and ESM natively and transpiles TypeScript without external tooling. By compiling `@turbo/gen` with `bun build --compile`, user generator configs (`.ts`, `.js`, `.cjs`) load correctly regardless of the project's `"type"` field — no `tsx` registration, no sucrase, no interop hacks.

## What changed

- **Build**: `bun build --compile` replaces tsdown for the CLI binary. tsdown is kept for DTS-only generation (the `PlopTypes` re-export).
- **Templates**: Embedded as string constants at build time via `scripts/embed-templates.ts`, eliminating the `__dirname`-relative copy that doesn't work in compiled binaries.
- **`plop.ts`**: Removed `tsx/cjs/api` registration (Bun handles `.ts` natively). Made `nodePlop()` and `plop.load()` calls async — Bun's `require()` of CJS modules with Babel interop returns Promises.
- **Distribution**: Platform binaries via `@turbo/gen-{os}-{arch}` optional dependency packages (same pattern as the Rust `turbo` binary).
- **`bin/turbo-gen`**: Node.js shim that resolves the correct platform binary via `require.resolve()` and executes it. Mirrors `packages/turbo/bin/turbo`.
- **Releaser**: Extended `@turbo/releaser` with a `gen` subcommand for packing and publishing `@turbo/gen` platform packages.
- **CI**: Added `build-gen` job with Bun cross-compilation for all 5 target platforms.
- **`package.json`**: All runtime deps moved to devDependencies (bundled into binary). `tsx` removed. `"files"` narrowed to only ship the shim + types (no 62MB binary in the main package).

## Backward compatibility

The Rust CLI (`generate.rs`) still uses `npx --yes @turbo/gen@{tag} raw ...`. The flow becomes:

```
turbo gen → npx @turbo/gen → bin/turbo-gen (Node.js shim) → @turbo/gen-darwin-arm64 (Bun binary)
```

No Rust-side changes needed. The `--version`, `--help`, `raw run`, and `raw workspace` commands all produce identical output.

## Verified

- **Exported types**: `diff` of old vs new `dist/types.d.ts` is empty
- **23 tests pass**: 9 existing unit tests + 14 new integration tests covering all ESM/CJS combinations
- **`check-types`** passes
- **`npm pack`** ships only the shim (2KB) + types — not the 62MB binary
- **Binary tested across**: CJS+.ts, ESM+.ts, ESM+.ts (no gen package.json), no-type+.js, CJS+.cjs, ESM+.js, ESM+.cjs

## How to test

```bash
# Build
turbo build --filter=@turbo/gen

# Run tests
turbo test --filter=@turbo/gen

# Manual test: run the binary directly against this repo
packages/turbo-gen/dist/turbo-gen --version
packages/turbo-gen/dist/turbo-gen --help
```


codegoyf:
This change broke the running of all of our generators at my job that we use extensively across multiple repos/apps. For such a massive change, why would this be published as a patch version? This is quite clearly a breaking change in the `turbo/gen` package IMO.

This seems like pretty poor usage of semantic versioning to me.

PR #11844, opened 2026-02-13
fix: Prevent pnpm overrides from corrupting resolved peer-dep variants

frankeld:
### Description

When a pnpm lockfile has overrides (e.g. `foo: ~1.0.0`), `turbo prune` can produce a broken lockfile that's missing snapshot entries for packages with multiple peer-dependency variants. This happens because `apply_overrides` corrupts already-resolved version strings like `1.0.0(peer-a@1.0.0)(peer-b@2.0.0)` by replacing them with the semver range from the override.

This fix makes `resolve_specifier` prefer the original specifier when it already matches a snapshot, so overrides only kick in when the specifier doesn't already resolve to a known package.

These issues are reoccurring for many different users. See ongoing bug reports on this closed issue: https://github.com/vercel/turborepo/issues/3382#issuecomment-3431554089

Here's a repository with a "real" minimal reproduction: https://github.com/frankeld/turbo-prune-repro

### Testing Instructions

New unit tests in `crates/turborepo-lockfiles/src/pnpm/data.rs`:

```
cargo test -p turborepo-lockfiles test_override_does_not_break_transitive_peer_variant
```

This test sets up a v9 lockfile with an override (`foo: ~1.0.0`) and two peer-dep variants of `foo`, then verifies that:
- `resolve_package` still resolves the transitive variant (`foo@1.0.0(peer-a@1.0.0)(peer-b@2.0.0)`)
- The transitive closure includes both variants and middleware

### DIFF
--- a/apps/docs/lib/ai-agent-detection.ts
+++ b/apps/docs/lib/ai-agent-detection.ts
@@ -1,0 +1,168 @@
+/**
+ * AI Agent Detection Utility
+ *
+ * Multi-signal detection for AI agents/bots. Used to serve markdown
+ * responses when agents request docs pages.
+ *
+ * Three detection layers:
+ * 1. Known UA patterns (definitive) — curated from https://bots.fyi/?tags=ai_assistant
+ * 2. Signature-Agent header (definitive) — catches ChatGPT agent (RFC 9421)
+ * 3. Missing browser fingerprint heuristic — catches unknown bots
+ *
+ * Optimizes for recall over precision: serving markdown to a non-AI bot
+ * is low-harm; missing an AI agent means a worse experience.
+ *
+ * Last reviewed: 2026-03-20 against bots.fyi + official vendor docs
+ */
+
+// Layer 1: Known AI agent UA substrings (lowercase).
+const AI_AGENT_UA_PATTERNS = [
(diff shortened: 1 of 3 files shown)


anthonyshew:
## E2E check-lockfiles validation

Added a `pnpm-override-peer-variant` fixture to `lockfile-tests/fixtures/` using the real reproduction from https://github.com/frankeld/turbo-prune-repro.

**Without the fix**, pruning `@repo/backend` and `server` fails:
```
ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY  Broken lockfile: no entry for
'inngest@3.51.0(@opentelemetry/core@2.5.1(...))(express@5.2.1)(fastify@4.29.1)
(h3@1.15.5)(hono@4.11.9)(next@15.5.12(...))(typescript@5.9.3)(zod@3.25.76)'
```

**With the fix**, all 3 workspaces pass. Full suite (221 tests across 71 fixtures) also clean:
```
Passed:            218
Expected failures: 3   (pre-existing bun issues)
Unexpected fails:  0
```

Branch with fixture + fix: [`shew/review-11844`](https://github.com/vercel/turborepo/tree/shew/review-11844)

PR #11860, opened 2026-02-15
chore: Use 2024 edition in more packages, do not ignore some clippy l…

ognevny:
…ints

### Description

with updated tonic build it's now possible to follow more clippy lints in truborepo-lib.

packages migrated to 2024 edition were easy to change, just putting `unsafe {}` for `env::{set,remove}_var` and let-chains

### Testing Instructions

CI


anthonyshew (approved):
Thank you! Had this on my to-do list - appreciate you beating me to it!

PR #11863, opened 2026-02-15
fix: Handle premature tag creation by external release notes service

anthonyshew:
## Summary

There's an external service that creates git tags that I just learned about. This makes it so that our release pipeline is the absolute truth. I'm handling the external git tagging separately.

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -73,7 +73,7 @@ create-release-tag:
 			echo "Tag v$(TURBO_VERSION) already exists on remote at the correct commit. Skipping."; \
 		else \
 			echo "Warning: Tag v$(TURBO_VERSION) exists on remote but points to $$REMOTE_TAG_SHA, expected $$LOCAL_SHA." && \
-			echo "Replacing tag (likely created prematurely by an external service)..." && \
+			echo "Replacing incorrect tag..." && \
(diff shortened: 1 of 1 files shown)


cli/Makefile
```diff
@@ -67,14 +67,17 @@ stage-release:
...
 	if [ -n "$$REMOTE_TAG_SHA" ]; then \
-		LOCAL_SHA=$$(git rev-parse HEAD); \
 		if [ "$$REMOTE_TAG_SHA" = "$$LOCAL_SHA" ]; then \
 			echo "Tag v$(TURBO_VERSION) already exists on remote at the correct commit. Skipping."; \
 		else \
-			echo "Error: Tag v$(TURBO_VERSION) exists on remote but points to $$REMOTE_TAG_SHA, expected $$LOCAL_SHA." && \
-			echo "This tag may have been pushed prematurely by a previous step." && \
-			exit 1; \
+			echo "Warning: Tag v$(TURBO_VERSION) exists on remote but points to $$REMOTE_TAG_SHA, expected $$LOCAL_SHA." && \
+			echo "Replacing tag (likely created prematurely by an external service)..." && \
```
anthonyshew:
```suggestion
			echo "Replacing incorrect tag..." && \
```

PR #11872, opened 2026-02-16
perf: Optimize hot-path hash computation by avoiding clones and using unstable sorts

anthonyshew:
## Summary

- Adds `From<&FileHashes>` impl so file inputs can be hashed by reference instead of cloning the entire HashMap per task
- Adds `LockFilePackagesRef` type so `get_external_deps_hash` borrows from the transitive dependency `HashSet` instead of cloning every `Package` (2 `String` allocations each)
- Replaces `sort_by` with `sort_unstable_by` throughout Cap'n Proto serialization paths — since all sorted collections have unique keys, stability is unnecessary, and unstable sort avoids auxiliary memory allocation

These changes target the per-task hash computation hot path that runs for every task in `turbo run`. In a large monorepo, the cumulative cost of HashMap clones and extra allocations is significant because each task independently serializes its inputs through Cap'n Proto before hashing with xxHash64.

## Benchmark results

Tested in a large monorepo:

```
Benchmark 1: ../../target/release/turbo run typecheck --dry --skip-infer --no-daemon
  Time (mean ± σ):      9.217 s ±  0.162 s    [User: 66.243 s, System: 18.195 s]
  Range (min … max):    8.784 s …  9.347 s    10 runs

Benchmark 2: turbo run typecheck --dry --skip-infer --no-daemon
  Time (mean ± σ):     12.277 s ±  2.471 s    [User: 77.811 s, System: 25.350 s]
  Range (min … max):    9.595 s … 16.435 s    10 runs

Summary
  ../../target/release/turbo run typecheck --dry --skip-infer --no-daemon ran
    1.33 ± 0.27 times faster than turbo run typecheck --dry --skip-infer --no-daemon
```

**~25% reduction in wall time** (9.2s vs 12.3s mean), with a notable improvement in both user CPU time (66s vs 78s) and system time (18s vs 25s), confirming reduced allocation pressure.

## Testing

- All existing hash tests pass — the reference-based impls produce identical hashes to the owned variants
- New tests verify `From<&FileHashes>` and `LockFilePackagesRef` produce the same hashes as their owned counterparts
- Integration tests show no changes to evaluated hashes


crates/turborepo-hash/src/lib.rs
```diff
@@ -105,6 +105,12 @@ pub struct GlobalHashable<'a> {
 
 pub struct LockFilePackages(pub Vec<turborepo_lockfiles::Package>);
 
+/// Reference-based variant of [`LockFilePackages`] that avoids cloning packages.
+///
+/// This is used in hot paths where we have a borrowed collection of packages
+/// (e.g. from a `HashSet`) and want to hash them without cloning each one.
```
anthonyshew:
```suggestion
```

crates/turborepo-hash/src/lib.rs
```diff
@@ -202,7 +240,54 @@ impl From<FileHashes> for Builder<HeapAllocator> {
             // and set the entries in the capnp message
 
             let mut hashable: Vec<_> = file_hashes.into_iter().collect();
-            hashable.sort_by(|(path_a, _), (path_b, _)| path_a.cmp(path_b));
+            // sort_unstable_by is faster than sort_by (no auxiliary allocation) and
+            // produces identical results here since HashMap keys are unique.
```
anthonyshew:
```suggestion
```

crates/turborepo-hash/src/lib.rs
```diff
@@ -202,7 +240,52 @@ impl From<FileHashes> for Builder<HeapAllocator> {
...
+
+        canon_builder
+    }
+}
+
+/// Reference-based hashing for [`FileHashes`] that avoids cloning the inner HashMap.
+///
+/// On the hot path, tasks need their file hashes both for computing the task hash
+/// and for storing in the expanded hashes tracker. Previously this required cloning
+/// the entire HashMap just to hash it. This implementation borrows instead.
```
anthonyshew:
```suggestion
```

crates/turborepo-run-summary/src/task_factory.rs
```diff
@@ -229,22 +229,19 @@ fn workspace_relative_log_file(task_name: &str) -> turbopath::AnchoredSystemPath
...
     let Some(transitive_dependencies) = transitive_dependencies else {
         return "".into();
     };
 
-    let mut transitive_deps = Vec::with_capacity(transitive_dependencies.len());
-
-    for dependency in transitive_dependencies.iter() {
-        transitive_deps.push(dependency.clone());
-    }
+    // Collect references instead of cloning each Package (which has two Strings).
```
anthonyshew:
```suggestion
```

PR #11883, opened 2026-02-17
feat: Make `--profile` and `--anon-profile` filename optional

anthonyshew:
## Summary

- `--profile` and `--anon-profile` no longer require a filename argument
- When no filename is provided, defaults to `profile.<unix_timestamp_ms>` (and `profile.<unix_timestamp_ms>.md` for the markdown summary)

### Usage

```sh
# Before: filename required
turbo run build --profile my-trace.json

# Now: filename is optional
turbo run build --profile
# Produces profile.1739836800000 and profile.1739836800000.md
```

### Testing

New tests cover:
- CLI parsing accepts `--profile` without a value
- Default filename generation produces `profile.<timestamp>` pattern
- Explicit filenames still work as before
- `--profile` and `--anon-profile` remain mutually exclusive

### DIFF
--- a/apps/agents/next-env.d.ts
+++ b/apps/agents/next-env.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
+import "./.next/types/routes.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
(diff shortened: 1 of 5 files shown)


crates/turborepo-lib/src/cli/mod.rs
```diff
@@ -3158,10 +3184,13 @@ mod test {
 
     #[test]
     fn test_profile_usage() {
-        assert!(Args::try_parse_from(["turbo", "build", "--profile", ""]).is_err());
-        assert!(Args::try_parse_from(["turbo", "build", "--anon-profile", ""]).is_err());
+        // Without a filename, profile should still be accepted
```
anthonyshew, line 3187:
```suggestion
```

anthonyshew:
The Windows failure is a flake. Merging.

PR #11887, opened 2026-02-17
perf: Batch per-package git subprocess calls into repo-wide index

anthonyshew:
## Summary

Turborepo spawns 2 git subprocesses (`git ls-tree` + `git status`) per package during file hashing. For large monorepos, this means hundreds of `fork`/`exec`/`waitpid` cycles that dominate the hot path. This PR replaces them with 2 repo-wide git commands whose results are cached and filtered per-package in-memory.

### 1. Profiling

CPU profiles of `turbo run --dry` on a large monorepo (~300 packages) showed that `append_git_status` (89.5% self time, 28 calls) and `git_ls_tree` (59.7%, 28 calls) were the top two functions by self time in the hot-cache path. Both are subprocess spawns that happen once per package in `get_package_file_hashes_from_index`.

### 2. Batching git calls into a repo-wide index

`RepoGitIndex` runs `git ls-tree -r -z HEAD` and `git status --untracked-files --no-renames -z` once at the git repo root (in parallel via `std::thread::scope`), then stores:
- A `BTreeMap<RelativeUnixPathBuf, String>` for committed file hashes (enables O(log N) prefix-range lookups per package)
- A sorted `Vec<RepoStatusEntry>` for working tree changes (enables binary-search + early-exit per package)

When `get_package_file_hashes_from_index` is called for a package, it filters the cached data by prefix range instead of spawning subprocesses.

### 3. Small repo opt-out

For repos with fewer than 16 packages, the overhead of repo-wide git commands + BTreeMap construction exceeds the per-package subprocess savings. These repos skip the index entirely and use the original code path.

### 4. File descriptor safety

Removing the per-package subprocesses eliminated the bottleneck that naturally throttled rayon parallelism — each subprocess had its own fd table and blocked its thread for hundreds of milliseconds. Without that bottleneck, all rayon threads reach the file-IO phase simultaneously, which can exhaust the OS file descriptor limit in two ways:

- **globwalk** (packages with custom inputs) uses rayon's `into_par_iter` internally, creating nested parallel directory traversal that holds many directory handles open. An `IoSemaphore` serializes these operations (limit=1). Packages without custom inputs — the common case — run fully parallel with no semaphore.

- **Cascading failures**: when `hash_objects` or globwalk hits EMFILE, the old code fell back to manual hashing via `ignore::WalkBuilder`, which opened even more files. `Error::is_resource_exhaustion` now detects OS resource errors (including through wax's wrapped error chains) and propagates them directly.

So, TL;DR: we now have a semaphore to ensure we don't exhaust fd.

## Benchmarks

All benchmarks: `hyperfine` with `--warmup`, `--dry` flag (no task execution), `--skip-infer --no-daemon`.

### Small repo (~6 packages)

```
Benchmark 1 (this PR):  652.3 ms ± 48.3 ms
Benchmark 2 (main):     667.2 ms ± 44.2 ms

Summary: 1.02 ± 0.10 times faster (neutral — index is skipped)
```

### Medium repo (~100 packages)

```
Benchmark 1 (this PR):  1.824 s ± 0.157 s    [User: 5.504 s, System: 0.910 s]
Benchmark 2 (main):     2.007 s ± 0.168 s    [User: 8.746 s, System: 3.321 s]

Summary: 1.10 ± 0.13 times faster
```

### Large repo (~300 packages)

```
Benchmark 1 (this PR):  6.796 s ± 0.046 s    [User: 40.302 s, System: 1.938 s]
Benchmark 2 (main):     9.318 s ± 0.092 s    [User: 67.119 s, System: 21.107 s]

Summary: 1.37 ± 0.02 times faster
```

The System time reductions on the large repo reflect the elimination of hundreds of subprocess spawns, and how the speedup scales with package count.

### DIFF
--- a/crates/turborepo-scm/src/package_deps.rs
+++ b/crates/turborepo-scm/src/package_deps.rs
@@ -131,12 +131,6 @@ impl GitRepo {
             return self.get_package_file_hashes_from_index(turbo_root, package_path, repo_index);
         }
 
-        // Paths below use globwalk, which does parallel directory traversal
-        // via rayon and can hold hundreds of directory handles open. Before
-        // the repo index optimization, per-package git subprocesses blocked
-        // rayon threads, limiting how many globwalks ran at once. With the
-        // index, threads reach globwalk near-instantly. Serialize globwalk
-        // operations to prevent fd exhaustion.
         let _permit = repo_index.map(|idx| idx.io_semaphore.acquire());
 
         // we have inputs, but no $TURBO_DEFAULT$


crates/turborepo-scm/src/package_deps.rs
```diff
@@ -109,12 +122,23 @@ impl GitRepo {
...
-            return self.get_package_file_hashes_from_index(turbo_root, package_path);
+            return self.get_package_file_hashes_from_index(turbo_root, package_path, repo_index);
         }
 
+        // Paths below use globwalk, which does parallel directory traversal
+        // via rayon and can hold hundreds of directory handles open. Before
+        // the repo index optimization, per-package git subprocesses blocked
+        // rayon threads, limiting how many globwalks ran at once. With the
+        // index, threads reach globwalk near-instantly. Serialize globwalk
+        // operations to prevent fd exhaustion.
```
anthonyshew:
```suggestion
```

PR #11898, opened 2026-02-18
perf: Optimize pnpm lockfile transitive closure resolution

anthonyshew:
## Summary

- Optimizes lockfile dependency resolution for pnpm repositories, reducing `transitive_closure` wall time by ~20%
- Five targeted changes to `turborepo-lockfiles` that eliminate redundant work in the hot path

## Benchmarks

**Large pnpm monorepo (~1000 packages):**
```
Benchmark 1 (this branch):   5.924s ± 0.231s
Benchmark 2 (main):          7.117s ± 0.147s

Summary: 1.20 ± 0.05 times faster
```

**Medium pnpm monorepo:**
```
Benchmark 1 (this branch):   1.326s ± 0.059s
Benchmark 2 (main):          1.548s ± 0.227s

Summary: 1.17 ± 0.18 times faster
```

User-space CPU time dropped from ~45s to ~25s on the large repo, confirming the reduction is in compute, not I/O.

## What changed

All changes are in `crates/turborepo-lockfiles`.

**Cross-workspace resolve cache** (`src/lib.rs`): `all_transitive_closures` runs workspace closures in parallel via rayon, but each workspace independently resolves the same transitive packages. A shared `DashMap` cache now deduplicates `resolve_package` calls across workspaces, reducing 726k calls to the number of unique `(workspace, name, specifier)` triples.

**HashMap for package/snapshot lookups** (`src/pnpm/data.rs`): Switched the `packages` and `snapshots` maps from `BTreeMap` to `HashMap` for O(1) lookups on the hot path. A custom serde serializer converts back to `BTreeMap` for deterministic YAML output.

**Pre-built dependency index** (`src/pnpm/data.rs`): The `all_dependencies` method was called 329k times, each time chaining iterators across dependency maps and cloning every key/value. A `HashMap<String, HashMap<String, String>>` is now built once at parse time and lookups just clone from the cache.

**Reusable key buffer** (`src/pnpm/data.rs`): `format_key` allocated a new `String` via `format!()` on every call (~2M allocations). `resolve_package` and `resolve_specifier` now share a single `String` buffer across all their internal lookups.

**Cached lockfile version** (`src/pnpm/data.rs`, `src/pnpm/mod.rs`): `version()` was recomputed via string matching on every call. Now computed once at parse time and stored.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7361,6 +7361,7 @@ dependencies = [
  "biome_formatter",
  "biome_json_formatter",
  "biome_json_parser",
+ "dashmap 5.5.3",
  "insta",
  "itertools 0.10.5",
  "nom",
--- a/crates/turborepo-lockfiles/Cargo.toml
+++ b/crates/turborepo-lockfiles/Cargo.toml
@@ -13,6 +13,7 @@ workspace = true
 biome_formatter = { workspace = true }
 biome_json_formatter = { workspace = true }
 biome_json_parser = { workspace = true }
+dashmap = { workspace = true }
 itertools = { workspace = true }
 nom = "7"
 pest = "2.8.6"
--- a/crates/turborepo-lockfiles/src/lib.rs
+++ b/crates/turborepo-lockfiles/src/lib.rs
@@ -26,6 +26,7 @@ use std::{
 
 pub use berry::{Error as BerryError, *};
 pub use bun::{BunLockfile, bun_global_change};
+use dashmap::DashMap;
 pub use error::Error;
 pub use npm::*;
 pub use pnpm::{PnpmLockfile, pnpm_global_change, pnpm_subgraph};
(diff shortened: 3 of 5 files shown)


kevva:
@anthonyshew, after upgrading to the newest version of `turbo` my Docker builds are now failing with `pnpm`. If a package has some module as a `peerDependency` (optional) and a `devDependency` and my app depends on this package but doesn't install the `peerDependency` I'm now getting a lockfile error. 

```
> ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY  Broken lockfile: no entry for '<pkg>@<version>' in pnpm-lock.yaml
```

To get it working I have to add the `peerDependency` as a `devDependency` in my app. I think `turbo` should take into account `devDependencies` of packages too, no, not just the pruned app/package?

anthonyshew:
@kevva Please open an issue with a minimal reproduction, thank you!

kevva:
@anthonyshew, done in https://github.com/vercel/turborepo/issues/12002.

anthonyshew:
Thank you!

PR #11916, opened 2026-02-19
perf: Reduce allocations in SCM hashing, glob preprocessing, and cache lookups

anthonyshew:
## Summary

Reduce heap allocations and syscalls across the hot paths of `turbo run`. These are mechanical changes — no behavioral differences, no new APIs.

### Benchmarks (`--dry` runs, `--skip-infer`, 10 runs each, 5 warmup)

| Repo | Packages | Tasks | Before | After | Delta |
|------|----------|-------|--------|-------|-------|
| Large | ~1000 | 1690 | 2.179s ± 0.073s | 2.130s ± 0.049s | **1.02x faster** |
| Medium | ~120 | ~200 | 1.235s ± 0.098s | 1.216s ± 0.079s | **1.02x faster** |
| Small | ~5 | ~5 | 834.1ms ± 28.4ms | 816.8ms ± 22.8ms | **1.02x faster** |

### Changes

**SCM parsing** (`ls_tree.rs`, `status.rs`): `entry.hash.to_vec()` + `String::from_utf8()` allocated twice per entry (intermediate `Vec<u8>` then `String`). Now uses `str::from_utf8()` + `.to_owned()` for a single allocation. BufReader buffer increased from 8KB to 64KB to reduce `read()` syscalls on large git output.

**File hashing** (`hash_object.rs`): `hashes.reserve(to_hash.len())` pre-allocates the HashMap to avoid rehashing during insertion.

**Package deps** (`package_deps.rs`): Input globs were cloned to `String` just to iterate them. Now works with `&str` references and reuses a single `String` buffer for the `"{package_path}/{glob}"` join instead of allocating per-iteration via `.join("/")`. Capacity hints added to `inclusions`, `exclusions`, `to_hash`, and `hashes` Vecs/HashMaps. When include globs overlap with the git index (the `$TURBO_DEFAULT$` + explicit inputs case), files already known from the index are skipped instead of being re-hashed.

**Task hashing** (`lib.rs`): Dependency hash deduplication now collects `&str` references under the mutex lock instead of cloning each `String` into the `HashSet`. Owned strings are only allocated after dedup, halving allocations. Capacity hint added.

**Glob preprocessing** (`globwalk/lib.rs`): Capacity hints on include/exclude path Vecs. Exclude path processing avoids an unnecessary `.to_string()` by using `Cow::into_owned()` which is free when the Cow is already borrowed.

**Cache lookups** (`fs.rs`): `FSCache::exists()` was doing 3 `format!()` + `join_component()` allocations per call (called once per task). Now reuses a single `String` buffer, truncating and re-appending suffixes for `.tar`, `.tar.zst`, and `-meta.json`.

### DIFF
--- a/crates/turborepo-scm/src/package_deps.rs
+++ b/crates/turborepo-scm/src/package_deps.rs
@@ -204,9 +204,6 @@ impl GitRepo {
         let extra_inputs = if include_configs { CONFIG_FILES } else { &[] };
         let total_inputs = inputs.len() + extra_inputs.len();
 
-        // Build glob lists directly from &str references — no need to clone
-        // every input into an owned String. We only allocate for the joined
-        // "package_path/glob" strings that globwalk requires.
         let mut inclusions = Vec::with_capacity(total_inputs);
         let mut exclusions = Vec::with_capacity(total_inputs);
         let mut glob_buf = String::with_capacity(package_unix_path.len() + 1 + 64);


crates/turborepo-scm/src/package_deps.rs
```diff
@@ -200,45 +200,33 @@ impl GitRepo {
...
         let package_unix_path = package_unix_path_buf.as_str();
 
-        let mut inputs = inputs
+        static CONFIG_FILES: &[&str] = &["package.json", "turbo.json", "turbo.jsonc"];
+        let extra_inputs = if include_configs { CONFIG_FILES } else { &[] };
+        let total_inputs = inputs.len() + extra_inputs.len();
+
+        // Build glob lists directly from &str references — no need to clone
+        // every input into an owned String. We only allocate for the joined
+        // "package_path/glob" strings that globwalk requires.
```
anthonyshew:
```suggestion
```

PR #11268, opened 2025-12-17
fix(bun): Add GitHub/git packages

Palid:
## Summary

Fixes multiple issues with `turbo prune` producing corrupted bun.lock files that fail `bun install --frozen-lockfile`.

While investigating #11266 (git repository package corruption), I discovered additional issues affecting the pruned lockfile format.

### 1. GitHub/git package format corruption (fixes #11266)

The deserializer was type-blind when processing package entries - it stored `registry: Some("")` for any package with an empty string at position 1, including GitHub/git packages that should never have a registry field.

Per bun.lock.zig, the format is:
- npm packages: 4 elements `[ident, registry, info, checksum]`
- github/git packages: 3 elements `[ident, info, .bun-tag]` — no registry

The serializer then output 4-element arrays for GitHub packages, which bun rejects.

### 2. Missing metadata sections

Pruned lockfiles were missing: `configVersion`, `trustedDependencies`, `overrides`, `catalog`, `catalogs`, and `patchedDependencies`.

### 3. JSON formatting issues

- Missing trailing commas in nested structures (bun's parser requires them)
- `optionalPeers` arrays formatted as `[ "a", "b", ]` instead of `["a", "b"]`

## Test Plan

- Red/green tests for all fixes (tests fail without fix, pass with fix)
- Comprehensive coverage for all 9 package types and 6 metadata sections
- Tested with real monorepo using `bun install --frozen-lockfile`
- Verified both `turbo prune` and `turbo prune --docker` work (same encode path)

## Related Issues

- Fixes #11266
- Related: #11007, #10782, #10783

PR #11917, opened 2026-02-19
fix(prune): Handle tarball URL deps in bun.lock

mimoo:
## Summary

Follow-up to #11268 -- that PR fixed `turbo prune` corrupting `bun.lock` for `@git+` and `@github:` packages, but missed tarball URL dependencies (e.g. `@mariozechner/pi-ai@https://github.com/.../pkg-1.0.0.tgz`). Same bug, different package type.

The root cause is the same: `is_git_or_github_package` only checked for git/github patterns, so tarball URLs still got an empty registry string injected, turning a 2-element array into a broken 4-element one. `bun install --production` then chokes on it:

```
error: Expected an object
InvalidPackageInfo: failed to parse lockfile: 'bun.lock'
```

The fix renames the function to `is_non_npm_package` and detects all non-npm package types (tarball, link, file, workspace, root) in addition to git/github. This way we handle the full bun lockfile schema rather than playing whack-a-mole with individual package types.

## Test plan

- [x] Added tarball URL test cases for both deserialization and serialization
- [x] Added corrupted-input test cases (tarball with extraneous empty string)
- [x] Existing git/github tests still pass
- [x] `cargo test -p turborepo-lockfiles --lib bun` -- all 64 tests pass

## How to reproduce the initial issue

see https://github.com/vercel/turborepo/issues/11266#issuecomment-3929094794

### DIFF
--- a/crates/turborepo-lockfiles/src/bun/de.rs
+++ b/crates/turborepo-lockfiles/src/bun/de.rs
@@ -2,7 +2,7 @@ use std::collections::VecDeque;
 
 use serde::Deserialize;
 
-use super::{PackageEntry, PackageInfo, is_git_or_github_package};
+use super::{PackageEntry, PackageInfo, is_non_npm_package};
 use crate::bun::RootInfo;
 // Comment explaining entry schemas taken from bun.lock.zig
 // first index is resolution for each type of package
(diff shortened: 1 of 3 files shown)


anthonyshew:
Hi, sorry for not seeing this sooner. #12076 addressed this. Apologies for taking so long to reply!

PR #11920, opened 2026-02-19
perf: Cache compiled env wildcard regexes for builtin pass-through patterns

anthonyshew:
## Summary

The ~30 builtin pass-through environment variable patterns (`HOME`, `PATH`, `VSCODE_*`, `DOCKER_*`, `GITHUB_*`, etc.) were recompiled into regexes on every `env()` call — once per task, 1690 times per run in a large monorepo. Now compiled once at `TaskHasher` construction and reused.

### Benchmarks (`--dry` runs, `--skip-infer`, 10 runs each, 5 warmup)

| Repo | Packages | Tasks | Before | After | Delta |
|------|----------|-------|--------|-------|-------|
| Large | ~1000 | 1690 | 2.149s ± 0.073s | 2.091s ± 0.174s | **1.03x faster** |
| Medium | ~120 | ~200 | 1.330s ± 0.142s | 1.249s ± 0.138s | **1.07x faster** |
| Small | ~5 | ~5 | 842.7ms ± 24.2ms | 813.0ms ± 62.1ms | **1.04x faster** |

### Changes

**`turborepo-env`**:
- New `CompiledWildcards` type that pre-compiles include/exclude wildcard patterns into reusable `Regex` objects.
- New `from_compiled_wildcards` and `pass_through_env_compiled` methods on `EnvironmentVariableMap` that use pre-compiled regexes instead of recompiling on each call.
- `BUILTIN_PASS_THROUGH_ENV` constant: the builtin pass-through pattern list now lives here instead of inline in task-hash. Single source of truth.

**`turborepo-task-hash`**:
- `TaskHasher` now holds a `compiled_builtins: CompiledWildcards` field, compiled once in `new()`.
- `env()` in Strict mode calls `pass_through_env_compiled` instead of `pass_through_env`, skipping regex compilation on every call.
- The inline 30-entry builtin list is removed in favor of `BUILTIN_PASS_THROUGH_ENV` from the env crate.

### Testing

8 new tests verifying the compiled path produces identical results to the original:
- `test_pass_through_env_compiled_matches_original` (4 test cases) — same scenarios as the existing `test_pass_through_env`
- `test_compiled_wildcards_matches_from_wildcards` — full `BUILTIN_PASS_THROUGH_ENV` list against a realistic env
- `test_builtin_pass_through_env_compiles` — the constant compiles without error
- `test_compiled_wildcards_with_excludes` — include + exclude equivalence
- `test_compiled_wildcards_empty_patterns` — empty patterns match nothing

### DIFF
--- a/crates/turborepo-env/src/lib.rs
+++ b/crates/turborepo-env/src/lib.rs
@@ -8,7 +8,7 @@ use std::{
     ops::{Deref, DerefMut},
 };
 
-use regex::RegexBuilder;
+use regex::{Regex, RegexBuilder};
 use serde::Serialize;
 use sha2::{Digest, Sha256};
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


anthonyshew:
The macOS test hanging on lockfile-aware-caching is a flake, and not related to this PR. Merging past it.

PR #11926, opened 2026-02-20
fix: Honor --output-logs errors-only for non-cached tasks

### LINKED ISSUE #6677: [turborepo] The --output-logs errors-only flag is ignored for non-cached tasks.

veeceey:
## Summary
Fixes #6677 — `--output-logs errors-only` was ignored for tasks with `cache: false`.

When caching is disabled, the log file was not created, so `on_error` had no log file to replay when a task failed. Since `errors-only` mode works by suppressing stdout during execution and replaying the log file on error, the missing log file meant error output was silently lost.

The fix adds an additional condition in `TaskCache::output_writer()`: the log file is also created when `task_output_logs == OutputLogsMode::ErrorsOnly`, ensuring the replay-on-error mechanism works for non-cached tasks.

## Test plan
- Added Rust integration tests in `crates/turborepo/tests/run_logging.rs` verifying:
  - Successful non-cached task with `--output-logs=errors-only` suppresses output
  - Failed non-cached task with `--output-logs=errors-only` shows output


veeceey:
updated the PR title to capitalize the first letter after the prefix — should fix the title validation check now

anthonyshew:
Hi, sorry for being way too slow to review here.

I moved the integration tests from the Prysk setup to pure Rust. I'll get that moved over for you to get that out of your way.

veeceey:
Thanks for merging and for handling the test migration! Appreciate it.

PR #11932, opened 2026-02-20
perf: Share HTTP client and improve profile instrumentation

anthonyshew:
## Summary

Two changes to improve turbo startup performance and profiling accuracy:

1. **Share a single `reqwest::Client`** across telemetry, API, and cache operations — eliminates redundant TLS initialization that was costing ~260ms on startup
2. **Improve `--profile` instrumentation coverage** — moves chrome tracing enablement earlier and instruments key functions so profiles capture the full wall-clock time

| Repo size | Before | After | Speedup |
|---|---|---|---|
| Small (6 packages) | 1.037s ± 0.079s | 868ms ± 27ms | **1.16x** |
| Medium (~120 packages) | 1.348s ± 0.065s | 1.299s ± 0.052s | **1.04x** |
| Large (~1000 packages) | 2.022s ± 0.152s | 1.933s ± 0.136s | **1.05x** |

The small repo shows the largest relative improvement because the redundant work was a fixed cost.

## Why the shared client matters

Previously, three separate `reqwest::Client` instances were built during startup, each independently initializing a TLS backend (loading system certs, creating crypto context):

- `initialize_telemetry_client` → `AnonAPIClient::new()` (~100ms)
- `RunBuilder::new` → `APIClient::new()` with 2 clients (~160ms)

That's ~260ms of sequential TLS initialization before any real work begins. Now a single client is built once in `cli::run()` and shared via `new_with_client()` constructors. Per-request timeouts replace per-client timeouts, preserving the existing timeout semantics.

## Profiling improvements

The `--profile` flag previously captured only ~50% of wall-clock time. This was caused by:

1. Chrome tracing was enabled too late — after clap parsing, config resolution, and repo inference
2. Many key functions in the startup and run paths lacked `#[tracing::instrument]`

Now the shim's argument parser extracts `--profile`/`--anon-profile` early and enables chrome tracing before any repo inference or CLI work begins. Profiles now capture the full run duration.

### DIFF
--- a/crates/turborepo-api-client/src/lib.rs
+++ b/crates/turborepo-api-client/src/lib.rs
@@ -108,10 +108,11 @@ pub trait TokenClient {
 #[derive(Clone)]
 pub struct APIClient {
     client: reqwest::Client,
-    cache_client: reqwest::Client,
     base_url: String,
     user_agent: String,
     use_preflight: bool,
+    timeout: Option<Duration>,
+    upload_timeout: Option<Duration>,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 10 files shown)


anthonyshew:
The hanging CI test is a flake and unrelated. Merging past.

PR #11998, opened 2026-02-25
fix: Resolve package-name-shaped tsconfig path aliases

### LINKED ISSUE #11906: `turbo boundaries` does not understand tsconfig path aliases, incorrectly flags local imports as undeclared dependencies
### Verify canary release

https://github.com/mentimeter/turborepo-boundaries-example

turbo@2.8.11-canary.4

### Environment information

```block
CLI:
   Version: 2.8.11-canary.4
   Path to executable: /Users/charlotte.andersson/workspace/my-turborepo/node_modules/turbo-darwin-arm64/bin/turbo
   Daemon status: Running
   Package manager: yarn

Platform:
   Architecture: aarch64
   Operating system: macos
   WSL: false
   Available memory (MB): 2943
   Available CPU cores: 10

Environment:
   CI: None
   Terminal program (TERM_PROGRAM): iTerm.app
   Terminal program version (TERM_PROGRAM_VERSION): 3.5.14
   Shell (SHELL): /bin/zsh
   stdin: false
```

### Expected behavior

turbo boundaries should either:
- Read and respect tsconfig.json path aliases when resolving imports, or
- Provide a config option in turbo.json to declare known path prefixes as local (e.g. ignoredImports: ["features/*"])

### Actual behavior

`turbo boundaries` reports violations for imports that use TypeScript path aliases resolving to local source folders, treating them as undeclared package dependencies.

Example:
`tsconfig.json` in apps/app-a:
```
{
  "compilerOptions": {
    "paths": {
      "*": [
        "./src/*"
      ],
    },
    
  }
}
```
A source file importing:
`import { feature_A } from "features/feature-a"; `

This resolves to ./src/features/feature-a -> it is entirely local to the workspace. TypeScript resolves it correctly. However, turbo boundaries flags it as:
```
Checking packages...
  × cannot import package `features` because it is not a dependency
   ╭─[/Users/charlotte.andersson/workspace/my-turborepo/apps/app-a/src/index.ts:1:
add workflow run
1]
 1 │ import { feature_A } from "features/feature-a";
   · ───────────────────────┬───────────────────────
   ·                        ╰── package imported here
 2 │
   ╰────

Checked 2 files in 5 packages, 1 issue found
```

### To Reproduce

1. Clone the [example repo.](https://github.com/mentimeter/turborepo-boundaries-example)
2. Install dependencies.
3. Open apps/app-a/src/index.ts.
4. Note the import from "features/..." which resolves to a local src folder via tsconfig path alias.
5. Run `yarn turbo boundaries`
6. Observe the error: “cannot import package features because it is not a dependency” even though the import is local and resolvable by TypeScript.

Also runs in CI: https://github.com/mentimeter/turborepo-boundaries-example/actions/workflows/boundaries.yml

### Additional context

Upgrading version by version, it seems it was introduced in turbo@v2.8.7, since turbo@v2.8.6 shows `no issues found`.

Current workaround: `npx @turbo/codemod migrate --to 2.8.6`

sleitor:
## Problem

When a `tsconfig.json` maps a bare specifier that looks like a package name to a local source path:

```json
{ "compilerOptions": { "paths": { "*": ["./src/*"] } } }
```

an import like:

```ts
import { feature_A } from "features/feature-a";
```

resolves locally (TypeScript handles it correctly), but `turbo boundaries` incorrectly reports:

```
× cannot import package `features` because it is not a dependency
```

This was introduced in v2.8.7. See #11906 for a reproduction.

## Root cause

`check_import_as_tsconfig_path_alias` had an early-return guard:

```rust
if import.starts_with('.') || BoundariesChecker::is_potential_package_name(import) {
    return Ok(false);
}
```

This prevented the resolver from attempting to resolve any import whose specifier looks like a package name (e.g. `features/feature-a`). Those imports then fell through to `check_package_import`, which flagged them as undeclared dependencies.

## Fix

Remove the `is_potential_package_name` guard so **all non-relative imports** are first tried against the tsconfig resolver:

- If the resolver resolves the import to a local file → the function returns `true` and the check stops (no false positive).
- If the resolver resolves to a path through `node_modules` → returns `false` (real npm package, falls through to dependency check).
- If the resolver fails → returns `false` (import falls through to `check_package_import` as before).

Additionally:
- **Structured error handling**: replaced blanket `Err(_) => Ok(false)` with explicit matching on expected errors (`NotFound`, `Builtin`, etc.) and `tracing::debug!` logging for unexpected errors (I/O, broken tsconfig) to aid debugging.
- **Doc comments**: added documentation to `check_import`, `check_file_import`, `get_package_name`, and `is_potential_package_name`.
- **Fixed pre-existing macOS test failures**: canonicalized temp dir paths in tests to match the resolver's symlink-resolved output (`/tmp` → `/private/tmp`).

## Tests

- Renamed `tsconfig_alias_check_skips_bare_package_names` → `tsconfig_alias_check_returns_false_for_unresolvable_package_imports` to reflect the corrected semantics (the behaviour is unchanged: unresolvable package imports still return `false`).
- Added `tsconfig_alias_resolves_package_name_shaped_path_alias`: regression test that mirrors the exact scenario in #11906 (`"*": ["./src/*"]` alias, `features/feature-a` import).
- Added `tsconfig_alias_check_returns_false_for_node_modules_resolution`: verifies the `node_modules` bail-out path with a real `node_modules/` fixture on disk.
- Added `tsconfig_alias_flags_boundary_violation_for_out_of_package_resolution`: verifies a tsconfig alias pointing outside the package root produces an `ImportLeavesPackage` diagnostic.
- Added `warnings.is_empty()` assertions to regression tests.

Fixes #11906

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7265,6 +7265,7 @@ dependencies = [
  "biome_deserialize_macros 0.6.0",
  "biome_diagnostics",
  "camino",
+ "dunce",
  "globwalk",
  "indicatif",
  "itertools 0.14.0",
--- a/crates/turborepo-boundaries/Cargo.toml
+++ b/crates/turborepo-boundaries/Cargo.toml
@@ -38,5 +38,6 @@ turborepo-ui = { workspace = true }
 unrs_resolver = { workspace = true }
 
 [dev-dependencies]
+dunce = { workspace = true }
 tempfile = { workspace = true }
 test-case = { workspace = true }
--- a/crates/turborepo-boundaries/src/imports.rs
+++ b/crates/turborepo-boundaries/src/imports.rs
@@ -85,11 +85,10 @@ impl<'a> DependencyLocations<'a> {
 /// `./src/features/foo`, and if so, checks the resolved path against package
 /// boundaries.
 ///
-/// Only called for package-name-shaped imports (after
-/// `is_potential_package_name` guard in `check_import`). This allows tsconfig
-/// `paths` entries that shadow package-name-shaped specifiers to be recognised
-/// as local imports instead of being incorrectly flagged as undeclared
-/// dependencies.
+/// Called for all non-relative imports in `check_import`. This allows tsconfig
+/// `paths` entries — whether they shadow package-name-shaped specifiers or use
+/// non-package-name patterns like `!` or `@/foo` — to be recognised as local
+/// imports instead of being incorrectly flagged as undeclared dependencies.
 ///
 /// Returns `Ok(true)` if the import was resolved as a tsconfig path alias
 /// (local or cross-package — the latter produces an `ImportLeavesPackage`
@@ -122,14 +121,22 @@ fn check_import_as_tsconfig_path_alias(
             // resolved to a real npm package rather than a tsconfig path alias
             // pointing to a local file.  Return false so the caller falls
             // through to `check_package_import`.
-            //
-            // We intentionally do NOT use `resolution.package_json().is_some()`
-            // because the resolver may return a package.json for any file
-            // inside a directory that has one — including local tsconfig alias
-            // targets.
             let path = resolution.path();
             if path.components().any(|c| c.as_os_str() == "node_modules") {
                 return Ok(false);
+            }
+            // Workspace packages are symlinked in node_modules, so the
+            // resolved path won't contain `node_modules` after symlink
+            // resolution. Detect these by checking if the resolution's
+            // package.json name matches the import's package name — if so,
+            // the resolver found the actual package, not a tsconfig alias.
+            if BoundariesChecker::is_potential_package_name(import) {
+                let import_pkg_name = get_package_name(import);
+                if let Some(pkg_json) = resolution.package_json()
+                    && pkg_json.name() == Some(import_pkg_name.as_str())
+                {
+                    return Ok(false);
+                }
             }
             let Some(utf8_path) = Utf8Path::from_path(path) else {
                 result.diagnostics.push(BoundariesDiagnostic::InvalidPath {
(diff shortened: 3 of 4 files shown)


sleitor:
Fixed in latest commit (83be5c2).

The issue was that `check_import_as_tsconfig_path_alias` returned `true` for any import the resolver could resolve — including bare package names like `lodash` that resolve to `node_modules/`. This bypassed `check_package_import` entirely.

The fix adds a check after resolution: if the resolved path contains a `node_modules` component, return `false` so the import falls through to proper dependency validation via `check_package_import`. Only genuinely local paths (resolved via tsconfig `paths` aliases) now return `true`.

sleitor:
Thanks for the review @anthonyshew! Addressed all three points in the latest commit:

**1. Failing regression test** — The only failing CI check now is "Validate PR title" (a format lint). All code tests pass locally (`cargo test -p turborepo-boundaries`: 24/24 ✅). If there's a specific regression test you had in mind from a previous CI run, please let me know and I'll fix it.

**2. `node_modules` check** — Switched to `resolution.package_json().is_some()` as suggested. If the resolver found a `package.json` the resolution went through node_modules (a real npm package), not a tsconfig path alias pointing to a local file. More reliable than the string path component check.

**3. `is_potential_package_name` performance** — Moved `check_import_as_tsconfig_path_alias` inside the `is_potential_package_name` guard. Non-package-name-shaped imports (`node:fs`, `bun:test`, etc.) now skip the resolver entirely, exactly as before. The resolver is only called for package-name-shaped imports — first trying tsconfig alias resolution, then falling through to `check_package_import` if no alias matches.

sleitor:
Good catch from VADE — fixed in the latest commit (500e7d9).

The `resolution.package_json().is_some()` check was indeed incorrect: oxc_resolver returns a `package_json` for *any* file within a package directory, not just node_modules packages. So in a real project (where the source package has its own `package.json`), a tsconfig path alias resolving to a local file would still cause `package_json()` to return `Some`, incorrectly returning `false` and falling through to `check_package_import`.

The fix switches to checking whether the resolved path contains a `node_modules` component — the canonical way to distinguish npm packages from local source files. Also added a regression test (`tsconfig_alias_resolves_with_package_json_present`) that replicates the real-project scenario and confirms the alias is now correctly resolved even when `package.json` is present.

anthonyshew (approved):
Hey, thanks for getting us started. I made a few more changes and now we're good to go. Thanks!

PR #12007, opened 2026-02-26
perf: Extract query module into turborepo-query crate

anthonyshew:
## Summary

- Extracts the GraphQL query layer (1,688 lines, 8 files) from `turborepo-lib` into a new `turborepo-query` crate
- Introduces a `QueryRun` trait that decouples the query layer from the concrete `Run` type
- Removes `swc_common`, `swc_ecma_ast`, `swc_ecma_parser`, `async-graphql-axum`, `axum`, `tower-http`, and `turbo-trace` from `turborepo-lib`'s direct dependencies

## Why

`turborepo-lib` is the bottleneck in the build graph — 42+ internal crates feed into it, and nothing downstream can compile until it finishes. The query module pulled in ~40s of CPU time in heavy dependencies (SWC, async-graphql, axum) that aren't needed by the rest of `turborepo-lib`.

After this change, those dependencies compile in the new `turborepo-query` crate **in parallel** with `turborepo-lib` instead of blocking it.

## Build time impact (clean build, macOS)

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Wall clock | 74.2s | 57.4s | **-22.6%** |
| Total CPU | 761.2s | 521.8s | **-31.5%** |
| `turborepo-lib` compile | 7.20s | 5.06s | **-30%** |
| `turborepo-lib` start time | 62.3s | 47.3s | **15s earlier** |

## Approach

The `QueryRun` trait exposes the read-only accessors that the GraphQL resolvers need (`repo_root()`, `pkg_dep_graph()`, `engine()`, `scm()`, `root_turbo_json()`), plus two methods that encapsulate `turborepo-lib`-internal logic:

- `calculate_affected_packages(base, head)` — wraps the `Opts` mutation + `RunBuilder::calculate_filtered_packages` call, keeping `Opts` out of the query crate
- `check_boundaries(show_progress)` — delegates to the existing `impl Run` method

The query crate uses `Arc<dyn QueryRun>` (object-safe dynamic dispatch). The vtable cost is negligible for network-bound GraphQL resolvers.

No behavioral changes. All 192 `turborepo-lib` unit tests pass.

## For reviewers

`async-graphql` remains in `turborepo-lib` because `commands/query.rs` directly constructs a `Schema` to execute inline queries. A follow-up could push that behind a higher-level API in `turborepo-query` to drop async-graphql from `turborepo-lib` entirely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5212,7 +5212,7 @@ dependencies = [
  "tokio-rustls",
  "tokio-util",
  "tower 0.5.2",
- "tower-http",
+ "tower-http 0.6.8",
  "tower-service",
  "url",
  "wasm-bindgen",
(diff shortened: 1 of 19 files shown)


anthonyshew:
The test is stuck on a known flake. We are safe to merge past.

PR #12009, opened 2026-02-26
fix: Resolve all lockfile pruning test failures

anthonyshew:
## Summary

- Fix Berry lockfile pruning to preserve `cacheKey` when the output contains packages with checksums
- Fix NPM lockfile pruning to preserve the original `lockfileVersion` instead of always forcing v3
- Regenerate all broken test fixtures with valid lockfiles while preserving their edge cases

## Code changes

**`crates/turborepo-lockfiles/src/berry/mod.rs`**: The `lockfile()` method was checking only `self.resolutions` for checksums to decide whether to strip `cacheKey`. When a pruned subgraph's resolutions only contained workspace references (no checksums), the key was incorrectly dropped — even though the output packages map did contain npm packages with checksums. Now checks the actual output packages instead.

**`crates/turborepo-lockfiles/src/npm.rs`**: `subgraph()` was hardcoding `lockfile_version: 3` in the pruned output. This broke `npm ci` for lockfileVersion 2 projects (npm@8). Now preserves the original version.

## Fixture changes

Many fixtures had fundamentally broken lockfiles (nonexistent npm packages, fake integrity hashes, missing patch files, missing transitive dependencies). These were regenerated with real package managers while preserving the specific edge cases each fixture tests:

| Fixture | Edge case preserved |
|---|---|
| `berry` | `patch:` protocol for lodash via `resolutions` field |
| `berry_semver` | Major-only semver range (`"2"`) |
| `minimal-berry-resolutions` | Resolution override forcing `debug@3.2.7` when workspace declares `^4.3.4` |
| `yarn4-resolution` | Resolution override pinning `react@18.2.0` against `^18.2.0` range |
| `yarn4-catalog` | `catalog:` protocol with definitions in `.yarnrc.yml` |
| `yarn4-patch` / `yarn4-direct-and-indirect` | Real `.yarn/patches/` files for `is-odd` |
| `npm-lock-workspace-variation` | Object-form `workspaces: { packages: [...] }` and differing lockfile/package name |
| `issue-10985` | Peer dep `next: ^14 \|\| ^15` with different resolved versions per consumer |

## Testing

171/171 lockfile tests pass. 146/146 Rust unit tests pass.

### DIFF
--- a/lockfile-tests/fixtures/berry/package.json
+++ b/lockfile-tests/fixtures/berry/package.json
@@ -11,8 +11,8 @@
     "prettier": "latest",
     "turbo": "latest"
   },
-  "packageManager": "yarn@3.6.0",
   "resolutions": {
     "lodash@^4.17.21": "patch:lodash@npm%3A4.17.23#./.yarn/patches/lodash-npm-4.17.23-50bdb1c01a.patch"
-  }
+  },
+  "packageManager": "yarn@3.6.0"
 }
--- a/lockfile-tests/fixtures/yarn4-catalog/package.json
+++ b/lockfile-tests/fixtures/yarn4-catalog/package.json
@@ -9,8 +9,8 @@
   "dependencies": {
     "turbo": "npm:^2.6.1"
   },
+  "packageManager": "yarn@4.12.0",
   "catalog": {
     "lodash": "^4.17.21"
-  },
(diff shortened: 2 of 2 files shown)


anthonyshew:
The failing test is a known flake. We will be fixing in a followup PR soon.

PR #12019, opened 2026-02-27
ci: Fix Rust build cache eviction on Windows CI

anthonyshew:
## Summary

- Replace `setup-node`'s built-in `cache: pnpm` with explicit `actions/cache` for the pnpm store
- Enable Rust cache saving on PR branches (not just main)

## Why

Windows CI builds take ~14 minutes per partition. Investigation revealed that **the Rust build cache was being evicted entirely** — zero Rust cache entries existed in the GitHub Actions cache. The 10GB cache budget was consumed by 14 node/pnpm cache entries at 425MB each (5.95GB), leaving no room for Rust artifacts.

Meanwhile, sccache achieves 87% hit rate on regular crate compilations, but **227 proc-macro and build-script crates are not cacheable by sccache** (they run on the host, not through `rustc`). These 227 crates recompile from scratch on every CI run, taking 2-3 minutes on Windows MSVC.

## What changed

**Node caching:** Replaced `actions/setup-node`'s built-in `cache: pnpm` with explicit `actions/cache@v4`. Same key format, same restore behavior, but we control the action rather than relying on setup-node's internal cache management. This prevents the opaque cache entry proliferation that was filling the cache budget.

**Rust caching:** The `save-if` expression previously hardcoded `github.ref == 'refs/heads/main'`, ignoring the `save-cache: true` input from `setup-environment`. Now it respects the input directly. Since `setup-environment` passes `save-cache: true` and the surrounding `if` already filters out fork PRs, Rust caches now save on all first-party PR branches.

With Rust caches present, Cargo can fingerprint-match unchanged proc-macro DLLs and build script outputs, skipping their compilation entirely — no sccache roundtrip needed, no MSVC invocation at all for unchanged crates.

## Expected impact

On the first run after merging, Rust caches populate. On subsequent runs (including PR branches reading from main's cache), the 227 proc-macro crates that sccache can't handle should be skipped via fingerprint matching. Estimated savings: **2-4 minutes on Windows**, bringing ~14 minute builds down to ~10-12 minutes. The impact compounds as the cache warms.

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -29,7 +29,6 @@ runs:
         # node-version-file is the default, but can be overridden using node-version
         node-version-file: "package.json"
         node-version: ${{ inputs.node-version }}
-        cache: pnpm
 
     - name: Upgrade corepack
       if: ${{ inputs.enable-corepack == 'true' }}
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -46,5 +46,4 @@ runs:
       with:
         shared-key: ${{ inputs.shared-cache-key }}
         key: ${{ inputs.cache-key }}
-        # the cache is huge and we only get 10gb max, so we only save on master
-        save-if: ${{ github.ref == 'refs/heads/main' && inputs.save-cache || 'false' }}
+        save-if: ${{ inputs.save-cache }}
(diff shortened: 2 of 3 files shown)


anthonyshew:
Closing — investigation showed this wouldn't materially improve build times. The Rust cache only stores 4.49MB of metadata (sccache intercepts compilation and stores artifacts in S3, leaving the local target/ empty). The 233 proc-macro crates that are the real bottleneck aren't helped by either caching layer. Pursuing larger CI runners instead (#12029).

PR #12028, opened 2026-02-27
ci: Increase Rust test partitions from 4 to 10

anthonyshew:
## Summary

- Increases Rust test nextest partitions from 4 to 10 across all CI jobs (macOS, Windows, Ubuntu) to reduce per-partition test time
- Updates the coverage merge step to combine all 10 partition lcov files


anthonyshew:
The failures are flakes that we are fixing in follow-up PRs.

PR #12031, opened 2026-02-27
ci: Use 8-core Windows runner for Rust tests

anthonyshew:
## Summary

Upgrade Windows test runner from `windows-latest` (4 vCPU) to `windows-latest-8-cores` (8 vCPU).

## Why

Windows Rust test partitions are the slowest CI jobs. The compilation phase is bottlenecked by core count — 233 proc-macro crates compile from scratch every run (sccache can't cache them), and parallel Rust compilation scales with available cores.

Doubling from 4 to 8 cores should cut the parallel compilation phase roughly in half, bringing total job time from ~14 minutes toward ~10 minutes per partition.


anthonyshew:
The failing test is a known flake. We can merge past it and fix in a follow-up.

PR #12038, opened 2026-02-27
fix(frameworks): also check devDependencies when inferring framework in single-package mode

### LINKED ISSUE #12023: Framework inference ignores devDependencies in single-package projects
### Verify canary release

 https://github.com/ysgk/turbo-framework-inference-repro

2.8.12-canary.2

### Environment information

```block
• turbo 2.8.12-canary.2
CLI:
   Version: 2.8.12-canary.2
   Path to executable: /private/tmp/turbo-repro/node_modules/turbo-darwin-arm64/bin/turbo
   Daemon status: Not running
   Package manager: npm

Platform:
   Architecture: aarch64
   Operating system: macos
   WSL: false
   Available memory (MB): 6733
   Available CPU cores: 12

Environment:
   CI: None
   Terminal program version (TERM_PROGRAM_VERSION): 466
   Shell (SHELL): /bin/zsh
   stdin: false
```

### Expected behavior

In a single-package (non-monorepo) project, frameworks listed in `devDependencies` (e.g. `vite`) should be detected by framework inference, and their environment variable wildcards (e.g. `VITE_*`) should be included in `inferred` env vars.

### Actual behavior

Framework inference does not detect frameworks in `devDependencies` for single-package projects. `turbo build --dry=json` shows `"framework": ""` and `"inferred": []`, so `VITE_*` environment variables are not passed to subprocesses in strict mode.

This works correctly in monorepo setups because `unresolved_external_dependencies` is used, which includes both `dependencies` and `devDependencies`.

### To Reproduce

1. Clone https://github.com/ysgk/turbo-framework-inference-repro
2. `npm install`
3. Run: `VITE_API_URL=http://localhost:3000 npx turbo build --dry=json | jq '{framework: .tasks[].framework, inferred: .tasks[].environmentVariables.inferred}'`
4. Observe: `{"framework": "", "inferred": []}`
5. Move `vite` from `devDependencies` to `dependencies` in package.json and repeat — framework is now detected correctly.

### Additional context

The root cause is in `Matcher::test` (`crates/turborepo-frameworks/src/lib.rs`). For single-package projects (`is_monorepo: false`), it only checks `package_json.dependencies`, ignoring `devDependencies`. This affects all frameworks typically installed as devDependencies: Vite, Astro, SvelteKit, Vue CLI, etc.

This was actually flagged during the original PR review (#5746):

> "I realize this is just a port of Go behavior, but do we want to also check the other places dependencies could appear? Primarily `devDependencies` and `optionalDependencies`"

However, no conclusion was reached and the PR was merged as-is. So this appears to be an oversight carried over from the Go implementation rather than an intentional design decision.

sleitor:
## Problem

In non-monorepo projects (single-package mode), framework inference silently fails for any framework installed as a `devDependency` (e.g. Vite, Astro, SvelteKit, Vue CLI).

```bash
# package.json has vite in devDependencies
turbo build --dry=json | jq '{framework, inferred}'
# => {"framework": "", "inferred": []}

# after moving vite to dependencies — works fine
```

Root cause: `Matcher::test` in `crates/turborepo-frameworks/src/lib.rs` only checks `package_json.dependencies` in single-package mode. `dev_dependencies` is not consulted.

The monorepo path was unaffected because `unresolved_external_dependencies` is populated from all dep fields during graph construction.

This was flagged in the original PR review (#5746) but never resolved:
> "I realize this is just a port of Go behavior, but do we want to also check the other places dependencies could appear? Primarily devDependencies and optionalDependencies"

## Fix

Replace the single `deps` map reference with a `contains` closure that checks both `dependencies` and `dev_dependencies` in single-package mode. The logic applies to both `Strategy::All` and `Strategy::Some` branches.

## Testing

Added a test case: `Finds vite in devDependencies in non-monorepo`. All 15 existing tests continue to pass.

Fixes #12023

### DIFF
--- a/crates/turborepo-frameworks/src/lib.rs
+++ b/crates/turborepo-frameworks/src/lib.rs
@@ -86,23 +86,32 @@ fn get_frameworks() -> &'static Vec<Framework> {
 impl Matcher {
     pub fn test(&self, workspace: &PackageInfo, is_monorepo: bool) -> bool {
         // In the case where we're not in a monorepo, i.e. single package mode
-        // `unresolved_external_dependencies` is not populated. In which
-        // case we should check `dependencies` instead.
-        let deps = if is_monorepo {
-            workspace.unresolved_external_dependencies.as_ref()
-        } else {
-            workspace.package_json.dependencies.as_ref()
+        // `unresolved_external_dependencies` is not populated. In which case we
+        // should check `dependencies` and `dev_dependencies` instead, as frameworks
+        // (e.g. Vite, Astro, SvelteKit) are typically installed as devDependencies.
+        let contains = |dep: &str| -> bool {
+            if is_monorepo {
+                workspace
+                    .unresolved_external_dependencies
+                    .as_ref()
+                    .is_some_and(|deps| deps.contains_key(dep))
+            } else {
+                workspace
+                    .package_json
+                    .dependencies
+                    .as_ref()
+                    .is_some_and(|deps| deps.contains_key(dep))
+                    || workspace
+                        .package_json
+                        .dev_dependencies
+                        .as_ref()
+                        .is_some_and(|deps| deps.contains_key(dep))
+            }
         };
 
         match self.strategy {
-            Strategy::All => self
-                .dependencies
-                .iter()
-                .all(|dep| deps.is_some_and(|deps| deps.contains_key(dep))),
-            Strategy::Some => self
-                .dependencies
-                .iter()
-                .any(|dep| deps.is_some_and(|deps| deps.contains_key(dep))),
+            Strategy::All => self.dependencies.iter().all(|dep| contains(dep)),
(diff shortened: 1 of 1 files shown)


anthonyshew:
Can you explain why there are two of these? https://github.com/vercel/turborepo/pull/12056

sleitor:
Hi @anthonyshew — sorry for the confusion! I filed #12038 first as a general devDependencies check, then realized the issue specifically targets **single-package mode** (non-monorepo projects where the framework list used different matcher logic). I filed #12056 with a more targeted fix for that specific code path and linked it to the issue report.

They both address the same underlying bug (framework inference ignoring devDependencies), but #12056 is the cleaner, more targeted fix specifically for single-package mode. I'll close this one (#12038) in favor of #12056 to avoid confusion.

PR #12056, opened 2026-02-28
fix(frameworks): Also check devDependencies when inferring framework in single-package mode

### LINKED ISSUE #12023: Framework inference ignores devDependencies in single-package projects

sleitor:
## Summary

In single-package mode (`is_monorepo: false`), framework inference only checked `package_json.dependencies`, missing frameworks installed as `devDependencies` (e.g. Vite, Astro, SvelteKit). This caused incorrect or missing framework detection for many common setups.

## Root Cause

In `crates/turborepo-frameworks/src/lib.rs`, the `Matcher::test` function uses `unresolved_external_dependencies` for monorepos (which includes all deps), but falls back to only `package_json.dependencies` for single-package projects.

## Fix

When not in a monorepo, check both `dependencies` and `dev_dependencies` when looking up whether a framework dep is present.

## Test

Added a test case: `finds_vite_in_devdependencies_in_non_monorepo` — verifies that Vite is detected when listed under `devDependencies` in single-package mode.

All 15 tests pass.

Fixes #12023

### DIFF
--- a/.github/workflows/turborepo-test.yml
+++ b/.github/workflows/turborepo-test.yml
@@ -315,7 +315,7 @@ jobs:
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
           node-version: "18.20.2"
-          capnproto: "false"
+          capnproto: ${{ github.event.pull_request.head.repo.fork && 'true' || 'false' }}
           package-install: "false"
 
       - name: Install Bun


anthonyshew:
Can you explain why there are two of these? https://github.com/vercel/turborepo/pull/12038

sleitor:
Hi @anthonyshew — there were two because I initially filed #12038 as a general fix, then realized the issue specifically affects **single-package mode** (non-monorepo projects) where the framework detection uses a separate code path. I filed this PR (#12056) with a more targeted fix for that specific path, linked to the issue (#12023).

I've now closed #12038 as a duplicate. This PR (#12056) is the canonical fix.

anthonyshew:
The failures on this PR are flakes associated with changes that I made that aren't compatible for forks. In the interest of getting this fix out, we will merge, and I will fix these flakes in follow-up PRs.

PR #12083, opened 2026-03-01
ci: Deduplicate Rust test compilation with nextest archive

anthonyshew:
## Summary

- Adds a `build_rust_test` job that compiles all test binaries once per OS (macOS, Windows) using `cargo nextest archive`, then caches the resulting archive via turbo remote cache
- Partition runners restore the archive from turbo cache and run `cargo nextest run --archive-file` with zero compilation
- Eliminates 18 redundant full workspace compilations per CI run (9 saved per OS × 2 OSes)

## How it works

1. New `build:test-archive` turbo task in `@turbo/cli` wraps `cargo nextest archive --workspace`
2. Task depends on `rust-src` (fingerprints all `.rs`, `Cargo.toml`, `Cargo.lock`) and is segmented by `RUNNER_OS` via `globalEnv`
3. Build job runs `turbo run build:test-archive --filter=@turbo/cli` → compiles + archives + pushes to remote cache
4. Each partition runs the same turbo command → cache hit → restores archive → runs tests from it

## What to watch for

- Verify the archive file path resolves correctly (`cli/nextest-archive.tar.zst`) in the partition runners
- The nextest archive is self-contained (test binaries + dynamic libs + build script outputs), so no `target/` directory is needed
- Ubuntu/coverage jobs are unchanged — `cargo llvm-cov nextest` requires compile-time instrumentation that doesn't work with the archive approach. That's a separate optimization.
- `turborepo-napi` tests are now excluded on all platforms (previously only excluded on Windows). This was an accepted tradeoff for using a single archive task.

### DIFF
--- a/.github/workflows/turborepo-test.yml
+++ b/.github/workflows/turborepo-test.yml
@@ -384,23 +384,35 @@ jobs:
         with:
           tool: cargo-llvm-cov
 
-      - name: Run tests with coverage
-        timeout-minutes: 45
+      - name: Install Global Turbo
+        uses: ./.github/actions/install-global-turbo
+
+      - name: Build coverage-instrumented test archive
         run: |
           if [ -z "${RUSTC_WRAPPER}" ]; then
             unset RUSTC_WRAPPER
           fi
-          cargo llvm-cov nextest \
-            --workspace \
-            --exclude turborepo-napi \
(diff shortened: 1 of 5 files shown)


anthonyshew:
The tests are hanging and are going to be addressed in a follow-up PR.

PR #12094, opened 2026-03-02
chore: Trim unused dependency features for faster compilation

anthonyshew:
## Summary

Trims unused default features from four workspace dependencies to reduce compilation overhead:

- **`async-graphql`**: Disables `playground`, `email-validator`, `tempfile` defaults — only `graphiql` is used (via `GraphiQLSource`). Removes `fast_chemail` and `ascii_utils` from the dep tree
- **`nix`**: Centralizes as a workspace dep with `default-features = false`. `turborepo-process` declared all 30+ features but only uses `sys::termios` (`term`). `turborepo-ui` only uses `sys::signal`. Removes `memoffset 0.7.1` from the dep tree. **0.95s saved**
- **`petgraph`**: Disables `graphmap`, `matrix_graph`, `stable_graph` defaults — only `Graph`/`DiGraph`/`NodeIndex` and algo/visit traits are used
- **`chrono`**: Disables `oldtime`, `js-sys`, `wasm-bindgen`, `android-tzdata`, `windows-targets` defaults — only `DateTime`, `Local`, `Utc`, `Duration` are used. **0.41s saved**

Zero code changes — only Cargo.toml feature declarations. 5 compilation units removed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,7 +100,9 @@ async-compression = { version = "0.3.13", default-features = false, features = [
   "gzip",
   "tokio",
 ] }
-async-graphql = { version = "7.0.17", default-features = false, features = ["graphiql"] }
+async-graphql = { version = "7.0.17", default-features = false, features = [
+  "graphiql",
+] }
 async-graphql-axum = "7.0.16"
 async-trait = "0.1.64"
 axum = "0.8.4"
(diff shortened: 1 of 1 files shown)


anthonyshew:
The test hanging on macOS is unrelated and is being addressed in a follow-up PR.

PR #12098, opened 2026-03-02
perf: Replace SWC with oxc for JS/TS parsing

anthonyshew:
## Summary

Replaces the SWC parser (`swc_ecma_parser`, `swc_ecma_ast`, `swc_ecma_visit`, `swc_common`) with oxc (`oxc_parser`, `oxc_ast`, `oxc_estree`, `oxc_span`, `oxc_syntax`) for all JS/TS parsing in the codebase.

SWC's deeply chained, macro-heavy crates were the single biggest bottleneck on the compilation critical path: `swc_ecma_ast` (8.4s, 93% frontend), `swc_ecma_parser` (5.1s), `swc_ecma_visit` (3.9s), `swc_common` (2.6s) = ~20s serial. oxc compiles with better parallelism and fewer total units. Clean build CPU time drops by ~21s.

We can generally get the same functionality for at a cheaper compilation time with oxc, so mind as well.

### What changed

**`turbo-trace`**: Rewrote import extraction to use oxc's `ModuleRecord` (pre-extracted import/export metadata from the parser) instead of a custom SWC visitor. `require()` calls are detected via direct AST pattern matching. AST serialization uses oxc's ESTree JSON format.

**`turborepo-boundaries`**: Replaced SWC parsing + `ImportFinder` visitor with calls to `turbo_trace::parse_file`. Comment detection for `@boundaries-ignore` directives now uses oxc's `Comment` spans with a backwards walk that respects blank line boundaries.

**`turborepo-query`**: `File.ast` is now `Option<serde_json::Value>` (eagerly serialized at parse time while the oxc allocator is alive) instead of `Option<swc_ecma_ast::Module>`. The `ast()` GraphQL endpoint now returns ESTree JSON.

### User-visible changes

- **`turbo query` AST output**: JSON format changed from SWC's custom format to ESTree standard. Top-level `"type"` is `"Program"` instead of `"Module"`. Field names follow ESTree conventions. This endpoint is undocumented and has no stability guarantees.
- **`@boundaries-ignore` with blank line**: A `@boundaries-ignore` comment separated from its import by a blank line no longer suppresses the violation. The old SWC behavior incorrectly attached the comment across the gap.

### Testing

- Added `require()` integration test (`require_example.js` fixture) — previously untested
- All 21 trace dependency snapshots unchanged (identical import resolution behavior)
- Updated AST snapshot to ESTree format
- Updated boundaries snapshot (+1 correct violation for blank-line-separated ignore)
- All 41 tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,18 +28,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
-name = "ahash"
-version = "0.8.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d713b3834d76b85304d4d525563c1276e2e30dc97cc67bfb4585a4a29fc2c89f"
-dependencies = [
- "cfg-if",
- "once_cell",
- "version_check",
- "zerocopy",
-]
-
-[[package]]
 name = "aho-corasick"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


anthonyshew:
Merging past flaky test that is being resolved in a different PR.

PR #12157, opened 2026-03-05
test: Add integration test coverage for `--anon-profile` markdown generation

anthonyshew:
## Summary

- Adds integration tests verifying `--anon-profile` generates both a trace file and companion `.md` file (with explicit filename and default timestamp filename)
- Adds a test for `--profile` with default timestamp filename (previously only tested with explicit filename)
- Extracts shared test helpers to reduce duplication across profile tests


anthonyshew:
Something is wrong with the Chocolatey CDN on this specific PR that I don't understand, but these are just tests, so we don't need to prove that Windows is still working. Merging past this.

PR #12164, opened 2026-03-05
perf: Defer lockfile await until after internal dependency resolution

anthonyshew:
## Summary

- Moves the lockfile future await from *before* `resolve_lockfile` to *inside* it, after `connect_internal_dependencies` completes
- Gives the lockfile YAML parse more time to run concurrently with useful work

## Why

The lockfile `spawn_blocking` read+parse was designed to overlap with `parse_package_jsons`. But when package.json parsing is fast (~20ms) and lockfile parsing is slow (~195ms for large pnpm lockfiles), the main thread blocks for ~175ms waiting for the lockfile.

`connect_internal_dependencies` doesn't need the lockfile — it only resolves workspace-internal dependency edges. By deferring the lockfile await until after it completes, the lockfile parse now overlaps with both phases:

```
Before:
  [parse_package_jsons 20ms] → [await lockfile ████ 175ms blocked] → [connect_internal_deps 46ms] → ...

After:
  [parse_package_jsons 20ms] → [connect_internal_deps 46ms] → [await lockfile ██ 129ms blocked] → ...
  [─────────── lockfile parse running in background (195ms) ───────────]
```

~46ms reduction in blocking wait on repos where lockfile parsing is the bottleneck.

> Timings are specific to the repo I was working with, but concepts apply to any Turborepo.

## Testing

All 130 `turborepo-repository` tests pass. No behavioral change — only the timing of when the lockfile future is awaited.

### DIFF
--- a/crates/turborepo-repository/src/package_graph/builder.rs
+++ b/crates/turborepo-repository/src/package_graph/builder.rs
@@ -516,9 +516,6 @@ impl<'a, T: PackageDiscovery> BuildState<'a, ResolvedWorkspaces, T> {
             .package_manager;
         self.connect_internal_dependencies(&package_manager)?;
 
-        // Collect the lockfile result now — after connect_internal_dependencies
-        // so the lockfile parse overlaps with both parse_package_jsons AND
-        // internal dependency resolution.
         if let Some(handle) = lockfile_future
             && let Ok(Some(lockfile)) = handle.await
         {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7161,6 +7161,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "turborepo-ai-agents"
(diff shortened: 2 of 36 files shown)


crates/turborepo-repository/src/package_graph/builder.rs
```diff
@@ -531,6 +516,15 @@ impl<'a, T: PackageDiscovery> BuildState<'a, ResolvedWorkspaces, T> {
             .package_manager;
         self.connect_internal_dependencies(&package_manager)?;
 
+        // Collect the lockfile result now — after connect_internal_dependencies
+        // so the lockfile parse overlaps with both parse_package_jsons AND
+        // internal dependency resolution.
```
anthonyshew:
```suggestion
```

PR #12170, opened 2026-03-05
feat: Replace local login server with OAuth 2.0 Device Authorization Grant

anthonyshew:
## Summary

For Vercel logins only, replaces the axum-based localhost HTTP callback server with the OAuth 2.0 Device Authorization Grant (RFC 8628). Instead of spawning a local server and redirecting the browser back to `127.0.0.1:9789`, the user visits a URL and enters a code directly in the browser.

Non-Vercel logins (self-hosted remote caches) are unchanged in behavior — they still use the localhost redirect flow with the same URL patterns and the same configurable callback port. The only difference is the implementation: a simple one-shot `TcpListener` instead of axum.

**Why:** The localhost server approach is fragile for Vercel logins (port conflicts, firewall rules, WSL networking, SSH sessions, containers). The device flow works everywhere — including headless environments — with no open ports required.

## Non-Vercel users

No behavioral change. The `TURBO_SSO_LOGIN_CALLBACK_PORT` env var and `sso_login_callback_port` config option are preserved and continue to work for non-Vercel login/SSO flows.

## Breaking Changes

- `LoginServer` trait removed — tests using `MockLoginServer` need to be updated
- `login()` and `sso_login()` now return `(Token, Option<TokenSet>)` instead of `Token`

## How to Test

```bash
cargo test -p turborepo-auth         # 65 tests pass
cargo build -p turborepo             # Build the CLI
./target/debug/turbo login           # Vercel: device flow in browser
./target/debug/turbo login --sso-team <team>  # Vercel SSO
```

Non-Vercel login can be tested by configuring a `loginUrl` that doesn't contain `vercel.com` — it will use the old localhost redirect flow with the same URL patterns.


apps/docs/content/docs/reference/system-environment-variables.mdx
```diff
@@ -325,10 +325,9 @@ System environment variables are always overridden by flag values provided direc
       </td>
       <td>
         Override the default port (9789) used for the SSO login callback server
-        during authentication.
+        during authentication with non-Vercel remote cache servers.
```
anthonyshew:
```suggestion
.
```

apps/docs/content/docs/reference/system-environment-variables.mdx
```diff
@@ -3,7 +3,7 @@ title: System environment variables
 description: Learn about system variables used by Turborepo.
 product: turborepo
 type: reference
-summary: All environment variables that control Turborepo behavior including Remote Cache, logging, and CI settings.
+sapps/docs/content/docs/reference/system-environment-variables.mdxummary: All environment variables that control Turborepo behavior including Remote Cache, logging, and CI settings.
```
anthonyshew:
```suggestion
summary: All environment variables that control Turborepo behavior including Remote Cache, logging, and CI settings.
```

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -76,30 +86,108 @@ pub async fn login<T: Client + TokenClient + CacheClient>(
...
+
+    let verification_url = device_auth
+        .verification_uri_complete
+        .as_deref()
+        .unwrap_or(&device_auth.verification_uri);
+
+    println!(
+        "\n  Visit {}",
+        color_config.apply(BOLD.apply_to(verification_url))
+    );
```
AAorris, line 129: Violation (Medium severity): In login.rs lines ~131-139, the code prefers verification_uri_complete and only prints that URL. The user_code is marked #[allow(dead_code)] in the DeviceAuthorizationResponse struct and is never displayed to the user. This directly violates the RFC's MUST requirement. The user code is an anti-phishing measure — the server shows it and the user is supposed to confirm it matches what their device displays.

It's ok to open the complete uri automatically, but the users should see instructions like
`Go to vercel.com/device and confirm code {user_code}`

This is a MUST in the RFC and that's mainly because we should be sympathetic to nontechnical users, and the device flow scenario where the browser must be opened by typing a URL into another device. Then it's easier to skip the query params and enter the code in the UI.

crates/turborepo-auth/src/device_flow.rs
```diff
@@ -0,0 +1,570 @@
...
+        ),
+        ("token_endpoint", &metadata.token_endpoint),
+        ("revocation_endpoint", &metadata.revocation_endpoint),
+        ("introspection_endpoint", &metadata.introspection_endpoint),
+    ] {
+        let endpoint_url = Url::parse(endpoint).map_err(|_| Error::DiscoveryFailed {
+            message: format!("invalid {name} URL: {endpoint}"),
+        })?;
+        let ep_host = endpoint_url.host_str().unwrap_or("");
+        let host_ok = ep_host == issuer_host || ep_host.ends_with(&format!(".{issuer_host}"));
```
AAorris, line 177: if issuer is vercel.com, the endswith check rejects `evil.com` but not `el.com` right? Might need a more formal check if this is to allow subdomains

crates/turborepo-auth/src/auth/sso.rs
```diff
@@ -86,22 +104,144 @@ pub async fn sso_login<T: Client + TokenClient + CacheClient>(
...
+        message: "missing client_id".to_string(),
+    })?;
+
+    let listener =
+        TcpListener::bind("127.0.0.1:0").map_err(|e| Error::DeviceAuthorizationFailed {
+            message: format!("failed to bind localhost server for SSO redirect: {e}"),
+        })?;
+    let port = listener.local_addr().unwrap().port();
+
+    let state = format!("{:x}{:x}", rand::random::<u64>(), rand::random::<u64>());
```
AAorris, line 186: rand::random() is cryptographically secure on modern platforms, but if the crate's rand were swapped for a non-CSPRNG this could be a problem. Might be worth a note, or an explicitly cryptographically secure pseudorandom number generator

crates/turborepo-auth/src/device_flow.rs
```diff
@@ -0,0 +1,570 @@
...
+/// use the default Vercel issuer. Otherwise fall back to the scheme + host
+/// of the login URL to support self-hosted/enterprise deployments.
+fn issuer_from_login_url(login_url: &str) -> String {
+    if login_url.contains("vercel.com") {
+        return DEFAULT_VERCEL_ISSUER.to_string();
+    }
+    if let Ok(parsed) = Url::parse(login_url) {
+        format!(
+            "{}://{}",
+            parsed.scheme(),
```
AAorris: warn: custom schemes and broadly allowing http

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -111,44 +199,76 @@ pub async fn login<T: Client + TokenClient + CacheClient>(
...
 
-    Ok(Token::new(token.clone()))
+    Ok((Token::new(token_string), None))
+}
+
+/// Accept a single HTTP request on the listener, extract the `token` query
+/// parameter, send a redirect response to the success URL, and return the
+/// token string.
+fn wait_for_login_redirect(listener: TcpListener, success_redirect: &str) -> Result<String, Error> {
+    let (stream, _) = listener.accept().map_err(|_| Error::FailedToGetToken)?;
```
AAorris: listener.accept is called exactly once? (same waiting for sso_redirect)
If a browser preflight or favicon request hits the port first, will we lose the token callback and hang for the 5 minute timeout?
Would the old axum-based server would have handled multiple requests more gracefully?

crates/turborepo-auth/src/auth/login.rs
```diff
@@ -76,30 +86,108 @@ pub async fn login<T: Client + TokenClient + CacheClient>(
...
+        &metadata,
+        &device_auth.device_code,
+        device_auth.interval,
+        expires_at,
+    )
+    .await?;
+
+    spinner.finish_and_clear();
+
+    let secret_token = turborepo_api_client::SecretString::new(token_set.access_token.clone());
```
AAorris, line 148: We request openid scope, which should return an id_token. The TokenSet struct doesn't include it, and we don't validate it. An OIDC-conformant implementation might validate it per OpenID Connect Core §3.1.3.7

It might be fine if we can guarantee get_user does not trust the id_token and does that work for us

crates/turborepo-auth/src/auth/sso.rs
```diff
@@ -1,16 +1,28 @@
...
 
+const DEFAULT_SSO_PROVIDER: &str = "SAML/OIDC Single Sign-On";
 const DEFAULT_HOST_NAME: &str = "127.0.0.1";
 const DEFAULT_PORT: u16 = 9789;
-const DEFAULT_SSO_PROVIDER: &str = "SAML/OIDC Single Sign-On";
+const SSO_REDIRECT_TIMEOUT: Duration = Duration::from_secs(300); // 5 minutes
+
+fn is_vercel(login_url: &str) -> bool {
+    login_url.contains("vercel.com")
+}
```
AAorris: 💭 duplicated in login.rs

crates/turborepo-auth/src/auth/sso.rs
```diff
@@ -116,53 +256,122 @@ pub async fn sso_login<T: Client + TokenClient + CacheClient>(
...
+/// If `expected_state` is provided (Vercel flow), validates the CSRF state
+/// param. For non-Vercel flows, state validation is skipped since the remote
+/// server may not support it.
+fn wait_for_sso_redirect(
+    listener: TcpListener,
+    expected_state: Option<&str>,
+) -> Result<String, Error> {
+    listener
+        .set_nonblocking(false)
+        .map_err(|_| Error::FailedToGetToken)?;
```
AAorris: 💭 Debugging might be easier if there were more specific error types

AAorris (approved):
One line change recommended: Display user_code on its own in the log.

crates/turborepo-auth/src/auth/login.rs
anthonyshew: This is the same as how it used to be handled from what I can tell, though you do seem right. I've never heard of or seen that failure mode, but I added a loop anyway for the sake of robustness.

crates/turborepo-auth/src/device_flow.rs
anthonyshew: I allowed this for the sake of self-hosted deployments not having a break, but this is ultimately 1) a security vulnerability and 2) device flow was never in Turborepo now, so this can't be considered breaking.

Let's make the call that its 2026 and its reasonable to only allow HTTPS.

crates/turborepo-auth/src/auth/sso.rs
anthonyshew: Added a note, Rust's rand::random uses ThreadRng.

crates/turborepo-auth/src/device_flow.rs
anthonyshew: Added some tests to make sure, here and below: https://github.com/vercel/turborepo/pull/12170/changes#diff-b494b4ff2e377c6e14e523fd2686eed9b335c2f31aec927eb568a569d4317aa1R547

crates/turborepo-auth/src/auth/login.rs
anthonyshew: Thank you. 🙏 Fixed up now.

crates/turborepo-auth/src/auth/login.rs
anthonyshew: I've removed id_token, since we don't need it. One less thing to worry about.

crates/turborepo-auth/src/auth/sso.rs
anthonyshew: This came from old code that was lazy, and I was continuing the laziness to try to keep the PR smaller-ish. But, hey, you gave an unreal review, let's do it!

anthonyshew:
Superceded by other PRs.

PR #12208, opened 2026-03-09
perf: Two-phase HTTP client init to avoid macOS Keychain blocking

anthonyshew:
## Summary

Splits HTTP client initialization into two phases to remove ~200ms of macOS Keychain enumeration from the critical path.

- **Phase 1 (instant):** Builds a client with bundled Mozilla CAs (webpki-roots). No system calls, no Keychain access.
- **Phase 2 (background):** Builds a full client with system Keychain CAs (native-roots) on `spawn_blocking`. Once ready, `get_or_init()` returns this client instead.

## Why

`reqwest::Client::builder().build()` with `rustls-tls-native-roots` takes ~200ms on macOS because `rustls-native-certs` enumerates the system Keychain via the Security framework. This was on the critical path during `activate()`, adding 200ms to every `turbo run`.

No HTTP request is made in the first 200ms of a run — remote cache requests happen after index construction (~350ms+) and telemetry flushes at shutdown (~1000ms+). The full client is always ready before it's needed.

## Impact

~100ms wall clock improvement on macOS (the HTTP init now fully overlaps with index construction instead of partially overlapping). On Linux, the native-root loading is already fast (<20ms), so this is a no-op.

## What changes

| File | Change |
|---|---|
| `Cargo.toml` | Added `reqwest/rustls-tls-webpki-roots` feature |
| `lib.rs` | Added `build_http_client_webpki_only()` (webpki-only, instant) |
| `shared_http_client.rs` | Two `OnceLock`s: `fast_client` (Phase 1) and `native_client` (Phase 2). `get_or_init()` prefers native if ready, falls back to fast. |

## Corporate proxy support

Corporate environments with custom root CAs are fully supported. The Phase 2 client loads all system CAs, including custom ones installed in the macOS Keychain. Since Phase 2 completes ~200ms after `activate()` and the earliest HTTP request happens ~350ms+ into the run, the full client is always available when needed.

### DIFF
--- a/crates/turborepo-api-client/Cargo.toml
+++ b/crates/turborepo-api-client/Cargo.toml
@@ -7,7 +7,10 @@ license = "MIT"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]
 native-tls = ["reqwest/native-tls"]
-rustls-tls = ["reqwest/rustls-tls-native-roots"]
+rustls-tls = [
+  "reqwest/rustls-tls-native-roots",
+  "reqwest/rustls-tls-webpki-roots",
+]
 
 [dev-dependencies]
 anyhow = { workspace = true }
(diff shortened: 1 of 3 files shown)


anthonyshew:
The Windows partition flake occurred because the Chocolatey CDN flakes, as it is known for doing. Merging past this.

